PR #447, opened 2020-01-13
format the header of files

yhmtsai:
This PR sorts the ordering of headers, use correct `"header"` or `<header>`, and correct the definition of header.

1. Add `hip/hip_runtime.h` into `hip/base/config.hip.hpp` as its main include to ensure it is in front of `hip/device_functions.h` mentioned in #412 
2. use `IncludeBlocks: Regroup` (need clang-format 6) to sort the header except the main include
3. add config (`dev_tools/scripts/config`) to set the rules how to find the corresponding main include
4. if there are something like comment of #if block before `#include`, the script will give alarm "sorting is maybe not correct"
5. Also alarm if there are multiple matches the main include.

Also mention this in #389 , but remove many if-else and unclear regex (which is almost handled by config)

add target `format_header` and `format_header_all` in CMake
`format_header` only process the changed files of branch
`format_header_all` process all code files.
They ignore build, examples, third_party, ginkgo.hpp

add `// force-top: on` and `// force-top: off` to keep some headers in the top (also before the main header).
I change the warning condition. If there are comments or #if block before codes, it gives warning.
The following is warning
```
Warning benchmark/utils/formats.hpp: sorting maybe is not correct
Warning core/base/extended_float.hpp: sorting maybe is not correct
Warning core/device_hooks/common_kernels.inc.cpp: sorting maybe is not correct
Warning include/ginkgo/core/base/abstract_factory.hpp: sorting maybe is not correct
Warning include/ginkgo/core/base/name_demangling.hpp: sorting maybe is not correct
Warning include/ginkgo/core/base/types.hpp: sorting maybe is not correct
Warning include/ginkgo/core/base/utils.hpp: sorting maybe is not correct
Warning reference/test/stop/combined.cpp: sorting maybe is not correct
Warning reference/test/stop/time.cpp: sorting maybe is not correct
```

PR #455, opened 2020-01-21
Add searching and merging functions

upsj:
To simplify reviewing the ParILUT branch, this PR contains a few cherry-picked algorithms that are used within ParILUT, but are also rather general-purpose. The kernels are based on ideas from Green et al., “GPU merge path: a GPU merging algorithm”

* Searching algorithms that, for a range and a corresponding predicate that is false on the lower part of the range and true on the upper part, identify the first index where the predicate is true. This is the most general form of a binary/k-ary search. There are three implementation variants as mentioned in the paper. (We could also potentially use them in our merge-path SpMV, but that's an idea for another time)

* Merging algorithms that merge two sorted ranges into a single sorted range (via a callback function)

TODO:
- [x] Add references to the implementations
- [x] Remove unnecessary changes

### DIFF
--- a/common/components/intrinsics.hpp.inc
+++ b/common/components/intrinsics.hpp.inc
@@ -34,10 +34,10 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  * @internal
  * Returns the number of set bits in the given mask.
  */
-inline __device__ int popcnt(uint32 mask) { return __popc(mask); }
+__forceinline__ __device__ int popcnt(uint32 mask) { return __popc(mask); }
 
 /** @copydoc popcnt */
-inline __device__ int popcnt(uint64 mask) { return __popcll(mask); }
+__forceinline__ __device__ int popcnt(uint64 mask) { return __popcll(mask); }
 
 
 /**
@@ -45,10 +45,10 @@ inline __device__ int popcnt(uint64 mask) { return __popcll(mask); }
  * Returns the (1-based!) index of the first set bit in the given mask,
  * starting from the least significant bit.
  */
-inline __device__ int ffs(uint32 mask) { return __ffs(mask); }
+__forceinline__ __device__ int ffs(uint32 mask) { return __ffs(mask); }
 
 /** @copydoc ffs */
-inline __device__ int ffs(uint64 mask)
+__forceinline__ __device__ int ffs(uint64 mask)
 {
     // the cast is necessary, as the overloads defined by HIP are ambiguous
     return __ffsll(static_cast<unsigned long long int>(mask));
@@ -60,7 +60,7 @@ inline __device__ int ffs(uint64 mask)
  * Returns the number of zero bits before the first set bit in the given mask,
  * starting from the most significant bit.
  */
-inline __device__ int clz(uint32 mask) { return __clz(mask); }
+__forceinline__ __device__ int clz(uint32 mask) { return __clz(mask); }
 
 /** @copydoc clz */
-inline __device__ int clz(uint64 mask) { return __clzll(mask); }
+__forceinline__ __device__ int clz(uint64 mask) { return __clzll(mask); }
--- a/common/components/intrinsics.hpp.inc
+++ b/common/components/intrinsics.hpp.inc
@@ -34,10 +34,10 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  * @internal
  * Returns the number of set bits in the given mask.
  */
-__device__ int popcnt(uint32 mask) { return __popc(mask); }
+inline __device__ int popcnt(uint32 mask) { return __popc(mask); }
 
 /** @copydoc popcnt */
-__device__ int popcnt(uint64 mask) { return __popcll(mask); }
+inline __device__ int popcnt(uint64 mask) { return __popcll(mask); }
 
 
 /**
@@ -45,10 +45,10 @@ __device__ int popcnt(uint64 mask) { return __popcll(mask); }
  * Returns the (1-based!) index of the first set bit in the given mask,
  * starting from the least significant bit.
  */
-__device__ int ffs(uint32 mask) { return __ffs(mask); }
+inline __device__ int ffs(uint32 mask) { return __ffs(mask); }
 
 /** @copydoc ffs */
-__device__ int ffs(uint64 mask)
+inline __device__ int ffs(uint64 mask)
 {
     // the cast is necessary, as the overloads defined by HIP are ambiguous
     return __ffsll(static_cast<unsigned long long int>(mask));
@@ -60,7 +60,7 @@ __device__ int ffs(uint64 mask)
  * Returns the number of zero bits before the first set bit in the given mask,
  * starting from the most significant bit.
  */
-__device__ int clz(uint32 mask) { return __clz(mask); }
+inline __device__ int clz(uint32 mask) { return __clz(mask); }
 
 /** @copydoc clz */
-__device__ int clz(uint64 mask) { return __clzll(mask); }
+inline __device__ int clz(uint64 mask) { return __clzll(mask); }
--- a/common/components/merging.hpp.inc
+++ b/common/components/merging.hpp.inc
@@ -133,11 +133,13 @@ __device__ detail::merge_result<ValueType> group_merge_step(ValueType a,
  * @param b_size  the size of the second range
  * @param group   the group that executes the merge
  * @param merge_fn  the callback that is being called for each merged element.
- *                  It takes five parameters:
- *                  `IndexType a_idx, ValueType a_val,
- *                   IndexType b_idx, ValueType b_val, IndexType c_index`.
- *                   `*_val` and `*_idx` are the values resp. the indices of the
- *                   values from a/b being compared at output index `c_index`.
+ *                  It takes six parameters:
+ *                  `IndexType a_idx, ValueType a_val, IndexType b_idx,
+ *                   ValueType b_val, IndexType c_index, bool valid`.
+ *                  `*_val` and `*_idx` are the values resp. the indices of the
+ *                  values from a/b being compared at output index `c_index`.
+ *                  `valid` specifies if the current thread has to merge an
+ *                  element (this is necessary for shfl and ballot operations).
  */
 template <int group_size, typename IndexType, typename ValueType,
           typename Group, typename Callback>
@@ -154,11 +156,9 @@ __device__ void group_merge(const ValueType *a, IndexType a_size,
     auto b_cur = detail::checked_load(b, b_begin + lane, a_size);
     while (c_begin < c_size) {
         auto merge_result = group_merge_step<group_size>(a_cur, b_cur, group);
-        if (c_begin + lane < c_size) {
-            merge_fn(merge_result.a_idx + a_begin, merge_result.a_val,
-                     merge_result.b_idx + b_begin, merge_result.b_val,
-                     c_begin + lane);
-        }
+        merge_fn(merge_result.a_idx + a_begin, merge_result.a_val,
+                 merge_result.b_idx + b_begin, merge_result.b_val,
+                 c_begin + lane, c_begin + lane < c_size);
         auto a_advance = merge_result.a_advance;
         auto b_advance = merge_result.b_advance;
         a_begin += a_advance;
--- a/common/components/merging.hpp.inc
+++ b/common/components/merging.hpp.inc
@@ -35,7 +35,7 @@ namespace detail {
 
 /**
  * @internal
- * The result from the @ref parallel_merge_step function.
+ * The result from the @ref group_merge_step function.
  */
 template <typename ValueType>
 struct merge_result {
@@ -71,8 +71,8 @@ checked_load(const ValueType *p, IndexType i, IndexType size,
  * Warp-parallel merge algorithm that merges the first `warp_size` elements from
  * two ranges, where each warp stores a single element from each range.
  * It assumes that the elements are sorted in ascending order, i.e. for i < j,
- * the value of `a` at thread i is smaller or equal the value at thread j, and
- * the same holds for `b`.
+ * the value of `a` at thread i is smaller or equal to the value at thread j,
+ * and the same holds for `b`.
  *
  * This implementation is based on ideas from  Green et al.,
  * "GPU merge path: a GPU merging algorithm", but uses random-access warp
@@ -115,7 +115,7 @@ __device__ detail::merge_result<ValueType> group_merge_step(ValueType a,
     auto a_val = group.shfl(a, a_idx);
     auto b_val = group.shfl(b, b_idx);
     auto cmp = a_val < b_val;
-    auto a_advance = int(popcnt(group.ballot(cmp)));
+    auto a_advance = popcnt(group.ballot(cmp));
     auto b_advance = int(group.size()) - a_advance;
 
     return {a_val, b_val, a_idx, b_idx, a_advance, b_advance};
@@ -125,7 +125,7 @@ __device__ detail::merge_result<ValueType> group_merge_step(ValueType a,
 /**
  * @internal
  * Warp-parallel merge algorithm that merges two sorted ranges of arbitrary
- * size. `merge_fun` will be called for each merged element.
+ * size. `merge_fn` will be called for each merged element.
  *
  * @param a       the first range
  * @param a_size  the size of the first range
@@ -195,7 +195,7 @@ __device__ void group_merge(const ValueType *a, IndexType a_size,
 /**
  * @internal
  * Sequential merge algorithm that merges two sorted ranges of arbitrary
- * size. `merge_fun` will be called for each merged element.
+ * size. `merge_fn` will be called for each merged element.
  *
  * @param a  the first range
  * @param a_size the size of the first range
--- a/common/components/merging.hpp.inc
+++ b/common/components/merging.hpp.inc
@@ -1,0 +1,373 @@
+/*******************************<GINKGO LICENSE>******************************
+Copyright (c) 2017-2020, the Ginkgo authors
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+
+1. Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+
+2. Redistributions in binary form must reproduce the above copyright
+notice, this list of conditions and the following disclaimer in the
+documentation and/or other materials provided with the distribution.
+
+3. Neither the name of the copyright holder nor the names of its
+contributors may be used to endorse or promote products derived from
+this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS
+IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED
+TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A
+PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+namespace detail {
+
+
+/**
+ * @internal
+ * The result from the @ref parallel_merge_step function.
+ */
+template <typename ValueType>
+struct merge_result {
+    /** The element of a being merged in the current thread. */
+    ValueType a_val;
+    /** The element of b being merged in the current thread. */
+    ValueType b_val;
+    /** The index from a that is being merged in the current thread. */
+    int a_idx;
+    /** The index from b that is being merged in the current thread. */
+    int b_idx;
+    /** The number of elements from a that have been merged in total. */
+    int a_advance;
+    /** The number of elements from b that have been merged in total. */
+    int b_advance;
+};
+
+
+template <typename ValueType, typename IndexType>
+__device__ ValueType
+checked_load(const ValueType *p, IndexType i, IndexType size,
+             ValueType sentinel = device_numeric_limits<ValueType>::max)
+{
+    return i < size ? p[i] : sentinel;
+}
+
+
+}  // namespace detail
+
+
+/**
+ * @internal
+ * Warp-parallel merge algorithm that merges the first `warp_size` elements from
+ * two ranges, where each warp stores a single element from each range.
+ * It assumes that the elements are sorted in ascending order, i.e. for i < j,
+ * the value of `a` at thread i is smaller or equal the value at thread j, and
+ * the same holds for `b`.
+ *
+ * @param a      the element from the first range
+ * @param b      the element from the second range
+ * @param size   the number of elements in the output range
+ * @param group  the cooperative group that executes the merge
+ * @return  a structure containing the merge result distributed over the group.
+ */
+template <typename ValueType, typename Group>
+__device__ detail::merge_result<ValueType> group_merge_step(ValueType a,
+                                                            ValueType b,
+                                                            int size,
+                                                            Group group)
+{
+    // round up to the next power of two
+    auto size_pow2 = 1 << (32 - clz(uint32(size - 1)));
+    // thread i takes care of ith element of the merged sequence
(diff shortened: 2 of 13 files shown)


common/components/merging.hpp.inc
```diff
@@ -0,0 +1,237 @@
...
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+namespace detail {
+
+
+/**
+ * @internal
+ * The result from the @ref parallel_merge_step function.
```
upsj:
```suggestion
 * The result from the @ref group_merge_step function.
```

cuda/test/components/searching.cu
```diff
@@ -0,0 +1,244 @@
...
+
+TEST_F(Searching, EmptySyncBinary)
+{
+    run_test(test_empty_sync_binary_search, 0, config::warp_size);
+}
+
+
+__global__ void test_warp_ary_search(bool *success, int offset, int size)
+{
+    // test binary search on [0, length)
```
tcojean: length is size I guess, also isn't it from offset rather than 0?

cuda/test/components/searching.cu
```diff
@@ -0,0 +1,244 @@
...
+{
+    for (auto size : sizes) {
+        run_test(test_warp_ary_search, 134, size, size + 1);
+    }
+}
+
+
+__global__ void test_warp_wide_search(bool *success, int offset, int size)
+{
+    // test binary search on [0, length)
```
tcojean: same

cuda/test/components/searching.cu
tcojean: Maybe these comments are better put in the respective TEST_F, since there is also a discrepancy  with the [0, warp_size) one.

common/components/searching.hpp.inc
```diff
@@ -0,0 +1,230 @@
...
+ * @param p  the predicate to be evaluated on the range - it should not have
+ *           side-effects and map from `IndexType` to `bool`
+ * @returns  the index of `middle`, i.e., the partition point
+ */
+template <typename IndexType, typename Group, typename Predicate>
+__device__ IndexType group_ary_search(IndexType offset, IndexType length,
+                                      Group group, Predicate p)
+{
+    IndexType end = offset + length;
+    // invariant: [offset, offset + length] contains middle
```
tcojean: I'm unsure of the phrasing here. Technically, it's true this is the pivot space (where a pivot could rightfully exist according to the definition), but "offset+length" is of course out of boundaries of the tested space.  Maybe it would be useful to specifically note somewhere that if the predicate isn't satisfied anywhere, then the pivot has to be `offset + length`.

upsj: I added a clarifying comment to the function documentations.

cuda/components/merging.cuh
```diff
@@ -0,0 +1,54 @@
...
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+#ifndef GKO_CUDA_COMPONENTS_MERGING_CUH_
+#define GKO_CUDA_COMPONENTS_MERGING_CUH_
+
+
+#include "cuda/components/intrinsics.cuh"
+#include "cuda/components/searching.cuh"
```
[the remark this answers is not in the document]
upsj: Good point, thanks!

hip/components/searching.hip.hpp
```diff
@@ -0,0 +1,55 @@
...
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+#ifndef GKO_HIP_COMPONENTS_SEARCHING_CUH_
+#define GKO_HIP_COMPONENTS_SEARCHING_CUH_
+
+
+#include "hip/base/config.hip.hpp"
+#include "hip/components/cooperative_groups.hip.hpp"
```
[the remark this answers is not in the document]
upsj: Good point, thanks!

cuda/test/components/merging.cu
```diff
@@ -0,0 +1,287 @@
...
+
+    void init_data()
+    {
+        std::uniform_int_distribution<gko::int32> dist(0, max_size);
+        std::fill_n(data1.get_data(), max_size, 0);
+        std::fill_n(data2.get_data(), max_size, 0);
+        for (auto i = 0; i < max_size; ++i) {
+            // here we also want to test some corner cases
+            // first two runs: zero data1
+            if (rng_run > 0) data1.get_data()[i] = dist(rng);
```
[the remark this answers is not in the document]
upsj: That's true, I fixed it, thanks!

cuda/test/components/merging.cu
```diff
@@ -0,0 +1,287 @@
...
+    auto warp = tiled_partition<config::warp_size>(this_thread_block());
+    auto i = warp.thread_rank();
+    auto result = group_merge_step<config::warp_size>(a[i], b[i], warp);
+    c[i] = min(result.a_val, result.b_val);
+}
+
+TEST_F(Merging, MergeStep)
+{
+    for (auto i = 0; i < rng_runs; ++i) {
+        init_data();
```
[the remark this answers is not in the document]
upsj: Good point, I moved it to a parameter.

cuda/test/components/searching.cu
```diff
@@ -0,0 +1,244 @@
...
+class Searching : public ::testing::Test {
+protected:
+    Searching()
+        : ref(gko::ReferenceExecutor::create()),
+          cuda(gko::CudaExecutor::create(0, ref)),
+          result(ref, 1),
+          dresult(cuda),
+          sizes(14203)
+    {
+        std::iota(sizes.begin(), sizes.end(), 0);
```
tcojean: iota needs `algorithm` include

upsj: while I think some stdlib implementations include it in `algorithm`, it is supposed to be located in `numeric`

common/components/merging.hpp.inc
```diff
@@ -0,0 +1,237 @@
...
+    /** The element of b being merged in the current thread. */
+    ValueType b_val;
+    /** The index from a that is being merged in the current thread. */
+    int a_idx;
+    /** The index from b that is being merged in the current thread. */
+    int b_idx;
+    /** The number of elements from a that have been merged in total. */
+    int a_advance;
+    /** The number of elements from b that have been merged in total. */
+    int b_advance;
```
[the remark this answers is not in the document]
upsj: These indices here and in `group_merge_step` are only warp-local (<= config::warp_size), so I wanted to keep the indexing separate. What do you think?

cuda/test/components/searching.cu
tcojean: cppreference says algorithm
https://en.cppreference.com/w/cpp/algorithm/iota

upsj: > Defined in header <numeric>

tcojean: my bad, you're right, it's algorithm category but numeric header

common/components/merging.hpp.inc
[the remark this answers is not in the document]
tcojean: Yes, I guess that makes sense. They are also used in a simple way so I guess there is no need to upgrade them to index type when using them in `group_merge`.

hip/test/components/merging.hip.cpp
```diff
@@ -0,0 +1,294 @@
...
+          drefidxs3(hip, 2 * max_size),
+          doutdata(hip, 2 * max_size)
+    {}
+
+    void init_data(int rng_run)
+    {
+        std::uniform_int_distribution<gko::int32> dist(0, max_size);
+        std::fill_n(data1.get_data(), max_size, 0);
+        std::fill_n(data2.get_data(), max_size, 0);
+        for (auto i = 0; i < max_size; ++i) {
```
thoasm: Please no `auto` in combination with an integer literal.
```suggestion
        for (int i = 0; i < max_size; ++i) {
```

cuda/test/components/merging.cu
```diff
@@ -0,0 +1,285 @@
...
+          drefidxs3(cuda, 2 * max_size),
+          doutdata(cuda, 2 * max_size)
+    {}
+
+    void init_data(int rng_run)
+    {
+        std::uniform_int_distribution<gko::int32> dist(0, max_size);
+        std::fill_n(data1.get_data(), max_size, 0);
+        std::fill_n(data2.get_data(), max_size, 0);
+        for (auto i = 0; i < max_size; ++i) {
```
thoasm:
```suggestion
        for (int i = 0; i < max_size; ++i) {
```

cuda/test/components/merging.cu
```diff
@@ -0,0 +1,285 @@
...
+{
+    auto warp = tiled_partition<config::warp_size>(this_thread_block());
+    auto i = warp.thread_rank();
+    auto result = group_merge_step<config::warp_size>(a[i], b[i], warp);
+    c[i] = min(result.a_val, result.b_val);
+}
+
+TEST_F(Merging, MergeStep)
+{
+    for (auto i = 0; i < rng_runs; ++i) {
```
thoasm:
```suggestion
    for (int i = 0; i < rng_runs; ++i) {
```

cuda/test/components/merging.cu
```diff
@@ -0,0 +1,285 @@
...
+    group_merge<config::warp_size>(
+        a, size, b, size, warp,
+        [&](int a_idx, gko::int32 a_val, int b_idx, gko::int32 b_val, int i) {
+            c[i] = min(a_val, b_val);
+        });
+}
+
+TEST_F(Merging, FullMerge)
+{
+    for (auto i = 0; i < rng_runs; ++i) {
```
thoasm:
```suggestion
    for (int i = 0; i < rng_runs; ++i) {
```

cuda/test/components/merging.cu
```diff
@@ -0,0 +1,285 @@
...
+    sequential_merge(
+        a, size, b, size,
+        [&](int a_idx, gko::int32 a_val, int b_idx, gko::int32 b_val, int i) {
+            c[i] = min(a_val, b_val);
+        });
+}
+
+TEST_F(Merging, SequentialFullMerge)
+{
+    for (auto i = 0; i < rng_runs; ++i) {
```
thoasm:
```suggestion
    for (int i = 0; i < rng_runs; ++i) {
```

cuda/test/components/merging.cu
```diff
@@ -0,0 +1,285 @@
...
+            aidxs[i] = a_idx;
+            bidxs[i] = b_idx;
+            cidxs[i] = i;
+            c[i] = min(a_val, b_val);
+        });
+}
+
+TEST_F(Merging, FullMergeIdxs)
+{
+    for (auto i = 0; i < rng_runs; ++i) {
```
thoasm:
```suggestion
    for (int i = 0; i < rng_runs; ++i) {
```

hip/test/components/merging.hip.cpp
```diff
@@ -0,0 +1,294 @@
...
+{
+    auto warp = tiled_partition<config::warp_size>(this_thread_block());
+    auto i = warp.thread_rank();
+    auto result = group_merge_step<config::warp_size>(a[i], b[i], warp);
+    c[i] = min(result.a_val, result.b_val);
+}
+
+TEST_F(Merging, MergeStep)
+{
+    for (auto i = 0; i < rng_runs; ++i) {
```
thoasm:
```suggestion
    for (int i = 0; i < rng_runs; ++i) {
```

hip/test/components/merging.hip.cpp
```diff
@@ -0,0 +1,294 @@
...
+    group_merge<config::warp_size>(
+        a, size, b, size, warp,
+        [&](int a_idx, gko::int32 a_val, int b_idx, gko::int32 b_val, int i) {
+            c[i] = min(a_val, b_val);
+        });
+}
+
+TEST_F(Merging, FullMerge)
+{
+    for (auto i = 0; i < rng_runs; ++i) {
```
thoasm:
```suggestion
    for (int i = 0; i < rng_runs; ++i) {
```

hip/test/components/merging.hip.cpp
```diff
@@ -0,0 +1,294 @@
...
+    sequential_merge(
+        a, size, b, size,
+        [&](int a_idx, gko::int32 a_val, int b_idx, gko::int32 b_val, int i) {
+            c[i] = min(a_val, b_val);
+        });
+}
+
+TEST_F(Merging, SequentialFullMerge)
+{
+    for (auto i = 0; i < rng_runs; ++i) {
```
thoasm:
```suggestion
    for (int i = 0; i < rng_runs; ++i) {
```

hip/test/components/merging.hip.cpp
```diff
@@ -0,0 +1,294 @@
...
+            aidxs[i] = a_idx;
+            bidxs[i] = b_idx;
+            cidxs[i] = i;
+            c[i] = min(a_val, b_val);
+        });
+}
+
+TEST_F(Merging, FullMergeIdxs)
+{
+    for (auto i = 0; i < rng_runs; ++i) {
```
thoasm:
```suggestion
    for (int i = 0; i < rng_runs; ++i) {
```

hip/test/components/searching.hip.cpp
```diff
@@ -0,0 +1,249 @@
...
+        auto success = *result.get_const_data();
+
+        ASSERT_TRUE(success);
+    }
+
+    std::shared_ptr<gko::ReferenceExecutor> ref;
+    std::shared_ptr<gko::HipExecutor> hip;
+    gko::Array<bool> result;
+    gko::Array<bool> dresult;
+    std::vector<int> sizes;
```
thoasm: You need to include `vector` for this one.

hip/test/components/merging.hip.cpp
```diff
@@ -0,0 +1,294 @@
...
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+// TODO remove when the HIP includes are fixed
+#include <hip/hip_runtime.h>
```
thoasm: Is this required to be included first, or could it be together with other third party includes?

common/components/searching.hpp.inc
```diff
@@ -0,0 +1,235 @@
...
+ * @internal
+ * Generic search that finds the first index where a predicate is true.
+ * It assumes that the predicate partitions the range [offset, offset + length)
+ * into two subranges [offset, middle), [middle, offset + length) such that
+ * the predicate is `false` for all elements in the first range and `true` for
+ * all elements in the second range. `middle` is called the partition point.
+ * If the predicate is `false` everywhere, `middle` equals `offset + length`.
+ *
+ * It executes `log2(length) / log2(group.size())` calls to `p` that effectively
+ * follow a random-access pattern.
```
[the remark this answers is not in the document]
upsj: I would rather leave that as it is (calls [...] that [...] follow)

hip/test/components/merging.hip.cpp
upsj: Yes, some HIP files are not self-contained, so they need the include from hip_runtime.h

common/components/merging.hpp.inc
```diff
@@ -0,0 +1,237 @@
...
+        return a_remote >= b_remote || x >= i;
+    });
+
+    auto a_idx = minx;
+    auto b_idx = max(i - minx, 0);
+    auto a_val = group.shfl(a, a_idx);
+    auto b_val = group.shfl(b, b_idx);
+    auto cmp = a_val < b_val;
+    auto a_advance = int(popcnt(group.ballot(cmp)));
+    auto b_advance = int(group.size()) - a_advance;
```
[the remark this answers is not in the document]
upsj: all involved values are <= 64, so I would not expect any issues here. Also, `group.size()` returns an unsigned value, so this conversion would fail. I could remove the cast from `popcnt` since it returns an `int` anyways, but would like to leave the rest as it is.

cuda/test/components/merging.cu
```diff
@@ -0,0 +1,285 @@
...
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+#include "cuda/components/merging.cuh"
+
+
+#include <algorithm>
+#include <memory>
+#include <random>
```
upsj:
```suggestion
#include <random>
#include <vector>
```

cuda/test/components/searching.cu
```diff
@@ -0,0 +1,244 @@
...
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+#include "cuda/components/searching.cuh"
+
+
+#include <memory>
+#include <numeric>
```
upsj:
```suggestion
#include <numeric>
#include <vector>
```

hip/test/components/merging.hip.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+
+#include "hip/components/merging.hip.hpp"
+
+
+#include <algorithm>
+#include <memory>
+#include <random>
```
upsj:
```suggestion
#include <random>
#include <vector>
```

hip/test/components/searching.hip.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+// TODO remove when the HIP includes are fixed
+#include <hip/hip_runtime.h>
+
+
+#include "hip/components/searching.hip.hpp"
+
+
+#include <memory>
+#include <numeric>
```
upsj:
```suggestion
#include <numeric>
#include <vector>
```

hip/test/components/searching.hip.cpp
upsj: Good point, thanks!

common/components/merging.hpp.inc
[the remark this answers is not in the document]
thoasm: If you really want to keep the current status, I will not stop you from doing so.
I simply dislike any kind of C-style casts, especially if it is easy to replace it with a C++ cast.

common/components/merging.hpp.inc
```diff
@@ -0,0 +1,237 @@
...
+        auto b_remote = group.shfl(b, max(i - x - 1, 0));
+        return a_remote >= b_remote || x >= i;
+    });
+
+    auto a_idx = minx;
+    auto b_idx = max(i - minx, 0);
+    auto a_val = group.shfl(a, a_idx);
+    auto b_val = group.shfl(b, b_idx);
+    auto cmp = a_val < b_val;
+    auto a_advance = int(popcnt(group.ballot(cmp)));
```
upsj:
```suggestion
    auto a_advance = popcnt(group.ballot(cmp));
```

thoasm: That is fine with me.
Just out of curiosity: Why are you proposing changes to your own PR? To keep taps on it?

upsj: Yes, I wanted to add all proposed small changes to a single commit instead of having to manually apply them on my system

common/components/merging.hpp.inc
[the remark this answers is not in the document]
upsj: For casting integer types, `static_cast<...>` looks unnecessarily verbose to me. In case you are not aware, for non-cv qualified integer types, a functional-style cast effectively is a `static_cast`, since [all other options](https://en.cppreference.com/w/cpp/language/explicit_cast) (const_cast, reinterpret_cast) do not apply there

yhmtsai:
non self-contained hip file is `hip/device_functions.h` which needs to include `hip/hip_runtime.h` first.
We only use `hip/device_functions.h` in the `config.hip.hpp` to get warpsize.
For temporary fix, put the `hip/hip_runtime.h` into `config.hip.hpp` before `hip/device_functions.h`.
It should fix the problem, and I also did it in PR #447 sorting and config changes (https://github.com/ginkgo-project/ginkgo/pull/447/files/55a80da0a4247401db21d5ae4932930a3d01b176)
Could you also try it here?

common/components/merging.hpp.inc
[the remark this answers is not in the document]
thoasm: Yes, I know that. However, I still prefer the C++ casts for 2 reasons:
1. For consistency, so it is unlikely I ever make the mistake to do a functional cast with a template parameter or an unknown type
2. I personally like the verbosity of these casts, since it is extremely easy to spot them (especially during debugging). Everyone who is reading `static_cast` knows instantly that a cast is happening, while it is easy to over-read `int()`.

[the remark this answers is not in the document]
upsj: I see, I need to think about the implications a bit. But would you consider it okay to leave this cast in?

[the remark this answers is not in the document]
thoasm: Yes, you can leave it in, it was just a sugesetion and I don't consider it absolutely mandatory to use C++ style casts (**in this context**).
As you said, there is pretty much next to no way that this cast can cause any issues, so you can leave it in.

common/components/intrinsics.hpp.inc
```diff
@@ -34,21 +34,21 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+inline __device__ int popcnt(uint64 mask) { return __popcll(mask); }
 
 
 /**
  * @internal
  * Returns the (1-based!) index of the first set bit in the given mask,
  * starting from the least significant bit.
  */
-__device__ int ffs(uint32 mask) { return __ffs(mask); }
+inline __device__ int ffs(uint32 mask) { return __ffs(mask); }
```
thoasm: Usually, we use `__forceinline__` for inlining in GPU kernels (at least we always did it in CUDA and we currently use it in pretty much all `common` kernels like in `atomic` or `cooperative_groups`).

common/components/intrinsics.hpp.inc
```diff
@@ -34,21 +34,21 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  * @internal
  * Returns the number of set bits in the given mask.
  */
-__device__ int popcnt(uint32 mask) { return __popc(mask); }
+inline __device__ int popcnt(uint32 mask) { return __popc(mask); }
```
upsj:
```suggestion
__forceinline__ __device__ int popcnt(uint32 mask) { return __popc(mask); }
```

common/components/intrinsics.hpp.inc
```diff
@@ -34,21 +34,21 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
 
 /** @copydoc popcnt */
-__device__ int popcnt(uint64 mask) { return __popcll(mask); }
+inline __device__ int popcnt(uint64 mask) { return __popcll(mask); }
```
upsj:
```suggestion
__forceinline__ __device__ int popcnt(uint64 mask) { return __popcll(mask); }
```

common/components/intrinsics.hpp.inc
```diff
@@ -34,21 +34,21 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
  * @internal
  * Returns the (1-based!) index of the first set bit in the given mask,
  * starting from the least significant bit.
  */
-__device__ int ffs(uint32 mask) { return __ffs(mask); }
+inline __device__ int ffs(uint32 mask) { return __ffs(mask); }
 
 /** @copydoc ffs */
-__device__ int ffs(uint64 mask)
+inline __device__ int ffs(uint64 mask)
```
upsj:
```suggestion
__forceinline__ __device__ int ffs(uint64 mask)
```

common/components/intrinsics.hpp.inc
```diff
@@ -60,7 +60,7 @@ __device__ int ffs(uint64 mask)
  * Returns the number of zero bits before the first set bit in the given mask,
  * starting from the most significant bit.
  */
-__device__ int clz(uint32 mask) { return __clz(mask); }
+inline __device__ int clz(uint32 mask) { return __clz(mask); }
```
upsj:
```suggestion
__forceinline__ __device__ int clz(uint32 mask) { return __clz(mask); }
```

common/components/intrinsics.hpp.inc
```diff
@@ -60,7 +60,7 @@ __device__ int ffs(uint64 mask)
...
 
 /** @copydoc clz */
-__device__ int clz(uint64 mask) { return __clzll(mask); }
+inline __device__ int clz(uint64 mask) { return __clzll(mask); }
```
upsj:
```suggestion
__forceinline__ __device__ int clz(uint64 mask) { return __clzll(mask); }
```

PR #419, opened 2019-12-09
Ensure non-zero diagonal when generating triangular matrices

upsj:
The generator for random triangular matrices currently does not ensure that the triangular matrix has non-zero diagonal entries, which means the resulting matrices can be singular.
In the same fashion, the generation of the ParILU factors sparsity pattern relies on the input matrix containing all diagonal entries.
This PR fixes this behavior by explicitly adding the diagonal entries if not present.

PR #456, opened 2020-01-23
Fix parilu by adding explicit zeros to diagonal

### LINKED ISSUE #445: Incorrect ParILU(T) sweep for zero diagonal entries
Currently, the ParILU(T) asynchronous iteration is only executed on non-zero locations of the system matrix, the fix in https://github.com/ginkgo-project/ginkgo/pull/419 was incomplete, as it only fixes the initialization of L and U, but not their updates.

Easiest solution: Add a kernel that introduces explicit zero values for the system matrix diagonal if they are not present already. This would also allow us to remove the special treatment for the diagonal introduced in #419

thoasm:
This PR adds a kernel to `ParIlu` (`add_diagonal_element`) which adds explicit zeros on the main diagonal if there was no element.
This ensures that the diagonal is always properly updated during the computation of both `L` and `U`.

Closes #445.

__TODO:__
- [x] Fix reference test case failure
- [x] Merge #455 and rebase on top of it


thoasm:
Currently, I did not touch the other kernels of `ParIlu`, meaning they still contain code which handles matrices without diagonal entries separately. Should I remove that part now, since diagonal elements are now always added before the other kernels are called?

upsj:
I would propose so, yes. I guess you can just revert the changes from https://github.com/ginkgo-project/ginkgo/pull/419/commits/f7836342c3e22ee73e03deb0351dd896f9af24e4

thoasm:
Thanks, that made it a lot easier!

cuda/factorization/par_ilu_kernels.cu
```diff
@@ -60,9 +65,92 @@ constexpr int default_block_size{512};
...
+        return;
+    }
+
+    prefix_sum(exec, cuda_row_ptrs_add, row_ptrs_size);
+    exec->synchronize();
+
+    IndexType total_additions{};
+    exec->get_master()->copy_from(
+        exec.get(), 1, cuda_row_ptrs_add + row_ptrs_size - 1, &total_additions);
+    size_type new_num_elems = static_cast<size_type>(total_additions) +
```
upsj: I know we use size_type for Array dimensions, but can this ever safely overflow IndexType if subsequent kernels use the same integer type?

cuda/factorization/par_ilu_kernels.cu
```diff
@@ -60,9 +65,92 @@ constexpr int default_block_size{512};
...
+                              mtx->get_num_stored_elements();
+
+
+    Array<ValueType> new_values{exec, new_num_elems};
+    Array<IndexType> new_col_idxs{exec, new_num_elems};
+    auto cuda_new_values = as_cuda_type(new_values.get_data());
+    auto cuda_new_col_idxs = as_cuda_type(new_col_idxs.get_data());
```
upsj: I guess you don't need the as_cuda_type here, since it is only necessary for complex

cuda/factorization/par_ilu_kernels.cu
```diff
@@ -60,9 +65,92 @@ constexpr int default_block_size{512};
...
+}
+
+GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
+    GKO_DECLARE_PAR_ILU_ADD_DIAGONAL_ELEMENTS_KERNEL);
+
+
 template <typename ValueType, typename IndexType>
 void initialize_row_ptrs_l_u(
-    std::shared_ptr<const CudaExecutor> exec,
+    std::shared_ptr<const DefaultExecutor> exec,
```
upsj: For consistency: Do we want to use this everywhere or revert these changes?

cuda/factorization/par_ilu_kernels.cu
```diff
@@ -88,7 +176,7 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
 
 
 template <typename ValueType, typename IndexType>
-void initialize_l_u(std::shared_ptr<const CudaExecutor> exec,
+void initialize_l_u(std::shared_ptr<const DefaultExecutor> exec,
```
upsj: Same here

cuda/factorization/par_ilu_kernels.cu
```diff
@@ -114,7 +202,7 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
 
 
 template <typename ValueType, typename IndexType>
-void compute_l_u_factors(std::shared_ptr<const CudaExecutor> exec,
+void compute_l_u_factors(std::shared_ptr<const DefaultExecutor> exec,
```
upsj: Same here

hip/factorization/par_ilu_kernels.hip.cpp
```diff
@@ -63,6 +67,92 @@ constexpr int default_block_size{512};
...
 
 
+template <typename ValueType, typename IndexType>
+void add_diagonal_elements(std::shared_ptr<const DefaultExecutor> exec,
+                           matrix::Csr<ValueType, IndexType> *mtx,
+                           bool is_sorted)
+{
+    // TODO: Runtime can be optimized by choosing a appropriate size for the
+    //       subwarp dependent on the matrix properties
+    constexpr int subwarp_size = config::warp_size / 2;
```
upsj: config::warp_size?

hip/factorization/par_ilu_kernels.hip.cpp
```diff
@@ -63,6 +67,92 @@ constexpr int default_block_size{512};
...
+
+    auto hip_old_values = as_hip_type(mtx->get_const_values());
+    auto hip_old_col_idxs = as_hip_type(mtx->get_const_col_idxs());
+    auto hip_old_row_ptrs = as_hip_type(mtx->get_row_ptrs());
+    auto hip_row_ptrs_add = as_hip_type(row_ptrs_addition.get_data());
+
+    const dim3 block_dim{default_block_size, 1, 1};
+    const dim3 grid_dim{
+        static_cast<uint32>(ceildiv(
+            num_rows, static_cast<size_type>(block_dim.x / subwarp_size))),
```
upsj: two casts?

hip/factorization/par_ilu_kernels.hip.cpp
```diff
@@ -63,6 +67,92 @@ constexpr int default_block_size{512};
...
+    auto hip_new_col_idxs = as_hip_type(new_col_idxs.get_data());
+
+    hipLaunchKernelGGL(
+        HIP_KERNEL_NAME(kernel::add_missing_diagonal_elements<subwarp_size>),
+        grid_dim, block_dim, 0, 0, num_rows, hip_old_values, hip_old_col_idxs,
+        hip_old_row_ptrs, hip_new_values, hip_new_col_idxs, hip_row_ptrs_add);
+
+    const dim3 grid_dim_row_ptrs_update{
+        static_cast<uint32>(
+            ceildiv(num_rows, static_cast<size_type>(block_dim.x))),
```
upsj: two casts?

omp/factorization/par_ilu_kernels.cpp
```diff
@@ -52,9 +58,164 @@ namespace omp {
...
+}
+
+GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
+    GKO_DECLARE_PAR_ILU_ADD_DIAGONAL_ELEMENTS_KERNEL);
+
+
 template <typename ValueType, typename IndexType>
 void initialize_row_ptrs_l_u(
-    std::shared_ptr<const OmpExecutor> exec,
+    std::shared_ptr<const DefaultExecutor> exec,
```
upsj: revert?

omp/factorization/par_ilu_kernels.cpp
```diff
@@ -92,7 +251,7 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
 
 
 template <typename ValueType, typename IndexType>
-void initialize_l_u(std::shared_ptr<const OmpExecutor> exec,
+void initialize_l_u(std::shared_ptr<const DefaultExecutor> exec,
```
upsj: revert?

omp/factorization/par_ilu_kernels.cpp
```diff
@@ -52,9 +58,164 @@ namespace omp {
...
+    {
+        return std::find(first, last, value) != last;
+    }
+};
+
+
+template <>
+struct find_helper<true> {
+    template <typename ForwardIt, typename IndexType>
+    static bool find(ForwardIt first, ForwardIt last, const IndexType &value)
```
upsj:
```suggestion
    static bool find(ForwardIt first, ForwardIt last, IndexType value)
```

omp/factorization/par_ilu_kernels.cpp
```diff
@@ -52,9 +58,164 @@ namespace omp {
...
 
 
+namespace kernel {
+namespace detail {
+
+
+template <bool IsSorted = false>
+struct find_helper {
+    template <typename ForwardIt, typename IndexType>
+    static bool find(ForwardIt first, ForwardIt last, const IndexType &value)
```
upsj:
```suggestion
    static bool find(ForwardIt first, ForwardIt last, IndexType value)
```

reference/factorization/par_ilu_kernels.cpp
```diff
@@ -52,9 +58,103 @@ namespace reference {
...
+}
+
+GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
+    GKO_DECLARE_PAR_ILU_ADD_DIAGONAL_ELEMENTS_KERNEL);
+
+
 template <typename ValueType, typename IndexType>
 void initialize_row_ptrs_l_u(
-    std::shared_ptr<const ReferenceExecutor> exec,
+    std::shared_ptr<const DefaultExecutor> exec,
```
upsj: revert?

reference/factorization/par_ilu_kernels.cpp
```diff
@@ -89,7 +185,7 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
 
 
 template <typename ValueType, typename IndexType>
-void initialize_l_u(std::shared_ptr<const ReferenceExecutor> exec,
+void initialize_l_u(std::shared_ptr<const DefaultExecutor> exec,
```
upsj: revert?

reference/factorization/par_ilu_kernels.cpp
```diff
@@ -148,7 +235,7 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
 
 
 template <typename ValueType, typename IndexType>
-void compute_l_u_factors(std::shared_ptr<const ReferenceExecutor> exec,
+void compute_l_u_factors(std::shared_ptr<const DefaultExecutor> exec,
```
upsj: revert?

common/factorization/par_ilu_kernels.hpp.inc
```diff
@@ -33,6 +33,203 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+template <bool IsSorted, int SubwarpSize, typename IndexType>
+__global__
+    __launch_bounds__(default_block_size) void find_missing_diagonal_elements(
+        IndexType num_rows, IndexType num_cols,
+        const IndexType *__restrict__ col_idxs,
+        const IndexType *__restrict__ row_ptrs,
+        IndexType *__restrict__ elements_to_add_per_row,
+        bool *__restrict__ changes_required)
+{
+    const auto total_thread_count =
```
upsj: Do we need this?

common/factorization/par_ilu_kernels.hpp.inc
```diff
@@ -33,6 +33,203 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 namespace kernel {
 
 
+namespace detail {
+
+
+// Default implementation for the unsorted case
+template <bool IsSorted>
+struct find_helper {
```
pratikvn: This `find_helper` looks pretty generic. Maybe it can be moved to a common_kernels or something ? A find function i\can be pretty useful in a lot of cases.

common/factorization/par_ilu_kernels.hpp.inc
```diff
@@ -33,6 +33,203 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+        threadIdx.x + static_cast<size_type>(blockIdx.x) * blockDim.x;
+    const auto begin_row = static_cast<IndexType>(tidx / SubwarpSize);
+
+    auto thread_block = group::this_thread_block();
+    auto subwarp_grp = group::tiled_partition<SubwarpSize>(thread_block);
+    const auto subwarp_idx = subwarp_grp.thread_rank();
+
+    bool local_change{false};
+    for (IndexType row = begin_row; row < num_rows; row += total_thread_count) {
+        if (row >= num_cols) {
```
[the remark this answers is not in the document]
pratikvn: I think it would be nice to have. But in any case, if this works only for square matrices, we should probably add a check before calling this kernel ?

common/factorization/par_ilu_kernels.hpp.inc
```diff
@@ -33,6 +33,203 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 namespace kernel {
 
 
+namespace detail {
+
+
+// Default implementation for the unsorted case
+template <bool IsSorted>
```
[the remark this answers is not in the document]
thoasm: Probably, but I did not want to take any chances with the compiler not optimizing it.

common/factorization/par_ilu_kernels.hpp.inc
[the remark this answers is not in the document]
thoasm: In the `generate_l_u` ([here](https://github.com/ginkgo-project/ginkgo/blob/develop/core/factorization/par_ilu.cpp#L77)), we do check if it is square or not.
Here, I simply thought it is extremely easy to also support non-square matrices, so I did it.

common/factorization/par_ilu_kernels.hpp.inc
```diff
@@ -33,6 +33,203 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+    __launch_bounds__(default_block_size) void add_missing_diagonal_elements(
+        IndexType num_rows, const ValueType *__restrict__ old_values,
+        const IndexType *__restrict__ old_col_idxs,
+        const IndexType *__restrict__ old_row_ptrs,
+        ValueType *__restrict__ new_values,
+        IndexType *__restrict__ new_col_idxs,
+        const IndexType *__restrict__ row_ptrs_addition)
+{
+    // Precaution in case not enough threads were created
+    const auto total_thread_count =
```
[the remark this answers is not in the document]
thoasm: I added it because I kind of remembered that HIP had issues with `gridDim.x` overflowing.
I just talked with @yhmtsai and apparently, each dimension is stored in a 32 bit integer (meaning `blockDim.x` has 10 bit, while `gridDim.x` has 22 bit).
I guess I can remove it, if you want. In theory, this allows us to also support huge matrices without needing to use the X or Z dimension, but so far, we never needed it.

core/factorization/par_ilu.cpp
```diff
@@ -82,15 +84,19 @@ ParIlu<ValueType, IndexType>::generate_l_u(
...
+    CsrMatrix *csr_system_matrix{};
+    if (csr_system_matrix_const == nullptr ||
+        csr_system_matrix_const->get_executor() != exec) {
         csr_system_matrix_unique_ptr = CsrMatrix::create(exec);
         as<ConvertibleTo<CsrMatrix>>(system_matrix.get())
             ->convert_to(csr_system_matrix_unique_ptr.get());
-        csr_system_matrix = csr_system_matrix_unique_ptr.get();
+    } else {
+        csr_system_matrix_unique_ptr = std::unique_ptr<CsrMatrix>{
+            static_cast<CsrMatrix *>(system_matrix->clone().release())};
```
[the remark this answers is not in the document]
thoasm: Yes, good point.

cuda/factorization/par_ilu_kernels.cu
thoasm: The constructor of `Array` uses an unsigned value (`size_type`) as the size, so I need to cast it to an unsigned anyway (to avoid a warning).
And I don't think it could "safely" overflow, but I still need the cast to avoid additional warnings.

cuda/factorization/par_ilu_kernels.cu
thoasm: That is true, however, I simply use `as_cuda_type` for all pointers I put into devices.

cuda/factorization/par_ilu_kernels.cu
```diff
@@ -60,9 +65,92 @@ constexpr int default_block_size{512};
...
 
 
+template <typename ValueType, typename IndexType>
+void add_diagonal_elements(std::shared_ptr<const DefaultExecutor> exec,
+                           matrix::Csr<ValueType, IndexType> *mtx,
+                           bool is_sorted)
+{
+    // TODO: Runtime can be optimized by choosing a appropriate size for the
+    //       subwarp dependent on the matrix properties
+    constexpr int subwarp_size = config::warp_size / 2;
```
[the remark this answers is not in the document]
thoasm: I just changed it to `warp_size`. I have no idea what a good value is, I just thought `warp_size / 2` might be a good fit without any benchmarking. You are probably right that the full warp might make more sense.

cuda/factorization/par_ilu_kernels.cu
```diff
@@ -60,9 +65,92 @@ constexpr int default_block_size{512};
...
+
+    auto cuda_old_values = as_cuda_type(mtx->get_const_values());
+    auto cuda_old_col_idxs = as_cuda_type(mtx->get_const_col_idxs());
+    auto cuda_old_row_ptrs = as_cuda_type(mtx->get_row_ptrs());
+    auto cuda_row_ptrs_add = as_cuda_type(row_ptrs_addition.get_data());
+
+    const dim3 block_dim{default_block_size, 1, 1};
+    const dim3 grid_dim{
+        static_cast<uint32>(ceildiv(
+            num_rows, static_cast<size_type>(block_dim.x / subwarp_size))),
```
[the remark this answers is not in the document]
thoasm: I forgot to remove the second cast. Previously, `num_rows` was also of type `size_type`. Now, it works perfectly without the second one (the first one is still required, otherwise, you get a narrowing warning).

common/factorization/par_ilu_kernels.hpp.inc
upsj: I would argue that in the current form, the function is not really general-purpose enough to be useful elsewhere. It might be helpful if it returned the index of the found value (or -1 if it didn't find the value), but that might complicate the rest of the code.

cuda/factorization/par_ilu_kernels.cu
thoasm: I reverted the changes. You are right, updating just one file is inconsistent, we should update all or none.

common/factorization/par_ilu_kernels.hpp.inc
pratikvn: Yes, now that I see it does return a boolean.

cuda/test/factorization/par_ilu_kernels.cpp
```diff
@@ -176,6 +220,63 @@ class ParIlu : public ::testing::Test {
...
+    gko::kernels::cuda::par_ilu_factorization::add_diagonal_elements(
+        cuda, gko::lend(mtx_cuda), false);
+
+    ASSERT_FALSE(mtx_ref->is_sorted_by_column_index());
+    GKO_ASSERT_MTX_NEAR(mtx_ref, mtx_cuda, 0.);
+    GKO_ASSERT_MTX_EQ_SPARSITY(mtx_ref, mtx_cuda);
+}
+
+
+TEST_F(ParIlu, CudaKernelAddDiagonalElementsAsymetricEquivalentToRef)
```
pratikvn: I think you mean non-square ? Square matrices can be asymmetric.

hip/test/factorization/par_ilu_kernels.hip.cpp
```diff
@@ -175,6 +219,66 @@ class ParIlu : public ::testing::Test {
...
+        hip, gko::lend(mtx_hip), false);
+    hip->synchronize();
+
+    ASSERT_FALSE(mtx_ref->is_sorted_by_column_index());
+    GKO_ASSERT_MTX_NEAR(mtx_ref, mtx_hip, 0.);
+    GKO_ASSERT_MTX_EQ_SPARSITY(mtx_ref, mtx_hip);
+}
+
+
+TEST_F(ParIlu, HipKernelAddDiagonalElementsAsymetricEquivalentToRef)
```
pratikvn: Same as in cuda

omp/test/factorization/par_ilu_kernels.cpp
```diff
@@ -175,6 +220,63 @@ class ParIlu : public ::testing::Test {
...
+    gko::kernels::omp::par_ilu_factorization::add_diagonal_elements(
+        omp, gko::lend(mtx_omp), false);
+
+    ASSERT_FALSE(mtx_ref->is_sorted_by_column_index());
+    GKO_ASSERT_MTX_NEAR(mtx_ref, mtx_omp, 0.);
+    GKO_ASSERT_MTX_EQ_SPARSITY(mtx_ref, mtx_omp);
+}
+
+
+TEST_F(ParIlu, OmpKernelAddDiagonalElementsAsymetricEquivalentToRef)
```
pratikvn: Same as in cuda

reference/test/factorization/par_ilu_kernels.cpp
```diff
@@ -186,6 +209,88 @@ class ParIlu : public ::testing::Test {
...
+
+    gko::kernels::reference::par_ilu_factorization::add_diagonal_elements(
+        ref, gko::lend(empty_mtx), true);
+
+    GKO_ASSERT_MTX_NEAR(empty_mtx, expected_mtx, 0.);
+    GKO_ASSERT_MTX_EQ_SPARSITY(empty_mtx, expected_mtx);
+}
+
+
+TEST_F(ParIlu, KernelAddDiagonalElementsAsymetric)
```
pratikvn: Non-square

cuda/test/factorization/par_ilu_kernels.cpp
thoasm: That is true, my bad.

Issue #430, opened 2019-12-19
Inconsistent use of integer types


upsj:
It seems to me Ginkgo is sometimes integer types inconsistently.
This includes both 32/64 bit as well as `signed`/`unsigned`. Examples:

* `dim` returns a `size_type` for all dimensions, even though we internally use the `IndexType` template parameter in almost all kernels.
* Most Reference and OMP kernels use `size_type` as loop variables.
* SellP stores its slice lengths and slice sets as `size_type`

My opinion: I would consistently use signed types everywhere, because we use them for indexing anyways and overflows on signed types are way easier to recognize/catch when debugging. Only for allocation sizes and/or launch bounds, it might make sense to use `size_type`. But that is often a very fundamental debate, so I would like to start it here.

On the question what type `dim` should use, I am conflicted, since it would probably be annoying having to type `dim<IndexType>` everywhere, but in my opinion, it should at least return a signed type.

upsj:
Coming back to this, since @thoasm and I discussed some related issues or questions about code style, I want to propose a few changes that might clean up our code base.

1. Change all integer types to `signed`. The rationale for this is simple: We do this for all (but `dense`) kernels anyways and many popular style guides ([Google](https://google.github.io/styleguide/cppguide.html#Integer_Types) and the [C++ Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-nonnegative)) recommend avoiding unsigned types (except for bit patterns).
This would also mean making the size type for `Array` signed and throwing an exception for `size<0`.
Concerning overflows, this is of course a complex topic, but I would leave it at the following: In case you have a matrix with nnz >= MAX_INT, we would expect you to use int64 types, otherwise there is nothing we can do. Since the largest indices that occur in our kernels are limited by nnz (neglecting "cutoff" thread IDs for large block sizes and counts and/or the use of a whole subwarp per row/non-zero, which we could solve by using the y and z indices), I see no reason why we should expect our code to overflow. And if it does (and the numbers are not extraordinarily big), a negative number is a better indicator for overflown values than a small value where you would not expect it.
2. Make sure all "external" functions we use have wrappers that return `signed` values in case they don't do that already. This includes the CUDA/HIP cooperative groups, intrinsics, but also the threadIdx and blockIdx/Dim variables, which we could wrap either in a `get_thread_id<IndexType>` function or use our cooperative_groups grid group everywhere.
[...]

PR #457, opened 2020-01-29
Use parallel merge in HIP advanced spgemm

upsj:
Lo and behold, the `group_merge` implementation has a tiny but nasty Guttenbug, which this PR fixes.
I found it when testing the HIP spgeam for advanced spgemm, which this PR also adds.
Finally, this PR improves the test reporting for mismatching sparsity patterns and fixes a small compiler warning in the test utils.

### DIFF
--- a/common/matrix/csr_kernels.hpp.inc
+++ b/common/matrix/csr_kernels.hpp.inc
@@ -494,95 +494,95 @@ __global__ __launch_bounds__(spmv_block_size) void abstract_classical_spmv(
 }
 
 
-template <typename IndexType>
+template <int subwarp_size, typename IndexType>
 __global__ __launch_bounds__(default_block_size) void spgeam_nnz(
     const IndexType *a_row_ptrs, const IndexType *a_col_idxs,
     const IndexType *b_row_ptrs, const IndexType *b_col_idxs,
     IndexType num_rows, IndexType *nnz)
 {
-    constexpr auto sentinel = device_numeric_limits<IndexType>::max;
-    auto row = threadIdx.x + blockDim.x * static_cast<size_type>(blockIdx.x);
+    auto row =
+        (threadIdx.x + blockDim.x * size_type(blockIdx.x)) / subwarp_size;
+    auto subwarp =
+        group::tiled_partition<subwarp_size>(group::this_thread_block());
     if (row >= num_rows) {
         return;
     }
 
     auto a_begin = a_row_ptrs[row];
     auto b_begin = b_row_ptrs[row];
-    auto a_end = a_row_ptrs[row + 1];
-    auto b_end = b_row_ptrs[row + 1];
-    auto a_size = a_end - a_begin;
-    auto b_size = b_end - b_begin;
-    auto a_ptr = a_begin;
-    auto b_ptr = b_begin;
-    auto a_col = a_ptr < a_end ? a_col_idxs[a_ptr] : sentinel;
-    auto b_col = b_ptr < b_end ? b_col_idxs[b_ptr] : sentinel;
+    auto a_size = a_row_ptrs[row + 1] - a_begin;
+    auto b_size = b_row_ptrs[row + 1] - b_begin;
     IndexType count{};
-    IndexType i{};
-    while (i < a_size + b_size) {
-        auto advance_a = (a_col <= b_col);
-        auto advance_b = (b_col <= a_col);
-        i += advance_a + advance_b;
-        a_ptr += advance_a;
-        b_ptr += advance_b;
-        if (advance_a) {
-            a_col = a_ptr < a_end ? a_col_idxs[a_ptr] : sentinel;
-        }
-        if (advance_b) {
-            b_col = b_ptr < b_end ? b_col_idxs[b_ptr] : sentinel;
-        }
-        ++count;
+    group_merge<subwarp_size>(
+        a_col_idxs + a_begin, a_size, b_col_idxs + b_begin, b_size, subwarp,
+        [&](IndexType, IndexType a_col, IndexType, IndexType b_col, IndexType,
+            bool valid) {
+            count += popcnt(subwarp.ballot(a_col != b_col && valid));
+        });
+
+    if (subwarp.thread_rank() == 0) {
+        nnz[row] = count;
     }
-    nnz[row] = count;
 }
 
 
-template <typename IndexType, typename ValueType>
+template <int subwarp_size, typename IndexType, typename ValueType>
 __global__ __launch_bounds__(default_block_size) void spgeam(
     ValueType alpha, const IndexType *a_row_ptrs, const IndexType *a_col_idxs,
     const ValueType *a_vals, ValueType beta, const IndexType *b_row_ptrs,
     const IndexType *b_col_idxs, const ValueType *b_vals, IndexType num_rows,
     const IndexType *c_row_ptrs, IndexType *c_col_idxs, ValueType *c_vals)
 {
-    constexpr auto sentinel = device_numeric_limits<IndexType>::max;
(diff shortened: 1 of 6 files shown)


common/matrix/csr_kernels.hpp.inc
```diff
@@ -494,95 +494,95 @@ __global__ __launch_bounds__(spmv_block_size) void abstract_classical_spmv(
...
+template <int subwarp_size, typename IndexType>
 __global__ __launch_bounds__(default_block_size) void spgeam_nnz(
     const IndexType *a_row_ptrs, const IndexType *a_col_idxs,
     const IndexType *b_row_ptrs, const IndexType *b_col_idxs,
     IndexType num_rows, IndexType *nnz)
 {
-    constexpr auto sentinel = device_numeric_limits<IndexType>::max;
-    auto row = threadIdx.x + blockDim.x * static_cast<size_type>(blockIdx.x);
+    auto row =
+        (threadIdx.x + blockDim.x * size_type(blockIdx.x)) / subwarp_size;
```
thoasm: nit: can you keep the `static_cast`?
Alternatively, if `blockIdx.x` is actually an unsigned, this should also work:
```suggestion
        (threadIdx.x + blockDim.x * size_type{blockIdx.x}) / subwarp_size;
```

common/matrix/csr_kernels.hpp.inc
```diff
@@ -494,95 +494,95 @@ __global__ __launch_bounds__(spmv_block_size) void abstract_classical_spmv(
...
 __global__ __launch_bounds__(default_block_size) void spgeam(
     ValueType alpha, const IndexType *a_row_ptrs, const IndexType *a_col_idxs,
     const ValueType *a_vals, ValueType beta, const IndexType *b_row_ptrs,
     const IndexType *b_col_idxs, const ValueType *b_vals, IndexType num_rows,
     const IndexType *c_row_ptrs, IndexType *c_col_idxs, ValueType *c_vals)
 {
-    constexpr auto sentinel = device_numeric_limits<IndexType>::max;
-    auto row = threadIdx.x + blockDim.x * static_cast<size_type>(blockIdx.x);
+    auto row =
+        (threadIdx.x + blockDim.x * size_type(blockIdx.x)) / subwarp_size;
```
thoasm: nit: same here. I would prefer to keep the `static_cast`, but it could also work with `{}`.
```suggestion
        (threadIdx.x + blockDim.x * size_type{blockIdx.x}) / subwarp_size;
```

common/matrix/csr_kernels.hpp.inc
```diff
@@ -494,95 +494,95 @@ __global__ __launch_bounds__(spmv_block_size) void abstract_classical_spmv(
...
+    auto subwarp =
+        group::tiled_partition<subwarp_size>(group::this_thread_block());
     if (row >= num_rows) {
         return;
     }
 
+    auto lane = IndexType(subwarp.thread_rank());
```
thoasm: Why not simply:
```suggestion
    IndexType lane = subwarp.thread_rank();
```
instead of a functional cast? Also, I do consider functional- & C-casts dangerous if done with a template parameter (which is not guaranteed to be a simple type).

hip/matrix/csr_kernels.hip.cpp
```diff
@@ -524,6 +529,53 @@ void spgemm(std::shared_ptr<const HipExecutor> exec,
...
+
+template <int subwarp_size, typename ValueType, typename IndexType>
+void spgeam(syn::value_list<int, subwarp_size>,
+            std::shared_ptr<const HipExecutor> exec, ValueType alpha,
+            const IndexType *a_row_ptrs, const IndexType *a_col_idxs,
+            const ValueType *a_vals, ValueType beta,
+            const IndexType *b_row_ptrs, const IndexType *b_col_idxs,
+            const ValueType *b_vals, matrix::Csr<ValueType, IndexType> *c)
+{
+    auto m = IndexType(c->get_size()[0]);
```
thoasm: Please use either a `static_cast` or
```suggestion
    IndexType m = c->get_size()[0];
```

cuda/matrix/csr_kernels.cu
```diff
@@ -57,6 +57,8 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "cuda/base/types.hpp"
 #include "cuda/components/atomic.cuh"
 #include "cuda/components/cooperative_groups.cuh"
+#include "cuda/components/intrinsics.cuh"
+#include "cuda/components/merging.cuh"
```
thoasm: Are you sure you need them in `csr_kernels.cu`, even though you did not add any code which uses it?

hip/matrix/csr_kernels.hip.cpp
```diff
@@ -60,6 +60,8 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "hip/base/types.hip.hpp"
 #include "hip/components/atomic.hip.hpp"
 #include "hip/components/cooperative_groups.hip.hpp"
+#include "hip/components/intrinsics.hip.hpp"
+#include "hip/components/merging.hip.hpp"
```
thoasm: Are you really using them here?

common/matrix/csr_kernels.hpp.inc
upsj: I guess that we've had this discussion often enough that it would make sense to find either a guideline or a non-guideline (i.e., we don't care) for this type of operation. To me, this is heavily connected with https://github.com/ginkgo-project/ginkgo/issues/430 since ideally, I would like to 1. maybe refactor our kernels so these types of calculations are executed in a device function or use cooperative_group grid groups and 2. get rid of as many static_cast (and of course functional-style cast) operations by making our size types consistent everywhere. I will elaborate a bit in #430.

After thinking about it for a while, I slightly agree with your viewpoint, but also to find the amount of static_casts in our code base too big.

common/matrix/csr_kernels.hpp.inc
upsj: Good point, that is probably easier. I am probably too heavily in the "[AAAARGH](https://www.youtube.com/watch?v=ZCGyvPDM0YY)" camp with my code. But I would seriously question every piece of code that used a non-primitive IndexType.

upsj: Wait no, thread_rank(), as well as most other functions of cooperative groups, returns an unsigned value, so this would produce a warning

thoasm: I agree that IndexType should be a primitive type.
I guess there are edge cases where if you call a function with a wrong parameter type, it could happen that `IndexType` is suddenly a pointer or something completely different. I hope it would not compile that way, but I guess it could happen.

hip/matrix/csr_kernels.hip.cpp
upsj: Yes, see spgeam in common/matrix/csr_kernels.hpp.inc

thoasm: Oh, my bad, thanks for clearing that up.

cuda/matrix/csr_kernels.cu
upsj: > /ginkgo/common/matrix/csr_kernels.hpp.inc(516): error: identifier "group_merge" is undefined
> /ginkgo/common/matrix/csr_kernels.hpp.inc(556): error: identifier "group_merge" is undefined

Yes, it seems like nvcc tries to instantiate the templates even though we never instantiate them.

thoasm: I completely missed the `common/matrix/csr_kernels.hpp.inc` file (thought it was also a `merge` file) when looking at use cases, by bad.

PR #460, opened 2020-02-12
Add Shuffle test

yhmtsai:
This PR add the shuffle test.

`shuffle_impl` before #439 is wrong when modifying the device array during the call of shuffle. It is only wrong in double, but correct in `complex<double>`. Also, it is correct when the value is on the register.

The failed [pipeline](https://gitlab.com/ginkgo-project/ginkgo-public-ci/pipelines/117269496) uses the `shuffle_impl` before #439.

### DIFF
--- a/hip/test/components/cooperative_groups.hip.cpp
+++ b/hip/test/components/cooperative_groups.hip.cpp
@@ -47,6 +47,10 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/base/executor.hpp>
 
 
+#include "core/test/utils.hpp"
+#include "hip/base/types.hip.hpp"
+
+
 namespace {
 
 
@@ -82,6 +86,18 @@ protected:
         hipLaunchKernelGGL(HIP_KERNEL_NAME(kernel), dim3(1),
                            dim3(config::warp_size / 2), 0, 0,
                            dresult.get_data());
+        result = dresult;
+        auto success = *result.get_const_data();
+
+        ASSERT_TRUE(success);
+    }
+
+    template <typename Kernel>
+    void test_several_block(Kernel kernel)
+    {
+        hipLaunchKernelGGL(HIP_KERNEL_NAME(kernel), dim3(10240),
+                           dim3(config::warp_size, 256 / config::warp_size), 0,
+                           0, dresult.get_data());
         result = dresult;
         auto success = *result.get_const_data();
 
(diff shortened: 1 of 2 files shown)


hip/test/components/cooperative_groups.hip.cpp
```diff
@@ -262,4 +266,53 @@ TEST_F(CooperativeGroups, SubwarpBallot) { test(cg_subwarp_ballot); }
...
+
+TEST_F(CooperativeGroups, ShuffleSumDouble)
+{
+    int num = 4;
+    uint64_t x = 0x401022C90008B240;
+    gko::Array<double> value(ref, config::warp_size);
+    gko::Array<double> answer(ref, config::warp_size);
+    gko::Array<double> dvalue(hip);
+    for (int i = 0; i < value.get_num_elems(); i++) {
+        value.get_data()[i] = *reinterpret_cast<double *>(&x);
```
thoasm: Here, I would not use `reinterpret_cast` since it results in *undefined behavior* (see [above *Defect Reports* here](https://en.cppreference.com/w/cpp/language/reinterpret_cast) and the linked [letter to SG12](http://www.open-std.org/pipermail/ub/2016-February/000565.html)). Since we don't care about runtime in this case, I would propose to do it the proper way (might even be better optimized):
```
#include <string> // needed for memcpy
// ...
uint64_t x = 0x401022C90008B240; // Not sure if this is necessary, but assume it is
double x_dbl{};
std::memcpy(&x_dbl, &x, sizeof(x_dbl)); // now, x_dbl contains the bits of x
```
This has 2 advantages:
- We comply with the standard
- You only do it once instead of `config::warp_size`  times (this should definitively change)

hip/test/components/cooperative_groups.hip.cpp
```diff
@@ -262,4 +266,53 @@ TEST_F(CooperativeGroups, SubwarpBallot) { test(cg_subwarp_ballot); }
...
+}
+
+TEST_F(CooperativeGroups, ShuffleSumComplexDouble)
+{
+    int num = 4;
+    uint64_t x = 0x401022C90008B240;
+    gko::Array<std::complex<double>> value(ref, config::warp_size);
+    gko::Array<std::complex<double>> answer(ref, config::warp_size);
+    gko::Array<std::complex<double>> dvalue(hip);
+    const double x_double = *reinterpret_cast<double *>(&x);
```
thoasm: This is a bit better than the test above (only a single `reinterpret_cast`), but I would prefer the `memcpy` way since it complies with the standard.

hip/test/components/cooperative_groups.hip.cpp
```diff
@@ -262,4 +266,53 @@ TEST_F(CooperativeGroups, SubwarpBallot) { test(cg_subwarp_ballot); }
...
+        group::tiled_partition<config::warp_size>(group::this_thread_block());
+    for (int ind = 0; ind < num; ind++) {
+        value[group.thread_rank()] += group.shfl(value[ind], ind);
+    }
+}
+
+TEST_F(CooperativeGroups, ShuffleSumDouble)
+{
+    int num = 4;
+    uint64_t x = 0x401022C90008B240;
```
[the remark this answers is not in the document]
yhmtsai: image
I pick up this value for the fixed internal result without rounding error.
Also, this value contains 1 in the last 32-bit which is lost in shuffle.
but others are randomly picked and point symmetric.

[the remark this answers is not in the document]
thoasm: Good to know, I was just curious. Thanks for explaining it to me.

hip/test/components/cooperative_groups.hip.cpp
```diff
@@ -262,4 +267,61 @@ TEST_F(CooperativeGroups, SubwarpBallot) { test(cg_subwarp_ballot); }
...
+__global__ void cg_shuffle_sum(const int num, ValueType *__restrict__ value)
+{
+    auto group =
+        group::tiled_partition<config::warp_size>(group::this_thread_block());
+    for (int ind = 0; ind < num; ind++) {
+        value[group.thread_rank()] += group.shfl(value[ind], ind);
+    }
+}
+
+TEST_F(CooperativeGroups, ShuffleSumDouble)
```
thoasm: For some reason, this file only has a single empty line between the tests (`TEST_F`).
In pretty much all other tests I was looking into, we have 2 empty lines to separate the tests. Could you also do that for this file?

yhmtsai: Sure, they are done.

PR #463, opened 2020-02-25
Fix sonarqube warnings

upsj:
Our develop branch currently has a bad SonarCloud reliability rating, and since I think these warnings actually have merit, this PR addresses them.

### DIFF
--- a/core/base/composition.cpp
+++ b/core/base/composition.cpp
@@ -43,12 +43,12 @@ namespace {
 template <typename ValueType, typename OpIterator, typename VecIterator>
 inline void allocate_vectors(OpIterator begin, OpIterator end, VecIterator res)
 {
-    for (auto it = begin; it != end; ++it, ++res) {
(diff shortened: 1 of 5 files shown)


core/base/composition.cpp
```diff
@@ -43,12 +43,13 @@ namespace {
...
 inline void allocate_vectors(OpIterator begin, OpIterator end, VecIterator res)
 {
-    for (auto it = begin; it != end; ++it, ++res) {
+    for (auto it = begin; it != end; ++it) {
         if (*res != nullptr && (*res)->get_size()[0] == (*it)->get_size()[0]) {
             continue;
         }
         *res = matrix::Dense<ValueType>::create(
             (*it)->get_executor(), gko::dim<2>{(*it)->get_size()[0], 1});
+        ++res;
```
tcojean: Is this correct? If `*res != nullptr` (and the other condition is satisfied), then we never increase `res` and check the next one.

upsj: Good catch, I now inverted the condition to keep the same code structure

PR #464, opened 2020-02-25
Extract thread ID computation from GPU kernels

upsj:
At the moment most of our code is filled with `threadIdx.x + blockDim.x * blockIdx.x` and similar expressions. There are some issues with this approach:

1. The product can overflow since the calculation is executed using unsigned 32 bit values. This is even true if the matrix dimensions are below MAX_INT, since we sometimes use multiple threads per row.
2. The resulting type of this statement is `unsigned` by default. It would instead be better if we could have a robust method to determine the thread or subwarp ID expressed in a given type.

This PR adds `get_thread_id_flat` and `get_subwarp_id_flat` functions that work for one-dimensional thread indexing (using only `*Idx.x`) and uses them where possible, which at the same time fixes some potential overflow bugs.

This is a first attempt to solve some of the issues listed in #430. But there might be some discussion warranted, especially in terms of the interaction with `cooperative_groups`, which always use 3D indexing.

Also feel free to add more locations where we could use these functions, as I only grepped for the most common patterns.

### DIFF
--- a/cuda/preconditioner/jacobi_kernels.cu
+++ b/cuda/preconditioner/jacobi_kernels.cu
@@ -57,7 +57,7 @@ namespace {
 
 
 // a total of 32 warps (1024 threads)
-constexpr int default_block_size = 32;
+constexpr int default_num_warps = 32;
 // with current architectures, at most 32 warps can be scheduled per SM (and
 // current GPUs have at most 84 SMs)
 constexpr int default_grid_size = 32 * 32 * 128;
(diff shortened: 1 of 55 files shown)


cuda/preconditioner/jacobi_kernels.cu
```diff
@@ -75,7 +76,7 @@ size_type find_natural_blocks(std::shared_ptr<const CudaExecutor> exec,
 
     Array<bool> matching_next_row(exec, mtx->get_size()[0] - 1);
 
-    const dim3 block_size(default_block_size, 1, 1);
+    const dim3 block_size(config::warp_size, 1, 1);
```
yhmtsai: It is caused by confusing name.
`default_block_size` is the number of warp not the block size.
It only affects `initialize_precisions` now.
Could you also change the name like `default_num_warp` or something else?

upsj: Good point, I renamed the variable.

PR #465, opened 2020-02-25
Improve isfinite

thoasm:
For some reason, compiling `::gko::isfinite` failed when using HIP in combination with `clang` and an NVIDIA GPU in some cases (See this [pipeline](https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/448090521) as an example). When HIP is not enabled, the pipeline example compiles without any errors.
This PR addresses that by simplifying some of the code we have for `isfinite` while improving the hand-written version. It now translates to the same PTX code as a call to `isfinite`. Additionally, both the hand-written and the default `gko::isfinite` are now tested simultaneously.

### DIFF
--- a/common/base/math.hpp.inc
+++ b/common/base/math.hpp.inc
@@ -116,6 +116,8 @@ custom_isfinite(T value)
 
 GKO_DEFINE_ISFINITE_FOR_TYPE(float)
 GKO_DEFINE_ISFINITE_FOR_TYPE(double)
+GKO_DEFINE_ISFINITE_FOR_TYPE(thrust::complex<float>)
+GKO_DEFINE_ISFINITE_FOR_TYPE(thrust::complex<double>)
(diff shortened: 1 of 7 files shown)


common/base/math.hpp.inc
```diff
@@ -108,10 +108,10 @@ custom_isfinite(T value)
...
 
 
-#define GKO_DEFINE_ISFINITE_FOR_TYPE(_type)          \
-    GKO_INLINE __device__ bool isfinite(_type value) \
-    {                                                \
-        return detail::custom_isfinite(value);       \
+#define GKO_DEFINE_ISFINITE_FOR_TYPE(_type)                 \
+    GKO_INLINE __device__ bool isfinite(const _type &value) \
+    {                                                       \
+        return detail::custom_isfinite(value);              \
```
yhmtsai: both versions work in MSVC

Issue #461, opened 2020-02-20
CUDA executor always calls `cudaDeviceReset()` when destroyed


thoasm:
Currently, the CUDA executor always calls `cudaDeviceReset()` ([documentation for `cudaDeviceReset()`](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1gef69dd5c6d0206c2b8d099abac61f217)) when it is destroyed (and the last executor on that device). This is done in [cuda/base/executor.cpp](https://github.com/ginkgo-project/ginkgo/blob/develop/cuda/base/executor.cpp#L67). The problem is that this call frees all allocated memory, including everything that was done by other libraries or the user itself.

We first need to investigate why we need this call to begin with. It is likely that we need it for `cuda-memcheck` for [Leak Checking](https://docs.nvidia.com/cuda/cuda-memcheck/index.html#leak-checking), but it is possible that it is not the only reason why it is there.

For multi-GPU, we might actually have to use a combination of [cudaGetDevice(int)](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1g80861db2ce7c29b6e8055af8ae01bc78) and [`cudaSetDevice(int)`](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1g159587909ffa0791bbe4b40187a4c6bb) to reset the correct GPU.

This issue was found by @nbeams

PR #467, opened 2020-02-27
Speed up GPU tests by avoiding device reset

upsj:
At the moment, all of our GPU tests store a device executor that gets destroyed after each test, which in turn executes `cuda/hipDeviceReset`. This causes each of our tests to take at least 800ms on NVIDIA GPUs (less on AMD GPUs), independent of the actual kernel runtime. I think this is not necessary and can be fixed by instantiating a second device executor in a global variable (other, cleaner solutions welcome!)

Somewhat related to #461 

Before:
```
151: [==========] Running 24 tests from 1 test case.
151: [----------] Global test environment set-up.
151: [----------] 24 tests from Dense
151: [ RUN      ] Dense.SingleVectorHipScaleIsEquivalentToRef
151: [       OK ] Dense.SingleVectorHipScaleIsEquivalentToRef (1012 ms)
...
151: [ RUN      ] Dense.CalculateTotalColsIsEquivalentToRef
151: [       OK ] Dense.CalculateTotalColsIsEquivalentToRef (864 ms)
151: [----------] 24 tests from Dense (22293 ms total)
```
After:
```
151: [==========] Running 24 tests from 1 test case.
151: [----------] Global test environment set-up.
151: [----------] 24 tests from Dense
151: [ RUN      ] Dense.SingleVectorHipScaleIsEquivalentToRef
151: [       OK ] Dense.SingleVectorHipScaleIsEquivalentToRef (95 ms)
...
151: [ RUN      ] Dense.CalculateTotalColsIsEquivalentToRef
151: [       OK ] Dense.CalculateTotalColsIsEquivalentToRef (20 ms)
151: [----------] 24 tests from Dense (1783 ms total)
```

### DIFF
--- a/cuda/test/utils.hpp
+++ b/cuda/test/utils.hpp
@@ -30,8 +30,8 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ******************************<GINKGO LICENSE>*******************************/
 
-#ifndef GKO_CUDA_TEST_NORESET_HPP_
-#define GKO_CUDA_TEST_NORESET_HPP_
+#ifndef GKO_CUDA_TEST_UTILS_HPP_
+#define GKO_CUDA_TEST_UTILS_HPP_
 
 
 #include <ginkgo/core/base/executor.hpp>
@@ -51,4 +51,4 @@ auto no_reset_exec =
 }  // namespace
 
 
-#endif  // GKO_CUDA_TEST_NORESET_HPP_
+#endif  // GKO_CUDA_TEST_UTILS_HPP_
--- a/cuda/test/utils.hpp
+++ b/cuda/test/utils.hpp
@@ -30,56 +30,24 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ******************************<GINKGO LICENSE>*******************************/
 
+#ifndef GKO_CUDA_TEST_NORESET_HPP_
+#define GKO_CUDA_TEST_NORESET_HPP_
 
-#include <core/test/utils/assertions.hpp>
-
-
-#include <gtest/gtest.h>
-
-
-#include <ginkgo/core/matrix/csr.hpp>
-#include <ginkgo/core/matrix/dense.hpp>
+#include <ginkgo/core/base/executor.hpp>
 
 
-#include "hip/test/utils.hip.hpp"
+#include "core/test/utils.hpp"
 
 
 namespace {
 
 
-class MatricesNear : public ::testing::Test {
-protected:
-    void SetUp()
-    {
-        ASSERT_GT(gko::HipExecutor::get_num_devices(), 0);
-        ref = gko::ReferenceExecutor::create();
-        hip = gko::HipExecutor::create(0, ref);
-    }
-
-    void TearDown()
-    {
-        if (hip != nullptr) {
-            ASSERT_NO_THROW(hip->synchronize());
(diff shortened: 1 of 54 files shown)


thoasm:
What about creating a `cuda/test/utils.hpp` header (which includes `core/test/utils.hpp`), and do the creation of the executor there? That would at least hide the not so pleasant code inside a header and would allow us to improve / remove that in one line instead of in every test-file.
This would obviously require us changing all includes of `core/test/utils.hpp` to `cuda/test/utils.hpp`, but that is easy to automate (only needs to be done once).

upsj:
@thoasm Good idea, I followed your suggestion and only had to patch a small compilation issue with `assert` and the comma operator in `constexpr` contexts for HIP.

yhmtsai:
Does `return GKO_ASSERT(...), value;` -> `GKO_ASSERT(...); return value;`  not work?
Or it can not be constexpr?

upsj:
In C++11, constexpr functions are only allowed to consist of a single return statement, so that unfortunately doesn't work. The expected behavior for `NDEBUG` would be that `assert` is replaced by something like `void(0)`, but this weird combination of includes seems to let some asserts vanish. I did not investigate this further yet, since the workaround works.

cuda/test/utils.hpp
```diff
@@ -0,0 +1,53 @@
...
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+#ifndef GKO_CUDA_TEST_NORESET_HPP_
+#define GKO_CUDA_TEST_NORESET_HPP_
+
```
tcojean: Missing one newline

cuda/test/utils.hpp
```diff
@@ -0,0 +1,54 @@
...
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+#ifndef GKO_CUDA_TEST_NORESET_HPP_
+#define GKO_CUDA_TEST_NORESET_HPP_
```
pratikvn:
```suggestion
#ifndef GKO_CUDA_TEST_UTILS_HPP_
#define GKO_CUDA_TEST_UTILS_HPP_
```

cuda/test/utils.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+// prevent device reset after each test
+auto no_reset_exec =
+    gko::CudaExecutor::create(0, gko::ReferenceExecutor::create());
+
+
+}  // namespace
+
+
+#endif  // GKO_CUDA_TEST_NORESET_HPP_
```
pratikvn:
```suggestion
#endif  // GKO_CUDA_TEST_UTILS_HPP_
```

cuda/test/utils.hpp
upsj: Good catch, thanks!

PR #468, opened 2020-02-28
Cuda10.1 generic api: CSR and COO SpMV

yhmtsai:
This PR add the new API of CSR and COO SpMV from cuda 10.1 updated 1.

### DIFF
--- a/benchmark/utils/cuda_linops.hpp
+++ b/benchmark/utils/cuda_linops.hpp
@@ -56,7 +56,10 @@ class CuspBase : public gko::LinOp {
 public:
     cusparseMatDescr_t get_descr() const { return this->descr_.get(); }
 
-    const gko::CudaExecutor *get_gpu_exec() const { return gpu_exec_.get(); }
+    std::shared_ptr<const gko::CudaExecutor> get_gpu_exec() const
+    {
+        return gpu_exec_;
+    }
 
 protected:
     void apply_impl(const gko::LinOp *, const gko::LinOp *, const gko::LinOp *,
@@ -481,7 +484,7 @@ private:
 
 
 template <typename ValueType>
-void cusp_generic_spmv(const gko::CudaExecutor *gpu_exec,
+void cusp_generic_spmv(std::shared_ptr<const gko::CudaExecutor> gpu_exec,
                        const cusparseSpMatDescr_t mat,
                        const gko::Array<ValueType> &scalars,
                        const gko::LinOp *b, gko::LinOp *x,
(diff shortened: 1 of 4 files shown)


benchmark/utils/cuda_linops.hpp
```diff
@@ -475,6 +477,197 @@ class CuspHybrid
...
+        mat, vecb, &scalars.get_const_data()[1], vecx, cu_value, alg, dBuffer));
+    GKO_ASSERT_NO_CUSPARSE_ERRORS(cusparseDestroyDnVec(vecx));
+    GKO_ASSERT_NO_CUSPARSE_ERRORS(cusparseDestroyDnVec(vecb));
+    GKO_ASSERT_NO_CUDA_ERRORS(cudaFree(dBuffer));
+}
+
+
+template <typename ValueType = gko::default_precision,
+          typename IndexType = gko::int32,
+          cusparseSpMVAlg_t Alg = CUSPARSE_MV_ALG_DEFAULT>
```
upsj: Could you not also make this a normal parameter instead of a template parameter? Or would this interfere with your other code?

benchmark/utils/cuda_linops.hpp
```diff
@@ -475,6 +477,197 @@ class CuspHybrid
...
+class CuspGenericCsr
+    : public gko::EnableLinOp<CuspGenericCsr<ValueType, IndexType, Alg>,
+                              CuspBase>,
+      public gko::EnableCreateMethod<CuspGenericCsr<ValueType, IndexType, Alg>>,
+      public gko::ReadableFromMatrixData<ValueType, IndexType> {
+    friend class gko::EnableCreateMethod<CuspGenericCsr>;
+    friend class gko::EnablePolymorphicObject<CuspGenericCsr, CuspBase>;
+
+public:
+    using csr = gko::matrix::Csr<ValueType, IndexType>;
```
upsj: nit: upper case?
```suggestion
    using Csr = gko::matrix::Csr<ValueType, IndexType>;
```

benchmark/utils/formats.hpp
```diff
@@ -166,8 +176,9 @@ std::unique_ptr<MatrixType> read_matrix_from_data(
 const std::map<std::string, std::function<std::unique_ptr<gko::LinOp>(
                                 std::shared_ptr<const gko::Executor>,
                                 const gko::matrix_data<> &)>>
-    matrix_factory{
-        {"csr", READ_MATRIX(csr, std::make_shared<csr::automatical>())},
+    matrix_factory
+{
+    {"csr", READ_MATRIX(csr, std::make_shared<csr::automatical>())},
```
pratikvn: weird indent ?

yhmtsai: clang-format seems to do different behavior between `#if` and `#ifdef`
If all begin with `#ifdef`, the format will be the same as before.
If some begin with `#if`, the format will be like here.
I use `#clang-format off` and `on` to format it now.
Does anyone have idea to handle `#if` and `#ifdef`?

benchmark/utils/cuda_linops.hpp
yhmtsai: I am not sure whether it is needed.
we use both of `using snake_case` and `using CamelCase`

benchmark/utils/cuda_linops.hpp
yhmtsai: I use it for the second algorithm.
`using cusp_gcsr2 =
    detail::CuspGenericCsr<double, gko::int32, CUSPARSE_CSRMV_ALG2>;`

upsj: I see, thanks for the clarification!

benchmark/utils/cuda_linops.hpp
upsj: Fair enough, you can ignore my comment then :)

benchmark/utils/cuda_linops.hpp
```diff
@@ -475,6 +477,197 @@ class CuspHybrid
...
+        &vecb, dense_b->get_num_stored_elements(),
+        as_culibs_type(const_cast<ValueType *>(db)), cu_value));
+
+    void *dBuffer = NULL;
+    size_t bufferSize = 0;
+    GKO_ASSERT_NO_CUSPARSE_ERRORS(cusparseSpMV_bufferSize(
+        gpu_exec->get_cusparse_handle(), trans, &scalars.get_const_data()[0],
+        mat, vecb, &scalars.get_const_data()[1], vecx, cu_value, alg,
+        &bufferSize));
+    GKO_ASSERT_NO_CUDA_ERRORS(cudaMalloc(&dBuffer, bufferSize));
```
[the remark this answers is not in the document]
yhmtsai: done

PR #469, opened 2020-02-28
Update Permutation class and kernels.

pratikvn:
This PR updates the permutation class and the kernels.

+ The constructor is updated to take in the type of permutation with the sizes.
+ Similar to the strategies, the permutation masks can be get and set to allow for some flexibility.
+ The CUDA and HIP permutation kernels for row, column and their inverses are added.

### TODO
+ [x] Add HIP kernels and tests.

### DIFF
--- a/common/matrix/dense_kernels.hpp.inc
+++ b/common/matrix/dense_kernels.hpp.inc
@@ -361,4 +361,80 @@ __global__ __launch_bounds__(default_block_size) void reduce_total_cols(
 }
 
 
+template <size_type block_size, typename IndexType, typename ValueType>
+__global__ __launch_bounds__(block_size) void row_permute(
+    size_type num_rows, size_type num_cols,
+    const IndexType *__restrict__ perm_idxs, const ValueType *__restrict__ orig,
+    size_type stride_orig, ValueType *__restrict__ result,
+    size_type stride_result)
+{
+    constexpr auto warps_per_block = block_size / config::warp_size;
+    const auto global_id =
+        thread::get_thread_id<config::warp_size, warps_per_block>();
+    const auto row_id = global_id / num_cols;
+    const auto col_id = global_id % num_cols;
+    if (row_id < num_rows) {
+        result[row_id * stride_result + col_id] =
+            orig[perm_idxs[row_id] * stride_orig + col_id];
+    }
+}
+
+
+template <size_type block_size, typename IndexType, typename ValueType>
+__global__ __launch_bounds__(block_size) void column_permute(
+    size_type num_rows, size_type num_cols,
+    const IndexType *__restrict__ perm_idxs, const ValueType *__restrict__ orig,
+    size_type stride_orig, ValueType *__restrict__ result,
+    size_type stride_result)
+{
+    constexpr auto warps_per_block = block_size / config::warp_size;
+    const auto global_id =
+        thread::get_thread_id<config::warp_size, warps_per_block>();
+    const auto row_id = global_id / num_cols;
+    const auto col_id = global_id % num_cols;
+    if (row_id < num_rows) {
+        result[row_id * stride_result + col_id] =
+            orig[row_id * stride_orig + perm_idxs[col_id]];
+    }
+}
+
+
+template <size_type block_size, typename IndexType, typename ValueType>
+__global__ __launch_bounds__(block_size) void inverse_row_permute(
+    size_type num_rows, size_type num_cols,
+    const IndexType *__restrict__ perm_idxs, const ValueType *__restrict__ orig,
+    size_type stride_orig, ValueType *__restrict__ result,
+    size_type stride_result)
+{
+    constexpr auto warps_per_block = block_size / config::warp_size;
+    const auto global_id =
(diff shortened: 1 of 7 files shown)


common/matrix/dense_kernels.hpp.inc
```diff
@@ -365,4 +365,76 @@ __global__ __launch_bounds__(default_block_size) void reduce_total_cols(
...
+    const IndexType *__restrict__ perm_idxs, const ValueType *__restrict__ orig,
+    ValueType *__restrict__ result, size_type stride_orig)
+{
+    constexpr auto warps_per_block = block_size / config::warp_size;
+    const auto global_id =
+        thread::get_thread_id<config::warp_size, warps_per_block>();
+    const auto row_id = global_id / num_cols;
+    const auto col_id = global_id % num_cols;
+    if (row_id < num_rows) {
+        result[row_id * stride_orig + col_id] =
```
upsj: Could we also allow different strides here? Otherwise we might need to make additional allocations in case the output doesn't have the correct stride. Or do we not want this?

pratikvn: Sorry, you are right. I should be using the stride of the result here.

core/test/matrix/permutation.cpp
```diff
@@ -100,6 +100,15 @@ TEST_F(Permutation, ReturnsNullValuesArrayWhenEmpty)
 }
 
 
+TEST_F(Permutation, FactorySetsCorrectPermuteMask)
```
tcojean: Now that the multiprecision testing is merged, this will need to slightly change. You only have to do it for these tests here, and you pretty much only have to do like what was there before. Hopefully it'll be rather fast.

As a measure of success, you can use the coverage report, if the coverage doesn't drop significantly then that means the templated tests are correct.

include/ginkgo/core/matrix/permutation.hpp
```diff
@@ -114,6 +114,23 @@ class Permutation : public EnableLinOp<Permutation<IndexType>>,
         return permutation_.get_num_elems();
     }
 
+    /**
+     * Get the permute masks
+     *
+     * @param permute_mask the permute masks
+     */
+    mask_type get_permute_mask() { return enabled_permute_; }
```
tcojean: Maybe const makes sense here.
```suggestion
    mask_type get_permute_mask() const { return enabled_permute_; }
```

include/ginkgo/core/matrix/permutation.hpp
```diff
@@ -122,21 +139,23 @@ class Permutation : public EnableLinOp<Permutation<IndexType>>,
      * @param exec  Executor associated to the LinOp
      */
     Permutation(std::shared_ptr<const Executor> exec)
-        : Permutation(std::move(exec), dim<2>{})
+        : Permutation(std::move(exec), dim<2>{}, row_permute)
```
tcojean: This change should be unneeded since you have the default properly set in the following ctor, but that's fine.

include/ginkgo/core/matrix/permutation.hpp
```diff
@@ -114,6 +114,23 @@ class Permutation : public EnableLinOp<Permutation<IndexType>>,
         return permutation_.get_num_elems();
     }
 
+    /**
+     * Get the permute masks
+     *
+     * @param permute_mask the permute masks
```
tcojean: This is `@return`

include/ginkgo/core/matrix/permutation.hpp
```diff
@@ -114,6 +114,23 @@ class Permutation : public EnableLinOp<Permutation<IndexType>>,
...
+     */
+    mask_type get_permute_mask() { return enabled_permute_; }
+
+    /**
+     * Set the permute masks
+     *
+     * @param permute_mask the permute masks
+     */
+    void set_permute_mask(mask_type permute_mask)
```
tcojean: You could actually also make this method const and change `enabled_permute_` to be mutable. Then you are sure this method will never touch the data (you have a core test which checks that, which is good).
That's up to you.

upsj:
Could you also add kernel invocations and tests for HIP?

include/ginkgo/core/matrix/permutation.hpp
thoasm: I would actually prefer if it stayed the way it is right now since `enabled_permute_` can change the behavior of this class drastically (see the `apply_impl`). If you could change that while it is const (e.g. in a `std::shared_ptr<const Permutation>`), you were allowed to change it anywhere which can lead to unwanted/wrong results without an easy way to track down what went wrong.

tcojean: Right, that makes sense.

pratikvn:
@thoasm 
> on a side note: Why did you skip the 2 (0b0010) in your mask options? (You have row_permute = 1, column_permute = 4 and inverse_permute = 8)

Initially, I had both `inverse_row_permute` and `inverse_column_permute` and then combined those into `inverse_permute`. I think I forgot to change the values.

PR #470, opened 2020-03-03
An OSX build with Github actions

pratikvn:
This PR adds an OSX build with github-actions. You can access it either by clicking on the Actions tab on top, or at the bottom, where it is run.

It runs our basic configure step without CUDA and HIP on the Apple Clang compiler.

We can additionally add a badge to say that we compile and test on Apple as well. 

Also modifies our OpenMP linking to use the modern CMake Targets.

### DIFF
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -3,13 +3,18 @@ name: OSX-build
 on: [push]
 
 jobs:
-  osx-clang-shared:
-
-    name: OSX build shared libs
+  osx-clang-omp:
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+        - {shared: "ON", build_type: "Debug", name: "omp/debug/shared"}
+        - {shared: "OFF", build_type: "Release", name: "omp/release/static"}
+    name: ${{ matrix.config.name }}
     runs-on: [macos-latest]
 
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
     - name: setup
       run: brew install libomp
     - name: info
@@ -20,27 +25,8 @@ jobs:
       run: |
         mkdir build
         cd build
-        cmake ..
-        make -j 10
-        ctest -j 10
-
-  osx-clang-static:
-
-    name: OSX build static libs
-    runs-on: [macos-latest]
-
-    steps:
-    - uses: actions/checkout@v1
-    - name: setup
-      run: brew install libomp
-    - name: info
-      run: |
-        g++ -v
-        cmake  --version
-    - name: configure
-      run: |
-        mkdir build
-        cd build
-        cmake -DBUILD_SHARED_LIBS=off ..
-        make -j 10
-        ctest -j 10
+        cmake .. -DBUILD_SHARED_LIBS=${{ matrix.config.shared }} -DCMAKE_BUILD_TYPE=${{ matrix.config.build_type }}
+        make -j8 
+        ctest -j10
+        make install
+        make test_install
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -1,0 +1,46 @@
+name: OSX-build
+
+on: [push]
+
+jobs:
+  osx-clang-shared:
+
+    name: OSX build shared libs
+    runs-on: [macos-latest]
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: setup
+      run: brew install libomp
+    - name: info
+      run: |
+        g++ -v
+        cmake --version
+    - name: configure
+      run: |
+        mkdir build
+        cd build
+        cmake ..
+        make -j 10
+        ctest -j 10
+
+  osx-clang-static:
+
+    name: OSX build static libs
+    runs-on: [macos-latest]
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: setup
+      run: brew install libomp
+    - name: info
+      run: |
+        g++ -v
+        cmake  --version
+    - name: configure
+      run: |
+        mkdir build
+        cd build
+        cmake -DBUILD_SHARED_LIBS=off ..
+        make -j 10
+        ctest -j 10
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,8 @@
-cmake_minimum_required(VERSION 3.9)
+if(APPLE)
+  cmake_minimum_required(VERSION 3.12)
+else()
+  cmake_minimum_required(VERSION 3.9)
+endif()
 
 project(Ginkgo LANGUAGES C CXX VERSION 1.1.1 DESCRIPTION "A numerical linear algebra library targeting many-core architectures")
 set(Ginkgo_VERSION_TAG "develop")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,8 +1,4 @@
-if(APPLE)
-  cmake_minimum_required(VERSION 3.12)
-else()
-  cmake_minimum_required(VERSION 3.9)
-endif()
+cmake_minimum_required(VERSION 3.9)
 
 project(Ginkgo LANGUAGES C CXX VERSION 1.1.1 DESCRIPTION "A numerical linear algebra library targeting many-core architectures")
 set(Ginkgo_VERSION_TAG "develop")
--- a/omp/CMakeLists.txt
+++ b/omp/CMakeLists.txt
@@ -27,8 +27,13 @@ target_sources(ginkgo_omp
         stop/residual_norm_reduction_kernels.cpp)
 
 ginkgo_compile_features(ginkgo_omp)
-target_link_libraries(ginkgo_omp PRIVATE OpenMP::OpenMP_CXX)
-target_compile_options(ginkgo_omp PRIVATE "${OpenMP_CXX_FLAGS}")
+
+target_link_libraries(ginkgo_omp PRIVATE "${OpenMP_CXX_LIBRARIES}")
+target_include_directories(ginkgo_omp PRIVATE "${OpenMP_CXX_INCLUDE_DIRS}")
+# We first separate the arguments, otherwise, the target_compile_options adds it as a string
+# and the compiler is unhappy with the quotation marks.
+separate_arguments(OpenMP_SEP_FLAGS NATIVE_COMMAND "${OpenMP_CXX_FLAGS}")
+target_compile_options(ginkgo_omp PRIVATE "${OpenMP_SEP_FLAGS}")
 target_compile_options(ginkgo_omp PRIVATE "${GINKGO_COMPILER_FLAGS}")
 
 # Need to link against ginkgo_cuda for the `raw_copy_to(CudaExecutor ...)` method
--- a/omp/CMakeLists.txt
+++ b/omp/CMakeLists.txt
@@ -27,7 +27,7 @@ target_sources(ginkgo_omp
         stop/residual_norm_reduction_kernels.cpp)
 
 ginkgo_compile_features(ginkgo_omp)
-target_link_libraries(ginkgo_omp PRIVATE "${OpenMP_CXX_LIBRARIES}")
+target_link_libraries(ginkgo_omp PRIVATE OpenMP::OpenMP_CXX)
 target_compile_options(ginkgo_omp PRIVATE "${OpenMP_CXX_FLAGS}")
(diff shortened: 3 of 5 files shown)


.github/workflows/main.yml
```diff
@@ -0,0 +1,23 @@
...
+    - uses: actions/checkout@v1
+    - name: info
+      run: |
+        g++ -v
+        cmake --version
+    - name: configure
+      run: |
+        mkdir build
+        cd build
+        cmake ..
```
upsj: AppleClang by default doesnt't support OpenMP, but with small modifications to the `CMake` invocations and installing the correct libraries, you can usually make it work:
https://gitlab.kitware.com/cmake/cmake/issues/18520

Do we have any influence over the build environment?

pratikvn: Thanks, i will see if I can add this.

.github/workflows/main.yml
```diff
@@ -0,0 +1,25 @@
...
+        make -j 10
+        make -j 10 test
```
upsj: Does CMake automatically use 10 threads for the tests here? I think make test is single-threaded. How about something like 
```suggestion
        ctest -j 10
```

.github/workflows/main.yml
upsj: A first step would be to see if Homebrew is installed and you can do `brew install libomp`

omp/CMakeLists.txt
```diff
@@ -27,8 +27,7 @@ target_sources(ginkgo_omp
         stop/residual_norm_reduction_kernels.cpp)
 
 ginkgo_compile_features(ginkgo_omp)
-target_link_libraries(ginkgo_omp PRIVATE "${OpenMP_CXX_LIBRARIES}")
-target_compile_options(ginkgo_omp PRIVATE "${OpenMP_CXX_FLAGS}")
+target_link_libraries(ginkgo_omp PRIVATE OpenMP::OpenMP_CXX)
```
upsj: Since we are already cleaning this up, might make sense to get rid of the other uses of these variables (below and in GinkgoConfig.cmake.in). @tcojean can we safely remove this from the installation config?

tcojean: As far as I know, no. I actually want this to stay the same. There were multiple bugs previously which made the `OpenMP::OpenMP_whatever` targets provide no flag on some compilers, which is why I went with this verison.

pratikvn: I will keep the flags. But instead of the `${OpenMP_CXX_LIBRARIES}`, I use the targets. It seems to work on the Gitlab CI as well.

pratikvn:
@yhmtsai , thanks. I have added the shared and static jobs. I am not so sure about CUDA. Does it make sense to compile if we know that there is no GPU to compile anyway ? We do test quite intensively the CUDA compilation on the fine-ci. Here it would just add more time for the job.

yhmtsai:
I think it still gives the information about whether we can compile cuda in OSX.
OSX also uses gcc, so maybe we do not need to check it.

upsj:
@yhmtsai Actually, OS X uses their own clone of clang mostly called AppleClang, but they add an alias `gcc=clang` for some reason.

.github/workflows/osx.yml
```diff
@@ -0,0 +1,50 @@
...
+on: [push]
+
+jobs:
+  osx-clang-shared:
+
+    name: OSX build shared libs
+    runs-on: [macos-latest]
+
+    steps:
+    - uses: actions/checkout@v1
```
yhmtsai:
```suggestion
    - uses: actions/checkout@v2
```

checkout@v2 which only pulls the current branch is faster than checkout@v1

.github/workflows/osx.yml
```diff
@@ -0,0 +1,50 @@
...
+        make install
+        make test_install
+
+  osx-clang-static:
+
+    name: OSX build static libs
+    runs-on: [macos-latest]
+
+    steps:
+    - uses: actions/checkout@v1
```
yhmtsai:
```suggestion
    - uses: actions/checkout@v2
```

.github/workflows/osx.yml
```diff
@@ -0,0 +1,32 @@
...
+      run: |
+        g++ -v
+        cmake --version
+    - name: configure
+      run: |
+        mkdir build
+        cd build
+        cmake .. -DBUILD_SHARED_LIBS=${{ matrix.config.shared }}
+        cmake --build . -j8 --config ${{ matrix.config.build_type }}
+        ctest . -C ${{ matrix.config.build_type }}
```
yhmtsai: cmake --build . -j8 --config build_type should work for mult-configuration toolset like MSVC or XCode. If you use gcc/clang, using the previous compile step sets correct build type.
I think clang from XCode is the type of clang not XCode
from `2020-03-04T12:25:58.9740980Z -- Setting build type to 'Release' as none was specified.` in [omp/debug/shared log](https://pipelines.actions.githubusercontent.com/dR6M3a2IKUyMCOpbBJ1SfpHlfxqOoIda0x3iCmpxIvMt04aaNr/_apis/pipelines/1/runs/35/signedlogcontent/3?urlExpires=2020-03-04T13%3A00%3A14.8711618Z&urlSigningMethod=HMACV1&urlSignature=DES9JRSAK8vmpn2EajqyLA18SWAhl1Ge9B6DnbkszxM%3D)

test_install/CMakeLists.txt
```diff
@@ -34,6 +34,10 @@ if((NOT GINKGO_BUILD_SHARED_LIBS) AND GINKGO_BUILD_HIP)
     endif()
 endif()
 
+if(GINKGO_BUILD_OMP)
+  find_package(OpenMP REQUIRED)
+endif()
```
tcojean: Do you really need to use the `OpenMP::OpenMP_CXX` target? I would rather we don't have an extra public OpenMP dependency in this way.

pratikvn: I could make [this linking](https://github.com/ginkgo-project/ginkgo/blob/083a5dee7a81aea745d02c9d58f81a6dec373f5b/omp/CMakeLists.txt#L30) `PUBLIC`, but I guess that is also a public dependency ?

tcojean: Well yes, that would make it spill over everywhere just like now. What I don't get is, is `OpenMP_CXX_LIBRARIES` wrong on MacOS? Or empty? What is the issue? I believe it should be available everywhere since it's a CMake standard from FindOpenMP.

upsj: One minor point: We don't set the OpenMP include paths at the moment

tcojean: In general with OpenMP that should not be needed, because OpenMP is integrated into the compilers and they take care of it. The OpenMP header usually sits in the compiler's private directory (you can activate it with flags, such as `-fopenmp` or whichever is your compiler's flag).

upsj: AppleClang is a bit... difficult in that regard ;) It also seems like the detected flag is incorrect: `clang: warning: argument unused during compilation: '-Xclang -fopenmp' [-Wunused-command-line-argument]`
[This article](https://iscinumpy.gitlab.io/post/omp-on-high-sierra/) suggests we should actually be using `-Xpreprocessor -fopenmp`

pratikvn: @upsj, looks like that minor point was the issue. It looks like it works now. :+1:

tcojean: Good find @upsj. I guess FindOpenMP's flags are for AppleClang then. They also don't seem to want to fix it:
https://github.com/Kitware/CMake/blob/b0e34a3160642a09728f1f56d93e2835443eb016/Modules/FindOpenMP.cmake#L100

upsj: I definitely remember `-Xclang -fopenmp` working with older versions of XCode/AppleClang, but this is something that should be solved upstream in CMake.

omp/CMakeLists.txt
```diff
@@ -27,7 +27,8 @@ target_sources(ginkgo_omp
         stop/residual_norm_reduction_kernels.cpp)
 
 ginkgo_compile_features(ginkgo_omp)
-target_link_libraries(ginkgo_omp PRIVATE "${OpenMP_CXX_LIBRARIES}")
+target_link_libraries(ginkgo_omp PRIVATE ${OpenMP_CXX_LIBRARIES})
+target_include_directories(ginkgo_omp PRIVATE ${OpenMP_CXX_INCLUDE_DIRS})
```
upsj:
```suggestion
target_link_libraries(ginkgo_omp PRIVATE "${OpenMP_CXX_LIBRARIES}")
target_include_directories(ginkgo_omp PRIVATE "${OpenMP_CXX_INCLUDE_DIRS}")
```

omp/CMakeLists.txt
```diff
@@ -28,7 +28,12 @@ target_sources(ginkgo_omp
 
 ginkgo_compile_features(ginkgo_omp)
 target_link_libraries(ginkgo_omp PRIVATE "${OpenMP_CXX_LIBRARIES}")
-target_compile_options(ginkgo_omp PRIVATE "${OpenMP_CXX_FLAGS}")
+target_include_directories(ginkgo_omp PRIVATE "${OpenMP_CXX_INCLUDE_DIRS}")
+if(APPLE)
```
tcojean: Actually, if you are going to do this, you should do it for AppleClang only. See: https://cmake.org/cmake/help/v3.0/variable/CMAKE_LANG_COMPILER_ID.html#variable:CMAKE_%3CLANG%3E_COMPILER_ID

```suggestion
if(CMAKE_CXX_COMPILER_ID MATCHES "AppleClang")
```

CMakeLists.txt
```diff
@@ -1,4 +1,8 @@
-cmake_minimum_required(VERSION 3.9)
+if(APPLE)
+  cmake_minimum_required(VERSION 3.12)
```
thoasm: Is there a reason why we need to have a later version for all Apple devices?
Or is it only for AppleClang?

.github/workflows/osx.yml
```diff
@@ -0,0 +1,32 @@
...
+        config:
+        - {shared: "ON", build_type: "Debug", name: "omp/debug/shared"}
+        - {shared: "OFF", build_type: "Release", name: "omp/release/static"}
+    name: ${{ matrix.config.name }}
+    runs-on: [macos-latest]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: setup
+      run: brew install libomp
```
thoasm: Does AppleClang really ship without OpenMP?

.github/workflows/osx.yml
```diff
@@ -0,0 +1,32 @@
...
+    - name: info
+      run: |
+        g++ -v
```
thoasm: Will `g++` actually be used for the compilation? I thought we are testing `AppleClang`.

tcojean: The default g++ on mac is actually AppleClang, which creates problems to a lot of unknowing mac users at first (thinking they have are using g++, but they are not by default).

.github/workflows/osx.yml
pratikvn: Yes, without this the server does not find OpenMP.

CMakeLists.txt
pratikvn: I needed it for linking to targets for AppleClang. I dont need it anymore, so I have removed it.

omp/CMakeLists.txt
```diff
@@ -27,8 +27,13 @@ target_sources(ginkgo_omp
         stop/residual_norm_reduction_kernels.cpp)
 
 ginkgo_compile_features(ginkgo_omp)
+
 target_link_libraries(ginkgo_omp PRIVATE "${OpenMP_CXX_LIBRARIES}")
-target_compile_options(ginkgo_omp PRIVATE "${OpenMP_CXX_FLAGS}")
+target_include_directories(ginkgo_omp PRIVATE "${OpenMP_CXX_INCLUDE_DIRS}")
+# We first separate the arguments, otherwise, the target_compile_options adds it as a string
+# and the compiler is unhappy with the quotation marks.
+separate_arguments(OpenMP_SEP_FLAGS NATIVE_COMMAND "${OpenMP_CXX_FLAGS}")
```
upsj: Nice solution! And the OpenMP target would not set these flags correctly?

pratikvn: It was, but there was a problem, when you would do the test_install step.

PR #471, opened 2020-03-03
add windows ci setting

yhmtsai:
~~This PR add the pipeline for windows gitlab shared runner.~~
This PR add the github action for windows
For cuda, it only test compilation. For nocuda, it test compilation and google test.

I only keep gitlab ci in commit.
gitlab ci does not give separate badge to show windows compiliation and gives 2000 min/month in our current gitlab plan.
Github action gives badge based on workflow and give unlimited min for open source.
However, it only gives 14GB for storage, so compiling windows with static and debug is failed.

github action strategy allows to evaluate all combination of matrix.
For example,
```
build_type: ["Debug", "Release"]
shared: ["ON", "OFF"]
```
gives four jobs.

I only specify only two config to reduce the list of pull request check.

~~Gitlab seems to have whole pipeline badge.
In #470 , Github action can give badges on different workflow.
Thus, I will try to see whether it also works on github action. 
By doing so, we can separate the status from different platform~~

Summary:
- MSVC: cuda (only compile), ref (build/test/install)
- MinGW: omp (build/test/install)
- Cygwin: omp (build/test/install)
- split install and compile in OSX-build

Note:
1. I do not find a way to use cygwin as a shell in github action, so I need to set cygwin path only to make everything work.
2. MSVC/mingw install Ginkgo into `C:\Program files(x86)\Ginkgo` not `C:\Program files\Ginkgo`.
I can not use `CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT` to set prefix force, which leads that user can only set `CMAKE_INSTALL_PREFIX` in the beginning. reference: https://gitlab.kitware.com/cmake/cmake/issues/18312
3.  MSVC nvcc has flags like `-Xcompiler="-O2 ..."`, so I add a macro to modify cuda_flag for GinkgoConfig.pc by `"` -> `\"`
4. Github action create a script containing the commands in the `run:` and use shell to run the script, so github action only catch the last error code. For example, in [workflow](https://github.com/ginkgo-project/ginkgo/runs/493420377?check_suite_focus=true), the parilu test is failed in configure step, but the last command `make install` is successful. Thus, the configure is successful.

TODO:
- [x] Merge #473 
- [x] MSVC env (for CUDA)
- [x] MinGW/CYGWIN env (for OpenMP)
- [x] MinGW/CYGWIN test_install
- [x] MSVC test_install
- [x] Ginkgo.pc warning

PR #473, opened 2020-03-04
Fix windows cuda generic

yhmtsai:
This PR add the windows condition on enabling cuda generic api

In the PR #468 , adding the cuda generic api.
However, cuda generic api does not support windows https://docs.nvidia.com/cuda/cusparse/index.html#cusparse-generic-api-reference in cuda 10.2
It lead the failed compilation in windows which is caught by [gitlab-ci](https://gitlab.com/ginkgo-project/ginkgo-public-ci/pipelines/123110689) or [github action](https://github.com/ginkgo-project/ginkgo/actions/runs/49335880) in #471 

Moreover, I add the newline in the final output of benchmark.
When benchmark writes the result to standard output, it gives `]` in the head of command line, which is sometimes annoying on some terminal.
```
]mike@workstation:~$ command||
```
when moving the cursor to left, it re-render the command without `]` in the head
like
```
]mike@workstation:~$ comm||nd
```
Move all letters 1 left.

### DIFF
--- a/benchmark/conversions/conversions.cpp
+++ b/benchmark/conversions/conversions.cpp
@@ -195,5 +195,5 @@ int main(int argc, char *argv[])
         }
     }
 
-    std::cout << test_cases;
+    std::cout << test_cases << std::endl;
 }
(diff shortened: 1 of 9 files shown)


yhmtsai:
It is verified by [github action](https://github.com/ginkgo-project/ginkgo/actions/runs/49383202)

PR #475, opened 2020-03-05
Add a private_ci tag for all our jobs.

tcojean:
Ensure all our jobs run on private CI with shared runners enabled. This is simply done with the addition of a `private_ci` tag for all the CI jobs which should run on our own machines.

This is to ensure we never run into an issue like we had recently after shared runners were enabled to test windows build:
https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/460277468

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,5 +1,3 @@
-image: localhost:5000/gko-cuda100-gnu7-llvm60
-
 stages:
   - sync
   - build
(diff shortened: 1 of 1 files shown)


yhmtsai:
There are two `ensure_private_ci` in the config. Are they only for checking this pr behavior?

tcojean:
Yes I wanted to make sure there would be no surprise. In particular since I now use another container which doesn't have the CUDA dependency for the sync and similar jobs. That way, we will get way faster syncs than now. From what the CI already did, we can see that it's working properly so I will remove these lines with a rebase once the full testsuite has completly ran.

pratikvn:
@tcojean, Could you also add a badge to show the test coverage status as we had before ?

yhmtsai:
@tcojean Got it. Thanks

PR #480, opened 2020-03-12
Using next_krylov_basis instead of krylov_bases to save unnecessary p…

fritzgoebel:
…adding in the preconditioned vector. This made it necessary to choose a different spmv strategy when using plain GMRES or with ilu preconditioner, since the default choice doesn't support multiple vectors.


fritzgoebel:
> LGTM! (assuming the tests succeed ;) ) Only minor comments. Can you imagine this causing any performance regressions due to the duplicated storage of the last Krylov vector?

No, the vector was already used before, just not initialized in the very beginning. I actually think performance will increase since this extra initialization saves n*restart padded entries in preconditioned_vector.

fritzgoebel:
Moved to PR 481

PR #485, opened 2020-03-23
Fix the view memory leak and array/view copy/move.

tcojean:
This fixes some of the discussed issues with arrays. The changes are noted in the list below. Some of these points should still be discussed.

+ Add tests for copy/move of all combinations of array/view.
+ Add a new exception helper macro for testing array dimensions.
+ When copying something into a view, do bound checking and ensure what is
  copied has a compatible amount of data.
+ When copying into an array, the proper amount of data is allocated and copied
   over.
+ When moving a into b, b becomes a and a is invalid afterwards.
+ Resize and reset for views throws an error.

### DIFF
--- a/core/test/base/array.cpp
+++ b/core/test/base/array.cpp
@@ -284,6 +284,17 @@ TYPED_TEST(Array, CanBeResized)
 }
 
 
+TYPED_TEST(Array, VewCanBeResetNotResized)
+{
+    TypeParam data[] = {1, 2, 3};
+    auto view = gko::Array<TypeParam>::view(this->exec, 3, data);
+    view.resize_and_reset(1);
+
+    EXPECT_EQ(view.get_const_data(), nullptr);
+    ASSERT_EQ(view.get_num_elems(), 1);
+}
+
+
 TYPED_TEST(Array, CanBeAssignedAnExecutor)
 {
     gko::Array<TypeParam> a;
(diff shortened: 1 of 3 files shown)


core/test/base/array.cpp
```diff
@@ -284,6 +284,17 @@ TYPED_TEST(Array, CanBeResized)
 }
 
 
+TYPED_TEST(Array, VewCanBeResetNotResized)
```
upsj:
```suggestion
TYPED_TEST(Array, ViewCanBeResetNotResized)
```

core/test/base/array.cpp
```diff
@@ -304,16 +315,168 @@ TYPED_TEST(Array, ChangesExecutors)
...
+    gko::Array<TypeParam> array2(this->exec, {5, 4, 2, 1});
+
+    array = array2;
+
+    EXPECT_EQ(array.get_data()[0], TypeParam{5});
+    EXPECT_EQ(array.get_data()[1], TypeParam{4});
+    EXPECT_EQ(array.get_data()[2], TypeParam{2});
+    EXPECT_EQ(array.get_data()[3], TypeParam{1});
+    EXPECT_EQ(array.get_num_elems(), 4);
+    ASSERT_EQ(array2.get_num_elems(), 4);
```
upsj: Maybe assert that the data pointers are not equal?

core/test/base/array.cpp
```diff
@@ -304,16 +315,168 @@ TYPED_TEST(Array, ChangesExecutors)
...
+
+    view = view2;
+    view2.get_data()[0] = 2;
 
     EXPECT_EQ(data[0], TypeParam{5});
     EXPECT_EQ(data[1], TypeParam{4});
     EXPECT_EQ(data[2], TypeParam{2});
+    EXPECT_EQ(view.get_num_elems(), 3);
+    EXPECT_EQ(view2.get_num_elems(), 3);
+    ASSERT_THROW(view2 = view_size4, gko::OutOfBoundsError);
```
upsj: Maybe assert that the data pointers have not been modified?

core/test/base/array.cpp
```diff
@@ -304,16 +315,168 @@ TYPED_TEST(Array, ChangesExecutors)
...
+    gko::Array<TypeParam> array_size4(this->exec, {5, 4, 2, 1});
+
+    view = std::move(array_size2);
+
+    EXPECT_EQ(view.get_data()[0], TypeParam{5});
+    EXPECT_EQ(view.get_data()[1], TypeParam{4});
+    EXPECT_EQ(view.get_num_elems(), 2);
+    EXPECT_NO_THROW(view = array_size4);
+    EXPECT_EQ(array_size2.get_data(), nullptr);
+    ASSERT_EQ(array_size2.get_num_elems(), 0);
```
upsj: Maybe check that `view` has the correct data pointer?

core/test/base/array.cpp
```diff
@@ -284,6 +284,17 @@ TYPED_TEST(Array, CanBeResized)
...
 
 
+TYPED_TEST(Array, VewCanBeResetNotResized)
+{
+    TypeParam data[] = {1, 2, 3};
+    auto view = gko::Array<TypeParam>::view(this->exec, 3, data);
+    view.resize_and_reset(1);
+
+    EXPECT_EQ(view.get_const_data(), nullptr);
+    ASSERT_EQ(view.get_num_elems(), 1);
```
upsj: nullptr strongly suggests that the view is empty, should it have non-zero size then?

core/test/base/array.cpp
```diff
@@ -304,16 +315,168 @@ TYPED_TEST(Array, ChangesExecutors)
...
+
+
+TYPED_TEST(Array, MoveViewToView)
+{
+    TypeParam data[] = {1, 2, 3, 4};
+    auto view = gko::Array<TypeParam>::view(this->exec, 4, data);
+    TypeParam data2[] = {5, 4, 2};
+    auto view2 = gko::Array<TypeParam>::view(this->exec, 3, data2);
+
+    view = std::move(view2);
```
pratikvn: I think we shoud also make sure that `data` still exists with a test such as `EXPECT_NE(data, nullptr)` or something like that and maybe that it does contain the values `{1,2,3,4}`

core/test/base/array.cpp
```diff
@@ -304,16 +315,168 @@ TYPED_TEST(Array, ChangesExecutors)
...
+
+
+TYPED_TEST(Array, MoveArrayToView)
+{
+    TypeParam data[] = {1, 2, 3};
+    auto view = gko::Array<TypeParam>::view(this->exec, 3, data);
+    gko::Array<TypeParam> array_size2(this->exec, {5, 4});
+    gko::Array<TypeParam> array_size4(this->exec, {5, 4, 2, 1});
+
+    view = std::move(array_size2);
```
pratikvn: Maybe check the type of `view` ?

include/ginkgo/core/base/array.hpp
```diff
@@ -294,14 +298,36 @@ class Array {
...
+     * means that in the following situation:
+     * ```cpp
+     *   gko::Array<int> a; // an existing array or view
+     *   gko::Array<int> b; // an existing array or view
+     *   b = std::move(a);
+     * ```
+     * Depending on whether `a` and `b` are array or view, this happens:
+     * + `a` and `b` are view, `b` becomes the only valid view of `a`
+     * + `a` and `b` are array, `b` becomes the only valid array of `a`
+     * + `a` is `a` view and `b` is an array, `b` frees its data and becomes the
```
pratikvn:
```suggestion
     * + `a` is a view and `b` is an array, `b` frees its data and becomes the
```

include/ginkgo/core/base/array.hpp
```diff
@@ -294,14 +298,36 @@ class Array {
...
+     *    only valid view of `a`
+     * + `a` is an array and `b` is `a` view, `b` becomes the only valid array
```
pratikvn:
```suggestion
     * + `a` is an array and `b` is a view, `b` becomes the only valid array
```

core/test/base/array.cpp
tcojean: Well yes, that depends if we do something or not with this size argument to the reset as you state earlie. For now I left it as is, but indeed I was considering always setting to 0 in this case. I guess another possibility is indeed to make this view into an array of 1 element. I'd like to have some feedback on what is the most useful here in fact.

core/test/base/array.cpp
tcojean: You would need to access the `data_` internal storage to check its deleter type, which we cannot with our interface and I'm not sure we want this to happen. So this is a bit tricky.

pratikvn:
Regarding when copying into a view, bounds checking and a throw looks reasonable to me. 

The `resize_and_reset` gets called when moving from an view, right ? In that case, intuitively, to me, setting the pointer to `nullptr` and not allocating makes sense. I am not sure if it makes sense to decide based on whether something makes implementing our algorithms easier because if the function does something un-intuitive, then using and debugging it becomes a lot harder.

core/test/base/array.cpp
pratikvn: Is is not possible to create a dummy array class in the test file and test locally ?

core/test/base/array.cpp
tcojean: So far, I decided to stick with the current model, but I'm making it clear that whenever the data is just a `nullptr`, then `num_elems_` is always 0. I think if you want to change your view to an array, there is always the move operation to do that now so there's no need to add this functionality. It also helps keeping the border between the two clear: it's crossed only thanks to the move operator=, and I think that's easy enough to understand and reason with.

upsj: An example where the `nullptr` behavior could cause issues:
If we have a CSR matrix that is built on top of `Array::view` col_idxs/values, using `apply` to compute the SpGEMM product with this matrix as output would segfault/fail: `csr_mtx->apply(csr_mtx2, csr_view_mtx)`, as the SpGEMM internally has to resize the arrays.

include/ginkgo/core/base/array.hpp
```diff
@@ -294,14 +298,36 @@ class Array {
             this->resize_and_reset(0);
             return *this;
         }
-        this->resize_and_reset(other.get_num_elems());
-        exec_->copy_from(other.get_executor().get(), num_elems_,
+        // We allow resizing only if we are an array, we cannot resize a view!
+        // If we are a view, we do bound checking to ensure the data can be
+        // copied over.
+        if (data_.get_deleter().target_type() != typeid(view_deleter)) {
```
thoasm: Can you export this in a helper function? We do that at least twice, and it is not a trivial operation. A helper function would also explain what is done here. Also, if we find a better way to do it, we only have to change a single function.

The problem I see is that we allow the user to use a default deleter with a data pointer (Argument `DeleterType deleter`). That could potentially be a self-written view, and almost impossible (currently, I don't know how) to detect.
We can, however, do the opposite: check if the deleter is the `default_deleter` (which we would treat normally) and if not, treat it as a view (prevent allocations / `resize_and_reset`).

In case we change it, we might have to rewrite the documentation: instead of writing `a view`, we should probably write `a view or an array with a custom deleter` to document that behavior.
If we keep the current handling (checking against `view_deleter`), we might have to document that using a custom deleter is UB when doing assignments (when on the left side of assignments).

include/ginkgo/core/base/array.hpp
```diff
@@ -294,14 +298,36 @@ class Array {
...
+     *   gko::Array<int> b; // an existing array or view
+     *   b = std::move(a);
+     * ```
+     * Depending on whether `a` and `b` are array or view, this happens:
+     * + `a` and `b` are view, `b` becomes the only valid view of `a`
+     * + `a` and `b` are array, `b` becomes the only valid array of `a`
+     * + `a` is a view and `b` is an array, `b` frees its data and becomes the
+     *    only valid view of `a`
+     * + `a` is an array and `b` is a view, `b` becomes the only valid array
+     *    of `a`
```
thoasm: I think it would be nice to mention that the pointer of `a` will be a `nullptr` (or, if we want to be vague and closer to the standard description: `a is in a valid but unspecified state afterwards`) after the move is done (in every case, right?).
I know that you try to say that with the "only valid [view, array] of `a`", but I expect that others might not understand the implications of that.

include/ginkgo/core/base/array.hpp
```diff
@@ -294,14 +298,36 @@ class Array {
...
+     * Moves data from another array or view. Only the pointer and deleter type
+     * change, a copy only happens when targeting another executor's data. This
+     * means that in the following situation:
+     * ```cpp
+     *   gko::Array<int> a; // an existing array or view
+     *   gko::Array<int> b; // an existing array or view
+     *   b = std::move(a);
+     * ```
+     * Depending on whether `a` and `b` are array or view, this happens:
+     * + `a` and `b` are view, `b` becomes the only valid view of `a`
```
thoasm: nit: use plural of `view`
```suggestion
     * + `a` and `b` are views, `b` becomes the only valid view of `a`
```

include/ginkgo/core/base/array.hpp
```diff
@@ -294,14 +298,36 @@ class Array {
...
+     * + `a` and `b` are array, `b` becomes the only valid array of `a`
```
thoasm: nit: use plural of `array`
```suggestion
     * + `a` and `b` are arrays, `b` becomes the only valid array of `a`
```

include/ginkgo/core/base/array.hpp
```diff
@@ -326,14 +352,14 @@ class Array {
...
         }
-        if (exec_ == other.get_executor() &&
-            data_.get_deleter().target_type() != typeid(view_deleter)) {
-            // same device and not a view, only move the pointer
+        if (exec_ == other.get_executor()) {
+            // same device, only move the pointer
             using std::swap;
             swap(data_, other.data_);
             swap(num_elems_, other.num_elems_);
+            other.resize_and_reset(0);
```
thoasm: nit: could you actually use the `clear()` function here? For some reason, we use `resize_and_reset(0)` in a lot of places instead of `clear()`, even though `clear()` has less side effects, is more light weight and more descriptive.

include/ginkgo/core/base/array.hpp
```diff
@@ -354,6 +380,8 @@ class Array {
 
     /**
      * Resizes the array so it is able to hold the specified number of elements.
+     * To a view, this only resets the pointer. Afterwards, the  only way to use
```
thoasm: nit: a space too much
Also, I would clarify it a bit more:
```suggestion
     * In a view, this resets the pointer, so it does not point to anything
     * (num_elems will also be set to 0).
```

include/ginkgo/core/base/array.hpp
```diff
@@ -354,6 +380,8 @@ class Array {
...
+     * a view in such a state is to `std::move` another view.
```
thoasm: You could also move another array into it. I would write something like:
`After a reset of a view, the view can be moved to in order to use it afterwards.`

include/ginkgo/core/base/array.hpp
```diff
@@ -371,10 +399,15 @@ class Array {
...
+
+        // For views, we should never allocate any data, all we can do is unset
+        // the pointer.
+        if (num_elems > 0 &&
+            data_.get_deleter().target_type() != typeid(view_deleter)) {
+            num_elems_ = num_elems;
             data_.reset(exec_->alloc<value_type>(num_elems));
         } else {
+            num_elems_ = 0;
             data_.reset(nullptr);
```
thoasm: nit: We could also use the `clear()` function here instead.

core/test/base/array.cpp
```diff
@@ -284,6 +284,17 @@ TYPED_TEST(Array, CanBeResized)
 }
 
 
+TYPED_TEST(Array, ViewCanBeResetNotResized)
+{
+    TypeParam data[] = {1, 2, 3};
+    auto view = gko::Array<TypeParam>::view(this->exec, 3, data);
+    view.resize_and_reset(1);
```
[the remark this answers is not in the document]
tcojean: Yes it becomes an array, because we do the following:
```cpp
Array tmp(std::move(exec));
tmp = *this;
```
The first line creates an **array** (with `default_deleter`) in the other executor. Then we copy the data to it which ensures that both objects keep their nature (`tmp` will stay an array after this copy). Thanks for this question actually I hadn't checked to make sure everything is proper!

About the `resize_and_reset`, I am still unsure what we should do with it. Maybe we should discuss it in today's call. Tobias at some point wanted it to change the nature of the object (view would become an array transparently), you want to disable it which makes sense, I am also fine with the current `clear()` version (the question is whether resizing an array or not is an error here, it can be considered both an error or not one depending on POV). I don't have a strong opinion on this, but picking something is tough because there are many valid options.

thoasm:
@tcojean Can you add a test with a custom deleter and copying of views with mismatching sizes?
That way, we document that this is a potential problem we need to address.
```c++
ASSERT_THROW(view1_custom_del = view2, gko::NotSupported);
```

tcojean:
Copying view test with mismatched size is already there I think
```c++
ASSERT_THROW(view2 = view_size4, gko::OutOfBoundsError);
```

tcojean:
Anyone has any idea why I get this weird Windows compilation issue? I don't think I changed anything relevant...?

```c++
D:\a\ginkgo\ginkgo\core/base/iterator_factory.hpp(297,2): error C2398: Element '2': conversion from 'size_t' to 'gko::detail::IteratorFactory<IndexType,ValueType>::Reference::array_index_type' requires a narrowing conversion [D:\a\ginkgo\ginkgo\build\reference\ginkgo_reference.vcxproj]
          with
          [
              IndexType=gko::int64,
              ValueType=std::complex<double>
          ]
D:\a\ginkgo\ginkgo\core/base/iterator_factory.hpp(296): message : while compiling class template member function 'gko::detail::IteratorFactory<IndexType,ValueType>::Reference gko::detail::IteratorFactory<IndexType,ValueType>::Iterator::operator [](size_t) const' [D:\a\ginkgo\ginkgo\build\reference\ginkgo_reference.vcxproj]
          with
          [
              IndexType=gko::int64,
              ValueType=std::complex<double>
          ]
C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.25.28610\include\algorithm(2749): message : see reference to function template instantiation 'gko::detail::IteratorFactory<IndexType,ValueType>::Reference gko::detail::IteratorFactory<IndexType,ValueType>::Iterator::operator [](size_t) const' being compiled [D:\a\ginkgo\ginkgo\build\reference\ginkgo_reference.vcxproj]
          with
          [
              IndexType=gko::int64,
              ValueType=std::complex<double>
          ]
```

upsj:
#490 simply rebase onto develop

tcojean:
Thanks... I thought I was already up to date with develop but I guess I missed this one.

core/test/base/array.cpp
```diff
@@ -304,16 +347,182 @@ TYPED_TEST(Array, ChangesExecutors)
...
+    auto size4_ptr = array_size4.get_data();
+
+    view = std::move(array_size2);
+
+    EXPECT_EQ(view.get_data()[0], TypeParam{5});
+    EXPECT_EQ(view.get_data()[1], TypeParam{4});
+    EXPECT_EQ(view.get_num_elems(), 2);
+    EXPECT_NE(view.get_data(), data);
+    EXPECT_EQ(view.get_data(), size2_ptr);
+    EXPECT_NO_THROW(view = std::move(array_size4));
```
yhmtsai, line 522: When move array to the view, the view become array. Is it correct?

tcojean: Yes.

yhmtsai: Got it. thanks

Issue #472, opened 2020-03-03
Improve ILU preconditioner interface


upsj:
Currently, `preconditioner::Ilu` has some shortcomings, as it only supports ILU(0), doesn't allow setting factorization parameters, needs multiple template parameters to ensure statically ensure equal `value_type` for its components, and at its core re-implements a two `Composition` of two `LinOp`.

The current workflow to set up a Jacobi preconditioner is very straightforward, and I would like to keep it that way for ILU(T):

```cpp
auto precond = preconditioner::Jacobi::build()
                    .with_parameter(...)
                    .on(exec)->generate(system_mtx); 
```

So I would like to propose two alternatives how we could handle factorization-based preconditioning:

1. Keep the separation into `preconditioner` and `factorization` and make `preconditioner::Ilu` more flexible or
2. Merge the functionality into a single class.

## Possible implementations:

**1a) Allow passing a factorization `LinOpFactory` to `preconditioner::Ilu` that builds a two-element `Composition`.**
This gives the user the ability to set parameters for the factorization, while still keeping the same preconditioner framework:
```cpp
auto precond = preconditioner::Ilu::build()
                    .with_factorization(
                        factorization::ParIlu::build()
                            .with_parameter(...)
                    )
                    .on(exec)->generate(system_mtx);
```
or
```cpp
auto fact_fact = factorization::ParIlu::build()
                   .with_parameter(...);
auto precond = preconditioner::Ilu::build()
                   .with_factorization(fact_fact)
                   .on(exec)->generate(system_mtx); 
```

**Advantages:** Extensible to new factorization types
**Disadvantages:** For me personally, this seems unnecessarily complicated. The only reason I can imagine for computing an ILU(T) is to use it as a preconditioner.

**1b) Duplicate all factorization parameters so they are available from the `preconditioner::Ilu` factory:**

```cpp
auto precond = preconditioner::Ilu::build()
                    .with_factorization_type(ILUT)
                    .with_parameter(...)
                    .on(exec)->generate(system_mtx); 
```

**Advantages:** In my opinion, this is closed to the interface I would want as a user
**Disadvantages:** Not extensible

**2) Deprecate `preconditioner::Ilu` and allow `factorization::ParIlu(t)` to generate a preconditioner** 

[...]

PR #400, opened 2019-11-25
Add ParILUT

upsj:
This PR adds ParILUT factorization support to Ginkgo.

Since this is a rather large PR, even though there are still a few parts missing, it might make sense to already start reviewing the kernel implementations.

TODO

- [x] Add core tests
- [x] Add benchmarks
- [x] Run benchmarks
- [x] Add more comments to complicated kernels
- [x] Add `__restrict__` modifiers where appropriate
- [x] ~~Integrate Array overallocation functionality (see [array_dynamic_functionality](https://github.com/ginkgo-project/ginkgo/tree/array_dynamic_functionality))~~
- [x] Increase number of sweep iterations to avoid spurious test failures
- [x] Add full tests with core interface
- [x] Add templated tests
- [x] Format header files (depends on #447)
- [x] Wait until the Array assignment operator gets updated (depends on #485)
- [x] Add ParICT for SPD matrices
- [x] Add ParICT tests
- [x] Add test for filter nullptr COO parameter

PR #487, opened 2020-03-24
Add sparselib ILU for benchmarks

upsj:
This PR adds a LinOp for cuSPARSE/hipSPARSE ILU factorizations and the new ILU preconditioner interface described in #472 for integrating it into the benchmarks.


benchmark/utils/cuda_linops.hpp
```diff
@@ -682,6 +684,190 @@ class CuspGenericCoo
...
+    void run(std::shared_ptr<const gko::CudaExecutor> exec) const override
+    {
+        op_(exec);
+    }
+    const char *get_name() const noexcept override { return name_.c_str(); }
+
+private:
+    std::string name_;
+    std::function<void(std::shared_ptr<const gko::CudaExecutor>)> op_;
+};
```
upsj: This might be a candidate for a more general utility, but how do we implement it for a single executor only?

benchmark/utils/hip_linops.hip.hpp
```diff
@@ -315,6 +320,191 @@ class HipspHybrid
...
+GKO_REGISTER_OPERATION(solve_u, upper_trs::solve);
+
+
+}  // namespace ilu
+}  // namespace hip
+
+
+template <typename ValueType = gko::default_precision,
+          typename IndexType = gko::int32>
+class HipspIlu
```
upsj: Should this be moved to Ginkgo itself?

benchmark/preconditioner/preconditioner.cpp
```diff
@@ -90,13 +90,33 @@ gko::precision_reduction parse_storage_optimization(const std::string &flag)
...
+        {"jacobi",
+         [](std::shared_ptr<const gko::Executor> exec) {
              return gko::preconditioner::Jacobi<etype>::build()
                  .with_max_block_size(FLAGS_max_block_size)
                  .with_storage_optimization(
                      parse_storage_optimization(FLAGS_storage_optimization))
                  .with_accuracy(FLAGS_accuracy)
                  .on(exec);
+         }},
+        {"ilu",
```
pratikvn: Maybe I am mistaken here, but isnt this actually a ParILU ? I guess this discussion should move out as I see that the class is [`named ILU`](https://github.com/ginkgo-project/ginkgo/blob/develop/include/ginkgo/core/preconditioner/ilu.hpp#L113) but the [documentation](https://github.com/ginkgo-project/ginkgo/blob/develop/include/ginkgo/core/preconditioner/ilu.hpp#L81) and [some parts](https://github.com/ginkgo-project/ginkgo/blob/develop/include/ginkgo/core/preconditioner/ilu.hpp#L112) mention ParILU

benchmark/utils/hip_linops.hip.hpp
pratikvn: Do you mean, both CuspspIlu and HIPspilu ? That would be nice. I guess then you have something similar to the strategies in CSR to select between the implementations ?

pratikvn:
Do you want to merge this before #400 or after ?

upsj:
I would probably merge it before, since it is much easier to review.

include/ginkgo/core/factorization/ilu.hpp
```diff
@@ -0,0 +1,153 @@
...
+    GKO_ENABLE_BUILD_METHOD(Factory);
+
+protected:
+    Ilu(const Factory *factory, std::shared_ptr<const gko::LinOp> system_matrix)
+        : Composition<ValueType>{factory->get_executor()},
+          parameters_{factory->get_parameters()}
+    {
+        if (parameters_.l_strategy == nullptr) {
+            parameters_.l_strategy =
+                std::make_shared<typename matrix_type::classical>();
```
pratikvn, line 123: Is `automatical` a better default ?

include/ginkgo/core/factorization/ilu.hpp
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    /**
+     * Generates the incomplete LU factors, which will be returned as a
+     * composition of the lower (first element of the composition) and the
+     * upper factor (second element). The dynamic type of L is l_matrix_type,
+     * while the dynamic type of U is u_matrix_type.
+     *
+     * @param system_matrix  the source matrix used to generate the factors.
+     *                       @note: system_matrix must be convertable to a Csr
```
pratikvn, line 139: nit:
```suggestion
     *                       @note: system_matrix must be convertible to a Csr
```

core/factorization/ilu.cpp
```diff
@@ -0,0 +1,131 @@
...
+#include "core/factorization/par_ilu_kernels.hpp"
+
+
+namespace gko {
+namespace factorization {
+namespace ilu_factorization {
+
+GKO_REGISTER_OPERATION(compute_ilu, ilu_factorization::compute_lu);
+GKO_REGISTER_OPERATION(add_diagonal_elements,
+                       par_ilu_factorization::add_diagonal_elements);
```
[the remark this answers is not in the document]
upsj: I planned to do that with the ParILUT PR, maybe "factorization_common_kernels"? In the end, these would be

1. adding diagonal zeros 
2. separation into L and U
3. extracting the lower triangle from a matrix

core/factorization/ilu.cpp
```diff
@@ -0,0 +1,131 @@
...
+    exec->run(ilu_factorization::make_initialize_row_ptrs_l_u(
+        local_system_matrix.get(), l_row_ptrs.get_data(),
+        u_row_ptrs.get_data()));
+
+    IndexType l_nnz_it{};
+    IndexType u_nnz_it{};
+    // Get nnz from device memory
+    host_exec->copy_from(exec.get(), 1, l_row_ptrs.get_data() + num_rows,
+                         &l_nnz_it);
+    host_exec->copy_from(exec.get(), 1, u_row_ptrs.get_data() + num_rows,
```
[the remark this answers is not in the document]
upsj: Yes, this is just pointer arithmetic, the pointer values themselves are on the CPU (and sizeof(IndexType) is the same)

hip/base/hipsparse_bindings.hip.hpp
```diff
@@ -680,6 +694,118 @@ GKO_BIND_HIPSPARSE_GATHER(std::complex<double>, hipsparseZgthr);
...
+GKO_BIND_HIPSPARSE_ILU0_ANALYSIS(float, hipsparseScsrilu02_analysis);
+GKO_BIND_HIPSPARSE_ILU0_ANALYSIS(double, hipsparseDcsrilu02_analysis);
+#if defined(hipsparseVersionMajor) && defined(hipsparseVersionMinor) && \
+    ((hipsparseVersionMajor > 1) ||                                     \
+     (hipsparseVersionMajor == 1 && hipsparseVersionMinor >= 4))
+GKO_BIND_HIPSPARSE_ILU0_ANALYSIS(std::complex<float>,
+                                 hipsparseCcsrilu02_analysis);
+GKO_BIND_HIPSPARSE_ILU0_ANALYSIS(std::complex<double>,
+                                 hipsparseZcsrilu02_analysis);
+#endif  // hipsparse version >= 1.4
```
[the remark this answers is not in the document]
upsj: That should be the default (see GKO_NOT_IMPLEMENTED above)

cuda/factorization/ilu_kernels.cu
```diff
@@ -0,0 +1,95 @@
...
+                            m->get_const_row_ptrs(), m->get_const_col_idxs(),
+                            info, CUSPARSE_SOLVE_POLICY_USE_LEVEL,
+                            buffer.get_data());
+
+    cusparse::ilu0(handle, num_rows, nnz, desc, m->get_values(),
+                   m->get_const_row_ptrs(), m->get_const_col_idxs(), info,
+                   CUSPARSE_SOLVE_POLICY_USE_LEVEL, buffer.get_data());
+
+    cusparse::destroy(info);
+    cusparse::destroy(desc);
```
thoasm, line 85: nit: Can we put this in a managing class (like a guard), so we are exception-safe here?

core/factorization/ilu.cpp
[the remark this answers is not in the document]
yhmtsai: Sorry, my bad. It declares the array location not summation of two value.

include/ginkgo/core/factorization/ilu.hpp
thoasm: I think we still can't use `automatical` on a reference executor when we did not compile a CUDA executor.

cuda/factorization/ilu_kernels.cu
upsj: While we could do this at some point, the only reason why this would be necessary would be if a sparselib invocation fails, either due to usage errors (unsupported index type, ...) or due to internal issues. IMO this is not really something we can recover from, so I am fine with leaking resources here.

core/factorization/ilu.cpp
```diff
@@ -0,0 +1,131 @@
...
+    GKO_ASSERT_IS_SQUARE_MATRIX(system_matrix);
+
+    const auto exec = this->get_executor();
+    const auto host_exec = exec->get_master();
+
+    // Only copies the matrix if it is not on the same executor or was not in
+    // the right format. Throws an exception if it is not convertible.
+    auto local_system_matrix = CsrMatrix::create(exec);
+    as<ConvertibleTo<CsrMatrix>>(system_matrix.get())
+        ->convert_to(local_system_matrix.get());
```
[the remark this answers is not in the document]
upsj: I will do that in this PR

core/factorization/ilu.cpp
```diff
@@ -0,0 +1,131 @@
...
+        exec, matrix_size, std::move(u_vals), std::move(u_col_idxs),
+        std::move(u_row_ptrs), parameters_.u_strategy);
+
+    // Separate L and U: columns and values
+    exec->run(ilu_factorization::make_initialize_l_u(
+        local_system_matrix.get(), l_factor.get(), u_factor.get()));
+
+    return Composition<ValueType>::create(std::move(l_factor),
+                                          std::move(u_factor));
+}
```
tcojean, line 123: Here I'm a bit dubious in terms of benchmark quality. While most of that is required for Ginkgo specifics of the interface, if we are to benchmark the vendor libraries ilu factorization, then pretty much most of the relevant time would be comprised in `compute_ilu` no? It's the same in par_ilut with the `compute_l_u_factors` function being the important function AFAIK. All of these allocations and so on to wrap the data into CSR matrices that we then put in a composition are purely Ginkgo specifics. Therefore the `generate` time would have little relevance when benchmarking.

upsj: The generate time would have little relevance, yes. But we can always just report the runtime of the `compute_ilu` operation (and I should note that the overall runtime is dominated by that and the triangular solve analysis phases)

tcojean: Yes, of course thanks to the loggers which report the time of separate operations. I think at the minimum we should have an immense warning somewhere, that the correct "generate" or factorization time at least for the factorizations should be taken from the specific `compute` kernels, and not the global generate time. That is different from say Jacobi, where the `generate` time is straight to the point a rather proper representation of the actual time.

Whereas here, due to the fact we want to separate the L U into Ginkgo CSR matrices among other quality checks, we can/could have a significant overhead which would blur the reality of the results.

upsj: I will add some performance numbers before we merge this PR, as far as I remember, the separation into L and U was really negligible

upsj: overhead

The actual kernels (`factorization::*`) look to be around 5% of the total runtime, the allocations and copy we would need to do anyways. The only thing that kind of surprises me is the `free` overhead, but I am not sure how much we can do about that.

tcojean: Thanks for these plots that's an interesting picture. Copy seems OK but free and allocate look rather significant, in particular the free indeed. For ani7 which I guess is a bigger matrix, it looks like the proportion of these "extra" operations are not so insignificant anymore except for the V100. The three architectures really don't behave in the same way.

upsj: Actually I accidentally added two different benchmark runs from the K20Xm to the plot, instead of correctly adding the Radeon7 timings. The plots should look like this:
overhead-abs
overhead-rel

hip/base/hipsparse_bindings.hip.hpp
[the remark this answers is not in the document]
yhmtsai: thanks

core/factorization/ilu.cpp
hartwiganzt: Wait - the generation time for ani7 on v100 is exceeding that of k20?

upsj: Yes

hartwiganzt: just that I understand it right - the timing includes the preparation of the triangular factors for the preconditioner application? In the performance graph you sent me, is it included there too?

upsj: Yes, these are the complete preconditioner generation runtimes, containing the elimination of structural zeros (add_diagonal_elements), the ILU generation (compute_lu), the separation into lower and upper triangular factors (initialize*_l_u) and the solver analysis phases (generate). I was just as surprised about this as you are. The setup is
```
auto ilu_fact = std::shared_ptr<gko::LinOpFactory>(
    gko::factorization::Ilu<etype>::build().on(exec));
gko::preconditioner::Ilu<>::build()
     .with_factorization_factory(ilu_fact)
     .on(exec);
```

hartwiganzt: hum... maybe we should exclude the preparation of the sparse systems in the plot for increasing ani dimension? How would it look like then? [https://www.overleaf.com/project/5e71c3074ed74a0001c2bbd8](https://www.overleaf.com/project/5e71c3074ed74a0001c2bbd8)

tcojean: As I was saying, I think for the sparselib case the `compute_lu` function is the most representative of the actual work. The problem is how relevant it is to compare against parilu's `compute_l_u_factors`, since this already writes in the separate factor instead of in place as sparselib do.

pratikvn: Doesn't this actually just show that the hipSPARSE ILU generation is faster than the CUSPARSE one ? With also that your compute_l_u is also faster on the AMD gpus. Maybe, the applies are different ? As you say, the apply on the K20x is much slower than on the V100. Is it similar for the Radeon 7 ?

upsj: The apply runtimes on the V100 and Radeon 7 are comparable, around 4ms. The astonishing thing is that apparently the analysis phases of ILU(0) and the triangular solves did not get any faster with at least 5 years of hardware development (The compute_lu kernel actually contains two phases: the analysis and the ILU solve. With the current setup, it is however not possible to separately measure their runtimes. Separating them into two kernels would be complicated, as this would require something similar to the SolveStruct)

pratikvn: The compute_lu is a ParILU isnt it ? So by nature isnt that faster than the level scheduled ILU(0) analysis and trisolve phases ?

upsj: No, this is still only the sparselib ILU, no ParILU in these plots

tcojean: Another interesting thing, for ani4 compute_lu (and everything else) seem to have shrinked quite a bit between V100 and K20. Weird that ani7 doesn't behave like that.

core/factorization/factorization_kernels.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+namespace gko {
+namespace kernels {
+
+
+#define GKO_DECLARE_FACTORIZATION_ADD_DIAGONAL_ELEMENTS_KERNEL(ValueType,   \
+                                                               IndexType)   \
+    void add_diagonal_elements(std::shared_ptr<const DefaultExecutor> exec, \
+                               matrix::Csr<ValueType, IndexType> *mtx,      \
+                               bool is_sorted)
```
yhmtsai, line 53:
```suggestion
                               bool is_sorted)
                               
```

core/factorization/factorization_kernels.hpp
```diff
@@ -0,0 +1,122 @@
...
+#define GKO_DECLARE_FACTORIZATION_INITIALIZE_ROW_PTRS_L_U_KERNEL(ValueType, \
+                                                                 IndexType) \
+    void initialize_row_ptrs_l_u(                                           \
+        std::shared_ptr<const DefaultExecutor> exec,                        \
+        const matrix::Csr<ValueType, IndexType> *system_matrix,             \
+        IndexType *l_row_ptrs, IndexType *u_row_ptrs)
```
yhmtsai, line 59:
```suggestion
        IndexType *l_row_ptrs, IndexType *u_row_ptrs)
        
```

upsj:
@yhmtsai The main point of `factorization_kernels` is to move kernels from ParILU to a more general file since they are used in many different algorithms (ParILU(T), sparselib ILU, ...?)

Issue #402, opened 2019-11-26
Possible interface breaking in Ginkgo 2.0.0


yhmtsai:
Track the possible interface breaking we find now.
1. csr strategy (#401)
2. csr clac_size -> calc_size

If there is anything miss, please feel free to edit it.

PR #489, opened 2020-03-27
Add ISAI reference implementation

thoasm:
This PR adds a reference implementation of ISAI (Incomplete Sparse Approximate Inverse).

Currently, I am not using a Composition internally because it is unable to handle multiple right hand sides. I will look into it and probably change that in the upcomming PR for the CUDA, HIP and OMP executor implementation.


reference/preconditioner/isai_kernels.cpp
```diff
@@ -0,0 +1,223 @@
...
+    for (IndexType col = 0; col < size[1] + 1; ++col) {
+        i_row_ptrs[col] = m_row_ptrs[col];
+    }
+    for (IndexType i = 0; i < num_elems; ++i) {
+        i_cols[i] = m_cols[i];
+    }
+
+    std::vector<ValueType> rhs;  // RHS for local trisystem
+    // memory for dense trisystem in column major:
+    std::vector<ValueType> trisystem;
```
upsj: Maybe use Array here so we can keep track of the allocations

reference/preconditioner/isai_kernels.cpp
```diff
@@ -0,0 +1,223 @@
...
+    for (IndexType row = 0; row < size[0]; ++row) {
+        const auto i_row_begin = i_row_ptrs[row];
+        const auto i_row_end = i_row_ptrs[row + 1];
+        const auto i_row_elems = i_row_end - i_row_begin;
+
+        trisystem.clear();
+        trisystem.reserve(i_row_elems * i_row_elems);
+        for (IndexType i = 0; i < i_row_elems * i_row_elems; ++i) {
+            trisystem.push_back(zero<ValueType>());
+        }
```
upsj: std::fill_n

reference/preconditioner/isai_kernels.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+        rhs.clear();
+        rhs.reserve(i_row_elems);
+
+        trs_solve(i_row_elems, trisystem, rhs);
+
+        // Drop RHS as a row to memory (since that is the computed inverse)
+        for (IndexType i = 0; i < i_row_elems; ++i) {
+            i_vals[i_row_begin + i] = rhs[i];
+        }
```
upsj: std::copy_n

reference/preconditioner/isai_kernels.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+    // expect mtx and inverse_mtx to have the same number of elems
+    const auto num_elems = mtx->get_num_stored_elements();
+    // Copy sparsity pattern of original into the inverse of L
+    for (IndexType col = 0; col < size[1] + 1; ++col) {
+        i_row_ptrs[col] = m_row_ptrs[col];
+    }
+    for (IndexType i = 0; i < num_elems; ++i) {
+        i_cols[i] = m_cols[i];
+    }
```
upsj: Channeling my inner Sean Parent: std::copy_n

reference/preconditioner/isai_kernels.cpp
```diff
@@ -0,0 +1,223 @@
...
+ * @brief The Isai preconditioner namespace.
+ *
+ * @ingroup isai
+ */
+namespace isai {
+
+
+template <typename ValueType, typename IndexType, typename Callable>
+void generic_generate(std::shared_ptr<const DefaultExecutor> exec,
+                      const matrix::Csr<ValueType, IndexType> *mtx,
```
upsj: For clarity: Maybe annotate the matrix
```suggestion
                      const matrix::Csr<ValueType, IndexType> *mtx_csc,
```
And do I understand correctly that inverse_mtx is also transposed/stored as CSC?

reference/preconditioner/isai_kernels.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+    Target: Solving (aiM * M = I)_{S} (aiM * M = I for the sparsity pattern S)
+    aiM[i, :] * D(i) = e(i)^T
+    <=> D(i)^T * vL[i, :]^T = e(i)   =^ Triangular system (Trs)
+    Solve Trs, fill in aiM row by row (coalesced access)
+    */
+    const auto size = mtx->get_size();
+    const auto m_row_ptrs = mtx->get_const_row_ptrs();
+    const auto m_cols = mtx->get_const_col_idxs();
+    const auto m_vals = mtx->get_const_values();
```
upsj: To maximize confusion: 
```suggestion
    const auto m_col_ptrs = mtx_csc->get_const_row_ptrs();
    const auto m_rows = mtx_csc->get_const_col_idxs();
    const auto m_vals = mtx_csc->get_const_values();
```
This is of course a matter of opinion, this suggestion aims at reducing the amount of abstractions (or hidden transpositions) you need to keep in mind to understand the code.

core/preconditioner/isai.cpp
```diff
@@ -0,0 +1,99 @@
...
+    auto exec = this->get_executor();
+    auto csr_l = copy_and_convert_to<Csr>(exec, to_invert_l);
+    const auto num_elems = csr_l->get_num_stored_elements();
+
+    std::shared_ptr<Csr> inverted_l =
+        Csr::create(exec, csr_l->get_size(), num_elems, csr_l->get_strategy());
+    exec->run(isai::make_generate_l(csr_l.get(), inverted_l.get()));
+
+    // call make_srow
+    inverted_l->set_strategy(inverted_l->get_strategy());
```
upsj: You could argue that this is the responsibility of the kernel (maybe by adding a CsrBuilder to it)

include/ginkgo/core/preconditioner/isai.hpp
```diff
@@ -0,0 +1,207 @@
...
+ *
+ * @tparam ValueType  precision of matrix elements
+ * @tparam IndexType  precision of matrix indexes
+ *
+ * @ingroup isai
+ * @ingroup precond
+ * @ingroup LinOp
+ */
+template <typename ValueType = default_precision, typename IndexType = int32>
+class Isai : public EnableLinOp<Isai<ValueType, IndexType>> {
```
upsj: This class combines two (imho completely distinct) responsibilities: The ISAI itself and the functionality of a `Composition`. I would argue it is cleaner to keep them separate and give the ISAI another factory parameter (or figure out a more general way of attaching *type* flags to matrices like with CUSPARSE_FILL_UPPER) to determine if it's a lower or upper triangular matrix.

reference/preconditioner/isai_kernels.cpp
```diff
@@ -0,0 +1,223 @@
...
+                const matrix::Csr<ValueType, IndexType> *l_csr,
+                matrix::Csr<ValueType, IndexType> *inverse_l)
+{
+    auto trs_solve = [](IndexType size, std::vector<ValueType> &trisystem,
+                        std::vector<ValueType> &rhs) {
+        // RHS is the identity: zero everywhere except for the last entry
+        // since that is the row we are searching the inverse for
+        for (IndexType d_row = 0; d_row < size - 1; ++d_row) {
+            rhs.push_back(zero<ValueType>());
+        }
```
upsj: std::fill_n

reference/preconditioner/isai_kernels.cpp
```diff
@@ -0,0 +1,223 @@
...
+                matrix::Csr<ValueType, IndexType> *inverse_u)
+{
+    auto trs_solve = [](IndexType size, std::vector<ValueType> &trisystem,
+                        std::vector<ValueType> &rhs) {
+        // RHS is the identity: zero everywhere except for the first entry
+        // since that is the row we are searching the inverse for
+        rhs.push_back(one<ValueType>());
+        for (IndexType d_row = 1; d_row < size; ++d_row) {
+            rhs.push_back(zero<ValueType>());
+        }
```
upsj: std::fill_n

reference/preconditioner/isai_kernels.cpp
thoasm: I had it like that at one point (when I had it in CSC), but now, everything is in CSR.

include/ginkgo/core/preconditioner/isai.hpp
thoasm: I think matrix tags in general would be a good idea to add to `LinOp` objects. I added it to the Ginkgo 2.0 Issue #402 
I will also add the possibility to generate the Preconditioner for only a lower or an upper triangular matrix.

thoasm:
All problems should now be resolved and most of the comments from @upsj are addressed.
We might have to discuss the responsibilities further.

reference/preconditioner/isai_kernels.cpp
```diff
@@ -0,0 +1,217 @@
...
+        // RHS is the identity: zero everywhere except for the last entry
+        // since that is the row we are searching the inverse for
+        std::fill_n(rhs, size - 1, zero<ValueType>());
+        rhs[size - 1] = one<ValueType>();
+
+        // Note: `trisystem` is an upper triangular matrix here (stored in
+        //       colum major)
+        for (IndexType d_col = size - 1; d_col >= 0; --d_col) {
+            const auto diag = trisystem[d_col * size + d_col];
+            const auto bot = rhs[d_col] / diag;
```
pratikvn, line 165: What happens if `diag` is zero (bot = nan) or close to zero (badly conditioned) ?

include/ginkgo/core/preconditioner/isai.hpp
```diff
@@ -0,0 +1,278 @@
...
+     *                 upper triangular matrix (L and U)
+     */
+    explicit Isai(const Factory *factory, std::shared_ptr<const LinOp> factors)
+        : EnableLinOp<Isai>(factory->get_executor(), factors->get_size()),
+          parameters_{factory->get_parameters()}
+    {
+        if (parameters_.exclusive_factor_l && parameters_.exclusive_factor_u) {
+            GKO_NOT_SUPPORTED(parameters_);
+        }
+        auto comp = dynamic_cast<const Composition<ValueType> *>(factors.get());
```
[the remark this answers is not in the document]
thoasm: I don't see any advantage for using `dynamic_pointer_cast` instead of `dynamic_cast` here.
I know that the object is at the very least valid until the end of the function (because the argument is a `std::shared_ptr`), and I don't need to store it anywhere or take ownership.
Using `dynamic_pointer_cast` has a bit more overhead (although that is neglectable here) and does not provide better readability than `dynamic_cast` in my opinion.

reference/preconditioner/isai_kernels.cpp
thoasm: If `diag` is zero, then bot is actually `+/-infinity`. However, I don't really see any way around it in this case. I don't think I can do any kind of pivoting in a simple triangular solve.
Unfortunately, we don't have any error-channel, so we can't directly tell the user that something went wrong. It might be better to have `infinity` or `NaN` in the matrix, so the problem can be traced back.

include/ginkgo/core/preconditioner/isai.hpp
```diff
@@ -0,0 +1,280 @@
...
+         *
+         * @note: Setting both exclusive_factor_l and exclusive_factor_u
+         *        results in an exception being thrown when generating the
+         *        preconditioner.
+         */
+        bool GKO_FACTORY_PARAMETER(exclusive_factor_u, false);
+
+        /**
+         * @brief Optimization parameter that skips the sorting if the input
+         *        matrix/matrices (required by the algorithm).
```
yhmtsai, line 131:
```suggestion
         *        matrix/matrices are sorted (required by the algorithm).
```
Maybe miss something here?

reference/preconditioner/isai_kernels.cpp
```diff
@@ -0,0 +1,229 @@
...
+    I := Identity matrix
+    e(i) := unit vector i (containing all zeros except for row i, which is one)
+    S := Sparsity pattern of the desired aiM
+    S(i) := Sparsity pattern of row i of aiM (Set of non-zero columns)
+    D(i) := L[S(i), S(i)]
+    aiM := approximate inverse of M
+
+    Target: Solving (aiM * M = I)_{S} (aiM * M = I for the sparsity pattern S)
+    aiM[i, :] * D(i) = e(i)^T
+    <=> D(i)^T * vL[i, :]^T = e(i)   =^ Triangular system (Trs)
```
yhmtsai, line 76: what is the vL here?

reference/preconditioner/isai_kernels.cpp
```diff
@@ -0,0 +1,229 @@
...
+        auto rhs = rhs_array.get_data();
+
+        trs_solve(i_row_elems, trisystem, rhs);
+
+        // Drop RHS as a row to memory (since that is the computed inverse)
+        for (size_type i = 0; i < i_row_elems; ++i) {
+            const auto new_val = rhs[i];
+            auto idx = i_row_begin + i;
+            // Check for non-finite elements which should not be copied over
+            if (isfinite(new_val)) {
```
yhmtsai, line 143:
```suggestion
            if (gko::isfinite(new_val)) {
```
use `gko::` should solve the MSVC compile problem

tcojean:
Here: https://github.com/ginkgo-project/ginkgo/blob/develop/include/ginkgo/core/base/dim.hpp#L100

What would make this undefined behavior? Is that not covered by 5.7.5 in the standard? The only issue would be bound checking, but we have the assert.
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3690.pdf

upsj:
I think the relevant part here would be § 9.2.13:

Nonstatic data members of a (non-union) class with the same access control (Clause 11) are allocated so that **later members have higher addresses** within a class object. The order of allocation of non-static data members with different access control is unspecified (Clause 11). Implementation alignment requirements **might cause two adjacent members not to be allocated immediately after each other**;

tcojean:
Yes indeed, with the code depending on `first_` and `rest_` being laid out in a strict consecutive fashion, that is rather broken. Your section shows it clearly. I had completely forgotten this had been implemented like that...

reference/preconditioner/isai_kernels.cpp
thoasm: Good catch. I was writing this for the L matrix first and modified it, so it would work for any matrix. `vL` was the abbreviation for in**V**erse **L** at the time before I changed it to the current version.

thoasm:
I also did not know or forgot that we store and access the dimensions that way. I assume that the intel compiler saw a chance to optimize there (to only copy the `first_` part), which caused the second dimension to never be initialized (valgrind also noted that there was a conditional jump on an uninitialized variable).

reference/preconditioner/isai_kernels.cpp
thoasm: Thanks!
I forgot about that when I wrote this code...

reference/test/preconditioner/isai_kernels.cpp
```diff
@@ -0,0 +1,626 @@
...
+    gko::kernels::reference::isai::generate_l_inverse(
+        this->exec, gko::lend(l_mtx), gko::lend(result));
+
+    // Results in a slightly different version than u_sparse_inv->transpose()
+    // because a different row-sparsity pattern is used in u_sparse vs. l_mtx
+    // (only one value changes compared to u_sparse_inv->transpose())
+    const auto expected = Csr::create(
+        this->exec, gko::dim<2>(4, 4),
+        I<value_type>{-.5, .125, .25, .3125, .125, -.5, .125, .25, .5},
+        I<index_type>{0, 0, 1, 0, 1, 2, 0, 2, 3}, I<index_type>{0, 1, 3, 6, 9});
```
tcojean, line 255: Same here, are the `I<...>` needed? `Csr` should already have the different types embeded in its template parameters, so I would be surprised.

reference/test/preconditioner/isai_kernels.cpp
```diff
@@ -0,0 +1,626 @@
...
+    GKO_ASSERT_MTX_NEAR(l_inv, this->l_sparse_inv, r<value_type>::value);
+    GKO_ASSERT_MTX_NEAR(u_inv, this->u_sparse_inv, r<value_type>::value);
+}
+
+
+TYPED_TEST(Isai, GeneratesWithUnsortedCsr)
+{
+    using Csr = typename TestFixture::Csr;
+    using Composition = typename TestFixture::Composition;
+    using value_type = typename TestFixture::value_type;
```
tcojean, line 365: you could also name this `T` since you only use it for the `r<...>::value`.

reference/test/preconditioner/isai_kernels.cpp
```diff
@@ -0,0 +1,626 @@
...
+    const auto comp = gko::share(Composition::create(
+        gko::share(this->l_dense), gko::share(this->u_dense)));
+    const auto isai = this->isai_factory->generate(comp);
+
+    isai->apply(gko::lend(vec1), gko::lend(res1));
+    isai->apply(gko::lend(vec2), gko::lend(res2));
+
+    GKO_ASSERT_MTX_NEAR(res1, l(I<T>{25, -40, -4}), r<T>::value);
+    GKO_ASSERT_MTX_NEAR(res2,
+                        l({I<T>{.625, 14.125}, I<T>{5, -43}, I<T>{-.5, -18.5}}),
```
tcojean, line 417: So far this one seems like the only one (with vec2 up here) where the `I<T>` should be needed because there can be list of complex of T vs list of list of T ambiguity.

include/ginkgo/core/preconditioner/isai.hpp
pratikvn: I have moved this to #495

reference/test/preconditioner/isai_kernels.cpp
thoasm: I use it here as an alias to `std::initializer_list<>`. The problem is that braced initialization lists do not work with unspecified template arguments because the type is unknown (therefore, it also does not work with perfect forwarding).
`auto` has a special behavior with braced initialization lists, it will automatically convert it to an `std::initializer_list`. For templates, this is not the case.

core/preconditioner/isai.cpp
```diff
@@ -0,0 +1,141 @@
...
+    static_assert(
+        std::is_same<Csr, matrix::Csr<typename Csr::value_type,
+                                      typename Csr::index_type>>::value,
+        "The given `Csr` type must be of type `matrix::Csr`!");
+    if (skip_sorting && exec == mtx->get_executor()) {
+        auto csr_mtx = dynamic_cast<const Csr *>(mtx);
+        if (csr_mtx) {
+            // Here, we can just forward the pointer with an empty deleter
+            // since it is already sorted and in the correct format
+            return {csr_mtx, [](const Csr *) {}};
```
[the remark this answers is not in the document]
thoasm: Yes, that is definitively better. I changed it to that.

reference/test/preconditioner/isai_kernels.cpp
thoasm: Unfortunately, I also need the `l()` here (which is defined in [`core/test/utils/assertions.hpp`](https://github.com/ginkgo-project/ginkgo/blob/develop/core/test/utils/assertions.hpp#L649) as a function taking and returning an `initializer_list`). I need it so the macro does not read the commas `,` as more arguments.

By the way: I did some testing, and we could use this `l(std::initializer_list<T> )` function instead of the alias `I<T>` (`using I<T> = std::initializer_list<T>`). The draw back, however, would be that we need an additional round brace in the end.
```c++
I<T>{1, 2, 3};
l<T>({1, 2, 3}); // equivalent to statement above
```
Personally, I would keep both variants (the alias is shorter when we need to specify the template argument), but I thought it is good to know what the `l()` function actually does.

reference/test/preconditioner/isai_kernels.cpp
```diff
@@ -0,0 +1,626 @@
...
+          u_s_unsorted{
+              Csr::create(exec, gko::dim<2>(4, 4),
+                          I<value_type>{-2., -1., 1., 1., 1., 4., -2., 1., 2.},
+                          I<index_type>{0, 2, 1, 3, 2, 1, 2, 3, 3},
+                          I<index_type>{0, 4, 6, 8, 9})},
+          u_sparse_inv{Csr::create(
+              exec, gko::dim<2>(4, 4),
+              I<value_type>{-.5, .125, .3125, .09375, .25, .125, -.5, .25, .5},
+              I<index_type>{0, 1, 2, 3, 1, 2, 2, 3, 3},
+              I<index_type>{0, 4, 6, 8, 9})}
```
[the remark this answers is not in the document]
thoasm: Yes, I do need it here as well because I create a CSR matrix, which uses the `create` function with perfect forwarding. Since template argument deduction fails / is illegal with braced init lists, I need to specify a type (and an `initializer_list` is ideal for that).
In the `ILU` case, we use [`matrix::initialize`](https://github.com/ginkgo-project/ginkgo/blob/develop/include/ginkgo/core/matrix/dense.hpp#L604), which has `initializer_list` as a parameter type, so braced init lists work.

core/preconditioner/isai.cpp
```diff
@@ -0,0 +1,146 @@
...
+ * If the given matrix was already sorted, is on the same executor and with a
+ * dynamic type of `const Csr`, the same pointer is returned with an empty
+ * deleter.
+ * In all other cases, a new matrix is created, which stores the converted Csr
+ * matrix.
+ * If `skip_sorting` is false, the matrix will be sorted by column index,
+ * otherwise, it will not be sorted.
+ */
+template <typename Csr>
+std::unique_ptr<const Csr, std::function<void(const Csr *)>>
```
pratikvn, line 75: This can probably marked as `@internal` or in case it is useful for other cases, then moved to a common file ?

core/preconditioner/isai.cpp
```diff
@@ -0,0 +1,146 @@
...
+    // Here, we assume that a sorted matrix converted to CSR will also be
+    // sorted
+    if (!skip_sorting) {
+        copy->sort_by_column_index();
+    }
+    return {copy.release(), std::default_delete<const Csr>{}};
+}
+
+
+template <isai_type IsaiType, typename ValueType, typename IndexType>
```
pratikvn, line 102: Does it make sense to instantiate this when `IsaiType = isai_type::upper` ?

include/ginkgo/core/preconditioner/isai.hpp
```diff
@@ -0,0 +1,206 @@
...
+    using Csr = matrix::Csr<ValueType, IndexType>;
+    static constexpr isai_type type{IsaiType};
+
+    /**
+     * Returns the approximate inverse of the given matrix (either L or U,
+     * depending on the template parameter IsaiType).
+     *
+     * @returns the generated approximate inverse
+     */
+    std::shared_ptr<const Csr> get_system_matrix() const
```
pratikvn, line 108: If this is the computed inverse, then I think the function should be named `get_matrix_inverse` or something like that. Or if this is actually returning the passed in `system_matrix`, then the documentation is incorrect

core/test/preconditioner/isai.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+TYPED_TEST(IsaiFactory, ThrowsNoConversionCsrU)
+{
+    using Csr = typename TestFixture::Csr;
+    auto mtx = DummyOperator::create(this->exec, gko::dim<2>{2, 2});
+
+    ASSERT_THROW(this->upper_isai_factory->generate(gko::share(mtx)),
+                 gko::NotSupported);
+}
+
```
pratikvn, line 137: I think you should also check if the factory parameter `skip_sorting` is being set properly.

core/preconditioner/isai.cpp
thoasm: I did not think about marking this as `@internal` because it is in the `.cpp` file, but I will add it.

core/preconditioner/isai.cpp
thoasm: It should not hurt since it is a non-`virtual` function. Additionally, this would complicate the code quite a lot more (I can only come up with a method that would probably be similar to writing both `LowerIsai` and `UpperIsai` by hand), and since it is a `private` function, I did not see the need for that.

core/test/preconditioner/isai.cpp
thoasm: Good idea, I will add this test.

include/ginkgo/core/preconditioner/isai.hpp
thoasm: Good point, it is returning the approximate inverse.
I will rename it to `get_approximate_inverse` to reduce confusion.

reference/preconditioner/isai_kernels.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+        for (size_type i = 0; i < i_row_elems; ++i) {
+            const auto col = i_cols[i_row_begin + i];
+            const auto m_row_end = m_row_ptrs[col + 1];
+            auto m_row_ptr = m_row_ptrs[col];
+            auto i_row_ptr = i_row_begin;
+            // Values in `trisystem` will be stored in transposed order
+            // to prevent the need for an explicit transpose.
+            // Since `trisystem` stores in column major, this is equivalent
+            // to storing it in row-major and reading in column major.
```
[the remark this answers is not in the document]
thoasm: Is it better now? This comment changed to:
```
// Instead of transposing `trisystem` after filling it
// (which the algorithm dictates), we simply store the values
// transposed, so we skip the explicit transpose.
// Since `trisystem` is stored in column major, we can write to
// it as if it was stored in row major (therefore, skipping the
// explicit transpose afterwards)
```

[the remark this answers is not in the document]
yhmtsai: LGTM

tcojean:
@thoasm for the Windows compilation issue, maybe for now you can add `-O2` as a flag for the windows debug + shared compilation, either in the github workflow or preferably directly in Ginkgo's CMake files. I prefer that to disabling it altogether since we already cannot run debug + static due to size issues.

hartwiganzt:
@thoasm this can be merged, or?

thoasm:
@yhmtsai Unfortunately, moving the `CXXFLAGS` flags to the `GINKGO_COMPILER_FLAGS` did not work, as you can see in the output the github workflow [here](https://github.com/ginkgo-project/ginkgo/pull/489/checks?check_run_id=604661676).
A reason might be that we do not use `GINKGO_COMPILER_FLAGS` when building the tests, but that is just a guess.
I will revert the change and leave it as was (with `cflags == "-O1"`), so it passes all workflows properly.

hartwiganzt:
ready to be merged?

thoasm:
@hartwig Now that all tests succeeded, yes, I will merge it.

PR #493, opened 2020-04-02
Add gko::as for smart pointers

upsj:
As far as I can tell, we currently have no nice way of casting a `std::unique_ptr` to another type. There is `std::*_ptr_cast`, but that only works for `std::shared_ptr`.
Instead we need workarounds like 
```cpp
std::unique_ptr<...> ptr2{gko::as<...>(ptr.release())};
```
This PR adds an overload for `gko::as` that takes a `unique_ptr` as an rvalue, resetting it if successful.

### DIFF
--- a/include/ginkgo/core/base/utils.hpp
+++ b/include/ginkgo/core/base/utils.hpp
@@ -352,6 +352,56 @@ inline std::unique_ptr<typename std::decay<T>::type> as(
 
 
 /**
+ * Performs polymorphic type conversion of a shared_ptr.
+ *
+ * @tparam T  requested result type
+ * @tparam U  static type of the passed object
+ *
+ * @param obj  the shared_ptr to the object which should be converted.
+ *
+ * @return If successful, returns a shared_ptr to the subtype, otherwise throws
+ *         NotSupported. This pointer shares ownership with the input pointer.
+ */
+template <typename T, typename U>
+inline std::shared_ptr<typename std::decay<T>::type> as(std::shared_ptr<U> obj)
+{
+    if (dynamic_cast<typename std::decay<T>::type *>(obj.get())) {
+        return std::static_pointer_cast<typename std::decay<T>::type>(obj);
+    } else {
+        throw NotSupported(__FILE__, __LINE__, __func__,
+                           name_demangling::get_type_name(typeid(obj.get())));
+    }
+}
+
+
+/**
+ * Performs polymorphic type conversion of a shared_ptr.
+ *
+ * This is the constant version of the function.
+ *
+ * @tparam T  requested result type
+ * @tparam U  static type of the passed object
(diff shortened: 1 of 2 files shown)


include/ginkgo/core/base/utils.hpp
```diff
@@ -325,6 +325,32 @@ inline const typename std::decay<T>::type *as(const U *obj)
...
+    if (auto p = dynamic_cast<typename std::decay<T>::type *>(obj.get())) {
+        obj.release();
+        return std::unique_ptr<typename std::decay<T>::type>{p};
+    } else {
+        throw NotSupported(__FILE__, __LINE__, __func__,
+                           name_demangling::get_type_name(typeid(obj.get())));
+    }
+}
+
+
```
tcojean, line 353: At least for uniformity, does it not make sense to also implement a `std::shared_ptr` version?

upsj: That's true, so you mean for a shared_ptr const/non-const lvalue?

nbeams: I agree, I think it'd be really nice to have `as` work for basically any type of smart or plain pointer.

include/ginkgo/core/base/utils.hpp
```diff
@@ -325,6 +325,82 @@ inline const typename std::decay<T>::type *as(const U *obj)
...
+ * @tparam U  static type of the passed object
+ *
+ * @param obj  the unique_ptr to the object which should be converted.
+ *             If successful, it will be reset to a nullptr.
+ *
+ * @return If successful, returns a unique_ptr to the subtype, otherwise throws
+ *         NotSupported.
+ */
+template <typename T, typename U>
+inline std::unique_ptr<typename std::decay<T>::type> as(
```
upsj, line 341: Somewhat related: Why do we need the `std::decay` here? T is always user-specified, so it should not be a reference in any case. Can we just drop it from `gko::as`?

include/ginkgo/core/base/utils.hpp
tcojean: For `std::shared_ptr` I guess both const/non-const can be supported, at least with `dynamic_pointer_cast` for example.

upsj: Yes, I decided to add them both, since we often use shared_ptr<const ...> in our code.

include/ginkgo/core/base/utils.hpp
tcojean: It also does `remove_cv` in particular. You do not want `const` to appear here so that seems fine to me.

upsj: Alright, so I'll leave it as-is.

PR #496, opened 2020-04-06
Fix some minor typos

pratikvn:
Fix typos:
+ Abstact to Abstract,
+ Gingko to Ginkgo.

### DIFF
--- a/cmake/information_helpers.cmake
+++ b/cmake/information_helpers.cmake
@@ -18,12 +18,12 @@ macro(ginkgo_git_information)
                 OUTPUT_VARIABLE GINKGO_GIT_BRANCH
                 OUTPUT_STRIP_TRAILING_WHITESPACE)
             execute_process(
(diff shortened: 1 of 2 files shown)


cmake/information_helpers.cmake
```diff
@@ -18,12 +18,12 @@ macro(ginkgo_git_information)
                 OUTPUT_VARIABLE GINKGO_GIT_BRANCH
                 OUTPUT_STRIP_TRAILING_WHITESPACE)
             execute_process(
-                COMMAND ${GIT_EXECUTABLE} log -1 --format=%H ${Gingko_SOURCE_DIR}
+                COMMAND ${GIT_EXECUTABLE} log -1 --format=%H ${Ginkgo_SOURCE_DIR}
```
upsj, line 21: So we didn't catch this before since it is already being executed in Ginkgo_SOURCE_DIR?

pratikvn: Yes, I guess so.

PR #497, opened 2020-04-07
Fix hip atomic add on complex in updated hip

yhmtsai:
hcc fixed the printf problem in https://github.com/RadeonOpenCompute/hcc/pull/1290 and https://github.com/RadeonOpenCompute/hcc/pull/1354

I rerun the script (`dev_tools/container/gko-amd-gnu7-llvm60.baseimage`) to build the image `localhost:5000/hip_update` with the updated hip.

They handle complex in more general way.
`&(addr->x)` gets `hip_impl::Scalar_accessor<double, double __attribute__((ext_vector_type(2))), 1>::Address` not `double *`, so compiler can not find a proper `atomic_add`. The failed log in https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/500647667
[The related code](https://github.com/ROCm-Developer-Tools/HIP/blob/5f1420a229b41662f1f3f178de56daa10028d39c/include/hip/hcc_detail/hip_vector_types.h#L62-L86)

I use `static_cast<double *>` to make the conversion work. Another way is to get the address first `double *addrx = &(addr->x)` and then call the `atomic_add`.

If someone would like to try the printf in `localhost:5000/hip_update`, you need to define `HIP_ENABLE_PRINTF` and `export HCC_ENABLE_PRINTF=1` to enable printf.

sample.cpp
```
#include "hip/hip_runtime.h"

__global__ void run_printf() {
    int idx = threadIdx.x;
    printf("Hello World, tid = %d\n", idx);
}

int main() {
    hipSetDevice(0);
    hipLaunchKernelGGL(HIP_KERNEL_NAME(run_printf), dim3(1), dim3(7), 0, 0);
    hipDeviceSynchronize();
    return 0;
}
```
```
export HCC_ENABLE_PRINTF=1
hipcc -DHIP_ENABLE_PRINTF sample.cpp
```

Todo:
- [ ] Finish the review
- [x] Update docker image (#499)
- [ ] Revert the `.gitlab-ci.yml`

### DIFF
--- a/hip/components/atomic.hip.hpp
+++ b/hip/components/atomic.hip.hpp
@@ -52,8 +52,8 @@ __forceinline__ __device__ thrust::complex<float> atomic_add(
 {
     hipComplex *addr = reinterpret_cast<hipComplex *>(address);
     // Separate to real part and imag part
-    auto real = atomic_add(&(addr->x), val.real());
-    auto imag = atomic_add(&(addr->y), val.imag());
+    auto real = atomic_add(static_cast<float *>(&(addr->x)), val.real());
+    auto imag = atomic_add(static_cast<float *>(&(addr->y)), val.imag());
     return {real, imag};
 }
 
@@ -68,8 +68,8 @@ __forceinline__ __device__ thrust::complex<double> atomic_add(
 {
     hipDoubleComplex *addr = reinterpret_cast<hipDoubleComplex *>(address);
     // Separate to real part and imag part
-    auto real = atomic_add(&(addr->x), val.real());
-    auto imag = atomic_add(&(addr->y), val.imag());
+    auto real = atomic_add(static_cast<double *>(&(addr->x)), val.real());
+    auto imag = atomic_add(static_cast<double *>(&(addr->y)), val.imag());
     return {real, imag};
 }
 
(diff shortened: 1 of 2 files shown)


hip/components/atomic.hip.hpp
```diff
@@ -52,8 +52,8 @@ __forceinline__ __device__ thrust::complex<float> atomic_add(
 {
     hipComplex *addr = reinterpret_cast<hipComplex *>(address);
     // Separate to real part and imag part
-    auto real = atomic_add(&(addr->x), val.real());
-    auto imag = atomic_add(&(addr->y), val.imag());
+    auto real = atomic_add(static_cast<float *>(&(addr->x)), val.real());
+    auto imag = atomic_add(static_cast<float *>(&(addr->y)), val.imag());
```
upsj, line 56: I really hate that we have to do this cast since thrust::complex gives no reference access to the data members. Or does anybody see a more sensible approach to accessing them?
This feels like borderline undefined behavior

tcojean:
I integrated this commit into #499 since this is also a container fix, and 499 is the purpose of this.

hip/components/atomic.hip.hpp
thoasm: The thrust part should be kind of fine since we reinterpret it to `hipComplex`. `addr->x` should have direct access (since we access a member variable and not a function).

yhmtsai:
@tcojean Do I need to keep this PR?

tcojean:
@yhmtsai I think this can be closed if you are OK with the updated HIP containers of #499.

yhmtsai:
Sure.
Close this PR which is embedded in the PR #499

PR #499, opened 2020-04-08
Update Ginkgo to the new containers status.

tcojean:
This updates the containers usage in our main repository due to some changes made while developing the automatic container building system which is available there for project members: https://gitlab.com/ginkgo-project/ginkgo-containers

Exemple of automatic build pipeline: https://gitlab.com/ginkgo-project/ginkgo-containers/pipelines/133883673

I welcome any feedback on either this PR or the containers project. Now is a good time to make changes.

### Summary of this PR's changes
+ Intel support has been made clear in container names
+ A repository stores all the containers and more, so remove them from Ginkgo
  itself.
+ Intel 2017 is already in CUDA 9.2 image, so do not use it in 9.1.
+ Image names are fixed for new llvm scheme from 7 onward, there is no ".0"
+ The old nocuda with gnu 8 image is useless now that we have proper tags for
  requesting an Intel capable image (on the CUDA machine).


tcojean:
Everything is now working. There were some pretty weird unforeseen issues that are now fixed:
+ Intel 19 and C++17 create a linking error when using gflags. Maybe Intel uses a high standard such as C++17 by default? Using the C++11 standard fixes the issue. It was basically this issue: https://software.intel.com/en-us/forums/intel-c-compiler/topic/810447, but on gflags linkage after compiling it without explicitly setting C++11 as standard.
+ For some reason, the previous way of handling the CUDA flags (host compilers and architecture) now created an issue. One character (the first of the `-DCMAKE_CUDA_HOST_COMPILER` argument value, in fact) was weirdly disappearing from the CMake command line.
+ @yhmtsai's fix for HIP is also required and has been integrated to this PR.

pratikvn:
As this PR is already related to CI and Build, you can also probably disable the failing Windows builds (due to the symbol limits) and add something to INSTALL.md/README.md ?

.gitlab-ci.yml
```diff
@@ -389,25 +382,26 @@ build/nocuda/clang/core/release/shared:
...
-    C_COMPILER: icc
-    CXX_COMPILER: icpc
+    C_COMPILER: "icc"
+    CXX_COMPILER: "icpc"
     BUILD_REFERENCE: "OFF"
-    BUILD_TYPE: Debug
+    BUILD_TYPE: "Debug"
   tags:
     - private_ci
+    - cuda
```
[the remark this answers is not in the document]
tcojean: `Intel` is only available in the `fineci` machine, so that's a way of asking for that without adding an extra intel tag to the runners

.gitlab-ci.yml
```diff
@@ -426,35 +420,36 @@ build/nocuda/clang/omp/debug/static:
...
-    CXX_COMPILER: icpc
+    C_COMPILER: "icc"
+    CXX_COMPILER: "icpc"
     BUILD_OMP: "ON"
-    BUILD_TYPE: Release
+    BUILD_TYPE: "Release"
     BUILD_SHARED_LIBS: "OFF"
   tags:
     - private_ci
+    - cuda
```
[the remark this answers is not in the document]
tcojean: same there

thoasm:
It would be nice if we could add a protection step before deploying in gitlab.
If it would be possible, deploying after a sucessfull PR (on the gitlab side) would be better because at least another person would read over it. That way, it is less likely that we override our images on the CI with images that do not work or fail the CI jobs.

tcojean:
I think that would be possible, here is what I think would make sense:
+ In the branch/PR, build and do preliminary tests of the containers.
+ People review the branch/PR for acceptance
+ When the branch is accepted and merged, the deploy step of the containers is ran (no rebuilding or retesting happens).

Another issue is concerning the tests. There is already a lot of them, all put together I'm fairly certain they cover most of Ginkgo's requirements. I would consider these tests unit tests in the sense that they check a lot of the single units present in the containers (documentation build, cmake, networking, ...). Should we add new tests? Does it make sense to add a test which tries to build and/or run Ginkgo (though that could a bit complex to write that should be possible, also one issue is to not add Ginkgo issues into the container builds). In general, that's going in the direction of integration testing for containers.

tcojean:
@pratikvn I think it would be better if someone with the issue can try it. I can change it but I cannot make sure it works properly.
Also, since we are discussing that windows issue, I think I would rather add `-O2` flags for these two compilers for sebug/shared in our CMake rather than disable this altogether. We cannot check `debug/static` due to size issues. Now we cannot check `debug/shared` due to the export ordinal issue. That means we are losing `debug` altogether...

tcojean:
I put together a MR on gitlab with an implementation proposition for the scheme I described. Feel free to discuss it there.
https://gitlab.com/ginkgo-project/ginkgo-containers/-/merge_requests/1

hip/components/atomic.hip.hpp
```diff
@@ -52,8 +52,8 @@ __forceinline__ __device__ thrust::complex<float> atomic_add(
 {
     hipComplex *addr = reinterpret_cast<hipComplex *>(address);
     // Separate to real part and imag part
-    auto real = atomic_add(&(addr->x), val.real());
-    auto imag = atomic_add(&(addr->y), val.imag());
+    auto real = atomic_add(static_cast<float *>(&(addr->x)), val.real());
+    auto imag = atomic_add(static_cast<float *>(&(addr->y)), val.imag());
```
yhmtsai, line 56: As @upsj mentioned, does anyone has any idea to handle it?
Maybe [the related code](https://github.com/ROCm-Developer-Tools/HIP/blob/5f1420a229b41662f1f3f178de56daa10028d39c/include/hip/hcc_detail/hip_vector_types.h#L62-L86) can give some idea.

upsj: To copy the original question: Can we not also do this another way, e.g. by casting the value to a `float*` or `float[2]`? What does the detour via `hipComplex` give us? Hopefully, thrust::complex should give some kind of guarantee about the data layout/packed storage

yhmtsai: Do you mean `float * addrx = &(addr->x)`?

tcojean: I think he means to directly use the `thrust::complex` and not use `hipComplex` at all. Thrust gives the functions `imag()` and `real()` to access the values, do they not work? Why do we need to cast `addr` to `hipComplex` ?
https://thrust.github.io/doc/group__complex__numbers.html

upsj: Because they are returned by value, not by reference, so there is no address for the atomic operation to use.

tcojean: If you store them in an extra variable that should work, though we would be using more storage than a single pointer.

upsj: I don't think that would work, since the values need to be stored in shared or global memory for the atomic operations to work. I guess this is not really an important point, since it works right now.

tcojean:
I will go ahead and merge this so that our containers are fixed.

PR #501, opened 2020-04-14
Simple overhead benchmarking facilities

tcojean:
This is done by modifying the `solver` benchmarks with:
+ A new `overhead` boolean parameter to ask whether or not an overhead benchmark should be done. This will use a simple matrix with a single `1.0` `value,` `0.0` as `x`, and `nan` as `b` (same setup as the `ginkgo_overhead` example). 
+ A new `Overhead` LinOP which does a normal solver setup in terms of Factory parameters and generation, except that it has an `apply_impl` method which calls four empty kernels. We have multiple runtime polymorphism calls which come into play here. I am most likely forgetting many of them: 
    + generate,
    + apply,
    + system_matrix and preconditioner's apply,
    + four kernel calls (`run()` functions), ...
+ This `Overhead` LinOp can be also be used as a preconditioner.

This allows the following:
+ Benchmark any solver/preconditioner with the dummy data.
+ Benchmark the overhead LinOp (optionally with preconditioner) with the dummy data.

We get the following results currently, benchmarked on BW Unicluster with 10.000 iterations:
+ CG with dummy data,
   + generate: `0.829 us`,
   + apply (1000 iterations): `1277.557 us`.
   + apply (2000 iterations) - apply (1000 iterations): `1200.118 us`
+ bicgstab with dummy data,
   + generate: `0.814 us`,
   + apply (1000 iterations): `1259.751 us`.
   + apply (2000 iterations) - apply (1000 iterations): `1239.370 us`
+ fcg with dummy data,
   + generate: `0.837 us`,
   + apply (1000 iterations): `1447.981 us`.
   + apply (2000 iterations) - apply (1000 iterations): `1410.630 us`
+ cgs with dummy data,
   + generate: `0.849 us`,
   + apply (1000 iterations): `1001.975 us`.
   + apply (2000 iterations) - apply (1000 iterations): `881.990 us`
+ gmres with dummy data,
   + generate: `0.836 us`,
   + apply (1000 iterations): `1510.327 us`.
   + apply (2000 iterations) - apply (1000 iterations): `1485.361 us`
+ Overhead "solver" with dummy data,
   + generate: `0.788 us`,
   + apply: `0.912 us`.
+ Overhead "solver" with overhead "preconditioner" with dummy data:
   + generate: `1.479 us`,
   + apply: `2.610 us`.

### DIFF
--- a/benchmark/solver/solver.cpp
+++ b/benchmark/solver/solver.cpp
@@ -294,13 +294,19 @@ void solve_system(const std::string &solver_name,
         }
 
         // warm run
+        auto it_logger = std::make_shared<IterationLogger>(exec);
         for (unsigned int i = 0; i < FLAGS_warmup; i++) {
             auto x_clone = clone(x);
             auto precond = precond_factory.at(precond_name)(exec);
             auto solver = solver_factory.at(solver_name)(exec, give(precond))
                               ->generate(system_matrix);
+            solver->add_logger(it_logger);
             solver->apply(lend(b), lend(x_clone));
             exec->synchronize();
+            solver->remove_logger(gko::lend(it_logger));
+        }
+        if (FLAGS_warmup > 0) {
+            it_logger->write_data(solver_json["apply"], allocator);
         }
 
         // detail run
--- a/benchmark/solver/solver.cpp
+++ b/benchmark/solver/solver.cpp
@@ -72,6 +72,10 @@ DEFINE_uint32(
     nrhs, 1,
     "The number of right hand sides. Record the residual only when nrhs == 1.");
 
+// This allows to benchmark the overhead of a solver by using the following
+// data: A=[1.0], x=[0.0], b=[nan]. This data can be used to benchmark normal
+// solvers or using the argument --solvers=overhead, a minimal solver will be
+// launched which contains only a few kernel calls.
 DEFINE_bool(overhead, false,
             "If set, uses dummy data to benchmark Ginkgo overhead");
 
--- a/benchmark/solver/solver.cpp
+++ b/benchmark/solver/solver.cpp
@@ -35,6 +35,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include <algorithm>
 #include <chrono>
+#include <cmath>
 #include <cstdlib>
 #include <exception>
 #include <fstream>
@@ -45,6 +46,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "benchmark/utils/formats.hpp"
 #include "benchmark/utils/general.hpp"
 #include "benchmark/utils/loggers.hpp"
+#include "benchmark/utils/overhead_linop.hpp"
 
 
 // some Ginkgo shortcuts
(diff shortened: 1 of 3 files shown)


hartwiganzt:
I like this! This might also be useful for debugging / performance engineering. Would it be possible to get the information on how often a runtime decision was taken? Obviously, the overhead of counting this would be much higher, so we may need two runs?

benchmark/solver/solver.cpp
```diff
@@ -58,21 +60,25 @@ DEFINE_uint32(max_iters, 1000,
...
+              "Supported values are: bicgstab, cg, cgs, fcg, gmres, overhead");
 
-DEFINE_string(
-    preconditioners, "none",
-    "A comma-separated list of preconditioners to use."
-    "Supported values are: none, jacobi, adaptive-jacobi, parilu, ilu");
+DEFINE_string(preconditioners, "none",
+              "A comma-separated list of preconditioners to use. "
+              "Supported values are: none, jacobi, adaptive-jacobi, parilu, "
+              "ilu, overhead");
```
hartwiganzt, line 70: Maybe add an example of how to measure the overhead.

tcojean:
I added an iteration logger to the solver benchmark. I put this as part of the warmup since we should pretty much always have warmup and that doing this simple logging of iteration counts is not a "detailed" run. We can maybe put this logger in the main benchmarking portion, but that would add a few extra virtual calls so I decided to not proceed with this.

This revealed that indeed, the solver results were accounting for 1000 iterations of a CG solver, which changes the picture significantly (1.2 ms overhead over 1000 iterations isn't too bad). I added numbers for all solvers, and also used the strategy of doing 2000 iterations - 1000 iterations, which shows that the allocations while numerous are indeed pretty much nothing. Some solvers do have more overhead than others though, the extremes being CGS vs GMRES.

benchmark/utils/loggers.hpp
```diff
@@ -229,4 +229,29 @@ struct ResidualLogger : gko::log::Logger {
...
 
 
+// Logs the number of iteration executed
+struct IterationLogger : gko::log::Logger {
+    void on_iteration_complete(const gko::LinOp *,
+                               const gko::size_type &num_iterations,
+                               const gko::LinOp *, const gko::LinOp *,
+                               const gko::LinOp *) const override
+    {
+        this->num_iters = num_iterations;
```
upsj, line 239: When we have multiple right-hand sides: Do I understand correctly that this logs the maximum number of iterations over all rhs?

benchmark/utils/overhead_linop.hpp
```diff
@@ -0,0 +1,196 @@
...
+
+
+template <typename ValueType = default_precision>
+class Overhead : public EnableLinOp<Overhead<ValueType>>,
+                 public Preconditionable {
+    friend class EnableLinOp<Overhead>;
+    friend class EnablePolymorphicObject<Overhead, LinOp>;
+
+public:
+    using value_type = ValueType;
```
yhmtsai, line 105:
```suggestion
```
`value_type` is unused here.

benchmark/utils/overhead_linop.hpp
```diff
@@ -0,0 +1,196 @@
+#ifndef GKO_BENCHMARK_UTILS_OVERHEAD_LINOP_
```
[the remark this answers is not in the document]
tcojean: Thanks, the `make format_header` worked as expected.

benchmark/utils/overhead_linop.hpp
tcojean: I removed it here since that is not something exposed to users, but as a side note these are usually unused in the code's class itself. They can be used from outside the class.

yhmtsai: You are right. I forgot that it can be used from outside.
I am okay for keeping or removing it.

PR #503, opened 2020-04-20
Allow benchmarking from a matrix list file

tcojean:
This adds more option to the `run_all_benchmarks.sh` script in order to be more practical.

In particular:
+ A new variable is added, `MATRIX_LIST_FILE` which can be set to a file containing a list of suitesparse matrices (one per line) to allow more targeted benchmarking. Either the ID, the name, or `group/name` form can be used to identify the matrix. As an example, the following list would be valid to benchmark the matrices for the TOMS interface paper:
  ```
    1903
    Freescale/circuit5M
    rajat31
    Freescale/FullChip
    Bodendiek/CurlCurl_4
    Janna/Bump_2911
    Janna/Cube_Coup_dt0
    Janna/StocF-1465
    Bourchtein/atmosmodj
    thermal2
  ```
+ Options are added to configure the solvers and matrix formats to use
+ Options are added to configure tolerance and maximum number of iterations for solvers
+ Options are added to run detailed solver benchmarks or not (the detailed default is now set to off as part of the script, though the runner still has the detailed default set to true, what is the best setting for this? It's a lot of overkill information)

Suggestions for other convenient features are welcome.

### DIFF
--- a/benchmark/run_all_benchmarks.sh
+++ b/benchmark/run_all_benchmarks.sh
@@ -62,7 +62,7 @@ fi
 
 # Control whether to run detailed benchmarks or not.
 # Default setting is detailed=false. To activate, set DETAILED=1.
-if  [ "${DETAILED}" -eq 0 ]; then
+if  [ ! "${DETAILED}" ] || [ "${DETAILED}" -eq 0 ]; then
     DETAILED_STR="--detailed=false"
 else
     DETAILED_STR="--detailed=true"
@@ -80,6 +80,9 @@ if [ ! "${MATRIX_LIST_FILE}" ]; then
     use_matrix_list_file=0
 elif [ -f "${MATRIX_LIST_FILE}" ]; then
     use_matrix_list_file=1
+else
+    echo -e "A matrix list file was set to ${MATRIX_LIST_FILE} but it cannot be found."
+    exit 1
 fi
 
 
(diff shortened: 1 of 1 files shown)


tcojean:
Another issue, I suppose the name of the suitesparse matrices contains only a combination of the following, is that true?
+ letter
+ number
+ underscore
+ dash

yhmtsai:
Yes, it should be. 
all group/name are passed by ^[0-9a-zA-Z/_-]+$ by a small script

tcojean:
Thanks for checking, Mike! I could not find a format specification on their website, I guess bruteforcing is a good solution for lack of a better one.

benchmark/run_all_benchmarks.sh
```diff
@@ -40,6 +60,28 @@ if [ ! "${DEVICE_ID}" ]; then
...
+#   id or group/name or name.
+# Example:
+# 1903
+# Freescale/circuit5M
+# thermal2
+if [ ! "${MATRIX_LIST_FILE}" ]; then
+    use_matrix_list_file=0
+elif [ -f "${MATRIX_LIST_FILE}" ]; then
+    use_matrix_list_file=1
+fi
```
yhmtsai, line 83: throw an error or give warning when the matrix list is not existed?

benchmark/run_all_benchmarks.sh
```diff
@@ -173,9 +215,42 @@ generate_suite_sparse_input() {
...
+    local source_list_file=$1
+    local benchmark_list=""
+    local id=0
+    for mtx in $(cat ${source_list_file}); do
+        if [[ ! "$mtx" =~ ^[0-9]+$ ]]; then
+            if [[ "$mtx" =~ ^[a-zA-Z0-9_-]+$ ]]; then
+                id=$(${SSGET} -s "[ @name == $mtx ]")
+            elif [[ "$mtx" =~ ^[a-zA-Z0-9_-]+/[a-zA-Z0-9_-]+$ ]]; then
+                local group=$(echo $mtx | cut -d"/" -f1)
+                local name=$(echo $mtx | cut -d"/" -f2)
```
yhmtsai, line 228:
```suggestion
            elif [[ "$mtx" =~ ^([a-zA-Z0-9_-]+)\/([a-zA-Z0-9_-]+)$ ]]; then
                local group="${BASH_REMATCH[1]}"
                local name="${BASH_REMATCH[2]}"
```
I prefer bash built-in function rather than system call although it does not hurt the performance much.

PR #500, opened 2020-04-08
Reverse Cuthill-McKee

lksriemer:
**Updated**

This PR implements the reverse Cuthill-McKee reordering. The initial commits were written by @pratikvn.
This PR also implements the unordered reverse Cuthill-McKee algorithm (see [here](https://ieeexplore.ieee.org/document/7013062), algorithm 6) for the OpenMP executor.

This is now ready to review. Some things I already thought about while integrating:
 - ~~Is passing vectors by reference, as is sometimes done here, in line with ginkgos coding guidelines?~~ Apparently yes.
 - ~~Is returning vectors by value, as is done here at least once, in line with ginkgos coding guidelines?~~ Apparently yes.
 - ~~Should there maybe be ... tests for the (main) test? It is one complex function after all.~~ Apparently no.

**TODO:**
+ [x] Rewrite OMP kernel
+ [x] Add proper documentation.
+ [x] Fix the templated tests.
+ [x] ~~Add an example to `examples/` to demonstrate usage.~~

PR #504, opened 2020-04-21
Add ExecutorAllocator

upsj:
This PR adds a C++ standard-library compatible allocator and uses it in all places where we use standard-library containers.

Additionally, it improves the operation logging for benchmarks by allowing the separate recording of nested operations (kernels -> allocate/free), which this PR will introduce in many additional places.

This was based on a discussion we had on #500

### DIFF
--- a/benchmark/solver/solver.cpp
+++ b/benchmark/solver/solver.cpp
@@ -81,8 +81,7 @@ DEFINE_bool(overhead, false,
 
 
 // input validation
-[[noreturn]] void print_config_error_and_exit()
-{
+[[noreturn]] void print_config_error_and_exit() {
     std::cerr << "Input has to be a JSON array of matrix configurations:\n"
               << "  [\n"
               << "    { \"filename\": \"my_file.mtx\",  \"optimal\": { "
@@ -318,7 +317,8 @@ void solve_system(const std::string &solver_name,
             // slow run, get the time of each functions
             auto x_clone = clone(x);
 
-            auto gen_logger = std::make_shared<OperationLogger>(exec);
+            auto gen_logger =
+                std::make_shared<OperationLogger>(exec, FLAGS_nested_names);
             exec->add_logger(gen_logger);
 
             auto precond = precond_factory.at(precond_name)(exec);
@@ -339,7 +339,8 @@ void solve_system(const std::string &solver_name,
                                    solver_json["preconditioner"], allocator);
             }
 
-            auto apply_logger = std::make_shared<OperationLogger>(exec);
+            auto apply_logger =
+                std::make_shared<OperationLogger>(exec, FLAGS_nested_names);
             exec->add_logger(apply_logger);
 
             solver->apply(lend(b), lend(x_clone));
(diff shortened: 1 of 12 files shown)


benchmark/solver/solver.cpp
```diff
@@ -81,8 +81,7 @@ DEFINE_bool(overhead, false,
 
 
 // input validation
-[[noreturn]] void print_config_error_and_exit()
-{
+[[noreturn]] void print_config_error_and_exit() {
```
upsj, line 84: We should do something about this formatting flip-flopping

pratikvn: Is it due to the combination of the `[[noreturn]]` and the `clang-format` version ?

upsj: Likely yes, this formatting has been changed a few times in the last months, see git blame ;)

yhmtsai: @upsj Do you use clang-format 6?

upsj: yes

core/base/allocator.hpp
```diff
@@ -0,0 +1,147 @@
...
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+#ifndef GKO_CORE_BASE_ALLOCATOR_HPP_
+#define GKO_CORE_BASE_ALLOCATOR_HPP_
+
+
+#include <ginkgo/core/base/executor.hpp>
+#include <memory>
```
thoasm, line 38: nit: 2 empty lines missing.
Also, do you consider `executor.hpp` to be directly related to this file? I can see an argument for both, but I slightly lean towards not directly related, in which case it should be included after the standard library headers.

core/base/allocator.hpp
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+private:
+    std::shared_ptr<const Executor> exec_;
+};
+
+
+// Convenience type alias
+template <typename T>
+using vector = std::vector<T, ExecutorAllocator<T>>;
```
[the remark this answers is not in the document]
upsj: Mostly because we don't use them in many places, but since their declaration is rather complicated, you are probably correct that they should be moved

reference/matrix/csr_kernels.cpp
```diff
@@ -129,9 +130,19 @@ void advanced_spmv(std::shared_ptr<const ReferenceExecutor> exec,
...
 
+template <typename Key>
+using unordered_set =
+    std::unordered_set<Key, std::hash<Key>, std::equal_to<Key>,
+                       gko::ExecutorAllocator<Key>>;
+
+template <typename Key, typename Value>
+using unordered_map =
+    std::unordered_map<Key, Value, std::hash<Key>, std::equal_to<Key>,
+                       gko::ExecutorAllocator<std::pair<const Key, Value>>>;
```
yhmtsai, line 141: I think they are general implementation of `unordered_set` and `unordered_map` on `gko::Allocator`, right?
Probably move them into `allocator.hpp`?

PR #506, opened 2020-04-22
Add vector copy to ilu preconditioner.

fritzgoebel:
This vector copy ensures linearity of the preconditioner when iterative solvers are used for the triangular systems.

Without the additional copy, the solution vector is used as initial guess to the second solver, no matter its content. When only doing a few iterations, this initial guess still has a strong influence on the result and, depending what is in the respective initial guess vecotrs, we cannot assume the following equality:
prec(x_1) + prec(x_2) = prec(x_1 + x_2)

(here, prec(x) means applying the preconditioner to a vector x)

Especially for preconditioned GMRES, this leads to errors in the result.

### DIFF
--- a/include/ginkgo/core/preconditioner/ilu.hpp
+++ b/include/ginkgo/core/preconditioner/ilu.hpp
@@ -176,9 +176,11 @@ protected:
         set_cache_to(b);
         if (!ReverseApply) {
             l_solver_->apply(b, cache_.intermediate.get());
+            x->copy_from(cache_.intermediate.get());
             u_solver_->apply(cache_.intermediate.get(), x);
         } else {
             u_solver_->apply(b, cache_.intermediate.get());
+            x->copy_from(cache_.intermediate.get());
             l_solver_->apply(cache_.intermediate.get(), x);
         }
     }


hartwiganzt:
@pratikvn Yes, I would argue that the overhead is acceptale.

include/ginkgo/core/preconditioner/ilu.hpp
```diff
@@ -176,9 +176,11 @@ class Ilu : public EnableLinOp<
         set_cache_to(b);
         if (!ReverseApply) {
             l_solver_->apply(b, cache_.intermediate.get());
+            x->copy_from(cache_.intermediate.get());
```
yhmtsai, line 179: I think we copy b to cache to make sure the cache is not random filled before.
Should the cache should be get data from x not b here?
first apply will use x as initial guess
second apply will use cache as initial guess

fritzgoebel: I'm not sure that I completely understand the question, but I'm fairly certain that the data from x should not be used since it can be any vector, e.g. the krylo vector from a previous iteration. This doesn't matter for exact triangular solves since they overwrite x with the solution anyway. However, when using iterative triangular solves, this causes problems. The first apply solves:
Ly = b
By setting the cache to b, we use the right hand side as initial guess. Now the second apply solves
Ux = y
and by copying x from the cache, we again use the right hand side as initial guess.

pratikvn: But I agree with @yhmtsai that `x` might be a better choice because, in the case of the solving `Ly=b` you are kind of hard-coding the init guess to `b`. But if you set the cache to `x`, would it not be possible for the function calling the preconditioner apply to set the initial guess and if you need it then you can set it to `b`, right ?

hartwiganzt:
We are applying a preconditioner to the Krylov vector `s` to get the preconditioned Krylov search direction `p`. So, in fact, we want to solve `Ap=s` - or more precisely `LUp=s`.

Now, applying a direct solver, we would just solve `Lt=s` (forward) and `Up=t` (backward) with `t` being some temporary vector. 

If we want to iteratively, we use some iteration method `L_method->apply(s,t)`, but here the initial guess plays a role. A smart initial guess for `t` is the right-hand side `s`. 
The same for the upper solve:  `U_method->apply(t,p)` and `t` is a good initial guess for `p`.

include/ginkgo/core/preconditioner/ilu.hpp
fritzgoebel: I see what you mean. However, using the data from x as initial guess for the first solve introduces the same nonlinearity as before, only in a different spot. This way, we would have to ensure linearity of the preconditioner, a "linear operator", from outside the preconditioner before calling it, by putting adequate restrictions to x. 
I checked and tried GMRES with the following configurations (without writing extra ode in GMRESpreparing the data in x):
- only `set_cache_to(b)`: this was the initial version, and gives errors with iterative triangular solves
- only `set_cache_to(x)`: also gives errors in the result
- `set_cache_to(x)` and `x->copy_from(cache_.intermediate.get())`: also gives errors in the result
- `set_cache_to(b)` and `x->copy_from(cache_.intermediate.get())`: This is the only configuration for which the result GMRES computes is correct when using a few block Jacobi sweeps as triangular solves in the preconditioner.

upsj:
I withdraw my concerns, but we should definitely mention that `LinOp::apply(b, x)` might still use the values of `x`.

hartwiganzt:
> I withdraw my concerns, but we should definitely mention that `LinOp::apply(b, x)` might still use the values of `x`.

Please add a comment where you feel it is appropriate!

include/ginkgo/core/preconditioner/ilu.hpp
fritzgoebel: I typed up an explanation, hopefully this clarifies:
explanation

PR #508, opened 2020-04-23
Extended sparsity pattern for ISAI

upsj:
This PR adds the capability to extend the sparsity pattern for ISAI from M to M^n. This works by repeatedly multiplying M by I (stored with the sparsity pattern of M!)

For this to work, we also need to require that the output of an SpGEMM is sorted (i.e. switching from hashmaps to searchtrees, the sparselibs seem to take care of that as well) and fix an issue with gko::as and smart pointers.


core/preconditioner/isai.cpp
```diff
@@ -97,39 +98,79 @@ convert_to_csr_and_sort(std::shared_ptr<const Executor> &exec, const LinOp *mtx,
...
+    auto strategy = csr_u->get_strategy();
+    auto csr_extended_u = extend_sparsity(exec, csr_u, power, false);
+    const auto num_elems = csr_extended_u->get_num_stored_elements();
 
     std::shared_ptr<Csr> inverted_u =
-        Csr::create(exec, csr_u->get_size(), num_elems, csr_u->get_strategy());
-    exec->run(isai::make_generate_u_inverse(csr_u.get(), inverted_u.get()));
+        Csr::create(exec, csr_extended_u->get_size(), num_elems, strategy);
+    exec->run(
+        isai::make_generate_u_inverse(csr_extended_u.get(), inverted_u.get()));
```
thoasm, line 173: What we could do alternatively is to forward the `csr_u` to the `generate_u_inverse`, and use the `csr_extended_u` as the template for the `inverted_u`.
We would need to remove the copying of the sparsity information inside the `generate_x_inverse`, and change the `clone_allocations` in the reference implementation to not override the sparsity pattern.

If you want, I can push these changes directly into this PR. The OMP and CUDA/HIP implementations would also be easy to change to that. PR #502 would then, however, depend on this PR to finish.

upsj: That is a really great idea, since otherwise we would mostly be reading zeros. Similar to what we do in ParILU vs. ParILUT

thoasm: Now that I think of it, it is probably better to do that at a later date, so we can move forward for now. This is just a small optimization, we could actually do that together with #511

core/preconditioner/isai.cpp
```diff
@@ -97,39 +98,79 @@ convert_to_csr_and_sort(std::shared_ptr<const Executor> &exec, const LinOp *mtx,
     if (!skip_sorting) {
         copy->sort_by_column_index();
     }
-    return {copy.release(), std::default_delete<const Csr>{}};
+    return copy;
```
thoasm, line 101: The intel compiler does not like the implicit conversion from `unique_ptr` to `shared_ptr` here (see [the pipeline for intel](https://gitlab.com/ginkgo-project/ginkgo-public-ci/pipelines/139151360)).
You could use the [(13.) constructor of shared_ptr](https://en.cppreference.com/w/cpp/memory/shared_ptr/shared_ptr), which uses an rvalue reference to a `unique_ptr` as a parameter. That should resolve the compilation issue.
```suggestion
    return {std::move(copy)};
```

thoasm: Alternatively, you can also generate `copy` as a `shared_ptr` (do the implicit conversion there):
```c++
auto copy = gko::share(Csr::create(exec));
```
or
```c++
std::shared_ptr<Csr> copy = Csr::create(exec);
```

common/matrix/csr_kernels.hpp.inc
```diff
@@ -860,6 +860,36 @@ __global__ __launch_bounds__(default_block_size) void fill_in_hybrid(
...
+    const IndexType *__restrict__ col_idxs, IndexType num_rows, bool *flag)
+{
+    __shared__ bool sh_flag;
+    auto block = group::this_thread_block();
+    if (block.thread_rank() == 0) {
+        sh_flag = *flag;
+    }
+    block.sync();
+
+    auto row = threadIdx.x + blockDim.x * static_cast<IndexType>(blockIdx.x);
```
upsj, line 875:
```suggestion
    auto row = thread::get_thread_id_flat<IndexType>();
```

PR #509, opened 2020-04-23
Add GDB pretty printer for gko::Array

upsj:
This PR adds a GDB pretty printer for gko::Arrays that can help debugging CPU code by outputting the array contents.

We could also collect ideas for other types that could use a pretty printer, in case anyone else here uses GDB ;)

Before:
```
(gdb) print x->values_
$2 = {num_elems_ = 500, data_ = std::unique_ptr<double []> = {get() = 0x562134b10680}, exec_ = std::shared_ptr<const gko::Executor> (use count 31, weak count 1) = {
    get() = 0x56212b088950}}
(gdb) print d_x->values_
$3 = {num_elems_ = 500, data_ = std::unique_ptr<double []> = {get() = 0x1303fa1600}, exec_ = 
std::shared_ptr<const gko::Executor> (use count 17, weak count 1) = {
    get() = 0x56213b2fb030}}
```
After:
```
(gdb) source ../dev_tools/scripts/gdb-ginkgo.py 
(gdb) print x->values_
$4 = gko::Array<double> of length 500 on gko::ReferenceExecutor (0x562134b10680) = {0.1617967567022226, 0.78603518073255585, -0.62451741825935325, 0.95091814339325187, 
  ...
(gdb) print d_x->values_
$5 = gko::Array<double> of length 500 on gko::CudaExecutor (0x1303fa1600)
```

### DIFF
--- a/dev_tools/scripts/gdb-ginkgo.py
+++ b/dev_tools/scripts/gdb-ginkgo.py
@@ -1,0 +1,130 @@
+# Pretty-printers for Ginkgo
+# Based on the pretty-printers for libstdc++.
+
+# Copyright (C) 2008-2020 Free Software Foundation, Inc.
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


upsj:
@tcojean Good idea, I added a few sentences to the [Developer Homepage](https://github.com/ginkgo-project/ginkgo/wiki/Developers-Homepage) in the Wiki

PR #510, opened 2020-04-23
Fix the windows cuda of GitHub action

yhmtsai:
The updated CMake (>= 3.17.0) can find the location of `CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES` which is not found in the older CMake version (< 3.17.0)

This PR handles `CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES` and `CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES` individually to fix this problem.

### DIFF
--- a/cuda/CMakeLists.txt
+++ b/cuda/CMakeLists.txt
@@ -9,9 +9,11 @@ endif()
 if(MSVC)
     # MSVC can not find CUDA automatically
     # Use CUDA_COMPILER PATH to define the CUDA TOOLKIT ROOT DIR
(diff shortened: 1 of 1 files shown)


yhmtsai:
Yes, `CMAKE_CUDA_ROOT_DIR` is our own variable, so I think it does not hurt anything.

hartwiganzt:
@yhmtsai  thanks for fixing this! I think this can be merged.

yhmtsai:
Yes, I think so because it does not affect linux compile

PR #512, opened 2020-04-24
Fix CUDA ISAI test by adding cuda include files during build

thoasm:
I forgot that `cuda/base/config.hpp` internally includes `thrust/complex.h`, which requires the CUDA headers to be present. This PR simply changes the create function of the test to one which forwards these headers to the compiler.

### DIFF
--- a/cuda/test/preconditioner/CMakeLists.txt
+++ b/cuda/test/preconditioner/CMakeLists.txt
@@ -1,2 +1,2 @@
 ginkgo_create_test(jacobi_kernels)
-ginkgo_create_test(isai_kernels)
+ginkgo_create_test_cpp_cuda_header(isai_kernels)


thoasm:
@upsj Yes, I need it for `warp_size`. For HIP, I use the file [`isai_kernels.hip.cpp`](https://github.com/ginkgo-project/ginkgo/blob/develop/hip/test/preconditioner/isai_kernels.hip.cpp) and use `ginkgo_create_hip_test` in the [CMakeLists.txt](https://github.com/ginkgo-project/ginkgo/blob/develop/hip/test/preconditioner/CMakeLists.txt), which will be compiled with HIP directly (needed because the [hip/base/config.hip.hpp](https://github.com/ginkgo-project/ginkgo/blob/develop/hip/base/config.hip.hpp#L37) includes `hip/runtime.h`).

PR #513, opened 2020-04-25
Templatize the examples.

pratikvn:
This PR adds template parameters to the declarations of the objects in the examples making it easier to change the precisions with the `using` statements in the beginning of the `main` function.

Only the `minimal-cuda-solver` is left as it is so that we demonstrate that we do have the defaults in place.


examples/custom-matrix-format/custom-matrix-format.cpp
```diff
@@ -42,8 +42,9 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 // A CUDA kernel implementing the stencil, which will be used if running on the
 // CUDA executor. Unfortunately, NVCC has serious problems interpreting some
 // parts of Ginkgo's code, so the kernel has to be compiled separately.
-extern void stencil_kernel(std::size_t size, const double *coefs,
-                           const double *b, double *x);
+template <typename ValueType>
+extern void stencil_kernel(std::size_t size, const ValueType *coefs,
+                           const ValueType *b, ValueType *x);
```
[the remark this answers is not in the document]
pratikvn: You would still need to compile the cuda code (calling the kernel with the <<<>>>) with nvcc , right ? I can remove the extern, I guess.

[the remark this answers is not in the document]
upsj: Ah I see your point, we also have OpenMP code in that file, so it probably wouldn't make sense to compile everything with nvcc. But yes, it might be possible to remove the `extern`.

examples/poisson-solver/poisson-solver.cpp
```diff
@@ -62,13 +63,14 @@ void generate_stencil_matrix(gko::matrix::Csr<> *matrix)
...
-void generate_rhs(Closure f, double u0, double u1, gko::matrix::Dense<> *rhs)
+template <typename Closure, typename ValueType>
+void generate_rhs(Closure f, ValueType u0, ValueType u1,
+                  gko::matrix::Dense<ValueType> *rhs)
 {
     const auto discretization_points = rhs->get_size()[0];
     auto values = rhs->get_values();
     const auto h = 1.0 / (discretization_points + 1);
-    for (int i = 0; i < discretization_points; ++i) {
+    for (size_type i = 0; i < discretization_points; ++i) {
```
upsj, line 73: My bad, this should probably be 
```suggestion
    for (gko::size_type i = 0; i < discretization_points; ++i) {
```

pratikvn: Sorry, I was too lazy :)

examples/custom-stopping-criterion/custom-stopping-criterion.cpp
```diff
@@ -93,9 +93,12 @@ void run_solver(volatile bool *stop_iteration_process,
...
     // Some shortcuts
-    using mtx = gko::matrix::Csr<>;
-    using vec = gko::matrix::Dense<>;
-    using bicg = gko::solver::Bicgstab<>;
+    using ValueType = double;
+    using IndexType = int;
+
+    using mtx = gko::matrix::Csr<ValueType, IndexType>;
+    using vec = gko::matrix::Dense<ValueType>;
+    using bicg = gko::solver::Bicgstab<ValueType>;
```
tcojean, line 101: further in the file, `gko::log::Stream<>` Also needs to be changed.

examples/ilu-preconditioned-solver/ilu-preconditioned-solver.cpp
```diff
@@ -92,10 +97,10 @@ int main(int argc, char *argv[])
     // if there are several very similar systems to solve, and the same
     // solver+preconditioner combination is expected to be effective.
     auto ilu_gmres_factory =
-        gko::solver::Gmres<>::build()
+        gmres::build()
             .with_criteria(
                 gko::stop::Iteration::build().with_max_iters(1000u).on(exec),
-                gko::stop::ResidualNormReduction<>::build()
+                gko::stop::ResidualNormReduction<ValueType>::build()
                     .with_reduction_factor(1e-15)
```
tcojean, line 104: This probably needs to be adaptated. You can use something like the r<valuetype>::value from the tests.

tcojean: This is the same basically everywhere, not just here.

examples/nine-pt-stencil-solver/nine-pt-stencil-solver.cpp
```diff
@@ -325,7 +335,7 @@ int main(int argc, char *argv[])
                  values.data(), rhs.data(), u.data(), 1e-12);
```
tcojean, line 335: Don't forget to also adapt this one.

examples/papi-logging/papi-logging.cpp
```diff
@@ -160,7 +163,7 @@ int main(int argc, char *argv[])
         cg::build()
             .with_criteria(
                 gko::stop::Iteration::build().with_max_iters(20u).on(exec),
-                gko::stop::ResidualNormReduction<>::build()
+                gko::stop::ResidualNormReduction<ValueType>::build()
                     .with_reduction_factor(1e-20)
```
tcojean, line 167: I don't remember why this is purposefully set high. Maybe we don't even need it anyway?

examples/nine-pt-stencil-solver/nine-pt-stencil-solver.cpp
```diff
@@ -284,35 +290,39 @@ int main(int argc, char *argv[])
                   << std::endl;
         std::exit(-1);
     }
+    using ValueType = double;
```
tcojean, line 293: Up here, line 273, `ResidualNormReduction<>` needs to be changed.

examples/performance-debugging/performance-debugging.cpp
```diff
@@ -343,8 +343,9 @@ int main(int argc, char *argv[])
     // Parametrize the benchmark here
     // Pick a value type
     using vtype = double;
+    using itype = int;
```
tcojean, line 346: Maybe for uniformization, also use ValueType and IndexType here.

tcojean: Down here, line `354` the reduction_factor also needs adaptation, although it's actually quite low.

tcojean: Up here, line `327`, missing Dense change for the `print_vector` function.

examples/twentyseven-pt-stencil-solver/twentyseven-pt-stencil-solver.cpp
```diff
@@ -334,6 +340,8 @@ void solve_system(const std::string &executor_string,
 
 int main(int argc, char *argv[])
 {
+    using ValueType = double;
```
tcojean, line 343: Up here, line `330`, missing `ResidualNormReduction<>` change. Same comment about the accuracy.

examples/ilu-preconditioned-solver/ilu-preconditioned-solver.cpp
pratikvn: Is it okay if I use a `ValueType(1e-15)` ? I dont want to complicate it with the `r::value` declarations. I guess to be uniform, I can use `1e-7` for everything as that makes sense for both `double` and `float`.

thoasm: You can use the C++ [`std::numeric_limits<ValueType>::digits10`](https://en.cppreference.com/w/cpp/types/numeric_limits/digits10), which should do basically the same that `r::value` does, just that it is from the standard library.

examples/custom-matrix-format/stencil_kernel.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+
+#define INSTANTIATE_FOR_EACH_VALUE_TYPE(_macro) \
+    template _macro(float);                     \
+    template _macro(double);
+
+
+#define STENCIL_KERNEL(_type)                                                 \
+    void stencil_kernel(std::size_t size, const _type *coefs, const _type *b, \
+                        _type *x);
```
thoasm, line 44: Why does this need an extra header? You only macros in this header, which should not be used in any other file. If you want an extra header, it should contain the definition of the stencil function of `examples/custom-matrix-format/custom-matrix-format.cpp`:
```
template <typename ValueType>
void stencil_kernel(std::size_t size, const ValueType *coefs,
                    const ValueType *b, ValueType *x);
```

examples/adaptiveprecision-blockjacobi/adaptiveprecision-blockjacobi.cpp
```diff
@@ -93,11 +92,11 @@ int main(int argc, char *argv[])
 
     // copy b again
     b->copy_from(host_x.get());
-
+    ValueType reduction_factor = 1e-7;
```
thoasm, line 95: Like I said in a comment, [`std::numeric_limit<ValueType>::epsilon`](https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon) should be more suitable here (hard-coding these in my opinion is bad practice):
```suggestion
    ValueType reduction_factor = std::numeric_limit<ValueType>::epsilon;
```

examples/adaptiveprecision-blockjacobi/adaptiveprecision-blockjacobi.cpp
```diff
@@ -93,11 +92,11 @@ int main(int argc, char *argv[])
...
     auto iter_stop =
         gko::stop::Iteration::build().with_max_iters(10000u).on(exec);
     auto tol_stop = gko::stop::ResidualNormReduction<ValueType>::build()
-                        .with_reduction_factor(static_cast<ValueType>(1e-12))
+                        .with_reduction_factor(reduction_factor)
```
thoasm, line 99: If this needs to be larger than `reduction_factor`, then maybe:
```suggestion
                        .with_reduction_factor(reduction_factor * 10e2)
```

examples/ilu-preconditioned-solver/ilu-preconditioned-solver.cpp
thoasm: Sorry, `digits10` is probably not the best here, [`std::numeric_limit<ValueType>::epsilon`](https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon) is a lot more fitting.

examples/performance-debugging/performance-debugging.cpp
```diff
@@ -324,7 +324,8 @@ void print_usage(const char *filename)
 }
 
 
-void print_vector(const gko::matrix::Dense<> *vec)
+template <typename ValueType>
```
tcojean, line 327: The only question left here concern the three functions `get_norm` and related on top of this file. We could make them return a `remove_complex` of `ValueType` instead of plain `double`, but if we don't want to complicate this too much as is should be fine.

examples/twentyseven-pt-stencil-solver/twentyseven-pt-stencil-solver.cpp
```diff
@@ -93,10 +93,12 @@ constexpr double default_delta = -1.0 / 24.0;
 
 // Creates a stencil matrix in CSR format for the given number of discretization
 // points.
-void generate_stencil_matrix(int dp, int *row_ptrs, int *col_idxs,
-                             double *values, double *coefs)
+template <typename ValueType, typename IndexType>
```
tcojean, line 96: In this and also the 9pt stencil example, line 82->85 we have all the default values declared as double. I'm not sure if in float that that creates issues due to the computed precision?

examples/performance-debugging/performance-debugging.cpp
thoasm: I kind of thought `ValueType` would only be `float` or `double`, but you are correct, if we also want to support `std::complex<float>` and `std::complex<double>`, we would also need the `remove_complex` for the type of the `reduction_factor`.

examples/nine-pt-stencil-solver/nine-pt-stencil-solver.cpp
```diff
@@ -202,18 +207,19 @@ double calculate_error(int dp, const double *u, Closure correct_u)
...
-    using val_array = gko::Array<double>;
-    using idx_array = gko::Array<int>;
+    using vec = gko::matrix::Dense<ValueType>;
+    using mtx = gko::matrix::Csr<ValueType, IndexType>;
+    using cg = gko::solver::Cg<ValueType>;
+    using bj = gko::preconditioner::Jacobi<ValueType, IndexType>;
+    using val_array = gko::Array<ValueType>;
+    using idx_array = gko::Array<IndexType>;
     const auto &dp = discretization_points;
     const size_t dp_2 = dp * dp;
```
thoasm, line 224: nit:
```suggestion
    const gko::size_type dp_2 = dp * dp;
```
Since it is used in `gko::dim`

examples/performance-debugging/performance-debugging.cpp
tcojean: Yes that's true, whenever we think of complex valuetypes, then we need remove complex for norms and reduction factors. Another thing to look out for is the matrix initialization sometimes. 

In general, Ginkgo should support complex computations (definitely reference kernels do). Of course not all of CUDA supports complex, but since we are going this direction ensuring that it works for these would be great. I'm fairly certain some users will try out complex in the examples at some point, so it's better if we tried it ourselves first.

examples/nine-pt-stencil-solver/nine-pt-stencil-solver.cpp
```diff
@@ -284,48 +290,52 @@ int main(int argc, char *argv[])
...
+    ValueType reduction_factor = 1e-7;
 
+    auto start_time = std::chrono::steady_clock::now();
     solve_system(executor_string, dp, row_ptrs.data(), col_idxs.data(),
-                 values.data(), rhs.data(), u.data(), 1e-12);
-
+                 values.data(), rhs.data(), u.data(), reduction_factor);
     auto stop_time = std::chrono::steady_clock::now();
-    double runtime_duration =
+    ValueType runtime_duration =
```
thoasm, line 338: nit: The time type should not be impacted by `ValueType`.
```suggestion
    auto runtime_duration =
```

examples/twentyseven-pt-stencil-solver/twentyseven-pt-stencil-solver.cpp
```diff
@@ -93,10 +93,12 @@ constexpr double default_delta = -1.0 / 24.0;
...
+void generate_stencil_matrix(IndexType dp, IndexType *row_ptrs,
+                             IndexType *col_idxs, ValueType *values,
+                             ValueType *coefs)
 {
-    int pos = 0;
+    IndexType pos = 0;
     size_t dp_2 = dp * dp;
```
thoasm, line 102: nit: this variable is unused.

examples/twentyseven-pt-stencil-solver/twentyseven-pt-stencil-solver.cpp
```diff
@@ -167,7 +171,7 @@ void generate_rhs(int dp, Closure f, ClosureT u, double *rhs, double *coefs)
         }
     }
 
-    // To avoid double counting we have to check if our previous calculations
+    // To avoid ValueType counting we have to check if our previous calculations
```
thoasm, line 174: Pretty sure this was a replace mistake, `double` was correct in this case.

examples/adaptiveprecision-blockjacobi/adaptiveprecision-blockjacobi.cpp
thoasm: Also, `reduction_factor` should be `const` everywhere.

examples/performance-debugging/performance-debugging.cpp
thoasm: Then we need to test every single example for compatibility with `std::complex<double>` since I am fairly certain we are missing some `gko::remove_complex<>` in the current examples code.
I think for now, checking if it compiles should be enough (to detect the missing `remove_complex` parts), but at some point, we need to test our whole library for compatibility with complex numbers.

examples/adaptiveprecision-blockjacobi/adaptiveprecision-blockjacobi.cpp
pratikvn: I dont think using `numeric_limits<>::epsilon()` makes sense here, because that says, atleast to me that the reduction_factor has to be the epsilon of that type, which is not true. 1e-7 makes perfect sense even if you are using just `double` . And it is definitely something the user would like to tweak.

examples/adaptiveprecision-blockjacobi/adaptiveprecision-blockjacobi.cpp
pratikvn: Sorry, but I dont understand why it should be `reduction_factor * 10e2` . It is for the CG solver. So, I think a normal `reduction_factor` should be valid.

examples/performance-debugging/performance-debugging.cpp
pratikvn: Yes, this makes sense. The problem is it might make some examples verbose. Maybe I can add it only for a few examples ? Or should I add it for all ?

examples/twentyseven-pt-stencil-solver/twentyseven-pt-stencil-solver.cpp
pratikvn: the coefs are stored as `ValueType` in line 355->374. So, I think it should be fine.

examples/custom-matrix-format/stencil_kernel.hpp
pratikvn: Yes, you are right. Thanks!

examples/adaptiveprecision-blockjacobi/adaptiveprecision-blockjacobi.cpp
thoasm: My bad, I thought `reduction_factor` was used for something else here as well, but apparently, it is only used for the stopping criterion.

pratikvn:
This should be ready for review again.

PR #515, opened 2020-04-27
Example for GPU kernel operations

upsj:
This PR is an example for how we could unify the kernel launch process between CUDA and HIP, and improve the portability (multi-stream!) and profilability (benchmarking) of individual GPU kernels.
If there are no major problems, this could allow us to merge most CUDA and HIP files outside `*/base` and `*/components`. The current syntax is of course only a simple example with much improvement potential.
@tcojean rightly mentioned that there might be a certain overhead associated with the logger invocations, so in case we launch a huge number of small kernels, this overhead might become noticable.

So Pros:
* Unifies the code base
* More portability (multiple CUDA/HIP streams)
* More logging potential
* No more need to use `as_cuda_type`/`as_hip_type` for complex types

Cons:
* Performance penalty?
* More verbose? (explicit template parameters, kernel name) We might be able to hide this in a macro.

### DIFF
--- a/cuda/base/executor.cuh
+++ b/cuda/base/executor.cuh
@@ -1,0 +1,60 @@
+/*******************************<GINKGO LICENSE>******************************
+Copyright (c) 2017-2020, the Ginkgo authors
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+
+1. Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+
+2. Redistributions in binary form must reproduce the above copyright
+notice, this list of conditions and the following disclaimer in the
+documentation and/or other materials provided with the distribution.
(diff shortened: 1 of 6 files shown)


cuda/base/executor.cuh
```diff
@@ -0,0 +1,60 @@
...
+
+#include "cuda/base/types.hpp"
+
+
+namespace gko {
+
+
+template <typename Kernel, typename... Args>
+void CudaExecutor::run_gpu(const char *kernel_name, Kernel kernel,
+                           size_type num_blocks, size_type block_size,
```
yhmtsai, line 47: should be dim3?

upsj: Yes, this is just a first sketch, I would probably provide an overload for `size_type` and `dim3`

upsj:
This is outdated and will most likely be superseded by a more generalized setup.

PR #516, opened 2020-04-28
Simplify executor copy interface

upsj:
This PR adds an executor-local `copy` function and a single-element `copy_to_master` that simplifies the common
```cpp
Type result{};
exec->get_master()->copy_from(exec.get(), 1, data, &result);
// and
exec->copy_from(exec.get(), size, src, target);
```
patterns into
```cpp
auto result = exec->copy_val_to_master(data);
// and
exec->copy(size, src, target);
```

### DIFF
--- a/hip/matrix/csr_kernels.hip.cpp
+++ b/hip/matrix/csr_kernels.hip.cpp
@@ -556,8 +556,7 @@ void spgeam(syn::value_list<int, subwarp_size>,
 
     // accumulate non-zeros for alpha * A + beta * B
     matrix::CsrBuilder<ValueType, IndexType> c_builder{c};
-    IndexType c_nnz{};
-    exec->get_master()->copy_from(exec.get(), 1, c_row_ptrs + m, &c_nnz);
+    auto c_nnz = exec->copy_val_to_host(c_row_ptrs + m);
     c_builder.get_col_idx_array().resize_and_reset(c_nnz);
     c_builder.get_value_array().resize_and_reset(c_nnz);
     auto c_col_idxs = c->get_col_idxs();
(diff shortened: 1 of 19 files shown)


hip/matrix/csr_kernels.hip.cpp
```diff
@@ -650,22 +649,19 @@ void advanced_spgemm(std::shared_ptr<const HipExecutor> exec,
...
-                      d_row_ptrs, d_col_idxs, d_vals, c);
+        select_spgeam(
+            spgeam_kernels(),
+            [&](int compiled_subwarp_size) {
+                return compiled_subwarp_size >= nnz_per_row ||
+                       compiled_subwarp_size == config::warp_size;
+            },
+            syn::value_list<int>(), syn::type_list<>(), exec, valpha,
+            c_tmp_row_ptrs, c_tmp_col_idxs, c_tmp_vals, vbeta, d_row_ptrs,
+            d_col_idxs, d_vals, c);
```
yhmtsai, line 664: This is another part which is different from different clang-format

pratikvn:
But maybe you can add some tests to be sure ?

include/ginkgo/core/base/executor.hpp
```diff
@@ -536,6 +536,40 @@ class Executor : public log::EnableLogging<Executor> {
...
+     * Retrieves a single element at the given location from executor memory.
+     *
+     * @tparam T  datatype to copy
+     *
+     * @param ptr  the pointer to the element to be copied
+     *
+     * @return the value stored at ptr
+     */
+    template <typename T>
+    T copy_to_master(const T *ptr) const
```
thoasm, line 566: I would prefer a name that tells that the copy only happens for a single element (so it is not mixed up with `copy`, but for different accessors):
```suggestion
    T copy_element_to_master(const T *ptr) const
```

Or
```suggestion
    T copy_elem_to_master(const T *ptr) const
```

upsj: Good point! I went with `copy_val_to_host`

PR #517, opened 2020-04-28
Fix CI core count

upsj:
Currently we determine the number of (logical) cores on the CI system incorrectly:

On fineci:
```
> grep "core id" /proc/cpuinfo | sort -u
core id         : 0
core id         : 1
core id         : 10
core id         : 11
core id         : 12
core id         : 2
core id         : 3
core id         : 4
core id         : 8
core id         : 9
```

The Core IDs doesn't seem suitable to determine the number of cores, since they contain duplicates.
**As far as I can see, this also causes our builds on both machines to be executed with only a single core.**

A better choice would be
```
> grep "^processor" /proc/cpuinfo | sort -u
processor       : 0
processor       : 1
processor       : 10
processor       : 11
processor       : 12
processor       : 13
processor       : 14
processor       : 15
processor       : 16
processor       : 17
processor       : 18
processor       : 19
processor       : 2
processor       : 20
processor       : 21
processor       : 22
processor       : 23
processor       : 24
processor       : 25
processor       : 26
processor       : 27
processor       : 28
processor       : 29
processor       : 3
processor       : 30
processor       : 31
processor       : 32
processor       : 33
processor       : 34
processor       : 35
processor       : 36
processor       : 37
processor       : 38
processor       : 39
processor       : 4
processor       : 5
processor       : 6
processor       : 7
processor       : 8
processor       : 9
```

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -25,8 +25,7 @@ stages:
   EXTRA_CMAKE_FLAGS: ""
 
 .before_script_template: &default_before_script
-  - export NUM_CORES=$(grep -E "^processor" /proc/cpuinfo | sort -u | wc -l)
-  - export NUM_CORES=$((NUM_CORES/10))
+  - export NUM_CORES=4
   - export OMP_NUM_THREADS=${NUM_CORES}
   - export CUDA_VISIBLE_DEVICES=0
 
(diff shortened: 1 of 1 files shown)


upsj:
Looking at the numbers again, I'm not convinced that a blanket `NUM_CORES/10` is the best solution, especially on amdci. How about using 4 cores everywhere and setting the gitlab-runner concurrency to around 10-15 and 6-9 so we have at most 50% oversubscription?

tcojean:
I'm fine with setting 4 cores everywhere and manage the complexity in the runner's configuration.

Also, to get the number of cores, another solution could be:
```bash
lscpu | sed -n 's/^CPU(s):[\t ]\+\([0-9]\+\)$/\1/p'
```

PR #49, opened 2018-05-10
Better support for custom value types

gflegar:
This PR adds fixes so Ginkgo (and all its modules - reference, CPU, GPU) can  be compiled with custom value types. I can confirm it compiles with [FloatX](https://github.com/oprecomp/FloatX).

However, we still do not support most of the GPU  implementations that rely on cuBLAS and cuSPARSE. I only added implementations of the dense BLAS 1 kernels, as this should allow us to run all the solvers as long as the system matrix is not `Dense` or `Csr`.

Note that we are not testing the custom implementation of the dot by default as it can only be tested with a custom value type, and we don't have any of those in Ginkgo. But if anyone wants to test it, you can easily add  `false &&` to the if statement in the implementation of the kernel, so the custom version gets selected instead of the cuBLAS one.

This partially implements #53 by adding `scale`, `add_scaled` and `compute_dot` implementations for custom datatypes.

--------------------------------

## How to test this with FloatX

Clone the [oprecomp/floatx](https://github.com/oprecomp/floatx) repository to the `third_party` directory:

```bash
git clone --depth=1 git@github.com:oprecomp/floatx.git -b develop third_party/floatx
```

Download [this patch](https://github.com/ginkgo-project/ginkgo/files/2032187/floatx.patch.txt) and apply it:

```bash
git apply floatx.patch.txt
```

You may want to modify one of the examples to use floatx. The simplest way is to download 
[this patch](https://github.com/ginkgo-project/ginkgo/files/2032205/floatx_example.patch.txt) (which will modify the `simple_solver` example) and apply it:

```bash
git apply floatx_example.patch.txt
```

PR #521, opened 2020-04-29
Add precision conversion functions

upsj:
This PR adds precision conversion support to the Ginkgo matrix formats

TODO:
- [x] Implement array conversions
- [x] Implement matrix conversions

### DIFF
--- a/core/base/array.cpp
+++ b/core/base/array.cpp
@@ -43,7 +43,7 @@ namespace gko {
 namespace conversion {
 
 
-GKO_REGISTER_OPERATION(convert, convert_precision);
+GKO_REGISTER_OPERATION(convert, components::convert_precision);
 
 
 }  // namespace conversion
--- a/core/base/array.cpp
+++ b/core/base/array.cpp
@@ -1,0 +1,71 @@
+/*******************************<GINKGO LICENSE>******************************
+Copyright (c) 2017-2020, the Ginkgo authors
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+
+1. Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+
+2. Redistributions in binary form must reproduce the above copyright
+notice, this list of conditions and the following disclaimer in the
+documentation and/or other materials provided with the distribution.
+
+3. Neither the name of the copyright holder nor the names of its
+contributors may be used to endorse or promote products derived from
+this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS
+IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED
+TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A
+PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+#include <ginkgo/core/base/array.hpp>
+
+
+#include <ginkgo/core/base/math.hpp>
+
+
+#include "core/components/precision_conversion.hpp"
+
+
+namespace gko {
+namespace conversion {
+
+
+GKO_REGISTER_OPERATION(convert, convert_precision);
+
+
+}  // namespace conversion
+
+
+namespace detail {
+
+
+template <typename SourceType, typename TargetType>
+void convert_data(std::shared_ptr<const Executor> exec, size_type size,
+                  const SourceType *src, TargetType *dst)
+{
+    exec->run(conversion::make_convert(size, src, dst));
+}
+
+
+#define GKO_DECLARE_ARRAY_CONVERSION(From, To)                              \
+    void convert_data<From, To>(std::shared_ptr<const Executor>, size_type, \
+                                const From *, To *)
+
+GKO_INSTANTIATE_FOR_EACH_VALUE_CONVERSION(GKO_DECLARE_ARRAY_CONVERSION);
+
+
+}  // namespace detail
+}  // namespace gko
--- a/core/device_hooks/common_kernels.inc.cpp
+++ b/core/device_hooks/common_kernels.inc.cpp
@@ -68,6 +68,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 namespace gko {
 namespace kernels {
 namespace GKO_HOOK_MODULE {
+namespace components {
 
 
 template <typename SourceType, typename TargetType>
@@ -75,13 +76,15 @@ GKO_DECLARE_CONVERT_PRECISION_KERNEL(SourceType, TargetType)
 GKO_NOT_COMPILED(GKO_HOOK_MODULE);
 GKO_INSTANTIATE_FOR_EACH_VALUE_CONVERSION(GKO_DECLARE_CONVERT_PRECISION_KERNEL);
 
-
 template <typename IndexType>
 GKO_DECLARE_PREFIX_SUM_KERNEL(IndexType)
 GKO_NOT_COMPILED(GKO_HOOK_MODULE);
 GKO_INSTANTIATE_FOR_EACH_INDEX_TYPE(GKO_DECLARE_PREFIX_SUM_KERNEL);
 // explicitly instantiate for size_type, as this is used in the SellP format
 template GKO_DECLARE_PREFIX_SUM_KERNEL(size_type);
+
+
+}  // namespace components
 
 
 namespace dense {
--- a/core/device_hooks/common_kernels.inc.cpp
+++ b/core/device_hooks/common_kernels.inc.cpp
@@ -33,6 +33,8 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/base/exception_helpers.hpp>
 
 
+#include "core/components/precision_conversion.hpp"
+#include "core/components/prefix_sum.hpp"
 #include "core/factorization/factorization_kernels.hpp"
 #include "core/factorization/ilu_kernels.hpp"
 #include "core/factorization/par_ilu_kernels.hpp"
@@ -66,6 +68,22 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 namespace gko {
 namespace kernels {
 namespace GKO_HOOK_MODULE {
+
+
+template <typename SourceType, typename TargetType>
+GKO_DECLARE_CONVERT_PRECISION_KERNEL(SourceType, TargetType)
+GKO_NOT_COMPILED(GKO_HOOK_MODULE);
+GKO_INSTANTIATE_FOR_EACH_VALUE_CONVERSION(GKO_DECLARE_CONVERT_PRECISION_KERNEL);
+
+
+template <typename IndexType>
+GKO_DECLARE_PREFIX_SUM_KERNEL(IndexType)
+GKO_NOT_COMPILED(GKO_HOOK_MODULE);
+GKO_INSTANTIATE_FOR_EACH_INDEX_TYPE(GKO_DECLARE_PREFIX_SUM_KERNEL);
+// explicitly instantiate for size_type, as this is used in the SellP format
+template GKO_DECLARE_PREFIX_SUM_KERNEL(size_type);
+
+
 namespace dense {
 
 
--- a/core/matrix/csr.cpp
+++ b/core/matrix/csr.cpp
@@ -122,6 +122,68 @@ void Csr<ValueType, IndexType>::apply_impl(const LinOp *alpha, const LinOp *b,
 
 template <typename ValueType, typename IndexType>
 void Csr<ValueType, IndexType>::convert_to(
+    Csr<next_precision<ValueType>, IndexType> *result) const
+{
+    result->values_ = this->values_;
+    result->col_idxs_ = this->col_idxs_;
+    result->row_ptrs_ = this->row_ptrs_;
+    result->set_size(this->get_size());
+    auto strat = this->get_strategy().get();
+    using Other = Csr<next_precision<ValueType>, IndexType>;
+    std::shared_ptr<typename Other::strategy_type> new_strat;
+    // TODO clean this up as soon as we improve strategy_type
+    if (dynamic_cast<classical *>(strat)) {
+        new_strat = std::make_shared<typename Other::classical>();
+    } else if (dynamic_cast<merge_path *>(strat)) {
+        new_strat = std::make_shared<typename Other::merge_path>();
+    } else if (dynamic_cast<cusparse *>(strat)) {
+        new_strat = std::make_shared<typename Other::cusparse>();
+    } else if (dynamic_cast<sparselib *>(strat)) {
+        new_strat = std::make_shared<typename Other::sparselib>();
+    } else {
+        auto rexec = result->get_executor();
+        auto cuda_exec = std::dynamic_pointer_cast<const CudaExecutor>(rexec);
+        auto hip_exec = std::dynamic_pointer_cast<const HipExecutor>(rexec);
+        auto lb = dynamic_cast<load_balance *>(strat);
+        if (cuda_exec) {
+            if (lb) {
+                new_strat =
+                    std::make_shared<typename Other::load_balance>(cuda_exec);
+            } else {
+                new_strat =
+                    std::make_shared<typename Other::automatical>(cuda_exec);
+            }
+        } else if (hip_exec) {
+            if (lb) {
+                new_strat =
+                    std::make_shared<typename Other::load_balance>(hip_exec);
+            } else {
+                new_strat =
+                    std::make_shared<typename Other::automatical>(hip_exec);
+            }
+        } else {
+            // FIXME this creates a long delay
+            if (lb) {
+                new_strat = std::make_shared<typename Other::load_balance>();
+            } else {
+                new_strat = std::make_shared<typename Other::automatical>();
+            }
+        }
+    }
+    result->set_strategy(new_strat);
+}
+
+
+template <typename ValueType, typename IndexType>
+void Csr<ValueType, IndexType>::move_to(
+    Csr<next_precision<ValueType>, IndexType> *result)
+{
+    this->convert_to(result);
+}
+
+
+template <typename ValueType, typename IndexType>
+void Csr<ValueType, IndexType>::convert_to(
     Coo<ValueType, IndexType> *result) const
 {
     auto exec = this->get_executor();
--- a/cuda/components/precision_conversion.cu
+++ b/cuda/components/precision_conversion.cu
@@ -40,6 +40,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 namespace gko {
 namespace kernels {
 namespace cuda {
+namespace components {
 
 
 #include "common/components/precision_conversion.hpp.inc"
@@ -60,6 +61,7 @@ void convert_precision(std::shared_ptr<const DefaultExecutor> exec,
 GKO_INSTANTIATE_FOR_EACH_VALUE_CONVERSION(GKO_DECLARE_CONVERT_PRECISION_KERNEL);
 
 
+}  // namespace components
 }  // namespace cuda
 }  // namespace kernels
 }  // namespace gko
--- a/cuda/components/precision_conversion.cu
+++ b/cuda/components/precision_conversion.cu
@@ -1,0 +1,65 @@
+/*******************************<GINKGO LICENSE>******************************
+Copyright (c) 2017-2020, the Ginkgo authors
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+
+1. Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+
+2. Redistributions in binary form must reproduce the above copyright
+notice, this list of conditions and the following disclaimer in the
+documentation and/or other materials provided with the distribution.
+
+3. Neither the name of the copyright holder nor the names of its
+contributors may be used to endorse or promote products derived from
+this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS
+IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED
+TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A
+PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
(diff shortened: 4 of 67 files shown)


upsj:
The PR is now ready for review. There is one rather difficult issue: `strategy_type` in `Csr` and `Hybrid` cannot be converted, so I built a very annoying if-else monster for `Csr` and left the strategy type out completely for `Hybrid`, because the `strategy_type` implementations hide their parameters.

core/device_hooks/common_kernels.inc.cpp
```diff
@@ -66,6 +68,22 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+GKO_NOT_COMPILED(GKO_HOOK_MODULE);
+GKO_INSTANTIATE_FOR_EACH_VALUE_CONVERSION(GKO_DECLARE_CONVERT_PRECISION_KERNEL);
+
+
+template <typename IndexType>
+GKO_DECLARE_PREFIX_SUM_KERNEL(IndexType)
+GKO_NOT_COMPILED(GKO_HOOK_MODULE);
+GKO_INSTANTIATE_FOR_EACH_INDEX_TYPE(GKO_DECLARE_PREFIX_SUM_KERNEL);
+// explicitly instantiate for size_type, as this is used in the SellP format
+template GKO_DECLARE_PREFIX_SUM_KERNEL(size_type);
```
yhmtsai, line 84: Should these be in a namespace?

core/matrix/csr.cpp
```diff
@@ -120,6 +120,68 @@ void Csr<ValueType, IndexType>::apply_impl(const LinOp *alpha, const LinOp *b,
...
+template <typename ValueType, typename IndexType>
+void Csr<ValueType, IndexType>::convert_to(
+    Csr<next_precision<ValueType>, IndexType> *result) const
+{
+    result->values_ = this->values_;
+    result->col_idxs_ = this->col_idxs_;
+    result->row_ptrs_ = this->row_ptrs_;
+    result->set_size(this->get_size());
+    auto strat = this->get_strategy().get();
+    using Other = Csr<next_precision<ValueType>, IndexType>;
```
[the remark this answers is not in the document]
pratikvn: Ah I get it now, forget I asked this question.

core/base/array.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+
+#include <ginkgo/core/base/math.hpp>
+
+
+#include "core/components/precision_conversion.hpp"
+
+
+namespace gko {
+namespace conversion {
```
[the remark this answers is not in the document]
upsj: Since it's only for the kernel operation, I think that's fine!

core/device_hooks/common_kernels.inc.cpp
upsj: Yes, I'll probably move them into a `precision_conversion` namespace. Not sure about `prefix_sum` though. That one I forgot to add to the `common_kernels`, but the name is already kind of self-explanatory. Do we need to put it into a namespace?

include/ginkgo/core/base/array.hpp
```diff
@@ -366,6 +383,54 @@ class Array {
...
+     * executor, it will inherit the executor of other.
+     *
+     * @param other  the Array to copy from
+     * @tparam OtherValueType  the value type of `other`
+     *
+     * @return this
+     */
+    template <typename OtherValueType>
+    xstd::enable_if_t<!std::is_same<ValueType, OtherValueType>::value, Array>
+        &operator=(const Array<OtherValueType> &other)
```
thoasm, line 405: Could you please also add a `operator=(Array<OtyerValueType> &&)` version?

include/ginkgo/core/base/math.hpp
```diff
@@ -202,6 +202,26 @@ GKO_INLINE GKO_ATTRIBUTES constexpr bool is_complex()
...
+
+template <>
+struct next_precision_impl<double> {
+    using type = float;
+};
+
+template <typename T>
+struct next_precision_impl<std::complex<T>> {
+    using type = std::complex<typename next_precision_impl<T>::type>;
+};
```
thoasm, line 222: The problem I see with this is that we could no longer use types like `floatX` as `ValueType`, which worked before (I am not 100% sure if we advertised that, but we aimed to support it, see #49 and #64).
Is it possible to add a default dummy-type, so everything would still compile / work?

Alternatively, we could add a helper `supports_precision_change`, which would work in combination with [`std::conditional`](https://en.cppreference.com/w/cpp/types/conditional) for the inheritance and by its own with the `Array` class.

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -35,6 +35,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
 #include <initializer_list>
+#include <type_traits>
```
thoasm, line 38: nit: Where do you use components from `type_traits`?
You are missing the header `ginkgo/core/base/math.hpp`, which contains the `next_precision` type.
Pretty much all of the matrix formats are missing the inclusion of that header.

omp/test/components/precision_conversion.cpp
```diff
@@ -0,0 +1,147 @@
...
+HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+#include "core/components/prefix_sum.hpp"
```
thoasm, line 33: Did you mean:
```suggestion
#include "core/components/precision_conversion.hpp"
```
Which you are actually not using, so can this header be removed?

upsj:
@yhmtsai float -> double is supposed to be deterministic, double -> float might depend on the rounding mode (which might depend on the issued instructions, flags set in the CPU, ...), i.e. rounding up, down, towards zero, nearest even/odd...

On the suggested tests, I need some clarification on how they extend what is already there

1. can you elaborate what you mean by *corresponding float* value for double -> float? And do you mean take double values that are not *exactly* representable as float (missing precision) or values that are not *at all* representable (outside value range)?

2. This seems to me like a subset of testing float -> double -> float for equality, since all of these double values are exactly representable as float

Maybe give a simple example with a single value?

reference/test/components/precision_conversion.cpp
```diff
@@ -0,0 +1,127 @@
...
+HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+#include "core/components/prefix_sum.hpp"
```
thoasm, line 33: Same with this header. Can it be removed?

reference/test/matrix/coo_kernels.cpp
```diff
@@ -98,6 +98,48 @@ class Coo : public ::testing::Test {
...
+    using IndexType = typename TestFixture::index_type;
+    using OtherType = typename gko::next_precision<ValueType>;
+    using Coo = typename TestFixture::Mtx;
+    using OtherCoo = gko::matrix::Coo<OtherType, IndexType>;
+    auto tmp = OtherCoo::create(this->exec);
+    auto res = Coo::create(this->exec);
+    // If OtherType is more precise: 0, otherwise r
+    auto residual = r<OtherType>::value < r<ValueType>::value
+                        ? gko::remove_complex<ValueType>{0}
+                        : gko::remove_complex<OtherType>{r<OtherType>::value};
```
thoasm, line 113: nit: The `r` struct already removes complex types (see definition [here](https://github.com/ginkgo-project/ginkgo/blob/develop/core/test/utils.hpp#L90))
```suggestion
                        : r<OtherType>::value;
```

reference/test/matrix/coo_kernels.cpp
```diff
@@ -98,6 +98,48 @@ class Coo : public ::testing::Test {
...
```
thoasm, line 134: nit: same here
```suggestion
                        : r<OtherType>::value;
```

reference/test/matrix/csr_kernels.cpp
```diff
@@ -601,6 +601,54 @@ TYPED_TEST(Csr, ApplyFailsOnWrongNumberOfCols)
...
+    using IndexType = typename TestFixture::index_type;
+    using OtherType = typename gko::next_precision<ValueType>;
+    using Csr = typename TestFixture::Mtx;
+    using OtherCsr = gko::matrix::Csr<OtherType, IndexType>;
+    auto tmp = OtherCsr::create(this->exec);
+    auto res = Csr::create(this->exec);
+    // If OtherType is more precise: 0, otherwise r
+    auto residual = r<OtherType>::value < r<ValueType>::value
+                        ? gko::remove_complex<ValueType>{0}
+                        : gko::remove_complex<OtherType>{r<OtherType>::value};
```
thoasm, line 616:
```suggestion
                        : r<OtherType>::value;
```

reference/test/matrix/csr_kernels.cpp
```diff
@@ -601,6 +601,54 @@ TYPED_TEST(Csr, ApplyFailsOnWrongNumberOfCols)
...
```
thoasm, line 640:
```suggestion
                        : r<OtherType>::value;
```

reference/test/matrix/dense_kernels.cpp
```diff
@@ -308,6 +308,46 @@ TYPED_TEST(Dense, ComputDotFailsOnWrongResultSize)
...
+    using Dense = typename TestFixture::Mtx;
+    using T = typename TestFixture::value_type;
+    using OtherT = typename gko::next_precision<T>;
+    using OtherDense = typename gko::matrix::Dense<OtherT>;
+    auto tmp = OtherDense::create(this->exec);
+    auto res = Dense::create(this->exec);
+    // If OtherT is more precise: 0, otherwise r
+    auto residual = r<OtherT>::value < r<T>::value
+                        ? gko::remove_complex<T>{0}
+                        : gko::remove_complex<T>{r<OtherT>::value};
```
thoasm, line 322:
```suggestion
                        : r<OtherT>::value;
```

reference/test/matrix/dense_kernels.cpp
```diff
@@ -308,6 +308,46 @@ TYPED_TEST(Dense, ComputDotFailsOnWrongResultSize)
...
```
thoasm, line 342:
```suggestion
                        : r<OtherT>::value;
```

reference/test/matrix/ell_kernels.cpp
```diff
@@ -211,6 +211,48 @@ TYPED_TEST(Ell, ApplyFailsOnWrongNumberOfCols)
...
+    using IndexType = typename TestFixture::index_type;
+    using OtherType = typename gko::next_precision<ValueType>;
+    using Ell = typename TestFixture::Mtx;
+    using OtherEll = gko::matrix::Ell<OtherType, IndexType>;
+    auto tmp = OtherEll::create(this->exec);
+    auto res = Ell::create(this->exec);
+    // If OtherType is more precise: 0, otherwise r
+    auto residual = r<OtherType>::value < r<ValueType>::value
+                        ? gko::remove_complex<ValueType>{0}
+                        : gko::remove_complex<OtherType>{r<OtherType>::value};
```
thoasm, line 226:
```suggestion
                        : r<OtherType>::value;
```

reference/test/matrix/ell_kernels.cpp
```diff
@@ -211,6 +211,48 @@ TYPED_TEST(Ell, ApplyFailsOnWrongNumberOfCols)
...
```
thoasm, line 247:
```suggestion
                        : r<OtherType>::value;
```

reference/test/matrix/hybrid_kernels.cpp
```diff
@@ -237,6 +237,48 @@ TYPED_TEST(Hybrid, ApplyFailsOnWrongNumberOfCols)
...
+    using IndexType = typename TestFixture::index_type;
+    using OtherType = typename gko::next_precision<ValueType>;
+    using Hybrid = typename TestFixture::Mtx;
+    using OtherHybrid = gko::matrix::Hybrid<OtherType, IndexType>;
+    auto tmp = OtherHybrid::create(this->exec);
+    auto res = Hybrid::create(this->exec);
+    // If OtherType is more precise: 0, otherwise r
+    auto residual = r<OtherType>::value < r<ValueType>::value
+                        ? gko::remove_complex<ValueType>{0}
+                        : gko::remove_complex<OtherType>{r<OtherType>::value};
```
thoasm, line 252:
```suggestion
                        : r<OtherType>::value;
```

reference/test/matrix/hybrid_kernels.cpp
```diff
@@ -237,6 +237,48 @@ TYPED_TEST(Hybrid, ApplyFailsOnWrongNumberOfCols)
...
```
thoasm, line 273:
```suggestion
                        : r<OtherType>::value;
```

reference/test/matrix/sellp_kernels.cpp
```diff
@@ -187,6 +187,48 @@ TYPED_TEST(Sellp, ApplyFailsOnWrongNumberOfCols)
...
+    using IndexType = typename TestFixture::index_type;
+    using OtherType = typename gko::next_precision<ValueType>;
+    using Sellp = typename TestFixture::Mtx;
+    using OtherSellp = gko::matrix::Sellp<OtherType, IndexType>;
+    auto tmp = OtherSellp::create(this->exec);
+    auto res = Sellp::create(this->exec);
+    // If OtherType is more precise: 0, otherwise r
+    auto residual = r<OtherType>::value < r<ValueType>::value
+                        ? gko::remove_complex<ValueType>{0}
+                        : gko::remove_complex<OtherType>{r<OtherType>::value};
```
thoasm, line 202:
```suggestion
                        : r<OtherType>::value;
```

reference/test/matrix/sellp_kernels.cpp
```diff
@@ -187,6 +187,48 @@ TYPED_TEST(Sellp, ApplyFailsOnWrongNumberOfCols)
...
```
thoasm, line 223:
```suggestion
                        : r<OtherType>::value;
```

include/ginkgo/core/matrix/dense.hpp
upsj: That's leftover from when I was using `std::conditional` instead of `next_precision`. Good point!

omp/test/components/precision_conversion.cpp
upsj: Copy-Paste... Yes, that shouldn't be there

include/ginkgo/core/base/math.hpp
upsj: We could add new types by adding them to the `next_precision` chain:
Instead of `float -> double -> float`, we could have `float -> double -> half -> bfloat16 -> unum -> double` or add multiple `next_x_precision` for larger steps in the chain (this would allow all pair-wise conversions directly)

include/ginkgo/core/base/array.hpp
upsj: Can you give a use-case for this operator (i.e. where providing a second overload gives us an advantage?) or how the move-assignment should differ from copy-assignment?

upsj:
@thoasm Thanks for your feedback! I would consider all of the `move_to` implementations something we should tackle in another PR, since as far as I can see, we currently don't really move anywhere (except for moves to the same type), and since our matrix classes have lots of assumptions/invariants about the stored data, we need to be very careful not to leave a matrix in an invalid state after a move (like a naive Coo -> Csr move would leave the row_idxs array unchanged, which gives inconsistent element numbers, a Csr -> * move must restore the row_ptrs or set the size to 0, ...)

yhmtsai:
@upsj I would like to see losing precision from these tests.
We can get the same result from float->float->float as float -> double -> float although I know it should be stored in double in the middle. These tests check whether float->double or double->float really changes the datatypes.
If double -> float is not deterministic, I think checking double->float(->double) with losing precision is enough. 

use double as X=(0x4218888000889111), float Y=(0x50c44400) and the double Z=(0x4218888000000000)?
Y=float(X), Z=double(Y) in Matlab.
Matlab turns this double value to the float value.
I get the error 1.2961e-09 from `abs(double(Y)-X)/X`, which is similar with the matrix test you did.

1. Doing the float->double on Y and comparing with Z.
2. Doing the double->float->double on X and comparing with X.

include/ginkgo/core/base/array.hpp
thoasm: With this operator I wasn't sure myself. At the time I wrote the comment, I simply thought: We support a `move` where we also support `operator=`, so the `move` is missing here.
I am actually fine if we skip the `move` here because we can never really reuse anything, so moving would not have any advantage over the copying.

include/ginkgo/core/base/math.hpp
thoasm: I don't necessarily want something truly genetic, but I would like a `Dense` matrix to still compile when we use `floatx` as a type instead of our defaults. Obviously, we could not use the precision change feature in that case (which is totally reasonable), but it would be nice if we could use all the other features that do not depend on the precision change.

upsj: Can you clarify how this would work together with our header/source separation? To me it looks like adding a new precision type requires changes at the implementation level, since it also requires modifying `dense.cpp` or re-implementing all member functions you want to use by yourself.

reference/test/matrix/coo_kernels.cpp
upsj: The point of these lines was actually to make sure that both branches of the ternary operator have the same type, so this was supposed to be
```suggestion
                        : gko::remove_complex<ValueType>{r<OtherType>::value};
```
Do you think that's okay? Or overly cautious?

thoasm: That is fine for me as well.

include/ginkgo/core/base/math.hpp
thoasm: Oh, my bad, I thought we could simply use Ginkgo inside CMake, and use a `floatx` type as a template parameter. After looking into the PRs that I referenced, we simply added the new type to the instantiate macro to make it work. In this case, we would also have to add it to the `next_precision` implementation, but that seems fine to me.
Sorry for causing this much confusion.

upsj:
I added tests for conversions via another executor (and also found a bug that way), rounding with precision loss and moved the component kernels to a namespace.

Thanks for your feedback, that was really helpful!

cuda/components/precision_conversion.cu
```diff
@@ -0,0 +1,67 @@
...
+namespace gko {
+namespace kernels {
+namespace cuda {
+namespace components {
+
+
+#include "common/components/precision_conversion.hpp.inc"
+
+
+constexpr int default_block_size = 512;
```
thoasm, line 49: nit: We usually have this defined before we include the `.inc` file, so we have only one defined variable `default_block_size` that we use in both files (cuda & components or hip & components).

hip/components/precision_conversion.hip.cpp
```diff
@@ -0,0 +1,68 @@
...
+namespace gko {
+namespace kernels {
+namespace hip {
+namespace components {
+
+
+#include "common/components/precision_conversion.hpp.inc"
+
+
+constexpr int default_block_size = 512;
```
thoasm, line 49: nit: see previous comment on moving this above the include to avoid having 2 `default_block_size`.

cuda/test/components/precision_conversion.cu
```diff
@@ -0,0 +1,173 @@
...
+        std::uniform_real_distribution<float> dist(-maxval, maxval);
+        for (gko::size_type i = 0; i < total_size; ++i) {
+            vals.get_data()[i] = dist(rand);
+            cvals.get_data()[i] = {dist(rand), dist(rand)};
+        }
+        dvals = vals;
+        dcvals = cvals;
+        gko::uint64 rawdouble = 0x4218888000889111ULL;
+        gko::uint32 rawfloat = 0x50c44400ULL;
+        gko::uint64 rawrounded = 0x4218888000000000ULL;
```
thoasm, line 78: On all of these, I would prefer the brace-initializer `{}` to make surer it can fit.
For `rawfloat`, your suffix is actually wrong (ULL is 64 bit):
```suggestion
        gko::uint64 rawdouble{0x4218888000889111ULL};
        gko::uint32 rawfloat{0x50c44400U};
        gko::uint64 rawrounded{0x4218888000000000ULL};
```
I think it should also be possible to remove all suffixes without any issues, but since I did not test it, I left them in.

hip/test/components/precision_conversion.hip.cpp
```diff
@@ -0,0 +1,173 @@
...
+        std::uniform_real_distribution<float> dist(-maxval, maxval);
+        for (gko::size_type i = 0; i < total_size; ++i) {
+            vals.get_data()[i] = dist(rand);
+            cvals.get_data()[i] = {dist(rand), dist(rand)};
+        }
+        dvals = vals;
+        dcvals = cvals;
+        gko::uint64 rawdouble = 0x4218888000889111ULL;
+        gko::uint32 rawfloat = 0x50c44400ULL;
+        gko::uint64 rawrounded = 0x4218888000000000ULL;
```
thoasm, line 78:
```suggestion
        gko::uint64 rawdouble{0x4218888000889111ULL};
        gko::uint32 rawfloat{0x50c44400U};
        gko::uint64 rawrounded{0x4218888000000000ULL};
```

omp/test/components/precision_conversion.cpp
```diff
@@ -0,0 +1,173 @@
...
+        std::uniform_real_distribution<float> dist(-maxval, maxval);
+        for (gko::size_type i = 0; i < total_size; ++i) {
+            vals.get_data()[i] = dist(rand);
+            cvals.get_data()[i] = {dist(rand), dist(rand)};
+        }
+        dvals = vals;
+        dcvals = cvals;
+        gko::uint64 rawdouble = 0x4218888000889111ULL;
+        gko::uint32 rawfloat = 0x50c44400ULL;
+        gko::uint64 rawrounded = 0x4218888000000000ULL;
```
thoasm, line 78:
```suggestion
        gko::uint64 rawdouble{0x4218888000889111ULL};
        gko::uint32 rawfloat{0x50c44400U};
        gko::uint64 rawrounded{0x4218888000000000ULL};
```

reference/test/components/precision_conversion.cpp
```diff
@@ -0,0 +1,150 @@
...
+    {
+        auto maxval = 1e10f;
+        std::uniform_real_distribution<float> dist(-maxval, maxval);
+        for (gko::size_type i = 0; i < total_size; ++i) {
+            vals.get_data()[i] = dist(rand);
+            cvals.get_data()[i] = {dist(rand), dist(rand)};
+        }
+        gko::uint64 rawdouble = 0x4218888000889111ULL;
+        gko::uint32 rawfloat = 0x50c44400ULL;
+        gko::uint64 rawrounded = 0x4218888000000000ULL;
```
thoasm, line 72:
```suggestion
        gko::uint64 rawdouble{0x4218888000889111ULL};
        gko::uint32 rawfloat{0x50c44400U};
        gko::uint64 rawrounded{0x4218888000000000ULL};
```

cuda/test/components/precision_conversion.cu
upsj: Small detail: `U` only guarantees 16 bit, it needs to be `UL`

thoasm: I don't think so, since the number that you wrote can't be stored in 16 bits, so C++ chooses the next larger type that can store the number.
Even for the others, I don't think it is necessary to have any postfix, but they don't hurt as long as they have the appropriate size (which I would like to have checked with the brace-initialization).

upsj: That is true. I went with braces in the final version ;)
Sometimes I like to be excessively verbose just to be sure :D

upsj:
@yhmtsai The tests make sure that conversions between different executors work. This actually uncovered a bug in the Array conversion-assignment implementation.

hartwiganzt:
@upsj ready to be merged!

PR #481, opened 2020-03-12
 Using next_krylov_basis instead of krylov_bases

fritzgoebel:
... to save unnecessary padding in the preconditioned vector. This made it necessary to choose a different spmv strategy when using plain GMRES or with ilu preconditioner, since the default choice doesn't support multiple vectors.

PR #523, opened 2020-04-30
Reorder krylov bases of gmres

yhmtsai:
This PR changes the order of storing krylov bases.

Originally, we store the Krylov bases as following:
K1 = the first vector of Krylov basis for each rhs => num_rows * num_rhs
K = [K1 K2 ... Kn], which lead the stride of Krylov still be larger than 1 even if the num_rhs = 1.
Thus, we can not get coalesced memory access.
Without changing the implementation of algorithm, change the Krylov order like
K = [K1; K2; ...; Kn] to get coalesced memory access when num_rhs = 1

When num_rhs = 1, use vendor dot not our own multidot_kernel which is parallel on #rhs.

Moreover, the stride of krylov_bases is same as stride of rhs.
Do not face the problem mentioned in #481, so next_krylov_basis is not needed now.

Todo:
- [x] change the algorithm comment


common/solver/gmres_kernels.hpp.inc
```diff
@@ -351,22 +329,20 @@ __global__ __launch_bounds__(block_size) void givens_rotation_kernel(
         return;
     }
 
-    const auto current_thread_block = group::this_thread_block();
+    // const auto current_thread_block = group::this_thread_block();
```
pratikvn, line 332: Can be removed ?

common/solver/gmres_kernels.hpp.inc
```diff
@@ -284,26 +271,22 @@ __device__ void calculate_sin_and_cos_kernel(
...
-        givens_cos[iter * stride_cos + col_idx] =
-            abs(hessenberg_iter[iter * stride_hessenberg + col_idx]) /
-            hypotenuse;
-        givens_sin[iter * stride_sin + col_idx] =
-            givens_cos[iter * stride_cos + col_idx] *
-            hessenberg_iter[(iter + 1) * stride_hessenberg + col_idx] /
-            hessenberg_iter[iter * stride_hessenberg + col_idx];
+        const auto next_hess =
+            hessenberg_iter[(iter + 1) * stride_hessenberg + col_idx];
+        auto hypotenuse = sqrt(this_hess * this_hess + next_hess * next_hess);
```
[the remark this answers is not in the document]
yhmtsai: Do you mean we still use this function even if `this_hess = 0`?

upsj:
I think the complex Givens rotations are still not 100% correct. What it looks like to me:
\begin{pmatrix}c&s\\-\overline{s}&\overline{c}\end{pmatrix}\begin{pmatrix}a\\b\end{pmatrix}=\begin{pmatrix}r\\0\end{pmatrix}
We need this form for the matrix to be unitary. Ansatz:
r=\sqrt{|a|^2 + |b|^2}, c=\overline{a} / r, s=\overline{b} / r
Then the update looks like this:
\begin{pmatrix}c&s\\-\overline{s}&\overline{c}\end{pmatrix}\begin{pmatrix}a\\b\end{pmatrix}=\begin{pmatrix}ca+sb\\-\overline{s}a + \overline{c}b\end{pmatrix}
So we need to also conjugate sin and cos for the second entry of H we are updating.

Please feel free to check my calculations!

yhmtsai:
@upsj The formula looks correct.
the current given rotation will rotate to the `a` direction.
That is, in your first formula -> [r (unit_vector of a) 0]'

I prefer your formula because of the same behaviour of cos and sin.

common/solver/gmres_kernels.hpp.inc
```diff
@@ -300,7 +299,7 @@ __device__ void calculate_residual_norm_kernel(
     const auto this_rnc =
         residual_norm_collection[iter * stride_residual_norm_collection +
                                  col_idx];
-    const auto next_rnc = -register_sin * this_rnc;
+    const auto next_rnc = -conj(register_sin) * this_rnc;
     residual_norm_collection[iter * stride_residual_norm_collection + col_idx] =
         register_cos * this_rnc;
```
yhmtsai, line 304: `norm_collection(iter) = cos * this_rnc + sin * next_rnc = cos * this_rnc;`
`norm_collection(iter+1) = -conj(sin) * this_rnc + cos * next_rnc = -conj(sin) * this_rnc;`
Is that correct?

cuda/solver/gmres_kernels.cu
```diff
@@ -144,35 +137,50 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE(GKO_DECLARE_GMRES_INITIALIZE_2_KERNEL);
...
-            as_cuda_type(next_krylov_basis->get_const_values()),
-            stride_next_krylov, as_cuda_type(krylov_bases->get_const_values()),
-            stride_krylov, as_cuda_type(hessenberg_iter->get_values()),
-            stride_hessenberg, as_cuda_type(stop_status));
+        const auto k_krylov_bases =
+            krylov_bases->get_const_values() +
+            k * num_rows * hessenberg_iter->get_size()[1];
+        if (hessenberg_iter->get_size()[1] > 1) {
+            // TODO: single rhs will use vendor's dot, otherwise, use our own
+            // multidot_kernel which parallelize multiple rhs.
```
tcojean, line 160: Isn't this now done?

cuda/solver/gmres_kernels.cu
```diff
@@ -144,35 +137,50 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE(GKO_DECLARE_GMRES_INITIALIZE_2_KERNEL);
...
+            zero_array(hessenberg_iter->get_size()[1],
+                       hessenberg_iter->get_values() + k * stride_hessenberg);
+            multidot_kernel<<<grid_size, block_size>>>(
+                k, num_rows, hessenberg_iter->get_size()[1],
+                as_cuda_type(k_krylov_bases), as_cuda_type(next_krylov_basis),
+                stride_krylov, as_cuda_type(hessenberg_iter->get_values()),
+                stride_hessenberg, as_cuda_type(stop_status));
+        } else {
+            for (size_type col = 0; col < hessenberg_iter->get_size()[1];
+                 ++col) {
```
tcojean, line 170: You don't need this `for` here anymore.

hip/solver/gmres_kernels.hip.cpp
```diff
@@ -141,45 +134,59 @@ void initialize_2(std::shared_ptr<const HipExecutor> exec,
...
-            as_hip_type(next_krylov_basis->get_const_values()),
-            stride_next_krylov, as_hip_type(krylov_bases->get_const_values()),
-            stride_krylov, as_hip_type(hessenberg_iter->get_values()),
-            stride_hessenberg, as_hip_type(stop_status));
+        const auto k_krylov_bases =
+            krylov_bases->get_const_values() +
+            k * num_rows * hessenberg_iter->get_size()[1];
+        if (hessenberg_iter->get_size()[1] > 1) {
+            // TODO: single rhs will use vendor's dot, otherwise, use our own
+            // multidot_kernel which parallelize multiple rhs.
```
tcojean, line 165: Same comment as for CUDA.

hip/solver/gmres_kernels.hip.cpp
```diff
@@ -141,45 +134,59 @@ void initialize_2(std::shared_ptr<const HipExecutor> exec,
...
+                       hessenberg_iter->get_values() + k * stride_hessenberg);
+            hipLaunchKernelGGL(
+                multidot_kernel, dim3(grid_size), dim3(block_size), 0, 0, k,
+                num_rows, hessenberg_iter->get_size()[1],
+                as_hip_type(k_krylov_bases), as_hip_type(next_krylov_basis),
+                stride_krylov, as_hip_type(hessenberg_iter->get_values()),
+                stride_hessenberg, as_hip_type(stop_status));
+        } else {
+            for (size_type col = 0; col < hessenberg_iter->get_size()[1];
+                 ++col) {
```
tcojean, line 176: Same comment as for CUDA.

cuda/solver/gmres_kernels.cu
yhmtsai: I would like to metion the condition should be tuned

PR #524, opened 2020-04-30
Add a SONAME to shared libraries

tamiko:
Ginkgo is written in C++ and does not guarantee any stable ABI.
Consequently, all shared libraries should be annotated by an appropriate
SONAME (this patch simply uses the release version). This ensures
conformity with best practices and library requirements for popular
Linux distributions (e.g., Debian/Ubuntu). It also prevents silent
breakage (at dynamic link time / run time) when the library ABI changed
and user code should have been recompiled.

### DIFF
--- a/cmake/build_helpers.cmake
+++ b/cmake/build_helpers.cmake
@@ -17,6 +17,9 @@ function(ginkgo_compile_features name)
     if(GINKGO_WITH_IWYU AND GINKGO_IWYU_PATH)
         set_property(TARGET "${name}" PROPERTY CXX_INCLUDE_WHAT_YOU_USE ${GINKGO_IWYU_PATH})
     endif()
+    # Set an appropriate SONAME
+    set_property(TARGET "${name}" PROPERTY
+        SOVERSION "${Ginkgo_VERSION}")
     if(GINKGO_CHANGED_SHARED_LIBRARY)
         # Put all shared libraries and corresponding imported libraries into the specified path
         set_property(TARGET "${name}" PROPERTY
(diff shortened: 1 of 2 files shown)


pratikvn:
@tamiko . Thank you for your contribution!

tamiko:
@pratikvn 
Added an entry to the contributors file.

@upsj 
> When linking an executable with a library that provides a versioned SONAME, will CMake then correctly set the NEEDED entry to include the version number?

Yes. By providing above target property CMake will create the shared library with a SONAME and set up appropriate symlinks:
```
% objdump -x libginkgo.so | grep SONAME
  SONAME               libginkgo.so.1.1.1
% ls -la libginkgo.so*
lrwxrwxrwx 1 root root      18 Apr 30 16:23 libginkgo.so -> libginkgo.so.1.1.1
-rwxr-xr-x 1 root root 7678896 Apr 30 16:23 libginkgo.so.1.1.1
```
The linker will now record `libginkgo.so.1.1.1` in the dynamic section of executables and libraries linking against `libginkgo.so`:
```
% objdump -x libdeal_II.so | grep libginkgo.so
  NEEDED               libginkgo.so.1.1.1
```
The dynamic linker/loader will now search for a shared object file named `libginkgo.so.1.1.1` instead of `libginkgo.so`:
```
% ldd libdeal_II.so | grep libginkgo.so
libginkgo.so.1.1.1 => /usr/lib64/libginkgo.so.1.1.1 (0x00007f4f59a4f000)
```

upsj:
We seem to have some issues with the interaction of HIP and nvcc as a linker.
[CI Pipeline](https://gitlab.com/ginkgo-project/ginkgo-public-ci/pipelines/142317252)

tcojean:
Yes I saw that, I'm looking into it.

tcojean:
@tamiko Sorry for the delay, could you rebase this to the latest develop status? The HIP issue which this PR highlighted should now be fixed.

tamiko:
@tcojean I rebase onto the current `develop` branch.

PR #527, opened 2020-05-07
Add setters and getters for some solver parameters

pratikvn:
~~This PR adds a macro to set solver parameters after the solver has been generated. This allows for changing the max iter count, residual norm goal or other parameters for successive calls to solver applies without generating the solver again.~~
It is now updated to add the setters and getter functions instead. It was decided that the macro would cause more confusion than acceptable.

Additionally, it also adds typed tests for the Bicg solver which were missing.

The generate of the stopping criterion is called within the apply of the solvers. So when set after generate, it should use the new stopping criteria. This holds for any of the other class member variables which have setters and getters as well.

### DIFF
--- a/core/test/solver/gmres.cpp
+++ b/core/test/solver/gmres.cpp
@@ -215,6 +215,31 @@ TYPED_TEST(Gmres, CanSetPreconditionerGenerator)
 }
 
 
+TYPED_TEST(Gmres, CanSetCriteriaAgain)
+{
+    using Solver = typename TestFixture::Solver;
+    std::shared_ptr<gko::stop::CriterionFactory> init_crit =
+        gko::stop::Iteration::build().with_max_iters(3u).on(this->exec);
+    auto gmres_factory =
+        Solver::build().with_criteria(init_crit).on(this->exec);
+
+    ASSERT_EQ((gmres_factory->get_parameters().criteria).back(), init_crit);
+
+    auto solver = gmres_factory->generate(this->mtx);
+    std::shared_ptr<gko::stop::CriterionFactory> new_crit =
+        gko::stop::Iteration::build().with_max_iters(5u).on(this->exec);
+
+    solver->set_stop_criterion_factory(new_crit);
+    auto new_crit_fac = solver->get_stop_criterion_factory();
+    auto niter =
+        static_cast<const gko::stop::Iteration::Factory *>(new_crit_fac.get())
+            ->get_parameters()
+            .max_iters;
+
+    ASSERT_EQ(niter, 5);
+}
+
+
 TYPED_TEST(Gmres, CanSetKrylovDim)
 {
     using Solver = typename TestFixture::Solver;
@@ -232,6 +257,25 @@ TYPED_TEST(Gmres, CanSetKrylovDim)
     auto krylov_dim = solver->get_krylov_dim();
 
     ASSERT_EQ(krylov_dim, 4);
+}
+
+
+TYPED_TEST(Gmres, CanSetKrylovDimAgain)
+{
+    using Solver = typename TestFixture::Solver;
+    std::shared_ptr<gko::stop::CriterionFactory> init_crit =
+        gko::stop::Iteration::build().with_max_iters(3u).on(this->exec);
+    auto gmres_factory =
+        Solver::build().with_criteria(init_crit).with_krylov_dim(10u).on(
+            this->exec);
+
+    ASSERT_EQ(gmres_factory->get_parameters().krylov_dim, 10);
+
+    auto solver = gmres_factory->generate(this->mtx);
+
+    solver->set_krylov_dim(20);
+
+    ASSERT_EQ(solver->get_krylov_dim(), 20);
 }
 
 
--- a/include/ginkgo/core/base/lin_op.hpp
+++ b/include/ginkgo/core/base/lin_op.hpp
@@ -838,6 +838,31 @@ public:                                                                      \
 
 
 /**
+ * Defines setters and getters for the parameters of a class, simplifying its
+ * construction by removing the repetitive typing of boiler plate code.
+ *
+ * @param _concrete_parameter_name  the parameter whose setter and getter is to
+ * be defined
+ *
+ * @ingroup LinOp
+ */
+#define GKO_ENABLE_SET_GET_PARAMETERS(_concrete_parameter_name, _type)       \
+public:                                                                      \
+    void set_##_concrete_parameter_name(_type other)                         \
+    {                                                                        \
+        this->_concrete_parameter_name##_ = other;                           \
+    }                                                                        \
+                                                                             \
+    _type get_##_concrete_parameter_name() const                             \
+    {                                                                        \
+        return this->_concrete_parameter_name##_;                            \
+    }                                                                        \
+    static_assert(true,                                                      \
+                  "This assert is used to counter the false positive extra " \
+                  "semi-colon warnings")
+
+
+/**
  * Defines a build method for the factory, simplifying its construction by
  * removing the repetitive typing of factory's name.
  *
--- a/include/ginkgo/core/base/lin_op.hpp
+++ b/include/ginkgo/core/base/lin_op.hpp
@@ -839,10 +839,18 @@ public:                                                                      \
 
 /**
  * Defines setters and getters for the parameters of a class, simplifying its
- * construction by removing the repetitive typing of boiler plate code.
+ * construction by removing the repetitive typing of boiler plate code. The
+ * setter and getter functions will be named set_parameter_name(...) and
+ * get_parameter_name().
+ *
+ * The passed in parameter should be a member variable of the class and not the
+ * factory parameter. These setters and getters can be used to re-set the member
+ * variables after the Factory has been generated. Hence, the generate step need
+ * not be called after using the setter.
  *
  * @param _concrete_parameter_name  the parameter whose setter and getter is to
  * be defined
+ * @param _type  the type of the parameter being passed in.
  *
  * @ingroup LinOp
  */
--- a/include/ginkgo/core/base/lin_op.hpp
+++ b/include/ginkgo/core/base/lin_op.hpp
@@ -838,40 +838,6 @@ public:                                                                      \
 
 
 /**
- * Defines setters and getters for the parameters of a class, simplifying its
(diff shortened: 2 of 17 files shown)


reference/test/solver/bicg_kernels.cpp
```diff
@@ -96,167 +99,209 @@ class Bicg : public ::testing::Test {
...
-    auto x = gko::initialize<Mtx>({0.0, 0.0, 0.0, 0.0, 0.0, 0.0}, exec);
+        {886630.5, -172578.0, 684522.0, -65310.5, 455487.5, 607436.0},
+        this->exec);
+    auto x = gko::initialize<Mtx>({0.0, 0.0, 0.0, 0.0, 0.0, 0.0}, this->exec);
 
     solver->apply(b.get(), x.get());
 
-    GKO_ASSERT_MTX_NEAR(x, l({33.0, -56.0, 81.0, -30.0, 21.0, 40.0}), 1e-10);
+    GKO_ASSERT_MTX_NEAR(x, l({33.0, -56.0, 81.0, -30.0, 21.0, 40.0}),
+                        r<value_type>::value * 1e2);
```
upsj, line 208: Here you are slightly inconsistent: 1e-10 -> r * 1e2
Below: 1e-14 -> r * 1e2

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -828,6 +828,31 @@ public:                                                                      \
...
+public:                                                                      \
+    void set_##_concrete_parameter_name(_type other)                         \
+    {                                                                        \
+        this->_concrete_parameter_name##_ = other;                           \
+    }                                                                        \
+                                                                             \
+    _type get_##_concrete_parameter_name() const                             \
+    {                                                                        \
+        return this->_concrete_parameter_name##_;                            \
+    }                                                                        \
```
upsj, line 850: Could we get by without specifying the type explicitly?
```suggestion
#define GKO_ENABLE_SET_GET_PARAMETERS(_concrete_parameter_name)              \
public:                                                                      \
    void set_##_concrete_parameter_name(                                     \
        decltype(this->_concrete_parameter_name##_) other)                   \
    {                                                                        \
        this->_concrete_parameter_name##_ = other;                           \
    }                                                                        \
                                                                             \
    auto get_##_concrete_parameter_name() const                              \
        -> decltype(this->_concrete_parameter_name##_)                       \
    {                                                                        \
        return this->_concrete_parameter_name##_;                            \
    }                                                                        \
```

pratikvn: I had also tried this but it failed with 
```
/home/public_ginkgo/include/ginkgo/core/base/lin_op.hpp:858:18: error: invalid use of ‘this’ at top level
                                 decltype(this->_concrete_parameter_name##_) other)                   \ 
```

Additonally, there were [other places](https://github.com/ginkgo-project/ginkgo/blob/develop/include/ginkgo/core/stop/criterion.hpp#L109) where we pass in the type as well, so I assumed it was okay.

upsj: Did you try leaving out the `this->` prefixes? That seems to work with the compilers I tested.

pratikvn: You mean like `decltype(_concrete_parameter_name##_) other` ? Then it does not know that the `concrete_paramater_name##_` is a member variable right ? I do get errors saying that it is not declared.

upsj: As far as I can tell, that's an issue of declaration order:
```cpp

class c {
    // works
    int i;
    auto f(decltype(i) a) -> decltype(i) {
        i = a;
    }
};

class c {
    // fails
    auto f(decltype(i) a) -> decltype(i) {
        i = a;
    }
    int i;
};
```

So if you put the getter/setter declaration after the member variable declaration, it should work

pratikvn: Yeah. I guess you could declare the member variables earlier, but I am not sure I am too happy about that.

upsj: (or declare the setters later, but I am also not happy about any of those approaches)

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -828,6 +828,31 @@ public:                                                                      \
                   "semi-colon warnings")
 
 
+/**
+ * Defines setters and getters for the parameters of a class, simplifying its
+ * construction by removing the repetitive typing of boiler plate code.
```
yhmtsai, line 833: I would add some description about accessing the variable `_concrete_parameter_name_` and the class can not use these parameter in the initial/generate step

include/ginkgo/core/solver/bicg.hpp
```diff
@@ -108,6 +108,8 @@ class Bicg : public EnableLinOp<Bicg<ValueType>>, public Preconditionable {
     };
     GKO_ENABLE_LIN_OP_FACTORY(Bicg, parameters, Factory);
     GKO_ENABLE_BUILD_METHOD(Factory);
+    GKO_ENABLE_SET_GET_PARAMETERS(
+        stop_criterion_factory, std::shared_ptr<const stop::CriterionFactory>);
```
yhmtsai, line 112: Should we keep the same name in the factory?

pratikvn: The set and get functions generated are for the member variables, so I would have to pass in the names and types of the member variables.

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -828,6 +828,39 @@ public:                                                                      \
...
+/**
+ * Defines setters and getters for the parameters of a class, simplifying its
+ * construction by removing the repetitive typing of boiler plate code. The
+ * setter and getter functions will be named set_parameter_name(...) and
+ * get_parameter_name().
+ *
+ * The passed in parameter should be a member variable of the class and not the
+ * factory parameter. These setters and getters can be used to re-set the member
+ * variables after the Factory has been generated. Hence, the generate step need
+ * not be called after using the setter.
```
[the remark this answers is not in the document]
tcojean: In addition, I think we need to be careful about this. I think we should have a big warning maybe with the following example to show that this can be dangerous.

This is not true for a lot of cases: if you consider the BlockJacobi, if you put a setter for the block pointers or storage optimization, then you would definitely need to call the generate method again (although I believe we never should do that to begin with, that is just to show that this can be used dangerously).

core/test/solver/gmres.cpp
```diff
@@ -209,6 +209,31 @@ TYPED_TEST(Gmres, CanSetPreconditionerGenerator)
...
+    auto gmres_factory =
+        Solver::build().with_criteria(init_crit).on(this->exec);
+
+    ASSERT_EQ((gmres_factory->get_parameters().criteria).back(), init_crit);
+
+    auto solver = gmres_factory->generate(this->mtx);
+    std::shared_ptr<gko::stop::CriterionFactory> new_crit =
+        gko::stop::Iteration::build().with_max_iters(5u).on(this->exec);
+
+    solver->set_stop_criterion_factory(new_crit);
```
tcojean, line 226: There is one issue here, I think, and I'm not sure what happens then:
All solvers if you look in their constructor, have these lines in their constructor (i.e. factory generate step):
https://github.com/ginkgo-project/ginkgo/blob/587df41c53f38b96b123cc8cbcf8544fbed926b6/include/ginkgo/core/solver/gmres.hpp#L158-L164

What happens in your case then? I'm quite sure that the solver's apply would still be using the wrong criterion, no?
https://github.com/ginkgo-project/ginkgo/blob/500d4a5719ab0b60f76f7f5cbc6f77fd6254f544/core/solver/gmres.cpp#L129-L131

tcojean: It's the same for `krylov_dim` btw.

pratikvn: When you use the `set_stop_criterion_factory(fac)`, it should overwrite the value in the current `stop_criterion_factory_` and hence in the `apply_impl` the new generate is called on the stop_crit_fac, right ?

reference/test/solver/bicg_kernels.cpp
```diff
@@ -96,167 +99,209 @@ class Bicg : public ::testing::Test {
...
+    auto b = gko::initialize<Mtx>(
+        {I<T>{-1.0, 1.0}, I<T>{3.0, 0.0}, I<T>{1.0, 1.0}}, this->exec);
+    auto x = gko::initialize<Mtx>(
+        {I<T>{0.5, 1.0}, I<T>{1.0, 2.0}, I<T>{2.0, 3.0}}, this->exec);
 
     solver->apply(alpha.get(), b.get(), beta.get(), x.get());
 
-    GKO_ASSERT_MTX_NEAR(x, l({{1.5, 1.0}, {5.0, 0.0}, {2.0, -1.0}}), 1e-14);
+    GKO_ASSERT_MTX_NEAR(x, l({{1.5, 1.0}, {5.0, 0.0}, {2.0, -1.0}}),
+                        r<value_type>::value * 1e1);
```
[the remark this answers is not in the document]
pratikvn: Yes, I think so.

core/test/solver/gmres.cpp
tcojean: My bad, that was a misunderstanding on my part. I now realize you do not change the factory parameters but only the generated solver's. 
I'm not sure I like that inconsistency though, what is the meaning of factories in that case if the solvers deviate from their factories? But indeed you don't have the issue pointed here.

pratikvn: I agree that there is a slight inconsistency, but I think it is useful when you dont want to re-generate the whole factory, but just want to reset the parameter when re-running the solver, when the matrix is the same and all other parameters are the same.

pratikvn: And by default, you always get whatever you set in the factory build. And these setters can affect only when you call the solver again.

yhmtsai:
Is using a class to contain these functions (and maybe get_system_matrix) suitable?

pratikvn:
@yhmtsai , Yes, we could do that. That would mean that the member variables would go into that new class and not be in the solver classes, which I am not a big fan of.  Moreover, to me, setters and getters should be local to the class.

PR #528, opened 2020-05-11
Amgx parallel graph match reference and fill_seq_array

yhmtsai:
This PR adds MultigridLevel and UseComposition class and AmgxPgm (AMGX parallel graph match) class.

Any class to represent one level of Multigrid should inherit from Multigrid class.
MultigridLevel contains fine_op and a composition of (prolong, coarse, restrict) by `set_multigrid_level`.
It also provides `get_*_op()` to get the component of MultigridLevel.
UseComposition considers end-class to contain composition not be a composition, which gives the end-class full linop functionalities.
fill_seq_array fills array from 0 to size-1

AmgxPgm only contains size=2 implementation which is shown on the paper.

need to extract diag from csr, which is fully implemented in the #563 
- [x] Merge #563 
- [x] Merge #636 to use absolute function


hip/matrix/csr_kernels.hip.cpp
```diff
@@ -654,14 +654,15 @@ void advanced_spgemm(std::shared_ptr<const HipExecutor> exec,
...
-                      d_row_ptrs, d_col_idxs, d_vals, c);
+        select_spgeam(
+            spgeam_kernels(),
+            [&](int compiled_subwarp_size) {
+                return compiled_subwarp_size >= nnz_per_row ||
+                       compiled_subwarp_size == config::warp_size;
+            },
+            syn::value_list<int>(), syn::type_list<>(), exec, valpha,
+            c_tmp_row_ptrs, c_tmp_col_idxs, c_tmp_vals, vbeta, d_row_ptrs,
+            d_col_idxs, d_vals, c);
```
upsj, line 665: formatting :) I am starting to think we need to take care of this from the CI side.

core/multigrid/amgx_pgm.cpp
```diff
@@ -0,0 +1,173 @@
...
+            break;
+        }
+        num_unagg_prev = num_unagg;
+    }
+    // Handle the left unassign points
+    if (num_unagg != 0 && intermediate_agg.get_num_elems() > 0) {
+        // copy the agg to intermediate_agg
+        intermediate_agg = agg_;
+    }
+    while (num_unagg != 0) {
```
upsj, line 130: Can it happen that we need to call assign_to_exist_agg multiple times? To me, it looks like every unassigned row will be assigned after the kernel finishes.

include/ginkgo/core/multigrid/restrict_prolong.hpp
```diff
@@ -0,0 +1,329 @@
...
+     * Applies a prolong operator to a vector (or a sequence of vectors).
+     *
+     * Performs the operation x = prolong(b) + x.
+     *
+     * @param b  the input vector(s) on which the operator is applied
+     * @param x  the output vector(s) where the result is stored
+     *
+     * @return this
+     */
+    RestrictProlong *prolong_applyadd(const LinOp *b, LinOp *x)
```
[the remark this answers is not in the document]
yhmtsai: I think `applyadd` should be more precise information on this behavior than `apply2`?
P = R^T in general, but some cases they can be different.

core/multigrid/amgx_pgm.cpp
yhmtsai: You are right. it will assign the unagg to one agg group or itself.

tcojean:
@yhmtsai can this PR be reviewed now that the diagonal PRs are merged or is this still a WIP? Also, you have a lot of conflicts right now.

yhmtsai:
@tcojean It is still in WIP. I need to modify this pr with diagonal format.

include/ginkgo/core/multigrid/amgx_pgm.hpp
```diff
@@ -0,0 +1,162 @@
...
+     */
+    const IndexType *get_const_agg() const noexcept
+    {
+        return agg_.get_const_data();
+    }
+
+    GKO_CREATE_FACTORY_PARAMETERS(parameters, Factory)
+    {
+        /**
+         * The maximum number of iteration.
```
pratikvn, line 111: nit:
```suggestion
         * The maximum number of iterations.
```

include/ginkgo/core/multigrid/amgx_pgm.hpp
```diff
@@ -0,0 +1,162 @@
...
+
+        /**
+         * The maximum ratio of unassigned number.
+         */
+        double GKO_FACTORY_PARAMETER(max_unassigned_percentage, 0.05);
+
+        /**
+         * Use the deterministic assign_to_exist_agg method or not.
+         */
+        bool GKO_FACTORY_PARAMETER(deterministic, false);
```
pratikvn, line 123: maybe you can also explain in short, the effects of using deterministic or non-deterministic ?

include/ginkgo/core/multigrid/restrict_prolong.hpp
```diff
@@ -0,0 +1,329 @@
...
+     * `restrict_apply` are of the wrong size.
+     *
+     * @param b  vector(s) on which the operator is applied
+     * @param x  output vector(s)
+     */
+    void validate_restrict_parameters(const LinOp *b, LinOp *x) const
+    {
+        auto restirct_dim = gko::dim<2>(coarse_dim_, fine_dim_);
+        GKO_ASSERT_CONFORMANT(restirct_dim, b);
+        GKO_ASSERT_EQUAL_ROWS(restirct_dim, x);
```
pratikvn, line 143:
```suggestion
        auto restrict_dim = gko::dim<2>(coarse_dim_, fine_dim_);
        GKO_ASSERT_CONFORMANT(restrict_dim, b);
        GKO_ASSERT_EQUAL_ROWS(restrict_dim, x);
```

include/ginkgo/core/multigrid/restrict_prolong.hpp
```diff
@@ -0,0 +1,329 @@
...
+    void set_coarse_fine(std::shared_ptr<const LinOp> coarse,
+                         size_type fine_dim)
+    {
+        coarse_ = coarse;
+        fine_dim_ = fine_dim;
+        coarse_dim_ = coarse->get_size()[0];
+    }
+
+    /**
+     * Creates a coarse fine.
```
pratikvn, line 201: I think this should be updated ?

include/ginkgo/core/multigrid/restrict_prolong.hpp
```diff
@@ -0,0 +1,329 @@
...
+    : public AbstractFactory<RestrictProlong, std::shared_ptr<const LinOp>> {
+public:
+    using AbstractFactory<RestrictProlong,
+                          std::shared_ptr<const LinOp>>::AbstractFactory;
+
+    std::unique_ptr<RestrictProlong> generate(
+        std::shared_ptr<const LinOp> input) const
+    {
+        auto generated = AbstractFactory::generate(input);
+        return generated;
```
pratikvn, line 225: I can this can be:
```suggestion
        return AbstractFactory::generate(input);
```

include/ginkgo/core/multigrid/amgx_pgm.hpp
```diff
@@ -0,0 +1,171 @@
...
+        : EnableRestrictProlong<AmgxPgm>(std::move(exec))
+    {}
+
+    explicit AmgxPgm(const Factory *factory,
+                     std::shared_ptr<const LinOp> system_matrix)
+        : EnableRestrictProlong<AmgxPgm>(factory->get_executor()),
+          parameters_{factory->get_parameters()},
+          system_matrix_{std::move(system_matrix)},
+          agg_(factory->get_executor(), system_matrix_->get_size()[0])
+    {
```
tcojean, line 152: Could you ensure that `max_unassigned_percent` is between 0.0 and 1.0?

include/ginkgo/core/multigrid/restrict_prolong.hpp
```diff
@@ -0,0 +1,328 @@
...
+
+
+namespace gko {
+namespace multigrid {
+
+
+/**
+ * The RestrictProlong class can be used to construct restrict_apply and
+ * prolong_applyadd. R is the restrict operator (fine level -> coarse
+ * level) and P is prolong operator (coarse level -> fine level).
```
tcojean, line 53:
```suggestion
 * level) and P is the prolong operator (coarse level -> fine level).
```

include/ginkgo/core/multigrid/restrict_prolong.hpp
```diff
@@ -0,0 +1,328 @@
...
+ * restrict_apply(b, x) -> x = R(b)
+ * prolong_applyadd(b, x) -> x = P(b) + x
+ *
+ * @ingroup RestrictProlong
```
tcojean, line 57: Maybe add a `@ingroup Multigrid` ?

include/ginkgo/core/multigrid/amgx_pgm.hpp
```diff
@@ -0,0 +1,171 @@
...
+ * elements.
+ * 1: get the strongest neighbor of each unaggregated element.
+ * 2: group the elements whose strongest neighbor is each other.
+ * repeating until reaching the given conditions. After that, assigns the
+ * unaggregated elements to aggregated group or stay alone.
+ *
+ * @tparam ValueType  precision of matrix elements
+ * @tparam IndexType  precision of matrix indexes
+ *
+ * @ingroup RestrictProlong
```
tcojean, line 68: Maybe also add a `@ingroup Multigrid`?

include/ginkgo/core/multigrid/amgx_pgm.hpp
```diff
@@ -0,0 +1,171 @@
...
+         * The maximum ratio of unassigned number.
+         */
+        double GKO_FACTORY_PARAMETER(max_unassigned_percentage, 0.05);
+
+        /**
+         * Use the deterministic assign_to_exist_agg method or not.
+         *
+         * If the deterministic is true, always get the same aggregated group
+         * from the same matrix. Otherwise, the aggregated group might be
+         * differenet depending on the execution ordering.
```
tcojean, line 130:
```suggestion
         * different depending on the execution ordering.
```

include/ginkgo/core/multigrid/restrict_prolong.hpp
```diff
@@ -0,0 +1,328 @@
...
+        GKO_ASSERT_EQUAL_ROWS(prolong_dim, x);
+        GKO_ASSERT_EQUAL_COLS(b, x);
+    }
+
+    /**
+     * Implementers of RestrictProlong should override this function instead
+     * of restrict_apply(const LinOp *, LinOp *).
+     *
+     * Performs the operation x = restrict(b), where op is the restrict
+     * operator.
```
tcojean, line 167:
```suggestion
     * Performs the operation x = restrict(b).
```

include/ginkgo/core/multigrid/restrict_prolong.hpp
```diff
@@ -0,0 +1,328 @@
...
+     * @param x  the output vector(s) where the result is stored
+     */
+    virtual void restrict_apply_impl(const LinOp *b, LinOp *x) const = 0;
+
+    /**
+     * Implementers of RestrictProlong should override this function instead
+     * of prolong_apply(const LinOp *, LinOp *).
+     *
+     * Performs the operation x = prolong(b), where op is the prolong
+     * operator.
```
tcojean, line 179:
```suggestion
     * Performs the operation x = prolong(b).
```

include/ginkgo/core/multigrid/restrict_prolong.hpp
```diff
@@ -0,0 +1,328 @@
...
+
+template <typename ConcreteFactory, typename ConcreteCriterion,
+          typename ParametersType,
+          typename PolymorphicBase = RestrictProlongFactory>
+using EnableDefaultRestrictProlongFactory =
+    EnableDefaultFactory<ConcreteFactory, ConcreteCriterion, ParametersType,
+                         PolymorphicBase>;
+
+
+template <typename ConcreteLinOp, typename PolymorphicBase = RestrictProlong>
```
tcojean, line 237: Some documentation?

include/ginkgo/core/multigrid/restrict_prolong.hpp
```diff
@@ -0,0 +1,328 @@
...
+        : EnableAbstractPolymorphicObject<RestrictProlong>(exec)
+    {}
+
+private:
+    std::shared_ptr<const LinOp> coarse_{};
+    size_type fine_dim_;
+    size_type coarse_dim_;
+};
+
+class RestrictProlongFactory
```
tcojean, line 215: Some documentation?

include/ginkgo/core/multigrid/restrict_prolong.hpp
```diff
@@ -0,0 +1,328 @@
...
+
+    std::unique_ptr<RestrictProlong> generate(
+        std::shared_ptr<const LinOp> input) const
+    {
+        return AbstractFactory::generate(input);
+    }
+};
+
+
+template <typename ConcreteFactory, typename ConcreteCriterion,
```
tcojean, line 229: Documentation?

include/ginkgo/core/multigrid/restrict_prolong.hpp
```diff
@@ -0,0 +1,328 @@
...
+        _factory_name, _restrict_prolong, _parameters_name##_type>;           \
+                                                                              \
+private:                                                                      \
+    _parameters_name##_type _parameters_name##_;                              \
+                                                                              \
+public:                                                                       \
+    static_assert(true,                                                       \
+                  "This assert is used to counter the false positive extra "  \
+                  "semi-colon warnings")
+
```
tcojean, line 323: missing blank line

tcojean (commented):
Some initial comments without looking yet into the algorithms

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -622,6 +622,25 @@ class DiagonalExtractable {
...
 
+
+/**
+ * The absolute of a LinOp implementing this interface can be computed.
+ * get_absolute get a new LinOp which contains absolute values from a LinOp.
+ *
+ *
+ * @ingroup LinOp
+ */
+class AbsoluteComputable {
```
pratikvn, line 633: Can we make this an interface ? `EnableAbsoluteComputation<ConcreteType>`. I would prefer this because then it can return the `ConcreteType` rather than a `LinOp` and we dont have to use `as` and dynamic casts everywhere.

tcojean: You mean a `mixin`. Yes, that does make sense.

yhmtsai: I thought it before, but it required the result type when the object is LinOp and it need to do dynamic_cast before using it.
```
// A is LinOp* currently
auto abs_obj = gko::as<AbsoluteComputable<csr<rmc_t, idx_t>>>(A);
auto abs_res = abs_obj->get_absolute()
```

reference/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,332 @@
...
+                       const Array<IndexType> &agg,
+                       matrix::Csr<ValueType, IndexType> *coarse)
+{
+    // agg[i] -> I, agg[j] -> J
+    const auto coarse_nrows = coarse->get_size()[0];
+    const auto source_nrows = source->get_size()[0];
+    const auto source_row_ptrs = source->get_const_row_ptrs();
+    const auto source_col_idxs = source->get_const_col_idxs();
+    const auto source_vals = source->get_const_values();
+    vector<map<IndexType, ValueType>> row_list(
```
pratikvn, line 291: If these are `gko` objects I would maybe prefer to use `gko::` here

reference/test/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,595 @@
...
+        cols[5] = 3;
+        cols[6] = 4;
+        cols[7] = 0;
+        cols[8] = 2;
+        cols[9] = 4;
+        cols[10] = 1;
+        cols[11] = 3;
+        cols[12] = 1;
+        cols[13] = 2;
+        cols[14] = 4;
```
pratikvn, line 167: Maybe it is better to create a dense matrix and then convert it for easier readability ? You could also use the [initialize method](https://github.com/ginkgo-project/ginkgo/blob/develop/include/ginkgo/core/matrix/dense.hpp#L658) instead.

pratikvn (approved):
Only some minor things. Otherwise, looks good to me!

core/multigrid/amgx_pgm.cpp
```diff
@@ -0,0 +1,193 @@
...
+    auto exec = this->get_executor();
+    const auto num = this->system_matrix_->get_size()[0];
+    Array<IndexType> strongest_neighbor(this->get_executor(), num);
+    Array<IndexType> intermediate_agg(this->get_executor(),
+                                      parameters_.deterministic * num);
+    // Only support csr matrix currently.
+    const matrix_type *amgxpgm_op = nullptr;
+    // Store the csr matrix if needed
+    auto amgxpgm_op_unique_ptr = matrix_type::create(exec);
+    if (!(amgxpgm_op =
```
upsj, line 106: nit: I would prefer if you could move this assignment out of the condition, since it is a bit difficult to read otherwise

core/multigrid/amgx_pgm.cpp
```diff
@@ -0,0 +1,193 @@
...
+            ->convert_to(amgxpgm_op_unique_ptr.get());
+        amgxpgm_op = amgxpgm_op_unique_ptr.get();
+    }
+
+    // Initial agg = -1
+    exec->run(amgx_pgm::make_fill_array(agg_.get_data(), agg_.get_num_elems(),
+                                        -one<IndexType>()));
+    size_type num_unagg{0};
+    size_type num_unagg_prev{0};
+    // TODO: if mtx is a hermitian matrix, weight_mtx = abs(mtx)
```
upsj, line 119: Is this TODO meant for later, to avoid the transposition if we know it is symmetric/hermitian?

core/multigrid/amgx_pgm.cpp
```diff
@@ -0,0 +1,193 @@
...
+    // Extract the diagonal value of matrix
+    auto diag = weight_mtx->extract_diagonal();
+    for (int i = 0; i < parameters_.max_iterations; i++) {
+        // Find the strongest neighbor of each row
+        exec->run(amgx_pgm::make_find_strongest_neighbor(
+            weight_mtx.get(), diag.get(), agg_, strongest_neighbor));
+        // Match edges
+        exec->run(amgx_pgm::make_match_edge(strongest_neighbor, agg_));
+        // Get the num_unagg
+        exec->run(amgx_pgm::make_count_unagg(agg_, &num_unagg));
```
upsj, line 138: nit: You could pass num_unagg as a reference instead of a pointer, which makes for a nicer interface that forbids passing nullptr :)

upsj (approved):
LGTM! I must say though that I am not 100% sure I like the whole RestrictProlong setup, since it basically circumvents our whole LinOp hierarchy, but we can maybe talk about that in more detail.

reference/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,332 @@
...
+                    matrix::Dense<ValueType> *x)
+{
+    auto x_vals = x->get_values();
+    const auto x_stride = x->get_stride();
+    auto x_dim = x->get_size();
+    for (size_type i = 0; i < x_dim[0]; i++) {
+        for (size_type j = 0; j < x_dim[1]; j++) {
+            x->at(i, j) = zero<ValueType>();
+        }
+    }
```
tcojean, line 77: Can't you use
```suggestion
    components::fill_array(exec, x->get_values(), x->get_num_stored_elements(),
                           zero<ValueType>());
```

reference/multigrid/amgx_pgm_kernels.cpp
tcojean: We don't have these? They should be `std` I believe.

tcojean (approved):
LGTM. Some small comments.

core/multigrid/amgx_pgm.cpp
```diff
@@ -0,0 +1,193 @@
...
+        // no new match, all match, or the ratio of num_unagg/num is lower
+        // than parameter.max_unassigned_percentage
+        if (num_unagg == 0 || num_unagg == num_unagg_prev ||
+            num_unagg < parameters_.max_unassigned_percentage * num) {
+            break;
+        }
+        num_unagg_prev = num_unagg;
+    }
+    // Handle the left unassign points
+    if (num_unagg != 0 && intermediate_agg.get_num_elems() > 0) {
```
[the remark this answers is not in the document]
tcojean: Similarly, for the related kernel with the internal logic which depends on the size of `intermediate_agg`, should we pass in the `deterministic` flag on top of the current arguments, in order to make the logic clearer?

reference/multigrid/amgx_pgm_kernels.cpp
upsj: These are typedefs using ExecutorAllocator

core/multigrid/amgx_pgm.cpp
yhmtsai: Yes, it will need the generic Linear Operator properties.
another way is to do a similar thing `is_hermitian and is_symmetric` like `skip_sorting` 
- is_symmetric: (A + A^T)/2 = A
- is_hermitian: (A + A^T)/2 = real(A)
- (Maybe) consider directed weight as non-directed weight: not need (A+A^T)/2. Not sure it makes sense or not.

reference/multigrid/amgx_pgm_kernels.cpp
yhmtsai: Yes, it is in `gko` namespace

reference/multigrid/amgx_pgm_kernels.cpp
yhmtsai: If x is a view, I can not use `fill_array`. Otherwise, it will touch the data out of the view.

reference/multigrid/amgx_pgm_kernels.cpp
tcojean: Thanks, I did not remember we had these.

yhmtsai:
@upsj It is possible to implement this in LinOp.
We have some discussion in https://github.com/ginkgo-project/ginkgo/issues/529
implement it in LinOp/LinOpFactory (implementation like ILU)
1. need another way to make sure the input is a CoarseFine not a Solver.
2. maybe use prolong->apply(x, b) as prolong_applyadd?

yhmtsai:
format!

yhmtsai:
format!

yhmtsai:
format!

include/ginkgo/core/multigrid/interface.hpp
```diff
@@ -0,0 +1,345 @@
...
+
+    /**
+     * Creates a RestrictOp with settings
+     *
+     * @param exec  Executor associated to the RestrictOp
+     * @param size  the size of RestrictOp.
+     *
+     * @note the multigrid is generated in the generation not the construction,
+     *       so needs to call `set_restrict_size` to set restrict size after
+     *       generation.
```
tcojean, line 90:
```suggestion
     *       so calling `set_restrict_size` is needed to set the restrict
     *       size after generation.
```

include/ginkgo/core/multigrid/interface.hpp
```diff
@@ -0,0 +1,345 @@
...
+        GKO_ASSERT_EQUAL_COLS(b, x);
+    }
+
+    std::shared_ptr<const Executor> exec_;
+    gko::dim<2> size_;
+};
+
+
+/**
+ * A class implementing this interface can be prolong operation.
```
tcojean, line 118: Similarly to before
```suggestion
 * This interface allows to implement the prolongation operation
 * of a multigrid method.
 *
 * Implementers of this interface need to overload both the
 * ProlongOp::prolong_apply_impl() and 
 * ProlongOp::prolong_applyadd_impl() methods.
```

include/ginkgo/core/multigrid/interface.hpp
```diff
@@ -0,0 +1,345 @@
...
+    }
+
+    std::shared_ptr<const Executor> exec_;
+    gko::dim<2> size_;
+    bool native_applyadd_;
+};
+
+
+/**
+ * A class implementing this interface can be coarse generation operation.
```
tcojean, line 207:
```suggestion
 * This class provides functionality for managing the coarse and fine operator
 * for a multigrid method.
```

include/ginkgo/core/multigrid/interface.hpp
```diff
@@ -0,0 +1,345 @@
...
+
+
+/**
+ * A class implementing this interface can contain restrict, prolong, coarse
+ * generation operation.
+ *
+ * @ingroup multigrid
+ */
+class MultigridLevel : public RestrictOp, public ProlongOp, public CoarseOp {
+protected:
```
tcojean, line 254: What's the use of this class if everything is protected? How do you use it? I think it would be useful to have more documentation for this class like the previous ones. Also, clarify what someone implementing this interface would need to overload.

Another aspect, while the previous classes do not need an advanced apply thus could not easily be a LinOp, this one could be a good candidate maybe? Though I'm not sure if there is much use in making this a LinOp.

include/ginkgo/core/multigrid/interface.hpp
```diff
@@ -0,0 +1,345 @@
...
+
+
+namespace gko {
+namespace multigrid {
+
+
+/**
+ * A class implementing this interface can be restrict operation.
+ *
+ * @ingroup multigrid
```
tcojean, line 53:
```suggestion
 * @ingroup Multigrid
```

include/ginkgo/core/multigrid/amgx_pgm.hpp
```diff
@@ -0,0 +1,194 @@
...
+    {
+        /**
+         * The maximum number of iterations.
+         */
+        unsigned GKO_FACTORY_PARAMETER(max_iterations, 15u);
+
+        /**
+         * The maximum ratio of unassigned number, which is valid in the
+         * interval 0.0 ~ 1.0. We use the same default value as NVIDIA AMGX
+         * Rerference Manual (October 2017, API Version 2).
```
tcojean, line 127:
```suggestion
         * Reference Manual (October 2017, API Version 2).
```

include/ginkgo/core/multigrid/amgx_pgm.hpp
```diff
@@ -0,0 +1,194 @@
...
+    {
+        return agg_.get_const_data();
+    }
+
+    GKO_CREATE_FACTORY_PARAMETERS(parameters, Factory)
+    {
+        /**
+         * The maximum number of iterations.
+         */
+        unsigned GKO_FACTORY_PARAMETER(max_iterations, 15u);
```
tcojean, line 122: Out of curiosity, does this come from somewhere?

include/ginkgo/core/multigrid/amgx_pgm.hpp
```diff
@@ -0,0 +1,175 @@
...
+
+    explicit AmgxPgm(const Factory *factory,
+                     std::shared_ptr<const LinOp> system_matrix)
+        : EnableRestrictProlong<AmgxPgm>(factory->get_executor()),
+          parameters_{factory->get_parameters()},
+          system_matrix_{std::move(system_matrix)},
+          agg_(factory->get_executor(), system_matrix_->get_size()[0])
+    {
+        GKO_ASSERT(parameters_.max_unassigned_percentage <= 1.0);
+        GKO_ASSERT(parameters_.max_unassigned_percentage >= 0.0);
```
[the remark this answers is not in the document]
tcojean: Good question.

core/multigrid/amgx_pgm.cpp
tcojean: Agree.

core/multigrid/amgx_pgm.cpp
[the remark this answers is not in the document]
tcojean: I feel like you had changed this previously but it's back to the previous version?

reference/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,332 @@
...
+}
+
+GKO_INSTANTIATE_FOR_EACH_INDEX_TYPE(GKO_DECLARE_AMGX_PGM_MATCH_EDGE_KERNEL);
+
+
+template <typename IndexType>
+void count_unagg(std::shared_ptr<const ReferenceExecutor> exec,
+                 const Array<IndexType> &agg, size_type *num_unagg)
+{
+    IndexType unagg = 0;
```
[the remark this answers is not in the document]
tcojean: You could have non matching types for `num_unagg` and `unagg`. I.e., int vs size_type. You should pick one and stick with it.

reference/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,332 @@
...
+        agg_map_vals[i] = 0;
+    }
+    for (size_type i = 0; i < num; i++) {
+        agg_map_vals[agg_vals[i]] = 1;
+    }
+    components::prefix_sum(exec, agg_map_vals, num + 1);
+    for (size_type i = 0; i < num; i++) {
+        agg_vals[i] = agg_map_vals[agg_vals[i]];
+    }
+    *num_agg = agg_map_vals[num];
```
tcojean, line 165: Here as well, `num_agg` is `size_type` while `agg_map_vals[num]` is `IndexType`.

reference/test/multigrid/amgx_pgm_kernels.cpp
tcojean: Agree. Another way would be to create a CSR matrix from scratch and pass three arrays for the values, column indices and row pointer separately. Same thing for the next ones. At least that would save a lot of space.

reference/test/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,644 @@
...
+    using Vec = typename TestFixture::Vec;
+    using value_type = typename TestFixture::value_type;
+    auto x = Vec::create_with_config_of(gko::lend(this->coarse_b));
+
+    gko::kernels::reference::amgx_pgm::restrict_apply(
+        this->exec, this->agg, this->fine_b.get(), x.get());
+
+    GKO_ASSERT_MTX_NEAR(x, this->restrict_ans, r<value_type>::value);
+}
+
```
tcojean, line 377:
```suggestion


```

reference/test/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,644 @@
...
+{
+    using value_type = typename TestFixture::value_type;
+    auto x = gko::clone(this->fine_x);
+
+    gko::kernels::reference::amgx_pgm::prolong_applyadd(
+        this->exec, this->agg, this->coarse_b.get(), x.get());
+
+    GKO_ASSERT_MTX_NEAR(x, this->prolong_ans, r<value_type>::value);
+}
+
```
tcojean, line 388:
```suggestion


```

reference/test/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,644 @@
...
+    gko::kernels::reference::amgx_pgm::match_edge(this->exec, snb, agg);
+
+    ASSERT_EQ(agg_val[0], 0);
+    ASSERT_EQ(agg_val[1], -1);
+    ASSERT_EQ(agg_val[2], 0);
+    ASSERT_EQ(agg_val[3], -1);
+    ASSERT_EQ(agg_val[4], -1);
+}
+
```
tcojean, line 413:
```suggestion


```

reference/test/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,644 @@
...
+
+TYPED_TEST(AmgxPgm, CoarseFineRestrictApply)
+{
+    auto amgx_pgm = this->amgxpgm_factory->generate(this->mtx);
+
+    // fine->coarse
+    using Vec = typename TestFixture::Vec;
+    using value_type = typename TestFixture::value_type;
+    auto x = Vec::create_with_config_of(gko::lend(this->coarse_b));
```
tcojean, line 476:
```suggestion
    using Vec = typename TestFixture::Vec;
    using value_type = typename TestFixture::value_type;
    auto x = Vec::create_with_config_of(gko::lend(this->coarse_b));
    
```

reference/test/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,644 @@
...
+    ASSERT_EQ(c[8], 2);
+    ASSERT_EQ(v[0], value_type{4});
+    ASSERT_EQ(v[1], value_type{-3});
+    ASSERT_EQ(v[2], value_type{-1});
+    ASSERT_EQ(v[3], value_type{-3});
+    ASSERT_EQ(v[4], value_type{5});
+    ASSERT_EQ(v[5], value_type{-1});
+    ASSERT_EQ(v[6], value_type{-2});
+    ASSERT_EQ(v[7], value_type{-2});
+    ASSERT_EQ(v[8], value_type{5});
```
tcojean, line 640: Maybe create a CSR or dense matrix with these and compare against it instead? That could probably save some lines.

tcojean (approved):
LGTM! I only have a few comments.

include/ginkgo/core/multigrid/amgx_pgm.hpp
yhmtsai: It is also from the NVIDIA AMGX Reference Manual

include/ginkgo/core/multigrid/interface.hpp
yhmtsai: I tried LinOp but it is failed.
When I use EnableLinOP, Ginkgo will create it as a real object in some class.
However, it still uses some virtual function from the previous Op but does not implement them, so it can not be real object.

The use case shorten the usage of the class.
dynamic_cast to this class is enough for all operations. do not need to cast RestrictOp for restrict, ProlonOp for prolong,..., three casting for all operation.
Is any idea to handle this situation?
This class is required to be used in the complete multigrid level.

include/ginkgo/core/multigrid/amgx_pgm.hpp
pratikvn: Can you maybe put a link to that manual here ?

include/ginkgo/core/multigrid/interface.hpp
pratikvn: I think it makes sense, but as Terry mentioned maybe you can write a small code snippet as to how one can use this class to construct a multigrid solver and this snippet to the this `MultigridLevel` class documentation. I mean some code snippets [similar to the one in linop](https://github.com/ginkgo-project/ginkgo/blob/develop/include/ginkgo/core/base/lin_op.hpp#L892)

include/ginkgo/core/multigrid/interface.hpp
```diff
@@ -0,0 +1,364 @@
...
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+#ifndef GKO_PUBLIC_CORE_MULTIGRID_INTERFACE_HPP_
+#define GKO_PUBLIC_CORE_MULTIGRID_INTERFACE_HPP_
+
```
pratikvn, line 35: I am not too sure if `interface.hpp` is the correct name for this file. Maybe something like `multigrid_base.hpp` is better ? Similar to `reordering_base.hpp` and so on.

pratikvn (approved):
Only some minor doc clarifications, otherwise LGTM!

include/ginkgo/core/multigrid/amgx_pgm.hpp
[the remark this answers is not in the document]
yhmtsai: it works because the percentage = 0 means to aggregate all pairs (under iter limit)
if it is equal to 1, the current implementation will run the match step once because only check the percent after the match step.

core/multigrid/amgx_pgm.cpp
yhmtsai: I prefer using the pointer here such that I can easily understand the variable is output.
If do not have strong opinion, I will stay this version.
If it is necessary, we can change it latter

yhmtsai:
I use `read` for the Csr setting in the test because `initialize` seems to contain the zero elements.
If there is any way to initialize Csr matrix, I can also try it.

include/ginkgo/core/multigrid/interface.hpp
yhmtsai: should I keep the `multigrid_` part?
There is already multigrid parent folder, but only `base.hpp` is maybe confusing...

upsj (commented):
Thanks for the modifications, again you did a really nice job with this implementation. The only feedback I have is related to the MultigridLevel + Restrict/Prolong/CoarseOp hierarchy. Right now, all of these operators are basically simplified re-implementations of LinOp. Do we get any benefit from them having their own interface instead of using LinOp?
For building MultigridLevel, I would suggest preferring composition over inheritance, i.e. if we can agree on what a MultigridLevel should do as a LinOp (something like restrict -> coarse -> prolong), we could add member functions get_restrict, get_coarse and get_prolong which basically provide names for the wrapped LinOps (which could internally also use the same data, just be wrappers  around everything).
This way, every one of the operators could be used in combination with all the other facilities that Ginkgo provides (e.g. building a complete V-cycle out of a Composition as an example for its versatility)

yhmtsai:
By default, they should be implemented with EnableLinOp such that use LinOpFactory helper macro.
Do you mean to implement the MultigridLevel as Composition special cases?
It will require all component to be implemented as LinOp separately, right?

yhmtsai:
@upsj @tcojean @pratikvn
I finish the structure draft
I do not clean the code and add all documentation yet, so I collect some points and corresponding file links and hope it can help review the design.
I will extract mapping to another pr if everyone are comfortable with this structure.

1. MultigridLevel will use composition (3 op)
https://github.com/ginkgo-project/ginkgo/blob/coarse_fine_agg/include/ginkgo/core/multigrid/multigrid_base.hpp#L315

2. implement restrict and prolong as LinOp (naming suggestion?)
https://github.com/ginkgo-project/ginkgo/blob/coarse_fine_agg/include/ginkgo/core/multigrid/mapping.hpp

3. any real multigrid class needs to inherit the EnableLinOp and MultigridLevel by itself.
Inherit EnableLinOp for full clone functionality.
Inherit MultigridLevel for the consistent usage in multigrid or additional information (maybe)
https://github.com/ginkgo-project/ginkgo/blob/coarse_fine_agg/include/ginkgo/core/multigrid/amgx_pgm.hpp#L76
https://github.com/ginkgo-project/ginkgo/blob/coarse_fine_agg/include/ginkgo/core/multigrid/amgx_pgm.hpp#L154

4. ApplyAddable Interface. it is for usage of prolong_op applyadd operation
https://github.com/ginkgo-project/ginkgo/blob/coarse_fine_agg/include/ginkgo/core/base/lin_op.hpp#L657

5. UseComposition
https://github.com/ginkgo-project/ginkgo/blob/coarse_fine_agg/include/ginkgo/core/base/composition.hpp#L170

6. the current usage of restrictapply and prolongapplyadd
https://github.com/ginkgo-project/ginkgo/blob/coarse_fine_agg/reference/test/multigrid/amgx_pgm_kernels.cpp#L405

include/ginkgo/core/multigrid/amgx_pgm.hpp
```diff
@@ -0,0 +1,198 @@
...
+     * @note This is the constant version of the function, which can be
+     *       significantly more memory efficient than the non-constant version,
+     *       so always prefer this version.
+     */
+    const IndexType *get_const_agg() const noexcept
+    {
+        return agg_.get_const_data();
+    }
+
+    GKO_CREATE_FACTORY_PARAMETERS(parameters, Factory)
```
Slaedr, line 118: I guess it would be possible to add a `cycle_type` parameter in future without breaking interface.

include/ginkgo/core/multigrid/multigrid_base.hpp
```diff
@@ -0,0 +1,381 @@
...
+/**
+ * This interface allows to implement the restrict operation of
+ * a multigrid method.
+ *
+ * Implementers of this interface only need to overload the
+ * RestrictOp::restrict_apply_impl() method.
+ *
+ * @ingroup Multigrid
+ */
+class RestrictOp {
```
Slaedr, line 62: I believe this class, as well as `ProlongOp` and `CoarseOp` below, are not being used anywhere. Would it be better to remove them?

core/multigrid/amgx_pgm.cpp
```diff
@@ -0,0 +1,214 @@
...
+    auto restrict_op = Mapping<ValueType, IndexType>::create(
+        exec, gko::dim<2>{coarse_dim, fine_dim}, agg_);
+    auto prolong_op = Mapping<ValueType, IndexType>::create(
+        exec, gko::dim<2>{fine_dim, coarse_dim}, agg_, false);
+    this->set_composition(share(prolong_op), coarse_matrix, share(restrict_op));
+}
+
+
+template <typename ValueType, typename IndexType>
+void AmgxPgm<ValueType, IndexType>::restrict_apply_impl(const LinOp *b,
```
Slaedr, line 174: This function seems to be redundant. Can it be removed?

core/multigrid/amgx_pgm.cpp
```diff
@@ -0,0 +1,214 @@
...
+{
+    auto exec = this->get_executor();
+    exec->run(amgx_pgm::make_prolong_applyadd(agg_,
+                                              as<matrix::Dense<ValueType>>(b),
+                                              as<matrix::Dense<ValueType>>(x)));
+}
+
+
+template <typename ValueType, typename IndexType>
+void AmgxPgm<ValueType, IndexType>::prolong_apply_impl(const LinOp *b,
```
Slaedr, line 196: Same here - redundant?

reference/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,332 @@
...
+/**
+ * @brief The AMGX_PGM solver namespace.
+ *
+ * @ingroup amgx_pgm
+ */
+namespace amgx_pgm {
+
+
+template <typename ValueType, typename IndexType>
+void restrict_apply(std::shared_ptr<const ReferenceExecutor> exec,
```
Slaedr, line 65: As I mentioned for `AmgxPgm` in core, is this function needed? As far as I can tell, it's only being called in `AmgxPgm::restrict_apply_impl`, which is then not called anywhere.

reference/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,332 @@
...
+        }
+    }
+}
+
+GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
+    GKO_DECLARE_AMGX_PGM_RESTRICT_APPLY_KERNEL);
+
+
+template <typename ValueType, typename IndexType>
+void prolong_applyadd(std::shared_ptr<const ReferenceExecutor> exec,
```
Slaedr, line 91: Similar to the previous comment, this function also appears redundant.

Slaedr (changes requested):
Great work! From my understanding, `MultigridLevel` is reasonably generic and can be used for different types of multigrid methods, but all other classes are pretty specific to AMGx, which should be fine. Also, I guess only 2-level multigrid is there at the moment, and multi-level will follow when you introduce the preconditioner class in future.

I'm a bit confused about a couple of things though.
(1) There seem to be multiple implementations of the restriction and prolongation operations using the integer mapping array (`agg_`, typically). One set of operations is in the `AmgxPgm` class and kernels and the other set is in the `Mapping` class and kernels. The one that's actually used in `AmgxPgm::apply_impl` is the set of functions in `Mapping`. Can we remove the redundant functions in `AmgxPgm`?
(2) I don't see the need for `RestrictOp`, `ProlongOp` and `CoarseOp` in `multigrid_base.hpp`.

Apart from that there are some minor things.

upsj (approved):
LGTM! Can we reuse the advanced apply support instead of adding partial apply2 support to the LinOps?

include/ginkgo/core/multigrid/amgx_pgm.hpp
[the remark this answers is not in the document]
pratikvn: okay, makes sense. thanks.

include/ginkgo/core/multigrid/mapping.hpp
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    void apply_impl(const LinOp *b, LinOp *x) const override
+    {
+        auto dense_x = as<matrix::Dense<ValueType>>(x);
+
+        // TODO: write dense fill kernel
+        auto zero = gko::initialize<matrix::Dense<ValueType>>(
+            {gko::zero<ValueType>()}, this->get_executor());
+        dense_x->scale(lend(zero));
```
pratikvn, line 131: You can use the dense fill now.

pratikvn (commented):
LGTM!

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -654,6 +654,28 @@ class Preconditionable {
 };
 
 
+class ApplyAddable {
+public:
+    void apply2(const LinOp *b, LinOp *x) const
```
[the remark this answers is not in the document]
Slaedr: Indeed, is there any advantage of having a separate `apply2` in comparison to an advanced `apply`?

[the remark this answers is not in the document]
yhmtsai: apply2 is to fit the coo already existed interface. I agree applyadd is a more meaningful name.
I thought advanced_apply before, too.
We need to use dense to represent 1 now.
it requires us to copy the val to the host first and check the value.
apply2 is helpful in the cases which needs atomic add in the end.
we do not need scale or fill zero in the beginning.

include/ginkgo/core/multigrid/mapping.hpp
yhmtsai: Yes. In fact, using scale is not always correct here if considering the initial vector of b is nan.
That's why I had duplicated work as you did :)

include/ginkgo/core/base/lin_op.hpp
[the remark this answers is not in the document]
upsj: Okay, I wasn't aware that your implementation uses atomics. Would it be possible to do this without them? The data structure you are using here (mapping row indices to their coarse index) doesn't look ideal for parallel processing to me, what if you "transposed" the structure instead to have a CSR-style small set of fine rows for each coarse row? Then you would even be able to reuse existing CSR kernels, using a constant 1 accessor for the values.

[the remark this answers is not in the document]
yhmtsai: It depends on the implementation. classical csr does not need atomic but load-balance csr does.
For coare->fine, it is column index of csr.
for fine->coarse, it is row index of csc.
It is good point. could you also point me which accessor I can use here?

| native impl        | apply_add  (atomic_add)     | apply               |
|--------------------|-----------------|---------------------|
| y = Ax + y         | 3n              | 3n (advanced_apply) |
| y = Ax             | n(fill) + 3n    | 2n                  |
| y = Ax + beta(1) y | 2n (scale) + 3n | 3n                  |

If using apply, it needs addition of 3 vectors in the end, so at least 4n with new kernel or 6n with two addition kernel.
If using apply_add, it needs addition of 2 vectors (3n)

apply_add method is to reduce the memory footprint for those implemented by atomic_add.
In this sense, how about do we add the special case for dense {1, 1} as scalar, consider the empty dense as one, additional flag of dense for 1?

include/ginkgo/core/multigrid/amgx_pgm.hpp
yhmtsai: what do you mean `cycle_type` here? I think `cycle_type` should be in the multigrid solver not the multigrid level generator

reference/multigrid/amgx_pgm_kernels.cpp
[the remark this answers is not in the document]
yhmtsai: I use IndexType for all

reference/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,332 @@
...
+            for (auto idx = row_ptrs[row]; idx < row_ptrs[row + 1]; idx++) {
+                auto col = col_idxs[idx];
+                if (col == row) {
+                    continue;
+                }
+                auto weight =
+                    vals[idx] / max(abs(diag_vals[row]), abs(diag_vals[col]));
+                if (agg.get_const_data()[col] == -1 &&
+                    (weight > max_weight_unagg ||
+                     (weight == max_weight_unagg && col > strongest_unagg))) {
```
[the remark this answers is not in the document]
yhmtsai: I think it should pick the max weight in the machine representation.

Slaedr (approved):
LGTM! Thanks for addressing all the points. I just have a few minor comments now.

common/components/fill_array.hpp.inc
```diff
@@ -45,4 +45,15 @@ __global__ __launch_bounds__(default_block_size) void fill_array(
 }
 
 
+template <typename ValueType>
+__global__ __launch_bounds__(default_block_size) void fill_seq_array(
+    size_type n, ValueType *__restrict__ array)
+{
+    const auto tidx = thread::get_thread_id_flat();
+    if (tidx < n) {
+        array[tidx] = tidx;
```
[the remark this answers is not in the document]
upsj: I think this is only used for index types, so we should be fine.

include/ginkgo/core/base/composition.hpp
```diff
@@ -167,6 +167,60 @@ class Composition : public EnableLinOp<Composition<ValueType>>,
...
+    }
+
+    /**
+     * Returns the operator at index-th poistion of composition
+     *
+     * @return index-th operator
+     *
+     * @note when this composition is not set, this function always returns
+     *       nullptr. However, when this composition is set, it will throw
+     *       exception when exceeding index.
```
tcojean, line 197: I guess we are not always consistent with this, but maybe add an `@throws` doxygen comment then?

include/ginkgo/core/multigrid/amgx_pgm.hpp
```diff
@@ -0,0 +1,191 @@
...
+#include <ginkgo/core/multigrid/multigrid_level.hpp>
+
+namespace gko {
+namespace multigrid {
+
+
+/**
+ * Amgx parallel graph match (AmgxPgm) is the aggregate method introduced in the
+ * paper M. Naumov et al., "AmgX: A Library for GPU Accelerated Algebraic
+ * Multigrid and Preconditioned Iterative Methods". Current implemenation only
```
tcojean, line 55:
```suggestion
 * Multigrid and Preconditioned Iterative Methods". Current implementation only
```

include/ginkgo/core/multigrid/amgx_pgm.hpp
```diff
@@ -0,0 +1,191 @@
...
+     */
+    std::shared_ptr<const LinOp> get_system_matrix() const
+    {
+        return system_matrix_;
+    }
+
+    /**
+     * Returns the aggregate group.
+     *
+     * Aggregate group whose size is same as the number of rows stores the
```
tcojean, line 97:
```suggestion
     * Aggregate group whose size is same as the number of rows. Stores the
```

include/ginkgo/core/multigrid/amgx_pgm.hpp
```diff
@@ -0,0 +1,191 @@
...
+         * interval 0.0 ~ 1.0. We use the same default value as NVIDIA AMGX
+         * Reference Manual (October 2017, API Version 2,
+         * https://github.com/NVIDIA/AMGX/blob/main/doc/AMGX_Reference.pdf).
+         */
+        double GKO_FACTORY_PARAMETER(max_unassigned_ratio, 0.05);
+
+        /**
+         * Use the deterministic assign_to_exist_agg method or not.
+         *
+         * If the deterministic is true, always get the same aggregated group
```
tcojean, line 137:
```suggestion
         * If deterministic is set to true, always get the same aggregated group
```

include/ginkgo/core/multigrid/multigrid_level.hpp
```diff
@@ -0,0 +1,148 @@
...
+#include <functional>
+#include <memory>
+
+
+#include <ginkgo/core/base/abstract_factory.hpp>
+#include <ginkgo/core/base/composition.hpp>
+#include <ginkgo/core/base/exception_helpers.hpp>
+#include <ginkgo/core/base/lin_op.hpp>
+#include <ginkgo/core/base/utils.hpp>
+#include <ginkgo/core/matrix/dense.hpp>
```
tcojean, line 46:
```suggestion
```

include/ginkgo/core/multigrid/multigrid_level.hpp
```diff
@@ -0,0 +1,148 @@
...
+     * @return  the prolong operator.
+     */
+    std::shared_ptr<const LinOp> get_prolong_op() const
+    {
+        return this->get_operator_at(0);
+    }
+
+protected:
+    /**
+     * Sets the multigrid level information.
```
tcojean, line 107: Minor: maybe add a comment stating that the parameters are stored in that order in the composition.

include/ginkgo/core/multigrid/multigrid_level.hpp
```diff
@@ -0,0 +1,148 @@
...
+        // check mg_size is the same as fine_size
+        this->set_composition(prolong_op, coarse_op, restrict_op);
+    }
+
+    explicit MultigridLevel() {}
+
+    /**
+     * Creates a MultigridLevel with the given fine operator
+     *
+     * @param fine_op  The fine operotor associated to the multigrid_level
```
tcojean, line 129:
```suggestion
     * @param fine_op  The fine operator associated to the multigrid_level
```

include/ginkgo/core/multigrid/multigrid_level.hpp
```diff
@@ -0,0 +1,148 @@
...
+     *
+     * @note the multigrid is generated in the generation not the construction,
+     *       so needs to call `set_multigrid_level` to set corresponding
+     *       information after generation.
```
tcojean, line 133:
```suggestion
     *       so the user needs to call `set_multigrid_level` to set the 
     *       corresponding information after generation.
```

tcojean (approved):
LGTM, only some small documentation suggestions.

dpcpp/test/components/fill_array.cpp
```diff
@@ -79,8 +82,19 @@ TYPED_TEST(FillArray, EqualsReference)
     using T = typename TestFixture::value_type;
     gko::kernels::dpcpp::components::fill_array(
         this->exec, this->dvals.get_data(), this->total_size, T(1523));
+
     GKO_ASSERT_ARRAY_EQ(this->vals, this->dvals);
 }
 
 
+TYPED_TEST(FillArray, EqualsReference)
```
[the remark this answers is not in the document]
tcojean: Good point, DPC++ compilation fails due to this.

include/ginkgo/core/base/lin_op.hpp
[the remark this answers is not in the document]
yhmtsai: Good catch. I forgot deleting it. I think we will have some discussion about this in the future when needed.
It is not urgent and we have many things to discuss, so I use csr way now.
Changing it to different matrix format does not change the interface.

include/ginkgo/core/multigrid/amgx_pgm.hpp
```diff
@@ -0,0 +1,191 @@
...
+namespace multigrid {
+
+
+/**
+ * Amgx parallel graph match (AmgxPgm) is the aggregate method introduced in the
+ * paper M. Naumov et al., "AmgX: A Library for GPU Accelerated Algebraic
+ * Multigrid and Preconditioned Iterative Methods". Current implementation only
+ * contains size = 2 version.
+ *
+ * AmgxPgm creates the aggreagate group according to the matrix value not the
```
pratikvn, line 58:
```suggestion
 * AmgxPgm creates the aggregate group according to the matrix value not the
```

pratikvn (approved):
LGTM!

include/ginkgo/core/multigrid/multigrid_level.hpp
```diff
@@ -0,0 +1,148 @@
...
+ *
+ * A class implementing this interface should use the
+ * this->get_compositions()->apply(...) as its own apply, which represents
+ * op(b) = prolong(coarse(restrict(b))). All MultigridLevel generator should
+ * inherit from this class.
+ *
+ * @ingroup Multigrid
+ */
+template <typename ValueType>
+class MultigridLevel : public UseComposition<ValueType> {
```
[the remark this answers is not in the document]
yhmtsai: Yes. the level's coarse op should be the next level's fine op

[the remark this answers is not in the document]
pratikvn: Okay, thanks

include/ginkgo/core/multigrid/multigrid_level.hpp
```diff
@@ -52,51 +52,75 @@ namespace multigrid {
...
  * This class represents two levels in a multigrid hierarchy.
  *
- * A class implementing this interface should use the
- * this->get_compositions()->apply(...) as its own apply, which represents
- * op(b) = prolong(coarse(restrict(b))). All MultigridLevel generator should
- * inherit from this class.
- *
- * @ingroup Multigrid
+ * The MultigridLevel is an interface that allows to get the individual
+ * components of multigrid level. Each Linop for MultigridLevel should inherit
```
Slaedr, line 56:
```suggestion
 * components of multigrid level. Each implementation of a multigrid level should inherit
```

include/ginkgo/core/multigrid/multigrid_level.hpp
```diff
@@ -52,51 +52,75 @@ namespace multigrid {
...
+ * from this interface. Use EnableMultigridLevel<ValueType> to implement this
+ * interface with composition by default.
  */
-template <typename ValueType>
-class MultigridLevel : public UseComposition<ValueType> {
+class MultigridLevel {
```
Slaedr, line 60: Why is a `MultigridLevel` without `ValueType` needed? Did you realize this was necessary for another pull request?

Slaedr (commented):
What's the main motivation for having a layer that does not depend on `ValueType`?

yhmtsai:
I would like to store all generated mg_level in the std::vector in multigrid solver.
I do not consider how to support the mix-precision inside multigrid clear yet, but I think storing `MultigridLevel` should be easier than `MultigridLevel<ValueType>`.
Otherwise, I need store them into LinOp and need dynamic_cast when using them.

only one function, `get_composition()`, is not in MultigridLevel Interface.
it should be acceptable because we do not use it in the multigrid solver.

pratikvn (approved):
LGTM!

PR #533, opened 2020-05-14
Improve compilation speed

tcojean:
Improves the compilation speed of our CI system by using the ninja build system.

Also generate a new AMD container with updated compilers.


tcojean:
This is now ready for review. I also pushed a fix for #524 which uses g++ for linking instead of nvcc. With my hand tests, that fixes the issue. Hopefully it does not create new ones.

tcojean:
This MR on gitlab also needs approval to replace the temporary container names:
https://gitlab.com/ginkgo-project/ginkgo-containers/-/merge_requests/5

PR #540, opened 2020-05-21
Optimize Composition intermediate storage

upsj:
Only store the minimal amount of memory necessary for Composition. This is either the maximum sum of input and output dimension for all but the last intermediate operator, or the output dimension
of the last intermediate operator if there is only one operator.

### DIFF
--- a/core/base/composition.cpp
+++ b/core/base/composition.cpp
@@ -89,7 +89,7 @@ std::unique_ptr<LinOp> apply_inner_operators(
         operators[i]->apply(lend(in), lend(out));
     }
 
-    return std::move(out);
+    return out;
 }
 
 
--- a/core/base/composition.cpp
+++ b/core/base/composition.cpp
@@ -89,7 +89,7 @@ std::unique_ptr<LinOp> apply_inner_operators(
         operators[i]->apply(lend(in), lend(out));
     }
 
-    return out;
+    return std::move(out);
 }
 
 
--- a/core/base/composition.cpp
+++ b/core/base/composition.cpp
@@ -33,50 +33,75 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/base/composition.hpp>
 
 
+#include <algorithm>
+
+
 #include <ginkgo/core/matrix/dense.hpp>
 
 
 namespace gko {
-namespace {
 
 
-template <typename ValueType, typename OpIterator, typename VecIterator>
-inline void allocate_vectors(OpIterator begin, OpIterator end, VecIterator res)
-{
-    for (auto it = begin; it != end; ++it) {
-        if (*res == nullptr || (*res)->get_size()[0] != (*it)->get_size()[0]) {
-            *res = matrix::Dense<ValueType>::create(
-                (*it)->get_executor(), gko::dim<2>{(*it)->get_size()[0], 1});
-        }
-        ++res;
-    }
-}
-
-
-inline const LinOp *apply_inner_operators(
+template <typename ValueType>
+std::unique_ptr<LinOp> apply_inner_operators(
     const std::vector<std::shared_ptr<const LinOp>> &operators,
-    const std::vector<std::unique_ptr<LinOp>> &intermediate, const LinOp *rhs)
+    Array<ValueType> &storage, const LinOp *rhs)
 {
-    for (auto i = operators.size() - 1; i > 0u; --i) {
-        auto solution = lend(intermediate[i - 1]);
-        operators[i]->apply(rhs, solution);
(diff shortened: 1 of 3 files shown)


core/base/composition.cpp
```diff
@@ -33,63 +33,89 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+                            num_rhs);
+        // apply operator
+        operators[i]->apply(lend(in), lend(out));
     }
-    return rhs;
-}
-
 
-}  // namespace
+    return std::move(out);
```
thoasm, line 92: nit: This move can hinder return value optimization. Since you are returning a `std::unique_ptr`, you can simply remove the move (you need the move in case you return a `std::shared_ptr` if `out` was a `std::unique_ptr` because there is no implicit cast for it):
```suggestion
    return out;
```

reference/test/base/composition.cpp
```diff
@@ -105,4 +118,124 @@ TYPED_TEST(Composition, AppliesLinearCombinationToVector)
...
+    auto alpha = gko::initialize<Mtx>({3.0}, this->exec);
+    auto beta = gko::initialize<Mtx>({-1.0}, this->exec);
+    auto x = clone(this->identity);
+    auto res = clone(x);
+
+    cmp->apply(lend(alpha), lend(x), lend(beta), lend(res));
+
+    GKO_ASSERT_MTX_NEAR(res, l({{161.0, 276.0}, {81.0, 137.0}}),
+                        r<TypeParam>::value);
+}
```
thoasm, line 238: Very nice testing and extensive!
However, I am missing the most basic case (a composition having only a single operator), which was not already present for some reason.

core/base/composition.cpp
```diff
@@ -33,63 +33,89 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+    using Dense = matrix::Dense<ValueType>;
+    // determine amount of necessary storage:
+    // maximum sum of two subsequent intermediate vectors
+    // (and the out dimension of the last op if we only have one operator)
+    auto num_rhs = rhs->get_size()[1];
+    auto max_intermediate_size = std::accumulate(
+        begin(operators) + 1, end(operators) - 1,
+        operators.back()->get_size()[0],
+        [](size_type acc, std::shared_ptr<const LinOp> op) {
+            return std::max(acc, op->get_size()[0] + op->get_size()[1]);
```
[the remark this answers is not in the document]
upsj: Yes, size[1] gives you the dimension of the input and size[0] the dimension of the output. So the maximum intermediate storage is the sum of both (or just size[0] in case of the first operator that we apply, i.e. operators.back())

core/base/composition.cpp
```diff
@@ -33,63 +33,89 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+        });
+    auto storage_size = max_intermediate_size * num_rhs;
```
yhmtsai, line 61: `max_intermediate_size` can be calculated in Composition creation.

reference/test/base/composition.cpp
```diff
@@ -105,4 +154,124 @@ TYPED_TEST(Composition, AppliesLinearCombinationToVector)
...
+
```
yhmtsai, line 275: I would like to add a test whose dim[0] (any -> 3 -> 2 -> 2) to make sure the reverse storage index correctly.

core/base/composition.cpp
upsj: I tried, but due to the variadic constructor, it is impossible to determine if the constructor call is the final one, thus making it impossible to "ignore" the first operator. So I decided to do it here, since the overhead is minimal and we don't need to duplicate any code.

reference/test/base/composition.cpp
upsj: Do you think adding another identity to the *Longest tests would be enough?

yhmtsai: Yes, adding another identity in the end should be enough

core/base/composition.cpp
yhmtsai: I see.

core/base/composition.cpp
upsj: Sorry, I have to revert that change: Intel 2018 doesn't like it.
The issue is most likely the type mismatch unique_ptr<LinOp> vs. unique_ptr<Dense>, so actually no RVO/elision can take place after all.

reference/test/base/composition.cpp
upsj: done

PR #543, opened 2020-05-22
Add support for clang-cuda

upsj:
This PR takes the first steps to full clang-cuda support in Ginkgo. It works with the current CMake master version and (slightly patched) clang-10 (https://github.com/llvm/llvm-project/commit/8e20516540444618ad32dd11e835c05804053697)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -380,7 +380,7 @@ build/cuda101/intel/cuda/debug/static:
 # clang-cuda with cuda 10.1 and friends
 build/clang-cuda101/gcc/all/release/shared:
   <<: *default_build_with_test
-  image: localhost:5000/add_clang_cuda_69d04ab3_gko-cuda101-gnu8-llvm10-intel2019
+  image: localhost:5000/gko-cuda101-gnu8-llvm10-intel2019
   variables:
     <<: *default_variables
     CUDA_COMPILER: "clang++"
@@ -399,7 +399,7 @@ build/clang-cuda101/gcc/all/release/shared:
 
 build/clang-cuda101/clang/cuda/debug/static:
   <<: *default_build_with_test
-  image: localhost:5000/add_clang_cuda_69d04ab3_gko-cuda101-gnu8-llvm10-intel2019
+  image: localhost:5000/gko-cuda101-gnu8-llvm10-intel2019
   variables:
     <<: *default_variables
     C_COMPILER: "clang"
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -15,6 +15,7 @@ stages:
   BENCHMARK_SERVER: "FINECI"
   C_COMPILER: "gcc"
   CXX_COMPILER: "g++"
+  CUDA_COMPILER: "nvcc"
   BUILD_TYPE: "Debug"
   BUILD_SHARED_LIBS: "ON"
   BUILD_REFERENCE: "ON"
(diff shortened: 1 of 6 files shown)


upsj:
I'll mark this as WIP until the issues with clang-cuda and HIP are solved

upsj:
Finally two issues: FindThreads doesn't deal with clang-cuda correctly ([soon to be fixed](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/4781)) and the CUDA compiler was not propagated on installation, so it needed to be set in GinkgoConfig.

upsj:
Since clang-cuda doesn't use a host compiler, the incompatibilities between Intel and clang/gcc symbol generation for constexpr globals ([CI job](https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/564751176)) makes the clang-cuda/intel combination infeasible.

.gitlab-ci.yml
```diff
@@ -329,6 +367,42 @@ build/cuda101/intel/cuda/debug/static:
...
     BUILD_TYPE: "Debug"
+    BUILD_SHARED_LIBS: "OFF"
+    CUDA_ARCH: 35
+  tags:
+    - private_ci
+    - cuda
+    - gpu
+
+# clang-cuda with cuda 10.1 and friends
+build/clang-cuda101/gcc/all/release/shared:
```
pratikvn, line 378: No debug /static builds ?

upsj: I had some initially, but removed them because clang-cuda really takes the "no optimizations" thing to the limit. But now that we only have two jobs, I guess it couldn't hurt.

upsj: I just realized that one of the jobs was mis-named anyways, it was a static build after all

pratikvn: Yeah, I think it should be fine. Unless it does really take the same time as 5 other equivalent jobs.

upsj: It's really not that bad, the test only takes 3x the usual time.

.gitlab-ci.yml
```diff
@@ -365,10 +367,43 @@ build/cuda101/intel/cuda/debug/static:
...
+    CXX_COMPILER: "clang++"
+    CUDA_COMPILER: "clang++"
+    BUILD_OMP: "ON"
+    BUILD_CUDA: "ON"
+    BUILD_TYPE: "Debug"
+    BUILD_SHARED_LIBS: "OFF"
     CUDA_ARCH: 35
-  only:
-    variables:
-      - $RUN_CI_TAG
```
tcojean, line 371: This should probably have stayed? Also, you will need to add this for the clang-cuda jobs.

examples/custom-matrix-format/CMakeLists.txt
```diff
@@ -8,4 +8,6 @@ if (GINKGO_BUILD_CUDA AND GINKGO_BUILD_OMP)
     target_link_libraries(custom-matrix-format ginkgo)
     target_include_directories(custom-matrix-format PRIVATE
         ${PROJECT_SOURCE_DIR} ${CMAKE_CURRENT_SOURCE_DIR})
+    # workaround for clang-cuda/g++ interaction
+    set_target_properties(custom-matrix-format PROPERTIES POSITION_INDEPENDENT_CODE ON)
```
tcojean, line 12: Just a comment here, in general my understanding is that this makes the code more robust but could create some performance hit. Nonetheless, it feels like position independent code is in general a good idea. I always wondered whether we should make this a default or not. I don't think it's the best place to discuss it, but if anyone has a strong opinion I would like to hear it.

upsj: This is only an issue because for some reason, g++ expects relocatable code from the CUDA file, but clang-cuda generates static code. For shared/static libraries, I think CMake already takes care of the PIC property.

tcojean: For that issue that's fine, what I meant was generally whether we should or not always put this flag on anyway.

upsj: I'm not really sure we need it, the only place where I actually had to manually set this was when generating an object library or an intermediate library that was being used to (statically) build a larger library. Do you have a situation in mind where this would be necessary?

PR #550, opened 2020-05-27
Generate initial guess and rhs instead of reading them

fritzgoebel:
This makes the test matrix exchangeable.


examples/ir-ilu-preconditioned-solver/ir-ilu-preconditioned-solver.cpp
```diff
@@ -76,8 +76,16 @@ int main(int argc, char *argv[])
 
     // Read data
     auto A = gko::share(gko::read<mtx>(std::ifstream("data/A.mtx"), exec));
-    auto b = gko::read<vec>(std::ifstream("data/b.mtx"), exec);
-    auto x = gko::read<vec>(std::ifstream("data/x0.mtx"), exec);
+    // Create RHS and initial guess as 1
+    gko::size_type size = A->get_size()[0];
+    auto host_x = vec::create(exec->get_master(), gko::dim<2>(size, 1));
+    for (auto i = 0; i < size; i++) {
```
upsj, line 82:
```suggestion
    for (gko::size_type i = 0; i < size; i++) {
```
0 is `int` by default, so probably not 100% what you want here.

examples/ir-ilu-preconditioned-solver/ir-ilu-preconditioned-solver.cpp
```diff
@@ -76,8 +76,16 @@ int main(int argc, char *argv[])
 
     // Read data
     auto A = gko::share(gko::read<mtx>(std::ifstream("data/A.mtx"), exec));
-    auto b = gko::read<vec>(std::ifstream("data/b.mtx"), exec);
-    auto x = gko::read<vec>(std::ifstream("data/x0.mtx"), exec);
+    // Create RHS and initial guess as 1
+    gko::size_type size = A->get_size()[0];
```
thoasm, line 80: Nit: Maybe call it `num_rows` instead of `size` to be more specific.

examples/ir-ilu-preconditioned-solver/ir-ilu-preconditioned-solver.cpp
```diff
@@ -76,8 +76,16 @@ int main(int argc, char *argv[])
...
+    auto host_x = vec::create(exec->get_master(), gko::dim<2>(size, 1));
+    for (auto i = 0; i < size; i++) {
+        host_x->at(i, 0) = 1.;
+    }
+    auto x = vec::create(exec);
+    auto b = vec::create(exec);
+    x->copy_from(host_x.get());
```
yhmtsai, line 87: use x = zeros before but ones now. Is it correct?
if it changes the result, you also need to change the result.dox.

fritzgoebel: Ones is the choice we took for the EuroPar paper, so it is correct now. I need to change the result.dox, thanks!

PR #551, opened 2020-05-28
Pass std::shared_ptr<Executor> by reference where sensible

upsj:
This PR avoids unnecessary `std::shared_ptr` copies by instead passing them as const references. Additionally, it makes sure that all kernel implementations are stored in `_kernel(s)` files.

Simple search/replace job, so there is no harm in discarding this PR if we consider it a bad idea ;)

Related reading: https://herbsutter.com/2013/06/05/gotw-91-solution-smart-pointer-parameters/


upsj:
I'll close this, since it doesn't have any effect on the implementation and is a pretty intrusive change without much pay-off (performance-wise). We have enough open PRs already :)

Issue #412, opened 2019-12-04
Some header files are not self-contained


upsj:
I came across some header files that only work correctly when another header file is included before. An example is `math.hpp`, which uses macros defined in `config.hpp` but doesn't include it.
It would probably make sense to go over the symbols and macros we define in different headers and make sure they are consistently including the headers defining them. 
Is this something `iwyu` can catch? Otherwise we might violate the ODR rule or introduce hard-to-catch compilation issues.

This is somewhat related to the AMD build issues in PR #410

upsj:
Other examples (although that is not our fault):

See https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/373718543

```
In file included from /opt/rocm/include/hip/device_functions.h:29:
In file included from /opt/rocm/include/hip/hcc_detail/device_functions.h:27:
/opt/rocm/include/hip/hcc_detail/math_fwd.h:36:5: error: no template named 'HIP_vector_base'
    HIP_vector_base<short, 2>::Native_vec_,
```

`device_functions.h` requires `hip_runtime.h` to be included such that the `HIP_vector_base` template is declared

upsj:
Revisiting this, there is actually a simple way to test if a header file. is self-contained: Try to compile it as if it were a source file. No idea why I didn't think of this earlier. We could let CMake take care of this in a separate CI step like the test for cyclical dependencies.

PR #552, opened 2020-05-28
Fix header self-sufficiency and add compile-time check

upsj:
This works by building a separate source file for each header that only includes this file. The PR also contains fixes for all found issues.

Fixes https://github.com/ginkgo-project/ginkgo/issues/412

### DIFF
--- a/include/CMakeLists.txt
+++ b/include/CMakeLists.txt
@@ -1,0 +1,6 @@
+if (GINKGO_CHECK_CIRCULAR_DEPS)
+    add_library(ginkgo_public_api INTERFACE) # dummy target
+    set_property(TARGET ginkgo_public_api APPEND PROPERTY INTERFACE_INCLUDE_DIRECTORIES "${CMAKE_CURRENT_SOURCE_DIR}")
+    set_property(TARGET ginkgo_public_api APPEND PROPERTY INTERFACE_INCLUDE_DIRECTORIES "${CMAKE_CURRENT_BINARY_DIR}")
+    ginkgo_check_headers(ginkgo_public_api)
+endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -179,6 +179,7 @@ if(GINKGO_BUILD_CUDA)
     add_subdirectory(cuda)       # High-performance kernels for NVIDIA GPUs
 endif()
 add_subdirectory(core)           # Core Ginkgo types and top-level functions
+add_subdirectory(include)        # Public API self-contained check
 if (GINKGO_BUILD_REFERENCE)
     add_subdirectory(reference)  # Reference kernel implementations
 endif()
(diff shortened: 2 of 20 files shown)


upsj:
The reason I didn't go for a GLOB is the following: I regularly develop on the build system where the files are being copied over from my editor via `scp`, but if I delete/rename a file, it's not automatically deleted -- GLOB would still find these files. But of course that's not a strong argument, since I rarely to never build with CIRCULAR_DEPS on. I am not sure how likely something like this would be in practice. Git usually takes care of this when changing branches, only maybe if somebody uses Ginkgo without git might we run into issues.

upsj:
I don't have too much of a strong opinion on GLOB vs. explicit list. I also think that the missing header files in IDEs were mostly an issue with older versions of CMake and/or Visual Studio, so that shouldn't matter nowadays.

I would however argue that iwyu and this approach achieve different (albeit related) things:
The declared goal of iwyu is to "thin out" the include dependency DAG of a project. It should also notice when a file is missing an include, but I can't really see these warnings stand out in any way among the huge number of warnings we currently get. Also in my experience (which is somewhat dated though), iwyu tends to produce a large number of false-positives.
This compilation step on the other hand is supposed to explicitly fail when a file is not self-contained, i.e. including it without some missing dependencies causes a build to fail. This step should not produce any false-positives as far as I can tell.

upsj:
So after a quick look over a few of the iwyu suggestions, it looks to me like some of them indeed have merit, especially related to standard library headers (missing <numeric> that was included somewhere else, ...), but there were also a few false positives (like proposing to include internal standard library headers) and a few false negatives (like wanting to remove a header declaring kernels, probably due to the macro usage).

So to me it looks like iwyu would be able to detect some of these issues, but we can't really use it as a general-purpose tool due to the large number of false suggestions.

tcojean:
I just found out that `iwyu` has quite a few pragmas which we could use to control its output significantly. I believe the issues we cannot fix with these are the false negatives. I admit though that getting the output to a usable state would take quite a bit of work since there is so much to fix, on top of having to tackle the bad suggestions it proposes.
https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md

upsj:
I moved from explicitly listing the headers to GLOBs now ;)
Regarding IWYU, while it might be possible, I believe this is too much work for too little pay-off in terms of build time.

include/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+if (GINKGO_CHECK_CIRCULAR_DEPS)
+    add_library(ginkgo_public_api INTERFACE) # dummy target
+    set_property(TARGET ginkgo_public_api APPEND PROPERTY INTERFACE_INCLUDE_DIRECTORIES "${CMAKE_CURRENT_SOURCE_DIR}")
+    set_property(TARGET ginkgo_public_api APPEND PROPERTY INTERFACE_INCLUDE_DIRECTORIES "${CMAKE_CURRENT_BINARY_DIR}")
+    ginkgo_check_headers(ginkgo_public_api)
```
tcojean, line 5: This is just a comment, but I think this target can be useful at some point since that would allow for example to use it to create the main `ginkgo.hpp` file, or to install our headers in a specific location. It's a good thing to have that available now.

upsj:
This is already part of the [check-circular-deps pipeline](https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/589260668). I will add comments to the INSTALL.md and CONTRIBUTING.md

PR #554, opened 2020-06-04
Add a MachineTopology class and HWLOC support

pratikvn:
This PR adds a machine topology class to the executor. This is done using the [HWLOC library](https://www.open-mpi.org/projects/hwloc/). This enables us to have finer control of the cores and GPUs and helps us bind executors to cores and such. 

Currently, HWLOC allows the system administrator to export the machine configuration into an XML file which we can export and get information about the system. In case, this XML file is not available, then we have to query and generate the topology with the HWLOC functions.

The core of the machine info class was written by @tcojean.  

TODO
+ [x] Add documentation
+ [x] Evaluate possibility of a global struct.

Edit: The exec_info struct now looks like this: 

| CPU | CUDA/HIP | DPCPP | Our Abstraction (Current) |
| :------------- | :----------: | :------------------: | -------------------: |
|  `num_cores` | `num_multiprocessors` (num_sms)   |  `num_compute_units`    | `num_compute_units`    |
| `num_threads_per_core`  | `num_warps_per_sm`  | -- | `num_pu_per_cu` |
| -- | -- | `subgroup_sizes` (array) | `subgroup_sizes` (array) |
| -- | `warp_size` (scalar)  | `max_sub_group_size` (scalar) | `max_subgroup_size` (scalar) |
| --  | `max_threads_per_block`  | `max_work_group_size` | `max_workgroup_size` |
| --  | `max_threads_per_block_dimension` (x,y,z) (array) | `max_work_item_sizes` (array) | `max_workitem_sizes` (array) |


pratikvn:
check_format!

pratikvn:
check-format!

pratikvn:
format!

CMakeLists.txt
```diff
@@ -194,6 +208,11 @@ include(cmake/package_helpers.cmake)
 ginkgo_find_package(GTest "GTest::GTest;GTest::Main" FALSE 1.8.1)
 ginkgo_find_package(gflags gflags FALSE 2.2.2)
 ginkgo_find_package(RapidJSON rapidjson TRUE 1.1.0)
+if(NOT GINKGO_FORCE_NO_HWLOC AND GINKGO_HAVE_HWLOC)
+  ginkgo_find_package(HWLOC hwloc FALSE 2.1)
+endif()
+set(GKO_HWLOC_XMLFILE "${HWLOC_XMLFILE}")
+message(STATUS "HWLOC XML file set to: ${GKO_HWLOC_XMLFILE}")
```
tcojean, line 215: Shouldn't it be `$ENV{HWLOC_XMLFILE}`? Also, isn't this unneeded since it's standard hwloc anyway?

cmake/get_info.cmake
```diff
@@ -127,7 +127,7 @@ foreach(log_type ${log_types})
     ginkgo_print_module_footer(${${log_type}} "User configuration:")
     ginkgo_print_module_footer(${${log_type}} "  Enabled modules:")
     ginkgo_print_foreach_variable(
-        "GINKGO_BUILD_OMP;GINKGO_BUILD_REFERENCE;GINKGO_BUILD_CUDA;GINKGO_BUILD_HIP;GINKGO_BUILD_DPCPP")
+        "GINKGO_BUILD_OMP;GINKGO_BUILD_REFERENCE;GINKGO_BUILD_CUDA;GINKGO_BUILD_HIP;GINKGO_BUILD_DPCPP;GINKGO_WITH_HWLOC")
```
tcojean, line 130: It's not a module though

cmake/package_helpers.cmake
```diff
@@ -173,7 +200,7 @@ macro(ginkgo_find_package package_name target_list header_only)
                     "${lib}" ${header_only})
              endforeach()
         else()
-            find_package(${package_name} QUIET ${ARGN})
+            find_package(${package_name} ${ARGN})
```
tcojean, line 203: Why not quiet?

core/devices/cuda/CMakeLists.txt
```diff
@@ -1,3 +1,3 @@
 ginkgo_add_object_library(ginkgo_cuda_device
-    executor.cpp)
+  executor.cpp)
```
tcojean, line 2: Keep 4 please

core/devices/dpcpp/CMakeLists.txt
```diff
@@ -1,2 +1,2 @@
 ginkgo_add_object_library(ginkgo_dpcpp_device
-    executor.cpp)
+  executor.cpp)
```
tcojean, line 2: same

core/devices/hip/CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
 ginkgo_add_object_library(ginkgo_hip_device
-    executor.cpp)
+  executor.cpp)
+
```
tcojean, line 3: same

core/devices/omp/CMakeLists.txt
```diff
@@ -1,2 +1,2 @@
 ginkgo_add_object_library(ginkgo_omp_device
-    executor.cpp)
+  executor.cpp)
```
tcojean, line 2: same

core/devices/machine_topology.cpp
```diff
@@ -0,0 +1,65 @@
...
+}  // namespace detail
+
+
+const MachineTopology *get_machine_topology()
+{
+    return detail::machine_topology.get();
+}
+
+
+std::shared_ptr<const MachineTopology> get_shared_machine_topology()
```
tcojean, line 59:
```suggestion
std::shared_ptr<const MachineTopology>& get_shared_machine_topology()
```

core/devices/reference/CMakeLists.txt
```diff
@@ -1,2 +1,2 @@
 ginkgo_add_object_library(ginkgo_reference_device
-    dummy.cpp)
+  dummy.cpp)
```
tcojean, line 2: same

dpcpp/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ endif()
 ginkgo_extract_dpcpp_version(${CMAKE_CXX_COMPILER} GINKGO_DPCPP_VERSION)
 set(GINKGO_DPCPP_VERSION ${GINKGO_DPCPP_VERSION} PARENT_SCOPE)
 
-add_library(ginkgo_dpcpp $<TARGET_OBJECTS:ginkgo_dpcpp_device> "")
+add_library(ginkgo_dpcpp $<TARGET_OBJECTS:ginkgo_core_device> $<TARGET_OBJECTS:ginkgo_dpcpp_device> "")
```
tcojean, line 9: Why do you need the object for `ginkgo_core_device` everywhere now?

hip/CMakeLists.txt
```diff
@@ -287,7 +287,8 @@ if(GINKGO_HIP_PLATFORM MATCHES "hcc")
         # directly use HIP
         ginkgo_hip_clang_ban_hip_device_flags()
     endif()
-    target_link_libraries(ginkgo_hip PRIVATE hip::device)
+    target_link_libraries(ginkgo_hip PRIVATE hip::device
+      )
```
tcojean, line 291:
```suggestion
    target_link_libraries(ginkgo_hip PRIVATE hip::device)
```

include/ginkgo/core/base/machine_topology.hpp
```diff
@@ -0,0 +1,323 @@
...
+
+
+namespace gko {
+
+
+class MachineTopology;
+
+extern const MachineTopology *get_machine_topology();
+
+extern std::shared_ptr<const MachineTopology> get_shared_machine_topology();
```
tcojean, line 74:
```suggestion
extern std::shared_ptr<const MachineTopology>& get_shared_machine_topology();
```

include/ginkgo/core/base/machine_topology.hpp
```diff
@@ -0,0 +1,323 @@
...
+    };
+
+public:
+    /**
+     * Creates a new MachineTopology object.
+     */
+    static std::shared_ptr<MachineTopology> create()
+    {
+        return std::shared_ptr<MachineTopology>(new MachineTopology());
+    }
```
tcojean, line 127:
```suggestion
protected:
    /**
     * Creates a new MachineTopology object.
     */
    static std::shared_ptr<MachineTopology> create()
    {
        return std::shared_ptr<MachineTopology>(new MachineTopology());
    }
    
public:
```

include/ginkgo/core/base/executor.hpp
```diff
@@ -1108,14 +1165,17 @@ class CudaExecutor : public detail::ExecutorBase<CudaExecutor>,
         return num_execs[device_id];
     }
 
+    void populate_exec_info(
+        std::shared_ptr<const MachineTopology> mach_topo) override;
```
tcojean, line 1169:
```suggestion
        std::shared_ptr<const MachineTopology>& mach_topo) override;
```
Or only use the pointer.

include/ginkgo/core/base/executor.hpp
```diff
@@ -1274,14 +1350,17 @@ class HipExecutor : public detail::ExecutorBase<HipExecutor>,
...
```
tcojean, line 1354:
```suggestion
        std::shared_ptr<const MachineTopology>& mach_topo) override;
```

include/ginkgo/core/base/executor.hpp
```diff
@@ -1350,57 +1429,63 @@ class DpcppExecutor : public detail::ExecutorBase<DpcppExecutor>,
...
-        : device_id_(device_id), master_(master), device_type_(device_type)
+        : master_(master)
     {
+        this->dpcpp_exec_info_.device_id = device_id;
+        this->dpcpp_exec_info_.device_type = device_type;
         this->set_device_property();
     }
 
+    void populate_exec_info(
+        std::shared_ptr<const MachineTopology> mach_topo) override;
```
tcojean, line 1472:
```suggestion
        std::shared_ptr<const MachineTopology>& mach_topo) override;
```

include/ginkgo/core/base/executor.hpp
```diff
@@ -959,6 +989,16 @@ class ReferenceExecutor : public OmpExecutor {
 
 protected:
     ReferenceExecutor() = default;
+
+    void populate_exec_info(std::shared_ptr<const MachineTopology>) override
```
tcojean, line 993:
```suggestion
    void populate_exec_info(std::shared_ptr<const MachineTopology>&) override
```

include/ginkgo/core/base/executor.hpp
```diff
@@ -684,6 +685,22 @@ class Executor : public log::EnableLogging<Executor> {
...
 #undef GKO_ENABLE_RAW_COPY_TO
 
+    /**
+     * Populates the executor specific info from the global machine topology
+     * object.
+     *
+     * @param mach_topo the machine topology object.
+     */
+    virtual void populate_exec_info(
+        std::shared_ptr<const MachineTopology> mach_topo) = 0;
```
tcojean, line 695:
```suggestion
        std::shared_ptr<const MachineTopology>& mach_topo) = 0;
```

include/ginkgo/core/base/executor.hpp
```diff
@@ -917,14 +934,27 @@ class OmpExecutor : public detail::ExecutorBase<OmpExecutor>,
...
+    int get_num_threads_per_core() const
+    {
+        return this->omp_exec_info_.num_work_groups_per_core;
+    }
+
 protected:
     OmpExecutor() = default;
 
+    void populate_exec_info(
+        std::shared_ptr<const MachineTopology> mach_topo) override;
```
tcojean, line 948:
```suggestion
        std::shared_ptr<const MachineTopology>& mach_topo) override;
```

core/device_hooks/cuda_hooks.cpp
```diff
@@ -62,6 +62,11 @@ std::shared_ptr<CudaExecutor> CudaExecutor::create(
 }
 
 
+void CudaExecutor::populate_exec_info(
+    std::shared_ptr<const MachineTopology> mach_topo)
```
tcojean, line 66:
```suggestion
    std::shared_ptr<const MachineTopology>& mach_topo)
```

core/device_hooks/dpcpp_hooks.cpp
```diff
@@ -59,6 +59,11 @@ std::shared_ptr<DpcppExecutor> DpcppExecutor::create(
 }
 
 
+void DpcppExecutor::populate_exec_info(
+    std::shared_ptr<const MachineTopology> mach_topo)
```
tcojean, line 63:
```suggestion
    std::shared_ptr<const MachineTopology>& mach_topo)
```

core/device_hooks/hip_hooks.cpp
```diff
@@ -59,6 +59,11 @@ std::shared_ptr<HipExecutor> HipExecutor::create(
 }
 
 
+void HipExecutor::populate_exec_info(
+    std::shared_ptr<const MachineTopology> mach_topo)
```
tcojean, line 63:
```suggestion
    std::shared_ptr<const MachineTopology>& mach_topo)
```

core/devices/omp/executor.cpp
```diff
@@ -44,6 +44,11 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 namespace gko {
 
 
+void OmpExecutor::populate_exec_info(
+    std::shared_ptr<const MachineTopology> mach_topo)
```
tcojean, line 48:
```suggestion
    std::shared_ptr<const MachineTopology>& mach_topo)
```

cuda/base/executor.cpp
```diff
@@ -71,6 +111,11 @@ std::shared_ptr<CudaExecutor> CudaExecutor::create(
 }
 
 
+void CudaExecutor::populate_exec_info(
+    std::shared_ptr<const MachineTopology> mach_topo)
```
tcojean, line 115:
```suggestion
    std::shared_ptr<const MachineTopology>& mach_topo)
```

dpcpp/base/executor.dp.cpp
```diff
@@ -84,6 +84,11 @@ std::shared_ptr<DpcppExecutor> DpcppExecutor::create(
 }
 
 
+void DpcppExecutor::populate_exec_info(
+    std::shared_ptr<const MachineTopology> mach_topo)
```
tcojean, line 88:
```suggestion
    std::shared_ptr<const MachineTopology>& mach_topo)
```

hip/base/executor.hip.cpp
```diff
@@ -71,6 +116,11 @@ std::shared_ptr<HipExecutor> HipExecutor::create(
 }
 
 
+void HipExecutor::populate_exec_info(
+    std::shared_ptr<const MachineTopology> mach_topo)
```
tcojean, line 120:
```suggestion
    std::shared_ptr<const MachineTopology>& mach_topo)
```

tcojean (commented):
Some small issues while going quickly through the code, and something to try whether it fixes the CUDA test issues?

cmake/autodetect_executors.cmake
```diff
@@ -2,6 +2,7 @@ set(GINKGO_HAS_OMP OFF)
 set(GINKGO_HAS_CUDA OFF)
 set(GINKGO_HAS_DPCPP OFF)
 set(GINKGO_HAS_HIP OFF)
+set(GINKGO_HAS_HWLOC OFF)
```
tcojean, line 5: It's technically not an executor though

core/devices/machine_topology.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+
+namespace gko {
+
+
+namespace detail {
+
+
+std::shared_ptr<const MachineTopology> machine_topology =
+    MachineTopology::create();
```
tcojean, line 48: For reference, another thing to try
```suggestion
volatile std::shared_ptr<const MachineTopology> machine_topology =
    MachineTopology::create();
```

pratikvn:
format!

include/ginkgo/core/base/machine_topology.hpp
```diff
@@ -0,0 +1,450 @@
...
+    MachineTopology();
+
+    /**
+     * @internal
+     *
+     * A helper function that binds the object with an id.
+     */
+    void hwloc_binding_helper(
+        const std::tuple<std::vector<MachineTopology::normal_obj_info>,
+                         MachineTopology::hwloc_manager<hwloc_bitmap_s>> &obj,
```
tcojean, line 364: Again, store this type somewhere.

include/ginkgo/core/base/machine_topology.hpp
```diff
@@ -0,0 +1,450 @@
...
+    hwloc_bitmap_s *init_bitmap();
+
+    /**
+     *
+     * @internal
+     *
+     * Get local id from the os index
+     */
+    inline int get_obj_local_id_by_os_index(
+        std::vector<normal_obj_info> &objects, size_type os_index) const;
```
tcojean, line 419: Rather than by os_index, we maybe want the logical index one instead?

core/devices/machine_topology.cpp
```diff
@@ -0,0 +1,349 @@
...
+        GKO_ASSERT(id[i] < std::get<0>(obj).size());
+        GKO_ASSERT(id[i] >= 0);
+        hwloc_bitmap_set(bitmap_toset, std::get<0>(obj)[id[i]].os_id);
+        // Check if this id has been bound before, Binding multiple threads to
+        // same object may not be suitable.
+        if (hwloc_bitmap_isset(std::get<1>(obj).get(),
+                               std::get<0>(obj)[id[i]].os_id)) {
+            hwloc_bitmap_clr(bitmap_toset, std::get<0>(obj)[id[i]].os_id);
+        }
+    }
```
tcojean, line 134: I think you are overdoing it with this here. I would remove the `bitmap` thing altogether. Although it's a good idea, this should probably be done one level up and not here. When asking to bind to a core, I would expect it to actually bind to this core. If it's overcrowded, too bad for me.

In general, this way of doing it has flaws as well:
1. You do this per type, but a CUDA host thread could be bound to the same core as a CPU processing executor.
2. In MPI mode, which is the main mode where these aspects need to be taken into account, this will not work anyway.
3. Again in MPI, being able to bind e.g., a CUDA GPU to multiple cores would allow to take care of the oversubscription issue without overcomplicating things: the OS will migrate the threads so that they don't step on each other.
4. I think this would fit better into the automatic `GenericDeviceExecutor` or other executor than here.

core/devices/machine_topology.cpp
```diff
@@ -0,0 +1,349 @@
...
+            io_obj_info{obj, obj->logical_index, obj->os_index, obj->gp_index,
+                        hwloc_bitmap_first(ancestor->nodeset), ancestor});
+        // Get the corresponding cpuset of the ancestor nodeset
+        hwloc_cpuset_t ancestor_cpuset = hwloc_bitmap_alloc();
+        hwloc_cpuset_from_nodeset(this->topo_.get(), ancestor_cpuset,
+                                  ancestor->nodeset);
+        // Find the cpu object closest to this device from the ancestor cpuset
+        // and store its id for binding purposes
+        vector.back().closest_cpu_id = get_obj_local_id_by_os_index(
+            std::get<0>(this->pus_), hwloc_bitmap_first(ancestor_cpuset));
```
tcojean, line 261: Here I think it would be better to simply store all the IDs as explained in other places.

core/devices/machine_topology.cpp
```diff
@@ -0,0 +1,349 @@
...
+        // Get type of the ancestor object and store it as a string.
+        char ances_type[24];
+        hwloc_obj_type_snprintf(ances_type, sizeof(ances_type), ancestor, 0);
+        vector.back().ancestor_type = std::string(ances_type);
+        // Write the PCI Bus ID from the object info.
+        char pci_bus_id[13];
+        snprintf(pci_bus_id, sizeof(pci_bus_id), "%04x:%02x:%02x.%01x",
+                 obj->attr->pcidev.domain, obj->attr->pcidev.bus,
+                 obj->attr->pcidev.dev, obj->attr->pcidev.func);
+        vector.back().pci_bus_id = std::string(pci_bus_id);
```
tcojean, line 284: You are adding a lot of information to this object. Aren't most of them unneeded, except for this portion? What comes after this code definitely looks like we do not need it for now.

I would rather we make this function more slim for now and only use what we need, and later if we need extra information for whichever reason we can always add them.

tcojean (changes requested):
LGTM in general, but I vouch for a simplification of some parts and a generalization of others. In addition, I would like to discuss some things. There are in particular two things which look fairly weird to me, putting hwloc to the Ginkgo RPATH and the `ifdef _ICC` thing for the CUDA tests.

Another minor comment: the CMake files have random formatting (sometimes 2 spaces, sometimes 4, ...).

cmake/Modules/FindHWLOC.cmake
```diff
@@ -0,0 +1,147 @@
+#
```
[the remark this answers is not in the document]
pratikvn: Is there any particular reason you prefer this over the current one ?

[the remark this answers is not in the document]
tcojean: Licensing. The license in this one is close to our own and more permissive than the LGPL. Since this is a core aspect of our library, we would pretty much always have it as part of our package, which could create future trouble.

I also know personally where this is from and used it myself so I can attest that it is solid. But it of course depends if it does all we need.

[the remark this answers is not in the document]
pratikvn: I use this now but modify it for our purposes.

core/test/base/executor.cpp
```diff
@@ -529,20 +594,61 @@ TEST(Executor, CanVerifyMemory)
...
+}
+
+
+TEST(Executor, CanVerifyMemoryBetweenTwoHipDevices)
+{
+    auto omp = gko::OmpExecutor::create();
+    auto hip = gko::HipExecutor::create(0, omp);
+    if (hip->get_num_devices() < 2) {
+        GTEST_SKIP() << "Only one hip device available";
+    }
```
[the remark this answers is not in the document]
pratikvn: The thing is you need to be able to get the PCI bus id from the device id. Currently this is done in the constructor which calls the `populate_exec_info` which calls the cuda api function. With an invalid device id this is no longer possible.

I understand that it changes the behaviour, but I would consider creating a CUDA/HIP executor with an invalid device id to be incorrect anyway. Is there a use case where this behaviour is needed ? If not I would prefer failing for an invalid id. 

If we really want to keep the same behaviour, then I think the user will have to explicitly call the `populate_exec_info` which I dont really like.

cmake/create_test.cmake
```diff
@@ -183,6 +189,7 @@ function(ginkgo_create_hip_test test_name)
         endif()
     endif()
 
-    target_link_libraries(${TEST_TARGET_NAME} PRIVATE ginkgo GTest::Main GTest::GTest ${ARGN})
+    target_link_libraries(${TEST_TARGET_NAME} PRIVATE ginkgo GTest::Main GTest::GTest
+      GMock::Main GMock::GMock ${ARGN})
```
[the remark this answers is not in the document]
pratikvn: I wanted to use the array feature of the `EXPECT_THAT`, but I guess indeed it is not necessary and we can do without it.

cuda/test/utils.hpp
```diff
@@ -42,11 +42,14 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
 namespace {
 
+#if defined(__ICC) || defined(__ICL)
+#else
 
 // prevent device reset after each test
 auto no_reset_exec =
     gko::CudaExecutor::create(0, gko::ReferenceExecutor::create(), true);
 
+#endif
```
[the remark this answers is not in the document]
pratikvn: Intel compilers when compiling for static libraries have issues figuring out the order these global objects should be freed. This conflicts with the freeing of the MachineTopology object.

include/ginkgo/core/base/machine_topology.hpp
pratikvn: I think binding is always done with the OS index and hence I need the `os_index` [HWLOC reference](https://www.open-mpi.org/projects/hwloc/doc/v2.4.0/a00175.php#ga062dbff93baeff3b425a7260c5463646)

include/ginkgo/core/base/executor.hpp
```diff
@@ -724,6 +725,20 @@ class Executor : public log::EnableLogging<Executor> {
...
+     * object.
+     *
+     * @param mach_topo the machine topology object.
+     */
+    virtual void populate_exec_info(const MachineTopology *mach_topo) = 0;
+
+    struct exec_info {
+        int device_id = -1;
+        int num_cores = 0;
+        int num_work_groups_per_core = 0;
```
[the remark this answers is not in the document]
pratikvn: Does `num_pu_per_core` make sense for GPU ? Isn't the `work_group` kind of equivalent to a `pu` ? I think `pu` does not really make sense for a GPU

core/devices/machine_topology.cpp
pratikvn: Sorry I didnt get this part 

> Again in MPI, being able to bind e.g., a CUDA GPU to multiple cores would allow to take care of the oversubscription issue without overcomplicating things: the OS will migrate the threads so that they don't step on each other.

This way of doing it allows you to bind to multiple cores, right ? Or did you mean something else ? 

But otherwise, I do get your point that this is bit over the top

core/devices/machine_topology.cpp
pratikvn: Initially, I planned to use the names of the io children objects to get the type of devices. I think that still makes sense for detecting network cards or Infiniband cards, as I am not sure how to detect those with just a PCI id. But I guess as currently we dont have any network based support, we can add it later when necessary.

include/ginkgo/core/base/executor.hpp
```diff
@@ -1062,6 +1111,8 @@ class CudaExecutor : public detail::ExecutorBase<CudaExecutor>,
     friend class detail::ExecutorBase<CudaExecutor>;
 
 public:
+    using cuda_exec_info = exec_info;
```
tcojean, line 1114: Why not use `exec_info` directly? That seems like much simpler, and shorter as well.

include/ginkgo/core/base/executor.hpp
```diff
@@ -1238,6 +1340,8 @@ class HipExecutor : public detail::ExecutorBase<HipExecutor>,
     friend class detail::ExecutorBase<HipExecutor>;
 
 public:
+    using hip_exec_info = exec_info;
```
tcojean, line 1343: Same here.

include/ginkgo/core/base/executor.hpp
```diff
@@ -1040,6 +1086,9 @@ class ReferenceExecutor : public OmpExecutor {
     GKO_DEFAULT_OVERRIDE_VERIFY_MEMORY(CudaExecutor, false);
 
     GKO_DEFAULT_OVERRIDE_VERIFY_MEMORY(HipExecutor, false);
+
+private:
+    exec_info ref_exec_info_;
```
tcojean, line 1091: Now that it's implemented using the same structure for everyone, you could make this a member of the Executor base class and add a public method which returns a reference/pointer to a const version of that structure, and a protected non const getter which you could use to set the parameters.

include/ginkgo/core/base/executor.hpp
```diff
@@ -1143,26 +1213,61 @@ class CudaExecutor : public detail::ExecutorBase<CudaExecutor>,
...
+     */
+    void bind_to_cpus(const int *ids, const size_type num_ids) const;
+
+    /**
+     * Bind the current executor thread to the list of cores
+     *
+     * @param  ids  the logical ids of the cores.
+     * @param  num_ids  the number of cores to bind
+     */
+    void bind_to_cores(const int *ids, const size_type num_ids) const;
```
thoasm, line 1247: This seems inconsistent: You return `std::vector<int>` at `get_closest_cpus()`, but here, you want to get pointers and the size separately (so you can't do `bind_to_cores(get_closest_cpus())`)?
Please use `std::vector` here as well (if you want support for arrays, add another function taking `std::array<int, N>` as an argument).

include/ginkgo/core/base/executor.hpp
```diff
@@ -1319,26 +1442,60 @@ class HipExecutor : public detail::ExecutorBase<HipExecutor>,
...
```
thoasm, line 1476: Same here (please use `std::vector` and maybe `std::array` as the argument).

include/ginkgo/core/base/executor.hpp
```diff
@@ -1143,26 +1213,61 @@ class CudaExecutor : public detail::ExecutorBase<CudaExecutor>,
...
-          warp_size_(0)
+        : EnableDeviceReset{device_reset}, master_(master)
     {
-        assert(device_id < max_devices && device_id >= 0);
+        this->cuda_exec_info_.device_id = device_id;
+        this->cuda_exec_info_.num_cores = 0;
+        this->cuda_exec_info_.num_work_groups_per_core = 0;
+        assert(this->cuda_exec_info_.device_id < max_devices &&
+               this->cuda_exec_info_.device_id >= 0);
+        this->CudaExecutor::populate_exec_info(get_machine_topology());
```
tcojean, line 1263: I would move this assert to the `populate_exec_info` overload and have this information be a no-op, or do almost nothing (maybe only set the device id) when the module itself is not compiled. This way, you get the executor instantiation failure *only* when you actually compile the module, which makes this more manageable.

This is how I somewhat implemented DPC++ actually.

include/ginkgo/core/base/machine_topology.hpp
```diff
@@ -0,0 +1,420 @@
...
+     * Initialize the topology object.
+     */
+    hwloc_topology *init_topology();
+
+    /**
+     * @internal
+     *
+     * Initialize the bitmap.
+     */
+    hwloc_bitmap_s *init_bitmap();
```
thoasm, line 384: These should not return plain (unmanaged) pointers since it would be prone to memory leaks, hard to use correctly / maintain and not thread safe.

include/ginkgo/core/base/machine_topology.hpp
```diff
@@ -0,0 +1,420 @@
...
+    {
+        hwloc_binding_helper(this->cores_, ids, num_ids);
+    }
+
+    /**
+     * Bind the object associated with the id to a Processing unit(PU).
+     *
+     * @param id  The id of the object to be bound.
+     */
+    void bind_to_pus(const int *ids, const size_type num_ids) const
```
thoasm, line 250: See previous comments about using `std::vector` and maybe `std::array` instead of pointer-size combination.

tcojean (approved):
LGTM. I have a few extra comments.

On the topic of making Executor initialization fail, I think it's better to have the executor creation fail only when the CUDA/HIP/DPC++ modules are actually compiled.

Otherwise, (maybe despite that even), the following changes are required I think:
+ executor.hpp, lines 151-160 need to change, since it would now pretty much always fail (you can't compile/have both DPC++ and CUDA together). I don't know to what we could change this honestly.
+ executor.hpp, lines 314-330, the full example main function needs to change
+ The DPC++ executor needs to be unified and fail at the same points as the CUDA/HIP executors now do
+ A `@note` should be added on when/how executors throw during instantiation in the main documentation
+ In the `simple_solver` example, or some other major one like it, we should detail that using a map such as this is one of the only way to have the code always work, since otherwise it would throw.

thoasm (dismissed):
Overall well done, but there are some inconsistencies in the interface I would like to have addressed before I approve it.
The main issue I have currently:
- Use of unmanaged plain pointers instead of wrappers with destructors
- Inconsistent `bind_to_cores` vs. `get_closest_cpus`

Suggestions:
- Change `MachineTopology` to a Singleton (and therefore move `get_machine_topology` to a `public static` member function)
- Use of `get_device_id()` (even internally)
- `GINKGO_FORCE_NO_HWLOC` -> `GINKGO_DISABLE_HWLOC` (I tried to mark every occurance)

tcojean:
@thoasm wouldn't a singleton anyway have the same implementation except that there would be no fast-path atomic-boolean once the class has been initialized, and therefore have sub-par performance?

thoasm:
@tcojean Yes, it can have pretty much the same implementation. The reason why I would prefer a Singleton is that the `get_machine_topology()` would be inside the class.
The `static MachineTopology instance;` line leaves it to the compiler to do a proper implementation, and at least in the Microbenchmark I did, it performed almost identical to @pratikvn's current implementation (I only tested single-threaded). It is only a suggestion from me (it is shorter an less error prone), but I am also totally fine if @pratikvn keeps his implementation.
Here is another take on it, including benchmarks: https://www.modernescpp.com/index.php/thread-safe-initialization-of-a-singleton
There, the compiler generated solution is always the fastest.

include/ginkgo/core/base/executor.hpp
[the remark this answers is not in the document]
tcojean: A workgroup is an expression of a unit of computation with internal parallelism which a single computation unit (i.e. core or SM) is able to execute as a standalone. In OpenCL, a computation unit pulls a workgroup (a bunch of workitems/logical threads) which it will execute.
In OpenCL, subgroup would be closer to PU.

Then, how about something like `subpu_per_core` or something like that. I think we agree that core represents both a CPU core and a SM fairly well, then what we need is a generic enough term to represent a sub processing unit of a core/SM, hence simply subPU, but maybe something else has a better idea.

test_install/CMakeLists.txt
```diff
@@ -24,6 +24,11 @@ add_executable(test_install test_install.cpp)
 target_compile_features(test_install PUBLIC cxx_std_14)
 target_link_libraries(test_install PRIVATE Ginkgo::ginkgo)
 
+if(GINKGO_HAVE_HWLOC)
+    find_package(HWLOC REQUIRED)
+    target_link_libraries(test_install PRIVATE hwloc)
+endif()
```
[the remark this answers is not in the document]
pratikvn: Yes, it should have been in `GinkgoConfig.cmake.in`. I will fix that.

include/ginkgo/core/base/executor.hpp
pratikvn: I wanted some leeway, in case we decide to add some executor specific info, we can create a new struct `cuda_exec_info` and still not break the interface.

core/devices/cuda/executor.cpp
```diff
@@ -50,20 +50,34 @@ std::shared_ptr<const Executor> CudaExecutor::get_master() const noexcept
...
 #else
     return false;
 #endif
 }
 
 
+void CudaExecutor::bind_to_cpus(const int *cpu_ids,
+                                const size_type num_ids) const
+{
+    get_machine_topology()->bind_to_pus(cpu_ids, num_ids);
```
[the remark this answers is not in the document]
pratikvn: I changed everything to `bind_to_pus` now.

include/ginkgo/core/base/executor.hpp
pratikvn: Good point. I have done this now.

include/ginkgo/core/base/executor.hpp
pratikvn: It turned out that I dont actually need this assert because there is an if clause that checks the same thing.

.gitlab-ci.yml
```diff
@@ -22,6 +22,7 @@ stages:
   BUILD_OMP: "OFF"
   BUILD_CUDA: "OFF"
   BUILD_HIP: "OFF"
+  FORCE_NO_HWLOC: "OFF"
```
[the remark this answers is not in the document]
pratikvn: Thanks. That is definitely better and I use it now.

core/devices/machine_topology.cpp
```diff
@@ -0,0 +1,264 @@
...
+    }
+#endif
+}
+
+
+hwloc_topology *MachineTopology::init_topology()
```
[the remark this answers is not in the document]
pratikvn: This is used in the creation of a unique ptr, so i think this is fine.

include/ginkgo/core/base/executor.hpp
pratikvn: Good point. I changed everything to take vector now and an additional function for a scalar.

pratikvn:
@tcojean and @thoasm , thank you for your comments. This should be ready for review again.

dpcpp/test/base/executor_topology.dp.cpp
```diff
@@ -0,0 +1,95 @@
...
+    std::shared_ptr<gko::Executor> omp;
+    std::shared_ptr<const gko::DpcppExecutor> dpcpp;
+    std::shared_ptr<const gko::DpcppExecutor> dpcpp2;
+};
+
+
+TEST_F(DpcppExecutor, CanGetExecInfo)
+{
+    dpcpp = gko::DpcppExecutor::create(0, omp);
+    dpcpp2 = gko::DpcppExecutor::create(0, omp);
```
tcojean, line 87: Unused

dpcpp/test/base/executor_topology.dp.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+    auto exec_info = dpcpp->get_exec_info();
+
+    ASSERT_TRUE(exec_info.max_workitem_sizes.size() == 3);
```
tcojean, line 91: I would just test that it's > 0, who knows what different hardware will give as result here
```suggestion
    ASSERT_TRUE(exec_info.max_workitem_sizes.size() > 0);
```

include/ginkgo/core/base/executor.hpp
```diff
@@ -646,6 +649,97 @@ class Executor : public log::EnableLogging<Executor> {
...
+         *       In DPCPP, this is the subgroup sizes for the device associated
+         *       with the dpcpp executor.
+         */
+        std::vector<int> subgroup_sizes{};
+
+        /**
+         * The sizes of the work items for the executor.
+         *
+         * @note In CPU executors this is invalid.
+         *       In CUDA and HIP executors this is invalid.
```
tcojean, line 705: Couldn't this be max thread per block?

include/ginkgo/core/base/executor.hpp
```diff
@@ -646,6 +649,97 @@ class Executor : public log::EnableLogging<Executor> {
...
+         * @note Only relevant for I/O devices (GPUs).
+         */
+        std::string pci_bus_id = std::string(13, 'x');
+
+        /**
+         * The processing units closest to the device.
+         *
+         * @note Only relevant for CUDA, HIP and DPCPP executors.
+         */
+        std::vector<int> closest_pu_ids{};
```
tcojean, line 733: As a note, later we could make this relevant for CPUs as well if we pass to the constructor some info to select a NUMA node or cpuids targeted which we bind to or something of that form.

include/ginkgo/core/base/executor.hpp
```diff
@@ -892,6 +1007,57 @@ class ExecutorBase : public Executor {
...
+     * Bind to core
+     *
+     * @param id  The id to be bound to the calling object.
+     */
+    void bind_to_core(const int &id) const
+    {
+        MachineTopology::get_instance()->bind_to_cores(std::vector<int>{id});
+    }
+};
+
```
tcojean, line 1060:
```suggestion


```

tcojean (approved):

dpcpp/test/base/executor_topology.dp.cpp
pratikvn, line 87:
```suggestion
```

hip/test/base/hip_executor.hip.cpp
```diff
@@ -40,9 +40,15 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
 #include <memory>
+#include <thread>
```
pratikvn, line 43:
```suggestion
```

hip/test/base/hip_executor.hip.cpp
```diff
@@ -40,9 +40,15 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
 #include <type_traits>
 
 
+#if defined(__unix__) || defined(__APPLE__)
+#include <utmpx.h>
+#endif
+
+
```
pratikvn, line 51:
```suggestion
```

include/ginkgo/core/base/executor.hpp
```diff
@@ -1073,7 +1267,7 @@ class CudaExecutor : public detail::ExecutorBase<CudaExecutor>,
         int device_id, std::shared_ptr<Executor> master,
         bool device_reset = false);
 
-    ~CudaExecutor() { decrease_num_execs(this->device_id_); }
+    ~CudaExecutor() { decrease_num_execs(this->get_exec_info().device_id); }
```
thoasm, line 1270: nit:
```suggestion
    ~CudaExecutor() { decrease_num_execs(this->get_device_id()); }
```

include/ginkgo/core/base/executor.hpp
```diff
@@ -1249,7 +1470,7 @@ class HipExecutor : public detail::ExecutorBase<HipExecutor>,
                                                std::shared_ptr<Executor> master,
                                                bool device_reset = false);
 
-    ~HipExecutor() { decrease_num_execs(this->device_id_); }
+    ~HipExecutor() { decrease_num_execs(this->get_exec_info().device_id); }
```
thoasm, line 1473: nit:
```suggestion
    ~HipExecutor() { decrease_num_execs(this->get_device_id()); }
```

include/ginkgo/core/base/machine_topology.hpp
```diff
@@ -0,0 +1,399 @@
...
+ * This class also provides functionalities to bind objects in the topology to
+ * the execution objects. Binding can enhance performance by allowing data to be
+ * closer to the executing object.
+ *
+ * See the hwloc documentation
+ * (https://www.open-mpi.org/projects/hwloc/doc/) for more detailed
+ * information on topology detection and binding interfaces.
+ *
+ * @note A global object of MachineTopology type is created when first needed
+ * and only destroyed at the end of the program. This means that any subsequent
```
thoasm, line 86: Maybe add `[...] created in a thread safe manner and only [...]`

include/ginkgo/core/base/machine_topology.hpp
```diff
@@ -0,0 +1,399 @@
...
+
+        /**
+         * The global persistent id assigned to the object by hwloc.
+         */
+        size_type gp_id;
+
+        /**
+         * The closest numa.
+         */
+        int numa;
```
thoasm, line 178: nit: Why is the closest numa just called `numa` instead of something like `closest_numa`?

include/ginkgo/core/base/machine_topology.hpp
```diff
@@ -0,0 +1,399 @@
...
+
+protected:
+    /**
+     * Creates a new MachineTopology object.
+     */
+    static std::shared_ptr<MachineTopology> create()
+    {
+        return std::shared_ptr<MachineTopology>(new MachineTopology());
+    }
+
```
thoasm, line 328: I don't think you use this function anymore (since you use a local `static` variable instead).
```suggestion
```

include/ginkgo/core/base/machine_topology.hpp
```diff
@@ -0,0 +1,399 @@
...
+    /**
+     * The default constructor that loads the different objects.
+     */
+    MachineTopology();
```
thoasm, line 332: The constructor must be `private`, otherwise, you could inherit from it and create more than 1 object.
Actually, I think you can make all functions private since you should not inherit from this object.

include/ginkgo/core/base/machine_topology.hpp
```diff
@@ -0,0 +1,399 @@
...
+                      std::vector<io_obj_info> &vector) const;
+
+    /**
+     *
+     * @internal
+     *
+     * Get object id from the os index
+     */
+    inline int get_obj_id_by_os_index(
+        const std::vector<normal_obj_info> &objects, size_type os_index) const;
```
thoasm, line 372: Are you sure `inline` actually does something here? If this function was defined here (implementation was in this header), I am pretty sure it would be inlined, however, since the implementation is in a completely different file, I don't think it is possible to inline.

thoasm (commented):
Good job implementing all the changes. I only have some minor comments, mostly formatting and some `auto` use in `for` loops.
I would, however, like to have the constructor of the Singleton change to private, after that, I will approve this PR.

include/ginkgo/core/base/executor.hpp
```diff
@@ -646,6 +649,97 @@ class Executor : public log::EnableLogging<Executor> {
...
+         */
+        std::vector<int> closest_pu_ids{};
+    };
+
+    /**
+     * Gets the exec info struct
+     *
+     * @return  the exec_info struct
+     */
+    const exec_info &get_exec_info() const { return this->exec_info_; }
```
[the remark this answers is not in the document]
pratikvn: I think we want the users to be able to query the information in the `exec_info` directly if they want to. It also helps in tests. This is a const getter so they cannot modify it anyway. The non-const getter is protected.

pratikvn:
Weird! The PR has not been updated even though I have pushed and the [Ginkgo CI is running](https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/pipelines/245415302)

include/ginkgo/core/base/executor.hpp
```diff
@@ -892,6 +1007,58 @@ class ExecutorBase : public Executor {
...
+ *
+ * The executor object can then be bound to the bind-able objects from the
+ * machine topology by calling the member functions of this class.
+ */
+class EnableBinding {
+public:
+    /**
+     * Bind to Processing units (PU)
+     *
+     * @param ids  The ids to be bound to the object.
```
Slaedr, line 1023:
```suggestion
     * @param ids  The ids of PUs to which to bind the current process or OS thread.
```

include/ginkgo/core/base/executor.hpp
```diff
@@ -892,6 +1007,58 @@ class ExecutorBase : public Executor {
...
+     */
+    void bind_to_pu(const int &id) const
+    {
+        MachineTopology::get_instance()->bind_to_pus(std::vector<int>{id});
+    }
+
+    /**
+     * Bind to cores
+     *
+     * @param ids  The ids to be bound to the calling object.
```
Slaedr, line 1043:
```suggestion
     * @param ids  The ids corresponding to cores to be bound to the calling process.
```

Slaedr (changes requested):
This is nice work - it will enable some nice capabilities. Here's my first review, where I have a few concerns at the moment. I've put the main one here:

I believe the documentation for binding (eg. `EnableBinding::bind_to_pus`) is misleading. We are not binding executor objects, but rather the current process or OS thread, to cores and PUs. From `hwloc_set_cpubind`:
> Bind current process or thread on cpus given in physical bitmap set.

If we create another executor on the same process and call these functions again, I guess it will also change the binding for the original executor. So I suggested some change to the documentation for `EnableBinding`. But this raises the larger question: should this kind of binding even be part of the executor, given that we can have multiple executors on a given process? Perhaps it makes sense to have a `Binding` class that's purely a helper for using `MachineTopology`, rather than the current `EnableBinding` that Executors inherit from?

Also, the name `num_pe_per_cu` is quite inconsistent between backends. Please see my main comment about this in the `executor.hpp` file and smaller comments in other places.

include/ginkgo/core/base/executor.hpp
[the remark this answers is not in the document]
thoasm: OK, that is fine with me.

core/devices/machine_topology.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+
+namespace detail {
+
+
+class topo_bitmap {
+public:
+#if GKO_HAVE_HWLOC
+    using bitmap_type =
+        std::remove_pointer<decltype(hwloc_bitmap_alloc())>::type;
```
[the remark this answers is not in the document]
thoasm: I think it is somewhat safe to assume that `hwloc_bitmap_alloc()` returns a pointer (since `alloc` usually returns a pointer). However, in this context, it is not necessary to know that it is a pointer, so removing it would be even more safe:
```suggestion
    using bitmap_type =
        decltype(hwloc_bitmap_alloc())::type;
```
Alternatively, you could obviously use `hwloc_bitmap_t` directly.
When I wrote this part as a suggestion, I had not idea what it actually returned, and I needed to remove the pointer in order to work with `unique_ptr`.

thoasm (approved):
LGTM!

include/ginkgo/core/base/executor.hpp
```diff
@@ -646,6 +649,97 @@ class Executor : public log::EnableLogging<Executor> {
...
+
+        /**
+         * The sizes of the work items for the executor.
+         *
+         * @note In CPU executors this is invalid.
+         *       In CUDA and HIP executors this is invalid.
+         *       In DPCPP, this is the maximum sizes for the workitems for the
+         *       device associated with the dpcpp executor.
+         */
+        std::vector<int> max_workitem_sizes{};
```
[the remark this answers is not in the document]
pratikvn: I am not too sure about this. @tcojean mentioned that it is possible that some dpcpp device might have a different value.

include/ginkgo/core/base/executor.hpp
```diff
@@ -646,6 +649,97 @@ class Executor : public log::EnableLogging<Executor> {
...
+         *       In CUDA and HIP executors this is the number of Streaming
+         *       Multiprocessors. In DPCPP, this is the number of computing
+         *       units.
+         */
+        int num_computing_units = -1;
+
+        /**
+         * The number of processing elements per computing unit in the executor.
+         *
+         * @note In CPU executors this is equivalent to the number of SIMD units
```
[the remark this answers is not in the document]
pratikvn: I think SMT is not an usual term for the GPU ? That is why I went for PE. Processing element (PE) to me abstracts nicely SMT in CPUs and warps in GPUs. 

But I do agree that `max_workgroup_size` is not `num_pe_per_cu`. I will add another variable to the struct and for CUDA/HIP set that to max threads per block as you suggested.

include/ginkgo/core/base/executor.hpp
```diff
@@ -724,6 +818,27 @@ class Executor : public log::EnableLogging<Executor> {
...
+
+    /**
+     * Gets the modifiable exec info object
+     *
+     * @return  the pointer to the exec_info object
+     */
+    exec_info &get_exec_info()
+    {
+        return const_cast<exec_info &>(
+            static_cast<const Executor &>(*this).get_exec_info());
```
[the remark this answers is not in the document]
pratikvn: Though this is ugly, I think this is valid code. Actually it is something recommended by Scott Meyers in his book (Effective C++).  Here is what he says: 

> Casting away the const on the return value is safe, in this case,
because whoever called the non- const getter must have had a non-
const object in the first place. Otherwise they couldn’t have called a
non- const function. So having the non- const getter call the const
version is a safe way to avoid code duplication, even though it requires
a cast. 

Moreover this is a protected member and hence accesible only within this class and its derived classes.

third_party/CMakeLists.txt
```diff
@@ -13,6 +13,10 @@ if(GINKGO_BUILD_TESTS)
     ginkgo_add_subdirectory(GTest gtest)
 endif()
 
+if(NOT GINKGO_FORCE_NO_HWLOC AND GINKGO_HAVE_HWLOC)
```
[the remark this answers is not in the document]
pratikvn: Yes, by default `GINKGO_HAVE_HWLOC` is set to `1` and `GINKGO_USE_EXTERNAL_HWLOC` is `ON`, so if found, the external is always used.

core/devices/machine_topology.cpp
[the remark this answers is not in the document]
pratikvn: I have updated this now.

core/devices/machine_topology.cpp
```diff
@@ -0,0 +1,263 @@
...
+            io_obj_info{obj, obj->logical_index, obj->os_index, obj->gp_index,
+                        hwloc_bitmap_first(ancestor->nodeset), ancestor});
+        // Get the corresponding cpuset of the ancestor nodeset
+        detail::topo_bitmap ancestor_cpuset;
+        hwloc_cpuset_from_nodeset(this->topo_.get(), ancestor_cpuset.get(),
+                                  ancestor->nodeset);
+        // Find the cpu objects closest to this device from the ancestor cpuset
+        // and store their ids for binding purposes
+        int closest_pu_id = -1;
+        int closest_os_id = hwloc_bitmap_first(ancestor_cpuset.get());
```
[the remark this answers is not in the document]
pratikvn: I am getting all the PUs in the ancestor cpuset in a for each type loop below. I just use the first bit set in the bitmap as a starting point.

core/test/base/executor.cpp
```diff
@@ -156,6 +162,55 @@ TEST(OmpExecutor, IsItsOwnMaster)
...
+    return gko::MachineTopology::get_instance()->get_core(log_id)->os_id;
+}
+
+
+TEST(OmpExecutor, CanBindToASpecificCore)
+{
+    auto omp = gko::OmpExecutor::create();
+    auto cpu_sys = sched_getcpu();
+
+    const int bind_core = 4;
```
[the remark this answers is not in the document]
pratikvn: Good point. I will change everything to 1 and 3

cuda/test/base/cuda_executor_topology.cu
```diff
@@ -0,0 +1,164 @@
...
+
+    std::vector<int> bind_cores = {6, 3};
+    cuda->bind_to_cores(bind_cores);
+
+    auto cpu_sys = sched_getcpu();
+    ASSERT_TRUE(cpu_sys == get_core_os_id(3) || cpu_sys == get_core_os_id(6));
+}
+
+
+TEST_F(CudaExecutor, ClosestCpusIsPopulated)
```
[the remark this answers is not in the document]
pratikvn: Not that I know of. If anyone is aware of some way I can do this, I would definitely be happy to add it here.

core/devices/machine_topology.cpp
```diff
@@ -0,0 +1,263 @@
...
+    auto id = bind_ids.data();
+    // Set the given ids to a bitmap
+    for (auto i = 0; i < num_ids; ++i) {
+        GKO_ASSERT(id[i] < obj.size());
+        GKO_ASSERT(id[i] >= 0);
+        hwloc_bitmap_set(bitmap_toset.get(), obj[id[i]].os_id);
+    }
+
+    // Singlify to reduce expensive migrations.
+    hwloc_bitmap_singlify(bitmap_toset.get());
```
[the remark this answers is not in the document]
pratikvn: Good point. I will make this an additional option and update the doc.

include/ginkgo/core/base/executor.hpp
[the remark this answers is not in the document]
thoasm: I don't understand why you go through so much casts for the simple line:
`return exec_info_;`
`get_exec_info()` only works for non-const `*this`, therefore, you don't need the `const_cast`.
Additionally, having 2 consecutive `const_cast`s does not make sense. You only use `const_cast` to remove const-ness, meaning the inner-cast is used in a wrong manner (if anything, it should be casted with a `static_cast`).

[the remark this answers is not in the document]
pratikvn: You are right. I am not really sure what I was thinking here. Sorry for the confusion.

include/ginkgo/core/base/executor.hpp
```diff
@@ -892,6 +1007,58 @@ class ExecutorBase : public Executor {
 
 #undef GKO_DECLARE_EXECUTOR_FRIEND
 
+
+/**
+ * This class provides some binding functions from the MachineTopology class and
+ * the executors can derive from this class as necessary.
+ *
+ * The executor object can then be bound to the bind-able objects from the
```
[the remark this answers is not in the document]
pratikvn: Now binding can only be performed by the MachineTopology class global instance.

Slaedr (approved):
Nice! Thanks a lot!

So now I guess the user is expected to use `MachineTopology` directly to set bindings, which is perfectly fine for me.

As I mentioned in the [discussion](https://github.com/ginkgo-project/ginkgo/discussions/694), my only concern now is the name `num_smt_per_cu`. Sorry about that!

include/ginkgo/core/base/executor.hpp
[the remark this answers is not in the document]
pratikvn: I think I will leave this as a vector, because there are essentially no performance impacts as this is done during creation and as it is also easier to manage in case something in the future changes.

pratikvn:
I think this is ready to be merged, unless someone has any more comments.

upsj (approved):
LGTM!

.gitlab-ci.yml
```diff
@@ -22,6 +22,7 @@ stages:
   BUILD_OMP: "OFF"
   BUILD_CUDA: "OFF"
   BUILD_HIP: "OFF"
+  DISABLE_HWLOC: "OFF"
```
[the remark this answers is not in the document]
Slaedr: Or `ENABLE_HWLOC`?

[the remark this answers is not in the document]
pratikvn: Yeah, it sounds reasonable.

pratikvn:
Regarding the licensing, I think it is nice to have all the licenses that Ginkgo uses in one place for anyone to refer to. In the future, if we use some code from any library with that license again, then we can again just refer to ABOUT-LICENSING without adding any more lines.

Issue #555, opened 2020-06-05
Adding Functionality needed by openCARP


fritzgoebel:
So far, I think we are missing the following functionalities:

- diagonal scaling: openCARP needs to compute LMR where L and R are diagonal matrices. This could potentially be done with a new diagonal matrix format

- matrix axpy: openCARP needs to compute a * A + b * B. This is not trivial as it potentially changes the sparsity pattern, if new zero entries occur even to something else that the union of the sparsity patterns of A and B

- extract the diagonal from a matrix

- initialize a matrix with all zeros

Any ideas / comments on how to handle a diagonal matrix format?

upsj:
Matrix axpy: For sparse matrices that's called spgeam and we already have an implementation hidden inside the HIP advanced spgemm and I could easily add it for the other executors as well. We only need to figure out how to represent/call it. If you call advanced Csr->apply(alpha, Identity, beta, Csr) maybe?

fritzgoebel:
> Matrix axpy: For sparse matrices that's called spgeam and we already have an implementation hidden inside the HIP advanced spgemm and I could easily add it for the other executors as well. We only need to figure out how to represent/call it. If you call advanced Csr->apply(alpha, Identity, beta, Csr) maybe?

That would be awesome! I think then we could just call the advanced spgemm, right?

fritzgoebel:
> Matrix axpy: For sparse matrices that's called spgeam and we already have an implementation hidden inside the HIP advanced spgemm and I could easily add it for the other executors as well. We only need to figure out how to represent/call it. If you call advanced Csr->apply(alpha, Identity, beta, Csr) maybe?

Your representation makes sense, it would be awesome if you could add this :)

pratikvn:
For diagonal scaling: I see two approaches:

1. Add a new diagonal matrix format as you suggest.
2. Add a diagonal scaling operation to all the matrix formats.

I personally think the second approach might be better because it would be easier to implement and I dont see any particular reason to add a new format for representing a diagonal matrix format. Additionally, you would also need to implement conversions from one diagonal to all other formats and vice versa which I am not sure is such a good idea. The diagonal scaling operation could look like :

```c++
auto mat = gko::matrix::Csr<>::create(...);
auto scaled_matmat = gko::matrix::Csr<>::create(...);
auto diag = gko::Array<>{exec, {values}};
mat->scale_diagonal(lend(diag), lend(scaled_mat));
```

[...]

PR #556, opened 2020-06-05
Add CSR SpGEAM

upsj:
This PR adds support for sparse matrix-matrix addition (spgeam) for CSR matrices.
It can be executed via `Csr::apply(Scalar, Identity, Scalar, Csr)`.
The algorithm only works correctly on (column-)sorted matrices, as it's basically a merge between the rows from both matrices.
The PR also removes special treatment for zeros in advanced spgemm.

This partially addresses #555


upsj:
I would say so, [cuSPARSE](https://docs.nvidia.com/cuda/cusparse/index.html#csrgeam2) and hipSPARSE use variations of the name. MKL uses `csradd` instead, but they also call their SpGEMM `csrmult*`

upsj:
That's already present in Ginkgo, as a workaround for hipSPARSE's missing advanced spgemm from a few versions ago

hip/matrix/csr_kernels.hip.cpp
```diff
@@ -653,14 +653,15 @@ void advanced_spgemm(std::shared_ptr<const HipExecutor> exec,
...
-                      d_row_ptrs, d_col_idxs, d_vals, c);
+        select_spgeam(
+            spgeam_kernels(),
+            [&](int compiled_subwarp_size) {
+                return compiled_subwarp_size >= nnz_per_row ||
+                       compiled_subwarp_size == config::warp_size;
+            },
+            syn::value_list<int>(), syn::type_list<>(), exec, valpha,
+            c_tmp_row_ptrs, c_tmp_col_idxs, c_tmp_vals, vbeta, d_row_ptrs,
+            d_col_idxs, d_vals, c);
```
upsj, line 664: Yes, I am aware of the formatting issues, I will push a fix when the pipeline finishes :)

omp/components/csr_spgeam.hpp
```diff
@@ -52,7 +52,8 @@ namespace omp {
  * Adds two (sorted) sparse matrices.
  *
  * Calls begin_cb(row) on each row to initialize row-local data
- * Calls entry_cb(a, b, a_val, b_val, local_data) on each non-zero of the output
+ * Calls entry_cb(row, col, a_val, b_val, local_data) on each output non-zero of
+ * the output.
```
fritzgoebel, line 56: I think there's an output too much here.

cuda/matrix/csr_kernels.cu
```diff
@@ -579,6 +582,75 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
...
+    c_builder.get_col_idx_array().resize_and_reset(c_nnz);
+    c_builder.get_value_array().resize_and_reset(c_nnz);
+    auto c_col_idxs = c->get_col_idxs();
+    auto c_vals = c->get_values();
+    kernel::spgeam<subwarp_size><<<num_blocks, default_block_size>>>(
+        as_cuda_type(alpha), a_row_ptrs, a_col_idxs, as_cuda_type(a_vals),
+        as_cuda_type(beta), b_row_ptrs, b_col_idxs, as_cuda_type(b_vals), m,
+        c_row_ptrs, c_col_idxs, as_cuda_type(c_vals));
+}
+
```
yhmtsai, line 619:
```suggestion
```

hip/matrix/csr_kernels.hip.cpp
```diff
@@ -670,6 +670,33 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
...
+template <typename ValueType, typename IndexType>
+void spgeam(std::shared_ptr<const DefaultExecutor> exec,
+            const matrix::Dense<ValueType> *alpha,
+            const matrix::Csr<ValueType, IndexType> *a,
+            const matrix::Dense<ValueType> *beta,
+            const matrix::Csr<ValueType, IndexType> *b,
+            matrix::Csr<ValueType, IndexType> *c)
+{
+    auto valpha = exec->copy_val_to_host(alpha->get_const_values());
+    auto vbeta = exec->copy_val_to_host(beta->get_const_values());
```
yhmtsai, line 682: do we have any rule or consideration to use the coefficient on device(POINTER_MODE_DEVICE) or host(POINTER_MODE_HOST)? AFAIK, we use the coefficient like pointer_mode_device in spmv. some routines are only for pointer_mode_host

reference/matrix/csr_kernels.cpp
```diff
@@ -294,12 +291,8 @@ void advanced_spgemm(std::shared_ptr<const ReferenceExecutor> exec,
...
     for (size_type a_row = 0; a_row < num_rows; ++a_row) {
         local_row_nzs.clear();
-        if (vbeta != zero(vbeta)) {
-            spgemm_accumulate_row(local_row_nzs, d, vbeta, a_row);
-        }
-        if (valpha != zero(valpha)) {
-            spgemm_accumulate_row2(local_row_nzs, a, b, valpha, a_row);
-        }
+        spgemm_accumulate_row(local_row_nzs, d, vbeta, a_row);
+        spgemm_accumulate_row2(local_row_nzs, a, b, valpha, a_row);
```
yhmtsai, line 295: no matter what alpha, beta are, always get the csr with the union sparsity of A, B?

upsj: Yes, I don't really consider the cases `alpha=0` or `beta=0` (effectively scaling all matrix elements) important enough to justify special treatment for them. cuSPARSE/hipSPARSE afaik does the same, unless alpha or beta are `nullptr`.

yhmtsai: I see.

hip/matrix/csr_kernels.hip.cpp
upsj: You're probably right, I'll load the values on the GPU instead

upsj:
I added a bit of documentation to the Csr LinOp so the SpGEMM and SpGEAM are not hidden to the users.
Screenshot_2020-06-19 Ginkgo gko matrix Csr ValueType, IndexType Class Template Reference

PR #557, opened 2020-06-05
Fix the (cuda|hip)DeviceReset issue.

### LINKED ISSUE #461: CUDA executor always calls `cudaDeviceReset()` when destroyed

tcojean:
To fix this problem, an optional boolean class is added which controls
whether to reset or not the device controlled by the executor. By 
default, the device is not reset.

All examples, benchmarks and the last executors from the tests are
modified to pass the correct value to reset the device after use. This
ensures that our CI tools and users can keep tracking memory leaks
internally.

Fixes #461

### DIFF
--- a/include/ginkgo/core/base/executor.hpp
+++ b/include/ginkgo/core/base/executor.hpp
@@ -785,7 +785,10 @@ private:
 
 /**
  * Controls whether the DeviceReset function should be called thanks to a
- * boolean.
+ * boolean. Note that in any case, `DeviceReset` is called only after destroying
+ * the last Ginkgo executor. Therefore, it is sufficient to set this flag to the
+ * last living executor in Ginkgo. Setting this flag to an executor which is not
+ * destroyed last has no effect.
  */
 class EnableDeviceReset {
 public:
--- a/include/ginkgo/core/base/executor.hpp
+++ b/include/ginkgo/core/base/executor.hpp
@@ -783,6 +783,40 @@ private:
 };
 
 
+/**
+ * Controls whether the DeviceReset function should be called thanks to a
+ * boolean.
+ */
+class EnableDeviceReset {
+public:
+    /**
+     * Set the device reset capability.
+     *
+     * @param device_reset  whether to allow a device reset or not
+     */
+    void set_device_reset(bool device_reset) { device_reset_ = device_reset; }
+
+    /**
+     * Returns the current status of the device reset boolean for this executor.
+     *
+     * @return the current status of the device reset boolean for this executor.
+     */
+    bool get_device_reset() { return device_reset_; }
+
+protected:
+    /**
+     * Instantiate an EnableDeviceReset class
+     *
+     * @param device_reset  the starting device_reset status. Defaults to false.
+     */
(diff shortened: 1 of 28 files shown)


include/ginkgo/core/base/executor.hpp
```diff
@@ -783,6 +783,39 @@ class ExecutorBase : public Executor {
...
+ * Controls whether the DeviceReset function should be called thanks to a
+ * boolean.
+ */
+class EnableDeviceReset {
+public:
+    /**
+     * Toggles the Device Reset capability. If it was previously true, set
+     * false, and vice versa.
+     */
+    void toggle_device_reset() { device_reset_ = !device_reset_; }
```
upsj, line 796: In my opinion, setting and getting are more common use cases than toggling and getting.

include/ginkgo/core/base/executor.hpp
```diff
@@ -783,6 +783,40 @@ class ExecutorBase : public Executor {
 };
 
 
+/**
+ * Controls whether the DeviceReset function should be called thanks to a
+ * boolean.
```
yhmtsai, line 788: Could you add some comments about the reset only considered in the last destroyed executor (device_id)?
If some executor with reset on but some with reset off, the resetting call only depends on the last one.
Also, no matter how users set these executors, ginkgo does not call reset when some executors (under ginkgo) are alive.

tcojean: Good point.

tcojean:
@yhmtsai I think changing the behavior is fine in this case since it's a fix (in most cases, resetting is more trouble than it's worth). It does break binary compatibility also since we now have an extra optional parameter. I don't think we can do much about this.

yhmtsai:
@tcojean thanks for explanation.

Issue #426, opened 2019-12-18
`Csr::automatical` behaves wrong when the same is used for multiple matrices


thoasm:
When multiple matrices use the same `Csr::automatical` strategy object, the last matrix that calls `process` determines the strategy for all other matrices that use the same strategy object.
This is because we use the `name_` of a strategy to determine the behavior, while each strategy object has one name. We set that name in `automatical` when we determined the best strategy (for example [the load balance here](https://github.com/ginkgo-project/ginkgo/blob/develop/include/ginkgo/core/matrix/csr.hpp#L494)). That means one `automatical` can only have one distinct strategy name, which means all matrices with this object will use the same strategy, even if it is not ideal for them.

A simple fix would be to copy the strategy in the constructor in Csr instead of using the same as a `std::shared_ptr`. However, this kind of defeats the purpose of having a `std::shared_ptr` to begin with, and might also cause some current tests to fail (because the pointer itself is not identical anymore).

PR #559, opened 2020-06-05
Fix the automatical strategy when shared.

tcojean:
The automatical strategy has an issue when shared, it behaves as the
wrong strategy if the matrices using it have different properties.
This fixes this issue only by extending the current interface, and not
changing any existing function. All other (more elegant) approaches
I thought of would break the current interface. In effect:
+ Add a new virtual `copy` function to the strategies in order to create
  a new shared pointer polymorphically.
+ The approach used is to always copy the strategy, whenever a CSR
  matrix is instantiated or copied. Thankfully, the strategies are
  currently light objects so this should create few overhead.
+ Add tests which ensure that matrices with difference properties work
  correctly with the automatical strategy.

Fixes https://github.com/ginkgo-project/ginkgo/issues/426

### DIFF
--- a/hip/test/matrix/csr_kernels.hip.cpp
+++ b/hip/test/matrix/csr_kernels.hip.cpp
@@ -652,7 +652,7 @@ TEST_F(Csr, SortUnsortedMatrixIsEquivalentToRef)
 
 TEST_F(Csr, OneAutomaticalWorksWithDifferentMatrices)
 {
-    auto automatical = std::make_shared<Mtx::automatical>();
+    auto automatical = std::make_shared<Mtx::automatical>(hip);
     auto row_len_limit = std::max(automatical->nvidia_row_len_limit,
                                   automatical->amd_row_len_limit);
     auto load_balance_mtx = Mtx::create(ref);
--- a/hip/test/matrix/csr_kernels.hip.cpp
+++ b/hip/test/matrix/csr_kernels.hip.cpp
@@ -650,4 +650,29 @@ TEST_F(Csr, SortUnsortedMatrixIsEquivalentToRef)
 }
 
 
+TEST_F(Csr, OneAutomaticalWorksWithDifferentMatrices)
+{
+    auto automatical = std::make_shared<Mtx::automatical>();
+    auto row_len_limit = std::max(automatical->nvidia_row_len_limit,
+                                  automatical->amd_row_len_limit);
+    auto load_balance_mtx = Mtx::create(ref);
+    auto classical_mtx = Mtx::create(ref);
+    load_balance_mtx->copy_from(
+        gen_mtx<Vec>(1, row_len_limit + 1000, row_len_limit + 1));
+    classical_mtx->copy_from(gen_mtx<Vec>(50, 50, 1));
+    auto load_balance_mtx_d = Mtx::create(hip);
+    auto classical_mtx_d = Mtx::create(hip);
+    load_balance_mtx_d->copy_from(load_balance_mtx.get());
+    classical_mtx_d->copy_from(classical_mtx.get());
+
+    load_balance_mtx_d->set_strategy(automatical);
+    classical_mtx_d->set_strategy(automatical);
+
+    EXPECT_EQ("load_balance", load_balance_mtx_d->get_strategy()->get_name());
+    EXPECT_EQ("classical", classical_mtx_d->get_strategy()->get_name());
+    ASSERT_NE(load_balance_mtx_d->get_strategy().get(),
+              classical_mtx_d->get_strategy().get());
+}
+
+
 }  // namespace
--- a/include/ginkgo/core/matrix/csr.hpp
+++ b/include/ginkgo/core/matrix/csr.hpp
@@ -597,13 +597,18 @@ public:
 
     void convert_to(Csr<ValueType, IndexType> *result) const override
     {
+        bool same_executor = this->get_executor() == result->get_executor();
         // NOTE: as soon as strategies are improved, this can be reverted
         result->values_ = this->values_;
         result->col_idxs_ = this->col_idxs_;
         result->row_ptrs_ = this->row_ptrs_;
         result->srow_ = this->srow_;
         result->set_size(this->get_size());
-        result->set_strategy(std::move(this->get_strategy()->copy()));
+        if (!same_executor) {
+            convert_strategy_helper(result);
+        } else {
+            result->set_strategy(std::move(this->get_strategy()->copy()));
+        }
         // END NOTE
     }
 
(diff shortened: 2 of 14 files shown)


include/ginkgo/core/matrix/csr.hpp
```diff
@@ -549,7 +598,13 @@ class Csr : public EnableLinOp<Csr<ValueType, IndexType>>,
     void convert_to(Csr<ValueType, IndexType> *result) const override
     {
         bool same_executor = this->get_executor() == result->get_executor();
-        EnableLinOp<Csr>::convert_to(result);
+        // NOTE: as soon as strategies are improved, this can be reverted
```
upsj, line 601: unrelated to your changes: Is there a reason why this function (and move_to(...)) is implemented in the header instead of csr.cpp?

include/ginkgo/core/matrix/csr.hpp
```diff
@@ -771,7 +826,7 @@ class Csr : public EnableLinOp<Csr<ValueType, IndexType>>,
           col_idxs_(exec, num_nonzeros),
           row_ptrs_(exec, size[0] + 1),
           srow_(exec, strategy->clac_size(num_nonzeros)),
-          strategy_(std::move(strategy))
+          strategy_(std::move(strategy->copy()))
```
upsj, line 829: I believe we no longer need to move it after the copy.
```suggestion
          strategy_(strategy->copy())
```

include/ginkgo/core/matrix/csr.hpp
```diff
@@ -804,7 +859,7 @@ class Csr : public EnableLinOp<Csr<ValueType, IndexType>>,
           col_idxs_{exec, std::forward<ColIdxsArray>(col_idxs)},
           row_ptrs_{exec, std::forward<RowPtrsArray>(row_ptrs)},
           srow_(exec),
-          strategy_(std::move(strategy))
+          strategy_(std::move(strategy->copy()))
```
upsj, line 862:
```suggestion
          strategy_(strategy->copy())
```

include/ginkgo/core/matrix/csr.hpp
```diff
@@ -549,7 +598,13 @@ class Csr : public EnableLinOp<Csr<ValueType, IndexType>>,
     void convert_to(Csr<ValueType, IndexType> *result) const override
     {
         bool same_executor = this->get_executor() == result->get_executor();
-        EnableLinOp<Csr>::convert_to(result);
+        // NOTE: as soon as strategies are improved, this can be reverted
+        result->values_ = this->values_;
+        result->col_idxs_ = this->col_idxs_;
+        result->row_ptrs_ = this->row_ptrs_;
+        result->srow_ = this->srow_;
+        result->strategy_ = std::move(this->get_strategy()->copy());
```
upsj, line 606: I guess you also need to set the LinOp size here

include/ginkgo/core/matrix/csr.hpp
tcojean: I'm unsure, usually we do, for example for `Coo`:
```c++
    using EnableLinOp<Coo>::convert_to;
    using EnableLinOp<Coo>::move_to;
```
So they are defined in the header in that case, maybe that was done intentionally to keep that situation when we fixed some strategy issues?

hip/test/matrix/csr_kernels.hip.cpp
```diff
@@ -638,4 +638,29 @@ TEST_F(Csr, SortUnsortedMatrixIsEquivalentToRef)
 }
 
 
+TEST_F(Csr, OneAutomaticalWorksWithDifferentMatrices)
+{
+    auto automatical = std::make_shared<Mtx::automatical>();
```
upsj, line 643:
```suggestion
    auto automatical = std::make_shared<Mtx::automatical>(hip);
```

tcojean: Thanks!

tcojean:
@yhmtsai I do not understand your issue properly. I think with the current implementation there is no way to change the executor dynamically. If the user wants to use `automatical` for both AMD and NVIDIA hardware, then he will need to anyway declare two different `make_shared<Csr::automatical>(cudaExec)` and `make_shared<Csr::automatical>(hipExec)` to have the proper values. In this case, there should be no issue with copies happening with the wrong values?

yhmtsai:
Originally, I think the case about amd_csr->copy_from(nvidia_csr) directly.
we do not support copy/move between amd/nvidia, right?
in host matrix -> amd matrix, the host matrix has warp_size = 32 by default. It will pass this to amd_csr, which leads inconsistent warp_size. `strategy_rebuild_helper` to reset the strategy by executor to make sure that the strategy uses the parameters from executor not others.

tcojean:
@yhmtsai Indeed, so far we do not even support `raw_copy_to` between AMD hardware and CUDA hardware. I think there would some work required to make it work.

Now I see the issue, I will try to see what I can do for that.

include/ginkgo/core/matrix/csr.hpp
```diff
@@ -597,13 +597,18 @@ class Csr : public EnableLinOp<Csr<ValueType, IndexType>>,
 
     void convert_to(Csr<ValueType, IndexType> *result) const override
     {
+        bool same_executor = this->get_executor() == result->get_executor();
         // NOTE: as soon as strategies are improved, this can be reverted
         result->values_ = this->values_;
         result->col_idxs_ = this->col_idxs_;
         result->row_ptrs_ = this->row_ptrs_;
         result->srow_ = this->srow_;
         result->set_size(this->get_size());
```
yhmtsai: Using `EnableLinOp<Csr>::convert_to(result);` only gets the shared_ptr not the copy, right?
Or keeping `EnableLinOp<Csr>::convert_to(result);` is also fine because you create a new strategy later

include/ginkgo/core/matrix/csr.hpp
```diff
@@ -549,10 +598,18 @@ class Csr : public EnableLinOp<Csr<ValueType, IndexType>>,
     void convert_to(Csr<ValueType, IndexType> *result) const override
     {
         bool same_executor = this->get_executor() == result->get_executor();
-        EnableLinOp<Csr>::convert_to(result);
+        // NOTE: as soon as strategies are improved, this can be reverted
+        result->values_ = this->values_;
+        result->col_idxs_ = this->col_idxs_;
+        result->row_ptrs_ = this->row_ptrs_;
+        result->srow_ = this->srow_;
+        result->set_size(this->get_size());
```
yhmtsai, line 606: Using `EnableLinOp<Csr>::convert_to(result);` only gets the shared_ptr not the copy, right?
Or keeping `EnableLinOp<Csr>::convert_to(result);` is also fine because you create a new strategy later

tcojean:
@yhmtsai I should have fixed that issue now. In the end, I put @upsj code from the `convert_to<Csr<next_precision...>>` into a helper, as that is exactly what needs to be called to fix this issue I believe.

Please tell me if I'm missing anything.

yhmtsai:
@tcojean I am fine with these changes. I do not have any question in this PR.

PR #560, opened 2020-06-07
Test all empty matrix conversions on Reference

### LINKED ISSUE #380: Make all conversions of empty matrices work
Some conversions will crash if the source matrix is empty. The ones I tested and failed resulted in a `Segmentation fault (core dumped)`. Example code:
```c++
    auto from_mtx = gko::matrix::Dense<>::create(exec, gko::dim<2>(0, 0));
    auto to_mtx = gko::matrix::Hybrid<>::create(exec, gko::dim<2>{0,0});

    from_mtx->convert_to(to_mtx.get());
```
I confirmed this bug with the following conversions for now:
`Dense` to: `Csr` (will be fixed in #379 ), `SparsityCsr`, `Hybrid`, `Sellp`.
The same applies to the `move_to` conversions.
I am sure there are more cases, these are just the ones I tested.

upsj:
This also fixes the found issues:
* The SparsityCsr value is sometimes uninitialized
* Dense convert_to skips empty matrices
* Conversions to sellp read from empty arrays (this bug is also present in the omp executor)

Closes #380

### DIFF
--- a/omp/matrix/dense_kernels.cpp
+++ b/omp/matrix/dense_kernels.cpp
@@ -472,8 +472,11 @@ void convert_to_sellp(std::shared_ptr<const OmpExecutor> exec,
             }
         }
     }
-    slice_sets[slice_num] =
-        slice_sets[slice_num - 1] + slice_lengths[slice_num - 1];
+
+    if (slice_num > 0) {
+        slice_sets[slice_num] =
+            slice_sets[slice_num - 1] + slice_lengths[slice_num - 1];
+    }
 }
 
 GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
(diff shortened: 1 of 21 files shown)


omp/matrix/dense_kernels.cpp
```diff
@@ -472,8 +472,11 @@ void convert_to_sellp(std::shared_ptr<const OmpExecutor> exec,
             }
         }
     }
-    slice_sets[slice_num] =
-        slice_sets[slice_num - 1] + slice_lengths[slice_num - 1];
+
+    if (slice_num > 0) {
+        slice_sets[slice_num] =
+            slice_sets[slice_num - 1] + slice_lengths[slice_num - 1];
+    }
```
tcojean, line 479: I think it would be nice if only for the conversion to sellp you could check at least for either CUDA or HIP that empty conversion also works, since you had to modify the kernels here for this. I looked only quickly and it should be fine as long as num_rows = 0, but making sure of it would maybe be a good idea.

PR #561, opened 2020-06-07
Provide initial guess for Composition

### LINKED ISSUE #541: Iterative solvers and Composition
Right now, the intermediate values used as x in composition-like LinOps (Composition, preconditioner::Ilu) are often uninitialized. We could use `apply_uses_initial_guess` to determine whether we need to initialize this with a sensible value.
For square matrices, we could use b, for non-square matrices, we could zero-initialize. Any other ideas?

upsj:
This PR provides an initial guess to Composition operators that need it. For square operators, this is the output of the previous operator, for rectangular, I don't see any better way than using zeros.

Also made two changes to Ginkgo overall:

* replace zero_array by a more general fill_array kernel that can be used in core code.
* turn GKO_ASSERT_ARRAY_EQ parameters from pointers to references

Fixes #541


core/components/fill_array.hpp
```diff
@@ -0,0 +1,100 @@
...
+}  // namespace components
+}  // namespace hip
+
+
+#undef GKO_DECLARE_ALL_AS_TEMPLATES
+
+
+}  // namespace kernels
+}  // namespace gko
+
```
yhmtsai, line 99:
```suggestion


```

core/base/composition.cpp
```diff
@@ -77,14 +87,26 @@ std::unique_ptr<LinOp> apply_inner_operators(
...
         auto out_data =
             data + (reversed_storage ? storage_size - out_size : size_type{});
         reversed_storage = !reversed_storage;
         out = Dense::create(exec, out_dim,
                             Array<ValueType>::view(exec, out_size, out_data),
                             num_rhs);
+        // for operators with initial guess: set initial guess
+        if (operators[i]->apply_uses_initial_guess()) {
+            if (op_size[0] == op_size[1]) {
+                // square matrix: we can use the previous output
```
yhmtsai, line 102: should we leave an option to use zero as initial guess in square matrix?

upsj: I would argue that this doesn't really make a difference, since for a few algorithms I can think of (IR, GMRES, CG, ...), the next iteration after `0` would always be `λb` for some scalar λ. I'm not sure we need to be *that* configurable.

cuda/test/solver/bicgstab_kernels.cpp
```diff
@@ -261,7 +261,7 @@ TEST_F(Bicgstab, CudaBicgstabInitializeIsEquivalentToRef)
     GKO_EXPECT_MTX_NEAR(d_beta, beta, 1e-14);
     GKO_EXPECT_MTX_NEAR(d_gamma, gamma, 1e-14);
     GKO_EXPECT_MTX_NEAR(d_omega, omega, 1e-14);
-    GKO_ASSERT_ARRAY_EQ(d_stop_status, stop_status);
+    GKO_ASSERT_ARRAY_EQ(*d_stop_status, *stop_status);
```
pratikvn, line 264: Line 151 above. `d_stop_status` is being created twice ? Maybe in other files as well.

upsj: It's being constructed once and assigned once (hence the *), so that should be fine

cuda/base/cublas_bindings.hpp
```diff
@@ -40,9 +40,9 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/base/exception_helpers.hpp>
 
 
+#include "core/components/fill_array.hpp"
```
tcojean, line 43: This is now unused, no?

cuda/matrix/sparsity_csr_kernels.cu
```diff
@@ -41,6 +41,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/matrix/dense.hpp>
 
 
+#include "core/components/fill_array.hpp"
```
tcojean, line 44: This also looks unused. Like most headers here actually since nothing is implemented.

cuda/preconditioner/isai_kernels.cu
```diff
@@ -38,6 +38,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/matrix/csr.hpp>
 
 
+#include "core/components/fill_array.hpp"
```
tcojean, line 41: unused

cuda/solver/gmres_kernels.cu
```diff
@@ -159,8 +159,9 @@ void finish_arnoldi(std::shared_ptr<const CudaExecutor> exec,
             // TODO: this condition should be tuned
             // single rhs will use vendor's dot, otherwise, use our own
             // multidot_kernel which parallelize multiple rhs.
-            zero_array(hessenberg_iter->get_size()[1],
-                       hessenberg_iter->get_values() + k * stride_hessenberg);
+            components::fill_array(
```
tcojean, line 162: The header probably should be taken from the `.hpp` file into this file.

cuda/test/solver/bicgstab_kernels.cpp
tcojean: If you look line 151, there really are two times `d_stop_status = std::unique_ptr<...`.

hip/base/hipblas_bindings.hip.hpp
```diff
@@ -41,9 +41,9 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/base/executor.hpp>
 
 
+#include "core/components/fill_array.hpp"
```
tcojean, line 44: Also unused.

hip/matrix/sparsity_csr_kernels.hip.cpp
```diff
@@ -41,6 +41,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/matrix/dense.hpp>
 
 
+#include "core/components/fill_array.hpp"
```
tcojean, line 44: Unused.

hip/preconditioner/isai_kernels.hip.cpp
```diff
@@ -41,6 +41,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/matrix/csr.hpp>
 
 
+#include "core/components/fill_array.hpp"
```
tcojean, line 44: unused

hip/solver/gmres_kernels.hip.cpp
```diff
@@ -164,8 +164,9 @@ void finish_arnoldi(std::shared_ptr<const HipExecutor> exec, size_type num_rows,
             // TODO: this condition should be tuned
             // single rhs will use vendor's dot, otherwise, use our own
             // multidot_kernel which parallelize multiple rhs.
-            zero_array(hessenberg_iter->get_size()[1],
-                       hessenberg_iter->get_values() + k * stride_hessenberg);
+            components::fill_array(
```
tcojean, line 167: Again header should probably be moved.

cuda/test/solver/bicgstab_kernels.cpp
upsj: I see, thanks! I was looking in the wrong file

PR #562, opened 2020-06-08
throw when all the criterion element are nullptr

### LINKED ISSUE #545: vector<pointer> in Factory initialized by nullptr contains one element
We uses vector<shared_ptr> on criteria with the nullptr initialization.
(I also use vector<shared_ptr> to allow multiple elements in one parameter in multigrid)
nullptr is also shared_ptr, so the criteria vector contains one nullptr element.
```
switch (factories.size()) {
    case 0:
        GKO_NOT_SUPPORTED(nullptr);
        return nullptr;
    case 1:
        return factories[0];
```
Do not get the error.
If we do not pass any criteria into solvers, get segmentation fault in criteria generation not `GKO_NOT_SUPPORTED`.

Two possible solutions:
1. add another `GKO_FACTORY_PARAMETER` without any initialization.
2. check whether the each content is nullptr or not and the number of available pointer. 

Option 1 still need to implement option 2 to avoid passing nullptr.
However, option 1 can know whether users do not use this parameter or users give some nullptr member.

I prefer option 2 because I think option 1's benefit is not needed.
any comment or idea?

yhmtsai:
This PR considers the nullptr of the criterion.
the first element of `combine` function must be valid such that the executor is captured correctly.
(I do not create a new object here)
If all elements of criteria are nullptr, it will throw `gko::NotSupported`.
It closes #545


include/ginkgo/core/stop/combined.hpp
```diff
@@ -120,12 +127,21 @@ std::shared_ptr<const CriterionFactory> combine(FactoryContainer &&factories)
...
+            GKO_NOT_SUPPORTED(nullptr);
+        }
         return factories[0];
     default:
-        auto exec = factories[0]->get_executor();
-        return Combined::build()
-            .with_criteria(std::forward<FactoryContainer>(factories))
-            .on(exec);
+        if (factories[0] == nullptr) {
+            // first factory must be vaild to capture executor
```
pratikvn, line 136:
```suggestion
            // first factory must be valid to capture executor
```

core/test/solver/bicg.cpp
```diff
@@ -185,7 +185,12 @@ TYPED_TEST(Bicg, CanSetPreconditionerGenerator)
                     .with_reduction_factor(
                         gko::remove_complex<value_type>(1e-6))
                     .on(this->exec))
-            .with_preconditioner(Solver::build().on(this->exec))
+            .with_preconditioner(
+                Solver::build()
+                    .with_criteria(
+                        gko::stop::Iteration::build().with_max_iters(3u).on(
```
hartwiganzt, line 191: I do understand correctly that you only run 3 iterations?

upsj: That's a core test, it doesn't actually run the kernels, it just checks the overall framework/setup.

yhmtsai: After this change, non-setting criteria throws `NotSupported`. Need to add a valid criteria to run the test.

PR #564, opened 2020-06-09
Fix norm calculation to always use non-complex values

### LINKED ISSUE #547: Adding `remove_complex` to norms in general inside Ginkgo
We need to change multiple parts of our implementation if we want to properly support complex:
1. `Dense::compute_norm2` should put the result into a `Dense<remove_complex<ValueType>>`
  While using `squared_norm()` instead of `compute_dot` inside the kernels. Reference kernel example:
   https://github.com/ginkgo-project/ginkgo/blob/develop/reference/matrix/dense_kernels.cpp#L187
2. The `Convergence` logger should also use the non-complex storage type here:
   https://github.com/ginkgo-project/ginkgo/blob/develop/core/log/convergence.cpp#L61
3. The `ResidualNormReduction` stopping criterion should use the non-complex vector type: `using Vector = matrix::Dense<remove_complex<ValueType>>` here:
   https://github.com/ginkgo-project/ginkgo/blob/develop/include/ginkgo/core/stop/residual_norm_reduction.hpp#L71

There might be more, but these are the ones I could find this far.

_Originally posted by @thoasm in https://github.com/ginkgo-project/ginkgo/issues/534#issuecomment-634031364_

thoasm:
Currently, all norm calculations result in a norm vector of `ValueType` instead of `remove_complex<ValueType>`. Also, some norms were calculated incorrectly (`sqrt(x * x)` instead of `x * conj(x)`).
This is all fixed in this PR. Additionally, `compute_norm2` is rewritten to no longer depend on `compute_dot` (on all executors).

TODO:
- [x] Test if the re-written CUDA and HIP norm2 implementations work properly (the tests only cover the call to the Cublas and Hipblas implementations).

Closes #547.


core/solver/fcg.cpp
```diff
@@ -67,7 +67,6 @@ void Fcg<ValueType>::apply_impl(const LinOp *b, LinOp *x) const
     constexpr uint8 RelativeStoppingId{1};
 
     auto exec = this->get_executor();
-    size_type num_vectors = dense_b->get_size()[1];
```
pratikvn, line 70: I guess this was unused ?

include/ginkgo/core/base/types.hpp
```diff
@@ -403,10 +416,9 @@ GKO_ATTRIBUTES constexpr bool operator!=(precision_reduction x,
  *                Should take one argument, which is replaced by the
  *                value type.
  */
-#define GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE(_macro) \
-    template _macro(float);                         \
-    template _macro(double);                        \
-    template _macro(std::complex<float>);           \
+#define GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE(_macro)          \
+    GKO_INSTANTIATE_FOR_EACH_NON_COMPLEX_VALUE_TYPE(_macro); \
```
pratikvn, line 420: :)

core/solver/fcg.cpp
thoasm: Yes, that is why I removed it.

PR #565, opened 2020-06-12
Prevent benchmark logger crashes with OpenMP

upsj:
As observed by @nbeams, some OpenMP algorithms may call executor operations (alloc and free) from multiple threads, which can crash the benchmark loggers.

This PR fixes these race conditions by ensuring exclusive access using a C++ standard library mutex.

The reported numbers for such operations are still bogus, though!

For a complete fix, we would need to pre-allocate all temporary data structures from the main thread. Offending kernels: [spgemm](https://github.com/ginkgo-project/ginkgo/blob/1af71c31c76905dccbd32a1d3c69a3396a35db42/omp/matrix/csr_kernels.cpp#L219) and [jacobi](https://github.com/ginkgo-project/ginkgo/blob/1af71c31c76905dccbd32a1d3c69a3396a35db42/omp/preconditioner/jacobi_kernels.cpp#L358)
We can definitely fix Jacobi, but for SpGEMM I would prefer moving to another algorithm.

### DIFF
--- a/omp/preconditioner/isai_kernels.cpp
+++ b/omp/preconditioner/isai_kernels.cpp
@@ -109,16 +109,20 @@ void generic_generate(std::shared_ptr<const DefaultExecutor> exec,
     const auto i_cols = inverse_mtx->get_const_col_idxs();
     auto i_vals = inverse_mtx->get_values();
 
+    auto num_threads = static_cast<size_type>(omp_get_max_threads());
+    // RHS for local trisystem
+    gko::Array<ValueType> rhs_array{exec, row_size_limit * num_threads};
+    // memory for dense trisystem
+    gko::Array<ValueType> trisystem_array{
+        exec, row_size_limit * row_size_limit * num_threads};
+
 #pragma omp parallel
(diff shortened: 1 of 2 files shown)


omp/preconditioner/isai_kernels.cpp
```diff
@@ -109,16 +109,21 @@ void generic_generate(std::shared_ptr<const DefaultExecutor> exec,
     const auto i_cols = inverse_mtx->get_const_col_idxs();
     auto i_vals = inverse_mtx->get_values();
 
+    auto num_threads = static_cast<size_type>(omp_get_max_threads());
+    // OpenMP seems to have issues copying the arrays, so we do it manually
```
upsj, line 113:
```suggestion
```

pratikvn: Does this still need to be updated ?

upsj: Yes, I just forgot to remove the comment :)

PR #566, opened 2020-06-14
Fix windows GitHub action

yhmtsai:
This PR fixes the issue of windows GitHub action

1. fix cuda version 10.2 in `Windows-build: cuda102/release/shared (only compile)`
2. enable cc1plus to use larger address space [reference](https://www.intel.com/content/www/us/en/programmable/support/support-resources/knowledge-base/embedded/2016/cc1plus-exe--out-of-memory-allocating-65536-bytes.html)
3. reduce the number of jobs

To be honest, I am not sure whether 2, 3 really help or not. 
I will rerun the workflow 2-3 times to check whether all are passed.

### DIFF
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -10,7 +10,7 @@ jobs:
     - uses: actions/checkout@v2
     - name: setup
       run: |
-        choco install cuda -y
+        choco install cuda --version=10.2.89.20191206 -y
     - name: configure
       run: |
         $env:ChocolateyInstall = Convert-Path "$((Get-Command choco).Path)\..\.."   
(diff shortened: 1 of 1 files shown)


.github/workflows/windows-build.yml
```diff
@@ -85,10 +85,12 @@ jobs:
       run: |
         set PATH=%PATH:C:\Program Files\Git\bin;=%
         set PATH=%PATH:C:\Program Files\Git\usr\bin;=%
+        bcdedit /set IncreaseUserVa 3072
```
pratikvn, line 88: Is this the maximum limit we can set ?

yhmtsai: `IncreaseUserVa` needs to be in 2048 ~ 3072 on https://docs.microsoft.com/en-us/windows-hardware/drivers/devtest/bcdedit--set
But I also see this command is only available on 32-bit OS on some comments.

PR #567, opened 2020-06-14
HIP Clang compatibility

tcojean:
Make Ginkgo compatible with HIP Clang compilation.

The MI50 machine available at ICL's cluster uses hip-clang as a compiler.
hip-clang actually fixes some of the most horrible thing with the classical
`hcc` compiler, in particular it's not necessary to ban flags anymore 
because all of these were `hcc` related.

Nonetheless, this is still an ongoing process. I believe what is here is mostly
what we need to do, but due to toolchain issue which took me a *very* long
time to debug, I cannot successfully compile everything on the MI50 machine
from ICL yet.

+ Add the new option to pass in HIP clang specific options
+ Properly detect and populate the new critical ROCM_PATH and
  HIP_CLANG_PATH variables.
+ Make HIP THRUST easier to be found thanks to using an optional environment variable.
+ Also add some missing test for the HipError and related classes
+ Also fix some missing includes which were found

As I believe this might be relevant to the future (for our CI systems or any
cluster), I will add in a comment the sneaky toolchain issue that hip clang has.


tcojean:
With the hip clang setup, the main compiler used by HIP is sitting in
`/opt/rocm/llvm/bin`. The basic issue is that it does not allow to
compile with c++11 at all which creates link issues in our library (we
rely on C++11 like many libraries nowadays). This should be
supported though. As we will see the root cause are toolchain issues.

### /opt/rocm/llvm/bin/clang++ cannot compile c++11 with a default installation
The basic fact that c++11 is not properly compiled is clear when we
compile a sample c++ code (the attached `c++11_issue.cpp` file). When
compiling it with `/opt/rocm/llvm/bin/clang++ -std=c++11` and running nm
on it, we get:
```
[tcojean@methane ~]$ nm compiled_with_clang++.exe | c++filt | grep char_traits
                 U std::basic_ostringstream<char, std::char_traits<char>, std::allocator<char> >::str() const@@GLIBCXX_3.4
                 U std::basic_ostream<char, std::char_traits<char> >::operator<<(int)@@GLIBCXX_3.4
                 U std::basic_string<char, std::char_traits<char>, std::allocator<char> >::~basic_string()@@GLIBCXX_3.4
                 U std::basic_ostringstream<char, std::char_traits<char>, std::allocator<char> >::basic_ostringstream(std::_Ios_Openmode)@@GLIBCXX_3.4
                 U std::basic_ostringstream<char, std::char_traits<char>, std::allocator<char> >::~basic_ostringstream()@@GLIBCXX_3.4
                 U std::basic_ostream<char, std::char_traits<char> >& std::operator<< <std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&, char const*)@@GLIBCXX_3.4
```

When compiling with a recent g++, we get instead:
```
[tcojean@methane ~]$ nm compiled_with_g++.exe | c++filt | grep char_traits
                 U std::__cxx11::basic_ostringstream<char, std::char_traits<char>, std::allocator<char> >::str() const@@GLIBCXX_3.4.21
                 U std::basic_ostream<char, std::char_traits<char> >::operator<<(int)@@GLIBCXX_3.4
                 U std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::~basic_string()@@GLIBCXX_3.4.21
                 U std::__cxx11::basic_ostringstream<char, std::char_traits<char>, std::allocator<char> >::basic_ostringstream()@@GLIBCXX_3.4.26
                 U std::__cxx11::basic_ostringstream<char, std::char_traits<char>, std::allocator<char> >::~basic_ostringstream()@@GLIBCXX_3.4.21
                 U std::basic_ostream<char, std::char_traits<char> >& std::operator<< <std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&, char const*)@@GLIBCXX_3.4
```

The important part here is that the clang++ version does not use
`std::__cxx11::...` as is typical of all c++11 code like we see with the
g++ version. In practice in the middle of a project which combines HIP
and non HIP code, this creates linker issues with `std::basic_string...`
expected but not found. This is actually really sneaky because you
pass in the flag requesting C++11 compilation and no error or anything
is thrown such as "I cannot compile C++11 code with this setup".

### Root cause: HIP clang does not know recent gcc installations
To understand the root of the problem, we can find which gcc installation
that clang is linked with:
```
[tcojean@methane ~]$ /opt/rocm/llvm/bin/clang -v
clang version 11.0.0 (/data/jenkins_workspace/centos_pipeline_job_3.5/rocm-rel-3.5/rocm-3.5-30-20200528/7.5/external/llvm-project/clang 6c08b900599eee52e12bce1e76b20dc413ce30e7)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /opt/rocm/llvm/bin
Found candidate GCC installation: /usr/lib/gcc/x86_64-redhat-linux/4.8.2
Found candidate GCC installation: /usr/lib/gcc/x86_64-redhat-linux/4.8.5
Selected GCC installation: /usr/lib/gcc/x86_64-redhat-linux/4.8.5
Candidate multilib: .;@m64
Candidate multilib: 32;@m32
Selected multilib: .;@m64
```

Even when loading the relevant modules (I have gcc 9.2.0 loaded), this
clang installation does not find any newer version of gcc to use as a
candidate installation. I believe that is because they only search in
the path `/usr/lib/gcc/xxxx` by default. There might be an easy way to
tell clang where to look, but the only thing I could find is the following.

### Unpractical solution: using a recent gcc with clang the manual way
There is one unpractical way for projects to target another gcc
toolchain, but which works fine when compiling one file. The command
looks like this:
```
[tcojean@methane ~]$
/opt/rocm/llvm/bin/clang++ c++11_issue.hip.cpp -std=c++11 --gcc-toolchain=/nfs/apps/spack/opt/spack/linux-centos7-x86_64/gcc-7.2.0/gcc-9.2.0-3mdu5gl5jymnhvqnmfaavtex2zguy2gt/lib/gcc/x86_64-pc-linux-gnu/9.2.0/ -I/nfs/apps/spack/opt/spack/linux-centos7-x86_64/gcc-7.2.0/gcc-9.2.0-3mdu5gl5jymnhvqnmfaavtex2zguy2gt/include/c++/9.2.0/ -I/nfs/apps/spack/opt/spack/linux-centos7-x86_64/gcc-7.2.0/gcc-9.2.0-3mdu5gl5jymnhvqnmfaavtex2zguy2gt/include/c++/9.2.0/x86_64-pc-linux-gnu/ -L/nfs/apps/spack/opt/spack/linux-centos7-x86_64/gcc-7.2.0/gcc-9.2.0-3mdu5gl5jymnhvqnmfaavtex2zguy2gt/lib/gcc/x86_64-pc-linux-gnu/9.2.0/ -B/nfs/apps/spack/opt/spack/linux-centos7-x86_64/gcc-7.2.0/gcc-9.2.0-3mdu5gl5jymnhvqnmfaavtex2zguy2gt/lib/gcc/x86_64-pc-linux-gnu/9.2.0/ -o compiled_with_clang++.exe
```

We need to pass in the `--gcc-toolchain` argument, and pass the location
of the c++ headers for this compiler version (split in two different
places) and library path. In addition, we also need to pass the -B
option (binary search path) to find ctrbegin.o. Most of these can be controled
with environment variables, but the `gcc-toolchain` would have to stay. After
doing all of this, we finally have a correct looking binary:
```
[tcojean@methane ~]$ nm compiled_with_clang++.exe | c++filt | grep char_traits
                 U std::__cxx11::basic_ostringstream<char, std::char_traits<char>, std::allocator<char> >::str() const@@GLIBCXX_3.4.21
                 U std::basic_ostream<char, std::char_traits<char> >::operator<<(int)@@GLIBCXX_3.4
                 U std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::~basic_string()@@GLIBCXX_3.4.21
                 U std::__cxx11::basic_ostringstream<char, std::char_traits<char>, std::allocator<char> >::basic_ostringstream()@@GLIBCXX_3.4.26
                 U std::__cxx11::basic_ostringstream<char, std::char_traits<char>, std::allocator<char> >::~basic_ostringstream()@@GLIBCXX_3.4.21
                 U std::basic_ostream<char, std::char_traits<char> >& std::operator<< <std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&, char const*)@@GLIBCXX_3.4

```

### Other unpractical solution (untested?)
Maybe using a CMake cross-compilation toolchain file would work?
Although we need to tell CMake to use this *only* when compiling
HIP code, and we're also not really cross-compiling...

### Summary and possible fixes
In summary, clang is unable to find a modern version of gcc. Here are some
possible fixes I can think of depending on how the HIP clang was installed.
1. When compiling clang, the option `-DGCC_INSTALL_PREFIX` can be given to
   CMake with the path for the GCC installation to use for this clang.
2. A dirty option I can think of is to try to create and maintain symlinks into
   /usr/lib/gcc and /usr/include/c++ and hopefully `clang -v` is able to find
   everything by itself. There are maybe cleaner ways to do this, maybe
   `alternatives` can be used for this?
3. Installing a modern g++ compiler from the redhat devtoolsets. I saw in LLVM's
   code that they specifically detect the devtoolset path for redhat systems.

upsj:
We might be able to make the unpractical solution work by determining the gcc toolchain path from the implicit include directories used by the CXX compiler:
```
message("${CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES}")
[...];/usr/include/x86_64-linux-gnu/c++/7;[...]
```

tcojean:
Indeed I guess that would work, though I think that's for a specific case of using clang with gcc, ad we need to populate all the separate `-B` `-L` and whatever option. I would rather if there is an easy way to fix this on the server side, but I will keep this in mind.

upsj:
That might be an issues either with LLVM or SPACK - a quick look through the clang source [[1]](https://github.com/llvm/llvm-project/blob/a974b33a10745b528c34f0accbd230b0a4e1fb87/clang/include/clang/Config/config.h.cmake#L52) [[2]](https://github.com/llvm/llvm-project/blob/a974b33a10745b528c34f0accbd230b0a4e1fb87/clang/lib/Driver/ToolChains/Gnu.cpp#L1654) suggests that this should be set to a suitable GCC toolchain location at configuration time.
If we go down this way, we should probably use `-isystem` instead of `-I`. Do we really need `-B` or is `-L` enough? Are we linking to non-library files, or is this needed for another step?

tcojean:
I don't think hip-clang can be installed from spack, so it cannot give the `-DGCC_INSTALL_PREFIX` at compile time.

We need the `-B` since that is used to locate necessary binaries, in particular `crtbegin.o`.

upsj:
Ah sorry, I guess you knew all that already, I must have skipped over your first comment too quickly.

tcojean:
Yes maybe I didn't work this properly, but the last three points are some other solutions I thought of which are rather on the server/system side. One of which include the `GCC_INSTALL_PREFIX` thing. 
What really surprised me is that looking at clang's code, it didn't seem like it tried to use the environment or something to add some runtime check. I guess there can be multiple issues with that.

CMakeLists.txt
```diff
@@ -46,6 +46,8 @@ set(GINKGO_HIP_NVCC_COMPILER_FLAGS "" CACHE STRING
     "Set the required HIP nvcc compiler flags. Current default is an empty string.")
 set(GINKGO_HIP_HCC_COMPILER_FLAGS "" CACHE STRING
     "Set the required HIPC HCC compiler flags. Current default is an empty string.")
+set(GINKGO_HIP_CLANG_COMPILER_FLAGS "" CACHE STRING
+    "Set the required HIPC CLANG compiler flags. Current default is an empty string.")
```
yhmtsai, line 50:
```suggestion
    "Set the required HIP CLANG compiler flags. Current default is an empty string.")
```

hip/CMakeLists.txt
```diff
@@ -1,20 +1,29 @@
...
+        set(ROCM_PATH $ENV{ROCM_PATH} CACHE PATH "Path to which ROCM has been installed")
+    elseif(DEFINED ENV{HIP_PATH})
+        set(ROCM_PATH "$ENV{HIP_PATH}/.." CACHE PATH "Path to which ROCM has been installed")
     else()
+        set(ROCM_PATH "/opt/rocm" CACHE PATH "Path to which ROCM has been installed")
+    endif()
+endif()
+
+if(NOT DEFINED HIPBLAS_PATH)
+    if(DEFINED ENV{HIPBLAS_PATH})
```
yhmtsai, line 12: I forgot to mention it before. In hipblas/hipsparse, they use `if (DEFINED $ENV{})` https://github.com/tcojean/hipBLAS/blob/ad14b3bf5ec6da92abd646a533268aace33e0902/CMakeLists.txt#L9
which should be `if (DEFINED ENV{})` here

tcojean:
Yes, 3.5 always uses hip-clang. They deprecated hcc. Technically, I guess it could still be installed manually though.
https://rocmdocs.amd.com/en/latest/Programming_Guides/Programming-Guides.html

hip/CMakeLists.txt
tcojean: Good find. Maybe they fixed it upstream. I need to update these repositories.

cmake/hip_helpers.cmake
```diff
@@ -1,11 +1,15 @@
 macro(ginkgo_hip_ban_link_hcflag target)
-    get_target_property(GINKGO_TARGET_ILL ${target} INTERFACE_LINK_LIBRARIES)
-    string(REPLACE "-hc " "" GINKGO_TARGET_NEW_ILL "${GINKGO_TARGET_ILL}")
-    set_target_properties(${target} PROPERTIES INTERFACE_LINK_LIBRARIES "${GINKGO_TARGET_NEW_ILL}")
+    if(TARGET ${target})
```
upsj, line 2: Can you explain what this change is necessary for?

core/device_hooks/cuda_hooks.cpp
```diff
@@ -30,8 +30,12 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ******************************<GINKGO LICENSE>*******************************/
 
+#include <memory>
+#include <string>
+
```
upsj, line 35:
```suggestion


```

core/device_hooks/hip_hooks.cpp
```diff
@@ -30,8 +30,12 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ******************************<GINKGO LICENSE>*******************************/
 
+#include <memory>
+#include <string>
+
```
upsj, line 35:
```suggestion


```

cmake/hip_helpers.cmake
tcojean: I can add a comment if needed in the code. In general, it's due to this comment in the PR text:

The MI50 machine available at ICL's cluster uses hip-clang as a compiler.
hip-clang actually fixes some of the most horrible thing with the classical
hcc compiler, in particular it's not necessary to ban flags anymore
because all of these were hcc related.

tcojean:
I can now confirm that this now fully works with a hip-clang setup. I ended up trying it out on our AMD system after updating the ROCm suite to 3.5. I had to fix a last naughty issue concerning the benchmarks linking to `hip`.

cmake/hip_helpers.cmake
upsj: I guess my question was a bit unspecific. I was especially wondering about the condition `TARGET ${target}`. But now that I looked at the call site, it makes perfect sense.

tcojean: To maybe give more detail, if the target doesn't exist as is the case with hip-clang, since there is no more `hcc` and related targets linked, this will do nothing. Otherwise, this banning process would create an error if left without protection.

Right now, this code is purely equivalent to a
```cmake
if (GINKGO_HIP_VERSION GREATER "3.5)
   # old code
endif()
```

PR #569, opened 2020-06-15
Fix Sonarqube warnings in examples/benchmarks

upsj:
We currently have a rather bad reliability rating again, probably due to some upstream changes in Sonarqube. This PR fixes [the reported "bugs"](https://sonarcloud.io/project/issues?id=ginkgo-project_ginkgo&resolved=false&types=BUG).

### DIFF
--- a/examples/poisson-solver/poisson-solver.cpp
+++ b/examples/poisson-solver/poisson-solver.cpp
@@ -69,9 +69,9 @@ void generate_rhs(Closure f, ValueType u0, ValueType u1,
 {
     const auto discretization_points = rhs->get_size()[0];
     auto values = rhs->get_values();
-    const ValueType h = 1.0 / (discretization_points + 1);
+    const ValueType h = 1.0 / static_cast<ValueType>(discretization_points + 1);
     for (gko::size_type i = 0; i < discretization_points; ++i) {
-        const auto xi = ValueType(i + 1) * h;
+        const auto xi = static_cast<ValueType>(i + 1) * h;
         values[i] = -f(xi) * h * h;
     }
     values[0] += u0;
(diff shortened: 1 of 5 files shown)


examples/poisson-solver/poisson-solver.cpp
```diff
@@ -69,9 +69,9 @@ void generate_rhs(Closure f, ValueType u0, ValueType u1,
 {
     const auto discretization_points = rhs->get_size()[0];
     auto values = rhs->get_values();
-    const ValueType h = 1.0 / (discretization_points + 1);
+    const ValueType h = 1.0 / static_cast<ValueType>(discretization_points + 1);
     for (gko::size_type i = 0; i < discretization_points; ++i) {
-        const auto xi = ValueType(i + 1) * h;
```
tcojean, line 74: They don't like constructor style cast?

upsj: They have no issue with this, but I thought since we already moved to `static_cast` mostly everywhere, I could do the same here. Not really a comprehensive change, though.

upsj:
@pratikvn Can you elaborate on that? Is it possible to hide such warnings from the web interface?

pratikvn:
@upsj , Yes, On the [web interface](https://sonarcloud.io/project/issues?id=ginkgo-project_ginkgo&pullRequest=569&resolved=false&types=BUG), you should see a list of the Bugs, Vulnerability and Code smells which can be navigated through with the left side menu. To mark any of these as false positives and otherwise, you need to be logged in and you can log in through your github account. 

Currently, these are the bugs that are shown: https://sonarcloud.io/project/issues?id=ginkgo-project_ginkgo&resolved=false&types=BUG. You can mark it as resolved/wont fix etc by selecting from the drop down menu under Open.

PR #573, opened 2020-06-17
Test workflow for a ginkgo 2.0.0 preparation branch

tcojean:
This is a test PR to see how we can manage a ginkgo version 2.0.0 preparation branch.

I propose this new target branch `ginkgo-2.0.0`. The idea for this branch is to collect all relevant commits and new features we want to be in an eventual Ginkgo 2.0.0, while having pull request and review capabilities like usual, maybe specific CI steps if required, all the while being able to play with the future ginkgo 2.0.0 to see how it behaves when everything is put together.

This branch should use the following workflow, maybe you have another idea:
+ It is based on top of develop
+ Reviews are required to push anything into this branch
+ Unlike develop/master, there should be no extra merge commit so that we maintain a clean history. That means all PR should use "rebase merge request" instead of the usual merge that we do.
+ We should be able to rebase this branch on top of the most recent `develop` through a PR (as this one does).

Now comes the critical part. Anyone working off this branch to add a new feature, let's say in a branch named `new-gko-v2-feature`, **should not** use `git rebase`. The issue is that, since this `ginkgo-2.0.0` will be rebased on top of develop (as this PR should do), all the history will become conflict commits since they will have disappeared in the base `ginkgo-2.0.0` branch itself once the rebase on top of `develop` has been done. That means that any feature branch such as this `new-gko-v2-feature` based off the `ginkgo-2.0.0` branch needs to use another updating scheme to stay up to date with this branch. Here is the most sensible one I believe:
+ Work normally in the `new-gko-v2-feature` based on top of `ginkgo-2.0.0`
+ When you want to update to the most recent `ginkgo-2.0.0`, create a duplicate branch somewhere, let's call this branch `backup_new-gko-v2-feature`.
+ Use `git reset --hard ginkgo-2.0.0` in the branch `new-gko-v2-feature`.
+ Use  `git cherry-pick` to pick your actual commits from the branch `backup_new-gko-v2-feature`. You can usually cherry-pick a range of commits, so you could do something like this: `git cherry-pick backup_new-gko-v2-feature..oldest_relevant_commit_sha`. I believe this will cherry-pick all commits from the last one in `backup_new-gko-v2-feature` to the first one of interest.
+ Push the new state of `new-gko-v2-feature` if everything went well. Delete `backup_new-gko-v2-feature`.

What do you think about this workflow?


pratikvn:
But this would mean that there would be some features in ginkgo-2.0.0 which would not be develop, right ? Would this make sense after the 1.2.0 release ? Because I expect that after releasing 1.2.0, we rebase develop on 1.2.0 and hence all the PR's and commits on develop would directly go on to become 2.0.0 unless we plan to have another minor release 1.x before that.

tcojean:
There will probably be a number of minor releases before 2.0.0. I don't think 2.0.0 will take a short time to get out at all. A lot of things are left to do and we also need to be confident enough that we will not need a 3.0 soon after releasing 2.0.

Develop will not be rebased on top of 1.2.0. 1.2.0 will be merged into both master and develop. And after that is done, this new `ginkgo-2.0.0` will be rebased on top of it again.

pratikvn:
Yes, that is true. I guess we could have some bug-fix releases. 

For the interface breaking changes, this new branch structure could make a lot of sense. The workflow with the new ginkgo-2.0.0 looks reasonable to me.

tcojean:
Not only bug fixes. There is still GMRES mixed, there will probably be sycl support, and probably much more things.

On top of it, even if 2.0.0 could happen soon, we would still need a specific branch because we cannot break develop to now be 2.0.0 by starting to drop interface compatibility all of a sudden. I think it's much more safe to work in a specific branch and eventually deprecate the old develop at some point or something like that.

yhmtsai:
does `rebase -i` failed in this case?
for example, I have multigrid<-amgx<-develop now. when develop is updated, amax can use `git rebase` directly but multigrid need to use `rebase -i` on amgx
For multigrid, I need to delete some commit which should be related to amgx in interactive mode.

tcojean:
Yes that is exactly the kind of issue that will happen with this workflow. It works fine to use `interactive rebase` to solve these issues in the short term, but if this new version 2.0 branch has a long life, it may have a lot of extra commits and deleting commits manually during rebase will have a limited scalability. That's why the workflow with `cherry-pick` can be useful for this case.

tcojean:
Since we are not using this for now, let me close and we might come back to it when we start working towards a 2.0 release.

Issue #147, opened 2018-10-12
Adding more stopping criteria taking into account residual norm


tcojean:
The current residual norm stopping criterion is a Residual Norm Reduction, which has the parameter `reduction_factor` and is defined as:
```
||Ax_n - b||
-------------    <= tau
||Ax_0 - b||
```
(Note: this needs to be fixed because it uses `<` instead of `<=` and this doesn't work when both `x` and `b` are 0)

We should have some more stopping criteria such as:

### Residual Norm
We can call the parameter we pass `tolerance` here?
```
||Ax_n - b|| <=  tau
```

### Normalized Residual Norm
A name needs to be found for the parameter
```
||Ax_n - b||
-------------    <=  tau
    ||b||
```
For this one we could also use ||A|| instead of ||b||, do we do this in the same criterion, keep only one use case, ... ?

PR #577, opened 2020-06-24
Add relative and absolute residual norm stopping criteria

fritzgoebel:
This PR adds stopping criteria for relative (to norm(b)) and absolute residual norm.
It abstracts the three criteria `ResidualNormReduction,` `RelativeResidualNorm` and `AbsoluteResidualNorm,` which only differ in the way they initialize `starting_tau_` from `ResidualNorm`. 

If someone has a better name for the base class, I am open to suggestions, maybe something line `ResidualNormCriterion`?

Closes https://github.com/ginkgo-project/ginkgo/issues/147


core/stop/residual_norm.cpp
```diff
@@ -30,53 +30,70 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
...
     }
     bool all_converged = true;
 
-    this->get_executor()->run(
-        residual_norm_reduction::make_residual_norm_reduction(
-            dense_tau, starting_tau_.get(), parameters_.reduction_factor,
-            stoppingId, setFinalized, stop_status, &this->device_storage_,
-            &all_converged, one_changed));
+    if (dense_tau->get_size()[1] > 1) {
+    }
```
yhmtsai, line 71: Is anything missing here?

include/ginkgo/core/stop/residual_norm.hpp
```diff
@@ -0,0 +1,265 @@
...
+#include <ginkgo/core/base/array.hpp>
+#include <ginkgo/core/base/math.hpp>
+#include <ginkgo/core/base/utils.hpp>
+#include <ginkgo/core/matrix/dense.hpp>
+#include <ginkgo/core/stop/criterion.hpp>
+
+
+namespace gko {
+namespace stop {
+
```
yhmtsai, line 49:
```suggestion


```

include/ginkgo/core/stop/residual_norm.hpp
```diff
@@ -0,0 +1,265 @@
...
+    std::unique_ptr<NormVector> starting_tau_{};
+    std::unique_ptr<NormVector> u_dense_tau_{};
+
+private:
+    remove_complex<ValueType> tolerance_{};
+    /* Contains device side: all_converged and one_changed booleans */
+    Array<bool> device_storage_;
+};
+
```
yhmtsai, line 84:
```suggestion


```

include/ginkgo/core/stop/residual_norm.hpp
yhmtsai: Could you add some description of ResidualNorm?

pratikvn:
I would still keep the abstract class name as `ResidualNormReduction` as we are still reducing the residual norm (but different types of the residual norm). The derived classes can be `AbsoluteResNormReduction` and `RelativeResNormReduction` or `AbsResNormReduction` and `RelResNormReduction` if the first two names are too long.

upsj:
@pratik I would disagree there, since to me the Reduction in the name refers to the amount we reduce the residual norm by, while e.g. for Absolute/RelativeResidualNorm we give an exact tolerance, not a reduction factor.

omp/test/stop/residual_norm_kernels.cpp
```diff
@@ -0,0 +1,342 @@
...
+
+template <typename T>
+class AbsoluteResidualNorm : public ::testing::Test {
+protected:
+    using Mtx = gko::matrix::Dense<T>;
+    using NormVector = gko::matrix::Dense<gko::remove_complex<T>>;
+
+    AbsoluteResidualNorm()
+    {
+        std::cout << "TOLERANCE: " << r<T>::value << "\n";
```
pratikvn, line 251: Can be removed ?

fritzgoebel:
> LGTM! Do you actually test with different values for the different tests? It didn't look to me like that was the case when skimming over the CUDA tests.

You mean different values for the tolerance? No, I use the same value as was used before for the ResidualNormReduction criterion..

include/ginkgo/core/stop/residual_norm_reduction.hpp
```diff
@@ -30,100 +30,5 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
...
-    Array<bool> device_storage_;
-};
-
-
-}  // namespace stop
-}  // namespace gko
-
-
-#endif  // GKO_CORE_STOP_RESIDUAL_NORM_REDUCTION_HPP_
+// This file is deprecated and will be removed in a later release.
```
yhmtsai, line 33: Maybe try `#warning this file ....` as @tcojean said
I prefer use `in a later major release` or `in 2.0.0 release`

cuda/test/stop/residual_norm_kernels.cpp
```diff
@@ -0,0 +1,366 @@
...
+            .residual_norm(d_res.get())
+            .check(RelativeStoppingId, true, &stop_status, &one_changed));
+    stop_status.set_executor(ref_);
+    ASSERT_TRUE(stop_status.get_data()[0].has_converged());
+    stop_status.set_executor(cuda_);
+    ASSERT_TRUE(one_changed);
+
+    res->at(0, 1) = tol * 1.0e+1;
+    d_res->copy_from(res.get());
+    ASSERT_TRUE(
```
yhmtsai, line 250: In the previous test, WaitsTillResidualGoal, `res` needs to be `tol` to satisfy the stopping criterion, but `res` in this test only needs to `tol * 10`. Do I misunderstand somthing?

cuda/test/stop/residual_norm_kernels.cpp
```diff
@@ -0,0 +1,366 @@
...
+    ASSERT_FALSE(
+        criterion->update()
+            .residual_norm(d_res.get())
+            .check(RelativeStoppingId, true, &stop_status, &one_changed));
+    stop_status.set_executor(ref_);
+    ASSERT_TRUE(stop_status.get_data()[0].has_converged());
+    stop_status.set_executor(cuda_);
+    ASSERT_TRUE(one_changed);
+
+    res->at(0, 1) = tol * 1.0e+1;
```
yhmtsai, line 354: Shouldn't it converge?

cuda/test/stop/residual_norm_kernels.cpp
fritzgoebel: I got these wrong before, my idea was to start with the same initial residual and right-hand side for all criteria and then check the different behavior of the criteria:
- ResidualReduction is satisfied when `|r| / |r_0| < tol`, so when `|r| < 100 * tol`
- RelativeResidualNorm is staisfied when `|r| / |b| < tol`, so when `|r| < 10 * tol`
- AbsoluteResidualNorm is satisfied when `|r| < tol`

include/ginkgo/core/stop/residual_norm_reduction.hpp
```diff
@@ -30,100 +30,10 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ******************************<GINKGO LICENSE>*******************************/
 
-#ifndef GKO_CORE_STOP_RESIDUAL_NORM_REDUCTION_HPP_
-#define GKO_CORE_STOP_RESIDUAL_NORM_REDUCTION_HPP_
```
yhmtsai, line 34: Keep this?

test_install/test_install.cpp
```diff
@@ -351,7 +351,7 @@ int main(int, char **)
         auto time = gko::stop::Time::build()
                         .with_time_limit(std::chrono::milliseconds(10))
                         .on(refExec);
-        // residual_norm_reduction.hpp
+        // residual_norm.hpp
         auto res_red = gko::stop::ResidualNormReduction<>::build()
```
tcojean, line 355: You need to add similar checks with the new RelativeResidualNorm and Absolute.

test_install/test_install_cuda.cu
```diff
@@ -343,7 +343,7 @@ int main(int, char **)
         auto time = gko::stop::Time::build()
                         .with_time_limit(std::chrono::milliseconds(10))
                         .on(cudaExec);
-        // residual_norm_reduction.hpp
+        // residual_norm.hpp
```
tcojean, line 346: Same here.

tcojean:
Looks like the residual_norm_reduction.hpp file is still included in the main ginkgo.hpp although it should be removed. I think this file is generated by a script so the script needs to be updated to not include this file.
https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/616917734

include/ginkgo/core/stop/residual_norm.hpp
```diff
@@ -0,0 +1,275 @@
...
+        this->u_dense_tau_ =
+            NormVector::create_with_config_of(this->starting_tau_.get());
+        dense_rhs->compute_norm2(this->starting_tau_.get());
+    }
+};
+
+
+/**
+ * The AbsoluteResidualNorm class is a stopping criterion which stops the
+ * iteration process when the residual norm is below a certain
```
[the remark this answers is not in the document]
fritzgoebel: I think this should be clear with the equation and the difference in residual norm / residual norm relative to the initial residual / residual norm relative to the right-hand side in the description.

PR #578, opened 2020-06-25
Add support for LLVM sanitizers

upsj:
This PR adds support for LLVM sanitizers and fixes the OpenMP ThreadSanitizer execution by using a separate libomp.so with TSAN support.


pratikvn:
Do we want this for 1.2.0 ?

upsj:
Not sure, I guess we can talk about that today. But at least the sanitizer results say "all is well" :)

PR #563, opened 2020-06-09
Add extract_diagonal for all matrix formats

fritzgoebel:
This PR introduces extraction of the diagonal (which is needed by openCARP) into a vector for all matrix formats.

### TODO
+ [x] Merge #580 
+ [x] Rebase and update to return diagonal matrix format.

PR #580, opened 2020-07-01
Add diagonal matrix format for diagonal scaling

fritzgoebel:
For now, Diagonal can be applied to Dense and Csr.
The normal apply does row wise scaling, the right apply does column wise scaling.


yhmtsai:
Does this pr should be merged first? After this, #563 `extract_diagonal` will return `Diagonal` not `Dense`

pratikvn:
I think we came to a conclusion to merge this first and make the extract diagonal return a Diagonal matrix format ?

include/ginkgo/core/matrix/diagonal.hpp
```diff
@@ -0,0 +1,183 @@
...
+
+template <typename ValueType>
+class Dense;
+
+
+/**
+ * This class is a utility which efficiently implements the diagonal matrix (a
+ * linear operator which scales a vector row wise).
+ *
+ * Objects of the Diagonal class always represent a square matrix, and
```
yhmtsai, line 54: I am not pretty sure here. In extract_diag, you can extract diag with small dim on non-square matrix. The diagonal may not apply/rapply original matrix?

fritzgoebel: You can extract the diagonal from a non square matrix, yes. The diagonal matrix will still be square, so for a skinny matrix, you can do a right apply with the resulting diagonal matrix, for a wide matrix you can do a normal apply.

core/matrix/diagonal.cpp
```diff
@@ -0,0 +1,202 @@
...
+    return std::move(tmp);
+}
+
+
+template <typename ValueType, typename IndexType>
+void Diagonal<ValueType, IndexType>::convert_to(
+    Csr<ValueType, IndexType> *result) const
+{
+    auto exec = this->get_executor();
+    auto tmp = Csr<ValueType, IndexType>::create(exec, this->get_size(),
```
yhmtsai, line 141: need to set the same strategy of result csr.

omp/matrix/diagonal_kernels.cpp
```diff
@@ -0,0 +1,188 @@
...
+                  const matrix::Csr<ValueType, IndexType> *b,
+                  matrix::Csr<ValueType, IndexType> *c)
+{
+    const auto diag_values = a->get_const_values();
+    c->copy_from(b);
+    auto csr_values = c->get_values();
+    const auto csr_row_ptrs = c->get_const_row_ptrs();
+
+#pragma omp parallel for
+    for (size_type row = 0; row < c->get_size()[0]; row++) {
```
yhmtsai, line 105:
```suggestion
    for (size_type row = 0; row < c->get_size()[0]; row++) {
        const auto scal = diag_values[row];
```
like apply_to_dense

reference/test/matrix/diagonal_kernels.cpp
```diff
@@ -0,0 +1,407 @@
...
+    auto result = gko::matrix::Csr<value_type, index_type>::create(
+        this->exec, gko::dim<2>{2});
+
+    // 2x3 times 3x3 = 2x2 --> mismatch for cols of diagonal and result
+    ASSERT_THROW(this->diag2->rapply(this->csr1.get(), result.get()),
+                 gko::DimensionMismatch);
+}
+
+
+TYPED_TEST(Diagonal, ConverstToCsr)
```
yhmtsai, line 348:
```suggestion
TYPED_TEST(Diagonal, ConvertsToCsr)
```

include/ginkgo/core/matrix/diagonal.hpp
```diff
@@ -0,0 +1,181 @@
...
+ * require one array to store their values.
+ *
+ * @tparam ValueType  precision of matrix elements
```
tcojean, line 58: missing `IndexType` param documentation

include/ginkgo/core/matrix/diagonal.hpp
```diff
@@ -0,0 +1,181 @@
...
+     * @param exec  Executor associated to the matrix
+     */
+    explicit Diagonal(std::shared_ptr<const Executor> exec)
+        : Diagonal(std::move(exec), size_type{})
+    {}
+
+    /**
+     * Creates an Diagonal matrix of the specified size.
+     *
+     * @param size  size of the matrix
```
tcojean, line 135: Missing `exec` parameter.

include/ginkgo/core/matrix/diagonal.hpp
```diff
@@ -0,0 +1,181 @@
...
+     * @note This is the constant version of the function, which can be
+     *       significantly more memory efficient than the non-constant version,
+     *       so always prefer this version.
+     */
+    const value_type *get_const_values() const noexcept
+    {
+        return values_.get_const_data();
+    }
+
+    void rapply(const LinOp *b, LinOp *x) const
```
tcojean, line 108: Missing documentation

reference/matrix/diagonal_kernels.cpp
```diff
@@ -0,0 +1,179 @@
...
+{
+    const auto diag_values = a->get_const_values();
+    c->copy_from(b);
+    auto csr_values = c->get_values();
+    const auto csr_row_ptrs = c->get_const_row_ptrs();
+
+    for (size_type row = 0; row < c->get_size()[0]; row++) {
+        for (size_type idx = csr_row_ptrs[row]; idx < csr_row_ptrs[row + 1];
+             idx++) {
+            csr_values[idx] *= diag_values[row];
```
tcojean, line 102: Can also be extracted above the second for loop

fritzgoebel:
The `IndexType` template is only needed for the CSR matrix the diagonal is potentially applied or converted to. One option is keeping it, the other is, similarly to dense, having multiple overloads for `convert_to`, `move_to`, `apply_impl` and `rapply_impl` to Csr.

I tend to keeping it, does someone disagree?

upsj:
I would prefer removing it, since it does not represent any inherent property of the matrix, but how it is being used. Also IMO there should be nothing preventing a conversion between DiagonalMatrix<double, int32> and DiagonalMatrix<double, int64>.

fritzgoebel:
That is true, but removing it means that if we wanted to add other index types at some point, we'd have to go and create new overloads for any of them instead of having it just work. 
I think I need to know the IndexType when calling *apply_impl to know if the matrix the diagonal is applied to is a `Csr<ValueType, IndexType>`, right?

upsj:
Adding another index type seems rather unlikely for me. Only mayybe something between 32 and 64 bit, but commonly aligned types are nicer from a memory access standpoint. You can try converting to both types Csr<ValueType, int32> and Csr<ValueType, int64> and apply the correct kernels to the casted object.

core/test/matrix/CMakeLists.txt
```diff
@@ -1,11 +1,12 @@
...
 ginkgo_create_test(ell)
 ginkgo_create_test(hybrid)
 ginkgo_create_test(identity)
 ginkgo_create_test(permutation)
 ginkgo_create_test(sellp)
 ginkgo_create_test(sparsity_csr)
 ginkgo_create_test(csr_builder)
-ginkgo_create_test(coo_builder)
\ No newline at end of file
+ginkgo_create_test(coo_builder)
```
tcojean, line 12: I believe this file should be sorted. Never saw before that `coo_builder` and `csr_builder` were at the very end

hip/test/matrix/diagonal_kernels.hip.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+namespace {
+
+
+class Diagonal : public ::testing::Test {
+protected:
+    using Csr = gko::matrix::Csr<>;
+    using Diag = gko::matrix::Diagonal<>;
+    using Dense = gko::matrix::Dense<>;
+    using Arr = gko::Array<int>;
```
pratikvn, line 61: Same comment regarding the types.

include/ginkgo/core/base/exception_helpers.hpp
```diff
@@ -211,6 +211,24 @@ inline dim<2> get_size(const dim<2> &size) { return size; }
     }
 
 
+/**
+ * Asserts that _op1 can be applied to _op2.
```
pratikvn, line 215: The documentation looks same as the macro above. Can you specify that this is for the outer dimension or something like that ?

include/ginkgo/core/matrix/diagonal.hpp
```diff
@@ -0,0 +1,191 @@
...
+                 public WritableToMatrixData<ValueType, int32>,
+                 public WritableToMatrixData<ValueType, int64> {
+    friend class Csr<ValueType, IndexType>;
+    friend class EnablePolymorphicObject<Diagonal, LinOp>;
+    friend class EnableCreateMethod<Diagonal>;
+
+public:
+    using value_type = ValueType;
+    using index_type = IndexType;
+    using mat_data = gko::matrix_data<ValueType, int64>;
```
pratikvn, line 80: I think `mat_data64` would be better

include/ginkgo/core/matrix/diagonal.hpp
```diff
@@ -0,0 +1,191 @@
...
+ * @ingroup diagonal
+ * @ingroup mat_formats
+ * @ingroup LinOp
+ */
+template <typename ValueType = default_precision, typename IndexType = int32>
+class Diagonal : public ConvertibleTo<Csr<ValueType, IndexType>>,
+                 public EnableLinOp<Diagonal<ValueType, IndexType>>,
+                 public EnableCreateMethod<Diagonal<ValueType, IndexType>>,
+                 public Transposable,
+                 public WritableToMatrixData<ValueType, int32>,
```
pratikvn, line 71: Any reason you dont implement the `ReadableFromMatrixData` interface ?

include/ginkgo/core/matrix/diagonal.hpp
fritzgoebel: I'd prefer leaving it as it is as this is consistent what is done in Dense

core/matrix/diagonal.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+
+namespace {
+
+
+template <typename MatrixType, typename MatrixData>
+inline void read_impl(MatrixType *mtx, const MatrixData &data)
+{
+    // Diagonal matrices are assumed to be square.
+    GKO_ASSERT_EQ(data.size[0], data.size[1]);
```
pratikvn, line 182: I think we have a [`GKO_ASSERT_IS_SQUARE_MATRIX`](https://github.com/ginkgo-project/ginkgo/blob/4df4dea1fbf142ddd3ee8aae9ff3f43c2a0829e1/core/factorization/ilu.cpp#L68) which throws `DimensionMismatch` which is better suited here. But maybe you have to add a `get_size()` method to `MatrixData`.

PR #581, opened 2020-07-02
Clean up sanitizers

upsj:
This PR attempts to get rid of most sanitizer warnings

* All sanitizers complain about the huge allocations we do in our executor tests
* UBSAN complains about empty memcpys on nullptrs
* TSAN complains about some write-after-write data races in the stopping conditions. This is not an issue, since we always write the same value, but it can be mitigated by making the for-loop a reduction
* TSAN complains about another write-after-write data race in bicgstab, where we do some unnecessarily duplicate work in parallel


upsj:
The PR looks to be ready to review, only the sanitizer flags are seemingly not passed to the executed tests. I will investigate this

upsj:
The result of this clean-up: Only the asynchronous iteration kernels (ParILU(T)/ParICT) are are complaining in TSAN, no further warnings except for the ones we deliberately produce in the sanitizer tests.

PR #583, opened 2020-07-06
Release/1.2.0

tcojean:
This merges all commits from develop into master for the release 1.2.0.

The documentation will be available here once updated:
https://ginkgo-project.github.io/ginkgo/doc/release/1.2.0/

We should see a  `based on branch master` in the title.


tcojean:
This should now be in a working state. The compiled documentation is available in the original post.

For some reason, a couple some commits got lost during the rebase so I had to manually fix the issues.

CMakeLists.txt
```diff
@@ -1,17 +1,23 @@
 cmake_minimum_required(VERSION 3.9)
 
-project(Ginkgo LANGUAGES C CXX VERSION 1.1.1 DESCRIPTION "A numerical linear algebra library targeting many-core architectures")
+project(Ginkgo LANGUAGES C CXX VERSION 1.2.0 DESCRIPTION "A numerical linear algebra library targeting many-core architectures")
```
thoasm, line 3: Will this automatically change `Ginkgo_VERSION_MAJOR` and so on?
Mainly asking to ensure that [`version_info::get_header_version()`](https://github.com/ginkgo-project/ginkgo/blob/develop/include/ginkgo/core/base/version.hpp#L216) returns the correct version.

thoasm:
After we are done with the release, I think we should increase the version in `develop` to `1.2.1`, so it is different from `master`.

CMakeLists.txt
tcojean: Yes that should work because we use `GKO_VERSION_MAJOR` etc in https://github.com/ginkgo-project/ginkgo/blob/daf1a4de99493914b16ea78f4fce014989d4e9fe/include/ginkgo/core/base/version.hpp#L216-L220 which is actually populated directly from CMake in the `config.hpp.in` file.

tcojean:
About the develop version, yes it would be nice but an issue is we do not know if we will have a `1.2.1` or if the next one will be `1.3.0`.

pratikvn:
Or we could name is `1.x.x-pre-release` or something like that.

yhmtsai:
I found we already set `Ginkgo_VERSION_TAG` master in `master` branch and develop in `develop` branch.
Maybe already solved it?

tcojean:
Yes actually we do have that `VERSION_TAG` distinction so that is exactly like @yhmtsai previously said, `1.2.0-develop` and `1.2.0-master`.

PR #585, opened 2020-07-07
Fix the CAS with HIP and NVIDIA backends.

tcojean:
Also include the CAS third party tool when building only HIP with CUDA backend.

When creating the spack package for Ginkgo which support HIP, I found out that we have an issue with the CAS dependency when only building HIP on an NVIDIA machine. Indeed, the CAS in required in `hip/CMakeLists.txt`, but in `third_party/CMakeLists.txt` we only include the CAS third party tool when building CUDA. This fixes that issue.

Building Ginkgo on a CUDA platform with HIP works when also building the CUDA backend (this is the case by default since we automatically detect the CUDA installation). Only in spack this is set the CUDA backend to `OFF` whenever a dependency to `CUDA` is not explicitly stated.


tcojean:
On a similar note, here is the spack package update for Ginkgo:
https://github.com/spack/spack/pull/17413

PR #586, opened 2020-07-07
Make criterion factory parameters work with CUDA

### LINKED ISSUE #491: Factory parameters and CUDA
We currently have a [workaround in place](https://github.com/ginkgo-project/ginkgo/blob/9435b453aa487fa74d2674d4275be2f76ac77716/include/ginkgo/core/base/lin_op.hpp#L874) that silently disables the parameter setup of `with_*` functions when compiling with `nvcc` or `hipcc`.
This might cause some unexpected issues (like [this CI job failure](https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/488830653)). I would like to propose a more robust workaround, but wanted to discuss the implications of it before:

```cpp
    auto with_##_name(decltype(_name) &&_value)
        const->const ::gko::xstd::decay_t<decltype(*this)> &
    {
        using type = decltype(this->_name);
        this->_name = type{std::move(_value)};
        return *this;
    }
```

This change would restrict the parameters to with_* functions to the actual parameter type, no longer allowing perfect forwarding to the parameter type's constructor.

As far as I can see, this is only relevant for the `with_criteria` function for all iterative solvers, which would change from 

```cpp
        .with_criteria(gko::stop::ResidualNormReduction<>::build()
                            .with_reduction_factor(FLAGS_rel_res_goal)
                            .on(exec),
                       gko::stop::Iteration::build()
                            .with_max_iters(FLAGS_max_iters)
                            .on(exec))
```
to

```cpp
        .with_criteria({gko::stop::ResidualNormReduction<>::build()
                            .with_reduction_factor(FLAGS_rel_res_goal)
                            .on(exec),
                        gko::stop::Iteration::build()
                            .with_max_iters(FLAGS_max_iters)
                            .on(exec)})
```
adding braces around the list of stopping criteria. What do you think? I guess the parameter type `decltype(_name) &&` might be a point of discussion.

upsj:
This PR adds new "overloads" `GKO_FACTORY_PARAMETER_SCALAR`/`_VECTOR` for types which can be initialized with a single value/multiple values which are aliased to by `GKO_FACTORY_PARAMETER` in C++ compilers.
It also adds a NotImplemented exception for nvcc using `GKO_FACTORY_PARAMETER`.

Fixes #491

### DIFF
--- a/include/ginkgo/core/base/lin_op.hpp
+++ b/include/ginkgo/core/base/lin_op.hpp
@@ -754,6 +754,9 @@ public:                                                                \
  *         // a factory parameter named "my_value", of type int and default
  *         // value of 5
  *         int GKO_FACTORY_PARAMETER_SCALAR(my_value, 5);
+ *         // a factory parameter named `my_pair` of type `std::pair<int,int>`
+ *         // and default value {5, 5}
+ *         std::pair<int, int> GKO_FACTORY_PARAMETER_VECTOR(my_pair, 5, 5);
  *     };
  *     // constructor needed by EnableLinOp
  *     explicit MyLinOp(std::shared_ptr<const Executor> exec) {
@@ -888,6 +891,7 @@ public:                                                                      \
     static_assert(true,                                                      \
                   "This assert is used to counter the false positive extra " \
                   "semi-colon warnings")
+
 /**
  * Creates a scalar factory parameter in the factory parameters structure.
  *
@@ -903,6 +907,7 @@ public:                                                                      \
  */
 #define GKO_FACTORY_PARAMETER_SCALAR(_name, _default) \
     GKO_FACTORY_PARAMETER(_name, _default)
+
 /**
  * Creates a vector factory parameter in the factory parameters structure.
  *
@@ -921,7 +926,8 @@ public:                                                                      \
 #else  // defined(__CUDACC__) || defined(__HIPCC__)
 // A workaround for the NVCC compiler - parameter pack expansion does not work
 // properly, because while the assignment to a scalar value is translated by
-// cudafe without parameter pack expansion, the corresponding parameter is not.
+// cudafe into a C-style cast, the parameter pack expansion is not removed and
+// `Args&&... args` is still kept as a parameter pack.
 #define GKO_FACTORY_PARAMETER(_name, ...)                                    \
     mutable _name{__VA_ARGS__};                                              \
                                                                              \
@@ -935,12 +941,13 @@ public:                                                                      \
     static_assert(true,                                                      \
                   "This assert is used to counter the false positive extra " \
                   "semi-colon warnings")
+
 #define GKO_FACTORY_PARAMETER_SCALAR(_name, _default)                        \
     mutable _name{_default};                                                 \
                                                                              \
     template <typename Arg>                                                  \
     auto with_##_name(Arg &&_value)                                          \
-        const->const ::gko::xstd::decay_t<decltype(*this)> &                 \
+        const->const std::decay_t<decltype(*this)> &                         \
     {                                                                        \
         using type = decltype(this->_name);                                  \
         this->_name = type{std::forward<Arg>(_value)};                       \
@@ -949,12 +956,13 @@ public:                                                                      \
     static_assert(true,                                                      \
                   "This assert is used to counter the false positive extra " \
                   "semi-colon warnings")
+
 #define GKO_FACTORY_PARAMETER_VECTOR(_name, ...)                             \
     mutable _name{__VA_ARGS__};                                              \
                                                                              \
     template <typename... Args>                                              \
     auto with_##_name(Args &&... _value)                                     \
-        const->const ::gko::xstd::decay_t<decltype(*this)> &                 \
+        const->const std::decay_t<decltype(*this)> &                         \
     {                                                                        \
         using type = decltype(this->_name);                                  \
         this->_name = type{std::forward<Args>(_value)...};                   \
--- a/include/ginkgo/core/base/lin_op.hpp
+++ b/include/ginkgo/core/base/lin_op.hpp
@@ -741,7 +741,7 @@ public:                                                                \
  *
  * The list of parameters for the factory should be defined in a code block
  * after the macro definition, and should contain a list of
- * GKO_FACTORY_PARAMETER declarations. The class should provide a constructor
+ * GKO_FACTORY_PARAMETER_* declarations. The class should provide a constructor
  * with signature
  * _lin_op(const _factory_name *, std::shared_ptr<const LinOp>)
  * which the factory will use a callback to construct the object.
@@ -753,7 +753,7 @@ public:                                                                \
  *     GKO_ENABLE_LIN_OP_FACTORY(MyLinOp, my_parameters, Factory) {
  *         // a factory parameter named "my_value", of type int and default
  *         // value of 5
- *         int GKO_FACTORY_PARAMETER(my_value, 5);
+ *         int GKO_FACTORY_PARAMETER_SCALAR(my_value, 5);
  *     };
  *     // constructor needed by EnableLinOp
  *     explicit MyLinOp(std::shared_ptr<const Executor> exec) {
@@ -787,8 +787,8 @@ public:                                                                \
  * std::cout << my_op->get_my_parameters().my_value;  // prints 0
  * ```
  *
- * @note It is possible to combine both the #GKO_CREATE_FACTORY_PARAMETER()
- * macro with this one in a unique macro for class __templates__ (not with
+ * @note It is possible to combine both the #GKO_CREATE_FACTORY_PARAMETER_*()
+ * macros with this one in a unique macro for class __templates__ (not with
  * regular classes). Splitting this into two distinct macros allows to use them
  * in all contexts. See <https://stackoverflow.com/q/50202718/9385966> for more
  * details.
@@ -870,6 +870,8 @@ public:                                                                      \
  *
  * @see GKO_ENABLE_LIN_OP_FACTORY for more details, and usage example
  *
+ * @deprecated Use GKO_FACTORY_PARAMETER_SCALAR or GKO_FACTORY_PARAMETER_VECTOR
+ *
  * @ingroup LinOp
  */
 #define GKO_FACTORY_PARAMETER(_name, ...)                                    \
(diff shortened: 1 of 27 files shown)


include/ginkgo/core/base/lin_op.hpp
```diff
@@ -855,49 +855,75 @@ public:                                                                      \
...
+    auto with_##_name(std::vector<_type> arg)                                \
+        const->const ::gko::xstd::decay_t<decltype(*this)> &                 \
+    {                                                                        \
+        this->_name = std::move(arg);                                        \
+        return *this;                                                        \
+    }                                                                        \
+    auto with_##_name(_type arg)                                             \
+        const->const ::gko::xstd::decay_t<decltype(*this)> &                 \
+    {                                                                        \
+        this->_name.push_back(std::move(arg));                               \
```
upsj, line 906: need to clear the vector first

upsj:
@yhmtsai Sure, let me know if it works.

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -855,49 +855,75 @@ public:                                                                      \
...
-// properly. You won't be able to use factories in code compiled with NVCC, but
-// at least this won't trigger a compiler error as soon as a header using it is
-// included. To not get a linker error, we provide a dummy body.
-#define GKO_FACTORY_PARAMETER(_name, ...)                                    \
-    mutable _name{__VA_ARGS__};                                              \
+
+
+/**
+ * Creates a factory parameter in the factory parameters structure that is a
+ * vector of multiple elements (at most 3).
```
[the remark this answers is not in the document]
tcojean: Maybe we should actually put back the recursive templated version with non-CUDA so that there is no change there and it works with any number of criteria. For CUDA, we can keep your new version so that it also works there.

thoasm:
Is it possible to still have the recursive version for the host compiler, and the fix you implemented for `nvcc`? That way, it is no longer interface breaking.
The problem is that we now can have up to 5 (not counting custom) stopping criterion pushed to master, meaning we would need to have at least 5 overloads (with 1-5 criterions) to not break potentially existing code.

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -855,49 +855,75 @@ public:                                                                      \
...
  * @param __VA_ARGS__  default value of the parameter
  *
  * @see GKO_ENABLE_LIN_OP_FACTORY for more details, and usage example
  *
  * @ingroup LinOp
  */
-#define GKO_FACTORY_PARAMETER(_name, ...)                                    \
-    mutable _name{__VA_ARGS__};                                              \
+#define GKO_FACTORY_PARAMETER(_name, _type, ...)                             \
+    _type mutable _name{__VA_ARGS__};                                        \
```
thoasm, line 870: This macro change definitively breaks the interface since it is defined in a public header and everyone who wants to add their own solver would use it. With this change, it would no longer compile.

tcojean: I have to agree with Thomas here, this is a definite interface break. I would rather that we go with a less potent version as suggested previously, so that we still fix this issue (for a set number of template parameters on CUDA) while not breaking interface.

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -753,7 +753,7 @@ public:                                                                \
  *     GKO_ENABLE_LIN_OP_FACTORY(MyLinOp, my_parameters, Factory) {
  *         // a factory parameter named "my_value", of type int and default
  *         // value of 5
- *         int GKO_FACTORY_PARAMETER(my_value, 5);
+ *         int GKO_FACTORY_PARAMETER_SCALAR(my_value, 5);
```
tcojean, line 756: Maybe add an example with `GKO_FACTORY_PARAMETER_VECTOR` to also showcase this one?

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -886,18 +888,76 @@ public:                                                                      \
...
+    }                                                                        \
+    static_assert(true,                                                      \
+                  "This assert is used to counter the false positive extra " \
+                  "semi-colon warnings")
+#define GKO_FACTORY_PARAMETER_VECTOR(_name, ...)                             \
+    mutable _name{__VA_ARGS__};                                              \
+                                                                             \
+    template <typename... Args>                                              \
+    auto with_##_name(Args &&... _value)                                     \
+        const->const ::gko::xstd::decay_t<decltype(*this)> &                 \
```
tcojean, line 957: Note, if for some reason this is merged after #604, you can update these to `std::decay_t`.

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -886,18 +888,76 @@ public:                                                                      \
...
+ */
+#define GKO_FACTORY_PARAMETER_VECTOR(_name, ...) \
+    GKO_FACTORY_PARAMETER(_name, __VA_ARGS__)
 #else  // defined(__CUDACC__) || defined(__HIPCC__)
 // A workaround for the NVCC compiler - parameter pack expansion does not work
-// properly. You won't be able to use factories in code compiled with NVCC, but
-// at least this won't trigger a compiler error as soon as a header using it is
-// included. To not get a linker error, we provide a dummy body.
+// properly, because while the assignment to a scalar value is translated by
+// cudafe without parameter pack expansion, the corresponding parameter is not.
```
tcojean, line 924: Nice find!

include/ginkgo/core/preconditioner/jacobi.hpp
```diff
@@ -338,7 +338,8 @@ class Jacobi : public EnableLinOp<Jacobi<ValueType, IndexType>>,
          *       has to be respected when setting this parameter. Failure to do
          *       so will lead to undefined behavior.
          */
-        gko::Array<index_type> GKO_FACTORY_PARAMETER(block_pointers, nullptr);
+        gko::Array<index_type> GKO_FACTORY_PARAMETER_SCALAR(block_pointers,
```
tcojean, line 341: Shouldn't this be considered a `_VECTOR`?

include/ginkgo/core/preconditioner/jacobi.hpp
```diff
@@ -435,7 +436,7 @@ class Jacobi : public EnableLinOp<Jacobi<ValueType, IndexType>>,
          * If the non-adaptive version of Jacobi is used, the
          * `storage_optimization.block_wise` Array will be empty.
          */
-        storage_optimization_type GKO_FACTORY_PARAMETER(
+        storage_optimization_type GKO_FACTORY_PARAMETER_SCALAR(
```
tcojean, line 439: This parameter is complex, it can be either a scalar or an array actually...

include/ginkgo/core/base/lin_op.hpp
upsj: Good point!

include/ginkgo/core/base/lin_op.hpp
upsj: Thanks for noticing, it most likely will be :)

include/ginkgo/core/preconditioner/jacobi.hpp
upsj: You are probably right. We only use it as a `_SCALAR` throughout our code base, but I believe that the bug actually manifests itself based on the type of the parameter (primitive vs. composite with multiple-element constructor), so `_VECTOR` should be more appropriate.

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -886,18 +888,76 @@ public:                                                                      \
...
+    }                                                                        \
+    static_assert(true,                                                      \
+                  "This assert is used to counter the false positive extra " \
+                  "semi-colon warnings")
+#define GKO_FACTORY_PARAMETER_SCALAR(_name, _default)                        \
+    mutable _name{_default};                                                 \
+                                                                             \
+    template <typename Arg>                                                  \
+    auto with_##_name(Arg &&_value)                                          \
+        const->const ::gko::xstd::decay_t<decltype(*this)> &                 \
```
upsj, line 943:
```suggestion
        const->const std::decay_t<decltype(*this)> &                 \
```

include/ginkgo/core/base/lin_op.hpp
upsj: Actually, I need to correct this one, it is not 100% accurate:
```suggestion
// cudafe into a C-style cast, the parameter pack expansion is not removed and
// `Args&&... args` is still kept as a parameter pack.
```

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -886,18 +888,76 @@ public:                                                                      \
...
+                  "semi-colon warnings")
+#define GKO_FACTORY_PARAMETER_VECTOR(_name, ...)                             \
+    mutable _name{__VA_ARGS__};                                              \
+                                                                             \
+    template <typename... Args>                                              \
+    auto with_##_name(Args &&... _value)                                     \
+        const->const ::gko::xstd::decay_t<decltype(*this)> &                 \
+    {                                                                        \
+        using type = decltype(this->_name);                                  \
+        this->_name = type{std::forward<Args>(_value)...};                   \
```
yhmtsai, line 960: `std::vector<shared_ptr> GKO_FACTORY_PARAMETER(a, nullptr)` is still be std::vector{nullptr} not nullptr in hipcc/nvcc/..., right?

include/ginkgo/core/base/lin_op.hpp
yhmtsai: I am confused with `std::pair` example.
I feel std::pair is a scalar containing two elements like complex type.
maybe like `std::pair<int, int> GKO_FACTORY_PARAMETER_SCALAR(my_value, make_pair(5, 5))`?

include/ginkgo/core/base/lin_op.hpp
upsj: Yes

include/ginkgo/core/base/lin_op.hpp
upsj: The difference between _SCALAR and _VECTOR comes from how cudafe++ translates the files. I'm not entirely sure about the formal criteria, but for `int` and other primitive types, the parameter pack expansion gets translated incorrectly (Into a C-style cast instead of a constructor call), while for composite types like `std::pair, std::vector, ...` it correctly invokes the constructor. The semantics of this are a bit fuzzy, though, that is true. You could also use SCALAR, but then you would have to use make_pair or braces { ... } for the default argument or with_my_value.

PR #587, opened 2020-07-08
Remove unused preconditioner parameter in LowerTrs

upsj:
When going through the factory parameters in Ginkgo, I noticed that the LowerTrs factory has a preconditioner parameter that is never used anywhere. This PR removes the unused parameter. Is this considered interface-breaking?


pratikvn:
Hm, Thanks for removing it. That should not have been there. But I think technically it is interface breaking, even though it was not used.

PR #589, opened 2020-07-13
Fix sellp allocator

yhmtsai:
This PR fixes the sellp allocator of `read` function to avoid allocate the vector on the device executor


upsj:
Do you know why we didn't have this issue sooner? Did we never try to benchmark SELL-P SpMV on the GPU before?

yhmtsai:
I did the spmv benchmark long time ago. And we usually use coo or csr to do the other benchmark.

upsj:
Ah I see, it was actually me who introduced that bug with the ExecutorAllocator PR. Thanks for finding this!

PR #593, opened 2020-07-17
Fix reduce_add, prefix_sum, and doc-build

yhmtsai:
This PR fixes some issue in `reduce_add` and `prefix_sum`.

1. `reduce_add` use deallocated memory if `size >  default_block_size` because we declare the array inside the if-block but use data pointer out of the if-block.
2. `prefix_sum` does not need to read the last element of the array (initcheck reports it) and compute the total sum of the last block.
Also, use `components::prefix_sum` not `start_prefix_sum` in `sampleselect_count`. otherwise, it gets the wrong result from this fix because the last block_sum will not include the last element.
3. `start_prefix_sum` uses `size_type` as shared_memory for `index_type` or `size_type`, which leads wrong result if some values are negative. This fix also makes it correct on floating point type.

minor fix: fix the doc-build


cuda/components/prefix_sum.cu
```diff
@@ -49,15 +49,21 @@ template <typename IndexType>
 void prefix_sum(std::shared_ptr<const CudaExecutor> exec, IndexType *counts,
                 size_type num_entries)
 {
+    // prefix_sum should be on the valid array
+    GKO_ASSERT(num_entries > 0);
```
upsj, line 53: How about just using
```suggestion
    if (num_entries > 0) {
```
This would make the behavior between cuda/hip and reference/omp match more closely, since there is nothing *inherently* wrong with computing the prefix sum of an empty array.

.gitlab-ci.yml
```diff
@@ -859,6 +859,27 @@ cudamemcheck:
     - cuda
     - gpu
 
+cudainitcheck:
+  stage: QoS_tools
+  image: localhost:5000/gko-cuda101-gnu8-llvm10-intel2019
+  before_script: *default_before_script
+  script:
+    - ctest -V -S cmake/CTestScript.cmake -DCTEST_BUILD_CONFIGURATION=RelWithDebInfo
+      -DCTEST_MEMORYCHECK_TYPE=CudaMemcheck -DCTEST_MEMORYCHECK_COMMAND_OPTIONS=--tool;initcheck
```
upsj, line 868:
```suggestion
      -DCTEST_MEMORYCHECK_TYPE=CudaMemcheck "-DCTEST_MEMORYCHECK_COMMAND_OPTIONS=--tool;initcheck"
```
Sorry, I made that mistake before and forgot to tell you about it :-D

.gitlab-ci.yml
```diff
@@ -859,6 +859,27 @@ cudamemcheck:
     - cuda
     - gpu
 
+cudainitcheck:
+  stage: QoS_tools
+  image: localhost:5000/gko-cuda101-gnu8-llvm10-intel2019
+  before_script: *default_before_script
+  script:
+    - ctest -V -S cmake/CTestScript.cmake -DCTEST_BUILD_CONFIGURATION=RelWithDebInfo
+      -DCTEST_MEMORYCHECK_TYPE=CudaMemcheck "-DCTEST_MEMORYCHECK_COMMAND_OPTIONS=--tool;initcheck"
```
upsj, line 868: Sorry, messed this up again, it should be
```suggestion
      -DCTEST_MEMORYCHECK_TYPE=CudaMemcheck "-DCTEST_MEMORYCHECK_COMMAND_OPTIONS=--tool initcheck"
```

common/factorization/par_ilut_select_kernels.hpp.inc
```diff
@@ -305,4 +305,4 @@ __global__ __launch_bounds__(config::warp_size) void find_bucket(
 }
 
 
-}  // namespace kernel
\ No newline at end of file
+}  // namespace kernel
```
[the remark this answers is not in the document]
upsj: I think this is more of an editor issue: Some editors always insert a newline at the end, some leave this out. We may want to enforce the newline at the end.

pratikvn:
Currently there is a problem with the documentation as the GLOB picks up the build-setup.sh as well. I think if you add build-setup.sh to this [line in doc/examples/CMakeLists.txt](https://github.com/ginkgo-project/ginkgo/blob/develop/doc/examples/CMakeLists.txt#L5), removing it, everything should work again.

common/components/prefix_sum.hpp.inc
```diff
@@ -130,8 +131,10 @@ __global__ __launch_bounds__(block_size) void start_prefix_sum(
     }
 
     if (element_id == 0) {
-        // Store the total sum
-        block_sum[blockIdx.x] = prefix_helper[block_size - 1];
+        // Store the total sum except the last block
+        if (blockIdx.x + 1 < gridDim.x) {
+            block_sum[blockIdx.x] = prefix_helper[block_size - 1];
+        }
```
thoasm, line 137: Could you also change the documentation of `block_sum` to have at least `ceildiv(num_elements, block_size) - 1` since you never access `block_size - 1`.

doc/examples/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@
 FILE(GLOB _ginkgo_examples
     ${CMAKE_SOURCE_DIR}/examples/*
     )
-LIST(REMOVE_ITEM _ginkgo_examples "${CMAKE_SOURCE_DIR}/examples/CMakeLists.txt")
+LIST(REMOVE_ITEM _ginkgo_examples "${CMAKE_SOURCE_DIR}/examples/CMakeLists.txt" "build-setup.sh")
```
upsj, line 5:
```suggestion
LIST(REMOVE_ITEM _ginkgo_examples "${CMAKE_SOURCE_DIR}/examples/CMakeLists.txt" "${CMAKE_SOURCE_DIR}/examples/build-setup.sh")
```

PR #603, opened 2020-07-30
Cuda11 support

### LINKED ISSUE #613: [JOSS REVIEW] Use of deprecated function `cusparseScsrmv ` in CUDA executor
This issue is part of the functionality aspect of a JOSS review (see #597 )

I have run into a new issue in building CUDA-enabled ginkgo while building.  After a successful `cmake`, an attempt to build with `make` gives
```
[  1%] Building CXX object cuda/CMakeFiles/ginkgo_cuda.dir/base/executor.cpp.o
In file included from /home/ajacobs/Reporoot/ginkgo/cuda/base/executor.cpp:43:
/home/ajacobs/Reporoot/ginkgo/cuda/base/cusparse_bindings.hpp: In function ‘void gko::kernels::cuda::cusparse::spmv(cusparseHandle_t, cusparseOperation_t, gko::int32, gko::int32, gko::int32, const float*, cusparseMatDescr_t, const float*, const int32*, const int32*, const float*, const float*, float*)’:
/home/ajacobs/Reporoot/ginkgo/cuda/base/cusparse_bindings.hpp:118:33: error: ‘cusparseScsrmv’ was not declared in this scope; did you mean ‘cusparseSbsrmv’?
  118 | GKO_BIND_CUSPARSE32_SPMV(float, cusparseScsrmv);
``` 

I deploy the latest stable CUDA tookit and dev environment, so when I looked at my cusparse header files I confirmed that `cusparseScsrmv` is no longer provided and was deprecated as mentioned [here (section 8.8.)](https://docs.nvidia.com/cuda/archive/10.1/pdf/CUSPARSE_Library.pdf).

NVIDIA suggests "use cusparseCsrmvEx() instead."

In the meantime, I will explore workarounds (and welcome suggestions).  As CUDA toolkits are made current, however, this will become an issue for more users/systems.

yhmtsai:
this PR add cuda11 support.

Summary:
1. add generic spmv/spmm into cusparse spmv/spmm
2. use cusp_coo/cusp_csr as cusp_gcoo/cusp_gcsr in benchmark/spmv 
3. use correct architecture flag in cuda/test
4. use cuda11's cooperative group because it support complex shuffle, but we remove the parent type in `tiled_partition`
5. add cuda 10.2/11 gitlab ci (linux) and cuda11 github action (windows)
6. replace deprecated spgemm2 and gthr by new cusparse generic interface.

Fixes #613

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -400,6 +400,126 @@ build/clang-cuda101/clang/cuda/debug/static:
     - cuda
     - gpu
 
+# cuda 10.2 and friends
+build/cuda102/gcc/all/debug/shared:
+  <<: *default_build_with_test
+  image: localhost:5000/gko-cuda102-gnu8-llvm8-intel2019
+  variables:
+    <<: *default_variables
+    BUILD_OMP: "ON"
+    BUILD_CUDA: "ON"
+    BUILD_HIP: "ON"
+    BUILD_TYPE: "Debug"
+    CUDA_ARCH: 35
+  only:
+    variables:
+      - $RUN_CI_TAG
+  tags:
+    - private_ci
+    - cuda
+    - gpu
+
+build/cuda102/clang/all/release/static:
+  <<: *default_build_with_test
+  image: localhost:5000/gko-cuda102-gnu8-llvm8-intel2019
+  variables:
+    <<: *default_variables
+    C_COMPILER: "clang"
+    CXX_COMPILER: "clang++"
+    BUILD_OMP: "ON"
+    BUILD_CUDA: "ON"
+    BUILD_HIP: "ON"
+    BUILD_TYPE: "Release"
+    BUILD_SHARED_LIBS: "OFF"
+    CUDA_ARCH: 35
+  only:
+    variables:
+      - $RUN_CI_TAG
+  tags:
+    - private_ci
+    - cuda
+    - gpu
+
+build/cuda102/intel/cuda/debug/static:
+  <<: *default_build_with_test
+  image: localhost:5000/gko-cuda102-gnu8-llvm8-intel2019
+  variables:
+    <<: *default_variables
+    C_COMPILER: "icc"
+    CXX_COMPILER: "icpc"
+    BUILD_OMP: "ON"
+    BUILD_CUDA: "ON"
+    BUILD_TYPE: "Debug"
+    BUILD_SHARED_LIBS: "OFF"
+    CUDA_ARCH: 35
+  only:
+    variables:
+      - $RUN_CI_TAG
+  tags:
+    - private_ci
+    - cuda
+    - gpu
+
+# cuda 11.0 and friends
+build/cuda110/gcc/all/debug/shared:
+  <<: *default_build_with_test
+  image: localhost:5000/gko-cuda110-gnu9-llvm9-intel2020
+  variables:
+    <<: *default_variables
+    BUILD_OMP: "ON"
+    BUILD_CUDA: "ON"
+    BUILD_HIP: "OFF"
+    BUILD_TYPE: "Debug"
+    CUDA_ARCH: 35
+  only:
+    variables:
+      - $RUN_CI_TAG
+  tags:
+    - private_ci
+    - cuda
+    - gpu
+
+build/cuda110/clang/all/release/static:
+  <<: *default_build_with_test
+  image: localhost:5000/gko-cuda110-gnu9-llvm9-intel2020
+  variables:
+    <<: *default_variables
+    C_COMPILER: "clang"
+    CXX_COMPILER: "clang++"
+    BUILD_OMP: "ON"
+    BUILD_CUDA: "ON"
+    BUILD_HIP: "OFF"
+    BUILD_TYPE: "Release"
+    BUILD_SHARED_LIBS: "OFF"
+    CUDA_ARCH: 35
+  only:
+    variables:
+      - $RUN_CI_TAG
+  tags:
+    - private_ci
+    - cuda
+    - gpu
+
+build/cuda110/intel/cuda/debug/static:
+  <<: *default_build_with_test
+  image: localhost:5000/gko-cuda110-gnu9-llvm9-intel2020
+  variables:
+    <<: *default_variables
+    C_COMPILER: "icc"
+    CXX_COMPILER: "icpc"
+    BUILD_OMP: "ON"
+    BUILD_CUDA: "ON"
+    BUILD_TYPE: "Debug"
+    BUILD_SHARED_LIBS: "OFF"
+    CUDA_ARCH: 35
+  only:
+    variables:
+      - $RUN_CI_TAG
+  tags:
+    - private_ci
+    - cuda
+    - gpu
+
 # HIP AMD
 build/amd/gcc/hip/debug/shared:
   <<: *default_build_with_test
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -339,26 +339,6 @@ build/cuda101/clang/all/release/static:
     - cuda
     - gpu
 
-build/cuda101/intel/cuda/debug/static:
-  <<: *default_build_with_test
-  image: localhost:5000/gko-cuda101-gnu8-llvm7-intel2019
-  variables:
-    <<: *default_variables
-    C_COMPILER: "icc"
-    CXX_COMPILER: "icpc"
-    BUILD_OMP: "ON"
-    BUILD_CUDA: "ON"
-    BUILD_TYPE: "Debug"
-    BUILD_SHARED_LIBS: "OFF"
-    CUDA_ARCH: 35
-  only:
-    variables:
-      - $RUN_CI_TAG
-  tags:
-    - private_ci
-    - cuda
-    - gpu
-
 # clang-cuda with cuda 10.1 and friends
 build/clang-cuda101/gcc/all/release/shared:
   <<: *default_build_with_test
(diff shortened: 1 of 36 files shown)


cuda/base/cusparse_bindings.hpp
```diff
@@ -629,6 +704,49 @@ inline void destroy(csrgemm2Info_t info)
 }
 
 
+#if defined(CUDA_VERSION) && (CUDA_VERSION >= 11000)
+
+
+inline cusparseDnVecDescr_t create_dnvec(int64_t size, void *values,
```
upsj, line 710: Suggestion: How about replacing void* values by a template parameter and inferring `valueType = cuda_data_type<ValueType>`? Same for create_csr

cuda/matrix/csr_kernels.cu
```diff
@@ -307,23 +307,56 @@ void spmv(std::shared_ptr<const CudaExecutor> exec,
...
+
+                cusparse::destroy(descr);
+#else  // CUDA_VERSION >= 11000
+                cusparseOperation_t trans = CUSPARSE_OPERATION_NON_TRANSPOSE;
+                cusparseSpMVAlg_t alg = CUSPARSE_CSRMV_ALG1;
+                auto row_ptrs =
+                    const_cast<IndexType *>(a->get_const_row_ptrs());
+                auto col_idxs =
+                    const_cast<IndexType *>(a->get_const_col_idxs());
+                auto values = const_cast<ValueType *>(a->get_const_values());
```
tcojean, line 336: Any idea why they take non const? That's a bit weird, do they want to change these somehow?

cuda/matrix/csr_kernels.cu
```diff
@@ -895,7 +1105,31 @@ void conj_transpose(std::shared_ptr<const CudaExecutor> exec,
...
             orig->get_const_values(), orig->get_const_row_ptrs(),
             orig->get_const_col_idxs(), trans->get_values(),
-            trans->get_col_idxs(), trans->get_row_ptrs(), copyValues, idxBase);
+            trans->get_row_ptrs(), trans->get_col_idxs(), copyValues, idxBase);
+#else  // CUDA_VERSION >= 11000
+        cudaDataType_t cu_value =
+            gko::kernels::cuda::cuda_data_type<ValueType>();
+        cusparseAction_t copyValues = CUSPARSE_ACTION_NUMERIC;
+        cusparseIndexBase_t idxBase = CUSPARSE_INDEX_BASE_ZERO;
+        cusparseCsr2CscAlg_t alg = CUSPARSE_CSR2CSC_ALG2;
```
upsj, line 1114: Important: We should always keep this at ALG1, since ALG2 doesn't guarantee any ordering of the columns. This could silently break a lot of algorithms! Can you maybe add a test for Csr::transpose (if we don't already have one) that tests that `is_sorted_by_column_index()` is true?
```suggestion
        cusparseCsr2CscAlg_t alg = CUSPARSE_CSR2CSC_ALG1;
```

cuda/matrix/csr_kernels.cu
upsj: They don't care about const correctness and want to allow themselves to reorder the columns if necessary, I would guess.

cuda/base/cusparse_bindings.hpp
```diff
@@ -226,6 +265,8 @@ GKO_BIND_CUSPARSE64_SPMM(ValueType, detail::not_implemented);
 #undef GKO_BIND_CUSPARSE32_SPMM
 #undef GKO_BIND_CUSPARSE64_SPMM
 
+#endif
```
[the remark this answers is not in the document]
yhmtsai: We currently put the function which does similar thing together.
It needs many `#if`-block but easily identify which function has different version.
Another way is to create a global `#if` to handle these function.
It can reduce the number of `#if`-block but maybe gives very complicated structure.
For example, we need to handle several version of benchmark/spmv.
1. old coo/csr/ell/hybrid is available in 9 ~ 10.2
2. generic coo/csr/csr is available after 10.2 in non-Windows or 11 in all platform.
3. coo/csr is alias for gcoo/gcsr after 11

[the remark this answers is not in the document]
tcojean: I mean, the previous one in this case has the same definition I think, so it wouldn't hurt to simply merge it with the one above. You do not have to change the file structure.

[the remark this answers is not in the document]
yhmtsai: I see

.gitlab-ci.yml
```diff
@@ -397,6 +397,126 @@ build/clang-cuda101/clang/cuda/debug/static:
...
+    BUILD_TYPE: "Debug"
+    BUILD_SHARED_LIBS: "OFF"
+    CUDA_ARCH: 35
+  only:
+    variables:
+      - $RUN_CI_TAG
+  tags:
+    - private_ci
+    - cuda
+    - gpu
```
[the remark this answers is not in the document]
yhmtsai: I remove the cuda 10.1 corresponding one because cuda 10.1 has additional clang-cuda test.

cuda/base/cusparse_bindings.hpp
```diff
@@ -504,101 +623,119 @@ GKO_BIND_CUSPARSE64_CSR2HYB(ValueType, detail::not_implemented);
...
+        const ValueType *OrigValA, const int32 *OrigRowPtrA,                  \
+        const int32 *OrigColIndA, ValueType *TransValA, int32 *TransRowPtrA,  \
+        int32 *TransColIndA, cusparseAction_t copyValues,                     \
+        cusparseIndexBase_t idxBase)                                          \
     {                                                                         \
         GKO_ASSERT_NO_CUSPARSE_ERRORS(                                        \
             CusparseName(handle, m, n, nnz, as_culibs_type(OrigValA),         \
                          OrigRowPtrA, OrigColIndA, as_culibs_type(TransValA), \
-                         TransRowPtrA, TransColIndA, copyValues, idxBase));   \
+                         TransColIndA, TransRowPtrA, copyValues, idxBase));   \
```
pratikvn, line 657: Why did this change ?

cuda/base/cusparse_handle.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+namespace gko {
+namespace kernels {
+namespace cuda {
+/**
+ * @brief The CUSPARSE namespace.
+ *
+ * @ingroup cusparse
+ */
+namespace cusparse {
```
pratikvn, line 52: Is there any specific reason you do this for cusparse and not for cublas ?

yhmtsai:
I also add the comments on `#endif` when the block only contains `#if-#endif`.
```c
#if condition
#endif // condition
```

cuda/matrix/csr_kernels.cu
```diff
@@ -868,7 +1053,31 @@ void transpose(std::shared_ptr<const CudaExecutor> exec,
             orig->get_size()[1], orig->get_num_stored_elements(),
             orig->get_const_values(), orig->get_const_row_ptrs(),
             orig->get_const_col_idxs(), trans->get_values(),
-            trans->get_col_idxs(), trans->get_row_ptrs(), copyValues, idxBase);
+            trans->get_row_ptrs(), trans->get_col_idxs(), copyValues, idxBase);
```
yhmtsai, line 1056: I also change the ordering here

cuda/base/cusparse_bindings.hpp
yhmtsai: cusparseCsr2csc uses csc by (CscVal, CscRowInd, CscColPtr) - (1)
cusparseCsr2cscEx2 uses csc by (CscVal, CscColPtr, CscRowInd) - (2)
We use it mainly for transposing, so I think using transposed_csr by (transVal, transRowPtr, transColInd), i.e. (2),should be much familiar and not confusing.

cuda/base/cusparse_handle.hpp
yhmtsai: maybe @upsj could answer it?

upsj: Yes, cusparse_bindings.hpp doesn't compile on plain C++ compilers because the NVCC version macros are not defined. So I had to separate these parts.

pratikvn: For uniformity, we could do the same for cublas as well ?

upsj: I don't think that's necessary (yet), since the interfaces are always the same.

cuda/matrix/csr_kernels.cu
```diff
@@ -385,6 +417,34 @@ void advanced_spmv(std::shared_ptr<const CudaExecutor> exec,
...
+            auto mat = cusparse::create_csr(a->get_size()[0], a->get_size()[1],
+                                            a->get_num_stored_elements(),
+                                            row_ptrs, col_idxs, values);
+            auto b_val = const_cast<ValueType *>(b->get_const_values());
+            auto c_val = c->get_values();
+            auto vecb =
+                cusparse::create_dnvec(b->get_num_stored_elements(), b_val);
+            auto vecc =
+                cusparse::create_dnvec(c->get_num_stored_elements(), c_val);
+            size_t buffer_size = 0;
```
thoasm, line 435: nit: Please either use `std::size_t` or `size_type`.
```suggestion
            std::size_t buffer_size = 0;
```

cuda/base/cusparse_handle.hpp
pratikvn: Ok

cuda/base/cusparse_bindings.hpp
yhmtsai: also another point is original we use transRowPtr as ColInd, which is wrong.

PR #607, opened 2020-07-31
Warn when CXX and CUDA host compiler do not match.

### LINKED ISSUE #605: Add a warning for mismatching CUDA host compiler and CXX version
There seems to be an incompatibility between GCC 10's glib version and GCC 8 which is typically used for CUDA host compilation. Somehow, this does not happen on all systems but could happen. In arch linux this always happen, in Ubuntu Thomas was able to make it happen on his computer, though in general it should work.

To help people identify this issue if they ever run into it, we can add a warning message whenever these compiler versions mismatch, as in general it would seem more logical to use the same CXX and CUDA host compiler version everywhere.

tcojean:
On some systems, this can cause problems if the two compilers are tied
to different glib versions. At linking time (esp. for examples or
tests), it is then possible that some symbols are not found.

Closes #605

### DIFF
--- a/cuda/CMakeLists.txt
+++ b/cuda/CMakeLists.txt
@@ -122,6 +122,15 @@ if (NOT CMAKE_CUDA_HOST_COMPILER AND NOT GINKGO_CUDA_DEFAULT_HOST_COMPILER)
 elseif(GINKGO_CUDA_DEFAULT_HOST_COMPILER)
     unset(CMAKE_CUDA_HOST_COMPILER CACHE)
 endif()
+
+if(CMAKE_CUDA_HOST_COMPILER AND NOT CMAKE_CXX_COMPILER STREQUAL CMAKE_CUDA_HOST_COMPILER)
+    message(WARNING "The CMake CXX compiler and CUDA host compiler do not match. "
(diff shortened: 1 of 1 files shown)


thoasm:
@upsj is correct, if `CUDA_HOST_COMPILER` is not set manually, it is not available here. I get the following output if I specify nothing (while setting `CXX=gcc-10` during `cmake`):
```
  The CXX compiler is /usr/local/bin/gcc-10 with version 10.2.1.

  The CUDA host compiler is .
```

tcojean:
Interesting, thanks for the info. I had not checked without passing the variable as my system requires this.

tcojean:
I cannot really find a way to identify which compiler `nvcc` will use by default, so if for some reason it is not set (at least when setting `GINKGO_CUDA_DEFAULT_HOST_COMPILER`) then I do not print anything.

tcojean:
@thoasm 
> I could not find a way to get the cuda host compiler either.
> Having this is definitively better than not having.

For reference, I found that you can find the host compiler by compiling some source code with `nvcc` and using the flag `--verbose`. Though it doesn't seem practical to extract the host compiler information from there (you need ot recognize the host compilation line, extract the command out of it, etc...).

PR #611, opened 2020-08-03
Use cxx14 standard

### LINKED ISSUE #604: Compile Ginkgo with C++14
It seems that thrust will be deprecating at a later point C++11 and force compilation with C++14. To prepare for this, we could consider moving Ginkgo to use the C++14 standard.

I looked it up, and all compilers we use on paper also support C++14, though we need testing to ensure that everything works.

There would be benefits to updating the C++ standard version in order to simplify the implementation of some of our extended standard header.

tcojean:
Upgrade Ginkgo from C++11 standard to C++14.

There are multiple reasons for this:
1. CUDA's thrust will be deprecating C++11 and moving to C++14 shortly. 
2. On paper and in practice, all the compiler versions we support already support C++14, so moving doesn't make us lose anything. The only failure is a corner case of CUDA 9.2 and clang 5.0 as CUDA host compiler which cannot be fixed, see the comments in the related PR on this issue.

What changed:
+ Upgrade any mention of C++11 to C++14
+ Review the `std_extensions` to use implementation from C++14 instead of the custom made one.
+ Update all code calls from `xstd` to `std` when the feature used was a C++14 feature.
+ Use Intel 2020 together with GCC 9 as there is an incompatibility when considering C++14 between Intel 2019 and GCC 9.
+ Ensure tests are always compiled with the C++14 standard
+ Throw a fatal_error when using CUDA 9.2 with clang 5.0 as CUDA host compiler.

Closes #604


tcojean:
Everything is working except for cuda 9.2 + clang 5.0 as host compiler. For some reason, there is an issue interpreting the standard headers and I cannot find the reason. See this pipeline: https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/pipelines/173724849

For some reason, a small example such as the following compiled with `nvcc -std=c++14 -ccbin=/usr/bin/clang++ test.cu` will fail, whereas `nvcc -std=c++14 -ccbin=/usr/bin/g++ test.cu` succeeds.
```c++
#include <complex>
int main() {
        return 0;
}
```

tcojean:
The issue is clearly that `nvcc` for some reason adds spaces between `operator` and `""` and more importantly `""` and `if` (etc). The operator is defined as `operator""if` without space (the space has its importance it seems, see https://en.cppreference.com/w/cpp/numeric/complex/operator%22%22i and https://en.cppreference.com/w/cpp/language/user_literal (especially the example down there). Clearly, `clang` is correct in saying that there is an error here, the issue is why does `nvcc 9.2` transform the perfectly valid header into 
```c++
constexpr complex< float>  operator "" if(long double __num) 
```

upsj:
This is a bug in the clang tokenization/parser: It consumes the "if" token as the if keyword instead of an identifier. Removing the space or changing the identifier to be a non-keyword fixes it.
Output of `clang++ --std=c++14 -fsyntax-only -Xclang -dump-tokens`:

```
string_literal '""'	 [LeadingSpace]	Loc=<test.cpp:1:33>
if 'if'	 [LeadingSpace]	Loc=<test.cpp:1:36>
```

tcojean:
For some reason, this example only produces a warning though, so the mystery isn't fully solved yet...
```c++
constexpr double operator "" i(unsigned long long d)
{
        return 0.0;
}

int main()
{
        return 0;
}
```

```bash
root@0d12ab8c785c:/# clang++ -std=c++14 test.cpp
test.cpp:1:18: warning: user-defined literal suffixes not starting with '_' are reserved [-Wuser-defined-literals]
constexpr double operator "" i(unsigned long long d)
                 ^
1 warning generated.
```

upsj:
Okay, I finished digging, this is a bug in cudafe++: `--gnu_version=70500` gives the correct translation, `--clang --clang_version=50000` inserts the space. I guess we can't do anything about that.

upsj:
Regarding the space: The standard is a bit unclear about the details, but as the [Literal operators](https://en.cppreference.com/w/cpp/language/user_literal#Literal_operators) section of the page you linked states, the *user-defined-string-literal* token was introduced to allow keywords to be used as literal suffixes, so I believe clang isn't at fault here. gcc emits a similar error message:
`unexpected keyword; remove space between quotes and suffix identifier`

tcojean:
Yes, I think `clang` is somewhat correct in the end, despite the bug being a weird parsing issue. Overall, this shouldn't be valid C++. My understanding is that having a space at this place `operator"" if` means that it's a `user defined literal`, which should always start with a underscore.

tcojean:
Tobias helped find a possible way to work around this issue, from the previous `nvcc` error example:
```c++
#include <complex>
int main() {
        return 0;
}
```

You can compile it successfully with the following arguments:
```bash
nvcc -ccbin=/usr/bin/clang++ -Xcudafe --gnu_version=70500 -Xcudafe --no_clang -std=c++14 test.cu
```

.gitlab-ci.yml
```diff
@@ -488,7 +467,7 @@ build/nocuda/clang/core/release/shared:
 
 build/nocuda/intel/core/debug/shared:
   <<: *default_build_with_test
-  image: localhost:5000/gko-nocuda-gnu9-llvm8
+  image: localhost:5000/add_cuda11_container_1b1857a4_gko-nocuda-gnu9-llvm8
```
upsj, line 470: Question: Are these changes necessary or left-over from the cuda11 branch?

.gitlab-ci.yml
```diff
@@ -536,7 +515,7 @@ build/nocuda/clang/omp/debug/static:
 
 build/nocuda/intel/omp/release/static:
   <<: *default_build_with_test
-  image: localhost:5000/gko-nocuda-gnu9-llvm8
+  image: localhost:5000/add_cuda11_container_1b1857a4_gko-nocuda-gnu9-llvm8
```
upsj, line 518: Same here

README.md
```diff
@@ -10,11 +10,11 @@
 [![CDash dashboard](https://img.shields.io/badge/CDash-Access-blue.svg)](http://my.cdash.org/index.php?project=Ginkgo+Project)
```
upsj, line 10: Unrelated, but maybe important: 
```suggestion
[![CDash dashboard](https://img.shields.io/badge/CDash-Access-blue.svg)](https://my.cdash.org/index.php?project=Ginkgo+Project)
```
[Pure HTTP](http://my.cdash.org/index.php?project=Ginkgo+Project) seems to be broken

tcojean:
The previous fix mentioned works for the `complex` header, but sadly it creates a lot of issues as soon as we include the `<memory>` header, so I went with the solution of throwing an error when using clang 5.0 as CUDA host compiler with CUDA 9.2.

.gitlab-ci.yml
tcojean: They are necessary, I had to upgrade the intel compiler to 2020 for this PR to work properly. I used the CUDA 11 PR to update this at the same time. The Intel 2019 has some incompatibility with GNU 9 backend when considering C++14 code. Therefore we will need to merge the CUDA11 container before this.

README.md
tcojean: Nice find

PR #612, opened 2020-08-03
Fix norm-type in the examples

### LINKED ISSUE #606: [Error] inverse-iteration
Hello everyone!

When I ran the inverse-iteration example, it reports: Operation gko::as<gko::matrix::Dense<double>> does not support parameters of type gko::matrix::Dense<std::complex<double> >. I have no idea about it. May you help me?

Thank you!

### LINKED ISSUE #609: Test execution of examples in the CI
Currently we do not test examples within our CI. To make sure that all the examples work as expected, it would be nice to execute the examples and compare the outputs.

Current ideas:

1. In the documentation we already have expected output of the examples. On the CI we can execute the examples and perform a simple diff to compare the differences.
1a. For examples, where non-deterministic behaviour is possible, we can also enforce checking within a tolerance.
2. Add a `examples` target to execute all examples (with `make examples`), with optional parameters to execute specific examples.

thoasm:
This PR updates the examples to always use a non-complex type for norms and makes sure that most of them compile and run properly with both `std::complex<double>` and `double` as `ValueType`.

Additionally done:
- Update the expected result in the documentation.
- Removed 27pt stencil example completely (since the error of the result is quite high)

Fixes #606


examples/twentyseven-pt-stencil-solver/doc/results.dox
```diff
@@ -1,9 +1,123 @@
...
+0.0291853 0.0463483 0.0689827 0.102026 0.150062 0.217457 0.308533 0.427951 0.581508 0.77596
+0.0491702 0.0708768 0.0972465 0.133535 0.184549 0.254761 0.348388 0.469474 0.621805 0.807902
+0.0814686 0.106594 0.135639 0.174097 0.226922 0.298608 0.393213 0.514273 0.664501 0.845005
+0.130712 0.15823 0.189002 0.228752 0.28255 0.354906 0.449762 0.57036 0.718928 0.896063
+0.201362 0.230216 0.261794 0.302066 0.356228 0.428829 0.523757 0.644108 0.791863 0.967311
+0.297681 0.326626 0.357976 0.397977 0.451965 0.524544 0.619611 0.740215 0.888206 1.06359
+0.423659 0.45105 0.48089 0.519702 0.572929 0.645238 0.74059 0.862101 1.01162 1.18894
+0.582779 0.606285 0.632968 0.669474 0.721228 0.792923 0.888647 1.01175 1.1644 1.34646
+0.777392 0.793673 0.81518 0.848059 0.897436 0.967948 1.06387 1.18904 1.34649 1.53734
+:
```
yhmtsai, line 28: is any idea why we use `:` to separate?

examples/performance-debugging/doc/results.dox
```diff
@@ -2,86 +2,90 @@
...
+
 @endcode
 
 Here is a sample output in the file log.txt:
 @code{.cpp}
-Generate time (ns): 3596
-Apply time (ns): 253224
+
+Generate time (ns): 30933
+Apply time (ns): 4367019
```
yhmtsai, line 27: the time is much slower than previous one. Could you rerun it when the system is idle? or original one is not on reference executor.

examples/preconditioned-solver/doc/results.dox
```diff
@@ -24,10 +25,10 @@ Solution (x):
 0.0107016
 0.0121141
 0.0123025
-Residual norm sqrt(r^T r): 
+Residual norm sqrt(r^T r):
 %%MatrixMarket matrix array real general
 1 1
-9.08137e-16
+4.82005e-08
```
yhmtsai, line 31: the original setting is `reduction_factor = 1e-20` and `max_iteration = 20` from https://github.com/ginkgo-project/ginkgo/blob/1cc73620651c447a8d4703b1045df41babd74d0a/examples/preconditioned-solver/preconditioned-solver.cpp. Thus, this residual changes makes sense in `reduction_factor = 1e-7`.

examples/ginkgo-overhead/doc/results.dox
```diff
@@ -2,8 +2,10 @@
 This is the expected output:
 
 @code{.cpp}
-Running 1000000 iterations of the CG solver took a total of 1.50535 seconds.
-        Average library overhead:     1505.35 [nanoseconds / iteration]
+
+Running 1000000 iterations of the CG solver took a total of 13.7016 seconds.
+	Average library overhead:     13701.6 [nanoseconds / iteration]
```
yhmtsai, line 7: Could you also rerun it ?

examples/ilu-preconditioned-solver/doc/results.dox
```diff
@@ -24,10 +25,10 @@ Solution (x):
 0.0107016
 0.0121141
 0.0123025
-Residual norm sqrt(r^T r): 
+Residual norm sqrt(r^T r):
 %%MatrixMarket matrix array real general
 1 1
-9.08137e-16
+1.46249e-08
```
yhmtsai, line 31: the original one is from `reduction_factor = 1e-15`. current `reduction_factor = 1e-7`

examples/inverse-iteration/inverse-iteration.cpp
```diff
@@ -126,7 +127,7 @@ int main(int argc, char *argv[])
     }();
     auto y = clone(x);
     auto tmp = clone(x);
-    auto norm = clone(one);
+    auto norm = gko::initialize<real_vec>({1.0}, exec);
     auto inv_norm = clone(this_exec, one);
```
pratikvn, line 131: I think technically this should also be of type `real_vec`, though it would not make a difference here.

examples/twentyseven-pt-stencil-solver/doc/results.dox
thoasm: I have no idea, I think that was done by a student assist.

upsj: It was a student from the HPC course, I believe.

examples/performance-debugging/doc/results.dox
thoasm: This was done on my laptop with the reference executor, which is likely the reason why it is a bit slow.
Do you think the runtime matters here?

examples/inverse-iteration/inverse-iteration.cpp
thoasm: I had it it like that, but since we use `inv_norm` to scale `x`, we need it in the same type `x->scale(lend(clone(exec, inv_norm)));`:
https://github.com/ginkgo-project/ginkgo/pull/612/files/3354330466daedb07cf8ed3221d0f287f758a78b#diff-8c65e0d7c939b9652bd35aae6dcef079R147

pratikvn: Ah, right. I forgot about that.

examples/ginkgo-overhead/doc/results.dox
thoasm: Updated now, this time, I compiled it as `RELEASE` and ran it on a Threadripper. The output I added previously was compiled in `DEBUG` mode, which explains why the overhead was that bad.

thoasm:
I just removed the 27pt stencil example, updated the timing output from the examples and rebased it.

PR #621, opened 2020-08-18
Use unified memory in CUDA debug builds

upsj:
This PR uses unified memory instead of device memory in CUDA debug builds, allowing you to directly access device memory in `cuda-gdb`.

It additionally fixes a bug in multi-GPU raw_copy_to, which used the wrong device ordinals due to a wrongly named parameter (src -> dest)

### DIFF
--- a/dev_tools/scripts/gdb-ginkgo.py
+++ b/dev_tools/scripts/gdb-ginkgo.py
@@ -106,14 +106,11 @@ class GkoArrayPrinter:
         self.val = val
         self.execname = str(self.val['exec_']['_M_ptr'].dereference().dynamic_type)
         self.pointer = get_unique_ptr_data_ptr(self.val['data_']);
-        self.is_cpu = re.match('gko::(Reference|Omp)Executor', str(self.execname)) is not None
 
     def children(self):
-        if self.is_cpu:
-            return self._iterator(self.pointer, self.val['num_elems_'])
-        return []
+        return self._iterator(self.pointer, self.val['num_elems_'])
 
-    def to_string(self):     
+    def to_string(self):
         return ('%s of length %d on %s (%s)' % (str(self.val.type), int(self.val['num_elems_']), self.execname, self.pointer))
 
     def display_hint(self):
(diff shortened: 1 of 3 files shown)


pratikvn:
I definitely like the idea of using UVM in debug builds, but I am not so sure if we should be implementing two different implementations for Release and Debug. To me debug builds are meant to have the same implementation but maybe with more possible checks and information, but in this case the implementation (std::memcpy vs CudaMemcpyPeer) is different which kind of violates that principle.

Nice job finding that bug for the multi-GPU raw_copy_to. I guess that was the issue with my CudaUVM memspace invalid device ordinal errors. 

Also you dont seem to be doing this for HIP device -->HIP device raw_copy_to.

upsj:
That's a good point, it might also be that you can just use cudaMemcpy as usual with UVM, I will check.

pratikvn (approved):
LGTM!

dev_tools/scripts/gdb-ginkgo.py
```diff
@@ -106,14 +106,11 @@ def __init__(self, val):
         self.val = val
         self.execname = str(self.val['exec_']['_M_ptr'].dereference().dynamic_type)
         self.pointer = get_unique_ptr_data_ptr(self.val['data_']);
-        self.is_cpu = re.match('gko::(Reference|Omp)Executor', str(self.execname)) is not None
 
     def children(self):
-        if self.is_cpu:
-            return self._iterator(self.pointer, self.val['num_elems_'])
-        return []
+        return self._iterator(self.pointer, self.val['num_elems_'])
```
yhmtsai, line 111: using gdb only in cuda not hip?

yhmtsai (approved):
LGTM

dev_tools/scripts/gdb-ginkgo.py
upsj: Yes, that probably makes sense. This is still left-over from when I thought AMD would work

dev_tools/scripts/gdb-ginkgo.py
```diff
@@ -106,14 +106,11 @@ def __init__(self, val):
         self.val = val
         self.execname = str(self.val['exec_']['_M_ptr'].dereference().dynamic_type)
         self.pointer = get_unique_ptr_data_ptr(self.val['data_']);
```
upsj, line 108:
```suggestion
        self.pointer = get_unique_ptr_data_ptr(self.val['data_']);
        self.is_cpu = re.match('gko::(Reference|Omp|Cuda)Executor', str(self.execname)) is not None
```

thoasm (approved):
LGTM!

PR #622, opened 2020-08-19
Fix warnings for diagonal matrix format

fritzgoebel:
The diagonal format was missing `using EnableLinOp<Diagonal>::convert_to` and `using EnableLinOp<Diagonal>::move_to` which caused a ton of warnings when compiling with cuda enabled.


hartwiganzt (approved):

upsj (approved):

yhmtsai (approved):

upsj:
We should maybe think about making the `warnings` CI job more strict by considering all warnings as errors (currently we only consider `-Wpedantic` errors)

PR #626, opened 2020-08-26
Add custom SpGEMM implementation

upsj:
This PR adds custom SpGEMM implementations to Ginkgo. For OpenMP, there is only a single implementation, for CUDA/HIP, the implementation is selected by the CSR strategy type:
* `classical` selects a two-pass (count nnz per row, then write) algorithm that uses `(3 * sizeof(index) + sizeof(value))*nnz(A)` bytes of auxiliary memory. It performs well on matrices with short and balanced rows.
* `loadbalance` selects a multi-pass (multiple intermediate matrices) algorithm whose intermediate storage depends on the problem. It performs well on all types of matrices, except for heavily imbalanced row lengths.
* `sparselib` selects the cuSPARSE/hipSPARSE implementation.

On the inner workings of the algorithm:
It's basically a multiway-merge SpGEMM, which means we compute the union of all rows from B matching column indices in a row from A. This also means that the rows of B need to be sorted, and the output A*B will be sorted.
* For short rows in A, we store the next unmerged entry (or entries) for each row from B in registers and compute the minimum in-place.
* For medium-sized rows, we store the next entry for each row from B in registers and shared memory (one block of k=4 entries per thread), where the register entry always has the minimum column within the block
* For long rows, we store entries larger than the shared-memory entries in global memory, as one k-ary heap per thread.
The OpenMP implementation uses a single binary heap for all row sizes, instead.

The benchmarks are still running, but from what I can tell, we almost consistently outperform cuSPARSE 10 and 11, but are at the moment still slower than some other SpGEMM libraries. This is hopefully subject to change :)


upsj:
What we should discuss here is how we want to deal with the wrappers for other libraries I implemented. Should they be maintained in `develop` or part of a separate branch we just keep for benchmark reproducibility?

upsj:
Figure_1
Some performance numbers for 32bit float SpGEMMon the V100 (I forgot to change it to 64bit and don't want to rerun all of it)
Visualization: window [flops / 1.2, flops * 1.2], showing median and 5% quantiles, then smoothed with a Gaussian kernel on the log values with sigma = 1/250 *  total FLOPs range

upsj:
Figure_Profile_Flops_1e5
Performance profile for all matrices which need more than 2e5 FLOPs for the SpGEMM

upsj:
I found a few failures among the SuiteSparse benchmarks, will investigate these before proceeding

upsj:
check-format!

upsj:
This branch is incredibly outdated, and I will probably rewrite it from scratch.

PR #632, opened 2020-08-31
Malicious change

tcojean:
Testing what happens with self hosted runners & doing a malicious change telling Github to run the action on a new pull request now.

### DIFF
--- a/.github/workflows/dummy.yml
+++ b/.github/workflows/dummy.yml
@@ -8,6 +8,7 @@ jobs:
       fail-fast: false
     name: dummy-echo
     runs-on: [self-hosted, cuda]
+
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/dummy.yml
```diff
@@ -1,6 +1,6 @@
 name: Dummy-echo
 
-on: [push]
+on: [push,pull_request]
```
tcojean, line 3: Looks like Github doesn't care if a modification of a workflow happens in a fork and tells to run events on a PR. This is really dangerous. We need to wait for some extra protection of workflows before using them with self hosted runners.

https://github.com/ginkgo-project/ginkgo/runs/1051608479

PR #640, opened 2020-09-14
Mixed Precision Gmres

thoasm:
This PR adds the compressed basis GMRES.

__TODO__:
- [ ] Merge #643
- [ ] Update accessor use to the merged version


tcojean:
First, nice that you have several examples. Quick question: do you think it makes sense to split this PR into one purely for the accessor, and another one using the accessor in the mixed precision GMRES? 8.5K lines is a bit much, though we've been having multiple such PR recently.

thoasm:
@tcojean thanks for the suggestion, it would definitively make sense.
I will look into extracting the accessor itself, so it is easier to discuss the changes.
It might also make sense to split the GMRES into a reference (maybe OpenMP as well) PR and a CUDA / HIP PullRequest, which should reduce the sizes as well.

tcojean:
Yes I guess we could also split the GMRES itself into at least Reference/base structure and then a PR for all the kernels. I don't think we would have to split the kernels PR though into OpenMP and another one HIP/CUDA, but we'll see depending on the size of each PR.

PR #641, opened 2020-09-14
Fix add_subdirectory builds for Ginkgo

upsj:
Currently, we often use `CMAKE_SOURCE_DIR` instead of the preferred `PROJECT_SOURCE_DIR` (relative to the last project() call) or `CMAKE_CURRENT_SOURCE_DIR` (relative to the current CMakeLists.txt)

This causes issues when Ginkgo is built as a subdirectory using `add_subdirectory(...)` instead of installing it.

TODO
- [x] Add CI job for subdir build similar to test_install
- [x] ~Fix git-cmake-format interaction with ExternalProject~ (**or just ignore it?**)

### DIFF
--- a/test_subdir/CMakeLists.txt
+++ b/test_subdir/CMakeLists.txt
@@ -1,0 +1,9 @@
+cmake_minimum_required(VERSION 3.14)
+project(GinkgoSubdirTest LANGUAGES CXX)
+file(CREATE_LINK "${CMAKE_CURRENT_SOURCE_DIR}/.." "${CMAKE_CURRENT_BINARY_DIR}/ginkgo" SYMBOLIC)
+
+add_subdirectory("${CMAKE_CURRENT_BINARY_DIR}/ginkgo")
+
+add_executable(test_subdir ../test_install/test_install.cpp)
+target_compile_features(test_subdir PUBLIC cxx_std_14)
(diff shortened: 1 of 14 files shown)


test_subdir/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+cmake_minimum_required(VERSION 3.14)
+project(GinkgoSubdirTest LANGUAGES CXX)
+file(CREATE_LINK "${CMAKE_CURRENT_SOURCE_DIR}/.." "${CMAKE_CURRENT_BINARY_DIR}/ginkgo" SYMBOLIC)
```
tcojean, line 3: Nice way using links for that.

tcojean (approved):
LGTM

test_subdir/CMakeLists.txt
upsj: Yes, but unfortunately it requires CMake >= 3.14, so we can't use that anywhere in the public part of the build system.

pratikvn (approved):
Can you confirm that the documentation is generated properly ? I mean hyperlinks for the examples and all the example files themselves. I have had some problems with relative paths and absolute paths with doxygen. 

If so, then this looks good to me!

upsj:
I can confirm that the hyperlinks to individual examples still work, and the 27pt-stencil is now finally gone from all of the documentation.

PR #642, opened 2020-09-15
Test execution of examples in the CI.

### LINKED ISSUE #609: Test execution of examples in the CI

pratikvn:
This PR adds custom targets for the example executables allowing to test the example execution within the CI.

1. Only certain examples are tested. These are passed in as a list.
2. The ability to run examples is added as an option. The example targets are added nonetheless.
3. Most examples are updated to use default values (reading files, output from the user and so on.)
4. The examples are executed in only one job and not in all.

TODO:

+ [x] Diffing the output and the expected output to check correctness. 
+ [x] Re-animate the .gitlab_ci files.

Closes #609.


pratikvn:
Currently the examples are run only on [one CI job](https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/755565989). 

The idea is to basically remove all the numbers with `sed` and compare the expected output in `<example>/doc/results.dox` with the output from the examples run with through the CI job.

For the case of examples, I believe this should be sufficient. For integration tests, if we add them in the future, this would need to be extended to compare the actual values (residuals, solution values and such).

examples/CMakeLists.txt
```diff
@@ -1,25 +1,64 @@
...
+set(run_list "")
+if(GINKGO_RUN_EXAMPLES)
+    foreach(example ${EXAMPLES_LIST})
+        add_custom_target("run-${example}"
+            COMMAND ${CMAKE_CURRENT_BINARY_DIR}/${example}/${example} > ${CMAKE_CURRENT_BINARY_DIR}/${example}/${example}.out
+            WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/${example})
+    endforeach()
+    foreach(example ${EXAMPLES_LIST})
+        file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/${example}/diff-command "#!/bin/bash
+diff <(sed -n '7,$p' ${CMAKE_CURRENT_BINARY_DIR}/${example}/${example}.out | sed -E 's/([0-9]+.)//g') <(sed -n '6,$p' ${CMAKE_SOURCE_DIR}/examples/${example}/doc/results.dox | head -n -4 | sed -E 's/([0-9]+.)//g')")
```
upsj, line 31: Instead of having a global diff command, we could also add individual scripts to the examples that do nothing else but read from stdin and compare the output to what we expect. For most examples, this could be a simple residual check, and for the more complex outputs, we could do some kind of pattern matching with expected output patterns.

examples/CMakeLists.txt
```diff
@@ -1,25 +1,64 @@
...
+    file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/validate_all_examples.sh
+        "NUM_EXAMPLES=$(wc -l < example-targets.txt)\n"
+        "for i in $(seq 1 $NUM_EXAMPLES)\n"
+        "do\n"
+        "  target_name=$(head -n $i example-targets.txt | tail -1)\n"
+        "  ex_target=diff-\$\{target_name\}\n"
+        "  cmake --build ${CMAKE_BINARY_DIR} --target $ex_target\n"
+        "done\n")
+    add_custom_target("validate_all_examples"
+        COMMAND chmod +x ${CMAKE_CURRENT_BINARY_DIR}/validate_all_examples.sh && ${CMAKE_CURRENT_BINARY_DIR}/validate_all_examples.sh
```
upsj, line 56: I would prefer if we could have one validation target per example and let `validate_all_examples` depend on these individual targets - then we wouldn't need these intermediate bash scripts, and it would be easier to re-run just a single validation as a developer.

upsj (commented):
Looks mostly good to my, one thing I would like to avoid though are the changes to the example CMakeLists.txt files:
They are supposed to show how you can interface with Ginkgo in a quick and easy fashion. So if we replace all the target setup by a black-box `setup_example_target`, this becomes less clear to users.

examples/CMakeLists.txt
pratikvn: Yes, each example has a separate validation target called `diff-<example_name>`. The `validate_all_examples.sh` just calls these targets. A dev can individually validate each example as well.

pratikvn:
As the number of examples increase, I think it will get difficult to maintain each `<example>/CMakeLists.txt` separately. Moreover, I dont think it will be difficult to look up what `setup_example_target`, as it just removes the common duplication among the different examples.

examples/CMakeLists.txt
upsj: Ah yes, I missed that. But we could still avoid the intermediate bash scripts by adding a dummy validate_all_examples target and using `add_dependency` to depend on all diff-* targets.

upsj:
But this is not a common duplication specific to Ginkgo, but to all of CMake. If we replace it by something Ginkgo-specific, it may surprise people, who expect add_executable and target_link_libraries. In the best case, they would be able to copy-paste the example code into their own project.
(Also now that I think of this use case, the `ginkgo` target in most example CMakeLists.txt should probably be replaced by `Ginkgo::ginkgo` so it is uniform between subdirectory and installed builds)

examples/CMakeLists.txt
```diff
@@ -1,25 +1,35 @@
 option(GINKGO_BUILD_EXTLIB_EXAMPLE "Build the external-lib-interfacing with deal.II, you need to link the deal.II library." OFF)
-add_subdirectory(custom-logger)
-add_subdirectory(custom-matrix-format)
-add_subdirectory(custom-stopping-criterion)
+option(GINKGO_RUN_EXAMPLES "Run the examples." OFF)
```
upsj, line 2: What is the intended effect of this flag? To me it looks like it only controls the generation of the run/validation targets, but doesn't cause them to be executed? I think we could also get away without this flag.

examples/CMakeLists.txt
```diff
@@ -1,25 +1,35 @@
...
+
+set(EXAMPLES_EXEC_LIST "adaptiveprecision-blockjacobi;custom-logger;ginkgo-overhead;ginkgo-ranges;ilu-preconditioned-solver;ir-ilu-preconditioned-solver;inverse-iteration;iterative-refinement;mixed-precision-ir;nine-pt-stencil-solver;poisson-solver;preconditioned-solver;simple-solver;three-pt-stencil-solver;")
```
upsj, line 4: I would suggest listing one example per line, this makes for nicer diffs in the future.
```suggestion
set(EXAMPLES_EXEC_LIST
    adaptiveprecision-blockjacobi
    custom-logger
    ginkgo-overhead
    ginkgo-ranges
    ilu-preconditioned-solver
    ir-ilu-preconditioned-solver
    inverse-iteration
    iterative-refinement
    mixed-precision-ir
    nine-pt-stencil-solver
    poisson-solver
    preconditioned-solver
    simple-solver
    three-pt-stencil-solver)
```

examples/CMakeLists.txt
```diff
@@ -1,25 +1,35 @@
...
+
+set(EXAMPLES_LIST "${EXAMPLES_EXEC_LIST};custom-matrix-format;custom-stopping-criterion;minimal-cuda-solver;papi-logging;performance-debugging;preconditioner-export;simple-solver-logging")
```
upsj, line 6:
```suggestion
set(EXAMPLES_LIST
    ${EXAMPLES_EXEC_LIST}
    custom-matrix-format
    custom-stopping-criterion
    minimal-cuda-solver
    papi-logging
    performance-debugging
    preconditioner-export
    simple-solver-logging)
```

examples/CMakeLists.txt
```diff
@@ -1,25 +1,35 @@
...
+
 if(GINKGO_BUILD_EXTLIB_EXAMPLE)
-    add_subdirectory(external-lib-interfacing)
+    set(EXAMPLES_LIST "${EXAMPLES_LIST};external-lib-interfacing")
```
upsj, line 9: You could also write
```suggestion
    list(APPEND EXAMPLES_LIST external-lib-interfacing)
```

examples/adaptiveprecision-blockjacobi/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
-add_executable(adaptiveprecision-blockjacobi adaptiveprecision-blockjacobi.cpp)
-target_link_libraries(adaptiveprecision-blockjacobi ginkgo)
-target_include_directories(adaptiveprecision-blockjacobi PRIVATE ${PROJECT_SOURCE_DIR})
+set(target_name "adaptiveprecision-blockjacobi")
+add_executable(${target_name} ${target_name}.cpp)
+target_link_libraries(${target_name} Ginkgo::ginkgo)
+target_include_directories(${target_name} PRIVATE ${PROJECT_SOURCE_DIR})
```
upsj, line 4: Do we even need the target_name somewhere? Otherwise, we could remove these changes entirely.

examples/custom-logger/custom-logger.cpp
```diff
@@ -232,7 +232,7 @@ int main(int argc, char *argv[])
                gko::HipExecutor::get_num_devices() > 0) {
         exec = gko::HipExecutor::create(0, gko::OmpExecutor::create());
     } else {
-        std::cerr << "Usage: " << argv[0] << " [executor]" << std::endl;
+        std::cerr << "Usage: executable [executor]" << std::endl;
```
upsj, line 235: What is the purpose of these changes? Do we test for the usage message somewhere?

upsj (approved):
LGTM! Aside from my previous comments, only code style suggestions and maybe a few changes we don't need.

examples/custom-logger/custom-logger.cpp
pratikvn: The usage is being printed when all the arguments are not being passed. And `argv[0]` prints the executable with the whole path.

examples/adaptiveprecision-blockjacobi/CMakeLists.txt
pratikvn: No, just locally in each example. But I would prefer to keep these changes instead of repeating the different names for the different examples.

examples/custom-logger/custom-logger.cpp
upsj: I know, but in a correct execution of these examples, the usage message should never be printed, right? I'm trying to understand why this change is necessary.

pratikvn: For most examples, yes. Some examples, like the [`three-pt-stencil-solver`](https://github.com/ginkgo-project/ginkgo/blob/develop/examples/three-pt-stencil-solver/three-pt-stencil-solver.cpp#L235) and a few others, the usage gets printed when the coefficients are not specified. So, I just wanted to make everything uniform.

upsj: Makes sense, so unless we match the output more fuzzily, these specific examples will have to be modified.

pratikvn: Yes, exactly.

upsj: But the rest of the changes could be reverted then?

pratikvn: Yes, I can do that. I just wanted the output to the uniform across all the examples, but i can revert the changes for the rest of the examples.

examples/CMakeLists.txt
tcojean: I do not agree here. In my view, the examples shouldn't really be integration tests. What we want with this system is to only make sure that examples still run successfully after every Ginkgo update which I believe this sort of check is enough.

examples/papi-logging/CMakeLists.txt
```diff
@@ -1,7 +1,8 @@
 if (GINKGO_HAVE_PAPI_SDE)
-	add_executable(papi-logging papi-logging.cpp)
-	target_link_libraries(papi-logging ginkgo PAPI::PAPI)
-	target_include_directories(papi-logging PRIVATE ${PROJECT_SOURCE_DIR})
+    set(target_name "papi-logging")
```
tcojean, line 2: nit: spaces

examples/poisson-solver/doc/results.dox
```diff
@@ -3,29 +3,8 @@ This is the expected output:
...
-0.296296
-0.364431
-0.442285
-0.530504
-0.629738
-0.740633
-0.863838
-1
-The average relative error is 1.87318e-15
+Usage: executable DISCRETIZATION_POINTS [executor]
```
tcojean, line 6: Why is the usage printed in these cases? The usage should be printed only if there was an error in the launch command, I would rather they do not appear here.

examples/poisson-solver/poisson-solver.cpp
```diff
@@ -122,10 +122,12 @@ int main(int argc, char *argv[])
     using cg = gko::solver::Cg<ValueType>;
     using bj = gko::preconditioner::Jacobi<ValueType, IndexType>;
 
+    // Print version information
+    std::cout << gko::version_info::get() << std::endl;
+
     if (argc < 2) {
-        std::cerr << "Usage: " << argv[0] << " DISCRETIZATION_POINTS [executor]"
+        std::cout << "Usage: executable DISCRETIZATION_POINTS [executor]"
```
tcojean, line 129: Please keep this std::cerr like the other ones

examples/three-pt-stencil-solver/three-pt-stencil-solver.cpp
```diff
@@ -232,10 +232,12 @@ int main(int argc, char *argv[])
     using ValueType = double;
     using IndexType = int;
 
+    // Print version information
+    std::cout << gko::version_info::get() << std::endl;
+
     if (argc < 2) {
-        std::cerr << "Usage: " << argv[0] << " DISCRETIZATION_POINTS [executor]"
+        std::cout << "Usage: executable DISCRETIZATION_POINTS [executor]"
```
tcojean, line 239: std::cerr

examples/nine-pt-stencil-solver/nine-pt-stencil-solver.cpp
```diff
@@ -294,11 +294,13 @@ void solve_system(const std::string &executor_string,
 
 int main(int argc, char *argv[])
 {
+    // Print version information
+    std::cout << gko::version_info::get() << std::endl;
+
     if (argc < 2) {
-        std::cerr << "Usage: " << argv[0] << " DISCRETIZATION_POINTS [executor]"
+        std::cout << "Usage: executable DISCRETIZATION_POINTS [executor]"
```
tcojean, line 301: std::cerr

Also, the discretization points seem to be optional now

examples/CMakeLists.txt
tcojean: Is this a list of examples which can take in executor as parameter? If so, I believe at least `simple-solver-logging` can be here as well.

tcojean (approved):
LGTM. One thing I'm wondering, is it possible to add a target to run the examples in `EXAMPLES_EXEC_LIST` with an executor as parameter? This way, we can in the CI test the examples with every executor and not just reference, as soon as the proper option is compiled.

My reasoning is that some examples good be poorly written (e.g. not using array for something) and fail especially on Cuda Executors, and I think it's important to test that.

examples/CMakeLists.txt
upsj: If I recall correctly, this effort was sparked at least in part by the fact that that the output of some examples was many orders off from what the documentation says during the JOSS review, so this would be a really cheap way to avoid such issues in the future.
I do agree that this is far (both in purpose and execution) from integration tests.

pratikvn: The main problem there was with the twenty-seven-pt stencil solver, which was actually wrong. I think the integration tests should definitely more comprehensive with testing the residuals and the convergence. But maybe for examples, i think this is sufficient.

examples/CMakeLists.txt
pratikvn: Yes, but `simple-solver-logging` prints out pointers and they can be different on each run.

examples/CMakeLists.txt
```diff
@@ -1,25 +1,53 @@
...
+    simple-solver-logging)
+
 if(GINKGO_BUILD_EXTLIB_EXAMPLE)
-    add_subdirectory(external-lib-interfacing)
+    list(APPEND EXAMPLES_LIST external-lib-interfacing)
+endif()
+
+if(GINKGO_RUN_EXAMPLES)
+    foreach(example ${EXAMPLES_LIST})
+        add_subdirectory(${example})
```
tcojean, line 36: I believe we should have this whether `GINKGO_RUN_EXAMPLES` is activated or not.

pratikvn:
For executor as a parameter, then i cannot add it as a target because `make` or `ninja` cannot take arguments. I am not sure how to do this properly.

tcojean:
I think you can pass it as a variable maybe? For example something like:
EXECUTOR="cuda" make run_all_examples

And then in the command you launch with $EXECUTOR as parameter, if it is set?

examples/adaptiveprecision-blockjacobi/CMakeLists.txt
upsj: Except for the ginkgo -> Ginkgo::ginkgo change, this makes no functional changes and I feel like it adds a bit of additional complexity to the CMakeLists.txt. Why not checkout the previous state and just update the Ginkgo target name? Again, these are the files that users are most likely to copy-paste into their CMake code, so keeping them simple and un-surprising feels important to me.

pratikvn: okay. I will change everything back to as it was before

examples/nine-pt-stencil-solver/nine-pt-stencil-solver.cpp
```diff
@@ -294,25 +294,29 @@ void solve_system(const std::string &executor_string,
...
-        std::cerr << "Usage: " << argv[0] << " DISCRETIZATION_POINTS [executor]"
-                  << " [stencil_alpha] [stencil_beta] [stencil_gamma]"
+    using ValueType = double;
+    using IndexType = int;
+
+    // Print version information
+    std::cout << gko::version_info::get() << std::endl;
+
+    if (argc == 2 && std::string(argv[1]) == "--help") {
+        std::cerr << "Usage: " << argv[0] << " [executor] DISCRETIZATION_POINTS"
```
tcojean, line 304: You are now missing alpha beta gamma

examples/papi-logging/papi-logging.cpp
```diff
@@ -138,23 +139,32 @@ int main(int argc, char *argv[])
...
-    } else if (argc == 2 && std::string(argv[1]) == "cuda" &&
-               gko::CudaExecutor::get_num_devices() > 0) {
-        exec = gko::CudaExecutor::create(0, gko::OmpExecutor::create(), true);
-    } else if (argc == 2 && std::string(argv[1]) == "hip" &&
-               gko::HipExecutor::get_num_devices() > 0) {
-        exec = gko::HipExecutor::create(0, gko::OmpExecutor::create(), true);
-    } else {
-        std::cerr << "Usage: " << argv[0] << " [executor]" << std::endl;
+    if (argc == 2 && (std::string(argv[1]) == "--help")) {
+        std::cerr << "Usage: " << argv[0] << " [executor] DISCRETIZATION_POINTS"
```
tcojean, line 143: there are no discretization points here

examples/nine-pt-stencil-solver/nine-pt-stencil-solver.cpp
tcojean: Also discretization points is now optional (with `[]`)

examples/poisson-solver/poisson-solver.cpp
```diff
@@ -122,16 +122,19 @@ int main(int argc, char *argv[])
     using cg = gko::solver::Cg<ValueType>;
     using bj = gko::preconditioner::Jacobi<ValueType, IndexType>;
 
-    if (argc < 2) {
-        std::cerr << "Usage: " << argv[0] << " DISCRETIZATION_POINTS [executor]"
+    // Print version information
+    std::cout << gko::version_info::get() << std::endl;
+
+    if (argc == 2 && (std::string(argv[1]) == "--help")) {
+        std::cerr << "Usage: " << argv[0] << " [executor] DISCRETIZATION_POINTS"
```
tcojean, line 129: discretization points now optional

examples/preconditioned-solver/preconditioned-solver.cpp
```diff
@@ -56,22 +57,32 @@ int main(int argc, char *argv[])
...
-    } else if (argc == 2 && std::string(argv[1]) == "cuda" &&
-               gko::CudaExecutor::get_num_devices() > 0) {
-        exec = gko::CudaExecutor::create(0, gko::OmpExecutor::create(), true);
-    } else if (argc == 2 && std::string(argv[1]) == "hip" &&
-               gko::HipExecutor::get_num_devices() > 0) {
-        exec = gko::HipExecutor::create(0, gko::OmpExecutor::create(), true);
-    } else {
-        std::cerr << "Usage: " << argv[0] << " [executor]" << std::endl;
+    if (argc == 2 && (std::string(argv[1]) == "--help")) {
+        std::cerr << "Usage: " << argv[0] << " [executor] DISCRETIZATION_POINTS"
```
tcojean, line 61: No discretization points

examples/three-pt-stencil-solver/three-pt-stencil-solver.cpp
```diff
@@ -232,15 +232,18 @@ int main(int argc, char *argv[])
     using ValueType = double;
     using IndexType = int;
 
-    if (argc < 2) {
-        std::cerr << "Usage: " << argv[0] << " DISCRETIZATION_POINTS [executor]"
+    // Print version information
+    std::cout << gko::version_info::get() << std::endl;
+
+    if (argc == 2 && std::string(argv[1]) == "--help") {
+        std::cerr << "Usage: " << argv[0] << " [executor] DISCRETIZATION_POINTS"
```
tcojean, line 239: discretization points now optional

examples/performance-debugging/performance-debugging.cpp
```diff
@@ -369,18 +369,32 @@ int main(int argc, char *argv[])
...
-        exec = gko::ReferenceExecutor::create();
-    } else if (argc > 1 && std::string(argv[1]) == "omp") {
-        exec = gko::OmpExecutor::create();
-    } else if (argc > 1 && std::string(argv[1]) == "cuda" &&
-               gko::CudaExecutor::get_num_devices() > 0) {
-        exec = gko::CudaExecutor::create(0, gko::OmpExecutor::create(), true);
-    } else {
-        print_usage(argv[0]);
+    if (argc == 2 && (std::string(argv[1]) == "--help")) {
+        std::cerr << "Usage: " << argv[0] << " [executor] DISCRETIZATION_POINTS"
```
tcojean, line 373: No discretization points

tcojean (approved):
Some errors in the example usage string.

PR #643, opened 2020-09-18
Add new accessor

thoasm:
Add both `reduced_row_major` and `scaled_reduced_row_major` accessors to Ginkgo (naming is up for debate).

The tests should cover everything now, and the public interface remains unchanged.

It would also be nice if the function/operator `range::operator=` would be marked as [deprecated](https://en.cppreference.com/w/cpp/language/attributes/deprecated) since it is not optimized at all.
However, this will lead to a lot of deprecated warnings in our own code (mainly `include/ginkgo/core/base/matrix_data.hpp` and `examples/ginkgo-ranges/ginkgo-ranges.cpp`), so it should be done in another PR.

__TODO__:
- [x] Extend `scaled_reduced_row_major` tests
- [x] Add operator tests (`*`, ..., `*=`, ...)
~~Add `constexpr` tests for both~~ (not possible to create a constexpr accessor)
~~Add `to_const()` to `range` (and test it)~~ (This would change the interface; Workaround: use the accessor `to_const()` directly works fine).
- [x] Make core tests typed, so more combinations are tested
- [x] Adapt code to deal with dimensionalities other than 3
- [x] Test (and likely change) the behaviour of `detail::enable_write_scalar` in case the accessor is `const`. Likely simply add boolean with `true` for `const`, and `false` for `non-const`. Also, rename `ScaleType` to `ScalarType`.


core/test/base/range_accessors.cpp
```diff
@@ -121,4 +139,569 @@ TEST_F(RowMajorAccessor, CanAssignSubranges)
...
+    EXPECT_EQ(subr(0, 0, 0), 22.);
+    EXPECT_EQ(subr(0, 1, 0), 26.);
+}
+
+
+TEST_F(ScaledReducedStorage3d, CanWriteScale)
+{
+    r->write_scale(0, 0, 10.);
+
+    EXPECT_EQ(r(0, 0, 0), 100.);
```
pratikvn, line 585: Maybe I am misunderstanding something here but I thought `write_scale` just writes a scalar value at a certain index [code doc](https://github.com/ginkgo-project/ginkgo/blob/add_new_accessor/include/ginkgo/core/base/range_accessors.hpp#L807). Here it seems to be scaling the existing the value by the passed in value at the index ?

pratikvn (commented):
Great work on this! Very nicely documented. 

Some minor nits and questions. A few naming suggestions. A lot of spam about new lines.

core/test/base/range_accessors.cpp
```diff
@@ -121,4 +139,569 @@ TEST_F(RowMajorAccessor, CanAssignSubranges)
...
+    EXPECT_EQ(r(0, 1, 0), 100.2);
+}
+
+
+TEST_F(ReducedStorage3d, Assignment)
+{
+    r(0, 0, 1) = 10.2;
+
+    check_accessor_correctness(r, t(0, 0, 1));
+    EXPECT_NEAR(r(0, 0, 1), 10.2, delta);
```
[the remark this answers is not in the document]
thoasm: Yes, but is not necessarily for `double` -> `float`. I was thinking about making this test for multiple precisions, and I going through `static_cast` twice is not guaranteed to result in the same value (I might have to change the `check_accessor_correctness` in that regard as well, but it is centralized, so changing that is easy).

core/test/base/range_accessors.cpp
```diff
@@ -121,4 +139,569 @@ TEST_F(RowMajorAccessor, CanAssignSubranges)
...
+    check_accessor_correctness(cr2);
+}
+
+
+TEST_F(ReducedStorage3d, CanWriteData)
+{
+    r(0, 1, 0) = 100.2;
+
+    check_accessor_correctness(r, t(0, 1, 0));
+    EXPECT_EQ(r(0, 1, 0), 100.2);
```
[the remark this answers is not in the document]
thoasm: I do that with `check_accessor_correctness` (I check everything except the index `0, 1, 0`).

include/ginkgo/core/base/range_accessors.hpp
```diff
@@ -186,6 +224,910 @@ class row_major {
...
 
 
+namespace detail {
+
+
+// tests if the cast operator to `ValueType` is present
+template <typename Ref, typename ValueType, typename = xstd::void_t<>>
+struct has_cast_operator : std::false_type {
+};
+
```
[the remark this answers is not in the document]
thoasm: Here, it is a specialization for `has_cast_operator`, so I consider these two code parts as together. But below, you are correct and I added it.

include/ginkgo/core/base/range_accessors.hpp
```diff
@@ -186,6 +224,910 @@ class row_major {
...
+    GKO_REFERENCE_ASSIGNMENT_OPERATOR_OVERLOAD(operator/=, /)
+    GKO_REFERENCE_ASSIGNMENT_OPERATOR_OVERLOAD(operator+=, +)
+    GKO_REFERENCE_ASSIGNMENT_OPERATOR_OVERLOAD(operator-=, -)
+#undef GKO_REFERENCE_ASSIGNMENT_OPERATOR_OVERLOAD
+
+    // TODO test if comparison operators need to be overloaded as well
+
+    friend GKO_ATTRIBUTES GKO_INLINE GKO_ENABLE_REFERENCE_CONSTEXPR
+        arithmetic_type
+        operator-(const Reference &ref)
```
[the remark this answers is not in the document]
thoasm: This overloads the unary minus operator (so something like: `-ref`).

core/test/base/range_accessors.cpp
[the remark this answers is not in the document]
pratikvn: Maybe then you can set `delta` to 0.0 for the functions where it is supposed to be exact ?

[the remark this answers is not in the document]
thoasm: The big question for that is if we want to have this test for multiple precisions, or not. If we don't, then testing for equality is fine. If we do want to test multiple types, we can set `delta` to `0.0` in case storage and arithmetic types are identical.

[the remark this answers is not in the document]
pratikvn: I think it makes sense to have tests for multiple precisions and set `delta` to zero when the storage and arithmetic types are the same.

core/test/base/range_accessors.cpp
thoasm: `write_scale` (now renamed to `write_scalar`) writes the given value to the scalar location. That means that the storage is interpreted differently, since all reads of the accessor (and not the scalar function) basically do `scalar * stored_value`, and all writes `stored_value = given_value / scalar`. That means a change in scale changes the way the memory is interpreted (that uses that specific scalar).

include/ginkgo/core/base/range_accessors.hpp
```diff
@@ -186,6 +224,910 @@ class row_major {
...
+to_value_type(const Ref &ref)
+{
+    return static_cast<ValueType>(ref);
+}
+
+/**
+ * This is a mixin which defines the binary operators for *, /, +, - for the
+ * Reference class, the unary operator -, and the assignment operators
+ * *=, /=, +=, -=
+ * All assignment operators expect an rvalue reference (Reference &&) for
```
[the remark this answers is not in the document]
thoasm: Since it is pronounced "ar-value reference", I think `an` is correct here.

include/ginkgo/core/base/range_accessors.hpp
```diff
@@ -186,6 +224,910 @@ class row_major {
...
+    : public detail::enable_reference<
+          reduced_storage<ArithmeticType, StorageType>, ArithmeticType> {
+public:
+    using arithmetic_type = std::remove_cv_t<ArithmeticType>;
+    using storage_type = StorageType;
+    // Allow move construction, so perfect forwarding is possible (required
+    // for `range` support)
+    reduced_storage(reduced_storage &&) = default;
+
+    reduced_storage() = delete;
```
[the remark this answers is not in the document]
thoasm: Inside a class, we are not supposed to have 2 consecutive empty lines according to our guidelines.

include/ginkgo/core/base/range_accessors.hpp
```diff
@@ -186,6 +224,910 @@ class row_major {
...
+ * @tparam StorageType  Type actually used as a storage, which is converted
+ * to ArithmeticType before usage
+ */
+template <typename ArithmeticType, typename StorageType>
+class reduced_storage
+    : public detail::enable_reference<
+          reduced_storage<ArithmeticType, StorageType>, ArithmeticType> {
+public:
+    using arithmetic_type = std::remove_cv_t<ArithmeticType>;
+    using storage_type = StorageType;
```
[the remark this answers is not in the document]
thoasm: Do you really want to separate `using` directives? I personally like grouping them.

include/ginkgo/core/base/range_accessors.hpp
```diff
@@ -186,6 +224,910 @@ class row_major {
...
+    using const_accessor =
+        reduced_row_major<dimensionality, arithmetic_type, const storage_type>;
+
+    static_assert(dimensionality == 3,
+                  "Only Dimensionality == 3 is currently supported");
+
+    friend class range<reduced_row_major>;
+
+protected:
+    using reference =
```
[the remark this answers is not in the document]
thoasm: Here, I kept `reference` since this is a `protected` member and here, it is hard to confuse it with the executor.

include/ginkgo/core/base/range_accessors.hpp
```diff
@@ -186,6 +224,910 @@ class row_major {
...
+
+// In case of a const type, do not provide a write function
+template <int Dimensionality, typename Accessor, typename ScaleType,
+          bool = std::is_const<ScaleType>::value>
+struct enable_write_scale {
+    static_assert(std::is_const<ScaleType>::value,
+                  "This class must have a constant ScaleType!");
+    using scale_type = ScaleType;
+};
+
```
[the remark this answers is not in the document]
thoasm: Since I consider these two classes related, I kept only a single empty line.

include/ginkgo/core/base/range_accessor_helper.hpp
```diff
@@ -0,0 +1,586 @@
...
+                     std::array<ValueType, N == 0 ? 0 : N - 1>>
+    extract_factorization(const dim<N> &size, Args... args)
+{
+    return {{args...}};
+}
+
+template <typename ValueType, size_type iter = 1, size_type N, typename... Args>
+constexpr GKO_ATTRIBUTES GKO_INLINE std::enable_if_t<
+    (iter < N) && (iter == sizeof...(Args) + 1), std::array<ValueType, N - 1>>
+extract_factorization(const dim<N> &size, Args... args)
```
tcojean, line 133: I'm again not sure about the naming here. What this does is call `mult_array` for every dimension successively, carry all those results around, until it's stored in an array with the previous call. I would name this `compute_default_strides` or something like that.

include/ginkgo/core/base/range_accessor_helper.hpp
```diff
@@ -0,0 +1,586 @@
...
+
+// bit for current dimensionality is set
+template <typename ValueType, size_type mask, size_type set_bits_processed,
+          size_type stride_size, size_type dim_idx, size_type total_dim>
+struct row_major_masked_helper_s<ValueType, mask, set_bits_processed,
+                                 stride_size, dim_idx, total_dim, true> {
+    static_assert(mask &
+                      (size_type{1}
+                       << (dim_idx < total_dim ? total_dim - 1 - dim_idx : 0)),
+                  "Do not touch the `mask_set` template parameter!");
```
tcojean, line 209: You mean `mask`?

include/ginkgo/core/base/range_accessor_helper.hpp
```diff
@@ -0,0 +1,586 @@
...
+{
+    return 1;
+}
+
+template <typename ValueType, size_type iter, size_type N>
+constexpr GKO_ATTRIBUTES GKO_INLINE std::enable_if_t<(iter < N), ValueType>
+mult_array(const dim<N> &size)
+{
+    return size[iter] * mult_array<ValueType, iter + 1>(size);
+}
```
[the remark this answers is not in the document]
tcojean: Looking at the rest of the file,  lot of functions are `dim` specific (such as successively computing all strides, or in a masked fashion), and could be moved to the dim header as well. While they are needed by the accessor, I don't think they are accessor specific.

include/ginkgo/core/base/range_accessor_helper.hpp
```diff
@@ -0,0 +1,586 @@
...
+template <typename... Args>
+struct are_all_integral : public std::true_type {
+};
+
+template <typename First, typename... Args>
+struct are_all_integral<First, Args...>
+    : public std::conditional<std::is_integral<std::decay_t<First>>::value,
+                              are_all_integral<Args...>,
+                              std::false_type>::type {
+};
```
tcojean, line 432: Maybe this one can be moved to `types.hpp`? This seems like something which could be reused, and I'm not sure we would look for it here.

include/ginkgo/core/base/range_accessors.hpp
```diff
@@ -62,7 +73,8 @@ namespace accessor {
  *          2-dimensional ranges.
  *
  * @tparam ValueType  type of values this accessor returns
- * @tparam Dimensionality  number of dimensions of this accessor (has to be 2)
+ * @tparam Dimensionality  number of dimensions of this accessor (has to be
+ * 2)
```
tcojean, line 77: same minor comment with text alignement in this file

include/ginkgo/core/base/range_accessors.hpp
```diff
@@ -186,6 +216,647 @@ class row_major {
...
+        return dimension < dimensionality ? size_[dimension] : 1;
+    }
+
+    /**
+     * Copies data from another accessor
+     *
+     * @warning Do not use this function since it is not optimized for a
+     *          specific executor. It will always be performed sequentially.
+     *          Please write an optimized version (adjusted to the architecture)
+     *          by iterating through the values yourself.
```
tcojean, line 344: It doesn't have to be part of the PR, but it would be nice to have an example on that topic at some point, I think.

include/ginkgo/core/base/range_accessors.hpp
```diff
@@ -186,6 +216,647 @@ class row_major {
...
+        helper::are_all_integral<Indices...>::value,
+        std::conditional_t<is_const, arithmetic_type, reference_type>>
+    operator()(Indices &&... indices) const
+    {
+        return reference_type{storage_ +
+                              compute_index(std::forward<Indices>(indices)...)};
+    }
+
+    /**
+     * Returns a sub-range spinning the current range (x1_span, x2_span, ...)
```
tcojean, line 378:
```suggestion
     * Returns a sub-range spanning the current range (x1_span, x2_span, ...)
```
? spinning looks like you shuffle them

core/test/base/range_accessors.cpp
```diff
@@ -121,4 +142,901 @@ TEST_F(RowMajorAccessor, CanAssignSubranges)
...
+
+TYPED_TEST(ScaledReducedStorage3d, CorrectStride)
+{
+    EXPECT_EQ(this->r->get_storage_stride()[0], this->size[1] * this->size[2]);
+    EXPECT_EQ(this->r->get_storage_stride().at(0),
+              this->size[1] * this->size[2]);
+    EXPECT_EQ(this->r->get_storage_stride()[1], this->size[2]);
+    EXPECT_EQ(this->r->get_storage_stride().at(1), this->size[2]);
+    EXPECT_EQ(this->r->get_scalar_stride(), this->scalar_stride);
+    EXPECT_EQ(this->r->get_scalar_stride(), this->scalar_stride);
```
tcojean, line 690: duplicate

tcojean (approved):
LGTM. Great work!

tcojean:
On the issue pointed by sonarcloud, maybe indeed you should use `Args... &&args` in some places?

include/ginkgo/core/base/range_accessor_references.hpp
```diff
@@ -0,0 +1,418 @@
...
+ * from class `Ref` before falling back to a simple
+ * `static_cast<ValueType>`.
+ *
+ * This function is only needed for CUDA TOOLKIT < 11 because
+ * thrust::complex has a constructor call: `template<T> complex(const T
+ * &other) : real(other), imag()`, which is always preferred over the
+ * overloaded `operator value_type()`.
+ */
+template <typename ValueType, typename Ref>
+GKO_ATTRIBUTES GKO_INLINE constexpr std::enable_if_t<
```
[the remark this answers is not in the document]
tcojean: I think that was fixed in between my review.

tcojean:
Another sonarqube related comment, I think it's possible to remove quite a bit of the "code smells" left, a lot of them amount to:
+ there is no need to add `inline` to a `constexpr` function (it's already necessarily inlined)
+ sometimes you declare the parameter name size but don't use it, removing the parameter name will remove this kind of warning

thoasm:
@tcojean I haven't actually looked too much into the sonarcloud Smells yet, thanks for pointing that out.
Do you know if the implicitly declared `inline` is as strong as the `__forceinline__` from CUDA?

Removing the name of the parameter was something I also did in the latest commit (I fixed perfect forwarding and unused parameters). I will look into the other smells today.

include/ginkgo/core/base/range_accessor_references.hpp
```diff
@@ -0,0 +1,417 @@
...
+ * *=, /=, +=, -=
+ * All assignment operators expect an rvalue reference (Reference &&) for
+ * the Reference class in order to prevent copying the Reference object.
+ *
+ * @tparam Reference  The reference class this mixin provides operator
+ * overloads for. The reference class needs to overload the cast operator
+ * to ValueType
+ *
+ * @tparam ArithmeticType  arithmetic type the Reference class is supposed
+ * to represent.
```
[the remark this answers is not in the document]
thoasm: Thanks, I usually try to follow that style, but sometimes I add more text and don't check if it adds line-breaks or not. I updated (hopefully) all documentation accordingly.

core/test/base/range_accessors.cpp
thoasm: I think I was about to simplify it by just checking both arrays for equality and did not finish it for some reason...
Good find.

thoasm (commented):
Replies addressing @tcojean 's comments.

include/ginkgo/core/base/range_accessors.hpp
thoasm: Yes, good find, it was a typo.

include/ginkgo/core/base/range_accessor_helper.hpp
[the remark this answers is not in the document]
thoasm: I didn't see another use-case for this functionality when I implemented it (now, I renamed it to `mult_size_upwards`).
I agree I would not search for it in this place in case I ever needed it (and would not have written it myself). The main reason why I did not add it do `dim` is that I try to put as little code into a public header as possible.

Now, thinking about it, it is useful to compute the total number of elements for a given size that way. I will need to think about a better name for it, but I think it makes sense to move this into `dim.hpp`.

The masked stuff is something I would keep here since I don't see any other use case for it, and I don't know if there will ever be one outside of accessors.

include/ginkgo/core/base/range_accessor_helper.hpp
thoasm: Excellent point, I renamed this to `compute_default_stride_array`.
I actually meant the opposite of `factorization` (`extract_recursive_product` might have been a bit more accurate), but I agree that even that is not stating the purpose of the function.

include/ginkgo/core/base/range_accessor_helper.hpp
```diff
@@ -0,0 +1,586 @@
...
+
+
+namespace gko {
+namespace accessor {
+/**
+ * This namespace contains helper functionality for the accessors.
+ *
+ * @note This namespace is not part of the public interface and can change
+ * without notice.
+ */
```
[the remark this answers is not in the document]
tcojean: No the current way is fine, thanks!

core/test/base/range_accessors.cpp
```diff
@@ -121,4 +142,896 @@ TEST_F(RowMajorAccessor, CanAssignSubranges)
...
+        if (ignore != t(3, 1, 1)) { EXPECT_EQ(a(3, 1, 1), st_type{-4.21});   }
+        if (ignore != t(3, 2, 0)) { EXPECT_EQ(a(3, 2, 0), st_type{3.22});    }
+        if (ignore != t(3, 2, 1)) { EXPECT_EQ(a(3, 2, 1), st_type{-2.23});   }
+        // clang-format on
+    }
+};
+
+using ReducedStorage3dTypes =
+    ::testing::Types<std::tuple<double, double>, std::tuple<double, float>,
+                     std::tuple<float, float>
```
tcojean, line 225: Quick question here, our custom float16/half implementation doesn't work for this? I believe it was made to work in combination with float, and to be used purely as a reduced storage.

thoasm: It should work here. Only the complex versions currently do not work because the ASSERT_NEAR don't work with complex.
I made it to work, just need to dial in the delta value.

core/base/extended_float.hpp
```diff
@@ -67,7 +67,8 @@ namespace detail {
 
 
 template <std::size_t, typename = void>
-struct uint_of_impl {};
+struct uint_of_impl {
```
upsj, line 70: `clang-format-8` seems to disagree on these change, which version are you using?

include/ginkgo/core/base/types.hpp
```diff
@@ -91,6 +93,14 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #endif  // defined(_WIN32) || defined(__CYGWIN__)
 
 
+/**
+ * Encapsulates all arguments in one macro, so a comma inside an argument (e.g.
+ * std::tuple<int, float>) still works as an argument and is not considered as
+ * multiple arguments.
+ */
+#define GKO_UNPACK(...) __VA_ARGS__
```
upsj, line 101: You can also do this without an additional macro by adding parentheses around the parameter:
`GKO_MACRO((parameter, with, commas), (next, parameter))`

omp/test/preconditioner/jacobi_kernels.cpp
```diff
@@ -308,7 +308,7 @@ TEST_F(Jacobi, OmpPreconditionerEquivalentToRefWithBlockSize32Sorted)
 TEST_F(Jacobi, OmpPreconditionerEquivalentToRefWithBlockSize32Unsorted)
 {
     std::ranlux48 engine(43);
-    initialize_data({0, 32, 64, 96, 128}, {}, {}, 32, 100, 110, 0.1, false);
+    initialize_data({0, 32, 64, 96, 128}, {}, {}, 32, 100, 110, 1, 0.1, false);
```
upsj, line 311: Good catch! This seems to have been a bug, right?

core/test/utils.hpp
```diff
@@ -90,23 +91,28 @@ using ComplexValueIndexTypes =
...
+template <typename Precision, typename OutputType>
 struct reduction_factor {
-    static constexpr gko::remove_complex<T> value =
-        std::is_same<gko::remove_complex<T>, float>::value ? 1.0e-7 : 1.0e-14;
+    static constexpr remove_complex<OutputType> value =
+        std::is_same<gko::remove_complex<Precision>, gko::half>::value
+            ? 1.0e-2
+            : std::is_same<gko::remove_complex<Precision>, float>::value
+                  ? 1.0e-6
+                  : 1.0e-14;
```
upsj, line 101: Can we get rid of these explicit numbers? specialization for `std::numeric_limits<gko::half>`?

include/ginkgo/core/base/range_accessor_references.hpp
```diff
@@ -0,0 +1,429 @@
...
+          reduced_storage<ArithmeticType, StorageType>, ArithmeticType> {
+public:
+    using arithmetic_type = std::remove_cv_t<ArithmeticType>;
+    using storage_type = StorageType;
+
+    // Allow move construction, so perfect forwarding is possible (required
+    // for `range` support)
+    reduced_storage(reduced_storage &&) = default;
+
+    reduced_storage() = delete;
```
upsj, line 228: not: I think you don't need to delete the default constructor if there is a constructor with parameters

upsj (commented):
First look over the implementation: Really good job, I guess this could significantly simplify the move to mixed precision with more complex storage types.

As usual, I have a naming suggestion and maybe a file structure discussion point:
* the name *references* is pretty overloaded in C++ and I feel like it doesn't exactly describe what your objects are doing (more akin to something like a *proxy* object?)
* Do we expect users to use these accessors themselves? Otherwise, we could also move the decoupled precision accessor implementations to core without impacting the public interface at all. This would also allow us to change the interface and implementation more freely.

omp/test/preconditioner/jacobi_kernels.cpp
thoasm: Yes, a but I introduced with the Jacobi fix.

core/base/extended_float.hpp
thoasm: I tried to force `clang-format-10`, but for some reason, it wants to use `clang-format-5` in the `format` portion. Any suggestions on how to change the version used? I was thinking about removing the `clang-format` binary and replace it with a symlink, but I would prefer to not do that.

upsj: You can replace the path in CMakeCache.txt

core/test/utils.hpp
thoasm: Sure, I can do that. It will likely be `10 * std::numeric_limits<T>` since we don't use the actual limits in our tests.

include/ginkgo/core/base/types.hpp
thoasm: I actually did not know that, I thought I need such a macro. I will remove it and see if the next PR (the CB-GMRES) can work with your suggestion.

include/ginkgo/core/base/range_accessor_references.hpp
thoasm: I think I need it. According to [cppreference/default_constructor](https://en.cppreference.com/w/cpp/language/default_constructor):
>If no user-defined constructors are present and the implicitly-declared default constructor is not trivial, the user may still inhibit the automatic generation of an implicitly-defined default constructor by the compiler with the keyword delete. 

I think the implicitly-declared constructor is probably trivial (assigning a `nullptr` to a `storage_type *const`). In any case, I like to specify it here since the rules are a bit complicated, so explicitly stating that I don't want to have a default constructor felt reasonable, especially since a reference to nothing does not make sense.

thoasm:
format!

thoasm:
format!

include/ginkgo/core/base/dim.hpp
```diff
@@ -242,6 +245,76 @@ constexpr GKO_ATTRIBUTES GKO_INLINE dim<2, DimensionType> transpose(
...
+ * @tparam dimensions  number of dimensions in the given size
+ *
+ * @tparam DimensionType  value type of the stored size
+ *
+ * @returns an std::array<ValueType, dimensions - 1> with the stride
+ *          information.
+ */
+template <typename ValueType, size_type dimensions, typename DimensionType>
+constexpr GKO_ATTRIBUTES
+    std::array<ValueType, (dimensions > 0 ? dimensions - 1 : 0)>
```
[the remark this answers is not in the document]
thoasm: The stride array always has one value less than the number of dimensions.
Example: 2D 4x6 Matrix, stored in row major (dim<2> == [4, 6])
The default stride we use here is 6 (which is a 1D array).
We compute the index like this:
```
compute_2d_index(x, y) {
    return x * stride + y; 
}
compute_3d_index(x, y, z) {
    return x * stride[0] + y * stride[1] + z; 
}
```

core/test/base/dim.cpp
```diff
@@ -79,6 +80,66 @@ TEST(Dim, ConstructsNullObject)
 }
 
 
+TEST(Dim, ComputeDefaultStrideArrayConstexpr)
+{
+    using value_type = unsigned int;
+    constexpr int dimens{4};
+    constexpr gko::dim<dimens, int> d{2, 3, 5, 7};
+    constexpr std::array<value_type, dimens - 1> expected{3 * 5 * 7, 5 * 7, 7};
```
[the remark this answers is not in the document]
thoasm: We could have that, but that would be useless information we would have to store that is not providing additional information.
It is easier to implement it that way, but we don't do it that way right now (and I don't do it that way in the accessor), and the computation of the actual index has an additional multiplication we don't need.
By having an additional `1` entry in every stride array, we would add 8 bytes to every accessor object that we would need to copy (for CUDA and HIP, this takes away space for the kernel parameters).

core/base/accessor_references.hpp
```diff
@@ -0,0 +1,469 @@
...
+/**
+ * This is a mixin which defines the binary operators for *, /, +, - for the
+ * Reference class, the unary operator -, and the assignment operators
+ * *=, /=, +=, -=
+ * All assignment operators expect an rvalue reference (Reference &&) for
+ * the Reference class in order to prevent copying the Reference object.
+ *
+ * @tparam Reference  The reference class this mixin provides operator overloads
+ *                    for. The reference class needs to overload the cast
+ *                    operatorto ValueType
```
pratikvn, line 117:
```suggestion
 *                    operator to ValueType
```

core/base/accessor_references.hpp
```diff
@@ -0,0 +1,469 @@
...
+ * Reference class for a different storage than arithmetic type with the
+ * addition of a scaling factor. The conversion between both formats is done
+ * with a static_cast to the ArithmeticType, followed by a multiplication
+ * of the scalar (when reading; for writing, the new value is divided by the
+ * scalar before casting to the StorageType).
+ *
+ * Copying this reference is disabled, but move construction is possible to
+ * allow for an additional layer (like gko::range).
+ * The assignment operator only works for an rvalue reference (&&) to
+ * prevent accidental copying the reference and working on a reference.
```
pratikvn, line 349:
```suggestion
 * prevent accidentally copying and working on the reference.
```

core/base/accessor_helper.hpp
```diff
@@ -0,0 +1,578 @@
...
+    return detail::row_major_helper_s<ValueType, total_dim>::compute(
+        size, stride, std::forward<Indices>(idxs)...);
+}
+
+
+namespace detail {
+
+
+/**
+ * This helper walkes through the index arguments from left to right (lowest to
```
pratikvn, line 127:
```suggestion
 * This helper walks through the index arguments from left to right (lowest to
```

pratikvn (approved):
Only some very minor nits. LGTM!

include/ginkgo/core/base/dim.hpp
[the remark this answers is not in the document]
pratikvn: This was in reference to the comment above, but as you mention there, because of the extra storage I guess it is unncessary.

upsj (approved):
LGTM! Nice job, especially juggling all the templates 😄

PR #646, opened 2020-10-09
Improve identity constructor and treatment of rectangular matrices

### LINKED ISSUE #645: Problematic interaction with implicit conversion gko::dim
I just ran into a really nasty usability issue which is something between an interface inconsistency for Identity and an effect of the implicit conversion gko::dim -> bool:

You write
```
gko::matrix::Csr<...>::create(exec, size);
```
to get a matrix of the given size. If you try the same for Identity
```
gko::matrix::Identity<...>::create(exec, size);
```
you get a 1x1 matrix instead. The reason for this is that Identity expects a size_type, not a gko::dim.
So since gko::dim is non-zero, it gets implicitly converted to bool (true) and then to size_type (1)

Is this something we would expect? To fix this, would it be better to remove the bool overload (which I would prefer, since this kind of implicit conversion from a 2D object to a 1D object should not happen) or change the interface to Identity?

upsj:
This PR adds a gko::dim<2> constructor for Identity and adds checks for square matrices into all solver constructors, which avoids opaque errors from within Solver::apply_impl by failing early.

Fixes #645

### DIFF
--- a/benchmark/utils/overhead_linop.hpp
+++ b/benchmark/utils/overhead_linop.hpp
@@ -208,7 +208,7 @@ protected:
                 parameters_.preconditioner->generate(system_matrix_));
         } else {
             set_preconditioner(matrix::Identity<ValueType>::create(
-                this->get_executor(), this->get_size()[0]));
+                this->get_executor(), this->get_size()));
         }
         stop_criterion_factory_ =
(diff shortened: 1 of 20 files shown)


pratikvn (approved):
LGTM!

tcojean (approved):

PR #647, opened 2020-10-13
Fix cuSPARSE CsrEx benchmark memory leak

upsj:
The CsrEx wrapper allocates memory in each `apply` operation, but only frees the last allocation in the destructor.
On the one hand, this is a memory leak, on the other hand, it adds allocation overheads to the benchmark also after the warmup runs.

This PR fixes the issues by using Array to only allocate once (for fixed size)

### DIFF
--- a/benchmark/utils/cuda_linops.hpp
+++ b/benchmark/utils/cuda_linops.hpp
@@ -328,21 +328,6 @@ public:
         return csr_->get_num_stored_elements();
     }
 
-    ~CuspCsrEx() override
-    {
-        const auto id = this->get_gpu_exec()->get_device_id();
-        if (set_buffer_) {
-            try {
-                gko::cuda::device_guard g{id};
-                GKO_ASSERT_NO_CUDA_ERRORS(cudaFree(buffer_));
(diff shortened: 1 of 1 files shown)


benchmark/utils/cuda_linops.hpp
```diff
@@ -370,14 +355,13 @@ class CuspCsrEx
             csr_->get_num_stored_elements(), &alpha, this->get_descr(),
             csr_->get_const_values(), csr_->get_const_row_ptrs(),
             csr_->get_const_col_idxs(), db, &beta, dx, &buffer_size);
-        GKO_ASSERT_NO_CUDA_ERRORS(cudaMalloc(&buffer_, buffer_size));
-        set_buffer_ = true;
+        buffer_.resize_and_reset(buffer_size);
```
yhmtsai, line 358: If buffersize does not depend on b/x data, putting it in the read stage may be reasonable.

thoasm (approved):
LGTM!

tcojean (approved):

yhmtsai (approved):

PR #649, opened 2020-10-16
fix omp local header and reorder_factory

yhmtsai:
This PR fixes the compile issue in Windows and MacOS
1. can not find the local header in MacOS
2. error inheriting constructors must be inherited from a direct base class in Windows (only cuda)


tcojean (approved):

pratikvn (approved):
Thanks for fixing this!

upsj (approved):

PR #650, opened 2020-10-16
Make gko::dim conversion to bool explicit

upsj:
Follow up to #646:

Currently, gko::dim is implicitly convertible to bool. We could avoid the gko::dim -> bool -> int conversion chain by making `operator bool` explicit. C++ allows for so-called contextual conversions to also use explicit conversion operators.
`<expression>` is contextually converted to bool in the following cases:

1. `if(<expression>), while(<expression>), for(...;`<expression>; ...)`
2. `!<expression>, ... && <expression>, || <expression>`
3. `<expression> ? ... : ...`
3. `static_assert/noexcept(<expression>)`

So these are all places where gko::dim is used directly as a boolean. Any other use looks to me like a bug waiting to happen. In this sense:
Is this interface-breaking?

### DIFF
--- a/include/ginkgo/core/base/dim.hpp
+++ b/include/ginkgo/core/base/dim.hpp
@@ -118,8 +118,11 @@ struct dim {
      * different than zero.
      *
      * @return true if and only if all dimensions evaluate to true
+     *
+     * @note This operator is explicit to avoid implicit dim-to-int casts.
+     *       It will still be used in contextual conversions (if, &&, ||, !)
      */
-    constexpr GKO_ATTRIBUTES operator bool() const
+    explicit constexpr GKO_ATTRIBUTES operator bool() const
     {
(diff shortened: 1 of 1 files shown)


pratikvn:
well-yes-butno

I think we have dont this kind of soft interface break before. If someone is using this bug as a feature, then I think it is better to break their interface to correct their usage

tcojean (approved):
I think that's a good approach to ensuring no further bugs appear from this. [Relevant godbolt](https://godbolt.org/z/6E6hhE).

On the other end, we have another operator bool in the same file, you probably want to apply the same treatment?

tcojean:
A bug isn't a feature.

pratikvn (approved):

upsj:
@tcojean Good catch! I added the second `explicit` as well.

yhmtsai (approved):
LGTM

PR #651, opened 2020-10-16
Fix BiCG and HIP CSR to work with complex matrices.

pratikvn:
This PR fixes a minor bug with the BiCG solver where we were using the normal transpose instead of the Hermitian transpose for the shadow space generation for BiCG. 
Some additional documentation is also added, along with a reference. 

A bug within the HIP CSR conj_transpose kernel is also fixed where the parameters were reversed, along with a test added for verification.

### DIFF
--- a/core/solver/bicg.cpp
+++ b/core/solver/bicg.cpp
@@ -95,7 +95,7 @@ std::unique_ptr<LinOp> Bicg<ValueType>::conj_transpose() const
  *                  before transposing it
  */
 template <typename CsrType>
-std::unique_ptr<LinOp> transpose_with_csr(const LinOp *mtx)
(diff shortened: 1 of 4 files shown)


upsj (approved):
LGTM!

core/solver/bicg.cpp
```diff
@@ -95,15 +95,15 @@ std::unique_ptr<LinOp> Bicg<ValueType>::conj_transpose() const
  *                  before transposing it
```
pratikvn, line 95:
```suggestion
 *                  before (conjugate-)transposing it
```

tcojean (approved):

hartwiganzt (approved):

PR #652, opened 2020-10-19
Add global constant and different executors on the same device do not need memory movement

yhmtsai:
This PR makes different executors on the same device do not need memory movement and add global constant.

- [x] executor can directly use the memory from other executors if they are on the same device (need the same Executor-type)
- [ ] allocate and deallocate in executor not global


include/ginkgo/core/base/executor.hpp
```diff
@@ -862,6 +908,17 @@ class OmpExecutor : public detail::ExecutorBase<OmpExecutor>,
...
+    GKO_DEFAULT_OVERRIDE_VERIFY_MEMORY_TO(ReferenceExecutor);
+
+    GKO_DEFAULT_OVERRIDE_VERIFY_MEMORY_TO(CudaExecutor);
+
+    GKO_DEFAULT_OVERRIDE_VERIFY_MEMORY_TO(HipExecutor);
+
+    bool verify_memory_to(const OmpExecutor *dest_exec) const override
+    {
+        return true;
+    }
```
tcojean, line 921: Maybe also for reference?

include/ginkgo/core/base/executor.hpp
```diff
@@ -895,6 +952,17 @@ class ReferenceExecutor : public OmpExecutor {
...
+    GKO_DEFAULT_OVERRIDE_VERIFY_MEMORY_TO(OmpExecutor);
+
+    GKO_DEFAULT_OVERRIDE_VERIFY_MEMORY_TO(CudaExecutor);
+
+    GKO_DEFAULT_OVERRIDE_VERIFY_MEMORY_TO(HipExecutor);
+
+    bool verify_memory_to(const ReferenceExecutor *dest_exec) const override
+    {
+        return true;
+    }
```
tcojean, line 964: Also for OpenMP?

include/ginkgo/core/base/executor.hpp
```diff
@@ -1027,6 +1095,17 @@ class CudaExecutor : public detail::ExecutorBase<CudaExecutor>,
 
     GKO_ENABLE_FOR_ALL_EXECUTORS(GKO_OVERRIDE_RAW_COPY_TO);
 
+    GKO_DEFAULT_OVERRIDE_VERIFY_MEMORY_TO(OmpExecutor);
+
+    GKO_DEFAULT_OVERRIDE_VERIFY_MEMORY_TO(HipExecutor);
```
tcojean, line 1100: `#if defined(GINKGO_HIP_PLATFORM_NVCC)` and both have the same id I would also return true.

include/ginkgo/core/base/executor.hpp
```diff
@@ -1193,6 +1272,17 @@ class HipExecutor : public detail::ExecutorBase<HipExecutor>,
 
     GKO_ENABLE_FOR_ALL_EXECUTORS(GKO_OVERRIDE_RAW_COPY_TO);
 
+    GKO_DEFAULT_OVERRIDE_VERIFY_MEMORY_TO(ReferenceExecutor);
+
+    GKO_DEFAULT_OVERRIDE_VERIFY_MEMORY_TO(OmpExecutor);
+
+    GKO_DEFAULT_OVERRIDE_VERIFY_MEMORY_TO(CudaExecutor);
```
tcojean, line 1279: Same here, depending on `GINKGO_HIP_PLATFORM_NVCC` and the ID.

tcojean:
Some comments here: 
I believe we should close this PR and move to #670. When implementing the global constant, I realized it's very complex code to make it work, and it would instead be very easy to implement with the concept of memory space, as that is exactly what a natural implementation of this feature would be. I would rather put this on hold until we have that, otherwise any implementation of this feature tries to provide something like memory spaces without properly integrating it into the executors to not create an interface breaking change...

PR #653, opened 2020-10-19
Add an index set class.

pratikvn:
This PR adds an index set class that stores subsets (contiguous or non-contiguous) by storing only their end-points. This is very useful for representing the indices of objects stored by different entities and is in preparation for distributed Ginkgo.

### TODO
+ [ ] Add GPU kernels and allow index sets to be operated on CUDA and HIP


pratikvn:
format!

pratikvn:
format!

pratikvn:
Closing due to the needed rework.

PR #654, opened 2020-10-20
Add a memory space abstraction

pratikvn:
This PR is a draft PR which showcases how a memory space abstraction would look like in Ginkgo. All the memory operations such as `alloc`, `free`  and `raw_copy_to`'s are move from the Executor to the MemorySpace class.

By default, Executors are created with separate new memory space objects thereby trying to maintain the same behaviour as before this PR. Executors can also be created with an associated memory space. In that case, with matching memory spaces, the data would not be copied. 

This PR is experimental and is available for anyone who wants to discuss and play around with the code. 

Please feel free to point out problems and issues with the current implementation and suggestions for improvements are welcome.

### TODO

+ [x] Add documentation.
+ [ ] Check whether polymorphic_object can also be completely abstracted.

Update: 
Interface breaking change, See #402

### DIFF
--- a/benchmark/utils/loggers.hpp
+++ b/benchmark/utils/loggers.hpp
@@ -48,41 +48,43 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 // A logger that accumulates the time of all operations
 struct OperationLogger : gko::log::Logger {
-    void on_allocation_started(const gko::Executor *exec,
+    void on_allocation_started(const gko::MemorySpace *mem_space,
                                const gko::size_type &) const override
     {
-        this->start_operation(exec, "allocate");
+        this->start_operation(mem_space, "allocate");
     }
 
-    void on_allocation_completed(const gko::Executor *exec,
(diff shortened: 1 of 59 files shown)


pratikvn:
Note: After thinking about this I decided that it does not really make sense to try to force this into a non-interface breaking addition. There are a couple of things that are problematic:

1. The Loggers `on_*` memory logging methods currently take `Executor *` and those will need to be updated. For a non-interface breaking addition, I can only add additional methods which take `MemorySpace *`, which considerably bloats the code.

2. The `alloc` , `free` and `copy_from` methods are in the public interface of the `Executor` class and hence I cannot simply move them to the `MemorySpace` class but will have to call the memory space methods from within the `Executor` methods, which kind of defeats the purpose in a  certain sense.  

So, I think maybe it is indeed better to wait to add this until 2.0.0. But I guess we could possibly merge this into the `ginkgo-2.0.0` branch.

pratikvn:
To reduce the number of PR's, I will close this for now as we dont want really long standing PR's anyway. We can open another one later maybe after 2.0

PR #655, opened 2020-10-20
Apply real matrices to complex vectors

fritzgoebel:
This PR adds the possibility to apply real matrices to complex vectors.

Note that in the apply, instead of using `this`, I need to use `as<Coo<remove_complex<ValueType>, IndexType>>(this)` (Or similar for the other formats).
This is because without doing this, the compiler does not realize that at this point in the code, the ValueType cannot be complex anymore and therefore complains that there's no implementation of the spmv with a complex matrix and real vectors.

### DIFF
--- a/core/matrix/coo.cpp
+++ b/core/matrix/coo.cpp
@@ -77,8 +77,18 @@ GKO_REGISTER_OPERATION(outplace_absolute_array,
 template <typename ValueType, typename IndexType>
 void Coo<ValueType, IndexType>::apply_impl(const LinOp *b, LinOp *x) const
 {
-    using Dense = Dense<ValueType>;
-    this->get_executor()->run(coo::make_spmv(this, as<Dense>(b), as<Dense>(x)));
+    using ComplexDense = Dense<to_complex<ValueType>>;
+
+    if (dynamic_cast<const Dense<ValueType> *>(b)) {
+        this->get_executor()->run(coo::make_spmv(this, as<Dense<ValueType>>(b),
+                                                 as<Dense<ValueType>>(x)));
+    } else {
+        auto dense_b = as<ComplexDense>(b);
+        auto dense_x = as<ComplexDense>(x);
+        this->get_executor()->run(coo::make_spmv(
+            as<Coo<remove_complex<ValueType>, IndexType>>(this),
+            dense_b->view_as_real().get(), dense_x->view_as_real().get()));
+    }
 }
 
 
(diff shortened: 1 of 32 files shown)


core/matrix/coo.cpp
```diff
@@ -77,37 +77,82 @@ GKO_REGISTER_OPERATION(outplace_absolute_array,
...
+
+    if (dynamic_cast<const Dense<ValueType> *>(b)) {
+        this->get_executor()->run(coo::make_spmv(this, as<Dense<ValueType>>(b),
+                                                 as<Dense<ValueType>>(x)));
+    } else {
+        auto dense_b = as<ComplexDense>(b);
+        auto dense_x = as<ComplexDense>(x);
+        this->get_executor()->run(coo::make_spmv(
+            as<Coo<remove_complex<ValueType>, IndexType>>(this),
+            dense_b->view_as_real().get(), dense_x->view_as_real().get()));
```
upsj, line 90: You could also write `this->apply(dense_b->view_as_real().get(), dense_x->view_as_real().get())`, not sure this is the cleanest solution, though. We should also keep in mind what the error case would be here, so what kind of error message does the user get when this fails due to the wrong type? Before, they were warned about ValueType, suddenly they are being warned about `to_complex<ValueType>`, might be slightly confusing.

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -441,6 +441,49 @@ class Dense
         return create_submatrix(rows, columns, this->get_stride());
     }
 
+    /*
+     * Create a real view of the (potentially) complex original matrix.
+     */
+    std::unique_ptr<Dense<remove_complex<ValueType>>> view_as_real()
```
upsj, line 447: Just the obligatory naming discussion: Since we already have a `create_submatrix`, how about naming this `create_real_view` with a more "imperative" base?

upsj (approved):
LGTM in general!

core/matrix/coo.cpp
fritzgoebel: I like it!

cuda/test/matrix/coo_kernels.cpp
```diff
@@ -231,6 +233,54 @@ TEST_F(Coo, AdvancedApplyAddToLargeDenseMatrixIsEquivalentToRef)
...
+    auto complex_x = gen_mtx<ComplexVec>(532, 3);
+    auto dcomplex_x = ComplexVec::create(cuda);
+    dcomplex_x->copy_from(complex_x.get());
+
+    mtx->apply(complex_b.get(), complex_x.get());
+    dmtx->apply(dcomplex_b.get(), dcomplex_x.get());
+
+    GKO_ASSERT_MTX_NEAR(dcomplex_x, complex_x, 1e-14);
+}
+
```
pratikvn, line 251: nit: All of them have only one new line. Can you add one more newline between the different tests ? This file and all the other files ?

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -441,6 +441,50 @@ class Dense
         return create_submatrix(rows, columns, this->get_stride());
     }
 
+    /*
+     * Create a real view of the (potentially) complex original matrix.
```
pratikvn, line 445: Can add a couple of lines saying that you are viewing the complex matrix with a `reinterpret_cast` with stride 2*original_stride ?

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -441,6 +441,50 @@ class Dense
...
+            this->get_executor(), dim<2>{num_rows, num_cols},
+            Array<remove_complex<ValueType>>::view(
+                this->get_executor(), num_rows * stride,
+                reinterpret_cast<remove_complex<ValueType> *>(
+                    this->get_values())),
+            stride);
+    }
+
+    /*
+     * Create a constant real view of the (potentially) complex original matrix.
```
pratikvn, line 466: You can use `copydoc` to copy from the previous doc

pratikvn (approved):
Some doc and new line nits. Otherwise looks good.

PR #658, opened 2020-10-26
Dense: conversion to complex and extraction of real and imaginary part

fritzgoebel:
This PR enables dense to:
- convert real to complex: add all zero imaginary part
- complex to complex: return a copy
- extract the real part: if real, return a copy, if complex, return the real part
- extract the imaginary part: if real, return an all zero matrix, if complex return the imaginary part.


cuda/test/matrix/dense_kernels.cpp
```diff
@@ -609,4 +609,37 @@ TEST_F(Dense, OutplaceAbsoluteMatrixIsEquivalentToRef)
...
 
 
+TEST_F(Dense, MakeComplexIsEquivalentToRef)
+{
+    set_up_apply_data();
+
+    auto complex_x = x->make_complex();
+    auto dcomplex_x = dx->make_complex();
+
+    GKO_ASSERT_MTX_NEAR(complex_x, dcomplex_x, 1e-14);
```
upsj, line 619: Could you try testing for exact equality? We shouldn't be doing any conversions here.
```suggestion
    GKO_ASSERT_MTX_NEAR(complex_x, dcomplex_x, 0.0);
```

upsj (approved):
LGTM!

cuda/test/matrix/dense_kernels.cpp
```diff
@@ -609,4 +609,37 @@ TEST_F(Dense, OutplaceAbsoluteMatrixIsEquivalentToRef)
...
+
+
+TEST_F(Dense, GetRealIsEquivalentToRef)
+{
+    set_up_apply_data();
+
+    auto real_x = x->get_real();
+    auto dreal_x = dx->get_real();
+
+    GKO_ASSERT_MTX_NEAR(real_x, dreal_x, 1e-14);
```
yhmtsai, line 630: also here
```suggestion
    GKO_ASSERT_MTX_NEAR(real_x, dreal_x, 0);
```

cuda/test/matrix/dense_kernels.cpp
```diff
@@ -609,4 +609,37 @@ TEST_F(Dense, OutplaceAbsoluteMatrixIsEquivalentToRef)
...
+
+
+TEST_F(Dense, GetImagIsEquivalentToRef)
+{
+    set_up_apply_data();
+
+    auto imag_x = x->get_imag();
+    auto dimag_x = dx->get_imag();
+
+    GKO_ASSERT_MTX_NEAR(imag_x, dimag_x, 1e-14);
```
yhmtsai, line 641:
```suggestion
    GKO_ASSERT_MTX_NEAR(imag_x, dimag_x, 0);
```

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -253,6 +254,12 @@ class Dense
 
     void compute_absolute_inplace() override;
 
+    std::unique_ptr<complex_type> make_complex() const;
+
+    std::unique_ptr<absolute_type> get_real() const;
+
+    std::unique_ptr<absolute_type> get_imag() const;
```
yhmtsai, line 261: Need the documentation because they are the first declaration.

yhmtsai (changes requested):
I forget the downside point if Convertible<to_complex<Dense>> and Convertible<to_real<Dense>>.
Does it destroy the default convert_to?
and maybe possible another operation is `make_imag_complex` for real->imag and complex->imag(complex type)

upsj:
@yhmtsai The issue here is that the conversion is not symmetric: `double -> complex<double>` is unambiguous, but `complex<double> -> double` has two candidates (real and imag), so we can't just say `complex->convert_to(double)`, which is the only interface `ConvertibleTo` provides. We also can't easily have just a one-sided conversion `double -> complex`, since this would require us to conditionally inherit from `ConvertibleTo` and conditionally override `convert_to` based on whether `ValueType` is complex.

core/matrix/dense.cpp
```diff
@@ -795,6 +798,81 @@ Dense<ValueType>::compute_absolute() const
...
+    exec->run(dense::make_make_complex(this, complex_dense.get()));
+
+    return complex_dense;
+}
+
+
+template <typename ValueType>
+void Dense<ValueType>::make_complex(Dense<to_complex<ValueType>> *result) const
+{
+    auto exec = this->get_executor();
```
upsj, line 818: `result` doesn't necessarily reside on the same executor as `this`, so in case the executors mismatch, you may need to do something like `result->copy_from(lend(this->get_real())`

upsj (changes requested):

core/matrix/dense.cpp
pratikvn: or a `make_temporary_clone`

upsj (approved):

pratikvn (approved):
LGTM!

common/matrix/dense_kernels.hpp.inc
```diff
@@ -539,4 +539,48 @@ __global__ __launch_bounds__(default_block_size) void outplace_absolute_dense(
 }
 
 
+template <typename ValueType, typename ComplexType>
+__global__ __launch_bounds__(default_block_size) void make_complex(
+    size_type num_rows, size_type num_cols, const ValueType *__restrict__ in,
+    size_type stride_in, ComplexType *__restrict__ out, size_type stride_out)
```
yhmtsai, line 545: If this ComplexType is only for complex<ValueType>, I prefer use one template<typename ValueType> and use to_complex<ValueType> to represent complex<ValueType>. 
Before Ginkgo 2.0.0, I think we need to discuss the precision(ValueType) conversion...

reference/test/matrix/dense_kernels.cpp
```diff
@@ -65,6 +65,8 @@ class Dense : public ::testing::Test {
 protected:
     using value_type = T;
     using Mtx = gko::matrix::Dense<value_type>;
+    using ComplexMtx = gko::matrix::Dense<gko::to_complex<value_type>>;
+    using RealMtx = gko::matrix::Dense<gko::remove_complex<value_type>>;
```
yhmtsai, line 69: These two can also use `gko::to_complex<Mtx>` and `gko::remove_complex<Mtx>`

yhmtsai (approved):
LGTM in general, some minor things are in the comments.
After rebasing, you will need to add these functionalities GKO_NOT_IMPLEMENTED in dpcpp kernel.

common/matrix/dense_kernels.hpp.inc
upsj: This doesn't work, since to_complex returns the CPU complex type, not thrust::complex. An alternative would be `thrust::complex<remove_complex<ValueType>>`

yhmtsai: you are right. maybe does another way `cuda_type<to_complex<ValueType>>` work?

dpcpp/matrix/dense_kernels.dp.cpp
```diff
@@ -308,6 +308,33 @@ void outplace_absolute_dense(std::shared_ptr<const DpcppExecutor> exec,
...
+void get_real(std::shared_ptr<const DpcppExecutor> exec,
+              const matrix::Dense<ValueType> *source,
+              matrix::Dense<remove_complex<ValueType>> *result)
+    GKO_NOT_IMPLEMENTED;
+
+GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE(GKO_DECLARE_GET_REAL_KERNEL);
+
+
+template <typename ValueType>
+void get_imag(std::shared_ptr<const CudaExecutor> exec,
```
upsj, line 330:
```suggestion
void get_imag(std::shared_ptr<const DpcppExecutor> exec,
```

PR #659, opened 2020-10-30
Add sorting option to both Jacobi preconditioner and the ILU factorization

### LINKED ISSUE #633: [MFEM example] Wrong result when change preconditioner in example MFEM using Ginkgo
Hi!

When I run ex1 (https://github.com/mfem/mfem/blob/master/examples/ginkgo/ex1.cpp) by reference excutor and cuda excutor, it's ok. But when I change preconditioner (Jacobi preconditioner), I ran by reference is ok. But with cuda excutor the result is wrong, it does not converge.

I think it's a error. You can try it and see.

Thank you!

thoasm:
This PR adds the option to sort the matrix before processing it inside the Jacobi preconditioner and the ILU factorization.
By default, it is assumed that the input matrix is not sorted, therefore, a sorting step is performed (unless the factory parameter `skip_sorting` is set to `true`).
Additionally, this PR adds functionality to unsort a given matrix for tests (currently only CSR and COO are supported). Thus functionality is then used in added tests that unsort the system matrix before generating / running for CSR, COO, Jacobi preconditioner and ILU factorization.

Closes #633


core/preconditioner/jacobi.cpp
```diff
@@ -201,12 +204,26 @@ void Jacobi<ValueType, IndexType>::detect_blocks(
...
+    decltype(copy_and_convert_to<csr_type>(exec, system_matrix)) csr_mtx{};
+
+    if (skip_sorting) {
+        csr_mtx = copy_and_convert_to<matrix::Csr<ValueType, IndexType>>(
+            exec, system_matrix);
+    } else {
+        auto editable_csr = csr_type::create(exec);
+        as<ConvertibleTo<csr_type>>(system_matrix)
+            ->convert_to(lend(editable_csr));
+        // Maybe check if it is sorted first
```
tcojean, line 222: I guess that should be a `TODO` note?

core/test/utils/unsort_matrix_test.cpp
```diff
@@ -0,0 +1,163 @@
...
+{
+    auto csr = Csr::create(exec);
+    mtx->convert_to(gko::lend(csr));
+    bool was_sorted = csr->is_sorted_by_column_index();
+
+    gko::test::unsort_matrix(gko::lend(csr), rand_engine);
+
+    ASSERT_FALSE(csr->is_sorted_by_column_index());
+    ASSERT_TRUE(was_sorted);
+    GKO_ASSERT_MTX_NEAR(csr, mtx, 0.);
```
tcojean, line 119: As the CI output says, since this is the core module you technically cannot use anything which calls a kernel, which I believe multiple parts in what I highlighted do. You should move all of this test to the reference module.

core/test/utils/unsort_matrix_test.cpp
```diff
@@ -0,0 +1,163 @@
...
+                                      {0, 0, 7, 0, 0},
+                                      {0, 0, 8, 9, 10}},
+                                     exec)),
+          empty(Dense::create(exec, gko::dim<2>(0, 0)))
+    {}
+
+    bool is_coo_matrix_sorted(Coo *mtx)
+    {
+        auto size = mtx->get_size();
+        auto vals = mtx->get_values();
```
tcojean, line 79: I don't think you use these?

core/test/utils/unsort_matrix.hpp
```diff
@@ -0,0 +1,153 @@
...
+    const auto &master = mtx->get_executor()->get_master();
+
+    if (exec != master) {
+        auto h_mtx = mtx->clone(master);
+        unsort_matrix(lend(h_mtx), engine);
+        mtx->copy_from(lend(h_mtx));
+        return;
+    }
+
+    auto size = mtx->get_size();
```
tcojean, line 128: I don't think you use this one.

tcojean (approved):
LGTM.

core/preconditioner/jacobi.cpp
thoasm: I was considering to check it beforehand, but I believe that just performing a sorting is likely faster than checking and sorting afterwards. I removed this comment.

core/test/utils/unsort_matrix_test.cpp
thoasm: Thanks, I implemented the check for a CSR matrix in this test myself to get around that limitation.

tcojean: Is this enough? Don't the conversion (from dense to CSR, here), also call kernels?

core/test/utils/unsort_matrix.hpp
```diff
@@ -0,0 +1,168 @@
...
+            const auto new_master = master->get_master();
+            if (new_master == master) {
+                actual_master_found = true;
+                break;
+            }
+            master = new_master;
+        }
+        if (!actual_master_found) {
+            GKO_NOT_SUPPORTED(exec);
+        }
```
pratikvn, line 81: Maybe I am missing something here, are there any executors with depth > 1 ? I think for all executors `exec->get_master()` should directly give you the master, right ?

pratikvn (approved):
LGTM!

core/test/utils/unsort_matrix.hpp
thoasm: Usually, yes, but there is also nothing preventing us from doing something like:
`auto exec = gko::CudaExecutor::create(0, gko::CudaExecutor::create(0, gko::OmpExecutor::create()))`.
We don't guarantee that `get_master()` is actually a CPU executor.

tcojean: That should not happen, the `get_master` should indeed directly be the CPU side executor.

thoasm: I tried it inside an example, I can create such a chain as an executor, but the example fails during runtime (but it compiles without any issue).
Do you want me to remove this part since it should not be necessary? (The current implementation should work either way.)

pratikvn: Yes, I think it is better to remove it. I dont think it makes sense to create an executor whose `get_master()` would be a non-CPU executor, as Terry mentions.

core/preconditioner/jacobi.cpp
```diff
@@ -201,12 +204,25 @@ void Jacobi<ValueType, IndexType>::detect_blocks(
...
-void Jacobi<ValueType, IndexType>::generate(const LinOp *system_matrix)
+void Jacobi<ValueType, IndexType>::generate(const LinOp *system_matrix,
+                                            bool skip_sorting)
 {
     GKO_ASSERT_IS_SQUARE_MATRIX(system_matrix);
+    using csr_type = matrix::Csr<ValueType, IndexType>;
     const auto exec = this->get_executor();
-    const auto csr_mtx = copy_and_convert_to<matrix::Csr<ValueType, IndexType>>(
-        exec, system_matrix);
+    decltype(copy_and_convert_to<csr_type>(exec, system_matrix)) csr_mtx{};
```
upsj, line 213: nit: Wouldn't it be shorter to be explicit here, i.e., `std::unique_ptr<csr_type>`?

core/test/utils/unsort_matrix.hpp
```diff
@@ -0,0 +1,140 @@
...
+            ++i;
+        }
+        current_row = rows[i];
+        auto end = i;
+        auto iterator = gko::detail::IteratorFactory<IndexType, ValueType>(
+            cols + start, vals + start, end - start);
+        // since the row entries are supposed to be the same, there is no need
+        // to swap
+        std::shuffle(iterator.begin(), iterator.end(), engine);
+    }
```
upsj, line 133: We could maybe simplify this into a std::shuffle followed by a std::stable_sort by only the first index? Might make the intent and correctness clearer?

core/preconditioner/jacobi.cpp
```diff
@@ -201,12 +204,25 @@ void Jacobi<ValueType, IndexType>::detect_blocks(
...
+        csr_mtx = copy_and_convert_to<matrix::Csr<ValueType, IndexType>>(
+            exec, system_matrix);
+    } else {
+        auto editable_csr = csr_type::create(exec);
+        as<ConvertibleTo<csr_type>>(system_matrix)
+            ->convert_to(lend(editable_csr));
+        editable_csr->sort_by_column_index();
+        csr_mtx = decltype(csr_mtx){editable_csr.release(),
+                                    std::default_delete<const csr_type>{}};
+    }
```
upsj, line 225: Refactoring suggestion: We have this type of scheme (convert to CSR, sort if requested, conditional ownership) in 5 places now: Jacobi, Ilu, Isai, ParIlu, ParIlut, ParIct. Do you think we can find a common interface we could use to extract this into a helper function?

core/test/utils/unsort_matrix.hpp
```diff
@@ -0,0 +1,140 @@
...
+
+    auto vals = mtx->get_values();
+    auto row_ptrs = mtx->get_row_ptrs();
+    auto cols = mtx->get_col_idxs();
+
+    for (index_type row = 0; row < size[0]; ++row) {
+        auto start = row_ptrs[row];
+        auto end = row_ptrs[row + 1];
+        auto iterator = gko::detail::IteratorFactory<IndexType, ValueType>(
```
upsj, line 82: This keeps on coming in handy, nice job! :smile: Small nit: this is not technically an iterator, but something like a range returning an iterator adapter, so the variable naming feels slightly off.

cuda/test/matrix/csr_kernels.cpp
```diff
@@ -219,6 +194,18 @@ TEST_F(Csr, SimpleApplyIsEquivalentToRefWithLoadBalance)
...
+    set_up_apply_data(std::make_shared<Mtx::load_balance>(cuda));
+    unsort_mtx();
+
+    mtx->apply(y.get(), expected.get());
+    dmtx->apply(dy.get(), dresult.get());
+
+    GKO_ASSERT_MTX_NEAR(dresult, expected, 1e-14);
+}
+
+
```
upsj, line 208: Question: Do we at any point make any assumptions about the column index order? It looks to me like we only use them to load the rhs, so I don't see how this could fail if the sorted one succeeds

benchmark/solver/solver.cpp
```diff
@@ -184,7 +184,9 @@ const std::map<std::string, std::function<std::unique_ptr<gko::LinOpFactory>(
         {"jacobi",
          [](std::shared_ptr<const gko::Executor> exec) {
              std::shared_ptr<const gko::LinOpFactory> f =
-                 gko::preconditioner::Jacobi<>::build().on(exec);
+                 gko::preconditioner::Jacobi<>::build()
+                     .with_skip_sorting(true)
```
upsj, line 188: I don't entirely agree with this change - we don't skip sorting for the other preconditioners, and as far as I can tell, we don't sort the input prior to passing it to the preconditioner, so if the input `.mtx` file is sorted incorrectly, the preconditioner will silently fail.

upsj (approved):
LGTM! Only some simplification potential and a small discussion point.

benchmark/solver/solver.cpp
thoasm: When reading a matrix from file, we always sort it by calling [`ensure_row_major_order()`](https://github.com/ginkgo-project/ginkgo/blob/develop/include/ginkgo/core/base/matrix_data.hpp#L463) inside [the `read()` function](https://github.com/ginkgo-project/ginkgo/blob/develop/core/base/mtx_io.cpp#L99). So whenever we read from a file, we can be sure  that we get a sorted input (which is why I was only able to reproduce the MFEM error inside MFEM and not when exporting it into a file, even though the file was unsorted).

upsj: Ah, thanks for correcting me - I was looking for `sort` in the IO file. `ensure_row_major_order` feels a bit inconsistent in that regard, but that's a point for another time 😄

core/preconditioner/jacobi.cpp
thoasm: No because it is not a default `unique_ptr`. `unique_ptr<csr_type, std::function<void(R*)>>` would be accurate, but in my opinion more complicated (and potentially loses the information why we need this type). That is the reason why I use `decltype`.

cuda/test/matrix/csr_kernels.cpp
thoasm: No, I don't think we do that, but I think it is always good to check if it still works.
Additionally, it is possible that we add another implementation where we do. In that case, it is more likely that we add an unsorted test for that as well since we see that it is tested for the others.

core/preconditioner/jacobi.cpp
upsj: I see. In case we had a separate function taking care of the copy and/or sorting, we could replace this by `auto`, right?

cuda/test/matrix/csr_kernels.cpp
upsj: I tend to disagree a bit there, since I can't imagine any benefit you gain from relying on the column order in an SpMV algorithm (which the cache hierarchy doesn't already provide you).

Anyways, the main question I wanted to ask here is: Do we need these changes in this PR, since they are unrelated to Jacobi and ILU? We don't test other parts of the implementation that might change depending on the column order either.

core/preconditioner/jacobi.cpp
thoasm: Yes, that is what I am working on right now (and I will definitively replace this with `auto` after I am done).

upsj: Perfect, thanks :)

core/test/utils/unsort_matrix.hpp
thoasm: Good idea, then we don't rely on the partially sorting of COO.

cuda/test/matrix/csr_kernels.cpp
thoasm: I can definitively remove it. We don't need it as part of this PR, i just added it to see if we depend on the sorting somewhere else.

core/test/utils/unsort_matrix.hpp
thoasm: I changed it a bit since I can't sort 3 values in different places according to just one of them (the self-written iterator helps only for 2 values at the same time).
Since performance is not important here (it is only for tests), I simply convert it to `matrix_data`, do the shuffling and sorting, and then write it to the matrix again.

core/test/utils/unsort_matrix.hpp
thoasm: Is `sort_wrapper` better? (I changed it to that.)

thoasm:
All comments should now be addressed. In summary, I:
- removed the reliance on kernels in the `unsort_matrix_test`
- added a helper `convert_to_with_sorting` that converts (and conditionally sorts) a given matrix
- use that helper in all factorizations and preconditioner where applicable
- improved the unsorting of the COO matrix
- rebased on top of `develop`

benchmark/solver/solver.cpp
```diff
@@ -170,6 +170,35 @@ const std::map<std::string, std::function<std::unique_ptr<gko::LinOpFactory>(
...
+
+protected:
+    std::unique_ptr<gko::LinOp> generate_impl(
+        std::shared_ptr<const gko::LinOp> op) const override
+    {
+        return base_factory->generate(gko::clone(exec, op));
+    }
+};
+
+
```
upsj, line 201: Something may have gone wrong during the rebase, I removed this on purpose, since all matrix format conversions are now implemented on the GPU

upsj (changes requested):

benchmark/solver/solver.cpp
thoasm: Yes, my bad, I did not pay close enough attention to the rebasing process.
Good catch!

upsj (approved):

PR #660, opened 2020-10-30
Improve solver and preconditioner benchmarks

upsj:
* Adds all ILU(T) factorizations with ISAI preconditioners to the solver and preconditioner benchmarks.
* Allows the user to specify the right-hand side to be used instead of a random vector
* Allows the user to specify whether the initial guess should be randomized
* Adds a GMRES restart/Krylov dimension flag.
* Consolidates the preconditioner setup and flags into a header file


tcojean (approved):
LGTM. 

Could you maybe also add some of these options to the `run_all_benchmarks.sh` script? The random RHS/initial guess in particular might be useful. That would also provide some documentation for the most convenient options.

benchmark/solver/solver.cpp
```diff
@@ -93,13 +94,14 @@ DEFINE_bool(overhead, false,
...
-              << "    { \"filename\": \"my_file.mtx\",  \"optimal\": { "
-                 "\"spmv\": \"<matrix format>\" } },\n"
-              << "    { \"filename\": \"my_file2.mtx\", \"optimal\": { "
-                 "\"spmv\": \"<matrix format>\" } }\n"
-              << "  ]" << std::endl;
+    std::cerr
+        << "Input has to be a JSON array of matrix configurations:\n"
+        << "  [\n"
+        << "    { \"filename\": \"my_file.mtx\",  \"optimal\": { "
+           "\"spmv\": \"<matrix format>\" }, \"rhs\": \"my_file_rhs.mtx\" },\n"
```
yhmtsai, line 101:
```suggestion
           "\"spmv\": \"<matrix format>\" },\n \"rhs\": \"my_file_rhs.mtx\" },\n"
```
small nit

benchmark/utils/preconditioners.hpp
```diff
@@ -0,0 +1,212 @@
...
+
+
+DEFINE_string(
+    preconditioners, "none",
+    "A comma-separated list of preconditioners to use. "
+    "Supported values are: none, jacobi, parict, parilu, parilut, ilu, "
+    "parict-isai, parilu-isai, parilut-isai, ilu-isai, overhead");
+
+DEFINE_uint32(parilu_iterations, 5,
+              "The number of iterations for ParICT/ParILU(T)");
```
yhmtsai, line 27:
```suggestion
              "The number of iterations for ParIC(T)/ParILU(T)");
```

yhmtsai (approved):
LGTM

benchmark/utils/preconditioners.hpp
upsj: This one is deliberate, since we don't have a ParIC implementation right now

PR #661, opened 2020-11-05
Fix new algorithm script and DPC++ flags.

tcojean:
Small PR with a few DPC++ related improvements.

+ Add DPC++ to the new algorithm script.
+ Extract DPC++ version in CMake
+ Add the proper flag for GPU complex upport.
+ Remove one of the issues from the customized DPC++
   `abs` implementation.


pratikvn (approved):
LGTM!

dpcpp/CMakeLists.txt
```diff
@@ -43,6 +46,10 @@ target_sources(ginkgo_dpcpp
 ginkgo_compile_features(ginkgo_dpcpp)
 
 set(GINKGO_DPCPP_FLAGS ${GINKGO_COMPILER_FLAGS} -fsycl)
+if (GINKGO_DPCPP_VERSION AND GINKGO_DPCPP_VERSION VERSION_GREATER_EQUAL 20201005)
```
yhmtsai, line 49: I think we will only support the first stable version not these beta afterwards?
maybe less than the version, throwing the error is enough.

yhmtsai (approved):
LGTM.
Could you also change the abs implementation?

dpcpp/CMakeLists.txt
tcojean: Since for now there are still the older versions on the devcloud, I thought maybe you would use them somehow. Since the flag didn't exist then, I'm not propagating it. That allows using any older version without having to apply a manual patch or something.

tcojean: In the end I moved everything to only use beta10. I also changed the math implementation. We still need to use `cl::sycl::sqrt` instead of plain `sqrt` AFAIK.

PR #662, opened 2020-11-06
Added localized warning ignore for GCC 9.x and above 

Slaedr:
for returning a local `initializer_list` object:

```
In file included from /home/aditya/projects/ginkgo/core/test/utils.hpp:49,
                 from /home/aditya/projects/ginkgo/core/test/matrix/identity.cpp:43:
/home/aditya/projects/ginkgo/core/test/utils/assertions.hpp: In instantiation of ‘std::initializer_list<std::initializer_list<_Tp> > gko::test::assertions::detail::l(std::initializer_list<std::initializer_list<_Tp> >) [with T = double]’:
/home/aditya/projects/ginkgo/core/test/matrix/identity.cpp:125:5:   required from ‘void {anonymous}::Identity_AppliesToMultipleVectors_Test<gtest_TypeParam_>::TestBody() [with gtest_TypeParam_ = std::complex<float>]’
/home/aditya/projects/ginkgo/core/test/matrix/identity.cpp:113:1:   required from here
/home/aditya/projects/ginkgo/core/test/utils/assertions.hpp:694:12: warning: returning local initializer_list variable ‘list’ does not extend the lifetime of the underlying array [-Winit-list-lifetime]
  694 |     return list;
      |            ^~~~
/home/aditya/projects/ginkgo/core/test/utils/assertions.hpp:692:53: note: declared here
  692 |     std::initializer_list<std::initializer_list<T>> list)
      |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~
/home/aditya/projects/ginkgo/core/test/utils/assertions.hpp: In instantiation of ‘T gko::test::assertions::detail::plain_ptr(T) [with T = std::initializer_list<std::initializer_list<double> >]’:
/home/aditya/projects/ginkgo/core/test/matrix/identity.cpp:125:5:   required from ‘void {anonymous}::Identity_AppliesToMultipleVectors_Test<gtest_TypeParam_>::TestBody() [with gtest_TypeParam_ = std::complex<float>]’
/home/aditya/projects/ginkgo/core/test/matrix/identity.cpp:113:1:   required from here
/home/aditya/projects/ginkgo/core/test/utils/assertions.hpp:724:12: warning: returning local initializer_list variable ‘ptr’ does not extend the lifetime of the underlying array [-Winit-list-lifetime]
  724 |     return ptr;
      |            ^~~
/home/aditya/projects/ginkgo/core/test/utils/assertions.hpp:722:15: note: declared here
  722 | T plain_ptr(T ptr)
```

I think this warning is spurious, since the underlying data we are returning is never created locally in the `l` functions. The pragmas help keep build output clean, and help for people using (Spac)Emacs to jump to errors. Thoughts?


upsj:
You make a good point in terms of making the build output less noisy - this is definitely something we should tackle!
I remember a while ago trying to fix the underlying issue, but I would need to take a look at the details again. Maybe passing by reference might be a first step.

pratikvn:
I think we discussed this some time back, as much as possible we would like to avoid ignoring warnings because in case there is a legitimate warning in the future, this may get ignored.

Slaedr:
Observations: GCC 9 gives this warning (and maybe GCC 10, not sure). nvcc with GCC 8, Clang 10 and Clang 12 do not.

@pratikvn I understand your concern; I don't really want to remove warnings either. But it's also easier to miss useful warnings when you have to sift through a number of false positives. Besides, the warning is only removed for those three specific functions (two `l`s and one `plain_ptr`). So if we can determine that the warning does not apply to these functions, maybe we should remove the warning just for them.

@upsj I'm wondering if these functions obey the standard. Going over [cppreference](https://en.cppreference.com/w/cpp/utility/initializer_list), I'm not completely sure, though I would guess they're fine.

upsj:
I tried out a fix for the underlying issue, which you can see in https://github.com/ginkgo-project/ginkgo/pull/663/commits/4e67674c011ca53cb6cfee8222e55e06c5c20b20. I believe this should not lead to any lifetime issues, since we only use `plain_ptr` and `l` on temporary initializer lists that live as long as the surrounding full-expression.

Slaedr:
Great, thanks! I was thinking of trying const references too, based on the warning message, but abandoned the idea because I don't entirely get the point of these functions and didn't want to risk it. If you're sure references are fine, maybe we can reject and close this PR. If I make the exact changes you made, I guess it should merge smoothly when I pull from you later.

upsj:
The `l` functions are there exclusively to help with template argument deduction: In templated tests, we need a way to transparently specify the ValueType which the `initializer_lists` should use. `plain_ptr` is used to uniformly use smart pointers, plain pointers and initializer lists in `GKO_ASSERT_MTX_NEAR`

PR #663, opened 2020-11-13
Improve and generalize permutation core

upsj:
This PR adds `Dense::row_gather` functions that generalize `row_permute` and improves the `core` functionality to also work with an `Array` from a different `Executor`. Also fixes the -Winit-list-lifetime warnings from GCC >= 9


include/ginkgo/core/base/array.hpp
```diff
@@ -563,6 +563,59 @@ class Array {
...
+    /**
+     * Creates a new deleter object.
+     *
+     * @param original  the origin object where the data will be copied before
+     *                  deletion
+     */
+    copy_back_deleter(pointer original) : original_{original} {}
+
+    /**
+     * Deletes the object.
```
tcojean, line 603:
```suggestion
     * Deletes the object after copying back to original.
```
?

include/ginkgo/core/base/utils.hpp
```diff
@@ -517,12 +530,14 @@ class temporary_clone {
...
-            handle_ = handle_type(ptr, [](pointer) {});
+            handle_ = handle_type(ptr, null_deleter<T>());
         } else {
             // clone the object to the new executor and make sure it's copied
             // back before we delete it
-            handle_ = handle_type(gko::clone(std::move(exec), ptr).release(),
-                                  copy_back_deleter<T>(ptr));
+            handle_ = handle_type(
+                temporary_clone_helper<T>::create(std::move(exec), ptr)
+                    .release(),
```
tcojean, line 539: Why using the helper here? It looks to me like clone already gave the proper type?

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -232,6 +232,16 @@ class Dense
     std::unique_ptr<LinOp> row_permute(
         const Array<int64> *permutation_indices) const override;
 
+    std::unique_ptr<Dense> row_gather(const Array<int32> *gather_indices) const;
```
tcojean, line 235: Since these are new and not an interface/override, documentation would be good. Also, I guess we expect to have this only with dense? Otherwise, how about creating an interface for this.

tcojean (commented):
Some small comments.

core/matrix/dense.cpp
```diff
@@ -646,38 +646,100 @@ std::unique_ptr<LinOp> Dense<ValueType>::row_permute(
...
-        dense::make_row_permute(permutation_indices, this, permute_cpy.get()));
+        dense::make_row_gather(make_temporary_clone(exec, row_indices).get(),
+                               this, row_gathered.get()));
+    return row_gathered;
+}
+
+
+template <typename ValueType>
+void Dense<ValueType>::row_gather(const Array<int32> *row_indices,
+                                  Dense<ValueType> *row_gathered) const
```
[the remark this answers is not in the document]
upsj: The reason I went for this interface is the return type: This returns `void`, the other function returns `std::unique_ptr<...>`. What would you return when `row_gathered` is not null?

include/ginkgo/core/base/utils.hpp
upsj: clone only works for polymorphic types, I need it to work for Array as well, where we involve a constructor instead of calling `object->copy_from`

include/ginkgo/core/matrix/dense.hpp
upsj: I only added the minimal interface I need to make distributed SpMV work, so far I couldn't imagine where we would use this for other matrix formats or columns instead of rows.

core/matrix/dense.cpp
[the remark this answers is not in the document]
tcojean: Ah, I missed that you were returning a unique pointer and that this is not.

include/ginkgo/core/matrix/dense.hpp
tcojean: Maybe block CSR could find some use in something related at some point for a distributed block ParILU of some sort? But I guess it wouldn't be row gather but block gather? I guess indeed there is little application to the same interface elsewhere.

core/test/utils/assertions.hpp
```diff
@@ -680,14 +680,14 @@ namespace detail {
 
 
 template <typename T>
-std::initializer_list<std::initializer_list<T>> l(
-    std::initializer_list<std::initializer_list<T>> list)
+const std::initializer_list<std::initializer_list<T>> &l(
+    const std::initializer_list<std::initializer_list<T>> &list)
```
yhmtsai, line 684: Could you explain these changes? avoid unnecessary copy?

yhmtsai (approved):
LGTM

core/matrix/dense.cpp
```diff
@@ -646,38 +646,100 @@ std::unique_ptr<LinOp> Dense<ValueType>::row_permute(
...
+{
+    auto exec = this->get_executor();
+    dim<2> expected_dim{row_indices->get_num_elems(), this->get_size()[1]};
+    GKO_ASSERT_EQUAL_DIMENSIONS(expected_dim, row_gathered);
```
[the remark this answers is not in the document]
upsj: There may be applications where the row_indices contain duplicates and even have more row_indices than total rows in extreme cases (even in distributed SpMV), so I think that would unnecessarily restrict the usage.

core/test/utils/assertions.hpp
upsj: Oh yeah, I forgot to mention these changes, sorry! They are supposed to fix the -Winit-list-lifetime warnings from GCC >= 9, see https://github.com/ginkgo-project/ginkgo/pull/662

thoasm (approved):
LGTM!

include/ginkgo/core/base/array.hpp
upsj: Good catch, thanks!

core/test/base/array.cpp
```diff
@@ -263,6 +263,38 @@ TYPED_TEST(Array, CanBeMovedFromExecutorlessArray)
...
+
+
+TYPED_TEST(Array, CanCreateTemporaryCloneOnDifferentExecutor)
+{
+    auto omp = gko::OmpExecutor::create();
+
+    auto tmp_clone = make_temporary_clone(omp, &this->x);
+
+    this->assert_equal_to_original_x(*tmp_clone.get());
+    ASSERT_NE(tmp_clone.get(), &this->x);
```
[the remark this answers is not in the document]
upsj: Yes, or instantiate the executors with two different memory spaces, if that's possible

core/test/base/array.cpp
```diff
@@ -263,6 +263,38 @@ TYPED_TEST(Array, CanBeMovedFromExecutorlessArray)
...
+TYPED_TEST(Array, CanCopyBackTemporaryCloneOnDifferentExecutor)
+{
+    auto omp = gko::OmpExecutor::create();
+
+    {
+        auto tmp_clone = make_temporary_clone(omp, &this->x);
+        this->x.get_data()[0] = 0;
+    }
+
+    this->assert_equal_to_original_x(this->x);
```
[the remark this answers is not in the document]
upsj: Thanks! I had to think about it for a while as well, so this probably warrants a comment.

tcojean (approved):

pratikvn (approved):
LGTM!

cuda/test/matrix/dense_kernels.cpp
```diff
@@ -591,7 +620,7 @@ TEST_F(Dense, IsInverseRowPermutable)
     set_up_apply_data();
 
     auto inverse_r_permute = x->inverse_row_permute(rpermute_idxs.get());
-    auto d_inverse_r_permute = dx->inverse_row_permute(drpermute_idxs.get());
+    auto d_inverse_r_permute = dx->inverse_row_permute(rpermute_idxs.get());
```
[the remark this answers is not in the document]
upsj: That's implicitly being checked in the reference implementation, though we might want to make it more explicit, that's true

PR #665, opened 2020-11-19
Fix ForwardIterator requirements in iterator_factory

upsj:
Our Windows builds are currently breaking, most likely because the iterator_factor `Iterator` is not *DefaultConstructible* and not *EqualityComparable* when const, which is a requirement of *LegacyForwardIterator*. This PR adds a default constructor which stores a nullptr and the necessary operator const-ness. This is most likely required for std::sort in the new version of Microsoft's std library.

### DIFF
--- a/core/base/iterator_factory.hpp
+++ b/core/base/iterator_factory.hpp
@@ -110,7 +110,7 @@ private:
 
         ~Reference() {}
 
-        Reference(IteratorFactory &parent, array_index_type array_index)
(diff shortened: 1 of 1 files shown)


thoasm (approved):
Thank you for fixing this!

tcojean (approved):
LGTM

pratikvn (approved):

Slaedr (commented):
It's good to have the `operator==` const anyway, irrespective of MSVC. Guess it's already merged, but LGTM too :)

PR #667, opened 2020-11-20
Add github bot

upsj:
This PR adds a Github Action-based bot to the Ginkgo repository.
You can test it out at https://github.com/upsj/ginkgo-bot-playground

By default, it does the following actions:
* On new PR by [OWNER, COLLABORATOR or MEMBER](https://docs.github.com/en/free-pro-team@latest/graphql/reference/enums#commentauthorassociation): check which files have been changed and add appropriate labels ([label.sh](https://github.com/ginkgo-project/ginkgo/blob/add_github_bot/.github/label.sh)). The same check happens if an authorized user comments with `label!` on the PR
* On new PR and sync (roughly after each push, if there are no merge conflicts to the base branch): check if any files need to be formatted ([check-format.sh](https://github.com/ginkgo-project/ginkgo/blob/add_github_bot/.github/check-format.sh)). The same check happens on a `check-format!` comment by an authorized user
* On `format!` command from an authorized user, push a commit with the changes found by `check-format!` ([format.sh](https://github.com/ginkgo-project/ginkgo/blob/add_github_bot/.github/format.sh))
* On `rebase!`, rebase the current branch onto develop ([rebase.sh](https://github.com/ginkgo-project/ginkgo/blob/add_github_bot/.github/rebase.sh))


tcojean:
Probably to make it work I would need to do some setup in terms of Github variables settings and token, right?

.github/check-format.sh
```diff
@@ -0,0 +1,61 @@
...
+git checkout -b format-tmp-$HEAD_BRANCH "fork/$HEAD_BRANCH"
+
+# restore files from develop
+cp ../add_license.sh dev_tools/scripts/
+cp ../update_ginkgo_header.sh dev_tools/scripts/
+cp ../format_header.sh dev_tools/scripts/
+
+# format files
+dev_tools/scripts/update_ginkgo_header.sh
+find . -type f \( -name '*.cuh' -o -name '*.hpp' -o -name '*.hpp.in' -o -name '*.cpp' -o -name '*.cu' -o -name '*.hpp.inc' \) -exec clang-format-8 -i "{}" \;
```
yhmtsai, line 49: Isn't clang-format clang-format-8?
If it is not, I think format_header use different clang-format.

yhmtsai: I put the comment in wrong place. This comment is for format.sh

upsj: I'm not sure, I wanted to make sure explicitly :)
I also decided to leave format_header out for now, since it takes much longer and is possibly less stable than clang-format.
Or do you think it can be safely run for every header?

upsj:
@tcojean as far as I know, we don't need to setup anything, the actions are executed automatically

.github/bot-pr-comment-base.sh
```diff
@@ -0,0 +1,57 @@
...
+BASE_REPO=$(echo "$PR_JSON" | jq -er .base.repo.full_name)
+echo -n .
+BASE_BRANCH=$(echo "$PR_JSON" | jq -er .base.ref)
+echo -n .
+HEAD_REPO=$(echo "$PR_JSON" | jq -er .head.repo.full_name)
+echo -n .
+HEAD_BRANCH=$(echo "$PR_JSON" | jq -er .head.ref)
+echo .
+
+BASE_URL="https://${GITHUB_ACTOR}:${GITHUB_TOKEN}@github.com/$BASE_REPO"
```
tcojean, line 41: I mean, where do these come from? Are they populated by the Github actions automatically?

upsj: Partially by default: https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables
Partially by the workflow files:
```yaml
      env:
        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

tcojean: I see thanks, actually what I was wondering about was where the token comes from as in Gitlab you need to generate that yourself. Looks like here it's automatically provided.
https://docs.github.com/en/free-pro-team@latest/actions/reference/authentication-in-a-workflow

pratikvn:
label!

upsj:
@pratikvn This won't work until the PR is merged, since `issue_comment` events are only dispatched to the default branch's workflows. You can test on my clone though, I just need to add you to the collaborator list :)

pratikvn:
Yes, I realized that after I commented. :)

.github/check-format.sh
yhmtsai: Maybe add another comment `format_header!` to execute it separately? we can also review the change from `format_header` clearly. 
we can only run the updated files in the branch not all files in repo in format related behavior.

.github/label.sh
```diff
@@ -0,0 +1,58 @@
...
+LABELS=$LABELS$(label_match type:factorization /factorization/)
+LABELS=$LABELS$(label_match type:matrix-format /matrix/)
+LABELS=$LABELS$(label_match type:multigrid /multigrid/)
+LABELS=$LABELS$(label_match type:preconditioner /preconditioner/)
+LABELS=$LABELS$(label_match type:reordering /reorder/)
+LABELS=$LABELS$(label_match type:solver /solver/)
+LABELS=$LABELS$(label_match type:stopping-criteria /stop/)
+
+# if all mod: labels present: replace by mod:all
+LABELS=$(echo "$LABELS" | sed 's/.*mod:.*mod:.*mod:.*mod:.*mod:[^"]*"\]/[]+["mod:all"]/')
```
Slaedr, line 55: I'm a regex noob, but do we need one more `mod` here, because there are 6 kinds of `mod`s apart from `mod:all`?

upsj: Great catch, I forgot `dpcpp`!
```suggestion
LABELS=$(echo "$LABELS" | sed 's/.*mod:.*mod:.*mod:.*mod:.*mod:.*mod:[^"]*"\]/[]+["mod:all"]/')
```

Slaedr (approved):
Nice! LGTM.

.github/check-format.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
+source $(dirname "${BASH_SOURCE[0]}")/bot-base.sh
+
+echo -n "Collecting information on triggering PR"
+PR_URL=$(jq -r .pull_request.url "$GITHUB_EVENT_PATH")
```
tcojean, line 6: I guess you cannot use `-e` here because the result can be null?

.github/check-format.sh
```diff
@@ -0,0 +1,61 @@
...
+
+bot_error() {
+  echo "$1"
+  bot_comment "Error: $1"
+  exit 1
+}
+
+# save scripts from develop
+cd dev_tools/scripts
+cp add_license.sh update_ginkgo_header.sh format_header.sh ../../../
```
tcojean, line 35: Instead of using relative paths could you store them in `/tmp` or something? Otherwise, using `$(dirname "${BASH_SOURCE[0]}")/..`. This would make the restauration part less confusing I think.

.github/format.sh
```diff
@@ -0,0 +1,52 @@
...
+# make sure branches are up-to-date
+git fetch origin $BASE_BRANCH
+git fetch fork $HEAD_BRANCH
+
+git config user.email "ginkgo.library@gmail.com"
+git config user.name "ginkgo-bot"
+
+# save scripts from develop
+cd dev_tools/scripts
+cp add_license.sh update_ginkgo_header.sh format_header.sh ../../../
```
tcojean, line 21: Same relative path comment.

.github/format.sh
```diff
@@ -0,0 +1,52 @@
...
+git checkout -b $LOCAL_BRANCH fork/$HEAD_BRANCH
+
+# restore files from develop
+cp ../add_license.sh dev_tools/scripts/
+cp ../update_ginkgo_header.sh dev_tools/scripts/
+cp ../format_header.sh dev_tools/scripts/
+
+# format files
+dev_tools/scripts/update_ginkgo_header.sh
+find . -type f \( -name '*.cuh' -o -name '*.hpp' -o -name '*.hpp.in' -o -name '*.cpp' -o -name '*.cu' -o -name '*.hpp.inc' \) -exec clang-format-8 -i "{}" \;
```
tcojean, line 34: Is it possible to store this command in a script as well for easier maintenance (since it's used two times)?

.github/rebase.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+source $(dirname "${BASH_SOURCE[0]}")/bot-pr-comment-base.sh
+
+if [[ "$PR_MERGED" == "true" ]]; then
+  bot_error "PR already merged!"
+fi
+
+git remote add base "$BASE_URL"
```
tcojean, line 9: In `format.sh` this is `origin`. Either are fine with me, but maybe picking one would be better.

.github/format.sh
```diff
@@ -0,0 +1,52 @@
...
+dev_tools/scripts/add_license.sh
+  
+# restore formatting scripts so they don't appear in the diff
+git checkout -- dev_tools/scripts/*.sh
+
+# check for changed files, replace newlines by \n
+LIST_FILES=$(git diff --name-only | sed '$!s/$/\\n/' | tr -d '\n')
```
tcojean, line 41: Maybe also this line as well.

tcojean (approved):
LGTM. Some minor comments.

.github/format.sh
```diff
@@ -0,0 +1,52 @@
...
+cp add_license.sh update_ginkgo_header.sh format_header.sh ../../../
+cd ../../
+
+LOCAL_BRANCH=format-tmp-$HEAD_BRANCH
+git checkout -b $LOCAL_BRANCH fork/$HEAD_BRANCH
+
+# restore files from develop
+cp ../add_license.sh dev_tools/scripts/
+cp ../update_ginkgo_header.sh dev_tools/scripts/
+cp ../format_header.sh dev_tools/scripts/
```
pratikvn, line 30: This is for security reasons ? So that only the scripts from develop are used ?

.github/check-format.sh
upsj: Exactly, usually I use -e to make the script fail fast if input is missing, i.e. `null` (also the reason I output dots between the steps), but in this situation I wanted to provide an alternative since `check-format` can be run both from a PR as well as a PR comment, which use different output formats.

.github/format.sh
upsj: Exactly, otherwise it would be possible to run code with read/write privileges on the repo if somebody accidentally comments `check-format!` on a malicious PR.

.github/rebase.sh
upsj: The reason for this is slightly convoluted: Due to the flat checkout `fetch-depth: 0` the branch `origin/develop` doesn't contain any history except for the top commit, so I needed to pull the base history manually, so the rebase works out of the box. `format.sh` on the other hand only needs `origin/develop` for the formatting files, the actual commits happen on the fork. The corresponding lines in `format.sh` are probably unnecessary.

.github/format.sh
upsj: good point!

.github/rebase.sh
tcojean: Ah I see, that makes more sense suddenly, thanks for the explanation!

pratikvn (approved):
LGTM!

.github/format.sh
upsj: This is necessary to correctly format the JSON output and most likely won't have to change

upsj:
I finished my changes, I would appreciate if you could take another look at it:
* Unify check-format and format formatting step
* Upload a patch file after a failed check-format run
* Incorporate format_header.sh and avoid header name collisions by adding `PUBLIC_` to public header file include guards
* Improve error output by linking to the failing Github Action.
* Improve security by avoiding executing bash scripts that may be overwritten by a checkout while they are running

PR #669, opened 2020-11-24
Add gpu timer and use seconds in benchmark

yhmtsai:
This PR adds gpu timer (based on event of cuda/hip) in benchmark.
`--gpu_timer=true/false` to choose which timer to use.
The gpu timer requires the first and the last op are on GPU.

**UPDATED** This PR also changes the timing result from nanoseconds to seconds

I only add it for major time measurement.
For the component time (detail), I keep original code because they usually contain some cpu operations.


benchmark/utils/timer.hpp
```diff
@@ -0,0 +1,281 @@
...
+class Timer {
+public:
+    void tic()
+    {
+        assert(tic_called_ == false);
+        this->tic_impl();
+        tic_called_ = true;
+    }
+
+    std::size_t toc()
```
Slaedr, line 82: Why is it nice to have the return type as `size_t`? I see that Cuda and Hip timers actually return `float` which are then cast to `size_t`. Maybe the return type can just be kept as `double`?

benchmark/utils/timer.hpp
```diff
@@ -0,0 +1,281 @@
...
+    {
+        assert(tic_called_ == true);
+        auto ns = this->toc_impl();
+        tic_called_ = false;
+        this->add_record(ns);
+        return ns;
+    }
+
+    std::size_t get_total_time() { return total_duration_ns_; }
```
Slaedr, line 91:
```suggestion
    std::size_t get_total_time() const { return total_duration_ns_; }
```

benchmark/utils/timer.hpp
```diff
@@ -0,0 +1,281 @@
...
+
+    std::size_t get_tictoc_num() { return duration_ns_.size(); }
```
Slaedr, line 93:
```suggestion
    std::size_t get_tictoc_num() const { return duration_ns_.size(); }
```

benchmark/utils/timer.hpp
```diff
@@ -0,0 +1,281 @@
...
+
+    double get_average_time()
```
Slaedr, line 95:
```suggestion
    double get_average_time() const
```

benchmark/utils/timer.hpp
yhmtsai: It is based on the chrono library.
In chrono, we can count the number of nanosecond, so I use size_t to represent it.

Slaedr: Okay. But it does not seem like good practice to cast a `float` to `size_t`... though the other way round (or better yet to `double`) is not too bad. [Side note: [cppreference](https://en.cppreference.com/w/cpp/chrono/duration) says `nanosecond` is a signed 64 bit type, not `size_t`].

benchmark/preconditioner/preconditioner.cpp
```diff
@@ -157,41 +158,32 @@ void run_preconditioner(const char *precond_name,
...
-
-            exec->synchronize();
-            auto g_tac = std::chrono::steady_clock::now();
+            generate_timer->toc();
 
             auto generate_time =
-                std::chrono::duration_cast<std::chrono::nanoseconds>(g_tac -
-                                                                     g_tic) /
-                FLAGS_repetitions;
+                generate_timer->get_total_time() / FLAGS_repetitions;
```
pratikvn, line 173: I guess you can use the `get_average_time` here as well ?

benchmark/utils/timer.hpp
```diff
@@ -0,0 +1,281 @@
...
+#ifdef HAS_HIP
+
+
+#include <hip/hip_runtime.h>
+
+
+#include "hip/base/device_guard.hip.hpp"
+
+
+#endif  // HAS_CUDA
```
pratikvn, line 65:
```suggestion
#endif  // HAS_HIP
```

benchmark/utils/timer.hpp
```diff
@@ -0,0 +1,281 @@
...
+DEFINE_bool(gpu_timer, false,
+            "use gpu timer based on event. It is valid only when "
+            "executor is cuda or hip");
+
+
+class Timer {
+public:
+    void tic()
+    {
+        assert(tic_called_ == false);
```
pratikvn, line 77: Not sure if this is useful, as if someone puts a `#define NDEBUG` at some point, it gets optimized out when in `Release`. Maybe you can define your own simple `GKO_ASSERT` macro ?

benchmark/utils/timer.hpp
pratikvn, line 93: Maybe you can rename this to `num_repetitions` or something like that ?

pratikvn (approved):
LGTM! Cleans up a lot of the benchmarking code!

benchmark/utils/timer.hpp
thoasm: I don't think it is a good idea to have `toc()` return something because it is not obvious what it returns and in any case, the return value is not really useful. It could be any of these return values:
- The number of times the time has already been recorded
- the average runtime up until this point
- the sum of the runtime until this point
- the runtime of the latest measurement (you have chosen this one)

Additionally, it is not clear that the unit returned is `nanoseconds`. All in all, I would recommend to have this function not return anything.
If you want to know the latest result, I would prefer an additional function (`std::chrono` does the same).

benchmark/utils/timer.hpp
thoasm: I agree, that would be more self-explanatory.

benchmark/utils/timer.hpp
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    virtual void tic_impl() = 0;
+
+    virtual std::size_t toc_impl() = 0;
+
+private:
+    std::vector<std::size_t> duration_ns_;
+    bool tic_called_;
+    std::size_t total_duration_ns_;
```
thoasm, line 125: I would prefer if we store this as `std::int64_t` (signed).
I am fine with storing this as nanoseconds because we don't lose precision when converting from `float` (or even `double`) to this, assuming we never measure something close to the nanosecond range (and if we would, we need a new implementation anyway because a virtual function call is too slow for that).
However, I would prefer a signed type for the same reason we want to use signed as the new `size_type` in Ginkgo 2.0:
- for 64 bit, we will never use the extra bit (like this, we can already measure up to 300 years)
- we prevent the many warnings for casting from unsigned to signed
Also, as @Slaedr already pointed out, `std::chrono` already uses a signed type (`std::int64_t`) for duration, so we prevent a casting warning here as well.

benchmark/utils/timer.hpp
```diff
@@ -0,0 +1,281 @@
...
+    void add_record(std::size_t ns)
+    {
+        // add the result;
+        duration_ns_.emplace_back(ns);
+        total_duration_ns_ += ns;
+    }
+
+    virtual void tic_impl() = 0;
+
+    virtual std::size_t toc_impl() = 0;
```
thoasm, line 120: `toc_impl()` returning is fine with me, but it should be both documented and return `std::int64_t`.

thoasm (changes requested):
When using the script, you currently can't measure with GPU time.

benchmark/utils/timer.hpp
Slaedr: I guess the point would be to avoid this check in release, so as to disturb the timing as little as possible? In case someone suspects issues, they can run in debug and check. Admittedly, the extra assert would not usually matter for the timing, but still, you never know how it might be used later.

benchmark/utils/timer.hpp
Slaedr: Technically `float` can go up to 10^49 or something which would overflow even `size_t`, and the GPU timers return `float`, I would consider just using `double`. I agree that in practice, we can probably also just use `int64_t`. :)

benchmark/preconditioner/preconditioner.cpp
yhmtsai: The timer is out of the repetition loop, so `total_time = average_time` here .
To get the correct average time, need `total_time/FLAGS_repetitions` or `average_time/FLAGS_repetitions`.
`total_time` is less confusing than `average_time`?

benchmark/utils/timer.hpp
thoasm: Yes, `float` can fit that, but we lose a lot of resolution in the process with float, which we won't with `int64`. Using plain `double` would obviously also work, and I would be fine with that as well.

Slaedr (approved):
Looks great! I would still prefer that `toc` etc. return `double`, but `int64_t` is also reasonable so I'll leave it to you.

benchmark/preconditioner/preconditioner.cpp
```diff
@@ -157,41 +158,32 @@ void run_preconditioner(const char *precond_name,
...
-            exec->synchronize();
-            auto a_tac = std::chrono::steady_clock::now();
-
-            auto apply_time =
-                std::chrono::duration_cast<std::chrono::nanoseconds>(a_tac -
-                                                                     a_tic) /
-                FLAGS_repetitions;
-            add_or_set_member(this_precond_data["apply"], "time",
-                              apply_time.count(), allocator);
+            auto apply_time = apply_timer->get_total_time() / FLAGS_repetitions;
```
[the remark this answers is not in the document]
tcojean: Since this is outside the loop and the timer doesn't know about `FLAGS_repetitions`, wouldn't this give a wrong result? (equivalent to `get_total_time()`).

benchmark/utils/timer.hpp
```diff
@@ -0,0 +1,283 @@
...
+
+#endif  // HAS_HIP
+
+
+DEFINE_bool(gpu_timer, false,
+            "use gpu timer based on event. It is valid only when "
+            "executor is cuda or hip");
+
+
+class Timer {
```
tcojean, line 73: Should we use the `gko` namespace?

tcojean (approved):
LGTM. Please add documentations though for the the different classes and public API. Also, please add some documentation to `BENCHMARKING.md`.

benchmark/utils/timer.hpp
```diff
@@ -0,0 +1,283 @@
...
+#endif  // HAS_CUDA
+
+#ifdef HAS_HIP
+        if (auto hip =
+                std::dynamic_pointer_cast<const gko::HipExecutor>(exec)) {
+            return std::make_shared<HipTimer>(hip);
+        }
+#endif  // HAS_HIP
+    }
+    // Not use gpu_timer or not cuda/hip executor
```
thoasm, line 281: nit:
```suggestion
    // No cuda/hip executor available or no gpu_timer used
```

thoasm (commented):
I agree with @tcojean: I would like to have more documentation.

benchmark/utils/timer.hpp
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    std::int64_t toc_impl() override
+    {
+        exec_->synchronize();
+        auto stop = std::chrono::steady_clock::now();
+        auto duration_time =
+            std::chrono::duration_cast<std::chrono::nanoseconds>(stop - start_)
+                .count();
+        return static_cast<std::int64_t>(duration_time);
```
Slaedr, line 150: I guess a cast is not required here anymore.
```suggestion
        return duration_time;
```

benchmark/preconditioner/preconditioner.cpp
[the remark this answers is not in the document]
thoasm: The `apply_timer` does know how often it was started and stopped, therefore, it is independent of the `FLAGS_repetitions` (and in my opinion safer) when using `get_average_time()`.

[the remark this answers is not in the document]
tcojean: https://github.com/ginkgo-project/ginkgo/pull/669/files#diff-d573d9686e53b5ce4b41de8061f46d1a30693d2dfa59795bec0680f6bdd4e4dcR178-R183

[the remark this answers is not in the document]
tcojean: It's independent only if you put it inside the for loop, AFAIK it is not here. See the context of the line highlighted:
https://github.com/ginkgo-project/ginkgo/pull/669/files#diff-d573d9686e53b5ce4b41de8061f46d1a30693d2dfa59795bec0680f6bdd4e4dcR178-R183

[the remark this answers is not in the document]
yhmtsai: We put the timer out of for-loop, so the apply_timer can not know the number of for-loop. As terry said, I keep the current timer for less overhead. If we need to refill/reset the x,b like spmv or solver, we can put the timer inside the for-loop

benchmark/utils/timer.hpp
yhmtsai: Should I put it in `gko::` namespace? I put the timer only in benchmark, so I am not sure whether it is suitable with gko namespace

yhmtsai:
format!

benchmark/utils/timer.hpp
```diff
@@ -0,0 +1,386 @@
...
+    double get_average_time() const
+    {
+        return static_cast<double>(this->get_total_time()) /
+               this->get_num_repetitions();
+    }
+
+    /**
+     * Get the vector containing the each time results in nanoseconds.
+     *
+     * @return the vector of each time results in nanoseconds
```
tcojean, line 128:
```suggestion
     * Get the vector containing the time of each repetition in nanoseconds.
     *
     * @return the vector of time for each repetition in nanoseconds
```

benchmark/utils/timer.hpp
```diff
@@ -0,0 +1,386 @@
...
+     */
+    std::vector<std::int64_t> get_time_detail() const { return duration_ns_; }
+
+    /**
+     * Get the latest result in nanoseconds. If there is not result yet, return
```
tcojean, line 133:
```suggestion
     * Get the latest result in nanoseconds. If there is no result yet, return
```

benchmark/utils/timer.hpp
```diff
@@ -0,0 +1,386 @@
...
+    /**
+     * Create a timer
+     */
+    Timer() : tic_called_(false), total_duration_ns_(0) {}
+
+protected:
+    /**
+     * Put the nanosecond result into vector
+     */
+    void add_record(std::int64_t ns)
```
tcojean, line 166:
```suggestion
    /**
     * Put the nanosecond result into vector
     *
     * @param ns  the nanosecond result to insert
     */
    void add_record(std::int64_t ns)
```

benchmark/utils/timer.hpp
```diff
@@ -0,0 +1,386 @@
...
+
+private:
+    std::vector<std::int64_t> duration_ns_;
+    bool tic_called_;
+    std::int64_t total_duration_ns_;
+};
+
+
+/**
+ * CpuTimer use synchronize of executor and std::chrono to measure the
```
tcojean, line 193:
```suggestion
 * CpuTimer use the synchronize of the executor and std::chrono to measure the
```

benchmark/utils/timer.hpp
```diff
@@ -0,0 +1,386 @@
...
+
+/**
+ * CudaTimer uses cuda executor and cudaEvent to measure the timing.
+ */
+class CudaTimer : public Timer {
+public:
+    /**
+     * Create a CudaTimer.
+     *
+     * @param exec Executor which is CudaExecutor indeed
```
tcojean, line 240:
```suggestion
     * @param exec  Executor which should be a CudaExecutor
```

benchmark/utils/timer.hpp
```diff
@@ -0,0 +1,386 @@
...
+     */
+    CudaTimer(std::shared_ptr<const gko::Executor> exec)
+        : CudaTimer(std::dynamic_pointer_cast<const gko::CudaExecutor>(exec))
+    {}
+
+    /**
+     * Create a CudaTimer.
+     *
+     * @param exec CudaExecutor associated to the timer
```
tcojean, line 249:
```suggestion
     * @param exec  CudaExecutor associated to the timer
```

benchmark/utils/timer.hpp
```diff
@@ -0,0 +1,386 @@
...
+
+/**
+ * HipTimer uses hip executor and hipEvent to measure the timing.
+ */
+class HipTimer : public Timer {
+public:
+    /**
+     * Create a HipTimer.
+     *
+     * @param exec Executor which is HipExecutor indeed
```
tcojean, line 305:
```suggestion
     * @param exec  Executor which should be a HipExecutor
```

benchmark/utils/timer.hpp
```diff
@@ -0,0 +1,386 @@
...
+     */
+    HipTimer(std::shared_ptr<const gko::Executor> exec)
+        : HipTimer(std::dynamic_pointer_cast<const gko::HipExecutor>(exec))
+    {}
+
+    /**
+     * Create a HipTimer.
+     *
+     * @param exec HipExecutor associated to the timer
```
tcojean, line 314:
```suggestion
     * @param exec  HipExecutor associated to the timer
```

tcojean (approved):
LGTM.

benchmark/utils/timer.hpp
```diff
@@ -0,0 +1,390 @@
...
+    {
+        exec_->synchronize();
+        start_ = std::chrono::steady_clock::now();
+    }
+
+    double toc_impl() override
+    {
+        exec_->synchronize();
+        auto stop = std::chrono::steady_clock::now();
+        auto duration_time = get_duration_in_seconds(stop - start_);
```
Slaedr, line 221: We could just use the standard way of doing this as shown here: https://en.cppreference.com/w/cpp/chrono/steady_clock/now
We don't really need an entire function and file to do this.
```suggestion
        std::chrono::duration<double> duration_time = stop - start_;
```

benchmark/utils/timer.hpp
```diff
@@ -0,0 +1,390 @@
...
+        return duration_time;
```
Slaedr, line 222:
```suggestion
        return duration_time.count();
```

benchmark/utils/timer.hpp
yhmtsai: thanks for the example, I did not know that before.

PR #670, opened 2020-11-24
Check if executors share the same memory

tcojean:
This PR contains only the first commit of #652.

This implements a new executor function `memory_accessible` between executors which verifies whether they share the same memory. Currently, this functionality is used in the `temporary_clone` function to save copies.
This currently means that:
+ Any OpenMP executor has the same memory has another OpenMP executor.
+ Same for Reference executors.
+ Same for CUDA executors if they have the same device id, and same for HIP executors.
+ OpenMP and Reference also share the same memory, and same for a host-side DPC++ executor.
+ HIP and CUDA have the same memory if they are the same device id and HIP actually uses the CUDA backend.


yhmtsai (changes requested):
how about we still isolate the reference executor?
If we optimize omp executor with some additional information like csr in cuda/hip,
considering omp/reference as the same memory will make the conversion failed.

upsj:
Awesome idea, this might also help later on with the MPI executor. Two comments before I review it in more detail:
1. This slightly changes the semantics of make_temporary_clone since the returned object no longer necessarily exists on the executor we passed in. Do we use this other executor somewhere?
2. I would prefer if we used a member function `memory_accessible` or something like that instead of overloading `operator==`, since that might prevent us from implementing full executor equality checks later.

include/ginkgo/core/base/executor.hpp
```diff
@@ -684,6 +695,33 @@ class Executor : public log::EnableLogging<Executor> {
...
+ * @internal
+ * Declares a verify_memory_to() overload for a specified Executor subclass.
+ *
+ * This is the second stage of the double dispatch emulation required to
+ * implement verify_memory_from().
+ *
+ * @param _exec_type  the Executor subclass
+ */
+#define GKO_ENABLE_VERIFY_MEMORY_TO(_exec_type, ...) \
+    virtual bool verify_memory_to(const _exec_type *dest_exec) const = 0
```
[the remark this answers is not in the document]
tcojean: Well the comment here is about the macro though, not the function? Also, I don't think we can put a comment between the macro definition and its body, that's why we have these macro commented everywhere.

tcojean:
> If we optimize omp executor with some additional information like csr in cuda/hip,
> considering omp/reference as the same memory will make the conversion failed.

I don't think this is such a big issue. If the algorithm needs to clone anyway, `gko::clone` will still work. If it needs to transform the memory without copies, it can do as long as it doesn't break anything, or it's put back to the previous state. You could always call in `make_srow()` equivalent in that situation to update the strategy related data? Also, it doesn't seem like we  have such a case for now.

>  This slightly changes the semantics of make_temporary_clone since the returned object no longer necessarily exists on the executor we passed in. Do we use this other executor somewhere?
Indeed, it does in  two ways: temporary clone is called after this PR only when there is no direct access to memory, and second we do not ensure anymore that LinOps have exactly the same executors when calling a function like `apply()`. So far in terms of tests that seems to work, but I'm indeed not sure whether we have such an extensive testing of all combinations to prove that at a 100% level. I believe that's where integration tests would be useful.

> I would prefer if we used a member function memory_accessible or something like that instead of overloading operator==, since that might prevent us from implementing full executor equality checks later.
Sure, I don't see how else we would use `operator==` than this but that does make it more future proof, so I did this change.

core/test/base/array.cpp
```diff
@@ -271,29 +272,30 @@ TYPED_TEST(Array, CanCreateTemporaryCloneOnSameExecutor)
...
     auto tmp_clone = make_temporary_clone(omp, &this->x);
 
     this->assert_equal_to_original_x(*tmp_clone.get());
-    ASSERT_NE(tmp_clone.get(), &this->x);
+    ASSERT_EQ(tmp_clone.get(), &this->x);
 }
 
 
-TYPED_TEST(Array, CanCopyBackTemporaryCloneOnDifferentExecutor)
+TYPED_TEST(Array, CanDoesNotCopyBackTemporaryCloneBetweenSameMemory)
```
Slaedr, line 287:
```suggestion
TYPED_TEST(Array, DoesNotCopyBackTemporaryCloneBetweenSameMemory)
```

include/ginkgo/core/base/executor.hpp
```diff
@@ -1371,19 +1470,30 @@ class DpcppExecutor : public detail::ExecutorBase<DpcppExecutor>,
...
+
 protected:
     void set_device_property();
 
     DpcppExecutor(int device_id, std::shared_ptr<Executor> master,
                   std::string device_type = "all")
         : device_id_(device_id), master_(master), device_type_(device_type)
     {
+        std::for_each(device_type_.begin(), device_type_.end(),
+                      [](char &c) { c = std::tolower(c); });
```
pratikvn, line 1496: :)

pratikvn (approved):
LGTM! But I have one question: This PR changes the behaviour of `make_temporary_clone` and `make_temporary_clone` seems to be part of the public interface.

Doesn't that mean we would be possibly breaking existing code ?

dpcpp/base/executor.dp.cpp
```diff
@@ -57,6 +57,7 @@ const std::vector<sycl::device> get_devices(std::string device_type)
         {"accelerator", sycl::info::device_type::accelerator},
         {"all", sycl::info::device_type::all},
         {"cpu", sycl::info::device_type::cpu},
+        {"host", sycl::info::device_type::host},
```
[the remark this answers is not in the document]
tcojean: Yes, it's a special SYCL "device", which works on the CPU but only sees one core, and has no subgroup at all (not even size 1). I see it as a lightweight version of our OmpExecutor or ReferenceExecutor which can be used when you mainly use a device code, but I'm not sure if that's the real purpose.

Honestly, I don't want to advertise this device type since it's so useless for general use, I put this here only to ensure compatibility and complete coverage.

tcojean:
I don't think changing the temporary clone interface is interface breaking, since the point of this function is to do a copy when you need it. This still happens now, all that changes is that you get less copies when the data is already sitting on memory that you can access.

Slaedr (commented):
Wow, this is an intricate piece of work.

My understanding of why we need a two-level dispatch:

Because we do not want to template `Executor` on the concrete executor type, we have an `ExecutorBase` template with CRTP, and this is a `friend` of `Executor` for all CRTP values. The `verify_memory_to` functions, which do the actual work, depend on being overloaded on the concrete executor types. This means calling one of them requires knowledge of the concrete type. Therefore, `Executor` cannot directly call `verify_memory_to`, so it calls `verify_memory_from`. This second function is not overloaded but is implemented in `ExecutorBase` and thus has access to its concrete type, and can call `verify_memory_to` of its argument, passing its concrete self as the argument in turn.

Is that correct? Does this get affected when actual memory space classes are introduced, or has the plan for that changed?

I have a minor reservation which is a point that @pratikvn raised, about consistency in using pointers and references for `verify_memory_to` and `verify_memory_from`.

tcojean:
@Slaedr yes you are pretty much exactly correct about that design, this is how it works. The idea of this double dispatch with the CRTP enabled class in the middle is to allow to evaluate to the actual executor type so that we can call the proper `verify_memory_to`, for the proper target executor.

About the memory space classes, either this will stay with memory spaces instead of executors, or, more likely, we build the memory spaces properly so that they are shared when they should be, and anyway we don't need all of this anymore.

About using pointers, yes I will change to that, this is a leftover of the previous design which I started changing in some recent commits, and had not done so for the deepest levels.

Slaedr:
> About the memory space classes, either this will stay with memory spaces instead of executors, or, more likely, we build the memory spaces properly so that they are shared when they should be, and anyway we don't need all of this anymore.

So all this is a only temporary measure until the full memory spaces are implemented? How much of this will be retained?

tcojean:
> > About the memory space classes, either this will stay with memory spaces instead of executors, or, more likely, we build the memory spaces properly so that they are shared when they should be, and anyway we don't need all of this anymore.
> 
> So all this is a only temporary measure until the full memory spaces are implemented? How much of this will be retained?

That is hard to say. First of all, the memory space is definitely interface breaking so it could be in a while still, whereas this is not. Removing this and replacing by memory spaces will be interface breaking as well, but not much more since memory space sare interface breaking to begin with. Second, it depends as I said on how the memory space implementation is done. You could keep this together with memory space, if we somehow want to keep a CudaMemory and HipMemory for example, you would still need to say Cuda == HIP when HIP is ran on CUDA and both have the same device id. If it's done in another way, where HIP executor gets a CudaMemory space while using a CUDA backend, then in this case you don't need the verify memory overload.

tcojean:
rebase!

tcojean:
format!

include/ginkgo/core/base/executor.hpp
```diff
@@ -633,6 +633,18 @@ class Executor : public log::EnableLogging<Executor> {
...
     virtual void synchronize() const = 0;
 
+    /**
+     * Verifies whether the executors share the same memory.
+     *
+     * @param other  the other Executor to compare against
+     *
+     * @return whether the executors this and other share the same memory.
+     */
+    bool memory_accessible(const std::shared_ptr<const Executor> &other) const
```
Slaedr, line 643: Just curious: why do we take a reference to the `shared_ptr` instead of passing the `shared_ptr` by value? We don't want the pointer's internal reference count to be updated here?

Slaedr (approved):
LGTM! Just one minor question.

include/ginkgo/core/base/executor.hpp
tcojean: Yes, we don't need to take ownership here since we don't store the other executor anywhere etc, we only want to access it. Since Ginkgo isn't multithreaded (at least in terms of executor management), there is no issue such as the executor being destructed in between our access as well.

dpcpp/base/executor.dp.cpp
```diff
@@ -150,6 +159,32 @@ int DpcppExecutor::get_num_devices(std::string device_type)
 }
 
 
+bool DpcppExecutor::verify_memory_to(const OmpExecutor *dest_exec) const
+{
+    auto device = detail::get_devices(device_type_)[device_id_];
+
+    return device.is_host() || device.is_cpu();
```
thoasm, line 166: nit: usually, you did not have an empty line between `device` and `return`:
```suggestion
    return device.is_host() || device.is_cpu();
```

dpcpp/base/executor.dp.cpp
```diff
@@ -150,6 +159,32 @@ int DpcppExecutor::get_num_devices(std::string device_type)
...
+bool DpcppExecutor::verify_memory_to(const DpcppExecutor *dest_exec) const
+{
+    auto device = detail::get_devices(device_type_)[device_id_];
+    auto other_device = detail::get_devices(
+        dest_exec->get_device_type())[dest_exec->get_device_id()];
+    return ((device.is_host() || device.is_cpu()) &&
+            (other_device.is_host() || other_device.is_cpu())) ||
+           (device.get_info<cl::sycl::info::device::device_type>() ==
+                other_device.get_info<cl::sycl::info::device::device_type>() &&
+            device.get() == other_device.get());
```
thoasm, line 184: I don't fully understand why you compare both the devices and the device info for equality. Isn't the info always the same when the devices are equal?

core/test/base/executor.cpp
```diff
@@ -474,6 +474,76 @@ TEST(DpcppExecutor, KnowsItsDeviceId)
...
+    auto cuda_1 = gko::CudaExecutor::create(1, omp);
+    std::shared_ptr<gko::DpcppExecutor> host_dpcpp;
+    std::shared_ptr<gko::DpcppExecutor> cpu_dpcpp;
+    std::shared_ptr<gko::DpcppExecutor> gpu_dpcpp;
+    if (gko::DpcppExecutor::get_num_devices("host"))
+        host_dpcpp = gko::DpcppExecutor::create(0, omp, "host");
+    if (gko::DpcppExecutor::get_num_devices("cpu"))
+        cpu_dpcpp = gko::DpcppExecutor::create(0, omp, "cpu");
+    if (gko::DpcppExecutor::get_num_devices("gpu"))
+        gpu_dpcpp = gko::DpcppExecutor::create(0, omp, "gpu");
```
thoasm, line 496: nit: According to our guidelines, we always want to use braces `{}`, even when we have just a single statement.

thoasm, line 496: Is it intentional, that the `gpu_dpcpp` executor is an `omp` executor?
Note: Comment only applies to last line.
```suggestion
    if (gko::DpcppExecutor::get_num_devices("host"))
        host_dpcpp = gko::DpcppExecutor::create(0, omp, "host");
    if (gko::DpcppExecutor::get_num_devices("cpu"))
        cpu_dpcpp = gko::DpcppExecutor::create(0, omp, "cpu");
    if (gko::DpcppExecutor::get_num_devices("gpu"))
        gpu_dpcpp = gko::DpcppExecutor::create(0, cuda, "gpu");
```

thoasm (approved):
LGTM, some minor comments.

core/test/base/executor.cpp
tcojean: Yes, the same way as with CUDA devices, we want to have access to a real host side executor for doing our Ginkgo copies through when we use an Intel GPU. In fact, such host side executor is not required when creating an `host` or `cpu` DPC++ executor, but we need to have something just in case. Another issue is that you can control what we see through the SYCL environment variables, and it's possible to enforce, up to a point, what device is actually selected from outside Ginkgo's scope...

dpcpp/base/executor.dp.cpp
tcojean: I am somewhat anticipating machines with multiple Intel GPUs, or even different vendors possibly. In this case, we need to distinguish whether we have an "accelerator", a "gpu" or other, but also if they actually are the same (same ID/vendor combination), or different. I have to say I obviously could not test this piece of code yet, as all machines we have access to only have one GPU.

tcojean (commented):
Thanks for the comments, I will fix the issues pointed out.

core/test/base/executor.cpp
thoasm: Can you add a comment for this case, to avoid this confusion?

tcojean: I'm not sure what clarification is needed here, the code isn't different in essence from:
```
auto cuda = gko::CudaExecutor::create(0, omp);
```

yhmtsai (approved):
LGTM. Thanks for handling this issue and add a lot of tests.
I have one question on the test.
we will make the reference executor isolated, right?

Slaedr:
Could you remind me why we need the reference executor to have a different memory space from the omp executor? Do we document this somewhere?

cuda/test/base/lin_op.cu
```diff
@@ -39,6 +39,110 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+    ASSERT_EQ(op->last_b_access, cuda);
+    ASSERT_EQ(op->last_x_access, cuda);
+}
+
+
+TEST_F(EnableLinOp, ApplyCopiesBackOnlyX)
+{
+    op->apply(gko::lend(b), gko::lend(x));
+
+    ASSERT_EQ(b->last_access, nullptr);
```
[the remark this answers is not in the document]
tcojean: That's the same as what was present before. It's nullptr because `b` was not copied back. The access was done on the clone in the `cuda` executor, but that information wasn't propagated back to the original object which is sitting in the reference executor (because there is no need, it's read only).

yhmtsai:
we have some specific data in different executor, like csr srow.
If we consider omp and reference in the same memory, when omp apply on reference data, the data will not be converted to the omp one. Thus, omp operation may access unallocated memory or wrong data.
Note. we only have it in cuda/hip not omp/reference now.

Another point is from testing or design.
We use reference executor/operation as our correct result to compare different executor.
May isolating the reference executor to avoid any unexpected operation or other executor make sense?

include/ginkgo/core/base/executor.hpp
```diff
@@ -1090,6 +1167,19 @@ class CudaExecutor : public detail::ExecutorBase<CudaExecutor>,
...
+    GKO_DEFAULT_OVERRIDE_VERIFY_MEMORY(ReferenceExecutor, false);
+
+    GKO_DEFAULT_OVERRIDE_VERIFY_MEMORY(DpcppExecutor, false);
+
+    bool verify_memory_to(const HipExecutor *dest_exec) const override;
+
+    bool verify_memory_to(const CudaExecutor *dest_exec) const override
+    {
+        return device_id_ == dest_exec->get_device_id();
+    }
```
upsj, line 1181: Why is this in the header file, but not the corresponding HIP function?

upsj (approved):
LGTM! Just a minor question

include/ginkgo/core/base/executor.hpp
tcojean: The issue is a chicken and egg problem. Who came first, CUDA or HIP? at most, I could do it for one of them, but not for both, so I decided to do it for none instead.

In more details, we have a predeclaration of all classes to allow for friendships. In order, CUDA Executor is declared then HIP. At the point in time where we want to call, inside the CudaExecutor, `HIP->get_device_id()`, we still only have incomplete information about HIP, and thus cannot find that function.

upsj: Could we define both in the source files instead?

tcojean: You mean also the Cuda->Cuda one? The annoying thing is we then also need to implement the `device_hook`, but I can do that, yes.

tcojean: I've done that by moving the two implementations into `devices/{cuda|hip}/executor.cpp`, I'm not sure if it's more or less confusing, but at least I don't have to duplicate the implementation this way.

cuda/test/base/lin_op.cu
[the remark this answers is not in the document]
yhmtsai: Thanks for explanation

PR #674, opened 2020-12-02
Format everything & delete old bot comments

upsj:
This PR formats all files and fixes a few broken formatting configs, as well as deleting old comments from check-format! and rebase!


upsj:
Okay, that was an interesting fail :laughing: I must have used the wrong clang-format version in format_header.sh

upsj:
format!

pratikvn:
So it also seems that `format!` comment does not trigger any CI pipelines ?

upsj:
Yes, that seems to be a Github bug, where pushes by Actions don't cause `push` events to be triggered.

upsj:
Related forum post: https://github.community/t/push-from-action-does-not-trigger-subsequent-action/16854

.github/bot-pr-base.sh
```diff
@@ -41,6 +41,13 @@ HEAD_URL="https://${GITHUB_ACTOR}:${GITHUB_TOKEN}@github.com/$HEAD_REPO"
 
 JOB_URL="https://github.com/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID"
 
+bot_delete_comments_matching() {
+  COMMENTS=$(api_get "$ISSUE_URL/comments" | jq -er '.[] | select((.user.login == "github-actions[bot]") and (.body | startswith('"\"$1\""'))) | .url')
```
tcojean, line 45: Maybe store `$1` into a variable to understand what it's doing? Like:
```suggestion
  local search_matching=$1
  COMMENTS=$(api_get "$ISSUE_URL/comments" | jq -er '.[] | select((.user.login == "github-actions[bot]") and (.body | startswith('"\"$search_matching\""'))) | .url')
```

reference/test/preconditioner/isai_kernels.cpp
```diff
@@ -47,11 +47,11 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/matrix/csr.hpp>
 #include <ginkgo/core/matrix/dense.hpp>
 #include <ginkgo/core/preconditioner/ilu.hpp>
-#include <ginkgo/core/preconditioner/isai.hpp>
 
 
 #include "core/test/utils.hpp"
 #include "matrices/config.hpp"
+#include "reference/preconditioner/isai_kernels.cpp"
```
tcojean, line 54: Is including the `.cpp` correct here? Surely not?

tcojean (approved):
LGTM.

reference/test/preconditioner/isai_kernels.cpp
upsj: Oh yeah, good catch. It was probably working before because the .cpp file includes the .hpp file, but still

pratikvn:
But I think the bug would mean that whenever a format commit is called atleast the actions would need to be triggered right ? Can we do that manually ? Maybe you can add a command (like `run-pipelines!`) to trigger the pipelines instead ?

upsj:
I am working on a workaround using a private access token, that should be simpler. In the future, I would probably add a `check-pr!` command to trigger Gitlab pipelines for PRs from forks.

omp/solver/idr_kernels.cpp
```diff
@@ -33,10 +33,10 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "core/solver/idr_kernels.hpp"
 
 
-#include <time.h>
 #include <algorithm>
 #include <iostream>
 #include <random>
+#include <time.h>
```
pratikvn, line 39: Maybe this should be in a separate section of its own as it is a C header ?

omp/test/solver/idr_kernels.cpp
```diff
@@ -48,6 +46,8 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/base/mtx_io.hpp>
 #include <ginkgo/core/matrix/csr.hpp>
 #include <ginkgo/core/matrix/dense.hpp>
+#include <ginkgo/core/solver/bicgstab.hpp>
+#include <ginkgo/core/solver/gmres.hpp>
 #include <ginkgo/core/stop/combined.hpp>
```
pratikvn, line 51: Maybe these headers are not needed

reference/solver/idr_kernels.cpp
```diff
@@ -33,10 +33,10 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "core/solver/idr_kernels.hpp"
 
 
-#include <time.h>
 #include <algorithm>
 #include <iostream>
 #include <random>
+#include <time.h>
```
pratikvn, line 39: Again, maybe better in its own section

pratikvn (approved):
LGTM! Only some minor things.

cuda/factorization/par_ilut_approx_filter_kernel.cu
```diff
@@ -44,7 +44,6 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
 #include "core/components/prefix_sum.hpp"
-#include "core/factorization/par_ilut_kernels.hpp"
```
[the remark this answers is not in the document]
upsj: I was as surprised as you are: This is a duplicate of the topmost header

dev_tools/scripts/config
```diff
@@ -6,10 +6,26 @@
...
+- "(cuda|hip|dpcpp)/factorization/par_ict_"
+    - FixInclude: "core/factorization/par_ict_kernels.hpp"
 - "(cuda|hip|dpcpp)/preconditioner/jacobi_"
     - FixInclude: "core/preconditioner/jacobi_kernels.hpp"
 - "core/test/base/(extended_float|iterator_factory)"
     - RemoveTest: "true"
+- "core/test/base/allocator"
+    - FixInclude: "core/base/allocator.hpp"
+- "reference/test/base/utils"
+    - FixInclude: "core/base/utils.hpp"
```
yhmtsai, line 28: at some point, I will check whether adding a special comment as the main header.
```
#include "core/base/utils.hpp" // MAIN
```
it will use it as main header no matter how the config sets

test_install/test_install.cpp
```diff
@@ -357,13 +357,13 @@ int main(int, char **)
...
+        auto rel_res =
+            gko::stop::RelativeResidualNorm<>::build().with_tolerance(1e-10).on(
+                refExec);
 
-        auto abs_res = gko::stop::AbsoluteResidualNorm<>::build()
-                           .with_tolerance(1e-10)
-                           .on(refExec);
+        auto abs_res =
+            gko::stop::AbsoluteResidualNorm<>::build().with_tolerance(1e-10).on(
+                refExec);
```
yhmtsai, line 366: It seems to be formatted not well, but my local formatted result is the same as it

yhmtsai (approved):
LGTM

omp/solver/idr_kernels.cpp
upsj: I guess the correct include here would be `ctime`, which should definitely be part of the same group, or even use std::random_device, but I wanted to keep the changes as minimal as possible. Maybe something for a later PR?

dev_tools/scripts/config
```diff
@@ -6,10 +6,26 @@
     - FixInclude: "hip/hip_runtime.h"
```
Slaedr, line 6: What's the purpose of this file?

upsj: You mean hip_runtime? It contains all the macros like hipLaunchKernelGGL, HIP_KERNEL_NAME, device-side assert etc.

test_install/test_install.cpp
Slaedr: I wonder why this was reformatted like this. The existing code is actually easier to read and looks better.

.github/workflows/bot-pr-comment.yml
```diff
@@ -14,7 +14,7 @@ jobs:
         fetch-depth: 0
     - name: Add appropriate labels
       env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        GITHUB_TOKEN: ${{ secrets.BOT_TOKEN }}
```
Slaedr, line 17: What's the difference between `secrets.GITHUB_TOKEN` and `secrets.BOT_TOKEN`?

Slaedr (approved):
LGTM. Just a couple of questions.

dev_tools/scripts/config
Slaedr: No, I meant this entire `config` file, haha. I guess it's unrelated to this PR but I have no idea what's going on here.

.github/workflows/bot-pr-comment.yml
upsj: GITHUB_TOKEN is an automatically generated and temporary token for **github-actions[bot]**, which has the annoying property that pushes from this token don't trigger subsequent Github Actions (like Windows/macOS CI jobs) to avoid infinite loops in Github Actions by default.
BOT_TOKEN is a manually generated personal authentication token for @ginkgo-bot which circumvents this limitation.
So in the future, comments will come from **ginkgo-bot** instead of **github-actions[bot]** and the CI jobs will run after every push

test_install/test_install.cpp
upsj: Yes, this is pretty unfortunate. Luckily, this is only part of our install test, which we only rarely need to modify anyways.

tcojean: Yes, when reviewing this I also thought the previous version looked better, but that's the usual `clang-format` shenaningans.

dev_tools/scripts/config
upsj: This file configures the `format_header.sh` script by fixing a few special-cases where the default detection of the main header belonging to a file (the one that is included at the top) doesn't work properly

PR #675, opened 2020-12-07
Allow benchmark to be compiled with different precisions

thoasm:
This PR exports the precision all benchmark code is compiled in into the file `benchmark/utils/types.hpp`.
The `etype` defines in which precision the benchmark code is compiled in, meaning a change to `using etype = float;` runs the benchmark for single precision (see [this pipeline](https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/pipelines/226257409) for this case).

__TODO__:
- [x] Generate 2 binaries: one for `float`, one for `double`
- [x] Add `remove_complex` to a lot of function return types and test benchmark with complex values


benchmark/utils/general.hpp
```diff
@@ -366,8 +368,8 @@ double compute_residual_norm(const gko::LinOp *system_matrix,
 
 
 template <typename ValueType>
-double compute_max_relative_norm2(vec<ValueType> *result,
-                                  const vec<ValueType> *answer)
+ValueType compute_max_relative_norm2(vec<ValueType> *result,
```
tcojean, line 371: As usual for most of the functions/variables in this file, should these be valuetype of `remove_complex<ValueType>`?

benchmark/utils/cuda_linops.hpp
```diff
@@ -709,10 +710,10 @@ using cusp_gcoo = detail::CuspGenericCoo<>;
 
 #if defined(CUDA_VERSION) && (CUDA_VERSION < 11000)
 using cusp_coo =
-    detail::CuspHybrid<double, gko::int32, CUSPARSE_HYB_PARTITION_USER, 0>;
+    detail::CuspHybrid<etype, gko::int32, CUSPARSE_HYB_PARTITION_USER, 0>;
 using cusp_ell =
-    detail::CuspHybrid<double, gko::int32, CUSPARSE_HYB_PARTITION_MAX, 0>;
-using cusp_hybrid = detail::CuspHybrid<>;
+    detail::CuspHybrid<etype, gko::int32, CUSPARSE_HYB_PARTITION_MAX, 0>;
```
tcojean, line 715: For both here and HIP, we are probably not going to have all valuetypes work right? Should we throw an exception when that's the case, if some, where would be the most appropriate?

tcojean (commented):
I think LGTM though I did not review in depth. I have a few comments for now.

benchmark/utils/cuda_linops.hpp
thoasm: I think it is easier to rely on the `GKO_NOT_SUPPORTED`, so we don't have to update it in multiple places.

benchmark/utils/general.hpp
thoasm: Yes, I did not think about that, good catch.
In that case, I also need to test if the benchmark works with complex types like `std::complex<double>`.

benchmark/utils/general.hpp
```diff
@@ -297,6 +297,27 @@ std::unique_ptr<vec<ValueType>> create_vector(
...
+{
+    struct count {
+        using itype = gko::int32;
+        itype idx;
+        count() = default;
+        itype get_next_value() { return idx++; }
+    };
+    auto my_dist = [](count &c) {
+        return std::sin(
+            static_cast<gko::remove_complex<ValueType>>(c.get_next_value()));
```
Slaedr, line 313: Maybe we should consider having a specialization for complex `ValueType`s and have sinusoidal distribution for the imaginary parts as well. That may be helpful in some cases. Just an idea - we could also delay that to another PR when needed.

benchmark/solver/solver.cpp
```diff
@@ -486,32 +490,49 @@ int main(int argc, char *argv[])
...
+                    gko::dim<2>{system_matrix->get_size()[0], FLAGS_nrhs};
                 if (test_case.HasMember("rhs")) {
                     std::ifstream rhs_fd{test_case["rhs"].GetString()};
                     b = gko::read<Vec>(rhs_fd, exec);
                 } else {
-                    b = create_matrix<etype>(
-                        exec,
-                        gko::dim<2>{system_matrix->get_size()[0], FLAGS_nrhs},
-                        engine, FLAGS_random_rhs);
+                    // TODO: Remove (only present to perform benchmark for
```
Slaedr, line 512: Is it possible to resolve this todo now? Seems like you're making the sin RHS available permanently, in which case you can remove this comment.

benchmark/run_all_benchmarks.sh
```diff
@@ -199,7 +221,7 @@ run_preconditioner_benchmarks() {
         for prec in ${PRECISIONS}; do
             echo -e "\t\t running jacobi ($prec) for block size ${bsize}" 1>&2
             cp "$1" "$1.imd" # make sure we're not loosing the original input
-            ./preconditioner/preconditioner \
+            ./preconditioner/preconditioner${BENCH_SUFFIX} \
```
Slaedr, line 224: I guess this will not work for `float`. I suppose an exception will be thrown for that somewhere?

Slaedr (approved):
LGTM! A few minor questions and comments.

benchmark/run_all_benchmarks.sh
```diff
@@ -65,8 +65,27 @@ if [ ! "${SOLVERS_RHS}" ]; then
...
+    BENCH_SUFFIX=""
+elif [ "${BENCHMARK_PRECISION}" == "single" ]; then
+    BENCH_SUFFIX="_single"
+elif [ "${BENCHMARK_PRECISION}" == "dcomplex" ]; then
+    BENCH_SUFFIX="_dcomplex"
+elif [ "${BENCHMARK_PRECISION}" == "scomplex" ]; then
+    BENCH_SUFFIX="_scomplex"
+else
+    echo "BENCHMARK_PRECISION is set to the not supported \"${BENCHMARK_PRECISION}\"." 1>&2
+    echo "Currently supported values: \"double\" and \"single\"" 1>&2
```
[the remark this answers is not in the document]
thoasm: Thanks, I guess I missed this documentation when adding the complex values!

benchmark/utils/general.hpp
thoasm: The sinus benchmark is not supposed to be part of this PR (I will remove it soon, needed it for a benchmark run for a paper).
The current implementation behaves a bit different for complex values:
```
for real value types: s(i) = sin(i)
for complex value types: s(i) = complex{sin(2*i), sin(2*i + 1)}
```
Since I did not care about the complex type for my benchmark, I ignored this behaviour (since it will not make it into develop).

benchmark/solver/solver.cpp
thoasm: I still want to remove it. The sin-benchmark was added as the last commit, which I will simply remove before actually merging it (this is the reason this PR is currently marked as `WIP`).
I plan to keep all other changes of this PR, just the sinus benchmark portion will be removed.

benchmark/run_all_benchmarks.sh
thoasm: Why would this not work for `float`? I use the suffix `_single` for these benchmarks (the suffixes are defined in the [CMakeLists.txt here](https://github.com/ginkgo-project/ginkgo/blob/change_benchmark_precision/benchmark/CMakeLists.txt#L92), and the translation in the [run_all_benchmarks.sh here](https://github.com/ginkgo-project/ginkgo/blob/change_benchmark_precision/benchmark/run_all_benchmarks.sh#L68))

Slaedr: I mean because the block-Jacobi test seems to be going down two precisions in the array on https://github.com/ginkgo-project/ginkgo/pull/675/files#diff-def408061e3ffca8b3e80d667589f88c99f23d6fd482c0dde9b60579872f2938R211 I guess at least one of 0,2 or 2,0 implies you'll need a float8 if you start at float32? Correct me if I'm wrong please.

benchmark/solver/solver.cpp
Slaedr: Oh sorry, I completely missed the label change. I should have waited a bit, but it should be fine I guess.

benchmark/run_all_benchmarks.sh
thoasm: Ah, you meant specifically for the various precision we try in this loop.
It will work without any exceptions since this is supported by the Jacobi preconditioner.
These numbers represent how many steps down we go in terms of precision. If we reach the lowest supported precision, we simply don't go further. That means in case of single precision (`Jacobi<float>`), `PRECISION=0,1` will lead to the same result as `PRECISION=0,2`, meaning they will both use half precision for storage. The `PRECISION` values will directly go to `precision_reduction`, which is detailed [in the Jacobi documentation](https://github.com/ginkgo-project/ginkgo/blob/develop/include/ginkgo/core/preconditioner/jacobi.hpp#L418).

If we specify a lower precision than double precision, we do some redundant work, but I think that is fine, especially since we could add a lower precision than half (not likely, but possible).

Slaedr: Okay, I see. Thanks.

thoasm:
format!

thoasm:
format!

upsj (approved):
LGTM! Did you test the execution with all precision types?

benchmark/utils/cuda_linops.hpp
hartwiganzt: @tcojean if you are ok with this decision, I guess we can resolve this discussion and the pull request is ready to be merged.

benchmark/CMakeLists.txt
```diff
@@ -54,6 +54,53 @@ function(ginkgo_benchmark_hipsp_linops name)
...
+        if (GINKGO_BUILD_CUDA)
+            ginkgo_benchmark_cusp_linops("${name}")
+        endif()
+        if (GINKGO_BUILD_HIP)
+            ginkgo_benchmark_hipsp_linops("${name}")
+        endif()
+    endif()
+
+
+endfunction(ginkgo_add_single_benchmark_executable)
```
tcojean, line 82:
```suggestion
endfunction(ginkgo_add_single_benchmark_executable)
```

tcojean (approved):
LGTM.

thoasm:
@upsj Yes, I ran a very small benchmark with all preconditioner and solver, and did not encounter any runtime issues.
However, I realized that I did not call the correct solver-binary in the `run_all_benchmarks.sh` script (I am sure I did when I performed all the tests, but it did get lost at some point). I just pushed the correct benchmark script (I also tested all precisions again without any issues).

PR #676, opened 2020-12-15
Add an IndexSet class.

pratikvn:
This PR adds an Index set class which stores non-contiguous subsets of the form `[begin,end)`. It is non-dynamic, which means that it is not possible to add and remove subsets after creation. 

Internally, the class stores three separate arrays, containing the starts of the subsets, the end of the subsets and the superset index of the start of the subset, respectively. The class is optimized for storage by storing only the end points and hence is more efficient than storing complete index ranges. It aims to provide the following functionalities:

1. Local to global index conversions.
2. Global to local index conversions.
3. Checking the number of stored indices.
4. If given index is an element of the IndexSet.

### TODO

+ [x] Add documentation.
+ [x] Check for additional creation methods.


core/base/index_set.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+template <typename IndexType>
+Array<IndexType> IndexSet<IndexType>::get_global_indices(
+    const Array<IndexType> &local_indices) const
+{
+    auto exec = this->get_executor();
+    auto global_indices =
+        gko::Array<IndexType>(exec, local_indices.get_num_elems());
+
+    GKO_ASSERT(this->get_num_subsets() >= 1);
```
tcojean, line 139: Do you need to check if the number of elements of `local_indices` is inferior or equal to `num_stored_indices` or is that not an error?

reference/base/index_set_kernels.cpp
```diff
@@ -0,0 +1,191 @@
...
+        GKO_ASSERT(index < index_space_size);
+        auto bucket =
+            std::distance(subset_begin->get_const_data(),
+                          std::upper_bound(subset_begin->get_const_data(),
+                                           subset_begin->get_const_data() +
+                                               subset_begin->get_num_elems(),
+                                           index));
+        auto shifted_bucket = bucket == 0 ? 0 : (bucket - 1);
+        if (subset_end->get_const_data()[shifted_bucket] <= index) {
+            local_indices->get_data()[i] = -1;
```
tcojean, line 143: Use the `invalid` thing?

reference/base/index_set_kernels.cpp
```diff
@@ -0,0 +1,191 @@
...
+{
+    for (auto i = 0; i < global_indices->get_num_elems(); ++i) {
+        auto index = global_indices->get_const_data()[i];
+        GKO_ASSERT(index < index_space_size);
+        auto bucket =
+            std::distance(subset_begin->get_const_data(),
+                          std::upper_bound(subset_begin->get_const_data(),
+                                           subset_begin->get_const_data() +
+                                               subset_begin->get_num_elems(),
+                                           index));
```
tcojean, line 140: In practical usage (i.e. your usage in MPI), will the indices be sorted? If so, maybe an optimization/a specific function for the sorted version would make sense? You could start over from your previous location if the data was sorted or other faster algorithms.

include/ginkgo/core/base/index_set.hpp
```diff
@@ -0,0 +1,317 @@
...
+    void populate_subsets(const gko::Array<index_type> &indices);
+
+    std::shared_ptr<const gko::Executor> exec_;
+
+    index_type index_set_id_;
+    index_type index_space_size_;
+    index_type num_stored_indices_;
+    gko::Array<index_type> subsets_begin_;
+    gko::Array<index_type> subsets_end_;
+    gko::Array<index_type> superset_cumulative_indices_;
```
tcojean, line 310: For the sake of clarity, could you explain these maybe in the class description, maybe reusing the example you show in `get_local_index`.

tcojean (approved):
LGTM. I only have some small remarks or suggestions.

core/base/index_set.cpp
pratikvn: I think if for some reason, someone wants to have the same indices, then it is technically possible that the size of the `local_indices` array is larger than the number of stored indices. So, I would not necessarily check for that.

include/ginkgo/core/base/index_set.hpp
```diff
@@ -0,0 +1,317 @@
...
+     */
+    Array<index_type> get_local_indices(
+        const Array<index_type> &global_indices) const;
+
+    /**
+     * Checks if the element exists in the index set.
+     *
+     * @return  whether the element exists in the index set.
+     */
+    bool is_element(const index_type &index) const;
```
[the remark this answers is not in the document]
pratikvn: Yeah, possibly, but that information can also be taken from the `global_to_local` conversion and the array operations inside the kernels will use this (`global_to_local`) instead. But if needed, I think it should be simple to add them later.

reference/base/index_set_kernels.cpp
```diff
@@ -0,0 +1,191 @@
...
+                      const Array<IndexType> *indices,
+                      Array<IndexType> *subset_begin,
+                      Array<IndexType> *subset_end,
+                      Array<IndexType> *superset_indices)
+{
+    auto num_indices = indices->get_num_elems();
+    auto tmp_indices = gko::Array<IndexType>(*indices);
+    GKO_ASSERT(*std::max_element(indices->get_const_data(),
+                                 indices->get_const_data() + num_indices) <=
+               index_space_size);
```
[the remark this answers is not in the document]
tcojean: I mean, except if I missed something, I think you you can now change the assert to:
```suggestion
    GKO_ASSERT(tmp_indices.get_const_data()[num_indices-1]) <=
               index_space_size);
```
It's the same in the OpenMP code then.

yhmtsai:
does allocator.hpp require public?

pratikvn:
No, not necessarily for this PR, but I think it would be nice to have in the public interface. Or should we just leave it usable only within Ginkgo ? 

I don't see why users should not be able to write `gko::vector`. It gives them more flexibility if they want to test stuff out.

upsj:
As it is only usable with CPU executors, I would prefer if we move this to a separate PR and discuss the consequences there in more detail, as it is no longer necessary for this PR.

core/device_hooks/common_kernels.inc.cpp
```diff
@@ -109,6 +110,11 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
     _macro(IndexType) GKO_NOT_COMPILED(GKO_HOOK_MODULE); \
     GKO_INSTANTIATE_FOR_EACH_INDEX_TYPE(_macro)
 
+#define GKO_STUB_INDEX_AND_SIZE_TYPE(_macro)                 \
+    template <typename IndexSizeType>                        \
+    _macro(IndexSizeType) GKO_NOT_COMPILED(GKO_HOOK_MODULE); \
+    GKO_INSTANTIATE_FOR_EACH_INDEX_AND_SIZE_TYPE(_macro)
+
```
upsj, line 117: This can be removed

```suggestion
```

core/test/utils.hpp
```diff
@@ -77,6 +77,17 @@ using ComplexValueTypes =
...
+    ::testing::Types<gko::int32, gko::int64, gko::uint32, gko::uint64>;
+
+using AllTemplateTypes =
+#if GINKGO_DPCPP_SINGLE_MODE
+    ::testing::Types<float, std::complex<float>, gko::int32, gko::int64,
+                     gko::uint32, gko::uint64>;
+#else
+    ::testing::Types<float, double, std::complex<float>, std::complex<double>,
+                     gko::int32, gko::int64, gko::uint32, gko::uint64>;
+#endif
```
upsj, line 90:
```suggestion
```

dpcpp/matrix/csr_kernels.dp.cpp
```diff
@@ -50,6 +50,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/matrix/sellp.hpp>
 
 
+#include "core/base/allocator.hpp"
```
upsj, line 53: are we using this?

include/ginkgo/core/base/index_set.hpp
```diff
@@ -0,0 +1,444 @@
...
+    /**
+     * This function allows the user obtain a decompresed global_indices Array
+     * from the indices stored in the index set
+     *
+     * @return  the decompressed set of indices.
+     */
+    Array<index_type> to_global_indices() const;
+
+    /**
+     * Checks if the global index exists in the index set.
```
upsj, line 361:
```suggestion
     * Checks if the individual global indices exists in the index set.
```

include/ginkgo/core/base/index_set.hpp
```diff
@@ -0,0 +1,444 @@
...
+ * > subsets_begin = {1, 10, 18, 42}
+ * > subsets_end = {9, 13, 22, 43}
+ * > superset_cumulative_indices = {0, 8, 11, 15, 16}
+ *
+ * @tparam index_type  type of the indices being stored in the index set.
+ *
+ * @ingroup IndexSet
+ */
+template <typename IndexType = int32>
+class IndexSet {
```
upsj, line 84: Now that I think of it, you could probably use PolymorphicObject to provide the executor functionality and avoid the need for all of these operator overloads :)

include/ginkgo/core/base/types.hpp
```diff
@@ -532,6 +532,20 @@ GKO_ATTRIBUTES constexpr bool operator!=(precision_reduction x,
...
+ *                (not including the leading `template` specifier).
+ *                Should take one argument, which is replaced by the
+ *                value type.
+ */
+#define GKO_INSTANTIATE_FOR_EACH_INDEX_AND_SIZE_TYPE(_macro) \
+    GKO_INSTANTIATE_FOR_EACH_INDEX_TYPE(_macro);             \
+    template _macro(uint32);                                 \
+    template _macro(uint64)
+
+
```
upsj, line 548:
```suggestion
```

include/ginkgo/core/base/types.hpp
```diff
@@ -626,8 +640,43 @@ GKO_ATTRIBUTES constexpr bool operator!=(precision_reduction x,
...
+inline IndexType invalid_index()
+{
+    static_assert(std::is_signed<IndexType>::value,
+                  "IndexType needs to be signed");
+    return static_cast<IndexType>(-1);
+}
+
+#define GKO_DECLARE_INVALID_TYPES(_itype) _itype invalid_index()
+
+GKO_INSTANTIATE_FOR_EACH_INDEX_TYPE(GKO_DECLARE_INVALID_TYPES);
```
upsj, line 659: This is a header, so explicit instantiations are not necessary.
```suggestion
```

include/ginkgo/core/base/types.hpp
```diff
@@ -626,8 +640,43 @@ GKO_ATTRIBUTES constexpr bool operator!=(precision_reduction x,
...
+
+
+/**
+ * Value for an invalid unsigned long.
+ */
+template <>
+inline uint64 invalid_index<uint64>()
+{
+    return std::numeric_limits<uint64>::max();
+}
```
upsj, line 679:
```suggestion
```

omp/test/components/fill_array.cpp
```diff
@@ -74,8 +74,7 @@ class FillArray : public ::testing::Test {
     gko::Array<value_type> seqs;
 };
 
-TYPED_TEST_SUITE(FillArray, gko::test::ValueAndIndexTypes,
-                 TypenameNameGenerator);
+TYPED_TEST_SUITE(FillArray, gko::test::AllTemplateTypes, TypenameNameGenerator);
```
upsj, line 77: this change seems left over from the unsigned index approach.

omp/test/components/reduce_array.cpp
```diff
@@ -74,7 +74,7 @@ class ReduceArray : public ::testing::Test {
     gko::Array<value_type> dvals;
 };
 
-TYPED_TEST_SUITE(ReduceArray, gko::test::ValueAndIndexTypes,
+TYPED_TEST_SUITE(ReduceArray, gko::test::AllTemplateTypes,
```
upsj, line 77: same here

reference/test/base/array.cpp
```diff
@@ -61,7 +61,7 @@ class Array : public ::testing::Test {
     gko::Array<T> x;
 };
 
-TYPED_TEST_SUITE(Array, gko::test::ValueAndIndexTypes, TypenameNameGenerator);
+TYPED_TEST_SUITE(Array, gko::test::AllTemplateTypes, TypenameNameGenerator);
```
upsj, line 64: same here

reference/test/components/fill_array.cpp
```diff
@@ -72,8 +72,7 @@ class FillArray : public ::testing::Test {
     gko::Array<value_type> seqs;
 };
 
-TYPED_TEST_SUITE(FillArray, gko::test::ValueAndIndexTypes,
-                 TypenameNameGenerator);
+TYPED_TEST_SUITE(FillArray, gko::test::AllTemplateTypes, TypenameNameGenerator);
```
upsj, line 75: same here

common/unified/base/index_set_kernels.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+template <typename IndexType>
+void compute_validity(std::shared_ptr<const DefaultExecutor> exec,
+                      const Array<IndexType>* local_indices,
+                      Array<bool>* validity_array)
+{
+    auto invalid_idx = invalid_index<IndexType>();
+    run_kernel(
+        exec,
+        [invalid_idx] GKO_KERNEL(auto elem, auto local_indices,
```
upsj, line 61: I'd avoid using Lambda captures in this context, since we encountered issues with CUDA before. Probably related to the fact that this is a generic Lambda?

core/base/index_set.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+template <typename IndexType>
+void IndexSet<IndexType>::populate_subsets(const gko::Array<IndexType>& indices,
+                                           const bool is_sorted)
+{
+    auto exec = this->get_executor();
+    this->num_stored_indices_ = indices.get_num_elems();
+    exec->run(index_set::make_populate_subsets(
+        this->index_space_size_, &indices, &this->subsets_begin_,
+        &this->subsets_end_, &this->superset_cumulative_indices_, is_sorted));
```
upsj, line 70: nit: why pass arrays as pointers? passing them as references saves & and *

core/base/index_set.cpp
```diff
@@ -0,0 +1,162 @@
...
+    auto exec = this->get_executor();
+    auto global_indices =
+        gko::Array<IndexType>(exec, local_indices.get_num_elems());
+
+    GKO_ASSERT(this->get_num_subsets() >= 1);
+    exec->run(index_set::make_local_to_global(
+        this->index_space_size_, &this->subsets_begin_, &this->subsets_end_,
+        &this->superset_cumulative_indices_, &local_indices, &global_indices,
+        is_sorted));
+    return std::move(global_indices);
```
upsj, line 137: nit: std::move makes copy elision/return value optimization more difficult for the compiler
```suggestion
    return global_indices;
```

core/base/index_set.cpp
```diff
@@ -0,0 +1,162 @@
...
+    auto exec = this->get_executor();
+    auto local_indices =
+        gko::Array<IndexType>(exec, global_indices.get_num_elems());
+
+    GKO_ASSERT(this->get_num_subsets() >= 1);
+    exec->run(index_set::make_global_to_local(
+        this->index_space_size_, &this->subsets_begin_, &this->subsets_end_,
+        &this->superset_cumulative_indices_, &global_indices, &local_indices,
+        is_sorted));
+    return std::move(local_indices);
```
upsj, line 154: same here

upsj (changes requested):
Sorry this review comes so late, I have a few suggestions and change requests for the public interface. Mostly related to index types, and type hierarchy. Do you still want to limit yourself to the case of global_index_type == local_index_type?

If you want to play around and get familiar with Thrust a bit, the vectorized binary searches etc. should give you a nice tool to implement most of the CUDA and HIP algorithms :)

include/ginkgo/core/base/index_set.hpp
pratikvn: That is a good point!

omp/test/components/fill_array.cpp
pratikvn: `ValueAndIndexTypes` still instantiates for `size_type`. Should we keep that ?

core/base/index_set.cpp
pratikvn: Just a personal preference from my side. It uniformizes reading the kernel when passing plain pointers, I can use `->` everywhere, instead of switching between `->` and `.`

pratikvn:
Regarding the global and local index_type, I think a single type would be okay, as this is meant to be used internally and I am not sure there is any benefit in separating the two types.

upsj (commented):
If you wait another day, I can add CUDA and HIP kernels :)

pratikvn:
@upsj, I was working on it as well. :) I wanted to create a separate PR for that.

PR #677, opened 2020-12-17
Support Mixed precision SpMV and BLAS operations

upsj:
This PR adds full mixed precision support to Ginkgo, at least in terms of LinOp compatibility.
By default, this happens with the new `make_temporary_conversion` helper that works similarly to `make_temporary_clone` and is wrapped in the `precision_dispatch` function, which applies the correct precision and complex-to-real conversions for SpMV and preconditioner applications. For solvers, a more explicit conversion is necessary.

The temporary conversion wrapper won't give ideal performance, since it requires additional conversions from and to the input/output vectors as well as associated allocations of temporary memory for each apply operation. For solvers, this might still pay off due to the long runtime of a single `apply`. Mixed Precision IR is then almost equivalent to the following
```cpp
auto solver =
    gko::solver::Ir<double>::build()
        .with_solver(
            gko::solver::Gmres<float>::build()
                .with_criteria(
                    gko::stop::ResidualNormReduction<SolverType>::build()
                        .with_reduction_factor(inner_reduction_factor)
                        .on(exec),
                    gko::stop::Iteration::build()
                        .with_max_iters(max_inner_iters)
                        .on(exec))
                .on(exec))
        .with_criteria(gko::stop::ResidualNormReduction<ValueType>::build()
                            .with_reduction_factor(outer_reduction_factor)
                            .on(exec),
                        gko::stop::Iteration::build()
                            .with_max_iters(max_outer_iters)
                            .on(exec))
        .on(exec)
        ->generate(give(A));
```
except for the fact that A is always stored and computed on in double precision.

TODO: 
- [x] Comprehensive Reference Tests
- [x] ~~No need to convert `x` to ValueType for LinOps where `apply_uses_initial_guess() == false`~~ This would be way too complex and probably be overkill, since this is not meant to provide good performance
- [x] ~~Modify MPIR example~~

### DIFF
--- a/core/matrix/diagonal.cpp
+++ b/core/matrix/diagonal.cpp
@@ -34,6 +34,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
 #include <ginkgo/core/base/exception_helpers.hpp>
+#include <ginkgo/core/base/precision_dispatch.hpp>
 #include <ginkgo/core/base/utils.hpp>
 #include <ginkgo/core/matrix/dense.hpp>
 
@@ -67,12 +68,8 @@ void Diagonal<ValueType>::apply_impl(const LinOp *b, LinOp *x) const
 {
     auto exec = this->get_executor();
 
-    if (dynamic_cast<const Dense<ValueType> *>(b) &&
-        dynamic_cast<Dense<ValueType> *>(x)) {
-        exec->run(diagonal::make_apply_to_dense(this, as<Dense<ValueType>>(b),
-                                                as<Dense<ValueType>>(x)));
-    } else if (dynamic_cast<const Csr<ValueType, int32> *>(b) &&
-               dynamic_cast<Csr<ValueType, int32> *>(x)) {
+    if (dynamic_cast<const Csr<ValueType, int32> *>(b) &&
+        dynamic_cast<Csr<ValueType, int32> *>(x)) {
         exec->run(diagonal::make_apply_to_csr(
             this, as<Csr<ValueType, int32>>(b), as<Csr<ValueType, int32>>(x)));
     } else if (dynamic_cast<const Csr<ValueType, int64> *>(b) &&
@@ -80,7 +77,12 @@ void Diagonal<ValueType>::apply_impl(const LinOp *b, LinOp *x) const
         exec->run(diagonal::make_apply_to_csr(
             this, as<Csr<ValueType, int64>>(b), as<Csr<ValueType, int64>>(x)));
     } else {
-        GKO_NOT_IMPLEMENTED;
+        precision_dispatch_real_complex<ValueType>(
+            [this, &exec](auto dense_b, auto dense_x) {
+                exec->run(
+                    diagonal::make_apply_to_dense(this, dense_b, dense_x));
+            },
+            b, x);
     }
 }
 
@@ -90,12 +92,8 @@ void Diagonal<ValueType>::rapply_impl(const LinOp *b, LinOp *x) const
 {
     auto exec = this->get_executor();
 
-    if (dynamic_cast<const Dense<ValueType> *>(b) &&
-        dynamic_cast<Dense<ValueType> *>(x)) {
-        exec->run(diagonal::make_right_apply_to_dense(
-            this, as<Dense<ValueType>>(b), as<Dense<ValueType>>(x)));
-    } else if (dynamic_cast<const Csr<ValueType, int32> *>(b) &&
-               dynamic_cast<Csr<ValueType, int32> *>(x)) {
+    if (dynamic_cast<const Csr<ValueType, int32> *>(b) &&
+        dynamic_cast<Csr<ValueType, int32> *>(x)) {
         exec->run(diagonal::make_right_apply_to_csr(
             this, as<Csr<ValueType, int32>>(b), as<Csr<ValueType, int32>>(x)));
     } else if (dynamic_cast<const Csr<ValueType, int64> *>(b) &&
@@ -103,7 +101,14 @@ void Diagonal<ValueType>::rapply_impl(const LinOp *b, LinOp *x) const
         exec->run(diagonal::make_right_apply_to_csr(
             this, as<Csr<ValueType, int64>>(b), as<Csr<ValueType, int64>>(x)));
     } else {
-        GKO_NOT_IMPLEMENTED;
+        // no real-to-complex conversion, as this would require doubling the
+        // diagonal entries for the complex-to-real columns
+        precision_dispatch<ValueType>(
+            [this, &exec](auto dense_b, auto dense_x) {
+                exec->run(diagonal::make_right_apply_to_dense(this, dense_b,
+                                                              dense_x));
+            },
+            b, x);
     }
 }
 
@@ -112,14 +117,14 @@ template <typename ValueType>
 void Diagonal<ValueType>::apply_impl(const LinOp *alpha, const LinOp *b,
                                      const LinOp *beta, LinOp *x) const
 {
-    if (dynamic_cast<const Dense<ValueType> *>(b) &&
-        dynamic_cast<Dense<ValueType> *>(x)) {
-        auto dense_x = as<Dense<ValueType>>(x);
-        dense_x->scale(beta);
-        dense_x->add_scaled(alpha, b);
-    } else {
-        GKO_NOT_IMPLEMENTED;
-    }
+    precision_dispatch_real_complex<ValueType>(
+        [this](auto dense_alpha, auto dense_b, auto dense_beta, auto dense_x) {
+            auto x_clone = dense_x->clone();
+            this->apply(dense_b, x_clone.get());
+            dense_x->scale(dense_beta);
+            dense_x->add_scaled(dense_alpha, x_clone.get());
+        },
+        alpha, b, beta, x);
 }
 
 
--- a/core/matrix/diagonal.cpp
+++ b/core/matrix/diagonal.cpp
@@ -120,7 +120,7 @@ void Diagonal<ValueType>::apply_impl(const LinOp *alpha, const LinOp *b,
     precision_dispatch_real_complex<ValueType>(
         [this](auto dense_alpha, auto dense_b, auto dense_beta, auto dense_x) {
             auto x_clone = dense_x->clone();
-            this->apply(dense_b, x_clone.get());
+            this->apply_impl(dense_b, x_clone.get());
             dense_x->scale(dense_beta);
             dense_x->add_scaled(dense_alpha, x_clone.get());
         },
--- a/core/preconditioner/jacobi.cpp
+++ b/core/preconditioner/jacobi.cpp
@@ -39,6 +39,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/base/exception_helpers.hpp>
 #include <ginkgo/core/base/executor.hpp>
 #include <ginkgo/core/base/math.hpp>
+#include <ginkgo/core/base/precision_dispatch.hpp>
 #include <ginkgo/core/base/utils.hpp>
 #include <ginkgo/core/matrix/csr.hpp>
 #include <ginkgo/core/matrix/dense.hpp>
@@ -71,11 +72,14 @@ GKO_REGISTER_OPERATION(initialize_precisions, jacobi::initialize_precisions);
 template <typename ValueType, typename IndexType>
 void Jacobi<ValueType, IndexType>::apply_impl(const LinOp *b, LinOp *x) const
 {
-    using dense = matrix::Dense<ValueType>;
-    this->get_executor()->run(jacobi::make_simple_apply(
-        num_blocks_, parameters_.max_block_size, storage_scheme_,
-        parameters_.storage_optimization.block_wise, parameters_.block_pointers,
-        blocks_, as<dense>(b), as<dense>(x)));
+    precision_dispatch_real_complex<ValueType>(
+        [this](auto dense_b, auto dense_x) {
+            this->get_executor()->run(jacobi::make_simple_apply(
+                num_blocks_, parameters_.max_block_size, storage_scheme_,
+                parameters_.storage_optimization.block_wise,
+                parameters_.block_pointers, blocks_, dense_b, dense_x));
+        },
+        b, x);
 }
 
 
@@ -84,12 +88,15 @@ void Jacobi<ValueType, IndexType>::apply_impl(const LinOp *alpha,
                                               const LinOp *b, const LinOp *beta,
                                               LinOp *x) const
 {
-    using dense = matrix::Dense<ValueType>;
-    this->get_executor()->run(jacobi::make_apply(
-        num_blocks_, parameters_.max_block_size, storage_scheme_,
-        parameters_.storage_optimization.block_wise, parameters_.block_pointers,
-        blocks_, as<dense>(alpha), as<dense>(b), as<dense>(beta),
-        as<dense>(x)));
+    precision_dispatch_real_complex<ValueType>(
+        [this](auto dense_alpha, auto dense_b, auto dense_beta, auto dense_x) {
+            this->get_executor()->run(jacobi::make_apply(
+                num_blocks_, parameters_.max_block_size, storage_scheme_,
+                parameters_.storage_optimization.block_wise,
+                parameters_.block_pointers, blocks_, dense_alpha, dense_b,
+                dense_beta, dense_x));
+        },
+        alpha, b, beta, x);
 }
 
 
@@ -220,8 +227,8 @@ void Jacobi<ValueType, IndexType>::generate(const LinOp *system_matrix,
 
     const auto all_block_opt = parameters_.storage_optimization.of_all_blocks;
     auto &precisions = parameters_.storage_optimization.block_wise;
-    // if adaptive version is used, make sure that the precision array is of the
-    // correct size by replicating it multiple times if needed
+    // if adaptive version is used, make sure that the precision array is of
+    // the correct size by replicating it multiple times if needed
     if (parameters_.storage_optimization.is_block_wise ||
         all_block_opt != precision_reduction(0, 0)) {
         if (!parameters_.storage_optimization.is_block_wise) {
--- a/core/solver/bicg.cpp
+++ b/core/solver/bicg.cpp
@@ -38,6 +38,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/base/executor.hpp>
 #include <ginkgo/core/base/math.hpp>
 #include <ginkgo/core/base/name_demangling.hpp>
+#include <ginkgo/core/base/precision_dispatch.hpp>
 #include <ginkgo/core/base/utils.hpp>
 
 
@@ -110,6 +111,18 @@ std::unique_ptr<LinOp> conj_transpose_with_csr(const LinOp *mtx)
 template <typename ValueType>
 void Bicg<ValueType>::apply_impl(const LinOp *b, LinOp *x) const
 {
+    precision_dispatch_real_complex<ValueType>(
+        [this](auto dense_b, auto dense_x) {
+            this->apply_dense_impl(dense_b, dense_x);
+        },
+        b, x);
+}
+
+
+template <typename ValueType>
+void Bicg<ValueType>::apply_dense_impl(const matrix::Dense<ValueType> *dense_b,
+                                       matrix::Dense<ValueType> *dense_x) const
+{
     using std::swap;
     using Vector = matrix::Dense<ValueType>;
     constexpr uint8 RelativeStoppingId{1};
@@ -119,8 +132,6 @@ void Bicg<ValueType>::apply_impl(const LinOp *b, LinOp *x) const
     auto one_op = initialize<Vector>({one<ValueType>()}, exec);
     auto neg_one_op = initialize<Vector>({-one<ValueType>()}, exec);
 
-    auto dense_b = as<const Vector>(b);
-    auto dense_x = as<Vector>(x);
     auto r = Vector::create_with_config_of(dense_b);
     auto r2 = Vector::create_with_config_of(dense_b);
     auto z = Vector::create_with_config_of(dense_b);
@@ -179,8 +190,9 @@ void Bicg<ValueType>::apply_impl(const LinOp *b, LinOp *x) const
     r2->copy_from(r.get());
     // r2 = r
     auto stop_criterion = stop_criterion_factory_->generate(
-        system_matrix_, std::shared_ptr<const LinOp>(b, [](const LinOp *) {}),
-        x, r.get());
+        system_matrix_,
+        std::shared_ptr<const LinOp>(dense_b, [](const LinOp *) {}), dense_x,
+        r.get());
 
     int iter = -1;
 
@@ -234,12 +246,14 @@ template <typename ValueType>
 void Bicg<ValueType>::apply_impl(const LinOp *alpha, const LinOp *b,
                                  const LinOp *beta, LinOp *x) const
 {
-    auto dense_x = as<matrix::Dense<ValueType>>(x);
-
-    auto x_clone = dense_x->clone();
-    this->apply(b, x_clone.get());
-    dense_x->scale(beta);
-    dense_x->add_scaled(alpha, x_clone.get());
+    precision_dispatch_real_complex<ValueType>(
+        [this](auto dense_alpha, auto dense_b, auto dense_beta, auto dense_x) {
+            auto x_clone = dense_x->clone();
+            this->apply(dense_b, x_clone.get());
+            dense_x->scale(dense_beta);
+            dense_x->add_scaled(dense_alpha, x_clone.get());
+        },
+        alpha, b, beta, x);
 }
 
 
--- a/core/solver/bicg.cpp
+++ b/core/solver/bicg.cpp
@@ -249,7 +249,7 @@ void Bicg<ValueType>::apply_impl(const LinOp *alpha, const LinOp *b,
     precision_dispatch_real_complex<ValueType>(
         [this](auto dense_alpha, auto dense_b, auto dense_beta, auto dense_x) {
             auto x_clone = dense_x->clone();
-            this->apply(dense_b, x_clone.get());
+            this->apply_dense_impl(dense_b, x_clone.get());
             dense_x->scale(dense_beta);
             dense_x->add_scaled(dense_alpha, x_clone.get());
         },
--- a/core/solver/bicgstab.cpp
+++ b/core/solver/bicgstab.cpp
@@ -228,7 +228,7 @@ void Bicgstab<ValueType>::apply_impl(const LinOp *alpha, const LinOp *b,
     precision_dispatch_real_complex<ValueType>(
         [this](auto dense_alpha, auto dense_b, auto dense_beta, auto dense_x) {
             auto x_clone = dense_x->clone();
-            this->apply(dense_b, x_clone.get());
+            this->apply_dense_impl(dense_b, x_clone.get());
             dense_x->scale(dense_beta);
             dense_x->add_scaled(dense_alpha, x_clone.get());
         },
--- a/core/solver/bicgstab.cpp
+++ b/core/solver/bicgstab.cpp
@@ -37,6 +37,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/base/exception_helpers.hpp>
 #include <ginkgo/core/base/executor.hpp>
 #include <ginkgo/core/base/math.hpp>
+#include <ginkgo/core/base/precision_dispatch.hpp>
 #include <ginkgo/core/base/utils.hpp>
 
 
@@ -87,6 +88,19 @@ std::unique_ptr<LinOp> Bicgstab<ValueType>::conj_transpose() const
 template <typename ValueType>
 void Bicgstab<ValueType>::apply_impl(const LinOp *b, LinOp *x) const
 {
+    precision_dispatch_real_complex<ValueType>(
+        [this](auto dense_b, auto dense_x) {
+            this->apply_dense_impl(dense_b, dense_x);
+        },
+        b, x);
+}
+
+
+template <typename ValueType>
+void Bicgstab<ValueType>::apply_dense_impl(
+    const matrix::Dense<ValueType> *dense_b,
+    matrix::Dense<ValueType> *dense_x) const
+{
     using std::swap;
     using Vector = matrix::Dense<ValueType>;
     using AbsVector = matrix::Dense<remove_complex<ValueType>>;
@@ -98,8 +112,6 @@ void Bicgstab<ValueType>::apply_impl(const LinOp *b, LinOp *x) const
     auto one_op = initialize<Vector>({one<ValueType>()}, exec);
     auto neg_one_op = initialize<Vector>({-one<ValueType>()}, exec);
 
-    auto dense_b = as<Vector>(b);
-    auto dense_x = as<Vector>(x);
     auto r = Vector::create_with_config_of(dense_b);
     auto z = Vector::create_with_config_of(dense_b);
     auto y = Vector::create_with_config_of(dense_b);
@@ -132,8 +144,9 @@ void Bicgstab<ValueType>::apply_impl(const LinOp *b, LinOp *x) const
 
     system_matrix_->apply(neg_one_op.get(), dense_x, one_op.get(), r.get());
     auto stop_criterion = stop_criterion_factory_->generate(
-        system_matrix_, std::shared_ptr<const LinOp>(b, [](const LinOp *) {}),
-        x, r.get());
+        system_matrix_,
+        std::shared_ptr<const LinOp>(dense_b, [](const LinOp *) {}), dense_x,
+        r.get());
     rr->copy_from(r.get());
 
     int iter = -1;
@@ -205,18 +218,21 @@ void Bicgstab<ValueType>::apply_impl(const LinOp *b, LinOp *x) const
             beta.get(), gamma.get(), omega.get(), &stop_status));
         swap(prev_rho, rho);
     }
-}  // namespace solver
+}
 
 
 template <typename ValueType>
 void Bicgstab<ValueType>::apply_impl(const LinOp *alpha, const LinOp *b,
                                      const LinOp *beta, LinOp *x) const
 {
-    auto dense_x = as<matrix::Dense<ValueType>>(x);
-    auto x_clone = dense_x->clone();
-    this->apply(b, x_clone.get());
-    dense_x->scale(beta);
-    dense_x->add_scaled(alpha, x_clone.get());
+    precision_dispatch_real_complex<ValueType>(
+        [this](auto dense_alpha, auto dense_b, auto dense_beta, auto dense_x) {
+            auto x_clone = dense_x->clone();
+            this->apply(dense_b, x_clone.get());
+            dense_x->scale(dense_beta);
+            dense_x->add_scaled(dense_alpha, x_clone.get());
+        },
+        alpha, b, beta, x);
 }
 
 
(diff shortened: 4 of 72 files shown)


core/base/temporary_conversion.hpp
```diff
@@ -0,0 +1,229 @@
...
+    temporary_conversion(handle_type handle) : handle_{std::move(handle)} {}
+
+    handle_type handle_;
+};
+
+
+}  // namespace detail
+
+
+}  // namespace gko
```
tcojean, line 226:
```suggestion
}  // namespace detail
}  // namespace gko
```

core/base/temporary_conversion.hpp
```diff
@@ -0,0 +1,229 @@
...
+
+
+#include <ginkgo/core/base/lin_op.hpp>
+#include <ginkgo/core/base/utils.hpp>
+
+
+namespace gko {
+
+
+namespace detail {
```
tcojean, line 47: I don't think we need to make the detail stand out so much.

```suggestion
namespace gko {
namespace detail {
```

core/base/precision_dispatch.hpp
```diff
@@ -0,0 +1,153 @@
...
+    }
+    return result;
+}
+
+
+template <typename ValueType, typename Function, typename... Args>
+void precision_dispatch(Function fn, Args *... linops)
+{
+    fn(make_temporary_conversion<ValueType>(linops).get()...);
+}
```
tcojean, line 77: I think these functions need some comments. On aspect is w.r.t. memory management. If I understood the code properly, the converted subtypes are valid only inside the scope of the function `fn`, they get initialized at the beginning (using this temporary conversion), and deleted at the end of the function?

core/base/precision_dispatch.hpp
```diff
@@ -0,0 +1,153 @@
...
+
+template <typename ValueType, typename Function>
+void precision_dispatch_spmv(Function fn, const LinOp *in, LinOp *out)
+{
+    // do we need to convert complex Dense to real Dense?
+    auto complex_to_real =
+        !(is_complex<ValueType>() ||
+          dynamic_cast<const ConvertibleTo<matrix::Dense<>> *>(in));
```
tcojean, line 85: In general I think this part really needs some explanation because it's a real trick. Either that or the longer implementation which checks the different combinations would be more clear. This reads (after integrating the not):
targeted type isn't complex and `in` isn't convertible to `Dense<double>`?

Basically, what I understand you are doing is depending on the fact that there is conversion to `Dense<next_precision<ValueType>>` which when float always evaluates to double, and vice versa. Thus, with the current implementation a `Dense` of `std::complex<double>` can convert to either `std::complex<double>` or  `std::complex<float>` and same for the simple `double` and `float` version. Thus, checking what you do here means checking against either `float` and `double` at the same time actually, although it appears to only check `double` at first.

This works of course but is maybe fragile (depends on the evolution of our conversion management) and is quite obfuscated. Something needs to be done to make this clearer at least a `WARN/NOTE` tag I think.

core/base/precision_dispatch.hpp
```diff
@@ -0,0 +1,153 @@
...
+        // If ValueType is real, the cast is a no-op
+        fn(dense_alpha.get(),
+           dynamic_cast<const Dense *>(dense_in->create_real_view().get()),
+           dense_beta.get(),
+           dynamic_cast<Dense *>(dense_out->create_real_view().get()));
+    } else {
+        precision_dispatch<ValueType>(fn, alpha, in, beta, out);
+    }
+}
+
```
tcojean, line 149:
```suggestion


```

core/base/precision_dispatch.hpp
```diff
@@ -0,0 +1,153 @@
...
+    return result;
+}
+
+
+template <typename ValueType, typename Function, typename... Args>
+void precision_dispatch(Function fn, Args *... linops)
+{
+    fn(make_temporary_conversion<ValueType>(linops).get()...);
+}
+
```
tcojean, line 78:
```suggestion


```

core/base/precision_dispatch.hpp
```diff
@@ -0,0 +1,153 @@
...
+        // These dynamic_casts are only needed to make the code compile
+        // If ValueType is complex, this branch will never be taken
+        // If ValueType is real, the cast is a no-op
+        fn(dynamic_cast<const Dense *>(dense_in->create_real_view().get()),
+           dynamic_cast<Dense *>(dense_out->create_real_view().get()));
+    } else {
+        precision_dispatch<ValueType>(fn, in, out);
+    }
+}
+
```
tcojean, line 99: As they are only different overloads not specialization I think we can separate them properly.
```suggestion


```

core/base/precision_dispatch.hpp
```diff
@@ -0,0 +1,153 @@
...
+        // If ValueType is complex, this branch will never be taken
+        // If ValueType is real, the cast is a no-op
+        fn(dense_alpha.get(),
+           dynamic_cast<const Dense *>(dense_in->create_real_view().get()),
+           dynamic_cast<Dense *>(dense_out->create_real_view().get()));
+    } else {
+        precision_dispatch<ValueType>(fn, alpha, in, out);
+    }
+}
+
```
tcojean, line 123:
```suggestion


```

core/matrix/diagonal.cpp
```diff
@@ -67,20 +68,21 @@ void Diagonal<ValueType>::apply_impl(const LinOp *b, LinOp *x) const
...
             this, as<Csr<ValueType, int64>>(b), as<Csr<ValueType, int64>>(x)));
     } else {
-        GKO_NOT_IMPLEMENTED;
+        precision_dispatch_spmv<ValueType>(
+            [&](auto dense_b, auto dense_x) {
+                exec->run(
+                    diagonal::make_apply_to_dense(this, dense_b, dense_x));
+            },
+            b, x);
     }
```
tcojean, line 86: In terms of functionality that should work as all formats can convert to Dense, but do we want that to happen, what about other formats? We  could keep a `GKO_NOT_IMPLEMENTED` as well.

tcojean: I guess the new form corresponds more to what is in other cases like for CSR where we assume all other cases to be dense.

core/preconditioner/jacobi.cpp
```diff
@@ -71,11 +72,14 @@ GKO_REGISTER_OPERATION(initialize_precisions, jacobi::initialize_precisions);
 template <typename ValueType, typename IndexType>
 void Jacobi<ValueType, IndexType>::apply_impl(const LinOp *b, LinOp *x) const
 {
-    using dense = matrix::Dense<ValueType>;
-    this->get_executor()->run(jacobi::make_simple_apply(
-        num_blocks_, parameters_.max_block_size, storage_scheme_,
-        parameters_.storage_optimization.block_wise, parameters_.block_pointers,
-        blocks_, as<dense>(b), as<dense>(x)));
+    precision_dispatch_spmv<ValueType>(
```
tcojean, line 75: Do we need the complex to real case here and would it/does it work?

core/solver/bicg.cpp
```diff
@@ -234,12 +237,14 @@ template <typename ValueType>
 void Bicg<ValueType>::apply_impl(const LinOp *alpha, const LinOp *b,
                                  const LinOp *beta, LinOp *x) const
 {
-    auto dense_x = as<matrix::Dense<ValueType>>(x);
-
-    auto x_clone = dense_x->clone();
-    this->apply(b, x_clone.get());
-    dense_x->scale(beta);
-    dense_x->add_scaled(alpha, x_clone.get());
+    precision_dispatch_spmv<ValueType>(
```
tcojean, line 240: The solvers do not take into account the complex to real case for the simple apply. We probably need to do the same in the advanced apply?

core/base/precision_dispatch.hpp
```diff
@@ -0,0 +1,153 @@
...
+
+template <typename ValueType, typename Function, typename... Args>
+void precision_dispatch(Function fn, Args *... linops)
+{
+    fn(make_temporary_conversion<ValueType>(linops).get()...);
+}
+
+template <typename ValueType, typename Function>
+void precision_dispatch_spmv(Function fn, const LinOp *in, LinOp *out)
```
tcojean, line 80: A comment on the name, if only SpMVs use it I guess the naming is fine, although not very descriptive. If it stays like now where some preconditioners as well as solvers use it in some cases, I think we should go for something else, like: `precision_dispatch_complex` or `precision_dispatch_manage_cpx` or whatever is a better name.

include/ginkgo/core/solver/idr.hpp
```diff
@@ -251,7 +251,8 @@ class Idr : public EnableLinOp<Idr<ValueType>>,
                     LinOp *x) const override;
 
     template <typename SubspaceType>
-    void iterate(const LinOp *b, LinOp *x) const;
+    void iterate(const matrix::Dense<SubspaceType> *dense_b,
+                 matrix::Dense<SubspaceType> *dense_x) const;
```
tcojean, line 255: Technically that's an interface break? Though at least IDR was not released yet.

tcojean (commented):
There is some mistakes I could find due to copy/paste. Of course tests are missing. In addition, I think you should change the mixed precision IR example to reflect the new structure, or add an extra example/case in the same example.

Another question, can anything be done about ISAI? I think changing the TRS might be enough for ILU and IC, but ISAI might need some more changes particularly since it stores the improximate inverse as a plain CSR matrix?

core/base/temporary_conversion.hpp
```diff
@@ -0,0 +1,229 @@
...
+
+    void operator()(pointer ptr) const { delete ptr; }
+};
+
+
+template <typename... ConversionCandidates>
+struct conversion_helper {
+    using first_candidate =
+        typename std::tuple_element<0,
+                                    std::tuple<ConversionCandidates...>>::type;
```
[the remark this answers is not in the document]
upsj: Yes, this was leftover from a previous implementation attempt

core/preconditioner/jacobi.cpp
upsj: I don't see any reason why it shouldn't, since multiple RHS seem to work. It would probably be necessary for BJ-preconditioned IDR

include/ginkgo/core/solver/idr.hpp
upsj: Good point, I thought that member function was private.

core/matrix/diagonal.cpp
yhmtsai: I prefer staying `GKO_NOT_IMPLEMENTED`. Converting to dense may needs a lot of storage

upsj: True, I hadn't even considered the sparse -> dense conversions here. So I will instead have to test against all Dense types instead.

include/ginkgo/core/solver/idr.hpp
pratikvn: As you mention, as IDR is not released, I also think it should be okay. Even if spack or someone else pulls develop, we never guarantee that develop has a consistent interface in itself.

core/base/temporary_conversion.hpp
```diff
@@ -0,0 +1,229 @@
...
+                                FirstCandidate>::value,
+                "ConvertibleTo not implemented");
+            static_assert(std::is_base_of<ConvertibleTo<FirstCandidate>,
+                                          TargetType>::value,
+                          "ConvertibleTo not symmetric");
+            return {converted.release(),
+                    convert_back_deleter<TargetType, candidate_type>{cast_obj}};
+        } else {
+            return conversion_helper<TrailingCandidates...>::template convert<
+                TargetType>(obj);
```
fritzgoebel, line 145: This is so cool!

core/matrix/diagonal.cpp
fritzgoebel: I agree, diagonal to dense might not be something one would want to do.

core/solver/bicg.cpp
fritzgoebel: Do you mean when applying a real solver to a complex vector? This could be an interesting case, although I'm not sure if this is something people actually do

core/matrix/diagonal.cpp
upsj: I actually misremembered my implementation, I never try to cast to ConvertibleTo in make_temporary_conversion, only to Dense directly, so this is not an issue.

tcojean: What I mean here is that when someone passes a `Csr` type or any other form of LinOp instead of a `Dense`, before we would got an exception thrown here `GKO_NOT_IMPLEMENTED`. Now what would happen, in what way would it fail, it's actually not clear looking at the code of `make_temporary_conversion`.

Exceptions are part of the interface AFAIK?

core/solver/bicgstab.cpp
```diff
@@ -87,136 +88,148 @@ std::unique_ptr<LinOp> Bicgstab<ValueType>::conj_transpose() const
...
+            auto z = Vector::create_with_config_of(dense_b);
+            auto y = Vector::create_with_config_of(dense_b);
+            auto v = Vector::create_with_config_of(dense_b);
+            auto s = Vector::create_with_config_of(dense_b);
+            auto t = Vector::create_with_config_of(dense_b);
+            auto p = Vector::create_with_config_of(dense_b);
+            auto rr = Vector::create_with_config_of(dense_b);
+
+            auto alpha =
+                Vector::create(exec, dim<2>{1, dense_b->get_size()[1]});
```
tcojean, line 114: Same issue here, but at least alpha is used.

To further state the potential issue, both beta and rho at least are used in `compute_dot` which could potentially be overriden by users as well, so should we consider these variables as "interface level" variables and thus they should have a polymorphic behavior?

tcojean (commented):
LGTM in terms of implementation, but seeing test/examples would be required to move the PR forward. I also have unrelated comments which arose while looking at the solvers' code.

core/base/precision_dispatch.hpp
```diff
@@ -0,0 +1,151 @@
...
+
+
+template <typename ValueType>
+detail::temporary_conversion<matrix::Dense<ValueType>>
+make_temporary_conversion(LinOp *matrix, bool initialize)
+{
+    return detail::temporary_conversion<matrix::Dense<ValueType>>::
+        template create<matrix::Dense<next_precision<ValueType>>>(matrix,
+                                                                  initialize);
+}
```
thoasm, line 53: I don't get the use of the `initialize` variable at all...
In `precision_dispatch`, you only forward one variable per call (the `linops`), and even in the `temporary_conversion`, you only take one parameter.

core/matrix/diagonal.cpp
upsj: Before my last commit, it would have thrown NotSupported in make_temporary_conversion. Now with the new changes in place, it would throw NotSupported in conversion_helper:::convert, but now that I think of it, the previous version might actually be better, together with a `make_temporary_conversion/precision_dispatch_nothrow` that just returns nullptr or something like that in case of an error.

core/base/temporary_conversion.hpp
```diff
@@ -0,0 +1,227 @@
...
+
+    template <typename TargetType, typename MaybeConstLinOp,
+              typename FirstCandidate, typename... TrailingCandidates>
+    static std::unique_ptr<TargetType, std::function<void(TargetType *)>>
+    convert_impl(MaybeConstLinOp *obj)
+    {
+        using candidate_type =
+            typename std::conditional<std::is_const<MaybeConstLinOp>::value,
+                                      const FirstCandidate,
+                                      FirstCandidate>::type;
```
thoasm, line 125: You need to include `type_traits` for this to work. Additionally, since we use C++14, I think this would be better:
```suggestion
            std::conditional_t<std::is_const<MaybeConstLinOp>::value,
                                      const FirstCandidate,
                                      FirstCandidate>;
```

core/base/temporary_conversion.hpp
```diff
@@ -0,0 +1,227 @@
...
+ *
+ * @tparam T  the type of object held in the temporary_conversion
+ */
+template <typename T>
+class temporary_conversion {
+public:
+    using value_type = T;
+    using pointer = T *;
+    using lin_op_type = typename std::conditional<std::is_const<T>::value,
+                                                  const LinOp, LinOp>::type;
```
thoasm, line 179: nit:
```suggestion
    using lin_op_type = std::conditional_t<std::is_const<T>::value,
                                                  const LinOp, LinOp>;
```

core/base/temporary_conversion.hpp
```diff
@@ -0,0 +1,227 @@
...
+
+    template <typename... Formats>
+    static temporary_conversion create(lin_op_type *ptr)
```
thoasm, line 182: Please add a small sentence what the `Formats...` types represent.
Also, I think renaming it to `FormatCandidates`, `ConversionCandidates` or similar, like you do in `conversion_helper` would be a lot better, otherwise, one might assume that you want to convert **TO** the Format (since most of the time, you only specify one type, which is `next_precision<[...]>`).

core/base/temporary_conversion.hpp
```diff
@@ -0,0 +1,227 @@
...
+     * Returns if the conversion was successful.
+     */
+    explicit operator bool() { return bool(handle_); }
+
+private:
+    // std::function deleter allows to decide the (type of) deleter at
+    // runtime
+    using handle_type = std::unique_ptr<T, std::function<void(T *)>>;
+
+    temporary_conversion(handle_type handle) : handle_{std::move(handle)} {}
```
thoasm, line 217: Requires inclusion of `memory` (both `std::move` and all smart pointer in here).

core/base/temporary_conversion.hpp
```diff
@@ -0,0 +1,227 @@
...
+    using value_type = T;
+    using pointer = T *;
+    using lin_op_type = typename std::conditional<std::is_const<T>::value,
+                                                  const LinOp, LinOp>::type;
+
+    template <typename... Formats>
+    static temporary_conversion create(lin_op_type *ptr)
+    {
+        T *cast_ptr{};
+        if ((cast_ptr = dynamic_cast<T *>(ptr))) {
```
thoasm, line 185: Is there a reason you use double parenthesis for this check? Is it to silence a bug-detection from sonarcloud?

core/solver/cb_gmres.cpp
```diff
@@ -188,320 +189,343 @@ struct helper<std::complex<T>> {
...
-                            // If convergence was detected earlier, or
-                            // at the very beginning:
-                            if (stop_encountered_rhs.get_data()[i] ||
-                                total_iter < start_force_reset) {
-                                fully_converged_rhs.get_data()[i] = true;
+    precision_dispatch_real_complex<ValueType>(
+        [&](auto dense_b, auto dense_x) {
+            // Current workaround to get a lambda with a template argument (only
+            // the type of `value` matters, the content does not)
+            auto apply_templated = [&](auto value) {
```
thoasm, line 196: I should look into an alternative, we have quite a lot of blocks opened here, which makes the readability quite hard (especially following the indentations). But for this PR, it is totally fine.

core/base/precision_dispatch.hpp
```diff
@@ -0,0 +1,151 @@
...
+
+
+template <typename ValueType>
+detail::temporary_conversion<const matrix::Dense<ValueType>>
+make_temporary_conversion(const LinOp *matrix, bool initialize)
+{
+    return detail::temporary_conversion<const matrix::Dense<ValueType>>::
+        template create<matrix::Dense<next_precision<ValueType>>>(matrix,
+                                                                  initialize);
```
thoasm, line 62: Please add a comment that currently only two precisions are supported: `ValueType` and `next_precision<ValueType>`, which is why we only need to specify `Dense<next_precision<ValueType>>` as a format candidate.
I needed quite some time to figure that out, assuming I am right. If I am wrong, it is even more reason to add a comment here.
This is related to this comment from @tcojean : https://github.com/ginkgo-project/ginkgo/pull/677/files#r589413029

core/base/temporary_conversion.hpp
```diff
@@ -0,0 +1,227 @@
...
+     * Calls a method on the underlying object.
+     *
+     * @return the underlying object
+     */
+    T *operator->() const { return handle_.get(); }
+
+    /**
+     * Returns if the conversion was successful.
+     */
+    explicit operator bool() { return bool(handle_); }
```
thoasm, line 210: nit: using C++ cast:
```suggestion
    explicit operator bool() { return static_cast<bool>(handle_); }
```

core/base/precision_dispatch.hpp
```diff
@@ -0,0 +1,151 @@
...
+}
+
+
+template <typename ValueType>
+detail::temporary_conversion<const matrix::Dense<ValueType>>
+make_temporary_conversion(const LinOp *matrix, bool initialize)
+{
+    return detail::temporary_conversion<const matrix::Dense<ValueType>>::
+        template create<matrix::Dense<next_precision<ValueType>>>(matrix,
```
thoasm, line 61: `next_precision<T>` requires the header `ginkgo/core/base/math.hpp`

thoasm (commented):
The conversion code is well done, including the helper functions.
However, I would like to have a bit more documentation since it is quite hard to follow the execution flow (the same problem with most of the Template Metaprogramming in C++).

core/base/precision_dispatch.hpp
upsj: Yes, that was an attempt to fix my second TODO in the PR, but I already reverted it, since there is no way to make this clean.

core/base/temporary_conversion.hpp
upsj: yes, this is afaik the default way to silence "assignment in conditional statement" warnings.

core/solver/cb_gmres.cpp
upsj: I had some similar thoughts, the only reason I need to use the Lambda here is that I wanted to support real-to-complex apply, which involves a temporary object I need to copy back to, and convert_back_deleter only stores a pointer. Maybe I should add a copy_back_temporary_deleter?

tcojean:
I definitely agree with Thomas on adding documentations, there is still at least one comment on that topic leftover from my previous pass.

core/solver/cb_gmres.cpp
upsj: I meant the surrounding Lambda, your code is fine :) copy_back_temporary_deleter would store a unique_ptr to the temporary instead of a raw pointer to the original object.

thoasm: I don't see how this would impact this code, you would still need the lambda function in order to have the correct type unless you plug it directly into `linop.hpp`.

upsj: you can see make_temporary_conversion like a more complex as<Dense<...>> call, the static type of dense_x and dense_b is always Dense<ValueType>

thoasm: Ah, you mean instead of having the surrounding lambda, you would change the `as<Dense<ValueType>>` calls with a `make_temporary_conversion<Dense<ValueType>>` call?

upsj: Yes, that's what I had in place previously, but that didn't support real-to-complex apply yet, but hopefully will soon.

thoasm: I think the current solution is fine, and since time is slightly pressing, I would prefer to merge the current (working) solution and update it if we feel the need / have a working, better alternative.

yhmtsai (approved):
LGTM.
But for the solver apply, how about to add another `apply_impl(const Dense *, Dense*)` and the lambda in `apply_impl(const LinOp*, LinOp*)` only does the dispatch and calls a few functions. If there are only a few functions call in lambda, it is also fine to me.
the long lambda and directly using the lambda in function are complex to me

upsj:
@tcojean I am a bit unsure about the MPIR example, since modifying it to use our mixed precision support would not be 100% equivalent: MPIR uses the matrix stored in float for the inner solves, but in double for the outer solves. So either we use the matrix in float, which gives us only float-level precision in the overall solve, or we store it in double, which uses "too precise" SpMVs in the inner solver.

tcojean:
For the example if it's not exactly the same effect as the old one, I understand it makes sense to keep the old one. The next question then is whether we need a new example to advertise these new features/help people play with it in a simple fashion. Otherwise, LGTM.

pratikvn:
In terms of performance, can we make sure that the base case of no mixed precision still has the same performance as before (in current develop) ? All the applies are now wrapped by the `precision_dispatch` lambdas, which is a major change. Can you maybe quickly run some benchmarks for some small matrices for all the applies ? I know that it shouldn't affect the performance, but I think it is better to be sure.

upsj:
@pratikvn Good point, I ran a small benchmark (ani4.mtx with CG on reference) to get some performance numbers:
Before `0.03631952699999999`
After `0.03605373300000001`

Looking at the overall code, this is not too surprising to me, since the only substantial changes happen in the code path that would previously fail (wrong Dense value type), everything else is almost equivalent to what we did previously with gko::as<Dense>

reference/test/matrix/coo_kernels.cpp
```diff
@@ -644,13 +695,38 @@ TYPED_TEST(Coo, AppliesToComplex)
...
+
 TYPED_TEST(Coo, AdvancedAppliesToComplex)
 {
     using value_type = typename TestFixture::value_type;
     using complex_type = gko::to_complex<value_type>;
-    using Mtx = typename TestFixture::Mtx;
-    using Scal = typename gko::matrix::Dense<value_type>;
-    using Vec = typename gko::matrix::Dense<complex_type>;
+    using Scal = gko::matrix::Dense<value_type>;
+    using Vec = gko::matrix::Dense<complex_type>;
```
tcojean, line 729: I find the naming Scal and Vec to be a bit confusing. They are indeed applied as either scalar or vector but what distinguishes them is whether they are value_type or complex_type, and technically they could be used the other way around just as well, no? Maybe Dense and DenseCpx or something like this would be simpler?

That pops up in all tests at similar places.

reference/test/matrix/coo_kernels.cpp
```diff
@@ -675,13 +751,98 @@ TYPED_TEST(Coo, AdvancedAppliesToComplex)
 }
 
 
+TYPED_TEST(Coo, AdvancedAppliesToMixedComplex)
+{
+    using mixed_value_type =
+        gko::next_precision<typename TestFixture::value_type>;
+    using mixed_complex_type = gko::to_complex<mixed_value_type>;
+    using MixedScal = gko::matrix::Dense<mixed_value_type>;
+    using MixedVec = gko::matrix::Dense<mixed_complex_type>;
```
tcojean, line 760: Same here.

tcojean (approved):
LGTM in general. A minor comment on variable naming.

One more important issue, I think I would really prefer if we used @yhmtsai 's idea of `apply(Dense*, Dense*)` such that the all form of dispatch code is isolated from the lengthy implementation. That would I think make the code much clearer and also hopefully help reduce the amount of code changes as well.

core/solver/cb_gmres.cpp
```diff
@@ -493,15 +505,16 @@ void CbGmres<ValueType>::apply_impl(const LinOp *b, LinOp *x) const
...
-    auto dense_x = as<matrix::Dense<ValueType>>(x);
-
-    auto x_clone = dense_x->clone();
-    this->apply(b, x_clone.get());
-    dense_x->scale(residual_norm_collection);
-    dense_x->add_scaled(alpha, x_clone.get());
+    precision_dispatch_real_complex<ValueType>(
+        [this](auto dense_alpha, auto dense_b, auto dense_beta, auto dense_x) {
+            auto x_clone = dense_x->clone();
+            this->apply(dense_b, x_clone.get());
```
tcojean, line 513: Some minor improvement? Now you could directly call `apply_dense_impl`. Same everywhere.
```suggestion
            this->apply_dense_impl(dense_b, x_clone.get());
```

tcojean (approved):
LGTM.

fritzgoebel (approved):
LGTM!

core/solver/cb_gmres.cpp
upsj: Good idea, thanks!

reference/test/matrix/sellp_kernels.cpp
```diff
@@ -96,6 +96,18 @@ TYPED_TEST(Sellp, AppliesToDenseVector)
 }
 
 
+TYPED_TEST(Sellp, AppliesToMixedDenseVector)
+{
+    using Vec = typename TestFixture::Vec;
```
Slaedr, line 101:
```suggestion
    using value_type = typename TestFixture::value_type;
    using Vec = gko::matrix::Dense<gko::next_precision<value_type>>;
```

reference/test/matrix/sellp_kernels.cpp
```diff
@@ -132,6 +144,20 @@ TYPED_TEST(Sellp, AppliesLinearCombinationToDenseVector)
 }
 
 
+TYPED_TEST(Sellp, AppliesLinearCombinationToMixedDenseVector)
+{
+    using Vec = typename TestFixture::Vec;
```
Slaedr, line 149:
```suggestion
    using value_type = typename TestFixture::value_type;
    using Vec = gko::matrix::Dense<gko::next_precision<value_type>>;
```

reference/test/matrix/sparsity_csr_kernels.cpp
```diff
@@ -174,6 +174,20 @@ TYPED_TEST(SparsityCsr, AppliesToDenseVector)
 }
 
 
+TYPED_TEST(SparsityCsr, AppliesToMixedDenseVector)
+{
+    using Vec = typename TestFixture::Vec;
+    using T = typename TestFixture::value_type;
```
Slaedr, line 180:
```suggestion
    using value_type = typename TestFixture::value_type;
    using T = gko::next_precision<value_type>;
    using Vec = gko::matrix::Dense<T>;
```
assuming you want `T` to be the new type.

reference/test/matrix/sparsity_csr_kernels.cpp
```diff
@@ -207,6 +221,22 @@ TYPED_TEST(SparsityCsr, AppliesLinearCombinationToDenseVector)
 }
 
 
+TYPED_TEST(SparsityCsr, AppliesLinearCombinationToMixedDenseVector)
+{
+    using Vec = typename TestFixture::Vec;
+    using T = typename TestFixture::value_type;
```
Slaedr, line 227:
```suggestion
    using value_type = typename TestFixture::value_type;
    using T = gko::next_precision<value_type>;
    using Vec = gko::matrix::Dense<T>;
```
assuming you want `T` to be the new type.

Slaedr (changes requested):
Great work! That is quite elegant. After seeing the code for the solver apply etc. I realize how good this is.

I think I found some bugs in a couple of tests. I also have a few minor clarifications.

I wonder if it makes sense to write a test to ensure that const arguments are not copied back, for performance reasons. Maybe what you could do is, in one test, write a dummy operator templated on value_type with an `apply(const LinOp *x, LinOp *y)`. Inside the apply, `const_cast` the `x` and change it. Outside, ensure that x was not modified. It might be nice to have such a test in case the conversion_helper etc. need to be modified in future.

Also, it would be nice to have the new tests that you added to one of the objects, maybe CSR matrix kernels, to the other backends as well. Just the CSR tests would be enough, I think, and would add almost nothing to the testing time and we could be sure all this works on the other backends too.

include/ginkgo/core/base/temporary_conversion.hpp
```diff
@@ -0,0 +1,253 @@
...
+                               const FirstCandidate, FirstCandidate>;
+        candidate_type *cast_obj{};
+        if ((cast_obj = dynamic_cast<candidate_type *>(obj))) {
+            // if the cast is successful, obj is of dynamic type candidate_type
+            // so we can convert from this type to TargetType
+            auto converted = TargetType::create(obj->get_executor());
+            cast_obj->convert_to(converted.get());
+            // Make sure ConvertibleTo<TargetType> is available and symmetric
+            static_assert(
+                std::is_base_of<ConvertibleTo<std::remove_cv_t<TargetType>>,
```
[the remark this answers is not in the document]
upsj: [polymorphic_object.hpp](https://github.com/ginkgo-project/ginkgo/blob/0c6b6dcfbbd78724574eba1d260c425eac4fd0ac/core/base/polymorphic_object.hpp#L364)

include/ginkgo/core/base/precision_dispatch.hpp
```diff
@@ -0,0 +1,232 @@
...
+#include <ginkgo/core/matrix/dense.hpp>
+
+
+namespace gko {
+
+
+/**
+ * Convert the given LinOp from matrix::Dense<...> to matrix::Dense<ValueType>.
+ * The conversion tries to convert the input LinOp to all Dense types with value
+ * type recursively reachable by next_precision<...> starting from the ValueType
```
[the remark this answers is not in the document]
upsj: Yes, the reason for this is that due to the general way of specifying Dense matrices, we can only implement symmetric conversions. This is also the reason for next_precision<...> existing at all.

include/ginkgo/core/base/temporary_conversion.hpp
```diff
@@ -0,0 +1,253 @@
...
+    void operator()(pointer ptr) const { delete ptr; }
+};
+
+
+/**
+ * @internal
+ *
+ * Helper type that attempts to statically find the dynamic type of a given
+ * LinOp from a list of ConversionCandidates and, on the first match, converts
+ * it to TargetType with an appropriate .
```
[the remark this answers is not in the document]
upsj: Good catch! That's supposed to say convert_back_deleter

include/ginkgo/core/base/temporary_conversion.hpp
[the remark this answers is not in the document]
Slaedr: Right, I completely forgot about this. Thanks.

upsj:
@Slaedr Thanks, all of those are really good suggestions, I incorporated all of them

cuda/test/matrix/dense_kernels.cpp
```diff
@@ -358,6 +380,19 @@ TEST_F(Dense, AdvancedApplyIsEquivalentToRef)
...
+TEST_F(Dense, AdvancedApplyMixedIsEquivalentToRef)
+{
+    set_up_apply_data();
+
+    x->apply(convert<MixedMtx>(alpha).get(), convert<MixedMtx>(y).get(),
+             convert<MixedMtx>(beta).get(), convert<MixedMtx>(expected).get());
+    dx->apply(convert<MixedMtx>(dalpha).get(), convert<MixedMtx>(dy).get(),
+              convert<MixedMtx>(dbeta).get(), convert<MixedMtx>(dresult).get());
+
+    GKO_ASSERT_MTX_NEAR(dresult, expected, 1e-14);
```
Slaedr, line 392: Should this be 1e-7, or can the previous test also be checked to 1e-14?
```suggestion
    GKO_ASSERT_MTX_NEAR(dresult, expected, 1e-7);
```

hip/test/matrix/dense_kernels.hip.cpp
```diff
@@ -353,6 +375,19 @@ TEST_F(Dense, AdvancedApplyIsEquivalentToRef)
...
+TEST_F(Dense, AdvancedApplyMixedIsEquivalentToRef)
+{
+    set_up_apply_data();
+
+    x->apply(convert<MixedMtx>(alpha).get(), convert<MixedMtx>(y).get(),
+             convert<MixedMtx>(beta).get(), convert<MixedMtx>(expected).get());
+    dx->apply(convert<MixedMtx>(dalpha).get(), convert<MixedMtx>(dy).get(),
+              convert<MixedMtx>(dbeta).get(), convert<MixedMtx>(dresult).get());
+
+    GKO_ASSERT_MTX_NEAR(dresult, expected, 1e-14);
```
Slaedr, line 387:
```suggestion
    GKO_ASSERT_MTX_NEAR(dresult, expected, 1e-7);
```

omp/test/matrix/dense_kernels.cpp
```diff
@@ -370,6 +394,19 @@ TEST_F(Dense, AdvancedApplyIsEquivalentToRef)
...
+TEST_F(Dense, AdvancedApplyMixedIsEquivalentToRef)
+{
+    set_up_apply_data();
+
+    x->apply(convert<MixedMtx>(alpha).get(), convert<MixedMtx>(y).get(),
+             convert<MixedMtx>(beta).get(), convert<MixedMtx>(expected).get());
+    dx->apply(convert<MixedMtx>(dalpha).get(), convert<MixedMtx>(dy).get(),
+              convert<MixedMtx>(dbeta).get(), convert<MixedMtx>(dresult).get());
+
+    GKO_ASSERT_MTX_NEAR(dresult, expected, 1e-14);
```
Slaedr, line 406:
```suggestion
    GKO_ASSERT_MTX_NEAR(dresult, expected, 1e-7);
```

Slaedr (approved):
LGTM! Just one set of small nits about the tolerance.

upsj:
@Slaedr Good catch! This was actually an accident, since our implementations seems to be bitwise equivalent, so we could even use zero tolerance. We're not comparing against a "ground truth", but two equally inexact implementations

PR #684, opened 2020-12-30
Add symmetric permutation kernels

upsj:
In many use cases for LinOp permutation, we want to symmetrically permute rows and columns.
This allows for a great amount of kernel fusion compared to permuting rows and then columns. This PR adds explicit kernels for symmetric forward and inverse permutation and cleans up the permutation kernels a bit.


include/ginkgo/core/base/lin_op.hpp
```diff
@@ -469,12 +469,44 @@ class Permutable {
 public:
     virtual ~Permutable() = default;
 
+    /**
+     * Returns a LinOp representing the symmetric row and column permutation of
+     * the Permutable object.
+     *
+     * @param permutation_indices  the array of indices contaning the
```
yhmtsai, line 476:
```suggestion
     * @param permutation_indices  the array of indices containing the
```

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -469,12 +469,44 @@ class Permutable {
...
+    {
+        return as<Permutable>(this->row_permute(permutation_indices))
+            ->column_permute(permutation_indices);
+    };
+
+    /**
+     * Returns a LinOp representing the symmetric inverse row and column
+     * permutation of the Permutable object.
+     *
+     * @param permutation_indices  the array of indices contaning the
```
yhmtsai, line 492:
```suggestion
     * @param permutation_indices  the array of indices containing the
```

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -469,12 +469,44 @@ class Permutable {
...
+    {
+        return as<Permutable>(this->inverse_row_permute(permutation_indices))
+            ->inverse_column_permute(permutation_indices);
+    };
+
     /**
      * Returns a LinOp representing the row permutation of the Permutable
      * object.
      *
      * @param permutation_indices  the array of indices contaning the
```
yhmtsai, line 508:
```suggestion
     * @param permutation_indices  the array of indices containing the
```
also applying to other contaning

core/matrix/csr.cpp
```diff
@@ -402,6 +403,48 @@ std::unique_ptr<LinOp> Csr<ValueType, IndexType>::conj_transpose() const
...
+        Csr::create(exec, this->get_size(), this->get_num_stored_elements(),
+                    this->get_strategy());
+    Array<IndexType> inv_permutation(exec, this->get_size()[1]);
+
+    exec->run(csr::make_invert_permutation(
+        this->get_size()[1],
+        make_temporary_clone(exec, permutation_indices)->get_const_data(),
+        inv_permutation.get_data()));
+    exec->run(csr::make_inv_symm_permute(inv_permutation.get_const_data(), this,
+                                         permute_cpy.get()));
```
yhmtsai, line 423: In practice, do we use inv_sym_permute more than sym_permute?

yhmtsai (approved):
LGTM.
Is implementing symm_permute reasonable?

core/matrix/csr.cpp
hartwiganzt: inv_symm_permute or inv_sym_permute?
I think "symm" is good as it avoids confusion with "symbolic"

hartwiganzt (approved):

core/matrix/csr.cpp
upsj: The reason we have a inv_symm_permute kernel, but not a symm_permute_kernel is that for the column permutation part, we need the inverse permutation to to
```
new_column[i] = inv_perm[old_column[i]]
```
due to the sparse nature of the operation, it doesn't work the other way round.

core/matrix/dense.cpp
```diff
@@ -814,15 +881,14 @@ std::unique_ptr<LinOp> Dense<ValueType>::inverse_column_permute(
 
 template <typename ValueType>
 std::unique_ptr<LinOp> Dense<ValueType>::inverse_column_permute(
-    const Array<int64> *inverse_permutation_indices) const
+    const Array<int64> *permutation_indices) const
```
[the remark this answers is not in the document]
upsj: The permutation_indices will be inverted internally, but the parameter name right now implies that it would take the inverse permutation as input, which is misleading.

PR #685, opened 2020-12-31
Fix GinkgoConfig.cmake. Test linking build dir.

tcojean:
As previously discussed, this fixes the problem with `GinkgoConfig.cmake` from a build directory perspective, while keeping a correct installation setup.

To ensure this works, I also test linking Ginkgo after exporting the build directory so that CMake is able to use this when doing a `find_package()`. We already had an option for this, it's only a matter of actually using it.

Issue found by: https://github.com/ginkgo-project/ginkgo/pull/680

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -26,6 +26,7 @@ stages:
   CONFIG_LOG: "ON"
   CXX_FLAGS: ""
   EXTRA_CMAKE_FLAGS: ""
+  EXPORT_BUILD_DIR: "OFF"
   CI_PROJECT_DIR_SUFFIX: ""
 
 .before_script_template: &default_before_script
@@ -65,8 +66,10 @@ stages:
         -DGINKGO_BUILD_TESTS=ON -DGINKGO_BUILD_EXAMPLES=ON
         -DGINKGO_RUN_EXAMPLES=${RUN_EXAMPLES}
         -DGINKGO_CONFIG_LOG_DETAILED=${CONFIG_LOG}
+        -DGINKGO_EXPORT_BUILD_DIR=${EXPORT_BUILD_DIR}
     - ninja -j${NUM_CORES} -l${CI_LOAD_LIMIT}
     - if [ ! -z ${SYCL_DEVICE_TYPE+x} ]; then unset SYCL_DEVICE_TYPE; fi
+    - if [ "${EXPORT_BUILD_DIR}" == "ON" ]; then ninja test_exportbuild; fi
   dependencies: []
   except:
       - schedules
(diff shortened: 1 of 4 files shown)


test_install/CMakeLists.txt
```diff
@@ -37,10 +37,9 @@ if(GINKGO_BUILD_CUDA)
     target_link_libraries(test_install_cuda PRIVATE Ginkgo::ginkgo)
 endif()
 
-if(GINKGO_BUILD_HIP
-   AND GINKGO_HIP_PLATFORM MATCHES "hcc"
-   AND GINKGO_HIP_VERSION VERSION_GREATER_EQUAL "3.5"
-   AND NOT GINKGO_BUILD_SHARED_LIBS)
+if(GINKGO_BUILD_HIP AND GINKGO_HIP_PLATFORM MATCHES "hcc"
+        AND GINKGO_HIP_VERSION VERSION_GREATER_EQUAL "3.5"
+        AND NOT GINKGO_BUILD_SHARED_LIBS)
```
yhmtsai, line 42: different format?

yhmtsai (approved):
LGTM!

test_install/CMakeLists.txt
tcojean: The format is supposed to be 4 spaces per block. Somehow this counts as two blocks on my setup (one the if, another the parenthesis?).

.gitlab-ci.yml
```diff
@@ -742,6 +747,28 @@ subdir-build:
...
+# Ensure Ginkgo can be used when exporting the build directory
+export-build:
+  <<: *default_build
+  stage: code_quality
+  image: localhost:5000/gko-cuda102-gnu8-llvm8-intel2019
+  variables:
+    <<: *default_variables
+    BUILD_OMP: "ON"
+    BUILD_CUDA: "ON"
+    BUILD_HIP: "ON"
```
pratikvn, line 759: Do we need both the HIP and CUDA modules to be compiled ? Can we not get away without them ?

pratikvn (approved):
LGTM!

.gitlab-ci.yml
tcojean: That's a good point, I think we can remove it from here and also from the subdir build.

upsj (approved):
LGTM!

Slaedr (commented):
Looks good. But will this (setting `-DGINKGO_EXPORT_BUILD_DIR=ON`) make the Ginkgo build directory searchable by `find_package` called anywhere by any project on the computer? Is that what we want? (Keeping in mind that `CMAKE_INSTALL_DIR` can also be set to a subdirectory of `$HOME` and `Ginkgo_ROOT` can exported by the user to make `find_package(Ginkgo)` find it.)

tcojean:
Yes, this does make the directory available for the user by any `find_package(Ginkgo)`. This is how this feature works by default. We set this to off by default so I don't see a problem with it. If the user wants to use this, that's his freedom to do so and we now know that the feature works properly thanks to this PR.

PR #687, opened 2021-01-04
Bump copyright year to 2021

yhmtsai:
This PR bumps the copyright year to 2021.
Edit the `LICENSE` and `make format`

@upsj do I also need to change the year of gdb-ginkgo.py? or does it need to wait the pretty-printers

Happy new year!

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2017-2020, the Ginkgo authors
+Copyright (c) 2017-2021, the Ginkgo authors
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
--- a/benchmark/conversions/conversions.cpp
+++ b/benchmark/conversions/conversions.cpp
@@ -1,5 +1,5 @@
 /*******************************<GINKGO LICENSE>******************************
-Copyright (c) 2017-2020, the Ginkgo authors
+Copyright (c) 2017-2021, the Ginkgo authors
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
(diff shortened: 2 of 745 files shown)


tcojean (approved):
LGTM.

Somehow, the bot wasn't able to detect that DPC++, OMP and Reference files changed. I think it's an issue with how the regex is written?

upsj (approved):

tcojean:
Too bad, using the `reviewers` github group doesn't work as we would expect: when one person approves, it's like the whole group approved. Whereas what we want is "get any two persons from the group to review this".

upsj:
Relating to the gdb pretty printers: I don't think we necessarily need to update the license date there unless we modify the file itself. I think this is only important for files which get formatted automatically.

tcojean:
label!

yhmtsai:
Is the bot from develop not this branch?

upsj:
The issue is that the bot only takes the first page of changed files for simplicity, since these huge diffs are rare, I think we can ignore this issue.

tcojean:
Yes I realized later that would not update it here. And I see, I thought that the issue would be that bash somehow interprets and gobbles the `^` in its own way, like it can happen to `!` and other characters sometimes.

hartwiganzt (approved):

pratikvn (approved):
LGTM!

yhmtsai:
The maximum of one page is 100 even if we specify it `&per_page=1000` in `label.sh`

yhmtsai:
label!

yhmtsai:
format!

yhmtsai:
Maybe I do the `format` too quickly such that github action only takes the original commit not the github-bot-format commit.
Windows: https://github.com/ginkgo-project/ginkgo/actions/runs/468671383
Mac: https://github.com/ginkgo-project/ginkgo/actions/runs/468671387
The format commit only changes the order of header, so I will merge it directly not push it again.

upsj:
Sounds sensible, also in #691 I will replace the headers by the correct C++-style headers, so this will not be an issue in the future.

PR #689, opened 2021-01-05
allow bot for all changed files and update label regex

yhmtsai:
This PR makes bot (label and format) get all changed files when the number of files is larger than 100, and update label regex.

if the regex use some special character like `\.`, `|`, or `$`, I will put the quote.
Also, add `.github` match into `reg:ci-cd` 

The log of this bot:
https://github.com/yhmtsai/ginkgo/pull/4/checks?check_run_id=1646511807
The result:
https://github.com/yhmtsai/ginkgo/pull/4
The file used in that repo:
https://github.com/yhmtsai/ginkgo/blob/a0d711041c159483025e66b0e1ee83186f0e653e/.github/label.sh#L5

### DIFF
--- a/.github/bot-pr-base.sh
+++ b/.github/bot-pr-base.sh
@@ -59,6 +59,27 @@ bot_error() {
   exit 1
 }
 
+bot_get_all_changed_files() {
+  local pr_url="$1"
+  local pr_files=""
+  local page="1"
+  while true; do
+    # this api allows 100 items per page
+    # github action uses `bash -e`. The last empty page will leads jq error, use `|| :` to ignore the error.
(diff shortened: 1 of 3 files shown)


upsj (approved):
LGTM! Thanks for fixing this, I was just too lazy to deal with pagination. A simplification suggestion: You could move the file list queries to a function in `bot_base.sh` that takes the PR_URL as input and writes the filenames either to a global variable or prints them to stdout.

tcojean (approved):
LGTM. I agree with Tobias, it looks like creating a helper function is the way to go here.

yhmtsai:
@upsj I move the helper function into bot-pr-base because it is related to pr.

upsj:
@yhmtsai Thanks! For some reason, I had thought that `label` doesn't use the `base-pr` script, but that is definitely the most appropriate place to put it.

PR #690, opened 2021-01-05
General isai

fritzgoebel:
This PR extends ISAI from triangular to general and spd matrices.
When generated for a general or spd matrix, it can be used as a preconditioner on its own, i.e. outside of a factorization.
In the spd case, an ISAI L is computed on the lower triangular half of A, a diagonal scaling step makes the result equivalent to an FSPAI preconditioner, so an approximation on the exact Cholesky factorization. In this case, the preconditioner is no longer a single CSR matrix but a composition of L^T and L.

For rows of size up to 32, the corresponding small dense systems are solved directly, for larger rows an excess system is built up as it is for triangular systems. The excess system is solved using GMRES with a block-Jacobi preconditioner.
A new factory parameter `excess_limit` sets a threshold to the size of the excess system. If set, the excess system gets solved as soon as this threshold is exceeded, the solution scattered to the result matrix and a new excess system is built up for the remaining rows, and so on.


fritzgoebel:
format!

core/preconditioner/isai_kernels.hpp
```diff
@@ -51,29 +51,41 @@ namespace kernels {
...
-
-#define GKO_DECLARE_ALL_AS_TEMPLATES                                      \
-    constexpr auto row_size_limit = 32;                                   \
-    template <typename ValueType, typename IndexType>                     \
-    GKO_DECLARE_ISAI_GENERATE_TRI_INVERSE_KERNEL(ValueType, IndexType);   \
-    template <typename ValueType, typename IndexType>                     \
-    GKO_DECLARE_ISAI_GENERATE_EXCESS_SYSTEM_KERNEL(ValueType, IndexType); \
-    template <typename ValueType, typename IndexType>                     \
+        matrix::Csr<ValueType, IndexType> *inverse, const size_type e_start,  \
+        const size_type e_end)
```
upsj, line 78: nit: `const` is an implementation detail and not part of the interface
```suggestion
        matrix::Csr<ValueType, IndexType> *inverse, size_type e_start,        \
        size_type e_end)
```

core/preconditioner/isai_kernels.hpp
```diff
@@ -51,29 +51,41 @@ namespace kernels {
...
 #define GKO_DECLARE_ISAI_GENERATE_EXCESS_SYSTEM_KERNEL(ValueType, IndexType) \
     void generate_excess_system(                                             \
         std::shared_ptr<const DefaultExecutor> exec,                         \
         const matrix::Csr<ValueType, IndexType> *input,                      \
         const matrix::Csr<ValueType, IndexType> *inverse,                    \
         const IndexType *excess_rhs_ptrs, const IndexType *excess_nz_ptrs,   \
         matrix::Csr<ValueType, IndexType> *excess_system,                    \
-        matrix::Dense<ValueType> *excess_rhs)
+        matrix::Dense<ValueType> *excess_rhs, const size_type excess_start,  \
+        const size_type excess_end)
```
upsj, line 70:
```suggestion
        matrix::Dense<ValueType> *excess_rhs, size_type excess_start,        \
        size_type excess_end)
```

include/ginkgo/core/preconditioner/isai.hpp
```diff
@@ -57,10 +57,10 @@ namespace preconditioner {
 /**
  * This enum lists the types of the ISAI preconditioner.
  *
- * ISAI can either generate a lower triangular matrix, or an upper triangular
- * matrix.
+ * ISAI can either generate a general matrix, a lower triangular matrix, or an
```
upsj, line 60: maybe
```suggestion
 * ISAI can either be generated for a general matrix, a lower triangular matrix, or an
```

include/ginkgo/core/preconditioner/isai.hpp
```diff
@@ -98,15 +98,18 @@ class Isai : public EnableLinOp<Isai<IsaiType, ValueType, IndexType>>,
...
+    friend class Isai<isai_type::general, ValueType, IndexType>;
+    friend class Isai<isai_type::lower, ValueType, IndexType>;
+    friend class Isai<isai_type::upper, ValueType, IndexType>;
 
 public:
     using value_type = ValueType;
     using index_type = IndexType;
     using transposed_type =
-        Isai<IsaiType == isai_type::lower ? isai_type::upper : isai_type::lower,
+        Isai<IsaiType == isai_type::general
```
upsj, line 109: We should note somewhere that ISAI(transpose) != transpose(ISAI)

common/components/warp_blas.hpp.inc
```diff
@@ -83,6 +83,52 @@ __device__ __forceinline__ void apply_gauss_jordan_transform(
...
+template <
+    int max_problem_size, typename Group, typename ValueType,
+    typename = std::enable_if_t<group::is_communicator_group<Group>::value>>
+__device__ __forceinline__ void apply_gauss_jordan_transform_with_rhs(
+    const Group &__restrict__ group, int32 key_row, int32 key_col,
+    ValueType *__restrict__ row, ValueType *__restrict__ rhs,
+    bool &__restrict__ status)
+{
+    auto key_col_elem = group.shfl(row[key_col], key_row);
+    auto key_rhs_elem = group.shfl(rhs[0], key_row);
```
upsj, line 105: Suggestion: Since you only access `rhs[0]`, you could also pass `rhs` by reference instead of by pointer, which saves the `[0]` in this function.

common/preconditioner/isai_kernels.hpp.inc
```diff
@@ -125,22 +129,26 @@ __forceinline__ __device__ void generic_generate(
                 [&](IndexType, IndexType m_idx, IndexType i_idx,
                     config::lane_mask_type, bool valid) {
                     if (valid) {
-                        trisystem(nz, i_idx) = m_values[m_row_begin + m_idx];
+                        if (m_col_idxs[m_row_begin + m_idx] < row && col == row)
+                            atomic_add(&rhs_one_idxs[row % subwarps_per_block],
+                                       size_type{1});
```
upsj, line 134: I would suggest you [(sub)warp-aggregate](https://developer.nvidia.com/blog/cuda-pro-tip-optimized-filtering-warp-aggregated-atomics/) the results, then you should be able to avoid using atomics and shared memory:
```cpp
count += popcnt(subwarp.ballot());
```
Alternative: Store partial counts and reduce afterwards
```cpp
local_count += (m_col_idxs[m_row_begin + m_idx] < row && col == row);
// ...
count = reduce(subwarp, local_count)
```

core/preconditioner/isai.cpp
```diff
@@ -108,17 +114,22 @@ std::shared_ptr<Csr> extend_sparsity(std::shared_ptr<const Executor> &exec,
 
 template <isai_type IsaiType, typename ValueType, typename IndexType>
 void Isai<IsaiType, ValueType, IndexType>::generate_inverse(
-    std::shared_ptr<const LinOp> input, bool skip_sorting, int power)
+    std::shared_ptr<const LinOp> input, bool skip_sorting, int power,
+    int excess_limit)
```
upsj, line 118:
```suggestion
    IndexType excess_limit)
```

core/preconditioner/isai.cpp
```diff
@@ -127,38 +138,86 @@ void Isai<IsaiType, ValueType, IndexType>::generate_inverse(
...
+                excess_block_ptrs.get_const_data() + block);
+            const auto nnz_offset = exec->copy_val_to_host(
+                excess_row_ptrs_full.get_const_data() + block);
+            while (excess_dim < excess_lim && block < num_rows) {
+                block++;
+                excess_dim = exec->copy_val_to_host(
+                                 excess_block_ptrs.get_const_data() + block) -
+                             block_offset;
+            }
+            if (excess_dim == 0) break;
```
upsj, line 171: We do a lot of individual copies here, which might not be too efficient. I would suggest that we instead create a host-side copy of excess_block_ptrs and use std::upper_bound to find the end of the current batch of blocks. Then we could also write this in a nicer loop:
```cpp
for (size_type excess_start = 0; excess_start < total_excess_dim; excess_start = find_next_block_index_lambda(excess_start))
```

core/preconditioner/isai.cpp
```diff
@@ -127,38 +138,86 @@ void Isai<IsaiType, ValueType, IndexType>::generate_inverse(
...
+            auto system_copy = Csr::create(exec->get_master());
+            system_copy->copy_from(excess_system.get());
+            auto rhs_copy = Dense::create(exec->get_master());
+            rhs_copy->copy_from(excess_rhs.get());
+            std::unique_ptr<LinOpFactory> trs_factory;
+            if (is_general) {
+                trs_factory =
+                    Gmres::build()
+                        .with_preconditioner(
+                            bj::build().with_max_block_size(32u).on(exec))
```
upsj, line 197: Two comments here:
1. we could make the solver factory a factory parameter of ISAI that defaults to this choice, then people can also use GMRES with BJ for triangular systems if they like, similar to what the ILU preconditioner does.
2. Since we already have block structure information, why not pass this on to the preconditioner as block pointers clamped at 32? That might significantly improve the preconditioner quality.

omp/preconditioner/isai_kernels.cpp
```diff
@@ -146,12 +151,14 @@ void generic_generate(std::shared_ptr<const DefaultExecutor> exec,
                     forall_matching(
                         m_cols + m_begin, m_size, i_cols + i_begin, i_size,
                         [&](IndexType, IndexType m_idx, IndexType i_idx) {
-                            trisystem(i, i_idx) = m_vals[m_idx + m_begin];
+                            if (m_cols[m_idx + m_begin] < row && col == row)
+                                rhs_one_idx++;
```
upsj, line 155: nit: formatting
```suggestion
                            if (m_cols[m_idx + m_begin] < row && col == row) {
                                rhs_one_idx++;
                            }
```

reference/preconditioner/isai_kernels.cpp
```diff
@@ -138,12 +142,14 @@ void generic_generate(std::shared_ptr<const DefaultExecutor> exec,
                 forall_matching(
                     m_cols + m_begin, m_size, i_cols + i_begin, i_size,
                     [&](IndexType, IndexType m_idx, IndexType i_idx) {
-                        trisystem(i, i_idx) = m_vals[m_idx + m_begin];
+                        if (m_cols[m_idx + m_begin] < row && col == row)
+                            rhs_one_idx++;
```
upsj, line 146: nit: formatting
```suggestion
                        if (m_cols[m_idx + m_begin] < row && col == row) {
                            rhs_one_idx++;
                        }
```

omp/preconditioner/isai_kernels.cpp
```diff
@@ -240,14 +247,111 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
...
+            // fill in transposed system
+            ValueType *transposed_system = new ValueType[size * size];
+            for (auto row = 0; row < size; row++) {
+                for (auto col = 0; col < size; col++) {
+                    transposed_system[row * size + col] = system(col, row);
+                }
+            }
+
+            // solve transposed system
+            std::vector<IndexType> perm(size, {});
```
upsj, line 304: nit: this is allocated outside of our control, you could use gko::vector instead. But it looks to me like we don't need the vector anyways.

The funny thing is, when I saw your reference implementation, I immediately thought: "This could maybe be improved with a permutation array", so nice 😆

omp/preconditioner/isai_kernels.cpp
```diff
@@ -240,14 +247,111 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
...
+            std::iota(begin(perm), end(perm), IndexType{0});
+            for (IndexType col = 0; col < size; col++) {
+                const auto row =
+                    choose_pivot(size - col,
+                                 transposed_system + col * (size + 1), size) +
+                    col;
+                swap_rows(col, row, size, transposed_system, size);
+                swap(perm[row], perm[col]);
```
upsj, line 312: You don't seem to be using perm anywhere?

upsj (approved):
LGTM - great job! I didn't look over the (especially reference) tests in too much detail, but the rest looked solid.

hip/preconditioner/isai_kernels.hip.cpp
```diff
@@ -168,6 +168,24 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
     GKO_DECLARE_ISAI_GENERATE_EXCESS_SYSTEM_KERNEL);
 
 
+template <typename ValueType, typename IndexType>
+void scale_excess_solution(std::shared_ptr<const DefaultExecutor>,
```
hartwiganzt, line 172: Do you anywhere document your strategy, like what you do with the long-row systems?

omp/preconditioner/isai_kernels.cpp
```diff
@@ -240,14 +248,115 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
...
+            if (size <= 0) {
+                return;
+            }
+            // RHS is the identity: zero everywhere except for the diagonal
+            // entry
+            std::fill_n(rhs, size, zero<ValueType>());
+            rhs[rhs_one_idx] = one<ValueType>();
+
+            // fill in transposed system
+            ValueType *transposed_system = new ValueType[size * size];
```
yhmtsai, line 297: Miss the deconstructor. Can it use the gko::Array?
gko::matrix can not be used here due to circular dependence. Does it also apply to gko::Array?

omp/preconditioner/isai_kernels.cpp
```diff
@@ -240,14 +248,115 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
...
+                rhs[col] /= d;
+                transposed_system[col * size + col] = one<ValueType>() / d;
+            }
+
+            if (spd) {
+                const auto scal = one<ValueType>() / sqrt(rhs[size - 1]);
+                for (IndexType row = 0; row < size; row++) {
+                    rhs[row] *= scal;
+                }
+            }
```
yhmtsai, line 340: omp seem to miss the spd cases test.

yhmtsai (changes requested):
Miss free memory in one lambda function

omp/preconditioner/isai_kernels.cpp
upsj: Good catch! No, you can use gko::Array, since it is a header-only type. I preallocated temporary storage for the triangular solve, you should be able to use the same approach here.

cuda/test/preconditioner/isai_kernels.cpp
```diff
@@ -102,6 +112,29 @@ class Isai : public ::testing::Test {
...
+        std::ifstream mtxstream{std::string{gko::matrices::location_isai_mtxs} +
+                                name};
+        auto result = gko::read<ReadMtx>(mtxstream, ref);
+        // to avoid removing 0s, the matrices store 12345 instead
+        for (gko::size_type i = 0; i < result->get_num_stored_elements(); ++i) {
+            auto &val = result->get_values()[i];
+            if (val == static_cast<value_type>(12345.0)) {
+                val = 0;
+            }
+        }
```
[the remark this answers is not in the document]
upsj: I'm to blame for this, I think I wanted to make the rows longer artificially by inserting numerical zeros into the sparsity pattern, but read() usually removes them.

reference/test/preconditioner/isai_kernels.cpp
```diff
@@ -614,14 +856,123 @@ TYPED_TEST(Isai, KernelGenerateExcessULongrow)
...
+    std::fill_n(a2_expect.get_data() + 36, 65, 338);
+
+    gko::kernels::reference::isai::generate_general_inverse(
+        this->exec, lend(this->spd_csr_longrow), lend(result), a1.get_data(),
+        a2.get_data(), true);
+
+    GKO_ASSERT_MTX_EQ_SPARSITY(result, this->spd_csr_longrow_inv_partial);
+    GKO_ASSERT_MTX_NEAR(result, this->spd_csr_longrow_inv_partial,
+                        r<value_type>::value);
+    // no row above the size limit -> zero array
```
fritzgoebel, line 942:
```suggestion
```

include/ginkgo/core/preconditioner/isai.hpp
```diff
@@ -57,22 +57,30 @@ namespace preconditioner {
...
+ * The sparsity pattern used for the approximate inverse of A, L and U is the
+ * same as the sparsity pattern of the respective matrix. For B, the sparsity
+ * pattern used for the approximate inverse is the same as the sparsity pattern
+ * of the lower triangular half of B.
+ *
+ * Note that, except for the spd case, for a matrix A generally
+ * ISAI(A)^T != ISAI(A^T).
  *
  * For more details on the algorithm, see the paper
  * <a href="https://doi.org/10.1016/j.parco.2017.10.003">
```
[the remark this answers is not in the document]
fritzgoebel: No, I think we can add the europar paper here later?

reference/preconditioner/isai_kernels.cpp
```diff
@@ -34,6 +34,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
 #include <algorithm>
+#include <iostream>
```
upsj, line 37:
```suggestion
```

reference/preconditioner/isai_kernels.cpp
```diff
@@ -228,13 +236,115 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
...
+            if (size <= 0) {
+                return;
+            }
+            // RHS is the identity: zero everywhere except for the diagonal
+            // entry
+            std::fill_n(rhs, size, zero<ValueType>());
+            rhs[rhs_one_idx] = one<ValueType>();
+
+            // fill in transposed system
+            ValueType *transposed_system = new ValueType[size * size];
```
upsj, line 285: Same as OpenMP: You should use pre-allocated storage here

reference/preconditioner/isai_kernels.cpp
```diff
@@ -228,13 +236,115 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
...
+            for (auto row = 0; row < size; row++) {
+                for (auto col = 0; col < size; col++) {
+                    transposed_system[row * size + col] = system(col, row);
```
upsj, line 288: How about using a range for transposed_system as well?

reference/test/preconditioner/isai_kernels.cpp
```diff
@@ -154,6 +219,16 @@ class Isai : public ::testing::Test {
...
+        a_csr_longrow = read<Csr>("isai_a.mtx");
+        a_csr_longrow_e = read<Csr>("isai_a_excess.mtx");
+        a_csr_longrow_e_rhs = read<Dense>("isai_a_excess_rhs.mtx");
+        a_csr_longrow_inv_partial = read<Csr>("isai_a_inv_partial.mtx");
+        a_csr_longrow_inv = read<Csr>("isai_a_inv.mtx");
+        spd_csr_longrow = read<Csr>("isai_spd.mtx");
+        spd_csr_longrow_e = read<Csr>("isai_spd_excess.mtx");
+        spd_csr_longrow_e_rhs = read<Dense>("isai_spd_excess_rhs.mtx");
+        spd_csr_longrow_inv_partial = read<Csr>("isai_spd_inv_partial.mtx");
+        spd_csr_longrow_inv = read<Csr>("isai_spd_inv.mtx");
```
upsj, line 231: We may want to think about ways to programatically generate these matrices?

include/ginkgo/core/preconditioner/isai.hpp
```diff
@@ -57,22 +57,30 @@ namespace preconditioner {
 /**
  * This enum lists the types of the ISAI preconditioner.
  *
- * ISAI can either generate a lower triangular matrix, or an upper triangular
- * matrix.
+ * ISAI can either be generated for a general square matrix, a lower triangular
+ * matrix, an upper triangular matrix or a spd matrix.
```
upsj, line 61: from the pronounciation, I would assume this is similar to `an essential`
```suggestion
 * matrix, an upper triangular matrix or an spd matrix.
```

include/ginkgo/core/preconditioner/isai.hpp
```diff
@@ -84,8 +92,9 @@ enum struct isai_type { lower, upper };
  *       matrix. If there are more than `width` elements per row, the remaining
  *       elements will be ignored.
  *
- * @tparam IsaiType  determines if the ISAI is generated for a lower triangular
- *                   matrix or an upper triangular matrix
+ * @tparam IsaiType  determines if the ISAI is generated for a general square
+ * matrix, a lower triangular matrix, an upper triangular matrix or an spd
```
upsj, line 96: nit: formatting

include/ginkgo/core/preconditioner/isai.hpp
```diff
@@ -143,6 +164,25 @@ class Isai : public EnableLinOp<Isai<IsaiType, ValueType, IndexType>>,
...
+         *
+         * For rows with more than 32 nonzero entries, the algorithm builds up
+         * an excess system which is solved with sparse triangular solves (for
+         * upper or lower ISAI) or GMRES (for general ISAI). If this parameter
+         * is set to some m > 0, the excess system is solved as soon
+         * as its size supersedes m. This is repeated until the complete excess
+         * solution has been computed.
+         * Must be at least 0, default value 0.
+         */
+        index_type GKO_FACTORY_PARAMETER_SCALAR(excess_limit, 0);
```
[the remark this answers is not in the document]
upsj: I would prefer size_type, but this will hopefully become a signed type in the future, which would require a check for negative values as well.

hip/test/preconditioner/isai_kernels.hip.cpp
```diff
@@ -53,7 +53,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 namespace {
 
 
-enum struct matrix_type { lower, upper };
+enum struct matrix_type { lower, upper, general };
```
upsj, line 56: spd

cuda/test/preconditioner/isai_kernels.cpp
```diff
@@ -41,19 +41,22 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
 #include "cuda/base/config.hpp"
 #include "cuda/test/utils.hpp"
+#include "matrices/config.hpp"
 
 
 namespace {
 
 
-enum struct matrix_type { lower, upper };
+enum struct matrix_type { lower, upper, general };
```
upsj, line 59: spd

common/preconditioner/isai_kernels.hpp.inc
```diff
@@ -233,6 +239,55 @@ __global__ __launch_bounds__(default_block_size) void generate_u_inverse(
...
+
+            apply_gauss_jordan_transform_with_rhs<subwarp_size>(
+                subwarp, piv, i, local_row, &rhs, status);
+        }
+
+        ValueType sol = subwarp.shfl(rhs, perm);
+
+        if (spd) {
+            sol /= sqrt(subwarp.shfl(rhs, num_elems - 1));
+        }
```
upsj, line 280: suggestion for clarity:
```suggestion
        if (spd) {
            auto diag = subwarp.shfl(rhs, num_elems - 1);
            sol /= sqrt(diag);
        }
```

common/preconditioner/isai_kernels.hpp.inc
```diff
@@ -283,36 +343,63 @@ __global__ __launch_bounds__(default_block_size) void generate_excess_system(
...
+    const int local_id = subwarp.thread_rank();
+
+    if (warp_id >= e_end - e_start) {
+        return;
+    }
+
+    const IndexType offset = excess_block_ptrs[e_start];
+    const IndexType block_begin = excess_block_ptrs[warp_id] - offset;
+    const IndexType block_end = excess_block_ptrs[warp_id + 1] - offset;
+    const ValueType scal =
```
upsj, line 382: same here

core/preconditioner/isai.cpp
```diff
@@ -108,17 +119,49 @@ std::shared_ptr<Csr> extend_sparsity(std::shared_ptr<const Executor> &exec,
...
+            to_invert.get(), inverted_row_ptrs.get_data()));
+
+        // Get nnz from device memory
+        auto inverted_nnz = static_cast<size_type>(
+            exec->copy_val_to_host(inverted_row_ptrs.get_data() + num_rows));
+
+        // Init arrays
+        Array<IndexType> inverted_col_idxs{exec, inverted_nnz};
+        Array<ValueType> inverted_vals{exec, inverted_nnz};
+        std::shared_ptr<Csr> inverted_base = Csr::create(
```
upsj, line 153: you could also use `auto` and `share(...)` here

core/preconditioner/isai.cpp
```diff
@@ -127,38 +170,94 @@ void Isai<IsaiType, ValueType, IndexType>::generate_inverse(
...
-        // solve it after transposing
-        std::unique_ptr<LinOpFactory> trs_factory;
-        if (is_lower) {
-            trs_factory = UpperTrs::build().on(exec);
-        } else {
-            trs_factory = LowerTrs::build().on(exec);
+    if (total_excess_dim > 0) {
+        bool done = false;
+        size_type block = 0;
+        while (true) {
```
upsj, line 190: can we write this in terms of a normal `while` loop instead? it seems the condition you are checking here is `while(block < num_rows)`

upsj (changes requested):
LGTM in general, only 3 points:

1. Do you test the excess_limit behavior somewhere? It looks to me like all tests use `0, num_rows` as the limit parameters in generate_excess_system.
2. Did you try out explicitly specifying the block pointers for Block Jacobi?
3. Could you try addressing the concerns about memory movement in the excess block pointers for excess limit handling?

fritzgoebel:
> LGTM in general, only 3 points:
> 
>     1. Do you test the excess_limit behavior somewhere? It looks to me like all tests use `0, num_rows` as the limit parameters in generate_excess_system.
> 
>     2. Did you try out explicitly specifying the block pointers for Block Jacobi?
> 
>     3. Could you try addressing the concerns about memory movement in the excess block pointers for excess limit handling?

1: good catch!
2: Not yet, I plan on doing that in a different pr later.
3: Please check if that's okay :)

thoasm (approved):
LGTM!
Some small nits, but nothing major.

common/components/warp_blas.hpp.inc
```diff
@@ -83,6 +83,52 @@ __device__ __forceinline__ void apply_gauss_jordan_transform(
...
+    if (key_col_elem == zero<ValueType>()) {
+        // TODO: implement error handling for GPUs to be able to properly
+        //       report it here
+        status = false;
```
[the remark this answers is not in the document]
upsj: For reference, that comment and `status` was already there 2 years ago, so I think Goran meant more general error handling mechanisms.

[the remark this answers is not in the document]
thoasm: Ah, good to know, thanks. In that case I will mark it as resolved.

upsj (approved):

common/components/warp_blas.hpp.inc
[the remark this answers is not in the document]
fritzgoebel: I don't think so, I carried this over from `apply_gauss_jordan_transform` and there it is also still present..

include/ginkgo/core/preconditioner/isai.hpp
```diff
@@ -143,6 +164,25 @@ class Isai : public EnableLinOp<Isai<IsaiType, ValueType, IndexType>>,
...
+         * upper or lower ISAI) or GMRES (for general ISAI). If this parameter
+         * is set to some m > 0, the excess system is solved as soon
+         * as its size supersedes m. This is repeated until the complete excess
+         * solution has been computed.
+         * Must be at least 0, default value 0.
+         */
+        size_type GKO_FACTORY_PARAMETER_SCALAR(excess_limit, 0u);
+
+        /**
+         * Factory for the Excess System solver
```
yhmtsai, line 182: Could you also add the default solver for different cases if it is not set by user

omp/test/preconditioner/isai_kernels.cpp
```diff
@@ -52,7 +52,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 namespace {
 
 
-enum struct matrix_type { lower, upper };
+enum struct matrix_type { lower, upper, general, spd };
 class Isai : public ::testing::Test {
```
yhmtsai, line 56:
```suggestion


class Isai : public ::testing::Test {
```

cuda/test/preconditioner/isai_kernels.cpp
```diff
@@ -41,19 +41,22 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
 #include "cuda/test/utils.hpp"
+#include "matrices/config.hpp"
 
 
 namespace {
 
 
-enum struct matrix_type { lower, upper };
+enum struct matrix_type { lower, upper, general, spd };
 class Isai : public ::testing::Test {
```
yhmtsai, line 60:
```suggestion


class Isai : public ::testing::Test {
```

hip/test/preconditioner/isai_kernels.hip.cpp
```diff
@@ -53,7 +53,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 namespace {
 
 
-enum struct matrix_type { lower, upper };
+enum struct matrix_type { lower, upper, general, spd };
 class Isai : public ::testing::Test {
```
yhmtsai, line 57:
```suggestion


class Isai : public ::testing::Test {
```

yhmtsai (approved):
LGTM. some minor format and documentation suggestion

PR #692, opened 2021-01-19
Tune and benchmark variables non intrusively.

tcojean:
This piece of code integrates and shows how to tune some variables,
namely the multiple used to decide on the number of warps used in COO.
This is a suggestion of what I could find as the simplest and most versatile
solution to this issue. If anyone has another idea I would be happy to try it out.

How: This is integrated into the SPMV benchmarks for now and is entirely
controlled with the new CMake option `GINKGO_BENCHMARK_ENABLE_TUNING`.

The setup works through the use of (dynamic) global variables so that
tuning is as little intrusive as possible. This should work even with the
"implementation selection" since this also has a dynamic component which
could use the same strategy as shown here.

Caveat: With this setup, only benchmarks can be compiled, other
executable codes such as tests would require to also add
`benchmark/utils/tuning_variables.cpp` or equivalent. This option should
of course not be turned on for general Ginkgo usage, but only for
tuning.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,6 +44,9 @@ option(GINKGO_CHECK_CIRCULAR_DEPS
     OFF)
 option(GINKGO_CONFIG_LOG_DETAILED
     "Enable printing of detailed configuration log to screen in addition to the writing of files," OFF)
+option(GINKGO_BENCHMARK_ENABLE_TUNING
+    "Enable tuning variables in the benchmarks. For specific use cases, manual code changes could be required."
+    OFF)
 set(GINKGO_VERBOSE_LEVEL "1" CACHE STRING
     "Verbosity level. Put 0 to turn off. 1 activates a few important messages.")
 if(MSVC)
@@ -73,6 +76,13 @@ option(BUILD_SHARED_LIBS "Build shared (.so, .dylib, .dll) libraries" ON)
 option(GINKGO_BUILD_HWLOC "Build Ginkgo with HWLOC. Default is ON. If a system HWLOC is not found, then we try to build it ourselves. Switch this OFF to disable HWLOC." ON)
 
 set(GINKGO_CIRCULAR_DEPS_FLAGS "-Wl,--no-undefined")
+
+if(GINKGO_BENCHMARK_ENABLE_TUNING)
+    # In this state, the tests and examples cannot be compiled without extra
+    # complexity/intrusiveness, so we simply disable them.
+    set(GINKGO_BUILD_TESTS OFF)
+    set(GINKGO_BUILD_EXAMPLES OFF)
+endif()
 
 if(BUILD_SHARED_LIBS AND (WIN32 OR CYGWIN) AND (GINKGO_BUILD_TESTS OR GINKGO_BUILD_EXAMPLES OR GINKGO_BUILD_BENCHMARKS))
     # Change shared libraries output only if this build has executable program
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -11,6 +11,12 @@ if (GINKGO_BUILD_CUDA AND GINKGO_BUILD_HIP AND GINKGO_HIP_PLATFORM MATCHES "hcc"
         "or use `export HIP_PLATFORM=nvcc` in your build environment instead.")
 endif()
 
+function(ginkgo_benchmark_add_tuning_maybe name)
+    if(GINKGO_BENCHMARK_ENABLE_TUNING)
+        target_sources(${name} PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/../utils/tuning_variables.cpp)
+    endif()
+endfunction()
+
 function(ginkgo_benchmark_cusp_linops name)
     target_compile_definitions("${name}" PRIVATE HAS_CUDA=1)
     target_link_libraries("${name}" ginkgo ${CUDA_RUNTIME_LIBS}
@@ -69,6 +75,7 @@ function(ginkgo_add_single_benchmark_executable name use_lib_linops macro_def)
     add_executable("${name}" ${ARGN})
     target_link_libraries("${name}" ginkgo gflags rapidjson)
     target_compile_definitions("${name}" PRIVATE "${macro_def}")
+    ginkgo_benchmark_add_tuning_maybe("${name}")
     if("${use_lib_linops}")
         if (GINKGO_BUILD_CUDA)
             ginkgo_benchmark_cusp_linops("${name}")
--- a/benchmark/matrix_generator/matrix_generator.cpp
+++ b/benchmark/matrix_generator/matrix_generator.cpp
@@ -43,6 +43,11 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "benchmark/utils/types.hpp"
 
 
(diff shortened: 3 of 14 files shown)


tcojean:
format!

tcojean:
I also added an R script to show how to plot JSON data such as this since it can be quite hard to get right. Note that I'm note an R expert and the code can probably be shrinked a bit. If you don't think that's too useful I can also remove this.

benchmark/spmv/spmv.cpp
```diff
@@ -96,6 +101,47 @@ void apply_spmv(const char *format_name, std::shared_ptr<gko::Executor> exec,
             system_matrix->apply(lend(b), lend(x_clone));
             exec->synchronize();
         }
+
+        // tuning run
+#ifdef GINKGO_BENCHMARK_ENABLE_TUNING
+        auto &format_case = spmv_case[format_name];
```
yhmtsai, line 107: add the check for coo and hybrid?

benchmark/matrix_generator/matrix_generator.cpp
```diff
@@ -42,6 +42,11 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "benchmark/utils/general.hpp"
 
 
+#ifdef GINKGO_BENCHMARK_ENABLE_TUNING
+#include "benchmark/utils/tuning_variables.hpp"
+#endif  // GINKGO_BENCHMARK_ENABLE_TUNING
```
yhmtsai, line 47: Is it needed? we do not use the matrix here

benchmark/matrix_statistics/matrix_statistics.cpp
```diff
@@ -44,6 +44,11 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "benchmark/utils/spmv_common.hpp"
 
 
+#ifdef GINKGO_BENCHMARK_ENABLE_TUNING
+#include "benchmark/utils/tuning_variables.hpp"
+#endif  // GINKGO_BENCHMARK_ENABLE_TUNING
```
yhmtsai, line 49: also here

benchmark/spmv/spmv.cpp
```diff
@@ -96,6 +101,47 @@ void apply_spmv(const char *format_name, std::shared_ptr<gko::Executor> exec,
...
+        add_or_set_member(tuning_case, "time",
+                          rapidjson::Value(rapidjson::kArrayType), allocator);
+        add_or_set_member(tuning_case, "values",
+                          rapidjson::Value(rapidjson::kArrayType), allocator);
+
+        // Enable tuning for this portion of code
+        gko::_tuning_flag = true;
+        // Select some values we want to tune.
+        std::vector<gko::size_type> tuning_values{
+            1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096};
```
yhmtsai, line 122: could you add the output to json for these tuning value?
I am not sure whether using command parameter to adjust these setting is necessary or not. Need pass via string and handle it.

yhmtsai (approved):
LGTM in general. Thanks for doing this.
Because this tuning factor only affect coo, maybe add `coo_` in the factor and the format check

benchmark/spmv/spmv.cpp
tcojean: This is supposed to be an example of tuning, and not specifically to tune that variable. The same scheme can be used to tune the SpMV formats in some other way, so I think a check for COO is not needed as if someone else tunes something else he will need the same code but maybe for CSR.

benchmark/matrix_generator/matrix_generator.cpp
tcojean: Yes it's needed, otherwise the executable will not compile. The library now has the symbols defined as extern but no actual definition of the symbols. Therefore, the executables need to also compile the variables and know the definition of the variables as well.

benchmark/spmv/spmv.cpp
tcojean: This is done, right now it creates two arrays, like the following:
```
         "tuning": {
            "time": [
                0.0179852619,
                0.010033904300000002,
                0.0064128235,
                0.005212025800000001,
                0.0049412771000000008,
                0.0045020674,
                0.0042101267,
                0.004052650200000001,
                0.0040126375,
                0.004072723,
                0.0039625192,
                0.0042129298,
                0.0043702082000000009
            ],
            "values": [
                1,
                2,
                4,
                8,
                16,
                32,
                64,
                128,
                256,
                512,
                1024,
                2048,
                4096
            ]
        }
```

tcojean: As for passing in the values via command line, I think it's better if it stays like this for now since anyway when someone will want to tune something else, the code for this might change.

CMakeLists.txt
```diff
@@ -73,6 +76,13 @@ option(BUILD_SHARED_LIBS "Build shared (.so, .dylib, .dll) libraries" ON)
 
 set(GINKGO_CIRCULAR_DEPS_FLAGS "-Wl,--no-undefined")
 
+if(GINKGO_BENCHMARK_ENABLE_TUNING)
+    # In this state, the tests and examples cannot be compiled without extra
+    # complexity/intrusiveness, so we simply disable them.
+    set(GINKGO_BUILD_TESTS OFF)
```
pratikvn, line 82: I think this should be mentioned in `BENCHMARKING.md`.

benchmark/CMakeLists.txt
```diff
@@ -11,6 +11,12 @@ if (GINKGO_BUILD_CUDA AND GINKGO_BUILD_HIP AND GINKGO_HIP_PLATFORM MATCHES "hcc"
         "or use `export HIP_PLATFORM=nvcc` in your build environment instead.")
 endif()
 
+function(ginkgo_benchmark_add_tuning_maybe name)
```
pratikvn, line 14: Why the `_maybe` at the end ?

pratikvn (approved):
I think the `BENCHMARKING.md` file needs to be updated. Otherwise, LGTM!

CMakeLists.txt
tcojean: I actually really do not want to document this feature as this benchmark feature is really for experts (i.e. us) if we need to tune some of our own kernels. Benchmark.md is also used for reproducing our paper results and so on, I think it would just confuse people if we talked about tuning or something there.

benchmark/CMakeLists.txt
tcojean: Because it doesn't always add the file, and does it only when needed (thankfully).

benchmark/matrix_generator/matrix_generator.cpp
yhmtsai: I see

benchmark/spmv/spmv.cpp
yhmtsai: you are right. I miss the value output. Thanks!

benchmark/spmv/spmv.cpp
yhmtsai: I think it is more difficult to use it on CSR because csr is required srow built on the construction.
But I am fine with leaving it.

upsj (approved):
LGTM - really nice approach!

PR #693, opened 2021-01-23
Add Compressed Basis GMRES (CB-GMRES)

thoasm:
This PR adds the compressed basis GMRES, which uses classical Gram-Schmidt GMRES where the basis can be stored in lower precision. The precision of the basis is decided by the user with a factory parameter.

This PR is an updated version of https://github.com/ginkgo-project/ginkgo/pull/640 because the underlying branch changed (to better reflect the content of the branch).


thoasm:
format!

thoasm:
label!

hartwiganzt:
I think all tests passed (except for the cleanup), only one windows test fails. Maybe you can already start reviewing this @tcojean @fritzgoebel @pratikvn @yhmtsai @upsj @Slaedr ?

examples/cb-gmres-solver/build.sh
```diff
@@ -0,0 +1,37 @@
...
+if [ $# -ne 1 ]; then
+    echo -e "Usage: $0 GINKGO_BUILD_DIRECTORY"
+    exit 1
+fi
+BUILD_DIR=$1
+THIS_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" &>/dev/null && pwd )
+
+# copy libraries
+LIBRARY_DIRS="core core/device_hooks reference omp cuda hip"
+LIBRARY_NAMES="ginkgo ginkgo_reference ginkgo_omp ginkgo_cuda ginkgo_hip"
```
yhmtsai, line 13: we need to update these example with dpcpp after we really have these dpcpp module.

examples/cb-gmres-solver/doc/tooltip
```diff
@@ -0,0 +1 @@
+Solve a simple linear system with CG. Read a matrix and right hand side from a file.
```
yhmtsai, line 1:
```suggestion
Solve a simple linear system with CB-GMRES. Read a matrix and right hand side from a file.
```

examples/gmres-solver/doc/intro.dox
```diff
@@ -0,0 +1,19 @@
+<a name="Intro"></a>
+<h1>Introduction</h1>
+This simple solver example should help you get started with Ginkgo. This example is meant for you to understand
```
yhmtsai, line 3:
```suggestion
This simple gmres example should help you get started with Ginkgo. This example is meant for you to understand
```

examples/gmres-solver/doc/tooltip
```diff
@@ -0,0 +1 @@
+Solve a simple linear system with CG. Read a matrix and right hand side from a file.
```
yhmtsai, line 1:
```suggestion
Solve a simple linear system with GMRES. Read a matrix and right hand side from a file.
```

examples/gmres-solver/gmres-solver.cpp
```diff
@@ -0,0 +1,152 @@
...
+    // multiple vectors is a now a natural extension of adding columns/rows are
+    // necessary.
+    using vec = gko::matrix::Dense<>;
+    // The gko::matrix::Csr class is used here, but any other matrix class such
+    // as gko::matrix::Coo, gko::matrix::Hybrid, gko::matrix::Ell or
+    // gko::matrix::Sellp could also be used.
+    using mtx = gko::matrix::Csr<>;
+    // The gko::solver::Cg is used here, but any other solver class can also be
+    // used.
+    using gmres = gko::solver::Gmres<>;
```
yhmtsai, line 58: Do all example need to add value_type such that user can switch the valuetype?

include/ginkgo/core/solver/cb_gmres.hpp
```diff
@@ -0,0 +1,192 @@
...
+
+
+namespace gko {
+namespace solver {
+
+
+constexpr size_type default_krylov_dim_mixed = 100u;
+
+
+enum class cb_gmres_storage_precision {
```
yhmtsai, line 58: Maybe I miss it in different files.
Could you also add these argument descriptions here or `with_storage_precision`?

reference/stop/residual_norm_kernels.cpp
```diff
@@ -34,6 +34,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
 #include <algorithm>
+#include <iostream>
```
yhmtsai, line 37: forget to clear it?

yhmtsai (commented):
the first part of my review
I do not go into the implementation and algorithm of CB-GMRES.
you may already do some comments.

common/solver/cb_gmres_kernels.hpp.inc
```diff
@@ -0,0 +1,1593 @@
...
+    if (row_idx < num_rows && col_idx < num_cols &&
+        !stop_status[col_idx].has_stopped() &&
+        !reorth_status[col_idx].has_stopped()) {
+        const auto next_krylov_idx = row_idx * stride_next_krylov + col_idx;
+        for (size_type k = 0; k < num_iters; ++k) {
+            // const auto krylov_idx =
+            //    row_idx * stride_krylov + k * num_cols + col_idx;
+            const auto hessenberg_idx = k * stride_hessenberg + col_idx;
+            const auto buffer_idx = k * stride_buffer + col_idx;
+            next_krylov_basis[next_krylov_idx] -=
```
yhmtsai, line 1155: register variable

common/solver/cb_gmres_kernels.hpp.inc
```diff
@@ -0,0 +1,1593 @@
...
+                buffer_iter[buffer_idx] * krylov_bases(k, row_idx, col_idx);
+            if ((row_idx == 0) && !reorth_status[col_idx].has_stopped()) {
+                hessenberg_iter[hessenberg_idx] += buffer_iter[buffer_idx];
+            }
+        }
+    }
+}
+
```
yhmtsai, line 1163:
```suggestion


```

common/solver/cb_gmres_kernels.hpp.inc
```diff
@@ -0,0 +1,1593 @@
...
+        if (row_idx < num_rows && col_idx < num_cols &&
+            !stop_status[col_idx].has_stopped() &&
+            !reorth_status[col_idx].has_stopped()) {
+            const auto next_krylov_idx = row_idx * stride_next_krylov + col_idx;
+            for (size_type k = 0; k < num_iters; ++k) {
+                // const auto krylov_idx =
+                //    row_idx * stride_krylov + k * num_cols + col_idx;
+                const auto hessenberg_idx = k * stride_hessenberg + col_idx;
+                const auto buffer_idx = k * stride_buffer + col_idx;
+                next_krylov_basis[next_krylov_idx] -=
```
yhmtsai, line 1227: also

yhmtsai (commented):
some comments and questions of gpu kernels

omp/solver/cb_gmres_kernels.cpp
```diff
@@ -0,0 +1,511 @@
...
+
+template <typename ValueType, typename Accessor3d>
+void finish_arnoldi_CGS2(matrix::Dense<ValueType> *next_krylov_basis,
+                         Accessor3d krylov_bases,
+                         matrix::Dense<ValueType> *hessenberg_iter,
+                         matrix::Dense<ValueType> *buffer_iter,
+                         matrix::Dense<remove_complex<ValueType>> *arnoldi_norm,
+                         size_type iter, const stopping_status *stop_status)
+{
+    const remove_complex<ValueType> eta = 1.0 / sqrt(2.0);
```
yhmtsai, line 79: use 1/sqrt(2) in omp, but use eta = 1/2 in cuda.
Are they for different purpose?

reference/solver/cb_gmres_kernels.cpp
```diff
@@ -0,0 +1,458 @@
...
+                         matrix::Dense<remove_complex<ValueType>> *arnoldi_norm,
+                         size_type iter, const stopping_status *stop_status)
+{
+    static_assert(
+        std::is_same<ValueType,
+                     typename Accessor3d::accessor::arithmetic_type>::value,
+        "ValueType must match arithmetic_type of accessor!");
+    const remove_complex<ValueType> eta = 1.0 / sqrt(2.0);
+    for (size_type i = 0; i < next_krylov_basis->get_size()[1]; ++i) {
+        arnoldi_norm->at(0, i) = zero<remove_complex<ValueType>>();
```
tcojean, line 77: Maybe store this type somewhere as you keep using it?

reference/solver/cb_gmres_kernels.cpp
```diff
@@ -0,0 +1,458 @@
...
+        }
+    }
+}
+
+GKO_INSTANTIATE_FOR_EACH_CB_GMRES_TYPE(
+    GKO_DECLARE_CB_GMRES_INITIALIZE_2_KERNEL);
+
+
+// template <typename ValueType, typename ValueTypeKrylovBases, bool
+// Reorthogonalization, bool MGS_CGS>
```
tcojean, line 398:
```suggestion
```

reference/test/solver/cb_gmres_kernels.cpp
```diff
@@ -0,0 +1,470 @@
...
+    using gmres_type = gko::solver::CbGmres<value_type>;
+
+    /**
+     * Used to extract the precision of the given value_type.
+     *
+     * @internal
+     * This struct is required because the `r<T>` stores a value of type
+     * remove_complex<T>, which is not possible for half since it can't be
+     * constructed as a constexpr.
+     */
```
tcojean, line 76: That's a good point. I think this is the first time we test half altogether. Should we move this into `r<T>` somehow?

omp/solver/cb_gmres_kernels.cpp
```diff
@@ -0,0 +1,511 @@
...
+        remove_complex<ValueType> nrm = zero<remove_complex<ValueType>>();
+#pragma omp parallel for reduction(addnc : nrm)
+        for (size_type j = 0; j < next_krylov_basis->get_size()[0]; ++j) {
+            nrm += squared_norm(next_krylov_basis->at(j, i));
+        }
+        arnoldi_norm->at(0, i) = nrm * eta;
+        // nrmP = norm(next_krylov_basis)
+#ifdef TIMING
+        auto start_1 = std::chrono::steady_clock::now();
+#endif
```
tcojean, line 98: Again, are the timing really needed? I think it's good that you checked the timing while developing though to ensure there was no slowdown of some loops as we saw sometimes. Thanks for that.

omp/solver/cb_gmres_kernels.cpp
```diff
@@ -0,0 +1,511 @@
...
+            << "time_1 / time_2(" << iter << ") = "
+            << std::chrono::duration_cast<double_seconds>(time_1).count() /
+                   std::chrono::duration_cast<double_seconds>(time_2).count()
+            << std::endl;
+#endif
+        // for i in 1:iter
+        //     next_krylov_basis  -= hessenberg(iter, i) * krylov_bases(:, i)
+        // end
+        // ValueType nrm = zero<ValueType>();
+        nrm = zero<remove_complex<ValueType>>();
```
tcojean, line 142: Again, storing the `rc_vtype` would probably help in terms of coding style since it's used everywhere.

tcojean (changes requested):
LGTM in general though I have some minor comments and I think some cleanup could be done as well (in core, examples, ...). I also did not review CUDA/HIP yet.

tcojean:
Another big question, since #691 will be merged soon, do we also need to do the same process here as well? Probably not as part of this PR?

benchmark/solver/solver.cpp
```diff
@@ -165,45 +229,112 @@ std::unique_ptr<gko::LinOpFactory> create_solver(
...
+             return gko::solver::Gmres<etype>::build()
+                 .with_criteria(create_criterion(exec))
+                 .with_krylov_dim(FLAGS_gmres_restart)
+                 .with_preconditioner(give(precond))
+                 .on(exec);
+         }},
+        {"cb_gmres_double",
+         [](std::shared_ptr<const gko::Executor> exec,
+            std::shared_ptr<const gko::LinOpFactory> precond) {
+             return gko::solver::CbGmres<double>::build()
```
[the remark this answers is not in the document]
thoasm: Yes, I failed rebasing here...

thoasm:
@tcojean Would probably be useful, but I would prefer if we / I did this in another PR.

common/solver/cb_gmres_kernels.hpp.inc
```diff
@@ -0,0 +1,1593 @@
...
+    const auto tidx = threadIdx.x;
+    const auto tidy = threadIdx.y;
+    const auto col_idx = blockIdx.x * default_dot_dim + tidx;
+    const auto num = ceildiv(num_rows, gridDim.y);
+    const auto start_row = blockIdx.y * num;
+    const auto end_row =
+        ((blockIdx.y + 1) * num > num_rows) ? num_rows : (blockIdx.y + 1) * num;
+    // Used that way to get around dynamic initialization warning and
+    // template error when using `reduction_helper_array` directly in `reduce`
+    __shared__ UninitializedArray<remove_complex<ValueType>,
```
tcojean, line 172: Maybe also store the `rc_vtype` for this function, this could improve coding style a bit.

common/solver/cb_gmres_kernels.hpp.inc
```diff
@@ -0,0 +1,1593 @@
...
+    if (col_idx < num_cols && !stop_status[col_idx].has_stopped()) {
+        for (size_type i = start_row + tidy; i < end_row;
+             i += default_dot_dim) {
+            const auto next_krylov_idx = i * stride_next_krylov + col_idx;
+            local_max = (local_max >= abs(next_krylov_basis[next_krylov_idx]))
+                            ? local_max
+                            : abs(next_krylov_basis[next_krylov_idx]);
+            // local_max = (local_max >= next_krylov_basis[next_krylov_idx])
+            //                 ? local_max
+            //                 : next_krylov_basis[next_krylov_idx];
```
tcojean, line 234:
```suggestion
```

common/solver/cb_gmres_kernels.hpp.inc
```diff
@@ -0,0 +1,1593 @@
...
+    if (col_idx < num_cols) {
+        for (size_type i = start_row + tidy; i < end_row;
+             i += default_dot_dim) {
+            const auto next_krylov_idx = i * stride_next_krylov + col_idx;
+            local_max = (local_max >= abs(next_krylov_basis[next_krylov_idx]))
+                            ? local_max
+                            : abs(next_krylov_basis[next_krylov_idx]);
+            // local_max = (local_max >= next_krylov_basis[next_krylov_idx])
+            //                 ? local_max
+            //                 : next_krylov_basis[next_krylov_idx];
```
tcojean, line 287:
```suggestion
```

common/solver/cb_gmres_kernels.hpp.inc
```diff
@@ -0,0 +1,1593 @@
...
+    if (tidx == 0 && new_col_idx < num_cols &&
+        !stop_status[new_col_idx].has_stopped()) {
+        const auto hessenberg_idx = k * stride_hessenberg + new_col_idx;
+        atomic_add(hessenberg_iter + hessenberg_idx, sum);
+    }
+}
+
+
+template <int block_size, typename ValueType, typename Accessor3d>
+__global__ __launch_bounds__(block_size) void singledot_kernel_num_iters_2(
```
tcojean, line 610: There are a lot of dot kernels. Are all really needed? I would expect you don't use all of these.

common/solver/cb_gmres_kernels.hpp.inc
```diff
@@ -0,0 +1,1593 @@
...
+        }
+    }
+}
+
+
+// Must be called with at least `num_rows * stride_next_krylov` threads in
+// total.
+template <int block_size, typename ValueType, typename Accessor3d>
+__global__
+    __launch_bounds__(block_size) void update_next_krylov_kernel_num_iters_4(
```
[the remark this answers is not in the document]
tcojean: Same here, I expect not all are used. I guess this is the one you mentioned during the call yesterday actually.

common/solver/cb_gmres_kernels.hpp.inc
```diff
@@ -0,0 +1,1593 @@
...
+            }
+        }
+    }
+}
+
+// Must be called with at least `num_rows * stride_next_krylov` threads in
+// total.
+template <int block_size, typename ValueType, typename Accessor3d>
+__global__
+    __launch_bounds__(block_size) void update_next_krylov_kernel_num_iters_and_add_1(
```
tcojean, line 1168: Again multiple versions (and check, and add, and add 1)

tcojean (commented):
Some final comments

thoasm (commented):
Answers / Comments on Review

examples/cb-gmres-solver/build.sh
thoasm: I completely removed these examples from this PR since a lot of it was outdated / not needed anymore.

omp/solver/cb_gmres_kernels.cpp
thoasm: The reason for that was actually that for GPUs, we did not compute the `sqrt` on both sides of the comparison. The results were the same, but the values differed. This is changed now, so all `arnoldi_norms` are now the same between all kernels.

benchmark/run_all_benchmarks.sh
```diff
@@ -85,16 +96,39 @@ else
...
 
+if [ "${SOLVERS_INITIAL_GUESS}" == "random" ]; then
+    SOLVERS_INITIAL_GUESS_FLAG="--initial_guess_generation=random"
+elif [ "${SOLVERS_INITIAL_GUESS}" == "0" ]; then
+    SOLVERS_INITIAL_GUESS_FLAG="--initial_guess_generation=0"
+elif [ "${SOLVERS_INITIAL_GUESS}" == "rhs" ]; then
+    SOLVERS_INITIAL_GUESS_FLAG="--initial_guess_generation=rhs"
+else
+    echo "SOLVERS_RHS does not support the value \"${SOLVERS_RHS}\"." 1>&2
+    echo "The following values are supported: \"1\", \"random\" and \"sinus\"" 1>&2
```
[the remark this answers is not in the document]
thoasm: It would be nice to have a better solution for that, but I currently don't know what the best approach would be. I assume we would have to re-write this script entirely for that.

benchmark/solver/solver.cpp
```diff
@@ -165,45 +229,112 @@ std::unique_ptr<gko::LinOpFactory> create_solver(
...
+                 .with_storage_precision(
+                     gko::solver::cb_gmres_storage_precision::keep)
```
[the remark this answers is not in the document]
thoasm: In order to have multiple precision run at the same time, the identifying string needs to be different (especially in the result file). I decided to make `cb_gmres_` a prefix, and create the appropriate enum variable according to the suffix (as @upsj suggested).

examples/gmres-solver/doc/intro.dox
```diff
@@ -0,0 +1,19 @@
...
+  <li> <b>Introduction:</b>This gives an overview of the example and mentions
+  any interesting aspects in the example that might help the reader.
+  <li> <b>The commented program:</b> This section is intended for you to
+  understand the details of the example so that you can play with it and understand
+  Ginkgo and its features better.
+  <li> <b>Results:</b> This section shows the results of the code when run. Though the
+  results may not be completely the same, you can expect the behaviour to be similar. 
+  <li> <b>The plain program:</b> This is the complete code without any comments to have
+  an complete overview of the code.
+  </ol>
```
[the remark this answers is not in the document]
thoasm: I removed the examples entirely since both did not add a lot of value compared to `simple-solver` (only one factory parameter would be added).

core/solver/cb_gmres.cpp
```diff
@@ -0,0 +1,641 @@
...
+        }
+        bool perform_reset{false};
+        decltype(krylov_dim_) forced_iterations{0};
+        const decltype(forced_iterations) forced_limit{krylov_dim_ / 10};
+        decltype(krylov_dim_) total_checks{0};  // TODO: Remove debug output
+        const char *type_str = type_string<storage_type>::get();
+        // TODO: take care of multiple RHS. Currently, we restart
+        // everything,
+        //       even though a lot of parts might have already converged!
+        //       Use `one_changed` to take care of that!
```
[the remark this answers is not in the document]
thoasm: This is actually already done, I just did not remove the comment. Thanks for reminding me and finding it.

reference/test/solver/cb_gmres_kernels.cpp
thoasm: We have this in `r<T>` now: You can specify `r<T, StorageType>`, which uses the `epsilon` from type `T`, and stores it as a `StorageType`. I added that in the accessor PR #643.

pratikvn (commented):
Some minor comments.

include/ginkgo/core/solver/cb_gmres.hpp
```diff
@@ -0,0 +1,228 @@
...
+        return storage_precision_;
+    }
+
+    GKO_CREATE_FACTORY_PARAMETERS(parameters, Factory)
+    {
+        /**
+         * Determines which storage type is used.
+         */
+        cb_gmres_storage_precision GKO_FACTORY_PARAMETER_SCALAR(
+            storage_precision, cb_gmres_storage_precision::keep);
```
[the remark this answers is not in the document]
thoasm: It makes it a GMRES which uses CGS instead of MGS. If the user does not specify a compression, I thought it is safer to not compress anything.

BENCHMARKING.md
```diff
@@ -295,17 +295,28 @@ The supported environment variables are described in the following list:
     library formats (cuSPARSE with `cusp_` prefix or hipSPARSE with `hipsp_`
     prefix) can be used as well. Multiple options can be passed. The default is
     `csr,coo,ell,hybrid,sellp`.
-* `SOLVERS={bicgstab,bicg,cg,cgs,fcg,gmres,lower_trs,upper_trs}` - the solvers
-    which should be benchmarked. Multiple options can be passed. The default
-    is `cg`. Note that `lower/upper_trs` by default don't use a preconditioner,
-    as they are by default exact direct solvers.
+* `SOLVERS={bicgstab,bicg,cg,cgs,fcg,gmres,cb_gmres_{keep,reduce1,reduce2,integer,ireduce1,ireduce2},lower_trs,upper_trs}`
+    - the solvers which should be benchmarked. Multiple options can be passed.
+    The default is `bicgstab,cg,cgs.fcg.gmres,idr`. Note that `lower/upper_trs`
```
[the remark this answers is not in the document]
thoasm: Thanks, I added `reduce1` to the default benchmark.

include/ginkgo/core/solver/cb_gmres.hpp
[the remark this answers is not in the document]
pratikvn: Being that the class itself is named Compressed basis GMRES, I guess by default, you would expect some compression ?

Also, we don't seem to mention in the regular GMRES documentation that we use MGS. Can you update that documentation to say so ?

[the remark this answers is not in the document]
thoasm: Good point, updated it accordingly.

benchmark/run_all_benchmarks.sh
```diff
@@ -2,18 +2,18 @@
...
 if [ ! "${DRY_RUN}" ]; then
-    echo "DRY_RUN     environment variable not set - assuming \"false\"" 1>&2
     DRY_RUN="false"
+    echo "DRY_RUN     environment variable not set - assuming \"${DRY_RUN}\"" 1>&2
 fi
 
 if [ ! "${EXECUTOR}" ]; then
-    echo "EXECUTOR    environment variable not set - assuming \"cuda\"" 1>&2
     EXECUTOR="cuda"
+    echo "EXECUTOR    environment variable not set - assuming \"${EXECUTOR}\"" 1>&2
```
tcojean, line 16: Nice way of not repeating these.

benchmark/solver/solver.cpp
```diff
@@ -154,61 +219,98 @@ std::shared_ptr<const gko::stop::CriterionFactory> create_criterion(
...
+            .with_num_rhs(FLAGS_nrhs)
+            .on(exec);
+    } else if (description == "upper_trs") {
+        return gko::solver::UpperTrs<etype>::build()
+            .with_num_rhs(FLAGS_nrhs)
+            .on(exec);
+    } else if (description == "overhead") {
+        return add_criteria_precond_finalize<gko::Overhead<etype>>(exec,
+                                                                   precond);
+    }
```
tcojean, line 310: For these, we could keep the map and pass anything there, but I guess this form is more generic.

benchmark/solver/solver.cpp
```diff
@@ -154,61 +219,98 @@ std::shared_ptr<const gko::stop::CriterionFactory> create_criterion(
...
+    std::shared_ptr<const gko::LinOpFactory> precond,
+    const std::string &description)
+{
+    std::string cb_gmres_prefix("cb_gmres_");
+    if (description.find(cb_gmres_prefix) == 0) {
+        auto s_prec = gko::solver::cb_gmres_storage_precision::keep;
+        const auto spec = description.substr(cb_gmres_prefix.length());
+        if (spec == "keep") {
+            s_prec = gko::solver::cb_gmres_storage_precision::keep;
+        } else if (spec == "reduce1") {
```
tcojean, line 252: Technically the first one is unneeded?
```suggestion
        if (spec == "reduce1") {
```

tcojean (approved):
LGTM. You removed all examples but I think it could be useful to add one specific to CbGmres since this uses new features for precision reduction, maybe an example which shows improved timing between `keep` and `reduce2`?

benchmark/solver/solver.cpp
thoasm: I thought about keeping the map, however, the map is less generic, a bit more complicated to use and more code, so I thought I would simply have all solvers in the same function.

benchmark/solver/solver.cpp
thoasm: We don't need the assignment, but we need the case `keep` because if the suffix does not match any of the choices, it throws a `std::range_error` exception.

tcojean:
Quick comment, we already have ani4 mtx, would it be possible to use the one we already have in the `matrices` directory? Otherwise, simply copying it when building the example would be easier.

thoasm:
@tcojean I suppose we could copy it from the `matrices/test` directory into the example build directory. The reason I did not do it is because it is harder to modify and play around with the example that way, and it is not easy to understand where the matrix is comming from.
My thought process was that since the matrix is already present, it should not add more storage requirements to the repository.

thoasm:
Small update: I can't access the header `matrices/config.hpp`, since it is not a public header (understandably so). I can simply copy the data from `matrices/test/ani4.mtx` to `example/cb-gmres/data/A.mtx`, however, I realized that it takes 37s to complete on reference when build on `Debug`, so I am experimenting with other matrices at the moment.

thoasm:
Example should be finished now. It is copying the matrix `matrices/test/ani1.mtx` into the test folder while building and is checked for validity. Every review comment should be addressed now.

pratikvn (commented):
Nice work! In general looks good to me.

I have a few questions:
1. Unification of kernels: It seems to me that many kernels are exactly the same between GMRES and CB-GMRES, example: `calculate_sin_and_cos_kernel`, `calculate_residual_norm_kernel`, `givens_rotation_kernel` and `solve_upper_triangular_kernel` Can we create a new file `common_gmres_kernels.hpp.inc` and move these kernels there so that the duplication is reduced ? 
2. In the documentation, you mention that this is CGS, but I think you should clarify that you use CGS+ reorthogonalization, as that is a different algorithm than plain CGS. 
3. I guess you already looked at using MGS instead of CGS and observed more benefits with CGS, but nevertheless if might be interesting to have the option to do MGS as well for this ? I guess you technically have four options (CGS, MGS) x (re-ortho, no re-ortho), with ortho stability, CGS < ( MGS =? CGS+re-ortho) < MGS + re-ortho and it might be interesting to explore the different options.
4. In the literature it seems that re-ortho might not always be necessary. The bound seems to be `|| I - Q^T*Q|| < mach_eps * cond(A)`. You seem to always perform a re-orthogonalization. Did you need to do this because of the loss of orthogonality due to the storage in reduced precision ? I would have assumed that `mach_eps` would be the one of the arithmetic type, because it seems that catastrophic cancellation is at play here. This is another case for making the re-orthogonalization loop variable `l` a factory parameter. If someone knows that their matrix is well-conditioned, then they can switch off the re-orthogonalization.

pratikvn (approved):
Looks like i clicked comment instead of approve before.

reference/test/solver/cb_gmres_kernels.cpp
```diff
@@ -0,0 +1,460 @@
...
+                    .with_reduction_factor(this->reduction_factor())
+                    .on(this->exec))
+            .on(this->exec);
+    auto solver = factory->generate(this->mtx2);
+    auto b = gko::initialize<Mtx>({33.0, 20.0, 20.0}, this->exec);
+    auto x = gko::initialize<Mtx>({0.0, 0.0, 0.0}, this->exec);
+
+    solver->apply(b.get(), x.get());
+
+    GKO_ASSERT_MTX_NEAR(x, l({1.0, 4.0, 8.0}), 8 * this->assert_precision());
```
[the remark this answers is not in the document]
thoasm: Yes, I basically played around with it, and 8 worked fine for all precisions (I could have also used `10` here). In a lot of solver tests, we do the same (in some cases even with `1e2`).

common/solver/cb_gmres_kernels.hpp.inc
```diff
@@ -0,0 +1,766 @@
...
+        !stop_status[new_col_idx].has_stopped()) {
+        const auto hessenberg_idx = k * stride_hessenberg + new_col_idx;
+        atomic_add(hessenberg_iter + hessenberg_idx, sum);
+    }
+}
+
+
+template <int block_size, typename ValueType, typename Accessor3d>
+__global__ __launch_bounds__(block_size) void singledot_kernel(
+    size_type num_iters, size_type num_rows,
```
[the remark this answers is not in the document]
thoasm: That is correct, we use `blockIdx.y` instead of `num_iters` here (which are supposed to be equal).

thoasm:
@pratikvn Regarding your questions:
1. Unification of kernels: Yes, I can and will do that (in the next iteration of the review update, I forgot to do it in the last one).
2. CGS+reorthogonalization: Thanks for noticing, I updated the documentation.
3. Option for both MGS & CGS: We originally had this as an option (by setting a Macro differently), but that got really hard to maintain (the current code is already hard to understand) and the result showed that the current implementation beats all others in close to all scenarios.
4. Optional reorthogonalization: We don't always perform a reorthogonalization, we test if it is necessary first (at least in the GPU implementation, in reference, we always reorthogonalize).

tcojean (approved):
LGTM. Thanks for the example.

core/solver/cb_gmres.cpp
```diff
@@ -0,0 +1,508 @@
...
+            //                         cos(j)*hessenberg(j+1)
+            //     hessenberg(j)    =  temp;
+            // end
+            // Calculate sin and cos
+            // hessenberg(restart_iter)   =
+            // cos(restart_iter)*hessenberg(restart_iter) +
+            //                      sin(restart_iter)*hessenberg(restart_iter)
+            // hessenberg(restart_iter+1) = 0
+            // End apply givens rotation
+            // Calculate residual norm
```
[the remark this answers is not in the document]
thoasm: Both of these variables are only helper for GPUs. I added a comment explaining that.

cuda/solver/cb_gmres_kernels.cu
```diff
@@ -0,0 +1,534 @@
...
+        <<<ceildiv(dim_size[1], default_block_size), default_block_size>>>(
+            as_cuda_type(arnoldi_norm->get_values()), stride_arnoldi,
+            as_cuda_type(hessenberg_iter->get_values()), stride_hessenberg,
+            iter + 1, as_cuda_accessor(krylov_bases), as_cuda_type(stop_status),
+            as_cuda_type(reorth_status), as_cuda_type(num_reorth->get_data()));
+    numReorth = 0;
+    exec->get_master()->copy_from(exec.get(), 1, num_reorth->get_const_data(),
+                                  &numReorth);
+    // numReorth <= number of next_krylov vector to be reorthogonalization
+    for (size_type l = 1; (numReorth > 0) && (l < 3); l++) {
```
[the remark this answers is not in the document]
thoasm: I don't really know why the `3` was added, but so far, but we do only perform the re-orthogonalization when necessary. I would prefer to keep it as is, and only change it when it is clear that this is a source for a problem.

reference/solver/cb_gmres_kernels.cpp
```diff
@@ -0,0 +1,460 @@
...
+            }
+        }
+        // for i in 1:iter
+        //     hessenberg(iter, i) = next_krylov_basis' * krylov_bases(:, i)
+        // end
+        for (size_type k = 0; k < iter + 1; ++k) {
+            for (size_type j = 0; j < next_krylov_basis->get_size()[0]; ++j) {
+                next_krylov_basis->at(j, i) -=
+                    hessenberg_iter->at(k, i) * krylov_bases(k, j, i);
+            }
```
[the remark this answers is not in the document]
thoasm: The problem is that even for CUDA and HIP, all these functions are unique, and we rarely re-use them. I decided to keep it as it is because understanding the for-loops is simple because each one is quite short. I think it would be more complicated to understand this `finish_arnoldi_CGS` method if it was littered with multiple function calls, where you would need to jump around to get the bigger picture.

cuda/solver/cb_gmres_kernels.cu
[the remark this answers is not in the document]
thoasm: From @josealiaga
> As Pratik Nayak says, the number 3 is the typical value in the literature ([link](https://www.sciencedirect.com/science/article/pii/S0898122105003366)). It could be defined as a parameter, but I don't know if it makes sense ...

tcojean:
I don't remember, is there any issue left here or can that be merged already?

pratikvn:
From my side it looks good and can be merged.

thoasm:
@tcojean @pratikvn I realized that @josealiaga is not in the `contributors.txt` list, so I am waiting for him to put himself in. After that, it can be merged.

PR #695, opened 2021-01-27
AmgxPgm cuda/hip/omp implementation and matrix/array utils

yhmtsai:
This PR implements the amgxpgm device code with aggregation size = 2 except for dpc++.

Also, collect those utils in core/test/utils
- array generator
- symmetric/Hermitian/diag_dominant/hpd(spd)
- add doc and type test

TODO:
- [x] Merge #528 
- [x] Add the test of the test utils


omp/test/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,323 @@
...
+    auto d_csr_temp = Csr::create(omp, gko::dim<2>{n, n},
+                                  d_weight_csr->get_num_stored_elements());
+
+    gko::kernels::omp::amgx_pgm::amgx_pgm_generate(
+        omp, d_weight_csr.get(), d_agg, d_csr_coarse.get(), d_csr_temp.get());
+    gko::kernels::reference::amgx_pgm::amgx_pgm_generate(
+        ref, weight_csr.get(), agg, csr_coarse.get(), csr_temp.get());
+
+    GKO_ASSERT_MTX_NEAR(d_csr_coarse, csr_coarse, 1e-14);
+}
```
tcojean, line 320: Could you also add a test `GenerateMgLevelIsEquivalentToRef` like for HIP/CUDA?

hip/test/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+    GKO_ASSERT_MTX_NEAR(d_csr_coarse, csr_coarse, 1e-14);
+}
+
+
+TEST_F(AmgxPgm, GenerateMgLevelIsEquivalentToRef)
+{
+    initialize_data();
+    auto mg_level_factory = gko::multigrid::AmgxPgm<double, int>::build()
+                                .with_deterministic(true)
```
tcojean, line 344: Maybe also a test with `false` here? Would that be possible or are the results way too different that you cannot find a tolerance for which it works?

cuda/test/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+
+TEST_F(AmgxPgm, GenerateMgLevelIsEquivalentToRef)
+{
+    initialize_data();
+    auto mg_level_factory = gko::multigrid::AmgxPgm<double, int>::build()
+                                .with_deterministic(true)
+                                .on(ref);
+    auto d_mg_level_factory = gko::multigrid::AmgxPgm<double, int>::build()
+                                  .with_deterministic(true)
```
tcojean, line 347: Same thing with deterministic false.

common/multigrid/amgx_pgm_kernels.hpp.inc
```diff
@@ -0,0 +1,341 @@
...
+        // Use the smaller index as agg point
+        auto group = min(tidx, neighbor);
+        agg_vals[tidx] = group;
+        agg_vals[neighbor] = group;
+    }
+}
+
+
+template <typename IndexType>
+__global__ __launch_bounds__(default_block_size) void activate_kernel(
```
tcojean, line 73: Isn't this exactly the same as `replace_kernel`?

tcojean (approved):
LGTM. I only have a few comments.

cuda/multigrid/amgx_pgm_kernels.cu
```diff
@@ -111,8 +177,55 @@ template <typename ValueType, typename IndexType>
...
+    // prefix sum of temp_row_ptrs
+    components::prefix_sum(exec, temp->get_row_ptrs(), coarse_nrows + 1);
+    // copy source -> to coarse and change column index
+    kernel::move_row_kernel<<<grid, default_block_size>>>(
+        source_nrows, agg.get_const_data(), row_map.get_const_data(),
+        source->get_const_row_ptrs(), source->get_const_col_idxs(),
+        as_cuda_type(source->get_const_values()), temp->get_const_row_ptrs(),
+        temp->get_col_idxs(), as_cuda_type(temp->get_values()));
+    // sort csr
+    csr::sort_by_column_index(exec, temp);
```
pratikvn, line 205: Maybe you can check if it is sorted, with `temp->is_sorted_by_column_index()` function of CSR, and sort it only if it not already sorted ? I guess if the `move_row_kernel` above does not give you a sorted output, then I guess you need this.

cuda/test/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,362 @@
...
+#include "core/multigrid/amgx_pgm_kernels.hpp"
+#include "core/test/utils/matrix_generator.hpp"
+#include "cuda/test/utils.hpp"
+
+
+namespace {
+
+
+template <typename Array, typename ValueDistribution, typename Engine>
+Array generate_random_array(gko::size_type num, ValueDistribution &&value_dist,
```
pratikvn, line 62: This function might be useful in other cases as well. Can you move it to `core/test/utils.hpp` ?

cuda/test/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,362 @@
...
+        d_unfinished_agg = unfinished_agg;
+        d_strongest_neighbor = strongest_neighbor;
+        d_coarse_vector->copy_from(coarse_vector.get());
+        d_fine_vector->copy_from(fine_vector.get());
+        d_weight_csr->copy_from(weight_csr.get());
+        d_weight_diag->copy_from(weight_diag.get());
+        d_system_mtx->copy_from(system_mtx.get());
+    }
+
+    void make_symetric(Mtx *mtx)
```
pratikvn, line 157:
```suggestion
    void make_symmetric(Mtx *mtx)
```

cuda/test/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,362 @@
...
+    {
+        for (int i = 0; i < mtx->get_size()[0]; ++i) {
+            for (int j = i + 1; j < mtx->get_size()[1]; ++j) {
+                mtx->at(i, j) = mtx->at(j, i);
+            }
+        }
+    }
+
+    // only for real value
+    void make_absoulte(Mtx *mtx)
```
pratikvn, line 167:
```suggestion
    void make_absolute(Mtx *mtx)
```

cuda/test/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,362 @@
...
+            for (int j = 0; j < mtx->get_size()[1]; ++j) {
+                sum += abs(mtx->at(i, j));
+            }
+            mtx->at(i, i) = sum;
+        }
+    }
+
+    void make_spd(Mtx *mtx)
+    {
+        make_symetric(mtx);
```
pratikvn, line 190:
```suggestion
        make_symmetric(mtx);
```

cuda/test/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,362 @@
...
+        make_diag_dominant(mtx);
+    }
+
+    void make_weight(Mtx *mtx)
+    {
+        make_symetric(mtx);
+        make_absoulte(mtx);
```
pratikvn, line 197:
```suggestion
        make_symmetric(mtx);
        make_absolute(mtx);
```

hip/multigrid/amgx_pgm_kernels.hip.cpp
```diff
@@ -62,25 +68,62 @@ namespace hip {
...
 
 template <typename IndexType>
 void renumber(std::shared_ptr<const HipExecutor> exec, Array<IndexType> &agg,
-              IndexType *num_agg) GKO_NOT_IMPLEMENTED;
+              IndexType *num_agg)
+{
+    const auto num = agg.get_num_elems();
+    Array<IndexType> agg_map(exec, num + 1);
+    components::fill_array(exec, agg_map.get_data(), agg_map.get_num_elems(),
+                           zero<IndexType>());
```
pratikvn, line 116: nit: can be simplified.
```suggestion
    agg_map.fill(zero<IndexType>());
```

hip/test/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,362 @@
...
+#include "core/multigrid/amgx_pgm_kernels.hpp"
+#include "core/test/utils/matrix_generator.hpp"
+#include "hip/test/utils.hip.hpp"
+
+
+namespace {
+
+
+template <typename Array, typename ValueDistribution, typename Engine>
+Array generate_random_array(gko::size_type num, ValueDistribution &&value_dist,
```
pratikvn, line 62: Again, if you move this to `core/test/utils.hpp` then you can remove this duplication.

hip/test/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,362 @@
...
+        d_unfinished_agg = unfinished_agg;
+        d_strongest_neighbor = strongest_neighbor;
+        d_coarse_vector->copy_from(coarse_vector.get());
+        d_fine_vector->copy_from(fine_vector.get());
+        d_weight_csr->copy_from(weight_csr.get());
+        d_weight_diag->copy_from(weight_diag.get());
+        d_system_mtx->copy_from(system_mtx.get());
+    }
+
+    void make_symetric(Mtx *mtx)
```
pratikvn, line 157:
```suggestion
    void make_symmetric(Mtx *mtx)
```

hip/test/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,362 @@
...
+    {
+        for (int i = 0; i < mtx->get_size()[0]; ++i) {
+            for (int j = i + 1; j < mtx->get_size()[1]; ++j) {
+                mtx->at(i, j) = mtx->at(j, i);
+            }
+        }
+    }
+
+    // only for real value
+    void make_absoulte(Mtx *mtx)
```
pratikvn, line 167:
```suggestion
    void make_absolute(Mtx *mtx)
```

hip/test/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,362 @@
...
+            for (int j = 0; j < mtx->get_size()[1]; ++j) {
+                sum += abs(mtx->at(i, j));
+            }
+            mtx->at(i, i) = sum;
+        }
+    }
+
+    void make_spd(Mtx *mtx)
+    {
+        make_symetric(mtx);
```
pratikvn, line 190:
```suggestion
        make_symmetric(mtx);
```

hip/test/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,362 @@
...
+        make_diag_dominant(mtx);
+    }
+
+    void make_weight(Mtx *mtx)
+    {
+        make_symetric(mtx);
+        make_absoulte(mtx);
```
pratikvn, line 197:
```suggestion
        make_symmetric(mtx);
        make_absolute(mtx);
```

omp/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -62,22 +68,62 @@ namespace amgx_pgm {
...
 template <typename IndexType>
 void renumber(std::shared_ptr<const OmpExecutor> exec, Array<IndexType> &agg,
-              IndexType *num_agg) GKO_NOT_IMPLEMENTED;
+              IndexType *num_agg)
+{
+    const auto num = agg.get_num_elems();
+    Array<IndexType> agg_map(exec, num + 1);
+    auto agg_vals = agg.get_data();
+    auto agg_map_vals = agg_map.get_data();
+    components::fill_array(exec, agg_map_vals, num + 1, zero<IndexType>());
```
pratikvn, line 115:
```suggestion
    agg_map_vals.fill(zero<IndexType>());
```

pratikvn (approved):
LGTM! Only some minor things.

omp/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -62,22 +68,62 @@ namespace amgx_pgm {
...
+    auto strongest_neighbor_vals = strongest_neighbor.get_const_data();
+#pragma omp parallel for
+    for (size_type i = 0; i < agg.get_num_elems(); i++) {
+        if (agg_vals[i] == -1) {
+            size_type neighbor = strongest_neighbor_vals[i];
+            if (neighbor != -1 && strongest_neighbor_vals[neighbor] == i) {
+                // Use the smaller index as agg point
+                auto group = min(i, neighbor);
+                agg_vals[i] = group;
+                agg_vals[neighbor] = group;
```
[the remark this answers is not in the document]
yhmtsai: it should be okay on the results aspect.
they always set the smaller index, so they should write the same result from two points.
I can do it with condition but I need to check what impact is.

hip/test/multigrid/amgx_pgm_kernels.cpp
yhmtsai: when it is false, the dimension may be different. Thus, I can only test whether it is valid in AssignToExistAggUnderteminsticIsEquivalentToRef

common/multigrid/amgx_pgm_kernels.hpp.inc
yhmtsai: Yes, it is the same. I forgot to delete it

cuda/multigrid/amgx_pgm_kernels.cu
yhmtsai: Yes, it is usually the unsorted case.

omp/multigrid/amgx_pgm_kernels.cpp
yhmtsai: I do not notice we add it in array header.
However, is it acceptable? Doesn't it lead the circular dependence?
I think we can use array in device folder because Array is header-only class.

tcojean: Good point, this will not work and create a circular dependency issue since `Array::fill` is implemented in the `core` module and not in the header.

omp/multigrid/amgx_pgm_kernels.cpp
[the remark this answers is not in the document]
tcojean: It depends entirely on what the neighbors set look like. If it's symmetric (neighbor[i] = j and neighbor[j] = i), then there should be no issue. But if you have cases where neighbor[i] = j, neighbor[j] = k and neighbor[k] = i, then you could get different results depending on what gets executed first, I think?

omp/multigrid/amgx_pgm_kernels.cpp
pratikvn: Yes, you are right. Forget about these fill changes that I asked for.

yhmtsai: Should I remove the fill member function out of array in this pr?

yhmtsai:
format!

omp/multigrid/amgx_pgm_kernels.cpp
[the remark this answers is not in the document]
yhmtsai: This case you mentioned does not work in the previous one, either.
Those points with circular strong neighbor will be assigned in the last step.
The original one: the both side of strongest connection will set the smaller index as aggregation group.
Current one: only the smaller one proceed settin the aggregation of both side.

upsj (changes requested):
First part of my review, there are a few questions and suggestions.

I still need to take a closer look at the GPU implementation of the coarse matrix computation

common/multigrid/amgx_pgm_kernels.hpp.inc
```diff
@@ -0,0 +1,327 @@
...
+    }
+    if (strongest_agg != -1) {
+        agg_val[row] = agg_const_val[strongest_agg];
+    } else {
+        agg_val[row] = row;
+    }
+}
+
+// This is the undeterminstic implementation which is the same implementation of
+// the previous one but agg_val == agg_const_val.
```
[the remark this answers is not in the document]
yhmtsai: Yes, it could be, but it will lose a little possibility such that compiler to optimize it

omp/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -108,8 +241,50 @@ template <typename ValueType, typename IndexType>
...
+                       matrix::Csr<ValueType, IndexType> *temp)
+{
+    // agg[i] -> I, agg[j] -> J
+    const auto coarse_nrows = coarse->get_size()[0];
+    const auto source_nrows = source->get_size()[0];
+    const auto source_row_ptrs = source->get_const_row_ptrs();
+    const auto source_col_idxs = source->get_const_col_idxs();
+    const auto source_vals = source->get_const_values();
+    vector<map<IndexType, ValueType>> row_list(
+        source_nrows, map<IndexType, ValueType>{exec}, exec);
```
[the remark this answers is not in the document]
yhmtsai: It should be almost the reference implementation.
I do not spend too much time for optimizing them (including CUDA generation)
the column index is not sorted and maybe repeated

omp/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -87,18 +133,105 @@ void find_strongest_neighbor(
...
+{
+    const auto row_ptrs = weight_mtx->get_const_row_ptrs();
+    const auto col_idxs = weight_mtx->get_const_col_idxs();
+    const auto vals = weight_mtx->get_const_values();
+    const auto agg_const_val = agg.get_const_data();
+    auto agg_val = (intermediate_agg.get_num_elems() > 0)
+                       ? intermediate_agg.get_data()
+                       : agg.get_data();
+    const auto diag_vals = diag->get_const_values();
+    for (IndexType row = 0; row < agg.get_num_elems(); row++) {
```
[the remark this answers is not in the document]
yhmtsai: Thanks. I delete it for debugging but forget to recover it

yhmtsai:
@tcojean @pratikvn I finish collection and testing of the matrix/array related helper or generator.

hip/test/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,308 @@
...
+    auto d_x = d_unfinished_agg;
+    index_type num_agg;
+    index_type d_num_agg;
+
+    gko::kernels::reference::amgx_pgm::renumber(ref, agg, &num_agg);
+    gko::kernels::hip::amgx_pgm::renumber(hip, d_agg, &d_num_agg);
+
+    ASSERT_EQ(d_num_agg, num_agg);
+    GKO_ASSERT_ARRAY_EQ(d_agg, agg);
+    ASSERT_LE(num_agg, 300);
```
pratikvn, line 215:
```suggestion
    ASSERT_LE(num_agg, this->n);
```

cuda/test/solver/bicgstab_kernels.cpp
```diff
@@ -69,7 +69,7 @@ class Bicgstab : public ::testing::Test {
         cuda = gko::CudaExecutor::create(0, ref);
 
         mtx = gen_mtx(123, 123);
-        make_diag_dominant(mtx.get());
+        gko::test::make_diag_dominant(mtx.get());
```
pratikvn, line 72: We should remove this. BiCGSTAB should be able to solve general systems. Or atleast add another test with a non-diagonally dominant system.

omp/test/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,299 @@
...
+        d_fine_vector->copy_from(fine_vector.get());
+        d_weight_csr->copy_from(weight_csr.get());
+        d_weight_diag->copy_from(weight_diag.get());
+        d_system_mtx->copy_from(system_mtx.get());
+    }
+
+    void make_weight(Mtx *mtx)
+    {
+        gko::test::make_symmetric(mtx);
+        // it is only works for realvalue case.
```
pratikvn, line 133:
```suggestion
        // it is only works for real value case.
```

omp/test/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,299 @@
...
+    auto d_x = d_unfinished_agg;
+    index_type num_agg;
+    index_type d_num_agg;
+
+    gko::kernels::reference::amgx_pgm::renumber(ref, agg, &num_agg);
+    gko::kernels::omp::amgx_pgm::renumber(omp, d_agg, &d_num_agg);
+
+    ASSERT_EQ(d_num_agg, num_agg);
+    GKO_ASSERT_ARRAY_EQ(d_agg, agg);
+    ASSERT_LE(num_agg, 300);
```
pratikvn, line 206:
```suggestion
    ASSERT_LE(num_agg, this->n);
```

omp/test/solver/bicgstab_kernels.cpp
```diff
@@ -68,7 +68,7 @@ class Bicgstab : public ::testing::Test {
         omp = gko::OmpExecutor::create();
 
         mtx = gen_mtx(123, 123);
-        make_diag_dominant(mtx.get());
+        gko::test::make_diag_dominant(mtx.get());
```
pratikvn, line 71: Same as for hip/cuda

omp/test/solver/cgs_kernels.cpp
```diff
@@ -67,7 +67,7 @@ class Cgs : public ::testing::Test {
         omp = gko::OmpExecutor::create();
 
         mtx = gen_mtx(123, 123);
-        make_diag_dominant(mtx.get());
+        gko::test::make_diag_dominant(mtx.get());
```
pratikvn, line 70: Same as for hip/cuda

pratikvn (approved):
LGTM! Only some minor things.

core/test/utils/value_generator.hpp
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * @return ValueType
+ */
+template <typename ValueType, typename ValueDistribution, typename Engine>
+typename std::enable_if<!is_complex_s<ValueType>::value, ValueType>::type
+get_rand_value(ValueDistribution &&value_dist, Engine &&gen)
+{
+    return value_dist(gen);
+}
+
```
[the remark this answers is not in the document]
yhmtsai: it is the specialization for the above one, so I keep the one linebreak here

cuda/test/solver/bicgstab_kernels.cpp
yhmtsai: When I removing it, it gives different result...
The matrix might be singular when it is not (strongly) diag dominant, so it has several solution in this sense?

pratikvn:
I guess this needs to only be rebased ?

common/multigrid/amgx_pgm_kernels.hpp.inc
```diff
@@ -0,0 +1,327 @@
...
+    const IndexType *__restrict__ result_row_ptrs,
+    IndexType *__restrict__ result_col_idxs,
+    ValueType *__restrict__ result_values)
+{
+    auto row = thread::get_thread_id_flat();
+    if (row >= source_nrows) {
+        return;
+    }
+    auto result_i = result_row_ptrs[agg_val[row]] + row_map[row];
+    for (auto i = source_row_ptrs[row]; i < source_row_ptrs[row + 1];
```
upsj, line 258: small nit: could you move result_i++ to the loop body? Otherwise this might be a bit hidden/unexpected to the user, as below with copy_to_coarse_kernel

omp/multigrid/amgx_pgm_kernels.cpp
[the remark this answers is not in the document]
upsj: I think especially considering the weaker memory model for C++ on ARM processors, we should give this some thought. There is a potential RAW race condition between the write in  `agg_vals[neighbor] = i;` and the read in `agg_vals[i] == -1`. This might be especially problematic if the reads and writes are torn, i.e. implemented as two distinct memory transactions, and potentially even reordered.
If any write to `agg_vals[neighbor]` become visible to the thread where `i == neighbor`, the comparison will become false. I would argue though that this cannot cause any issues, since in the writing thread, `i < neighbor`, so the reading thread would have `i > neighbor` and thus would not write to memory itself. Does that sound right to you?

This will definitely produce ThreadSanitizer warnings, though, not sure how to silence those.

upsj:
@yhmtsai I think I figured out why this was confusing to me, and I believe we can provide a better framework by slightly reordering the operations:
Currently you proceed with
1. computing the coarse matrix
2. putting the aggregation information into the prolongation matrix (which has exactly one non-zero per row in AMGX)
3. transposing the prolongation matrix to get the restriction matrix
I think there is more potential for reuse of existing code if you compute the coarse matrix only after setting up both restriction and prolongation matrix. Then, you can directly express your operations as a special type of three-way SpGEMM: `P^T * A * P`.
Especially having the restriction matrix available here might help a lot, since it gives you easy access to the mapping `coarse index -> fine indices`. Does that make sense?

omp/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -62,22 +69,62 @@ namespace amgx_pgm {
...
+              IndexType *num_agg)
+{
+    const auto num = agg.get_num_elems();
+    Array<IndexType> agg_map(exec, num + 1);
+    auto agg_vals = agg.get_data();
+    auto agg_map_vals = agg_map.get_data();
+    components::fill_array(exec, agg_map_vals, num + 1, zero<IndexType>());
+#pragma omp parallel for
+    for (size_type i = 0; i < num; i++) {
+        agg_map_vals[agg_vals[i]] = 1;
```
upsj, line 119: Do I understand correctly that for each (un-remapped) coarse index `i`, `agg_vals[i] == i` holds? This kind of reminds me of a disjoint set/union find data structure. In case that's true, do you think it would make sense conceptually to make this write conditional, then? That would eliminate all duplicate writes
```suggestion
        if (agg_vals[i] == i) {
            agg_map_vals[agg_vals[i]] = 1;
        }
```

yhmtsai: I think so, `agg_vals[i] == i` should hold in algorithm.
It really a good idea!
agg_map_val is IndexType, so I can also use masking way to do it `(agg_val[i] == i) * 1` to eliminate the branch.

upsj: Yes, I only proposed this solution, since it silences SonarQube, and the compiler should translate this to similar code.

yhmtsai:
It can be reviewed again

cuda/test/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -98,13 +98,35 @@ class AmgxPgm : public ::testing::Test {
...
+    {
+        auto agg_array = gen_array(num, 0, num_agg - 1);
+        auto agg_array_val = agg_array.get_data();
+        std::vector<index_type> select_agg(num);
+        std::iota(select_agg.begin(), select_agg.end(), 0);
+        // use the first num_agg item as the aggregated index.
+        std::shuffle(select_agg.begin(), select_agg.end(), rand_engine);
+        // the value of agg_array is the i-th of aggregate group
+        for (gko::size_type i = 0; i < num; i++) {
+            agg_array_val[i] = select_agg[agg_array_val[i]];
```
upsj, line 112: Couldn't it happen that this is also equal to `i`, thus breaking your assumption for `num_agg == n` below?

upsj (approved):
LGTM!

cuda/test/multigrid/amgx_pgm_kernels.cpp
yhmtsai: I think it should be fine.
the array contains at most num_agg distinct value.
later operation set the aggregated index such that the array contains exactly num_agg distinct value.
each group may has 1 ~ x elements

upsj: if select_agg has a fixed point `select_agg[i] == i` larger than num_agg, wouldn't this lead to `num_agg + 1` distinct values?

yhmtsai: if i > num_agg, select_agg[i] will not be used. agg_array has 0 ~ num-1 value, so it will use the first num item from select_agg

upsj: Oh yes sorry, I didn't realize that agg_array only contains values from 0 to num_agg - 1.

omp/multigrid/amgx_pgm_kernels.cpp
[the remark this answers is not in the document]
yhmtsai: Sorry for late reply.
Yes, read may be after or before the write from the smaller threads and it will not cause any issue as the reason you said.
I am not sure how ThreadSanitizer warning shows on this part.

Issue #179, opened 2018-12-10
Alternatives to raw pointers


gflegar:
@upsj and I discussed a bit today about Ginkgo's _lend_ semantics. The problem we currently have is that `lend` has to be spammed everywhere. For example:

```c++
y = mtx::create(...);
x->copy_from(lend(y));
// vs
x->copy_from(give(y));
// another example:
mtx->apply(lend(x), lend(y));
```

The question is whether we can write this as:

```c++
y = mtx::create(...);
x->copy_from(y);
// vs
x->copy_from(give(y));
// another example:
mtx->apply(x, y);
```

I.e. to have `lend` as the default behavior wherever possible.

### Option 1: use `std::unique_ptr` / `std::shared_ptr` for everything

The first idea discussed (considering only unique ownership) is to use `const std::unique_ptr<T> &`  instead of `T *`. Theoretically, this works: we just avoid using plain pointers, and use `std::unique_ptr` for everywhere. However, problems begin as soon as another ownership mode (e.g. `std::shared_ptr`) is added, as the number of overloads needed for each such function becomes at most **n<sup>k</sup>** (where the upper limit is reached for functions that do not depend on ownership), where **n** is the number of ownership modes, and **k** is the number of parameters carrying ownership information. A way to alleviate the issue slightly would be to use template parameters, but it only works for non-virtual functions, and may introduce quite a bit of issues with reference collapsing rules if the users are not extremely careful.

### Option 2: create a smart pointer type with ownership-less semantics

Something like this:

```c++
template <typename T>
class pointer {
public:
    pointer(std::shared_ptr<T> &p) : ptr{p.get()} {}
    pointer(std::unique_ptr<T> &p) : ptr{p.get()} {}
    pointer(T *p) : ptr{p} {}
    T *operator->() const { return ptr; };
    T *get() const { return ptr; };
private:
    T *ptr;
};
```

This would solve the problem of exponential number of overloads, but due to an implicit conversions from `std::shared_ptr<T>` and `std::unique_ptr<T>` to `pointer<T>`, users may misuse the pointer:

```c++
std::unique_ptr<T> func();

pointer<T> p = func();
p->whatever();    // segfault
```

Making clear that `pointer` should only be used as a parameter (e.g. by renaming it to `pointer_param<T>`) seems to fix the issue, but the following problem still remains:

```c++
std::unique_ptr<T> func();

pointer_param<T> modify(pointer_param<T> p) {
    p->whatever();    // ok
[...]

PR #698, opened 2021-01-30
Add heat equation example

upsj:
Adds an example showing the solution of a 2D heat equation with finite differences and implicit Euler integration using CG with IC preconditioner.


upsj:
Here's what the output looks like:
https://user-images.githubusercontent.com/1693511/106369097-f9887280-634e-11eb-9adb-f3e6ce4030fc.mp4

examples/heat-equation/heat-equation.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+    // Problem parameters:
+    // simulation length
+    auto t0 = 5.0;
+    // diffusion factor
+    auto diffusion = 0.0005;
+    // scaling factor for heat source
+    auto source_scale = 2.5;
+    // Simulation parameters:
+    // grid size for inner points
```
hartwiganzt, line 134: inner points per discretization direction

hartwiganzt (approved):
LGTM

examples/heat-equation/heat-equation.cpp
```diff
@@ -0,0 +1,209 @@
...
+            auto c_val = diffusion * tau * 4.0 / h2 + 1.0;
+            auto off_val = -diffusion * tau / h2;
+            // for each grid point: insert 5 stencil points
+            // with Dirichlet boundary conditions, i.e. with zero boundary value
+            // clang-format off
+            if (i > 0) mtx_data.nonzeros.emplace_back(c, c - n, off_val);
+            if (j > 0) mtx_data.nonzeros.emplace_back(c, c - 1, off_val);
+            mtx_data.nonzeros.emplace_back(c, c, c_val);
+            if (j < n - 1) mtx_data.nonzeros.emplace_back(c, c + 1, off_val);
+            if (i < n - 1) mtx_data.nonzeros.emplace_back(c, c + n, off_val);
```
yhmtsai, line 175: we have the brackets rule for control flow in [control-flow-constructs](https://github.com/ginkgo-project/ginkgo/wiki/Contributing-guidelines#control-flow-constructs).
Do you think we can allow something for readability?

examples/heat-equation/heat-equation.cpp
```diff
@@ -0,0 +1,209 @@
...
+        // if enough time has passed, output the next frame
+        if (t - last_t > 1.0 / fps) {
+            last_t = t;
+            std::cout << t << std::endl;
+            output_timestep(output, n, in_vector->get_const_values());
+        }
+        // add heat
+        in_vector->add_scaled(lend(tau_source_scalar), lend(source));
+        // execute Euler step
+        solver->apply(lend(in_vector), lend(out_vector));
```
yhmtsai, line 205:
```suggestion
        in_vector->add_scaled(gko::lend(tau_source_scalar), gko::lend(source));
        // execute Euler step
        solver->apply(gko::lend(in_vector), gko::lend(out_vector));
```

yhmtsai (approved):
LGTM. minor format changes might be required

pratikvn:
Is there some way we can add the video to the documentation/ website or both ? That would be nice. You could convert it into a gif with reduced frames as mp4 is probably unnecessarily heavy in this case.

examples/heat-equation/heat-equation.cpp
upsj: This was mainly meant for the poster, where vertical space seems to be more limited, iirc. I have no issue with reformatting this at all.

examples/heat-equation/heat-equation.cpp
upsj: I know we had this discussion before somewhere, but I don't really like this coding style, primarily because it complicates what it originally aimed to simplify/unify: This is more "noisy" than the inlined
```suggestion
        in_vector->add_scaled(tau_source_scalar.get(), source.get());
        // execute Euler step
        solver->apply(in_vector.get(), out_vector.get());
```
Maybe it would also make sense to revisit #179 again to work around this issue?

upsj:
The GIF is actually larger than the input video :-D 
heat

examples/heat-equation/doc/intro.dox
```diff
@@ -0,0 +1,39 @@
...
+$
+
+We then build an implicit Euler integrator by discretizing with time step $\tau$
+
+$
+    \frac{u_{i,j}^{k+1} - u_{i,j}^k}{\tau} =
+    \alpha\frac{u_{i-1,j}^{k+1}+u_{i+1,j}^{k+1}
+               -u_{i,j-1}^{k+1}-u_{i,j+1}^{k+1}+4u_{i,j}^{k+1}}{h^2}
+    +f_{i,j}^{k+1}
+$
```
nbeams, line 28: Is it worth adding somewhere that this example uses a constant-in-time source?  Denoting f_{i,j}^{k+1} isn't incorrect, of course, but it perhaps implies that f will be changing with time, which it is not.

nbeams (approved):

examples/heat-equation/heat-equation.cpp
upsj: I'll change this for now, but this is definitely something we can improve upon.

examples/heat-equation/doc/intro.dox
upsj: That's if course correct, thanks for pointing it out!

pratikvn (approved):
LGTM!

PR #702, opened 2021-02-06
Add an implicit residual norm criterion.

pratikvn:
This PR adds an implicit residual norm convergence criterion, which checks the reduction in the implicit residual calculated in a solver. Some libraries (example: MFEM) use an implicit residual as their main criterion. This criterion class will allow apples to apples comparison of our solvers to theirs. 

The idea is that for solvers such as CG or extensions of CG (BiCG, BiCGSTAB, CGS, FCG...) we can use scalars which are representative of the residual norm of the solution. For example, in CG, the scalar `rho` is always computed in each iteration and closely follows the error in the solution. 

This allows us to save computing the residual norm from the residual vector, which is what we do in most cases to check for convergence in the `ResidualNormReduction` stopping criterion. 

Of course, for some solvers, there is no implicit residual norm available, and for those solvers, this stopping criterion will throw a `NOT_SUPPORTED` exception.

### Edited

Updated modifications:

1. The `ResidualNorm` class now can perform all three types of criterion checks: `absolute`, `relative` and `reduction`.
2. ~~The previous classes are marked with the `[[deprecated("message")]]` attribute. Some byproducts of this are:
     i. Whenever this class is used a warning is thrown at compile time.
     ii. As we need to still test the older class functionality, we will always throw the warning during compilation.~~
    A deprecated note is added to the deprecated classes.
3. As we use `ResidualNormReduction` kernels everywhere, these are now replaced by `ResidualNorm`. The default behaviour of `ResidualNorm` should be the same as `ResidualNormReduction`, so the change in function name should not affect where `ResidualNormReduction` was previously called.
4. `ResidualNorm` and `ImplicitResidualNorm` use the same class structure and functions except that they call different `check_impl` kernels. In the future we might want to create a `ResidualNormBase` class and derive both from that. But I think currently that breaks interface, so I more or less duplicated the code.

Unrelated modifications:
1. `fill` with value functions were added for `Array` and `Dense`. I think these are really helpful to just fill the entire values arrays with a single value. This just uses the `fill_array` kernel that we previously had, so no new kernels are added.
2. In IDR, we were computing the residual norm in each iteration, but were not updating the stopping criterion with this information, which is now fixed.


hartwiganzt:
I think having this is very useful! However, I am unsure whether "implicit" is the best name, or whether "recurrence residual" is better.

pratikvn:
We always pass in the recurrent residual (the explicitly computed residual vector) to the stopping criterion. This is what we check in the relative residual norm reduction stopping criterion. Even though this is also a recurrent residual, I think "implicit residual" might be better suited because the residual norm is implicitly computed, so for example in case of preconditioned CG, the implicit residual norm would be `r'*P*r`, while the recurrent residual vector would be `r` and its norm `r'*r`

upsj:
Here I would note the distinction between *residual*, *residual norm* and *implicit residual norm*:
*residual* is exact modulo rounding errors (every solver)
*residual norm* is exact modulo rounding errors (GMRES)
*implicit residual norm* is only correlated with the exact residual

hartwiganzt (approved):
LGTM

core/solver/idr.cpp
```diff
@@ -178,6 +179,7 @@ void Idr<ValueType>::iterate(const LinOp *b, LinOp *x) const
         if (stop_criterion->update()
                 .num_iterations(total_iter)
                 .residual(residual.get())
+                .residual_norm(residual_norm.get())
```
Slaedr, line 182: Why is this better? You had to add an extra `compute_norm2` anyway.

hip/stop/residual_norm_kernels.hip.cpp
```diff
@@ -125,6 +125,86 @@ GKO_INSTANTIATE_FOR_EACH_NON_COMPLEX_VALUE_TYPE(
...
+        const ValueType *__restrict__ tau,
+        const ValueType *__restrict__ orig_tau, uint8 stoppingId,
+        bool setFinalized, stopping_status *__restrict__ stop_status,
+        bool *__restrict__ device_storage)
+{
+    const auto tidx = thread::get_thread_id_flat();
+    if (tidx < num_cols) {
+        if (sqrt(tau[tidx]) < rel_residual_goal * orig_tau[tidx]) {
+            stop_status[tidx].converge(stoppingId, setFinalized);
+            device_storage[1] = true;
```
Slaedr, line 154: If two RHS converge at the same time, there would a write conflict here, no? Or does it not matter since all threads are writing the same value, similar to below? Maybe use an atomic write here anyway - this is not performance-sensitive at all.

reference/test/solver/bicg_kernels.cpp
```diff
@@ -89,6 +89,16 @@ class Bicg : public ::testing::Test {
                           .with_reduction_factor(r<value_type>::value)
                           .on(exec))
                   .on(exec)),
+          bicg_factory_big2(
+              Solver::build()
+                  .with_criteria(
+                      gko::stop::Iteration::build().with_max_iters(100u).on(
+                          exec),
+                      gko::stop::ImplicitResidualNormReduction<
```
Slaedr, line 97: Wow, that is awful formatting. It's tolerable, I guess.

reference/test/stop/residual_norm_kernels.cpp
```diff
@@ -302,6 +302,134 @@ TYPED_TEST(RelativeResidualNorm, WaitsTillResidualGoalMultipleRHS)
...
+
+
+TYPED_TEST(ImplicitResidualNormReduction, CannotCreateCriterionWithoutB)
+{
+    ASSERT_THROW(this->factory_->generate(nullptr, nullptr, nullptr, nullptr),
+                 gko::NotSupported);
+}
+
+
+TYPED_TEST(ImplicitResidualNormReduction, CanCreateCriterionWithB)
```
Slaedr, line 344: For this test and the ones above: Do we need them for both reference and omp, or can we remove them from one of these two files?

Slaedr (approved):
LGTM, though I do have a few comments. Thanks!

Slaedr:
> Here I would note the distinction between _residual_, _residual norm_ and _implicit residual norm_:
> _(recurrent) residual_ is exact modulo rounding errors (every solver)
> _(recurrent) residual norm_ is exact modulo rounding errors (GMRES)
> _implicit residual norm_ is only correlated with the exact residual norm (CG, ...)

I would note that, only in the case of left preconditioned GMRES, the recurrent residual norm (from the Hessenberg least-squares RHS) is the exact *preconditioned* residual norm. So when convergence is detected based on it, the true residual would typically still not be converged.

core/solver/idr.cpp
pratikvn: No, that norm added is outside the loop. And that was so that a valid value is passed in the first iteration. In the loop, a residual_norm was already being computed in every iteration.

reference/test/stop/residual_norm_kernels.cpp
pratikvn: Both are calling different kernels, so I would prefer to have it similar to as it is now for the other stopping criteria.

reference/test/solver/bicg_kernels.cpp
pratikvn: Yeah, sometimes clang-format cannot do too much, but it tries to keep atleast the namespaces and the object name together.

include/ginkgo/core/stop/residual_norm.hpp
```diff
@@ -226,6 +226,84 @@ class RelativeResidualNorm : public ResidualNorm<ValueType> {
...
+ * stops the iteration process when the residual norm is below a certain
+ * threshold recurrent to the norm of the right-hand side, i.e. when
+ * norm(residual) / norm(right_hand_side) < threshold.
+ * For better performance, the checks are run thanks to kernels on
+ * the executor where the algorithm is executed.
+ *
+ * @note To use this stopping criterion there are some dependencies. The
+ * constructor depends on `b` in order to compute the norm of the
+ * right-hand side. If this is not correctly provided, an exception
+ * ::gko::NotSupported() is thrown.
```
nbeams, line 240: I don't understand what "recurrent to the norm of the right-hand side" means. 

I also think either this comment isn't correctly describing the class, or I am not understanding what the class is supposed to do (certainly possible!).  

My understanding is that what makes it an "implicit residual norm" class is that it's checking some value which is computed by the iterative solver -- the exact value being computed depends on the solver -- which we expect to tell us about the current residual, though it is not found by computing the norm of the residual itself.  For people familiar with other CG, GMRES, etc. implementations, this value is probably the one they're "used to" seeing as the value used for convergence checking.  Is this correct?  If so, I think the documentation should definitely mention this, so users realize this is likely the stopping criterion they want in order to replicate results they've gotten from other solvers.  

But why is it tied to the right hand side? The `ResidualNormReduction` class compares the residual to the initial residual, while the `RelativeResidualNorm` class compares to the rhs, which I feel makes this class's behavior confusing with respect to its name.  But also, can't we have both versions for the `ImplicitResidualNorm`?  One that compares to rhs and one that compares to initial residual?

nbeams (commented):
Thanks @pratikvn, it seems this will be useful for the MFEM integration, but also more generally for all Ginkgo users in the future.

hip/stop/residual_norm_kernels.hip.cpp
pratikvn: I think atomic is not necessary here for the same reason as mentioned below. I think adding an atomic write here might be a bit more confusing and I feel as it is now is a bit more clear.

include/ginkgo/core/stop/residual_norm.hpp
pratikvn: Yeah, I can do that. I have updated the factory to now take an additional parameter to select the comparison norm. I also updated the doc to be more clear.

nbeams: Cool, thanks!

hip/test/stop/residual_norm_kernels.cpp
```diff
@@ -259,6 +259,114 @@ TEST_F(RelativeResidualNorm, WaitsTillResidualGoalMultipleRHS)
...
+    gko::Array<gko::stopping_status> stop_status(ref_, 1);
+    stop_status.get_data()[0].reset();
+    stop_status.set_executor(hip_);
+
+    ASSERT_FALSE(
+        criterion->update()
+            .implicit_sq_residual_norm(d_res.get())
+            .check(RelativeStoppingId, true, &stop_status, &one_changed));
+
+    res->at(0) = std::pow(tol * 1.1e+1, 2);
```
thoasm, line 302: nit: again, using `rhs_norm` would make it easier to understand.

include/ginkgo/core/stop/residual_norm.hpp
```diff
@@ -226,6 +226,122 @@ class RelativeResidualNorm : public ResidualNorm<ValueType> {
...
+         * Implicit residual norm reduction factor
+         */
+        remove_complex<ValueType> GKO_FACTORY_PARAMETER_SCALAR(
+            reduction_factor, static_cast<remove_complex<ValueType>>(1e-15));
+
+        /**
+         * The quantity the reduction is relative to. Choices include "rhs" and
+         * "initial_residual"
+         */
+        std::string GKO_FACTORY_PARAMETER_SCALAR(relative_to, "rhs");
```
thoasm, line 264: Why do you use a string for that? I would personally prefer an Enum because it is easier and more efficient to compare while also having compile-time checks if the value is valid (if not, it simply fails to compile). With a string, it is easy to mistype a name and get a runtime error, which is a lot harder to debug (especially since the only information you get is `NOT_SUPPORTED` where you need to read the file to understand where the error is).
If you only ever want to have `rhs` and `initial_residual` as options, a `bool` value might be better (`relative_to_initial_residual` might be a good name, which you already use internally).

reference/test/stop/residual_norm_kernels.cpp
```diff
@@ -302,6 +302,163 @@ TYPED_TEST(RelativeResidualNorm, WaitsTillResidualGoalMultipleRHS)
...
+{
+    using Mtx = typename TestFixture::Mtx;
+    std::shared_ptr<gko::LinOp> scalar =
+        gko::initialize<Mtx>({1.0}, this->exec_);
+    auto criterion =
+        this->factory_->generate(nullptr, scalar, nullptr, nullptr);
+    ASSERT_NE(criterion, nullptr);
```
thoasm, line 370: Nit:
```suggestion
    auto criterion =
        this->factory_->generate(nullptr, scalar, nullptr, nullptr);

    ASSERT_NE(criterion, nullptr);
```

reference/test/stop/residual_norm_kernels.cpp
```diff
@@ -302,6 +302,163 @@ TYPED_TEST(RelativeResidualNorm, WaitsTillResidualGoalMultipleRHS)
...
+}
+
+
+TYPED_TEST(ImplicitResidualNormReduction, CanCreateCriterionWithInitialRes)
+{
+    using Mtx = typename TestFixture::Mtx;
+    auto initial_res = gko::initialize<Mtx>({100.0}, this->exec_);
+    auto criterion = this->factory_2_->generate(nullptr, nullptr, nullptr,
+                                                initial_res.get());
+    ASSERT_NE(criterion, nullptr);
```
thoasm, line 380: nit:
```suggestion
    auto criterion = this->factory_2_->generate(nullptr, nullptr, nullptr,
                                                initial_res.get());

    ASSERT_NE(criterion, nullptr);
```

reference/test/stop/residual_norm_kernels.cpp
```diff
@@ -302,6 +302,163 @@ TYPED_TEST(RelativeResidualNorm, WaitsTillResidualGoalMultipleRHS)
...
+    constexpr gko::uint8 RelativeStoppingId{1};
+    gko::Array<gko::stopping_status> stop_status(this->exec_, 1);
+    stop_status.get_data()[0].reset();
+
+    ASSERT_FALSE(
+        criterion->update()
+            .implicit_sq_residual_norm(res_norm.get())
+            .check(RelativeStoppingId, true, &stop_status, &one_changed));
+
+    res_norm->at(0) = std::pow(r<TypeParam>::value * 1.1e+1, 2);
```
thoasm, line 403: nit: same here again. I would prefer a varialbe `rhs_norm` and have it be multiplied by `1.1 * tol` instead of hard-coding the value to understand the code better.

thoasm (commented):
Nice work!

My biggest concern is the usage of a string instead of an enum or boolean to decide between the initial guess or the right-hand side as the reference. Otherwise, I only have minor nits.

core/solver/bicgstab.cpp
```diff
@@ -140,16 +142,18 @@ void Bicgstab<ValueType>::apply_impl(const LinOp *b, LinOp *x) const
         ++iter;
         this->template log<log::Logger::iteration_complete>(this, iter, r.get(),
                                                             dense_x);
+        rr->compute_dot(r.get(), rho.get());
+        rho_abs = rho->compute_absolute();
```
upsj, line 146: Can we take care of computing the absolute in-place in the stopping criteria? This would eliminate the need for additional alloctions for the absolute vector.

core/solver/cg.cpp
```diff
@@ -138,6 +138,7 @@ void Cg<ValueType>::apply_impl(const LinOp *b, LinOp *x) const
         if (stop_criterion->update()
                 .num_iterations(iter)
                 .residual(r.get())
+                .implicit_sq_residual_norm(rho->compute_absolute().get())
```
upsj, line 141: same here

core/solver/cgs.cpp
```diff
@@ -173,6 +173,7 @@ void Cgs<ValueType>::apply_impl(const LinOp *b, LinOp *x) const
         if (stop_criterion->update()
                 .num_iterations(iter)
                 .residual(r.get())
+                .implicit_sq_residual_norm(rho->compute_absolute().get())
```
upsj, line 176: same here

core/solver/fcg.cpp
```diff
@@ -141,6 +141,7 @@ void Fcg<ValueType>::apply_impl(const LinOp *b, LinOp *x) const
         if (stop_criterion->update()
                 .num_iterations(iter)
                 .residual(r.get())
+                .implicit_sq_residual_norm(rho->compute_absolute().get())
```
upsj, line 144: same here

core/stop/residual_norm_kernels.hpp
```diff
@@ -63,12 +63,39 @@ namespace residual_norm {
 }  // namespace residual_norm
 
 
+namespace implicit_residual_norm {
```
upsj, line 66: nit: This namespace is unnecessary, as macros don't have namespaces. Might still keep it around to make it easier to understand which namespaces the kernels live in?

cuda/stop/residual_norm_kernels.cu
```diff
@@ -121,6 +121,85 @@ GKO_INSTANTIATE_FOR_EACH_NON_COMPLEX_VALUE_TYPE(
...
+    __launch_bounds__(default_block_size) void implicit_residual_norm_kernel(
+        size_type num_cols, ValueType rel_residual_goal,
+        const ValueType *__restrict__ tau,
+        const ValueType *__restrict__ orig_tau, uint8 stoppingId,
+        bool setFinalized, stopping_status *__restrict__ stop_status,
+        bool *__restrict__ device_storage)
+{
+    const auto tidx = thread::get_thread_id_flat();
+    if (tidx < num_cols) {
+        if (sqrt(tau[tidx]) < rel_residual_goal * orig_tau[tidx]) {
```
upsj, line 148:
```suggestion
        if (sqrt(abs(tau[tidx])) < rel_residual_goal * orig_tau[tidx]) {
```

hip/stop/residual_norm_kernels.hip.cpp
```diff
@@ -125,6 +125,86 @@ GKO_INSTANTIATE_FOR_EACH_NON_COMPLEX_VALUE_TYPE(
...
+    __launch_bounds__(default_block_size) void implicit_residual_norm_kernel(
+        size_type num_cols, ValueType rel_residual_goal,
+        const ValueType *__restrict__ tau,
+        const ValueType *__restrict__ orig_tau, uint8 stoppingId,
+        bool setFinalized, stopping_status *__restrict__ stop_status,
+        bool *__restrict__ device_storage)
+{
+    const auto tidx = thread::get_thread_id_flat();
+    if (tidx < num_cols) {
+        if (sqrt(tau[tidx]) < rel_residual_goal * orig_tau[tidx]) {
```
upsj, line 152:
```suggestion
        if (sqrt(abs(tau[tidx])) < rel_residual_goal * orig_tau[tidx]) {
```

hip/stop/residual_norm_kernels.hip.cpp
upsj: Only cuda-memcheck/ThreadSanitizer complaining, but that should not be an issue.

omp/stop/residual_norm_kernels.cpp
```diff
@@ -88,6 +88,54 @@ GKO_INSTANTIATE_FOR_EACH_NON_COMPLEX_VALUE_TYPE(
...
+                            Array<stopping_status> *stop_status,
+                            Array<bool> *device_storage, bool *all_converged,
+                            bool *one_changed)
+{
+    static_assert(is_complex_s<ValueType>::value == false,
+                  "ValueType must not be complex in this function!");
+    bool local_one_changed = false;
+#pragma omp parallel for reduction(|| : local_one_changed)
+    for (size_type i = 0; i < tau->get_size()[1]; ++i) {
+        if (sqrt(tau->at(i)) < rel_residual_goal * orig_tau->at(i)) {
```
upsj, line 116:
```suggestion
        if (sqrt(abs(tau->at(i))) < rel_residual_goal * orig_tau->at(i)) {
```

reference/stop/residual_norm_kernels.cpp
```diff
@@ -85,6 +85,49 @@ GKO_INSTANTIATE_FOR_EACH_NON_COMPLEX_VALUE_TYPE(
...
+                            Array<stopping_status> *stop_status,
+                            Array<bool> *device_storage, bool *all_converged,
+                            bool *one_changed)
+{
+    static_assert(is_complex_s<ValueType>::value == false,
+                  "ValueType must not be complex in this function!");
+    *all_converged = true;
+    *one_changed = false;
+    for (size_type i = 0; i < tau->get_size()[1]; ++i) {
+        if (sqrt(tau->at(i)) < rel_residual_goal * orig_tau->at(i)) {
```
upsj, line 113:
```suggestion
        if (sqrt(abs(tau->at(i))) < rel_residual_goal * orig_tau->at(i)) {
```

upsj (dismissed):
LGTM in general, great idea! I would prefer if we could remove the compute_absolute calls, since they allocate memory in each allocation. Another question would be: Do we want to report these implicit residuals to the loggers? Finally, in your stopping criterion, you currently represent relative residuals and residual reduction. What do you think about making this consistent with the regular residual stopping criteria and adding 3 separate stopping criteria based on the baseline value (absolute, relative, residual reduction)?

core/solver/bicgstab.cpp
thoasm: Good point, would be nice to have that in the stopping criterion.

core/stop/residual_norm_kernels.hpp
pratikvn: Yes, for clarity, I think it is better to have the namespace.

core/solver/bicgstab.cpp
pratikvn: Good catch. I have updated this now.

include/ginkgo/core/stop/residual_norm.hpp
pratikvn: Good point. I use an `enum` now.

pratikvn:
Regarding reporting the implicit residuals to the loggers is a good point and I did have a look at it, but I think that breaks our interface as we cannot add another parameter to the `Updater` class.

Regarding adding three separate criterion classes, I think that would be a lot of duplication and unnecessary code. But if everyone feels that they would prefer to have separate classes, I can do that.

pratikvn:
format!

pratikvn:
Some new modifications:

1. The `ResidualNorm` class now can perform all three types of criterion checks: `absolute`, `relative` and `reduction`.
2. The previous classes are marked with the `[[deprecated("message")]]` attribute. Some byproducts of this are:
     i. Whenever this class is used a warning is thrown at compile time.
     ii. As we need to still test the older class functionality, we will always throw the warning during compilation.
3. As we use `ResidualNormReduction` kernels everywhere, these are now replaced by `ResidualNorm`. The default behaviour of `ResidualNorm` should be the same as `ResidualNormReduction`, so the change in function name should not affect where `ResidualNormReduction` was previously called.
4. `ResidualNorm` and `ImplicitResidualNorm` use the same class structure and functions except that they call different `check_impl` kernels. In the future we might want to create a `ResidualNormBase` class and derive both from that. But I think currently that breaks interface, so I more or less duplicated the code.

Unrelated modifications:
1. `fill` with value functions were added for `Array` and `Dense`. I think these are really helpful to just fill the entire values arrays with a single value. This just uses the `fill_array` kernel that we previously had, so no new kernels are added.

include/ginkgo/core/stop/residual_norm.hpp
```diff
@@ -77,17 +100,200 @@ class ResidualNorm
...
+          device_storage_{factory->get_executor(), 2},
+          parameters_{factory->get_parameters()},
+          reduction_factor_{factory->get_parameters().reduction_factor}
+    {
+        this->baseline_rhs_norm_ =
+            factory->get_parameters().baseline == mode::rhs_norm;
+        this->baseline_init_resnorm_ =
+            factory->get_parameters().baseline == mode::initial_resnorm;
+        this->baseline_absolute_ =
+            factory->get_parameters().baseline == mode::absolute;
```
thoasm, line 121: I don't understand why you still use these. Is it not easier to store the `enum` do a `switch` `case` over the options? This would also prevent the future error of having multiple of these `bool` set to `true`.

include/ginkgo/core/stop/residual_norm.hpp
```diff
@@ -77,17 +100,200 @@ class ResidualNorm
...
+    }
+
+    std::unique_ptr<NormVector> starting_tau_{};
+    std::unique_ptr<NormVector> u_dense_tau_{};
+
+private:
+    remove_complex<ValueType> reduction_factor_{};
+    bool baseline_rhs_norm_{};
+    bool baseline_init_resnorm_{};
+    bool baseline_absolute_{};
```
thoasm, line 171:
```suggestion
    mode baseline_{};
```

include/ginkgo/core/stop/residual_norm.hpp
```diff
@@ -110,7 +316,11 @@ class ResidualNorm
  * @ingroup stop
  */
 template <typename ValueType = default_precision>
-class ResidualNormReduction : public ResidualNorm<ValueType> {
```
thoasm, line 113: Why are you publicly inheriting from `ResidualNorm`? You are exposing `ResidualNorm::Factory` and `ResidualNorm::build` to the public interface, which should not happen.
I am currently doing some tests with it and will come back to you if this is indeed a big issue or not (and how to fix it).

include/ginkgo/core/stop/residual_norm.hpp
```diff
@@ -77,17 +100,200 @@ class ResidualNorm
...
+
     std::unique_ptr<NormVector> starting_tau_{};
     std::unique_ptr<NormVector> u_dense_tau_{};
 
 private:
-    remove_complex<ValueType> tolerance_{};
+    remove_complex<ValueType> reduction_factor_{};
+    bool baseline_rhs_norm_{};
+    bool baseline_init_resnorm_{};
+    bool baseline_absolute_{};
```
thoasm, line 296:
```suggestion
     mode baseline_{};
```

tcojean:
About reporting to the logger the implicit residual, I think there should be ways to do this without breaking interface? Either by adding an extra overload with the information while keeping the old one, or adding a new parameter with a default value so that the old function calls still works. Since it's only  about adding something new and not removing or changing something, there should be a way to make it work?

reference/test/stop/residual_norm_kernels.cpp
```diff
@@ -50,54 +50,324 @@ namespace {
...
     using NormVector = gko::matrix::Dense<gko::remove_complex<T>>;
 
-    ResidualNormReduction()
+    ResidualNorm()
     {
         exec_ = gko::ReferenceExecutor::create();
-        factory_ = gko::stop::ResidualNormReduction<T>::build()
+        factory_ = gko::stop::ResidualNorm<T>::build()
                        .with_reduction_factor(r<T>::value)
                        .on(exec_);
```
thoasm, line 63: Could you rename this factory, so it is clear that it is using `mode::rhs_norm`? I was a bit confused in the beginning.

omp/test/stop/residual_norm_kernels.cpp
```diff
@@ -45,6 +45,249 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+protected:
+    using Mtx = gko::matrix::Dense<T>;
+    using NormVector = gko::matrix::Dense<gko::remove_complex<T>>;
+
+    ResidualNorm()
+    {
+        exec_ = gko::OmpExecutor::create();
+        factory_ = gko::stop::ResidualNorm<T>::build()
+                       .with_reduction_factor(r<T>::value)
+                       .on(exec_);
```
thoasm, line 59: Same here: Renaming `factory_` to something that indicates that `mode::rhs_norm` is used.

thoasm (changes requested):
LGTM except for the public inheritance of `ResidualNorm`, which now has a `Factory`, which would get exposed for `ResidualNormReduction`, `RelativeNormReduction` and `AbsoluteNormReduction` (even though they are deprecated). I will experiment with this now an come back to you soon with more insight and hopefully a solution.

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -244,6 +244,13 @@ class Dense
     std::unique_ptr<LinOp> row_permute(
         const Array<int64> *permutation_indices) const override;
 
+    /**
+     * Fill the dense matrix with a given value.
+     *
+     * @param value  the value to be filled
+     */
+    void fill(const ValueType value) { this->values_.fill(value); }
```
thoasm, line 252: This **ONLY** works if no padding is used (`stride == size[1]`), which we can't expect. If you want to keep this functionality, you need to add a test with a stride and add an additional kernel for each executor respecting the stride.

thoasm (changes requested):
`Dense::fill` function does not work when a `stride != size[1]` is used!

thoasm:
@pratikvn I played around with it a bit, and I think I found a good solution: I created a new branch (I did not want to push into yours, and that way, you can decide what to keep) named [recur-res-stop-crit_improve](https://github.com/ginkgo-project/ginkgo/tree/recur-res-stop-crit_improve). Here are the changes (actually the commit itself) I did: https://github.com/ginkgo-project/ginkgo/commit/d749e01e2066c3002d323e5df94f38a3e08e6b4c

I tried to remove as much code duplication as possible while keeping the interface. Debatable is my solution for `ImplicitResidualNorm`: I inherit from `ResidualNormBase` and overwrite the `check_impl` to have the creation of the initial vectors depending on `baseline` from `ResidualNormBase` while having the proper `check_impl`.

It does not follow the Inheritance IS-A relationship (especially for `ResidualNormBase -> ImplicitResidualNorm`), but I think we avoid the most code duplication that way. It is possible to have an abstract class, which does not implement `check_impl` and therefore following IS-A inheritance principle, but that would force all deprecated `*ResidualNorm` classes to overwrite `check_impl`, creating more duplicated code.

Note: please also look at the comments I wrote, I am not 100% sure that they are all correct.

include/ginkgo/core/matrix/dense.hpp
pratikvn: good catch. i will fix this

pratikvn:
@thoasm thank you for the updates and the fixes. I think your solution is good. I think the fact that we dont have to duplicate `check_impl` for the deprecated criteria is good. 

I also added the strided kernels, which I had missed. I think this should be ready for reviewing again.

pratikvn:
@tcojean , regarding the logger. Maybe I misunderstand your comment. As I see it, I need to update the `criterion_check_started` to also take the implicit_residual_norm from the updater, which is passed to the logger in `include/.../stop/criterion.hpp` line 157. I think I cannot add another `GKO_REGISTER_EVENT` with `criterion_check_started` as that sets some variables with the same name so overloading is not possible. Adding something like `criterion_check_started_new` is possible, I guess without breaking interface, but that again will be in the public interface and will cause naming issues in the future. I would prefer to modify the `criterion_check_started` and break the interface for good in 2.0.

thoasm (approved):
LGTM!

tcojean:
@pratikvn What I mean is that maybe it works if you add the implicit residual at the end and use an ` = nullptr` default case, both for the logger event and the criterion updater? Both the current and new interface would work. Of course, that's maybe not optimal in terms of positioning etc.

upsj (commented):
LGTM mostly, I have two ideas how we could improve and future-proof this approach:

1. logging: You can add another parameter to a logger without breaking interface as follows:

```
    GKO_LOGGER_REGISTER_EVENT(21, iteration_complete, ..., const LinOp *implicit_tau_sq = nullptr)
protected:                                                           \
    virtual void on_iteration_complete(const LinOp *solver, const size_type &it,
            const LinOp *r, const LinOp *x, const LinOp *tau) const {
        this->on_iteration_complete(solver, it, r, x, tau, nullptr);
    }
```
2. stopping criterion generation: We could avoid the with_baseline factory parameter call if we had more speaking with_tolerance functions:
```
Factory& with_reduction_factor(double) { baseline = initial_resnorm; ... }
Factory& with_relative_residual(double) { baseline = rhs_norm; ... }
Factory& with_absolute_residual(double) { baseline = absolute; ... }
```
and adding another value `none` to the baseline enum. When we try to generate a residual norm on that enum value, we throw an exception: "Baseline not specified" or something like that.

cuda/test/utils.hpp
```diff
@@ -51,4 +51,8 @@ auto no_reset_exec =
 }  // namespace
 
 
+template <typename T>
+using I = std::initializer_list<T>;
+
+
```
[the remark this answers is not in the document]
pratikvn: Yes, but we dont include `core/test/utils.hpp` in cuda tests

reference/test/matrix/dense_kernels.cpp
```diff
@@ -109,6 +109,51 @@ class Dense : public ::testing::Test {
...
+    m->fill(value_type{42});
+
+    ASSERT_EQ(m->get_size(), gko::dim<2>(3, 2));
+    EXPECT_EQ(m->get_num_stored_elements(), 12);
+    EXPECT_EQ(m->at(0), value_type{42.0});
+    EXPECT_EQ(m->at(1), value_type{42.0});
+    EXPECT_EQ(m->at(2), value_type{42.0});
+    ASSERT_EQ(m->at(3), value_type{42.0});
+    EXPECT_EQ(m->at(4), value_type{42.0});
+    EXPECT_EQ(m->at(5), value_type{42.0});
```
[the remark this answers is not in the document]
pratikvn: I do check it with `in_stride`, making sure that it is the same as before the fill was called.

core/solver/bicgstab.cpp
```diff
@@ -169,6 +171,7 @@ void Bicgstab<ValueType>::apply_impl(const LinOp *b, LinOp *x) const
             stop_criterion->update()
                 .num_iterations(iter)
                 .residual(s.get())
+                .implicit_sq_residual_norm(rho.get())
```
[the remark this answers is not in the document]
pratikvn: good point

pratikvn:
@upsj , Regarding the logger, I guess @tcojean was also talking about `iteration_complete`. I completely missed that. I was thinking of `criterion_check_x`. I will update that.

Regarding the factory parameters, I prefer the `with_baseline` than separate parameters for the different baselines.

reference/test/matrix/dense_kernels.cpp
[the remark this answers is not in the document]
upsj: Okay, I'm blind, thanks 🤦

cuda/test/utils.hpp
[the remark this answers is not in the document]
upsj: I think `cuda/test/utils.hpp` includes that file as well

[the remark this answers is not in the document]
tcojean: There is nothing which prevents you to, some like `idr_kernels` include it. Also, it seems `cuda/test/utils.hpp` includes `core/test/utils.hpp` (line 37).

pratikvn:
Actually, having another look at the `iteration_complete` and that it is a public function which is overridden in other loggers, adding parameters, actually breaks interface, I think.

cuda/test/utils.hpp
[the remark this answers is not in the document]
pratikvn: Yes, you are right. I am not sure why I was getting an error then. I will fix and remove this. Thanks.

upsj:
@pratikvn That's what the second on_iteration_complete overload is for. Though I actually got it the wrong way round, this current implementation might cause iteration events to get lost on old loggers. The correct solution would be
```
    GKO_LOGGER_REGISTER_EVENT(21, iteration_complete, ...)
protected:                                                           \
    virtual void on_iteration_complete(const LinOp *solver, const size_type &it,
            const LinOp *r, const LinOp *x, const LinOp *tau, const LinOp *implicit_tau_sq = nullptr) const {
        this->on_iteration_complete(solver, it, r, x, tau);
    }
```
and overriding the new `on_iteration_complete` where necessary

pratikvn:
@upsj , I think the problem is that the call of `this->on_iteration_complete(solver, it, r, x, tau)` inside the new `on_iteration_complete` is ambiguous because the compiler cannot resolve between the two when you call it with just non-specific parameters. Also I think any call to `on_iteration_complete` will be ambiguous unless all the parameters are specified.

upsj:
@pratikvn Oh yeah, when you drop the default value for the last parameter, the ambiguity goes away. You only need to make sure when overriding the new interface that you override both functions, one with your actual implementation, and the other one with 
```
void on_iteration_complete(const LinOp *solver, const size_type &it,
            const LinOp *r, const LinOp *x = nullptr, const LinOp *tau = nullptr) const {
        this->on_iteration_complete(solver, it, r, x, tau, nullptr);
    }
```

include/ginkgo/core/stop/residual_norm.hpp
```diff
@@ -49,47 +49,236 @@ namespace stop {
...
+ * The mode for the residual norm criterion.
+ *
+ * - absolute:        Check for tolerance against residual norm.
+ *                    || r || < tau
+ *
+ * - initial_resnorm: Check for tolerance relative to the initial residual norm.
+ *                    || r ||/|| r_0|| < tau
+ *
+ * - rhs_resnorm:     Check for tolerance relative to the rhs norm.
+ *                    || r ||/|| b || < tau
```
upsj, line 61: Can you check how this is formatted in Doxygen? It might be helpful to add $ $ around the mathematical expressions

pratikvn:
Maybe I am still misunderstanding what you mean here.

Consider a [logging call from the gmres apply](https://github.com/ginkgo-project/ginkgo/blob/develop/core/solver/gmres.cpp#L164). This essentially the [`log` function](https://github.com/ginkgo-project/ginkgo/blob/develop/include/ginkgo/core/log/logger.hpp#L547), which calls the [`on_x` based on the `Event` template parameter](https://github.com/ginkgo-project/ginkgo/blob/develop/include/ginkgo/core/log/logger.hpp#L132), therefore calling `on_iteration_complete` with the parameters passed into the `log` function.

The question is as for example in the call to [`log` in gmres](https://github.com/ginkgo-project/ginkgo/blob/develop/core/solver/gmres.cpp#L164),  where you pass in 4 parameters in addition to the solver, the [`on` in the logger](https://github.com/ginkgo-project/ginkgo/blob/develop/include/ginkgo/core/log/logger.hpp#L132) does not know which `on_iteration_complete` to call, because both the `on_iteration_complete` have the same parameters and have only one extra in the second overload. 

Unless we specify all the parameters for the `iteration_complete` in the `log` call, this will be ambiguous. And adding parameters to all `log` calls will break interface, because in MFEM and deal.ii, we do derive and use from the previous loggers.

upsj:
The new overload with the additional parameter must not have any default parameters, then there is no ambiguity between them. If you specify all parameter, the new overload is used, otherwise the old one is called.

pratikvn:
Okay, I think I finally get you mean. :)

include/ginkgo/core/stop/residual_norm.hpp
pratikvn: I added the $ $ and frac to make it prettier in doxygen.

upsj (approved):
LGTM! I like the unification of the interface, probably something we can finish with 2.0
Only the two overloads of iteration_complete need to be swapped, otherwise old loggers can miss new events.

core/solver/bicgstab.cpp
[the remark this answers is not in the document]
pratikvn: I need to always report this so that `check_impl` can capture it. If it is `nullptr`, then the `check_impl` for `ImplicitResidualNorm` will fail.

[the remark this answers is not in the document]
upsj: Do you see any downside to checking for nullptr in check_impl and, in that case, doing nothing? If the stopping status is not changed, the iteration should continue normally, right? Only then we can't stop at half-iterations with BiCGSTAB

PR #703, opened 2021-02-08
Parallel prefix sum for OpenMP backend

Slaedr:
Simple algorithm with O(n log n) work and O(log n) depth or span.

Edit: It was decided to change the algorithm to one with O(n) work and O(p) depth (p = number of processors). Even though it's less parallel in a sense, it now has less memory accesses.

### DIFF
--- a/omp/components/prefix_sum.cpp
+++ b/omp/components/prefix_sum.cpp
@@ -33,9 +33,11 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "core/components/prefix_sum.hpp"
 
 
-#include <omp.h>
 #include <algorithm>
 #include <vector>
+
+
+#include <omp.h>
 
 
 namespace gko {
--- a/omp/components/prefix_sum.cpp
+++ b/omp/components/prefix_sum.cpp
@@ -92,8 +92,8 @@ void prefix_sum(std::shared_ptr<const OmpExecutor> exec,
 
     counts[0] = 0;
 
-    const int levels = static_cast<IndexType>(std::ceil(std::log(nthreads)));
-    for (IndexType ilvl = 0; ilvl < levels; ilvl++) {
+    const auto levels = static_cast<int>(std::ceil(std::log(nthreads)));
+    for (int ilvl = 0; ilvl < levels; ilvl++) {
         const IndexType factor = power(2, (ilvl + 1));
         const IndexType lvl_num_witems = factor * def_num_witems;
         const int ntasks = (nthreads - 1) / factor + 1;
--- a/omp/components/prefix_sum.cpp
+++ b/omp/components/prefix_sum.cpp
@@ -35,23 +35,13 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include <omp.h>
 #include <algorithm>
+#include <vector>
 
 
 namespace gko {
 namespace kernels {
 namespace omp {
 namespace components {
-
-
-template <typename T>
-static inline T power(const T x, const int exponent)
-{
-    T ans = static_cast<T>(1);
-    for (int i = 0; i < exponent; i++) {
-        ans *= x;
-    }
-    return ans;
-}
 
 
 /*
@@ -62,11 +52,11 @@ void prefix_sum(std::shared_ptr<const OmpExecutor> exec,
                 IndexType *const counts, const size_type num_entries)
 {
     const auto nentries = static_cast<IndexType>(num_entries);
-    if (num_entries <= 1) {
-        return;
-    }
+    // the operation only makes sense for arrays of size at least 2
+    GKO_ENSURE_COMPATIBLE_BOUNDS(2, nentries);
 
     const int nthreads = omp_get_max_threads();
+    std::vector<IndexType> proc_sums(nthreads, 0);
     const IndexType def_num_witems = (num_entries - 1) / nthreads + 1;
 
 #pragma omp parallel
@@ -75,41 +65,33 @@ void prefix_sum(std::shared_ptr<const OmpExecutor> exec,
         const IndexType startidx = thread_id * def_num_witems;
         const IndexType endidx =
             std::min(nentries, (thread_id + 1) * def_num_witems);
-        const IndexType startval = counts[startidx];
 
-#pragma omp barrier
-
-        IndexType partial_sum = startval;
-        for (IndexType i = startidx + 1; i < endidx; ++i) {
+        IndexType partial_sum{0};
+        for (IndexType i = startidx; i < endidx; ++i) {
             auto nnz = counts[i];
             counts[i] = partial_sum;
             partial_sum += nnz;
         }
-        if (thread_id != nthreads - 1) {
-            counts[endidx] = partial_sum;
+
+        proc_sums[thread_id] = partial_sum;
+
+#pragma omp barrier
+
+#pragma omp single
+        {
+            for (int i = 0; i < nthreads - 1; i++) {
+                proc_sums[i + 1] += proc_sums[i];
+            }
+        }
+
+        if (thread_id > 0) {
+            for (IndexType i = startidx; i < endidx; i++) {
+                counts[i] += proc_sums[thread_id - 1];
+            }
         }
     }
 
     counts[0] = 0;
-
-    const auto levels = static_cast<int>(std::ceil(std::log(nthreads)));
-    for (int ilvl = 0; ilvl < levels; ilvl++) {
-        const IndexType factor = power(2, (ilvl + 1));
-        const IndexType lvl_num_witems = factor * def_num_witems;
-        const int ntasks = (nthreads - 1) / factor + 1;
-
-#pragma omp parallel for
-        for (int itask = 0; itask < ntasks; itask++) {
-            const IndexType startidx = std::min(
-                nentries, lvl_num_witems / 2 + itask * lvl_num_witems + 1);
-            const IndexType endidx =
-                std::min(nentries, (itask + 1) * lvl_num_witems + 1);
-            const IndexType baseval = counts[startidx - 1];
-            for (int i = startidx; i < endidx; i++) {
-                counts[i] += baseval;
-            }
-        }
-    }
 }
 
 GKO_INSTANTIATE_FOR_EACH_INDEX_TYPE(GKO_DECLARE_PREFIX_SUM_KERNEL);
--- a/omp/components/prefix_sum.cpp
+++ b/omp/components/prefix_sum.cpp
@@ -34,10 +34,12 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
 #include <algorithm>
-#include <vector>
 
 
 #include <omp.h>
+
+
+#include "core/base/allocator.hpp"
 
 
 namespace gko {
@@ -53,23 +55,29 @@ template <typename IndexType>
 void prefix_sum(std::shared_ptr<const OmpExecutor> exec,
                 IndexType *const counts, const size_type num_entries)
 {
-    const auto nentries = static_cast<IndexType>(num_entries);
     // the operation only makes sense for arrays of size at least 2
-    GKO_ENSURE_COMPATIBLE_BOUNDS(2, nentries);
+    if (num_entries < 2) {
+        if (num_entries == 0) {
+            return;
+        } else {
+            counts[0] = 0;
+            return;
+        }
+    }
 
     const int nthreads = omp_get_max_threads();
-    std::vector<IndexType> proc_sums(nthreads, 0);
-    const IndexType def_num_witems = (num_entries - 1) / nthreads + 1;
+    vector<IndexType> proc_sums(nthreads, 0, {exec});
+    const size_type def_num_witems = (num_entries - 1) / nthreads + 1;
 
 #pragma omp parallel
     {
         const int thread_id = omp_get_thread_num();
-        const IndexType startidx = thread_id * def_num_witems;
-        const IndexType endidx =
-            std::min(nentries, (thread_id + 1) * def_num_witems);
+        const size_type startidx = thread_id * def_num_witems;
+        const size_type endidx =
+            std::min(num_entries, (thread_id + 1) * def_num_witems);
 
         IndexType partial_sum{0};
-        for (IndexType i = startidx; i < endidx; ++i) {
+        for (size_type i = startidx; i < endidx; ++i) {
             auto nnz = counts[i];
             counts[i] = partial_sum;
             partial_sum += nnz;
@@ -87,13 +95,11 @@ void prefix_sum(std::shared_ptr<const OmpExecutor> exec,
         }
 
         if (thread_id > 0) {
-            for (IndexType i = startidx; i < endidx; i++) {
+            for (size_type i = startidx; i < endidx; i++) {
                 counts[i] += proc_sums[thread_id - 1];
             }
         }
     }
-
-    counts[0] = 0;
 }
 
 GKO_INSTANTIATE_FOR_EACH_INDEX_TYPE(GKO_DECLARE_PREFIX_SUM_KERNEL);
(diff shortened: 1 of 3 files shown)


upsj:
Great idea, we can really use this. I would suggest a slightly modified algorithm though, which would reduce the amount of synchronization and memory access by adding a small sequential part:
* Each thread gets a block of data (like right now)
* Each thread computes the local prefix sum + total block sum
* Each thread writes the total block sum into `total_sums[thread_id]`
* One thread computes the prefix-sum over total_sums
* Each thread adds `total_sums[thread_id]` to each of its local entries.

This would mean 2n reads and 2n writes in parallel plus p reads and writes sequentially for p threads

Slaedr:
> Great idea, we can really use this. I would suggest a slightly modified algorithm though, which would reduce the amount of synchronization and memory access by adding a small sequential part:
> 
>     * Each thread gets a block of data (like right now)
> 
>     * Each thread computes the local prefix sum + total block sum
> 
>     * Each thread writes the total block sum into `total_sums[thread_id]`
> 
>     * One thread computes the prefix-sum over total_sums
> 
>     * Each thread adds `total_sums[thread_id]` to each of its local entries.
> 
> 
> This would mean 2n reads and 2n writes in parallel plus p reads and writes sequentially for p threads

Thanks for taking a look at it. I had actually thought of this first and it's simpler to implement. The advantage is the total work would be O(n) instead of O(n log n), but the disadvantage is the max depth would be O(p). In a weak scaling scenario, which to me is the most important for exascale etc., the max depth would be O(n). The extra memory allocation also becomes O(n). The current approach might be slower for some small problem sizes, but might be better suited for Ginkgo, no? The current approach also avoids dynamic memory allocation altogether.

upsj:
Can you clarify what you mean by depth? It sounds like you are implying a size limit for this approach? If necessary, the prefix sum on total_sums can also be parallelized if the number of cores is too large (even though this sounds unlikely on-node). Finally, we only need O(p) additional memory, so this should really be negligible.

omp/components/prefix_sum.cpp
```diff
@@ -33,21 +33,82 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
 {
-    IndexType partial_sum{};
-    for (size_type i = 0; i < num_entries; ++i) {
-        auto nnz = counts[i];
-        counts[i] = partial_sum;
-        partial_sum += nnz;
+    const auto nentries = static_cast<IndexType>(num_entries);
+    if (num_entries <= 1) {
+        return;
+    }
```
upsj, line 67:
```suggestion
    if (num_entries <= 1) {
        counts[0] = 0;
        return;
    }
```

Slaedr:
> Can you clarify what you mean by depth? It sounds like you are implying a size limit for this approach? If necessary, the prefix sum on total_sums can also be parallelized if the number of cores is too large (even though this sounds unlikely on-node). Finally, we only need O(p) additional memory, so this should really be negligible.

By depth I mean the number of sequential steps in the kernel. You make a good point - because this is CPUs we're talking about, the number of cores should typically be very small. In the largest cases maybe it's up to 100 or so but seldom more than that. 

But by having an extra array in main memory, would we not be introducing more latency? Right now, what you would be storing in the global `total_sums` array, I just store in a local register. I think pretty much all of my accesses are cache-friendly. (There would be a few cache-line conflicts which would happen in both implementations, but that's possibly avoidable in the current implementation by moving the boundaries of work chunks a bit.) I wonder if the extra cost of 1 logarithm and log n number of power operations are worth avoiding the O(p) uncached main memory accesses.

upsj:
The reads and writes into `total_sums` should be negligible, since after the small prefix sum computation, each thread only needs to read a single element from `total_sums`, so it can be kept in a register, too. If you look at line 109, if I see this correctly, each element is read and written O(log n) times instead of 1 time in the two-level approach.

Slaedr:
format!

omp/components/prefix_sum.cpp
```diff
@@ -33,22 +33,67 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+        }
+
+        if (thread_id > 0) {
+            for (IndexType i = startidx; i < endidx; i++) {
+                counts[i] += proc_sums[thread_id - 1];
+            }
+        }
     }
+
+    counts[0] = 0;
```
upsj, line 96: I believe this is not necessary, since L74 already sets the first entry to 0

omp/components/prefix_sum.cpp
```diff
@@ -33,22 +33,67 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+                IndexType *const counts, const size_type num_entries)
 {
-    IndexType partial_sum{};
-    for (size_type i = 0; i < num_entries; ++i) {
-        auto nnz = counts[i];
-        counts[i] = partial_sum;
-        partial_sum += nnz;
+    const auto nentries = static_cast<IndexType>(num_entries);
+    // the operation only makes sense for arrays of size at least 2
+    GKO_ENSURE_COMPATIBLE_BOUNDS(2, nentries);
```
upsj, line 58: I don't think we should restrict ourselves here: There are some cases (conversion between empty matrices, ...) where we call prefix_sum on 0 or 1-sized arrays, the expected output would not be a failure, but writing 0 or nothing to the output array.

omp/components/prefix_sum.cpp
```diff
@@ -33,22 +33,67 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+
+    const int nthreads = omp_get_max_threads();
+    std::vector<IndexType> proc_sums(nthreads, 0);
+    const IndexType def_num_witems = (num_entries - 1) / nthreads + 1;
```
upsj, line 62: Random thought: Can nthreads ever be 0? Does the runtime make any guarantees for mis-configured OMP environments?

omp/components/prefix_sum.cpp
```diff
@@ -33,22 +33,67 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-    for (size_type i = 0; i < num_entries; ++i) {
-        auto nnz = counts[i];
-        counts[i] = partial_sum;
-        partial_sum += nnz;
+    const auto nentries = static_cast<IndexType>(num_entries);
+    // the operation only makes sense for arrays of size at least 2
+    GKO_ENSURE_COMPATIBLE_BOUNDS(2, nentries);
+
+    const int nthreads = omp_get_max_threads();
+    std::vector<IndexType> proc_sums(nthreads, 0);
```
upsj, line 61: We could use a custom allocator here, allowing us to track the allocations through loggers
```suggestion
    vector<IndexType> proc_sums(nthreads, 0, {exec});
```

omp/components/prefix_sum.cpp
```diff
@@ -33,22 +33,67 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "core/components/prefix_sum.hpp"
 
 
+#include <algorithm>
+#include <vector>
```
upsj, line 37:
```suggestion
#include <ginkgo/core/base/allocator.hpp>
```
(modulo formatting)

upsj (approved):
LGTM! Only small suggestions

omp/components/prefix_sum.cpp
Slaedr: I don't think so. All I could find in the spec (4.5) is
> The return value of the `omp_get_max_threads` routine can be used to dynamically allocate sufficient storage for all threads in the team formed at the subsequent active parallel region.

If the OpenMP implementation is correct, there's no way for `nthreads` to be zero or less. Should we worry about improperly configured OpenMP? I guess I could add a check and set it to 1 by default. I could add that to omp/components or something.

upsj, line 62: SonarQube raises a good point here: While our values are stored as IndexType, we might have more than INT_MAX of them (think many zeros in the input array). I think in this case using `auto` and letting the [integer conversion rules](https://en.cppreference.com/w/cpp/language/operator_arithmetic#Conversions) take care of it or using `size_type` would be more appropriate 
Same for startidx and endidx

omp/components/prefix_sum.cpp
```diff
@@ -33,21 +33,72 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-        partial_sum += nnz;
+    // the operation only makes sense for arrays of size at least 2
+    if (num_entries < 2) {
+        if (num_entries == 0) {
+            return;
+        } else {
+            counts[0] = 0;
+            return;
+        }
+    }
```
tcojean, line 66: For 1 entry shouldn't you still return `counts[0]`. In effect, I don't understand the sequential implementation of our prefix sum. It always puts `0` in `counts[0]` and has a lagging behavior. What gives? Usually you would simply do:
```
for (size_type i = 1; i < num_entries ; i++) {
    counts[i] += counts[i-1]; 
}
```

Slaedr: We're not actually returning a value, but I guess you're talking about setting `counts[0]`. Since `num_entries` is assumed to be input array size that we are supposed to operate on, it being zero is taken to mean 'do nothing'. If there is one entry, I indeed just do what is equivalent to the reference implementation. IMO, this operation only makes sense for `num_entries >= 2`.

In general, the idea of the implementation is to not only compute a running sum, but also shift the whole array to the right by one position. I thought of exactly the same thing as your loop. But the input/output semantics of wherever this function is used, implies we also need to shift the array and set the first position zero. So, an array containing the lengths of each row can be converted to `row_ptrs` as long as it has an extra allocation in the end. I guess this is referred to as an exclusive scan, as opposed to an inclusive scan.

tcojean: You're not using the `return` statement but `counts` is where what you computed is passed to the user (returned).

Also, if the purpose of this is to be implemented as it is, then I would really not call this `prefix_sum` because the shifting to the right part is not natural and not how actual prefix sums are implemented/used, it's really confusing.
In addition, what about the final value stored in `counts[num_entries-1]`? This one should also be off by `counts[num_entries-1]` won't it, since the sum has this lagging behavior. You would need to store into `counts[num_entries]` to have the actual full sum.

Slaedr: I'm not sure about the name `prefix_sum` either. I would prefer `exclusive_scan`; see https://en.cppreference.com/w/cpp/algorithm/exclusive_scan. Indeed, the typical exclusive scan does not take into account the final value, which is simply lost.

How it's normally used in Ginkgo is: an extra slot is left empty at the end of the input array. So for an array of row sizes for `N` rows, one allocates an array of size `N+1` with the first `N` entries containing useful data. Then the `counts[N]` on output will contain the total number of nonzeros in all rows, as we need for `row_ptrs`. It's a consequence of how this function is used currently throughout the code.

upsj: What we have here is an *inclusive* prefix sum, what you are describing is an *exclusive* prefix sum. Looking at Ginkgo, I don't think we use exclusive prefix sums anywhere, so this choice seems sensible to me. Most calls that need the full sum just use num_elements + 1 as the input size.

tcojean: I mean, I understand that it's useful for building CSR row pointers in particular, but the way it is this is really unlike a  `prefix_sum`, you need to have extra space available and add a `+1` to the element counts, knowing that the last entry will be discarded and everything is shifted... I also did not even see any disclaimer or note anywhere hinting in this direction, either.

Slaedr: I did add a comment on top of my function body. But yes, I'd agree with documenting this in `core/components/prefix_sum.hpp`, if that would be better. I'd also be okay with renaming this everywhere to `exclusive_scan`.

tcojean: I guess using `exclusive_scan` or `exclusive_prefix_sum` as a name would solve the issue for me, but when reviewing this I was expected to find the canonical (inclusive) prefix sum which made this really confusing. Another way would be to implement the usual prefix sum and when using it start at index 1 of the array, no?

upsj: I find the naming *scan* much less intuitive than *prefix sum*, since it neither describes the operation (summing up) nor over which elements it happens (the previous elements, be it inclusive or exclusive). The common use cases for prefix sums, work and data distribution, all lead to simpler implementations when you use an exclusive scan, because you can get both the beginning and end of a range from the same array without special case treatment. Do you see any use cases for inclusive prefix sums in Ginkgo?

tcojean: I'm not saying we need to implement the inclusive one necessarily, we don't seem to require it right now. It's just that for me since my early algorithmic courses, and as far as I can see in many places on the web (starting with wikipedia), `prefix_sum` is reserved for the inclusive form, and does not have data loss and other such pitfalls which you need to take into account for this one. I think at least having a name which reflects that would be helpful for the future, that's all.

thoasm: I actually thought that a prefix sum is the sum of all "prefix" elements, meaning all elements before `i`. So I thought it is always an **"exclusive"** prefix sum.
Maybe a template parameter `bool inclusive` is all we need to have both implementations available? We already have this in [`common/components/prefix_sum.hpp.inc` L44](https://github.com/ginkgo-project/ginkgo/blob/develop/common/components/prefix_sum.hpp.inc#L44)
Also, I agree that we should have the behavior documented in [the prefix header](https://github.com/ginkgo-project/ginkgo/blob/develop/core/components/prefix_sum.hpp)

Slaedr: I agree that `scan` is a weird name for this operation. But the advantage of `exclusive_scan` is it's shorter than `exclusive_prefix_sum`, and consistent with std which means it's kind of standard nomenclature - people would either know it or it would help them to look it up. Of course we could also document what it does. I guess we would also change the name of the relevant files if we change this name.

I'm also okay with just leaving the name as it is and adding proper documentation.

tcojean: I believe `scan` is called as such because it's not only a `prefix_sum` but it can also use `max` instead or any other operator. Maybe we could consider going that way as well if we want generality, although I'm not sure we need it now.

Slaedr: Oh yes, I see. The C++ std implementation takes an optional BinaryOp. We don't need it right now, but it would be an option to template prefix_sum on a BinaryOp (though it would be a bit complicated). Maybe for now we keep it as it is, and decide on a name and/or add documentation, unless people feel the added generality is worth the extra time and changes.

thoasm: I agree with @Slaedr : I would keep it as it is while documenting the behavior (would be nice if you could do that in this PR).
We can't be as general as `std::scan`with `BinaryOp` because we need to make sure it works for both CPU and GPU.

fritzgoebel (approved):
LGTM

omp/test/components/prefix_sum.cpp
```diff
@@ -89,8 +89,13 @@ class PrefixSum : public ::testing::Test {
 TYPED_TEST_SUITE(PrefixSum, gko::test::IndexTypes);
 
 
-TYPED_TEST(PrefixSum, SmallEqualsReference) { this->test(100); }
+TYPED_TEST(PrefixSum, TrivialCasesEqualReference)
+{
+    this->test(0);
+    this->test(1);
+}
```
pratikvn, line 96:
```suggestion
}

```

omp/test/components/prefix_sum.cpp
```diff
@@ -89,8 +89,13 @@ class PrefixSum : public ::testing::Test {
...
 
+TYPED_TEST(PrefixSum, SmallEqualsReference) { this->test(100); }
```
pratikvn, line 98:
```suggestion
TYPED_TEST(PrefixSum, SmallEqualsReference) { this->test(100); }

```

pratikvn (approved):
LGTM!

PR #705, opened 2021-02-11
Fix build issues related to CUDA Generic API in CUDA 10.1.105

upsj:
This PR fixes the build issues caused by CUDA having added 64 bit index support only in version > 10.1.105 with no way to differentiate between the two patch versions based on macros: The 64 bit index support in SpMV benchmarks is only added from CUDA 10.2 onwards


thoasm (approved):
LGTM!

pratikvn (approved):
LGTM

tcojean (approved):
LGTM

PR #706, opened 2021-02-18
Make heat-equation example run on GPUs

upsj:
This PR adds GPU support to the heat-equation example.

### DIFF
--- a/examples/heat-equation/heat-equation.cpp
+++ b/examples/heat-equation/heat-equation.cpp
@@ -148,8 +148,8 @@ int main(int argc, char *argv[])
     // time step size for the simulation
     auto tau = 1.0 / steps_per_sec;
 
-    // create an OpenMP executor
-    auto exec = gko::OmpExecutor::create();
+    // create a CUDA executor with an associated OpenMP host executor
+    auto exec = gko::CudaExecutor::create(0, gko::OmpExecutor::create());
     // load heat source and initial state vectors
     std::ifstream source_stream("data/source.mtx");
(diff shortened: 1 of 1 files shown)


yhmtsai (approved):
LGTM. Does adding executor selection in this example make sense?

upsj:
I think since this is meant as an example for a tutorial setting, keeping it as simple as possible is desirable (similar to the minimal-cuda-solver example). But maybe @hartwiganzt also has an opinion about that.

pratikvn (approved):
LGTM!

Slaedr (changes requested):
Looks good; just one concern. Looks like this example can no longer be run without a NVidia GPU. Maybe it would be nice to enable different executors based on a command line argument, or at least the omp executor along with cuda.

upsj:
@Slaedr Mike had the same concern, though I think keeping this as simple as possible for an introductory tutorial is also important. Maybe let's discuss that tomorrow.

Slaedr:
Okay, I understand. Even for the tutorial, maybe it's nice to just select between `omp` and `cuda` (not bother with the whole `map`)?

yhmtsai:
Could you check whether using cg without precondition on coo matrix works or not?
For dpcpp porting PR (not just in one branch), I plan to work dense->coo->cg first.
Thus, if it is fine, could you also change it?

Slaedr (approved):
So I'll leave it to you to decide about the executor then, though I'd personally like to see some (simplified) form of backend selection to at least enable running on CPU. In any case, thanks for the nice artistic example!

PR #707, opened 2021-02-18
Multidimensional row-major and column-major accessors

Slaedr:
This PR adds a multidimensional generalization of the existing 2D row-major accessor. The old one is preserved as a template specialization to not break interface.

There is also a new multidimensional 'column-major' accessor. It is very similar to the row-major accessor except that the innermost two dimensions are flipped. So eg., a 4D column-major accessor would be a set of column-major blocks arranged in a block-row-major fashion with respect to each other.

### DIFF
--- a/core/test/base/range_accessors.cpp
+++ b/core/test/base/range_accessors.cpp
@@ -275,4 +275,137 @@ TEST_F(RowMajorAccessor3d, CanAssignSubranges)
 }
 
 
+class ColMajorAccessor3d : public ::testing::Test {
+protected:
+    using span = gko::span;
+    static constexpr gko::size_type dimensionality{3};
+
+    using col_major_range =
+        gko::range<gko::accessor::col_major<int, dimensionality>>;
+
+    // clang-format off
+    int data[2 * 3 * 4]{
+         1, 3, 5,
+         2, 4, 6,
+		-1,-2,-3,
+		11,12,13,
+
+		21,25,29,
+		22,26,30,
+		23,27,31,
+		24,28,32
+
+		/* This matrix actually looks like
+        1, 2, -1, 11,
+        3, 4, -2, 12,
+        5, 6, -3, 13,
+
+        21, 22, 23, 24,
+        25, 26, 27, 28,
+        29, 30, 31, 32
+	    */
+    };
+    // clang-format on
+    const gko::dim<dimensionality> dim1{2, 3, 4};
+    const gko::dim<dimensionality> dim2{2, 2, 3};
+    col_major_range default_r{data, dim1};
+    col_major_range custom_r{
+        data, dim2,
+        std::array<const gko::size_type, dimensionality - 1>{12, 3}};
+};
+
+
+TEST_F(ColMajorAccessor3d, ComputesCorrectStride)
+{
+    auto range_stride = default_r.get_accessor().stride;
+    auto check_stride = std::array<const gko::size_type, 2>{12, 3};
+
+    ASSERT_EQ(range_stride, check_stride);
+}
+
+
+TEST_F(ColMajorAccessor3d, CanAccessData)
+{
+    EXPECT_EQ(default_r(0, 0, 0), 1);
+    EXPECT_EQ(custom_r(0, 0, 0), 1);
+    EXPECT_EQ(default_r(0, 1, 0), 3);
+    EXPECT_EQ(custom_r(0, 1, 0), 3);
+    EXPECT_EQ(default_r(0, 1, 1), 4);
+    EXPECT_EQ(default_r(0, 1, 3), 12);
+    EXPECT_EQ(default_r(0, 2, 2), -3);
+    EXPECT_EQ(default_r(1, 2, 1), 30);
+    EXPECT_EQ(default_r(1, 2, 2), 31);
+    EXPECT_EQ(default_r(1, 2, 3), 32);
+}
+
+
+TEST_F(ColMajorAccessor3d, CanWriteData)
+{
+    default_r(0, 0, 0) = 4;
+    custom_r(1, 1, 1) = 100;
+
+    EXPECT_EQ(default_r(0, 0, 0), 4);
+    EXPECT_EQ(custom_r(0, 0, 0), 4);
+    EXPECT_EQ(default_r(1, 1, 1), 100);
+    EXPECT_EQ(custom_r(1, 1, 1), 100);
+}
+
+
+TEST_F(ColMajorAccessor3d, CanCreateSubrange)
+{
+    auto subr = custom_r(span{0, 2}, span{1, 2}, span{1, 3});
+
+    EXPECT_EQ(subr(0, 0, 0), 4);
+    EXPECT_EQ(subr(0, 0, 1), -2);
+    EXPECT_EQ(subr(1, 0, 0), 26);
+    EXPECT_EQ(subr(1, 0, 1), 27);
+}
+
+
+TEST_F(ColMajorAccessor3d, CanCreateRowVector)
+{
+    auto subr = default_r(1u, 2u, span{0, 2});
+
+    EXPECT_EQ(subr(0, 0, 0), 29);
+    EXPECT_EQ(subr(0, 0, 1), 30);
+}
+
+
+TEST_F(ColMajorAccessor3d, CanCreateColumnVector)
+{
+    auto subr = default_r(span{0u, 2u}, 1u, 3u);
+
+    EXPECT_EQ(subr(0, 0, 0), 12);
+    EXPECT_EQ(subr(1, 0, 0), 28);
+}
+
+
+TEST_F(ColMajorAccessor3d, CanAssignValues)
+{
+    default_r(1, 1, 1) = default_r(0, 0, 0);
+
+    EXPECT_EQ(data[16], 1);
+}
+
+
+TEST_F(ColMajorAccessor3d, CanAssignSubranges)
+{
+    default_r(1u, span{0, 2}, span{0, 3}) =
+        custom_r(0u, span{0, 2}, span{0, 3});
+
+    EXPECT_EQ(data[12], 1);
+    EXPECT_EQ(data[15], 2);
+    EXPECT_EQ(data[18], -1);
+    EXPECT_EQ(data[21], 24);
+    EXPECT_EQ(data[13], 3);
+    EXPECT_EQ(data[16], 4);
+    EXPECT_EQ(data[19], -2);
+    EXPECT_EQ(data[22], 28);
+    EXPECT_EQ(data[14], 29);
+    EXPECT_EQ(data[17], 30);
+    EXPECT_EQ(data[20], 31);
+    EXPECT_EQ(data[23], 32);
+}
+
+
 }  // namespace
--- a/core/test/base/range_accessors.cpp
+++ b/core/test/base/range_accessors.cpp
@@ -275,13 +275,13 @@ TEST_F(RowMajorAccessor3d, CanAssignSubranges)
 }
 
 
-class ColMajorAccessor3d : public ::testing::Test {
+class BlockColMajorAccessor3d : public ::testing::Test {
 protected:
     using span = gko::span;
     static constexpr gko::size_type dimensionality{3};
 
-    using col_major_range =
-        gko::range<gko::accessor::col_major<int, dimensionality>>;
+    using blk_col_major_range =
+        gko::range<gko::accessor::block_col_major<int, dimensionality>>;
 
     // clang-format off
     int data[2 * 3 * 4]{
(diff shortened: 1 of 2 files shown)


Slaedr:
The build with Intel 17 and CUDA 9.2 is failing with an error that I don't understand. It seems to think an `array<size_type,N>` is a `size_type` for some reason.. unless I'm mistaken.

include/ginkgo/core/base/range_accessors.hpp
```diff
@@ -86,9 +312,179 @@ class row_major {
...
+     *                the dimensions (i.e.
+     *                `x_1 * stride_1 + x_2 * stride_2 * ... + x_n`
+     *                points to the element at (x_1, x_2, ..., x_n))
+     */
+    constexpr GKO_ATTRIBUTES explicit row_major(data_type data,
+                                                dim<dimensionality> size,
+                                                stride_type stride)
+        : data{data},
+          lengths{detail::to_array<const size_type>(size)},
+          stride{stride}
```
thoasm, line 336:
```suggestion
          lengths(detail::to_array<const size_type>(size)),
          stride(stride)
```

include/ginkgo/core/base/range_accessors.hpp
```diff
@@ -86,9 +312,179 @@ class row_major {
...
+     * Creates a row_major accessor with a default stride (assumes no padding)
+     *
+     * @param data  pointer to the block of memory containing the data
+     * @param lengths size / length of the accesses of each dimension
+     */
+    constexpr GKO_ATTRIBUTES explicit row_major(data_type data,
+                                                dim<dimensionality> size)
+        : data{data},
+          lengths{detail::to_array<const size_type>(size)},
+          stride{detail::compute_default_stride_array(lengths)}
```
thoasm, line 349:
```suggestion
          lengths(detail::to_array<const size_type>(size)),
          stride(detail::compute_default_stride_array(lengths))
```

thoasm (commented):
Maybe the intel compiler expects a list of `gko::size_type` values with the list-initialization, and tries to convert the `std::array` to one element of it. If that is the case, this should fix it (you would also need to do that for all constructor calls, I only did it exemplary for `row_major<2>`).

Slaedr:
It looks like there's a lot of code duplication. Some of it is difficult to avoid until we remove the dimension 2 specialization of `row_major` and remove the copy functions.

Some of it comes from `core/base/accessor_helpers.hpp`. At first glance that file now appears redundant. @thoasm Do you think some stuff in that file can now be removed?

thoasm:
@Slaedr At the current state, no, it can't be removed because the `accessor_helper.hpp` uses `dim<>` as the size, while the `row_major<>` uses `std::array`. When moving the accessor to the independent folder (PR https://github.com/ginkgo-project/ginkgo/pull/708), I changed it to `std::array` as well, so some of it could be re-used. However, since I want to keep it independent of the rest of Ginkgo, I don't know a way to re-use it while keeping it independent (especially since it is in a public header here).

Slaedr:
Yes, sorry, soon after posting that comment I realized that that file would be part of the new separate-but-not-really-separate accessor library. I guess the duplication is required for the separation.

Slaedr:
Right now, the type of stride is `size_type`. So, the index computations (multidimensional index tuple -> flat 1D index) will happen with 64-bit unsigned integers, regardless of the actual type of index passed into the accessor `operator()`.

Question: would it be better to have a template parameter for the stride type instead? Typically, `int` would be enough. Would using `int` instead of `size_t` save some cycles for compute-bound situations? In some situations we might be able to save on casts as well.

thoasm:
@Slaedr The problem with 32bit integers is that you simply can't address more than 2GiB of memory. How do you want to determine when to use 32, and when to use 64 bit? You can't determine it automatically, so I guess the only option is another template parameter.
If it is compute bound, it will not matter if you use 32 or 64bit because you obviously do more than one computation with the data you read, making the index computation totally irrelevant. Also, on 64bit machines, your address will always be 64bit, so at least the last indexing will be performed in 64bit arithmetic.
I will do some testing while benchmarking to see if there is any difference. I did a small amount of research, and some even say that using 32bit while computing the indexing might be slower because you need to convert it to 64bit.

thoasm:
A thought I had recently: would it be possible to move all of your new accessors to the `accessor` folder (see PR #708)? That would allow us to make more changes to the accessors, AND reduce the amount of code duplication.

Slaedr:
Okay, that's a good point about compute-bound situations - even there the cost of index computation is probably a tiny fraction of the overall cost. I don't think the 2 billion limit is an issue though - I doubt you would ever need more than 2 billion entries on one node. In any case, I did mean making it a template parameter, but I see that it might not be worth the effort.

Slaedr:
About moving the column-major accessor to the folder as well, that is a good idea. I originally did not want my `DenseBlocksView` class to be public and therefore had put it inside core (rather than include/ginkgo/core). Maybe we could merge this extend_row_major branch into your independent_accessor branch, and then finally both together into develop.

include/ginkgo/core/base/range_accessors.hpp
```diff
@@ -205,6 +631,289 @@ class row_major {
...
+ * This helper runs from first to last dimension in order to compute the index.
+ * The index is computed like this:
+ * indices: x1, x2, x3, ..., xn
+ * compute(stride, x1, x2, x3, ..., x(n-1), xn) ->
+ *  x1 * stride[0] + x2 * stride[1] + ...
+ *    + x(n-2) * stride[n-3] + x(n-1) + xn * stride[n-2]
+ * Note that swap of the last two strides, making this 'column major'.
+ */
+template <typename ValueType, size_type total_dim, size_type current_iter = 1>
+struct index_helper_s {
```
thoasm, line 647: Instead of going for a new helper structure, I would recommend rename the `compute` of the `row_major` helper to `compute_row_major`, and name the `column_major` one `compute_column_major`. That would save quite a bit of code duplication.

include/ginkgo/core/base/range_accessors.hpp
```diff
@@ -205,6 +631,289 @@ class row_major {
...
+               first * stride[dim_idx - 1];
+    }
+};
+
+/**
+ * Computes the storage index for the given indices with respect to the given
+ * stride array
+ */
+template <typename ValueType, size_type total_dim, typename... Indices>
+constexpr GKO_ATTRIBUTES ValueType compute_index(
```
thoasm, line 696: Maybe rename the previous one to `compute_row_major_index` and this one to `compute_col_major_index`, so both can be in the namespace `detail`, and it is clear what each is for.

thoasm (commented):
My main concern is that the name `col_major` does not properly describe the the access pattern (also, it is quite confusing to me). Would it be possible to make it a "true" column-major, and adapt the algorithm you need it for? Alternatively, the name needs to change IMO (unfortunately, I can't come up with a good one right now).
The rest looks good.

thoasm:
Name proposal from @upsj that I fully support: `block_col_major`.

include/ginkgo/core/base/range_accessors.hpp
Slaedr: I'm not sure what you mean. This `index_helper_s` does not do the same thing as `row_major_helper_s`. Instead of adding a prefix like 'colmajor' to everything, I just decided on a `detail_colmajor` namespace.

include/ginkgo/core/base/range_accessors.hpp
```diff
@@ -205,6 +631,289 @@ class row_major {
...
+        return GKO_ASSERT(first < size[dim_idx]),
+               first * stride[dim_idx] +
+                   index_helper_s<ValueType, total_dim, current_iter + 1>::
+                       compute(size, stride, std::forward<Indices>(idxs)...);
+    }
+};
+
+template <typename ValueType, size_type total_dim>
+struct index_helper_s<ValueType, total_dim, total_dim> {
+    static_assert(total_dim >= 2, "Dimensionality must be >= 2");
```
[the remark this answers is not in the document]
Slaedr: Yes. This concept of 'column-major', where the second-last index is the innermost one, is meaningless for 1D arrays where the 'last' index is necessarily the innermost one and there's no 'second-last' index.

include/ginkgo/core/base/range_accessors.hpp
Slaedr: Well.. there's still stuff like `default_stride_array_impl` below which will become too long if I do that. Especially if we rename the accessor as block_col_major. Maybe I can just add a line of documentation for `detail_colmajor` to make things clear? (I guess this namespace will then be renamed `detail_block_col_major`.)

include/ginkgo/core/base/range_accessors.hpp
```diff
@@ -205,6 +631,289 @@ class row_major {
...
+ * You should not try to explicitly create an instance of this accessor.
+ * Instead, supply it as a template parameter to a range, and pass the
+ * constructor parameters for this class to the range (it will forward it to
+ * this class).
+ *
+ * @tparam ValueType  type of values this accessor returns
+ * @tparam Dimensionality  number of dimensions of this accessor
+ */
+template <typename ValueType, size_type Dimensionality>
+class col_major {
```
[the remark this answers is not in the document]
Slaedr: I don't think so.. I need the blocks to be contiguous, not interleaved, at least for now. A true 3D layout_right, for example, would have the (0,0) entry of all blocks first, then the (1,0) entry of all blocks, and so on. That's suitable sometimes, but not what I'm looking for right now. I also think  "row major" and "column major" really only make sense in 2D. But I'm okay with `block_col_major` for this.

include/ginkgo/core/base/range_accessors.hpp
thoasm: The structure provides the same framework for both row- and column-major (`ValueType`, `total_dim` and `current_iter`, which is counted upwards). That is why I suggested to merge them together and just have a different function name. If you want to keep it totally separate, I can understand.

include/ginkgo/core/base/range_accessors.hpp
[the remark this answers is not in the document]
thoasm: Row- and column-major do exist for more than 2D (see the wiki page about [Row- and column-major](https://en.wikipedia.org/wiki/Row-_and_column-major_order)). It might be useful to have a full column-major implementation in the future, which is why I would prefer if you would rename it to `block_col_major`.

include/ginkgo/core/base/range_accessors.hpp
thoasm: I think a comment for `compute_storage_index` might be needed, since there is no information if it is column- or row-major.
For this one, you are right, the namespace should be enough (in my mind, I was fusing both namespaces and would have a clash there).

include/ginkgo/core/base/range_accessors.hpp
Slaedr: Ah, I think I see what you mean now. Yes, it might be more readable to keep them separate at the cost of a few lines of duplication in the struct definition itself.

thoasm (approved):
LGTM!

core/test/base/range_accessors.cpp
```diff
@@ -135,4 +164,248 @@ TEST_F(RowMajorAccessor, CanAssignSubranges)
...
+    static constexpr gko::size_type dimensionality{3};
+
+    using blk_col_major_range =
+        gko::range<gko::accessor::block_col_major<int, dimensionality>>;
+
+    // clang-format off
+    int data[2 * 3 * 4]{
+         1, 3, 5,
+         2, 4, 6,
+		-1,-2,-3,
```
upsj, line 290: nit: maybe clean up the formatting a bit here

upsj (approved):
LGTM!

tcojean (approved):
LGTM.

include/ginkgo/core/base/range_accessors.hpp
```diff
@@ -205,6 +635,301 @@ class row_major {
 };
 
 
+/**
+ * Namespace for helper functions and structs for
+ * the block column major accessor.
+ */
+namespace detail_block_col_major {
```
pratikvn, line 642: Why is this namespace itself `detail_block_col_major` instead of `namespace detail { namespace block_col_major {}} ` ?

pratikvn (approved):
LGTM!

include/ginkgo/core/base/range_accessors.hpp
Slaedr: Yes, I guess one option would be to put the row-major helpers in `detail::row_major::` and block-col-major helpers in `detail::block_col_major::`. Would that be better?

pratikvn: Yes, I think that would be better.

PR #709, opened 2021-02-22
Update HIP and hwloc in INSTALL.md

tcojean:
Update `INSTALL.md`'s HIP and hwloc details to the current setup

+ We forgot to mention `rocRAND` in the dependencies and since some issues were fixed, in particular with the CUDA setup, there is no need anymore to mention the forks of `hipSPARSE` and `hipBLAS`.
+ In addition, `hwloc` was not added as dependency whereas it should have been (thanks Mike!).
+ Also set `.md` file change to add the label `reg:documentation`.


tcojean:
This PR now also updates the hwloc portion of the text as it was missing, thanks @yhmtsai

INSTALL.md
```diff
@@ -177,57 +179,13 @@ of HIP either at `/opt/rocm/hip` or at the path specified by `HIP_PATH` as a
...
-
-```bash
-# HIP
-git clone https://github.com/ROCm-Developer-Tools/HIP.git
-pushd HIP && mkdir build && pushd build
-cmake .. && make install
-popd && popd
-
-# hipBLAS
-git clone https://github.com/tcojean/hipBLAS.git
```
yhmtsai, line 220: Can we directly use official hip without any issue now?
something like using hip on nvidia card does not need rocm dependency

INSTALL.md
```diff
@@ -281,16 +248,22 @@ packages can be turned off by disabling the relevant options.
...
 Ginkgo to try to use an external version of a package. For this, Ginkgo provides
 two ways to find packages. To rely on the CMake `find_package` command, use the
-CMake option `-DGINKGO_USE_EXTERNAL_<package>=ON`. Note that, if the external
-packages were not installed to the default location, the CMake option
-`-DCMAKE_PREFIX_PATH=<path-list>` needs to be set to the semicolon (`;`)
-separated list of install paths of these external packages. For more
-Information, see the [CMake documentation for
+CMake option `-DGINKGO_USE_EXTERNAL_<package>=ON`. `HWLOC` works the opposite
+way, Ginkgo always looks for the system's `hwloc` first. Set
+`-DGINKGO_USE_EXTERNAL_HWLOC=OFF` to not use the system provided one.
```
yhmtsai, line 261: does it work?
we set the GINKGO_USE_EXTERNAL_HWLOC=1 when enabling GINKGO_BUILD_HWLOC.
I thought it will override the parameter or have two copies in different scope?

yhmtsai (approved):
LGTM

.github/label.sh
```diff
@@ -28,7 +28,7 @@ LABELS=$LABELS$(label_match reg:benchmarking '^benchmark/')
 LABELS=$LABELS$(label_match reg:example '^examples/')
 LABELS=$LABELS$(label_match reg:build '(cm|CM)ake')
 LABELS=$LABELS$(label_match reg:ci-cd '(^\.github/|\.yml$)')
-LABELS=$LABELS$(label_match reg:documentation '^doc/')
+LABELS=$LABELS$(label_match reg:documentation '^doc/|\.md$')
```
Slaedr, line 31: Do you think we should also include `*.in` and `*.hpp*`, for the files in `headers` and `examples`?

INSTALL.md
Slaedr: @yhmtsai  I have not tried explicitly setting `-DGINKGO_USE_EXTERNAL_HWLOC=OFF`, but without setting this, it does use the system HWLOC at least on Ubuntu 20.04. So at least the default behaviour works correctly.

Slaedr (approved):
LGTM.

INSTALL.md
```diff
@@ -41,6 +41,8 @@ Ginkgo adds the following additional switches to control what is being built:
     `OFF` otherwise.
 *   `-DGINKGO_HIP_AMDGPU="gpuarch1;gpuarch2"` the amdgpu_target(s) variable
     passed to hipcc for the `hcc` HIP backend. The default is none (auto).
+*   `-DGINKGO_BUILD_HWLOC={ON, OFF}` builds Ginkgo with HWLOC. If system HWLOC
```
pratikvn, line 44: I have been thinking about this and now I think we should maybe set this to be `OFF` by default ? We might run into issues when inter-operating with libraries such as MFEM and deal.ii as they also have some hwloc support.

INSTALL.md
pratikvn: Yes, if a suitable HWLOC is found, then that is **always** used. setting `GINKGO_USE_EXTERNAL_HWLOC=off` makes no difference. 

Do we want the user to be able to force to use the internal version, even if the external version is suitable ? This will probably cause issues, will not have the required functionality and also cause havoc when interfacing with other libraries.

pratikvn (approved):
LGTM! One comment about changing the default HWLOC building.

thoasm (approved):
LGTM!

INSTALL.md
tcojean: Yes, all of these were fixed recently. The only one we found recently (yesterday) to not work in this case is`hipFFT`, but the changes aren't merged yet.

.github/label.sh
tcojean: I would not consider hpp files to be documentation. We generate documentation from them but they're not documentation only. Some hpp files can be sitting in private Ginkgo space as well where documentation policies can be more lax.

INSTALL.md
tcojean: Would we have issues? The big question is if they also download/install their own hwloc and cannot find ours. In another PR I fixed this (I think) as I found that we don't install the hwloc that we build whereas we should, considering that it's a public dependency.

INSTALL.md
tcojean: I guess I will simply remove this comment then.

PR #710, opened 2021-02-24
Dpcpp port dense

yhmtsai:
This PR is WIP. It is runnable on Gen9 GPU in current state.

This is mainly for dense porting and related kernels.

TODO:
- [x] dpcpp fix the issue on cpu side in next release.
- [x] device config selection?
- [x] merge #757
- [x] merge #733  (can be separate


yhmtsai:
format!

yhmtsai:
this can be reviewed.
For those kernels are unifying in #733 , I do not add the config selection and they do not need to be reviewed.

dpcpp/components/reduction.dp.hpp
```diff
@@ -0,0 +1,301 @@
...
+template <std::uint32_t cfg = KCFG_1D::encode(256, 32), typename ValueType>
+void reduce_add_array(dim3 grid, dim3 block, size_t dynamic_shared_memory,
+                      sycl::queue *stream, size_type size,
+                      const ValueType *source, ValueType *result)
+{
+    stream->submit([&](sycl::handler &cgh) {
+        sycl::accessor<UninitializedArray<ValueType, KCFG_1D::decode<0>(cfg)>,
+                       0, sycl::access::mode::read_write,
+                       sycl::access::target::local>
+            block_sum_acc_ct1(cgh);
```
tcojean, line 229: Does mixing buffer and USM work here? That sounds a bit suspicious. Btw, is this the failing function which requires the `queue->wait_and_throw()`?

dpcpp/components/reduction.dp.hpp
```diff
@@ -0,0 +1,301 @@
...
+ *
+ * @return the reduction result
+ */
+template <typename ValueType>
+ValueType reduce_add_array(std::shared_ptr<const DpcppExecutor> exec,
+                           size_type size, const ValueType *source)
+{
+    auto block_results_val = source;
+    size_type grid_dim = size;
+    auto block_results = Array<ValueType>(exec);
```
tcojean, line 263: Can't this form of initialization have overhead? Technically, it's doing a copy from one to the other which could be avoided.

dpcpp/components/reduction.dp.hpp
```diff
@@ -0,0 +1,301 @@
...
+
+        block_results.resize_and_reset(grid_dim);
+
+        reduce_add_array_call(cfg, grid_dim, wg_size, 0, exec->get_queue(),
+                              size, source, block_results.get_data());
+
+        block_results_val = block_results.get_const_data();
+    }
+
+    auto d_result = Array<ValueType>(exec, 1);
```
tcojean, line 287: Same here

dpcpp/components/thread_ids.dp.hpp
```diff
@@ -0,0 +1,320 @@
...
+namespace dpcpp {
+/**
+ * @brief The DPCPP thread namespace.
+ *
+ * @ingroup dpcpp_thread
+ */
+namespace thread {
+
+
+// #include "common/components/thread_ids.hpp.inc"
```
tcojean, line 55:
```suggestion
```

dpcpp/components/thread_ids.dp.hpp
```diff
@@ -0,0 +1,320 @@
...
+/**
+ * @internal
+ *
+ * Returns the ID of the block group this thread belongs to.
+ *
+ * @return the ID of the block group this thread belongs to
+ *
+ * @note Assumes that grid dimensions are in standard format:
+ *       `(block_group_size, first_grid_dimension, second grid_dimension)`
```
tcojean, line 64: Isn't this reversed with DPC++?

dpcpp/components/thread_ids.dp.hpp
```diff
@@ -0,0 +1,320 @@
...
+
+/**
+ * @internal
+ *
+ * Returns the ID of the block this thread belongs to.
+ *
+ * @return the ID of the block this thread belongs to
+ *
+ * @note Assumes that grid dimensions are in standard format:
+ *       `(block_group_size, first_grid_dimension, second grid_dimension)`
```
tcojean, line 81: Same thing. And everywhere else in the file.

dpcpp/components/thread_ids.dp.hpp
```diff
@@ -0,0 +1,320 @@
...
+ *         config::warp_size)`
+ */
+template <int subwarp_size>
+__dpct_inline__ size_type get_local_subwarp_id(sycl::nd_item<3> item_ct1)
+{
+    // dpcpp does not have subwarp.
+    constexpr auto subwarps_per_warp = subwarp_size / subwarp_size;
+    return get_local_warp_id(item_ct1) * subwarps_per_warp +
+           item_ct1.get_local_id(1);
+}
```
tcojean, line 131: Should we remove this function?

dpcpp/components/thread_ids.dp.hpp
```diff
@@ -0,0 +1,320 @@
...
+ *       respectively.
+ */
+template <int subwarp_size, int warps_per_block>
+__dpct_inline__ size_type get_subwarp_id(sycl::nd_item<3> item_ct1)
+{
+    // dpcpp dose not have subwarp
+    constexpr auto subwarps_per_warp = subwarp_size / subwarp_size;
+    return get_warp_id<warps_per_block>(item_ct1) * subwarps_per_warp +
+           item_ct1.get_local_id(1);
+}
```
tcojean, line 203: Maybe we remove it?

dpcpp/components/thread_ids.dp.hpp
```diff
@@ -0,0 +1,320 @@
...
+template <int subwarp_size, typename IndexType = size_type>
+__dpct_inline__ IndexType get_subwarp_id_flat(sycl::nd_item<3> item_ct1)
+{
+    static_assert(!(subwarp_size & (subwarp_size - 1)),
+                  "subwarp_size must be a power of two");
+    return item_ct1.get_local_id(2) / subwarp_size +
+           static_cast<IndexType>(item_ct1.get_local_range().get(2) /
+                                  subwarp_size) *
+               item_ct1.get_group(2);
+}
```
tcojean, line 289: Also this one

dpcpp/components/thread_ids.dp.hpp
```diff
@@ -0,0 +1,320 @@
...
+ * @tparam IndexType  the index type
+ */
+template <int subwarp_size, typename IndexType = size_type>
+__dpct_inline__ IndexType get_subwarp_num_flat(sycl::nd_item<3> item_ct1)
+{
+    static_assert(!(subwarp_size & (subwarp_size - 1)),
+                  "subwarp_size must be a power of two");
+    return item_ct1.get_local_range().get(2) / subwarp_size *
+           static_cast<IndexType>(item_ct1.get_group_range(2));
+}
```
tcojean, line 311: And this one

dpcpp/components/uninitialized_array.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+#include "dpcpp/base/dpct.hpp"
+
+
+namespace gko {
+namespace kernels {
+namespace dpcpp {
+
+
+// #include "common/components/uninitialized_array.hpp.inc"
```
tcojean, line 48:
```suggestion
```

dpcpp/components/uninitialized_array.hpp
```diff
@@ -0,0 +1,118 @@
...
+     *
+     * @return a reference to the array entry at the given index.
+     */
+    __dpct_inline__ ValueType &operator[](size_type pos) noexcept
+    {
+        return data_[pos];
+    }
+
+private:
+    // unsigned char data_[sizeof(ValueType) / sizeof(unsigned char) * size];
```
tcojean, line 108:
```suggestion
```

include/ginkgo/core/synthesizer/containers.hpp
```diff
@@ -113,6 +114,13 @@ template <typename T>
 using as_list = typename detail::as_list_impl<T>::type;
 
 
+template <typename T, T... Value>
+constexpr std::array<T, sizeof...(Value)> as_array(value_list<T, Value...> vl)
+{
```
tcojean, line 119: Missing documentation, since it's a public header and not in `detail`.

include/ginkgo/core/synthesizer/containers.hpp
```diff
@@ -113,6 +114,13 @@ template <typename T>
 using as_list = typename detail::as_list_impl<T>::type;
```
tcojean, line 114: Actually, this one technically needs to be documented as well.

dpcpp/matrix/dense_kernels.dp.cpp
```diff
@@ -65,11 +68,926 @@ namespace dpcpp {
...
+    syn::value_list<std::uint32_t, KCFG_1D::encode(512, 64),
+                    KCFG_1D::encode(512, 32), KCFG_1D::encode(512, 16),
+                    KCFG_1D::encode(256, 32), KCFG_1D::encode(256, 16),
+                    KCFG_1D::encode(256, 8)>();
+constexpr auto subgroup_list = syn::value_list<std::uint32_t, 64, 32, 16, 8>();
+constexpr auto kcfg_1d_array = as_array(kcfg_1d_list);
+constexpr auto default_block_size = 256;
+
+
+// #include "common/matrix/dense_kernels.hpp.inc"
```
tcojean, line 82:
```suggestion
```

dpcpp/matrix/dense_kernels.dp.cpp
```diff
@@ -65,11 +68,926 @@ namespace dpcpp {
...
+                         const ValueType *x, size_type stride_x,
+                         const ValueType *y, size_type stride_y,
+                         ValueType *work)
+{
+    constexpr auto wg_size = KCFG_1D::decode<0>(cfg);
+    stream->submit([&](sycl::handler &cgh) {
+        sycl::accessor<UninitializedArray<ValueType, wg_size>, 0,
+                       sycl::access::mode::read_write,
+                       sycl::access::target::local>
+            tmp_work_acc_ct1(cgh);
```
tcojean, line 251: Same question with mixing USM and buffers? And later in the file.

dpcpp/matrix/dense_kernels.dp.cpp
```diff
@@ -65,11 +68,926 @@ namespace dpcpp {
...
+
+        cgh.parallel_for(
+            sycl_nd_range(grid, block), [=](sycl::nd_item<3> item_ct1) {
+                compute_partial_dot<cfg>(
+                    num_rows, x, stride_x, y, stride_y, work, item_ct1,
+                    (UninitializedArray<ValueType, wg_size> *)
+                        tmp_work_acc_ct1.get_pointer());
```
tcojean, line 258: C-style cast? Also the same for all later ones.

dpcpp/matrix/dense_kernels.dp.cpp
```diff
@@ -65,11 +68,926 @@ namespace dpcpp {
...
+
+
+template <std::uint32_t cfg, typename ValueType, typename IndexType>
+void count_nnz_per_row(size_type num_rows, size_type num_cols, size_type stride,
+                       const ValueType *__restrict__ work,
+                       IndexType *__restrict__ result,
+                       sycl::nd_item<3> item_ct1)
+{
+    constexpr auto sg_size = KCFG_1D::decode<1>(cfg);
+    const auto row_idx = thread::get_subwarp_id_flat<sg_size>(item_ct1);
```
tcojean, line 477: We should maybe not use subwarp here? Either we change the term to be subgroup or we use another function altogether.

dpcpp/matrix/dense_kernels.dp.cpp
```diff
@@ -112,16 +1082,92 @@ template <typename ValueType>
...
+
+GKO_BIND_DOT(float, dot, oneapi::mkl::blas::row_major::dot);
+GKO_BIND_DOT(double, dot, oneapi::mkl::blas::row_major::dot);
+GKO_BIND_DOT(std::complex<float>, dot, oneapi::mkl::blas::row_major::dotu);
+GKO_BIND_DOT(std::complex<double>, dot, oneapi::mkl::blas::row_major::dotu);
+GKO_BIND_DOT(float, conj_dot, oneapi::mkl::blas::row_major::dot);
+GKO_BIND_DOT(double, conj_dot, oneapi::mkl::blas::row_major::dot);
+GKO_BIND_DOT(std::complex<float>, conj_dot, oneapi::mkl::blas::row_major::dotc);
+GKO_BIND_DOT(std::complex<double>, conj_dot,
+             oneapi::mkl::blas::row_major::dotc);
```
tcojean, line 1121: This should be in a corresponding blas file?

dpcpp/matrix/dense_kernels.dp.cpp
```diff
@@ -206,16 +1470,53 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
...
 
 
 template <typename ValueType>
 void calculate_max_nnz_per_row(std::shared_ptr<const DpcppExecutor> exec,
                                const matrix::Dense<ValueType> *source,
-                               size_type *result) GKO_NOT_IMPLEMENTED;
+                               size_type *result)
+{
+    const auto num_rows = source->get_size()[0];
+    auto nnz_per_row = Array<size_type>(exec, num_rows);
```
tcojean, line 1492: Same remark with Array copy style of initialization. Maybe the compiler optimizes it away, but I don't know if we should rely on that. Writing it without the copy is as short or even shorter? Same comments in other places in the file.

dpcpp/matrix/dense_kernels.dp.cpp
```diff
@@ -33,25 +33,28 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "core/matrix/dense_kernels.hpp"
 
 
-#include <algorithm>
-
-
 #include <CL/sycl.hpp>
+#include <oneapi/mkl.hpp>
```
tcojean, line 37: This should be interfaced through the BLAS file.

dpcpp/test/matrix/dense_kernels.cpp
```diff
@@ -0,0 +1,975 @@
...
+// TEST_F(Dense, IsConjugateTransposable)
+// {
+//     set_up_apply_data();
+
+//     auto trans = c_x->conj_transpose();
+//     auto dtrans = dc_x->conj_transpose();
+
+//     GKO_ASSERT_MTX_NEAR(static_cast<ComplexMtx *>(dtrans.get()),
+//                         static_cast<ComplexMtx *>(trans.get()), 0);
+// }
```
tcojean, line 574: Why?

tcojean (dismissed):
Nice work!

I think we need to improve in a few places:
+ The MKL interaction should be better done, the interface should be isolated in a blas file like we do with cuBLAS etc and we should have some finder for it even if it's really simple for now.
+ I wonder if there is a way specify the shared memory in any other way, for now it looks like we are mixing buffer access with USM and I'm not sure how happy DPC++/SYCL is with that.
+ I think we should at least remove instances of `subwarp` in kernel names and other such things, possibly removing all subwarp functions if they are now useless.

On the subwarp topic, there is now an explicit SIMD interface in Intel, I don't know how usable that is and if that will allow subwarp functionality on GPUs later on. For now, it's experimental. Maybe we should experience with it a little to see if this is useful?
https://intel.github.io/llvm-docs/doxygen/classcl_1_1sycl_1_1ext_1_1intel_1_1experimental_1_1esimd_1_1simd.html

dpcpp/CMakeLists.txt
```diff
@@ -57,6 +58,7 @@ target_compile_options(ginkgo_dpcpp PRIVATE "${GINKGO_DPCPP_FLAGS}")
 target_compile_features(ginkgo_dpcpp PRIVATE cxx_std_17)
 target_link_options(ginkgo_dpcpp PRIVATE -fsycl-device-lib=all)
 target_link_options(ginkgo_dpcpp PRIVATE -fsycl-device-code-split=per_kernel)
+target_link_libraries(ginkgo_dpcpp PRIVATE "mkl_sycl;mkl_intel_ilp64;mkl_sequential;mkl_core")
```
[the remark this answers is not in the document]
upsj: Intel [provides one in their oneMKL library](https://github.com/oneapi-src/oneMKL/tree/develop/cmake), but _of course_ they don't ship it (or even better, a `MKLConfig.cmake`). In the past you [could get MKL](https://cmake.org/cmake/help/git-stage/module/FindBLAS.html) with
```
set(BLA_VENDOR Intel_*)
find_package(BLAS)
```
but they didn't update this setup for oneAPI yet.

dpcpp/base/helper.hpp
```diff
@@ -40,12 +40,36 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+ * kernels which require encoded config but do not need explicit template
+ * parameter and share memory
+ *
+ * @param name_  the name of the host function with config
+ * @param kernel_  the kernel name
+ */
+#define GKO_ENABLE_DEFAULT_HOST(name_, kernel_)                     \
+    template <typename... InferredArgs>                             \
+    void name_(dim3 grid, dim3 block, size_t dynamic_shared_memory, \
+               sycl::queue *queue, InferredArgs... args)            \
```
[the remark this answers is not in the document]
yhmtsai: Yes, and it somehow will be replaced by another pull request

dpcpp/base/helper.hpp
```diff
@@ -115,18 +139,19 @@ namespace dpcpp {
...
-        device.get_info<cl::sycl::info::device::sub_group_sizes>();
-    auto max_workgroup_size =
-        device.get_info<sycl::info::device::max_work_group_size>();
-    bool allowed = false;
-    for (auto &i : subgroup_size_list) {
-        allowed |= (i == subgroup_size);
+    for (auto &cfg : arr) {
+        if (verify(cfg)) {
+            return cfg;
+        }
```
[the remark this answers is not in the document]
yhmtsai: Yes, if considering the usage from current dense.
It will depends on the verify function and it is changeable in the host

dpcpp/components/reduction.dp.hpp
yhmtsai: I am not sure it is buffer or not. but it is generated by DPCT.
Not sure for the reason

dpcpp/components/thread_ids.dp.hpp
yhmtsai: I don't update them, they all in cuda sense

dpcpp/matrix/dense_kernels.dp.cpp
yhmtsai: dpct did, I think the get_pointer will not return the actual type pointer directly?

dpcpp/CMakeLists.txt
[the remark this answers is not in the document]
tcojean: Great finding that they have one. I guess we can pay the price of including their license when using DPC++ in that case until something better is done about this.

dpcpp/test/matrix/dense_kernels.cpp
yhmtsai: they do not support axpy with transpose, so we need to do trans(A) * I to get transpose or conj_transpose from oneMKL

yhmtsai:
esimd seems to have limitation there. https://software.intel.com/content/www/us/en/develop/documentation/oneapi-dpcpp-cpp-compiler-dev-guide-and-reference/top/optimization-and-programming-guide/vectorization/explicit-vector-programming/explicit-simd-sycl-extension.html
it only supports intel gpu, it only enable N=8, 16, 32, and many functions are not supported.

tcojean:
Thanks for looking int it. I guess in that case we don't have a reason to keep `subwarp` related stuff?

dpcpp/base/config.hpp
```diff
@@ -49,6 +49,12 @@ struct config {
      */
     using lane_mask_type = uint64;
 
+
+    /**
+     * The number of threads within a CUDA warp.
+     */
+    static constexpr uint32 warp_size = 32;
+
```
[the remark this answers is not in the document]
yhmtsai: I am thinking to keep it as 16 or 32 for a while.
Although some CPU only support 4,8, it accidently can handle 16.
Use it as default without doing the config selection like this pr might be quick push 1.4 forward.

dpcpp/components/prefix_sum.dp.hpp
```diff
@@ -0,0 +1,255 @@
...
+ * @param subwarp     the cooperative group representing the subwarp.
+ *
+ * @tparam inclusive  if this is true, the computed prefix sum will be
+ *                    inclusive, otherwise it will be exclusive.
+ *
+ * @note For this function to work on architectures with independent thread
+ * scheduling, all threads of the subwarp have to execute it.
+ */
+template <bool inclusive, typename ValueType, typename Group>
+__dpct_inline__ void subwarp_prefix_sum(ValueType element,
```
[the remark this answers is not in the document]
yhmtsai: let's keep it in the beginning and change it later. maybe after know how dpcpp will use them in the kernel.
hope only change subwarp -> subgroup

dpcpp/components/reduction.dp.hpp
yhmtsai: I think there is no copy? it is temporary workspace to store the block result from the reduction first step

dpcpp/CMakeLists.txt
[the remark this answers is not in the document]
yhmtsai: it does not link explicitly, so it will require compile flag public. Otherwise, it can not link to oneMKL in ginkgo_core
If you click link with oneMKL libraries explicitly, we do not need to make the compile flag public

dpcpp/test/matrix/dense_kernels.cpp
upsj: Can we write our own naive transpose kernel in the meantime?

upsj (commented):
I have a few small questions and comments. In general, our conversion kernels are mostly implemented as 1D kernels. Do we expect any upside from using the configuration selection there? This seems unnecessarily complex to me. Also, it seems like we are using oneMKL only partially?

yhmtsai:
For those 1D kernel, it should be handled by the common kernel, right?
I do not think it is necessary for every kernel. It's more like showing the flexibility.
we do not test the performance against the config.
If the cpu subgroup size issue is a issue, do not become a feature.
By the way, I found 2d kernel still uses 2d range, is it missed?
I think 1d range gives better performance.
For the transpose, I could implement one but it will not be in this pr.

yhmtsai:
for oneMKL, https://software.intel.com/content/www/us/en/develop/documentation/onemkl-windows-developer-guide/top/getting-started/cmake-config-for-onemkl.html
we can use this after update docker image/container.
they seems to link implicitly, so the flags/linking are public not private.
the downside of it: we have dpcpp version requirement. I think it should be okay because dpcpp is still updating and fixing issue frequently.

upsj:
@yhmtsai No, the conversion kernels are all implemented explicitly, in part since we don't have a device mapping for sparse matrix types yet.

yhmtsai:
do we need the device map for other format?
it only uses one time and do not have the reused parameter, so we can just pass the valuetype and something into common kernel.

upsj:
No, it would work without as well. It's just that I didn't want to overuse the common kernels, since they were mainly meant for pointwise Dense kernels, and moving them to common kernels makes later optimization more work.

yhmtsai:
I already update the cmake to use MKLConfig.cmake from oneMKL. It does not work on the dgpu because the dgpu need to wait we push the image to the dockerhub.

dpcpp/matrix/dense_kernels.dp.cpp
```diff
@@ -78,17 +1013,93 @@ template <typename ValueType>
...
+                 matrix::Dense<ValueType> *result)
+{
+    if (0) {
+        // TODO: write a custom kernel which does this more efficiently
+        for (size_type col = 0; col < x->get_size()[1]; ++col) {
+            onemkl::dot(*exec->get_queue(), x->get_size()[0],
+                        x->get_const_values() + col, x->get_stride(),
+                        y->get_const_values() + col, y->get_stride(),
+                        result->get_values() + col);
+        }
```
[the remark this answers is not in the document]
yhmtsai: what do you mean?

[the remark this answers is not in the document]
upsj: The code is disabled. Does it not work, or should we be using it? I am not a fan of leaving unused code, especially if you can get it back from the git history.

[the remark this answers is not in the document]
yhmtsai: I see. I put it in (0) because it depends on we use malloc_shared or malloc_device
if we would like to use malloc_shared for oneMKL csr ability, use onemkl routine might be better.
but if we ignore the malloc_shared, use our routine might be better. (I am not certain)

include/ginkgo/core/synthesizer/containers.hpp
```diff
@@ -46,24 +47,55 @@ namespace gko {
...
+ */
 template <typename List1, typename List2>
 struct concatenate_impl;
 
+/**
+ * concatenate_impl specializes for two value_list with the same value type.
+ *
+ * @tparam T  the value type of two value_list
+ * @tparam T...  the values of the first list
+ * @tparam T...  the values of the second list
```
thoasm, line 97: nit: Wouldn't the `tparam` be the name of the types (and not just the types)?
```suggestion
 * @tparam Values1  the values of the first list
 * @tparam Values2  the values of the second list
```

include/ginkgo/core/synthesizer/containers.hpp
```diff
@@ -73,33 +105,70 @@ struct concatenate_impl<value_list<T, Values1...>, value_list<T, Values2...>> {
...
+ * @tparam T  the input template
+ */
 template <typename T, typename = void>
 struct as_list_impl;
 
+/**
+ * as_list_impl specializes for the value_list
+ *
+ * @tparam T  the value_list type
+ * @tparam T...  the values of value_list
```
thoasm, line 133:
```suggestion
 * @tparam Values  the values of value_list
```

include/ginkgo/core/synthesizer/containers.hpp
```diff
@@ -109,10 +178,34 @@ struct as_list_impl<range<Start, End, Step>, std::enable_if_t<(Start >= End)>> {
...
 template <typename T>
 using as_list = typename detail::as_list_impl<T>::type;
 
 
+/**
+ * as_array<T> returns the array from value_list. It will be helpful if using
+ * for in runtime on the array.
+ *
+ * @tparam T  the type of value_list
+ * @tparam T...  the values of value_list
```
thoasm, line 196:
```suggestion
 * @tparam Value  the values of value_list
```

thoasm (changes requested):
For the most part, it looks good to me, but I really dislike the C-style casts and want them changed before approving.
Other than that, I mostly have typos and other small nits.

dpcpp/base/helper.dp.cpp
```diff
@@ -0,0 +1,64 @@
...
+            device.get_info<cl::sycl::info::device::sub_group_sizes>();
+        auto max_workgroup_size =
+            device.get_info<sycl::info::device::max_work_group_size>();
+        bool allowed = false;
+        for (auto &i : subgroup_size_list) {
+            allowed |= (i == subgroup_size);
+        }
+        return allowed && (workgroup_size <= max_workgroup_size);
+    }
+}
```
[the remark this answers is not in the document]
yhmtsai: good catch! just moving codes result

dpcpp/base/helper.hpp
```diff
@@ -115,18 +139,19 @@ namespace dpcpp {
...
     }
-    return allowed && (workgroup_size <= max_workgroup_size);
+    GKO_NOT_SUPPORTED(arr);
+    return 0;
```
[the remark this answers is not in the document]
yhmtsai: return a certain value if do not catch exception outside?

dpcpp/components/reduction.dp.hpp
```diff
@@ -0,0 +1,300 @@
...
+ * magnitude among all the threads in the group.
+ * Only the values from threads which set `is_pivoted` to `false` will be
+ * considered.
+ */
+template <
+    typename Group, typename ValueType,
+    typename = std::enable_if_t<group::is_communicator_group<Group>::value>>
+__dpct_inline__ int choose_pivot(const Group &group, ValueType local_data,
+                                 bool is_pivoted)
+{
```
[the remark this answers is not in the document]
yhmtsai: good point. some of them (those copy-pasting) can be put in common but some can not (not direct way).
- config:warp_size: we can still provide a common choice in dpcpp too (maybe 16, although some cpu in theory do not support it)
- shared_memory: need to separate function again - first for allocation and give it as argument

dpcpp/components/uninitialized_array.hpp
```diff
@@ -0,0 +1,118 @@
...
+     */
+    __dpct_inline__ ValueType &operator[](size_type pos) noexcept
+    {
+        return data_[pos];
+    }
+
+private:
+    // if dpcpp uses char to represent data in char, compiling gives error.
+    // Thanksfully, dpcpp support complex data allocation directly.
+    ValueType data_[size];
```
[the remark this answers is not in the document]
yhmtsai: it is mostly for porting phase. dpcpp indeed supports ValueType data_[size] directly for standard format.

dpcpp/components/uninitialized_array.hpp
```diff
@@ -0,0 +1,118 @@
...
+template <typename ValueType, size_type size>
+class UninitializedArray {
+public:
+    /**
+     * Operator for casting an UninitializedArray into its constexpr value
+     * pointer.
+     *
+     * @return the constexpr pointer to the first entry of the array.
+     */
+    constexpr __dpct_inline__ operator ValueType *() const noexcept
```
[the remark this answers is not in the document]
yhmtsai: this should be const ValueType. Do you know why dpcpp/cuda does not complain it?

dpcpp/matrix/dense_kernels.dp.cpp
```diff
@@ -65,11 +69,630 @@ namespace dpcpp {
...
+                 const ValueType *__restrict__ source,
+                 IndexType *__restrict__ row_ptrs,
+                 IndexType *__restrict__ col_idxs,
+                 ValueType *__restrict__ values, sycl::nd_item<3> item_ct1)
+{
+    const auto tidx = thread::get_thread_id_flat(item_ct1);
+
+    if (tidx < num_rows) {
+        auto write_to = row_ptrs[tidx];
+        for (auto i = 0; i < num_cols; i++) {
```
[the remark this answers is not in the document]
yhmtsai: I also fix it in other files

dpcpp/components/uninitialized_array.hpp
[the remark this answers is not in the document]
thoasm: No, I don't know why these compiler don't complain.

dpcpp/components/thread_ids.dp.hpp
thoasm: If we don't change the indexing, we will have lower performance because data accesses will not be linearized properly.

dpcpp/components/uninitialized_array.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+    /**
+     * constexpr array access operator.
+     *
+     * @param pos The array index. Using a value outside [0, size) is undefined
+     * behavior.
+     *
+     * @return a reference to the array entry at the given index.
+     */
+    constexpr __dpct_inline__ ValueType &operator[](size_type pos) const
```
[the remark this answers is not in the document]
yhmtsai: cuda complains this part

dpcpp/components/thread_ids.dp.hpp
yhmtsai: it is in cuda sense if you use dim3 here.
It will auto reverse in sycl, so it is still linear

dpcpp/components/prefix_sum.dp.hpp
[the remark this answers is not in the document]
yhmtsai: I prefer keeping this name currently for porting. At least, check these function with others like block jacobi. and then decide how to handle it.

dpcpp/base/helper.hpp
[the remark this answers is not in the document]
thoasm: When you throw, you don't continue the function, even if you catch the exception. But this is a very minor issue, you can also leave as it is.

dpcpp/components/reduction.dp.hpp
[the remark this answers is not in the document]
thoasm: OK, I see. In that case, leave the file as it is if you need to modify a lot.

dpcpp/matrix/dense_kernels.dp.cpp
yhmtsai: use `tmp_work_acc_ct1.get_pointer().get()` currently, which return `pointer_t` and it can be found in line 187 of https://intel.github.io/llvm-docs/doxygen/access_8hpp_source.html

dpcpp/matrix/dense_kernels.dp.cpp
yhmtsai: I do not fully understand what you mean. Could you give more description on it?

dpcpp/test/matrix/dense_kernels.cpp
yhmtsai: I implement it in another pr, which will also complete some necessary functions of solver.

dpcpp/base/config.hpp
```diff
@@ -49,6 +49,12 @@ struct config {
      */
     using lane_mask_type = uint64;
 
+
+    /**
+     * The number of threads within a CUDA warp.
+     */
+    static constexpr uint32 warp_size = 16;
```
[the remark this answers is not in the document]
yhmtsai: I remove it in this pr. I will maybe use it again in another pr if I feel it can port easily in this moment

dpcpp/matrix/dense_kernels.dp.cpp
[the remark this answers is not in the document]
yhmtsai: Sorry for confusing or different comment from my side. I enable the mkl routine when the `num_cols == 1`

common/factorization/par_ilut_select_kernels.hpp.inc
```diff
@@ -62,7 +62,8 @@ __global__ __launch_bounds__(searchtree_width) void build_searchtree(
     // assuming rounding towards zero
     auto stride = double(size) / sample_size;
 #pragma unroll
-    for (auto i = 0; i < sampleselect_oversampling; ++i) {
+    for (auto i = decltype(sampleselect_oversampling){0};
+         i < sampleselect_oversampling; ++i) {
```
upsj, line 66: that completely destroys readability IMO.
```suggestion
    for (int i = 0; i < sampleselect_oversampling; ++i) {
```

common/factorization/par_ilut_select_kernels.hpp.inc
```diff
@@ -119,7 +120,8 @@ __global__ __launch_bounds__(default_block_size) void count_buckets(
         auto el = abs(input[i]);
         IndexType tree_idx{};
 #pragma unroll
-        for (auto level = 0; level < sampleselect_searchtree_height; ++level) {
+        for (auto level = decltype(sampleselect_searchtree_height){0};
+             level < sampleselect_searchtree_height; ++level) {
```
upsj, line 124:
```suggestion
        for (int level = 0; level < sampleselect_searchtree_height; ++level) {
```

common/factorization/par_ilut_select_kernels.hpp.inc
```diff
@@ -168,7 +170,7 @@ __global__ __launch_bounds__(default_block_size) void block_prefix_sum(
     // compute prefix sum over warp-sized blocks
     IndexType total{};
     auto base_idx = warp_idx * work_per_warp * warp.size();
-    for (auto step = 0; step < work_per_warp; ++step) {
+    for (auto step = decltype(work_per_warp){0}; step < work_per_warp; ++step) {
```
upsj, line 173:
```suggestion
    for (IndexType step = 0; step < work_per_warp; ++step) {
```

cuda/test/components/sorting_kernels.cu
```diff
@@ -99,7 +99,7 @@ protected:
     {
         // we want some duplicate elements
         std::uniform_int_distribution<gko::int32> dist(0, num_elements / 2);
-        for (auto i = 0; i < num_elements; ++i) {
+        for (auto i = decltype(num_elements){0}; i < num_elements; ++i) {
```
upsj, line 102:
```suggestion
        for (int i = 0; i < num_elements; ++i) {
```

dpcpp/components/prefix_sum.dp.hpp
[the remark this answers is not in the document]
upsj: Small note: This function is only used in ParILUT kernels

dpcpp/components/reduction.dp.hpp
upsj: This is only an accessor, not a buffer, so I don't think this is an issue.

yhmtsai:
@ginkgo-project/reviewers I finish all reviews.
I do not change the name of thread_ids.dp.hpp due to future porting concern.
I also add some `// TODO: porting ...` to remind that we should revisit them when we are close to fully porting dpcpp or after more porting involving.

dpcpp/components/reduction.dp.hpp
upsj: It's a move, not a copy, so if the initialization doesn't happen in-place (which I think it does), it would have negligible overhead.

dpcpp/test/matrix/dense_kernels.cpp
```diff
@@ -0,0 +1,951 @@
...
+    auto ddiag = gko::matrix::Diagonal<Mtx::value_type>::create(dpcpp);
+    ddiag->copy_from(diag.get());
+    dalpha = Mtx::create(dpcpp);
+    dalpha->copy_from(alpha.get());
+
+    mat->add_scaled(alpha.get(), diag.get());
+    dmat->add_scaled(dalpha.get(), ddiag.get());
+
+    GKO_ASSERT_MTX_NEAR(mat, dmat, r<vtype>::value);
+}
```
upsj, line 315: (Some of) these are already present in `test/matrix/dense_kernels.cpp`, and should maybe be moved there since they are implemented in common

dpcpp/test/matrix/dense_kernels.cpp
```diff
@@ -0,0 +1,951 @@
...
+    set_up_apply_data();
+
+    x->apply(y.get(), expected.get());
+    dx->apply(dy.get(), dresult.get());
+
+    GKO_ASSERT_MTX_NEAR(dresult, expected, r<vtype>::value);
+}
+
+
+#if !GINKGO_DPCPP_SINGLE_MODE
```
upsj, line 387: Maybe use `GTEST_SKIP` in a helper `skip_if_no_double` (or something similar) function instead of macros.

dpcpp/test/matrix/dense_kernels.cpp
```diff
@@ -0,0 +1,951 @@
...
+{
+    set_up_apply_data();
+
+    auto imag_x = Mtx::create(ref, x->get_size());
+    x->get_imag(imag_x.get());
+    auto dimag_x = Mtx::create(dpcpp, dx->get_size());
+    dx->get_imag(dimag_x.get());
+
+    GKO_ASSERT_MTX_NEAR(imag_x, dimag_x, 0);
+}
```
upsj, line 948: Also in `test/matrix/dense_kernels.cpp`

upsj (changes requested):
LGTM in general, nice job keeping the interface so consistent w.r.t CUDA.
The `auto` changes in `common` decrease readibility quite a lot, so I would prefer to use explicit types. In general, integer literals are typed, so `0` is `int`, even with `auto`, which is why we could keep this as it is, but I understand the objection that it obscures the type of the parameters (overflows are not really a risk in almost all of these cases, since we are only dealing with local data or small sizes as in the tests).
Most of the `dense_kernels` tests are already present in `test/matrix/dense_kernels.cpp`, so you can probably remove them or move them there if they are not yet present.

common/factorization/par_ilut_select_kernels.hpp.inc
thoasm: I agree, directly stating the type is better IMO, even if it makes it slightly less adaptable (to type changes).

cmake/create_test.cmake
```diff
@@ -42,6 +42,10 @@ function(ginkgo_create_dpcpp_test test_name)
     target_compile_options(${test_target_name} PRIVATE "${GINKGO_DPCPP_FLAGS}")
     target_link_options(${test_target_name} PRIVATE -fsycl-device-code-split=per_kernel)
     ginkgo_set_test_target_properties(${test_name} ${test_target_name})
+    # Note: MKL_ENV is empty on linux. Maybe need to apply MKL_ENV to all test.
+    if (MKL_ENV)
+       set_tests_properties(${test_target_name} PROPERTIES ENVIRONMENT "${MKL_ENV}")
+    endif()
```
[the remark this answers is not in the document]
yhmtsai: this is from the MKL example https://software.intel.com/content/www/us/en/develop/documentation/onemkl-windows-developer-guide/top/getting-started/cmake-config-for-onemkl.html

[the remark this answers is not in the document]
upsj: Setting environment variables is not the responsibility of the build system. I would prefer if we remove this altogether, as we can't guarantee users set it the same way for their application.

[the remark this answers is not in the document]
yhmtsai: the input env is defined by MKL cmake and not all parameters from environment, which means no difference even if user gives same env with different input.
I am not sure whether it is only for test or for all application.
I agree with you if it only contains LD_LIBRARY_PATH
but it also contains MKL_INTERFACE_LAYER, MKL_THREADING_LAYER, MKL_BLACS_MPI depends MKL input or platform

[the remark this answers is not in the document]
upsj: > I am not sure whether it is only for test or for all application.

It is not possible to set the environment variables for an application that is built for CMake. Thus, introducing this change would mean that the tests potentially behave differently than application codes.
This is a really bad CMake setup, since it mixes runtime behavior with compile-time options. I should probably bring this up in the next Intel meeting.

dpcpp/CMakeLists.txt
```diff
@@ -54,13 +58,22 @@ target_sources(ginkgo_dpcpp
...
-target_link_options(ginkgo_dpcpp PRIVATE -fsycl-device-code-split=per_kernel)
+# When building ginkgo as a static library, we need to use dpcpp and per_kernel
+# link option when the program uses dpcpp related function.
+if (BUILD_SHARED_LIBS)
+    target_link_options(ginkgo_dpcpp PRIVATE -fsycl-device-code-split=per_kernel)
+else ()
+    target_link_options(ginkgo_dpcpp PUBLIC -fsycl-device-code-split=per_kernel)
+endif()
 target_link_libraries(ginkgo_dpcpp PUBLIC ginkgo_device)
+target_link_libraries(ginkgo_dpcpp PRIVATE $<LINK_ONLY:MKL::MKL_DPCPP>)
```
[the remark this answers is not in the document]
upsj: As far as I understand it, this only links against MKL, but doesn't add any include paths or other compiler flags associated with the MKL_DPCPP target. Why is this necessary, as opposed to plain
```suggestion
target_link_libraries(ginkgo_dpcpp PRIVATE MKL::MKL_DPCPP)
```

dpcpp/test/matrix/dense_kernels.cpp
```diff
@@ -0,0 +1,608 @@
...
+
+TEST_F(Dense, MakeComplexIsEquivalentToRef)
+{
+    set_up_apply_data();
+
+    auto complex_x = x->make_complex();
+    auto dcomplex_x = dx->make_complex();
+
+    GKO_ASSERT_MTX_NEAR(complex_x, dcomplex_x, 0);
+}
```
upsj, line 605: This should also be in `test/...`

upsj (approved):
LGTM! Only the CMake-specific discussions should be resolved from my viewpoint.

cmake/create_test.cmake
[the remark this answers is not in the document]
yhmtsai: I agree with you.

thoasm (approved):
LGTM!

dpcpp/CMakeLists.txt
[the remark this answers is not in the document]
yhmtsai: I try both version. `LINK_ONLY` is usually used by `INTERFACE_LINK_LIBRARIES` in `set_targert_properties`
using target_link_libraries will handle it automatically.
previous one gives `$<LINK_ONLY:<$<LINK_ONLY>:MKL::MKL_DPCPP>>`
updated one gives `$<LINK_ONLY>:MKL::MKL_DPCPP>`

PR #393, opened 2019-11-19
port cuda solver without triangular solver to hip

yhmtsai:
This PR ports cuda solver without triangular solver to hip

The test of preconditioner needs to use `ginkgo_create_test` not `ginkgo_create_hip_test`.
Otherwise, the stop factory is always empty.

The triangular solver needs some extra works, so I do not put it in this PR.

Note: Some conversion fails from cuda to hip.
The block_size or grid_size is too long to hipify-perl can not recognize it correctly, so need some fixes by hand.
[fail1-1](https://github.com/ginkgo-project/ginkgo/blob/78d2113dca67afea0d5567dfd511b89bb083e2b9/cuda/solver/gmres_kernels.cu#L165) and [fail1-2](https://github.com/ginkgo-project/ginkgo/blob/78d2113dca67afea0d5567dfd511b89bb083e2b9/cuda/solver/gmres_kernels.cu#L188): splitting is failed.
should be dim3(ceildiv(dim_size[0] * stride_next_krylov, default_block_size)), dim3(default_block_size), 0, 0
but get dim3(ceildiv(dim_size[0] * stride_next_krylov)), dim3(default_block_size)), default_block_size, 0

[fail2](https://github.com/ginkgo-project/ginkgo/blob/78d2113dca67afea0d5567dfd511b89bb083e2b9/cuda/solver/gmres_kernels.cu#L245): hipify does not know it is a device call, so hipify does nothing on it.

PR #712, opened 2021-02-27
Fix HIP, hwloc and test_install

tcojean:
HIP overhaul, Hwloc and test_install fixes.

### On hwloc

There were a few issues with the CMake hwloc setup.
+ hwloc was always built even when a valid version already existed on
the system. 
+ The FindHWLOC module had a few issues in terms of organization, search
paths as well as variable management. This has been fixed.
+ hwloc is a public dependency of Ginkgo but was not installed when we
use the one from our build directory, that would create errors with
missing libraries in some cases.
+ The `HWLOC_XMLFILE` should not be managed by Ginkgo, hwloc already
takes care of it, if the environment variable `HWLOC_XMLFILE` is set, it
uses that as base, otherwise it uses the machine topology it can
detect.
+ The core/devices which are always compiled were moved into their own
root directory. This better reflects the new library that it creates as
well as the fact all libraries depend on it.

### On test install

Test install was missing a HIP version which I added. In addition, the
commands were not properly triggered thus only the base (non CUDA
version) was tested, that is now fixed.

### On HIP

This part saw a lot of changes.

+ ROCm < 3.5 support is dropped as it is now old and all HPC clusters
always update the ROCm version since they know the ecosystem is somewhat
unstable.
+ Instead of linking against `hip::device` in `ginkgo_hip`, I link
against `libamdhip64.so` publicly.
+ In general, the purpose of this rework is to depend more on CMake
transitive properties such that we remove special cases. Thanks to this
and the previous point as well, it is possible to remove entirely the
`hip_helpers.cmake` file and remove all of the bans we were doing.
+ Thanks to relying on transitive properties, there is no need for the
`special_linkage` test case anymore.
+ Instead of building the `RPATH_FOR_HIP` chain, in Ginkgo everything
uses the `CXX` language for linking, which properly takes into account
RPATH information unlike when liking with the `HIP` language. For users,
a simple string `GINKGO_INSTALL_RPATH_FOR_HIP` is provided which is a
linking flag to be given to the `HIP` language linker after installing
Ginkgo.
+ In general, there is also better management or relying more on the
defaults for some environment variables and HIP variables.

### Open issues

An annoying issue (still), although it's mitigated thanks to linking
publicly against `libamdhip64.so` which doesn't create inconvenience to
the user, is that we have to include `hip/hip_runtime.hpp` in our public
headers (to find `__forceinline__` as well as `HIP_DEVICE_COMPILE`, see
https://github.com/ginkgo-project/ginkgo/blob/develop/include/ginkgo/core/base/types.hpp#L46-L85.
In addition, the `GKO_ASSERT` cannot be reworked as well otherwise it
creates mysterious issues in the `check_headers` test, see:
https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/1073846520


tcojean:
This should now be ready for review.

tcojean:
I added some explicit RPATH management for hwloc & the HIP libraries but I'm not sure if it should stay: I'm not sure if CMake is actually populating the `RUNPATH` in which case it's fine, or if it's the `RPATH` which the user cannot configure (as would be needed when copying the libraries to another platform).

cmake/Modules/FindHWLOC.cmake
```diff
@@ -40,21 +40,18 @@ include(hwloc_helpers)
...
+    PATH_SUFFIXES lib lib64
     DOC "Find the hwloc library"
     )
 
 if (HWLOC_INCLUDE_DIRS)
-    unset(HWLOC_FOUND CACHE)
-    set(HWLOC_FOUND 1)
-
-    # Find the version of hwloc found
+        # Find the version of hwloc found
```
upsj, line 54: nit: formatting

cmake/Modules/FindHWLOC.cmake
```diff
@@ -68,25 +65,26 @@ if (HWLOC_INCLUDE_DIRS)
...
-        endif()
+if (NOT HWLOC_FIND_QUIETLY)
+    if (HWLOC_INCLUDE_DIRS AND HWLOC_LIBRARIES)
+        message(STATUS "Looking for HWLOC - found version ${HWLOC_VERSION}")
+    else()
+        message(STATUS "${Magenta}Looking for HWLOC - not found"
+            "\n   Please check that your environment variable HWLOC_DIR"
+            "\n   has been set properly.${ColourReset}")
+        unset(HWLOC_LIBRARIES CACHE)
+        unset(HWLOC_INCLUDE_DIRS CACHE)
```
upsj, line 78: I could imagine that this might cause issues when users are trying to debug their FindHWLOC via CMakeCache.txt, when only one of the two paths is found correctly.

hip/CMakeLists.txt
```diff
@@ -144,8 +120,7 @@ list(APPEND CMAKE_PREFIX_PATH
 set(CMAKE_MODULE_PATH "${CMAKE_MODULE_PATH}" PARENT_SCOPE)
 set(CMAKE_PREFIX_PATH "${CMAKE_PREFIX_PATH}" PARENT_SCOPE)
 
-# setting the default flags like CMAKE_{LANG}_FLAGS_{TYPE}
-# the setting is copied from the default CMAKE_CXX_FLAGS_{TYPE}
+# NOTE: without this, HIP jacobi build takes a *very* long time
```
upsj, line 123: Could you figure out why that is the case? Too low optimization level?

upsj (approved):
LGTM, great job!

cmake/Modules/FindHWLOC.cmake
tcojean: Yes, but on the other hand if these are not cleaned then Ginkgo will build/install hwloc, but there is a risk that it will consider the headers or library from the installed version which wasn't complete, and the other one will come from the hwloc that we built... I've seen that happening quite a few times, which is why I decided to make sure the finder doesn't leave things in a weird state.

cmake/install_helpers.cmake
```diff
@@ -9,7 +9,14 @@ set(GINKGO_INSTALL_CONFIG_DIR "${CMAKE_INSTALL_LIBDIR}/cmake/Ginkgo")
 set(GINKGO_INSTALL_MODULE_DIR "${CMAKE_INSTALL_LIBDIR}/cmake/Ginkgo/Modules")
 
 function(ginkgo_install_library name subdir)
-
+    if (BUILD_SHARED_LIBS)
+        if (APPLE)
+            set(ORIGIN_OR_LOADER_PATH "@loader_path")
+        else()
+            set(ORIGIN_OR_LOADER_PATH "$ORIGIN")
+        endif()
```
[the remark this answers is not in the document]
tcojean: Sure I will.

Basically all this does is say something like "if you're looking for a library, look first into the directory this one will be installed into", thus this helps finding `libginkgo_hip.so` and other ginkgo sub-libraries. For all the other libraries which are not found, the `LD_LIBRARY_PATH` setting can be used by the user to state where to go find them. Note that with this setting, it's not possible to change anymore which Ginkgo library will be found. As long as they exist within the same directory, the libraries of the directory will always be found first and `LD_LIBRARY_PATH` will have no effect (the RPATH property is populated, not the RUNPATH).

hip/CMakeLists.txt
tcojean: Yes, AFAIK there is literally no flags passed to any compiler by default in HIP, thus by passing these flags we tell the HIP compilers to actually use optimization levels, which make things much better.
See: https://github.com/ROCm-Developer-Tools/HIP/blob/main/cmake/FindHIP.cmake#L9-L23

I wonder if we would get any benefit from passing in any device flag as well. I think for `nvcc` it doesn't matter, but I don't know about `clang`.

cmake/install_helpers.cmake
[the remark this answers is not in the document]
tcojean: Well the last note only applies to Linux, as MacOS looks first into the environment variable.

tcojean:
There is one extra thing to consider for this PR: what default RPATH behavior we want to consider for our dependencies, e.g. HIP, and if we want optional "convenient" features to manage the RPATH directly.

### Context
The RPATH is a library information field that states into which PATH the dependencies can be searched into first and foremost. You can usually read that information by doing `objdump -x libginkgo.so` for example.
On Linux, the order of searching is somewhat like this:
+ DT_RPATH (the one we said above, from the dynamic library itself)
+ LD_LIBRARY_PATH (the user defined environment variable)
+ DR_RUNPATH.


This means that if we do not provide any `RPATH` information by default, the user *needs* to provide all the correct PATH of our dependencies (e.g. hipSPARSE, rocSPARSE, ...) into the `LD_LIBRARY_PATH` environment variable.

### Issue: an installed library should be relocatable. Should it always though?
The title summarizes the main problem, when we bundle RPATH information (since I do not know how to provide the RUNPATH in CMake and it's also not as powerful as RPATH due to some other differences), we basically prevent the user from customizing the dependencies, e.g. copying an installed ginkgo library into another platform with potentially different paths/libraries and have it automatically adapt to this new setup (through the `LD_LIBRARY_PATH`). On the other hand, putting `RPATH` information for some dependencies is convenient for users which can live with a "broken" system with missing `LD_LIBRARY_PATH` settings.

### Question

Thus the question is, which path should we take forward:
+ Leave things as is (let users always populate a correct `LD_LIBRARY_PATH`)
+ Have a CMake setting for adding`RPATH` information and make it default.
+ Have a CMake setting for adding `RPATH` information and isn't default.

Note that actually setting all the RPATH information of our dependencies isn't always enough for the users' convenience, because Ginkgo itself will depend on `libhipsparse.so`, but `libhipsparse.so` doesn't bundle any `RPATH` information and thus will fail to find, for example, `librocsparse.so`. If you do an install of Ginkgo the commit e6505da on  a HIP based system (e.g. amdci) and use `lddtree` with an empty `LD_LIBRARY_PATH` you will see that happen.

tcojean:
As some extra consideration on the topic of `RPATH` and what Ginkgo's behavior should be, it seems that `MFEM` and `libCEED` both encode all the `RPATH` information into their library by default, thus creating non relocatable libraries. If many prominent packages decided to remove possibility for customization/relocatable code in order to simplify their user's life, maybe we should consider going that way as well.

cmake/create_test.cmake
```diff
@@ -101,63 +101,31 @@ function(ginkgo_create_cuda_test test_name)
     add_test(NAME ${REL_BINARY_DIR}/${test_name} COMMAND ${TEST_TARGET_NAME})
 endfunction(ginkgo_create_cuda_test)
 
-function(ginkgo_create_hip_test_special_linkage test_name)
```
yhmtsai, line 104: Do you know why special link works?

test_install/test_install.hip.cpp
```diff
@@ -0,0 +1,387 @@
...
+                      "mtx_io.hpp not included properly!");
+    }
+
+    // core/base/name_demangling.hpp
+    {
+        auto testVar = 3.0;
+        gko::name_demangling::get_static_type(testVar);
+    }
+
+
```
yhmtsai, line 137: nit
```suggestion
```

cmake/create_test.cmake
tcojean: What I understand is that the special linkage was required because it added the libraries of `HIP`, like `libamdhip64.so` in an automatic fashion. What happens now is that there is a public dependency to that library directly (the `.so` and not a target), which means that this dependency is always propagated including to our tests. Thus, the tests can just link like normal, they will anyway also link against the HIP libraries through that, which makes it all work in the default case.

See: https://github.com/ginkgo-project/ginkgo/blob/fix_hip/hip/CMakeLists.txt#L264

core/device_hooks/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ if(NOT GINKGO_BUILD_CUDA)
     add_library(ginkgo_cuda
         $<TARGET_OBJECTS:ginkgo_cuda_device>
         cuda_hooks.cpp)
-    target_link_libraries(ginkgo_cuda PUBLIC ginkgo_core_device)
+    target_link_libraries(ginkgo_cuda PRIVATE ginkgo_core_device)
```
pratikvn, line 5: `core/device` does not really exist right ? So maybe we can change this library name to `ginkgo_device` or something like that ?

pratikvn (approved):
LGTM! Thanks for fixing all of this!

yhmtsai:
Could you also update the ReadMe for hip and maybe INSTALL for hwloc?

tcojean:
All requested changes should have been done for now.

pratikvn (approved):
LGTM! 

Did you figure out why @nbeams was having an issue with the `get_pci_device` function in MFEM ?

cmake/create_test.cmake
```diff
@@ -101,63 +101,31 @@ function(ginkgo_create_cuda_test test_name)
...
         ${PROJECT_BINARY_DIR} ${CMAKE_CURRENT_BINARY_DIR})
     string(REPLACE "/" "_" TEST_TARGET_NAME "${REL_BINARY_DIR}/${test_name}")
 
     set_source_files_properties(${test_name}.hip.cpp PROPERTIES HIP_SOURCE_PROPERTY_FORMAT TRUE)
+    hip_add_executable(${TEST_TARGET_NAME} ${test_name}.hip.cpp
+        # With how HIP works, to prevent redifinition we need to comment these.
+        # They might need to be used later though.
+        #HIPCC_OPTIONS ${GINKGO_HIPCC_OPTIONS}
+        #NVCC_OPTIONS  ${GINKGO_HIP_NVCC_OPTIONS}
+        #CLANG_OPTIONS ${GINKGO_HIP_CLANG_OPTIONS}
```
yhmtsai, line 116: what do you mean redefinition here?
If it does not contain these option, can it get the arch flag and optimization?

cmake/create_test.cmake
yhmtsai: I see. I also put the previous comments link here for reference.
https://github.com/ginkgo-project/ginkgo/pull/393#issuecomment-558626356

test_install/CMakeLists.txt
```diff
@@ -37,17 +37,25 @@ if(GINKGO_BUILD_CUDA)
...
-    # -x;hip;--hip-device-lib-path=/opt/rocm/lib;--cuda-gpu-arch=gfx900;
-    # --cuda-gpu-arch=gfx906
-    set_target_properties(hip::device PROPERTIES INTERFACE_COMPILE_OPTIONS "")
-    # In addition, link libraries have a similar problem. We only keep
-    # `hip::host`. Currently, the flags should be:
-    # hip::host;--hip-device-lib-path=/opt/rocm/lib;--hip-link;
-    # --cuda-gpu-arch=gfx900;--cuda-gpu-arch=gfx906
-    set_target_properties(hip::device PROPERTIES INTERFACE_LINK_LIBRARIES "hip::host")
+if(GINKGO_BUILD_HIP)
+    find_package(HIP REQUIRED)
```
yhmtsai, line 41: do we need to add version check here?

yhmtsai (approved):
LGTM in general.
I would like check whether hip test is still compiled with correct arch flags or not.
For RPATH thing, I would like to depends on `LD_LIBRARY_PATH` by default.
also, I think if we upload compiled ginkgo library via github release, the package should use `LD_LIBRARY_PATH` to find hip stuff.
the one benefit of embedding all info into ginkgo is maybe for application with many library which use different `backend` libs.
(I am not sure whether it really works or not)

pratikvn:
@nbeams , but it was when you were building this branch and not develop right ? Do you still get it with the current status of the `fix_hip` branch ?

nbeams:
I just tried with the latest version of this branch and also the latest version of develop; they both behave the same way.  If I compile with gcc (9.2.0), I see the errors, but if I compile with ROCm clang, I don't.  It only happens when building the examples, e.g.

```
[ 84%] Linking CXX executable ginkgo-ranges
../../hip/libginkgo_hip.so.1.3.0: undefined reference to `gko::MachineTopology::get_pci_device(std::string const&) const'
collect2: error: ld returned 1 exit status
make[2]: *** [examples/ginkgo-ranges/ginkgo-ranges] Error 1
make[1]: *** [examples/ginkgo-ranges/CMakeFiles/ginkgo-ranges.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
[ 84%] Linking CXX executable cb-gmres
../../hip/libginkgo_hip.so.1.3.0: undefined reference to `gko::MachineTopology::get_pci_device(std::string const&) const'
collect2: error: ld returned 1 exit status
make[2]: *** [examples/cb-gmres/cb-gmres] Error 1
make[1]: *** [examples/cb-gmres/CMakeFiles/cb-gmres.dir/all] Error 2
```
So whatever is happening, whether it's my configuration on this machine or the hwloc I'm using or something else, it's not related to any changes in this branch.

test_install/CMakeLists.txt
tcojean: The HIP version isn't the same as ROCm and in some cases defaults to 0 so this is hard to do.

pratikvn:
I see. thanks. I will look into that.

cmake/create_test.cmake
tcojean: I have trouble explaining how it works but indeed with the current setting everything is passed correctly. I can only explain it as a bug in FindHIP.cmake, that bug would explain why we have so many duplications of all the flags as well.

For example:
```
/opt/rocm-4.0.0/hip/bin/hipcc -M /home/terry/ginkgo/hip/test/base/hip_executor_topology.hip.cpp -o /home/terry/ginkgo/build/hip/test/base/CMakeFiles/hip_test_base_hip_executor_topology.dir//hip_test_base_hip_executor_topology_generated_hip_executor_topology.hip.cpp.o.depend.pre -std=c++14 -fPIC --amdgpu-target=gfx906 --amdgpu-target=gfx908 -fPIC -O2 -DNDEBUG -I/home/terry/ginkgo/build -I/opt/rocm-4.0.0/include -I/opt/rocm-4.0.0/hipblas/include -I/opt/rocm-4.0.0/hiprand/include -I/opt/rocm-4.0.0/hipsparse/include -I/home/terry/ginkgo/build/include -I/home/terry/ginkgo/include -I/home/terry/ginkgo -I/usr/local/include -I/home/terry/ginkgo/build/include -I/home/terry/ginkgo/include -I/home/terry/ginkgo -I/usr/local/include -I/usr/local/include -I/home/terry/ginkgo/build/include -I/home/terry/ginkgo/include -I/home/terry/ginkgo -I/usr/local/include -I/home/terry/ginkgo/build/include -I/home/terry/ginkgo/include -I/home/terry/ginkgo -I/usr/local/include -I/home/terry/ginkgo/build/include -I/home/terry/ginkgo/include -I/home/terry/ginkgo -I/usr/local/include -I/opt/rocm-4.0.0/hip/include -I/home/terry/ginkgo/build/include -I/home/terry/ginkgo/include -I/home/terry/ginkgo -I/usr/local/include -I/home/terry/ginkgo/build/include -I/home/terry/ginkgo/include -I/home/terry/ginkgo -I/usr/local/include -I/home/terry/ginkgo/build/third_party/gtest/src/googletest/include -I/home/terry/ginkgo/build/third_party/gtest/src/googletest/include
```

tcojean: Note that If I pass the flags again (at least HIPCC_OPTIONS), there is two `std=c++14` and somehow that makes the command invalid. I can maybe reactivate the other flags at least for safety?

tcojean: I didn't try it to make 100% sure but I think the problem is due to these:
https://github.com/ROCm-Developer-Tools/HIP/blob/main/cmake/FindHIP.cmake#L463-L466
https://github.com/ROCm-Developer-Tools/HIP/blob/main/cmake/FindHIP.cmake#L381-L397

By looking only at these two, since no reset seem to be done, they *always* append the flags that they find to the *global* variables `HIP_HIPCC_FLAGS` and whatnot. The macro doing this is called whenever we call a `hip_add_library` or `hip_add_executable`. Thus, when we do `hip_add_executable` for the test, we already had the definitions from `libginkgo_hip.so` which get reused and thus we do not need to pass the options again (and it creates an error in the case of `-std=c++14`).
If they used a function that would make things much simpler and they could use a scoped variable to manage the flags that they find. Or they could just add them to the property of the object they are defining. I do not understand what they are doing.

In any case, I will try to put them back and see if it works, although that will duplicate everything at least twice.

tcojean: For completeness, here is the failing build, I think it's only `nvcc` which creates issues so I will make a special case out of it:
https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/1060533273
```
nvcc fatal   : redefinition of argument 'std'
```

cmake/install_helpers.cmake
```diff
@@ -8,27 +8,32 @@ set(GINKGO_INSTALL_PKGCONFIG_DIR "${CMAKE_INSTALL_LIBDIR}/pkgconfig")
...
+function(ginkgo_add_install_rpath name subdir)
     if (BUILD_SHARED_LIBS)
-        set (HWLOC_LIB_PATH "")
-        if  (GINKGO_HAVE_HWLOC AND GINKGO_USE_EXTERNAL_HWLOC)
-            get_filename_component(HWLOC_LIB_PATH ${HWLOC_LIBRARIES} DIRECTORY)
+        if (GINKGO_INSTALL_RPATH_ORIGIN)
+            if (APPLE)
+                set(ORIGIN_OR_LOADER_PATH "@loader_path")
+            else()
+                set(ORIGIN_OR_LOADER_PATH "$ORIGIN")
```
upsj, line 17: I guess this would be a no-op on Windows?

cmake/install_helpers.cmake
```diff
@@ -8,27 +8,32 @@ set(GINKGO_INSTALL_PKGCONFIG_DIR "${CMAKE_INSTALL_LIBDIR}/pkgconfig")
...
 
-function(ginkgo_install_library name subdir)
-    # This setting sets the directory of the library as the `INSTALL_RPATH`.
-    # This is required for `libginkgo.so` to find its dependencies such as
-    # `libginkgo_omp.so` etc. Note that with this setting it's not possible on
-    # Linux to control which Ginkgo libraries are found and where: if all the
-    # libraries are present in the directory, these will be used first and
-    # foremost and the environment variable `LD_LIBRARY_PATH` will have no
-    # effect.
+function(ginkgo_add_install_rpath name subdir)
```
upsj, line 11: I can't see the subdir parameter being used anywhere. Is it still relevant here?

core/CMakeLists.txt
```diff
@@ -64,18 +64,22 @@ add_library(Ginkgo::ginkgo ALIAS ginkgo)
...
+    list(APPEND GKO_RPATH_ADDITIONS "${GKO_PAPI_LIBDIR}")
 endif()
 
-# Since we have a public dependency on HIP, this dependency appears here as well
-set(RPATH_ADD_HIP_PATH "")
+# Since we have a public dependency on HIP, this dependency appears
+# here as well
 if (GINKGO_BUILD_HIP AND GINKGO_HIP_PLATFORM MATCHES "hcc")
-    set(RPATH_ADD_HIP_PATH "${HIP_PATH}/lib")
+    list(APPEND GKO_RPATH_ADDITIONS "${HIP_PATH}/lib")
```
upsj, line 78: Do we always add this to the RPATH, or just in case GINKGO_INSTALL_RPATH_DEPENDENCIES is ON?

upsj (approved):
LGTM! Only a few clarifying questions

cmake/install_helpers.cmake
tcojean: Yes there is anyway no RPATH support on Windows (except some experimental one somewhere).

core/CMakeLists.txt
tcojean: Yes, this is passed to the `ginkgo_install_library` which gets then passed to the `ginkgo_add_install_rpath`, and nothing gets done when we are not either in `BUILD_SHARED_LIBS` or `GINKGO_INSTALL_RPATH`.
https://github.com/ginkgo-project/ginkgo/blob/fix_hip/cmake/install_helpers.cmake#L27-L30

yhmtsai: I think it already did. it passed by ARGN.

yhmtsai (approved):
LGTM. Do you think it is possible to add some related tests for rpath check? (In CI not cmake, I guess)

cmake/install_helpers.cmake
```diff
@@ -8,27 +8,32 @@ set(GINKGO_INSTALL_PKGCONFIG_DIR "${CMAKE_INSTALL_LIBDIR}/pkgconfig")
...
+            if  (GINKGO_HAVE_HWLOC AND GINKGO_USE_EXTERNAL_HWLOC)
+                get_filename_component(HWLOC_LIB_PATH ${HWLOC_LIBRARIES} DIRECTORY)
+                list(APPEND RPATH_DEPENDENCIES "${HWLOC_LIBRARIES}")
+            endif()
         endif()
-        if (APPLE)
-            set(ORIGIN_OR_LOADER_PATH "@loader_path")
-        else()
-            set(ORIGIN_OR_LOADER_PATH "$ORIGIN")
+        if (GINKGO_INSTALL_RPATH)
```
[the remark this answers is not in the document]
tcojean: Yes that's the case. I thought it was already clear, I will add more details then.

core/CMakeLists.txt
tcojean: I will add more description on that topic.

cmake/install_helpers.cmake
tcojean: Right, it's a relica of `ginkgo_install_library` where it's also useless now. Since this has been annoying me for a while, I'm removing this parameter everywhere.

cmake/create_test.cmake
```diff
@@ -108,13 +108,26 @@ function(ginkgo_create_hip_test test_name)
...
+    # creates a redefinition of all flags. This creates some issues with `nvcc`
+    # (particularly the flags in `HIPCC_OPTIONS`), but `clang` is fine with the
+    # redefinitions.
+    if (GINKGO_HIP_PLATFORM STREQUAL "nvcc")
+        hip_add_executable(${TEST_TARGET_NAME} ${test_name}.hip.cpp
+            # If `FindHIP.cmake`, namely `HIP_PARSE_HIPCC_OPTIONS` macro and
+            # call gets fixed, uncomment this.
+            # HIPCC_OPTIONS ${GINKGO_HIPCC_OPTIONS}
+            # NVCC_OPTIONS  ${GINKGO_HIP_NVCC_OPTIONS}
+            # CLANG_OPTIONS ${GINKGO_HIP_CLANG_OPTIONS}
```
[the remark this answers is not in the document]
tcojean: No it's in everything. `HIPCC` is in the name of the macro but it's used for all options, including `NVCC` and `CLANG`. The main issue actually comes from `nvcc` (also clang if I recall) which really dislikes having two times some important flags like `-std=c++14`.

tcojean (commented):
On the topic of testing RPATH in CI, I don't think it can be easily done since it depends on platform, where libraries are located, etc...

cmake/create_test.cmake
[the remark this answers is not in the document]
yhmtsai: I see. I also see what you wrote in the previous conversation.
We still have these flag added by hip on Nvidia (including optimization/arch).
The benchmark and test should have the same settings if they do not fix or change hip.
If we need to compare hip and nvcc on NVIDIA GPU again, we should remember to check these flags are still in library/benchmark/test.
Thanks

cmake/install_helpers.cmake
yhmtsai: side note, Windows can use the registry to register the library path.
I do not want to change any system settings, so I do not dig into this direction.

PR #686, opened 2021-01-02
Add support for C++ package manager(s)

upsj:
This PR adds support for the Conan Package manager to Ginkgo. I also plan to add vcpkg support, but didn't completely understand their build process yet.

TODO:
- [ ] Make tests optional
- [x] Add to CI setup (unnecessary, they do their own CI)
- [x] Add vcpkg support

PR #713, opened 2021-03-01
Clean up CMake installation step

upsj:
I was building a [vcpkg port for Ginkgo](https://github.com/upsj/vcpkg) (see #686) where I stumbled across a few issues with the current build setup, which this PR fixes:

* install and use windows and HIP helpers only when necessary
* use GNUInstallDirs provided values instead of hard-coded paths
* correctly quote paths (even though they are most likely relative)
* Avoid creating `include/CMakeFiles` by manually installing `ginkgo/config.hpp`
* Use SameMajorVersion instead of AnyNewerVersion to find potential interface breaks at configure time already

### DIFF
--- a/cmake/install_helpers.cmake
+++ b/cmake/install_helpers.cmake
@@ -2,28 +2,28 @@ include(CMakePackageConfigHelpers)
 include(GNUInstallDirs)
 
 
-set(GINKGO_INSTALL_INCLUDE_DIR "include")
-set(GINKGO_INSTALL_LIBRARY_DIR "lib")
-set(GINKGO_INSTALL_PKGCONFIG_DIR "lib/pkgconfig")
-set(GINKGO_INSTALL_CONFIG_DIR "lib/cmake/Ginkgo")
-set(GINKGO_INSTALL_MODULE_DIR "lib/cmake/Ginkgo/Modules")
+set(GINKGO_INSTALL_INCLUDE_DIR "${CMAKE_INSTALL_INCLUDEDIR}")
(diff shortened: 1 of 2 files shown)


cmake/install_helpers.cmake
```diff
@@ -60,7 +59,7 @@ function(ginkgo_install)
     write_basic_package_version_file(
         "${Ginkgo_BINARY_DIR}/GinkgoConfigVersion.cmake"
         VERSION "${PROJECT_VERSION}"
-        COMPATIBILITY AnyNewerVersion
+        COMPATIBILITY SameMajorVersion
```
yhmtsai, line 62: how does it detect interface breaking?

yhmtsai (approved):
LGTM. 
I do not understand how it detect interface breaking clearly

cmake/install_helpers.cmake
upsj: any find_package call using a specific version currently only checks that the installed version is at least as new as the desired version:
```
# installed Ginkgo 2.0
find_package(Ginkgo 1.3) # works
```
Since we do semantic versioning, a change in the major version points to an interface break, so in my view, we should fail already at this point if the user expects 1.3, but gets 2.0

tcojean (approved):
LGTM. It's important to change the COMPATIBILITY to `SameMajorVersion` before any major release, thanks for finding that.

cmake/install_helpers.cmake
yhmtsai: I see. Thanks for explanation

PR #714, opened 2021-03-03
Add implicit residual log to solvers and benchmarks

upsj:
This was something we overlooked in #702: We can also log the implicit residuals and report them in the solver benchmark.


yhmtsai (approved):
LGTM

pratikvn (approved):
Can you also update the [`custom-logger` example](https://github.com/ginkgo-project/ginkgo/blob/develop/examples/custom-logger/custom-logger.cpp#L123) ?

LGTM!

upsj:
@pratikvn Good point, when testing this out I noticed that the estimates we provide here are actually not Energy norm estimates but true residual estimates, since we need to compute r^T r in almost all of the short recurrence solvers anyways.

examples/custom-logger/custom-logger.cpp
```diff
@@ -93,38 +93,51 @@ struct ResidualLogger : gko::log::Logger {
         std::cout << "Recurrent vs true residual norm:" << std::endl;
```
pratikvn, line 93:
```suggestion
        std::cout << "Recurrent vs true residual norm vs Implicit residual norm:" << std::endl;
```

pratikvn (approved):
LGTM! You also need to update the [`results.dox` documentation file](https://github.com/ginkgo-project/ginkgo/blob/develop/examples/custom-logger/doc/results.dox#L29)

tcojean (approved):

PR #715, opened 2021-03-04
Make symbol visibility explicit

upsj:
Currently, we export all of our symbols (even internal ones), which is not really good practice and probably causes the symbol ordinal overflows and the need for `bigobj` on Windows. This PR attempts to avoid these issues by explicitly specifying the symbol visibility for all parts of Ginkgo that are used across library boundaries.

TODO:
- [x] Fix cb_gmres issues with accessor
- [x] Fix static member issues on Windows


upsj:
This is a much larger issue than expected, which cannot reasonably be solved without an interface-break. Among the issues are:

* Array is an almost header-only type (except for `fill` and `convert`), but uses std::function to store the deleter for its data. We then determine based on the type ID of its deleter whether it is an executor_deleter or something else to implement `is_owned`. This comparison is based on RTTI symbols for std::function internals, whose visibility we can't control
* The separation between Ginkgo core and its executors is not clean enough sometimes: MSVC doesn't allow different export specifiers on classes and their member functions, so we can't implement OmpExecutor::raw_copy_to(CudaExecutor) in a different executor as the rest. The interface must thus happen over free functions. This would also allow us to untangle dependencies between different executors.

Aside from that a general observation:

All of our header files are full of inline implementations of functions, which often makes it difficult to identify where a bit of functionality is instantiated and where symbols are emitted. I think it would make for a cleaner interface and faster build performance if we moved all of these implementations except for the template member functions ones into `core`. I need to check how this would impact circular dependencies, but I believe our header files are packed enough already, and will probably only grow unless we do something about this.

upsj:
This will need to be done in 2.0, but we have a draft how it could work here :)

PR #671, opened 2020-11-24
Public interface and basic reference kernels for fixed-size block CSR

Slaedr:
Adds a fixed block CSR matrix format. "Fixed" refers to the fact that the block size is uniform for the entire matrix, as opposed to what is sometimes called variable BCSR.

This PR adds the core functions and reference kernels.

TODO:
- [x] Squash some unnecessary commits

PR #716, opened 2021-03-08
Parallel block-ILU(0) factorization for reference, OpenMP and CUDA backends

Slaedr:
Two public classes have been added for factorization - `Bilu` which represents 'exact' block-ILU factorization, and `ParBilu` which represents iterative approximate block-ILU factorization. Only block-ILU(0) is supported. Various kernels implementing the required functionality for the classes have been added to three backends. For now, the kernels may not be most efficient possible.

Other required functionality has been lumped in here:
- Cusparse bindings for BSR matrix format
- Some `Fbcsr` kernels for `omp` and `cuda`

TODO:
- [ ] Add `syn`-based code to uniformly instantiate the kernels for the supported block sizes
- [ ] Maybe add a CMake interface for users to specify the block sizes they need?


pratikvn:
I think this PR is a too large to be reviewed at once. Even without the large .mtx file, it is over 10k lines. Can you please break up into 2 at least PR's ? One with core + reference kernels and the second with CUDA, HIP and OpenMP kernels ?

tcojean:
I think it's based on some other PRs like #707 #671 and it's not cleaned up yet, but indeed that would still be way too big considering all of that.

Slaedr:
Yes, this is based on those two pull requests that @tcojean pointed out. Let's take a call on splitting this one up after those are merged. And yes, I have clean-up work to do on this, in addition to the todo's.

Slaedr:
This work will be split into two separate PRs related to (1) Fbcsr kernels on Cuda and OpenMP and (2) ParBilu kernels.

PR #717, opened 2021-03-08
Mixed precision ELL

upsj:
This PR adds a full mixed precision ELL SpMV. It is based on #677

To show the potential of this approach, I implemented a full mixed precision SpMV for the ELL format, which can already be used in the benchmarks:
```
"ell": {
    "storage": 22014828,
    "max_relative_norm2": 1.2637828574791889e-16,
    "time": 2090340.0
},
"fell": {
    "storage": 14676552,
    "max_relative_norm2": 1.2333668343137365e-8,
    "time": 1610980.0
}
```

The core of this is hidden in the mixed_precision_dispatch function, which is currently still pretty limited, but can probably be extended to more general multiple dispatch, basically turning the dynamic type of a LinOp into a static Dense pointer.


fritzgoebel:
format!

include/ginkgo/core/base/types.hpp
```diff
@@ -488,6 +488,57 @@ GKO_ATTRIBUTES constexpr bool operator!=(precision_reduction x,
     template _macro(std::complex<double>, int64)
 
 
+#define GKO_INSTANTIATE_FOR_EACH_MIXED_VALUE_AND_INDEX_TYPE(_macro) \
```
thoasm, line 491: I would prefer to have this macro in the `ell_kernels` header (and named something with `INSTANTIATE_ELL_FOR`...) since it is currently an ELL exclusive.

omp/matrix/ell_kernels.cpp
```diff
@@ -56,59 +57,91 @@ namespace omp {
...
+    const auto num_stored_elements_per_row =
+        a->get_num_stored_elements_per_row();
+    const auto stride = a->get_stride();
+    const auto num_rows = b->get_size()[0];
+    const auto a_vals = gko::acc::range<a_accessor>(
+        std::array<size_type, 1>{num_stored_elements_per_row * stride},
+        a->get_const_values());
+    const auto b_vals = gko::acc::range<b_accessor>(
+        std::array<size_type, 2>{num_rows, b->get_stride()},
+        b->get_const_values());
```
thoasm, line 81: I forgot to change this one:
```suggestion
    const auto b_vals = gko::acc::range<b_accessor>(
        std::array<size_type, 2>{{b->get_size()[0], b->get_size()[1]]},
        b->get_const_values(), std::array<size_type, 1>{{b->get_stride()}});
```

omp/matrix/ell_kernels.cpp
```diff
@@ -56,59 +57,91 @@ namespace omp {
...
```
thoasm, line 125: Same here:
```suggestion
    const auto b_vals = gko::acc::range<b_accessor>(
        std::array<size_type, 2>{{b->get_size()[0], b->get_size()[1]]},
        b->get_const_values(), std::array<size_type, 1>{{b->get_stride()}});
```

reference/matrix/ell_kernels.cpp
```diff
@@ -50,57 +53,89 @@ namespace reference {
...
+    const auto num_stored_elements_per_row =
+        a->get_num_stored_elements_per_row();
+    const auto stride = a->get_stride();
+    const auto num_rows = b->get_size()[0];
+    const auto a_vals = gko::acc::range<a_accessor>(
+        std::array<size_type, 1>{num_stored_elements_per_row * stride},
+        a->get_const_values());
+    const auto b_vals = gko::acc::range<b_accessor>(
+        std::array<size_type, 2>{num_rows, b->get_stride()},
+        b->get_const_values());
```
thoasm, line 77: And here as well:
```suggestion
    const auto b_vals = gko::acc::range<b_accessor>(
        std::array<size_type, 2>{{b->get_size()[0], b->get_size()[1]]},
        b->get_const_values(), std::array<size_type, 1>{{b->get_stride()}});
```

reference/matrix/ell_kernels.cpp
```diff
@@ -50,57 +53,89 @@ namespace reference {
...
```
thoasm, line 120: And here:
```suggestion
    const auto b_vals = gko::acc::range<b_accessor>(
        std::array<size_type, 2>{{b->get_size()[0], b->get_size()[1]]},
        b->get_const_values(), std::array<size_type, 1>{{b->get_stride()}});
```

core/base/precision_dispatch.hpp
```diff
@@ -0,0 +1,204 @@
...
+        // If ValueType is complex, this branch will never be taken
+        // If ValueType is real, the cast is a no-op
+        fn(dynamic_cast<const Dense *>(dense_in->create_real_view().get()),
+           dynamic_cast<Dense *>(dense_out->create_real_view().get()));
+    } else {
+        mixed_precision_dispatch<ValueType>(fn, in, out);
+    }
+}
+
+}  // namespace gko
```
thoasm, line 201: nit: Missing newline
```suggestion


}  // namespace gko
```

thoasm (commented):
I have some nits and some corrections, but overall, it looks good to me.

include/ginkgo/core/base/types.hpp
fritzgoebel: I would like to keep it here as this will hopefully generalize to more matrix formats in the future

thoasm: Can you add a small name change to it, so it is clear that you need to have 3 different type template parameter? The current name is not really clear how many parameters you need for the macro.

thoasm: The other reason why I would like to have this in the `ell_kernels` header is because this is a public header, meaning we have to keep it in here unchanged until we have Ginkgo 2.0.

thoasm (approved):
LGTM except for the public macro `GKO_INSTANTIATE_FOR_EACH_MIXED_VALUE_AND_INDEX_TYPE` (if others are fine with it / don't mind, I will follow).

cuda/test/matrix/ell_kernels.cpp
```diff
@@ -160,6 +248,39 @@ TEST_F(Ell, SimpleApplyWithStrideIsEquivalentToRef)
 }
 
 
+TEST_F(Ell, MixedSimpleApplyWithStrideIsEquivalentToRef1)
+{
+    set_up_apply_data(532, 231, 1, 300, 600);
```
pratikvn, line 253: I guess you can extract all these sizes to variables and unify them across tests ?

hip/matrix/ell_kernels.hip.cpp
```diff
@@ -111,40 +112,69 @@ using compiled_kernels = syn::value_list<int, 0, 1, 2, 4, 8, 16, 32>;
 namespace {
 
 
-template <int info, typename ValueType, typename IndexType>
+template <int dim, typename Type1, typename Type2>
+GKO_INLINE auto as_hip_accessor(
```
pratikvn, line 116: Maybe this and the `as_cuda_accesor` can be moved to some common hip and common cuda files, as these will probably be useful for other mixed precision algorithms as well ?

hip/test/matrix/ell_kernels.hip.cpp
```diff
@@ -160,6 +249,39 @@ TEST_F(Ell, SimpleApplyWithStrideIsEquivalentToRef)
 }
 
 
+TEST_F(Ell, MixedSimpleApplyWithStrideIsEquivalentToRef1)
+{
+    set_up_apply_data(532, 231, 1, 300, 600);
```
pratikvn, line 254: Same comment as for cuda

include/ginkgo/core/base/temporary_conversion.hpp
```diff
@@ -30,8 +30,8 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ******************************<GINKGO LICENSE>*******************************/
 
-#ifndef GKO_PUBLIC_CORE_BASE_TEMPORARY_CONVERSION_HPP_
-#define GKO_PUBLIC_CORE_BASE_TEMPORARY_CONVERSION_HPP_
+#ifndef GKO_CORE_BASE_TEMPORARY_CONVERSION_HPP_
```
pratikvn, line 33: This header is not public anymore ?

include/ginkgo/core/base/types.hpp
pratikvn: Looking at this huge list, I would definitely agree with @upsj that we should allow disabling of these instantiations and also disable them by default.

omp/matrix/ell_kernels.cpp
```diff
@@ -56,59 +57,91 @@ namespace omp {
...
 
 #pragma omp parallel for
     for (size_type row = 0; row < a->get_size()[0]; row++) {
         for (size_type j = 0; j < c->get_size()[1]; j++) {
-            c->at(row, j) = zero<ValueType>();
+            c->at(row, j) = zero<OutputValueType>();
         }
         for (size_type i = 0; i < num_stored_elements_per_row; i++) {
-            auto val = a->val_at(row, i);
+            auto val = a_vals(row + i * stride);
```
pratikvn, line 89: Just an idea here. Instead of adding `a_vals` variable, which is an accessor, can we add a additional templated `val_at` function to the ELL class which takes in the indices and forwards them to the accessor and returns the reference to the value ? I think that would make the code more clear, and also maybe reduce code changes when adapting to other algorithms.

pratikvn (approved):
LGTM! Only some minor things.

hip/matrix/ell_kernels.hip.cpp
thoasm: I will do this in PR #727:
https://github.com/ginkgo-project/ginkgo/blob/4408ac930431e9edf46cf1ebf5ef25e9e2dbd909/accessor/cuda_helper.hpp#L122

omp/matrix/ell_kernels.cpp
thoasm: The problem with that is that you can't have an accessor in the ELL class since it could change the data or even the executor. An accessor can't be edited after it was created, meaning you have to create it when you know the data will not change. It would be better to use a `reduced_column_major<2, ...>` accessor, which would be able to compute the indices, however, that does not exist yet.

upsj:
@ginkgo-project/reviewers Somewhere in separating the two PRs, the actual mixed precision dispatch got lost, so I added it together with moving the mixed precision instantiation macro into private headers.

Finally, I added a GINKGO_MIXED_PRECISION flag, which is disabled by default, but enabled in all except for a handful of CI jobs, since this is exclusively a core-level change.

.gitlab-ci.yml
```diff
@@ -24,6 +24,7 @@ stages:
   BUILD_HIP: "OFF"
   BUILD_HWLOC: "ON"
   FAST_TESTS: "OFF"
+  MIXED_PRECISION: "ON"
```
pratikvn, line 27: Can we do it the other way ? Switch it `ON` only for a few jobs and `OFF` by default ? I don't see any particular need to have it on for all jobs.

CMakeLists.txt
```diff
@@ -33,6 +33,7 @@ option(GINKGO_BUILD_CUDA "Compile kernels for NVIDIA GPUs" ${GINKGO_HAS_CUDA})
 option(GINKGO_BUILD_HIP "Compile kernels for AMD or NVIDIA GPUs" ${GINKGO_HAS_HIP})
 option(GINKGO_BUILD_DOC "Generate documentation" OFF)
 option(GINKGO_FAST_TESTS "Reduces the input size for a few tests known to be time-intensive" OFF)
+option(GINKGO_MIXED_PRECISION "Add true mixed-precision kernel (otherwise they will be conversion-based)" OFF)
```
pratikvn, line 36: nit: Is this more clear ? 
```suggestion
option(GINKGO_MIXED_PRECISION "Instantiate true mixed-precision kernels (otherwise they will be conversion-based)" OFF)
```

cmake/get_info.cmake
```diff
@@ -127,7 +127,7 @@ foreach(log_type ${log_types})
     ginkgo_print_module_footer(${${log_type}} "User configuration:")
     ginkgo_print_module_footer(${${log_type}} "  Enabled modules:")
     ginkgo_print_foreach_variable(${${log_type}}
-        "GINKGO_BUILD_OMP;GINKGO_BUILD_REFERENCE;GINKGO_BUILD_CUDA;GINKGO_BUILD_HIP;GINKGO_BUILD_DPCPP")
+        "GINKGO_BUILD_OMP;GINKGO_BUILD_REFERENCE;GINKGO_BUILD_CUDA;GINKGO_BUILD_HIP;GINKGO_BUILD_DPCPP;GINKGO_MIXED_PRECISION")
```
pratikvn, line 130: I would maybe move this to line 133, and modify the heading there, as it is technically not a module.

include/ginkgo/core/base/precision_dispatch.hpp
```diff
@@ -226,6 +226,111 @@ void precision_dispatch_real_complex(Function fn, const LinOp *alpha,
 }
 
 
+/**
+ * Calls the given function with each given argument LinOp
+ * converted into matrix::Dense<ValueType> as parameters.
+ *
+ * If GINKGO_MIXED_PRECISION is defined, this means that the function will be
+ * called with its dynamic type with its static type, so the (templated/generic)
```
pratikvn, line 234:
```suggestion
 * called with its dynamic type instead of with its static type, so the (templated/generic)
```

pratikvn (approved):
LGTM! Only a minor thing about CI job defaults.

include/ginkgo/core/base/precision_dispatch.hpp
upsj: I think I made a small typo there, what I meant to write is
```suggestion
 * called with its dynamic type as a static type, so the (templated/generic)
```

.gitlab-ci.yml
upsj: The reason I chose this default is that `MIXED_PRECISION` has two impacts:
1. it limits the number of kernels that will be instantiated to a smaller number
2. it changes the `core` implementation to use only the same-precision kernels.
 
To test that everything works correctly, we need to check that both parts work correctly. If the device kernel compile with `MIXED_PRECISION=ON`, then they will also compile with `OFF`, and the kernels might have their own mixed-precision quirks, so I would like to test them on different platforms.

When testing with `MIXED_PRECISION=OFF`, we only test that `make_temporary_conversion` and the non-mixed precision kernels work correctly. Since this behavior is already tested elsewhere, I wanted the focus to be on mixed-precision kernels, with only one instantiation of the conversion-based mode for each compiler vendor to be sure it works. I should do the same for Windows and MacOS, though!

CMakeLists.txt
upsj: Maybe even 
```suggestion
option(GINKGO_MIXED_PRECISION "Instantiate true mixed-precision kernels (otherwise they will be conversion-based using implicit temporary storage)" OFF)
```

core/matrix/ell.cpp
```diff
@@ -101,39 +102,29 @@ size_type calculate_max_nnz_per_row(
...
+        [&](auto dense_b, auto dense_x) {
+            auto converted_alpha = make_temporary_conversion<ValueType>(alpha);
+            auto converted_beta =
+                make_temporary_conversion<typename std::remove_reference_t<
+                    decltype(*dense_x)>::value_type>(beta);
+            this->get_executor()->run(
+                ell::make_advanced_spmv(converted_alpha.get(), this, dense_b,
+                                        converted_beta.get(), dense_x));
+        },
+        b, x);
```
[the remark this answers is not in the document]
upsj: Oh right, this question was actually for me (and is still relevant for the current state, even though the above code is outdated).

We are doing two different things here:
1. `dense_b` and `dense_x` are `x` and `b` cast "from their dynamic type to their static type", i.e. if `x` is `Dense<float>`, then the lambda instantiation with `Dense<float> dense_x` parameter type is called.
2. Due to their small size (and not wanting to have 4x more parameter type combinations), the scalars are converted to the correct type instead of using the above procedure. So they do indeed need to be treated differently for this to work and make sense.

include/ginkgo/core/base/precision_dispatch.hpp
```diff
@@ -226,6 +226,100 @@ void precision_dispatch_real_complex(Function fn, const LinOp *alpha,
...
+ * Dense<next_precision<ValueType>> parameter types, and the appropriate
+ * overload will be called based on the dynamic type of the parameter.
+ *
+ * If GINKGO_MIXED_PRECISION is not defined, it will behave exactly like
+ * precision_dispatch.
+ *
+ *
+ * @param fn  the given function. It will be passed one (potentially const)
+ *            matrix::Dense<...>* parameter per parameter in the parameter
+ *            pack `linops`.
```
thoasm, line 245: Currently, you are not even using a parameter pack `linops`, so this is wrong either way.

Also nit: That is a lot of `parameter`
Maybe "[...] parameter per argument in the parameter pack `linops`" (if you would use a parameter pack).

include/ginkgo/core/base/precision_dispatch.hpp
```diff
@@ -226,6 +226,100 @@ void precision_dispatch_real_complex(Function fn, const LinOp *alpha,
...
+ * @param linops  the given arguments to be converted and passed on to fn.
```
thoasm, line 246: This parameter does not exist. You are not using parameter packs here, you are only using `in` and `out` as parameters.

include/ginkgo/core/base/precision_dispatch.hpp
```diff
@@ -226,6 +226,100 @@ void precision_dispatch_real_complex(Function fn, const LinOp *alpha,
...
+#else
+    precision_dispatch<ValueType>(fn, in, out);
+#endif
+}
+
+
+/**
+ * Calls the given function with the given LinOps cast to their dynamic type
+ * matrix::Dense<ValueType>* as parameters.
+ * If ValueType is real and both `in` and `out` are complex, uses
```
thoasm, line 287: You are only testing if `in` is complex.

include/ginkgo/core/base/precision_dispatch.hpp
```diff
@@ -302,13 +302,24 @@ void mixed_precision_dispatch_real_complex(Function fn, const LinOp *in,
     if (complex_to_real) {
         mixed_precision_dispatch<to_complex<ValueType>>(
             [&fn](auto dense_in, auto dense_out) {
-                using Dense = matrix::Dense<ValueType>;
+                using DenseIn = typename std::decay_t<decltype(*dense_in)>;
+                using DenseOut = typename std::decay_t<decltype(*dense_out)>;
+                using DenseDummy = matrix::Dense<ValueType>;
```
thoasm, line 307: Maybe add a comment that you need `DenseDummy`, so it compiles properly (at least I think that is the reason).
Additionally, think about using SFINAE on this whole function `mixed_precision_dispatch_real_complex` to ensure that `ValueType` is complex, or not. IMO, this would make it more readable and avoid the work-around of the `dynamic_cast` (and potentially the `DenseDummy`).

thoasm (approved):
LGTM.
I have some documentation corrections and some nits about your work-around, but nothing serious.

thoasm:
@upsj Additionally, could you remove the listed CodeSmells? Basically, 4x removing the unused variable `num_rows` for reference and OMP ELL.

include/ginkgo/core/base/precision_dispatch.hpp
upsj: The test for `out` is well-hidden and happens kind of implicitly -- within the `mixed_precision_dispatch`, which attempts to cast to all value types reachable via `next_precision`.

include/ginkgo/core/base/precision_dispatch.hpp
upsj: Good point. For the older implementation, this direct cast solution is preferable, but for this more complex setting, SFINAE/enable_if seems preferable

cmake/get_info.cmake
upsj: Good point, though I think it doesn't fit into "Test, benchmark and examples" either, so I created a new heading

benchmark/utils/formats.hpp
```diff
@@ -59,7 +59,8 @@ namespace formats {
 
 
 std::string available_format =
-    "coo, csr, ell, sellp, hybrid, hybrid0, hybrid25, hybrid33, hybrid40, "
+    "coo, csr, ell, fell, sellp, hybrid, hybrid0, hybrid25, hybrid33, "
```
yhmtsai, line 62:
```suggestion
    "coo, csr, ell, ell-mixed, sellp, hybrid, hybrid0, hybrid25, hybrid33, "
```

benchmark/utils/formats.hpp
```diff
@@ -90,6 +91,8 @@ std::string format_description =
     "csrm: Ginkgo's CSR implementation with merge_path strategy.\n"
     "ell: Ellpack format according to Bell and Garland: Efficient Sparse "
     "Matrix-Vector Multiplication on CUDA.\n"
+    "fell: Mixed Precision Ellpack format according to Bell and Garland: "
```
yhmtsai, line 94:
```suggestion
    "ell-mixed: Mixed Precision Ellpack format according to Bell and Garland: "
```

yhmtsai (commented):
minor changes on fell->ell-mixed

benchmark/utils/formats.hpp
```diff
@@ -204,6 +207,23 @@ const std::map<std::string, std::function<std::unique_ptr<gko::LinOp>(
         {"csrc", READ_MATRIX(csr, std::make_shared<csr::classical>())},
         {"coo", read_matrix_from_data<gko::matrix::Coo<etype>>},
         {"ell", read_matrix_from_data<gko::matrix::Ell<etype>>},
+        {"ell_mixed",
```
upsj, line 210:
```suggestion
        {"ell-mixed",
```

include/ginkgo/core/base/precision_dispatch.hpp
```diff
@@ -226,6 +226,110 @@ void precision_dispatch_real_complex(Function fn, const LinOp *alpha,
...
+ * If ValueType is real and both `in` and `out` are complex, uses
+ * matrix::Dense::get_real_view() to convert them into real matrices after
+ * precision conversion.
+ *
+ * @see mixed_precision_dispatch()
+ */
+template <typename ValueType, typename Function,
+          std::enable_if_t<is_complex<ValueType>()> * = nullptr>
+void mixed_precision_dispatch_real_complex(Function fn, const LinOp *in,
+                                           LinOp *out)
```
thoasm, line 301: nit: Why don't you use the return type for SFINAE? IMO, it looks a bit cleaner:
```suggestion
template <typename ValueType, typename Function>
std::enable_if_t<is_complex<ValueType>()> 
    mixed_precision_dispatch_real_complex(Function fn, const LinOp *in,
                                          LinOp *out)
```

include/ginkgo/core/base/precision_dispatch.hpp
```diff
@@ -226,6 +226,110 @@ void precision_dispatch_real_complex(Function fn, const LinOp *alpha,
...
+#else
+    precision_dispatch<ValueType>(fn, in, out);
+#endif
+}
+
+
+template <typename ValueType, typename Function,
+          std::enable_if_t<!is_complex<ValueType>()> * = nullptr>
+void mixed_precision_dispatch_real_complex(Function fn, const LinOp *in,
+                                           LinOp *out)
```
thoasm, line 314: Same here:
```suggestion
template <typename ValueType, typename Function>
std::enable_if_t<!is_complex<ValueType>()>
    mixed_precision_dispatch_real_complex(Function fn, const LinOp *in,
                                          LinOp *out)
```

core/base/mixed_precision_types.hpp
```diff
@@ -0,0 +1,108 @@
...
+    template _macro(std::complex<float>, std::complex<double>,      \
+                    std::complex<double>, int32);                   \
+    template _macro(std::complex<double>, std::complex<float>,      \
+                    std::complex<float>, int32);                    \
+    template _macro(std::complex<double>, std::complex<float>,      \
+                    std::complex<double>, int32);                   \
+    template _macro(std::complex<double>, std::complex<double>,     \
+                    std::complex<float>, int32);                    \
+    template _macro(std::complex<double>, std::complex<double>,     \
+                    std::complex<double>, int32);                   \
```
tcojean, line 66: Minor, this could be one macro with an `int_type` parameter such that there is no repetition in all these combinations.

thoasm (approved):
Minor nits about SFINAE style.

tcojean (approved):
LGTM!

common/matrix/ell_kernels.hpp.inc
```diff
@@ -34,29 +34,30 @@ namespace kernel {
...
 
-template <int num_thread_per_worker, bool atomic, typename ValueType,
-          typename IndexType, typename Closure>
+template <int num_thread_per_worker, bool atomic, typename b_accessor,
+          typename a_accessor, typename OutputValueType, typename IndexType,
+          typename Closure>
 __device__ void spmv_kernel(
     const size_type num_rows, const int num_worker_per_row,
-    const ValueType *__restrict__ val, const IndexType *__restrict__ col,
+    acc::range<a_accessor> val, const IndexType *__restrict__ col,
```
yhmtsai, line 42: after using accessor, it can not have `const *__restrict__` anymore, right?

core/base/mixed_precision_types.hpp
```diff
@@ -0,0 +1,108 @@
...
+    template _macro(float, float, float, int64);                    \
+    template _macro(float, float, double, int64);                   \
+    template _macro(float, double, float, int64);                   \
+    template _macro(float, double, double, int64);                  \
+    template _macro(double, float, float, int64);                   \
```
yhmtsai, line 71: Maybe I remember something wrong.
Does accessor already have storing value in double and using value as float?
or it only miss benchmark there

core/base/mixed_precision_types.hpp
upsj: Great idea! Then we can also repurpose it for Dense stuff

yhmtsai (approved):
LGTM.
If we consider row_major is left_major and use the left is leading index, we can use the same accessor for col_major.
it will use vals(col_idx, row_idx) for col_major.
Is it too misleading?

core/base/mixed_precision_types.hpp
upsj: No, we did not benchmark this. But I think in case there were some performance issues, this should be fixed in the accessor :)

common/matrix/ell_kernels.hpp.inc
upsj: Yes, but I think the accessor itself uses `__restrict__`, so we shouldn't need to worry about it.

include/ginkgo/core/base/precision_dispatch.hpp
upsj: Sadly, [MSVC doesn't like this at all](https://github.com/ginkgo-project/ginkgo/runs/2499508262?check_suite_focus=true), so I'll have to go with the original solution.

PR #718, opened 2021-03-10
Give `create_submatrix` a virtual behavior.

tcojean:
This is useful for users to be able to derive from Dense in order to
specify the behavior of our solvers (i.e., GMRES).

### DIFF
--- a/include/ginkgo/core/matrix/dense.hpp
+++ b/include/ginkgo/core/matrix/dense.hpp
@@ -510,19 +510,7 @@ public:
                                             const span &columns,
                                             const size_type stride)
     {
-        row_major_range range_this{this->get_values(), this->get_size()[0],
-                                   this->get_size()[1], this->get_stride()};
-        auto range_result = range_this(rows, columns);
-        // TODO: can result in HUGE padding - which will be copied with the
-        // vector
-        return Dense::create(
-            this->get_executor(),
-            dim<2>{range_result.length(0), range_result.length(1)},
-            Array<ValueType>::view(
-                this->get_executor(),
-                range_result.length(0) * range_this.length(1) - columns.begin,
-                range_result->data),
-            stride);
+        return this->create_submatrix_impl(rows, columns, stride);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


tcojean:
My understanding of this issue is that for the purpose of @nbeams integration of Ginkgo with MFEM it should be enough to copy/paste the implementation of `create_submatrix_impl` into her own wrapper since no actual allocation or anything specific is required, all she wants is to obtain the proper type (her type).

yhmtsai (approved):
LGTM.
Maybe I remember something wrong or there is already alternative way to work.
gmres also needs `create_with_same_config_of` for more columns.

pratikvn (commented):
Would this be enough ? Is there already a definition of `span` that can be used externally ? If not we can provide an additional function with expanded arguments (`row_start`, `row_end`, `col_start` , `col_end`) instead of a `span` ?

tcojean:
format!

tcojean:
Your assumptions are correct Natalie, but I think you don't need to actually "create an MFEM vector" in the `create_submatrix_impl`, what this does is only return a Dense which reuses the same data but through a different shape. Hopefully you can simply copy/paste the code that is in Dense for that function except you would create a `VectorWrapper` instead. All you need is that it uses the VectorWrapper type I think right? MFEM doesn't need to see that part necessarily (I think?)

tcojean:
format!

nbeams:
No, I do need to create an MFEM Vector.  The problem right now happens when I have an MFEM matrix-free operator, which means MFEM will handle the matvec itself, in which case I need an MFEM Vector.

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -161,6 +161,26 @@ class Dense
...
+     * @param size  The size of the new matrix.
+     * @param stride  The stride of the new matrix.
+     *
+     * @returns a Dense matrix with the type of other.
+     */
+    static std::unique_ptr<Dense> create_with_type_of(
+        const Dense *other, std::shared_ptr<const Executor> exec,
+        const dim<2> &size = dim<2>{}, size_type stride = 0)
+    {
+        auto used_stride = stride == 0 ? size[1] : stride;
```
upsj, line 179: From the interface documentation view, this might be a bit confusing, how about having two overloads (one with size = dim<2>{}, one with all parameters, since value 0 as a special indicator for the stride to me is less clear than indicating this by the presence/absense of a parameter.

upsj (approved):
LGTM!

include/ginkgo/core/matrix/dense.hpp
tcojean: I added an extra overload which takes all parameters whereas the other ones still uses a default `dim<2>{}` if needed but takes no stride.

tcojean:
If @nbeams can confirm that technically there is everything needed for the purpose of the MFEM-Ginkgo integration, I think this would be ready to be merged.

include/ginkgo/core/matrix/dense.hpp
upsj: That's exactly what I meant, thanks :)

thoasm (approved):
LGTM!

pratikvn (approved):
LGTM!

PR #719, opened 2021-03-10
Mixed Precision ISAI

fritzgoebel:
This PR makes the ISAI preconditioner support reduced precision storage format.
It is based on Tobias' mp spmv approach for ell, which is slightly changed to use an accessor on both the matrix values and the input vector, for now always using the ValueType of the output vector as arithmetic type.

Points I would like to discuss are:
- The ISAI matrix is now stored in ell. Without breaking interface, we would need to convert this to csr when retrieving the preconditioner matrix with get_approximate_inverse(). 
- I introduced `GKO_INSTANTIATE_FOR_EACH_VALUE_INDEX_AND_STORAGE_TYPE` in types.hpp to make this work. I'm not really happy with having this additional to the mixed value and index types Tobias already introduced, maybe we can find a better way.
- The preconditioner matrix is stored in the precision of StorageType. When retrieving the preconditioner matrix of a preconditioner with ValueType, I'm not 100% sure which precision the returned matrix should be stored in. For now, I went with converting it to ValueType.
- For now, the storage type of the preconditioner matrix is a template. Another option would be to have it as a factory parameter somehow, for that I didn't find a nice way to do it yet though.
- To support half precision storage, I think Ell will have to fully support half.

Todos:
- [x] Wait for https://github.com/ginkgo-project/ginkgo/pull/690 to be merged
- [x] Wait for https://github.com/ginkgo-project/ginkgo/pull/708 to be merged
- [x] Wait for https://github.com/ginkgo-project/ginkgo/pull/717 to be merged

### DIFF
--- a/core/preconditioner/isai.cpp
+++ b/core/preconditioner/isai.cpp
@@ -268,16 +268,19 @@ void Isai<IsaiType, ValueType, IndexType>::generate_inverse(
 template <isai_type IsaiType, typename ValueType, typename IndexType>
 std::unique_ptr<LinOp> Isai<IsaiType, ValueType, IndexType>::transpose() const
 {
+    auto exec = this->get_executor();
     auto is_spd = IsaiType == isai_type::spd;
     if (is_spd) {
         return this->clone();
     }
 
-    std::unique_ptr<transposed_type> transp{
-        new transposed_type{this->get_executor()}};
+    std::unique_ptr<transposed_type> transp{new transposed_type{exec}};
     transp->set_size(gko::transpose(this->get_size()));
-    transp->approximate_inverse_ =
-        share(as<Csr>(this->get_approximate_inverse())->transpose());
+    auto csr_transp =
+        share(as<Csr>(as<Csr>(this->get_approximate_inverse())->transpose()));
+    auto ell_transp = Ell::create(exec);
+    csr_transp->convert_to(ell_transp.get());
+    transp->approximate_inverse_ = share(ell_transp);
 
     return std::move(transp);
 }
@@ -287,16 +290,19 @@ template <isai_type IsaiType, typename ValueType, typename IndexType>
 std::unique_ptr<LinOp> Isai<IsaiType, ValueType, IndexType>::conj_transpose()
     const
 {
+    auto exec = this->get_executor();
     auto is_spd = IsaiType == isai_type::spd;
     if (is_spd) {
         return this->clone();
     }
 
-    std::unique_ptr<transposed_type> transp{
-        new transposed_type{this->get_executor()}};
+    std::unique_ptr<transposed_type> transp{new transposed_type{exec}};
     transp->set_size(gko::transpose(this->get_size()));
-    transp->approximate_inverse_ =
-        share(as<Csr>(this->get_approximate_inverse())->conj_transpose());
+    auto csr_transp = share(
+        as<Csr>(as<Csr>(this->get_approximate_inverse())->conj_transpose()));
+    auto ell_transp = Ell::create(exec);
+    csr_transp->convert_to(ell_transp.get());
+    transp->approximate_inverse_ = share(ell_transp);
 
     return std::move(transp);
 }
--- a/core/preconditioner/isai.cpp
+++ b/core/preconditioner/isai.cpp
@@ -117,8 +117,9 @@ std::shared_ptr<Csr> extend_sparsity(std::shared_ptr<const Executor> &exec,
 }
 
 
-template <isai_type IsaiType, typename ValueType, typename IndexType>
-void Isai<IsaiType, ValueType, IndexType>::generate_inverse(
+template <isai_type IsaiType, typename ValueType, typename IndexType,
+          typename StorageType>
+void Isai<IsaiType, ValueType, IndexType, StorageType>::generate_inverse(
     std::shared_ptr<const LinOp> input, bool skip_sorting, int power,
     IndexType excess_limit)
 {
@@ -265,8 +266,10 @@ void Isai<IsaiType, ValueType, IndexType>::generate_inverse(
 }
 
 
-template <isai_type IsaiType, typename ValueType, typename IndexType>
-std::unique_ptr<LinOp> Isai<IsaiType, ValueType, IndexType>::transpose() const
+template <isai_type IsaiType, typename ValueType, typename IndexType,
+          typename StorageType>
+std::unique_ptr<LinOp>
+Isai<IsaiType, ValueType, IndexType, StorageType>::transpose() const
 {
     auto exec = this->get_executor();
     auto is_spd = IsaiType == isai_type::spd;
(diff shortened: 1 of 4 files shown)


fritzgoebel:
format!

core/preconditioner/isai.cpp
```diff
@@ -117,8 +117,9 @@ std::shared_ptr<Csr> extend_sparsity(std::shared_ptr<const Executor> &exec,
 }
 
 
-template <isai_type IsaiType, typename ValueType, typename IndexType>
-void Isai<IsaiType, ValueType, IndexType>::generate_inverse(
+template <isai_type IsaiType, typename ValueType, typename IndexType,
+          typename StorageType>
+void Isai<IsaiType, ValueType, IndexType, StorageType>::generate_inverse(
```
thoasm, line 122: Missing documentation (in the header file) for `power` and `excess_limit`.

include/ginkgo/core/preconditioner/isai.hpp
```diff
@@ -102,15 +103,17 @@ enum struct isai_type { lower, upper, general, spd };
  * @ingroup precond
  * @ingroup LinOp
  */
-template <isai_type IsaiType, typename ValueType, typename IndexType>
-class Isai : public EnableLinOp<Isai<IsaiType, ValueType, IndexType>>,
-             public Transposable {
+template <isai_type IsaiType, typename ValueType, typename IndexType,
+          typename StorageType = ValueType>
```
thoasm, line 107: Missing documentation for `StorageType`.

core/preconditioner/isai.cpp
```diff
@@ -265,58 +266,84 @@ void Isai<IsaiType, ValueType, IndexType>::generate_inverse(
...
     }
 
-    std::unique_ptr<transposed_type> transp{
-        new transposed_type{this->get_executor()}};
+    std::unique_ptr<transposed_type> transp{new transposed_type{exec}};
     transp->set_size(gko::transpose(this->get_size()));
-    transp->approximate_inverse_ =
-        share(as<Csr>(this->get_approximate_inverse())->transpose());
+    auto csr_transp =
+        share(as<Csr>(as<Csr>(this->get_approximate_inverse())->transpose()));
```
thoasm, line 283: In general in this PR: you use too much `as<>()` and `share()`, which makes it harder to read.

include/ginkgo/core/preconditioner/isai.hpp
```diff
@@ -122,10 +125,11 @@ class Isai : public EnableLinOp<Isai<IsaiType, ValueType, IndexType>>,
                        ? isai_type::spd
                        : IsaiType == isai_type::lower ? isai_type::upper
                                                       : isai_type::lower,
-             ValueType, IndexType>;
+             ValueType, IndexType, StorageType>;
```
thoasm, line 128: `transposed_type` can be simplified by only having special treatment for `lower` and `upper`, while all others remain the same.

thoasm (changes requested):
There are some parts that I would really like to change before merging this PR.

Additionally, I realized that it might make sense to make `SpdIsai` and `GeneralIsai` separate classes or at least specialization of the current `Isai` because:
1. they seem to **not** share too much with `LowerIsai` and `UpperIsai`
2. there is a lot of branching in all functions 
3. **all** `Isai` preconditioner (including `LowerIsai` and `UpperIsai`) are stored in ELL in the current implementation
4. `get_approximate_inverse` has a lot of overhead now

thoasm:
@fritzgoebel All dependencies for this PR are merged. Can you rebase it, so we can properly review it?

include/ginkgo/core/preconditioner/isai.hpp
```diff
@@ -186,6 +199,15 @@ class Isai : public EnableLinOp<Isai<IsaiType, ValueType, IndexType>>,
...
             excess_solver_factory, nullptr);
+
+        /**
+         * @brief Parameter for reducing preconditioner storage precision.
+         *
+         * Reduces the preconditioner storage precision up to reduce_precision
+         * times. For value types double and complex<double>, storage precision
+         * can be reduced up to two times, for float and complex<float> once.
+         */
+        int GKO_FACTORY_PARAMETER_SCALAR(reduce_precision, 0);
```
tcojean, line 210: This is unused and I think also unneeded. The reason being that `StorageType` is set as template parameter anyway.

```suggestion
```

tcojean (approved):
LGTM. Some minor comments.

include/ginkgo/core/preconditioner/isai.hpp
```diff
@@ -211,15 +233,21 @@ class Isai : public EnableLinOp<Isai<IsaiType, ValueType, IndexType>>,
...
           parameters_{factory->get_parameters()}
     {
+        const auto exec = this->get_executor();
         const auto skip_sorting = parameters_.skip_sorting;
         const auto power = parameters_.sparsity_power;
         const auto excess_limit = parameters_.excess_limit;
         generate_inverse(system_matrix, skip_sorting, power, excess_limit);
+
+        auto csr_inv = as<Csr>(approximate_inverse_);
+        auto ell_inv = convert_csr_to_ell(approximate_inverse_.get());
```
tcojean, line 243: One thing to note is that `generate_inverse` creates a `Csr<ValueType, IndexType>` but `convert_csr_to_ell()` takes in a `Csr<StorageType, IndexType>`? That makes the functions `convert_csr_to_ell()` and `convert_ell_to_csr()` non symmetric and it's probably unneeded as well? It's the same thing in other places as well.

fritzgoebel: Actually, `convert_csr_to_ell()` takes in a `Csr<ValueType, IndexType>` and converts it to an `Ell<StorageType, IndexType>`. `convert_ell_to_csr()` in turn converts an `Ell<StorageType, IndexType>` to a `Csr<ValueType, IndexType>`.

thoasm (approved):
LGTM!

include/ginkgo/core/preconditioner/isai.hpp
tcojean: Ah right, I missed some lines. Thanks for the explanation.

tcojean (approved):
LGTM.

include/ginkgo/core/preconditioner/isai.hpp
```diff
@@ -95,37 +96,39 @@ enum struct isai_type { lower, upper, general, spd };
...
-        Isai<IsaiType == isai_type::general
-                 ? isai_type::general
-                 : IsaiType == isai_type::spd
-                       ? isai_type::spd
-                       : IsaiType == isai_type::lower ? isai_type::upper
-                                                      : isai_type::lower,
-             ValueType, IndexType>;
+        Isai<IsaiType == isai_type::lower
+                 ? isai_type::upper
+                 : IsaiType == isai_type::upper ? isai_type::lower : IsaiType,
```
upsj, line 125: could we move this into a
```
constexpr isai_type transpose(isai_type) {
    
}
```
function?

upsj (commented):
LGTM in general. To be honest though, I would rather want to wait for full mixed-precision support in Csr instead of choosing Ell just because it is the only type where we put in the effort to make it mixed-precision capable. Coo would probably also be alternative with low development effort for the changes, and could provide much better worst-case performance.

fritzgoebel:
Yes, I will close it.

PR #720, opened 2021-03-11
Simplify Windows CMake setup + Stand-alone examples

upsj:
I attempted to simplify our current Windows CMake setup a bit:

- Require OpenMP >= 3.0 to avoid OpenMP being enabled erroneously on MSVC
- Add `/bigobj` flags by default to avoid users having to specify them
- Always link dynamically against the C runtime library (CRT); If users really want static linking, they just have to specify their own `CMAKE_CXX_<CONFIG>_FLAGS` and disable tests. We could also try to identify based on the compiler flags whether we need this, but GTest decided to make this a CMake option instead of relying on the compiler flags, so this would need additional code 🤷‍♂️ 
- Install DLLs into the binary instead of library path. This is what vcpkg expects by default, and Craig Scott recommends in his book, so I guess this is the right way to go
- Create all of Ginkgo's libraries in `lib/` by default, which allows us to avoid having to specify the `PATH` in tests, instead using the `WORKING_DIRECTORY` parameter of `add_test` to allow the libraries to be found.

Finally, I turned all examples into stand-alone projects which conditionally call `find_package(Ginkgo)` if we are not in the build tree.

All of these changes are implemented in individual commits, so it should be easy to remove one of them if we consider them unnecessary or harmful.

- [x] Keep in sync with #734


yhmtsai:
Is any reason to delete the current workaround?
I do not like we leave the users to set `CMAKE_CXX_...` or `CMAKE_CUDA_...`
(You also need to modify `CMAKE_CUDA_..` when linking crt.)
CMAKE_CXX_ has some default setting, so it is hard to set it correctly with only change crt in the beginning via command.
And we still need tests when we build static ginkgo.

I prefer keeping CMAKE_CXX_FLAGS empty such that user can add anything without covering original settings.
(users still need to set the bigobj by command or env in current ginkgo though)

upsj:
**/MD, /MT**
Yes, the CRT linking issues are probably the most contentious change. Just so we are talking about the same thing:
Both dynamic library + dynamic CRT and static library + dynamic CRT work fine and can build the tests without any modifications.
GTest is the only library that sets the CRT-related flags itself instead of using what is provided in `CMAKE_CXX_<CONFIG>_FLAGS`. With 2-3 lines of code, we could also make it work for GTest.

In general, I think we should give users the freedom to choose which CRT to use, which means not changing the settings ourselves based on `BUILD_SHARED_LIBRARIES`, at the same time, it is a property that must be the same across the entire library + application. We could add something like `GINKGO_SHARED_CRT`, but that again breaks transitivity of compiler flags, which is what caused the GTest issues in the first place. Basically most of this PR is removing parts from the CMake configuration which I believe have more canonical or common ways to do this.

**/bigobj**
The ideal setting for users would be to just call `mkdir build && cd build && cmake ..`. That currently doesn't work because of the missing `/bigobj` flag. `CMAKE_CXX_FLAG` is only modified internally and the change is not visible in the `CMakeCache.txt` file. Can you explain what you mean by covering original settings?
Finally, I think ginkgo_cuda also works without the `/bigobj` flag, but I can add them of course.

yhmtsai:
thanks, I never know static library works with dynamic crt on windows.
I also did some changes on cuda runtime choice according to the shared/static.
Could you also check it? I am wondering there is another issue.

`\bigobj`:
I mean cmake -DCMAKE_CXX_FLAGS="<some-flags>" .. will lose the `big-obj` from ginkgo's cmake.

upsj:
I didn't see any issues with CUDA static/dynamic linking so far, but I will check.

On `/bigobj`: CMake variables are scoped, so everything in CMakeCache.txt is used as the input for the CMake scripts, but all changes within the scripts are only temporary during execution, so `/bigobj` will always be added, even if it is already present in CMAKE_CXX_FLAGS.

upsj:
The build itself seems to work, even though the CUDA tests segfault during execution. This is the same as a few months ago when I tried to build on Windows first, so I think this is not related to the build configuration. As soon as I manage to fix my CUDA installation, I'll try to identify the issue.

cmake/create_test.cmake
```diff
@@ -14,7 +14,9 @@ function(ginkgo_create_test test_name)
         target_link_libraries(${TEST_TARGET_NAME} PRIVATE "${GINKGO_CIRCULAR_DEPS_FLAGS}")
     endif()
     target_link_libraries(${TEST_TARGET_NAME} PRIVATE ginkgo GTest::Main GTest::GTest ${ARGN})
-    add_test(NAME ${REL_BINARY_DIR}/${test_name} COMMAND ${TEST_TARGET_NAME})
+    add_test(NAME ${REL_BINARY_DIR}/${test_name}
+        COMMAND ${TEST_TARGET_NAME}
+        WORKING_DIRECTORY "$<TARGET_FILE_DIR:ginkgo>")
```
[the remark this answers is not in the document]
upsj: Either that, or use `ctest -C Release|Debug` since that uses the correct `WORKING_DIRECTORY` specified in `add_test`. As far as I can tell, that issue is only relevant to Windows, though.

examples/adaptiveprecision-blockjacobi/CMakeLists.txt
```diff
@@ -1,5 +1,12 @@
-set(target_name "adaptiveprecision-blockjacobi")
-add_executable(${target_name} ${target_name}.cpp)
-target_link_libraries(${target_name} Ginkgo::ginkgo)
-target_include_directories(${target_name} PRIVATE ${PROJECT_SOURCE_DIR})
+cmake_minimum_required(VERSION 3.9)
+project(adaptiveprecision-blockjacobi)
+
+# We only need to find Ginkgo if we build this example stand-alone
+if (NOT GINKGO_BUILD_EXAMPLES)
+    find_package(Ginkgo 1.3.0 REQUIRED)
```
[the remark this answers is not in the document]
upsj: There would be variables available, however, I would like to avoid using them here, because otherwise we cannot build the examples stand-alone.

cmake/build_helpers.cmake
```diff
@@ -26,26 +26,11 @@ function(ginkgo_compile_features name)
...
-        endif()
-        if(GINKGO_CHECK_PATH)
-            ginkgo_check_shared_library("${CMAKE_SHARED_LIBRARY_PREFIX}${name}${CMAKE_SHARED_LIBRARY_SUFFIX}")
-        endif()
-    endif()
+    # Put all shared libraries and corresponding imported libraries into the specified path
+    set_property(TARGET "${name}" PROPERTY
+        RUNTIME_OUTPUT_DIRECTORY "${GINKGO_LIBRARY_PATH}")
+    set_property(TARGET "${name}" PROPERTY
+        ARCHIVE_OUTPUT_DIRECTORY "${GINKGO_LIBRARY_PATH}")
```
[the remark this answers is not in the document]
upsj: Yes, though I don't see any harm in having a common `lib` output path, that might even make it easier for users to see which libraries we are generating.

examples/custom-stopping-criterion/CMakeLists.txt
```diff
@@ -1,12 +1,20 @@
...
-	${target_name}.cpp)
-target_link_libraries(${target_name} Ginkgo::ginkgo
+
+add_executable(custom-stopping-criterion
+	custom-stopping-criterion.cpp)
+
+target_link_libraries(custom-stopping-criterion Ginkgo::ginkgo
 	Threads::Threads)
-target_include_directories(${target_name}
-	PRIVATE ${PROJECT_SOURCE_DIR})
```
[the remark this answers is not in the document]
upsj: I am not sure about that. I removed the include since we should not use any code from `core` in the examples.

examples/ginkgo-ranges/CMakeLists.txt
```diff
@@ -1,4 +1,10 @@
...
+cmake_minimum_required(VERSION 3.9)
+project(ginkgo-ranges)
+
+# We only need to find Ginkgo if we build this example stand-alone
+if (NOT GINKGO_BUILD_EXAMPLES)
+    find_package(Ginkgo 1.3.0 REQUIRED)
+endif()
+add_executable(ginkgo-ranges ginkgo-ranges.cpp)
+target_link_libraries(ginkgo-ranges Ginkgo::ginkgo)
+target_include_directories(ginkgo-ranges PRIVATE ${PROJECT_SOURCE_DIR})
```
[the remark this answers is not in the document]
upsj: No, I just forgot it. Good catch, thanks :)

upsj:
@ginkgo-project/reviewers This PR is now up to date and ready to review again. I think it's easiest if you go through the individual commits, since there are multiple changes mixed together.

examples/adaptiveprecision-blockjacobi/CMakeLists.txt
```diff
@@ -1,5 +1,12 @@
-set(target_name "adaptiveprecision-blockjacobi")
-add_executable(${target_name} ${target_name}.cpp)
-target_link_libraries(${target_name} Ginkgo::ginkgo)
-target_include_directories(${target_name} PRIVATE ${PROJECT_SOURCE_DIR})
+cmake_minimum_required(VERSION 3.9)
+project(adaptiveprecision-blockjacobi)
+
+# We only need to find Ginkgo if we build this example stand-alone
+if (NOT GINKGO_BUILD_EXAMPLES)
+    find_package(Ginkgo 1.4.0 REQUIRED)
```
pratikvn, line 6: Is there a need to require a specific version ?

pratikvn (approved):
LGTM! Just a question whether this change of library paths affects external libraries.

examples/adaptiveprecision-blockjacobi/CMakeLists.txt
upsj: This means that we find Ginkgo with version number at least 1.4.0 and the same major version (i.e. no 2.0), so we can be sure that the interfaces we use are available.

pratikvn: Yes, I mean do we need 1.4.0 ? 1.3.0 should also work ? Because, technically these are standalone and can be used with an already existing, possibly older version of Ginkgo ?

cmake/build_helpers.cmake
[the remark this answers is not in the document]
upsj: This does not affect where Ginkgo gets installed to, so it is only relevant if somebody hard-coded the library paths, which is not a usage method we recommend. Before, it was scattered `core/libginkgo.so`, `reference/libginkgo_reference.so`, ..., now everything is in a single folder. Your question did make me realize though that we still need to update the build.sh scripts, so thanks for that!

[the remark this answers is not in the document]
pratikvn: But for example, MFEM does do the hard-coding for their Makefile scripts, so we need to be aware of that.

examples/adaptiveprecision-blockjacobi/CMakeLists.txt
upsj: That is a good point, in general I agree. However, with the addition of `DpcppExecutor` to most examples, they are no longer backwards-compatible with 1.3.0, so we need to modify it in a few places at least.

cmake/build_helpers.cmake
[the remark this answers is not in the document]
upsj: Does MFEM even use the build directory? Or does it rely on an installed Ginkgo? In the past, I had assumed the latter. I can't find any hard-coded paths like `reference` in the MFEM source, only [this](https://github.com/mfem/mfem/blob/11b3e0b4b5d21544c5c03f8d6f8f72d703676134/config/defaults.mk#L285).

examples/adaptiveprecision-blockjacobi/CMakeLists.txt
tcojean: For ease of maintainance and version bumps, can you create a shared `.cmake` file somewhere to define a variable like ${GINKGO_VERSION_REQUIRED}` or whatever the name could be? Just so that it's in one place.

You would also need to set that file as installed or it can be to the directory when building the example.

examples/adaptiveprecision-blockjacobi/CMakeLists.txt
```diff
@@ -1,5 +1,12 @@
-set(target_name "adaptiveprecision-blockjacobi")
-add_executable(${target_name} ${target_name}.cpp)
-target_link_libraries(${target_name} Ginkgo::ginkgo)
-target_include_directories(${target_name} PRIVATE ${PROJECT_SOURCE_DIR})
+cmake_minimum_required(VERSION 3.9)
```
tcojean, line 1: Ginkgo itself requires 3.13 now. I guess there is no need, but you could upgrade this as well.

examples/build-setup.sh
```diff
@@ -2,27 +2,25 @@
 
 # copy libraries
 LIBRARY_DIRS="core core/device_hooks reference omp cuda hip"
```
tcojean, line 4: This can be removed?

examples/custom-matrix-format/CMakeLists.txt
```diff
@@ -1,14 +1,16 @@
...
-    target_include_directories(${target_name} PRIVATE
-        ${PROJECT_SOURCE_DIR} ${CMAKE_CURRENT_SOURCE_DIR})
-    # workaround for clang-cuda/g++ interaction
-    set_target_properties(${target_name} PROPERTIES POSITION_INDEPENDENT_CODE ON)
+cmake_minimum_required(VERSION 3.9)
+project(custom-matrix-format CXX CUDA)
+
+# We only need to find Ginkgo if we build this example stand-alone
+if (NOT GINKGO_BUILD_EXAMPLES)
+    find_package(Ginkgo 1.4.0 REQUIRED)
```
tcojean, line 6: Don't we still need to make sure that we have `GINKGO_BUILD_CUDA AND GINKGO_BUILD_OMP`?

examples/papi-logging/CMakeLists.txt
```diff
@@ -1,9 +1,15 @@
...
-    target_include_directories(${target_name} PRIVATE ${PROJECT_SOURCE_DIR})
-    configure_file(data/A.mtx data/A.mtx COPYONLY)
-    configure_file(data/b.mtx data/b.mtx COPYONLY)
-    configure_file(data/x0.mtx data/x0.mtx COPYONLY)
+cmake_minimum_required(VERSION 3.9)
+project(papi-logging)
+
+# We only need to find Ginkgo if we build this example stand-alone
+if (NOT GINKGO_BUILD_EXAMPLES)
+    find_package(Ginkgo 1.4.0 REQUIRED)
```
tcojean, line 6: same thing, don't we need to ensure we have `GINKGO_HAVE_PAPI_SDE` ?

examples/minimal-cuda-solver/CMakeLists.txt
```diff
@@ -1,7 +1,15 @@
-set(target_name "minimal-cuda-solver")
-add_executable(${target_name} ${target_name}.cpp)
-target_link_libraries(${target_name} Ginkgo::ginkgo)
-target_include_directories(${target_name} PRIVATE ${PROJECT_SOURCE_DIR})
+cmake_minimum_required(VERSION 3.9)
+project(minimal-cuda-solver)
+
+# We only need to find Ginkgo if we build this example stand-alone
+if (NOT GINKGO_BUILD_EXAMPLES)
+    find_package(Ginkgo 1.4.0 REQUIRED)
```
tcojean, line 6: Don't we need to make sure we have `GINKGO_BUILD_CUDA`? Or we just let it fail if that's not the case?

tcojean (approved):
LGTM. Some small comments.

examples/adaptiveprecision-blockjacobi/CMakeLists.txt
upsj: My idea behind these changes is to make the CMake files as self-sufficient as possible. So I would like to avoid such external dependencies. Another point is that this way, you only need to increment the version requirement if you actually modify the example to use newer parts of Ginkgo, but whether that is useful would be another question entirely.

examples/minimal-cuda-solver/CMakeLists.txt
upsj: I would just let it fail in this case, since CUDA is not a compile-time dependency.

examples/adaptiveprecision-blockjacobi/CMakeLists.txt
tcojean: My fear is that this will easily become hard to manage and some aspects might be missed which make examples actually incompatible but we did not realize that. For that reason, I think I would pretty much always bump the version requirement if it is there.

upsj: I see. So would you be okay with keeping the values in sync manually, but leaving them in here explicitly?

examples/papi-logging/CMakeLists.txt
```diff
@@ -6,6 +6,10 @@ if (NOT GINKGO_BUILD_EXAMPLES)
     find_package(Ginkgo 1.4.0 REQUIRED)
 endif()
 
+if (NOT GINKGO_HAVE_PAPI_SDE)
+    message(FATAL_ERROR "This example needs Ginkgo built with PAPI support")
+endif()
```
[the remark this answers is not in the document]
tcojean: My bad forget it, the file is ignored altogether.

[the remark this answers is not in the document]
upsj: You mean without PAPI support? Then the subdirectory will not be included in the parent CMakeLists.txt

examples/adaptiveprecision-blockjacobi/CMakeLists.txt
tcojean: I think that's still somewhat error prone particularly as the number of examples increase, but if you think there is no easy way to integrate a macro in that workflow, fine then.

.github/workflows/windows-msvc-ref.yml
```diff
@@ -32,16 +32,15 @@ jobs:
 
     - name: configure
       run: |
-        $env:PATH="$env:PATH;$pwd\build\windows_shared_library"
         mkdir build
         cd build
-        cmake  -DCMAKE_CXX_FLAGS=/bigobj -DBUILD_SHARED_LIBS=${{ matrix.config.shared }} -DCMAKE_CXX_FLAGS_DEBUG="/MDd /Zi /Ob1 /Od /RTC1" -DGINKGO_BUILD_CUDA=OFF -DGINKGO_BUILD_OMP=OFF -DGINKGO_MIXED_PRECISION=${{ matrix.config.mixed }} ..
+        cmake  -DBUILD_SHARED_LIBS=${{ matrix.config.shared }} -DCMAKE_CXX_FLAGS_DEBUG="/MDd /Zi /Ob1 /Od /RTC1" -DGINKGO_BUILD_CUDA=OFF -DGINKGO_BUILD_OMP=OFF -DGINKGO_MIXED_PRECISION=${{ matrix.config.mixed }} ..
```
yhmtsai, line 37:
```suggestion
        cmake  -DBUILD_SHARED_LIBS=${{ matrix.config.shared }} -DCMAKE_CXX_FLAGS_DEBUG="/MTd /Zi /Ob1 /Od /RTC1" -DGINKGO_BUILD_CUDA=OFF -DGINKGO_BUILD_OMP=OFF -DGINKGO_MIXED_PRECISION=${{ matrix.config.mixed }} ..
```
such that it test static

upsj: Unfortunately, that won't work because GTest does not respect the CXX_FLAGS CRT settings

yhmtsai (approved):
LGTM. I still have some concern build ginkgo with static runtime library.
I do not have the usage case about requirement of static runtime library on Windows.
It simplifies CMakeFiles a lot, so I am fine with the current one.

PR #721, opened 2021-03-12
Improve test install

tcojean:
Provide test_install improvements in multiple ways.

+ No more file duplication, everything uses one file.
+ Test most SpMV and solvers when building for accelerators (check against reference).
+ Keep multiple dedicated targets (`test_install_hip` and `test_install_cuda`) for simplicity of running tests in spack.
+ Install the test install into the directory `${CMAKE_INSTALL_PREFIX}/smoke_tests`.
+ Use proper RPATH information when installing the binaries such that they work without environment settings.

There is also some irrelevant Executor documentation fix.

TODO:
+ [x] - Install all executable and data (keep all RPATH information if that isn't done).
+ [x] - Ensure this works in all cases (in particular when building both CUDA and HIP).


tcojean:
format!

tcojean:
format!

tcojean:
format!

test_install/CMakeLists.txt
```diff
@@ -33,29 +43,55 @@ if(GINKGO_BUILD_CUDA)
...
-    set_source_files_properties(test_install.hip.cpp PROPERTIES HIP_SOURCE_PROPERTY_FORMAT TRUE)
+    set_source_files_properties(test_install.cpp PROPERTIES HIP_SOURCE_PROPERTY_FORMAT TRUE)
     # The library was compiled with `-fPIC` when in shared library mode
     if (GINKGO_BUILD_SHARED_LIBS)
         set (GINKGO_PIC_OPTION "-fPIC")
         set (GINKGO_CUDA_PIC_OPTION "-Xcompiler '-fPIC'")
     endif()
-    hip_add_executable(test_install_hip test_install.hip.cpp
+    if (CMAKE_CUDA_COMPILER_ID STREQUAL "NVIDIA")
+        set(TESTINSTALL_CUDA_ARCH_FLAGS "${GINKGO_CUDA_ARCH_FLAGS}")
```
yhmtsai, line 64: Do you want to use the minimal flags on test_install?
otherwise, maybe reuse `GINKGO_HIP_NVCC_OPTIONS`

test_install/CMakeLists.txt
```diff
@@ -33,29 +43,55 @@ if(GINKGO_BUILD_CUDA)
             ginkgo_switch_to_windows_static("CUDA")
         endif()
     endif()
-    add_executable(test_install_cuda test_install_cuda.cu)
+    set_source_files_properties(test_install.cpp PROPERTIES LANGUAGE CUDA)
+    add_executable(test_install_cuda test_install.cpp)
+    target_compile_options(test_install_cuda
+        PRIVATE "$<$<COMPILE_LANGUAGE:CUDA>:${GINKGO_CUDA_ARCH_FLAGS}>")
+    target_compile_definitions(test_install_cuda PRIVATE HAS_CUDA=1)
```
yhmtsai, line 50: Is the HAS_REFERENCE still declared in it?

test_install/CMakeLists.txt
```diff
@@ -33,29 +43,55 @@ if(GINKGO_BUILD_CUDA)
...
+        set(TESTINSTALL_CUDA_ARCH_FLAGS "${GINKGO_CUDA_ARCH_FLAGS}")
+    endif()
+    hip_add_executable(test_install_hip test_install.cpp
         HIPCC_OPTIONS "-std=c++14"
         CLANG_OPTIONS "${GINKGO_PIC_OPTION}"
-        NVCC_OPTIONS "${GINKGO_CUDA_PIC_OPTION}")
+        NVCC_OPTIONS "${GINKGO_CUDA_PIC_OPTION}" "${TESTINSTALL_CUDA_ARCH_FLAGS}")
 
     target_link_libraries(test_install_hip PRIVATE Ginkgo::ginkgo)
+    target_compile_definitions(test_install_hip PRIVATE HAS_HIP=1)
```
yhmtsai, line 72: also here

test_install/test_install.cpp
```diff
@@ -42,13 +44,114 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+void check_solver(std::shared_ptr<gko::Executor> exec,
+                  gko::matrix_data<double> &A_raw, gko::matrix::Dense<> *b,
+                  gko::matrix::Dense<> *x)
+{
+    using Mtx = gko::matrix::Csr<>;
+    auto A =
+        gko::share(Mtx::create(exec, std::make_shared<Mtx::load_balance>()));
+#if HAS_REFERENCE
+    A->read(A_raw);
+#endif  // HAS_REFERENCE
```
yhmtsai, line 94: move it to the place before x_clone to reduce the #if condition

test_install/test_install.cpp
```diff
@@ -42,13 +44,114 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+                gko::stop::ResidualNorm<>::build()
+                    .with_reduction_factor(reduction_factor)
+                    .on(exec))
+            .on(exec);
+#if HAS_REFERENCE
+    auto x_clone = gko::clone(x);
+    solver_gen->generate(A)->apply(b, gko::lend(x_clone));
+
+#if defined(HAS_HIP) || defined(HAS_CUDA)
+    auto A_ref =
```
yhmtsai, line 112: the thing with `_ref` are different.

test_install/test_install.cpp
```diff
@@ -42,13 +44,114 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+        gko::share(Mtx::create(exec, std::make_shared<Mtx::load_balance>()));
+    A_ref->read(A_raw);
+    auto refExec = exec->get_master();
```
yhmtsai, line 115:
```suggestion
    auto exec_ref= exec->get_master();
```

test_install/test_install.cpp
```diff
@@ -42,13 +44,114 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+                    exec),
+                gko::stop::ResidualNorm<>::build()
+                    .with_reduction_factor(reduction_factor)
+                    .on(refExec))
+            .on(refExec);
+    auto x_ref = gko::clone(exec->get_master(), x);
+    solver_gen->generate(A_ref)->apply(b, gko::lend(x_ref));
+
+    auto x_clone_ref = gko::clone(exec->get_master(), gko::lend(x_clone));
+    assert_similar_matrices(gko::lend(x_clone_ref), gko::lend(x_ref), 1e-12);
```
yhmtsai, line 129: Does it only test the solver on reference and the copy function on CUDA/HIP?

test_install/test_install.cpp
```diff
@@ -42,13 +44,114 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+    test->apply(b, gko::lend(x_clone));
+
+#if defined(HAS_HIP) || defined(HAS_CUDA)
+    auto test_ref = Mtx::create(exec->get_master());
+    auto x_ref = gko::clone(exec->get_master(), x);
+    test_ref->read(A_raw);
+    test_ref->apply(b, gko::lend(x_ref));
+
+    auto x_clone_ref = gko::clone(exec->get_master(), gko::lend(x_clone));
+    assert_similar_matrices(gko::lend(x_clone_ref), gko::lend(x_ref), 1e-14);
```
yhmtsai, line 78: the same question as check_solver

yhmtsai (changes requested):
some questions in comments. Does it also need `HAS_OMP` test?

test_install/CMakeLists.txt
tcojean: There is a bug here. `hipcc` always uses `nvcc`, even if you set `CMAKE_CUDA_COMPILER` to be clang. If you used `clang` for compiling Ginkgo, `${GINKGO_CUDA_ARCH_FLAGS}` will work only with `clang` and `hipcc` will fail to compile because the flags are for `clang` and not `nvcc`. On the other hand, this is such a corner case I think it's not worth the trouble of exporting the CudaArchitectureSelector to the test_install.

test_install/test_install.cpp
tcojean: I only compare CUDA/HIP to reference. Reference launches the kernels but does not check if they are correct. It's the same for the solvers.

test_install/test_install.cpp
tcojean: What do you mean here?

yhmtsai:
format!

yhmtsai:
The format and check format issue should be related to our check list.
The error is from empty changed file.

test_install/test_install.cpp
yhmtsai: doesn't running the solver on cuda/hip?
The smoke test only involve the copy/allocation on devices side. I wonder whether it is too small on the test.

yhmtsai (approved):
LGTM

test_install/test_install.cpp
```diff
@@ -192,202 +305,177 @@ int main(int, char **)
...
 
     // core/solver/ir.hpp
     {
         using Solver = gko::solver::Ir<>;
-        auto test = Solver::build()
-                        .with_criteria(
-                            gko::stop::Iteration::build().with_max_iters(1u).on(
-                                refExec))
-                        .on(refExec);
+        auto test =
```
pratikvn, line 433: Can we also not test with IR with some inner solver ?

pratikvn (approved):
LGTM!

test_install/test_install.cpp
tcojean: We would need a special case just for IR, since the purpose isn't to cover everything but instead to make sure most stuff work, I found it not really required to test IR here. If multiple people think we should do it, I can add it.

include/ginkgo/core/base/executor.hpp
```diff
@@ -342,7 +342,7 @@ private:                                                                     \
             ::gko::syn::as_list<::gko::syn::range<0, sizeof...(Args)>>;       \
                                                                               \
     public:                                                                   \
-        explicit _name##_operation(Args &&... args)                           \
+        explicit _name##_operation(Args &&...args)                            \
```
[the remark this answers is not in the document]
tcojean: Yes, somehow my clang-format 11.0.1 keeps messing up these `&&...`.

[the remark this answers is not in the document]
upsj: And it seems that clang-format-8 is missing from the Github Action image?

test_install/CMakeLists.txt
```diff
@@ -33,29 +43,57 @@ if(GINKGO_BUILD_CUDA)
             ginkgo_switch_to_windows_static("CUDA")
         endif()
     endif()
-    add_executable(test_install_cuda test_install_cuda.cu)
+    configure_file(test_install.cpp test_install.cu COPYONLY)
+    add_executable(test_install_cuda test_install.cu)
```
upsj, line 47: Does this place (and use) the file in the source or binary dir?

upsj (approved):
LGTM!

test_install/CMakeLists.txt
tcojean: This uses the one in the binary dir without issue. Possibly since there is no `test_install.cu` in the source dir? I was surprised about that as well.

include/ginkgo/core/base/executor.hpp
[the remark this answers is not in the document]
tcojean: I think the issue is that I deleted files, and the script isn't able to cope with that. See Mike's comment earlier. https://github.com/ginkgo-project/ginkgo/pull/721#issuecomment-808256826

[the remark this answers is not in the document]
tcojean: One way to fix this issue might be to append the parameter `&hide-deleted-files=true` to the api request here: https://github.com/ginkgo-project/ginkgo/blob/develop/.github/bot-pr-base.sh#L69.

I did not check whether that fixes the issue though, and I'm also not sure how to properly test the bot setup without doing it on a separate repository?

test_install/test_install.cpp
pratikvn: Yeah, I guess it is not really required as we test separately it anyway.

include/ginkgo/core/base/executor.hpp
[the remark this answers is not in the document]
upsj: That's what I did in the past, I can reactivate the repo if possible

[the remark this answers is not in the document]
tcojean: It seems like that option isn't available in the API itself. But `jq` produces a similar result:
`jq -er '.[] | select(.status != "removed") | .filename'`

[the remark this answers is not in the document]
upsj: I have the repo set up and can test it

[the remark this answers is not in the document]
upsj: Looks like it's two issues in one: clang-format-8 got removed and the deleted file being listed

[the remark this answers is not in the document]
yhmtsai: If you would like to test it on local environment, you only need to delete the TOKEN and set the number manually.
Unauthentic API can be used 60/hour.
If you need post action (not this case), you need the repo or the token to test the action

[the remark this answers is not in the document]
upsj: https://github.com/upsj/ginkgo-bot-playground/commit/29ebb83f42246c7620cd09b40060a1e515818527
This seems to fix the issues, thanks for the suggestion Terry!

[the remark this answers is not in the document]
yhmtsai: for example, the L69 will be
`curl -X GET -s -H "Accept: application/vnd.github.v3+json" "https://api.github.com/repos/ginkgo-project/ginkgo/pulls/721/files?&per_page=100&page=1" | jq -er '.[] | .filename' || :`
and get
```
.gitlab-ci.yml
CMakeLists.txt
include/ginkgo/core/base/executor.hpp
test_exportbuild/CMakeLists.txt
test_install/CMakeLists.txt
test_install/data/A.mtx
test_install/data/b.mtx
test_install/data/x0.mtx
test_install/test_install.cpp
test_install/test_install.hip.cpp
test_install/test_install_cuda.cu

File: test_install/CMakeLists.txt (hunk 11 above)
Comment (upsj) [reply test_install/CMakeLists.txt:47]: Just to be sure can we maybe use `CMAKE_CURRENT_BINARY_DIR` instead? otherwise this might be ambiguous if for some reason you still have the old file lying around.

File: include/ginkgo/core/base/executor.hpp (hunk 10 above)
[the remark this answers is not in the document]
Comment (tcojean) [reply include/ginkgo/core/base/executor.hpp:345]: Do you want to create a tiny separate PR for the github action fixes @upsj, or would you rather these changes are carried through an existing PR? It's the same question as the git-cmake-format changes.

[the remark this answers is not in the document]
Comment (upsj) [reply include/ginkgo/core/base/executor.hpp:345]: I'll make a PR

File: test_install/test_install.cpp [hunk 12]
```diff
@@ -42,13 +44,128 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+
+
+template <typename Solver>
+void check_solver(std::shared_ptr<gko::Executor> exec,
+                  gko::matrix_data<double> &A_raw, gko::matrix::Dense<> *b,
+                  gko::matrix::Dense<> *x)
+{
+    using Mtx = gko::matrix::Csr<>;
+    auto A =
+        gko::share(Mtx::create(exec, std::make_shared<Mtx::load_balance>()));
```
thoasm, line 96: nit: `std::make_shared` requires the header `memory`.

test_install/test_install.cpp
```diff
@@ -42,13 +44,128 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+#endif  // HAS_REFERENCE
+}
+
+
 // core/base/polymorphic_object.hpp
 class PolymorphicObjectTest : public gko::PolymorphicObject {};
 
 
 int main(int, char **)
 {
```
thoasm, line 147: nit: Since we don't use the parameter, can we simply change it to:
```suggestion
int main()
{
```

thoasm (approved):
LGTM!
I only have minor nits.

tcojean:
format!

tcojean:
format!

PR #724, opened 2021-03-23
Add a publications dropdown in doxygen documentation page

pratikvn:
This PR adds a dropdown for Ginkgo's publications, which can showcase the item that is the [white paper of Ginkgo rendered nicely.](https://www.arxiv-vanity.com/papers/2006.16852/) 

A screenshot

### DIFF
--- a/doc/DoxygenLayout.xml
+++ b/doc/DoxygenLayout.xml
@@ -5,7 +5,11 @@
     <tab type="mainpage" visible="yes" title=""/>
     <tab type="user" visible="yes" title="Tutorial" url="https://github.com/ginkgo-project/ginkgo/wiki/Tutorial:-Building-a-Poisson-Solver" />
     <tab type="user" visible="yes" title="Examples" url="@ref Examples" />
-    <tab type="user" visible="yes" title="Citing Ginkgo" url="@ref citing_ginkgo" />
+    <tab type="user" visible="yes" title="Publications" url="@ref citing_ginkgo" >
+      <tab type="user" visible="yes" title="Citing Ginkgo" url="@ref citing_ginkgo" />
+      <tab type="user" visible="yes" title="Ginkgo White Paper" url="https://www.arxiv-vanity.com/papers/2006.16852/" />
+      <tab type="user" visible="yes" title="Slides and Outreach" url="https://github.com/ginkgo-project/outreach" />
+    </tab>
     <tab type="user" visible="yes" title="Contributing To Ginkgo" url="@ref contributing_guidelines" />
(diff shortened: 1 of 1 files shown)


thoasm (approved):
LGTM!

doc/DoxygenLayout.xml
```diff
@@ -5,7 +5,11 @@
     <tab type="mainpage" visible="yes" title=""/>
     <tab type="user" visible="yes" title="Tutorial" url="https://github.com/ginkgo-project/ginkgo/wiki/Tutorial:-Building-a-Poisson-Solver" />
     <tab type="user" visible="yes" title="Examples" url="@ref Examples" />
-    <tab type="user" visible="yes" title="Citing Ginkgo" url="@ref citing_ginkgo" />
+    <tab type="user" visible="yes" title="Publications" url="@ref citing_ginkgo" >
+      <tab type="user" visible="yes" title="Citing Ginkgo" url="@ref citing_ginkgo" />
+      <tab type="user" visible="yes" title="Ginkgo White Paper" url="https://www.arxiv-vanity.com/papers/2006.16852/" />
```
upsj, line 10: I looked at the formatting, and it seems like they have some issues with code formatting (they center-align the code and have fixed line numbers that don't move when horizontal-scrolling), so maybe the arxiv URL might be better?

upsj (approved):
LGTM!

doc/DoxygenLayout.xml
pratikvn: Yes, but I think overall this just allows people to read the paper directly from the browser instead of needing to download it first as a pdf. It also seems like this is optimized for mobile ? I can confirm that it looks much better on mobile.

yhmtsai: Does mobile browser can change the CSS?
Adding the `ltx_align_left` after `ltx_lst_language_c` or removing `ltx_align_center` in `<figure id="LST1"...>` looks pretty good on computer

yhmtsai (approved):
LGTM

doc/DoxygenLayout.xml
pratikvn: I am not sure what you mean. Where do you want to add the `ltx_align_left` ?

yhmtsai: image
The above figure is the result when doing the previous commit's change.
I would like to know whether it has bad impact on mobile side. (Could mobile browser modify the css?)
Moreover, we can create an issue on arxiv-vanity.

pratikvn: Ah, okay. Yeah, I am not sure if mobile browsers have inspect element or something similar. But I think it is a good idea to create an issue on arxiv vanity.

PR #725, opened 2021-03-24
Add ccache support for CMake

upsj:
This PR adds [ccache](https://ccache.dev/) as a compiler launcher if it is available. Things we would probably need to decide: When do we want to enable this? GINKGO_DEVEL_TOOLS? GINKGO_WITH_CCACHE (default to ON or OFF?) If ccache was found?

Some related reading: https://crascit.com/2016/04/09/using-ccache-with-cmake/
He suggests there using the RULE_LAUNCH_COMPILE property, though I am not if this is a good choice for us? It gives us no way to differentiate between different languages, the [language-specific variables](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_LAUNCHER.html) seem more specific. If users have CMake 3.17 or newer, they can even use an [environment variable](https://cmake.org/cmake/help/latest/envvar/CMAKE_LANG_COMPILER_LAUNCHER.html) to set it manually.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,6 +39,7 @@ option(GINKGO_EXPORT_BUILD_DIR
     OFF)
 option(GINKGO_WITH_CLANG_TIDY "Make Ginkgo call `clang-tidy` to find programming issues." OFF)
 option(GINKGO_WITH_IWYU "Make Ginkgo call `iwyu` (Include What You Use) to find include issues." OFF)
+option(GINKGO_WITH_CCACHE "Use ccache if available to speed up C++ and CUDA rebuilds by caching compilations." ON)
 option(GINKGO_CHECK_CIRCULAR_DEPS
     "Enable compile-time checks detecting circular dependencies between libraries and non-self-sufficient headers."
     OFF)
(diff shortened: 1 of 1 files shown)


yhmtsai:
what do you mean for "no way to differentiate between different languages"?
It still uses c++/nvcc as compiler for C++ and cuda, doesn't it?
I did not use ccache before, so I do not know how stability it perform.
If it does not require rebuild the whole library when disabling ccache, I think turning it on as default should be fine.
Maybe only turn it off when we use benchmark to make sure everything is updated

upsj:
I meant that when using the RULE_LAUNCH_COMPILE properties, all languages use the same launcher. Using the variable gives us more fine control over which compilers should use ccache. ccache seems to be pretty stable, their only caveats are set around race conditions when a file is created during compilation. In the worst case, you can always force a full recompile with `ccache -c`. If they encounter an unknown compiler flag, they just fall back to calling the compiler directly without caching.

yhmtsai (approved):
LGTM. thanks for explanation.
Yes, I agree with using `CMAKE_<LANG>_COMPILER_LAUNCHER ` gives the control.
but if we would like to use it on hip, do we need `RULE_LAUNCH_COMPILE` for it?
If it works for every language (up to it pass the job to compiler entirely), maybe `RULE_LAUNCH_COMPILE` is good enough.

upsj:
Since HIP currently doesn't have CMake or ccache support at all, this would probably not matter in the near time.

tcojean (approved):
LGTM.

I wouldn't go with the full `RULE_LAUNCH_COMPILE` since our setup is really complex.

thoasm (approved):
LGTM!

PR #726, opened 2021-03-26
Support ROCm 4.1.x and breaking HIP_PLATFORM change

tcojean:
Since @ROCm-Developer-Tools has broken all HIP dependent packages build in version 4.1, we need to add new platform checks to support the new variables in the future Ginkgo releases. We need to keep in mind that before this commit, all Ginkgo code will not be compatible with ROCm >= 4.1.

https://github.com/ROCm-Developer-Tools/HIP/blob/rocm-4.1.x/bin/hipvars.pm#L89-L100

- [x] TODO: check that this works by using the newly built AMD container which will use ROCM 4.1.


tcojean:
Yes we test both ROCM 4.0 (all CUDA containers) and ROCm 4.1 (AMD container).

hip/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
-if (GINKGO_HIP_PLATFORM MATCHES "nvcc" AND CMAKE_CUDA_COMPILER_VERSION VERSION_LESS 9.2)
+if ((GINKGO_HIP_PLATFORM MATCHES "nvcc" OR GINKGO_HIP_PLATFORM MATCHES "nvidia")
```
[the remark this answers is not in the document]
tcojean: That's a good idea thanks. I did the last suggestion.

.gitlab-ci.yml
```diff
@@ -545,7 +545,7 @@ build/cuda110/intel/cuda/debug/static:
 # HIP AMD
 build/amd/gcc/hip/debug/shared:
   <<: *default_build_with_test
-  image: localhost:5000/gko-amd-gnu8-llvm7
+  image: localhost:5000/intel_dgpu_6ee821ee_gko-amd-gnu8-llvm7
```
Slaedr, line 548: Do we have a common container for Intel DPCPP builds and HIP builds?

CMakeLists.txt
```diff
@@ -196,10 +198,12 @@ if(GINKGO_BUILD_HIP)
             "Set and export the environment variable HIP_PLATFORM.")
     endif()
     message(STATUS "HIP platform set to ${GINKGO_HIP_PLATFORM}")
+    set(HIP_PLATFORM_AMD_REGEX "hcc|amd")
+    set(HIP_PLATFORM_NVIDIA_REGEX "nvcc|nvidia")
 
-    if (GINKGO_HIP_PLATFORM STREQUAL "hcc")
+    if (GINKGO_HIP_PLATFORM MATCHES "${HIP_PLATFORM_AMD_REGEX}")
```
Slaedr, line 204: Nice!

INSTALL.md
```diff
@@ -220,26 +220,26 @@ imposed by the `HIP` tool suite. The variables are the following:
...
-export HIP_PLATFORM=nvcc
+export HIP_PLATFORM=nvcc # or nvidia for ROCM >= 4.1
 ```
 
-When using `HIP_PLATFORM=hcc`, note that two `HIP` compilers can be set, the old
-`hcc` or since ROCm 3.5, `clang`. Ginkgo is only compatible with the `clang`
-based installations. Although this setting should be automatically done, it is
-also possible to manually set the `HIP` compiler to `clang`:
+When using `HIP_PLATFORM=hcc` (or `amd`), note that two `HIP` compilers can be
+set, the old `hcc` or since ROCm 3.5, `clang`. Ginkgo is only compatible with
```
Slaedr, line 233:
```suggestion
set: the old `hcc`, or since ROCm 3.5, `clang`. Ginkgo is only compatible with
```

hip/CMakeLists.txt
```diff
@@ -205,8 +207,9 @@ if (GINKGO_HIP_PLATFORM MATCHES "nvcc")
     # Remove false positive CUDA warnings when calling one<T>() and zero<T>()
     list(APPEND GINKGO_HIP_NVCC_ADDITIONAL_FLAGS "--expt-relaxed-constexpr")
 
-    if (GINKGO_HIP_PLATFROM MATCHES "nvcc" AND CMAKE_CUDA_COMPILER_VERSION
-            MATCHES "9.2" AND CMAKE_CUDA_HOST_COMPILER MATCHES ".*clang.*" )
+    if ((GINKGO_HIP_PLATFROM MATCHES "nvcc" OR GINKGO_HIP_PLATFROM MATCHES "nvidia")
```
Slaedr, line 210: We can't use the new variable here?
```suggestion
    if (GINKGO_HIP_PLATFORM MATCHES "${HIP_PLATFORM_NVIDIA_REGEX}"
```

Slaedr (approved):
LGTM.

.gitlab-ci.yml
tcojean: There is no such thing. It's only that as a side effect of the containers being regenerated for the new DPC++ CI, ROCm updated as well and everything broke (pretty much).

hip/CMakeLists.txt
tcojean: Thanks, I simply missed this due to the typo.

pratikvn (approved):
LGTM! I do not see any updates to the yaml file. I guess the container names are the same but the containers themselves have been upgraded ? Maybe it is better to test against 4.0 on the amd machine as well ? At least I think we should have the container with 4.0 because I remember someone saying that 4,1 has some performance issues.

upsj (approved):

tcojean:
@pratikvn I removed the updates to the yaml file before merging. The containers weren't upgraded yet otherwise all CI would break without this PR merged first.

On the topic of testing 4.0 on the AMD system, that would require a lot more work as we use the packages provided by the package manager, and now only 4.1 is available. We would thus need to compile by hand all of AMD's tools. In addition, some packages of 4.0 are left over on the system itself.

PR #728, opened 2021-03-29
Add more data to convergence logger

pratikvn:
This PR adds the implicit residual norm to the convergence logger so that the implicit squared residual norm can be queried from a getter. It also adds a `has_converged` function which returns `true` only when the solver has converged and `false` otherwise.

It also adds a reset_convergence_status function to reset the convergence status in cases of subsequent solves.


upsj (approved):
LGTM!

tcojean (approved):
LGTM.

include/ginkgo/core/log/logger.hpp
```diff
@@ -388,6 +388,18 @@ public:                                                              \
         const uint8 &stopping_id, const bool &set_finalized,
         const Array<stopping_status> *status, const bool &one_changed,
         const bool &all_converged)
+protected:
+    virtual void on_criterion_check_completed(
```
Slaedr, line 392: Either here, or in the block above for `GKO_LOGGER_REGISTER_EVENT` for `criterion_check_completed`, it would be nice if you document what `implicit_sq_tau` is, so that people who create new solvers/wrappers know what this is about. If you add it to the block above, you could somehow include it as an optional argument by just adding a line without `\param`.

Slaedr (approved):
LGTM! This is a nice way of addressing the issue. Just one nit below.

include/ginkgo/core/log/convergence.hpp
```diff
@@ -94,6 +103,13 @@ class Convergence : public Logger {
             new Convergence(exec, enabled_events));
     }
 
+    /**
+     * Returns true if the solver has converged.
```
yhmtsai, line 107: The converge is only true when all solutions are converged, right?

yhmtsai (approved):
LGTM. only nit for the blank lines.

include/ginkgo/core/log/convergence.hpp
pratikvn: yes, but I would say that the solver has converged when only all solutions have converged.

nbeams:
Looks like just what I need for MFEM, thanks!

include/ginkgo/core/log/convergence.hpp
```diff
@@ -94,6 +102,18 @@ class Convergence : public Logger {
             new Convergence(exec, enabled_events));
     }
 
+    /**
+     * Returns true if the solver has stopped.
```
upsj, line 106: nit: this seems to be left-over from before
```suggestion
     * Returns true if the solver has converged.
```

PR #730, opened 2021-03-30
Fix for issue with corner case of reading empty Fbcsr matrix

Slaedr:
This is part of an attempt to understand an issue flagged by SonarCloud related to a null reference in the Fbcsr read function. One possibility is that it's catching the case where the input is empty, because that was causing issues which this PR fixes. Before SonarCloud is run on this PR, it is unclear whether the null reference issue will be fixed. Regardless, this small fix is good to have.

The tests related to the read functions (and all core Fbcsr tests) are Valgrind-clean.

### DIFF
--- a/core/test/matrix/fbcsr.cpp
+++ b/core/test/matrix/fbcsr.cpp
@@ -460,6 +460,24 @@ TYPED_TEST(Fbcsr, CanBeReadFromMatrixData)
 }
 
 
+TYPED_TEST(Fbcsr, CanBeReadFromEmptyMatrixData)
+{
+    using Mtx = typename TestFixture::Mtx;
+    using MtxData = typename TestFixture::MtxData;
+    auto m = Mtx::create(this->exec);
+    m->set_block_size(this->fbsample.bs);
+    MtxData mdata;
+    mdata.size = gko::dim<2>{0, 0};
+
+    m->read(mdata);
+
+    ASSERT_EQ(m->get_size(), (gko::dim<2>{0, 0}));
+    ASSERT_EQ(m->get_const_row_ptrs()[0], 0);
+    ASSERT_EQ(m->get_const_col_idxs(), nullptr);
+    ASSERT_EQ(m->get_const_values(), nullptr);
+}
+
+
 TYPED_TEST(Fbcsr, GeneratesCorrectMatrixData)
 {
     using value_type = typename TestFixture::value_type;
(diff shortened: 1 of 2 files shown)


Slaedr:
Indeed, it looks like SonarCloud detected the zero-size-input case. Pretty impressive.

core/test/matrix/fbcsr.cpp
```diff
@@ -460,6 +460,29 @@ TYPED_TEST(Fbcsr, CanBeReadFromMatrixData)
...
+    auto m = Mtx::create(this->exec);
+    m->set_block_size(this->fbsample.bs);
+    MtxData mdata;
+    mdata.size = gko::dim<2>{0, 0};
+
+    m->read(mdata);
+
+    const value_type *const vals = m->get_const_values();
+    const index_type *const rowptrs = m->get_const_row_ptrs();
+    const index_type *const colidxs = m->get_const_col_idxs();
```
pratikvn, line 478: nit: auto ? 
```suggestion
    auto vals = m->get_const_values();
    auto rowptrs = m->get_const_row_ptrs();
    auto colidxs = m->get_const_col_idxs();
```

pratikvn (approved):
LGTM!

core/test/matrix/fbcsr.cpp
```diff
@@ -460,6 +460,26 @@ TYPED_TEST(Fbcsr, CanBeReadFromMatrixData)
 }
 
 
+TYPED_TEST(Fbcsr, CanBeReadFromEmptyMatrixData)
+{
+    using Mtx = typename TestFixture::Mtx;
+    using MtxData = typename TestFixture::MtxData;
+    using value_type = typename TestFixture::value_type;
+    using index_type = typename TestFixture::index_type;
```
tcojean, line 468: Unused

thoasm (approved):
LGTM!

tcojean (approved):
LGTM

upsj (approved):
LGTM! Small nit, don't put too much time into it: I am a fan of algorithms that don't need special-case treatment for input size 0. Do you think it would be possible to modify the algorithm to avoid this? Something like setting cur_bcol to a sentinel value -1, and checking if the current non-zero is the last overall block to avoid having to write the row pointers outside of the loop

Slaedr:
Actually, yes, I agree. I guess I wanted to get the bugfix out quickly first. I'll try to come back to this again; maybe as part of the next Fbcsr PR.

PR #733, opened 2021-04-02
Add common interface for simple kernels

upsj:
## What?

This PR adds the capability to implement simple kernels with a uniform syntax across all executors:

```
exec->run_kernel([] GKO_KERNEL(auto row, auto col, auto alpha, auto x, auto y) {
                     x(row, col) = alpha(0, col) * x(row, col) + y(row, col);
                 }, x->get_size(), alpha, x, y);
```

## Why?
This might help simplify our solver porting efforts for SYCL, and also allow us to execute custom operations on our data in examples transparently across executors, like I am doing in #701 on the CPU.

## How?

It works in ~~two~~ three steps:
1. a uniform kernel dispatch approach for 1D and 2D kernels that call the provided lambda for all entries of the index range
1. mapping the input parameters `alpha, x, y` to intermediate device representations. The mappings right now are:
    * `matrix::Dense<ValueType>*` -> `matrix_accessor<ValueType>`
    * `Array<ValueType>*` -> `ValueType*`
    * other `T*` -> `T*` ~~(the disambiguation here is currently a bit hacky and limited to arithmetic types)~~
1. Mapping the device representation to kernel parameters. The mapped parameters are then passed to the kernel lambda function. This is only relevant for `compact(...)` and `vector(...)` which omit (`vector -> ValueType*`) or infer the stride from kernel dimensions (`compact -> matrix_accessor<ValueType>`)

By wrapping the input objects in wrapper types, we could implement more specific semantics like
`pointwise` for `(row, col)`, colwise for `(0, col)`, rowwise for `(row, 0)` or scalar for `(0, 0)`. This could simplify the example to
```
exec->run_kernel([] GKO_KERNEL(auto row, auto col, auto alpha, auto x, auto y) {
                     x = alpha * x + y;
                 }, x->get_size(), colwise(alpha), pointwise(x), pointwise(y));
```
or even
 ```
exec->run_kernel([] GKO_KERNEL(auto alpha, auto x, auto y) { x = alpha * x + y; },
                  x->get_size(), colwise(alpha), pointwise(x), pointwise(y));
```
if we added optional `row_index()` and `column_index()` pseudo parameters to `run_kernel`

I had to re-implement a tiny accessor replacement, since the value semantics implemented by `gko::range` don't work well in this generic context.

Finally, `GKO_KERNEL` is defined to be empty for CPU and SYCL or `__device__` for CUDA and HIP.

## TODO
- [x] Fix nvcc flags on Windows
- [x] Check if this has any performance impact:
    * The last time I checked this, I couldn't see any difference in code generation, but that might have changed?
    * ~~We now specify a different stride for each matrix parameter. Does that matter, or is constant memory fast enough anyways?~~ This is no longer true, we use compact representations now
- [x] Add tests
   - [x] Kernel launch
   - [x] Stride handling
   - [x] Division by zero handling
- [x] Add documentation

### CODE REVIEW

**Comment by upsj** [comment]
The code that triggers the compiler error on Intel:
```cpp
template <typename F, typename... Args>
void run(F f, Args... args) {
    #pragma omp parallel for
    for (unsigned i = 0; i < 100; i++) {
        f(args...);
    }
}

int main() {
    int a;
    run([](auto j) {}, a);
}
```

dpcpp/base/kernel_launch.dp.hpp
```diff
@@ -0,0 +1,176 @@
...
+
+#include <ginkgo/core/base/executor.hpp>
+#include <ginkgo/core/matrix/dense.hpp>
+
+
+#ifdef GKO_KERNEL
+#error "Only one kernel_launch.hpp file can be included at a time."
+#else
+#define GKO_KERNEL
+#endif
```
yhmtsai, line 48: Is it to avoid include different kernel_launch from different executor?

dpcpp/base/kernel_launch.dp.hpp
```diff
@@ -0,0 +1,176 @@
...
+{
+    this->get_queue()->submit([&](sycl::handler &cgh) {
+        kernels::dpcpp::generic_kernel_2d(
+            cgh, size[0], size[1], fn, kernels::dpcpp::map_to_device(args)...);
+    });
+}
+
+
+}  // namespace gko
+
```
yhmtsai, line 175:
```suggestion


```

cuda/base/kernel_launch.hpp
```diff
@@ -0,0 +1,182 @@
...
+__global__ __launch_bounds__(default_block_size) void generic_kernel_2d(
+    size_type rows, size_type cols, KernelFunction fn, KernelArgs... args)
+{
+    auto tidx = thread::get_thread_id_flat();
+    auto col = tidx % cols;
+    auto row = tidx / cols;
+    if (row >= rows) {
+        return;
+    }
+    fn(row, col, args...);
```
yhmtsai, line 96: Maybe also pass tidx to kernel? tidx might be useful for (row, 0) array

yhmtsai (commented):
I would not say this reduce the porting effort, but it really reduce a lot code from porting. Thanks!
Originally, our porting script always generates 1 ~ 2 layers to keep the same interface and template usage as CUDA/HIP. Thus, it will gives many layer just for internal, which is somehow annoying.
I think we can leave these kernels launch args to compiler decide and it should not give any performance impact.
Do you have any idea such that we do not need to copy the same thing over different executors?
Instantiating in core directly seems to be impossible because we need different compiler to compile it.
Is putting it in generic folder like common a possible way? but it will increase the complexity to find the actual kernel code

dpcpp/base/kernel_launch.dp.hpp
upsj: This is mainly a precaution to prevent conflicting GKO_KERNEL definitions.

cuda/base/kernel_launch.hpp
upsj: Either that or add a `index_parameter` pseudo-parameter that transforms into the thread index, so we don't have too many parametery by default. I have an idea for a more general approach for specifying different types of kernels, but that will need some more work.

upsj:
@yhmtsai Yes, that was the idea, we could basically have a core/device/simple_headers.hpp file that includes the correct executor's headers etc. dependent on which executor is being compiled.

cuda/base/kernel_launch.hpp
```diff
@@ -0,0 +1,182 @@
...
+
+
+template <typename ValueType>
+matrix_accessor<cuda_type<ValueType>> map_to_device(
+    matrix::Dense<ValueType> *mtx)
+{
+    return {as_cuda_type(mtx->get_values()), mtx->get_stride()};
+}
+
+
```
yhmtsai, line 107:
```suggestion

```
the following are also overloading, so it only needs 1 linebreak

cuda/base/kernel_launch.hpp
```diff
@@ -0,0 +1,182 @@
...
+void CudaExecutor::run_kernel(KernelFunction fn, size_type size,
+                              KernelArgs &&... args) const
+{
+    constexpr auto block_size = kernels::cuda::default_block_size;
+    auto num_blocks = ceildiv(size, block_size);
+    kernels::cuda::generic_kernel_1d<<<num_blocks, block_size>>>(
+        size, fn, kernels::cuda::map_to_device(args)...);
+}
+
+
```
yhmtsai, line 168: also here

dpcpp/base/kernel_launch.dp.hpp
```diff
@@ -0,0 +1,176 @@
...
+#define GKO_KERNEL
+#endif
+
+
+namespace gko {
+namespace kernels {
+namespace dpcpp {
+
+
+constexpr auto default_block_size = 512;
```
yhmtsai, line 56:
```suggestion
```
the dpcpp will automatically decide it

yhmtsai (changes requested):
LGTM.
Could you also add the documentation for these function and Array needs the object not pointer?
Array can still use Array.get_data() to pass, right? it will be the normal value pointer.
Will you change other in different PR? like dense strided_fill, etc..
I am not sure whether `core` is good place to store the generic imp because we have many files with different purpose there. Does the file will be `.hpp` or `.hpp.inc`.
It somehow needs to be in namespace, right? use DefaultExecutor and namespace will give the proper Executor type.
Does run_kernel needs to be under Executor?
the executor type is determined in device function, so is `run_kernel(exec, ....)` already enough?

cuda/base/kernel_launch.hpp
upsj: Come to think of it again, wouldn't it be more appropriate to use a generic_kernel_1d for this?

yhmtsai: yes, if only (row, 0). but maybe others (row, col) will still need 2d.
Or, do you mean we leave the control to each accessor?
every item takes tidx and does its own condition.
the branches in theory are the same as origin one, but I am not sure whether several condition branch (the condition are identical) gives the same performance as one condition branch

upsj: Can you give an example? I don't see the use case yet.

upsj:
@yhmtsai Before adding documentation, I wanted to make sure the interface is fine. Yes, you can both pass Arrays or pointers to them directly, both results in the same pointer on the device side.

Using a free function run_kernel actually sounds like a great idea, that would also decouple things from the public interface. Still, how much of this do we want to provide to the users?

upsj:
@yhmtsai For your other questions: Since it is a complete source file, I put them into `.cpp` files for now, which also makes them easier to distinguish from other files in `common`, though it might not be ideal to put them both into the same folder.

One thing I should add is that pointers can currently be passed through only partially (right now limited to value and index types), since otherwise we would also pass through matrix::Dense pointers. A workaround for that would be using a struct instead of functions, but I don't want to complicate things too much.

cuda/base/kernel_launch.hpp
yhmtsai: I do not find the corresponding use case, either.
Another kernel may be `add_scaled` of dense.
Do you think generic_kernel need the ability of plain value into `map_to_device`?
I think this kernel should be the simple but needs the num_alpha_col.
plain variable must be limited to ValueType/IndexType/size_type
and if have the plain variable ability, some kernels with different size should also work on generic_kernel. 
initialize_1 of gmres needs krylov_dim and num_row for different matrix

upsj: Plain values are a good point. I think this is a good argument for making "pass-through" the default and only adding a special case for Array and Dense.

yhmtsai:
Users implement kernel via `run_kernel` always needs `cuda/base/launch_kernel.cuh`, so they are requires the same headers.
For the place, additional folder gives clear usage but needs the same folder structure like others.
using the existed folder does not create a full folder structure again but the codes for different purpose are in the same place.

upsj:
@yhmtsai that's correct. For pointers and values, I only need to add that they are passed through as_cuda_type/as_hip_type if necessary

tcojean (changes requested):
In general, that's an interesting approach at least for all the code reduction it brings. I am of course curious about the performance results in order to make sure that this does not come with a cost. I have some remarks:


+ We should separate common device kernels from common launch. Maybe in `common/kernels` and `common/impl` or something like this.
+ You claim that this would help write kernels from the user side (e.g., in `examples`), for now I fail to see how because the user would still need to compile a separate file/module specifically for `hip`, `dpcpp` or `cuda`, right? If you think it's possible without that complexity, can you add an example?
+ Can you add tests for all of this? E.g., that you can use `run_kernel` in a few settings, also that types are properly expanded.

hip/base/kernel_launch.hip.hpp
```diff
@@ -0,0 +1,261 @@
...
+    GKO_ASSERT(mtx->get_size()[0] == 1 ||
+               (mtx->get_size()[1] == 1 && mtx->get_stride() == 1));
+    return as_hip_type(mtx->get_const_values());
+}
+
+
+template <typename T>
+struct device_unpack_1d_impl {
+    using type = T;
+    static __device__ __forceinline__ type unpack(T param, size_type, size_type)
```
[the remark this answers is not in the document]
upsj: Can you elaborate? For the function attributes, the whole struct or the implementation? The former two should work, the latter would probably difficult, since this is used in a template context.

common/solver/bicg_kernels.cpp
```diff
@@ -33,18 +33,15 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "core/solver/bicg_kernels.hpp"
 
 
-#include <omp.h>
+#include <ginkgo/core/base/math.hpp>
 
 
-#include <ginkgo/core/base/array.hpp>
```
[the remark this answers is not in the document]
upsj: I was relying on that to be imported from bicg_kernels.hpp for the interface, but it probably still makes sense from the `iwyu` standpoint.

common/solver/cg_kernels.cpp
```diff
@@ -54,85 +51,82 @@ namespace cg {
...
+        exec,
+        [] GKO_KERNEL(auto row, auto col, auto b, auto r, auto z, auto p,
+                      auto q, auto prev_rho, auto rho, auto stop) {
+            if (row == 0) {
+                rho[col] = zero(rho[col]);
+                prev_rho[col] = one(prev_rho[col]);
+                stop[col].reset();
+            }
+            r(row, col) = b(row, col);
+            z(row, col) = p(row, col) = q(row, col) = zero(z(row, col));
```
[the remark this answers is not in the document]
upsj: Safe in what way? the accessors return references to actual objects, so since `=` is right-associative, we first assign to q, take the resulting zero and assign it to p, then to z.

common/solver/cgs_kernels.cpp
```diff
@@ -66,91 +63,84 @@ void initialize(std::shared_ptr<const OmpExecutor> exec,
...
+                rho[col] = zero(rho[col]);
+                prev_rho[col] = alpha[col] = beta[col] = gamma[col] =
+                    one(prev_rho[col]);
+                stop[col].reset();
+            }
+            r(row, col) = r_tld(row, col) = b(row, col);
+            u(row, col) = u_hat(row, col) = p(row, col) = q(row, col) =
+                v_hat(row, col) = t(row, col) = zero(u(row, col));
+        },
+        p->get_size(), b, compact(r), compact(r_tld), compact(p), compact(q),
```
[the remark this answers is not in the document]
upsj: No particular reason, just copy-pasted, since `b` or `x` are not available in all kernels.

common/solver/fcg_kernels.cpp
```diff
@@ -53,89 +51,83 @@ namespace fcg {
...
-                auto tmp = rho->at(j) / beta->at(j);
-                auto prev_r = r->at(i, j);
-                x->at(i, j) += tmp * p->at(i, j);
-                r->at(i, j) -= tmp * q->at(i, j);
-                t->at(i, j) = r->at(i, j) - prev_r;
+    run_kernel(
+        exec,
+        [] GKO_KERNEL(auto row, auto col, auto x, auto r, auto t, auto p,
+                      auto q, auto beta, auto rho, auto stop) {
+            if (!stop[col].has_stopped()) {
```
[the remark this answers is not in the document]
upsj: Yes, my thought here was that it does not make a difference from the memory access pattern standpoint, and we don't have any more predicated writes/reads, which might be easier on the memory controller.

cuda/base/kernel_launch.cuh
```diff
@@ -0,0 +1,257 @@
...
+template <typename T>
+typename device_map_impl<T>::type map_to_device(T &&param)
+{
+    return device_map_impl<T>::map_to_device(param);
+}
+
+
+template <typename ValueType>
+struct compact_dense_wrapper {
+    ValueType *data;
```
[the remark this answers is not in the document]
upsj: yes, most likely the best place to put it, though I am not sure we should even expect an impact of this, since we are mostly memory-bound and do one read and one write per variable at most.

upsj:
@tcojean For user-facing kernel launches, there are two ways we can approach this: 1. require the users to decide on a single executor at compile-time, and include `executor/base/kernel_launch.hpp` (which kind of mixed public and private headers) or have some machinery for compiling the same source file for multiple executors using `core/base/simple_kernels.hpp` and some `core`-like code in between. But I would personally like to avoid the latter for now.

omp/base/kernel_launch.hpp
```diff
@@ -0,0 +1,240 @@
...
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+#ifndef GKO_OMP_BASE_KERNEL_LAUNCH_HPP_
+#define GKO_OMP_BASE_KERNEL_LAUNCH_HPP_
+
+
```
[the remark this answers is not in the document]
upsj: This is not necessary right now, since we don't use any OpenMP functions, but it could be helpful if we decide to add more complex kernels, like for reductions etc.

common/solver/bicg_kernels.cpp
```diff
@@ -63,57 +60,60 @@ void initialize(std::shared_ptr<const OmpExecutor> exec,
...
+            r(row, col) = b(row, col);
+            r2(row, col) = b(row, col);
+            z(row, col) = zero(z(row, col));
+            p(row, col) = zero(p(row, col));
+            q(row, col) = zero(q(row, col));
+            z2(row, col) = zero(z2(row, col));
+            p2(row, col) = zero(p2(row, col));
+            q2(row, col) = zero(q2(row, col));
+        },
+        p->get_size(), b, compact(r), compact(z), compact(p), compact(q),
```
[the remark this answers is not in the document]
yhmtsai: I think we would like to have new create_with_same_size_of(b), so the others except for b have the same stride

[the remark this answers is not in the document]
upsj: Yes, the implementation currently does not respect non-default strides in `b` yet. I would personally prefer changing the behavior of `create_with_same_config` to not copy the stride, since this would prevent any issues related to users overriding Dense::create_with_same_config (considering MFEM especially)

hip/CMakeLists.txt
```diff
@@ -179,19 +178,22 @@ set(GINKGO_HIP_SOURCES
...
     solver/gmres_kernels.hip.cpp
     solver/cb_gmres_kernels.hip.cpp
     solver/idr_kernels.hip.cpp
-    solver/ir_kernels.hip.cpp
     solver/lower_trs_kernels.hip.cpp
     solver/upper_trs_kernels.hip.cpp
     stop/criterion_kernels.hip.cpp
-    stop/residual_norm_kernels.hip.cpp)
+    stop/residual_norm_kernels.hip.cpp
+    ../common/components/precision_conversion.cpp
```
yhmtsai, line 188: Will hip take all *.cpp file not only *.hip.cpp?

common/solver/cgs_kernels.cpp
```diff
@@ -66,91 +63,84 @@ void initialize(std::shared_ptr<const OmpExecutor> exec,
...
-            if (stop_status->get_const_data()[j].has_stopped()) {
-                continue;
+    run_kernel(
+        exec,
+        [] GKO_KERNEL(auto row, auto col, auto r, auto u, auto p, auto q,
+                      auto beta, auto rho, auto prev_rho, auto stop) {
+            if (!stop[col].has_stopped()) {
+                auto tmp = safe_divide(rho[col], prev_rho[col]);
+                if (row == 0) {
+                    beta[col] = tmp;
```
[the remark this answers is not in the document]
upsj: beta is not used outside of this kernel, so this doesn't make any difference (we wouldn't even need to store it, except when we want to inspect the internal state)

common/solver/cgs_kernels.cpp
```diff
@@ -66,91 +63,84 @@ void initialize(std::shared_ptr<const OmpExecutor> exec,
...
-            if (stop_status->get_const_data()[j].has_stopped()) {
-                continue;
+    run_kernel(
+        exec,
+        [] GKO_KERNEL(auto row, auto col, auto u, auto v_hat, auto q, auto t,
+                      auto alpha, auto rho, auto gamma, auto stop) {
+            if (!stop[col].has_stopped()) {
+                auto tmp = safe_divide(rho[col], gamma[col]);
+                if (row == 0) {
+                    alpha[col] = tmp;
```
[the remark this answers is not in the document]
upsj: I am not entirely sure it is a good idea to leave alpha unmodified in case gamma is zero.

hip/CMakeLists.txt
upsj: Yes, `hipcc` doesn't care about the file extension

common/solver/cgs_kernels.cpp
```diff
@@ -66,91 +63,84 @@ void initialize(std::shared_ptr<const OmpExecutor> exec,
...
+        [] GKO_KERNEL(auto row, auto col, auto r, auto u, auto p, auto q,
+                      auto beta, auto rho, auto prev_rho, auto stop) {
+            if (!stop[col].has_stopped()) {
+                auto tmp = safe_divide(rho[col], prev_rho[col]);
+                if (row == 0) {
+                    beta[col] = tmp;
+                }
+                u(row, col) = r(row, col) + tmp * q(row, col);
+                p(row, col) =
+                    u(row, col) + tmp * (q(row, col) + tmp * p(row, col));
```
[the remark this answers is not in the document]
upsj: I guess you mean beta?

common/solver/cg_kernels.cpp
```diff
@@ -54,85 +51,82 @@ namespace cg {
...
-            }
-            if (beta->at(j) != zero<ValueType>()) {
-                auto tmp = rho->at(j) / beta->at(j);
-                x->at(i, j) += tmp * p->at(i, j);
-                r->at(i, j) -= tmp * q->at(i, j);
+    run_kernel(
+        exec,
+        [] GKO_KERNEL(auto row, auto col, auto x, auto r, auto p, auto q,
+                      auto beta, auto rho, auto stop) {
+            if (!stop[col].has_stopped()) {
```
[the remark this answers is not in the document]
upsj: The behavior is equivalent unless p or q contain non-finite numbers

common/solver/cgs_kernels.cpp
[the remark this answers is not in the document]
yhmtsai: sorry for the late reply. (I should figure how to filter a tons of github mail)
Yes, it is beta.

upsj:
I ran a few benchmarks for solvers and Dense kernels, and got the following speedups:

BLAS operations

cuda-blas
omp-blas

Solver kernels with different #rhs:

cuda-solver
omp-solver

So there is definitely something going wrong with OpenMP BLAS operations, I need to check these in more detail. These differences might also be due to weird NUMA effects, since I didn't pin my threads and ran on a two-socket system. I will try rerunning everything with more controlled settings tomorrow.

upsj:
After some deeper investigation, it looks like the following change brings us back to original performance, tested with `OMP_NUM_THREADS=8 OMP_PLACES=cores OMP_PROC_BIND=close`
Probably the index computations using modulo prohibit some optimizations.
```diff
diff --git a/omp/base/kernel_launch.hpp b/omp/base/kernel_launch.hpp
index 546d53adfd..b307576b46 100644
--- a/omp/base/kernel_launch.hpp
+++ b/omp/base/kernel_launch.hpp
@@ -222,15 +222,15 @@ template <typename KernelFunction, typename... KernelArgs>
 void run_kernel(std::shared_ptr<const OmpExecutor> exec, KernelFunction fn,
                 dim<2> size, KernelArgs &&... args)
 {
-#pragma omp parallel for
-    for (size_type i = 0; i < size[0] * size[1]; i++) {
-        auto row = i / size[1];
-        auto col = i % size[1];
-        [&]() {
-            fn(row, col,
-               kernels::omp::map_unpack_to_device(args, row, col, size[0],
-                                                  size[1])...);
-        }();
+#pragma omp parallel for collapse(2)
+    for (size_type row = 0; row < size[0]; row++) {
+       for (size_type col = 0; col < size[1]; col++) {
+            [&]() {
+                fn(row, col,
+                   kernels::omp::map_unpack_to_device(args, row, col,
+                                                     size[0], size[1])...);
+            }();
+        }
     }
 }
```

yhmtsai:
or maybe increase the chunk_size? collapse also looks good to me.

dpcpp/base/kernel_launch.dp.hpp
```diff
@@ -0,0 +1,242 @@
...
+    GKO_ASSERT(mtx->get_size()[0] == 1 ||
+               (mtx->get_size()[1] == 1 && mtx->get_stride() == 1));
+    return mtx->get_const_values();
+}
+
+
+template <typename T>
+struct device_unpack_1d_impl {
+    using type = T;
+    static type unpack(T param, size_type, size_type) { return param; }
```
Slaedr, line 164: Again, I don't get the reason for all these unused parameters in this function and the next.

hip/base/kernel_launch.hip.hpp
```diff
@@ -0,0 +1,260 @@
...
+namespace gko {
+namespace kernels {
+namespace hip {
+
+
+constexpr auto default_block_size = 512;
+
+
+template <typename ValueType>
+struct matrix_accessor {
```
Slaedr, line 65: Except for the actual kernel launch with `hipKernelLaunchGGL`, is it not possible to somehow reuse the cuda code for all this, and move them to common? That's quite a bit of nearly-duplicated code.

hip/base/kernel_launch.hip.hpp
```diff
@@ -0,0 +1,260 @@
...
+
+
+template <typename KernelFunction, typename... KernelArgs>
+void run_kernel(std::shared_ptr<const HipExecutor> exec, KernelFunction fn,
+                size_type size, KernelArgs &&... args)
+{
+    hip::device_guard guard{exec->get_device_id()};
+    constexpr auto block_size = kernels::hip::default_block_size;
+    auto num_blocks = ceildiv(size, block_size);
+    hipLaunchKernelGGL(kernels::hip::generic_kernel_1d, num_blocks, block_size,
```
Slaedr, line 241: It's funny that `hipLaunchKernelGGL` itself is a higher-order function to launch other functions, but in your case it launches a function that "launches" yet another function.  😆

include/ginkgo/core/base/math.hpp
```diff
@@ -1008,6 +1008,24 @@ is_finite(const T &value)
...
+ *
+ * @param a  the dividend
+ * @param b  the divisor
+ *
+ * @return the value of `a / b` if b is non-zero, zero otherwise.
+ */
+template <typename T>
+GKO_INLINE GKO_ATTRIBUTES T safe_divide(T a, T b)
+{
+    return b == zero<T>() ? zero<T>() : a / b;
```
Slaedr, line 1025: Any votes for leaving `inf` as it is, but doing this only for `0/0`? Maybe that's a separate PR, though.

omp/base/kernel_launch.hpp
```diff
@@ -0,0 +1,240 @@
...
+        }();
+    }
+}
+
+template <typename KernelFunction, typename... KernelArgs>
+void run_kernel(std::shared_ptr<const OmpExecutor> exec, KernelFunction fn,
+                dim<2> size, KernelArgs &&... args)
+{
+#pragma omp parallel for
+    for (size_type i = 0; i < size[0] * size[1]; i++) {
```
Slaedr, line 226: You're going to replace this with a 2D loop and `collapse`, right?

cuda/base/kernel_launch.cuh
```diff
@@ -0,0 +1,256 @@
...
+    return to_device_type_impl<T>::map_to_device(param);
+}
+
+
+template <typename ValueType>
+struct compact_dense_wrapper {
+    ValueType *data;
+};
+
+
```
Slaedr, line 130: Can we document somewhere what exactly `compact` is, especially to differentiate it from just passing a `Dense`?
```suggestion
/**
 * @fn compact
 * Creates and returns a simple device-friendly representation of a dense matrix
 * whose stride exactly equals its number of columns.
 */
```
Though I'm not sure about the best place to put it. I'm thinking of how to inform developers on how to use `compact()` versus just passing plain `Dense` while supplying their lamda parameters, essentially.

Slaedr (commented):
This is really good stuff. Great work! Nice job removing so many files! I think it mostly looks good to go, especially since you already did performance comparisons.

However, I do have some concerns below. It would also be nice if you documented the Hip performance difference and added the final Omp comparison after changing the kernel launch.

Finally, maybe as an extension in a separate PR, you could look at a general kernel launch for simple reductions. That *might* be enough, for example, to implement custom norms in applications.

common/matrix/dense_kernels.cpp
```diff
@@ -0,0 +1,340 @@
...
+
+
+template <typename ValueType>
+void fill(std::shared_ptr<const DefaultExecutor> exec,
+          matrix::Dense<ValueType> *mat, ValueType value)
+{
+    run_kernel(
+        exec,
+        [] GKO_KERNEL(auto row, auto col, auto mat, auto value) {
+            mat(row, col) = value;
```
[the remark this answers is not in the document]
upsj: Sure, that's both possible. Since we only use this kernel in something matrix-like, I went for row and column, since it looks slightly more descriptive to me.

common/solver/cg_kernels.cpp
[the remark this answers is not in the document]
upsj: Unfortunately no, since on CUDA and HIP, we need `as_cuda/hip_type<ValueType>` instead

cuda/base/kernel_launch.cuh
```diff
@@ -0,0 +1,256 @@
...
+template <typename T>
+struct to_device_type_impl {
+    using type = std::decay_t<cuda_type<T>>;
+    static type map_to_device(T in) { return as_cuda_type(in); }
+};
+
+template <typename ValueType>
+struct to_device_type_impl<matrix::Dense<ValueType> *&> {
```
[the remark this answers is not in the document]
upsj: That's kind of necessary due to the fact that we want to be able to pass both references and pointers at the same time. To make sure that values are actually passed by reference and not by value, we need to use forwarding references `T&&...` everywhere, which means that T will always be a reference type. It might be possible to work around this, but I don't want to complicate the code too much.

cuda/base/kernel_launch.cuh
```diff
@@ -0,0 +1,256 @@
...
+    GKO_ASSERT(mtx->get_size()[0] == 1 ||
+               (mtx->get_size()[1] == 1 && mtx->get_stride() == 1));
+    return as_cuda_type(mtx->get_const_values());
+}
+
+
+template <typename T>
+struct device_unpack_1d_impl {
+    using type = T;
+    static __device__ __forceinline__ type unpack(T param, size_type, size_type)
```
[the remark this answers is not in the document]
upsj: Ah right, in 1D we don't necessarily need the unpacking step. I only added it to make things like `run_kernel(size, [](auto& x) { x = 1; }, pointwise(x));` possible, but they are unnecessarily fragile. I guess I'll remove it.

hip/base/kernel_launch.hip.hpp
upsj: Yes, that's definitely worth a consideration. I wanted to keep the files as self-contained as possible for now, but we can definitely improve upon the duplication.

include/ginkgo/core/base/math.hpp
upsj: This would allow solvers to introduce `NaN` into vectors, which might not be the best choice? Right now, that is only possible in very limited circumstances (overflows in vector entries or b != 0 but a/b overflows)

omp/base/kernel_launch.hpp
upsj: yes

cuda/base/kernel_launch.cuh
upsj: Yes, I will move the kernel interface description to a common header for documentation. Also in general, I only left most of the code undocumented because I wasn't certain it would stay this way. This will change before the merge.

common/solver/cg_kernels.cpp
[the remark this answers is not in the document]
Slaedr: Can we define a struct called `as_device_type` that does the correct thing based on which backend is being compiled? That could be used here, and it could also help unify your Cuda and hip simple kernel infrastructure.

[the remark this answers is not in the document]
upsj: Yes, that makes sense. But I would still prefer `zero(z(row, col))` over `zero<as_device_type<ValueType>>()`

include/ginkgo/core/base/math.hpp
yhmtsai: I think it is to fit the original behavior mostly

yhmtsai (approved):
LGTM in general.

core/test/matrix/dense.cpp
```diff
@@ -138,6 +138,18 @@ TYPED_TEST(Dense, CanBeConstructedFromExistingData)
 }
 
 
+TYPED_TEST(Dense, CreateWithSameConfigDoesntKeepStride)
+{
+    auto m =
+        gko::matrix::Dense<TypeParam>::create(this->exec, gko::dim<2>{2, 3}, 4);
+    auto m2 = gko::matrix::Dense<TypeParam>::create_with_config_of(m.get());
```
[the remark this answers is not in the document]
upsj: Unfortunately, that is not possible due to our MFEM wrappers only implementing this function.

upsj:
I finished tests for the solvers and simple kernel launch. The solver tests are just merged versions of the individual executor kernels plus some special cases for division by zero and stopped columns added. The kernel launch tests test both static and dynamic behaviors, but the code itself is the same across all of them except for a workaround that was necessary due to some nvcc device-lambda-in-classes-with-protected-members issues

common/solver/cg_kernels.cpp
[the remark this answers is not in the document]
Slaedr: I just thought it's un-intuitive to have a "zero of some vector", while the lambda is written on a scalar level. Otherwise it's fine.

cuda/base/kernel_launch.cuh
```diff
@@ -0,0 +1,247 @@
...
+        return param;
+    }
+};
+
+template <typename ValueType>
+struct device_unpack_2d_impl<compact_dense_wrapper<ValueType>> {
+    using type = matrix_accessor<ValueType>;
+    static __device__ __forceinline__ type
+    unpack(compact_dense_wrapper<ValueType> param, size_type, size_type,
+           size_type, size_type num_cols)
```
Slaedr, line 146: Here, too, are these extra parameters necessary?
```suggestion
    unpack(compact_dense_wrapper<ValueType> param, size_type num_cols)
```

cuda/base/kernel_launch.cuh
```diff
@@ -0,0 +1,247 @@
...
+{
+    auto tidx = thread::get_thread_id_flat();
+    auto col = tidx % cols;
+    auto row = tidx / cols;
+    if (row >= rows) {
+        return;
+    }
+    fn(row, col,
+       device_unpack_2d_impl<KernelArgs>::unpack(args, row, col, rows,
+                                                 cols)...);
```
Slaedr, line 177: And here you could simply have
```suggestion
       device_unpack_2d_impl<KernelArgs>::unpack(args, cols)...);
```

cuda/base/kernel_launch.cuh
upsj: At the moment no, they could only be useful for row-major storage at some point.

Slaedr: Good point. But then you need just one more parameter for `num_rows`, because the lambda would be the current row and column indices anyway.
```suggestion
    unpack(compact_dense_wrapper<ValueType> param,
           size_type, size_type num_cols)
```

upsj:
So CUDA looks good, I haven't tested HIP yet but that will follow. I did some more extensive OpenMP experiments on an exclusive `single` node on bwUniCluster with GCC 10.2:
* x axis: number of rows
* y axis: bandwidth (B/s)
* label: number of columns
* omp-4/40: Number of OpenMP threads
* nocollapse: remove `collapse(2)`
* nolambda: remove immediately invoked lambda expression in the loop body that needed to be added because of a bug in ICC
* 1d: replace 2 nested loops by a single loop of size rows * cols, compute row and column index with division.

output

Take-aways: nocollapse-nolambda looks equivalent to the old implementation and generally provides the best performance
1d seems to make it really hard on the optimizer, nolambda seems to have negligible effect, but collapse(2) really hampers performance. I will dig a bit deeper into the small dips we can see at some places, but in general this looks okay to me. I will run the same thing through a few different compilers and check how far we can get with this.

upsj:
Of course missing from the comparison: with lambda, but without collapse
4 threads
omp-4-nocollapse-axpy
40 threads
omp-40-nocollapse-axpy

looks slightly worse than without lambda

omp/base/kernel_launch.hpp
```diff
@@ -0,0 +1,304 @@
...
+{
+    const auto rows = size[0];
+    const auto cols = size[1];
+    const auto rounded_cols = cols / 4 * 4;
+    GKO_ASSERT(rounded_cols + remainder_cols == cols);
+#pragma omp parallel for
+    for (size_type row = 0; row < rows; row++) {
+        for (size_type base_col = 0; base_col < rounded_cols; base_col += 4) {
+#pragma unroll
+            for (size_type i = 0; i < 4; i++) {
```
Slaedr, line 230: Nice! I guess a good aspect of common kernels is that you can now afford to spend time putting all these optimizations in one place and they pay off everywhere they're used. I guess this is to aid interleaving, but does this still vectorize?

cuda/base/kernel_launch.cuh
```diff
@@ -0,0 +1,243 @@
...
+}  // namespace kernels
+
+
+template <typename ValueType>
+kernels::cuda::compact_dense_wrapper<kernels::cuda::cuda_type<ValueType>>
+compact(matrix::Dense<ValueType> *mtx)
+{
+    GKO_ASSERT(mtx->get_stride() == mtx->get_size()[1]);
+    return {kernels::cuda::as_cuda_type(mtx->get_values())};
+}
```
thoasm, line 187: This function needs a more descriptive name, or a separate namespace (maybe even both since this is in the top-level namespace). I still don't quite understand why it is necessary to have such a function. Is the intention to avoid an additional parameter for a stride that is known to be `size[1]`?

cuda/base/kernel_launch.cuh
```diff
@@ -0,0 +1,243 @@
...
+
+template <typename KernelFunction, typename... KernelArgs>
+void run_kernel(std::shared_ptr<const CudaExecutor> exec, KernelFunction fn,
+                size_type size, KernelArgs &&... args)
+{
+    cuda::device_guard guard{exec->get_device_id()};
+    constexpr auto block_size = kernels::cuda::default_block_size;
+    auto num_blocks = ceildiv(size, block_size);
+    kernels::cuda::generic_kernel_1d<<<num_blocks, block_size>>>(
+        size, fn, kernels::cuda::map_to_device(args)...);
```
thoasm, line 226: Not having access to the size of the array prevents certain implementations. For example, if I want to initialize a reverse-counter (counting backwards from N) or want to do a bound-check myself, I would have to forward the size twice (once for the `generic_kernel`, the other time for my own lambda `fn`). If I had access to the size all the time, there would also be no need for the functions `compact` and `vector` as the author of the kernel could implement the accesses him-/herself (this would also make it obvious that it is an optimization, which `compact` or `vector` does not communicate).
Therefore, I would propose that all `KernelFunction`s get the size as the first argument, followed by the current index.
(same applies for 2D case.)

cuda/base/kernel_launch.cuh
```diff
@@ -0,0 +1,243 @@
...
+
+template <typename KernelFunction, typename... KernelArgs>
+void run_kernel(std::shared_ptr<const CudaExecutor> exec, KernelFunction fn,
+                dim<2> size, KernelArgs &&... args)
+{
+    cuda::device_guard guard{exec->get_device_id()};
+    constexpr auto block_size = kernels::cuda::default_block_size;
+    auto num_blocks = ceildiv(size[0] * size[1], block_size);
+    kernels::cuda::generic_kernel_2d<<<num_blocks, block_size>>>(
+        size[0], size[1], fn, kernels::cuda::map_to_device(args)...);
```
thoasm, line 237: Why are breaking `dim<2>` open instead of giving the kernel a single `dim<2>` argument?
If you would give the auther the size into the `KernelFunction`, this would also only occupy a single argument instead of two.

thoasm (changes requested):
My thoughts so far on the `kernel_launch` portion (commented only on the `cuda` version).
I do like the idea and fully support this PR, but I would like to discuss a few points before green-lighting this PR.

CMakeLists.txt
```diff
@@ -330,56 +333,69 @@ ginkgo_modify_flags(CMAKE_CUDA_FLAGS_DEBUG)
...
+set(GINKGO_TEST_INSTALL_SRC_DIR "${Ginkgo_SOURCE_DIR}/test/build_install/")
+set(GINKGO_TEST_INSTALL_BIN_DIR "${Ginkgo_BINARY_DIR}/test/build_install/")
+set(GINKGO_TEST_EXPORTBUILD_SRC_DIR "${Ginkgo_SOURCE_DIR}/test/build_export/")
+set(GINKGO_TEST_EXPORTBUILD_BIN_DIR "${Ginkgo_BINARY_DIR}/test/build_export/")
 if(MSVC)
-    # Set path/command with $<CONFIG>
-    set(GINKGO_TEST_INSTALL_COMMAND "${Ginkgo_BINARY_DIR}/test_install/$<CONFIG>/test_install")
-    set(GINKGO_TEST_EXPORTBUILD_COMMAND "${Ginkgo_BINARY_DIR}/test_exportbuild/$<CONFIG>/test_exportbuild")
+    set(GINKGO_TEST_INSTALL_CMD ${GINKGO_TEST_INSTALL_BIN_DIR}/$<CONFIG>/test_install)
+    set(GINKGO_TEST_EXPORTBUILD_CMD ${GINKGO_TEST_EXPORTBUILD_BIN_DIR}/$<CONFIG>/test_exportbuild)
```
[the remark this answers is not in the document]
upsj: I had a bit of a misunderstanding regarding spaces in CMake in the past, I assumed everything needs to be quoted to avoid issues with spaces, but spaces in expanded variables `${var_name}` already get quoted correctly. So we do not need them here.

cuda/base/kernel_launch.cuh
```diff
@@ -0,0 +1,243 @@
...
+};
+
+template <typename ValueType>
+struct to_device_type_impl<Array<ValueType> &> {
+    using type = cuda_type<ValueType> *;
+    static type map_to_device(Array<ValueType> &array)
+    {
+        return as_cuda_type(array.get_data());
+    }
+};
```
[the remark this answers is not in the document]
upsj: In Ginkgo, we always pass polymorphic objects as pointers (almost always) arrays by reference. So I wanted to use the same convention.

cuda/base/kernel_launch.cuh
```diff
@@ -0,0 +1,243 @@
...
+        return as_cuda_type(array.get_const_data());
+    }
+};
+
+
+template <typename T>
+typename to_device_type_impl<T>::type map_to_device(T &&param)
+{
+    return to_device_type_impl<T>::map_to_device(param);
+}
```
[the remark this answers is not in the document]
upsj: I actually do something forward-like, though it's well-hidden: I have a T&& parameter, but take T as a template parameter. If I didn't have that parameter, passing in a Array would create a copy and thus prevent kernels from modifying arrays. I only use forwarding references where they are necessary.

cuda/base/kernel_launch.cuh
upsj: Yes, it avoids having a separate stride parameter for each parameter, which might become significant for solver kernels with many parameters.

cuda/base/kernel_launch.cuh
upsj: I think what you are suggesting here would exceed the intended purpose of the framework, as simple kernels like these should not need to do any bounds checks, and would additionally add 2 unused parameters to all kernels. If the implementations do need access to the size, we can pass it explicitly as a parameter.

cuda/base/kernel_launch.cuh
upsj: I tried to avoid using more complex types on the device. Since this is the backend, where do you see the difference in using a single argument instead of two?

omp/base/kernel_launch.hpp
upsj: Exactly, this was one of the ideas behind this approach :) I did not check the generated code in detail, it should definitely be possible to auto-vectorize, but since the kernels are memory-bound, I am not sure how much of a difference vectorization would make here.

cuda/base/kernel_launch.cuh
thoasm: In that case, I would remove these functions or, at the very least, move them into the `dense` namespace (we should not pollute the `gko` namespace with something only used for `Dense` matrices).
Personally, I would remove them since they are purely for niche optimization purposes and they make the code a lot harder to read.

cuda/base/kernel_launch.cuh
thoasm: Better self-documentation and extensibility (for 3D, 4D, ...).
Also, with a `dim`, it is impossible to accidentally swap rows and cols (otherwise, you could swap them in the kernel call or in the variable naming of the kernel).
But this is obviously a very minor thing and I would be fine to keep it like it is now, but I would prefer `dim<2>`.

pratikvn (commented):
This looks really good! It will simplify a lot of things.

Only minor things and one major aspect. As this is a major change in Ginkgo, it might make sense to first run the whole Suitesparse benchmarks for the affected kernels (solvers and dense kernels) in the current develop and store them somewhere (ginkgo-data repo), so that we can come back to it if there are any issues at a later point.

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -827,14 +827,14 @@ class Dense
     }
 
     /**
-     * Creates a Dense matrix with the same configuration as the callers matrix.
+     * Creates a Dense matrix with the same size as the callers matrix.
+     * Note that the stride is not preserved!
```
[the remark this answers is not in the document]
upsj: We basically have a choice here between
1. adding a new function create_with_same_size, keeping the original behavior in create_with_same_config
2. modifying create_with_same_config to discard original stride
3. keep behavior as-is

1. breaks MFEM support, 2. changes behavior slightly, 3. can give bad memory usage and performance when applying solvers on a system with huge padding, and makes the `compact` optimization pretty much impossible, so I'd probably need to add something specific like `run_solver_kernel` expicitly that provides the correct stride in the device_unpack step.

cuda/base/kernel_launch.cuh
thoasm: It would be only a single (mostly useless) parameter (in case of `dim<N>` usage), but I understand your point.

include/ginkgo/core/matrix/dense.hpp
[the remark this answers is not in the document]
pratikvn: I think we should still go for 1. We have not yet released Ginkgo 1.4.0. So we should fix this now so that we dont have something whose name says that we preserve the stride. We can create a small PR on MFEM's side as well.

cuda/base/kernel_launch.cuh
upsj: What do you think about the new `solver::rowvector` and `solver::default_stride` wrappers instead?

cuda/base/kernel_launch.cuh
```diff
@@ -0,0 +1,243 @@
...
+{
+    GKO_ASSERT(mtx->get_stride() == mtx->get_size()[1]);
+    return {kernels::cuda::as_cuda_type(mtx->get_const_values())};
+}
+
+
+template <typename ValueType>
+kernels::cuda::cuda_type<ValueType> *vector(matrix::Dense<ValueType> *mtx)
+{
+    GKO_ASSERT(mtx->get_size()[0] == 1 ||
```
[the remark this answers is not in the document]
upsj: I changed this to be a `rowvector`, so I think this should no longer apply?

cuda/base/kernel_launch.cuh
upsj: Since we have individual functions for all of these types of kernels, I would prefer leaving it as it is.

cuda/base/kernel_launch.cuh
thoasm: I like them, and since they are in a different namespace, I am fine with them.
One nit: Why `rowvector` and not `row_vector`? I feel like it is easier to read.

upsj: Yes, that's probably better

cuda/base/kernel_launch.cuh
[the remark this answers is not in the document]
thoasm: I did not think about `Array`s being given directly instead of as pointers. In that case, I agree, there is no way around using `T &&args` if you want to avoid code duplication.

upsj:
I restructured the code a bit, which hopefully reduces the amount of code duplication. Still to be done is the following:
- [ ] Document how to write kernels using this setup
- [ ] Document internals a bit more
- [ ] Add tests for simple Dense kernels

CMakeLists.txt
```diff
@@ -330,56 +333,69 @@ ginkgo_modify_flags(CMAKE_CUDA_FLAGS_DEBUG)
 ginkgo_modify_flags(CMAKE_CUDA_FLAGS_RELEASE)
 ginkgo_install()
 
+set(GINKGO_TEST_INSTALL_SRC_DIR "${Ginkgo_SOURCE_DIR}/test/build_install/")
+set(GINKGO_TEST_INSTALL_BIN_DIR "${Ginkgo_BINARY_DIR}/test/build_install/")
```
tcojean, line 337: The naming here is a bit weird? `build_install` doesn't sound like what this is doing: testing the installed Ginkgo. Maybe just `install` or repeating the test with `test_install`?

tcojean (approved):
LGTM!

common/solver/cg_kernels.cpp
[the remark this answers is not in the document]
upsj: We now have `as_device_type` and `device_type` type aliases, but I guess it's still shorter to use an existing variable.

CMakeLists.txt
upsj: Yes, I'm also not happy with the name. `test/test_install` just felt a bit redundant. I wanted to emphasize that we test whether the build with installed Ginkgo works, because `test_subdir` doesn't clearly state that it tests the build process.

omp/test/base/kernel_launch.cpp
```diff
@@ -0,0 +1,236 @@
...
+#include "core/test/utils.hpp"
+#include "omp/base/kernel_launch_solver.hpp"
+
+
+namespace {
+
+
+using gko::dim;
+using gko::size_type;
+using std::is_same;
```
thoasm, line 52: See the other include issues.

test/utils/executor.hpp
```diff
@@ -30,49 +30,49 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
...
-namespace gko {
-namespace kernels {
-namespace hip {
-/**
- * @brief The IR solver namespace.
- *
- * @ingroup ir
- */
-namespace ir {
+void init_executor(std::shared_ptr<gko::ReferenceExecutor> ref,
```
thoasm, line 42: nit: missing include for `<memory>`

thoasm (approved):
LGTM, really nice work on unifying all tests as well.
I mainly have one question that is not in the comments:
What about the script `dev_tools/scripts/create_new_algorithm.sh`? Should it not include a check if a common kernel or test file exists and if so, copy that as well?

cuda/base/kernel_launch.cuh
```diff
@@ -0,0 +1,256 @@
...
+    using type = matrix_accessor<cuda_type<ValueType>>;
+    static type map_to_device(matrix::Dense<ValueType> *mtx)
+    {
+        return {as_cuda_type(mtx->get_values()), mtx->get_stride()};
+    }
+};
+
+template <typename ValueType>
+struct to_device_type_impl<const matrix::Dense<ValueType> *&> {
+    using type = matrix_accessor<const cuda_type<ValueType>>;
```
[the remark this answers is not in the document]
upsj: Yes, definitely. The main reason I had to build a separate accessor is that the existing Ginkgo ranges have value semantics, i.e. they would copy the data by default, instead of just copying pointers. That would make writing kernels much more fragile.

cuda/base/kernel_launch_solver.cuh
```diff
@@ -30,45 +30,49 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
...
+                       KernelFunction fn, dim<2> size, size_type default_stride,
+                       KernelArgs &&... args)
 {
-    const dim3 block_size(default_block_size, 1, 1);
-    const dim3 grid_size(ceildiv(stop_status->get_num_elems(), block_size.x), 1,
-                         1);
-
-    initialize_kernel<<<grid_size, block_size, 0, 0>>>(
-        stop_status->get_num_elems(), stop_status->get_data());
+    gko::cuda::device_guard guard{exec->get_device_id()};
```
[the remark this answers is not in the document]
upsj: This is mainly relevant if we later decide to add this capability also for users, otherwise they might execute on the wrong GPU if this is not executed in the context of an `Operation`.

common/components/precision_conversion.cpp
```diff
@@ -33,26 +33,29 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+
 namespace gko {
 namespace kernels {
-namespace omp {
+namespace GKO_DEVICE_NAMESPACE {
 namespace components {
 
 
 template <typename SourceType, typename TargetType>
 void convert_precision(std::shared_ptr<const DefaultExecutor> exec,
```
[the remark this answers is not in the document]
upsj: It is present in precision_conversion.hpp, so I would prefer relying on it in there.

PR #731, opened 2021-04-01
fix memory leak and use ref not omp in device test

yhmtsai:
This PR fixes memory leak in destructor of executor and `get_gpu_property`
and uses the reference executor not omp executor in the device test.
(mainly for rcm, which only the one use the function other than copy/alloc from host)
This should also solve the windows running issue with cuda.

Summary:
- fix memory leak in destructor of executor and get_gpu_property
- use reference executor in device test
- disable the device executor in global scope with MSVC
MSVC describe the constructor of mutex are constexpr, but they comment constexpr in mutex header in particular. 
- move third-party build after modifying C/CXX flags such that third party use updated flags
- add description of possible issue when switching static/shared with MSVC. It should not occur again after the previous changes
- always turn off GINKGO_BUILD_HWLOC and set off as default on Windows/MacOS
- recover msvc/reference/release/static test

PR #734, opened 2021-04-06
Bump the current version to 1.4.0

tcojean:
This will be the next master release.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,7 @@ if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.18)
     cmake_policy(SET CMP0104 OLD)
 endif()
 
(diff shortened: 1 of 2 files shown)


upsj (approved):

pratikvn (approved):

nbeams (approved):
Thanks!

yhmtsai (approved):
LGTM. Do you think we should also include the #731 for memory leak?

tcojean:
@yhmtsai This is not a proper release per se, this is only for convenience. It's fine that #731 or any other PR is merged afterwards. This is done so that we can identify the already done changes compared to 1.3.0 as part of the future 1.4.0, as well as the coming changes. I will consider actually releasing 1.4.0 when we have more DPC++ features available in `develop`.

thoasm (approved):
LGTM!

yhmtsai:
@tcojean I see. Thanks!

PR #735, opened 2021-04-06
Log first residual norm in CGS

upsj:
We are a bit inconsistent in our solvers: All of them log the value and check the stopping criterion before updating `x` and `r` for the first time except for CGS, where the check and log only happens after the first iteration.


yhmtsai (approved):
LGTM. I think line 152 should be `int iter = -1;`

upsj:
@yhmtsai great catch!

pratikvn (approved):
LGTM!

tcojean (approved):

PR #736, opened 2021-04-07
Add new Intel CI support

tcojean:
Make DPC++ tests work with the new system.

The new Intel CI system adds two new checks: the integrated and discrete GPU.
Some changes to the DPC++ tests are required:
+ Make the DPCPP executor tests work with GPUs.
+ Allow compiling and running the DPC++ tests for float only (double is not supported in either GPU).


yhmtsai:
short comment:
when enabling DPCPP_SINGLE_MODE, it will also remove the double related kernels of omp/reference.
Maybe dpcpp always uses DPCPP_related_arg? we can still have double support on the other module

tcojean:
No it does not, double is properly ran for omp/reference. Note that this works when passing `target_compile_definitions(target GINKGO_DPCPP_SINGLE_MODE=1)`. This is done only:
+ For the Dpcpp module (target `ginkgo_dpcpp`)
+ `ginkgo_create_test` which are part of the `dpcpp` folder or tests made using `ginkgo_create_dpcpp_test`.

yhmtsai:
@tcojean thanks for explanation. You are right, it only affect the DPCPP part.
I do not need the part for changing the macro of script now.

.gitlab-ci.yml
```diff
@@ -678,7 +679,7 @@ build/nocuda/intel/omp/release/static:
     - cuda
     - cpu
 
-build/dpcpp/cpu/release/static:
+build/dpcpp/igpu/release/static:
```
yhmtsai, line 682: also leave one dpcpp test on cpu?

dpcpp/CMakeLists.txt
```diff
@@ -51,11 +51,15 @@ target_sources(ginkgo_dpcpp
 
 ginkgo_compile_features(ginkgo_dpcpp)
 
-set(GINKGO_DPCPP_FLAGS ${GINKGO_COMPILER_FLAGS} -fsycl -fsycl-device-lib=libc)
+set(GINKGO_DPCPP_FLAGS ${GINKGO_COMPILER_FLAGS} -fsycl)
 set(GINKGO_DPCPP_FLAGS ${GINKGO_DPCPP_FLAGS} PARENT_SCOPE)
 target_compile_options(ginkgo_dpcpp PRIVATE "${GINKGO_DPCPP_FLAGS}")
 target_compile_features(ginkgo_dpcpp PRIVATE cxx_std_17)
+target_link_options(ginkgo_dpcpp PRIVATE -fsycl-device-lib=all)
```
yhmtsai, line 58: Should we provide option for DPCPP_JIT_OPTION or DPCPP_LINK_OPTION?
With specific JIT_OPTION can allow us to compile full ginkgo with double support but does not face the double emulation issue on float-only device

dpcpp/test/base/executor.dp.cpp
```diff
@@ -60,19 +60,23 @@ class DpcppExecutor : public ::testing::Test {
...
     }
 
     void TearDown()
     {
-        if (dpcpp != nullptr) {
-            // ensure that previous calls finished and didn't throw an error
-            ASSERT_NO_THROW(dpcpp->synchronize());
-        }
+        // ensure that previous calls finished and didn't throw an error
+        ASSERT_NO_THROW(dpcpp->synchronize());
```
yhmtsai, line 79:
```suggestion
        ASSERT_NO_THROW(dpcpp->synchronize());
        ASSERT_NO_THROW(dpcpp2->synchronize());
```

yhmtsai: Does it not need to check the dpcpp is existed? or will it stop at throw error.

dpcpp/test/base/executor.dp.cpp
```diff
@@ -208,30 +200,6 @@ TEST_F(DpcppExecutor, CopiesDataToCPU)
     dpcpp->free(copy);
 }
 
-
-TEST_F(DpcppExecutor, CopiesDataToGPU)
```
yhmtsai, line 212: Should we keep this?
creating executor with arg "cpu" or "gpu" ignore the SYCCL_DEVICE_TYPE

yhmtsai (changes requested):
LGTM in general. some concerns are in the comments

dpcpp/CMakeLists.txt
tcojean: I do not understand what this is, what option do you mean from the compiler?

yhmtsai: `-fsycl-device-code-split=per_kernel` allow us to compile ginkgo dpcpp with double kernel. It only gives error when the program really use the kernel with double.

upsj (approved):
LGTM! The DPCPP_SINGLE_MODE integration is surprisingly smooth, I like it!

tcojean:
format!

tcojean:
format!

tcojean:
format!

tcojean:
The floating point issues of the dpcpp compiler has been solved by using `-ffp-model=precise` everywhere. It's still surprising to me that `-O2` breaks the floating point standard, but maybe that's the way they want to go.

upsj:
Sound similar to what you need to do in GLSL shaders to enforce bitwise equality between floats: https://www.khronos.org/opengl/wiki/Type_Qualifier_(GLSL)#Precise_qualifiers. By default, they don't adhere strictly to IEEE 754 specifications.

dpcpp/CMakeLists.txt
```diff
@@ -51,11 +51,16 @@ target_sources(ginkgo_dpcpp
...
+target_link_options(ginkgo_dpcpp PRIVATE -fsycl-device-code-split=per_kernel)
```
yhmtsai, line 59: dpcpp also provides ahead-of-time (AOT) to select the device and compile the assembly to eliminate the JIT time.
but I need to check whether AOT allowed some unused illegal subgroup function existing.

dpcpp/CMakeLists.txt
```diff
@@ -51,11 +51,16 @@ target_sources(ginkgo_dpcpp
...
 target_link_libraries(ginkgo_dpcpp PUBLIC ginkgo_device)
+if (GINKGO_DPCPP_SINGLE_MODE)
```
yhmtsai, line 61: Because `-fsycl-device-code-split=per_kernel`, should we separate the SINGLE_MODE to build/test usage?
To test full build on float-only device

include/ginkgo/core/base/math.hpp
```diff
@@ -98,6 +98,20 @@ using std::sqrt;
 }  // namespace kernels
 
 
+namespace dpcpp {
+namespace reference {
```
yhmtsai, line 102: where does the code use this namespace?

dpcpp/test/base/executor.dp.cpp
```diff
@@ -60,31 +60,42 @@ class DpcppExecutor : public ::testing::Test {
 
     void SetUp()
     {
-        ASSERT_GT(gko::DpcppExecutor::get_num_devices("cpu"), 0);
-        dpcpp = gko::DpcppExecutor::create(0, omp, "cpu");
         if (gko::DpcppExecutor::get_num_devices("gpu") > 0) {
```
yhmtsai, line 63: Is `get_num_devices()` affected by DEVICE_FILTER?

yhmtsai (changes requested):

.gitlab-ci.yml
```diff
@@ -680,19 +681,51 @@ build/nocuda/intel/omp/release/static:
...
+    - .quick_test_condition
+    - .use_gko-oneapi-dgpu
+  variables:
+    <<: *default_variables
+    C_COMPILER: "gcc"
+    CXX_COMPILER: "dpcpp"
+    BUILD_DPCPP: "ON"
+    BUILD_TYPE: "Debug"
+    DPCPP_SINGLE_MODE: "ON"
+    SYCL_DEVICE_FILTER: "Level_Zero:GPU"
```
yhmtsai, line 727: When I test it on devcloud, the Level_Zero:GPU gives me FPGA acc not gpu.

CMakeLists.txt
```diff
@@ -78,6 +81,7 @@ if(MSVC OR WIN32 OR CYGWIN OR APPLE)
 else()
     option(GINKGO_BUILD_HWLOC "Build Ginkgo with HWLOC. Default is ON. If a system HWLOC is not found, then we try to build it ourselves. Switch this OFF to disable HWLOC." ON)
 endif()
+option(GINKGO_DPCPP_SINGLE_MODE "Do not compile double kernels for the DPC++ backend." OFF)
```
yhmtsai, line 84: Adding it into minimal or detail log will be good.

dpcpp/CMakeLists.txt
tcojean: I think that mode will be quite complicated to setup, particularly since there is nothing to help set it up in CMake, thus it would require architecture detection like the CAS and possibly other additions as well?

dpcpp/CMakeLists.txt
tcojean: Sorry, I do not understand this one.

dpcpp/test/base/executor.dp.cpp
tcojean: Everything is affected by `SYCL_DEVICE_FILTER` or `SYCL_DEVICE_TYPE`. If you set `CPU` as the filter/type, there will always be 0 GPU found, which is why I need to write this test like this.

include/ginkgo/core/base/math.hpp
tcojean: I failed when copy/pasting, thanks!

tcojean:
This is now fully working and ready for review again. It also fixes a few bugs in the CMake setup.

dpcpp/CMakeLists.txt
yhmtsai: we add the link flag into ginkgo library, which allows us to use single kernel (test) on single-only device even if compiling ginkgo with double kernel

dpcpp/test/base/executor.dp.cpp
yhmtsai: I see. Thanks

.gitlab-ci.yml
yhmtsai: it's related to dpcpp version. the filter works after 2021.2.0 (included)

.gitlab/image.yml
```diff
@@ -75,9 +75,21 @@
...
     - private_ci
-    - bwuni
+    - intelci-igpu
     - cpu
+
+.use_gko-oneapi-igpu:
+  image: ginkgohub/oneapi:latest
+  tags:
+    - private_ci
+    - intel-igpu
```
yhmtsai, line 89: one is intelci-igpu but another is intel-igpu

dpcpp/CMakeLists.txt
yhmtsai: Yes, It should need more changes. the current version is enough for our usage. 
AOT may be helpful to detect error for those devices we do not have or eliminate the JIT time in the future

hip/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 if (GINKGO_HIP_PLATFORM MATCHES "${HIP_PLATFORM_NVIDIA_REGEX}"
-    AND CMAKE_CUDA_COMPILER_VERSION VERSION_LESS 9.2)
+    AND GINKGO_BUILD_CUDA AND CMAKE_CUDA_COMPILER_VERSION VERSION_LESS 9.2)
```
yhmtsai, line 2: could we have method to check nvcc version from hip? it will skip the check when disabling the `GINKGO_BUILD_CUDA`

yhmtsai (approved):
LGTM

include/ginkgo/core/base/math.hpp
```diff
@@ -913,13 +927,7 @@ GKO_INLINE GKO_ATTRIBUTES constexpr xstd::enable_if_t<is_complex_s<T>::value,
                                                       remove_complex<T>>
 abs(const T &x)
 {
-#ifdef CL_SYCL_LANGUAGE_VERSION
-    // FIXME: This implementation is due to a DPC++ issue:
-    // plain `sqrt` call evaluates to `std::sqrt` which fails on GPUs
-    return cl::sycl::sqrt(squared_norm(x));
-#else
     return sqrt(squared_norm(x));
```
yhmtsai, line 930: I am not sure absolute array will call this or not due to namespace issue from dpcpp (sycl).
Also, if the sqrt is not the `std::sqrt` from `kernels::dpcpp`, it will give type error.
I write the above only as note because absolute_array test seems to be passed correctly

hip/CMakeLists.txt
tcojean: In that case, we already do that:
`find_package(CUDA 9.2 REQUIRED)`

.gitlab/image.yml
tcojean: In the second case any IGPU would work, but for the first one I want to make sure the job is executed on that machine so that I can reevaluate where to put this job if we get another Intel CPU.
For example, if we have a runner on an Intel Cascadelake or another new CPU, this might be a good place to try to run this job on.

include/ginkgo/core/base/math.hpp
tcojean: Yes that's a good point, my expectation is now things are fixed and the `sqrt` is actually the `std::sqrt` we specified above. If you have any issue in practice it's good to keep in mind that this may have to change.

tcojean:
The recent logger merge is creating a lot of failures everywhere, I will try to fix this.

tcojean:
format!

.gitlab/image.yml
yhmtsai: I see.

PR #738, opened 2021-04-08
Update format scripts

upsj:
* update to the latest git-cmake-format
* bump the clang-format requirements
* ignore deleted files in formatting bot
* update clang-format version to 9 in bot, since 8 got deleted

### DIFF
--- a/.github/bot-pr-base.sh
+++ b/.github/bot-pr-base.sh
@@ -66,7 +66,7 @@ bot_get_all_changed_files() {
   while true; do
     # this api allows 100 items per page
     # github action uses `bash -e`. The last empty page will leads jq error, use `|| :` to ignore the error.
-    local pr_page_files=$(api_get "$pr_url/files?&per_page=100&page=${page}" | jq -er '.[] | .filename' || :)
+    local pr_page_files=$(api_get "$pr_url/files?&per_page=100&page=${page}" | jq -er '.[] | select(.status != "removed") | .filename' || :)
     if [ "${pr_page_files}" = "" ]; then
       break
     fi
--- a/third_party/git-cmake-format/CMakeLists.txt
+++ b/third_party/git-cmake-format/CMakeLists.txt
@@ -1,7 +1,7 @@
 message(STATUS "Configuring and building git-cmake-format")
 ginkgo_load_git_package(git-cmake-format
     "https://github.com/ginkgo-project/git-cmake-format.git"
-    "29c23665d624e1cae1308bec651706fdaa8fe38b"
-    "-DGCF_CLANGFORMAT_MINIMAL_VERSION=5.0.0")
+    "be9554a5d71030dcc23f1eb355c41ab9d876e776"
+    "-DGCF_CLANGFORMAT_MINIMAL_VERSION=8.0.0")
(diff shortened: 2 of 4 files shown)


tcojean (approved):
LGTM

upsj:
check https://github.com/upsj/ginkgo-bot-playground/pull/12 to see it working :)

third_party/git-cmake-format/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 message(STATUS "Configuring and building git-cmake-format")
 ginkgo_load_git_package(git-cmake-format
     "https://github.com/ginkgo-project/git-cmake-format.git"
-    "29c23665d624e1cae1308bec651706fdaa8fe38b"
-    "-DGCF_CLANGFORMAT_MINIMAL_VERSION=5.0.0")
```
yhmtsai, line 5: Should we change it from 5 to 8?
5 works most of ginkgo but some corner cases are different from 8.
If we decide to require 8+, also need to update the ReadME

yhmtsai (approved):
LGTM. but updating the minimal requirement of clang-format may be unnecessary.
Otherwise, updating the README is required

.github/bot-pr-base.sh
```diff
@@ -66,7 +66,7 @@ bot_get_all_changed_files() {
   while true; do
     # this api allows 100 items per page
     # github action uses `bash -e`. The last empty page will leads jq error, use `|| :` to ignore the error.
-    local pr_page_files=$(api_get "$pr_url/files?&per_page=100&page=${page}" | jq -er '.[] | .filename' || :)
+    local pr_page_files=$(api_get "$pr_url/files?&per_page=100&page=${page}" | jq -er '.[] | select(.status != "removed") | .filename' || :)
```
yhmtsai, line 69: Do you know whether the deleted file is always in the end of list?
if one pages contains 100 deleted files, it will stop earlier than we thought.

third_party/git-cmake-format/CMakeLists.txt
upsj: Good point! Since clang-format only matters for `GINKGO_DEVEL_TOOLS=ON`, I would still require 8 though, otherwise code needs to be reformatted every time somebody creates a PR with one of those files changed.

.github/bot-pr-base.sh
upsj: Yes, but this would only happen if all files on a page are deleted, which is very unlikely. I was a bit too lazy to make it fully correct 😁

pratikvn (approved):

.github/bot-pr-base.sh
yhmtsai: Yes. it is pretty corner cases. I am totally fine with the current changes

third_party/git-cmake-format/CMakeLists.txt
yhmtsai: make sense. thanks

thoasm (commented):
LGTM!

thoasm (approved):

upsj:
rebase!

PR #739, opened 2021-04-09
Add CCACHE support in CI

tcojean:
CCACHE is in fact already available, but a few cosmetic changes are required.

Most of the support is provided in the background by the gitlab-runner setup. You can see its effect on a few jobs where the cache was already built like all jobs on the AMD machine:
https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/1166039002 (2 minutes to compile/test ginkgo core).

What this PR does on top of it:
+ Propagate some CCACHE environment variables set by gitlab-runner to ensure they are available/used
+ Review the way developer tools information is printed and add a bunch of CCACHE related information.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -34,6 +34,8 @@ stages:
   - export NUM_CORES=${CI_PARALLELISM}
   - export OMP_NUM_THREADS=${NUM_CORES}
   - export CUDA_VISIBLE_DEVICES=0
+  - export CCACHE_DIR=${CCACHE_DIR}
+  - export CCACHE_MAXSIZE=${CCACHE_MAXSIZE}
 
 .before_script_git_template: &git_before_script
   - eval $(ssh-agent -s)
--- a/cmake/get_info.cmake
+++ b/cmake/get_info.cmake
@@ -77,10 +77,10 @@ function(ginkgo_print_env_variable log_type var_name)
     string(SUBSTRING
         "
 --        ${var_name}:                                                          " 0 55 upd_string)
-    if(NOT ENV{${var_name}})
-        set(str_value "<empty>")
-    else()
+    if(DEFINED ENV{${var_name}})
         set(str_value "$ENV{${var_name}}")
+    else()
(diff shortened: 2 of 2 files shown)


tcojean:
I will write a separate writeup on how this works behind the scene. Together with other CI improvements, our setup is quite complex so a general review document is long overdue at least for safekeeping...

upsj (approved):
LGTM!

pratikvn (approved):
LGTM! Do we have some numbers on how much we save for CUDA/HIP builds ? Or do we need to have a later version of ccache for that ? 

There are a few places where we still use the deprecated `TYPED_TEST_CASE`. Can we update that to `TYPED_TEST_SUITE` ? 

typed-test-case

upsj:
@pratikvn that is already part of another PR, #721 I think. HIP sadly won't get any performance improvements due to their custom build system, but CUDA might :)

yhmtsai:
Is the CCACHE_DIR unique for different compiler and library?
Is the CCACHE_DIR read-only in these job? there is no race condition when two jobs with the same property compile.

tcojean:
The CCACHE DIR is read-write, the jobs need to update the cache to the last state.

In short:
+ CCACHE uses the build directories (fullpaths) in their hash: different jobs (which use different build directories) create different entries in CCACHE, see [this link](https://ccache.dev/manual/4.2.html#_compiling_in_different_directories)
+ CCACHE supports concurrent write, can be shared, and the default mode takes precaution against the few race conditions which can happen, see  [this link](https://ccache.dev/manual/4.2.html#_sharing_a_cache).

tcojean:
About speedups on CUDA, see this job: https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/1166112739
5 minutes to build Ginkgo with CUDA and OpenMP. The testing seems to take the most time now, probably because of contention on the CUDA card.

yhmtsai (approved):
@tcojean , thanks for explanation. LGTM.

PR #742, opened 2021-04-11
Decrease test size for a few OpenMP tests

upsj:
Now that the builds are blazingly fast thanks to #739 the tests are starting to dominate our runtime, and there are a few outliers that IMO don't justify their long runtime.

This PR adds the (off-by-default) CMake flag GINKGO_FAST_TESTS that reduces the input size for a few tests known to be slow and uses it in Debug CI jobs. An example why I would consider this necessary is https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/1169715092 which takes more than 2 hours to complete.

The tests in question are:

GPU
* ParILUT
* CSR
* Diagonal
* (CB)GMRES

OpenMP
* ParICT
* ParILUT
* CSR
* Dense
* Diagonal
* (CB)GMRES

The reason I would not consider this an issue is that

1. CUDA compiles the kernels in Release mode independent of the CMake build type, so we would catch failures with Release builds, too
2. OpenMP doesn't have any parameters like block sizes we need to exceed to cover all possible scenarios.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,7 @@ option(GINKGO_BUILD_DPCPP
 option(GINKGO_BUILD_CUDA "Compile kernels for NVIDIA GPUs" ${GINKGO_HAS_CUDA})
 option(GINKGO_BUILD_HIP "Compile kernels for AMD or NVIDIA GPUs" ${GINKGO_HAS_HIP})
 option(GINKGO_BUILD_DOC "Generate documentation" OFF)
+option(GINKGO_FAST_TESTS "Reduces the input size for a few tests known to be time-intensive" OFF)
 option(GINKGO_SKIP_DEPENDENCY_UPDATE
     "Do not update dependencies each time the project is rebuilt" ON)
 option(GINKGO_EXPORT_BUILD_DIR
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -143,6 +143,7 @@ if(GINKGO_BUILD_TESTS)
 
     enable_testing()
     include(CTest)
+    add_custom_target(quick_test "${CMAKE_CTEST_COMMAND}" -R 'core|reference')
 endif()
 
 if(GINKGO_WITH_CLANG_TIDY)
--- a/cuda/test/matrix/csr_kernels.cpp
+++ b/cuda/test/matrix/csr_kernels.cpp
@@ -67,7 +67,14 @@ protected:
     using ComplexVec = gko::matrix::Dense<std::complex<double>>;
     using ComplexMtx = gko::matrix::Csr<std::complex<double>>;
 
-    Csr() : mtx_size(532, 231), rand_engine(42) {}
+    Csr()
+#ifdef GINKGO_FAST_TESTS
+        : mtx_size(152, 231),
+#else
+        : mtx_size(532, 231),
+#endif
+          rand_engine(42)
+    {}
 
     void SetUp()
     {
@@ -464,10 +471,10 @@ TEST_F(Csr, AdvancedApplyToIdentityMatrixIsEquivalentToRef)
 TEST_F(Csr, ApplyToComplexIsEquivalentToRef)
 {
     set_up_apply_data(std::make_shared<Mtx::automatical>());
-    auto complex_b = gen_mtx<ComplexVec>(231, 3, 1);
+    auto complex_b = gen_mtx<ComplexVec>(this->mtx_size[1], 3, 1);
     auto dcomplex_b = ComplexVec::create(cuda);
     dcomplex_b->copy_from(complex_b.get());
-    auto complex_x = gen_mtx<ComplexVec>(532, 3, 1);
+    auto complex_x = gen_mtx<ComplexVec>(this->mtx_size[0], 3, 1);
     auto dcomplex_x = ComplexVec::create(cuda);
     dcomplex_x->copy_from(complex_x.get());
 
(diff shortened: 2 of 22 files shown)


tcojean (approved):
LGTM!

Thanks for looking into this. My understanding of the problem is that one big contention point now is the GPU card which is stressed more often (in proportion) during a job lifetime due to the compilation being much faster. I've seen a few times where concurrent access to the card by ParILU/GMRES tests created a deadly combo making some test case take upwards of 5 minutes or more for no reason. I think this approach is a good one, since it reduces the chance of this happening by reducing the time spent on the GPU card by some of the jobs.

I wonder if there would be a better way to do the GPU sharing setup, such that we have real time-sharing for example. Maybe this approach is enough already, though. We will see during real usage.

upsj:
I think the more important part are the slow Debug Reference kernels, since the CUDA GPU kernels are equivalent between Reference and Debug, but we don't see the same issue in Reference jobs.

thoasm (approved):
LGTM!

pratikvn (approved):
LGTM!

upsj:
@pratikvn I added your suggestion for a `quick_test` target, what do you think about the realization?

pratikvn:
Thanks. That looks good to me.

cuda/test/matrix/csr_kernels.cpp
```diff
@@ -67,7 +67,14 @@ class Csr : public ::testing::Test {
     using ComplexVec = gko::matrix::Dense<std::complex<double>>;
     using ComplexMtx = gko::matrix::Csr<std::complex<double>>;
 
-    Csr() : mtx_size(532, 231), rand_engine(42) {}
+    Csr()
+#ifdef GINKGO_FAST_TESTS
+        : mtx_size(152, 231),
```
yhmtsai, line 72: Should we keep the same property, nrows > ncols, in these matrices?

hip/test/matrix/csr_kernels.hip.cpp
```diff
@@ -67,7 +67,14 @@ class Csr : public ::testing::Test {
     using ComplexVec = gko::matrix::Dense<std::complex<double>>;
     using ComplexMtx = gko::matrix::Csr<std::complex<double>>;
 
-    Csr() : mtx_size(532, 231), rand_engine(42) {}
+    Csr()
+#ifdef GINKGO_FAST_TESTS
+        : mtx_size(152, 231),
```
yhmtsai, line 72: also here

hip/test/matrix/diagonal_kernels.hip.cpp
```diff
@@ -62,7 +62,14 @@ class Diagonal : public ::testing::Test {
     using Dense = gko::matrix::Dense<ValueType>;
     using ComplexDiag = gko::matrix::Diagonal<ComplexValueType>;
 
-    Diagonal() : mtx_size(532, 231), rand_engine(42) {}
+    Diagonal()
+#ifdef GINKGO_FAST_TESTS
+        : mtx_size(152, 231),
```
yhmtsai, line 67: also

omp/test/factorization/par_ilut_kernels.cpp
```diff
@@ -72,7 +72,11 @@ class ParIlut : public ::testing::Test {
     using Csr = gko::matrix::Csr<value_type, index_type>;
 
     ParIlut()
+#ifdef GINKGO_FAST_TESTS
+        : mtx_size(152, 231),
```
yhmtsai, line 76: also

omp/test/matrix/csr_kernels.cpp
```diff
@@ -67,7 +67,14 @@ class Csr : public ::testing::Test {
     using ComplexVec = gko::matrix::Dense<std::complex<double>>;
     using ComplexMtx = gko::matrix::Csr<std::complex<double>>;
 
-    Csr() : mtx_size(532, 231), rand_engine(42) {}
+    Csr()
+#ifdef GINKGO_FAST_TESTS
+        : mtx_size(152, 185),
```
yhmtsai, line 72: also

omp/test/matrix/diagonal_kernels.cpp
```diff
@@ -62,7 +62,14 @@ class Diagonal : public ::testing::Test {
     using Dense = gko::matrix::Dense<ValueType>;
     using ComplexDiag = gko::matrix::Diagonal<ComplexValueType>;
 
-    Diagonal() : mtx_size(532, 231), rand_engine(42) {}
+    Diagonal()
+#ifdef GINKGO_FAST_TESTS
+        : mtx_size(152, 185),
```
yhmtsai, line 67: same here

CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ option(GINKGO_BUILD_DPCPP
 option(GINKGO_BUILD_CUDA "Compile kernels for NVIDIA GPUs" ${GINKGO_HAS_CUDA})
 option(GINKGO_BUILD_HIP "Compile kernels for AMD or NVIDIA GPUs" ${GINKGO_HAS_HIP})
 option(GINKGO_BUILD_DOC "Generate documentation" OFF)
+option(GINKGO_FAST_TESTS "Reduces the input size for a few tests known to be time-intensive" OFF)
```
yhmtsai, line 35: Could you also add it into INSTALL.md?

yhmtsai (approved):
LGTM in general. should we keep the test matrices structure like nrows > ncols?

cuda/test/matrix/csr_kernels.cpp
upsj: I'm not sure it matters here, since mostly the rows are handled independently, and having a more diverse set of test cases shouldn't hurt?

CMakeLists.txt
pratikvn: And also to the `get_info.cmake` ?

cuda/test/matrix/csr_kernels.cpp
yhmtsai: it should not hurt anything if still gives the same residual or uses the same strategy during the computing.
When having a diverse set in fast_test, I would consider whether we should include it in normal test.

tcojean: You raise a good point in the datasets we use for testing, but I think it's outside the scope of this PR. Technically for at least some kernels, we might need to test matrices with different properties, shape, sizes as well, ... I don't think it hurts if we do it like this for now, the fast test testing ncols > nrows and normal tests nrows > ncols. Later we might have to think if it makes sense to do a general dataset improvement.

yhmtsai: I see.

PR #743, opened 2021-04-14
Fix criterion check logger

upsj:
The logger::on_criterion_check_complete function accidentally calls itself recursively. This PR adds a test to detect that behavior and fixes it together with a `partially overridden` warning in record.hpp


pratikvn (approved):
LGTM! Good catch! Thanks for fixing it.

thoasm (approved):
LGTM!

PR #737, opened 2021-04-07
Segment CI runs into quick and full runs

upsj:
This PR segments CI pipeline that only runs the full job pipeline on demand.

After this changes:
- Uses `ST:run-full-test` or `ST:ready-for-merge` to activate the remaining pipeline
- there will be three pipeline related to gitlab, quick/full/gitlab. note gitlab status might be quick or full, which depends on the last one of quick/full finishes the job.
ref:
https://gitlab.com/gitlab-org/gitlab/-/issues/216629
https://docs.github.com/en/rest/reference/repos#statuses
- extract the condition and image to another file with `extends:` usage

PR #744, opened 2021-04-15
Fix ROCm device visibility in CI

upsj:
This is an attempt to fix the random issues with hidden GPUs due to CUDA_VISIBLE_DEVICES being interpreted by ROCm as well.


yhmtsai:
does gitlab ci variable allow RANDOM?
By the way, it is fixed in #737 by different implementation

upsj:
@yhmtsai yes, looks like it doesn't work. I had hoped this would be simple text replacement, but looks like gitlab does something more complex there.

PR #750, opened 2021-04-26
Add exact forward error computation for trisolve benchmark

upsj:
Checking the triangular solvers for correctness is a bit annoying, since some of the triangular systems we are checking are very ill-conditioned. Thus this PR adds a direct forward error calculation for these solvers based on the reference executor, which should give bitwise identical results for all straightforward trisolve implementations (not including blocked and warp-parallel solves here)

- [x] Merge #753

### DIFF
--- a/benchmark/solver/solver.cpp
+++ b/benchmark/solver/solver.cpp
@@ -476,14 +476,15 @@ void solve_system(const std::string& solver_name,
         auto generate_timer = get_timer(exec, FLAGS_gpu_timer);
         auto apply_timer = ic.get_timer();
         auto x_clone = clone(x);
+        std::shared_ptr<const gko::LinOp> solver;
         for (auto status : ic.run(false)) {
             x_clone = clone(x);
 
             exec->synchronize();
             generate_timer->tic();
             auto precond = precond_factory.at(precond_name)(exec);
-            auto solver = generate_solver(exec, give(precond), solver_name)
-                              ->generate(system_matrix);
+            solver = generate_solver(exec, give(precond), solver_name)
+                         ->generate(system_matrix);
             generate_timer->toc();
 
             exec->synchronize();
@@ -492,6 +493,14 @@ void solve_system(const std::string& solver_name,
             apply_timer->toc();
         }
         if (b->get_size()[1] == 1 && !FLAGS_overhead) {
+            // a solver is considered direct if it didn't log any iterations
+            if (solver_json["apply"].HasMember("iterations") &&
+                solver_json["apply"]["iterations"].GetInt() == 0) {
+                auto error =
+                    compute_direct_error(lend(solver), lend(b), lend(x_clone));
+                add_or_set_member(solver_json, "forward_error", error,
+                                  allocator);
+            }
             auto residual = compute_residual_norm(lend(system_matrix), lend(b),
                                                   lend(x_clone));
             add_or_set_member(solver_json, "residual_norm", residual,
--- a/benchmark/utils/general.hpp
+++ b/benchmark/utils/general.hpp
@@ -413,7 +413,7 @@ std::unique_ptr<vec<ValueType>> create_vector(
 template <typename ValueType>
 ValueType get_norm(const vec<ValueType>* norm)
 {
-    return clone(norm->get_executor()->get_master(), norm)->at(0, 0);
+    return norm->get_executor()->copy_val_to_host(norm->get_const_values());
 }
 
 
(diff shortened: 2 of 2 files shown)


benchmark/utils/general.hpp
```diff
@@ -392,6 +392,21 @@ gko::remove_complex<ValueType> compute_norm2(const vec<ValueType> *b)
...
+                                                    const vec<ValueType> *b,
+                                                    const vec<ValueType> *x)
+{
+    auto exec = solver->get_executor();
+    auto ref_solver = gko::clone(gko::ReferenceExecutor::create(), solver);
+    auto one = gko::initialize<vec<ValueType>>({1.0}, exec);
+    auto neg_one = gko::initialize<vec<ValueType>>({-1.0}, exec);
+    auto res = clone(x);
+    ref_solver->apply(lend(one), lend(b), lend(neg_one), lend(res));
+    return compute_norm2(lend(res));
```
pratikvn, line 406: I think you should clone `res` also to be on reference executor ? Otherwise the norm2 gets computed on the main executor and might have differences due to partial reduction orderings ?

pratikvn (commented):
Nice! Maybe this is useful in all circumstances. It might be nice to have it as a user configurable on/off parameter for all solvers ?

benchmark/utils/general.hpp
```diff
@@ -392,6 +392,22 @@ gko::remove_complex<ValueType> compute_norm2(const vec<ValueType> *b)
...
+gko::remove_complex<ValueType> compute_direct_error(const gko::LinOp *solver,
+                                                    const vec<ValueType> *b,
+                                                    const vec<ValueType> *x)
+{
+    auto ref_exec = gko::ReferenceExecutor::create();
+    auto exec = solver->get_executor();
+    auto ref_solver = gko::clone(ref_exec, solver);
+    auto one = gko::initialize<vec<ValueType>>({1.0}, exec);
+    auto neg_one = gko::initialize<vec<ValueType>>({-1.0}, exec);
+    auto res = gko::clone(ref_exec, x);
```
Slaedr, line 405:
```suggestion
    auto error = gko::clone(ref_exec, x);
```

benchmark/utils/general.hpp
```diff
@@ -392,6 +392,22 @@ gko::remove_complex<ValueType> compute_norm2(const vec<ValueType> *b)
...
+    ref_solver->apply(lend(one), lend(b), lend(neg_one), lend(res));
+    return compute_norm2(lend(res));
```
Slaedr, line 407: It would be nice to just return the error vector here as a unique_ptr, and compute the norm in the `if(is_direct_solver)` block in the solver.cpp file. In some cases, it might be of interest to have access to the actual error vector.

benchmark/solver/solver.cpp
```diff
@@ -490,6 +490,15 @@ void solve_system(const std::string &solver_name,
 
             if (b->get_size()[1] == 1 && i == FLAGS_repetitions - 1 &&
                 !FLAGS_overhead) {
+                // a solver is considered direct if it didn't log any iterations
+                auto is_direct_solver =
+                    solver_json["true_residuals"].GetArray().Size() == 0;
+                if (is_direct_solver) {
+                    auto error = compute_direct_error(lend(solver), lend(b),
+                                                      lend(x_clone));
+                    add_or_set_member(solver_json, "forward_error", error,
```
Slaedr, line 499: It would be nice to separately compute the error norm here.

Slaedr (commented):
Great idea! I think it might be better for the new function to return the actual error vector though, and we compute the norm outside separately.

upsj:
@Slaedr What use case would you have in mind for returning the full forward error vector? The benchmarks only output a json file, so there is not much we can do with this data, unless you want to use it for debugging?

upsj:
Turns out that there is a whole lot more going on here - simply cloning a solver doesn't work, because that doesn't clone the preconditioner and system matrix correctly. So this will take a lot more work, and I'll have to think a bit about how to approach this most consistently. It will probably involve breaking our current interface.

pratikvn:
I think you are free to break the benchmark interface, as that is not a part of our release.

upsj:
@pratikvn The issue is much deeper than that, we can't clone solvers across executors, and any suitable solution I can imagine involves at least some interface changes :-/

Slaedr:
@upsj It's fine, I guess, but there might be other functionals of interest, not just norms. For example, you might want the dominant frequency for some analyses.

pratikvn:
@Slaedr , maybe you mean write the error to a file so that it can be post-processed later ?

upsj:
What I had in mind for this option is a quick check whether the result of a triangular solve is correct, as for badly conditioned matrices, the residual is not really useful, but the forward error compared to Reference is usually 0 or very small depending on whether the triangular solve uses one or multiple threads per row.

Slaedr:
Yes, an option for that would nice. Or, maybe we add functions to compute different kinds of functionals in the future as the need arises, and we keep the mechanism flexible for that.

pratikvn (approved):
LGTM!

benchmark/utils/general.hpp
```diff
@@ -392,6 +392,22 @@ gko::remove_complex<ValueType> compute_norm2(const vec<ValueType> *b)
...
+                                                    const vec<ValueType> *b,
+                                                    const vec<ValueType> *x)
+{
+    auto ref_exec = gko::ReferenceExecutor::create();
+    auto exec = solver->get_executor();
+    auto ref_solver = gko::clone(ref_exec, solver);
+    auto one = gko::initialize<vec<ValueType>>({1.0}, exec);
+    auto neg_one = gko::initialize<vec<ValueType>>({-1.0}, exec);
+    auto res = gko::clone(ref_exec, x);
+    ref_solver->apply(lend(one), lend(b), lend(neg_one), lend(res));
```
Slaedr, line 406: I guess you don't need to wait to resolve the `make_temporary_clone` question; we can update this later once it's resolved.

Slaedr (approved):
LGTM. Since this is only in the benchmarking code, we can change it later if need be to output the error vector etc.

PR #751, opened 2021-04-26
Run the CUDA jobs on the CI controllers.

tcojean:
Run our CI on both ForHLR2 and BWuni.

Most of the setup happened in the background, eitheir through the container generation or to make enroot work, this required environment settings as well as modification of the custom executor scripts. CCACHE support seems to be working. For now, everything runs on my account.

Everything seems to be working other than DPC++ which I removed for now. It will come back as part of #736.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -680,20 +680,6 @@ build/nocuda/intel/omp/release/static:
     BUILD_TYPE: "Release"
     BUILD_SHARED_LIBS: "OFF"
 
-build/dpcpp/cpu/release/static:
-  <<: *default_build_with_test
-  extends:
-    - .quick_test_condition
-    - .use_gko-oneapi
-  variables:
-    <<: *default_variables
-    C_COMPILER: "gcc"
-    CXX_COMPILER: "dpcpp"
-    BUILD_DPCPP: "ON"
-    BUILD_TYPE: "Release"
-    BUILD_SHARED_LIBS: "OFF"
-    SYCL_DEVICE_TYPE: "CPU"
-
 # Job with important warnings as error
 warnings:
   <<: *default_build
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -80,7 +80,7 @@ include:
         -DGINKGO_RUN_EXAMPLES=${RUN_EXAMPLES}
         -DGINKGO_CONFIG_LOG_DETAILED=${CONFIG_LOG}
         -DGINKGO_EXPORT_BUILD_DIR=${EXPORT_BUILD_DIR}
-    - ninja -j${NUM_CORES} -l${CI_LOAD_LIMIT}
+    - ninja -j${NUM_CORES} -l${CI_LOAD_LIMIT} install
     - if [ ! -z ${SYCL_DEVICE_TYPE+x} ]; then unset SYCL_DEVICE_TYPE; fi
     - if [ "${EXPORT_BUILD_DIR}" == "ON" ]; then ninja test_exportbuild; fi
   dependencies: []
@@ -251,7 +251,7 @@ trigger_pipeline:
 # Build jobs
 # Job with example runs.
 build/cuda90/gcc/all/debug/shared:
-  <<: *default_build_with_test
+  <<: *default_build
   extends:
     - .quick_test_condition
     - .use_gko-cuda90-gnu5-llvm39
@@ -265,7 +265,7 @@ build/cuda90/gcc/all/debug/shared:
     CUDA_ARCH: 35
 
 build/cuda90/clang/all/release/static:
-  <<: *default_build_with_test
+  <<: *default_build
   extends:
     - .full_test_condition
     - .use_gko-cuda90-gnu5-llvm39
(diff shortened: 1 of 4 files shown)


.gitlab-ci.yml
```diff
@@ -251,7 +251,7 @@ trigger_pipeline:
 # Build jobs
 # Job with example runs.
 build/cuda90/gcc/all/debug/shared:
-  <<: *default_build_with_test
+  <<: *default_build
```
pratikvn, line 254: I guess you plan to change this back to `default_build_with_test`  ?

.gitlab/image.yml
```diff
@@ -1,82 +1,83 @@
...
+    - cpu
 
 .use_gko-cuda102-gnu8-llvm8-intel2019:
-  image: localhost:5000/gko-cuda102-gnu8-llvm8-intel2019
+  image: ginkgohub/cuda:102-gnu8-llvm8-intel2019
   tags:
     - private_ci
-    - cuda
-    - gpu
+    - controller
```
pratikvn, line 61: So, both the FH2 and the bwuni have the controller tags ?

.gitlab/image.yml
```diff
@@ -1,82 +1,83 @@
...
 
 .use_gko-cuda90-gnu5-llvm39:
-  image: localhost:5000/gko-cuda90-gnu5-llvm39
+  image: ginkgohub/cuda:90-gnu5-llvm39
   tags:
     - private_ci
-    - cuda
-    - gpu
+    - controller
+    - cpu
```
pratikvn, line 20: Is this and the other cuda jobs supposed to be `cpu`

.gitlab-ci.yml
tcojean: For now we cannot, because we do not have a GPU to run CUDA/HIP.

.gitlab/image.yml
tcojean: yes, since I'm using only the controllers now

.gitlab/image.yml
tcojean: Yes, because we cannot run on GPUs with the current settings (it's also tied to `default_build`).

upsj (approved):
LGTM!

third_party/hwloc/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@ else()
     message(STATUS "Downloading HWLOC tars and building")
     ginkgo_load_and_configure_package(hwloc_external "https://download.open-mpi.org/release/hwloc/v2.3/hwloc-2.3.0.tar.gz"
         "SHA1=69d8e2f088d365cf5e86d04f049c6b1e09e2b9c3"
-        "${TPL_HWLOC_PATH}/src/configure"
+        "${TPL_HWLOC_PATH}/src/configure --disable-nvml --disable-cuda --disable-rsmi"
```
yhmtsai, line 18: Will it only consider the cpu?

yhmtsai (approved):
LGTM. I have some nit comments on runner tags.
Could we have more detail on tags such that we can use device property not the specific runner name ?
for example
```
fineci: fineci, privateci, gpu, cpu, intel-cpu, nvidia-gpu
amdci: amdci, privateci, gpu, cpu, amd-cpu, amd-gpu
bwuni: bwuni, privateci, cpu, controller, intel-cpu (I guess?)
fh2: fh2, privateci, cpu, controller, amd-cpu
intelci-igpu: intelci-igpu, privateci, gpu, cpu, intel-cpu, intel-igpu
intelci-dgpu: intelci-dgpu, privateci, gpu, intel-dgpu
```
Not sure whether putting cpu in fineci because we would like to eliminate the load of fineci before.

third_party/hwloc/CMakeLists.txt
tcojean: We do not need the backend support with our current HWLOC implementation, because we only rely on the PCI device ID and the actual runtimes when needed (e.g. CUDA or HIP runtime). Since somehow the automatic support addition was creating issue with some containers, I decided to try to disable it.

.gitlab/image.yml
```diff
@@ -1,82 +1,83 @@
...
 
 .use_gko-amd-gnu8-llvm7:
-  image: localhost:5000/gko-amd-gnu8-llvm7
+  image: ginkgohub/rocm:gnu8-llvm7
   tags:
     - private_ci
-    - amd
+    - amdci
     - gpu
```
yhmtsai, line 76: we can also use `amd-gpu` for these tags if we have more machine equipped with amd-gpu in the future

.gitlab-ci.yml
pratikvn: Don't the bwuni/FH2 controllers have V100's ? Or are those only for Level 3 CI and you are using Level 1 right now ?

upsj: Unfortunately no, those are just rebranded login nodes

pratikvn (approved):
LGTM!

.gitlab-ci.yml
tcojean: We are using Level 2 now, i.e. directly running on the controllers (login nodes). Level 3, we cannot access ourselves and need the admins to register the runner themselves.

PR #752, opened 2021-04-27
Add batching functionality to Ginkgo

pratikvn:
This PR is the start of batching functionality in Ginkgo. It adds

1. A `BatchLinOp` to represent batched operators.
    a. `batch_dim` class to represent `dim` for the different batch objects, both uniform and non-uniform.
    b. `BatchStruct` struct that unifies the entries of the batch for all concrete batched operators.
    c. `BatchEntry` struct that is a wrapper for a single entry of the batched operators providing abstracted access to the underlying concrete batch entry type. 
2. A `BatchDense` class to represent a batch Dense matrix.
    a. `batch_stride` class to represent the batched stride. This follows a similar suite to the `batch_dim` class, where you do not pay for the uniform batch case, but pay an additional storage + validation cost for the non-uniform case. 
    b. `BatchDense` can store non-uniform batches, but currently, the `apply` only supports uniform batches.
3. A `BatchCsr` class to represent a batch CSR matrix.
    a. `BatchCsr` can only store the same sparsity pattern for the different batch entries. The values stored by the different batch entries can be different.
4. Most of the functionality has been tested.

Finally, all this functionality is still experimental and the interface is highly volatile. Use at your own risk in production code!


pratikvn:
format!

pratikvn:
format!

include/ginkgo/core/matrix/batch_dense.hpp
```diff
@@ -0,0 +1,1013 @@
...
+    Array<size_type> num_elems_per_batch_cumul_;
+    Array<value_type> values_;
+};
+
+
+}  // namespace matrix
+
+
+/**
+ * Creates and initializes a column-vector.
```
Slaedr, line 729:
```suggestion
 * Creates and initializes a batch of column-vectors.
```

include/ginkgo/core/matrix/batch_dense.hpp
```diff
@@ -0,0 +1,1013 @@
...
+ * This function first creates a temporary Dense matrix, fills it with passed in
+ * values, and then converts the matrix to the requested type.
+ *
+ * @tparam Matrix  matrix type to initialize
+ *                 (Dense has to implement the ConvertibleTo<Matrix> interface)
+ * @tparam TArgs  argument types for Matrix::create method
+ *                (not including the implied Executor as the first argument)
+ *
+ * @param stride  row stride for the temporary Dense matrix
+ * @param vals  values used to initialize the vector
```
Slaedr, line 740:
```suggestion
 * @param vals  a two-dimensional list of values used to initialize the batch.
 *               Each inner list corresponds to one vector.
```

include/ginkgo/core/matrix/batch_dense.hpp
```diff
@@ -0,0 +1,1013 @@
...
+    std::shared_ptr<const Executor> exec, TArgs &&... create_args)
+{
+    return batch_initialize<Matrix>(std::vector<size_type>(vals.size(), 1),
+                                    vals, std::move(exec),
+                                    std::forward<TArgs>(create_args)...);
+}
+
+
+/**
+ * Creates and initializes a matrix.
```
Slaedr, line 817:
```suggestion
 * Creates and initializes a batch of matrices.
```

include/ginkgo/core/matrix/batch_dense.hpp
```diff
@@ -0,0 +1,1013 @@
...
+        strides[ind] = begin(b)->size();
+        ++ind;
+    }
+    return batch_initialize<Matrix>(strides, vals, std::move(exec),
+                                    std::forward<TArgs>(create_args)...);
+}
+
+
+/**
+ * Creates and initializes a column-vector from copies of a given vector.
```
Slaedr, line 920:
```suggestion
 * Creates and initializes a batch of column-vectors from copies of a given vector.
```

include/ginkgo/core/matrix/batch_dense.hpp
```diff
@@ -0,0 +1,1013 @@
...
+        }
+    }
+    auto mtx = Matrix::create(exec, std::forward<TArgs>(create_args)...);
+    tmp->move_to(mtx.get());
+    return mtx;
+}
+
+
+/**
+ * Creates and initializes a column-vector from copies of a given vector.
```
Slaedr, line 975:
```suggestion
 * Creates and initializes a batch of column-vectors from copies of a given vector.
```

include/ginkgo/core/base/lin_op_helpers.hpp
```diff
@@ -0,0 +1,351 @@
+/*******************************<GINKGO LICENSE>******************************
```
Slaedr, line 1: It does not seem like this file belongs in this PR.

core/test/base/batch_lin_op.cpp
```diff
@@ -0,0 +1,413 @@
...
+                 gko::DimensionMismatch);
+}
+
+
+// For tests between different memory, check cuda/test/base/batch_lin_op.cu
+TEST_F(EnableBatchLinOp, ApplyDoesNotCopyBetweenSameMemory)
+{
+    op->apply(gko::lend(b), gko::lend(x));
+
+    ASSERT_EQ(op->last_b_access, ref);
```
Slaedr, line 300: This is not clear to me. How do we know that no copy happened from this assertion?

include/ginkgo/core/base/dim.hpp
```diff
@@ -200,11 +215,154 @@ struct dim<1u, DimensionType> {
...
+
+    /**
+     * Checks if the batch_dim object stores equal sizes.
+     *
+     * @return bool representing whether equal sizes are being stored
+     */
+    bool stores_equal_sizes() const { return equal_sizes_; }
+
+    /**
+     * Get the number of batches stored
```
Slaedr, line 251:
```suggestion
     * Get the number of operators (individual matrices, vectors etc.) in the batch.
```

include/ginkgo/core/base/dim.hpp
```diff
@@ -200,11 +215,154 @@ struct dim<1u, DimensionType> {
...
+     *
+     * @return num_batches
```
Slaedr, line 253:
```suggestion
     * @return The size of the batch.
```

include/ginkgo/core/base/dim.hpp
```diff
@@ -200,11 +215,154 @@ struct dim<1u, DimensionType> {
...
+     */
+    size_type get_num_batches() const { return num_batches_; }
```
Slaedr, line 255: IMO, "num_batches" is counter intuitive and makes one think of many batches. To have consistent English semantics, we should call this the number of batch-entries, but I felt "batch size" would be better suited for code. That is, a batch is made of `n` matrices, so the size of the batch is `n`.
```suggestion
    size_type get_batch_size() const { return num_batches_; }
```

include/ginkgo/core/base/dim.hpp
```diff
@@ -200,11 +215,154 @@ struct dim<1u, DimensionType> {
...
+
+    /**
+     * Get the batch sizes as a std::vector.
+     *
+     * @return  the std::vector of batch sizes
+     */
+    std::vector<dim<dimensionality, dimension_type>> get_batch_sizes() const
```
Slaedr, line 262:
```suggestion
    std::vector<dim<dimensionality, dimension_type>> get_sizes() const
```
We could also use `get_entry_sizes` if you want to be more specific.

include/ginkgo/core/base/dim.hpp
```diff
@@ -200,11 +215,154 @@ struct dim<1u, DimensionType> {
...
+
+    /**
+     * Get the number of batches stored
+     *
+     * @return num_batches
+     */
+    size_type get_num_batches() const { return num_batches_; }
+
+    /**
+     * Get the batch sizes as a std::vector.
```
Slaedr, line 258:
```suggestion
     * Get the sizes of all batch entries as a std::vector.
```

include/ginkgo/core/base/dim.hpp
```diff
@@ -200,11 +215,154 @@ struct dim<1u, DimensionType> {
...
+                return std::vector<dim<dimensionality, dimension_type>>{
+                    common_size_};
+            }
+        } else {
+            return sizes_;
+        }
+    }
+
+    /**
+     * Get the batch size at a particular index.
```
Slaedr, line 278:
```suggestion
     * Get the size of the entry at a particular batch-index.
```

include/ginkgo/core/base/dim.hpp
```diff
@@ -200,11 +215,154 @@ struct dim<1u, DimensionType> {
...
+     *
+     * @param batch the index whose size is needed
+     *
+     * @return  the size of the batch at the requested index
```
Slaedr, line 282:
```suggestion
     * @return  the size of the entry at the requested index
```

include/ginkgo/core/base/dim.hpp
```diff
@@ -200,11 +215,154 @@ struct dim<1u, DimensionType> {
...
+        } else {
+            return x.sizes_ == y.sizes_;
+        }
+    }
+
+    /**
+     * Creates a batch_dim object which stores uniform batch sizes.
+     *
+     * @param num_batches  number of batches to be stored
+     * @param size  the size of all the batches stored
```
Slaedr, line 317:
```suggestion
     * @param batch_size  number of objects/entries to be stored in the batch.
     * @param size  the common size of all the objects stored
```

include/ginkgo/core/base/dim.hpp
```diff
@@ -200,11 +215,154 @@ struct dim<1u, DimensionType> {
...
+     *
+     * @note  Use this constructor when uniform batches need to be stored.
+     */
+    batch_dim(const size_type num_batches = 0,
```
Slaedr, line 321:
```suggestion
    batch_dim(const size_type batch_size = 0,
```

Slaedr (changes requested):
Great work! I think this is a good base to build the batch capabilities on. Thanks for including a lot of the helper code for making further development easier. Thanks also for adopting and implementing my suggestion about batch_dim.

However, I do have some concerns before we move ahead. One is about the English language terminology we use about batches. See my comments in the `dim.hpp` file. Essentially, "batch" should consistently refer to the entire collection of small objects taken together.

The other point is about changes unrelated to the batch functionality, primarily in lin_op.hpp and lin_op_helpers.hpp. I suspect some of them are there as vestiges from the time when `BatchDense` etc. were `LinOp`s.

In addition, there are several minor comments.

core/matrix/batch_dense.cpp
```diff
@@ -0,0 +1,401 @@
...
+
+
+template <typename ValueType>
+void BatchDense<ValueType>::move_to(BatchCsr<ValueType, int32> *result)
+{
+    this->convert_to(result);
+}
+
+
+// template <typename ValueType>
```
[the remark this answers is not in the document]
pratikvn: Yeah, I guess we can remove this for now

core/test/matrix/batch_dense.cpp
```diff
@@ -0,0 +1,454 @@
...
+
+    auto m = gko::matrix::BatchDense<TypeParam>::create(
+        this->exec,
+        std::vector<gko::dim<2>>{gko::dim<2>{2, 2}, gko::dim<2>{2, 2}},
+        gko::Array<value_type>::view(this->exec, 12, data),
+        std::vector<size_type>{3, 3});
+
+    ASSERT_EQ(m->get_const_values(), data);
+    ASSERT_EQ(m->at(0, 0, 1), value_type{2.0});
+    ASSERT_EQ(m->at(0, 1, 2), value_type{-1.0});
```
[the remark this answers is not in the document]
pratikvn: No, we have the same behaviour for Dense as well, because technically the view has the data, so we should be able to access it ? And also because the stride > number of columns

cuda/test/matrix/batch_dense_kernels.cpp
```diff
@@ -0,0 +1,393 @@
...
+    std::unique_ptr<Mtx> dx;
+    std::unique_ptr<ComplexMtx> dc_x;
+    std::unique_ptr<Mtx> dy;
+    std::unique_ptr<Mtx> dalpha;
+    std::unique_ptr<Mtx> dbeta;
+    std::unique_ptr<Mtx> dsquare;
+};
+
+
+// TEST_F(BatchDense, SingleVectorCudaScaleIsEquivalentToRef)
```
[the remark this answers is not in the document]
pratikvn: Okay, I will remove all the commented tests.

core/matrix/dense_kernels.hpp
```diff
@@ -198,17 +198,17 @@ namespace kernels {
         const matrix::Dense<_vtype> *source,               \
         matrix::Dense<remove_complex<_vtype>> *result)
 
-#define GKO_DECLARE_MAKE_COMPLEX_KERNEL(_vtype)                    \
+#define GKO_DECLARE_MAKE_COMPLEX_DENSE_KERNEL(_vtype)              \
```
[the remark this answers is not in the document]
pratikvn: I think this should be okay, I like it for the clarity. If needed we can move this to another PR, but I dont think it hurts to leave this here.

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -42,6 +42,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/base/abstract_factory.hpp>
 #include <ginkgo/core/base/dim.hpp>
 #include <ginkgo/core/base/exception_helpers.hpp>
+#include <ginkgo/core/base/lin_op_helpers.hpp>
```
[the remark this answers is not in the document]
pratikvn: I split the main lin_op file into two to reduce the duplication.

core/test/base/batch_lin_op.cpp
pratikvn: Because `op` is created on `ref2` and on on `ref`. Therefore we check that b is not copied to `ref2` because `ref` and `ref2` have the same memory space.

include/ginkgo/core/base/lin_op_helpers.hpp
pratikvn: Yes, I needed to do this to reduce duplication for the batch linop.

include/ginkgo/core/base/dim.hpp
pratikvn: I dont think size here is very clear, because size can also mean the dimension of the entries or of the entire batch object. Maybe number of batch entries is better ?

include/ginkgo/core/base/dim.hpp
pratikvn: I guess `get_num_batch_entries` is better, but I feel that `batch_size` is confusing. I will change it to `get_num_batch_entries`

include/ginkgo/core/base/dim.hpp
pratikvn: I think here batch_sizes is appropriate as you are getting the sizes of all the batches in a std::vector.

pratikvn:
@Slaedr , I think most changes should be addressed/replied to, Let me know if you still want to discuss something.

core/test/matrix/batch_dense.cpp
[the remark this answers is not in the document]
Slaedr: Eg. with the row_major and block_col_major accessors, if the the index exceeds the stated size, it throws. If Dense allows it, I guess it' fine. But IMO it should be regarded as an error that we "protect" the user against (in debug builds). If the user created a 2x2 view, they must want to access only those 4 entries through the view.

include/ginkgo/core/base/dim.hpp
Slaedr: It should be fine. I'm just saying that there's only one batch here, not many batches. There are many entries in the batch. But alright, I think in this case get_batch_sizes works regardless.

core/test/matrix/batch_dense.cpp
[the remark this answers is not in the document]
pratikvn: But the thing is the user is not creating a (2x2) view. Here the user is creating a view of 9 elements, and the matrix size is indeed 2x2, but it has a stride of 3.

include/ginkgo/core/base/lin_op_helpers.hpp
```diff
@@ -0,0 +1,351 @@
...
+ * @tparam ParametersType  a subclass of enable_parameters_type template which
+ *                         defines all of the parameters of the factory
+ * @tparam PolymorphicBase  parent of ConcreteFactory in the polymorphic
+ *                          hierarchy, has to be a subclass of LinOpFactory
+ *
+ * @ingroup LinOp
+ */
+template <typename ConcreteFactory, typename ConcreteLinOp,
+          typename ParametersType, typename PolymorphicBase = LinOpFactory>
+using EnableDefaultLinOpFactory =
```
Slaedr, line 76: I guess this can remain in lin_op.hpp, because BatchLinOps don't need this?

include/ginkgo/core/base/lin_op_helpers.hpp
```diff
@@ -0,0 +1,351 @@
...
+ * @param _parameters_name  name of the parameters member in the class
+ *                          (its type is `<_parameters_name>_type`, the
+ *                          protected member's name is `<_parameters_name>_`,
+ *                          and the public getter's name is
+ *                          `get_<_parameters_name>()`)
+ * @param _factory_name  name of the generated factory type
+ *
+ * @ingroup LinOp
+ */
+#define GKO_ENABLE_LIN_OP_FACTORY(_lin_op, _parameters_name, _factory_name)  \
```
Slaedr, line 177: This can also remain in lin_op.hpp as this is not relevant to Batch LinOps.

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -828,292 +829,6 @@ class EnableLinOp
...
- * @tparam ParametersType  a subclass of enable_parameters_type template which
- *                         defines all of the parameters of the factory
- * @tparam PolymorphicBase  parent of ConcreteFactory in the polymorphic
- *                          hierarchy, has to be a subclass of LinOpFactory
- *
- * @ingroup LinOp
- */
-template <typename ConcreteFactory, typename ConcreteLinOp,
-          typename ParametersType, typename PolymorphicBase = LinOpFactory>
-using EnableDefaultLinOpFactory =
```
Slaedr, line 850: This can still remain here, though.

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -828,292 +829,6 @@ class EnableLinOp
...
- * @param _parameters_name  name of the parameters member in the class
- *                          (its type is `<_parameters_name>_type`, the
- *                          protected member's name is `<_parameters_name>_`,
- *                          and the public getter's name is
- *                          `get_<_parameters_name>()`)
- * @param _factory_name  name of the generated factory type
- *
- * @ingroup LinOp
- */
-#define GKO_ENABLE_LIN_OP_FACTORY(_lin_op, _parameters_name, _factory_name)  \
```
Slaedr, line 950: And this can also be kept in this file, as it's specific to LinOps.

core/test/matrix/dense.cpp
```diff
@@ -135,6 +135,7 @@ TYPED_TEST(Dense, CanBeConstructedFromExistingData)
 
     ASSERT_EQ(m->get_const_values(), data);
     ASSERT_EQ(m->at(2, 1), value_type{6.0});
+    ASSERT_EQ(m->at(2, 2), value_type{-1.0});
```
Slaedr, line 138: Honestly I don't think we should be testing for this. If a developer implements something that violates this somehow, I don't think that should be counted as a bug. I agree that we should not be modifying data that falls outside this matrix, so m[2,2] should still be -1. But I'd rather check it as follows:
```suggestion
    ASSERT_EQ(m->get_const_values()[2*m->get_stride() + 2], value_type{-1.0});
```
That is, the underlying array should be unchanged were the view is not supposed to touch it, but we don't guarantee access to it through the view. That said, I'm okay with whatever you decide here.

Slaedr (approved):
LGTM, thanks a lot! I just have some comments about the movement of code from `lin_op.hpp`. I'm building this on a V100 node to run the tests once.

core/test/matrix/batch_dense.cpp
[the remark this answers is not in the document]
Slaedr: Well, if the user passed the dimensions of the two matrices as (2,2), even though they themselves are providing the values array, clearly they don't want to access the remaining entries in the original array through this Dense matrix. Essentially, the gko Dense/BatchDense matrix should be oblivious to and should not affect the entries that are outside the user-specified size. Testing this implies we promise the user they'll be able to access the entries beyond their specified size, which would (or should) never be their intent.

cuda/test/matrix/batch_dense_kernels.cpp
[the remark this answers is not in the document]
Slaedr: Thanks!

Slaedr:
The debug build works on a V100.
Screenshot_20210428_batchtest

Slaedr:
The debug build works on a V100.
Screenshot_20210428_batchtest

Slaedr:
The debug build works on a V100 node.
Screenshot_20210428_batchtest

include/ginkgo/core/base/lin_op_helpers.hpp
pratikvn: I think it still makes sense to separate these mixins into a different file. This is still a `lin_op_helpers.hpp` file. It is just that this is included in `batch_lin_op_helpers.hpp` as well. Only the ones that were really specific to `LinOp`, I keep in `LinOp`. Anything that can be possibly general is moved to the helpers file.

core/test/matrix/dense.cpp
pratikvn: For this case, isn't it that we are just making sure that the underlying data is not modified ? Both the ways, check for the same thing. For me, the second one is a bit more complex to understand, so I would prefer to keep it the simpler way.

pratikvn:
@Slaedr, great! thanks for checking !

PR #753, opened 2021-04-27
Improve value semantics for LinOp types

### LINKED ISSUE #747: Consistent empty state for `clear` and `move_to`
Related to #746, we currently don't have a clear concept of what an empty object is. This is relevant in two places: `PolymorphicObject::clear` resets the object by assigning a newly created instance `*this = Object{exec}`, which uses the move-assignment operator, and `move_to` implemented by `EnablePolymorphicAssignment` uses it directly `*this = std::move(*other)`.

This might leave us with inconsistent states: `Ell::move_to(Ell)` moves the value arrays, but leaves the size variables unmodified. `Csr::move_to(Csr)` leaves the matrix with empty row pointers, while `Csr::clear()` resets it to a valid state.

What we would need to do to fix this:
* Figure out what an empty state is for all LinOps (0x0, do nothing or throw exception on operations)
* Implement move-assignment operator and move-constructor to build this state.

My suggestions here would be:

### All LinOps
* Size set to 0x0

### Matrix formats
* Empty column indices and values
* Size values set to zero (nonzeros per row in Ell)
* Singleton row/slice pointers (Csr, Sellp)

### Factorizations
* Empty or nullptr underlying matrix (Ilu, Ic)

### Preconditioners
* Empty data
* Singleton block pointers (Jacobi)
* (Maybe) empty caches, but don't move them! (Ilu, Ic)

### Solvers
* Empty or nullptr system matrix
* Empty or nullptr preconditioner
* Empty or nullptr solve_struct (Trisolve)

### Composition etc.
* Empty list of operators
* (Maybe) empty caches, don't move them!

upsj:
This PR improves the value semantics of Ginkgo polymorphic objects by making sure that their contents are always stored on the correct executor. This fixes the issues caused by `gko::clone` not copying data that is stored in a `shared_ptr`. Additionally, I added a few helpers to reduce code duplication in solvers.

The PR is still work in progress, I only added the basic implementation and an example for Bicg, but I wanted to open the floor to discussions already

TODO:
- [x] moving Preconditionable to solver_base.hpp is interface-breaking, I guess we can do without the move?
- [x] Add tests
- [x] Bicgstab
- [x] Cg
- [x] Cgs
- [x] Fcg
- [x] Gmres
- [x] CbGmres
- [x] Idr
- [x] Ir
- [x] Lower/UpperTrs
- [x] Ilu/Ic factorizations
- [x] Ilu/Ic preconditioners
- [x] Isai preconditioner
- [x] Jacobi (size members)
- [x] Sellp (size members)
- [x] Dense (stride-preserving copy)
- [x] Hybrid (Coo, Ell, strategy)
- [x] Fbcsr (size members)
- [x] Ell (size members)
- [x] Csr (strategy)
- [x] Composition
- [x] Combination
- [x] Pertubation
- [x] Multigrid
- [ ] UseComposition is still broken, but there is no interface break-less approach to fix this. EnableMultigridLevel needs access to the object's executor, thus needs to become a mix-in.
- [ ] MultigridLevel cloning is currently very hacky, we probably need a better approach here.

Fixes #747


upsj:
It would also make sense for simplification to have a general utility class that provides correct value semantics to a `std::unique_ptr<LinOp>` or `std::shared_ptr<LinOp>` member.

include/ginkgo/core/solver/ir.hpp
```diff
@@ -145,33 +139,15 @@ class Ir : public EnableLinOp<Ir<ValueType>>, public Transposable {
...
-    void set_stop_criterion_factory(
-        std::shared_ptr<const stop::CriterionFactory> other)
-    {
-        stop_criterion_factory_ = std::move(other);
-    }
+    Ir& operator=(Ir&& other);
+
+    Ir(const Ir& other);
+
+    Ir(Ir&& other);
```
pratikvn, line 150: It might make sense to have a `IterativeSolverBase`/`EnableIterativeSolverBase` to move these copy and move assignment and constructors to those mixins.

pratikvn (commented):
Really nice work on this. Should streamline a lot of things internally. The interface and the mixins look good to me. I did not look into Multigrid yet.

include/ginkgo/core/solver/ir.hpp
upsj: Ir is kind of a special case, since it has both a system matrix and a solver, so we need assignment operators specialized for them. For all the other solvers, that is true :)

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -0,0 +1,371 @@
...
+        if (&other != this) {
+            set_preconditioner(other.get_preconditioner());
+        }
+        return *this;
+    }
+
+    EnablePreconditionable& operator=(EnablePreconditionable&& other)
+    {
+        if (&other != this) {
+            set_preconditioner(other.get_preconditioner());
```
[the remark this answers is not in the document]
upsj: I think that isn't necessary, since get_preconditioner() already returns a temporary.

core/matrix/ell.cpp
```diff
@@ -101,6 +101,50 @@ size_type calculate_max_nnz_per_row(
...
+template <typename ValueType, typename IndexType>
+Ell<ValueType, IndexType>& Ell<ValueType, IndexType>::operator=(Ell&& other)
+{
+    if (&other != this) {
+        EnableLinOp<Ell>::operator=(std::move(other));
+        values_ = std::move(other.values_);
+        col_idxs_ = std::move(other.col_idxs_);
+        num_stored_elements_per_row_ =
+            std::exchange(other.num_stored_elements_per_row_, 0);
+        stride_ = std::exchange(other.stride_, 0);
```
yhmtsai, line 128: Aren't num_stored_elements_per_row_ and stride_ zero from original move?

core/base/combination.cpp
```diff
@@ -60,6 +60,67 @@ inline void initialize_scalars(std::shared_ptr<const Executor> exec,
 }  // namespace
 
 
+template <typename ValueType>
+Combination<ValueType>& Combination<ValueType>::operator=(
+    const Combination& other)
```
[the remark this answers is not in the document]
upsj: yes, and another issue there is that the type implementing operator= needs to know the executor, because otherwise you are able to assign a CUDA system matrix to a OpenMP solver that was initialized as an empty object (create_default()).

core/matrix/ell.cpp
upsj: They are members directly inside the Ell type, thus they need to be handled inside its assignment operator. Primitive types don't reset on move assignment afaik.

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -241,6 +241,22 @@ class LinOp : public EnableAbstractPolymorphicObject<LinOp> {
      */
     virtual bool apply_uses_initial_guess() const { return false; }
 
+    LinOp &operator=(const LinOp &) = default;
+
+    LinOp &operator=(LinOp &&other)
+    {
```
pratikvn, line 247: I guess we still preserve the executor ?

include/ginkgo/core/solver/idr.hpp
```diff
@@ -262,33 +232,18 @@ class Idr : public EnableLinOp<Idr<ValueType>>,
...
-            stop::combine(std::move(parameters_.criteria));
-        subspace_dim_ = parameters_.subspace_dim;
-        kappa_ = parameters_.kappa;
-        deterministic_ = parameters_.deterministic;
-        complex_subspace_ = parameters_.complex_subspace;
-    }
+          subspace_dim_{parameters_.subspace_dim},
+          kappa_{parameters_.kappa},
+          deterministic_{parameters_.deterministic},
+          complex_subspace_{parameters_.complex_subspace}
```
pratikvn, line 241: I think we talked discussed this before. Given that the parameters struct is public, we can just use that and not have duplicate member variables.

include/ginkgo/core/solver/ir.hpp
```diff
@@ -212,6 +188,9 @@ class Ir : public EnableLinOp<Ir<ValueType>>, public Transposable {
     void apply_impl(const LinOp* alpha, const LinOp* b, const LinOp* beta,
                     LinOp* x) const override;
 
+    void set_relaxation_factor(
+        std::shared_ptr<const matrix::Dense<ValueType>> new_factor);
```
pratikvn, line 192: Do we need this to take a `Dense` ? It might be better to take a scalar value ?

include/ginkgo/core/solver/lower_trs.hpp
```diff
@@ -142,26 +137,16 @@ class LowerTrs : public EnableLinOp<LowerTrs<ValueType, IndexType>>,
                       std::shared_ptr<const LinOp> system_matrix)
         : EnableLinOp<LowerTrs>(factory->get_executor(),
                                 gko::transpose(system_matrix->get_size())),
-          parameters_{factory->get_parameters()},
-          system_matrix_{}
+          EnableSolverBase<LowerTrs<ValueType, IndexType>,
+                           matrix::Csr<ValueType, IndexType>>{
```
pratikvn, line 141:
```suggestion
                          CsrMatrix>{
```

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -0,0 +1,371 @@
...
+};
+
+
+/**
+ * A LinOp implementing this interface stores a system matrix.
+ *
+ * @ingroup solver
+ * @ingroup LinOp
+ */
+template <typename DerivedType, typename MatrixType = LinOp>
```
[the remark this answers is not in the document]
pratikvn: I think all the current solvers still use default LinOp, so it should be okay.

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -0,0 +1,371 @@
...
+    {
+        system_matrix_ = std::move(system_matrix);
+    }
+
+    std::shared_ptr<const MatrixType> system_matrix_;
+};
+
+
+/**
+ * A LinOp implementing this interface stores a system matrix.
```
pratikvn, line 166: Missing doc for template types

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -0,0 +1,371 @@
...
+        std::shared_ptr<const stop::CriterionFactory> new_stop_factory)
+    {
+        stop_factory_ = new_stop_factory;
+    }
+
+private:
+    std::shared_ptr<const stop::CriterionFactory> stop_factory_;
+};
+
+
```
pratikvn, line 260: Missing doc

include/ginkgo/core/solver/upper_trs.hpp
```diff
@@ -142,26 +136,16 @@ class UpperTrs : public EnableLinOp<UpperTrs<ValueType, IndexType>>,
                       std::shared_ptr<const LinOp> system_matrix)
         : EnableLinOp<UpperTrs>(factory->get_executor(),
                                 gko::transpose(system_matrix->get_size())),
-          parameters_{factory->get_parameters()},
-          system_matrix_{}
+          EnableSolverBase<UpperTrs<ValueType, IndexType>,
+                           matrix::Csr<ValueType, IndexType>>{
```
pratikvn, line 140:
```suggestion
                           CsrMatrix>{
```

test/solver/solver.cpp
```diff
@@ -0,0 +1,979 @@
...
+        ASSERT_EQ(mtx->get_stopping_criterion_factory(), nullptr);
+    }
+};
+
+
+struct Cg : SimpleSolverTest<gko::solver::Cg<solver_value_type>> {};
+
+
+struct Cgs : SimpleSolverTest<gko::solver::Cgs<solver_value_type>> {
+    static double tolerance() { return 1e5 * r<value_type>::value; }
```
pratikvn, line 139: Isn't this tolerance a bit too high for `float` ? It might make sense to have a `value_type` specific tolerance ? Something already in the `r<>` struct to get something like `r<value_type>::tolerance` ?

test/solver/solver.cpp
```diff
@@ -0,0 +1,979 @@
...
+                Config::build(ref, 0).on(ref)->generate(mtx.ref),
+                Config::build(exec, 0).on(exec)->generate(mtx.dev)};
+            pair.ref->clear();
+            pair.dev->clear();
+            guarded_fn(std::move(pair));
+        }
+        {
+            SCOPED_TRACE("Unpreconditioned solver with 0 iterations via clone");
+            guarded_fn(test_pair<SolverType>{
+                Config::build(ref, 0).on(ref)->generate(mtx.ref), exec});
```
pratikvn, line 601: Missing device element ?

pratikvn (commented):
Mostly looks good. Some minor comments. Some questions:
1. Should we add documentation for all the move/copy assignment/constructors ? 
2. Should the move/copy tests be in reference or core ? 

Nit; Maybe rename the PR name to general value semantics, as a significant part of the code also fixes Matrix formats as well.

test/solver/solver.cpp
upsj: that is being generated implicitly by the test_pair constructor calling ref->clone(exec);

include/ginkgo/core/base/lin_op.hpp
upsj: yes, that's an invariant of PolymorphicObject. I should probably call its move-assignment operator here. Great catch!
```suggestion
    {
        EnableAbstractPolymorphicObject<LinOp>::operator=(std::move(other));
```

core/test/matrix/dense.cpp
```diff
@@ -346,6 +346,15 @@ TYPED_TEST(Dense, CanCreateSubmatrix)
 }
 
 
+TYPED_TEST(Dense, CanCreateEmptySubmatrix)
+{
+    using value_type = typename TestFixture::value_type;
+    auto submtx = this->mtx->create_submatrix(gko::span{0, 0}, gko::span{1, 1});
+
+    EXPECT_FALSE(submtx->get_size());
```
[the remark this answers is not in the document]
upsj: I should note that the tests are mostly part of #973, but it seems like you commented specifically on those related to value semantics, thanks!

core/solver/cb_gmres.cpp
```diff
@@ -221,16 +223,16 @@ void CbGmres<ValueType>::apply_dense_impl(
         auto one_op = initialize<Vector>({one<ValueType>()}, exec);
         auto neg_one_op = initialize<Vector>({-one<ValueType>()}, exec);
 
+        const auto num_rows = this->get_size()[1];
```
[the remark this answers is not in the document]
upsj: good catch! that probably caused some test Issues I didn't look at yet

core/base/combination.cpp
```diff
@@ -60,6 +60,67 @@ inline void initialize_scalars(std::shared_ptr<const Executor> exec,
...
+        operators_ = other.operators_;
+        // if the operators are on the wrong executor, copy them over
+        if (other.get_executor() != exec) {
+            for (auto &coef : coefficients_) {
+                coef = gko::clone(exec, coef);
+            }
+            for (auto &op : operators_) {
+                op = gko::clone(exec, op);
+            }
+        }
```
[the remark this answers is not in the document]
upsj: the assignment coefficients_ = other.coefficients_ is only a shallow copy, i.e. it copies the pointer, nothing else.
Only this block actually creates a new object on the new executor.

core/solver/cb_gmres.cpp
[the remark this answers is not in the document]
upsj: ah nevermind, the matrix is symmetric. But it's still reasonable to address this.

test/solver/solver.cpp
upsj: Yes, this is kind of the issue in #973, there are some significant differences between Reference and other kernels which I didn't have time to investigate yet

core/matrix/dense.cpp
```diff
@@ -317,49 +317,74 @@ void Dense<ValueType>::compute_norm1_impl(LinOp* result) const
...
+            stride_ = other.get_stride();
+            values_.resize_and_reset(this->get_size()[0] * stride_);
+        }
         // we need to create a executor-local clone of the target data, that
         // will be copied back later.
-        auto exec = this->get_executor();
-        auto result_array = make_temporary_output_clone(exec, &result->values_);
+        auto exec = other.get_executor();
+        auto exec_values_array =
+            make_temporary_output_clone(exec, &this->values_);
```
upsj, line 334:
```suggestion
            make_temporary_clone(exec, &this->values_);
```

core/base/combination.cpp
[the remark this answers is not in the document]
pratikvn: Ah, right, I missed that.

yhmtsai (approved):
LGTM in general.
from this pr, the convert_to can also be done by `*mtx = *mtx2` if they are the same class, right?
some public function in solver_base.hpp miss document.
and also worried about the executor

core/test/solver/cb_gmres.cpp
```diff
@@ -334,7 +334,7 @@ TYPED_TEST(CbGmres, ThrowsOnWrongPreconditionerInFactory)
     using Mtx = typename TestFixture::Mtx;
     using Solver = typename TestFixture::Solver;
     std::shared_ptr<Mtx> wrong_sized_mtx =
-        Mtx::create(this->exec, gko::dim<2>{1, 3});
+        Mtx::create(this->exec, gko::dim<2>{2, 2});
```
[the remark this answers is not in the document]
upsj: It needs to be square, otherwise already setting the system matrix fails, which produces the wrong kind of exception.

test/matrix/matrix.cpp
```diff
@@ -99,12 +112,43 @@ struct DenseWithCustomStride : DenseWithDefaultStride {
...
+    static void assert_empty_state(const matrix_type* mtx)
+    {
+        ASSERT_FALSE(mtx->get_size());
+        ASSERT_EQ(mtx->get_num_stored_elements(), 0);
+        ASSERT_NE(mtx->get_const_row_ptrs(), nullptr);
+        ASSERT_EQ(
+            mtx->get_executor()->copy_val_to_host(mtx->get_const_row_ptrs()),
+            0);
+        ASSERT_EQ(mtx->get_const_col_idxs(), nullptr);
+        ASSERT_EQ(mtx->get_const_values(), nullptr);
```
MarcelKoch, line 147: perhaps you could extract that into a separate function which will be called by all the different CSR variants

test/matrix/matrix.cpp
```diff
@@ -277,6 +459,22 @@ struct HybridColumnLimitStrategy
...
+        ASSERT_EQ(mtx->get_coo_num_stored_elements(), 0);
+        ASSERT_EQ(mtx->get_const_coo_row_idxs(), nullptr);
+        ASSERT_EQ(mtx->get_const_coo_col_idxs(), nullptr);
+        ASSERT_EQ(mtx->get_const_coo_values(), nullptr);
+        ASSERT_FALSE(mtx->get_ell()->get_size());
+        ASSERT_EQ(mtx->get_ell_num_stored_elements_per_row(), 0);
+        ASSERT_EQ(mtx->get_ell_num_stored_elements(), 0);
+        ASSERT_EQ(mtx->get_ell_stride(), 0);
+        ASSERT_EQ(mtx->get_const_ell_col_idxs(), nullptr);
+        ASSERT_EQ(mtx->get_const_ell_values(), nullptr);
```
MarcelKoch, line 476: this could again be extracted from all the hyrbid variants.

MarcelKoch: Also, should the strategy be checked?

test/matrix/matrix.cpp
MarcelKoch: should `srow` also be checked here?

MarcelKoch (approved):
lgtm. I only have some smaller remarks.

In retrospective, I think it would have been easier to review this if it were split into a matrix value semantics and a solver base PR.

pratikvn (approved):
LGTM! You want to merge this into `solver-test` and not into develop ? Is that mainly because of the convergence test issue ?

upsj:
No, this is a technique @MarcelKoch introduced me to: I originally wanted to merge `solver-test` first, so this PR would only show the differences to `solver-test`, not all of the changes `solver-test` itself introduces. If the first PR is merged, the base ref is automatically updated to `develop`

core/test/utils/matrix_utils.hpp
```diff
@@ -47,108 +47,221 @@ namespace test {
 
 
 /**
```
thoasm, line 49: If you decide to move this out of test, maybe also change the namespace `test` to `utils`?
Also, it seems like you no longer need to include `core/test/utils/value_generator.hpp`.

thoasm (approved):
LGTM!
I really like the new changes, this should make moving defined behavior.

upsj:
format-rebase!

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -0,0 +1,371 @@
...
+ * criterion factory.
+ *
+ * @ingroup solver
+ * @ingroup LinOp
+ */
+template <typename ValueType, typename DerivedType>
+class EnablePreconditionedIterativeSolver
+    : public EnableSolverBase<DerivedType>,
+      public EnableIterativeBase<DerivedType>,
+      public EnablePreconditionable<DerivedType> {
```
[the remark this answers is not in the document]
upsj: I disagree, since if you were to do this not via CRTP, but purely using interfaces, you would have PreconditionedIterativeSolver deriving from IterativeSolver deriving from Solver, so solver would be the first base class.

include/ginkgo/core/solver/ir.hpp
upsj: Let's address this together with #820

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -241,6 +241,26 @@ class LinOp : public EnableAbstractPolymorphicObject<LinOp> {
...
+
+    LinOp(const LinOp&) = default;
+
+    LinOp(LinOp&& other)
+        : EnableAbstractPolymorphicObject<LinOp>{std::move(
+              static_cast<EnableAbstractPolymorphicObject<LinOp>&&>(other))},
+          size_{other.size_}
+    {
+        other.set_size({});
+    }
```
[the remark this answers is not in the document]
upsj: I still want to derive the same behavior w.r.t. loggers that PolymorphicObject provides

test/matrix/matrix.cpp
```diff
@@ -214,6 +338,11 @@ struct FbcsrBlocksize1
     {
         ASSERT_EQ(mtx->get_block_size(), 1);
     }
+
+    static void assert_empty_state(const matrix_type* mtx)
+    {
+        ASSERT_FALSE(mtx->get_size());
```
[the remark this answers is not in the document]
upsj: I don't think I follow?

test/matrix/matrix.cpp
upsj: I don't think it needs to be checked, it's more of an internal property.

test/matrix/matrix.cpp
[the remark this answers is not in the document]
upsj: Ah right, I think I wanted to check the data pointers as well, thanks!

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -241,6 +241,26 @@ class LinOp : public EnableAbstractPolymorphicObject<LinOp> {
      */
     virtual bool apply_uses_initial_guess() const { return false; }
 
+    LinOp& operator=(const LinOp&) = default;
```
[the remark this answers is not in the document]
upsj: I tend to agree about the need for documentation, though I'm not sure this is the right place for it - the actual executor handling happens in PolymorphicObject, and in practice it matters for concrete LinOps. So I'm not sure I would expect people to check for documentation "in the middle"

include/ginkgo/core/base/lin_op.hpp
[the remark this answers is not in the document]
thoasm: In that case, don't you want to `std::move` `other.size_` as well: `size_{std::move(other.size_}}`?

core/solver/ir.cpp
```diff
@@ -53,13 +53,86 @@ GKO_REGISTER_OPERATION(initialize, ir::initialize);
...
+
+template <typename ValueType>
+void Ir<ValueType>::set_relaxation_factor(
+    std::shared_ptr<const matrix::Dense<ValueType>> new_factor)
+{
+    auto exec = this->get_executor();
+    if (new_factor && new_factor->get_executor() != exec) {
+        new_factor = gko::clone(exec, new_factor);
+    }
+    relaxation_factor_ = new_factor;
```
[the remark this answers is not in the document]
upsj: I think if it's empty, we at least shouldn't leave relaxation_factor unchanged. This seems like the least surprising behavior to me.

PR #758, opened 2021-05-05
Allow using device memory allocations in debug mode.

pratikvn:
Currently, we always use Unified memory for debug builds. While this helps for debugging in some cases, it can be a bit annoying because release and debug builds have different behaviours. 

This PR allows the user to specify whether or not to use the allocation using the unified memory. The default behaviour is the same as current behaviour currently controlled by pre-processor defines.

### DIFF
--- a/cuda/test/base/cuda_executor.cu
+++ b/cuda/test/base/cuda_executor.cu
@@ -98,7 +98,8 @@ protected:
         cuda = gko::CudaExecutor::create(0, omp);
         cuda2 = gko::CudaExecutor::create(
             gko::CudaExecutor::get_num_devices() - 1, omp);
-        cuda3 = gko::CudaExecutor::create(0, omp, false, true);
+        cuda3 = gko::CudaExecutor::create(0, omp, false,
+                                          gko::allocation_mode::unified_host);
     }
 
     void TearDown()
--- a/cuda/test/base/cuda_executor.cu
+++ b/cuda/test/base/cuda_executor.cu
@@ -86,7 +86,10 @@ public:
 class CudaExecutor : public ::testing::Test {
 protected:
     CudaExecutor()
-        : omp(gko::OmpExecutor::create()), cuda(nullptr), cuda2(nullptr)
+        : omp(gko::OmpExecutor::create()),
+          cuda(nullptr),
+          cuda2(nullptr),
+          cuda3(nullptr)
     {}
 
     void SetUp()
@@ -95,6 +98,7 @@ protected:
         cuda = gko::CudaExecutor::create(0, omp);
         cuda2 = gko::CudaExecutor::create(
             gko::CudaExecutor::get_num_devices() - 1, omp);
+        cuda3 = gko::CudaExecutor::create(0, omp, false, true);
     }
 
     void TearDown()
@@ -108,6 +112,7 @@ protected:
     std::shared_ptr<gko::Executor> omp;
     std::shared_ptr<gko::CudaExecutor> cuda;
     std::shared_ptr<gko::CudaExecutor> cuda2;
+    std::shared_ptr<gko::CudaExecutor> cuda3;
 };
 
 
@@ -163,6 +168,7 @@ __global__ void check_data(int *data)
     }
 }
 
+
 TEST_F(CudaExecutor, CopiesDataToCuda)
 {
     int orig[] = {3, 8};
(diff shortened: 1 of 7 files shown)


upsj:
I really like the idea and the implementation so far, one thing I would like to avoid though is a function with multiple `bool` parameters. I think this isn't necessarily a binary choice, since there are different allocation modes for unified memory (cudaMemAttachGlobal or cudaMemAttachHost)
```cpp
enum class allocation_mode { device, unified_host, unified_device };
```
Also, macros should be namespaced `GKO_...`.

include/ginkgo/core/base/executor.hpp
```diff
@@ -50,6 +50,52 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+
+}  // namespace gko
+
+
+#ifdef NDEBUG
+
+// When in release, always prefer device allocations
+#define GKO_DEFAULT_CUDA_ALLOC_MODE allocation_mode::device
+
+#define GKO_DEFAULT_HIP_ALLOC_MODE allocation_mode::device
```
upsj, line 79: These could also be constexpr global variables

upsj (approved):
LGTM!

cuda/test/base/cuda_executor.cu
```diff
@@ -176,6 +183,29 @@ TEST_F(CudaExecutor, CopiesDataToCuda)
 }
 
 
+__global__ void check_data2(int *data)
+{
+    if (data[0] != 4 || data[1] != 8) {
+        asm("trap;");
```
yhmtsai, line 189: Is it to make sure the compiler does not ignore this kernel?

cuda/test/base/cuda_executor.cu
```diff
@@ -95,6 +98,8 @@ protected:
         cuda = gko::CudaExecutor::create(0, omp);
         cuda2 = gko::CudaExecutor::create(
             gko::CudaExecutor::get_num_devices() - 1, omp);
+        cuda3 = gko::CudaExecutor::create(0, omp, false,
+                                          gko::allocation_mode::unified_host);
```
yhmtsai, line 102: should it also test `unified_global` ?
`unified_host` require device with `cudaDevAttrConcurrentManagedAccess`, or, the device can not access the `unified_host` memory.

include/ginkgo/core/base/executor.hpp
```diff
@@ -50,6 +50,55 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+ * host through the Unified memory model.
+ *
+ * `unified_host` allocates memory on the
+ * host and it is not available on devices which do not have concurrent acesses
+ * switched on, but this access can be explictly switched on, when necessary.
+ */
+enum class allocation_mode { device, unified_global, unified_host };
+
+
+#ifdef NDEBUG
```
yhmtsai, line 71: Do you think we should use another cmake variable to control it? or user change this file to change the behavior

yhmtsai (changes requested):
LGTM in general, but I would like to make sure the cudaMemAttachGlobal things.

cuda/test/base/cuda_executor.cu
upsj: This is meant to fail if the data contains the incorrect value

include/ginkgo/core/base/executor.hpp
pratikvn: No, the user controls it by creating a executor capable of allocating on unified memory. I can add a setter, but I think that is confusing and error-prone. We can anyway create multiple executors on one CUDA device

cuda/test/base/cuda_executor.cu
pratikvn: Ah, yes, sorry. It was meant to be `unified_global` .

include/ginkgo/core/base/executor.hpp
yhmtsai: I see

hip/base/executor.hip.cpp
```diff
@@ -124,11 +125,16 @@ void *HipExecutor::raw_alloc(size_type num_bytes) const
...
-#else
-    auto error_code = hipMallocManaged(&dev_ptr, num_bytes);
-#endif
+    int error_code = 0;
+    if (this->alloc_mode_ == allocation_mode::unified_host) {
+        error_code = hipMallocManaged(&dev_ptr, num_bytes, hipMemAttachHost);
+    } else if (this->alloc_mode_ == allocation_mode::unified_global) {
+        error_code = hipMallocManaged(&dev_ptr, num_bytes, hipMemAttachGlobal);
+    } else if (this->alloc_mode_ == allocation_mode::device) {
+        error_code = hipMalloc(&dev_ptr, num_bytes);
```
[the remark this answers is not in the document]
yhmtsai: does it use the wrong commit? it seemed to contain the @upsj comments yesterday, but does not now

include/ginkgo/core/base/executor.hpp
```diff
@@ -1445,9 +1496,10 @@ class HipExecutor : public detail::ExecutorBase<HipExecutor>,
      * @param master  an executor on the host that is used to invoke the device
      *                kernels
      */
-    static std::shared_ptr<HipExecutor> create(int device_id,
-                                               std::shared_ptr<Executor> master,
-                                               bool device_reset = false);
+    static std::shared_ptr<HipExecutor> create(
```
yhmtsai, line 1499: Could you also add the documentation for device_reset and alloc_mode of this function? also Cuda

yhmtsai (approved):
LGTM, is it based on the old commit? some changes seem to be lost.

pratikvn:
Yes, sorry, looks like my work laptop was on a older commit and I feel into the peril of the force push.

pratikvn:
I guess the rebase does not run the complete CI. But as this was a small PR, should I go ahead and merge this ?

upsj:
For some reason, the rebase doesn't use the @ginkgo-bot token, I guess I'll need to look at that some time. But I would be fine with merging it directly, as it already worked before, and the only change from Thomas' PR was unrelated

PR #761, opened 2021-05-06
Make array constructor explicit

upsj:
This PR prevents Executors from being accidentally converted to empty Arrays. It is technically interface breaking, but like the explicit bool conversion previously, I think it only prevents accidental misuse


pratikvn (approved):
LGTM!

thoasm (approved):
LGTM!

yhmtsai (approved):

PR #762, opened 2021-05-06
Code structure for batch solvers, preconditioners, scaling, stopping criteria and logging

Slaedr:
This long PR lays out the proposed structure for handling solvers, preconditioners, stopping criteria, loggers and scaling for Ginkgo batch solvers. Reference and Cuda implementations (and core interfaces where applicable) of examples of each category are included to show the structure - batch Richardson for the solver, batch (scalar) Jacobi for the preconditioner and so on.

Code design
-----------------
The structure looks quite different from other (non-batch) Ginkgo components, mainly for performance reasons. Most of the design is based around having only one kernel launch to solve all the "batch entries" (individual systems in the batch) and attempting to inline as many calls to components (preconditioners, loggers etc.) as possible into the solver kernel. Therefore, the solver kernel function is templated on the preconditioner type, matrix type, stopping criterion type and logger type. For now, only the solver and parts of the logger have any user-facing interface in the core module. The rest of the components are defined individually for each backend with a "developer's pledge" to keep the interfaces same for all implementations for all backends. Suggestions are solicited on how to improve this. The choice among the provided options is exercised by the user with simple `enum` parameters.

~The preconditioner setup is intended to happen in two phases. An initial "pattern setup" phase involves a separate kernel launch for a "small" operation for common setup on the sparsity pattern.~ The actual numerical generation of the preconditioner happens inside the solver kernels before the actual solve starts.

For the purpose of diagonal scaling of matrices, the solver explicitly scales the matrices before the iteration loop starts. However, since the input matrices are stored using a pointer to `const BatchLinOp`, this requires the use of a `const_cast` by the solver before kernel launch. Thus if the user requests scaling, the matrix is indeed modified. Suggestions on a better approach are welcome.

Limitations
--------------
- In general, the user cannot provide their own components such as preconditioners, loggers etc., as they must be known when the solver is compiled.
- The number of right-hand-sides for solvers is limited to 32 in order to use `uint32_t`s for keeping track of convergence states, as well as for more efficiency using warp collectives.
- As mentioned below, shared memory is used for intermediate vectors on GPUs, which limits the problem size, the number of such vectors and/or the occupancy. The default build now uses dynamic shared memory, so that this issue is now pushed to runtime. In general, dynamic memory should lead to better occupancy.

Performance considerations
-------------------------------------
Currently, one thread-block handles one batch entry (one small system in the batch). Immutable data, such as the RHS vectors and matrix nonzero values are kept in global memory with the expectation that they will be cached favourably. Changing data, such as intermediate vectors used by the solver and the solution vector, are stored in shared memory ~using compile-time sizes~.

Some kernels are currently inefficient. Most such places are marked with 'todo's. These are planned to be addressed in future PRs.

[The following is no longer relevant because of the move to dynamic shared memory]
At least for an odd number of compile-time max RHS, using `UninitializedArray` leads to misaligned warp memory access. However, simply using `__shared__` variables directly seems to fix the issue at the cost of copious warnings. Perhaps it is the case that if the compiler does not know the type of the data, it does not care about properly aligning it. Since `UninitializedArray` uses `char` internally, perhaps no alignment requirement is conveyed to the compiler.

TODO:
- [x] Consider switching to dynamic shared memory.


Slaedr:
Some changes are needed to how the preconditioners are implemented.
[ ] We need a way for the user to give the array of block pointers/sizes
[ ] We need the initial pattern generation function to be more general than just one global kernel call.

core/test/utils/batch.hpp
```diff
@@ -92,6 +117,68 @@ std::unique_ptr<MatrixType> generate_uniform_batch_random_matrix(
...
+        for (int i = 1; i < nrows - 1; i++) {
+            const int rstart = row_ptrs.get_const_data()[i];
+            va[rstart] = -1.0;
+            va[rstart + 1] = 2.0;
+            va[rstart + 2] = -1.0;
+        }
+        const auto lrstart = row_ptrs.get_const_data()[nrows - 1];
+        va[lrstart] = -1.0;
+        va[lrstart + 1] = 2.0;
+    }
```
pratikvn, line 173: I think you can simplify this with a `matrix_assembly_data`, where you can create triplets `(row_idx, col_idx, value)` depending on the index. https://github.com/ginkgo-project/ginkgo/blob/develop/include/ginkgo/core/base/matrix_assembly_data.hpp#L119

cuda/matrix/batch_csr_kernels.cu
```diff
@@ -206,6 +204,28 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE_AND_INT32_INDEX(
...
+                 matrix::BatchCsr<ValueType, IndexType> *const mat)
+{
+    if (!left_scale->get_size().stores_equal_sizes()) GKO_NOT_IMPLEMENTED;
+    if (!right_scale->get_size().stores_equal_sizes()) GKO_NOT_IMPLEMENTED;
+
+    const auto m_ub = get_batch_struct(mat);
+    const auto left_ub = get_batch_struct(left_scale);
+    const auto right_ub = get_batch_struct(right_scale);
+
+    constexpr int block_size = 512;
```
pratikvn, line 220: It might make sense to make this tunable.

pratikvn (commented):
Really great work on this! Some issues here and there, but nothing major. I still need to look into the main headers, core and the reference,omp,hip

include/ginkgo/core/base/batch_lin_op.hpp
```diff
@@ -515,8 +515,152 @@ class BatchTransposable {
...
+                          const BatchLinOp *const right_scale_op) override
+    {
+        // auto sysmat =
+        // std::static_pointer_cast<ConcreteSolver*>()->system_matrix_; auto
+        // orig_const_mat = 	std::dynamic_pointer_cast<const
+        // BatchScalable>(sysmat); if (!orig_const_mat) {
+        // 	GKO_NOT_SUPPORTED(sysmat);
+        // }
+        // auto mat = std::const_pointer_cast<BatchScalable>(orig_const_mat);
+        // mat->batch_scale(left_op, right_op);
```
pratikvn, line 649: Can be removed ?

include/ginkgo/core/matrix/batch_dense.hpp
```diff
@@ -1010,6 +1010,101 @@ std::unique_ptr<Matrix> batch_initialize(
...
+ * @param exec  Executor associated to the vector
+ * @param create_args  additional arguments passed to Matrix::create, not
+ *                     including the Executor, which is passed as the first
+ *                     argument
+ *
+ * @ingroup LinOp
+ * @ingroup mat_formats
+ */
+template <typename Matrix, typename... TArgs>
+std::unique_ptr<Matrix> batch_initialize(
```
pratikvn, line 1037: I am not sure, but maybe you need to combine this with the previous initialize where we have the duplicate the vector ? Because otherwise there might be ambiguity here.

reference/matrix/batch_csr_kernels.hpp
```diff
@@ -43,12 +43,6 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 namespace gko {
 namespace kernels {
 namespace reference {
-/**
- * @brief The Compressed sparse row matrix format namespace.
- * @ref BatchCsr
- * @ingroup batch_csr
- */
-namespace batch_csr {
```
pratikvn, line 51: Don't we need these kernels to be in this namespace ?

pratikvn (commented):
Some final minor comments. But otherwise, it looks good.

common/solver/batch_richardson_kernels.hpp.inc
```diff
@@ -0,0 +1,178 @@
...
+                                                                    1, 1};
+
+        __shared__ real_type norms[max_nrhs];
+        for (int iz = threadIdx.x; iz < max_nrhs; iz += blockDim.x) {
+            norms[iz] = 0.0;
+        }
+        const gko::batch_dense::BatchEntry<real_type> norms_b{
+            norms, static_cast<size_type>(max_nrhs), 1, num_rhs};
+
+        prec.generate(a_b, prec_work);
```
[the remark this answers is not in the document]
Slaedr: That might be useful in some situations, but for now that's not exactly the problem. It's more in the constructor that's called on the host. For the constructor, every batch preconditioner can have its own signature anyway. The problem with the curent batch block Jacobi (not in this PR) is that we need extra input from the user specifically for the preconditioner.

common/solver/batch_richardson_kernels.hpp.inc
```diff
@@ -0,0 +1,178 @@
...
+        // initial residual
+        for (int iz = threadIdx.x; iz < num_rows * num_rhs; iz += blockDim.x) {
+            const int i = iz / num_rhs;
+            const int j = iz % num_rhs;
+            r_b.values[i * r_b.stride + j] = b_b.values[i * b_b.stride + j];
+        }
+        compute_norm2<ValueType>(gko::batch::to_const(r_b), norms_b);
+        __shared__ real_type init_rel_res_norm[max_nrhs];
+        for (int iz = threadIdx.x; iz < num_rows * num_rhs; iz += blockDim.x) {
+            init_rel_res_norm[iz] = sqrt(norms_b.values[iz]);
```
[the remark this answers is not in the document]
Slaedr: Great catch! This fixed a lot of things.

common/solver/batch_richardson_kernels.hpp.inc
```diff
@@ -0,0 +1,178 @@
...
+                }
+            }
+            __syncthreads();
+
+            add_scaled(relax_b, gko::batch::to_const(dx_b), x_b);
+            __syncthreads();
+            iter++;
+        }
+
+        if (left.values) {
```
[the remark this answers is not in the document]
Slaedr: No, I want to use left here again. On the host side I'm taking care that they're both always equal.

common/stop/batch_criteria.hpp.inc
```diff
@@ -0,0 +1,147 @@
...
+// to include "cuda|hip/matrix/batch_dense_kernels.hpp"
+
+namespace stop {
+
+
+/**
+ * A stopping criterion for batch solvers that comprises a
+ * maximum iteration count as well as relative residual tolerance.
+ *
+ * At most 32 right-hand-side vectors are supported.
```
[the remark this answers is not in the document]
Slaedr: Nice catch. This made me realize that this will not work on AMD. We still might need to do something about the intrinsics, but at least we're closer now.

common/stop/batch_criteria.hpp.inc
```diff
@@ -0,0 +1,147 @@
...
+        bitset_type &converged) const
+    {
+        if (iter >= max_its - 1) {
+            return true;
+        }
+
+        if (residual_norms) {
+            check_norms(residual_norms, converged);
+        } else {
+            real_type norms[32];
```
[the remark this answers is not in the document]
Slaedr: Good catch. The issue if I've not tested for this case yet. I should add a test for this.

core/solver/batch_richardson.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+    log::BatchLogData<ValueType> logdata;
+    // allocate logging arrays assuming uniform size batch
+    const size_type num_rhs = dense_b->get_size().at(0)[1];
+    const size_type num_batches = dense_b->get_num_batch_entries();
+    batch_dim<> sizes(num_batches, dim<2>{1, num_rhs});
+    logdata.res_norms =
+        matrix::BatchDense<real_type>::create(this->get_executor(), sizes);
+    logdata.iter_counts.set_executor(this->get_executor());
+    logdata.iter_counts.resize_and_reset(num_rhs * num_batches);
```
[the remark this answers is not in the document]
Slaedr: The allocation time should be negligible compared to the entire solve. Also, I think it would be best to wipe the logger after one solve before starting the next anyway.

core/solver/batch_richardson_kernels.hpp
```diff
@@ -0,0 +1,132 @@
...
+/**
+ * Options controlling the batch Richardson solver.
+ */
+template <typename RealType>
+struct BatchRichardsonOptions {
+    std::string preconditioner;
+    int max_its;
+    RealType rel_residual_tol;
+    RealType relax_factor;
+};
```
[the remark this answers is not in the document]
Slaedr: It could, but it's used in the kernel function, so I'd prefer it's in the same namespace as the function itself. It's in an internal header anyway, so it should not be an issue.

reference/matrix/batch_csr_kernels.hpp
Slaedr: I'm moving towards using overloading as much as possible instead of namespaces for these small "device-side" kernels. In that regard, reference is inconsistent with common/cuda. I should probably fix that now.

common/solver/batch_richardson_kernels.hpp.inc
```diff
@@ -0,0 +1,178 @@
...
+    const gko::batch_dense::UniformBatch<ValueType> x)
+{
+    using real_type = typename gko::remove_complex<ValueType>;
+    const auto nbatch = mat.num_batch;
+    const auto num_rows = mat.num_rows;
+    const auto num_rhs = b.num_rhs;
+    constexpr int max_nrhs = batch_config<ValueType>::max_num_rhs;
+    constexpr int max_nrows = batch_config<ValueType>::max_num_rows;
+
+    if (batch_config<ValueType>::max_num_rows < num_rows * num_rhs) {
```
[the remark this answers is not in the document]
Slaedr: Ah yes, I meant to fix this, but looks like I missed it.

pratikvn (commented):
Some more minor things.

reference/matrix/batch_dense_kernels.hpp
```diff
@@ -57,9 +61,9 @@ using BatchEntry = gko::batch_dense::BatchEntry<ValueType>;
 
 
 template <typename ValueType>
-void simple_apply(const BatchEntry<const ValueType> &a,
-                  const BatchEntry<const ValueType> &b,
-                  const BatchEntry<ValueType> &c)
+inline void simple_apply(const BatchEntry<const ValueType> &a,
```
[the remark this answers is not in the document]
Slaedr: It's in a header file and is non-local to the file. So if it's not inline and included in more than one file, it will create duplicate definitions and the linker will complain.

reference/preconditioner/batch_jacobi.hpp
```diff
@@ -0,0 +1,113 @@
...
+        for (int i = 0; i < r.num_rows; i++) {
+            for (int j = 0; j < r.num_rhs; j++) {
+                z.values[i * z.stride + j] =
+                    work_[i] * r.values[i * r.stride + j];
+            }
+        }
+    }
+
+private:
+    ValueType *work_;
```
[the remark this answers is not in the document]
Slaedr: It's more flexible this way. It can be static shared memory, dynamic shared memory or even global memory, depending on what's passed into the `generate` function.

Slaedr:
format!

common/solver/batch_richardson_kernels.hpp.inc
```diff
@@ -0,0 +1,174 @@
...
+        //__shared__ UninitializedArray<ValueType, max_nrows> delta_x_un;
+        //__shared__ UninitializedArray<ValueType, PrecType::work_size>
+        //    prec_work_un;
+        // ValueType *const residual = residual_un;
+        // ValueType *const delta_x = delta_x_un;
+        // ValueType *const prec_work = prec_work_un;
+
+        __shared__ ValueType residual[max_nrows * max_nrhs];
+        __shared__ ValueType delta_x[max_nrows * max_nrhs];
+        __shared__ ValueType prec_work[PrecType::work_size];
```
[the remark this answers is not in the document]
Slaedr: I'm getting more and more convinced that we should move to dynamic shared memory at the cost of runtime failures, as long as we can give a good error message. There are a few intricacies to maximize performance, which I'm looking into.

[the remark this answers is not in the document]
pratikvn: I agree. The flexibility of dynamic shared memory trumps the delay in error reporting.

reference/matrix/batch_dense_kernels.hpp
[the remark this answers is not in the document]
pratikvn: Ah, right, I forgot about that.

common/solver/batch_richardson_kernels.hpp.inc
```diff
@@ -0,0 +1,178 @@
...
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+template <typename StopType, typename PrecType, typename LogType,
+          typename MatrixType, typename ValueType>
```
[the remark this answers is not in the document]
Slaedr: For now, I added some static asserts. Also, the last parameter is deduced from other stuff, so I think it does not compile if the value types are inconsistent anyway.

core/solver/batch_richardson_kernels.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+
+template <typename ValueType>
+inline int local_memory_requirement(const int num_rows, const int num_rhs)
+{
+    return (2 * num_rows * num_rhs) * sizeof(ValueType) +
```
pratikvn, line 66: Maybe you can add a comment here showing the calculation  ?

cuda/base/config.hpp
```diff
@@ -74,6 +74,21 @@ struct config {
...
+    /**
+     * Max number of rows per matrix in a batch of (small) sparse matrices.
+     */
+    static constexpr int max_num_rows = 128;
+
+    /**
+     * Max number of RHS vectors in a linear system. This can be at most
+     * `config::warp_size`.
+     */
+    static constexpr int max_num_rhs = 5;
```
pratikvn, line 88: Is there any reason, you set this to 5, rather than `config::warp_size` ?

cuda/base/kernel_config.cuh
```diff
@@ -0,0 +1,65 @@
...
+namespace cuda {
+
+
+/**
+ * Set shared memory bank configuration.
+ *
+ * \tparam ValueType  The scalar type used for computations.
+ */
+template <typename ValueType>
+inline void configure_shared_memory()
```
pratikvn, line 51: Maybe this should be in detail ?

include/ginkgo/core/base/exception_helpers.hpp
```diff
@@ -255,6 +255,57 @@ inline std::tuple<bool, int> compare_batch_cols(const batch_dim<2> &size1,
...
+    if (size.stores_equal_sizes()) {
+        if (size.at(0)[0] != size.at(0)[1]) {
+            return std::tuple<bool, int>{false, 0};
+        } else {
+            return std::tuple<bool, int>{true, 0};
+        }
+    }
+
+    for (auto i = 0; i < size.get_num_batch_entries(); ++i) {
+        if (size.at(i)[1] != size.at(i)[1]) {
```
pratikvn, line 269: I think this is supposed to be
```suggestion
        if (size.at(i)[1] != size.at(i)[0]) {
```

include/ginkgo/core/preconditioner/batch_preconditioner_types.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+namespace gko {
+namespace preconditioner {
+namespace batch {
+
+
+/**
+ * Types of batch preconditioners available.
+ */
+enum Type { none, jacobi };
```
pratikvn, line 48: Prefer `enum class` to `enum`

include/ginkgo/core/preconditioner/batch_preconditioner_types.hpp
```diff
@@ -0,0 +1,72 @@
...
+ * Get a string name of an available batch preconditioner type.
+ */
+inline std::string get_string_of(Type type)
+{
+    if (type == none) {
+        return none_str;
+    } else {
+        return jacobi_str;
+    }
+}
```
pratikvn, line 65: Do you need the string setup ? Can you not directly compare with the enum ?

reference/solver/batch_richardson_kernels.cpp
```diff
@@ -0,0 +1,257 @@
...
+        const auto norms = reinterpret_cast<real_type *>(
+            prec_work + PrecType::dynamic_work_size(nrows, a.num_nnz));
+        real_type *const init_rel_res_norm = norms + nrhs;
+#else
+        ValueType residual[max_nrows * max_nrhs];
+        ValueType delta_x[max_nrows * max_nrhs];
+        ValueType prec_work[PrecType::work_size];
+        real_type norms[max_nrhs];
+        real_type init_rel_res_norm[max_nrhs];
+#endif
```
pratikvn, line 106: I think if this is not needed for reference, then we should just remove it and simplify the code here.

cuda/base/config.hpp
Slaedr: The static allocation depended on this. Setting it to 32 would make us run out shared memory much quicker for no good reason. If we choose to permanently remove static allocation, this whole struct can be removed. Since this is not user-facing, we could also do this later.

include/ginkgo/core/preconditioner/batch_preconditioner_types.hpp
Slaedr: The thing is, it's already in the `gko::preconditioner::batch` namespace. Since the user needs to use these, it'll be quite annoying to add yet another scope there.

include/ginkgo/core/preconditioner/batch_preconditioner_types.hpp
Slaedr: It's only there for possible error-reporting purposes.

reference/solver/batch_richardson_kernels.cpp
Slaedr: I was thinking of keeping the reference code kind of consistent with cuda code with regard to dynamic memory usage, so maybe we could remove this once the question is settled for Cuda? (But you're right that apart from a vague consistency argument, there's no real reason for this.)

cuda/base/kernel_config.cuh
Slaedr: Is it necessary? This is file is not included in any interface file.

reference/solver/batch_richardson_kernels.cpp
pratikvn: In general, I think there is no need to be consistent with CUDA. Actually, it is better to have code that is as simple as possible in reference. If there is some bug in CUDA and reference uses code similar to that, then this makes things complicated to test and debug.

include/ginkgo/core/preconditioner/batch_preconditioner_types.hpp
pratikvn: I think it is okay to have an additional scope there, because `enum classes` are bit safer, so I think that is acceptable.

cuda/base/kernel_config.cuh
pratikvn: I think it makes the code clearer, because you know that it is internal.

cuda/base/config.hpp
pratikvn: Yeah, we can decide this later.

include/ginkgo/core/preconditioner/batch_preconditioner_types.hpp
Slaedr: My understanding is that enum classes are safer only because the enum constants are within the class scope (public). Here, because it's already in the preconditioner::batch:: scope, it should be fine. 

But nevertheless, I'm okay with having an enum class I guess, if only to silence the Sonarcloud error.

reference/solver/batch_richardson_kernels.cpp
Slaedr: Then I would probably just keep the dynamic allocation version here. Then we could test any matrix size.

pratikvn:
I guess everything mostlty looks good from my side. The CI is failing, 
1. Because you are calling some of the apply kernels in `core/test/solver/batch_richardson.cpp`, which you probably need to move to the reference tests.
2. In [batch_preconditioner_types](https://github.com/ginkgo-project/ginkgo/pull/762/files#diff-585f8853801214a33ca6935038f7801466a92b2ff65fb4ddecfa9e8c4af6a8d1R58), you are inlining that function, so the I think  scope is lost, and hence you get the error that `type` is not an class, namespce or enum ... Of course this seems to only show up on the cuda-9.0 job. Using the complete namespace `gko::preconditioner::batch::type` should resolve the issue.

yhmtsai:
For the `__shared__`, does it mean we only support real valuetype in batch system?

Slaedr:
In the static shared memory setup, it does throw out a bunch of warnings when complex numbers are used. However, all the toolchains except the ones using clang as device compiler work correctly. But indeed, if we decide to use static shared memory again, I'll do something about that. Right now it's irrelevant because we are using dynamic shared memory.

PR #763, opened 2021-05-06
Add `set_krylov_dim` to CB-GMRES

thoasm:
This PR adds a `set_krylov_dim` function to CB-GMRES (which is already present in GMRES).


include/ginkgo/core/solver/cb_gmres.hpp
```diff
@@ -144,6 +144,13 @@ class CbGmres : public EnableLinOp<CbGmres<ValueType>>,
      */
     size_type get_krylov_dim() const { return krylov_dim_; }
 
+    /**
+     * Sets the krylov dimension
+     *
+     * @param other  the new krylov dimension
+     */
+    void set_krylov_dim(const size_type &other) { krylov_dim_ = other; }
```
tcojean, line 152: I only saw now that we use reference here, I don't think it's really needed for a simple number like that, but that's fine with me.

tcojean (approved):
LGTM.

pratikvn (approved):

include/ginkgo/core/solver/cb_gmres.hpp
thoasm: I simply copy-pasted the same function from `GMRES` to `CB-GMRES`.
I agree it is a bit weird to use a reference here, but it should not impact the usage.
I will change it for both to be a regular value. Thanks for mentioning it.

nbeams:
Thanks, I will go ahead and add this to the MFEM interface for both GMRES classes now.

yhmtsai (approved):
LGTM

include/ginkgo/core/solver/cb_gmres.hpp
```diff
@@ -144,6 +144,13 @@ class CbGmres : public EnableLinOp<CbGmres<ValueType>>,
      */
     size_type get_krylov_dim() const { return krylov_dim_; }
 
+    /**
+     * Sets the krylov dimension
```
upsj, line 148: nit: (also in GMRES)
```suggestion
     * Sets the Krylov dimension
```

upsj (approved):
LGTM!

PR #764, opened 2021-05-07
Naive lower triangular solver + updated cuSPARSE solvers

lksriemer:
This PR adds a minimal lower triangular solver to the CUDA backend, to have baseline to benchmark against going forward, as discussed with @upsj. 
For all non-`"sparselib"` matrices this will be the solver used.

This includes explicit store/load functions to get around using `volatile` on the `ValueType`, these functions were written by @upsj. I don't think anyone feels qualified to make any guarantees there, but they seem to work.

~A warning about unavailable dynamic initialization in shared memory is silenced. That should not be an issue, the values are manually initialized anyway.~

- [x] add support for multiple RHS


cuda/solver/lower_trs_kernels.cu
```diff
@@ -62,6 +62,127 @@ namespace cuda {
...
+    if (row >= n) {
+        return;
+    }
+
+    const auto self_shmem_id = gid / default_block_size;
+    const auto self_shid = gid % default_block_size;
+
+    volatile __shared__ uint32 ready_s[default_block_size];
+#pragma diag_suppress static_var_with_dynamic_init
+    __shared__ ValueType x_s[default_block_size];
```
upsj, line 123: we usually use `UninitializedArray` as a workaround for that :)

cuda/solver/lower_trs_kernels.cu
```diff
@@ -62,6 +62,127 @@ namespace cuda {
...
+    store(imag, 2 * index, value.imag());
+}
+
+template <typename ValueType, typename IndexType>
+__global__ void sptrsv_naive_caching_kernel(
+    const IndexType *const rowptrs, const IndexType *const colidxs,
+    const ValueType *const vals, const ValueType *const b, ValueType *const x,
+    volatile uint32 *const ready, const size_type n)
+{
+    const auto gid = blockIdx.x * blockDim.x + threadIdx.x;
```
upsj, line 111:
```suggestion
    const auto gid = thread::get_thread_id_flat();
```
from `components/thread_ids.cuh`

cuda/solver/lower_trs_kernels.cu
```diff
@@ -88,16 +209,20 @@ void generate(std::shared_ptr<const CudaExecutor> exec,
 GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
```
upsj, line 209: You could also make the generate_kernel call conditional on the strategy type

cuda/solver/lower_trs_kernels.cu
```diff
@@ -88,16 +209,20 @@ void generate(std::shared_ptr<const CudaExecutor> exec,
...
     GKO_DECLARE_LOWER_TRS_GENERATE_KERNEL);
 
-
```
upsj, line 91: nit: this line should stay to separate functions more clearly

upsj (commented):
Nice work! I need to go through the concurrency questions again, but a few suggestions for the surrounding code

upsj:
label!

upsj:
check-format!

cuda/solver/lower_trs_kernels.cu
```diff
@@ -62,6 +62,127 @@ namespace cuda {
...
+
+    __syncthreads();
+
+    const auto row_start = rowptrs[row];
+    const auto row_end = rowptrs[row + 1] - 1;
+
+    ValueType sum = 0.0;
+    for (IndexType i = row_start; i < row_end; ++i) {
+        const auto dependency = colidxs[i];
+        volatile auto is_ready = &ready[dependency];
```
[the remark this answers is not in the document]
lksriemer: Exactly, I wanted to circumvent a branch in the tight loop and instead use generic addressing. I measured different versions though, and if I recall correctly, they all were reasonably close in performance. Which would make sense, given that memory is the bottleneck.

cuda/solver/lower_trs_kernels.cu
lksriemer: That seems sensible. I wanted to leave all that to an analysis-phase PR, but on second thought, this will come in handy when comparing total runtime.

cuda/solver/lower_trs_kernels.cu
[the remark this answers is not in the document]
upsj: Another thing that just came to my mind - based on the thread ordering, shouldn't the shared-memory dependencies be the last to be fulfilled? Could we split this into a global memory, then shared memory wait?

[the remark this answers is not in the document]
lksriemer: That should be possible, and I believe I also did this in some versions. I just checked, it does not alter performance in one case that heavily benefits from shared memory, if anything I measure a slight degradation. The unified version however has the advantage of allowing for different orders, which might come up in the future.

lksriemer:
I just pushed a fallback kernel for pre-volta architectures, employing outer iteration. I tested it to work on a GTX760 (Kepler, CC 3.0) and MX150 (Pascal, CC 6.1). It doesn't include shared memory, but that should be fine, it's only a fallback after all.

cuda/solver/lower_trs_kernels.cu
```diff
@@ -62,6 +64,173 @@ namespace cuda {
...
+        auto col = colidxs[j];
+        while (ready[col]) {
+            sum += vals[j] * load(x, col);
+            ++j;
+            col = colidxs[j];
+        }
+        if (row == col) {
+            store(x, row, (b[row] - sum) / vals[row_end]);
+            ++j;
+            __threadfence();
```
upsj, line 196: Does it matter whether the threadfence is predicated?

cuda/solver/lower_trs_kernels.cu
```diff
@@ -62,6 +64,173 @@ namespace cuda {
...
+__device__ void store(
+    ValueType *values, int index,
+    std::enable_if_t<std::is_floating_point<ValueType>::value, ValueType> value)
+{
+    volatile ValueType *val = values + index;
+    *val = value;
+}
+
+template <typename ValueType>
+__device__ void store(thrust::complex<ValueType> *values, int index,
```
upsj, line 98: Do you still see performance degredation from using these separate stores/loads? Should we just remove them?

cuda/solver/lower_trs_kernels.cu
lksriemer: I did a quick test on one matrix, there performance indeed improves by ~5% by factoring it out. Good catch!

cuda/solver/lower_trs_kernels.cu
lksriemer: No, at least not from quick tests. Forceinlining also doesn't change it, so I guess we're good to go with these.

cuda/solver/lower_trs_kernels.cu
lksriemer: I have to correct this: I by accident noticed that, at least for one matrix, the fallback version demands the minimal block size of 32 threads for acceptable performance. In that scenario, the threadfence should be predicated. Then again, these measurements are from my post-volta GPU, perhaps it looks different on a P100.

upsj:
trisolve
performance profile: csrc is ours, csrs is cuSPARSE (pre-11.3 or >= 11.3)

thoasm (dismissed):
Well done on the code and good documentation!

However, it looks like your `sptrsv_naive_caching_kernel` and `sptrsv_naive_legacy_kernel` kernels could loop indefinitely (according to the CUDA documentation). I assume that on current architectures, this might never happen, but we should make sure to also support future architectures.

cuda/solver/common_trs_kernels.cuh
```diff
@@ -229,6 +320,210 @@ void solve_kernel(std::shared_ptr<const CudaExecutor> exec,
...
+            (dependency_gid / default_block_size) == self_shmem_id;
+        if (shmem_possible) {
+            const auto dependency_shid = dependency_gid % default_block_size;
+            x_p = &x_s[dependency_shid];
+        }
+
+        ValueType x = *x_p;
+        while (is_nan(x)) {
+            x = load(x_p, 0);
+        }
```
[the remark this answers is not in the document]
upsj: I pushed a commit to address this issue using a atomic thread block counter. @lksriemer can you give this a second look? Maybe also check for the performance impact on a V100? I haven't checked there yet.

cuda/solver/common_trs_kernels.cuh
```diff
@@ -366,9 +368,15 @@ __global__ void sptrsv_naive_caching_kernel(
     const IndexType* const rowptrs, const IndexType* const colidxs,
     const ValueType* const vals, const ValueType* const b, size_type b_stride,
     ValueType* const x, size_type x_stride, const size_type n,
-    const size_type nrhs, bool* nan_produced)
+    const size_type nrhs, bool* nan_produced, IndexType* atomic_counter)
 {
-    const auto full_gid = thread::get_thread_id_flat<IndexType>();
+    __shared__ IndexType block_base_idx;
```
lksriemer, line 373: As discussed, we need to move the `x_s_array` declaration above this one, to preserve it's alignment.

cuda/solver/common_trs_kernels.cuh
```diff
@@ -366,9 +368,15 @@ __global__ void sptrsv_naive_caching_kernel(
...
+    if (threadIdx.x == 0) {
+        block_base_idx =
+            atomic_add(atomic_counter, IndexType{1}) * default_block_size;
```
lksriemer, line 376: Doesn't `block_base_idx` need to be marked as volatile, or otherwise reads to it could be optimized away? Though this behavior does not seem to occur.

cuda/solver/common_trs_kernels.cuh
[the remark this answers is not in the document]
lksriemer: To document this here: I already checked the performance impact on the V100 some time ago, there is no significant measurable degradation.

cuda/solver/common_trs_kernels.cuh
yhmtsai: what do you mean here? the block_base_idx is only shared in thread-block, and there is a `__syncthreads()` so I think it should be okay?

yhmtsai (approved):
LGTM. only some comments/format requirement is not applied.
Thanks for handling it. 
If it does not require too much work, could you also update the hip part except for cuda 11 part?
If it does, we can also move to another pr to update hip part

cuda/solver/common_trs_kernels.cuh
```diff
@@ -492,8 +506,10 @@ void sptrsv_naive_caching(std::shared_ptr<const CudaExecutor> exec,
     x->fill(nan<ValueType>());
 
     Array<bool> nan_produced(exec, {false});
+    Array<IndexType> atomic_counter(exec, {IndexType{}});
```
thoasm: Comment: For small matrices, both allocating and copying the data over takes a long time.
For initialization, I recommend using a initialize kernel, which is significantly faster (I don't have the numbers anymore, but it made quite a big difference using a kernel compared to copying the data).

lksriemer: Are you talking about the marked section? This is just two single values, did you mean those with "small matrices"?

thoasm: Yes, I am talking about the marked section.
Both boil down to one `cudaMalloc` and one `cudaMemcpy` each (since you provide an initial value, this will be copied through `cudaMemcpy`).
If you remove the initialization (making it `Array<type> val(exec)`), you don't have the `cudaMemcpy`.
Having a kernel that initializes the data instead of copying is faster than copying single values, so only allocating (no initialization in the `gko::Array`) followed by an initialization kernel is faster than the current method. However, that is only noticeable for small matrices, so I wouldn't consider important. I just wanted to mention it.

lksriemer: Now I see what you mean. Sure, I'll include that in the review changes, which I will push probably tomorrow.

cuda/solver/common_trs_kernels.cuh
```diff
@@ -229,6 +320,210 @@ void solve_kernel(std::shared_ptr<const CudaExecutor> exec,
...
+
+    ValueType sum = 0.0;
+    auto j = row_begin;
+    while (j != row_diag + row_step) {
+        auto col = colidxs[j];
+        while (!is_nan(load(x, col * x_stride + rhs))) {
+            sum += vals[j] * load(x, col * x_stride + rhs);
+            j += row_step;
+            col = colidxs[j];
+        }
```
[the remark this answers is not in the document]
thoasm: nit: this is still using two `load`s when only one would be necessary.

thoasm (approved):
LGTM, good work!

cuda/solver/lower_trs_kernels.cu
```diff
@@ -69,20 +69,16 @@ void should_perform_transpose(std::shared_ptr<const CudaExecutor> exec,
...
 template <typename ValueType, typename IndexType>
 void generate(std::shared_ptr<const CudaExecutor> exec,
               const matrix::Csr<ValueType, IndexType>* matrix,
-              solver::SolveStruct* solve_struct, const gko::size_type num_rhs)
+              std::shared_ptr<solver::SolveStruct>& solve_struct,
+              const gko::size_type num_rhs)
 {
-    generate_kernel<ValueType, IndexType>(exec, matrix, solve_struct, num_rhs,
-                                          false);
+    if (matrix->get_strategy()->get_name() == "sparselib") {
```
Slaedr, line 78: There was some talk earlier about overhauling the strategy system. Are we are keeping it as it is? I guess it does sound like a good idea to have the strategy control not only the spmv but also the trsv.

upsj: We were planning on using a factory parameter instead of the strategy, so yes, that makes sense.

upsj:
rebase!

PR #765, opened 2021-05-08
Parallelize COO OpenMP SpMV

upsj:
This PR parallelizes the COO OpenMP SpMV by using atomics for rows that get handled by multiple threads.

### DIFF
--- a/omp/components/atomic.hpp
+++ b/omp/components/atomic.hpp
@@ -1,0 +1,73 @@
+/*******************************<GINKGO LICENSE>******************************
+Copyright (c) 2017-2021, the Ginkgo authors
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+
+1. Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+
+2. Redistributions in binary form must reproduce the above copyright
+notice, this list of conditions and the following disclaimer in the
+documentation and/or other materials provided with the distribution.
+
+3. Neither the name of the copyright holder nor the names of its
+contributors may be used to endorse or promote products derived from
+this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS
+IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED
+TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A
+PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+#ifndef GKO_OMP_COMPONENTS_ATOMIC_HPP_
+#define GKO_OMP_COMPONENTS_ATOMIC_HPP_
+
+
+#include <type_traits>
+
+
+#include <ginkgo/core/base/types.hpp>
+
+
+namespace gko {
+namespace kernels {
+namespace omp {
+
+
+template <typename ValueType,
+          std::enable_if_t<!is_complex<ValueType>()> * = nullptr>
+void atomic_add(ValueType &out, ValueType val)
+{
+#pragma omp atomic
+    out += val;
+}
+
+template <typename ValueType,
+          std::enable_if_t<is_complex<ValueType>()> * = nullptr>
+void atomic_add(ValueType &out, ValueType val)
+{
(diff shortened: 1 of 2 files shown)


omp/matrix/coo_kernels.cpp
```diff
@@ -103,15 +104,45 @@ void spmv2(std::shared_ptr<const OmpExecutor> exec,
...
+    const auto coo_val = a->get_const_values();
+    const auto coo_col = a->get_const_col_idxs();
+    const auto coo_row = a->get_const_row_idxs();
+    const auto num_rhs = b->get_size()[1];
+    const auto sentinel_row = a->get_size()[0] + 1;
+    const auto nnz = a->get_num_stored_elements();
+    const auto num_threads = omp_get_max_threads();
+    const auto work_per_thread = ceildiv(nnz, num_threads);
+
+#pragma omp parallel num_threads(num_threads)
```
upsj, line 116: I didn't bother to dive too deep into the OpenMP Specs to see if num_threads is necessary here. Probably not (:

omp/matrix/coo_kernels.cpp
```diff
@@ -103,15 +104,46 @@ void spmv2(std::shared_ptr<const OmpExecutor> exec,
...
+        if (begin < end) {
+            const auto first = begin > 0 ? coo_row[begin - 1] : sentinel_row;
+            const auto last = end < nnz ? coo_row[end] : sentinel_row;
+            auto nz = begin;
+            for (; nz < end && coo_row[nz] == first; nz++) {
+                const auto row = first;
+                const auto col = coo_col[nz];
+                for (size_type rhs = 0; rhs < num_rhs; rhs++) {
+                    atomic_add(c->at(row, rhs), coo_val[nz] * b->at(col, rhs));
+                }
```
upsj, line 131: Here we could optimize for small numbers of RHS, keeping the partial sum in registers and doing only a single atomic_add afterwards

omp/matrix/coo_kernels.cpp
Slaedr: True, you don't need it.

omp/matrix/coo_kernels.cpp
```diff
@@ -103,15 +104,46 @@ void spmv2(std::shared_ptr<const OmpExecutor> exec,
...
-    for (size_type j = 0; j < num_cols; j++) {
-        for (size_type i = 0; i < a->get_num_stored_elements(); i++) {
-            c->at(coo_row[i], j) += coo_val[i] * b->at(coo_col[i], j);
+    const auto coo_val = a->get_const_values();
+    const auto coo_col = a->get_const_col_idxs();
+    const auto coo_row = a->get_const_row_idxs();
+    const auto num_rhs = b->get_size()[1];
+    const auto sentinel_row = a->get_size()[0] + 1;
+    const auto nnz = a->get_num_stored_elements();
+    const auto num_threads = omp_get_max_threads();
```
Slaedr, line 113:
```suggestion
```

omp/matrix/coo_kernels.cpp
```diff
@@ -126,17 +158,49 @@ void advanced_spmv2(std::shared_ptr<const OmpExecutor> exec,
...
+    const auto work_per_thread =
+        static_cast<size_type>(ceildiv(nnz, num_threads));
+    const auto scale = alpha->at(0, 0);
+
+#pragma omp parallel num_threads(num_threads)
```
Slaedr, line 172:
```suggestion
#pragma omp parallel
```

Slaedr (approved):
LGTM! Nice job separating the work into atomic and non-atomic parts. At some point later, I guess we should adapt the GPU algorithm for vectorization.

omp/components/atomic.hpp
```diff
@@ -0,0 +1,73 @@
...
+{
+#pragma omp atomic
+    out += val;
+}
+
+template <typename ValueType,
+          std::enable_if_t<is_complex<ValueType>()> * = nullptr>
+void atomic_add(ValueType &out, ValueType val)
+{
+    auto values = reinterpret_cast<gko::remove_complex<ValueType> *>(&out);
```
upsj, line 60: Just a note here in case somebody is worried about undefined behavior:
The C++ standard contains [explicit provisions for reinterpret-casting std::complex](https://eel.is/c++draft/complex.numbers#general), so we should be fine :)

omp/matrix/coo_kernels.cpp
upsj: I think still, the most stable solution would be to compute the work per thread inside the parallel region, then we know how many threads were actually spawned. Otherwise we might run into issues if this gets called from within a parallel region?

omp/matrix/coo_kernels.cpp
yhmtsai: Yes, 
the atomic is only needed when the row is the head/end of the handled queue. I also try it in cuda kernel but it does not give much difference on NVIDIA GPU

omp/matrix/coo_kernels.cpp
```diff
@@ -103,15 +104,46 @@ void spmv2(std::shared_ptr<const OmpExecutor> exec,
...
+                const auto col = coo_col[nz];
+                for (size_type rhs = 0; rhs < num_rhs; rhs++) {
+                    c->at(row, rhs) += coo_val[nz] * b->at(col, rhs);
+                }
+            }
+            for (; nz < end; nz++) {
+                const auto row = last;
+                const auto col = coo_col[nz];
+                for (size_type rhs = 0; rhs < num_rhs; rhs++) {
+                    atomic_add(c->at(row, rhs), coo_val[nz] * b->at(col, rhs));
```
yhmtsai, line 144:
```suggestion
                    atomic_add(c->at(last, rhs), coo_val[nz] * b->at(col, rhs));
```
maybe directly use `last`?

omp/matrix/coo_kernels.cpp
yhmtsai: maybe use one temp array not register to store the result?

yhmtsai (approved):
Adding atomic add for complex is useful

omp/matrix/coo_kernels.cpp
upsj: I think in the context of few cores and many rows, this should not be too much of a performance limit. If it does turn out later to cause performance issues, we can definitely change that. A special case for 1 or few columns might also be worthwile then.

omp/matrix/coo_kernels.cpp
upsj: I personally prefer keeping `row`, since it makes the code more uniform compared to the other loops.

PR #768, opened 2021-05-11
Add benchmarking support for Batch matrix formats

pratikvn:
This PR adds benchmarking functionality to BatchCsr and possibly any other future batch matrix formats. 

To allow for easy benchmarking, two new constructors were added to `BatchCsr` allowing duplication of CSR matrices or duplication of `BatchCsr` matrices. 

To facilitate benchmarking with the Suitesparse matrix collection, a new `NUM_BATCHES` environment variable can be set, which reads the input matrix ( in CSR ) and duplicates `NUM_BATCHES` of it and stores it in the `BatchCsr` format. 

### TODO

+ [x] Read different matrices for different batch entries. 
+ [x] Allow reading of scaled matrices and right hand sides.
+ [x] Allow reading of scaling vectors.

### DIFF
--- a/benchmark/run_all_benchmarks.sh
+++ b/benchmark/run_all_benchmarks.sh
@@ -35,9 +35,9 @@ if [ ! "${FORMATS}" ]; then
     FORMATS="batch_csr"
 fi
 
-if [ ! "${NUM_BATCHES}" ]; then
-    NUM_BATCHES="1"
-    echo "NUM_BATCHES environment variable not set - assuming \"${NUM_BATCHES}\"" 1>&2
+if [ ! "${NUM_BATCH_DUP}" ]; then
+    NUM_BATCH_DUP="1"
+    echo "NUM_BATCH_DUP environment variable not set - assuming \"${NUM_BATCH_DUP}\"" 1>&2
 fi
 
 if [ ! "${SOLVERS}" ]; then
@@ -72,7 +72,7 @@ fi
 
 if [ ! "${SOLVERS_JACOBI_MAX_BS}" ]; then
     SOLVERS_JACOBI_MAX_BS="32"
-    "SOLVERS_JACOBI_MAX_BS environment variable not set - assuming \"${SOLVERS_JACOBI_MAX_BS}\"" 1>&2
+    echo "SOLVERS_JACOBI_MAX_BS environment variable not set - assuming \"${SOLVERS_JACOBI_MAX_BS}\"" 1>&2
 fi
 
 if [ ! "${BENCHMARK_PRECISION}" ]; then
@@ -141,6 +141,20 @@ else
     DETAILED_STR="--detailed=true"
 fi
 
+if  [ ! "${BATCH_SCALING}" ] || [ "${BATCH_SCALING}" -eq 0 ]; then
+    BATCH_SCALING_STR="--batch_scaling=false"
+else
+    BATCH_SCALING_STR="--batch_scaling=true"
+fi
+
+if  [ ! "${USE_SUITE_SPARSE}" ] || [ "${USE_SUITE_SPARSE}" -eq 1 ]; then
+    SS_STR="--using_suite_sparse=true"
+    echo "Using matrices from SuiteSparse"
+else
+    SS_STR="--using_suite_sparse=false"
+    echo "Not using matrices from SuiteSparse"
+fi
+
 # This allows using a matrix list file for benchmarking.
 # The file should contains a suitesparse matrix on each line.
 # The allowed formats to target suitesparse matrix is:
@@ -155,6 +169,42 @@ elif [ -f "${MATRIX_LIST_FILE}" ]; then
     use_matrix_list_file=1
 else
     echo -e "A matrix list file was set to ${MATRIX_LIST_FILE} but it cannot be found."
+    exit 1
+fi
+
+if [ ! "${BATCH_MATRIX_LIST_FILE}" ]; then
+    use_batch_matrix_list_file=0
+elif [ -f "${BATCH_MATRIX_LIST_FILE}" ]; then
+    use_batch_matrix_list_file=1
+    echo "A batch matrix list file was set to ${BATCH_MATRIX_LIST_FILE}"
+else
+    echo -e "A batch matrix list file was set to ${BATCH_MATRIX_LIST_FILE} but it cannot be found."
+    exit 1
+fi
+
+# This allows using a folder to automatically read the files in the folder into a matrix.
+# The folder structure for a matrix class with two batch entries could look like this:
+#
+# matrix_class
+# \__1
+#     \__A.mtx
+#     \__b.mtx
+#     \__x0.mtx
+# \__2
+#     \__A.mtx
+#     \__b.mtx
+#     \__x0.mtx
+if [ ! "${BATCH_MATRIX_FOLDER}" ]; then
+    use_batch_matrix=0
+elif [ -d "${BATCH_MATRIX_FOLDER}" ]; then
+    use_batch_matrix=1
+    echo "A batch matrix folder was set to ${BATCH_MATRIX_FOLDER}"
+    if [ -f "${MATRIX_LIST_FILE}" ]; then
+        echo "Cannot use both matrix file list and batch matrix folder. The matrix file list will be ignored."
+        use_matrix_list_file=0
+    fi
+else
+    echo -e "A matrix folder was set to ${BATCH_MATRIX_FOLDER} but it cannot be found."
     exit 1
 fi
 
@@ -238,7 +288,8 @@ run_batch_spmv_benchmarks() {
     cp "$1" "$1.imd" # make sure we're not loosing the original input
     ./spmv/batch_spmv${BENCH_SUFFIX} --backup="$1.bkp" --double_buffer="$1.bkp2" \
                 --executor="${EXECUTOR}" --formats="${FORMATS}" \
-                --num_batches="${NUM_BATCHES}" \
+                --num_duplications="${NUM_BATCH_DUP}" "${BATCH_SCALING_STR}" \
+                --num_batches="${NUM_BATCH_ENTRIES}" "${SS_STR}" \
                 --device_id="${DEVICE_ID}" --gpu_timer=${GPU_TIMER} \
                 <"$1.imd" 2>&1 >"$1"
     keep_latest "$1" "$1.bkp" "$1.bkp2" "$1.imd"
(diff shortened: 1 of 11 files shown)


include/ginkgo/core/matrix/batch_csr.hpp
```diff
@@ -247,6 +247,77 @@ class BatchCsr
...
+    BatchCsr(std::shared_ptr<const Executor> exec,
+             const size_type num_batch_entries,
+             const matrix::Csr<value_type, index_type> *csr_mat)
+        : EnableBatchLinOp<BatchCsr>(
+              exec, batch_dim<2>(num_batch_entries, csr_mat->get_size())),
+          values_(exec, csr_mat->get_num_stored_elements() * num_batch_entries),
+          col_idxs_(exec, csr_mat->get_num_stored_elements()),
+          row_ptrs_(exec, (csr_mat->get_size()[0]) + 1)
+    {
+        size_type offset = 0;
```
Slaedr, line 267: Minor note: you could put this code block in a common templated function and just call that function in the two new constructors.

core/test/matrix/batch_csr.cpp
```diff
@@ -152,6 +163,59 @@ TYPED_TEST(BatchCsr, CanBeEmpty)
...
+    auto batch_mtx = gko::matrix::BatchCsr<value_type, index_type>::create(
+        this->exec, 2, gko::dim<2>{3, 2},
+        gko::Array<value_type>::view(this->exec, 8, values),
+        gko::Array<index_type>::view(this->exec, 4, col_idxs),
+        gko::Array<index_type>::view(this->exec, 4, row_ptrs));
+
+    auto mtx = gko::matrix::BatchCsr<value_type, index_type>::create(
+        this->exec, 3, batch_mtx.get());
+
+    this->assert_equal_data_array(24, bvalues, mtx->get_values());
```
Slaedr, line 213: You could also test the size of the created matrix to make sure the metadata is populated correctly.

benchmark/utils/general.hpp
```diff
@@ -406,6 +447,42 @@ gko::remove_complex<ValueType> compute_residual_norm(
 }
```
Slaedr, line 449:
```suggestion

/**
 * Computes the 2-norms of the difference between corresponding pairs of vectors
 * in two given batches, and returns the maximum relative norm among the different RHS
 * for each pair.
 */
```

benchmark/utils/formats.hpp
```diff
@@ -156,6 +159,29 @@ namespace formats {
...
+ * @param data  the data represented in the intermediate representation format
+ *
+ * @tparam MatrixType  the Ginkgo matrix type (such as `gko::matrix::Csr<>`)
+ *
+ * @return a `unique_pointer` to the created matrix
+ */
+template <typename MatrixType>
+std::unique_ptr<MatrixType> read_batch_matrix_from_data(
+    std::shared_ptr<const gko::Executor> exec, const int num_duplications,
+    const gko::matrix_data<etype> &data)
```
Slaedr, line 178: Wouldn't it be more useful to take a vector of matrix_data (representing an entire batch) as input, and duplicate the entire batch?

benchmark/spmv/batch_spmv.cpp
```diff
@@ -0,0 +1,270 @@
...
+            if (!FLAGS_overwrite &&
+                all_of(begin(formats), end(formats),
+                       [&spmv_case](const std::string &s) {
+                           return spmv_case.HasMember(s.c_str());
+                       })) {
+                continue;
+            }
+            std::clog << "Running test case: " << test_case << std::endl;
+            std::ifstream mtx_fd(test_case["filename"].GetString());
+            auto data = gko::read_raw<etype>(mtx_fd);
```
Slaedr, line 226: I guess you're currently working on this. Maybe we can have new read_raw that takes a file name prefix, appends numeric suffixes to it to read each of the batch entries (assuming the file names are like that), and then returns a vector<matrix_data>?

Or, we can have a descriptor file that has the metadata for the batch, and we point read_raw to that file. That file can then be parsed to get the number of small matrices in the batch and the prefix for files containing each of the matrices.

Slaedr (commented):
Good progress! Some initial thoughts.

include/ginkgo/core/matrix/batch_csr.hpp
pratikvn: Good point.

core/test/matrix/batch_csr.cpp
pratikvn: I have added this now

benchmark/run_all_benchmarks.sh
```diff
@@ -2,7 +2,7 @@
 # Environment variable detection
 
 if [ ! "${BENCHMARK}" ]; then
-    BENCHMARK="spmv"
+    BENCHMARK="batch_spmv"
```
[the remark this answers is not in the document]
pratikvn: I think it makes sense to have some custom defaults for us. When merging it into `develop`, we can switch back to the normal `spmv`

benchmark/spmv/batch_spmv.cpp
pratikvn: Yeah, I was planning on doing more or less the above. Now we populate the input file class and read the `matrix_data`. And we can either benchmark by duplicating either the input batch matrices as well or by using ssget and duplicating that.

benchmark/spmv/batch_spmv.cpp
```diff
@@ -0,0 +1,484 @@
...
+                    apply_spmv(format_name.c_str(), exec, data[0], lend(b),
+                               lend(x), lend(answer), test_case, allocator);
+                } else {
+                    apply_spmv(format_name.c_str(), exec, data, lend(b),
+                               lend(x), lend(answer), test_case, allocator);
+                }
+                std::clog << "Current state:" << std::endl
+                          << test_cases << std::endl;
+                if (spmv_case[format_name.c_str()]["completed"].GetBool()) {
+                    auto performance =
```
Slaedr, line 472: Is there something missing here? It looks like `performance` will be simply be deleted.

benchmark/spmv/batch_spmv.cpp
```diff
@@ -0,0 +1,484 @@
...
+                const batch_vec<etype> *b, const batch_vec<etype> *x,
+                const batch_vec<etype> *answer, rapidjson::Value &test_case,
+                rapidjson::MemoryPoolAllocator<> &allocator)
+{
+    try {
+        auto &spmv_case = test_case["spmv"];
+        add_or_set_member(spmv_case, format_name,
+                          rapidjson::Value(rapidjson::kObjectType), allocator);
+
+        auto nbatch = FLAGS_num_duplications;
```
Slaedr, line 212: Just for clarity, as this is no longer the number of matrices in the batch.
```suggestion
        auto n_dup = FLAGS_num_duplications;
```

benchmark/spmv/batch_spmv.cpp
```diff
@@ -0,0 +1,484 @@
...
+        // check the residual
+        if (FLAGS_detailed) {
+            auto x_clone = clone(x);
+            exec->synchronize();
+            system_matrix->apply(lend(b), lend(x_clone));
+            exec->synchronize();
+            auto max_relative_norm2 =
+                compute_batch_max_relative_norm2(lend(x_clone), lend(answer));
+            // FIXME
+            // add_or_set_member(spmv_case[format_name], "max_relative_norm2",
```
Slaedr, line 238: I guess this whole if-block can be commented out if this is commented out?

Slaedr (changes requested):
Overall, this looks good. I guess it will be straightforward to extend this for the solvers, except that we'll have to push the new batch logger.

However, I believe there is an inconsistency between the file-list setting and the folder setting in the run script, though I might have misunderstood something. Please see below.

benchmark/run_all_benchmarks.sh
```diff
@@ -153,6 +185,54 @@ else
...
+# \__2
+#     \__A.mtx
+#     \__b.mtx
+#     \__x0.mtx
+if [ ! "${BATCH_MATRIX_FOLDER}" ]; then
+    use_batch_matrix=0
+elif [ -d "${BATCH_MATRIX_FOLDER}" ]; then
+    use_batch_matrix=1
+    echo "A batch matrix folder was set to ${BATCH_MATRIX_FOLDER}"
+    if [ -f "${MATRIX_LIST_FILE}" ]; then
```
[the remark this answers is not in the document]
pratikvn: No, this is a check so that if `MATRIX_LIST_FILE` has also been set, then we explicitly throw a warning and set `use_matrix_list_file` to 0

benchmark/run_all_benchmarks.sh
```diff
@@ -153,6 +185,54 @@ else
...
+        echo "Cannot use both matrix file list and batch matrix folder. The matrix file list will be ignored."
+        use_matrix_list_file=0
```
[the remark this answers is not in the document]
pratikvn: same as above

benchmark/run_all_benchmarks.sh
```diff
@@ -273,120 +372,180 @@ run_preconditioner_benchmarks() {
...
-        benchmark_list="$benchmark_list $id"
+        benchmark_list="$benchmark_list $mtx"
     done
     echo "$benchmark_list"
 }
 
-if [ $use_matrix_list_file -eq 1 ]; then
-    MATRIX_LIST=($(parse_matrix_list $MATRIX_LIST_FILE))
-    NUM_PROBLEMS=${#MATRIX_LIST[@]}
+if [ $use_batch_matrix -eq 1 ]; then
```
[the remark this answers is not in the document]
pratikvn: Yeah, sorry for the confusion here. I will slightly re-organize this.

benchmark/spmv/batch_spmv.cpp
pratikvn: Yes, you are right. This was for showing the best spmv format, which is not relevant for batch matrices, atleast not right now.

core/test/matrix/batch_dense.cpp
```diff
@@ -172,6 +172,29 @@ TYPED_TEST(BatchDense, CanBeConstructedFromExistingData)
...
+    auto m = gko::matrix::BatchDense<TypeParam>::create(
+        this->exec, std::vector<DenseMtx *>{mat1.get(), mat2.get()});
+    auto m_ref = gko::matrix::BatchDense<TypeParam>::create(
+        this->exec,
+        std::vector<DenseMtx *>{mat1.get(), mat2.get(), mat1.get(), mat2.get(),
+                                mat1.get(), mat2.get()});
+    auto m2 =
+        gko::matrix::BatchDense<TypeParam>::create(this->exec, 3, m.get());
+
+    GKO_ASSERT_BATCH_MTX_NEAR(m2.get(), m_ref.get(), 1e-14);
```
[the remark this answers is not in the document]
pratikvn: I think we are comparing the matrices themselves, so it should be fine, I think. Because the macro, internally does check the stride, size and others

benchmark/run_all_benchmarks.sh
```diff
@@ -273,120 +372,180 @@ run_preconditioner_benchmarks() {
...
+[{
+    "problem": "${BASE_DIR}/${INPUT}"
+}]
+EOT
+}
 
-    echo -e "${PREFIX}Extracting statistics for ${GROUP}/${NAME}" 1>&2
-    compute_matrix_statistics "${RESULT_FILE}"
+if [ $use_batch_matrix -eq 1 ]; then
+    for (( p=0; p < ${NUM_BATCH_MATS}; ++p )); do
```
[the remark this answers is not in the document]
pratikvn: No, the `BATCH_MATRIX_FOLDER` and `BATCH_MATRIX_LIST` are two separate things. `X_FOLDER` contains the list of classes. Only some of them might be benchmarked. The list of matrix classes to be benchmarked needs to be specified by the user in a text file stored in the env var `BATCH_MATRIX_LIST_FILE` in a similar fashion to `MATRIX_LIST_FILE`.

benchmark/run_all_benchmarks.sh
```diff
@@ -273,120 +370,180 @@ run_preconditioner_benchmarks() {
...
+        benchmark_list="$benchmark_list $mtx"
     done
     echo "$benchmark_list"
 }
 
-if [ $use_matrix_list_file -eq 1 ]; then
-    MATRIX_LIST=($(parse_matrix_list $MATRIX_LIST_FILE))
-    NUM_PROBLEMS=${#MATRIX_LIST[@]}
+if [ $use_batch_matrix -eq 1 ]; then
+    BATCH_MATRIX_LIST=$(parse_batch_matrix_list $BATCH_MATRIX_LIST_FILE)
```
Slaedr, line 391: I guess this is the list of matrix classes? In that case, it would be easier to understand with
```suggestion
    BATCH_MATRIX_TYPES_LIST=$(parse_batch_matrix_list $BATCH_MATRIX_LIST_FILE)
```

benchmark/run_all_benchmarks.sh
```diff
@@ -273,120 +370,180 @@ run_preconditioner_benchmarks() {
...
+    NUM_BATCH_MATS=${#BATCH_MATRIX_LIST[@]}
```
Slaedr, line 392:
```suggestion
    NUM_BATCH_MAT_TYPES=${#BATCH_MATRIX_LIST[@]}
```

Slaedr (approved):
LGTM! I now see that to do the batch benchmark you always need either (1) both a matrix file and a matrix folder, or (2) a file of suitesparse matrix names/IDs. In case 1, I guess we need to generate this file, which contains the matrix classes we want to benchmark, and set an environment variable pointing to the folder containing all the systems in subfolders. It would be nice to document this somewhere.

I also have a few minor documentation-related suggestions.

PR #769, opened 2021-05-20
Fix the error size of extract_diagonal

yhmtsai:
This PR uses the diag size not matrix size for extract_diagonal to avoid the segmentation fault when `#rows > #cols`.

TODO:
- [x] also add test case `#cols > #rows`


pratikvn (approved):
LGTM!

tcojean (approved):

upsj (approved):

PR #771, opened 2021-05-20
Add benchmarking for batch solvers

pratikvn:
This PR adds benchmarking support for batch solvers. Currently, only Richardson is supported.


pratikvn:
The benchmarks work now, and here is an example of the output:

[1138_bus.json.txt](https://github.com/ginkgo-project/ginkgo/files/6530855/1138_bus.json.txt)

benchmark/solver/batch_solver.cpp
```diff
@@ -0,0 +1,517 @@
...
+    if (description == "richardson") {
+        using Solver = gko::solver::BatchRichardson<etype>;
+        return Solver::build()
+            .with_max_iterations(static_cast<int>(FLAGS_max_iters))
+            .with_rel_residual_tol(
+                static_cast<gko::remove_complex<etype>>(FLAGS_rel_res_goal))
+            .with_preconditioner(gko::preconditioner::batch::type::jacobi)
+            .with_relaxation_factor(
+                static_cast<gko::remove_complex<etype>>(0.95))
+            // FLAGS_relaxation_factor))
```
Slaedr, line 247: I guess you need to re-enable this, because it might be good to briefly look at whether simple preconditioned Richardson can solve those systems and how it compares with GMRES etc.

examples/batched-solver/batched-solver.cpp
```diff
@@ -0,0 +1,192 @@
...
+    using real_vec = gko::matrix::Dense<RealValueType>;
+    using real_bvec = gko::matrix::BatchDense<RealValueType>;
+    // The gko::matrix::Csr class is used here, but any other matrix class such
+    // as gko::matrix::Coo, gko::matrix::Hybrid, gko::matrix::Ell or
+    // gko::matrix::Sellp could also be used.
+    using mtx = gko::matrix::Csr<ValueType, IndexType>;
+    using bmtx = gko::matrix::BatchCsr<ValueType, IndexType>;
+    // The gko::solver::Cg is used here, but any other solver class can also be
+    // used.
+    using cg = gko::solver::Cg<ValueType>;
```
Slaedr, line 69:
```suggestion
```

examples/batched-solver/batched-solver.cpp
```diff
@@ -0,0 +1,192 @@
...
+    // read function.
+    // @note Ginkgo uses C++ smart pointers to automatically manage memory. To
+    // this end, we use our own object ownership transfer functions that under
+    // the hood call the required smart pointer functions to manage object
+    // ownership. The gko::share , gko::give and gko::lend are the functions
+    // that you would need to use.
+    auto A = share(gko::read<mtx>(std::ifstream("data/A.mtx"), exec));
+    auto b = gko::read<vec>(std::ifstream("data/b.mtx"), exec);
+    auto x = gko::read<vec>(std::ifstream("data/x0.mtx"), exec);
+
```
Slaedr, line 126:
```suggestion
// Duplicate the read matrices to create batches.
```

examples/batched-solver/batched-solver.cpp
```diff
@@ -0,0 +1,192 @@
...
+    // to be built on.
+    const RealValueType reduction_factor{1e-7};
+    // auto solver_gen =
+    //     cg::build()
+    //         .with_criteria(
+    //             gko::stop::Iteration::build().with_max_iters(20u).on(exec),
+    //             gko::stop::ResidualNorm<ValueType>::build()
+    //                 .with_reduction_factor(reduction_factor)
+    //                 .on(exec))
+    //         .on(exec);
```
Slaedr, line 147:
```suggestion
```

examples/batched-solver/batched-solver.cpp
```diff
@@ -0,0 +1,192 @@
...
+    auto solver_gen = rich::build()
+                          .with_max_iterations(500)
+                          .with_rel_residual_tol(reduction_factor)
+                          .with_relaxation_factor(RealValueType{0.95})
+                          .on(exec);
+    // Generate the solver from the matrix. The solver factory built in the
```
Slaedr, line 153:
```suggestion
    // Generate the solver from the batch of matrices.
```

examples/batched-solver/batched-solver.cpp
```diff
@@ -0,0 +1,192 @@
...
+    // previous step takes a "matrix"(a gko::LinOp to be more general) as an
+    // input. In this case we provide it with a full matrix that we previously
+    // read, but as the solver only effectively uses the apply() method within
+    // the provided "matrix" object, you can effectively create a gko::LinOp
+    // class with your own apply implementation to accomplish more tasks. We
+    // will see an example of how this can be done in the custom-matrix-format
+    // example
+    // auto solver = solver_gen->generate(A);
```
Slaedr, line 161:
```suggestion
```

examples/batched-solver/batched-solver.cpp
```diff
@@ -0,0 +1,192 @@
...
+    auto solver = solver_gen->generate(bA);
+
+    // Finally, solve the system. The solver, being a gko::LinOp, can be applied
+    // to a right hand side, b to
+    // obtain the solution, x.
+    solver->apply(lend(bb), lend(bx));
+
+    // Print the solution to the command line.
+    // std::cout << "Solution (x):\n";
+    // write(std::cout, lend(x));
```
Slaedr, line 171: We should probably demonstrate how to write the batch solution too.

examples/batched-solver/batched-solver.cpp
```diff
@@ -0,0 +1,192 @@
...
+    // @sect3{Reading your data and transfer to the proper device.}
+    // Read the matrix, right hand side and the initial solution using the @ref
+    // read function.
+    // @note Ginkgo uses C++ smart pointers to automatically manage memory. To
+    // this end, we use our own object ownership transfer functions that under
+    // the hood call the required smart pointer functions to manage object
+    // ownership. The gko::share , gko::give and gko::lend are the functions
+    // that you would need to use.
+    auto A = share(gko::read<mtx>(std::ifstream("data/A.mtx"), exec));
+    auto b = gko::read<vec>(std::ifstream("data/b.mtx"), exec);
```
Slaedr, line 124: I think for the RHS and initial guesses, since they're trivial, we should remove the mtx files and just generate them in the example. Reading the matrix is sufficient demonstration of how to read from file, I think.

examples/batched-solver/data/b.mtx
```diff
@@ -0,0 +1,21 @@
+%%MatrixMarket matrix array real general
```
Slaedr, line 1: I think we should remove this file and the one for x0.

Slaedr (commented):
Nice work, thanks! I think what remains is to
- [x] Add the batch logger and its output to the JSON file
- [x] Add the option to explicitly scale the matrix within the solver

in the batch-solver benchmark.

Apart from that, I have some comments, and it would be nice to update the example documentation.

benchmark/run_all_benchmarks.sh
```diff
@@ -509,14 +543,25 @@ EOT
...
+            echo -e "${PREFIX}Running SpMV for ${GROUP}/${NAME}" 1>&2
+            run_spmv_benchmarks "${RESULT_FILE}"
+            echo -e "${PREFIX}Running solvers for ${GROUP}/${NAME}" 1>&2
+            run_solver_benchmarks "${RESULT_FILE}"
+        fi
+
+        if [ "${BENCHMARK}" == "batch_solver" ]; then
+            echo -e "${PREFIX}Running Batch SpMV for ${GROUP}/${NAME}" 1>&2
+            NUM_BATCH_ENTRIES=${NUM_BATCH_DUP}
+            run_batch_spmv_benchmarks "${RESULT_FILE}"
```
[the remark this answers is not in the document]
pratikvn: For a solver benchmark, I think it makes sense to run the SpMV to just to have some info about the cost of an SpMV. The cost of the SpMV benchmark is minimal anyway compared to the solver benchmark.

benchmark/solver/batch_solver.cpp
```diff
@@ -269,14 +268,21 @@ void solve_system(const std::string &sol_name,
...
+        size_type nbatch = system_matrix->get_num_batch_entries();
+        if (FLAGS_detailed && b->get_size().at(0)[1] == 1 && !FLAGS_overhead) {
+            add_or_set_member(solver_json, "rhs_norm",
+                              rapidjson::Value(rapidjson::kObjectType),
+                              allocator);
+            for (size_type i = 0; i < nbatch; ++i) {
+                add_or_set_member(
+                    solver_json["rhs_norm"], std::to_string(i).c_str(),
+                    rapidjson::Value(rapidjson::kArrayType), allocator);
+                auto unbatch_b = b->unbatch();
```
Slaedr, line 315: You could move this line outside the loop.

benchmark/solver/batch_solver.cpp
```diff
@@ -269,14 +268,21 @@ void solve_system(const std::string &sol_name,
...
-        add_or_set_member(solver_json, "recurrent_residuals",
-                          rapidjson::Value(rapidjson::kArrayType), allocator);
-        add_or_set_member(solver_json, "true_residuals",
-                          rapidjson::Value(rapidjson::kArrayType), allocator);
-        add_or_set_member(solver_json, "implicit_residuals",
-                          rapidjson::Value(rapidjson::kArrayType), allocator);
-        add_or_set_member(solver_json, "iteration_timestamps",
-                          rapidjson::Value(rapidjson::kArrayType), allocator);
+        size_type nbatch = system_matrix->get_num_batch_entries();
+        if (FLAGS_detailed && b->get_size().at(0)[1] == 1 && !FLAGS_overhead) {
```
Slaedr, line 307: Are we limited to one RHS because it's harder to have arrays of arrays in rapidJSON? I think that's fine for now.

benchmark/solver/batch_solver.cpp
```diff
@@ -286,18 +292,35 @@ void solve_system(const std::string &sol_name,
...
-            solver.get()->apply(lend(b), lend(x));
+            solver->add_logger(logger);
+            solver->apply(lend(b), lend(x));
+            solver->remove_logger(gko::lend(logger));
             exec->synchronize();
         }
-        // if (FLAGS_warmup > 0) {
-        //     it_logger->write_data(solver_json["apply"], allocator);
-        // }
+        size_type nrhs = b->get_size().at(0)[1];
```
Slaedr, line 348:
```suggestion
        const size_type nrhs = b->get_size().at(0)[1];
```

Slaedr (approved):
Great job! Thanks for additionally extracting the implicit residuals, and the example looks good.

Just one or two minor points.

benchmark/solver/batch_solver.cpp
```diff
@@ -0,0 +1,646 @@
...
+                              allocator);
+        }
+
+        std::shared_ptr<gko::log::BatchConvergence<etype>> logger =
+            gko::log::BatchConvergence<etype>::create(exec);
+        // warm run
+        for (unsigned int i = 0; i < FLAGS_warmup; i++) {
+            auto x_clone = clone(x);
+            auto solver =
+                generate_solver(exec, sol_name)->generate(system_matrix);
```
Slaedr, line 336: This is a problem, I think. You need to create a copy of the system matrix here, because with scaling, it will be modified by the solver. Then all the further solves will be wrong.

benchmark/solver/batch_solver.cpp
```diff
@@ -0,0 +1,646 @@
...
+        // detail run
+        if (FLAGS_detailed && !FLAGS_overhead) {
+            // slow run, get the time of each functions
+            auto x_clone = clone(x);
+
+            auto gen_logger =
+                std::make_shared<OperationLogger>(exec, FLAGS_nested_names);
+            exec->add_logger(gen_logger);
+            auto solver =
+                generate_solver(exec, sol_name)->generate(system_matrix);
```
Slaedr, line 376: Same here. We should pass a deep copy of `system_matrix`.

benchmark/solver/batch_solver.cpp
```diff
@@ -0,0 +1,646 @@
...
+        // timed run
+        auto generate_timer = get_timer(exec, FLAGS_gpu_timer);
+        auto apply_timer = get_timer(exec, FLAGS_gpu_timer);
+        for (unsigned int i = 0; i < FLAGS_repetitions; i++) {
+            auto x_clone = clone(x);
+
+            exec->synchronize();
+            generate_timer->tic();
+            auto solver =
+                generate_solver(exec, sol_name)->generate(system_matrix);
```
Slaedr, line 435: ...and here.

Slaedr (changes requested):
Sorry, it turned out I was not reviewing the latest version earlier. I think there's an issue with how the scaling is being done - please see below. Apart from that, it seems good to go.

benchmark/solver/batch_solver.cpp
pratikvn: Good catch. Should be fixed now.

benchmark/solver/batch_solver.cpp
pratikvn: Yes, I think one rhs is enough, otherwise we have unnecessary lengthy arrays or wider arrays. It is not too difficult to add if we later decide we want it. It would just need another loop over the number of rhs.

Slaedr (approved):
LGTM!

Issue #748, opened 2021-04-21
Consistent treatment of strides in `Dense`


upsj:
I believe our treatment of strides in `Dense` vectors could be improved. Sometimes we ignore strides and copy padding data directly, sometimes we overwrite existing values, causing potential reallocations. And sometimes, it looks like we could save on some storage. I see two important areas where this pops up:

### Matrix conversions
`EnablePolymorphicAssignment` implements `convert_to` via the assignment operator. This means that if you copy from a Dense matrix to another matrix, this operation will overwrite the stride and potentially reallocate the target, copying all padding data over. I believe we could improve our memory footprint if we reused the existing storage and didn't modify the target stride. At least, we shouldn't be touching padding data, neither reading nor writing.
The same is also true for value type conversions and (potentially) other places I can't think of right now.

### Solvers
In solvers, we use `Dense::create_with_config_of` to create intermediate vectors of the same size and stride. This means that if our input vector has a huge stride, we will use this value for all internal vectors, potentially creating a large amount of unnecessary padding. I guess the idea behind this is that we only need to pass a single stride value to the kernels, instead of having a distinct stride for each vector? Do you think we can find a more memory-efficient solution for the worst case?

PR #772, opened 2021-05-22
Add benchmark for BLAS operations

upsj:
This PR adds benchmarks for Ginkgo's BLAS-like operations (copy, axpy, scale, dot, norm, gemm) ~~as well as a Dense::convert_to(Dense) implementation that doesn't modify the stride of the result vector if unnecessary (thus partially addressing #748).~~ This benchmark can be used later to check whether #733 provides decent implementations for Dense operations.

The BLAS benchmarks are not the most sophisticated thing out there, since they don't flush the cache in any way, or check for correctness/precision (though the current design allows for that), but that should not matter too much as soon as we scale up.

A benchmark test case consists entirely of problem dimensions and strides.
We have input vectors `x, y: nxr, a, b: 1xr or 1x1, A: nxk, B: kxm, C:nxm`, for which we can specify individual strides `stride_x/y/A/B/C` with `stride` being a shorthand for `stride_x/y`.

The benchmark uses the amount of FLOPs and memory movement to compute FLOP/s and B/s numbers in addition to the runtime. Please double-check those (found in `get_memory()` and `get_flops()`, I just claimed that `copy` does one FLOP per entry), though.

Example output on my GPU:
```
[
    {
        "n": 1000000,
        "r": 10,
        "stride_x": 20,
        "stride_y": 15,
        "blas": {
            "copy": {
                "time": 0.019050621900000003,
                "flops": 524917246.92725118,
                "bandwidth": 8398675950.836019,
                "completed": true
            },
            "axpy": {
                "time": 0.0243559142,
                "flops": 821155791.3929588,
                "bandwidth": 9853869496.715506,
                "completed": true
            },
            "scal": {
                "time": 0.018665861,
                "flops": 535737408.5235072,
                "bandwidth": 8571798536.376115,
                "completed": true
            }
        }
    }
]
```

### DIFF
--- a/benchmark/blas/blas.cpp
+++ b/benchmark/blas/blas.cpp
@@ -69,9 +69,6 @@ class BenchmarkOperation {
 public:
     virtual ~BenchmarkOperation() = default;
 
-    virtual void write_error(rapidjson::Value &blas_case,
-                             rapidjson::MemoryPoolAllocator<> &allocator)
-    {}
     virtual gko::size_type get_flops() const = 0;
     virtual gko::size_type get_memory() const = 0;
     virtual void prepare(){};
@@ -442,6 +439,7 @@ void apply_blas(const char *operation_name, std::shared_ptr<gko::Executor> exec,
 
         // warm run
         for (unsigned int i = 0; i < FLAGS_warmup; i++) {
+            op->prepare();
             exec->synchronize();
             op->run();
             exec->synchronize();
@@ -450,6 +448,7 @@ void apply_blas(const char *operation_name, std::shared_ptr<gko::Executor> exec,
         // timed run
         auto timer = get_timer(exec, FLAGS_gpu_timer);
         for (unsigned int i = 0; i < FLAGS_repetitions; i++) {
+            op->prepare();
             exec->synchronize();
             timer->tic();
             op->run();
@@ -481,7 +480,17 @@ int main(int argc, char *argv[])
 {
     std::string header =
         "A benchmark for measuring performance of Ginkgo's BLAS-like "
-        "operations.\n";
+        "operations.\nParameters for a benchmark case are:\n"
+        "    n: number of rows for vectors and gemm output (required)\n"
+        "    r: number of columns for vectors (optional, default 1)\n"
+        "    m: number of columns for gemm output (optional, default n)\n"
+        "    k: inner dimension of the gemm (optional, default n)\n"
+        "    stride: storage stride for both vectors (optional, default r)\n"
+        "    stride_x: stride for input vector x (optional, default r)\n"
+        "    stride_y: stride for in/out vector y (optional, default r)\n"
+        "    stride_A: stride for A matrix in gemm (optional, default k)\n"
+        "    stride_B: stride for B matrix in gemm (optional, default m)\n"
+        "    stride_C: stride for C matrix in gemm (optional, default m)\n";
     std::string format = std::string() + "  [\n    { \"n\": 100 },\n" +
                          "    { \"n\": 200, \"m\": 200, \"k\": 200 }\n" +
                          "  ]\n\n";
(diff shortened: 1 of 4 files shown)


cuda/test/matrix/dense_kernels.cpp
```diff
@@ -199,7 +216,7 @@ TEST_F(Dense, CudaFillIsEquivalentToRef)
     dx->fill(42);
     result->copy_from(dx.get());
 
-    GKO_ASSERT_MTX_NEAR(result, x, 1e-14);
+    GKO_ASSERT_MTX_NEAR(result, x, 0);
```
upsj, line 219: Should probably change that for the other executors as well

benchmark/blas/blas.cpp
```diff
@@ -0,0 +1,540 @@
...
+    try {
+        auto &blas_case = test_case["blas"];
+        add_or_set_member(blas_case, operation_name,
+                          rapidjson::Value(rapidjson::kObjectType), allocator);
+
+        auto op = operation_map[operation_name](exec, parse_dims(test_case));
+
+        // warm run
+        for (unsigned int i = 0; i < FLAGS_warmup; i++) {
+            exec->synchronize();
```
yhmtsai, line 445:
```suggestion
            op->prepare();
            exec->synchronize();
```

benchmark/blas/blas.cpp
```diff
@@ -0,0 +1,540 @@
...
+            op->run();
+            exec->synchronize();
+        }
+
+        // timed run
+        auto timer = get_timer(exec, FLAGS_gpu_timer);
+        for (unsigned int i = 0; i < FLAGS_repetitions; i++) {
+            exec->synchronize();
```
yhmtsai, line 453:
```suggestion
            op->prepare()
            exec->synchronize();
```

core/matrix/dense.cpp
```diff
@@ -327,6 +328,34 @@ void Dense<ValueType>::compute_norm2_impl(LinOp *result) const
 }
 
 
+template <typename ValueType>
+void Dense<ValueType>::convert_to(Dense<ValueType> *result) const
+{
+    if (result->get_size() && result->get_size() == this->get_size()) {
```
yhmtsai, line 334: Did we face the issue somewhere?

benchmark/blas/blas.cpp
```diff
@@ -0,0 +1,540 @@
...
+        A_->fill(1);
+        B_->fill(1);
+        alpha_->fill(1);
+        beta_->fill(1);
+    }
+
+    gko::size_type get_flops() const override
+    {
+        return A_->get_size()[0] * A_->get_size()[1] * B_->get_size()[1] * 2 +
+               C_->get_size()[0] * C_->get_size()[1] * 3;
```
yhmtsai, line 310: this is 2 multiplication + 1 addition from alpha*(AB)_ij + beta * C_ij, right?

benchmark/blas/blas.cpp
```diff
@@ -0,0 +1,540 @@
...
+                  << "what(): " << e.what() << std::endl;
+    }
+}
+
+
+int main(int argc, char *argv[])
+{
+    std::string header =
+        "A benchmark for measuring performance of Ginkgo's BLAS-like "
+        "operations.\n";
```
yhmtsai, line 484: Could you write what arguments should be in json? I think it is fine to combine this with the doc from gflags.

yhmtsai (approved):
LGTM

core/matrix/dense.cpp
upsj: I decided to move this to a new PR: #774

benchmark/blas/blas.cpp
upsj: exactly

tcojean (approved):
LGTM.

I guess you don't want to advertise it which is why you didn't add it to neither `run_all_benchmarks.sh` or `BENCHMARKING.md`?

upsj:
@tcojean Yes, I would like to finish some interface extensions (non-inplace or non-object-creating operations) and implementation changes (convert_to and move_to, mostly) first, so we can also represent conversions, `transpose`, `permute` etc. correctly.

PR #773, opened 2021-05-23
Bugfixes for reference batch functionality and improvements to testing

Slaedr:
This PR 
- [x] fixes an issue where reference batch Jacobi was reporting the wrong amount of memory it needs for the work vector
- [x] improves the integration test in `reference/test/solvers/batch_richardson_kernels` to also demonstrate the use of the batch convergence logger
- [x] Fixed batch logger has a corner case bug by resetting the stored convergence state everytime a new solve begins. The beginning of a new solve is equated with the iteration count being zero.
- [x] Simplified Cuda batch Jacobi implementation.


pratikvn (approved):
LGTM!

PR #778, opened 2021-05-31
Add Batch Solvers: CG, BiCGSTAB, GMRES, IDR

aggarwal2000:
This PR adds:

### A.
-OpenMp kernels for batch dense and batch csr matrices

-Some additions to/changes in  batch dense and batch csr Cuda kernels

### B.
-Batch identity preconditioner to allow for unpreconditioned solves

### C.
-A new stopping criterion for batch solvers which checks convergence on the basis of either relative or absolute tolerance and the maximum number of iterations allowed; the type of tolerance to be considered could be set as a solver factory parameter.

### D.
-Batch CG solver

-Batch BiCGSTAB solver

-Batch GMRES solver

-Batch IDR solver

### TODO:
Have an option to allocate large intermediate vectors and matrices in global memory for solvers such as batch GMRES and batch IDR when the shared memory limits are exceeded.


core/solver/batch_cg_kernels.hpp
```diff
@@ -0,0 +1,161 @@
...
+ * Options controlling the batch Cg solver.
+ */
+template <typename RealType>
+struct BatchCgOptions {
+    preconditioner::batch::type preconditioner;
+    int max_its;
+    RealType rel_residual_tol;
+    RealType abs_residual_tol;
+    ::gko::stop::batch::ToleranceType tol_type;
+};
```
pratikvn, line 59: Maybe we can unify this for all the batch solvers ?

core/solver/batch_gmres.cpp
```diff
@@ -0,0 +1,143 @@
...
+    logdata.res_norms =
+        matrix::BatchDense<real_type>::create(this->get_executor(), sizes);
+    logdata.iter_counts.set_executor(this->get_executor());
+    logdata.iter_counts.resize_and_reset(num_rhs * num_batches);
+
+    exec->run(batch_gmres::make_apply(
+        opts, system_matrix_.get(), this->get_left_scaling_vector(),
+        this->get_right_scaling_vector(), dense_b, dense_x, logdata));
+
+    this->template log<log::Logger::batch_solver_completed>(
```
pratikvn, line 117: Shouldn't there also be a `batch_solver_started` log ?

core/solver/batch_gmres_kernels.hpp
```diff
@@ -0,0 +1,167 @@
...
+
+#include <ginkgo/core/preconditioner/batch_preconditioner_types.hpp>
+#include <ginkgo/core/stop/batch_stop_enum.hpp>
+#include "core/log/batch_logging.hpp"
+
+namespace gko {
+namespace kernels {
+namespace batch_gmres {
+
+constexpr int max_restart = 5;
```
pratikvn, line 48: Is this bound because of the amount of shared memory ? Can we maybe set this to 10 or 20 ?

core/solver/batch_bicgstab.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+
+template <typename ValueType>
+std::unique_ptr<BatchLinOp> BatchBicgstab<ValueType>::transpose() const
+{
+    return build()
+        .with_preconditioner(parameters_.preconditioner)
+        .with_max_iterations(parameters_.max_iterations)
+        .with_rel_residual_tol(parameters_.rel_residual_tol)
+        .with_abs_residual_tol(parameters_.abs_residual_tol)
```
pratikvn, line 60: I dont think it is possible to use both at the same time, so maybe just one parameter called `tolerance` should be sufficient, I think . Then you can decide whether to use `tolerance` for absolute or relative based on the `tolerance_type` parameter.

pratikvn (commented):
Excellent work! Most of the framework looks good to me. Some minor comments on the main classes and core. I still need to review the kernels and their tests.

omp/test/matrix/batch_dense_kernels.cpp
```diff
@@ -186,116 +152,96 @@ class BatchDense : public ::testing::Test {
     std::unique_ptr<Mtx> dalpha;
     std::unique_ptr<Mtx> dbeta;
     std::unique_ptr<Mtx> dsquare;
-    std::unique_ptr<Arr> rpermute_idxs;
-    std::unique_ptr<Arr> cpermute_idxs;
-    std::unique_ptr<Arr> rgather_idxs;
 };
 
 
 TEST_F(BatchDense, SingleVectorOmpScaleIsEquivalentToRef)
```
Slaedr, line 158:
```suggestion
TEST_F(BatchDense, SingleVectorScaleIsEquivalentToRef)
```

omp/test/matrix/batch_dense_kernels.cpp
```diff
@@ -186,116 +152,96 @@ class BatchDense : public ::testing::Test {
...
+    set_up_vector_data(20);
+
+    x->scale(alpha.get());
+    dx->scale(dalpha.get());
+
+    GKO_ASSERT_BATCH_MTX_NEAR(dx, x, 1e-14);
+}
 
 
 TEST_F(BatchDense, MultipleVectorOmpScaleWithDifferentAlphaIsEquivalentToRef)
```
Slaedr, line 182:
```suggestion
TEST_F(BatchDense, MultipleVectorScaleWithDifferentAlphaIsEquivalentToRef)
```

omp/test/matrix/batch_dense_kernels.cpp
```diff
@@ -306,737 +252,177 @@ GKO_NOT_IMPLEMENTED;
...
-//    rmtx->move_to(srmtx.get());
-//    omtx->move_to(somtx.get());
-//    srmtx->move_to(drmtx.get());
-//    somtx->move_to(domtx.get());
-//
-//    GKO_ASSERT_MTX_NEAR(drmtx, domtx, 1e-14);
-//    GKO_ASSERT_MTX_NEAR(srmtx, somtx, 1e-14);
-//    GKO_ASSERT_MTX_NEAR(domtx, omtx, 1e-14);
-//}
+TEST_F(BatchDense, OmpComputeNorm2IsEquivalentToRef)
```
Slaedr, line 276:
```suggestion
TEST_F(BatchDense, ComputeNorm2IsEquivalentToRef)
```

omp/test/preconditioner/batch_identity_kernels.cpp
```diff
@@ -0,0 +1,157 @@
...
+{
+    return {5.0, 1.5};
+}
+
+template <typename T>
+T get_num(T)
+{
+    return 5.0;
+}
+
```
Slaedr, line 64:
```suggestion
```

omp/test/preconditioner/batch_identity_kernels.cpp
```diff
@@ -0,0 +1,157 @@
...
+                if (col_idxs[iz] == irow) {
+                    for (size_t ibatch = 0; ibatch < nbatch; ibatch++) {
+                        // TODO: take care of any padding here
+                        const size_t valpos = iz + ibatch * nnz;
+                        vals[valpos] =
+                            get_num(T{}) + static_cast<T>(std::sin(irow));
+                    }
+                }
+            }
+        }
```
Slaedr, line 99:
```suggestion
```

omp/test/solver/batch_cg_kernels.cpp
```diff
@@ -0,0 +1,504 @@
...
+    auto left_scale = gko::batch_initialize<BDense>(
+        this->nbatch, {0.8, 0.9, 0.95}, this->exec);
+    auto right_scale = gko::batch_initialize<BDense>(
+        this->nbatch, {1.0, 1.5, 1.05}, this->exec);
+
+    Result result =
+        this->solve_poisson_uniform_1(left_scale.get(), right_scale.get());
+
+    GKO_ASSERT_BATCH_MTX_NEAR(result.x, this->sys_1.xex,
+                              1e-3 /*r<value_type>::value*/);
```
Slaedr, line 415: If this is the best tolerance we can use, it's concerning.

omp/test/solver/batch_gmres_kernels.cpp
```diff
@@ -0,0 +1,47 @@
+/*******************************<GINKGO LICENSE>******************************
```
Slaedr, line 1: Let's just remove this file for now.

omp/test/solver/batch_idr_kernels.cpp
```diff
@@ -0,0 +1,47 @@
+/*******************************<GINKGO LICENSE>******************************
```
Slaedr, line 1: Same here- we can remove this file.

reference/preconditioner/batch_identity.hpp
```diff
@@ -0,0 +1,94 @@
...
+ */
+template <typename ValueType>
+class BatchIdentity final {
+public:
+    using value_type = ValueType;
+
+    /**
+     * The size of the work vector required in case of static allocation.
+     */
+    static constexpr int work_size = 1;
```
Slaedr, line 59:
```suggestion
    static constexpr int work_size = 0;
```

reference/preconditioner/batch_identity.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * The size of the work vector required in case of dynamic allocation.
+     */
+    static int dynamic_work_size(const int nrows, int) { return 1; }
```
Slaedr, line 64:
```suggestion
    static int dynamic_work_size(int, int) { return 0; }
```

reference/test/preconditioner/batch_identity_kernels.cpp
```diff
@@ -0,0 +1,131 @@
...
+        value_type *const vals = mat->get_values();
+        // clang-format off
+		row_ptrs[0] = 0; row_ptrs[1] = 2; row_ptrs[2] = 4; row_ptrs[3] = 6;
+		col_idxs[0] = 0; col_idxs[1] = 1; col_idxs[2] = 0; col_idxs[3] = 1;
+		col_idxs[4] = 0; col_idxs[5] = 2;
+		vals[0] = 2.0; vals[1] = 0.25; vals[2] = -1.0; vals[3] = -3.0;
+		vals[4] = 2.0; vals[5] = 0.2;
+		vals[6] = -1.5; vals[7] = 0.55; vals[8] = -1.0; vals[9] = 4.0;
+		vals[10] = 2.0; vals[11] = -0.25;
+        // clang-format on
```
Slaedr, line 83: I guess you don't actually need any of this, because batch identity should not care about the matrix itself, apart from its size.
```suggestion
```

reference/test/stop/batch_additional_criteria.cpp
```diff
@@ -0,0 +1,343 @@
...
+    using BatchStop = typename TestFixture::BatchStop;
+    const int maxits = 10;
+    const int iter = 5;
+    uint32_t converged;
+    std::vector<real_type> resnv(this->nrhs);
+    for (int i = 0; i < this->nrhs; i++) {
+        resnv[i] = 6.0;
+    }
+    const int conv_col = 2;
+    resnv[conv_col] = (1 / 2) * this->abs_tol;
```
Slaedr, line 126: This would simply be zero, but it's still fine, haha.

omp/stop/batch_criteria.hpp
```diff
@@ -0,0 +1,197 @@
...
+                             bitset_type &converge_bitset,
+                             const real_type *const rhs_b_norms)
+        : nrhs{num_rhs},
+          rel_tol{rel_res_tol},
+          abs_tol{abs_res_tol},
+          tol_type{type_of_tol},
+          max_its{max_iters},
+          rhs_norms{rhs_b_norms}
+    {
+        if (nrhs > 32) {
```
Slaedr, line 115:
```suggestion
        if (nrhs > max_nrhs) {
```

omp/stop/batch_criteria.hpp
```diff
@@ -0,0 +1,197 @@
...
+        bitset_type &converged) const
+    {
+        if (iter >= max_its - 1) {
+            return true;
+        }
+
+        if (residual_norms) {
+            check_norms(residual_norms, converged);
+        } else {
+            real_type norms[32];
```
Slaedr, line 145:
```suggestion
            real_type norms[batch_config<ValueType>::max_nrhs];
```

omp/stop/batch_criteria.hpp
```diff
@@ -0,0 +1,197 @@
...
+            batch_dense::compute_norm2<ValueType>(residual,
+                                                  {norms, 32, 1, nrhs});
```
Slaedr, line 147:
```suggestion
                                                  {norms, static_cast<size_type>(nrhs), 1, nrhs});
```

omp/stop/batch_criteria.hpp
```diff
@@ -0,0 +1,197 @@
...
+
+
+private:
+    int nrhs;
+    int max_its;
+    const real_type rel_tol;
+    const real_type abs_tol;
+    const tolerance tol_type;
+    const real_type *const rhs_norms;
+    static constexpr uint32 all_true = detail::power(2, 32) - 1;
```
Slaedr, line 166:
```suggestion
    static constexpr bitset_type all_true = ~static_cast<bitset_type>(0);
```

omp/stop/batch_criteria.hpp
```diff
@@ -0,0 +1,197 @@
...
+constexpr uint32 power(const uint32 x, const int N)
+{
+    uint32 ans = 1;
+    for (int i = 0; i < N; i++) {
+        ans *= x;
+    }
+    return ans;
+}
+
+
```
Slaedr, line 63:
```suggestion
```

Slaedr (changes requested):
Excellent work! Looks good in general. Thanks for upgrading some of the pre-existing kernels as well. Here's the first part of my review - I have not yet looked at the actual device-level kernels for the Krylov solvers.

One potentially serious issue I see is that scaling appears to change the solution for some Krylov solvers. Could you rework the tests maybe to check that if the linear system is solved to a good enough tolerance, the solution stays very close, at least 1e-6 for double? I just realized that I had also been somewhat sloppy with the same test for batch Richardson. We should try to compare the result to some epsilon-dependent tolerance.

In general, a few other points:
- For solver settings such as the restart count, there should only be one variable in the class that stores it. Since there is already `parameters_.restart` doing exactly that (in this example), we should avoid having another class member for storing a copy of the same thing.
- You could remove all unnecessary newlines.
- For any commented line, remove it unless you think it will certainly be useful in future.

Later after benchmarking, we can remove all the `GKO_CUDA_BATCH_USE_DYNAMIC_SHARED_MEM`.

Other notes for later:
- For all the batch solvers, 
  1. all the `core/solver/batch_<solver>.cpp`s look almost the same. We can probably move the `apply_impl` to a base class, most likely `EnableBatchScaledSolver`;
  2. similarly, the .cu files for all the solvers look essentially the same;
  3. finally, a lot of the testing code can be extracted into a separate header file and called for each of the solver tests. Let's do all that in a separate PR.
- Use `syn` to dispatch to the correct preconditioner; I'll do this later in a separate PR.

core/solver/batch_gmres.cpp
Slaedr: Right now, there's actually no such log function, because it only logs the iteration counts and residual norms. For the timings, I guess it's the BatchLinOp `apply` logging that's being used?

core/solver/batch_cg_kernels.hpp
Slaedr: Actually, I'm not completely sure of that. We'd have unused parameters lying around for many solvers. Especially IDR has parameters that no one else needs. Maybe it's fine to have separate structs.

pratikvn:
format!

common/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -0,0 +1,847 @@
...
+    const gko::batch_dense::BatchEntry<typename gko::remove_complex<ValueType>>
+        &res_norms_shared_entry)
+{
+    for (int c = threadIdx.x; c < rho_old_shared_entry.num_rhs;
+         c += blockDim.x) {
+        rho_old_shared_entry.values[0 * rho_old_shared_entry.stride + c] =
+            one<ValueType>();
+        omega_old_shared_entry.values[0 * omega_old_shared_entry.stride + c] =
+            one<ValueType>();
+        alpha_shared_entry.values[0 * alpha_shared_entry.stride + c] =
```
pratikvn, line 85: I guess for readability you added the mutliplication by 0, but I think you can remove it.

common/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -0,0 +1,847 @@
...
+{
+    for (int c = threadIdx.x; c < beta_shared_entry.num_rhs; c += blockDim.x) {
+        const uint32 conv = converged & (1 << c);
+
+        if (conv) {
+            continue;
+        }
+
+        beta_shared_entry.values[0 * beta_shared_entry.stride + c] =
+            (rho_new_shared_entry.values[0 * rho_new_shared_entry.stride + c] /
```
pratikvn, line 143: Same here as well

common/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -0,0 +1,847 @@
...
+
+        const gko::batch_dense::BatchEntry<real_type> rhs_norms_shared_entry{
+            norms_rhs_sh, static_cast<size_type>(nrhs), 1, nrhs};
+
+        const gko::batch_dense::BatchEntry<real_type> res_norms_shared_entry{
+            norms_res_sh, static_cast<size_type>(nrhs), 1, nrhs};
+
+        const gko::batch_dense::BatchEntry<real_type>
+            res_norms_temp_shared_entry{norms_res_temp_sh,
+                                        static_cast<size_type>(nrhs), 1, nrhs};
```
pratikvn, line 636: Can we abstract these initializations away in a function ?

common/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -0,0 +1,847 @@
...
+                     gko::batch::to_const(omega_new_shared_entry),
+                     gko::batch::to_const(t_shared_entry), r_shared_entry,
+                     converged);
+
+            __syncthreads();
+
+            compute_norm2<ValueType>(
+                gko::batch::to_const(r_shared_entry),
+                res_norms_temp_shared_entry);  // residual norms
+            __syncthreads();
```
pratikvn, line 819: This is only needed for the logging, right ? Can we make this optional ? This would reduce the overhead a bit, as you have a norm computation in every iteration, which seems unnecessary.

common/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -0,0 +1,847 @@
...
+
+            __syncthreads();
+
+
+            // p = r + beta*(p - omega_old * v)
+            update_p(gko::batch::to_const(r_shared_entry),
+                     gko::batch::to_const(beta_shared_entry),
+                     gko::batch::to_const(omega_old_shared_entry),
+                     gko::batch::to_const(v_shared_entry), p_shared_entry,
+                     converged);
```
pratikvn, line 708: `beta` unlike `alpha` is used only once. Do we need to store it separately ? Or can it just be computed implicitly in the `update_p` function ?

common/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -0,0 +1,847 @@
...
+                     gko::batch::to_const(s_hat_shared_entry),
+                     gko::batch::to_const(alpha_shared_entry),
+                     gko::batch::to_const(omega_new_shared_entry), converged);
+
+
+            // r = s - omega_new*t
+            update_r(gko::batch::to_const(s_shared_entry),
+                     gko::batch::to_const(omega_new_shared_entry),
+                     gko::batch::to_const(t_shared_entry), r_shared_entry,
+                     converged);
```
pratikvn, line 812: Here as well I think you can combine these three functions, compute `omega_new` implicitly, and store it in `omega_old` and only update entries of `r` and `x` in one loop.

pratikvn (commented):
Some comments on the bicgstab kernel.

include/ginkgo/core/solver/batch_gmres.hpp
```diff
@@ -0,0 +1,189 @@
...
+
+        /**
+         * Absolute residual tolerance.
+         */
+        real_type GKO_FACTORY_PARAMETER_SCALAR(abs_residual_tol, 1e-11);
+
+        /**
+         * Restart parameter for Gmres
+         */
+        int GKO_FACTORY_PARAMETER_SCALAR(restart, 2);
```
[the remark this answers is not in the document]
aggarwal2000: Most of the systems are around 50, so I have set the default to 10 now.

include/ginkgo/core/solver/batch_idr.hpp
```diff
@@ -0,0 +1,297 @@
...
+         * To enable smoothing
+         */
+        bool GKO_FACTORY_PARAMETER_SCALAR(smoothing, true);
+
+
+        /**
+        * If set to true, the vectors spanning the subspace S are chosen
+        * deterministically. This is mostly needed for testing purposes.
+        *
+        * Note: If 'deterministic' is set to true, the subspace vectors are
```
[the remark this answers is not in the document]
aggarwal2000: Completed this doc string!

omp/matrix/batch_csr_kernels.cpp
```diff
@@ -68,7 +69,47 @@ template <typename ValueType, typename IndexType>
...
+    //                  k < static_cast<size_type>(row_ptrs[row + 1]); ++k) {
+    //                 auto val = vals[offset + k];
+    //                 auto col = col_idxs[k];
+    //                 for (size_type j = 0; j < c->get_size().at(batch)[1];
+    //                 ++j) {
+    //                     c->at(batch, row, j) += val * b->at(batch, col, j);
+    //                 }
+    //             }
+    //         }
+    //     }
```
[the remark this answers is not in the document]
aggarwal2000: Done!

core/solver/batch_gmres_kernels.hpp
aggarwal2000: I have set the max to 20 now, considering we would be using dynamic shared memory.

common/matrix/batch_csr_kernels.hpp.inc
```diff
@@ -42,21 +42,51 @@ __device__ __forceinline__ void spmv_kernel(
...
-        for (auto k = a.row_ptrs[row]; k < a.row_ptrs[row + 1]; ++k) {
-            auto val = a.values[k];
-            auto col = a.col_idxs[k];
-            for (int j = 0; j < b.num_rhs; ++j) {
-                c.values[row * c.stride + j] +=
-                    val * b.values[col * b.stride + j];
-            }
+
+        // subwarp_grp level reduction
+        for (int i = subwarp_grp.size() / 2; i > 0; i /= 2) {
```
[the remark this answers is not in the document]
aggarwal2000: Yes, but I have changed row_end - row_start to a number rounded off to closest power of 2, to avoid missing few elements during reduction, when i is getting halved every iteration. Would that be okay?

common/matrix/batch_csr_kernels.hpp.inc
```diff
@@ -84,21 +114,52 @@ __device__ __forceinline__ void advanced_spmv_kernel(
...
-    for (int row = threadIdx.x; row < a.num_rows; row += blockDim.x) {
-        for (int j = 0; j < c.num_rhs; ++j) {
-            c.values[row * c.stride + j] *= beta;
+    constexpr auto tile_size = config::warp_size;
+
+    auto thread_block = group::this_thread_block();
+
+    auto subwarp_grp = group::tiled_partition<tile_size>(thread_block);
+
+    const int subwarp_grp_id = floor(static_cast<double>(threadIdx.x) /
```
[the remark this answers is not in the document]
aggarwal2000: Changed as suggested.

cuda/base/config.hpp
```diff
@@ -79,13 +79,16 @@ struct batch_config {
...
-    static constexpr int max_num_rows = 128;
+    static constexpr int max_num_rows = 150;
 
     /**
      * Max number of RHS vectors in a linear system. This can be at most
      * `config::warp_size`.
      */
-    static constexpr int max_num_rhs = 5;
+    static constexpr int max_num_rhs =
+        6;  // NOTE: max_num_rhs has to be an even number to avoid cuda
```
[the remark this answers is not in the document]
aggarwal2000: Done

cuda/test/solver/batch_bicgstab_kernels.cpp
```diff
@@ -0,0 +1,515 @@
...
+        }
+    }
+
+    std::shared_ptr<gko::ReferenceExecutor> exec;
+    std::shared_ptr<const gko::CudaExecutor> cuexec;
+
+    const size_t nbatch = 2;
+    const int nrows = 3;
+    const Options opts_1{gko::preconditioner::batch::type::none, 10, 1e-6,
+                         1e-11, gko::stop::batch::ToleranceType::absolute};
```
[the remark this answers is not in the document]
aggarwal2000: Now every solver tests for both absolute and relative tolerance.

core/solver/batch_idr_kernels.hpp
```diff
@@ -0,0 +1,182 @@
...
+    RealType kappa_val;
+    bool to_use_smoothing;
+    bool deterministic_gen;
+    ::gko::stop::batch::ToleranceType tol_type;
+};
+
+/**
+ * Calculates the amount of in-solver storage needed by batch-Idr.
+ *
+ * The calculation includes multivectors for
```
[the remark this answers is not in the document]
aggarwal2000: Done!

cuda/solver/batch_idr_kernels.cu
```diff
@@ -0,0 +1,286 @@
...
+                Subspace_vectors_entry);
+    } else {
+        GKO_NOT_IMPLEMENTED;
+    }
+}
+
+
+template <typename ValueType, typename Distribution, typename Generator>
+typename std::enable_if<!is_complex_s<ValueType>::value, ValueType>::type
+get_rand_value(Distribution &&dist, Generator &&gen)
```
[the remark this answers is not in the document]
aggarwal2000: Now these two are in an unnamed namespace.

cuda/test/solver/batch_cg_kernels.cpp
```diff
@@ -0,0 +1,508 @@
...
+    auto left_scale = gko::batch_initialize<BDense>(
+        this->nbatch, {0.8, 0.9, 0.95}, this->exec);
+    auto right_scale = gko::batch_initialize<BDense>(
+        this->nbatch, {1.0, 1.5, 1.05}, this->exec);
+
+    Result result =
+        this->solve_poisson_uniform_1(left_scale.get(), right_scale.get());
+
+    GKO_ASSERT_BATCH_MTX_NEAR(result.x, this->sys_1.xex,
+                              1e-3 /*r<value_type>::value*/);
```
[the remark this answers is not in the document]
aggarwal2000: Changed as suggested.

hip/base/config.hip.hpp
```diff
@@ -89,6 +89,19 @@ struct config {
 };
 
 
+template <typename ValueType>
+struct batch_config {
+    /**
+     * Max number of rows per matrix in a batch of (small) sparse matrices
+     */
+    static constexpr int max_num_rows = 150;
+
```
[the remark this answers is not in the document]
aggarwal2000: May be we could do this after the benchmarking  results.

hip/matrix/batch_struct.hip.hpp
```diff
@@ -0,0 +1,140 @@
...
+HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+#ifndef GKO_HIP_MATRIX_BATCH_STRUCT_HPP_
```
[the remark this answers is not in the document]
aggarwal2000: I have removed this file now.

omp/test/solver/batch_gmres_kernels.cpp
aggarwal2000: Done

omp/test/solver/batch_idr_kernels.cpp
aggarwal2000: Removed.

include/ginkgo/core/solver/batch_idr.hpp
```diff
@@ -0,0 +1,297 @@
...
+     *
+     * @return true as iterative solvers use the data in x as an initial guess.
+     */
+    bool apply_uses_initial_guess() const override { return true; }
+
+    /**
+     * Gets the subspace dimension of the solver.
+     *
+     * @return the subspace Dimension*/
+    size_type get_subspace_dim() const { return subspace_dim_; }
```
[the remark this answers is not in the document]
aggarwal2000: I am using parameters_ struct to access these in getters /setters now.

cuda/test/solver/batch_gmres_kernels.cpp
```diff
@@ -0,0 +1,518 @@
...
+    auto left_scale = gko::batch_initialize<BDense>(
+        this->nbatch, {0.8, 0.9, 0.95}, this->exec);
+    auto right_scale = gko::batch_initialize<BDense>(
+        this->nbatch, {1.0, 1.5, 1.05}, this->exec);
+
+    Result result =
+        this->solve_poisson_uniform_1(left_scale.get(), right_scale.get());
+
+    GKO_ASSERT_BATCH_MTX_NEAR(result.x, this->sys_1.xex,
+                              1e-3 /*r<value_type>::value*/);
```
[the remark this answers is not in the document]
aggarwal2000: Checked that 1e2 * eps does work here!

omp/test/solver/batch_cg_kernels.cpp
aggarwal2000: I guess 1e3 * eps is the best tol, we could use here.

core/solver/batch_cg_kernels.hpp
aggarwal2000: Could we do something like- have a BatchSolverOptions struct which houses all common parameters, and then derive  Gmres and IDR Options struct from it and just add extra members while doing so?

core/solver/batch_bicgstab.cpp
aggarwal2000: May be we could improve upon this later ?

common/solver/batch_gmres_kernels.hpp.inc
```diff
@@ -0,0 +1,985 @@
...
+
+            ValueType h_k_i_scalar = h_k_i_shared_entry.values[c];
+
+            w_shared_entry.values[r * w_shared_entry.stride + c] -=
+                h_k_i_scalar *
+                v_k_shared_entry.values[r * v_k_shared_entry.stride + c];
+        }
+
+        __syncthreads();
+    }
```
pratikvn, line 305: I am not sure how feasible it is, but maybe we can implement some form of multi-dot. Maybe that gives us some advantage in terms of lesser syncs due to lesser reductions. But it is also something we can look into later.

pratikvn (commented):
Some comments on CG and part of GMRES.

pratikvn (commented):
Some comments on GMRES

reference/solver/batch_gmres_kernels.cpp
```diff
@@ -0,0 +1,853 @@
...
+                }
+            }
+
+            if (inner_loop_break_flag == true) {
+                break;
+            }
+
+            // y = Hessenburg(0:restart - 1,0:restart - 1) \ s(0:restart-1)
+            // x = x + KrylovBasis(:,0 : restart - 1) * y
+            update_x(x_entry, restart - 1, gko::batch::to_const(H_entry),
```
Slaedr, line 741: Here, it might be better to separate the triangular solve from the update of `x`. Right now, at the end of every restart, it looks like you're solving the triangular system again for no reason.

common/solver/batch_gmres_kernels.hpp.inc
```diff
@@ -0,0 +1,985 @@
...
+    const int num_rhs = w_shared_entry.num_rhs;
+
+    const gko::batch_dense::BatchEntry<const ValueType> v_i_shared_entry{
+        &V_shared_entry.values[i * V_shared_entry.stride * num_rows],
+        V_shared_entry.stride, num_rows, V_shared_entry.num_rhs};
+
+    __shared__
+        UninitializedArray<ValueType, batch_config<ValueType>::max_num_rows *
+                                          batch_config<ValueType>::max_num_rhs>
+            w_temp_sh;
```
Slaedr, line 253: I think you should allocate this in dynamic shared memory and pass it in. Something proportional to the number of rows might be too big to allocate statically without affecting occupancy.

common/solver/batch_gmres_kernels.hpp.inc
```diff
@@ -0,0 +1,985 @@
...
+            cs[c] = one<ValueType>();
+            sn[c] = zero<ValueType>();
+        } else if (abs(y[c]) > abs(x[c])) {
+            ValueType temp = x[c] / y[c];
+            sn[c] = one<ValueType>() / sqrt(one<ValueType>() + temp * temp);
+            cs[c] = temp * sn[c];
+        } else {
+            ValueType temp = y[c] / x[c];
+            cs[c] = one<ValueType>() / sqrt(one<ValueType>() + temp * temp);
+            sn[c] = temp * cs[c];
```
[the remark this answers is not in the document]
Slaedr: It would be nice if complex could be addressed, but I don't think it's urgent right now. We could come back to it later too. If that is the case, perhaps we should throw an exception in the apply_impl for complex instantiations.

common/solver/batch_gmres_kernels.hpp.inc
```diff
@@ -0,0 +1,985 @@
...
+            y_shared_entry.values[r * y_shared_entry.stride + rhs_index] =
+                (s_shared_entry.values[r * s_shared_entry.stride + rhs_index] -
+                 temp_sum) /
+                H_shared_entry.values[r * H_shared_entry.stride +
+                                      r * y_shared_entry.num_rhs + rhs_index];
+        }
+    }
+
+    __syncthreads();
+
```
Slaedr, line 454: Like my comment for the reference code, I think it would be good to split this function into the triangular solve separate from the update of `x`. That way, we would not be repeating the triangular solve at the end of each restart cycle. That may be somewhat more important here because we expect short restart cycles.

common/solver/batch_gmres_kernels.hpp.inc
```diff
@@ -0,0 +1,985 @@
...
+            __syncthreads();
+
+
+            compute_norm2<ValueType>(gko::batch::to_const(r_shared_entry),
+                                     res_norms_temp_shared_entry);
+
+            __syncthreads();
+
+            copy(gko::batch::to_const(res_norms_temp_shared_entry),
+                 res_norms_shared_entry, converged);
```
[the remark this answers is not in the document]
Slaedr: At some point we should discuss this. Is it so important for GMRES to stop the update on RHS that have converged? If we can just continue to update them, we can simplify and reduce the cost of the solver. Another way to address this would be for `compute_norm2`, `compute_dot` etc. to take a bit-mask and only update those RHS for whom the bit is set.

common/solver/batch_gmres_kernels.hpp.inc
```diff
@@ -0,0 +1,985 @@
...
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+namespace {
+
+
+template <typename ValueType>
+__device__ __forceinline__ void copy(
```
Slaedr, line 37: This should be moved to `batch_dense_kernels.hpp.inc` and reused for all the solvers.

omp/test/matrix/batch_dense_kernels.cpp
```diff
@@ -186,857 +152,276 @@ class BatchDense : public ::testing::Test {
...
-//    srmtx->move_to(drmtx.get());
-//    somtx->move_to(domtx.get());
+//    mat->add_scaled(alpha.get(), diag.get());
+//    dmat->add_scaled(dalpha.get(), ddiag.get());
 //
-//    GKO_ASSERT_MTX_NEAR(drmtx, domtx, 1e-14);
-//    GKO_ASSERT_MTX_NEAR(srmtx, somtx, 1e-14);
-//    GKO_ASSERT_MTX_NEAR(domtx, omtx, 1e-14);
+//    GKO_ASSERT_MTX_NEAR(mat, dmat, 1e-14);
 //}
```
Slaedr, line 250:
```suggestion
```

omp/test/matrix/batch_dense_kernels.cpp
```diff
@@ -186,857 +152,276 @@ class BatchDense : public ::testing::Test {
...
+//    dalpha = Mtx::create(omp);
+//    dalpha->copy_from(alpha.get());
 //
-//    rmtx->move_to(srmtx.get());
-//    omtx->move_to(somtx.get());
-//    srmtx->move_to(drmtx.get());
-//    somtx->move_to(domtx.get());
+//    mat->add_scaled(alpha.get(), diag.get());
+//    dmat->add_scaled(dalpha.get(), ddiag.get());
 //
```
Slaedr, line 248:
```suggestion
```

omp/test/matrix/batch_dense_kernels.cpp
```diff
@@ -186,857 +152,276 @@ class BatchDense : public ::testing::Test {
...
+//    auto diag =
+//        gko::matrix::Diagonal<Mtx::value_type>::create(ref, 532, diag_values);
+//    alpha = gko::initialize<Mtx>({2.0}, ref);
+//    auto dmat = Mtx::create(omp);
+//    dmat->copy_from(mat.get());
+//    auto ddiag = gko::matrix::Diagonal<Mtx::value_type>::create(omp);
+//    ddiag->copy_from(diag.get());
+//    dalpha = Mtx::create(omp);
+//    dalpha->copy_from(alpha.get());
 //
```
Slaedr, line 245:
```suggestion
```

include/ginkgo/core/solver/batch_gmres.hpp
```diff
@@ -0,0 +1,189 @@
...
+
+        /**
+         * To specify which tolerance is to be considered.
+         *
+         */
+        ::gko::stop::batch::ToleranceType GKO_FACTORY_PARAMETER_SCALAR(
```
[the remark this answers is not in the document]
Slaedr: As we discussed it might actually make sense to take them separately for combined convergence criteria in the future. But I am okay either way.

include/ginkgo/core/solver/batch_gmres.hpp
```diff
@@ -0,0 +1,187 @@
...
+        real_type GKO_FACTORY_PARAMETER_SCALAR(abs_residual_tol, 1e-11);
+
+        /**
+         * Restart parameter for Gmres
+         */
+        int GKO_FACTORY_PARAMETER_SCALAR(restart, 10);
+
+        /**
+         * To specify which tolerance is to be considered.
+         *
```
Slaedr, line 149:
```suggestion
```

core/test/solver/batch_idr.cpp
```diff
@@ -0,0 +1,287 @@
...
+}
+
+
+TYPED_TEST(BatchIdr, ApplyUsesInitialGuessReturnsTrue)
+{
+    ASSERT_TRUE(this->solver->apply_uses_initial_guess());
+}
+
+
+TYPED_TEST(BatchIdr, CanSetCriteria)
```
Slaedr, line 185: It would be good to have another similar test where you set the criteria using the setter functions after the solver is generated.

Slaedr (commented):
Next part of my review. Very good work on the solver kernels! Thanks for looking into the test tolerances; I think they're good now. For all the solvers, the scaling tolerance is at worst 1e3*epsilon, which I think is good enough for those scaling vectors. Let's see if they work on the CI machines. Thanks also for dealing with the repeated parameters in the public solver interfaces, great job.

I have some suggestions in some places; please see below. However, I have not yet reviewed the IDR kernel nor the OpenMP implementations.

common/matrix/batch_csr_kernels.hpp.inc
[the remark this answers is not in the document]
Slaedr: Thanks for going over it. I think it's okay, but there are now a lot of special functions there. I think it's fine for now though.

reference/solver/batch_bicgstab_kernels.cpp
```diff
@@ -0,0 +1,711 @@
...
+    const auto nrhs = b.num_rhs;
+
+
+    // TODO: Remove these assert statements once you make sure that there are no
+    // static allocations (in device functions and stopping criterion
+    // check_converged) which use the values in batch config struct.
+    GKO_ASSERT((batch_config<ValueType>::max_num_rows *
+                    batch_config<ValueType>::max_num_rhs >=
+                nrows * nrhs));
+    GKO_ASSERT(batch_config<ValueType>::max_num_rows >= nrows);
```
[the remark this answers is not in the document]
aggarwal2000: Removed.(Now there are no static allocations (not even proportional to the num_rhs) except for the one in the stopping criterion.

common/solver/batch_bicgstab_kernels.hpp.inc
aggarwal2000: Removed.

common/solver/batch_bicgstab_kernels.hpp.inc
aggarwal2000: Done!

common/solver/batch_bicgstab_kernels.hpp.inc
aggarwal2000: Now it is computed implicitly in the update_p function.

common/solver/batch_bicgstab_kernels.hpp.inc
aggarwal2000: I guess that is not possible as discussed.

common/solver/batch_bicgstab_kernels.hpp.inc
aggarwal2000: Merged updates of x and r, and also now I have only one omega as omega_old is no longer needed once omega_new is computed.

common/solver/batch_cg_kernels.hpp.inc
```diff
@@ -0,0 +1,601 @@
...
+
+            // x = x + alpha * p
+            update_x(x_shared_entry, gko::batch::to_const(p_shared_entry),
+                     gko::batch::to_const(alpha_shared_entry), converged);
+
+
+            // r = r - alpha * Ap
+            update_r(gko::batch::to_const(Ap_shared_entry),
+                     gko::batch::to_const(alpha_shared_entry), r_shared_entry,
+                     converged);
```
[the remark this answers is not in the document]
aggarwal2000: Done!

common/solver/batch_gmres_kernels.hpp.inc
```diff
@@ -0,0 +1,985 @@
...
+        if (conv) {
+            continue;
+        }
+
+        s_shared_entry.values[r * s_shared_entry.stride + c] =
+            zero<ValueType>();
+
+        if (r == 0) {
+            s_shared_entry.values[r * s_shared_entry.stride + c] =
+                z_norms_shared_entry.values[c];
```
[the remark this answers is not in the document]
aggarwal2000: Done

common/solver/batch_gmres_kernels.hpp.inc
aggarwal2000: This is not very clear to me. What is the exact meaning of multi-dot kernel? Here, there is a kind of dependency, value of w computed in every iteration depends upon the previous value of w.

common/solver/batch_gmres_kernels.hpp.inc
```diff
@@ -0,0 +1,985 @@
...
+{
+    for (int c = threadIdx.x; c < nrhs; c += blockDim.x) {
+        const uint32 conv = converged & (1 << c);
+
+        if (conv) {
+            continue;
+        }
+
+        ValueType temp = cs[c] * x[c] + sn[c] * y[c];
+        y[c] = -one<ValueType>() * sn[c] * x[c] + cs[c] * y[c];
```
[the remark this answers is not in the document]
aggarwal2000: Done!

common/solver/batch_gmres_kernels.hpp.inc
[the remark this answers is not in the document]
aggarwal2000: I have now tried adapting it for complex systems.

common/solver/batch_gmres_kernels.hpp.inc
```diff
@@ -0,0 +1,985 @@
...
+
+        v_i_plus_1_shared_entry.values[r * v_i_plus_1_shared_entry.stride + c] =
+            w_shared_entry.values[r * w_shared_entry.stride + c] /
+            h_i_plus_1_i_shared_entry.values[c];
+    }
+}
+
+
+template <typename ValueType>
+__device__ __forceinline__ void GeneratePlaneRotation(
```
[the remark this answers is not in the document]
aggarwal2000: Changed to snake_case.

reference/solver/batch_bicgstab_kernels.cpp
```diff
@@ -0,0 +1,711 @@
...
+        const uint32 conv = converged & (1 << c);
+
+        if (conv) {
+            continue;
+        }
+        beta_entry.values[0 * beta_entry.stride + c] =
+            (rho_new_entry.values[0 * rho_new_entry.stride + c] /
+             rho_old_entry.values[0 * rho_old_entry.stride + c]) *
+            (alpha_entry.values[0 * alpha_entry.stride + c] /
+             omega_old_entry.values[0 * omega_old_entry.stride + c]);
```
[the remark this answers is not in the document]
aggarwal2000: Done

reference/solver/batch_bicgstab_kernels.cpp
```diff
@@ -0,0 +1,711 @@
...
+        for (int c = 0; c < p_entry.num_rhs; c++) {
+            const uint32 conv = converged & (1 << c);
+
+            if (conv) {
+                continue;
+            }
+
+            p_entry.values[r * p_entry.stride + c] =
+                r_entry.values[r * r_entry.stride + c] +
+                beta_entry.values[0 * beta_entry.stride + c] *
```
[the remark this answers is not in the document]
aggarwal2000: Done

reference/solver/batch_bicgstab_kernels.cpp
```diff
@@ -0,0 +1,711 @@
...
+                    (p_entry.values[r * p_entry.stride + c] -
+                     omega_old_entry.values[0 * omega_old_entry.stride + c] *
```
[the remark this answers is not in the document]
aggarwal2000: Done

common/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -0,0 +1,847 @@
...
+        alpha_shared_entry.values[0 * alpha_shared_entry.stride + c] =
+            one<ValueType>();
+    }
+
+
+    // Compute norms of rhs
+    compute_norm2<ValueType>(b_global_entry, rhs_norms_shared_entry);
+
+    // copy x from global to shared memory
+    copy(x_global_entry, x_shared_entry, 0x00000000);
```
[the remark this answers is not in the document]
aggarwal2000: I have tried this many times, but doing so leads to compilation errors- undefined reference. So for the time being, I am using this bit-mak dependent copy here.

reference/solver/batch_cg_kernels.cpp
```diff
@@ -0,0 +1,539 @@
...
+
+    // rho_old = r' * z
+    batch_dense::compute_dot_product(gko::batch::to_const(r_entry),
+                                     gko::batch::to_const(z_entry),
+                                     rho_old_entry);
+}
+
+
+template <typename ValueType>
+inline void compute_beta(
```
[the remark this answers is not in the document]
aggarwal2000: Done

reference/solver/batch_cg_kernels.cpp
```diff
@@ -0,0 +1,539 @@
...
+                          gko::batch::to_const(p_entry),
+                          gko::batch::to_const(Ap_entry), alpha_entry,
+                          converged);
+
+            // x = x + alpha * p
+            update_x(x_entry, gko::batch::to_const(p_entry),
+                     gko::batch::to_const(alpha_entry), converged);
+
+            // r = r - alpha * Ap
+            update_r(gko::batch::to_const(Ap_entry),
```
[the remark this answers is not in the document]
aggarwal2000: Done

reference/solver/batch_gmres_kernels.cpp
```diff
@@ -0,0 +1,853 @@
...
+    //         {
+    //             H_entry.values[row_idx * H_entry.stride + col_idx*num_rhs +
+    //             rhs_idx] = zero<ValueType>();
+    //         }
+    //     }
+    // }
+}
+
+template <typename ValueType>
+inline void GeneratePlaneRotation(const ValueType *const x,
```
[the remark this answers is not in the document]
aggarwal2000: Changed all to snake_case.

reference/solver/batch_gmres_kernels.cpp
```diff
@@ -0,0 +1,853 @@
...
+    using real_type = typename gko::remove_complex<ValueType>;
+    const int &i = inner_iter;
+    const int num_rows = w_entry.num_rows;
+    const int num_rhs = w_entry.num_rhs;
+
+    const gko::batch_dense::BatchEntry<const ValueType> v_i_entry{
+        &V_entry.values[i * V_entry.stride * num_rows], V_entry.stride,
+        num_rows, V_entry.num_rhs};
+
+    ValueType w_temp[batch_config<ValueType>::max_num_rows *
```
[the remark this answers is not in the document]
aggarwal2000: This now uses dynamic memory allocation.

reference/solver/batch_gmres_kernels.cpp
```diff
@@ -0,0 +1,853 @@
...
+
+
+            // KrylovBasis_0 = z/norm(z)
+            // s -> fill with zeroes
+            // s(0) = norm(z)
+            update_v_naught_and_s(V_entry, s_entry,
+                                  gko::batch::to_const(z_entry), converged);
+
+            for (int inner_iter = -1; inner_iter < restart - 1;) {
+                inner_iter++;
```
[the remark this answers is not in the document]
aggarwal2000: I just wanted to make it look similar to the while loop, where we start from -1 and update the counter inside the loop, but I have changed this now.

common/solver/batch_gmres_kernels.hpp.inc
aggarwal2000: Done

reference/solver/batch_gmres_kernels.cpp
aggarwal2000: This is not very clear to me, why don't we want the triangular solve at the end of every restart? Isn't that necessary for the update of x?

cuda/test/solver/batch_bicgstab_kernels.cpp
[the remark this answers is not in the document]
Slaedr: Nice.

Slaedr (commented):
That is good progress. Please see my comments below.

My concerns are:
- We can remove the `omp` "device" level kernels, mainly because I'm not sure how good `parallel for` will be at the "device" level, as we are already using it at the "global" level. We need SIMD at the "device" level, but that's a whole different thing that we don't want to deal with right now. If you are really interested in working on SIMD for the batch stuff, you can keep the extra code, but otherwise I think it's best to remove the copies in omp for now.
- You could move the mask-dependent copy function to the batch dense kernels and reuse that for all the solvers.

common/solver/batch_cg_kernels.hpp.inc
```diff
@@ -0,0 +1,601 @@
...
+    const gko::batch_dense::BatchEntry<ValueType> &x_shared_entry,
+    const gko::batch_dense::BatchEntry<ValueType> &r_shared_entry,
+    const PrecType &prec_shared,
+    const gko::batch_dense::BatchEntry<ValueType> &z_shared_entry,
+    const gko::batch_dense::BatchEntry<ValueType> &rho_old_shared_entry,
+    const gko::batch_dense::BatchEntry<ValueType> &p_shared_entry,
+    const gko::batch_dense::BatchEntry<typename gko::remove_complex<ValueType>>
+        &rhs_norms_shared_entry,
+    const gko::batch_dense::BatchEntry<typename gko::remove_complex<ValueType>>
+        &res_norms_shared_entry)
```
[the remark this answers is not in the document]
pratikvn: I think with the `__restrict__`, the compiler can do the optimizations, that you were expecting it to do ? With for example, using the same pointers for the different batch entry stuff without filling up the register files. Maybe that helps in reducing the register usage as well.

common/solver/batch_gmres_kernels.hpp.inc
pratikvn: No, let's look into this later. It is just an idea, which would help in doing the different dot products at once.

common/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -0,0 +1,847 @@
...
+
+        // stopping criterion object
+        StopType stop(nrhs, max_iter, abs_tol, rel_tol,
+                      static_cast<stop::tolerance>(tol_type),
+                      rhs_norms_shared_entry.values, converged);
+
+        int iter = -1;
+
+
+        while (1) {
```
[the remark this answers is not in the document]
pratikvn: Yes, good point. I think we should log the residual only once. Of course we will need to update the convergence state twice.

common/solver/batch_bicgstab_kernels.hpp.inc
pratikvn: Okay, I think we can leave it as it now for now and adapt this later, if necessary.

common/solver/batch_gmres_kernels.hpp.inc
```diff
@@ -0,0 +1,870 @@
...
+                         .values[inner_iter * cs_shared_entry.stride],
+                    &sn_shared_entry
+                         .values[inner_iter * sn_shared_entry.stride],
+                    nrhs,
+                    &H_shared_entry.values[inner_iter * H_shared_entry.stride +
+                                           inner_iter * nrhs],
+                    &H_shared_entry
+                         .values[(inner_iter + 1) * H_shared_entry.stride +
+                                 inner_iter * nrhs],
+                    converged);
```
pratikvn, line 758: For readability, this makes sense. For performance, though it might not matter too much here, you can do both of these in one loop.

common/solver/batch_gmres_kernels.hpp.inc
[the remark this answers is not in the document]
pratikvn: Yeah, that is difficult to say. For correctness, I think it makes sense to stop updating the vectors which have converged. If for performance reasons not updating is better, then provided that the solutions are still correct, I guess it would be okay to continue to update them. But if possible, I think a bit-mask approach is definitely preferable.

pratikvn (commented):
Thanks for addressing the changes! It mostly looks good from my side, only some previous comments still need to be addressed.

core/solver/batch_bicgstab.cpp
```diff
@@ -56,8 +56,7 @@ std::unique_ptr<BatchLinOp> BatchBicgstab<ValueType>::transpose() const
     return build()
         .with_preconditioner(parameters_.preconditioner)
         .with_max_iterations(parameters_.max_iterations)
-        .with_rel_residual_tol(parameters_.rel_residual_tol)
-        .with_abs_residual_tol(parameters_.abs_residual_tol)
+        .with_residual_tol(parameters_.residual_tol)
```
Slaedr, line 59: Is it possible to take separate tolerances as before here, but have only one tolerance in the kernel based on the tolerance type? As a user interface, it makes sense to take two separate tolerances because we might need to use them both in future. Within the kernel, we prioritize efficiency for now and keep only one, but we could change that any time without breaking interface.

common/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -461,9 +461,8 @@ __global__ void apply_kernel(
 
 
         // stopping criterion object
-        StopType stop(nrhs, max_iter, abs_tol, rel_tol,
-                      static_cast<stop::tolerance>(tol_type),
-                      rhs_norms_shared_entry.values, converged);
+        StopType stop(nrhs, max_iter, tol, rhs_norms_shared_entry.values,
+                      converged);
```
Slaedr, line 465: This looks good.

common/stop/batch_criteria.hpp.inc
```diff
@@ -262,11 +235,7 @@ private:
                 ichk = 1;
             }
 
-            if (tol_type == tolerance::relative) {
-                ichk = (res_norms[rhs_index] / rhs_norms[rhs_index] < rel_tol);
-            } else if (tol_type == tolerance::absolute) {
-                ichk = (res_norms[rhs_index] < abs_tol);
-            }
+            ichk = (res_norms[rhs_index] < abs_tol);
```
Slaedr, line 238: Looks good.

common/solver/batch_cg_kernels.hpp.inc
```diff
@@ -0,0 +1,601 @@
...
+        const uint32 conv = converged & (1 << c);
+
+        if (conv) {
+            continue;
+        }
+
+        dest_shared_entry.values[r * dest_shared_entry.stride + c] =
+            src_shared_entry.values[r * src_shared_entry.stride + c];
+    }
+}
```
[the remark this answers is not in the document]
aggarwal2000: I have moved this function to batch_dense.hpp.inc now.

common/solver/batch_cg_kernels.hpp.inc
```diff
@@ -0,0 +1,601 @@
...
+        }
+        return;
+    }
+
+    if (batch_config<ValueType>::max_num_rhs < nrhs) {
+        if (blockIdx.x == 0 && threadIdx.x == 0) {
+            printf(
+                " Compile time constant: max_rhs < actual rhs in "
+                "matrix!\n");
+        }
```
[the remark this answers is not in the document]
aggarwal2000: Changed printf to assert statements.

common/solver/batch_cg_kernels.hpp.inc
```diff
@@ -0,0 +1,601 @@
...
+
+            __syncthreads();
+
+            compute_norm2<ValueType>(
+                gko::batch::to_const(r_shared_entry),
+                res_norms_temp_shared_entry);  // residual norms
+            __syncthreads();
+            copy(gko::batch::to_const(res_norms_temp_shared_entry),
+                 res_norms_shared_entry,
+                 converged);  // update res norms only for those RHSs which have
```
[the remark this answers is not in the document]
aggarwal2000: Done!

common/solver/batch_gmres_kernels.hpp.inc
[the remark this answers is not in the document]
aggarwal2000: So now all batch dense functions have an overload which takes a bit-mask.

reference/solver/batch_bicgstab_kernels.cpp
```diff
@@ -0,0 +1,712 @@
...
+            r_entry.values[r * r_entry.stride + c] =
+                s_entry.values[r * s_entry.stride + c] -
+                omega_new_entry.values[c] *
+                    t_entry.values[r * t_entry.stride + c];
+        }
+    }
+}
+
+template <typename ValueType>
+inline void copy(
```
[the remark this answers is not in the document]
aggarwal2000: Done!

common/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -0,0 +1,847 @@
...
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+
+namespace {
+
+
+template <typename ValueType>
+__device__ __forceinline__ void copy(
```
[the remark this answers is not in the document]
aggarwal2000: Done!

common/solver/batch_bicgstab_kernels.hpp.inc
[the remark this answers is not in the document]
aggarwal2000: Yes, that works now.

common/stop/batch_criteria.hpp.inc
```diff
@@ -150,3 +141,141 @@ private:
...
+        if (rhs_index < nrhs) {
+            int ichk = 0;
+
+            if (converged &
+                (1 << rhs_index)) {  // don't check for convergence again for
+                                     // RHSs which have already converged.
+                ichk = 1;
+            }
+
+            if (tol_type == tolerance::relative) {
```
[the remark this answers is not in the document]
aggarwal2000: Done!

reference/solver/batch_cg_kernels.cpp
```diff
@@ -0,0 +1,539 @@
...
+ *
+ * @ingroup batch_cg
+ */
+namespace batch_cg {
+
+namespace {
+
+
+template <typename ValueType>
+inline void copy(
```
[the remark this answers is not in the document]
aggarwal2000: Done

core/test/solver/batch_idr.cpp
aggarwal2000: I have added such tests now.

common/preconditioner/batch_jacobi.hpp.inc
```diff
@@ -39,11 +39,6 @@ class BatchJacobi final {
 public:
     using value_type = ValueType;
 
-    /**
-     * The size of the work vector required in case of static allocation.
-     */
-    static constexpr int work_size = batch_config<ValueType>::max_num_rows;
-
```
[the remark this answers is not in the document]
aggarwal2000: Okay, re-added this.

common/solver/batch_cg_kernels.hpp.inc
```diff
@@ -0,0 +1,441 @@
...
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+
+namespace {
+
+template <typename ValueType>
+__device__ __forceinline__ void copy(
```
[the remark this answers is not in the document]
aggarwal2000: Done

common/solver/batch_cg_kernels.hpp.inc
```diff
@@ -0,0 +1,441 @@
...
+__device__ __forceinline__ void compute_alpha(
+    const gko::batch_dense::BatchEntry<const ValueType> &rho_old_shared_entry,
+    const gko::batch_dense::BatchEntry<const ValueType> &p_shared_entry,
+    const gko::batch_dense::BatchEntry<const ValueType> &Ap_shared_entry,
+    const gko::batch_dense::BatchEntry<ValueType> &alpha_shared_entry,
+    const uint32 &converged)
+{
+    compute_dot_product<ValueType>(p_shared_entry, Ap_shared_entry,
+                                   alpha_shared_entry);
+    __syncthreads();
```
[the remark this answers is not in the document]
aggarwal2000: Done

common/solver/batch_cg_kernels.hpp.inc
```diff
@@ -0,0 +1,441 @@
...
+
+    for (int c = threadIdx.x; c < alpha_shared_entry.num_rhs; c += blockDim.x) {
+        const uint32 conv = converged & (1 << c);
+
+        if (conv) {
+            continue;
+        }
+        alpha_shared_entry.values[c] =
+            rho_old_shared_entry.values[c] / alpha_shared_entry.values[c];
```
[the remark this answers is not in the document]
aggarwal2000: Yes, merged computation of alpha with the update_x_and_r kernel.

common/solver/batch_gmres_kernels.hpp.inc
```diff
@@ -0,0 +1,870 @@
...
+    // {
+    //     int row_index = ( tid / V_shared_entry.num_rhs ) %
+    //     x_shared_entry.num_rows; int basis_vec_index =  ( tid /
+    //     V_shared_entry.num_rhs ) / x_shared_entry.num_rows ; int rhs_index =
+    //     tid % V_shared_entry.num_rhs;
+
+    //     V_shared_entry.values[ ( basis_vec_index * x_shared_entry.num_rows  +
+    //     row_index ) * V_shared_entry.stride + rhs_index ] =
+    //     zero<ValueType>();
+    // }
```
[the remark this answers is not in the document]
aggarwal2000: Removed.

omp/log/batch_logger.hpp
```diff
@@ -0,0 +1,119 @@
...
+     *
+     * @param batch_idx  The index of linear system in the batch to log.
+     * @param iter  The current iteration count.
+     * @param res_norm  Norms of current residuals for each RHS.
+     * @param converged  Bitset representing convergence state for each RHS.
+     */
+    void log_iteration(const size_type batch_idx, const int iter,
+                       const real_type *const res_norm, const uint32 converged)
+    {
+        if (converged != init_converged_ || iter >= max_iters_ - 2) {
```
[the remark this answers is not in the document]
aggarwal2000: Changed.

core/solver/batch_bicgstab.cpp
aggarwal2000: I have changed it back now.

common/matrix/batch_csr_kernels.hpp.inc
```diff
@@ -42,21 +42,57 @@ __device__ __forceinline__ void spmv_kernel(
...
+        }
+
+
+        const int gap_rounded_off = static_cast<int>(
+            pow(2, ceil(log2(static_cast<double>(
+                       row_end -
+                       row_start)))));  // round off to the closest power of 2
+        // subwarp_grp level reduction
+        for (int i = static_cast<int>(
+                         min(static_cast<double>(gap_rounded_off),
```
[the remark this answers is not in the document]
aggarwal2000: I have now removed casts which are not required.

common/solver/batch_gmres_kernels.hpp.inc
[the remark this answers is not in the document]
pratikvn: Nice! Thanks

omp/solver/batch_bicgstab_kernels.cpp
```diff
@@ -0,0 +1,652 @@
...
+
+
+    const int local_size_bytes =
+        gko::kernels::batch_bicgstab::local_memory_requirement<ValueType>(
+            nrows, nrhs) +
+        PrecType::dynamic_work_size(nrows, a.num_nnz) * sizeof(ValueType);
+    using byte = unsigned char;
+
+
+#pragma omp parallel for firstprivate(logger)
```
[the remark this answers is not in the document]
aggarwal2000: Now omp backend reuses the reference kernels as well as the loop body for all the solvers.

reference/solver/batch_bicgstab_body.hpp
```diff
@@ -0,0 +1,254 @@
...
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+
+for (size_type ibatch = 0; ibatch < nbatch; ibatch++) {
```
Slaedr, line 34: It would be better to make this a proper function and put it in `batch_bicgstab_kernels.hpp`.

reference/solver/batch_bicgstab_body.hpp
```diff
@@ -0,0 +1,254 @@
...
+    byte *const shared_space = local_space.get_data();
```
Slaedr, line 35: It would be better to move the body of this loop into `batch_bicgstab_kernels.hpp`.

reference/solver/batch_bicgstab_body.hpp
Slaedr, line 34: The loop itself could moved to the `apply_impl` in `batch_bicgstab_kernels.cpp`.

Slaedr (approved):
Great work! Thanks for all the work in addressing most of the comments. I just have one issue with the file structure, but apart from that we can merge it, and address the remaining issues and optimizations in separate PRs.

reference/solver/batch_bicgstab_body.hpp
aggarwal2000: Done!

reference/solver/batch_bicgstab_body.hpp
aggarwal2000: Done!

pratikvn (approved):
An [minor issue on clang-cuda](https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/1329501014#L616) with the batch_idr_kernels with the way you initialize `mul_sh`. Otherwise, looks good. Great work!

omp/solver/batch_bicgstab_kernels.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+    const int local_size_bytes =
+        gko::kernels::batch_bicgstab::local_memory_requirement<ValueType>(
+            nrows, nrhs) +
+        PrecType::dynamic_work_size(nrows, a.num_nnz) * sizeof(ValueType);
+    using byte = unsigned char;
+
+    Array<byte> local_space(exec, local_size_bytes);
+
+#pragma omp parallel for firstprivate(logger) firstprivate(local_space)
```
Slaedr, line 110: I hope the `firstprivate` here will not do a deep copy of the `local_space` for each thread.

PR #779, opened 2021-05-31
Dpcpp port ELL and the porting script

yhmtsai:
This PR adds the dpcpp ported ELL and the porting script
For review before merging dense, the https://github.com/ginkgo-project/ginkgo/pull/779/files/7ac6d0909f006b52c5516e63b50cf0dff6b18012..3fc8bea0d5d6f81ef4be661aade1254ad1d54355 changes only contains the ell part

TODO:
- [x] clear some unused code in script
- [ ] add ConfigSelection into ELL (maybe not)
- [x] merge dense


yhmtsai:
format!

dev_tools/oneapi/add_host_function.sh
```diff
@@ -0,0 +1,305 @@
+#!/bin/bash
+
+extract_varname() {
+    local str="$1"
+    # local GET_PARAM=" *([^ ]*) *$"
+    # Need to remove the  = ....
+    # note. it only remove the simple one
```
tcojean, line 7: Here you mean that it doesn't work with `x = y = 3` sort of cases?

dev_tools/oneapi/add_host_function.sh
```diff
@@ -0,0 +1,305 @@
...
+                        (get_block_size(exec_config) >= get_block_size(config));
+                },
+            ::gko::syn::value_list<${list_list[bool]}>(),
+            ::gko::syn::value_list<${list_list[int]}>(),
+            ::gko::syn::value_list<${list_list[size_type]}>(),
+            ::gko::syn::type_list<${list_list[typename]}>(),
+            grid, block, dynamic_shared_memory, stream, ${var_input});
+    }
+    "
+}
```
tcojean, line 143: In general, is this function still used? You have macros for defining this now, no?

dev_tools/oneapi/add_host_function.sh
```diff
@@ -0,0 +1,305 @@
...
+IN_FUNC=0
+STORE_LINE=""
+STORE_REGEX="__ *$"
+HOST_SUFFIX="_AUTOHOSTFUNC"
+EXTRACT_KERNEL="false"
+GINKGO_LICENSE_BEACON="******************************<GINKGO LICENSE>******************************"
+DURING_LICENSE="false"
+SKIP="false"
+MAP_FILE="map_list"
+rm "${MAP_FILE}"
```
tcojean, line 168: Some of the previous definitions can be shared as well, in particular the LICENSE stuff

dev_tools/oneapi/add_host_function.sh
```diff
@@ -0,0 +1,305 @@
...
+        fi
+        if [[ "${line}" =~ ${SCOPE_START} ]]; then
+            IN_FUNC=$((IN_FUNC+1))
+        fi
+        if [[ "${line}" =~ ${SCOPE_END} ]]; then
+            IN_FUNC=$((IN_FUNC-1))
+        fi
+        # output to new file
+        continue
+    fi
```
tcojean, line 223: In the end, all the previous things are skipped, i.e. not printed right? I
don't see any `echo`?

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
+#!/bin/bash
+
```
tcojean, line 2: I think you should comment on what the file takes as parameter. At least, `$1`
should be a CUDA file to convert?

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+# the following parameters set by environment
```
tcojean, line 3: ```# the following are parameters set by environment variables```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+# MUST:
```
tcojean, line 4: ```# REQUIRED:```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+#   CUDA_HEADER_DIR: contains the cuda headers
+# OPTIONAL:
+#   ROOT_DIR: the ginkgo folder. The default is current path
+#   BUILD_DIR: the ginkgo build folder, which needs cmake to generate config.hpp and gtest include. The default is "build"
+#              Note. It requires GINKGO_BUILD_TESTS=ON to download gtest but does not need to compile ginkgo.
```
tcojean, line 9: ```#              Note. It requires GINKGO_BUILD_TESTS=ON to download gtest but it is not required to compile ginkgo.```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+#                    If GTEST_HEADER_DIR is from other place, does not require GINKGO_BUILD_TESTS.
```
tcojean, line 10: ```#                    If GTEST_HEADER_DIR is available elsewhere GINKGO_BUILD_TESTS is not required.```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+#                    If copy the ginkgo config.hpp from other ginkgo build into "${ROOT_DIR}/include/ginkgo/", do not require cmake.
```
tcojean, line 11: ```#                    CMake is not required if copying the ginkgo config.hpp from another ginkgo build into "${ROOT_DIR}/include/ginkgo/".```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+    echo "SCRIPT_DIR ${SCRIPT_DIR}"
+    echo "ROOT_DIR ${ROOT_DIR}"
+    echo "ROOT_BUILD_DIR ${ROOT_BUILD_DIR}"
+    echo "GTEST_HEADER_DIR ${GTEST_HEADER_DIR}"
+    echo "CUDA_HEADER_DIR ${CUDA_HEADER_DIR}"
+    echo "ROOT_BUILD_DIR ${ROOT_BUILD_DIR}"
+    echo "CLANG_FORMAT ${CLANG_FORMAT}"
+fi
+if [[ "${CUDA_HEADER_DIR}" == "" ]]; then
+    echo "Please set the CUDA_HEADER_DIR in environment"
```
tcojean, line 36: ```    echo "Please set the environment variable CUDA_HEADER_DIR"```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+    exit 1
+fi
+# move to working_directory
+cd working_directory
```
tcojean, line 40: Where is this supposed to happen, in the source directory?
Also, there is no `mkdir` for this, shouldn't there be?

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+
+KERNEL_SYNTAX_START="<<<"
+KERNEL_SYNTAX_END=">>>"
+DEVICE_CODE_SYNTAX="#include \"(common.*)\""
+FUNCTION_END=");"
+SPECIAL_SUFFIX="_AUTOHOSTFUNC"
+EXTRACT_KERNEL="false"
+GLOBAL_FILE="global_kernel"
+CONFIG_SELECTION_SUFFIX="_CONFIG"
+
```
tcojean, line 51: Exemple of a comment for this function:
```
Checks if $1 is self contained code, that is it does not have an open and unclosed
code portion, e.g. `my_struct->my_func(xxx,` should fail.
```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+check_closed() {
+    local str="$1"
+    str="${str//->}"
```
tcojean, line 54: Extra comment? ```    # Replace everything except begin or end characters, resp. (<[ and )>]```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+    str_start="${str//[^(<\[]}"
+    str_end="${str//[^>)\]]}"
```
tcojean, line 56: Extra comment? ```    # Check that there are as many begin as end characters````

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+    if [[ "${#str_start}" -eq "${#str_end}" ]]; then
+        echo "true"
+    else
+        echo "false"
+    fi
+}
+
```
tcojean, line 63: Possible comment:
```# Converts a CUDA kernel call to the DPC++ equivalent. Also takes care of the config selection calls and DPC++ queue manipulation if needed.```

File: dev_tools/oneapi/convert_source.sh [hunk 17]
```diff
@@ -0,0 +1,429 @@
...
+convert_syntax() {
```
tcojean, line 64: Tentative other name:
`convert_cuda_kernel_call`

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+    local syntax_regex="([^<>]*)(<[^<>]*>)?<<<(.*)>>>(.*)"
```
tcojean, line 65: Maybe a comment for the regex?
``` # This regex matches the kernel calls, with:
# [1]: kernel name
# [2]: template parameters if it exists
# [3]: CUDA parameters
# [4]: the rest of the function call (parameters, ...)
```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+    local str="$1"
```
tcojean, line 66: As a note, couldn't the template parameters call another templated struct? In
that case, maybe the check `[^<>]*` isn't warranted for the template portion?

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+    str=$(echo "${str}" | sed -E 's/ +/ /g')
+    # echo "str ${str}"
+    local temp=""
+    local num=0
+    local var=""
+    if [[ "${str}" =~ ${syntax_regex} ]]; then
+        content="${BASH_REMATCH[3]}"
+        # echo "// content ${content}"
```
tcojean, line 74: ```        local content="${BASH_REMATCH[3]}"```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+        IFS=',' read -ra par <<< "$content"
+        for variable in "${par[@]}"; do
+            if [ -n "${temp}" ]; then
+                temp="${temp},"
+            fi
+            temp="${temp}${variable}"
+            # echo "// temp ${temp}"
+            is_closed=$(check_closed "$temp")
+            if [[ "${is_closed}" = "true" ]]; then
```
tcojean, line 83: I think a comment is warranted here:
```# Do not fall into the trap of counting parameters in a sub-expression, such as a function call inside the CUDA kernel parameters```

File: dev_tools/oneapi/convert_source.sh [hunk 22]
```diff
@@ -0,0 +1,429 @@
...
+                else
+                    if [ -n "${var}" ]; then
+                        var="${var},"
+                    fi
+                    var="${var}${temp}"
+                fi
+                temp=""
+            fi
+        done
+        # echo "// var ${var}"
```
tcojean, line 98: If I'm correct, all you do in the previous loop is to count the number of
variables in the list, and you reconstruct it all in `$var`? In that case,
isn't it possible to use three regex on `$content` to cleanup the annoying
parts before counting? Like:

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+        # echo "var ${var}"
```
tcojean, line 99:
```
temp=${content//\[*\]/}
temp=${temp//\(*\)/}
temp=${temp//<*>/}
num=$((${#temp//[^,]} + 1))
# You can also do checks for when num == 4
```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+                var="${var}, GET_QUEUE"
+                num=$((num+1))
+            fi
+        fi
+        local suffix=""
+        local function_name=$(echo "${BASH_REMATCH[1]}" | sed -E 's/(.*::)| //g')
+        # echo "// functioname ${function_name}"
+        MAP_FILE="map_list"
+        suffix=$(cat "${MAP_FILE}" | sed -nE "s/${function_name} -> ${function_name}(.*)/\1/p")
+        suffix_matches=$(echo "${suffix}" | wc -l)
```
tcojean, line 118: This seems to come from somewhere else, maybe put it as a global variable or
function variable?

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+        if [[ "${suffix_matches}" -gt 1 ]]; then
+            echo "static_assert(false, \"Has ${suffix_matches} matches for ${function_name}\");"
```
tcojean, line 120: ```        local suffix_matches=$(echo "${suffix}" | wc -l)```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+        fi
+    fi
+}
+
+
+convert_regex_allowed() {
+    local str="$1"
+    str=$(echo "${str}" | sed -E 's/(\/|\(|\)|\.)/\\\1/g')
+    echo "$str"
+}
```
tcojean, line 143: This one is sort of hard, I would appreciate a comment. Why use `\\\1`? It
looks like you are looking for different patterns using `|`. Would something
like the following work?
``` sed -E 's:([/().]{1)):\\\1/g' ```

File: dev_tools/oneapi/convert_source.sh [hunk 27]
```diff
@@ -0,0 +1,429 @@
...
+
+# Transfer header file to the correct one
+# common -> dpcpp_common
+# cuda -> dpcpp
```
tcojean, line 147: ```# Transfer header file to the correct location```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+# add sycl header
+
+input="$1"
+filename="${ROOT_DIR}/$input"
+if [[ "${VERBOSE}" == 1 ]]; then
+    echo "Porting file ${filename}"
+fi
+# check file exists
+
```
tcojean, line 156: Add Blankline?

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+if [ ! -f "${filename}" ]; then
+    echo "${filename} does not exist"
```
tcojean, line 158: Remove blankline?

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+    echo "the original file ${UNFORMAT_FILE}"
+    echo "the formatted file ${FORMAT_FILE}"
+    echo "collect common/*.inc in file ${EMBED_FILE}"
+    echo "add autohost func in file ${EMBED_HOST_FILE}"
+    echo "convert original CUDA call in file ${OUTPUT_FILE}, which is the file for dpct"
+fi
+rm "${OUTPUT_FILE}"
+# echo "#include \"trick/dim3.hpp\"" >> "${OUTPUT_FILE}"
+echo "#define GET_QUEUE 0" >> "${OUTPUT_FILE}"
+# add empty ginkgo license such that format_header recognize some header before header def macro
```
tcojean, line 183: It's the same in other places, if lines are not needed anymore it's better to remove them.

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+echo "/*${GINKGO_LICENSE_BEACON}" >> "${OUTPUT_FILE}"
+echo "${GINKGO_LICENSE_BEACON}*/" >> "${OUTPUT_FILE}"
+rm "${GLOBAL_FILE}"
+rm "${EMBED_FILE}"
+while IFS='' read -r line; do
+    if [[ "${line}" =~ $DEVICE_CODE_SYNTAX ]]; then
+        # hold the command to easy replace
```
tcojean, line 190: Maybe a comment here?
```     # Embed shared kernels into ${EMBED_FILE}```

File: dev_tools/oneapi/convert_source.sh [hunk 32]
```diff
@@ -0,0 +1,429 @@
...
+        if [ -n "${DEVICE_FILE}" ]; then
+            DEVICE_FILE="${DEVICE_FILE};"
+        fi
+        DEVICE_FILE="${DEVICE_FILE}${device_file}"
+    else
+        echo "${line}" >> "${EMBED_FILE}"
+    fi
+done < "${UNFORMAT_FILE}"
+${CLANG_FORMAT} -style=file "${EMBED_FILE}" > "${FORMAT_FILE}"
+${SCRIPT_DIR}/add_host_function.sh "${FORMAT_FILE}" > "${EMBED_HOST_FILE}"
```
tcojean, line 205: Maybe a comment here?
```# Call clang-format for better formatting. ```

File: dev_tools/oneapi/convert_source.sh [hunk 33]
```diff
@@ -0,0 +1,429 @@
...
+
```
tcojean, line 206: Maybe a newline and a comment here (correct it if I'm wrong)?
```# Add an extra host function so that the converted DPC++ code will look like CUDA.```

File: dev_tools/oneapi/convert_source.sh [hunk 34]
```diff
@@ -0,0 +1,429 @@
...
+CONFIG_REGEX="constexpr *Config *([a-zA-Z0-9_]*) *= *([a-zA-Z][a-zA-Z_:,\(\) ]*)"
+while IFS='' read -r line; do
```
tcojean, line 208: Maybe a comment?
```# Convert kernel calls and manage DPC++ config based kernel generation.```

File: dev_tools/oneapi/convert_source.sh [hunk 35]
```diff
@@ -0,0 +1,429 @@
...
+replace_regex="s/dim3/dim3_t/g"
+replace_regex="${replace_regex};s/this_thread_block/this_thread_block_t/g"
+replace_regex="${replace_regex};s/this_grid/this_grid_t/g"
+replace_regex="${replace_regex};s/tiled_partition/tiled_partition_t/g"
+replace_regex="${replace_regex};s/thread::/thread_t::/g"
+replace_regex="${replace_regex};s/bitonic_sort/bitonic_sort_t/g"
+replace_regex="${replace_regex};s/reduction_array/reduction_array_t/g"
+replace_regex="${replace_regex};s/cuda\/components\/thread_ids.cuh/trick\/thread_ids.hpp/g"
+replace_regex="${replace_regex};s/cuda\/components\/cooperative_groups.cuh/trick\/cooperative_groups.hpp/g"
+replace_regex="${replace_regex};s/cuda\/components\/sorting.cuh/trick\/sorting.hpp/g"
```
tcojean, line 252: ```replace_regex="${replace_regex};s:cuda/components/thread_ids.cuh:trick/thread_ids.hpp:g"```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+replace_regex="${replace_regex};s/cuda\/components\/reduction.cuh/trick\/reduction.hpp/g"
```
tcojean, line 253: ```replace_regex="${replace_regex};s:cuda/components/cooperative_groups.cuh:trick/cooperative_groups.hpp:g"```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+replace_regex="${replace_regex};s/CUH_/DP_HPP_/g"
```
tcojean, line 254: ```replace_regex="${replace_regex};s:cuda/components/sorting.cuh:trick/sorting.hpp:g"```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+replace_regex="${replace_regex};s/(.*template.*)(Config * [a-zA-Z0-9_]*) * = *[a-zA-Z0-9_:]*(\(+[^\)]*\))?/\1\2/g"
```
tcojean, line 255: ```replace_regex="${replace_regex};s:cuda/components/reduction.cuh:trick/reduction.hpp:g"```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+# keep using original xxx.sync(); ->xxx;//.sync();
+replace_regex="${replace_regex};s/(\.sync\(\);)/;\/\/.sync()/g"
+# template macro(); lead std::length_error
+replace_regex="${replace_regex};s/(template GKO.*;)/\/\/ \1/g"
```
tcojean, line 259: ```replace_regex="${replace_regex};s|\.sync\(\);|;//.sync()|g"```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+sed -i -E "${replace_regex}" "${OUTPUT_FILE}"
+if grep -Eq "dim3" ${OUTPUT_FILE}; then
```
tcojean, line 261: ```replace_regex="${replace_regex};s|(template GKO.*;)|// \1|g"```
Maybe a comment for the next line?
```# Add the trick dim3_t header if dim3 is used.```

File: dev_tools/oneapi/convert_source.sh [hunk 41]
```diff
@@ -0,0 +1,429 @@
...
+    sed -i '1 i#include "trick/dim3_t.hpp"' ${OUTPUT_FILE}
+fi
+# add the cooperative group header according to group:: because some sources forget to add it
+if grep -Eq "group::" ${OUTPUT_FILE}; then
+    # Found
+    sed -i '1 i#include "trick/cooperative_groups.hpp"' ${OUTPUT_FILE}
+fi
+# exit 0
+
+OUTPUT_FOLDER="output"
```
tcojean, line 272: Can be removed

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+    echo "The dpct calling:"
+    echo "dpct --extra-arg=\"-std=c++14\" --extra-arg=\"-I ${ROOT_DIR}\" --extra-arg=\"-I ${ROOT_DIR}/include\" --extra-arg=\"-I ${ROOT_DIR}/dev_tools/oneapi\" --extra-arg=\"-I ${GTEST_HEADER_DIR}\" --cuda-include-path=\"${CUDA_HEADER_DIR}\" --format-range=none ${OUTPUT_FILE} --out-root=${OUTPUT_FOLDER}"
+fi
+
+# Delete output/source.cpp
+rm "${OUTPUT_FOLDER}/${OUTPUT_FILE}"
+rm "${OUTPUT_FOLDER}/${OUTPUT_FILE}.dp.cpp"
+
+dpct --extra-arg="-std=c++14" --extra-arg="-I ${ROOT_DIR}" --extra-arg="-I ${ROOT_DIR}/include" --extra-arg="-I ${ROOT_DIR}/dev_tools/oneapi" --extra-arg="-I ${GTEST_HEADER_DIR}" --cuda-include-path="${CUDA_HEADER_DIR}" --format-range=none ${OUTPUT_FILE} --out-root=${OUTPUT_FOLDER}
+
```
tcojean, line 283: Maybe add a comment or separate this clearly? You could extract all of this
into multiple scripts I think, one with preprocessing, one with post
processing, and clearly show the call to DPCT.
```############################################
# Call DPCT on the previously generated file.
############################################```

File: dev_tools/oneapi/convert_source.sh [hunk 43]
```diff
@@ -0,0 +1,429 @@
...
+# dim3_t -> dim3
+# this_thread_block_t -> this_thread_block
+# tiled_partition_t -> tiled_partition
+# thread_id
+# cooperative_group
+# replace_regex="1 i\#include <CL/sycl.hpp>"
+# replace_regex="s/cuda/dpcpp/g"
+# replace_regex="${replace_regex};s/#include \\\"common/#include \\\"dpcpp_code/g"
+replace_regex="s/dim3_t/dim3/g"
+replace_regex="${replace_regex};s/trick\/dim3\.hpp/dpcpp\/base\/dim3\.dp\.hpp/g"
```
tcojean, line 309: I think the previous issues should be documented somewhere,
in a markdown file or something. Maybe detailed a little if you have an
example of failures?

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+replace_regex="${replace_regex};s/this_thread_block_t/this_thread_block/g"
+replace_regex="${replace_regex};s/thread_t::/thread::/g"
```
tcojean, line 311: ```replace_regex="${replace_regex};s:trick/dim3\.hpp:dpcpp/base/dim3\.dp\.hpp:g"```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+# echo "global_range ${global_range}"
+# local_range="sycl::range<3>(dpct_local_range.get(2), dpct_local_range.get(1), dpct_local_range.get(0))"
+# local_range=$(convert_regex_allowed "${local_range}")
+# echo "local_range ${local_range}"
+# replace_regex="${replace_regex};s/${global_range}/global_range/g"
+# replace_regex="${replace_regex};s/${local_range}/local_range/g"
+replace_regex="${replace_regex};s/tiled_partition_t/tiled_partition/g"
+replace_regex="${replace_regex};s/trick\/thread_ids.hpp/dpcpp\/components\/thread_ids.dp.hpp/g"
+replace_regex="${replace_regex};s/trick\/cooperative_groups.hpp/dpcpp\/components\/cooperative_groups.dp.hpp/g"
+replace_regex="${replace_regex};s/trick\/sorting.hpp/dpcpp\/components\/sorting.dp.hpp/g"
```
tcojean, line 330: ```replace_regex="${replace_regex};s:trick/thread_ids.hpp:dpcpp/components/thread_ids.dp.hpp:g"```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+replace_regex="${replace_regex};s/trick\/reduction.hpp/dpcpp\/components\/reduction.dp.hpp/g"
```
tcojean, line 331: ```replace_regex="${replace_regex};s:trick/cooperative_groups.hpp:dpcpp/components/cooperative_groups.dp.hpp:g"```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+replace_regex="${replace_regex};s/#define GET_QUEUE 0//g"
```
tcojean, line 332: ```replace_regex="${replace_regex};s:trick/sorting.hpp:dpcpp/components/sorting.dp.hpp:g"```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+replace_regex="${replace_regex};s/GET_QUEUE/exec->get_queue()/g"
```
tcojean, line 333: ```replace_regex="${replace_regex};s:trick/reduction.hpp:dpcpp/components/reduction.dp.hpp:g"```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+replace_regex="${replace_regex};s/cuda/dpcpp/g"
+replace_regex="${replace_regex};s/Cuda/Dpcpp/g"
+replace_regex="${replace_regex};s/CUDA/DPCPP/g"
+replace_regex="${replace_regex};s/\.cuh/\.dp\.hpp/g"
+replace_regex="${replace_regex};s/\.cu/\.dp\.cpp/g"
+replace_regex="${replace_regex};s/${SPECIAL_SUFFIX}//g"
+replace_regex="${replace_regex};s/#include <dpct\/dpct\.hpp>//g"
+replace_regex="${replace_regex};s/#include \"dpcpp\/base\/types\.hpp\"//g"
+replace_regex="${replace_regex};s/#include \"dpcpp\/test\/utils\.hpp\"/#include \"core\/test\/utils\.hpp\"/g"
+# remove as_dpcpp_type\(content\) -> content
```
tcojean, line 343: ```replace_regex="${replace_regex};s:#include \"dpcpp/base/types\.hpp\"::g"```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+# If the content use some brackets, only allowed one nested bracket now.
```
tcojean, line 344: ```replace_regex="${replace_regex};s:#include \"dpcpp/test/utils\.hpp\":#include \"core/test/utils\.hpp\":g"```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+# echo "replace_regex ${quote}${replace_regex}${quote}"
+sed -i -E "${replace_regex}" "${dpct_file}"
+
+# auto dpct_global_range = grid * block;
+# auto dpct_local_range = block;
+# sycl::nd_range<3>(sycl::range<3>(dpct_global_range.get(2), dpct_global_range.get(1), dpct_global_range.get(0)), sycl::range<3>(dpct_local_range.get(2), dpct_local_range.get(1), dpct_local_range.get(0))),
+
+# check whether containing __dpct_inline__ or __dpct_align__
+need_dpct=$(grep -Eq "__(dpct_align|dpct_inline)__" ${dpct_file})
+if grep -Eq "__(dpct_align|dpct_inline)__" ${dpct_file}; then
```
tcojean, line 384: ```# Include dpct.hpp if __dpct_inline__ or __dpct_align__ are found```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+    # Found
+    sed -i '1 i#include "dpcpp/base/dpct.hpp"' ${dpct_file}
+fi
+
+# extract device_code
+if [ "${EXTRACT_KERNEL}" = "true" ]; then
+    IFS=';' read -ra individual_deivce <<< "${DEVICE_FILE}"
+    for variable in "${individual_deivce[@]}"; do
+        device_regex=$(convert_regex_allowed "${variable}")
```
tcojean, line 393: ```    IFS=';' read -ra individual_device <<< "${DEVICE_FILE}" ```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+        # echo "device_regex ${device_regex}"
```
tcojean, line 394: ```    for variable in "${individual_device[@]}"; do```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+        # echo "dpct_device_file ${dpct_device_file}"
+        cat ${dpct_file} | sed -n "/${device_regex} - start/,/${device_regex} - end/p" | sed "1d;\$d" > ${dpct_device_file}
+        sed -i "/${device_regex} - start/,/${device_regex} - end/d;s~// *#include \"${device_regex}\"~#include \"${dpct_device_path}\"~g" ${dpct_file}
+        dpct_dir=$(dirname "${dpct_device_path}")
+        mkdir -p "${ROOT_DIR}/${dpct_dir}"
+        cp "${dpct_device_file}" "${ROOT_DIR}/${dpct_device_path}"
+    done
+fi
+
+target_file=$(echo "${input}" | sed 's/cuda\//dpcpp\//g;s/\.cuh/\.dp\.hpp/g;s/\.cu/\.dp\.cpp/g')
```
tcojean, line 408: I fail to see for now, what is this for?

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+target_dir=$(dirname "${target_file}")
```
tcojean, line 409: Maybe a comment?
```# Integrate the new DPC++ file into the correct place```

File: dev_tools/oneapi/convert_source.sh [hunk 56]
```diff
@@ -0,0 +1,429 @@
...
+
```
tcojean, line 410: ```target_file=$(echo "${input}" | sed 's:cuda/:dpcpp/:g;s:\.cuh:\.dp\.hpp:g;s:\.cu:\.dp\.cpp:g')```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+convert_syntax() {
+    local syntax_regex="([^<>]*)(<[^<>]*>)?<<<(.*)>>>(.*)"
+    local str="$1"
+    str=$(echo "${str}" | sed -E 's/ +/ /g')
+    # echo "str ${str}"
+    local temp=""
+    local num=0
+    local var=""
+    if [[ "${str}" =~ ${syntax_regex} ]]; then
+        content="${BASH_REMATCH[3]}"
```
tcojean, line 73: ```        local content="${BASH_REMATCH[3]}"```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+        # echo "// content ${content}"
+        IFS=',' read -ra par <<< "$content"
+        for variable in "${par[@]}"; do
+            if [ -n "${temp}" ]; then
+                temp="${temp},"
+            fi
+            temp="${temp}${variable}"
+            # echo "// temp ${temp}"
+            is_closed=$(check_closed "$temp")
```
tcojean, line 82: I think a comment is warranted here:
```# Do not fall into the trap of counting parameters in a sub-expression, such as a function call inside the CUDA kernel parameters```

File: dev_tools/oneapi/convert_source.sh [hunk 59]
```diff
@@ -0,0 +1,429 @@
...
+                    num=$((num-1))
+                else
+                    if [ -n "${var}" ]; then
+                        var="${var},"
+                    fi
+                    var="${var}${temp}"
+                fi
+                temp=""
+            fi
+        done
```
tcojean, line 97: If I'm correct, all you do in the previous loop is to count the number of
variables in the list, and you reconstruct it all in `$var`? In that case,
isn't it possible to use three regex on `$content` to cleanup the annoying
parts before counting? Like:
```
temp=${content//\[*\]/}
temp=${temp//\(*\)/}
temp=${temp//<*>/}
num=$((${#temp//[^,]} + 1))
# You can also do checks for when num == 4
```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+            if [[ ${num} -eq 3 ]]; then
+                # var="${var}, exec->get_queue()"
+                var="${var}, GET_QUEUE"
+                num=$((num+1))
+            fi
+        fi
+        local suffix=""
+        local function_name=$(echo "${BASH_REMATCH[1]}" | sed -E 's/(.*::)| //g')
+        # echo "// functioname ${function_name}"
+        MAP_FILE="map_list"
```
tcojean, line 116: This seems to come from somewhere else, maybe put it as a global variable or
function variable?

dev_tools/oneapi/convert_source.sh
tcojean, line 118: ```        local suffix_matches=$(echo "${suffix}" | wc -l)```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+        else
+            echo "${result}"
+        fi
+    fi
+}
+
+
+convert_regex_allowed() {
+    local str="$1"
+    str=$(echo "${str}" | sed -E 's/(\/|\(|\)|\.)/\\\1/g')
```
tcojean, line 141: This one is sort of hard, I would appreciate a comment. Why use `\\\1`? It
looks like you are looking for different patterns using `|`. Would something
like the following work?
``` sed -E 's:([/().]{1)):\\\1/g' ```

File: dev_tools/oneapi/convert_source.sh [hunk 62]
```diff
@@ -0,0 +1,429 @@
...
+    echo "$str"
+}
+
+# Transfer header file to the correct one
```
tcojean, line 145: ```# Transfer header file to the correct location```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+# common -> dpcpp_common
+# cuda -> dpcpp
+# add sycl header
+
+input="$1"
+filename="${ROOT_DIR}/$input"
+if [[ "${VERBOSE}" == 1 ]]; then
+    echo "Porting file ${filename}"
+fi
```
tcojean, line 154: Add Blankline?

dev_tools/oneapi/convert_source.sh
tcojean, line 156: Remove blankline?

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+EMBED_HOST_FILE="embed_host.cu"
+if [[ "${VERBOSE}" == 1 ]]; then
+    echo "the original file ${UNFORMAT_FILE}"
+    echo "the formatted file ${FORMAT_FILE}"
+    echo "collect common/*.inc in file ${EMBED_FILE}"
+    echo "add autohost func in file ${EMBED_HOST_FILE}"
+    echo "convert original CUDA call in file ${OUTPUT_FILE}, which is the file for dpct"
+fi
+rm "${OUTPUT_FILE}"
+# echo "#include \"trick/dim3.hpp\"" >> "${OUTPUT_FILE}"
```
tcojean, line 181: It's the same in other places, if lines are not needed anymore it's better to remove them.

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+echo "#define GET_QUEUE 0" >> "${OUTPUT_FILE}"
+# add empty ginkgo license such that format_header recognize some header before header def macro
+echo "/*${GINKGO_LICENSE_BEACON}" >> "${OUTPUT_FILE}"
+echo "${GINKGO_LICENSE_BEACON}*/" >> "${OUTPUT_FILE}"
+rm "${GLOBAL_FILE}"
+rm "${EMBED_FILE}"
+while IFS='' read -r line; do
```
tcojean, line 188: Maybe a comment here?
```     # Embed shared kernels into ${EMBED_FILE}```

File: dev_tools/oneapi/convert_source.sh [hunk 66]
```diff
@@ -0,0 +1,429 @@
...
+        cat "${ROOT_DIR}/${device_file}" >> "${EMBED_FILE}"
+        [ "${EXTRACT_KERNEL}" == "true" ] && echo "/**** ${device_file} - end ****/" >> "${EMBED_FILE}"
+        if [ -n "${DEVICE_FILE}" ]; then
+            DEVICE_FILE="${DEVICE_FILE};"
+        fi
+        DEVICE_FILE="${DEVICE_FILE}${device_file}"
+    else
+        echo "${line}" >> "${EMBED_FILE}"
+    fi
+done < "${UNFORMAT_FILE}"
```
tcojean, line 203: Maybe a comment here?
```# Call clang-format for better formatting. ```

File: dev_tools/oneapi/convert_source.sh [hunk 67]
```diff
@@ -0,0 +1,429 @@
...
+${CLANG_FORMAT} -style=file "${EMBED_FILE}" > "${FORMAT_FILE}"
```
tcojean, line 204: Maybe a newline and a comment here (correct it if I'm wrong)?
```# Add an extra host function so that the converted DPC++ code will look like CUDA.```

File: dev_tools/oneapi/convert_source.sh (hunk 33 above)
Comment (tcojean) [inline dev_tools/oneapi/convert_source.sh:206]: Maybe a comment?
```# Convert kernel calls and manage DPC++ config based kernel generation.```

File: dev_tools/oneapi/convert_source.sh [hunk 68]
```diff
@@ -0,0 +1,429 @@
...
+# thread_id.cuh -> use local
+# cooperative_group.cuh -> use local
+replace_regex="s/dim3/dim3_t/g"
+replace_regex="${replace_regex};s/this_thread_block/this_thread_block_t/g"
+replace_regex="${replace_regex};s/this_grid/this_grid_t/g"
+replace_regex="${replace_regex};s/tiled_partition/tiled_partition_t/g"
+replace_regex="${replace_regex};s/thread::/thread_t::/g"
+replace_regex="${replace_regex};s/bitonic_sort/bitonic_sort_t/g"
+replace_regex="${replace_regex};s/reduction_array/reduction_array_t/g"
+replace_regex="${replace_regex};s/cuda\/components\/thread_ids.cuh/trick\/thread_ids.hpp/g"
```
tcojean, line 250: ```replace_regex="${replace_regex};s:cuda/components/thread_ids.cuh:trick/thread_ids.hpp:g"```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+replace_regex="${replace_regex};s/cuda\/components\/cooperative_groups.cuh/trick\/cooperative_groups.hpp/g"
```
tcojean, line 251: ```replace_regex="${replace_regex};s:cuda/components/cooperative_groups.cuh:trick/cooperative_groups.hpp:g"```

dev_tools/oneapi/convert_source.sh
tcojean, line 252: ```replace_regex="${replace_regex};s:cuda/components/sorting.cuh:trick/sorting.hpp:g"```

dev_tools/oneapi/convert_source.sh
tcojean, line 253: ```replace_regex="${replace_regex};s:cuda/components/reduction.cuh:trick/reduction.hpp:g"```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+replace_regex="${replace_regex};s/cuda\/components\/sorting.cuh/trick\/sorting.hpp/g"
+replace_regex="${replace_regex};s/cuda\/components\/reduction.cuh/trick\/reduction.hpp/g"
+replace_regex="${replace_regex};s/CUH_/DP_HPP_/g"
+replace_regex="${replace_regex};s/(.*template.*)(Config * [a-zA-Z0-9_]*) * = *[a-zA-Z0-9_:]*(\(+[^\)]*\))?/\1\2/g"
+# keep using original xxx.sync(); ->xxx;//.sync();
+replace_regex="${replace_regex};s/(\.sync\(\);)/;\/\/.sync()/g"
```
tcojean, line 257: ```replace_regex="${replace_regex};s|\.sync\(\);|;//.sync()|g"```

dev_tools/oneapi/convert_source.sh
tcojean, line 259: ```replace_regex="${replace_regex};s|(template GKO.*;)|// \1|g"```
Maybe a comment for the next line?
```# Add the trick dim3_t header if dim3 is used.```

File: dev_tools/oneapi/convert_source.sh [hunk 71]
```diff
@@ -0,0 +1,429 @@
...
+if grep -Eq "dim3" ${OUTPUT_FILE}; then
+    # Found
+    sed -i '1 i#include "trick/dim3_t.hpp"' ${OUTPUT_FILE}
+fi
+# add the cooperative group header according to group:: because some sources forget to add it
+if grep -Eq "group::" ${OUTPUT_FILE}; then
+    # Found
+    sed -i '1 i#include "trick/cooperative_groups.hpp"' ${OUTPUT_FILE}
+fi
+# exit 0
```
tcojean, line 270: Can be removed

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+OUTPUT_FOLDER="output"
+if [[ "${VERBOSE}" == 1]]; then
+    echo "The dpct calling:"
+    echo "dpct --extra-arg=\"-std=c++14\" --extra-arg=\"-I ${ROOT_DIR}\" --extra-arg=\"-I ${ROOT_DIR}/include\" --extra-arg=\"-I ${ROOT_DIR}/dev_tools/oneapi\" --extra-arg=\"-I ${GTEST_HEADER_DIR}\" --cuda-include-path=\"${CUDA_HEADER_DIR}\" --format-range=none ${OUTPUT_FILE} --out-root=${OUTPUT_FOLDER}"
+fi
+
+# Delete output/source.cpp
+rm "${OUTPUT_FOLDER}/${OUTPUT_FILE}"
+rm "${OUTPUT_FOLDER}/${OUTPUT_FILE}.dp.cpp"
+
```
tcojean, line 281: Maybe add a comment or separate this clearly? You could extract all of this
into multiple scripts I think, one with preprocessing, one with post
processing, and clearly show the call to DPCT.
```############################################
# Call DPCT on the previously generated file.
############################################```

File: dev_tools/oneapi/convert_source.sh [hunk 73]
```diff
@@ -0,0 +1,429 @@
...
+fi
+# global reverse fix
+# dim3_t -> dim3
+# this_thread_block_t -> this_thread_block
+# tiled_partition_t -> tiled_partition
+# thread_id
+# cooperative_group
+# replace_regex="1 i\#include <CL/sycl.hpp>"
+# replace_regex="s/cuda/dpcpp/g"
+# replace_regex="${replace_regex};s/#include \\\"common/#include \\\"dpcpp_code/g"
```
tcojean, line 307: I think the previous issues should be documented somewhere,
in a markdown file or something. Maybe detailed a little if you have an
example of failures?

dev_tools/oneapi/convert_source.sh
tcojean, line 309: ```replace_regex="${replace_regex};s:trick/dim3\.hpp:dpcpp/base/dim3\.dp\.hpp:g"```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+# global_range="sycl::range<3>(dpct_global_range.get(2), dpct_global_range.get(1), dpct_global_range.get(0))"
+# global_range=$(convert_regex_allowed "${global_range}")
+# echo "global_range ${global_range}"
+# local_range="sycl::range<3>(dpct_local_range.get(2), dpct_local_range.get(1), dpct_local_range.get(0))"
+# local_range=$(convert_regex_allowed "${local_range}")
+# echo "local_range ${local_range}"
+# replace_regex="${replace_regex};s/${global_range}/global_range/g"
+# replace_regex="${replace_regex};s/${local_range}/local_range/g"
+replace_regex="${replace_regex};s/tiled_partition_t/tiled_partition/g"
+replace_regex="${replace_regex};s/trick\/thread_ids.hpp/dpcpp\/components\/thread_ids.dp.hpp/g"
```
tcojean, line 328: ```replace_regex="${replace_regex};s:trick/thread_ids.hpp:dpcpp/components/thread_ids.dp.hpp:g"```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+replace_regex="${replace_regex};s/trick\/cooperative_groups.hpp/dpcpp\/components\/cooperative_groups.dp.hpp/g"
```
tcojean, line 329: ```replace_regex="${replace_regex};s:trick/cooperative_groups.hpp:dpcpp/components/cooperative_groups.dp.hpp:g"```

dev_tools/oneapi/convert_source.sh
tcojean, line 330: ```replace_regex="${replace_regex};s:trick/sorting.hpp:dpcpp/components/sorting.dp.hpp:g"```

dev_tools/oneapi/convert_source.sh
tcojean, line 331: ```replace_regex="${replace_regex};s:trick/reduction.hpp:dpcpp/components/reduction.dp.hpp:g"```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+replace_regex="${replace_regex};s/#define GET_QUEUE 0//g"
+replace_regex="${replace_regex};s/GET_QUEUE/exec->get_queue()/g"
+replace_regex="${replace_regex};s/cuda/dpcpp/g"
+replace_regex="${replace_regex};s/Cuda/Dpcpp/g"
+replace_regex="${replace_regex};s/CUDA/DPCPP/g"
+replace_regex="${replace_regex};s/\.cuh/\.dp\.hpp/g"
+replace_regex="${replace_regex};s/\.cu/\.dp\.cpp/g"
+replace_regex="${replace_regex};s/${SPECIAL_SUFFIX}//g"
+replace_regex="${replace_regex};s/#include <dpct\/dpct\.hpp>//g"
+replace_regex="${replace_regex};s/#include \"dpcpp\/base\/types\.hpp\"//g"
```
tcojean, line 341: ```replace_regex="${replace_regex};s:#include \"dpcpp/base/types\.hpp\"::g"```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+replace_regex="${replace_regex};s/#include \"dpcpp\/test\/utils\.hpp\"/#include \"core\/test\/utils\.hpp\"/g"
```
tcojean, line 342: ```replace_regex="${replace_regex};s:#include \"dpcpp/test/utils\.hpp\":#include \"core/test/utils\.hpp\":g"```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+
+quote="'"
+# echo "replace_regex ${quote}${replace_regex}${quote}"
+sed -i -E "${replace_regex}" "${dpct_file}"
+
+# auto dpct_global_range = grid * block;
+# auto dpct_local_range = block;
+# sycl::nd_range<3>(sycl::range<3>(dpct_global_range.get(2), dpct_global_range.get(1), dpct_global_range.get(0)), sycl::range<3>(dpct_local_range.get(2), dpct_local_range.get(1), dpct_local_range.get(0))),
+
+# check whether containing __dpct_inline__ or __dpct_align__
```
tcojean, line 382: ```# Include dpct.hpp if __dpct_inline__ or __dpct_align__ are found```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+need_dpct=$(grep -Eq "__(dpct_align|dpct_inline)__" ${dpct_file})
+if grep -Eq "__(dpct_align|dpct_inline)__" ${dpct_file}; then
+    # Found
+    sed -i '1 i#include "dpcpp/base/dpct.hpp"' ${dpct_file}
+fi
+
+# extract device_code
+if [ "${EXTRACT_KERNEL}" = "true" ]; then
+    IFS=';' read -ra individual_deivce <<< "${DEVICE_FILE}"
```
tcojean, line 391: ```    IFS=';' read -ra individual_device <<< "${DEVICE_FILE}" ```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+    for variable in "${individual_deivce[@]}"; do
```
tcojean, line 392: ```    for variable in "${individual_device[@]}"; do```

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+        dpct_device_file="output/${dpct_device_file}"
+        # echo "dpct_device_path ${dpct_device_path}"
+        # echo "dpct_device_file ${dpct_device_file}"
+        cat ${dpct_file} | sed -n "/${device_regex} - start/,/${device_regex} - end/p" | sed "1d;\$d" > ${dpct_device_file}
+        sed -i "/${device_regex} - start/,/${device_regex} - end/d;s~// *#include \"${device_regex}\"~#include \"${dpct_device_path}\"~g" ${dpct_file}
+        dpct_dir=$(dirname "${dpct_device_path}")
+        mkdir -p "${ROOT_DIR}/${dpct_dir}"
+        cp "${dpct_device_file}" "${ROOT_DIR}/${dpct_device_path}"
+    done
+fi
```
tcojean, line 406: I fail to see for now, what is this for?

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,429 @@
...
+
```
tcojean, line 407: Maybe a comment?
```# Integrate the new DPC++ file into the correct place```

File: dev_tools/oneapi/convert_source.sh (hunk 54 above)
Comment (tcojean) [inline dev_tools/oneapi/convert_source.sh:408]: ```target_file=$(echo "${input}" | sed 's:cuda/:dpcpp/:g;s:\.cuh:\.dp\.hpp:g;s:\.cu:\.dp\.cpp:g')```

**Review by tcojean — COMMENTED** [review]
Continued

File: dev_tools/oneapi/add_host_function.sh (hunk 1 above)
Comment (yhmtsai) [reply dev_tools/oneapi/add_host_function.sh:7]: Yes, but I think the template will not have this case?

File: dev_tools/oneapi/add_host_function.sh [hunk 83]
```diff
@@ -0,0 +1,305 @@
...
+        temp="${temp}${var}"
+        # only handle one pair <> currently
+        if [[ "${temp}" =~ "<" ]] && [[ ! "${temp}" =~ ">" ]]; then
+            continue
+        fi
+        # If the string contains typename, do not extract it.
+        # It should automatically be decided from argument
+        # Also need to ignore = ...
+        if [[ "${temp}" =~ "typename" ]]; then
+            :
```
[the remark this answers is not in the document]
yhmtsai: no, I need to reset the temp to empty

dev_tools/oneapi/add_host_function.sh
```diff
@@ -0,0 +1,305 @@
...
+    str="${str//->}"
+    str_start="${str//[^(<\[]}"
+    str_end="${str//[^>)\]]}"
+    if [[ "${#str_start}" -eq "${#str_end}" ]]; then
+        echo "true"
+    else
+        echo "false"
+    fi
+}
+CONFIG_SELECTION_SUFFIX="_CONFIG"
```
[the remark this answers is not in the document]
yhmtsai: the situation is complex, and we do not add the config to cuda.
Thus, I will delete the adding Config part first.

dev_tools/oneapi/add_host_function.sh
```diff
@@ -0,0 +1,305 @@
...
+                        parameter_name=", ${parameter_name}"
+                    fi
+                    list_list[${identification}]="${list_list[${identification}]}${parameter_name}"
+                else
+                    echo "static_assert(false, \"Need to follow bool, int, size_type, type ordering\");"
+                fi
+            fi
+        fi
+        temp=""
+    done
```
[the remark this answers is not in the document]
yhmtsai: no, it is to match the config selection input arg (template)
the original function is not changed but the config call contains value_list(bool, ...), value_list(int, ...).

dev_tools/oneapi/add_host_function.sh
```diff
@@ -0,0 +1,305 @@
...
+TEMPLATE_REGEX="^ *template <*"
+FUNCTION_START="^ *(template *<|${GLOBAL_KEYWORD}|void)"
+FUNCTION_NAME_END=".*\{.*"
+SCOPE_START="${FUNCTION_NAME_END}"
+SCOPE_END=".*\}.*"
+CHECK_GLOBAL_KEYWORD=".*${GLOBAL_KEYWORD}.*"
+FUNCTION_HANDLE=""
+DURING_FUNCNAME="false"
+ANAYSIS_FUNC=" *(template *<(.*)>)?.* (.*)\((.*)\)"
+START_BLOCK_REX="^( *\/\*| *\/\/)"
```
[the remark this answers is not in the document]
yhmtsai: Yes, it will check whether the same line is end or not.
the `//` comments are start/end together

dev_tools/oneapi/add_host_function.sh
yhmtsai: it print the line in the upper else.
this is to skip those comments and not go into next block for handling add host

dev_tools/oneapi/convert_source.sh
yhmtsai: the working directory is put in the dev_tools/oneapi
It is happened in the script dir. we change directory to script dir in the begining.

dev_tools/oneapi/convert_source.sh
yhmtsai: you are right, it can not handle `call<t<1>, ..><<<>>>()`
should we consider it? it maybe is possible to handle up to one level but arbitrary template is hard or impossible?

dev_tools/oneapi/convert_source.sh
yhmtsai: it needs to escape `/`, right?

dev_tools/oneapi/convert_source.sh
yhmtsai: it is optional and off.
it will move the common code to common/dpcpp_code if enabling it.

dev_tools/oneapi/convert_source.sh
yhmtsai: I still need to escape the `.`, right?
although it may not change the behavior in general

dpcpp/base/config.hpp
```diff
@@ -65,6 +65,11 @@ struct config {
      * to maximize GPU occupancy.
      */
     static constexpr uint32 min_warps_per_block = 4;
+
+    /**
+     * The maximal number of threads allowed in a CUDA warp.
```
pratikvn, line 70: DPCPP is also 256 ?

dpcpp/matrix/ell_kernels.dp.cpp
```diff
@@ -57,12 +72,580 @@ namespace dpcpp {
...
+/**
+ * A compile-time list of sub-warp sizes for which the spmv kernels should be
+ * compiled.
+ * 0 is a special case where it uses a sub-warp size of warp_size in
+ * combination with atomic_adds.
+ */
+using compiled_kernels = syn::value_list<int, 0, 8, 16, 32>;
+
+
+// #include "common/matrix/ell_kernels.hpp.inc"
```
pratikvn, line 109: debug leftover ?

accessor/range.hpp
```diff
@@ -69,7 +69,8 @@ class range {
      *
      * @param params  parameters forwarded to Accessor constructor.
      */
-    template <typename... AccessorParams>
+    template <typename... AccessorParams,
+              std::enable_if_t<(sizeof...(AccessorParams) > 1), bool> = true>
```
pratikvn, line 73: This change would make the `include/ginkgo/core/base/range.hpp` diverge ? Maybe we should either note this somewhere and in 2.0 fix it, or adapt the `include/ginkgo/core/base/range.hpp` as well.

pratikvn (commented):
This is a bit out of my depth, but I just had a cursory look. :)

dpcpp/components/prefix_sum.dp.hpp
```diff
@@ -0,0 +1,254 @@
...
+ * @param subwarp     the cooperative group representing the subwarp.
+ *
+ * @tparam inclusive  if this is true, the computed prefix sum will be
+ *                    inclusive, otherwise it will be exclusive.
+ *
+ * @note For this function to work on architectures with independent thread
+ * scheduling, all threads of the subwarp have to execute it.
+ */
+template <bool inclusive, typename ValueType, typename Group>
+__dpct_inline__ void subwarp_prefix_sum(ValueType element,
```
pratikvn, line 73: Is it possible to move this `__dpct_inline__` to `GKO_INLINE` ?

pratikvn (commented):
Some more minor comments

accessor/range.hpp
yhmtsai: what do you mean diverge here? the usage of constructor keeps the same due to the accessor always require more than 1 args

dev_tools/oneapi/fake_interface/cooperative_groups.cuh
```diff
@@ -0,0 +1,57 @@
...
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+#ifndef GKO_FAKE_INTERFACE_COOPERATIVE_GROUPS_CUH_
+#define GKO_FAKE_INTERFACE_COOPERATIVE_GROUPS_CUH_
+
+#include <cuda/components/cooperative_groups.cuh>
```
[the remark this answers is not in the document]
yhmtsai: In general, yes.
we have cuda code and corresponding dpcpp code but dpct does not allow us to provide custom mapping.
This file only involves dpct process such that dpct has the information on interface but does not try converting real cuda code. In the end, we replace the fake interface to our real dpcpp impl

dev_tools/oneapi/working_directory/.clang-format
```diff
@@ -0,0 +1,108 @@
+---
+Language:        Cpp
```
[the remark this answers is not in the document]
yhmtsai: keep the porting script simple and correct.
I add something to make sure the script be able to catch whole function but there's always some edge case such that I need to add more stuff to handle it.
this is mainly for columnlimit : 1000. template or function name with its args will be in one line.

dpcpp/base/config.hpp
yhmtsai: no, it depends on the device. but 256 is common acceptable block size for dpcpp device.
it is only for pushing dpcpp in the first step.

accessor/range.hpp
pratikvn: I believe this is a duplication of the `include/ginkgo/core/base/range.hpp`, as it was needed to be separate for the Accessor. I meant now the typically the same class (range) has two different versions and we should probably unify them. Also I think you are changing one of the copy constructors below ?

dpcpp/base/config.hpp
pratikvn: Okay, that makes sense. Maybe you can update the doc strings ?

dev_tools/oneapi/working_directory/.clang-format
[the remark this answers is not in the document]
pratikvn: I see. I guess that is okay. Thanks for the clarification.

dev_tools/oneapi/fake_interface/cooperative_groups.cuh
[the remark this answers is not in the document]
pratikvn: But `cuda/components/cooperative_groups.cuh` has cuda syntax, right ? Do you convert/replace that or dpct does that ?

fritzgoebel (approved):
The ELL kernels look good to me. On the rest I think others are better qualified to review

accessor/range.hpp
yhmtsai: Yes, it uses enable_if_t to avoid compiler use this constructor as copy constuctor. By doing so, compiler can auto generate the copy/move constructor, so we do not need to manually point copy constructor to default behavior.
I can also change the part in `include/ginkgo/core/base/range.hpp`

dev_tools/oneapi/fake_interface/cooperative_groups.cuh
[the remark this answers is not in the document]
yhmtsai: dpct has different behavior on system header and local header. moreover, they use keyword to capture cuda function.
when function is in local header, dpct will try to convert them.
but in system header, dpct will only consider it is a function without checking anything.
cooperative group will in their keyword list, so it throws error.
we put the real cuda cooperative group as system header to avoid dpct touching and fake_interface with alternative name to avoid dpct detect.
we also add threadIdx in the function such that dpct can add the nd_item correctly.
in dpct view, it is a normal function with a system header function and threadIdx.
Thus, dpct does not really see cuda/components/cooperative_groups.cuh.
and porting script will replace fake_interface with corresponding dpcpp impl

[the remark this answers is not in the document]
pratikvn: I see. Thanks for the explanation.

accessor/range.hpp
pratikvn: Yeah, I guess we need to be sure that it does not break interface, I would also maybe ping @thoasm for his comments here.

yhmtsai: I do not think it break any interface. Do we have two range class for different purpose?

thoasm: Divergence between the two should not matter much. The accessor in `accessor/` is not considered a public interface, so interface breaking changes in there are fine.
However, I dislike the current change as it prevents any accessor of taking a single argument. I will shortly post a solution to that (basically making sure that the argument given is not a range, but forwarding everything else to the accessor constructor).

yhmtsai: I see. I can also do that if you want

thoasm: Why exactly is it actually necessary that the compiler generates the copy constructor instead of setting it as `= default`?
We could definitively write the condition more specific, so only single range objects are rejected, but to me this feels like a lot of effort for simply avoiding writing `= default` (I would reject the current version, however, because you forbid an accessor with only a single argument).

yhmtsai: for dpcpp usage, I need to use two version copy constructor
1. declare and impl them in the class
```
range(const range &other) = default;
range(range &other) = default;
```
linux is okay if declare and define them in the class
but MSVC is not happy. please see [ref1](https://developercommunity.visualstudio.com/t/false-error-c2580-msvc-doesnt-allow-a-constructor-1/1271214), [ref2](https://developercommunity.visualstudio.com/t/error-c2580-multiple-versions-of-a-defaulted-speci/468431) and some stackoverflow
MSVC only fixes it in /c++latest

2. declare in the class but impl out of the class
```
class range{...};
template <class Accessor>
range<Accessor>(const range &other) = default;
template <class Accessor>
range<Accessor>(range &other) = default;
```
the workaround is mentioned in MSVC such that use two default on copy constructor.
However, dpcpp can not recognize the copy constructor is default behavior, so dpcpp will gives an error about the class is not trivial ...

3. make all copy constructor generated by compiler
the current one is implemented by this way. Avoid the templated constructor is seen as copy constructor such that compiler can automatically handle it.
MSVC and DPCPP are happy with this.

upsj: Can you give a bit of background why we need a copy constructor and a mutable lvalue ref constructor?

yhmtsai: disabling Outer(const Outer&) 
```
error: no matching constructor for initialization of 'Outer'
       run([&]() { [=]() { func(a); }; });
                                ^
candidate constructor not viable: 1st argument ('const Outer') would lose const qualifier
```
disabling Outer(Outer&) 
```
note: in instantiation of function template specialization 'Outer::Outer<Outer &>' requested here
  run([&]() { [=]() { func(a); }; });
```
this is simple version for emulating the dpcpp submit the job.
As far as I remember, two default impl out of class is okay in C++ code, but dpcpp can not allow it.

upsj: Can you post a longer snippet that contains the entire context? Why do you need to modify the element you are copying in a copy constructor?

thoasm: Right, we talked about something like that previously.
If I remember correctly, the problem is that you need both `const range &` and `range &` for them to be used, otherwise, for a `range &`, the forwarding constructor is taken. Additionally, for some reason, the visual studio compiler does not accept both copy constructors `range(const range &) = default` and `range(range &) = default` (apparently, implementing them instead of `=default` does not help as well).
Was that about right, @yhmtsai?
I would prefer to find an alternative solution and will try something tomorrow, otherwise, we can add the SFINAE work-around to only exclude single `range` parameter for this constructor.

dev_tools/oneapi/convert_source.sh
```diff
@@ -0,0 +1,374 @@
...
+# EnvironementSet ./dev_tools/oneapi/convert_source.sh <the_file_in_cuda>
+# <the_file_in_cuda> can be .hpp/.cpp/.cu/.cuh
+
+# the following are parameters set by environment variables
+# REQUIRED:
+#   CUDA_HEADER_DIR: contains the cuda headers
+# OPTIONAL:
+#   ROOT_DIR: the ginkgo folder. The default is current path
+#   BUILD_DIR: the ginkgo build folder, which needs cmake to generate config.hpp and gtest include. The default is "build"
+#              Note. It requires GINKGO_BUILD_TESTS=ON to download gtest but it is not required to compile ginkgo.
```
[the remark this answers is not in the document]
yhmtsai: which indentation? I use two space here

accessor/range.hpp
```diff
@@ -42,6 +42,29 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+struct check_if_valid : public std::true_type {};
+
+/**
+ * check_if_valid gives false if the decay type of input is the same type as
+ * Ref.
+ *
+ * @tparam Ref  the reference type
+ */
+template <typename Ref>
+struct check_if_valid<Ref, Ref> : public std::false_type {};
```
thoasm, line 64: Please move this to the private section of `range<Accessor>`. Currently, it pollutes the namespace and is not descriptive enough to know what it does.
Also, I just realized that `check_if_equal` might be a better name. Anyway, it must either be moved to the `range` class or into a `detail` namespace (with strong preference to the `range` class).

accessor/range.hpp
```diff
@@ -66,14 +89,16 @@ class range {
      * Creates a new range.
      *
      * @tparam AccessorParam  types of parameters forwarded to the accessor
-     *                        constructor
+     *                        constructor. It does not allow range itself.
```
thoasm, line 92: I would remove that comment and write:
```
@note  We use SFINAE to allow for a default copy and move constructor to be generated,
       so a `range` is trivially copyable if the `Accessor` is trivially copyable.
```
After that, I am fine with keeping this without checking for Visual Studio compiler.

thoasm (changes requested):
I would prefer to move the `check_if_valid` helper structure.

dpcpp/matrix/ell_kernels.dp.cpp
```diff
@@ -57,12 +72,567 @@ namespace dpcpp {
...
+    UninitializedArray<OutputValueType,
+                       default_block_size / num_thread_per_worker> *storage)
+{
+    const OutputValueType alpha_val = alpha(0);
+    const OutputValueType beta_val = beta[0];
+    // Because the atomic operation changes the values of c during computation,
+    // it can not do the right alpha * a * b + beta * c operation.
+    // Thus, the dpcpp kernel only computes alpha * a * b when it uses atomic
+    // operation.
+    if (atomic) {
```
upsj, line 250: Maybe instead note that `c` needs to be scaled beforehand, otherwise this might lead to confusion. Also:
```suggestion
    // Because the atomic operation changes the values of c during computation,
    // it can not do the right alpha * a * b + beta * c operation.
    // Thus, the dpcpp kernel only adds alpha * a * b when it uses atomic
    // operation.
    if (atomic) {
```

upsj (approved):
LGTM! Does it make sense to keep the conversion scripts in-tree in the long run? We're not maintaining them, and they will not be used after we finished porting.

dev_tools/oneapi/add_host_function.sh
```diff
@@ -0,0 +1,176 @@
+#!/bin/bash
+
+# add_host_function adds a host function to wrap the cuda kernel call with template and parameter configuration.
+#
+# For example
```
pratikvn, line 5: Thanks! This looks good!

dev_tools/oneapi/convert_source.sh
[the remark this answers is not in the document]
pratikvn: I mean of the `Note. ...` Its only a minor thing and you can leave it as it is.

pratikvn (approved):
LGTM! Thanks for adding the documentation.

accessor/range.hpp
yhmtsai: should I also change the ginkgo range? or changing this for accessor should be enough

thoasm: If we don't have a problem with the public `range`, I would leave it as it is for now. Should we encounter the same problem there (I assume we would as soon as we use it the same way we use `acc::range`), we can simply copy it from `acc::range`.

dpcpp/matrix/ell_kernels.dp.cpp
yhmtsai: I will move this part into the atomic scope and the following is what I use (including c be scaled)
```
// Because the atomic operation changes the values of c during
// computation, it can not directly do alpha * a * b + beta * c
// operation. The beta * c needs to be done before calling this kernel.
// Then, this kernel only adds alpha * a * b when it uses atomic
// operation.
```

accessor/range.hpp
yhmtsai: I see

yhmtsai:
@upsj I think we put it here like cuda2hip. it will be hard to use on adaptive changes.
implement cuda -> apply cuda2dpcpp -> delete those porting on unchanged function -> fix the porting
maybe some script to only extract the changed function or selection help this situtation

PR #781, opened 2021-06-02
Run the CUDA tests with a few configurations

tcojean:
Use `amdci` to run some CUDA tests.

Thanks to the new card, we can use `amdci` to run some CUDA tests.
In order to not overload the machines, I only use it with CUDA 11.0 for now.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -64,7 +64,6 @@ include:
     - if [ -n "${CUDA_ARCH}" ]; then
       CUDA_ARCH_STR=-DGINKGO_CUDA_ARCHITECTURES=${CUDA_ARCH};
       CUDA_HOST_STR=-DCMAKE_CUDA_HOST_COMPILER=$(which ${CXX_COMPILER});
-      export CUDA_VISIBLE_DEVICES=$((RANDOM % 2));
       fi
     - cmake ${CI_PROJECT_DIR}${CI_PROJECT_DIR_SUFFIX}
         -GNinja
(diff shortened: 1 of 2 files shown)


upsj (approved):

yhmtsai (approved):
CUDA_ARCH should be 61?
or, we can also add this argument to image but it needs to handle the CUDA_VISIBLE_DEVICES carefully.

pratikvn (approved):
LGTM!

thoasm (approved):
LGTM!

tcojean:
I was a bit too quick in the setup on the machine. We are missing the nvidia-container runtime, thus the GPUs are not correctly passed to the containers.

yhmtsai:
in cuda/hip container, do they both only see only one device?
If so, we may introduce the device_number inside the image such that it uses the number to set the visible device.
otherwise, 
```
if [ -n "${CUDA_ARCH}" ]; then
    CUDA_ARCH_STR=-DGINKGO_CUDA_ARCHITECTURES=${CUDA_ARCH};
    CUDA_HOST_STR=-DCMAKE_CUDA_HOST_COMPILER=$(which ${CXX_COMPILER});
    export CUDA_VISIBLE_DEVICES=$((RANDOM % 2));
fi
```
sometimes it will fail.

upsj:
Since it's likely we will soon move to another system, can we just remove the CUDA device randomization altogether?

yhmtsai:
Will the old machine with 2 gpus not come back?
Sure, we can delete them now.
if we need them again, we can add more general rules for setting it.

PR #783, opened 2021-06-04
Fix no-circular-deps header compilation for CUDA

### LINKED ISSUE #782: Build Failure with GINKGO_CHECK_CIRCULAR_DEPS enabled
Building the default target with `GINKGO_CHECK_CIRCULAR_DEPS=ON` results in a compilation error. More specifically, the target `ginkgo_cuda_headers` fails with the error message:

```
ginkgo/cuda/base/math.hpp:40:10: fatal error: thrust/complex.h: No such file or directory
   40 | #include <thrust/complex.h>
      |          ^~~~~~~~~~~~~~~~~~
```

My guess is that this is caused by missing includes from cuda. Checking the verbose make output shows that the default C++ compiler is used and no include path for cuda are given.

upsj:
This change ensures that all CUDA header files with NO_CIRCULAR_DEPS enabled are being compiled using `nvcc`

Fixes #782


yhmtsai:
or we can glob the `CUDA_HEADERS` from `cuda/*` not `*.cuh`?

upsj:
@yhmtsai Unfortunately that doesn't work, since the paths are relative to `cuda`. We could alternatively add a separate function for CUDA

Slaedr:
Just out of curiosity, why do we compile these header-only objects? To be able to catch some obvious compile-time errors independent of usage (to the extent that anything is actually compiled)?

upsj:
@Slaedr We want to ensure that all of our header files are self-sufficient, i.e. including any of them in any file doesn't cause build issues (excluding really esoteric scenarios)

thoasm (approved):
LGTM

tcojean (approved):

upsj:
rebase!

yhmtsai (approved):
LGTM.

PR #785, opened 2021-06-05
Use another toolset of msvc in cuda job

yhmtsai:
This PR uses `-T version=14.25` to use another one toolset in github action.
Also, pass the toolset selection into other cmake command if it exists.
It avoid the error from the latest compiler from 16.10 with cuda 11. [ref](https://github.com/actions/virtual-environments/issues/3485) 
Github action install the toolset 14.25 from this [issue](https://github.com/actions/virtual-environments/issues/1076)  
Another workaround is to use standard 17 not 14.

### DIFF
--- a/.github/workflows/windows-msvc-cuda.yml
+++ b/.github/workflows/windows-msvc-cuda.yml
@@ -45,5 +45,5 @@ jobs:
         refreshenv
         mkdir build
         cd build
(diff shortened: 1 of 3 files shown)


yhmtsai:
rebase!

upsj (approved):

tcojean (approved):

PR #786, opened 2021-06-06
Simplify third-party integration

upsj:
inspired by the changes in #785 and some thoughts on packaging and dependency management:
Right now, we are building our third-party dependencies during configure time, manually pass all CMake parameters and construct targets out of the generated libraries manually.
Since all dependencies except for `hwloc` are only used inside Ginkgo, I think it might make more sense to build them as subdirectories, since then all CMake-specific flags are forwarded directly and consistently.

This changes the behavior of googletest slightly, making it use shared libraries by default, but otherwise everything should stay the same.

Another thing that we might want to address here is that currently, compiling Ginkgo (even without tests, benchmarks) on CUDA systems without internet connectivity fails, since we try to pull in CAS. Since we only ever need a single file from that repository, why don't we include it directly in our source instead, and set up a sync action that either syncs CAS from our repo or creates a PR if CAS was changed?

Finally, Ginkgo 1.4.0 is the first release with an actual library dependency exceeding vendor libraries (`hwloc`), so we should check whether the current design is future proof (also thinking of library separation and #715), since we don't seem to be using any hwloc functionality yet, right now. It looks to me like the functionality is most useful for `GenericExecutor` #704 and MPI support #760 ?

- [x] Add a CI job checking compatibility with pre-installed dependencies
- [x] Consider vendoring CAS
- [x] ~~Consider improving separation of `hwloc` component~~ I think I misunderstood some details there, the separation looks fine


yhmtsai:
When looking into ExternalProject, I think we do not use ExternalProject fully in ginkgo.
Can FetchContent set the config type of MSVC?
MSVC in linking will require they use same crt in debug mode (with d) or others (without d). or can it mix them together?

upsj:
Since everything is built as part of the same CMake workflow, everything uses the same config. The only issue is GTest, which overwrites the existing CRT flags. Just in general, their CMake setup is a bit of a mess.

third_party/gtest/CMakeLists.txt
```diff
@@ -1,21 +1,21 @@
...
+    include(FetchContent)
+    FetchContent_Declare(
+        googletest
+        GIT_REPOSITORY https://github.com/google/googletest.git
+        GIT_TAG        6a7ed316a5cdc07b6d26362c90770787513822d4
+    )
+    set(gtest_disable_pthreads ON)
+    if(MSVC)
+        set(gtest_force_shared_crt ON)
+    endif()
```
yhmtsai, line 12: should it be `googletest_gtest_disable_pthreads` and `googletest_force_shared_crt`?

upsj: Where does that prefix come from?

yhmtsai: googletest  is from the name declaration of `FetchContent_Declare`
I found it from the example, so I guess they follow this way to distinguish the variable from different project? 
image

upsj: that is the actual name of the variable in protobuf: https://github.com/protocolbuffers/protobuf/blob/a1d8f8f96900468f09ced2ce9d23dea5a1bc070c/cmake/CMakeLists.txt#L273

yhmtsai: Ah, I see. Sorry for wrong statement.

yhmtsai:
the google test does not use the value from ginkgo due to CMP0077?
From this https://cmake.org/pipermail/cmake/2019-March/069206.html and also CMP0077 is shown in the log.
make the variable be in the CACHE or to set CMP0077 outside to affect googletest CMakeList?

upsj:
I added CAS directly to the repository and a sync step with the CAS repository. You can see it working in my forks https://github.com/upsj/ginkgo-bot-playground and https://github.com/upsj/CudaArchitectureSelector

yhmtsai:
Could it be also done by github submodule?

upsj:
git submodules have the same issue as FetchContent: You need an internet connection, plus any change in the submodule requires a change in the main repo as well. This way, we make it easier to change CAS and have a self-contained library for builds that don't enable tests and benchmarks.

yhmtsai:
the update from CAS to ginkgo may be true.
but I think the internet somehow depends the download ginkgo in the beginning.
you can use `git clone --recurse-submodules` to download them together.

upsj:
That doesn't work if you download Ginkgo as a tarball (as many package managers do). I have gotten the impression that git submodules are frowned upon for dependency management (and found them overly complicated myself), and since we control this dependency completely, I think it might be easier to vendor it directly.

yhmtsai:
I see. one nit for the update, could you also add the commit sha of CAS into playground? like you have the commit sha of playground inside CAS

upsj:
After some discussion with @yhmtsai I decided to keep the CAS sync out of the library for now.

core/test/accessor/CMakeLists.txt
```diff
@@ -14,7 +14,9 @@ function(create_accessor_test test_name)
     set_target_properties("${TEST_TARGET_NAME}" PROPERTIES
         OUTPUT_NAME "${test_name}")
     target_link_libraries("${TEST_TARGET_NAME}" PRIVATE GTest::Main GTest::GTest ${ARGN})
-    add_test(NAME "${REL_BINARY_DIR}/${test_name}" COMMAND "${TEST_TARGET_NAME}")
+    add_test(NAME "${REL_BINARY_DIR}/${test_name}"
+        COMMAND "${TEST_TARGET_NAME}"
+        WORKING_DIRECTORY "$<TARGET_FILE_DIR:ginkgo>")
```
tcojean, line 19: Why set the working directory here, that would make it be `build/core` if I understand this directive?

third_party/gflags/CMakeLists.txt
```diff
@@ -1,36 +1,18 @@
...
-        "-DGFLAGS_BUILD_PACKAGING=OFF" "-DCMAKE_CXX_FLAGS_DEBUG=${CMAKE_CXX_FLAGS_DEBUG}"
-        "-DCMAKE_CXX_FLAGS_RELEASE=${CMAKE_CXX_FLAGS_RELEASE}" "-DCMAKE_C_FLAGS_DEBUG=${CMAKE_C_FLAGS_DEBUG}"
-        "-DCMAKE_C_FLAGS_RELEASE=${CMAKE_C_FLAGS_RELEASE}")
-else()
-    # There is a weird issue with Intel 19 and c++17 causing a linking error.
-    # Use c++14 instead.
-    set(INTEL19_STD_FIX "")
-    if (CMAKE_CXX_COMPILER_ID STREQUAL "Intel" AND CMAKE_CXX_COMPILER_VERSION MATCHES "19")
-        set(INTEL19_STD_FIX "-DCMAKE_CXX_FLAGS=-std=c++14")
+if(NOT gflags_FOUND)
```
tcojean, line 1: So we reverse the current behavior in that case, which we had previously defined to be use ours first except if explicitly requested. I think that eases package management, but might surprise some users. You should document this in the relevant `.md` file(s).

Maybe to make this clearer, we could put all these `if(NOT xxx_FOUND)` in `third_party/CMakeLists.txt` as well?

third_party/git-cmake-format/CMakeLists.txt
```diff
@@ -1,7 +1,12 @@
...
-    ${CMAKE_CURRENT_BINARY_DIR}/build EXCLUDE_FROM_ALL)
+include(FetchContent)
+FetchContent_Declare(
+    git_cmake_format
+    GIT_REPOSITORY https://github.com/ginkgo-project/git-cmake-format.git
+    GIT_TAG        be9554a5d71030dcc23f1eb355c41ab9d876e776
+)
+FetchContent_GetProperties(git_cmake_format)
+if(NOT git_cmake_format_POPULATED)
+    message(STATUS "Fetching git-cmake-format")
```
tcojean, line 9: Shouldn't the message be at the top level? e.g., before the `FetchContent_Declare`. Also to uniformize with the other packages.

third_party/gtest/CMakeLists.txt
```diff
@@ -1,21 +1,27 @@
...
+    # need to set the variables in CACHE due to CMP0077
+    set(gtest_disable_pthreads ON CACHE INTERNAL "")
+    if(MSVC)
+        set(gtest_force_shared_crt ON CACHE INTERNAL "")
+    endif()
+    FetchContent_GetProperties(googletest)
+    if(NOT googletest_POPULATED)
+        FetchContent_Populate(googletest)
+    endif()
+    add_subdirectory(${googletest_SOURCE_DIR} ${googletest_BINARY_DIR} EXCLUDE_FROM_ALL)
```
tcojean, line 18: Shouldn't the `add_subdirectory` be done only if it's not populated already? Otherwise that would be available already no? See example here:
https://cmake.org/cmake/help/latest/module/FetchContent.html#overview

third_party/gtest/CMakeLists.txt
```diff
@@ -1,21 +1,27 @@
...
+    # make sure the tests DLLs are placed in the working path for CTest
+    set_target_properties(gtest gtest_main PROPERTIES
+        RUNTIME_OUTPUT_DIRECTORY "${GINKGO_LIBRARY_PATH}"
+        ARCHIVE_OUTPUT_DIRECTORY "${GINKGO_LIBRARY_PATH}"
+        LIBRARY_OUTPUT_DIRECTORY "${GINKGO_LIBRARY_PATH}")
```
tcojean, line 23: Is this for Windows specifically then? Or is there some RPATH issue on linux as well?

tcojean (approved):
LGTM in general.

Nice catch that we can now use `FetchContent` due to all the CMake version bumps, that is much less awkward to use than `ExternalProject`.

I have a few comments and questions.

CMakeLists.txt
```diff
@@ -216,22 +220,17 @@ endif()
 
 # Try to find the third party packages before using our subdirectories
 include(cmake/package_helpers.cmake)
-ginkgo_find_package(GTest "GTest::GTest;GTest::Main" FALSE 1.8.1)
-ginkgo_find_package(gflags gflags FALSE 2.2.2)
-ginkgo_find_package(RapidJSON rapidjson TRUE 1.1.0)
+find_package(GTest 1.10.0)
+find_package(gflags 2.2.2 QUIET)
+find_package(RapidJSON 1.1.0 QUIET)
```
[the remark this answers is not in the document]
upsj: GTest has a FindGTest module in CMake, so it will work in any situation and only output "GTest not found" by default. gflags and RapidJSON don't, so if not found, they will output a long error message. We could of course decide to make all of them quiet.

core/test/accessor/CMakeLists.txt
upsj: There is a longer story behind this which I should mention: Contrary to the CMake default, GTest builds static libraries by default, if BUILD_SHARED_LIBS is unspecified. That means that previously, all tests were linked statically in all settings and we didn't need to worry about the location. With ExternalProject, we didn't forward our BUILD_SHARED_LIBS settings, but with add_subdirectory we do, which means we are now linking dynamically to GTest.
To make tests work on Windows without having to copy around DLLs all the time, we placed all libraries and DLLs into ´lib/`, which is also where `libginkgo.dll` and more importantly `libgtest.dll` will be stored. So this change is necessary for all tests to find `libgtest.dll`

third_party/gtest/CMakeLists.txt
upsj: see my explanation for test WORKING_DIRECTORY above

third_party/gtest/CMakeLists.txt
upsj: This is just me misunderstanding the purpose of *_POPULATED
I thought it meant "has this code already been downloaded in a previous CMake run" while it means "has this code already been downloaded previously in the same CMake run", i.e. if multiple CMake files load the same package. I will move to the documented version.

third_party/gflags/CMakeLists.txt
upsj: good idea. In which cases did we actually use the existing libraries then?

third_party/git-cmake-format/CMakeLists.txt
upsj: Good point, that was an oversight

third_party/rapidjson/CMakeLists.txt
```diff
@@ -1,10 +1,14 @@
...
+    rapidjson
+    GIT_REPOSITORY https://github.com/Tencent/rapidjson.git
+    GIT_TAG        88bd956d66d348f478bceebfdadb8e26c6844695
+)
+FetchContent_GetProperties(rapidjson)
+if(NOT rapidjson_POPULATED)
+    FetchContent_Populate(rapidjson)
+endif()
+set(RapidJSON_INCLUDE_DIR "${rapidjson_SOURCE_DIR}/include")
+add_library(rapidjson INTERFACE)
```
yhmtsai, line 13: doesn't rapidjson provide their target?

yhmtsai (approved):
LGTM. 
the gflags/gtest are in `build/_dep/<>-src` and `build/_dep/<>-build` not the directory inside the third_party.
I think it is okay?

yhmtsai:
could FetchContent handle non-cmake configure project?
change the `CONFIGURE_COMMAND` to `"${TPL_HWLOC_PATH}/src/configure --disable-nvml --disable-cuda --disable-rsmi"`

third_party/rapidjson/CMakeLists.txt
upsj: Unfortunately no, they only provide a target in their RapidJSONConfig.cmake, which is only available in installed RapidJSON, but not if we include it in our build.

third_party/gflags/CMakeLists.txt
tcojean: Most of the time that's already in `third_party/CMakeLists.txt`. Adding `AND NOT xxx_FOUND` to the existing `if` conditions should just work.

CMakeLists.txt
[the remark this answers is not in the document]
tcojean: Thanks for the explanation. Maybe write a comment about that in that case so that someone doesn't get surprised (or maybe if `FindGTest` somehow gets removed).

core/test/accessor/CMakeLists.txt
tcojean: Thanks for the detailed explanation. If that's the case, maybe it makes sense to explicitly add an `if (WIN32)` to highlight that fact, and keep the defaults otherwise?

CMakeLists.txt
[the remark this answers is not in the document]
upsj: I added some comments :)

third_party/gflags/CMakeLists.txt
upsj: I think `ginkgo_add_subdirectory` did the same thing (only include if `NOT xxx_FOUND`), so this should not change the behavior.

core/test/accessor/CMakeLists.txt
upsj: The working directory does not matter except for dynamic linking, so I would prefer to keep this, as the explicit conditional would make the code more complex.

upsj:
So I think my changes here are now complete, I simplified the HWLOC integration by
* Removing all MSVC-specific code, since we disable it for Windows/Mac anyways.
* Merging the GINKGO_USE_EXTERNAL_HWLOC variable with `HWLOC_FOUND`, because they will both always have the same value
* Removing GINKGO_USE_EXTERNAL_* and TPL_* variables from the installed Ginkgo config, since the test and benchmark environment is not installed, and users should now care whether the hwloc we are linking to internally was existing or built by Ginkgo.
* Removing unused CMake code
* Checking for packages with find_package only if they are needed
@tcojean or @yhmtsai could you give the latest version another read?

CMakeLists.txt
```diff
@@ -182,11 +182,8 @@ if(GINKGO_BUILD_HWLOC AND (MSVC OR WIN32 OR CYGWIN OR APPLE))
     message(WARNING "Ginkgo does not support HWLOC on Windows/MacOS, switch GINKGO_BUILD_HWLOC to OFF")
 endif()
 if(GINKGO_BUILD_HWLOC)
-    # By default always use external HWLOC
-    set(GINKGO_USE_EXTERNAL_HWLOC 1)
     set(GINKGO_HAVE_HWLOC 1)
```
yhmtsai, line 185: can this variable `GINKGO_HAVE_HWLOC` also be deleted? `GINKGO_HAVE_HWLOC` seems to be equal to `GINKGO_BUILD_HWLOC`

tcojean: We use this to have a `define` which is set to `0` or `1`. If we use the `BUILD` instead we also need to change all macros to use `#ifdef GKO_HAVE_HWLOC` rather than `#if GKO_HAVE_HWLOC`.

tcojean (approved):
LGTM I think. I will need to check what spack thinks of these changes.

Issue #483, opened 2020-03-20
Unifying kernel tests


upsj:
At the moment, all of our non-reference kernel tests are basically the same (except for the templated tests, but I will get to that): We generate some random matrices or load one of the ani matrices, execute each of our kernels, maybe in multiple configurations on reference and the executor we are testing and compare the result for equality or near-equality.

This approach means a lot of code duplication, a lot of potential for inconsistencies and a lot of work in case we need to change an internal kernel interface.

With a very hacky and inaccurate bash one-liner, I looked at the amount of different lines between CUDA and HIP as well as OMP and CUDA/HIP for each test:
<details>

```
> for f in cuda/test/**/*.c*; do echo $f; ff=${f##cuda/}; fff=${ff%%.c*}; c1=$(diff -y --suppress-common-lines $f hip/${fff}.* | grep "^" | wc -l);c2=$(cat $f | wc -l); echo "$c1 / $c2" | bc -l; done
cuda/test/base/exception_helpers.cu
.18072289156626506024
cuda/test/base/math.cu
.07228915662650602409
cuda/test/components/cooperative_groups.cu
.33716475095785440613
cuda/test/components/merging.cu
.13103448275862068965
cuda/test/components/prefix_sum.cu
.03191489361702127659
cuda/test/components/searching.cu
.06530612244897959183
cuda/test/components/sorting.cu
.12056737588652482269
cuda/test/factorization/par_ilu_kernels.cpp
.26011560693641618497
cuda/test/matrix/coo_kernels.cpp
.04580152671755725190
cuda/test/matrix/csr_kernels.cpp
.11893583724569640062
cuda/test/matrix/dense_kernels.cpp
.12231759656652360515
cuda/test/matrix/ell_kernels.cpp
.04624277456647398843
cuda/test/matrix/hybrid_kernels.cpp
.07272727272727272727
cuda/test/matrix/sellp_kernels.cpp
.06060606060606060606
cuda/test/preconditioner/jacobi_kernels.cpp
.13129973474801061007
cuda/test/solver/bicg_kernels.cpp
.12885154061624649859
cuda/test/solver/bicgstab_kernels.cpp
.16713091922005571030
cuda/test/solver/cg_kernels.cpp
.15073529411764705882
cuda/test/solver/cgs_kernels.cpp
.17191977077363896848
cuda/test/solver/fcg_kernels.cpp
.14736842105263157894
cuda/test/solver/gmres_kernels.cpp
.15107913669064748201
[...]

PR #787, opened 2021-06-06
Create example for unified test execution

upsj:
As a first step towards addressing #483, this PR adds an example for how unified kernel execution could look.

Upsides of this approach:

1. Drastically reduced code duplication
2. Faster execution time (Reference is only executed once)

Downsides:

1. Test code is slightly more complex
2. Test messages are slightly worse
3. "catastrophic failure" (Segfault etc.) in one executor prevents other tests from running


tcojean:
As a small comment, I don't think I like this. Test code being duplicated is a non issue for me because it's only test code, all it should be is easily understandable and accessible. In addition, I believe we will want to have different features/tests for different hardware: OpenMP and DPC++ are slower than the other executors. Also, I think will want more feature into our tests, like testing with real matrices or varying sizes at some point, which will very greatly vary depending on the executor due to the time constraints, among other things. Finally, I'm in general not too happy about moving most of Ginkgo into `common/xxx` in terms of library design.

upsj:
Let me start off with all the things I agree with you on:

* `common/test`: Yes, that is a bad name, maybe something like `ginkgo/test` or `ginkgo/unittest` might be a more suitable location.
* We definitely need a way to test certain inputs only on some executors out of runtime considerations
* The test cases should be easy to understand

I think the only thing I really disagree on is your assessment that code duplication in tests in Ginkgo specifically is not an issue. Most test files in omp, hip and cuda are virtually identical, except maybe for small differences in input sizes, seeds or structure of the test fixture. Adding a new feature or a new test for an existing feature means I have to make the same modifications to 3 (soon 4) files, often involving both test fixture and test cases themselves. Even more so if we decided to make more fundamental changes to the test framework (templated tests, reordering/restructuring test cases, more inputs, ...). Individual test cases may be close to write-only code, but our tests as a whole still need to be maintained like regular code.

Especially now that we are close to adding a significant part of DPC++ support, I wanted to jump ahead trying to add individual test files for this executor as well, and instead merge the majority of test files into a single framework.

There are several possible approaches how this could be implemented, from the top of my head:

1. similar to the simple kernel interface, compile the same source file with different `#define`s to provide a uniform interface
1. loop over executors in some way: either explicitly, by templating the tests on the executor type or by using GoogleTest Parameterized tests
1. (what I started in this PR) implement *IsEquivalentToRef tests by providing a AAA-like framework based on per-executor test fixture state and Lambdas that are executed for Acting and Asserting in each executor separately.

I went for the last solution mainly because it avoids multiple executions of the random matrix generation and Reference code and simplifies the Fixture setup for these types of `IsEquivalentToRef` operations, with the potential to copy back the data to the host and perform explicit data assertion checks also for device executors (this would allow us to run reference tests on all devices to test small-scale behavior on top of comparisons against Reference). I am not too attached to this approach though, since the others I mentioned also seem useful, and keep the code closer to the status quo.

Going forward, you raise good points about future extensibility and handling of complexity in light of a larger number of inputs, which I want to address in the following:

If I take a 10k ~~feet~~ meter view of Ginkgo tests, I see three dimensions: Executors, Operations and Inputs (four if you count value types). We only have a handful of executors and operations, so what you describe means drastically extending the input dimension and maybe adding a few limitations on input size out of runtime considerations. I don't think that this disqualifies a unification approach though, since we can still find ways to list inputs or input sizes that should be skipped for certain executors.

Honestly though, I believe if we want to move to more complex combinations of inputs and executors in integration-like tests, GoogleTest may not provide the generality we need. If I look at what Catch2 provides with its [Generators and Sections](https://github.com/catchorg/Catch2/blob/devel/docs/generators.md), I believe this might be more suitable and expressive both in terms of code and test output for settings where we want to test the same operation on a larger number of inputs both synthetic and application-provided.

common/test/matrix/coo_kernels.cpp
```diff
@@ -0,0 +1,475 @@
...
+            dev.copy_from(ref_data);
+        }
+        auto ref_result = run(ref_data);
+        std::vector<decltype(ref_result)> dev_results;
+        for (auto &dev : device_data) {
+            SCOPED_TRACE(dev.get_name());
+            dev_results.push_back(run(dev));
+        }
+        int i = 0;
+        for (auto &dev : device_data) {
```
Slaedr, line 192: Sometimes, old is gold.
```suggestion
        for (size_t i = 0; i < device_data.size(); i++) {
            auto &dev = device_data[i];
```

common/test/matrix/coo_kernels.cpp
```diff
@@ -0,0 +1,475 @@
...
+            SCOPED_TRACE(dev.get_name());
+            test(dev_results[i], ref_result);
+            i++;
```
Slaedr, line 195:
```suggestion
```

Slaedr (commented):
I like this! As long as we don't insist on using this for every single kind of test, I think it's good.

Some considerations:
- You could try to do this for a couple of solver tests to see if any changes are needed to the overall framework.
- Can some of this be moved to a header file (like `common/test/matrix/setup.hpp`), that can be included from, say, `cuda/test/matrix/coo_kernels.cpp`? If there are some tests that are specific to cuda, it would be nice to use some of the functions here like `set_up_apply_data` by just including a header.
- I am actually okay with putting this in `common`. That directory is indeed slowly becoming more and more common, and it does seem like a natural place to put this.

common/test/matrix/coo_kernels.cpp
Slaedr, line 195: Sometimes, old is gold.
```suggestion
        for (size_t i = 0; i < device_data.size(); i++) {
            SCOPED_TRACE(device_data[i].get_name());
            test(dev_results[i], ref_result);
```

upsj:
With the changes in #733, this PR is mostly moot.

PR #788, opened 2021-06-09
Parallel reductions for batched operations

Slaedr:
The following changes are introduced:
- Parallel 2-norms and dot products replace the earlier sequential placeholders.
- Minor improvement to row reduction in batch CSR SpMV
- Fix for Clang-Cuda build issue in batch IDR


Slaedr:
format!

omp/preconditioner/batch_jacobi_kernels.cpp
```diff
@@ -37,7 +37,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
 #include "core/matrix/batch_struct.hpp"
-#include "omp/matrix/batch_struct.hpp"
+#include "reference/matrix/batch_struct.hpp"
```
pratikvn, line 40: Maybe if this is common in reference and omp, we can move it to `common/` instead ?

reference/matrix/batch_struct.hpp
```diff
@@ -44,7 +44,11 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 namespace gko {
 namespace kernels {
-namespace reference {
+/**
+ * @brief A namespace for shared functionality between omp and reference
+ *  executors.
+ */
+namespace host {
```
pratikvn, line 51: I think it makes sense to move this to a common folder.

pratikvn (approved):
LGTM! Only a minor thing. We might need to benchmark this for multiple rhs at a later point to be sure that this does not degrade the performance too much.

omp/preconditioner/batch_jacobi_kernels.cpp
Slaedr: But `common` is currently used for cuda and hip. This cannot be used for those, but only on the "host". Maybe we create a new folder called `common_host` or something?

pratikvn: I mean you can put it in common and include it only in reference and omp. They dont have to be with the `.inc` extension either. Just a normal `.hpp` should be good.

Slaedr: I still think it would be a bit confusing that something in `common` does not work on cuda and hip. Would you be okay with a `common_host` or `common_cpu` folder for this?

PR #790, opened 2021-06-11
Add some micro-benchmarking support to Ginkgo

pratikvn:
This PR adds some micro-benchmarking support to benchmark all the small functionalities of Ginkgo. For example, you can get information like the following:

1. Executor creation and run timings and overheads.
mbench

2. Array creation, copy, move timings and differences.
array-mbench

3. More specific backend-specific timings as well are possible, example for arrays with CUDA:
cuda-array-mbench

Of course there are a few things that we need to discuss:

1. Timings for GPUs is from the host side, but that may be okay.
2. The repetitions for the function calls is not something that can be clearly controlled through google benchmark, which as discussed before is a downside of this.

I think this type of micro-benchmarking information can be really useful in detecting performance degradations and can atleast give us a hint if some commit/change regressed the performance.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,6 +59,9 @@ option(GINKGO_CONFIG_LOG_DETAILED
 option(GINKGO_BENCHMARK_ENABLE_TUNING
     "Enable tuning variables in the benchmarks. For specific use cases, manual code changes could be required."
     OFF)
+option(GINKGO_BUILD_MICRO_BENCHMARKS
+    "Build Ginkgo's micro-benchmarks. This benchmarks the smaller functionalities to ensure their performance."
+    OFF)
 set(GINKGO_VERBOSE_LEVEL "1" CACHE STRING
     "Verbosity level. Put 0 to turn off. 1 activates a few important messages.")
 if(MSVC)
@@ -224,6 +227,9 @@ endif()
 if(GINKGO_BUILD_BENCHMARKS)
     find_package(gflags 2.2.2 QUIET)
     find_package(RapidJSON 1.1.0 QUIET)
+endif()
+if(GINKGO_BUILD_MICRO_BENCHMARKS)
+    find_package(gbench QUIET)
 endif()
 if(GINKGO_BUILD_HWLOC)
     find_package(HWLOC 2.1) # No need for QUIET as we ship FindHWLOC
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -119,9 +119,9 @@ add_custom_command(
             ${CMAKE_CURRENT_SOURCE_DIR}/run_all_benchmarks.sh
             ${CMAKE_CURRENT_BINARY_DIR}/run_all_benchmarks.sh)
 
-add_custom_target(benchmark)
+add_custom_target(gko-benchmark)
 add_custom_command(
-    TARGET benchmark POST_BUILD
+    TARGET gko-benchmark POST_BUILD
     COMMAND bash run_all_benchmarks.sh >/dev/null
     DEPENDS make_run_all_benchmarks
     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
--- a/cmake/create_mbench.cmake
+++ b/cmake/create_mbench.cmake
@@ -1,0 +1,21 @@
+function(ginkgo_create_mbench mbench_name)
+    set(THREADS_PREFER_PTHREAD_FLAG ON)
+    find_package(Threads REQUIRED)
+    file(RELATIVE_PATH REL_BINARY_DIR
+        ${PROJECT_BINARY_DIR} ${CMAKE_CURRENT_BINARY_DIR})
+    string(REPLACE "/" "_" MBENCH_TARGET_NAME "${REL_BINARY_DIR}/${mbench_name}")
+    add_executable(${MBENCH_TARGET_NAME} ${mbench_name}.cpp)
+    target_compile_features("${MBENCH_TARGET_NAME}" PUBLIC cxx_std_14)
+    target_include_directories("${MBENCH_TARGET_NAME}"
+        PRIVATE
+        "$<BUILD_INTERFACE:${Ginkgo_BINARY_DIR}>"
+        )
+    set_target_properties(${MBENCH_TARGET_NAME} PROPERTIES
+        OUTPUT_NAME ${mbench_name})
+    if (GINKGO_CHECK_CIRCULAR_DEPS)
+        target_link_libraries(${MBENCH_TARGET_NAME} PRIVATE "${GINKGO_CIRCULAR_DEPS_FLAGS}")
+    endif()
+    target_link_libraries(${MBENCH_TARGET_NAME} PRIVATE ginkgo GBench::GBench GBench::Main
+        Threads::Threads ${ARGN})
+    add_test(NAME ${REL_BINARY_DIR}/${mbench_name} COMMAND ${MBENCH_TARGET_NAME})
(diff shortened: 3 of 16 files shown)


cmake/create_mbench.cmake
```diff
@@ -0,0 +1,44 @@
...
+    set_target_properties(${MBENCH_TARGET_NAME} PROPERTIES
+        OUTPUT_NAME ${mbench_name})
+    if (GINKGO_CHECK_CIRCULAR_DEPS)
+        target_link_libraries(${MBENCH_TARGET_NAME} PRIVATE "${GINKGO_CIRCULAR_DEPS_FLAGS}")
+    endif()
+    target_link_libraries(${MBENCH_TARGET_NAME} PRIVATE ginkgo GBench::GBench GBench::Main
+        Threads::Threads ${ARGN})
+endfunction(ginkgo_create_mbench)
+
+function(ginkgo_create_cuda_mbench mbench_name)
```
upsj, line 22: Do we need cuda microbenchmarks specifically, as opposed to benchmarking C++ code calling ginkgo_cuda functionality?

CMakeLists.txt
```diff
@@ -224,6 +227,9 @@ if(GINKGO_BUILD_BENCHMARKS)
     find_package(gflags 2.2.2 QUIET)
     find_package(RapidJSON 1.1.0 QUIET)
 endif()
+if(GINKGO_BUILD_MICRO_BENCHMARKS)
+    find_package(gbench QUIET)
```
upsj, line 231: nit: maybe add a specific version matching the commit hash you provided?

benchmark/CMakeLists.txt
```diff
@@ -117,9 +117,9 @@ add_custom_command(
             ${CMAKE_CURRENT_SOURCE_DIR}/run_all_benchmarks.sh
             ${CMAKE_CURRENT_BINARY_DIR}/run_all_benchmarks.sh)
 
-add_custom_target(benchmark)
+add_custom_target(gko-benchmark)
 add_custom_command(
-    TARGET benchmark POST_BUILD
+    TARGET gko-benchmark POST_BUILD
```
upsj, line 122: why rename these?

core/CMakeLists.txt
```diff
@@ -89,3 +89,7 @@ endif()
 if(GINKGO_BUILD_TESTS)
     add_subdirectory(test)
 endif()
+
+if(GINKGO_BUILD_MICRO_BENCHMARKS)
+    add_subdirectory(micro_bench)
```
upsj, line 94: nit: I would prefer `micro_benchmarks`

core/micro_bench/base/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+ginkgo_create_mbench(array)
```
upsj, line 1: same here: maybe use `microbenchmark`?

core/micro_bench/base/array.cpp
```diff
@@ -0,0 +1,252 @@
...
+}
+
+BENCHMARK_REGISTER_F(Array, Clear)->RangeMultiplier(8)->Range(64, 64 << 12);
+
+
+BENCHMARK_DEFINE_F(Array, ResizeAndReset)(benchmark::State &state)
+{
+    auto x = gko::Array<int>{this->exec, state.range(0)};
+    for (auto _ : state) {
+        x.resize_and_reset(state.range(0) + this->dis(this->gen));
```
upsj, line 246: Can we make this benchmark deterministic instead? Or are you benchmarking `malloc` here?

core/micro_bench/base/array.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+using exec_ptr = std::shared_ptr<gko::Executor>;
+
+
+class Array : public benchmark::Fixture {
+public:
+    void SetUp(const ::benchmark::State &state)
+    {
+        gen();
+        exec = gko::ReferenceExecutor::create();
```
upsj, line 55: Is it possible to parametrize the benchmark on different executors? Then we could get allocation times for CUDA, ROCm, DPC++ and CPU code.

third_party/gbench/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+    add_subdirectory(${googlebenchmark_SOURCE_DIR} ${googlebenchmark_BINARY_DIR} EXCLUDE_FROM_ALL)
+endif()
+# make sure the tests DLLs are placed in the working path for CTest
+set_target_properties(benchmark benchmark_main PROPERTIES
+    RUNTIME_OUTPUT_DIRECTORY "${GINKGO_LIBRARY_PATH}"
+    ARCHIVE_OUTPUT_DIRECTORY "${GINKGO_LIBRARY_PATH}"
+    LIBRARY_OUTPUT_DIRECTORY "${GINKGO_LIBRARY_PATH}")
+# by default, the outdated targets are not being exported
+add_library(GBench::Main ALIAS benchmark_main)
+add_library(GBench::GBench ALIAS benchmark)
```
upsj, line 35: Can we use the original names instead? Otherwise this becomes confusing. CMake already deprecated the GTest::Main target, so we have to use it for backwards-compatibility only, but we are starting from scratch here, so we are free to choose. Or are those the targets that are also exported by installed GBench?

third_party/gbench/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+#         "-DCMAKE_CXX_FLAGS_DEBUG=${CMAKE_CXX_FLAGS_DEBUG}"
+#         "-DCMAKE_CXX_FLAGS_RELEASE=${CMAKE_CXX_FLAGS_RELEASE}"
+#         "-DCMAKE_C_FLAGS_DEBUG=${CMAKE_C_FLAGS_DEBUG}"
+#         "-DCMAKE_C_FLAGS_RELEASE=${CMAKE_C_FLAGS_RELEASE}")
+
+# ginkgo_add_external_target(GBench::GBench benchmark src/include build/src
+#    STATIC "d" gbench_external FALSE)
+# ginkgo_add_external_target(GBench::Main benchmark_main src/include build/src
+#     STATIC "d" gbench_external FALSE)
+
```
upsj, line 15: Do we need this still?

upsj (commented):
LGTM! Some ideas for benchmarks we could add:
* 0x0 SpMVs for all matrix types
* 0x0 Solver runs with different numbers of iterations
* make_temporary_clone and make_temporary_conversion runtimes for tiny matrices
* All of the above with and without attached loggers (probably only ones with little state and no output make sense)

pratikvn:
@upsj, Thank you for the review and the comments. I will try to update it soon.

tcojean:
Thanks for this, I think using google benchmark for these parts can make sense, since that's pretty much what it was designed with. I have a few questions, although I didn't look into the code in depth:
+ I like the name `micro_benchmarks` for the directory as that explains what should go in there.
+ I added some overhead benchmarks to our base benchmark line, maybe that can go in there instead (or in addition?).
+ We maybe should use the full name `googlebenchmark` for the targets and some other aspects, at first it wasn't clear to me what `GBench` and `mbench` were. I would stick to using `micro_benchmarks` always for the Ginkgo side, and the full `googlebenchmark` or whichever is the original name for the third party.
+ A big question is CUDA: by default the timer is CPU only and that's terrible for doing GPU micro-benchmarking. I know it's possible to add custom timers but I'm not sure how extensible they are, back when I had tried it (many years ago) that aspect wasn't satisfactory, maybe it's better now?

pratikvn:
@tcojean, thanks. I will look into GPU timers and see if it extensible from within googlebenchmark. I agree with your other suggestions.

PR #794, opened 2021-06-17
Changes in batch GMRES and parallel reductions for batch dense

aggarwal2000:
This PR:

A. Merges generate and apply plane rotation kernels in batch GMRES

B. Offers a slight improvement in batch dense compute norm and dot kernels
   The difference between the current and the new approach for reduction could be seen in the attached plots. Approach of using one warp per right hand side is faster, though the overall difference in timings for the batch solvers this would lead to is only a very few milliseconds. 
[batched_reduction_comparison.pdf](https://github.com/ginkgo-project/ginkgo/files/6668029/batched_reduction_comparison.pdf)


common/matrix/batch_dense_kernels.hpp.inc
```diff
@@ -147,11 +122,38 @@ __device__ __forceinline__ void compute_norm2(
...
-        one_norm2(x, col, result.values);
+    using real_type = typename gko::remove_complex<ValueType>;
+    constexpr auto tile_size = config::warp_size;
+
+    auto thread_block = group::this_thread_block();
+
+    auto subwarp_grp = group::tiled_partition<tile_size>(thread_block);
+
+    const auto subwarp_grp_id = static_cast<int>(threadIdx.x / tile_size);
+
```
Slaedr, line 133: Nit: you don't need to leave that many empty lines.
```suggestion
    auto subwarp_grp = group::tiled_partition<tile_size>(thread_block);
    const auto subwarp_grp_id = static_cast<int>(threadIdx.x / tile_size);
```

common/matrix/batch_dense_kernels.hpp.inc
```diff
@@ -166,15 +168,44 @@ __device__ __forceinline__ void compute_norm2(
...
+        }
+
+        // subwarp_grp level reduction
+        for (int j = subwarp_grp.size() / 2; j > 0; j /= 2) {
+            val += subwarp_grp.shfl_down(val, j);
+        }
+
+        if (subwarp_grp.thread_rank() == 0) {
+            result.values[rhs_index] = sqrt(val);
+        }
```
Slaedr, line 204: Please put this in a separate `__device__ __forceinline__` function. That was the purpose of the `one_norm2` function earlier.

common/matrix/batch_dense_kernels.hpp.inc
```diff
@@ -264,11 +270,39 @@ __device__ __forceinline__ void compute_dot_product(
...
+
+        // subwarp_grp level reduction
+        for (int j = subwarp_grp.size() / 2; j > 0; j /= 2) {
+            val += subwarp_grp.shfl_down(val, j);
+        }
+
+        if (subwarp_grp.thread_rank() == 0) {
+            result.values[rhs_index] = val;
+        }
```
Slaedr, line 301: Again, this should be its own inline function and reused below.

Slaedr (changes requested):
Excellent work! The only issue I have is with the code duplication. If you could move the common part of the norm functions (and it's really the common part that's the most important part) to a common function, that would be great. Similar for the dot functions.

Slaedr:
It would also be good if you could take a look at the Sonarcloud issues and see if any of them can be fixed without too much effort.

common/matrix/batch_dense_kernels.hpp.inc
aggarwal2000: Removed all unnecessary lines.

common/matrix/batch_dense_kernels.hpp.inc
aggarwal2000: Done!

common/matrix/batch_dense_kernels.hpp.inc
aggarwal2000: Done!

Slaedr (approved):
LGTM!

pratikvn (approved):
LGTM! But maybe we should one cuda job to run the cuda tests as well.

PR #793, opened 2021-06-16
Fix some sanitizer failures

upsj:
This PR fixes a (probably harmless) UBSAN/ASAN overflow/out-of-bounds memory access in ISAI and an uninitialized memory read in IDR. Additionally, it warns when using a build type that is not supported by the compiler (an issue found by @fritzgoebel), and runs the `cuda-memcheck` tasks on amdci.

PR #795, opened 2021-06-17
Run cuda-memcheck with CUDA 10.1 on amdci

upsj:
HIP doesn't support CUDA 11.*, so the fix in #793 was incomplete

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -976,7 +976,10 @@ cudamemcheck:
   stage: QoS_tools
   extends:
     - .deploy_condition
-    - .use_gko-cuda110-gnu9-llvm9-intel2020
+  image: ginkgohub/cuda:101-gnu8-llvm10-intel2019
+  tags:
+    - private_ci
+    - nvidia-gpu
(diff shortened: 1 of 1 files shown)


Slaedr (approved):
LGTM.

.gitlab-ci.yml
```diff
@@ -976,7 +976,10 @@ cudamemcheck:
   stage: QoS_tools
   extends:
     - .deploy_condition
-    - .use_gko-cuda110-gnu9-llvm9-intel2020
+  image: ginkgohub/cuda:101-gnu8-llvm10-intel2019
+  tags:
+    - private_ci
+    - nvidia-gpu
```
yhmtsai, line 982: Could you move it to image.yml?

upsj: No, because I don't want to move the other jobs of that image to amdci. This is only temporary until we have more GPU-capable CI resources.

yhmtsai: if we are only using it for this, I am fine with it

yhmtsai (approved):

PR #797, opened 2021-06-17
Core solver tests for all backends are moved to a common function

Slaedr:
Re-formulated the test to check whether calling the batch solvers through their public interface works. The reformulation is in terms of checking whether the logger contains reasonable values. The loose reasoning is as follows:
img
It's not as rigorous as Latex makes it look, but it should do for now.

All this is put into a common location to remove duplication.


Slaedr:
format!

PR #799, opened 2021-06-18
DPCPP SpGEMM, SpGEAM, Transpose, Sort

upsj:
This PR adds SpGEMM, SpGEAM, transpose, sort and is_sorted kernels to DPC++. They don't give great performance, but they work.
* SpGEAM uses a simple two-way merge algorithm
* SpGEMM uses a binary heap-based multiway merge algorithm like in OpenMP
* Sort uses Max-Heapsort, which is the simplest asymptotically optimal in-place sorting algorithm
* Transpose uses atomics to count the number of non-zeros per columns and assign unique indices in a second pass, followed by sorting

### DIFF
--- a/dpcpp/components/atomic.dp.hpp
+++ b/dpcpp/components/atomic.dp.hpp
@@ -1,0 +1,214 @@
+/*******************************<GINKGO LICENSE>******************************
+Copyright (c) 2017-2021, the Ginkgo authors
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+
+1. Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+
+2. Redistributions in binary form must reproduce the above copyright
+notice, this list of conditions and the following disclaimer in the
+documentation and/or other materials provided with the distribution.
+
+3. Neither the name of the copyright holder nor the names of its
+contributors may be used to endorse or promote products derived from
+this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS
+IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED
+TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A
+PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+#ifndef GKO_DPCPP_COMPONENTS_ATOMIC_DP_HPP_
+#define GKO_DPCPP_COMPONENTS_ATOMIC_DP_HPP_
+
+
+#include <type_traits>
+
+
+#include <CL/sycl.hpp>
+
+
+#include "dpcpp/base/dpct.hpp"
+
+// #include "dpct/atomic.hpp"
+
+namespace gko {
+namespace kernels {
+namespace dpcpp {
+namespace atomic {
+constexpr cl::sycl::access::address_space local_space =
+    cl::sycl::access::address_space::local_space;
+constexpr cl::sycl::access::address_space global_space =
+    cl::sycl::access::address_space::global_space;
+}  // namespace atomic
+
+namespace {
+
+
+template <cl::sycl::access::address_space addressSpace = atomic::global_space,
+          typename T>
+T atomic_compare_exchange_strong(
+    cl::sycl::multi_ptr<T, addressSpace> addr, T expected, T desired,
+    cl::sycl::memory_order success = cl::sycl::memory_order::relaxed,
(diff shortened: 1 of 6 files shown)


dpcpp/base/executor.dp.cpp
```diff
@@ -102,6 +102,7 @@ void DpcppExecutor::populate_exec_info(const MachineTopology *mach_topo)
 
 void DpcppExecutor::raw_free(void *ptr) const noexcept
 {
+    queue_->wait_and_throw();
```
pratikvn, line 105: Just curious here. Isn't this supposed to be `noexcept` ?

upsj: yeah, I would kind of like to blame SYCL/the DPC++ runtime here, since it frees data before the previous kernel finished running.

upsj: I guess this is a temporary workaround we shouldn't merge.

dpcpp/components/atomic.dp.hpp
```diff
@@ -0,0 +1,214 @@
...
+
+
+#define GKO_BIND_ATOMIC_HELPER_STRUCTURE(CONVERTER_TYPE)                   \
+    template <cl::sycl::access::address_space addressSpace,                \
+              typename ValueType>                                          \
+    struct atomic_helper<                                                  \
+        addressSpace, ValueType,                                           \
+        std::enable_if_t<(sizeof(ValueType) == sizeof(CONVERTER_TYPE))>> { \
+        __dpct_inline__ static ValueType atomic_add(                       \
+            ValueType *__restrict__ addr, ValueType val)                   \
```
tcojean, line 138: Do we need this? Since they have fetch_type and their stuff is templated everywhere, can't we use the base type directly?
https://intel.github.io/llvm-docs/doxygen/classcl_1_1sycl_1_1atomic.html

tcojean (approved):
LGTM. A few comments or questions.

dpcpp/components/atomic.dp.hpp
yhmtsai: IsValidAtomicType does not support complex and some of them do not support float.( __SYCL_STATIC_ASSERT_NOT_FLOAT)
I do not do it pretty well by template. the complex<float> is done by 8 byte type but complex<double> is two 8 byte impl.

dpcpp/matrix/csr_kernels.dp.cpp
```diff
@@ -88,37 +91,263 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
...
+                           const typename HeapElement::value_type *b_vals,
+                           HeapElement *heap, InitCallback init_cb,
+                           StepCallback step_cb, ColCallback col_cb)
+    -> decltype(init_cb(0))
+{
+    auto a_begin = a_row_ptrs[row];
+    auto a_end = a_row_ptrs[row + 1];
+
+    using index_type = typename HeapElement::index_type;
+    constexpr auto sentinel = std::numeric_limits<index_type>::max();
```
thoasm, line 234: nit: needs include of `<limits>`

dpcpp/matrix/csr_kernels.dp.cpp
```diff
@@ -207,10 +631,64 @@ void transpose_and_transform(std::shared_ptr<const DpcppExecutor> exec,
...
+    const auto row_ptrs = orig->get_const_row_ptrs();
+    const auto cols = orig->get_const_col_idxs();
+    const auto vals = orig->get_const_values();
+
+    Array<IndexType> counts{exec, num_cols + 1};
+    auto tmp_counts = counts.get_data();
+    auto out_row_ptrs = trans->get_row_ptrs();
+    auto out_cols = trans->get_col_idxs();
+    auto out_vals = trans->get_values();
+    components::fill_array(exec, tmp_counts, num_cols, IndexType{});
```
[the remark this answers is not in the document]
thoasm: I think it would be better if we would use `queue->memset()` inside `components::fill_array` for DPC++ and keep the call here (IMO, the current code is more descriptive and I doubt we would get performance benefits from `memset()`).

dpcpp/matrix/csr_kernels.dp.cpp
```diff
@@ -323,8 +852,30 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
...
+                for (auto i = begin; i < end - 1; i++) {
+                    if (cols[i] > cols[i + 1]) {
+                        *is_sorted_device = false;
+                        break;
+                    }
+                }
+            }
+        });
+    });
+    exec->get_master()->copy_from(exec.get(), 1, is_sorted_device, is_sorted);
```
thoasm, line 877: nit: maybe easier to read and understans:
```suggestion
    *is_sorted = exec->copy_val_to_host(is_sorted_device);
```

thoasm (approved):
LGTM!
I have some nits, but nothing important.

dpcpp/components/atomic.dp.hpp
```diff
@@ -0,0 +1,214 @@
...
+        fake_complex<real_type> *fake_addr =
+            reinterpret_cast<fake_complex<real_type> *>(addr);
+        // Separate to real part and imag part
+        auto real = atomic_helper<addressSpace, real_type>::atomic_add(
+            &(fake_addr->x), val.real());
+        auto imag = atomic_helper<addressSpace, real_type>::atomic_add(
+            &(fake_addr->y), val.imag());
+        return {real, imag};
+    }
+};
```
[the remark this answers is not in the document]
yhmtsai: which part? the fake_addr or the whole thing.
If it is for fake_addr, we need it because I can not get the reference from std::complex

dpcpp/components/atomic.dp.hpp
```diff
@@ -0,0 +1,214 @@
...
+constexpr cl::sycl::access::address_space local_space =
+    cl::sycl::access::address_space::local_space;
+constexpr cl::sycl::access::address_space global_space =
+    cl::sycl::access::address_space::global_space;
+}  // namespace atomic
+
+namespace {
+
+
+template <cl::sycl::access::address_space addressSpace = atomic::global_space,
```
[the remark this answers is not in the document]
yhmtsai: when we use the function from kernel, the pointer is plain. we use addressSpace to specify the storage location.
I do not investigate whether we can use multi_ptr in all kernel to reduce the template specification.

dpcpp/matrix/csr_kernels.dp.cpp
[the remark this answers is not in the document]
upsj: no, SYCL has a typed memset template

dpcpp/components/atomic.dp.hpp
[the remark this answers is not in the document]
upsj: You don't need fake_complex, you can reinterpret_cast into ´real_type*` directly, see OpenMP atomic_add.

dpcpp/base/executor.dp.cpp
upsj: To be fixed in #832

PR #801, opened 2021-06-22
Improve benchmark output

upsj:
This PR adds two matrix dimension output for `matrix_statistics`


MarcelKoch:
I think the first part of this PR is rather uncontroversial, but for the second part I'm not so sure. Would it make sense to split this PR, as the two parts already have few things in common?

upsj:
@MarcelKoch you're of course totally correct, I was just a bit lazy :)

MarcelKoch (approved):
LGTM, seems like a useful addition

yhmtsai (approved):
LGTM

PR #802, opened 2021-06-22
Benchmark work estimate

upsj:
This PR adds work estimates to the executor Operations, implements them for a few Dense kernels and outputs them in benchmark loggers.

### DIFF
--- a/benchmark/utils/loggers.hpp
+++ b/benchmark/utils/loggers.hpp
@@ -120,17 +120,17 @@ struct OperationLogger : gko::log::Logger {
         for (const auto &entry : work) {
             add_or_set_member(work_object, entry.first.c_str(),
                               rapidjson::Value(rapidjson::kObjectType), alloc);
-            work_object[entry.first.c_str()].AddMember(
-                "flops", entry.second.flops / repetitions, alloc);
-            work_object[entry.first.c_str()].AddMember(
-                "memory", entry.second.memory_volume / repetitions, alloc);
+            add_or_set_member(work_object[entry.first.c_str()], "flops",
+                              entry.second.flops / repetitions, alloc);
+            add_or_set_member(work_object[entry.first.c_str()], "memory",
+                              entry.second.memory_volume / repetitions, alloc);
         }
         add_or_set_member(work_object, "total",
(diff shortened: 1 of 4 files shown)


tcojean:
I think one limit with this technic if  I understand properly is that it requires two consistencies:
1. All executors will have a similar implementation so that they have the same work estimates
2. Every operation is tied to only one kernel.

I'm not sure about 1. for now, I think most of our algorithms have roughly the same order of magnitude of work between executors, but that means a design like the CSR SpMV  (classical, imbalance, ...) with strategies would be a no go, and instead we would need an operation for each strategy and switch strategy at the core/algorithm level (which is maybe the best thing to do anyway).

I don't think it's a downside, I thought I should just mention it.

pratikvn:
I think instead of artificially classifying operations into likely compute bound and likely memory bound (which is very much dependent on hardware), IMO a better approach would be to just calculate the work and memory complexities of the operations and register them for each operation. We can then have a roofline estimator (which could take in the hardware properties), to estimate whether the operation is memory-bound or compute bound.

upsj:
I don't think the classification is particularly artificial, let me formulate it as
* compute-bound means the compute complexity grows asymptotically faster than the memory complexity
* memory-bound means the memory complexity grows asymptotically at least as fast as the compute complexity

There are many kernels where it doesn't make sense to talk of FLOPS, or that don't allow for a nice closed-form expression of their memory footprint/compute complexity, which is why I want to leave the option open to either not annotate kernels at all, or to annotate them with custom metrics.

upsj:
Yes, there are more precise models or exact performance counters. This is aiming mostly to provide a rootline-like approximation to the performance, to quickly highlight kernels that are significantly below expected performance, enabling users to highlight possible optimization opportunities. We can use this framework to capture such information on an application level without the need for executing with a profiler, which requires additional tooling for analysis.

The BLAS 1/2 and solver kernels should be pretty accurate, only the SpMVs undercount accesses to the input vector, which should ideally be served from cache though. IIRC, the footprints are equivalent to what we used in our ACM TOMS paper to report achieved bandwidths for different SpMVs and solvers.

PR #803, opened 2021-06-23
Add tests for bit-mask dependent batch-dense kernels

aggarwal2000:
This PR adds Reference, OpenMp and Cuda tests for all the batch dense functions which accept an additional convergence bit-set parameter.

### DIFF
--- a/cuda/test/matrix/batch_dense_kernels.cpp
+++ b/cuda/test/matrix/batch_dense_kernels.cpp
@@ -348,6 +348,11 @@ TEST_F(BatchDense, CudaConvergenceComputeNorm2IsEquivalentToRef)
 
     const gko::uint32 converged = 0xbfa00f0c | (0 - (1 << num_rhs));
 
+    for (int ibatch = 0; ibatch < batch_size; ibatch++) {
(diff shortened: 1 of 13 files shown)


cuda/test/matrix/batch_dense_kernels.cpp
```diff
@@ -168,6 +168,22 @@ TEST_F(BatchDense, SingleVectorCudaAddScaledIsEquivalentToRef)
 }
 
 
+TEST_F(BatchDense, SingleVectorCudaConvergenceAddScaledIsEquivalentToRef)
+{
+    const int num_rhs = 1;
+    set_up_vector_data(num_rhs);
+
```
Slaedr, line 175:
```suggestion
```
Normally, Ginkgo tests follow
```
<setup>
<one blank line>
<run test and, if necessary, reference code>
<one blank line>
<test result>
```
with on other extra blank lines.

Slaedr (approved):
Looks good, thanks for putting in all the tests. But now that we are planning to remove multiple right hand sides, some of this might soon be removed 😅 I think it makes sense to merge this for now, and we then decide what to keep and what to remove.

I have one issue below with the one of the reference tests, and some other minor comments.

PR #807, opened 2021-06-30
Fix Coo out of bounds access for strided vectors

upsj:
This PR prevents the Coo CPU kernels from overwriting data past the end of the vector, since they took the amount of storage, not the actual number of elements into account.
The commit message is actually slightly inaccurate, since the linearized index skips padding, but may write past the end of the vector.

### DIFF
--- a/cuda/test/matrix/coo_kernels.cpp
+++ b/cuda/test/matrix/coo_kernels.cpp
@@ -141,6 +141,24 @@ TEST_F(Coo, SimpleApplyIsEquivalentToRef)
 }
 
 
+TEST_F(Coo, SimpleApplyDoesntOverwritePadding)
+{
+    set_up_apply_data();
+    auto dresult_padded =
+        Vec::create(cuda, dresult->get_size(), dresult->get_stride() + 1);
+    dresult_padded->copy_from(dresult.get());
+    double padding_val{1234.0};
+    cuda->copy_from(cuda->get_master().get(), 1, &padding_val,
+                    dresult_padded->get_values() + 1);
+
+    mtx->apply(y.get(), expected.get());
+    dmtx->apply(dy.get(), dresult.get());
+
+    GKO_ASSERT_MTX_NEAR(dresult, expected, 1e-14);
+    ASSERT_EQ(cuda->copy_val_to_host(dresult_padded->get_values() + 1), 1234.0);
+}
+
+
 TEST_F(Coo, SimpleApplyIsEquivalentToRefUnsorted)
 {
     set_up_apply_data();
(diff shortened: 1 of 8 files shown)


MarcelKoch:
This change seems necessary, but just out of curiosity, have you looked at the performance impact of this? My guess is that it would be a bit slower for single right-hand sides.

upsj:
Yes, all of these kernels would profit heavily of specialization for single RHS, but that is the topic for another PR. The performance would probably make too much of a difference, since the previous version involves integer divisions, which the compilers seemed to have issues with before.

omp/matrix/coo_kernels.cpp
```diff
@@ -67,11 +67,12 @@ void spmv(std::shared_ptr<const OmpExecutor> exec,
...
 {
-#pragma omp parallel for
-    for (size_type i = 0; i < c->get_num_stored_elements(); i++) {
-        c->at(i) = zero<ValueType>();
+#pragma omp parallel for collapse(2)
+    for (size_type i = 0; i < c->get_size()[0]; i++) {
+        for (size_type j = 0; j < c->get_size()[1]; j++) {
+            c->at(i, j) = zero<ValueType>();
+        }
     }
```
yhmtsai, line 75: you can also use the dense::fill for it

omp/matrix/coo_kernels.cpp
```diff
@@ -87,9 +88,11 @@ void advanced_spmv(std::shared_ptr<const OmpExecutor> exec,
...
     auto beta_val = beta->at(0, 0);
-#pragma omp parallel for
-    for (size_type i = 0; i < c->get_num_stored_elements(); i++) {
-        c->at(i) *= beta_val;
+#pragma omp parallel for collapse(2)
+    for (size_type i = 0; i < c->get_size()[0]; i++) {
+        for (size_type j = 0; j < c->get_size()[1]; j++) {
+            c->at(i, j) *= beta_val;
+        }
     }
```
yhmtsai, line 96: dense::scale

yhmtsai (changes requested):
omp initial/scale can be replaced by dense::fill or dense::scale.
Could you also add the corresponding test to cuda/hip?

cuda/test/matrix/coo_kernels.cpp
```diff
@@ -141,6 +141,24 @@ TEST_F(Coo, SimpleApplyIsEquivalentToRef)
...
+    set_up_apply_data();
+    auto dresult_padded =
+        Vec::create(cuda, dresult->get_size(), dresult->get_stride() + 1);
+    dresult_padded->copy_from(dresult.get());
+    double padding_val{1234.0};
+    cuda->copy_from(cuda->get_master().get(), 1, &padding_val,
+                    dresult_padded->get_values() + 1);
+
+    mtx->apply(y.get(), expected.get());
+    dmtx->apply(dy.get(), dresult.get());
```
yhmtsai, line 155: should it be dresult_padded? also for others?

yhmtsai (approved):
LGTM. one question about the apply output object

pratikvn (approved):
LGTM!

PR #808, opened 2021-06-30
Scalar Jacobi specialization and kernels

pratikvn:
This PR adds special kernels for the scalar Jacobi preconditioner. It extracts the diagonal from the system matrix and applies the inverse of the diagonal to the rhs. The generate and apply overhead compared to the usual block-Jacobi should be lesser due to fewer storage requirements and memory movement in general.

This should in combination with the `Ir` solver, provide for some cheap and effective relaxation methods, useful in multi-grid and coarse grid solvers.


omp/preconditioner/jacobi_kernels.cpp
```diff
@@ -696,6 +696,45 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
...
+template <typename ValueType>
+void scalar_apply(std::shared_ptr<const DefaultExecutor> exec,
+                  const matrix::Diagonal<ValueType> *diag,
+                  const matrix::Dense<ValueType> *alpha,
+                  const matrix::Dense<ValueType> *b,
+                  const matrix::Dense<ValueType> *beta,
+                  matrix::Dense<ValueType> *x)
+{
+#pragma omp parallel for
+    for (size_type i = 0; i < x->get_size()[0]; ++i) {
```
upsj, line 708: This would be a prime candidate for #733 :)

include/ginkgo/core/preconditioner/jacobi.hpp
```diff
@@ -604,6 +608,7 @@ class Jacobi : public EnableLinOp<Jacobi<ValueType, IndexType>>,
     size_type num_blocks_;
     Array<value_type> blocks_;
     Array<remove_complex<value_type>> conditioning_;
+    std::shared_ptr<matrix::Diagonal<value_type>> diag_;
```
upsj, line 611: How about reusing the `blocks` array instead? This avoids a few issues related to copy/move_to

hip/test/preconditioner/jacobi_kernels.cpp
```diff
@@ -461,6 +492,44 @@ TEST_F(Jacobi, HipLinearCombinationApplyEquivalentToRef)
 }
 
 
+TEST_F(Jacobi, HipScalarLinearCombinationApplyEquivalentToRef)
```
upsj, line 495: Can you try using different non-default strides for the two vectors?

upsj (changes requested):
LGTM in general, this makes for a really neat interface :)
Only two specific things come to my mind:
1. do we need to do something special for 0 diagonal entries? Maybe set them to 1 (implicitly)? Maybe we should consider the same for singular blocks (setting them to the identity matrix)?
2. I would prefer if we could reuse the existing data structures to avoid complicating the value semantics of the objects, see #753. Basically ´Array` does the right thing, while `std::shared_ptr` doesn't respect executors, which can cause issues when copying around `Jacobi` objects.

upsj (approved):

hartwiganzt (approved):
LGTM.
There is no test whether this kernel is invoked when the BS is set to 1, or did I miss that?

pratikvn:
@hartwiganzt, I tested that manually by printing from within the newly added kernels.

pratikvn:
rebase!

Issue #320, opened 2019-06-27
Improving Csr::strategy_type


thoasm:
Currently, all strategies for CSR are tailored for CUDA, which is fine for the most part.
However, also the names are CUDA specific (see `Csr::cusparse`), which should be changed in my opinion since we want to support multiple platforms. I would prefer to go with a more neutral name like `Csr::sparse_library` in  that case (maybe there is even an OpenMP sparse library we could use for that).
Also, `automatical` only works on CUDA devices and even requires a `CudaExecutor` to work. I would prefer a solution where it is possible to also adapt to certain OpenMP properties (which we should have as soon as we have a more sophisticated SpMV there).

Additionally, I am not sure why we use an `std::shared_ptr` for these strategies. Currently, we always have to call `std::make_shared` to generate a strategy, which is both not intuitive and not necessary since there is not much stored inside a strategy object (at most an `std::string` and an `int64_t`). I think copying the object would be faster than allocating memory on the heap, although it should not really matter much (the more important part for me is the intuitiveness).
We could also encapsulate the strategies in a class named `strategy`, so it is clear that `Csr::spmv_strategy::automatical` is an SpMV strategy.

In summary, I think the following changes should be introduced:
- Change the names of the strategies to more neutral ones, e.g. `cusparse` -> `sparse_library`
- Make `automatical` to actually be automatical and dependent on the executor (CUDA vs. OpenMP vs. Reference) without requiring a CudaExecutor
- Change the type of the strategy from `std::shared_ptr` to just a plain object since the most one of these objects contain is an `int64_t` and an `std::string`
- put all strategy classes in a separate class `spmv_strategy` (or similar), so you call it with `Csr::spmv_strategy::automatical`, which is more descriptive

yhmtsai:
The strategy `automatical` or  `load_balance` compute the `srow` for the GPU kernel but the matrix is maybe on the host memory.

strategy creates a CUDA handle to take the parameter of GPU [code](https://github.com/ginkgo-project/ginkgo/blob/cd7465db0c6e51eb92b154b5bc4146fd3aee08e4/include/ginkgo/core/matrix/csr.hpp#L163)
bechmark/spmv does not pass the executor to strategy when reading the matrix data. [code](https://github.com/ginkgo-project/ginkgo/blob/cd7465db0c6e51eb92b154b5bc4146fd3aee08e4/benchmark/spmv/spmv.cpp#L101)

[...]

Issue #171, opened 2018-12-03
Keep benchmarking errors


gflegar:
Currently, we're dropping all the errors that happen in the benchmarks.

The `run_all_benchmarks.sh` script redirects all the `stderr` output from benchmark runners to `stdout`, and the `stdout` of `run_all_benchmarks.sh` is redirected to `/dev/null` in the CMake target for running the benchmarks. Instead, we should create a temporary file and redirect the output there.

PR #812, opened 2021-07-02
Various benchmark improvements

upsj:
## Avoid benchmarking ELL format with heavy imbalance
ELL is the only format that has a really significant worst-case runtime behavior, so this PR avoids benchmarking ELL on heavily imbalanced matrices, measured by the quotient _max nnz / mean nnz_ over all rows.

## Fix Hybrid/Coo -> Csr conversions
The OpenMP kernel gets stuck with empty Coo matrices due to `num_nonzeros - 1` overflowing or produces incomplete data if the last row is not present in `row_idxs`.

## Fix IterationControl
conversion benchmarks used the timer counts instead of the IterationControl counts, which leads to incorrect repetition counts in the JSON output and probably incorrect averages.

## Fix accidental CudaExecutor generation
The Csr strategies create a Cuda executor as a default value, which breaks things in benchmarking environment with BUILD_CUDA enabled, but without CUDA support. I added a function that takes care of selecting the correct strategy depending on the executor type. Related to #320 

## Write exception messages into JSON output
Partially addressing #171, store the benchmarking exception messages next to `completed: false`

### DIFF
--- a/benchmark/conversions/conversions.cpp
+++ b/benchmark/conversions/conversions.cpp
@@ -88,9 +88,9 @@ void convert_matrix(const gko::LinOp *matrix_from, const char *format_to,
             matrix_to->copy_from(matrix_from);
         }
         add_or_set_member(conversion_case[conversion_name], "time",
-                          timer->compute_average_time(), allocator);
+                          ic.compute_average_time(), allocator);
         add_or_set_member(conversion_case[conversion_name], "repetitions",
-                          timer->get_num_repetitions(), allocator);
+                          ic.get_num_repetitions(), allocator);
 
         // compute and write benchmark data
         add_or_set_member(conversion_case[conversion_name], "completed", true,
--- a/benchmark/conversions/conversions.cpp
+++ b/benchmark/conversions/conversions.cpp
@@ -98,6 +98,12 @@ void convert_matrix(const gko::LinOp *matrix_from, const char *format_to,
     } catch (const std::exception &e) {
         add_or_set_member(test_case["conversions"][conversion_name],
                           "completed", false, allocator);
+        if (FLAGS_keep_errors) {
+            rapidjson::Value msg_value;
+            msg_value.SetString(e.what(), allocator);
+            add_or_set_member(test_case["conversions"][conversion_name],
+                              "error", msg_value, allocator);
+        }
         std::cerr << "Error when processing test case " << test_case << "\n"
                   << "what(): " << e.what() << std::endl;
     }
--- a/benchmark/conversions/conversions.cpp
+++ b/benchmark/conversions/conversions.cpp
@@ -156,8 +156,7 @@ int main(int argc, char *argv[])
             try {
                 auto matrix_from =
                     share(formats::matrix_factory.at(format_from)(exec, data));
-                for (const auto &format : formats::matrix_factory) {
-                    const auto format_to = std::get<0>(format);
+                for (const auto &format_to : formats) {
                     if (format_from == format_to) {
                         continue;
                     }
--- a/benchmark/run_all_benchmarks.sh
+++ b/benchmark/run_all_benchmarks.sh
@@ -21,6 +21,11 @@ if [ ! "${REPETITIONS}" ]; then
     echo "REPETITIONS    environment variable not set - assuming ${REPETITIONS}" 1>&2
 fi
 
+if [ ! "${SOLVER_REPETITIONS}" ]; then
+    SOLVER_REPETITIONS=1
+    echo "SOLVER_REPETITIONS    environment variable not set - assuming ${SOLVER_REPETITIONS}" 1>&2
+fi
+
 if [ ! "${SEGMENTS}" ]; then
     echo "SEGMENTS    environment variable not set - running entire suite" 1>&2
     SEGMENTS=1
@@ -253,7 +258,7 @@ run_solver_benchmarks() {
                     --gpu_timer=${GPU_TIMER} \
                     --jacobi_max_block_size=${SOLVERS_JACOBI_MAX_BS} --device_id="${DEVICE_ID}" \
                     --gmres_restart="${SOLVERS_GMRES_RESTART}" \
-                    --repetitions="${REPETITIONS}" \
+                    --repetitions="${SOLVER_REPETITIONS}" \
                     <"$1.imd" 2>&1 >"$1"
     keep_latest "$1" "$1.bkp" "$1.bkp2" "$1.imd"
 }
--- a/benchmark/run_all_benchmarks.sh
+++ b/benchmark/run_all_benchmarks.sh
@@ -35,6 +35,11 @@ if [ ! "${FORMATS}" ]; then
     FORMATS="csr,coo,ell,hybrid,sellp"
 fi
 
+if [ ! "${ELL_IMBALANCE_LIMIT}" ]; then
+    echo "ELL_IMBALANCE_LIMIT    environment variable not set - assuming 100" 1>&2
+    ELL_IMBALANCE_LIMIT=100
+fi
+
 if [ ! "${SOLVERS}" ]; then
     SOLVERS="bicgstab,cg,cgs,fcg,gmres,cb_gmres_reduce1,idr"
     echo "SOLVERS    environment variable not set - assuming \"${SOLVERS}\"" 1>&2
@@ -67,7 +72,7 @@ fi
 
 if [ ! "${SOLVERS_JACOBI_MAX_BS}" ]; then
     SOLVERS_JACOBI_MAX_BS="32"
-    "SOLVERS_JACOBI_MAX_BS environment variable not set - assuming \"${SOLVERS_JACOBI_MAX_BS}\"" 1>&2
+    echo "SOLVERS_JACOBI_MAX_BS environment variable not set - assuming \"${SOLVERS_JACOBI_MAX_BS}\"" 1>&2
 fi
 
 if [ ! "${BENCHMARK_PRECISION}" ]; then
@@ -191,6 +196,17 @@ compute_matrix_statistics() {
 }
 
 
+remove_ell_worstcase() {
+    local IMBALANCE=$(jq '.[0].problem.row_distribution | .max / (.median + 1) | floor' $1)
+    # if the imbalance is too large, remove ELL formats from the list.
+    if [[ "${IMBALANCE}" -gt "${ELL_IMBALANCE_LIMIT}" ]]; then
+        echo -n $FORMATS | tr ',' '\n' | grep -vE '^ell' | tr '\n' ',' | sed 's/,$//'
+    else
+        echo -n $FORMATS
+    fi
+}
+
+
 # Runs the conversion benchmarks for all matrix formats by using file $1 as the
 # input, and updating it with the results. Backups are created after each
 # benchmark run, to prevent data loss in case of a crash. Once the benchmarking
@@ -199,8 +215,9 @@ compute_matrix_statistics() {
 run_conversion_benchmarks() {
     [ "${DRY_RUN}" == "true" ] && return
     cp "$1" "$1.imd" # make sure we're not loosing the original input
+    local LOCAL_FORMATS=$(remove_ell_worstcase "$1")
     ./conversions/conversions${BENCH_SUFFIX} --backup="$1.bkp" --double_buffer="$1.bkp2" \
-                --executor="${EXECUTOR}" --formats="${FORMATS}" \
+                --executor="${EXECUTOR}" --formats="${LOCAL_FORMATS}" \
                 --device_id="${DEVICE_ID}" --gpu_timer=${GPU_TIMER} \
                 <"$1.imd" 2>&1 >"$1"
     keep_latest "$1" "$1.bkp" "$1.bkp2" "$1.imd"
@@ -214,9 +231,10 @@ run_conversion_benchmarks() {
 # taken as the final result.
 run_spmv_benchmarks() {
     [ "${DRY_RUN}" == "true" ] && return
+    local LOCAL_FORMATS=$(remove_ell_worstcase "$1")
     cp "$1" "$1.imd" # make sure we're not loosing the original input
     ./spmv/spmv${BENCH_SUFFIX} --backup="$1.bkp" --double_buffer="$1.bkp2" \
-                --executor="${EXECUTOR}" --formats="${FORMATS}" \
+                --executor="${EXECUTOR}" --formats="${LOCAL_FORMATS}" \
                 --device_id="${DEVICE_ID}" --gpu_timer=${GPU_TIMER} \
                 <"$1.imd" 2>&1 >"$1"
     keep_latest "$1" "$1.bkp" "$1.bkp2" "$1.imd"
--- a/benchmark/run_all_benchmarks.sh
+++ b/benchmark/run_all_benchmarks.sh
@@ -16,6 +16,11 @@ if [ ! "${EXECUTOR}" ]; then
     echo "EXECUTOR    environment variable not set - assuming \"${EXECUTOR}\"" 1>&2
 fi
 
+if [ ! "${REPETITIONS}" ]; then
+    REPETITIONS=10
+    echo "REPETITIONS    environment variable not set - assuming ${REPETITIONS}" 1>&2
+fi
+
 if [ ! "${SEGMENTS}" ]; then
     echo "SEGMENTS    environment variable not set - running entire suite" 1>&2
     SEGMENTS=1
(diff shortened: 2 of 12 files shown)


benchmark/run_all_benchmarks.sh
```diff
@@ -35,6 +35,11 @@ if [ ! "${FORMATS}" ]; then
     FORMATS="csr,coo,ell,hybrid,sellp"
 fi
 
+if [ ! "${ELL_IMBALANCE_LIMIT}" ]; then
+    echo "ELL_IMBALANCE_LIMIT    environment variable not set - assuming 100" 1>&2
+    ELL_IMBALANCE_LIMIT=100
```
yhmtsai, line 40: I would prefer give `-1` to run all matrices for ell
Also, we have some stuff to give better performance in these cases than the vendors'

upsj: If you want to run all matrices, you can use a number like `1000000000` which we will definitely not exceed. IMO it does not make any sense to run ELL benchmarks for such imbalanced matrices. ELL can be hundreds to thousands of times slower than any other matrix format, since almost all entries are padding zeroes then.

yhmtsai: It's slow but we still need the result if we want a complete benchmark. 
`1000000000` should be enough but we do not really know it's enough or not.

upsj: The max value I found in ginkgo-data/V100 is 100M (note that this is bounded by the number of columns in the matrix), so this is a safe margin, and setting it to 100 would only exclude 450 matrices. Imagine a matrix with a single dense row. Storing it in ELL would mean storing the matrix as Dense. Attempting to build such a matrix would throw `bad_alloc` in the best case and cause the OOM killer to terminate the benchmark in the worst case, leading to incomplete data. Waiting an additional 2 hours for the benchmark to finish for cases that fit into memory just to have complete numbers we already know to be very slow doesn't seem worth it to me.

yhmtsai: bad_alloc or OOM killer depends on the server setting, but it requires the admin permission.
If those 450 matrices are all OOM case, it should be fine. It's not worth but it needed if we would like to claim we run all real matrices. we should not kick out the matrices when the format in device is available to handle it.
ELL can also handle some cases well with the max/medium > 100, like half of them are 1, some of is 200.
To be clear, it is useful if we just want a quick and close to full benchmark.

upsj: To give some background: This was motivated by my recent OpenMP benchmarks. I saw runtimes of 400s for matrix conversion, SpMVs failing due to huge CUDA allocations as well as the OOM killer terminating benchmark processes. All of those should be avoided in real benchmarks. max / median being large means that the matrix consists of at least 50% zeros. For 380 of these matrices, the matrix is even at least 75% empty (max / q3), and that is a very low estimate, since many of those have only a single dense row or power-law distributed row sizes. I wouldn't consider ELL a relevant format for these matrices.

upsj: Now that I think of it, I guess `.max / .mean` is probably a more suitable value, since it actually measures the ratio non-zero vs. empty storage. 100x storage overhead should be a good exclusion criterion, right?

yhmtsai: In general, yes. I do not judge the functionality. mainly for what should be the default behavior.
For me, the benchmark only needs to run through all inputs and faces error according to the system.
Using `-1` does not need to depend on the matrix_analysis to skip the selection. (need to manually disable the matrix_analysis though)
For CPU allocation, adjusting `/proc/sys/vm/overcommit_memory` might change the behvaior.
the default behavior, malloc does not guarantee the memory is enough. It starts fills the memory when touch it, so it will be killed by OOM. If all are in CUDA, the benchmark can catch the exception and does not kill the program.

upsj: I am aware of the default allocation behavior on Linux. The thing is, we have two issues here:
1. OOM killer leads to all matrix formats after `ell` not being benchmarked. Modifying system flags is not an issue for us. Might also be that SLURM gets involved, but I haven't checked that in detail.
2. Jobs that don't hit the OOM limit, but allocate huge amounts of memory and take forever (again, I had runtimes of 400 s for a single repetition of a single conversion, i.e. 1h for a _single_ format conversion operation on a _single_ matrix), and we should not run them by default.

pratikvn: Currently, you are skipping the matrix if `.max/.median` is above 100 and hence you don't write anything in the `ell { }` for the matrix, right ? Maybe it makes sense to have another parameter `suitability` and set that to `true` or `false` to record these types of situations ?

upsj: Yes, more precisely `.max / .mean`, since this measures storage overhead. The difference happens only on the parameter level, we do not modify the JSON input, since it only describes format-agnostic information. The different approaches I considered for this are:
1. have an ELL size limit inside the `spmv` benchmark executable itself. That would require us to determine the row size distribution every time, though.
2. do this check only where we have the necessary data available, namely inside `run_all_benchmark.sh` after we computed the `matrix_statistics`.

Where would you put this `suitability` parameter?

yhmtsai: for the .max/.mean, if the matrices contains dense row contiguously and less(or zero) contiguously, the ell pack should give the similar performance as classical csr.

upsj: That is only partially true, since the classical CSR kernel only needs to read the non-zeros that are actually there, while the ELL kernel explicitly reads all the zero values and column indices. So the thread-to-row mapping is similar, but the memory access differs a lot.

yhmtsai: I am also thinking about what is the difference between limiting ell in these cases and limiting cg in symmetric.
ell is designed as matrix format but good for balanced matrix originally.
cg is designed as krylov solver for symmetric matrices.

yhmtsai: no, our ell has early stopping, so it only touches the first one zero.

pratikvn: I think one of the concerns that Mike has, is the uniformity of a benchmark, in the sense that there should always be a output available for all the different parameters (ex. csr, ell, coo etc.) of an operation (spmv) being benchmarked. If you have a category called `suitability` in the parameter, it would make it easier to just say and record that the matrix format was not suitable for that particular operation. So, it would be inside the matrix format itself. So `ell { "suitability": , }`

hartwiganzt: I feel it makes sense to skip these matrices in the benchmarks. It is true that we would outperform the competitors, but this is really not the target problem where we are interested in outperforming competitors. So I am ok with skipping those automatically (even though this breaks - as Mike points out) the assumption that all benchmarks are processed.

upsj: I see, sorry about the confusion, the check is rather well-hidden and assumes sorted column indices, it seems? We should note that somewhere. But I still feel like we are getting away from the point here. Classical CSR would also be rather unsuitable for this problem, but it does not have any storage overhead compared to other formats, so there is no potential for crashing due to memory limitations.

yhmtsai: Yes, if it needs the early stopping.
The performance and suitability needs some knowledge about the benchmark set, the format and maybe implementation. the option is advanced option for me. 
@pratikvn comment is good. some output in the json gives the user what happened in the benchmark. also need to add "completed": false. 
for OOM issue, I prefer using some system call to get free space and ell storage, require 2x ell space in conversion and 1x ell + something space in the spmv.

upsj: I'll be honest: While I understand the intention behind the `skipped` flag, I am not sure it justifies the effort of either
* adding a new flag `skipped_formats` to the benchmarks together with the associated logic or
* modifying JSON data inside the `run_all_benchmarks.sh` script.
I feel like the absense of the `ell` entry speaks for itself. Of course, we should note this in the BENCHMARKING.md.

And as a general comment, we all have limited time to work on things, and this discussion already took us more time (and probably headspace) than I spent creating the entire PR.

For a bit of history, I had similar issues with the SpGEMM benchmarks, where some matrices took huge amounts of time (literally hours) or crashed after a while, causing subsequent benchmarks to fail, and I had to manually exclude them, but here where we have a nice heuristic that can easily identify unsuitable matrices, excluding which avoids any potential for crashes and huge runtimes, so I would heavily prefer doing that, also since we are very much targeting automation with this script.

tcojean: I'm fine with this behavior but agree with Mike that more is needed to make the information useful while plotting and analyzing a complete benchmark run for a paper. Normally, we use `completed` to check which matrix format went through. `completed: false` means that there was an exception, in practice that is almost always an OOM condition. In the best case, we would give a third value to `completed` to something like `skipped` with your condition. Of course, we cannot because `completed` is a boolean.

AFAIR, missing entries altogether in the JSON file make the GPE and other plotter quite annoying to manage for some type of script.

benchmark/utils/formats.hpp
```diff
@@ -181,6 +181,27 @@ std::unique_ptr<MatrixType> read_matrix_from_data(
...
+ * falls back to csr::classical for executors without support for this strategy.
+ *
+ * @tparam Strategy  one of csr::automatical or csr::load_balance
+ */
+template <typename Strategy>
+std::shared_ptr<csr::strategy_type> create_gpu_strategy(
+    std::shared_ptr<const gko::Executor> exec)
+{
+    if (auto cuda = dynamic_cast<const gko::CudaExecutor *>(exec.get())) {
+        return std::make_shared<Strategy>(cuda->shared_from_this());
```
pratikvn, line 196: This is a good find! We should definitely document it in `csr.hpp` so that we remember to fix this when we fix #320

benchmark/conversions/conversions.cpp
```diff
@@ -156,8 +156,7 @@ int main(int argc, char *argv[])
             try {
                 auto matrix_from =
                     share(formats::matrix_factory.at(format_from)(exec, data));
-                for (const auto &format : formats::matrix_factory) {
-                    const auto format_to = std::get<0>(format);
+                for (const auto &format_to : formats) {
```
pratikvn, line 159: Does it automatically use `<0>` th element in the map ?

pratikvn (approved):
LGTM!

benchmark/conversions/conversions.cpp
upsj: No, std::get<0> gets the key out of the (key, value) pair from the map

benchmark/run_all_benchmarks.sh
upsj: What do you think of the latest solution? Moving the ELL imbalance check into formats.hpp plus storing the exception message for `completed: false`? @yhmtsai @pratikvn @tcojean

yhmtsai: I think it is good. storing exception will be considered in #171. for now it can just throw the exception and mark `completed: false`. For `skipped_formats`, I prefer `-1` to represent non-skipped to reduce the dependency on matrix_statistics

yhmtsai: ah, I do not check the latest changes. You seems to store the error already?

omp/test/matrix/hybrid_kernels.cpp
```diff
@@ -230,6 +238,49 @@ TEST_F(Hybrid, CountNonzerosIsEquivalentToRef)
...
+    auto dcsr_mtx = gko::matrix::Csr<>::create(omp);
+
+    balanced_mtx->convert_to(csr_mtx.get());
+    dbalanced_mtx->convert_to(dcsr_mtx.get());
+
+    GKO_ASSERT_MTX_NEAR(csr_mtx.get(), dcsr_mtx.get(), 1e-14);
+}
+
+
+TEST_F(Hybrid, ConvertWithEmptyFirstAndLastRowToCsrIsEquivalentToRef)
```
yhmtsai, line 259: do you have the test failed result?

benchmark/conversions/conversions.cpp
```diff
@@ -156,8 +162,7 @@ int main(int argc, char *argv[])
...
```
yhmtsai, line 165: does it get the pair (key, value)? what is the string of format_from and format_to

yhmtsai (changes requested):
LGTM, but I have some questions in the comments

benchmark/run_all_benchmarks.sh
```diff
@@ -239,6 +253,7 @@ run_solver_benchmarks() {
                     --gpu_timer=${GPU_TIMER} \
                     --jacobi_max_block_size=${SOLVERS_JACOBI_MAX_BS} --device_id="${DEVICE_ID}" \
                     --gmres_restart="${SOLVERS_GMRES_RESTART}" \
+                    --repetitions="${REPETITIONS}" \
```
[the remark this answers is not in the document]
upsj: yes, that is the default everywhere, set by `general.hpp`

benchmark/utils/formats.hpp
```diff
@@ -181,6 +186,51 @@ std::unique_ptr<MatrixType> read_matrix_from_data(
...
+        return;
+    }
+    std::vector<gko::size_type> row_lengths(data.size[0]);
+    for (auto nz : data.nonzeros) {
+        row_lengths[nz.row]++;
+    }
+    auto max_len = *std::max_element(row_lengths.begin(), row_lengths.end());
+    auto avg_len =
+        std::max<gko::size_type>(data.nonzeros.size() / data.size[0], 1);
+    if (max_len / avg_len > FLAGS_ell_imbalance_limit) {
```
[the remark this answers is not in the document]
upsj: They don't really need to be, since this is only a heuristic.

benchmark/conversions/conversions.cpp
upsj: before, we tested the conversions from the input formats to all available formats, which is pretty much overkill since most of them are duplicate, and only differ in their strategy settings. `formats` is the list of formats from `FLAGS_formats`, so instead of doing FLAGS_formats x all_formats, we now just do FLAGS_formats x FLAGS_formats conversions.

benchmark/run_all_benchmarks.sh
upsj: Yes, I had the same idea :) Just not certain whether `true` or `false` should be the default for `keep_error`.

omp/test/matrix/hybrid_kernels.cpp
upsj: Do you mean whether I tested if it failed before? Yes, I did, at least partially. The failures first came up in the benchmarks, but I think I had to work on the fix a while before the tests succeeded.

benchmark/utils/formats.hpp
```diff
@@ -152,6 +153,10 @@ std::string format_command =
 // the formats command-line argument
 DEFINE_string(formats, "coo", formats::format_command.c_str());
 
+DEFINE_uint64(
+    ell_imbalance_limit, 100,
+    "Maximal storage overhead above which ELL benchmarks will be skipped");
```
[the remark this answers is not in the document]
upsj: Do you mean not skipping anything by default? I would probably rather use `std::numeric_limits<uint64_t>::max()` for that.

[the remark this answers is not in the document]
yhmtsai: what value is for not skipping
100 as default is fine.
max() can not be used from command line, can it?

benchmark/conversions/conversions.cpp
yhmtsai: I see. the formats and namespaces have the same name such that I am confused with

benchmark/run_all_benchmarks.sh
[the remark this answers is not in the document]
yhmtsai: maybe we separate them? one for conversion/spmv and one for others.

benchmark/utils/formats.hpp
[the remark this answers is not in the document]
upsj: 1000000000 is sufficient for any inputs, even if it is way below the maximum value.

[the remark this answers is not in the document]
yhmtsai: nit: the counter matrix is existed.
and using -1 can skip going through all nonzero second times.

benchmark/run_all_benchmarks.sh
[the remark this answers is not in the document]
upsj: That is the same behavior we have right now. I don't want to change everything, but the benchmark script does require a lot more work in general (also related to the BJ-specific preconditioner setup etc.)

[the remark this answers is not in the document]
yhmtsai: we do not have this argument before, so it will use the solver default repetition, which is 1 not 10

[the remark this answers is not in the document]
upsj: oh right, we modify the default before parsing. I'll add it.

benchmark/run_all_benchmarks.sh
```diff
@@ -265,6 +280,7 @@ run_preconditioner_benchmarks() {
                 --jacobi_max_block_size="${bsize}" \
                 --jacobi_storage="${prec}" \
                 --device_id="${DEVICE_ID}" --gpu_timer=${GPU_TIMER} \
+                --repetitions="${REPETITIONS}" \
```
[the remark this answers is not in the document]
upsj: This doesn't run multiple SpMVs, so I would use the default number of repetitions.

yhmtsai (approved):
some nits but LGTM

Issue #796, opened 2021-06-17
Clarify the behavioral differences between a dense matrix and a multivector


MarcelKoch:
The current `Dense` matrix can be used simultaneously as a regular dense matrix and as multiple column vectors combined (multivector/MV). In the class documentation, the case of multiple column vectors is even highlighted as a common use case. For most methods, there is no difference in the behavior of a dense matrix and a multivector. However, I think for the following methods some differences exists (**impl** marks the currently implemented version):

| method | Dense behavior | MV behavior|
|-|-|-|
| `apply`| matrix * vector product (**impl**) | - |
|`scale&add_scaled` | Apply the scaling row wise (corresponds to `diag(alpha)*A`, could also support `alpha` as column vector) | Apply the scaling column-wise (**impl**) |
|`dot` | Compute `sum_ij a_ij * b_ij`, i.e. the dot product of the flat vectors | Compute a column-wise dot product (**impl**)|
| `norm` | Compute the appropriate (frobenius, l1, l_inf, ...) matrix norm | Compute a column-wise norm (**impl**)|

Sidenote: The documentation of each method already specifies which behavior is used, but I don't think that is sufficient. I guess most users won't consult the documentation for these methods, which could cause some confusion if the dense behavior is assumed.

IMO, as the name already suggest the `Dense` class should model a dense matrix and the methods should have the expected behavior. If the behavior of a MV is wanted, e.g. for computing the norm of multiple vectors simultaneously, the name of the method should reflect that behavior. Since this change would break interface, it needs to be put on hold until ginkgo 2.0. Nevertheless, I propose the foilowing changes:

1. add new methods with more descriptive names
2. add a deprecation warning to the current methods stating that their behavior will change with the next major release and that the methods from 1. should be used.
3. for the 2.0 release, the behavior of the current methods is changed to the one corresponding to a dense matrix.

upsj:
Stealing a bit of MATLAB's notation, we could also add a (defaulted) parameter to compute_norm2 that for a nxm matrix specifies whether we want column norms (output 1xm), row norms (output nx1) or frobenius norm (output 1x1).
With the Diagonal matrix, we already have a representation of row and column scaling we could use there.
Do you have any use cases in mind for dot products of flattened matrices?

yhmtsai:
compute_norm2 can relay on the output size, so we do not need to introduce another parameter

MarcelKoch:
[...]

PR #820, opened 2021-07-06
Scalar parameter handling

### LINKED ISSUE #816: Simplify the handling of scalar parameters
Passing a scalar to some top-level functions can be quite bothersome. For example, the functions  `LinOp::apply(alpha, b, beta, x)`, `Dense::scale(alpha)`, `Dense::add_scaled(alpha, b)` only accept a `LinOp*` as scalar. Also, passing a `double` would result in a compile-time error. Since this is quite unintuitive, a way to pass a single scalar into these functions should be introduced.

Current ideas:
1. add a new mixin `EnableScalarApplyLinOp` which implements overloads for `LinOp::apply` + add overloads to `Dense`
  - Pro:
    - simple to implement
  - Con:
    - need to add mixin to all matrix formats, solvers, preconditioners, etc
    - re-implementation of `apply`, which is already implemented twice
    - need to add overloads for new functions/mixin to new classes
2. replace `LinOp*` parameter type by a variant-like type, e.g `std::variant<LinOp*, double>`
   - Pro:
     - no additions to exisiting code
     - only requires one definition for new functions
     - clearer distinction that the `alpha` parameter should not be any LinOp
  - Con:
    - requires implementation of `std::variant` (perhaps not all of that)
    - changes API, although user code should not break
3. introduce a function `LinOp* gko::scalar(T)` which creates the corresponding LinOp in a global cache
  - Pro:
    - no overloads/additions to exisiting code
 - Con:
   - handling global cache

I'm mostly in favor of 2., due to the noted pros. The second-best choice would be 1., and 3. is just for reference.

MarcelKoch:
This PR implements solution 1. from #816.

Todo:
- [x] adjust tests to call new overloads
- [ ] (maybe) test all derived classes if the overloads work correctly

Fixes #816


MarcelKoch:
Most solvers and preconditioners use `friend class EnableLinOp<...>`. I don't think this needs to be changed, but I might be wrong on that.

upsj (commented):
LGTM, nice job so far! You could probably also add these changes to some examples, where we had to create scalars for 0, 1 and -1 before. I guess scalar_to_dense is your way of decoupling the dense dependency from lin_op? And as a final comment, we usually try to keep two empty lines between functions and other declarations unless we are inside a class declaration or the two adjacent declarations are directly related, i.e. overloads, template specializations or similar. For more details, see https://github.com/ginkgo-project/ginkgo/wiki/Contributing-guidelines

core/base/utils.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+
+namespace gko {
+
+template <typename ValueType>
+std::unique_ptr<matrix::Dense<ValueType>> scalar_to_dense(
+    const ValueType val, std::shared_ptr<const Executor> exec)
+{
+    return initialize<matrix::Dense<ValueType>>({val}, exec);
+}
```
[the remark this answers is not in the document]
MarcelKoch: I added this, since I would require including dense.hpp if EnableValueTypedLinOp stays where it is. To be honest, I would like to keep it in lin_op.hpp, since it fits there context-wise.

MarcelKoch:
I'm working through the examples, as @upsj suggested, and I'm realizing that there are still some cases where this approach is not helpful. For example, in `custom-logger` the class `ResidualLogger` wants to compute something like `matrix->apply(1, b, -1, x)`, but `matrix` is only available as `LinOp*` and thus the overload does not exist in that case. One could use a dynamic cast to `EnableValuedTypedLinOp` for that case, but I think that is rather clunky.

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -827,6 +827,66 @@ class EnableLinOp
...
+    using EnableLinOp<ConcreteLinOp, PolymorphicBase>::EnableLinOp;
+    using EnableLinOp<ConcreteLinOp, PolymorphicBase>::apply;
+
+    const ConcreteLinOp *apply(const ValueType alpha, const LinOp *b,
+                               const ValueType beta, LinOp *x) const
+    {
+        auto dense_alpha = scalar_to_dense(alpha, this->get_executor());
+        auto dense_beta = scalar_to_dense(beta, this->get_executor());
+        return this->apply(dense_alpha.get(), b, dense_beta.get(), x);
+    }
```
yhmtsai, line 851: I think this part can be templated and put it in the linop directly?
and the scalar_to_dense is free function, so there is no cycle dep

MarcelKoch: I've tried that, and it is possible. However, one problem I haven't figured out is how to handle `EnableLinOp`. If this is part of linop, it should be implemented again in enablelinop (or am I wrong with that?). This is more difficult, since the template needs to be instantiated explicitly somehow, and I don't know how to do this for a member function of an already templated class.

MarcelKoch:
Currently, I think testing only the apply overloads of `EnableValueTypedLinOp` directly is sufficient. These overloads are just wrappers that call `scalar_to_dense` if necessary, so I think they should work in the derived classes.

include/ginkgo/core/base/lin_op.hpp
yhmtsai: I think you does not need to instantiate the template member function

MarcelKoch: After some discussion with Tobias, I've decided to keep this in a separate mixin. Perhaps when linop gets some information about the used value type something like this could be implemented.

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -694,6 +741,20 @@ class Dense
...
+     */
+    void compute_norm2(ValueType *result) const
+    {
+        GKO_ASSERT_EQUAL_COLS(this, dim<2>(1, 1));
+        auto exec = this->get_executor();
+        auto dense_result = Dense::create(
+            exec->get_master(), dim<2>{1, 1},
+            Array<ValueType>::view(exec->get_master(), 1, result), 1);
+        this->compute_norm2(dense_result.get());
+    }
```
upsj, line 756: I'm not sure about this one (same for dot): So far, I think we don't have any places where we pass pointers into the public interface except to initialize arrays. Without any information associated to it, a pointer can be on the host or a device, and we have no way to differentiate between them. Users might try to pass a device pointer to this function by accident?
When we have Frobenius norm calculation in there, we can definitely add an overload that returns a single value, but for now, I think this might be a step in the wrong direction.

include/ginkgo/core/base/mtx_io.hpp
```diff
@@ -141,6 +141,26 @@ inline void write(StreamType &&os, MatrixType *matrix,
...
+ * @param layout  the layout used in the output
+ */
+template <typename ValueType, typename StreamType>
+inline void write(StreamType &&os, ValueType val,
+                  layout_type layout = layout_type::array)
+{
+    matrix_data<ValueType, int32> data{{val}};
+    write_raw(os, data, layout);
+}
+
```
upsj, line 163: If we remove the pointer overloads to compute_dot etc, this will probably also no longer be necessary

upsj (approved):
LGTM for the most part! I'd like to discuss the pointer out-parameter interface though.

include/ginkgo/core/base/lin_op.hpp
yhmtsai: Is there any reason for that?
or any downside from templated member function?
If user does not use the same type as LinOp, ginkgo will handle it via mix_and_match, I think?

upsj: Only within a group (real or complex), but there is no implicit real -> complex or complex -> real conversion

include/ginkgo/core/matrix/dense.hpp
MarcelKoch: You're right, that overload (and the one for dot) is nor really fitting. I will remove it. Perhaps as part of #796 the overload with a single return value can be added.

examples/adaptiveprecision-blockjacobi/adaptiveprecision-blockjacobi.cpp
```diff
@@ -156,7 +154,7 @@ int main(int argc, char *argv[])
     std::cout << "Final residual norm sqrt(r^T r):\n";
     write(std::cout, lend(res));
     std::cout << "Implicit residual norm squared (r^2):\n";
-    write(std::cout, lend(impl_res));
+    gko::write(std::cout, lend(impl_res));
```
upsj, line 157: I think that is no longer necessary due to ADL, and makes it a bit inconsistent with the previous lines. You could either remove `gko::` or add it to all ADL-found functions (except for lend etc)

examples/cb-gmres/cb-gmres.cpp
```diff
@@ -194,25 +194,20 @@ int main(int argc, char *argv[])
...
     auto tmp = gko::clone(gko::lend(b));
 
     // tmp = Ax - tmp
-    A->apply(lend(one), lend(x_keep), lend(neg_one), lend(tmp));
+    A->apply(1.0, lend(x_keep), -1.0, lend(tmp));
     tmp->compute_norm2(lend(res_norm_keep));
 
     std::cout << "\nResidual norm without compression:\n";
-    write(std::cout, lend(res_norm_keep));
+    gko::write(std::cout, lend(res_norm_keep));
```
upsj, line 207: same here

upsj (approved):

examples/adaptiveprecision-blockjacobi/adaptiveprecision-blockjacobi.cpp
MarcelKoch: Of course, seems like I didn't get all of them...

include/ginkgo/core/base/lin_op.hpp
yhmtsai: I think mix_and_match should be able to add the real->complex and limited usage for complex->real when the complex does not contain imag part.
does the mixin allow complex -> real?
I think the compiler does not allow std::complex<double>{3, 0} -> double?

upsj: No, since all conversions need to be symmetric (conditionally inheriting from a type is really difficult), we can't represent this in the type hierarchy. Adding real-to-complex conversion to make_temporary_conversion would be possible though, I'll take a look at that. This would probably need to be implemented in `core/base/lin_op.cpp`, though I am not entirely happy with this function failing e.g. if you call `apply(1, x, 0, y)` with a linker error on the missing `apply(int, LinOp, int, LinOp)`.

core/base/utils.cpp
```diff
@@ -0,0 +1,63 @@
...
+#define GKO_DECLARE_SCALAR_TO_DENSE(_type)                 \
+    std::unique_ptr<matrix::Dense<_type>> scalar_to_dense( \
+        const _type, std::shared_ptr<const Executor>)
+GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE(GKO_DECLARE_SCALAR_TO_DENSE);
+
```
yhmtsai, line 59:
```suggestion


```

core/base/utils.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include <memory>
+
+
+#include <ginkgo/core/base/executor.hpp>
+#include <ginkgo/core/matrix/dense.hpp>
+
+
+namespace gko {
+
```
yhmtsai, line 48:
```suggestion


```

core/test/base/value_typed_lin_op.cpp
```diff
@@ -0,0 +1,176 @@
...
+class DummyValueTypedLinOp
+    : public gko::EnableValueTypedLinOp<DummyValueTypedLinOp<T>, T>,
+      public gko::EnableCreateMethod<DummyValueTypedLinOp<T>> {
+    using Self = DummyValueTypedLinOp<T>;
+
+public:
+    explicit DummyValueTypedLinOp(std::shared_ptr<const gko::Executor> exec,
+                                  gko::dim<2> size = gko::dim<2>{})
+        : gko::EnableValueTypedLinOp<Self, T>(exec, size), value_()
+    {}
```
yhmtsai, line 68:
```suggestion
    {}

```

core/test/base/value_typed_lin_op.cpp
```diff
@@ -0,0 +1,176 @@
...
+template <typename T>
+class DummyValueTypedLinOp
+    : public gko::EnableValueTypedLinOp<DummyValueTypedLinOp<T>, T>,
+      public gko::EnableCreateMethod<DummyValueTypedLinOp<T>> {
+    using Self = DummyValueTypedLinOp<T>;
+
+public:
+    explicit DummyValueTypedLinOp(std::shared_ptr<const gko::Executor> exec,
+                                  gko::dim<2> size = gko::dim<2>{})
+        : gko::EnableValueTypedLinOp<Self, T>(exec, size), value_()
```
yhmtsai, line 67: maybe give zero here?

core/test/base/value_typed_lin_op.cpp
```diff
@@ -0,0 +1,176 @@
...
+};
+
+
+TYPED_TEST_SUITE(EnableValueTypedLinOp, gko::test::ValueTypes);
+
+
+TYPED_TEST(EnableValueTypedLinOp, CanCallExtendedApplyImplLinopLinop)
+{
+    using value_type = typename TestFixture::value_type;
+    this->op->apply(lend(this->alpha), lend(this->b), lend(this->beta),
```
yhmtsai, line 131: this->op can not be LinOP, right?
it needs to be the explicit type or EnableValuedTypeLinOp<T> type

examples/cb-gmres/cb-gmres.cpp
```diff
@@ -138,7 +138,7 @@ int main(int argc, char *argv[])
         b_host->at(i, 0) =
             ValueType{1} / std::sqrt(static_cast<ValueType>(A_size[0]));
     }
-    auto b_norm = gko::initialize<real_vec>({0.0}, exec);
+    auto b_norm = gko::initialize<real_vec>({0.0}, exec->get_master());
```
yhmtsai, line 141: b_norm can stay on exec not host

examples/cb-gmres/cb-gmres.cpp
```diff
@@ -194,25 +194,20 @@ int main(int argc, char *argv[])
...
-    // uniform interface when computing on any device. To compute the residual,
-    // the (advanced) apply method is used.
-    auto one = gko::initialize<vec>({1.0}, exec);
-    auto neg_one = gko::initialize<vec>({-1.0}, exec);
-
-    auto res_norm_keep = gko::initialize<real_vec>({0.0}, exec);
-    auto res_norm_reduce = gko::initialize<real_vec>({0.0}, exec);
+    // To compute the residual, the (advanced) apply method is used.
+    auto res_norm_keep = gko::initialize<real_vec>({0.0}, exec->get_master());
+    auto res_norm_reduce = gko::initialize<real_vec>({0.0}, exec->get_master());
```
yhmtsai, line 199: also here

examples/custom-logger/custom-logger.cpp
```diff
@@ -343,18 +343,14 @@ int main(int argc, char *argv[])
...
-    auto one = gko::initialize<vec>({1.0}, exec);
-    auto neg_one = gko::initialize<vec>({-1.0}, exec);
-    auto res = gko::initialize<real_vec>({0.0}, exec);
-    A->apply(gko::lend(one), gko::lend(x), gko::lend(neg_one), gko::lend(b));
-    b->compute_norm2(gko::lend(res));
+    // the error of the solution. To compute the residual, all you need to do is
+    // call the apply method, which in this case is an spmv and equivalent to
+    // the LAPACK z_spmv routine. Finally, you compute the euclidean 2-norm with
+    // the compute_norm2 function.
+    auto res = gko::initialize<real_vec>({0.0}, exec->get_master());
```
yhmtsai, line 350: also

examples/custom-logger/custom-logger.cpp
```diff
@@ -343,18 +343,14 @@ int main(int argc, char *argv[])
...
+    A->apply(1.0, gko::lend(x), -1.0, gko::lend(b));
+    b->compute_norm2(lend(res));
 
     std::cout << "Residual norm sqrt(r^T r):\n";
-    write(std::cout, gko::lend(res));
+    write(std::cout, lend(res));
```
yhmtsai, line 355: this is confusing or not strong concern in ginkgo.
we use `gko::lend` not `lend` even if the compiler can find the correct lend from the input type.

examples/custom-stopping-criterion/custom-stopping-criterion.cpp
```diff
@@ -127,10 +127,8 @@ void run_solver(volatile bool *stop_iteration_process,
     write(std::cout, lend(x));
 
     // Calculate residual
-    auto one = gko::initialize<vec>({1.0}, exec);
-    auto neg_one = gko::initialize<vec>({-1.0}, exec);
-    auto res = gko::initialize<real_vec>({0.0}, exec);
-    A->apply(lend(one), lend(x), lend(neg_one), lend(b));
+    auto res = gko::initialize<real_vec>({0.0}, exec->get_master());
```
yhmtsai, line 130: also

examples/mixed-precision-ir/mixed-precision-ir.cpp
```diff
@@ -168,24 +165,24 @@ int main(int argc, char *argv[])
         inner_solution->convert_to(lend(outer_delta));
 
         // x = x + inner_solution
-        x->add_scaled(lend(one), lend(outer_delta));
+        x->add_scaled(1.0, lend(outer_delta));
 
         // residual = b - A * x
         outer_residual->copy_from(lend(b));
-        A->apply(lend(neg_one), lend(x), lend(one), lend(outer_residual));
+        A->apply(-1.0, lend(x), 1.0, lend(outer_residual));
```
yhmtsai, line 172: we report timing here, so maybe staying original one should be better

examples/mixed-spmv/mixed-spmv.cpp
```diff
@@ -257,31 +255,31 @@ int main(int argc, char *argv[])
...
     lplp_diff->compute_norm2(lend(lplp_diff_norm));
-    lphp_diff->add_scaled(lend(neg_one), lend(lphp_x));
+    lphp_diff->add_scaled(-1.0, lend(lphp_x));
     lphp_diff->compute_norm2(lend(lphp_diff_norm));
     exec->synchronize();
 
     std::cout.precision(10);
     std::cout << std::scientific;
     std::cout << "High Precision time(s): " << hp_sec << std::endl;
-    std::cout << "High Precision result norm: " << hp_x_norm->at(0)
```
yhmtsai, line 273: at(0) is equal to get_values()[0], I think.

examples/mixed-spmv/mixed-spmv.cpp
```diff
@@ -257,31 +255,31 @@ int main(int argc, char *argv[])
...
+              << hplp_diff_norm->get_values()[0] / hp_x_norm->get_values()[0]
+              << "\n";
     std::cout << "Lp * Lp -> Hp time(s): " << lplp_sec << std::endl;
     std::cout << "Lp * Lp -> Hp relative error: "
-              << lplp_diff_norm->at(0) / hp_x_norm->at(0) << "\n";
-    std::cout << "Lp * Hp -> Hp time(s): " << lplp_sec << std::endl;
-    std::cout << "Lp * Hp -> Hp relative error: "
-              << lphp_diff_norm->at(0) / hp_x_norm->at(0) << "\n";
+              << lplp_diff_norm->get_values()[0] / hp_x_norm->get_values()[0]
+              << "\n";
```
yhmtsai, line 284: some additional delete?

include/ginkgo/core/base/lin_op.hpp
yhmtsai: In current one, you also need to know which type of linop at least which valuetype of linop.
both cases you need to know about that.
it throws error when create dense with int.
if scalar_to_dense have a specialized to handle it, it should be okay.
integer will use float to represent.

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -827,6 +827,66 @@ class EnableLinOp
...
+          typename PolymorphicBase = LinOp>
+class EnableValueTypedLinOp
+    : public EnableLinOp<ConcreteLinOp, PolymorphicBase> {
+public:
+    using value_type = ValueType;
+
+    using EnableLinOp<ConcreteLinOp, PolymorphicBase>::EnableLinOp;
+    using EnableLinOp<ConcreteLinOp, PolymorphicBase>::apply;
+
+    const ConcreteLinOp *apply(const ValueType alpha, const LinOp *b,
```
yhmtsai, line 845: need documentation

include/ginkgo/core/base/mtx_io.hpp
```diff
@@ -120,10 +120,10 @@ inline std::unique_ptr<MatrixType> read(StreamType &&is, MatrixArgs &&... args)
 
 
 /**
- * Reads a matrix stored in matrix market format from an input stream.
+ * Writes a matrix into an output in the matrix market format.
```
yhmtsai, line 123: good catch!

include/ginkgo/core/base/utils_helper.hpp
```diff
@@ -432,6 +432,15 @@ class null_deleter<T[]> {
...
 
+namespace matrix {
+template <typename ValueType>
+class Dense;
+}
+
+template <typename ValueType>
+std::unique_ptr<matrix::Dense<ValueType>> scalar_to_dense(
+    ValueType val, std::shared_ptr<const Executor> exec);
+
```
yhmtsai, line 443:
```suggestion
namespace matrix {


template <typename ValueType>
class Dense;


}


template <typename ValueType>
std::unique_ptr<matrix::Dense<ValueType>> scalar_to_dense(
    ValueType val, std::shared_ptr<const Executor> exec);


```

yhmtsai (changes requested):
nit format and need some additional documentation.
some discussion on the EnableValueTypeLinOp

include/ginkgo/core/base/lin_op.hpp
upsj: To make my example more concrete:
```cpp
    template <typename ValueType>
    void apply(const ValueType alpha, const LinOp *b,
                               const ValueType beta, LinOp *x) const
    {
        auto dense_alpha = scalar_to_dense(alpha, this->get_executor());
        auto dense_beta = scalar_to_dense(beta, this->get_executor());
        return this->apply(dense_alpha.get(), b, dense_beta.get(), x);
    }
    ...
    op->apply(1, x, 0, y);
```
causes a linker error because scalar_to_dense<int> is not instantiated.

On the other hand, in a non-templated ValueType overload
```cpp
   op->apply(1, x, 0, y);
```
casts 1 and 0 to ValueType and at worst gives a warning about the mismatching types.

I'm trying to avoid the use of unrestricted template member functions, since users don't get a compilation error at compile time, but only a much less useful error at link time complaining about a missing scalar_to_dense<int> symbol. I believe having the concrete type in here makes for much better type-safety in the common use case, which is using concrete types instead of generic LinOp types.

yhmtsai: the template one should be good for using.
```
auto dense_alpha = scalar_to_dense(int, this->get_executor);
// gives Dense<float> or Dense<double> with some addition on scalar_to_dense
```
it will not complain the scalar_to_dense<int>, when we add the conversion.
`op->apply(1, x, 0, y);` will gives implicitly conversion to ValueType, I think.

yhmtsai: the apply will accept different input when the object is concrete type or linop
what's this target for? for all user, the developer using ginkgo or ginkgo's developer.

upsj: > it will not complain the scalar_to_dense, when we add the conversion.
> op->apply(1, x, 0, y); will gives implicitly conversion to ValueType, I think.

Yes, that is intended. Template type deduction requires the exact type to match, so scalar_to_dense(...) would try to use the non-existent `scalar_to_dense<int>` instantiation and cause a linker error. On the other hand, with an explicit non-template overload, even implicit conversions can be applied, so for a complex LinOp, you could have something like
`apply(1.0, x, std::complex<double>{0.0, 1.0}, y)` which works in the non-templated setting but doesn't compile in the templated setting.

> what's this target for? for all user, the developer using ginkgo or ginkgo's developer.

I would imagine that this is most useful for users and maybe example development. For generic library code and things like solvers, where the repeated allocations are not negligible, I would use explicit LinOps anyways.

> the apply will accept different input when the object is concrete type or linop

We have similar differences e.g. in the `apply` function, where `LinOp::apply` returns `LinOp*`, but `EnableLinOp::apply` returns a `ConcreteLinOp*`. More generally, only concrete types expose all functionality that the type has explicitly, which I would be totally fine with.

yhmtsai: maybe I put thing together such that it is misleading.
```
// for integer
Dense<float> scalar_to_dense<int>(...)
// for floating
Dense<float> scalar_to_dense<float>(...)
```
If we have this, the conversion should be okay.
current `op->apply(1, x, 0, y)` gives implicitly conversion to ValueType as templated one.
current pr conversion is on host but templated one may be on device.

`op->apply(1.0, x, std::complex<double>{0.0, 1.0}, y)` can be handled by two templated parameter for alpha and beta.

It can used when user only create the object and then use it.
If they do some combination, the type could be `LinOp*`

In my feeling, when I know concrete type, it is easy to use scalar_to_dense.
when I have concrete type, it is easy to know the valuetype and its executor, so I can easily to use these parameter to generate alpha/beta.
but when I have LinOp, I only have executor, so I need to use `dynamic_cast` through all possible types (and maybe more template params needs to know) or pick one valuetype to generate via `scalar_to_dense` thanks to `mixed_and_match`.

If we does not want `scalar_to_dense<int>` return `Dense<float>`, maybe the templated one needs to request the input always be floating point. Also, it will lead `op->apply(scalar_to_dense(1, ref), ...)` fail and `op->apply(1, ...)` work in current pr.
If we allow `scalar_to_dense<int>` return `Dense<float>`, it will be good for templated one.

It's good for example, but we start to have some LinOp like Multigrid without ValueType, so it will not always usable.
(Out of topic, I am thinking whether CG does really need ValueType because it just depends on the input matrix not the operation).
I think it is like the mixed_and_match place. It is a good tool to investigate ginkgo or develop something in ginkgo.

current pr:
pro: the conversion is host, do not contain any conversion on device.
con: needs to be concrete type or know the valuetype
templated:
pro: work for all linop
con: does not use the value type if it has one, so it might need conversion on device again.
If the linop does not implement the mixed_and_match, it will throw the link issue or runtime error?

Dense->apply return Dense and LinOp->apply return LinOP seems to be the same for me
I agree with you. only concreate type contains all functionalities but originally it's means different function name of the class to me. (maybe related to it, the naming issue `compute_absolute_linop` and `compute_absoulte`, sorry for out of topic again)

MarcelKoch:
I'm closing this for now. ATM, there is no good way of adding the functionality.

PR #821, opened 2021-07-06
Some utilities for span and dim

pratikvn:
This PR adds some useful utilities for dim and span.

1. An ostream << operator overload for dim.
~~2. Adding two spans.~~
3. Getting the length of a span.

### DIFF
--- a/include/ginkgo/core/base/dim.hpp
+++ b/include/ginkgo/core/base/dim.hpp
@@ -164,8 +164,7 @@ struct dim {
      *
      * @return a stream object appended with the dim output
      */
-    friend GKO_ATTRIBUTES std::ostream &operator<<(std::ostream &os,
-                                                   const dim &x)
+    friend std::ostream &operator<<(std::ostream &os, const dim &x)
     {
         os << "(";
         x.print_to(os);
@@ -229,8 +228,7 @@ struct dim<1u, DimensionType> {
         return dim(x.first_ * y.first_);
     }
 
-    friend GKO_ATTRIBUTES std::ostream &operator<<(std::ostream &os,
-                                                   const dim &x)
+    friend std::ostream &operator<<(std::ostream &os, const dim &x)
     {
         os << "(";
         x.print_to(os);
--- a/include/ginkgo/core/base/dim.hpp
+++ b/include/ginkgo/core/base/dim.hpp
@@ -34,6 +34,9 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define GKO_PUBLIC_CORE_BASE_DIM_HPP_
 
 
+#include <iostream>
+
+
 #include <ginkgo/core/base/types.hpp>
 
 
@@ -51,6 +54,7 @@ namespace gko {
 template <size_type Dimensionality, typename DimensionType = size_type>
 struct dim {
     static constexpr size_type dimensionality = Dimensionality;
+    friend class dim<dimensionality + 1>;
 
     using dimension_type = DimensionType;
 
@@ -163,10 +167,20 @@ struct dim {
     friend GKO_ATTRIBUTES std::ostream &operator<<(std::ostream &os,
                                                    const dim &x)
     {
-        return os << "(" << x.first_ << " , " << x.rest_ << ")";
+        os << "(";
+        x.print_to(os);
+        os << ")";
+        return os;
     }
 
 private:
+    void inline print_to(std::ostream &os) const
+    {
+        os << first_ << ", ";
+        rest_.print_to(os);
+    }
+
+
     constexpr GKO_ATTRIBUTES dim(const dimension_type first,
                                  dim<dimensionality - 1> rest)
         : first_{first}, rest_{rest}
@@ -181,6 +195,7 @@ private:
 template <typename DimensionType>
 struct dim<1u, DimensionType> {
     static constexpr size_type dimensionality = 1u;
+    friend class dim<2>;
 
     using dimension_type = DimensionType;
 
@@ -217,10 +232,15 @@ struct dim<1u, DimensionType> {
     friend GKO_ATTRIBUTES std::ostream &operator<<(std::ostream &os,
                                                    const dim &x)
     {
-        return os << x.first_;
+        os << "(";
+        x.print_to(os);
+        os << ")";
+        return os;
     }
 
 private:
+    void inline print_to(std::ostream &os) const { os << first_; }
+
     dimension_type first_;
 };
 
--- a/include/ginkgo/core/base/dim.hpp
+++ b/include/ginkgo/core/base/dim.hpp
@@ -152,6 +152,20 @@ struct dim {
         return dim(x.first_ * y.first_, x.rest_ * y.rest_);
     }
 
+    /**
+     * A stream operator overload for dim
+     *
+     * @param os  stream object
+     * @param x  dim object
+     *
+     * @return a stream object appended with the dim output
+     */
+    friend GKO_ATTRIBUTES std::ostream &operator<<(std::ostream &os,
+                                                   const dim &x)
+    {
+        return os << "(" << x.first_ << " , " << x.rest_ << ")";
+    }
+
 private:
     constexpr GKO_ATTRIBUTES dim(const dimension_type first,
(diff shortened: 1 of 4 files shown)


include/ginkgo/core/base/range.hpp
```diff
@@ -155,6 +166,13 @@ GKO_ATTRIBUTES GKO_INLINE constexpr bool operator!=(const span &first,
 }
 
 
+GKO_ATTRIBUTES GKO_INLINE constexpr span operator+(const span &first,
+                                                   const span &second)
+{
+    return gko::span{first.begin + second.begin, first.end + second.end};
+}
+
+
```
upsj, line 175: Could you make this a `friend` inside `span` instead? That prevents accidental implicit conversions when looking for suitable `operator+` overloads. Also, what use case do you have in mind for this operation? span + index I can see, but span + span looks a bit unusual to me. Maybe make this a named operation instead?

include/ginkgo/core/base/dim.hpp
```diff
@@ -152,6 +152,20 @@ struct dim {
...
+     *
+     * @param os  stream object
+     * @param x  dim object
+     *
+     * @return a stream object appended with the dim output
+     */
+    friend GKO_ATTRIBUTES std::ostream &operator<<(std::ostream &os,
+                                                   const dim &x)
+    {
+        return os << "(" << x.first_ << " , " << x.rest_ << ")";
```
upsj, line 166: That will lead to a gang of owls `(0,(0,(0,0)))` for higher dimensionalities. I guess that might not be what we want?

include/ginkgo/core/base/range.hpp
pratikvn: Mainly when combining two contiguous spans. Yeah, I think it might make sense to make this a named operation.

upsj: Combining two contiguous spans to me sounds like [a,b) + [b,c) = [a,c), but that is not what this operation is intended for, right?

include/ginkgo/core/base/dim.hpp
pratikvn: I am not sure how else to accurately represent this. I guess our main use case would be `dim<2>`, for which this output should be okay.

include/ginkgo/core/base/range.hpp
pratikvn: Yes, good catch!  I confused this with my usage of separate usages of row_spans and column_spans. `[a,b) + [b,c) = [a,c)` is the right way to go. I will update this.

upsj: But that operation is only defined if the end of one range does not match the beginning of the next, so I would be really careful about adding something like this.

pratikvn: Yeah, I agree. I removed it for now.

include/ginkgo/core/base/dim.hpp
thoasm: I would suggest this:
Unfortunately, I need the `friend` declaration with the `dim<N+1>`, so I can call `rest_.print_to(os)`. It does seem to work properly, but it is more code than your suggestion.
```
#include <iostream>

template <int N>
class dim {
    static_assert(N > 1, "Internal Assert!");
    friend class dim<N + 1>;

   public:
    template <typename... Args>
    dim(int first, Args &&... args)
        : current_{first}, rest_(std::forward<Args>(args)...) {}

    friend std::ostream &operator<<(std::ostream &os, const dim &x) {
        os << '(';
        x.print_to(os);
        os << ')';
        return os;
    }

   private:
    void inline print_to(std::ostream &os) const {
        os << current_ << ", ";
        rest_.print_to(os);
    }

    int current_;
    dim<N - 1> rest_;
};

template <>
class dim<1> {
    friend class dim<2>;

   public:
    dim(int first) : current_{first} {}

    friend std::ostream &operator<<(std::ostream &os, const dim &x) {
        os << '(';
        x.print_to(os);
        os << ')';
        return os;
    }

   private:
    void inline print_to(std::ostream &os) const { os << current_; }

    int current_;
};

template <>
class dim<0> {
   public:
    dim() {}

    friend std::ostream &operator<<(std::ostream &os, const dim &x) {
        os << '(';
        os << ')';
        return os;
    }

   private:
    void inline print_to(std::ostream &os) const {}
};

int main() {
    dim<3> test3(1, 2, 3);
    dim<2> test2(5, 6);
    dim<1> test1(7);
    dim<0> test0;

    std::cout << test3 << '\n'
              << test2 << '\n'
              << test1 << '\n'
              << test0 << '\n';
}
```

pratikvn: Thanks! That works quite nicely.

include/ginkgo/core/base/range.hpp
```diff
@@ -101,6 +101,17 @@ struct span {
...
     constexpr bool is_valid() const { return begin < end; }
 
+    /**
+     * Returns the length of a span.
+     *
+     * @return `this->end - this->begin`
+     */
+    constexpr size_type length() const
+    {
+        GKO_ASSERT(is_valid());
```
tcojean, line 111: `static_assert`?

tcojean (approved):
LGTM. Do we add them to the accessor equivalents as well? E.g., the length to `index_span.hpp`.

include/ginkgo/core/base/range.hpp
pratikvn: I dont think that is possible as begin and end need to be accessed, which are not always constexpr

tcojean: I guess we use it in a dynamic context as well even if this is constexpr. Thanks for checking.

pratikvn: Yes, in our code we use runtime defined versions in many places. But in some user's code, if the span in defined at compile time, the constexpr would kick in, I think.

include/ginkgo/core/base/dim.hpp
```diff
@@ -152,7 +156,31 @@ struct dim {
...
 
+    /**
+     * A stream operator overload for dim
+     *
+     * @param os  stream object
+     * @param x  dim object
+     *
+     * @return a stream object appended with the dim output
+     */
+    friend GKO_ATTRIBUTES std::ostream &operator<<(std::ostream &os,
```
thoasm, line 167: nit: this overload is not working on devices, therefore:
```suggestion
    friend std::ostream &operator<<(std::ostream &os,
```

include/ginkgo/core/base/dim.hpp
```diff
@@ -200,7 +229,18 @@ struct dim<1u, DimensionType> {
         return dim(x.first_ * y.first_);
     }
 
+    friend GKO_ATTRIBUTES std::ostream &operator<<(std::ostream &os,
```
thoasm, line 232:
```suggestion
    friend std::ostream &operator<<(std::ostream &os,
```

include/ginkgo/core/base/range.hpp
```diff
@@ -101,6 +101,17 @@ struct span {
      */
     constexpr bool is_valid() const { return begin < end; }
```
thoasm, line 102: Maybe add `GKO_ATTRIBUTES` to this function.

include/ginkgo/core/base/range.hpp
```diff
@@ -101,6 +101,17 @@ struct span {
...
 
+    /**
+     * Returns the length of a span.
+     *
+     * @return `this->end - this->begin`
+     */
+    constexpr size_type length() const
```
thoasm, line 109: Same here:
```suggestion
    constexpr GKO_ATTRIBUTES size_type length() const
```

thoasm (approved):
LGTM!

yhmtsai:
It needs to rebase and then can be merged?

pratikvn:
Yes, but it does not seem to run the full test ?

pratikvn:
rebase!

yhmtsai:
if gitlab pipeline runs after you adding the label `ST:ready-to-merge` or `ST:run-full-test`, the pipeline will trigger the full pipeline.
Otherwise, you need to manually trigger it by clicking rerun of `trigger-pipeline` after adding the label.

yhmtsai:
rebase!

PR #828, opened 2021-07-07
Use consistent matrix index type in benchmark

upsj:
To prepare for larger benchmark types, I added a index type (32 bit for now) to all benchmarks and use it in all related types (Feel free to double-check I didn't miss anything!)

### DIFF
--- a/benchmark/utils/cuda_linops.hpp
+++ b/benchmark/utils/cuda_linops.hpp
@@ -685,11 +685,11 @@ private:
 
 
 // Some shortcuts
-using cusp_csrex = detail::CuspCsrEx<etype>;
+using cusp_csrex = detail::CuspCsrEx<etype, itype>;
 #if defined(CUDA_VERSION) && (CUDA_VERSION < 11000)
-using cusp_csr = detail::CuspCsr<etype>;
-using cusp_csrmp = detail::CuspCsrmp<etype>;
-using cusp_csrmm = detail::CuspCsrmm<etype>;
+using cusp_csr = detail::CuspCsr<etype, itype>;
+using cusp_csrmp = detail::CuspCsrmp<etype, itype>;
+using cusp_csrmm = detail::CuspCsrmm<etype, itype>;
 #endif  // defined(CUDA_VERSION) && (CUDA_VERSION < 11000)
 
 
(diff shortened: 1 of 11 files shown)


benchmark/utils/cuda_linops.hpp
```diff
@@ -710,10 +709,10 @@ using cusp_gcoo = detail::CuspGenericCoo<etype>;
 
 #if defined(CUDA_VERSION) && (CUDA_VERSION < 11000)
 using cusp_coo =
-    detail::CuspHybrid<etype, gko::int32, CUSPARSE_HYB_PARTITION_USER, 0>;
+    detail::CuspHybrid<etype, itype, CUSPARSE_HYB_PARTITION_USER, 0>;
```
yhmtsai, line 712: do they support int64?

benchmark/utils/hip_linops.hip.hpp
```diff
@@ -322,14 +322,14 @@ class HipspHybrid
...
 
 
 using hipsp_coo =
-    detail::HipspHybrid<etype, gko::int32, HIPSPARSE_HYB_PARTITION_USER, 0>;
+    detail::HipspHybrid<etype, itype, HIPSPARSE_HYB_PARTITION_USER, 0>;
 using hipsp_ell =
-    detail::HipspHybrid<etype, gko::int32, HIPSPARSE_HYB_PARTITION_MAX, 0>;
-using hipsp_hybrid = detail::HipspHybrid<etype>;
+    detail::HipspHybrid<etype, itype, HIPSPARSE_HYB_PARTITION_MAX, 0>;
+using hipsp_hybrid = detail::HipspHybrid<etype, itype>;
```
yhmtsai, line 333: also here.

yhmtsai (approved):
matrix_statistics extract_matrix_statistics use matrix_data<etype, int64>.
does it always use int64?

benchmark/utils/cuda_linops.hpp
upsj: I am not certain, I mainly wanted to avoid compilation issues with the wrong matrix_data index type if we do add int64 benchmarks at some point.

upsj:
@yhmtsai Yes, I would leave the index type for matrix_statistics, since this avoids any values out of range for the following operations.

benchmark/utils/cuda_linops.hpp
yhmtsai: I see, when we really this for int64 and get error, we can add the throw at that point

thoasm (approved):
LGTM!

upsj:
rebase!

PR #830, opened 2021-07-09
Fixes handling of multiple worktrees

MarcelKoch:
Uses the correct .git directory in case of multiple worktrees created by `git worktree add`.

WIP: Waiting on ginkgo-project/git-cmake-format#5

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -17,7 +17,9 @@ Ginkgo adds the following additional switches to control what is being built:
 
 *   `-DGINKGO_DEVEL_TOOLS={ON, OFF}` sets up the build system for development
     (requires clang-format, will also download git-cmake-format),
-    default is `OFF`.
+    default is `OFF`. If it is set to `ON`, a pre-commit hook for formatting
+    will be installed. This hook may overwrite a user defined pre-commit hook
+    when Ginkgo is used as a submodule.
 *   `-DGINKGO_MIXED_PRECISION={ON, OFF}` compiles true mixed-precision kernels
     instead of converting data on the fly, default is `OFF`.
     Enabling this flag increases the library size, but improves performance of
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -17,9 +17,11 @@ Ginkgo adds the following additional switches to control what is being built:
 
 *   `-DGINKGO_DEVEL_TOOLS={ON, OFF}` sets up the build system for development
     (requires clang-format, will also download git-cmake-format),
-    default is `OFF`. If it is set to `ON`, a pre-commit hook for formatting
-    will be installed. This hook may overwrite a user defined pre-commit hook
-    when Ginkgo is used as a submodule.
+    default is `OFF`. The default behavior installs a pre-commit hook, which
+    disables git commits.  If it is set to `ON`, a new pre-commit hook for
+    formatting will be installed (enabling commits again). In both cases the
+    hook may overwrite a user defined pre-commit hook when Ginkgo is used as
+    a submodule.
 *   `-DGINKGO_MIXED_PRECISION={ON, OFF}` compiles true mixed-precision kernels
     instead of converting data on the fly, default is `OFF`.
     Enabling this flag increases the library size, but improves performance of
--- a/third_party/dummy-hook/CMakeLists.txt
+++ b/third_party/dummy-hook/CMakeLists.txt
@@ -1,18 +1,27 @@
-if(EXISTS "${Ginkgo_SOURCE_DIR}/.git")
-    set(ADD_HOOK FALSE)
-    set(HOOK_LOCATION "${Ginkgo_SOURCE_DIR}/.git/hooks/pre-commit")
-    if(NOT EXISTS "${HOOK_LOCATION}")
-        set(ADD_HOOK TRUE)
-    else()
-        # check if the correct hook is installed
-        execute_process(COMMAND grep git-cmake-format.py "${HOOK_LOCATION}"
-                        RESULT_VARIABLE res OUTPUT_QUIET)
-        # return value =/= 0 means the pattern was not found
-        if(NOT res EQUAL 0)
+find_package(Git QUIET)
+if(GIT_FOUND)
+    execute_process(COMMAND ${GIT_EXECUTABLE} rev-parse --git-path hooks
+      WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+      OUTPUT_VARIABLE Ginkgo_GIT_HOOKS_DIR
+      OUTPUT_STRIP_TRAILING_WHITESPACE)
+    get_filename_component(Ginkgo_GIT_HOOKS_DIR ${Ginkgo_GIT_HOOKS_DIR} REALPATH BASE_DIR ${CMAKE_SOURCE_DIR})
+
+    if(EXISTS "${Ginkgo_GIT_HOOKS_DIR}")
+        set(ADD_HOOK FALSE)
+        set(HOOK_LOCATION "${Ginkgo_GIT_HOOKS_DIR}/pre-commit")
+        if(NOT EXISTS "${HOOK_LOCATION}")
             set(ADD_HOOK TRUE)
+        else()
+            # check if the correct hook is installed
+            execute_process(COMMAND grep git-cmake-format.py "${HOOK_LOCATION}"
+                            RESULT_VARIABLE res OUTPUT_QUIET)
+            # return value =/= 0 means the pattern was not found
+            if(NOT res EQUAL 0)
+                set(ADD_HOOK TRUE)
+            endif()
         endif()
-    endif()
-    if(ADD_HOOK)
-        configure_file(dummy_hook "${HOOK_LOCATION}" COPYONLY)
+        if(ADD_HOOK)
+            configure_file(dummy_hook "${HOOK_LOCATION}" COPYONLY)
+        endif()
     endif()
 endif()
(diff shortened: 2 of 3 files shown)


third_party/dummy-hook/CMakeLists.txt
```diff
@@ -1,6 +1,11 @@
-if(EXISTS "${Ginkgo_SOURCE_DIR}/.git")
+execute_process(COMMAND ${GIT_EXECUTABLE} rev-parse --absolute-git-dir
+        WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+        OUTPUT_VARIABLE Ginkgo_GIT_DIR
+        OUTPUT_STRIP_TRAILING_WHITESPACE)
+
+if(EXISTS "${Ginkgo_GIT_DIR}")
     set(ADD_HOOK FALSE)
-    set(HOOK_LOCATION "${Ginkgo_SOURCE_DIR}/.git/hooks/pre-commit")
+    set(HOOK_LOCATION "${Ginkgo_SOURCE_DIR}/hooks/pre-commit")
```
upsj, line 8:
```suggestion
    set(HOOK_LOCATION "${Ginkgo_GIT_DIR}/hooks/pre-commit")
```

upsj (approved):
LGTM! And the dummy commits should probably go, as well ;)

yhmtsai (commented):
LGTM

yhmtsai (approved):
LGTM. (forget to change selection)

upsj (changes requested):
I just tested this with a worktree `ginkgo-tmp2`, and the pre-commit dummy hook doesn't get created: `Ginkgo_GIT_DIR` points to `/home/tobi/ginkgo-tmp/.git/worktrees/ginkgo-tmp2` instead of `.home/tobi/ginkgo-tmp/.git/` directly

upsj:
Now that I think of it, other cases we may need to catch: Ginkgo as a `git submodule`?

MarcelKoch:
@upsj I think that is exactly how it is supposed to work. The git-dir for the worktree resides within the git-dir of the root repository, which has the effect you mentioned. The hook should then be installed within this directory, ie `.../ginkgo/.git/ginkgo-tmp/hooks`.

On your second point regarding submodules, I need to think about that.

upsj:
It looks like `git` doesn't execute hooks stored in the worktree subdirectory, only ones in the root `.git/hooks`. For submodules, we should probably just skip installing the hook altogether.

third_party/dummy-hook/CMakeLists.txt
```diff
@@ -1,6 +1,11 @@
-if(EXISTS "${Ginkgo_SOURCE_DIR}/.git")
+execute_process(COMMAND ${GIT_EXECUTABLE} rev-parse --absolute-git-dir
+        WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+        OUTPUT_VARIABLE Ginkgo_GIT_DIR
```
yhmtsai, line 3: in the first cmake from scratch, the Ginkgo_GIT_DIR is empty

yhmtsai (changes requested):

yhmtsai:
which is the path you want to use?
I run `git rev-parse --absolute-git-dir` in the worktree folder.
it gives `/home/mike/worktree/ginkgo/.git/worktrees/second`, is it correct?

third_party/dummy-hook/CMakeLists.txt
MarcelKoch: Sorry about the confusion, but this PR is still WIP. It needs https://github.com/ginkgo-project/git-cmake-format/pull/6 and some additional changes to actually work

yhmtsai: I change the commit and the repo in local for testing.
you need to add `find_package(Git QUIET)`
the first run will not include Git yet, which is included in ginkgo_git_information, so the GIT_EXECUTABLE is notfound
Thus, git is available in the second time.
and maybe you also need to change the ginkgo_git_information there.

MarcelKoch:
Another edge case, that could be thought about: what should happen if the user switches between `GINKGO_DEVEL_TOOLS=ON|OFF` within the same repository? If the user goes `OFF->ON->OFF` in the current version, the dummy hook does not get installed, i.e. the user stays with `ON`. Should that be the case, or should the hook be reset?

upsj:
Yes, that is (was?) intended behavior. When I added it, I wanted to avoid accidentally committing unformatted code, but not unnecessarily hinder commits even after turning it off, since the necessary files were already present. Though of course that broke when recreating the `build` directory completely, which is why your changes are really useful.

third_party/dummy-hook/CMakeLists.txt
```diff
@@ -1,6 +1,12 @@
-if(EXISTS "${Ginkgo_SOURCE_DIR}/.git")
+execute_process(COMMAND ${GIT_EXECUTABLE} rev-parse --git-path hooks
```
yhmtsai, line 1: you need to find git first

third_party/dummy-hook/CMakeLists.txt
```diff
@@ -1,18 +1,27 @@
...
+        if(NOT EXISTS "${HOOK_LOCATION}")
             set(ADD_HOOK TRUE)
+        else()
+            # check if the correct hook is installed
+            execute_process(COMMAND grep git-cmake-format.py "${HOOK_LOCATION}"
+                            RESULT_VARIABLE res OUTPUT_QUIET)
+            # return value =/= 0 means the pattern was not found
+            if(NOT res EQUAL 0)
+                set(ADD_HOOK TRUE)
+            endif()
```
yhmtsai, line 21: this does not work as expect

MarcelKoch: Could you describe the problem? For me, it does the right thing, i.e. adding the dummy hook if pre-commit does not contain `git-cmake-format`.

third_party/git-cmake-format/CMakeLists.txt
```diff
@@ -3,10 +3,14 @@ include(FetchContent)
...
-    GIT_TAG        be9554a5d71030dcc23f1eb355c41ab9d876e776
+    GIT_TAG        069338e8afd47e6a27e661940cd36e940253f81c
 )
 FetchContent_GetProperties(git_cmake_format)
 if(NOT git_cmake_format_POPULATED)
     FetchContent_Populate(git_cmake_format)
+
+    if(NOT DEFINED GCF_FORCE_OVERWRITE)
+        set(GCF_FORCE_OVERWRITE ON CACHE INTERNAL "If true, always overwrite pre-commit hook and script")
+    endif()
```
yhmtsai, line 14: I think it needs to be before calling `FetchContent_Populate` or `FetchContent_GetProperties`.
Using the defined might be confusing.
`cmake -DGCF_FORCE_OVERWRITE=ON ..` it still turn off GCF_FORCE_OVERWRITE.
I am not sure whether FetchContent can realize the config change in the second time or not

MarcelKoch: As far as I understand it, FetchContent does not execute any cmake code of the fetched content. That happens only in the `add_subdirectory` call, so I think it should be fine to set the variable here. I'm not sure what you mean in your `cmake -DGCF_FORCE_OVERWRITE=ON ..` part (assuming you also set `GINKGO_DEVEL_TOOLS`). One problem might be the `INTERNAL` option, which hides `GCF_FORCE_OVERWRITE` at least when using ccmake. But at least for me, it works as expected, i.e. setting it to `ON` if `-DGCF_FORCE_OVERWRITE` is not used and using that value otherwise.

third_party/dummy-hook/CMakeLists.txt
yhmtsai: I see. Sorry, I misunderstand this line.
I thought it check git-cmake-format.py is existed or not.
but you check the git-cmake-format.py inside the pre-commit file

third_party/git-cmake-format/CMakeLists.txt
yhmtsai: does cmake allow -DGCF_FORCE_OVERWRITE directly?
my cmake complain it `VAR:type=value`
I mean `cmake -DGCF_FORCE_OVERWRITE=OFF ..`  and `cmake -DGCF_FORCE_OVERWRITE=OFF ..` give the same behavior

MarcelKoch: Which cmake version are you using, I'm using 3.17.0 without complaints? Also does changing `INTERNAL` to `BOOL` change the behavior?

yhmtsai: cmake 3.13.

I rerun it again. it does not work as expected due to my typo :man_facepalming:
and use -DGCF_FORCE_OVERWRITE=ON/OFF will overwrite the internal one, and it will change the setting in git-cmake-format.
Is empty for OFF or ON in cmake?

MarcelKoch: Empty would correspond to OFF, if you test it in an if condition, i.e. in `if(<var>)` it would evaluate to false if `<var>` is empty or not defined.

yhmtsai: but it will depends on how you detect the VARIABLE.
anyway, it is the cmake usage and git-cmake-format claims it should be ON/OFF

yhmtsai (approved):
LGTM. sorry for my wrong comments

INSTALL.md
```diff
@@ -17,7 +17,9 @@ Ginkgo adds the following additional switches to control what is being built:
 
 *   `-DGINKGO_DEVEL_TOOLS={ON, OFF}` sets up the build system for development
     (requires clang-format, will also download git-cmake-format),
-    default is `OFF`.
+    default is `OFF`. If it is set to `ON`, a pre-commit hook for formatting
+    will be installed. This hook may overwrite a user defined pre-commit hook
+    when Ginkgo is used as a submodule.
```
upsj, line 22: Nit: We should also mention the dummy hook

upsj (approved):

MarcelKoch:
rebase!

PR #831, opened 2021-07-09
Add reduction for arrays.

pratikvn:
This PR adds some reduction kernels for arrays. For CUDA and HIP the existing reduction kernels are reduced, with only one kernel added.

### TODO
+ [ ] Add a return overload.


yhmtsai:
I want to ask whether we make Array class related to the device (not header only) or not? (before release 1.4)
for now, we have fill and this pr reduce.
- have fill and reduce in class:
good for usage, 
but need to link all ginkgo always and still use free function to use fill/reduce in the device side
- only have fill and reduce free function
Array is header-only class
but always use free function to call fill/reduce.

upsj:
Generally, I like to think of gko::Array as a the only *vocabulary type* that Ginkgo exposes. Something that can be used universally for interoperability with other libraries and user data. Now we have a few additions, namely value-converting `Array::operator=` and `Array::fill` that rely on actual device kernel and only work for specific `ValueType`s. The more I think about it, the more I think it might be better to keep these out (replacing them by explicit conversion kernels for arrays only used internally, and maybe a datatype-agnostic `Array::fill` that acts like repeated `memcpy`)
All of this becomes especially important in the context of #715, where `Array` is the fundamental type that gets passed around between all shared libraries.
My point here is: If people want to compute a reduction over an array, I believe they should use more high-level functionality, like Dense or (maybe later) the simple kernel interface. Otherwise we risk overloading the `Array` type.

pratikvn:
I agree with making not overloading Array, but I think that can severely restrict functionality. For example, the only way to store integer types right now is with arrays. And there are many situations where you need reductions for these integer arrays.

upsj:
In that case, wouldn't a free function `ValueType reduce(const Array<ValueType>&)` be more suitable? What use case do you have in mind for index calculations from the user side?

pratikvn:
Yes, I guess we could add a free function instead of member functions. I am not sure where to add them though. I think there are multiple use cases for reduction over integer arrays. To calculate the total number of non-zeros in a matrix, if you have the number of nonzeros per row to give a very specific example :) 

But in general, I believe this is something useful not only from a user's perspective but also for us to write our algorithms more easily in our core.

upsj:
In core algorithms, we usually need prefix sums, not plain reductions. And these computations are mostly necessary in the context of a matrix conversion, which we represent with a high-level interface already. If we want to use this operation inside `core`, we can use existing kernels.

I think we need to find a balance of how much we want to enable to write their own complex algorithms, and how generic we want to make that. Sum reductions are for example not the only important operation, think of (arg)min/max for building ELL.

hip/components/reduce_array.hip.cpp
```diff
@@ -0,0 +1,98 @@
...
+#include "hip/components/cooperative_groups.hip.hpp"
+#include "hip/components/thread_ids.hip.hpp"
+#include "hip/components/uninitialized_array.hip.hpp"
+
+
+namespace gko {
+namespace kernels {
+namespace hip {
+namespace components {
+
```
yhmtsai, line 57:
```suggestion


```

hip/test/components/reduce_array.hip.cpp
```diff
@@ -0,0 +1,96 @@
...
+          exec(gko::HipExecutor::create(0, ref)),
+          total_size(6344),
+          vals(ref, total_size),
+          dvals(exec, total_size)
+    {
+        std::fill_n(vals.get_data(), total_size, 3);
+        dvals = vals;
+        out = T(2);
+        gko::kernels::reference::components::reduce_array(
+            ref, vals.get_const_data(), total_size, &out);
```
yhmtsai, line 69: we usually put it into TEST not init

hip/test/components/reduce_array.hip.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+TYPED_TEST_SUITE(ReduceArray, gko::test::ValueAndIndexTypes);
+
+
+TYPED_TEST(ReduceArray, EqualsReference)
+{
+    using T = typename TestFixture::value_type;
+    auto dval = gko::Array<T>(this->exec, I<T>{2});
+    auto val = gko::Array<T>(this->ref, 1);
+    gko::kernels::hip::components::reduce_array(
```
yhmtsai, line 88:
```suggestion
    auto val = gko::Array<T>(this->ref, 1);

    gko::kernels::hip::components::reduce_array(
```

hip/test/components/reduce_array.hip.cpp
```diff
@@ -0,0 +1,96 @@
...
+        this->exec, this->dvals.get_data(), this->total_size, dval.get_data());
+    val = dval;
+
+    ASSERT_EQ(T(this->out), T(val.get_data()[0]));
```
yhmtsai, line 92: it does not need `T()`
GKO_ASSERT_ARRAY_EQ also works for device memory

omp/components/reduce_array.cpp
```diff
@@ -0,0 +1,62 @@
...
+namespace omp {
+namespace components {
+
+
+template <typename ValueType>
+void reduce_array(std::shared_ptr<const DefaultExecutor> exec,
+                  const ValueType* array, size_type n, ValueType* val)
+{
+    ValueType out = *val;
+#pragma omp declare reduction(add:ValueType : omp_out = omp_out + omp_in)
```
yhmtsai, line 47: does the usual addition need this?

reference/components/reduce_array.cpp
```diff
@@ -0,0 +1,58 @@
...
+namespace kernels {
+namespace reference {
+namespace components {
+
+
+template <typename ValueType>
+void reduce_array(std::shared_ptr<const DefaultExecutor> exec,
+                  const ValueType* array, size_type n, ValueType* val)
+{
+    *val = std::accumulate(array, array + n, ValueType(*val));
```
yhmtsai, line 49:
```suggestion
    *val = std::accumulate(array, array + n, *val);
```

dpcpp/components/reduce_array.dp.cpp
```diff
@@ -0,0 +1,56 @@
...
+namespace gko {
+namespace kernels {
+namespace dpcpp {
+namespace components {
+
+
+template <typename ValueType>
+void reduce_array(std::shared_ptr<const DefaultExecutor> exec,
+                  const ValueType* array, size_type n,
+                  ValueType* val) GKO_NOT_IMPLEMENTED;
```
yhmtsai, line 48: need implement?

core/base/array.cpp
```diff
@@ -92,9 +94,42 @@ void Array<ValueType>::fill(const ValueType value)
 }
 
 
+template <typename ValueType>
+void reduce(const Array<ValueType>& input_arr, ValueType* value)
+{
+    auto exec = input_arr.get_executor();
+    exec->run(array::make_reduce_array(input_arr.get_const_data(),
+                                       input_arr.get_num_elems(), value));
```
yhmtsai, line 102: value needs to be the device pointer, right?
if only the pointer in the public interface, we do not know whether it is on the device or not.

core/base/array.cpp
```diff
@@ -92,9 +94,42 @@ void Array<ValueType>::fill(const ValueType value)
...
+
+template <typename ValueType>
+ValueType reduce(const Array<ValueType>& input_arr, const ValueType init_value)
+{
+    auto exec = input_arr.get_executor();
+    auto value = Array<ValueType>(exec, {init_value});
+    exec->run(array::make_reduce_array(input_arr.get_const_data(),
+                                       input_arr.get_num_elems(),
+                                       value.get_data()));
+    return exec->copy_val_to_host(value.get_data());
```
yhmtsai, line 114: It can be done with normal reduction and report init_value + result?
the normal reduction do not have additional init in the beginning

yhmtsai (changes requested):
need another name for reduce2 and how many changes will be related to #833

hip/test/components/reduce_array.hip.cpp
yhmtsai: also apply to other test

omp/components/reduce_array.cpp
pratikvn: I think we do because of the templating and omp is not able to detect the types to use the in-built overloaded reduction operators.

core/base/array.cpp
pratikvn: Yes, the idea was that if you needed the value on the host, then you would use the reduce function which returns a value. Yes, I think this would throw a seg-fault, but I did not want to wrap this in an Array again.

core/base/array.cpp
pratikvn: Yes, the idea of having a host return was why this function was added.

include/ginkgo/core/base/array.hpp
```diff
@@ -570,6 +571,31 @@ class Array {
...
+/**
+ * Reduce the values in the array
+ *
+ * @tparam The type of the input data
+ *
+ * @param [in] input_arr the input array to be reduced
+ * @param [out,in] value the reduced value
+ */
+template <typename ValueType>
+void reduce(const Array<ValueType>& input_arr, ValueType* value);
```
thoasm, line 596: - Missing description that the initial value of `value` is considered (and that you only consider the first value).
- Why do you have the Input as an `Array`, while the output is only a pointer?

cuda/test/components/reduce_array.cpp
```diff
@@ -0,0 +1,96 @@
...
+TYPED_TEST(ReduceArray, EqualsReference)
+{
+    using T = typename TestFixture::value_type;
+    auto dval = gko::Array<T>(this->exec, I<T>{2});
+    auto val = gko::Array<T>(this->ref, 1);
+    gko::kernels::cuda::components::reduce_array(
+        this->exec, this->dvals.get_data(), this->total_size, dval.get_data());
+    val = dval;
+
+    ASSERT_EQ(T(this->out), T(val.get_data()[0]));
```
thoasm, line 92: nit: spacing
```suggestion
    auto val = gko::Array<T>(this->ref, 1);

    gko::kernels::cuda::components::reduce_array(
        this->exec, this->dvals.get_data(), this->total_size, dval.get_data());

    val = dval;
    ASSERT_EQ(T(this->out), T(val.get_data()[0]));
```

cuda/test/components/reduce_array.cpp
```diff
@@ -0,0 +1,96 @@
...
+          exec(gko::CudaExecutor::create(0, ref)),
+          total_size(6344),
+          vals(ref, total_size),
+          dvals(exec, total_size)
+    {
+        std::fill_n(vals.get_data(), total_size, 3);
+        dvals = vals;
+        out = T(2);
+        gko::kernels::reference::components::reduce_array(
+            ref, vals.get_const_data(), total_size, &out);
```
thoasm, line 69: This should be part of the test itself (right next to the cuda kernel call), so it is clear what you compare.

hip/test/components/reduce_array.hip.cpp
thoasm: Yes, I agree.

reference/test/base/array.cpp
```diff
@@ -76,4 +76,21 @@ TYPED_TEST(Array, CanBeFilledWithValue)
 }
 
 
+TYPED_TEST(Array, CanBeReduced)
+{
+    TypeParam out = 1;
+    gko::reduce(this->x, &out);
```
thoasm, line 82:
```suggestion
    TypeParam out = 1;

    gko::reduce(this->x, &out);
```

omp/test/components/reduce_array.cpp
```diff
@@ -0,0 +1,92 @@
...
+        : ref(gko::ReferenceExecutor::create()),
+          exec(gko::OmpExecutor::create()),
+          total_size(6355),
+          vals(ref, total_size),
+          dvals(exec, total_size)
+    {
+        std::fill_n(vals.get_data(), total_size, 3);
+        dvals = vals;
+        out = std::accumulate(vals.get_data(), vals.get_data() + total_size,
+                              T(2));
```
thoasm, line 67: Please use the reference kernel for the comparison, and move it into the test itself.

omp/test/components/reduce_array.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+TYPED_TEST_SUITE(ReduceArray, gko::test::ValueAndIndexTypes);
+
+
+TYPED_TEST(ReduceArray, EqualsReference)
+{
+    using T = typename TestFixture::value_type;
+    auto val = T(2);
+    gko::kernels::omp::components::reduce_array(
+        this->exec, this->dvals.get_data(), this->total_size, &val);
```
thoasm, line 86: Also needs the reference kernel as part of it.
```suggestion
    auto val = T(2);

    gko::kernels::omp::components::reduce_array(
        this->exec, this->dvals.get_data(), this->total_size, &val);
```

thoasm (commented):
Mostly minor comments, but some of them I would like to have resolved before approving.

hip/components/reduce_array.hip.cpp
```diff
@@ -0,0 +1,99 @@
...
+#include <ginkgo/core/base/array.hpp>
+#include <ginkgo/core/base/executor.hpp>
+#include <ginkgo/core/base/math.hpp>
+
+
+#include "hip/base/types.hip.hpp"
+#include "hip/components/cooperative_groups.hip.hpp"
+#include "hip/components/thread_ids.hip.hpp"
+#include "hip/components/uninitialized_array.hip.hpp"
```
tcojean, line 50: I think most of these are not needed? Same in other files

cuda/test/base/array.cu
```diff
@@ -92,3 +92,29 @@ TYPED_TEST(Array, CanCopyBackTemporaryCloneOnDifferentExecutor)
...
+
+TYPED_TEST(Array, CanBeReduced2)
+{
+    using T = TypeParam;
+    auto cuda = gko::CudaExecutor::create(0, this->exec);
+    auto arr = gko::Array<TypeParam>(cuda, I<T>{T(4), T(6)});
+
+    auto out = gko::reduce(arr, TypeParam{T(3)});
+
+    ASSERT_EQ(out, TypeParam{13});
```
tcojean, line 119: Like in other places

```suggestion
    auto out = gko::reduce(arr, T{3});

    ASSERT_EQ(out, T{13});
```

cuda/test/base/array.cu
```diff
@@ -92,3 +92,29 @@ TYPED_TEST(Array, CanCopyBackTemporaryCloneOnDifferentExecutor)
...
+    out.set_executor(cuda->get_master());
+    ASSERT_EQ(out.get_data()[0], TypeParam{12});
+}
+
+
+TYPED_TEST(Array, CanBeReduced2)
+{
+    using T = TypeParam;
+    auto cuda = gko::CudaExecutor::create(0, this->exec);
+    auto arr = gko::Array<TypeParam>(cuda, I<T>{T(4), T(6)});
```
tcojean, line 115: Is `T()` really required when you have `I<T>{}`? Same above.

tcojean (approved):
LGTM, some comments.

Another question, should this be only called `reduce` since it specifically targets a sum? What about min/max etc if we would add that as well?

hip/components/reduce_array.hip.cpp
pratikvn: I think we do because the kernels in `reduction.hpp.inc` use them.

tcojean: Ah right I missed that there are a bunch of existing codes there already.

pratikvn:
@tcojean, The question regarding the name `reduce` is a good point. I will try to add a third lambda parameter to allow for the reduction operation to be set.

dpcpp/components/reduce_array.dp.cpp
pratikvn: Yes, I need to. But I haven't gotten used to dpcpp yet, so it would take more some more time. Maybe we can a future TODO ? Or would you rather have this now ?

hip/test/components/reduce_array.hip.cpp
```diff
@@ -0,0 +1,97 @@
...
+    gko::Array<value_type> dvals;
+};
+
+TYPED_TEST_SUITE(ReduceArray, gko::test::ValueAndIndexTypes);
+
+
+TYPED_TEST(ReduceArray, EqualsReference)
+{
+    using T = typename TestFixture::value_type;
+    auto dval = gko::Array<T>(this->exec, I<T>{2});
```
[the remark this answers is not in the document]
pratikvn: No, we need to have an initializer list. We don't have an Array constructor for just one value.

pratikvn:
Okay, then I will update that to an Array as well. Thanks for the suggestions.

PR #832, opened 2021-07-12
Fix dpcpp memory issue and capture exception in raw_free

yhmtsai:
This PR limits the direct memcpy between different backend or device, add free after kernel test and capture exception in raw_free

the memcpy issue between backend on the same device is on https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/1413646244
the `raw_free` isssue is on https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/1415200944

the `codeplay_host_task` does not work as expectation yet.
need to investigate furthermore and use the new api.
Thus, we stay wait workaround to ensure free is after kernel

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -722,6 +722,7 @@ build/dpcpp/cpu/release/static:
     BUILD_SHARED_LIBS: "ON"
     SYCL_DEVICE_TYPE: "CPU"
 
+# It gives two available backends of GPU on tests
 build/dpcpp/igpu/release/shared:
   <<: *default_build_with_test
   extends:
@@ -768,6 +769,7 @@ build/dpcpp/level_zero_igpu/debug/static:
     DPCPP_SINGLE_MODE: "ON"
     SYCL_DEVICE_FILTER: "Level_Zero:GPU"
 
+# It gives two available backends of GPU on tests
 build/dpcpp/dgpu/release/static:
   <<: *default_build_with_test
   extends:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -100,6 +100,7 @@ include:
       fi
     - if [ -n "${SYCL_DEVICE_TYPE}" ]; then export SYCL_DEVICE_TYPE; fi
     - if [ -n "${SYCL_DEVICE_FILTER}" ]; then export SYCL_DEVICE_FILTER; fi
+    - if [ -n "${SYCL_DEVICE_TYPE}" ]; then export SYCL_DEVICE_TYPE; fi
     - cmake ${CI_PROJECT_DIR}${CI_PROJECT_DIR_SUFFIX}
         -GNinja
         -DCMAKE_C_COMPILER=${C_COMPILER} -DCMAKE_CXX_COMPILER=${CXX_COMPILER}
@@ -148,6 +149,7 @@ include:
       fi
     - if [ -n "${SYCL_DEVICE_TYPE}" ]; then unset SYCL_DEVICE_TYPE; fi
     - if [ -n "${SYCL_DEVICE_FILTER}" ]; then unset SYCL_DEVICE_FILTER; fi
+    - if [ -n "${SYCL_DEVICE_TYPE}" ]; then unset SYCL_DEVICE_TYPE; fi
     - if [ "${EXPORT_BUILD_DIR}" == "ON" ]; then ninja test_exportbuild; fi
   dependencies: []
   except:
@@ -750,7 +752,7 @@ build/dpcpp/dgpu/debug/shared:
     BUILD_DPCPP: "ON"
     BUILD_TYPE: "Debug"
     DPCPP_SINGLE_MODE: "ON"
-    SYCL_DEVICE_FILTER: "Level_Zero:GPU"
+    SYCL_DEVICE_TYPE: "GPU"
 
 
 # Job with important warnings as error
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -100,7 +100,6 @@ include:
       fi
     - if [ -n "${SYCL_DEVICE_TYPE}" ]; then export SYCL_DEVICE_TYPE; fi
     - if [ -n "${SYCL_DEVICE_FILTER}" ]; then export SYCL_DEVICE_FILTER; fi
-    - if [ -n "${SYCL_DEVICE_TYPE}" ]; then export SYCL_DEVICE_TYPE; fi
     - cmake ${CI_PROJECT_DIR}${CI_PROJECT_DIR_SUFFIX}
         -GNinja
         -DCMAKE_C_COMPILER=${C_COMPILER} -DCMAKE_CXX_COMPILER=${CXX_COMPILER}
@@ -149,7 +148,6 @@ include:
       fi
     - if [ -n "${SYCL_DEVICE_TYPE}" ]; then unset SYCL_DEVICE_TYPE; fi
     - if [ -n "${SYCL_DEVICE_FILTER}" ]; then unset SYCL_DEVICE_FILTER; fi
-    - if [ -n "${SYCL_DEVICE_TYPE}" ]; then unset SYCL_DEVICE_TYPE; fi
     - if [ "${EXPORT_BUILD_DIR}" == "ON" ]; then ninja test_exportbuild; fi
   dependencies: []
   except:
@@ -724,7 +722,7 @@ build/dpcpp/cpu/release/static:
     BUILD_SHARED_LIBS: "ON"
     SYCL_DEVICE_TYPE: "CPU"
 
-build/dpcpp/igpu/release/static:
+build/dpcpp/igpu/release/shared:
   <<: *default_build_with_test
   extends:
     - .full_test_condition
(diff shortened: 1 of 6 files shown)


dpcpp/base/executor.dp.cpp
```diff
@@ -102,8 +102,29 @@ void DpcppExecutor::populate_exec_info(const MachineTopology *mach_topo)
...
+                  << "Exiting program" << std::endl;
+#endif  // GKO_VERBOSE_LEVEL >= 1
+        // OpenCL error code use 0 for CL_SUCCESS and negative number for others
+        // error. if the error is not from OpenCL, it will return CL_SUCCESS.
+        int err_code = err.get_cl_code();
+        // if return CL_SUCCESS, exit 1 as DPCPP error.
+        if (err_code == 0) {
+            err_code = 1;
+        }
+        std::exit(err_code);
```
upsj, line 126: We need to think about other ways to approach this at some point, since killing the program may not be desirable for application users.

dpcpp/test/base/executor.dp.cpp
```diff
@@ -269,4 +269,31 @@ TEST_F(DpcppExecutor, Synchronizes)
...
+            gko::Array<float> y(dpcpp, length);
+            auto x_val = x.get_data();
+            auto y_val = y.get_data();
+            dpcpp->get_queue()->submit([&](sycl::handler &cgh) {
+                cgh.parallel_for(sycl::range<1>{length},
+                                 [=](sycl::id<1> i) { y_val[i] += x_val[i]; });
+            });
+        }
+        // to ensure everything on queue is finished.
+        dpcpp->synchronize();
```
upsj, line 294: Is this necessary? The case we want to catch is the one without synchronization, isn't it?

upsj (approved):
LGTM!

dpcpp/test/base/executor.dp.cpp
yhmtsai: Not sure honestly, I do not know whether queue wait the all kernel finished when queue destroy.
If it is like cuda, just terminate. It is possible that we only have allocation and free but no execution.
I use curly brackets to contain the Array allocation to make sure the free is called before the synchronize.
Yes, it catches the free is not after the kernel finishes

upsj: When the array gets freed, we wait on the queue, that isn't sufficient?

dpcpp/base/executor.dp.cpp
yhmtsai: Probably, I am not sure whether the application allows the error situation is happened or not.

dpcpp/test/base/executor.dp.cpp
yhmtsai: Yes, it is fixed version. when the free does not contain synchronization, we will probably not wait the kernel.

yhmtsai:
After more test, the combination of device gives different result:
1. `opencl:gpu` vs `opencl:cpu` -> use `opencl:gpu` queue can get the `opencl:cpu` memory but reverse case can not. (`opencl:cpu` is like normal host memory)
2. `opencl:gpu` vs `opencl:gpu` -> no matter whether two queue is mapped to the same device, they can not transfer data to each other
3. `level_zero:gpu` vs `opencl:cpu` -> similar to 1.

host option does not work in filter_selector ([ref](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/FilterSelector/FilterSelector.adoc)), so I need to check it.

I only find one way to get the backend, which using dynamic_cast (with backend type) to type provided by level_zero.
Or, we always use host memory to transfer data when the queues are different (for handling opencl cases)

I put the code here. we can use it check. 
Compile: `dpcpp <file>`
Run: `./a.out <first_device_filter> <second_device_filter>`
```
#include <CL/sycl.hpp>
#include <iostream>

{
    if (argc != 3) {
        std::cout << "Usage: " << argv[0] << " first_filter second_filter"
                  << std::endl;
        std::exit(1);
    }
    auto first_filter = sycl::ONEAPI::filter_selector(argv[1]);
    auto second_filter = sycl::ONEAPI::filter_selector(argv[2]);
    auto *first_queue =
        new sycl::queue{first_filter, sycl::property::queue::in_order{}};
    auto *second_queue =
        new sycl::queue{second_filter, sycl::property::queue::in_order{}};
    std::cout << "First queue device name: "
              << first_queue->get_device().get_info<sycl::info::device::name>()
              << std::endl;
    std::cout << "Second queue device name: "
              << second_queue->get_device().get_info<sycl::info::device::name>()
              << std::endl;
    const int n = 32;

    float data[n];
    for (int i = 0; i < n; i++) {
        data[i] = static_cast<float>(i) / n;
    }
    float *first_data = sycl::malloc_device<float>(n, *first_queue);
    float *first_data2 = sycl::malloc_device<float>(n, *first_queue);
    float *second_data = sycl::malloc_device<float>(n, *second_queue);
    float *second_data2 = sycl::malloc_device<float>(n, *second_queue);
    std::cout << "First queue test:" << std::endl;
    std::cout << "From Host to Device" << std::endl;
    first_queue->memcpy(first_data, data, n * sizeof(float)).wait();
    std::cout << "From Device to Device (self-queue)" << std::endl;
    first_queue->memcpy(first_data2, first_data, n * sizeof(float)).wait();
    std::cout << "From Device to Host" << std::endl;
    first_queue->memcpy(data, first_data2, n * sizeof(float)).wait();
    first_queue->wait_and_throw();
    std::cout << "Second queue test:" << std::endl;
    std::cout << "From Host to Device" << std::endl;
    second_queue->memcpy(second_data, data, n * sizeof(float)).wait();
    std::cout << "From Device to Device (self-queue)" << std::endl;
    second_queue->memcpy(second_data2, second_data, n * sizeof(float)).wait();
    std::cout << "From Device to Host" << std::endl;
    second_queue->memcpy(data, second_data2, n * sizeof(float)).wait();
    second_queue->wait_and_throw();
    std::cout << "Communication between first/second queue:" << std::endl;
    std::cout << "device check: "
              << (first_queue->get_device() == second_queue->get_device())
              << std::endl;
    std::cout << "From second to first (handled by first queue)" << std::endl;
    first_queue->memcpy(first_data, second_data, n * sizeof(float)).wait();
    std::cout << "From fisrt to second (handled by second queue)" << std::endl;
    second_queue->memcpy(second_data, first_data, n * sizeof(float)).wait();
    first_queue->wait_and_throw();
    second_queue->wait_and_throw();
    sycl::free(first_data, first_queue->get_context());
    sycl::free(first_data2, first_queue->get_context());
    sycl::free(second_data, second_queue->get_context());
    sycl::free(second_data2, second_queue->get_context());
    return 0;
}
```

upsj:
A round-trip via host memory sounds sensible, I mean we do the same thing for ROCm <-> CUDA copies.

yhmtsai:
I forgot to update level_zero information.
level_zero backend allows to transfer data on the same device, but opencl cannot.
get_device() are the same but get_context() are different

.gitlab-ci.yml
```diff
@@ -748,10 +793,24 @@ build/dpcpp/dgpu/debug/shared:
     C_COMPILER: "gcc"
     CXX_COMPILER: "dpcpp"
     BUILD_DPCPP: "ON"
-    BUILD_TYPE: "Debug"
+    BUILD_TYPE: "Release"
     DPCPP_SINGLE_MODE: "ON"
     SYCL_DEVICE_FILTER: "Level_Zero:GPU"
```
tcojean, line 798: Does this work, or do you need to use the device_type as well?

.gitlab-ci.yml
```diff
@@ -733,13 +733,58 @@ build/dpcpp/igpu/release/static:
...
+    - .use_gko-oneapi-igpu
+  variables:
+    <<: *default_variables
+    C_COMPILER: "gcc"
+    CXX_COMPILER: "dpcpp"
+    BUILD_DPCPP: "ON"
+    BUILD_TYPE: "Debug"
     BUILD_SHARED_LIBS: "OFF"
     DPCPP_SINGLE_MODE: "ON"
     SYCL_DEVICE_FILTER: "Level_Zero:GPU"
```
tcojean, line 769: Same here

.gitlab-ci.yml
```diff
@@ -733,13 +733,58 @@ build/dpcpp/igpu/release/static:
...
 
-build/dpcpp/dgpu/debug/shared:
+build/dpcpp/dgpu/release/static:
```
tcojean, line 771: The point of this job is to pick either OpenCL or Level Zero? Why have this when there are also specific OpenCL and Level Zero jobs?

dpcpp/base/executor.dp.cpp
```diff
@@ -144,7 +165,21 @@ void DpcppExecutor::raw_copy_to(const DpcppExecutor *dest, size_type num_bytes,
...
+        auto dest_queue = dest->get_queue();
+        auto device = queue->get_device();
+        auto dest_device = dest_queue->get_device();
+        if (((device.is_host() || device.is_cpu()) &&
+             (dest_device.is_host() || dest_device.is_cpu())) ||
+            (queue == dest_queue)) {
+            dest->get_queue()->memcpy(dest_ptr, src_ptr, num_bytes).wait();
+        } else {
+            // the memcpy only support host<->device or itself memcpy
+            GKO_NOT_SUPPORTED(dest);
```
tcojean, line 181: I think we should do a copy through the host here like in the ROCM <-> NVIDIA cases. We should also add a debug level message to state that this might not be optimized. I expect that later direct GPU copies will be supported?

tcojean (approved):
LGTM. Some small comments.

.gitlab-ci.yml
yhmtsai: yes for the result.
but only Level_Zero has effect.
the GPU is the only one which has level_zero backend, so it gives the same result

.gitlab-ci.yml
yhmtsai: It depends on the intel.
this is to make opencl and level_zero available on the executor test for checking the copy between opencl and level_zero.

dpcpp/base/executor.dp.cpp
yhmtsai: should we add the message?
for the GPU direct copies, it seems to use context to take care (identify) the memory.
If I pass the context to another one, they can communicate without segmentation fault.
but I can also pass the opencl context to level_zero queue and then they communicate without segfault, which is weird.
context should be under the platform but opencl and level zero should be the different platforms

yhmtsai: I use `#if (GKO_VERBOSE_LEVEL >= 1) && !defined(NDEBUG)` to enable output clog

.gitlab-ci.yml
tcojean: I see that makes sense.

dpcpp/base/executor.dp.cpp
tcojean: To make it work as you say, we would need a global context which everything uses? Or can you have one context per device and use the correct context so that the copy works?

yhmtsai: I think it should be a context per platform.
but I need to check the level_zero/opencl context behavior
two queues with the same context and then they can communicate

yhmtsai:
rebase!

dpcpp/test/base/executor.dp.cpp
upsj: Can you remove this line to make sure the program can actually fail due to use-after-free issues?

yhmtsai: it's shown in the https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/1415200944
and the code contains sync https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/commit/f9da28bc5b8f6847b4d587b0283b2cc7f7e2f60a
I mean fixed version is the we wait on the queue when we call free.
(the following description is from cuda thinking)
Without sync before program end
1. when free does not contain wait:
submit job (not run) -> free -> program end (the job may not run, so we may does not face the free after use)
2. when free does contain wait: (fixed version)
submit job (not run) -> run job (due to free contain wait) -> free -> program end
this case does not need sync in the end

upsj: I see, thanks! I completely missed that the arrays have their own scope, so this is fine to me

PR #834, opened 2021-07-14
Fix scalar jacobi generate on different precision and add DiagonalLinOpExtractable

yhmtsai:
This pr adds DiagonalLinOpExtractable and fix scalar Jacobi issue on generation on different precision matrix.

the failed test can be found in https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/pipelines/336902618

### DIFF
--- a/core/matrix/diagonal.cpp
+++ b/core/matrix/diagonal.cpp
@@ -302,4 +302,24 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE(GKO_DECLARE_DIAGONAL_MATRIX);
 
 
 }  // namespace matrix
+
+
+// Implement DiagonalExtractable for LinOp when Diagonal is complete class
+template <typename ValueType>
+std::unique_ptr<LinOp> DiagonalExtractable<ValueType>::extract_diagonal_linop()
+    const
+{
+    auto diag = this->extract_diagonal();
+    auto p = dynamic_cast<LinOp *>(diag.get());
+    diag.release();
+    return std::unique_ptr<LinOp>{p};
+}
+
+
+#define GKO_DECLARE_DIAGONAL_EXTRACTABLE(value_type) \
+    std::unique_ptr<LinOp>                           \
+    DiagonalExtractable<value_type>::extract_diagonal_linop() const
+GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE(GKO_DECLARE_DIAGONAL_EXTRACTABLE);
+
+
 }  // namespace gko
--- a/core/matrix/diagonal.cpp
+++ b/core/matrix/diagonal.cpp
@@ -147,6 +147,22 @@ std::unique_ptr<LinOp> Diagonal<ValueType>::conj_transpose() const
 
 
 template <typename ValueType>
+void Diagonal<ValueType>::convert_to(
+    Diagonal<next_precision<ValueType>> *result) const
+{
+    result->values_ = this->values_;
+    result->set_size(this->get_size());
+}
+
+
+template <typename ValueType>
+void Diagonal<ValueType>::move_to(Diagonal<next_precision<ValueType>> *result)
+{
+    this->convert_to(result);
+}
+
+
+template <typename ValueType>
 void Diagonal<ValueType>::convert_to(Csr<ValueType, int32> *result) const
 {
     auto exec = this->get_executor();
--- a/core/matrix/diagonal.cpp
+++ b/core/matrix/diagonal.cpp
@@ -325,10 +325,7 @@ template <typename ValueType>
 std::unique_ptr<LinOp> DiagonalExtractable<ValueType>::extract_diagonal_linop()
     const
 {
-    auto diag = this->extract_diagonal();
-    auto p = dynamic_cast<LinOp *>(diag.get());
-    diag.release();
-    return std::unique_ptr<LinOp>{p};
+    return this->extract_diagonal();
 }
 
 
--- a/core/preconditioner/jacobi.cpp
+++ b/core/preconditioner/jacobi.cpp
@@ -231,11 +231,23 @@ void Jacobi<ValueType, IndexType>::generate(const LinOp *system_matrix,
     using csr_type = matrix::Csr<ValueType, IndexType>;
     const auto exec = this->get_executor();
     if (parameters_.max_block_size == 1) {
-        auto diag = as<DiagonalExtractable<ValueType>>(system_matrix)
-                        ->extract_diagonal();
-        auto temp = gko::Array<ValueType>::view(
-            exec, system_matrix->get_size()[0], diag->get_values());
-        this->blocks_ = temp;
+        auto diag = share(as<DiagonalLinOpExtractable>(system_matrix)
+                              ->extract_diagonal_linop());
+        using next_type = next_precision<ValueType>;
+        if (auto diag_dense =
+                std::dynamic_pointer_cast<matrix::Diagonal<ValueType>>(diag)) {
+            auto temp = gko::Array<ValueType>::view(
+                exec, system_matrix->get_size()[0], diag_dense->get_values());
+            this->blocks_ = temp;
+        } else if (auto diag_dense =
+                       std::dynamic_pointer_cast<matrix::Diagonal<next_type>>(
+                           diag)) {
+            auto temp = gko::Array<next_type>::view(
+                exec, system_matrix->get_size()[0], diag_dense->get_values());
+            this->blocks_ = temp;
+        } else {
+            GKO_NOT_SUPPORTED(system_matrix);
+        }
     } else {
         auto csr_mtx = convert_to_with_sorting<csr_type>(exec, system_matrix,
                                                          skip_sorting);
--- a/core/preconditioner/jacobi.cpp
+++ b/core/preconditioner/jacobi.cpp
@@ -237,13 +237,15 @@ void Jacobi<ValueType, IndexType>::generate(const LinOp *system_matrix,
         if (auto diag_dense =
                 std::dynamic_pointer_cast<matrix::Diagonal<ValueType>>(diag)) {
             auto temp = gko::Array<ValueType>::view(
-                exec, system_matrix->get_size()[0], diag_dense->get_values());
+                diag_dense->get_executor(), system_matrix->get_size()[0],
+                diag_dense->get_values());
             this->blocks_ = temp;
         } else if (auto diag_dense =
                        std::dynamic_pointer_cast<matrix::Diagonal<next_type>>(
                            diag)) {
             auto temp = gko::Array<next_type>::view(
-                exec, system_matrix->get_size()[0], diag_dense->get_values());
+                diag_dense->get_executor(), system_matrix->get_size()[0],
+                diag_dense->get_values());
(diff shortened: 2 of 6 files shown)


yhmtsai:
rebase!

pratikvn:
I think the [`omp/debug/shared` job was failing](https://github.com/ginkgo-project/ginkgo/runs/3067008668?check_suite_focus=true) ?

yhmtsai:
yes

CMakeLists.txt
```diff
@@ -278,7 +278,10 @@ if(GINKGO_DEVEL_TOOLS)
     add_custom_target(add_license
         COMMAND ${Ginkgo_SOURCE_DIR}/dev_tools/scripts/add_license.sh
         WORKING_DIRECTORY ${Ginkgo_SOURCE_DIR})
-    add_dependencies(format add_license)
+    # if git-cmake-format can not build format target, do not add the dependencies
+    if(TARGET format)
```
pratikvn, line 282: I guess this is from another PR ?

core/matrix/diagonal.cpp
```diff
@@ -302,4 +302,24 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE(GKO_DECLARE_DIAGONAL_MATRIX);
...
 
 }  // namespace matrix
+
+
+// Implement DiagonalExtractable for LinOp when Diagonal is complete class
+template <typename ValueType>
+std::unique_ptr<LinOp> DiagonalExtractable<ValueType>::extract_diagonal_linop()
+    const
+{
+    auto diag = this->extract_diagonal();
```
pratikvn, line 312: Am I missing something here ? `Diagonal` does not have a `extract_diagonal` method, right ?

core/preconditioner/jacobi.cpp
```diff
@@ -231,11 +231,23 @@ void Jacobi<ValueType, IndexType>::generate(const LinOp *system_matrix,
     using csr_type = matrix::Csr<ValueType, IndexType>;
     const auto exec = this->get_executor();
     if (parameters_.max_block_size == 1) {
-        auto diag = as<DiagonalExtractable<ValueType>>(system_matrix)
-                        ->extract_diagonal();
-        auto temp = gko::Array<ValueType>::view(
-            exec, system_matrix->get_size()[0], diag->get_values());
-        this->blocks_ = temp;
+        auto diag = share(as<DiagonalLinOpExtractable>(system_matrix)
+                              ->extract_diagonal_linop());
```
pratikvn, line 235: Is there some reason we need the `extract_diagonal_linop` and `extract_diagonal` will not work ?

pratikvn (commented):
Some questions.

core/preconditioner/jacobi.cpp
```diff
@@ -231,11 +231,23 @@ void Jacobi<ValueType, IndexType>::generate(const LinOp *system_matrix,
...
+        using next_type = next_precision<ValueType>;
+        if (auto diag_dense =
+                std::dynamic_pointer_cast<matrix::Diagonal<ValueType>>(diag)) {
+            auto temp = gko::Array<ValueType>::view(
+                exec, system_matrix->get_size()[0], diag_dense->get_values());
```
upsj, line 240: The source of the segfault we just saw:
```suggestion
                diag_dense->get_executor(), system_matrix->get_size()[0], diag_dense->get_values());
```

core/preconditioner/jacobi.cpp
```diff
@@ -231,11 +231,23 @@ void Jacobi<ValueType, IndexType>::generate(const LinOp *system_matrix,
...
+            this->blocks_ = temp;
+        } else if (auto diag_dense =
+                       std::dynamic_pointer_cast<matrix::Diagonal<next_type>>(
+                           diag)) {
+            auto temp = gko::Array<next_type>::view(
+                exec, system_matrix->get_size()[0], diag_dense->get_values());
```
upsj, line 246: same here

core/preconditioner/jacobi.cpp
yhmtsai: extract_diagonal require the valuetype, but extract_diagonal_linop does not. we can get the diag first and then check which precision to work.
I am thinking do the dynamic_cast to check the matrix type and use corresponding extract_diag.

core/matrix/diagonal.cpp
yhmtsai: no, but the DiagonalExtractable use `as` or `dynamic_cast` to convert the output from Diagonal to Linop.
then it requires Diagonal be a complete class.
I put the implementation into Diagonal and instantiate it.

CMakeLists.txt
yhmtsai: yes

core/preconditioner/jacobi.cpp
yhmtsai: good catch!

upsj: Could you add a test for this behavior (generating with block size 1 on CUDA from a Reference/OMP matrix)?

core/preconditioner/jacobi.cpp
upsj: An alternative to DiagonalLinOpExtractable would be using
```
auto csr_mtx = convert_to_with_sorting<csr_type>(exec, system_matrix, /*skip_sorting=*/true);
```
from the non-scalar branch and extracting from that. That would also make for more consistent behavior between scalar and block Jacobi.

core/preconditioner/jacobi.cpp
pratikvn: Good catch. I think we should think about having a better way to test these cross-executor combinations ? I think we hardly test any cross-executor functionality now.

core/preconditioner/jacobi.cpp
yhmtsai: I would like to avoid the converting matrix self because we need to pay more for it against converting diag

core/preconditioner/jacobi.cpp
yhmtsai: for easy test, we can use `get_test_executor()` in random or in order to avoid specify the executor in each place

core/preconditioner/jacobi.cpp
pratikvn: Could we also not convert the extracted diagonal after calling the `extract_diagonal`. Currently, `Diagonal` does not have mixed precision convert methods, but we could easily add those

upsj: Oh yes, I think I like that idea best of all suggestions yet :)

yhmtsai: but the issue is different, right?
you still need to dynamic_cast to the correct type first.
and it is the same impl in the current one if we use default behavior
if we use direct mapping, we need to add all matrix to support it (at least csr)

pratikvn: Actually, i am a bit confused as to how the current implementation works. `extract_diagonal_linop` should return a `Diagonal` of `ValueType` right ? Because it is just calling the `extract_diagonal` of `Csr<ValueType, IndexType>`. So, the else if branch in line 243, will never be taken ?

core/preconditioner/jacobi.cpp
pratikvn: Yeah, I think you can do a specific test for this and we can think discuss a more generic approach later.

core/preconditioner/jacobi.cpp
upsj: You could use something similar to `make_temporary_conversion`. The framework also supports other matrix types, you only need to wrap it like you see in that function. Precision conversions in Diagonal is definitely a good idea, I forgot to suggest those when we added the format.

core/matrix/diagonal.cpp
```diff
@@ -302,4 +318,24 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE(GKO_DECLARE_DIAGONAL_MATRIX);
...
+    auto p = dynamic_cast<LinOp *>(diag.get());
+    diag.release();
+    return std::unique_ptr<LinOp>{p};
```
upsj, line 331: Isn't it possible to return `diag` directly? std::unique_ptr has a up-casting constructor that should take care of that.

core/preconditioner/jacobi.cpp
```diff
@@ -231,11 +232,20 @@ void Jacobi<ValueType, IndexType>::generate(const LinOp *system_matrix,
...
+        auto diag = share(as<DiagonalLinOpExtractable>(system_matrix)
+                              ->extract_diagonal_linop());
+        auto diag_vt =
+            ::gko::detail::temporary_conversion<matrix::Diagonal<ValueType>>::
+                template create<matrix::Diagonal<next_precision<ValueType>>>(
+                    diag.get());
+        if (!diag_vt) {
+            GKO_NOT_SUPPORTED(system_matrix);
+        }
+        auto temp = gko::Array<ValueType>::view(diag_vt->get_executor(),
```
upsj, line 244: Is the namespace necessary here?
```suggestion
        auto temp = Array<ValueType>::view(diag_vt->get_executor(),
```

core/preconditioner/jacobi.cpp
```diff
@@ -231,11 +232,20 @@ void Jacobi<ValueType, IndexType>::generate(const LinOp *system_matrix,
...
     const auto exec = this->get_executor();
     if (parameters_.max_block_size == 1) {
-        auto diag = as<DiagonalExtractable<ValueType>>(system_matrix)
-                        ->extract_diagonal();
-        auto temp = gko::Array<ValueType>::view(
-            exec, system_matrix->get_size()[0], diag->get_values());
+        auto diag = share(as<DiagonalLinOpExtractable>(system_matrix)
+                              ->extract_diagonal_linop());
+        auto diag_vt =
+            ::gko::detail::temporary_conversion<matrix::Diagonal<ValueType>>::
```
upsj, line 238: same here: Is the fully qualified name necessary?
```suggestion
            detail::temporary_conversion<matrix::Diagonal<ValueType>>::
```

upsj (approved):
LGTM! Thanks also for fixing the wrong executor in the array view :)

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -654,6 +654,28 @@ class Preconditionable {
...
 
 
+/**
+ * The diagonal of a LinOp can be extracted. It will be implemented by
+ * DiagonalExtractable<ValueType>, so the class does not need to implement it.
+ * extract_diagonal_linop returns a linop which extracts the elements whose col
+ * and row index are the same and stores the result in a min(nrows, ncols) x 1
+ * dense matrix.
+ *
+ * @ingroup LinOp
```
tcojean, line 664:
```suggestion
 * @ingroup diagonal
 * @ingroup LinOp
```

tcojean (approved):
LGTM.

core/preconditioner/jacobi.cpp
yhmtsai: I need to put `gko::`, or I will get the error
` error: 'temporary_conversion' is not a member of 'gko::preconditioner::detail'`
it look for the function in `gko::preconditioner::detail` not `gko::detail`

PR #835, opened 2021-07-15
Remove multiple RHS capability from and optimizations to batch GMRES, BiCGStab and Richardson

Slaedr:
The changes are as follows.

- Because allowing multiple right-hand sides (RHS) was slowing down batch Richardson and biCGStab significantly, it was decided to remove that from Cuda implementations for now. This has only been done for BiCGStab, GMRES and Richardson for now.
- This also makes the `BatchEntry` structs unnecessary, and removing these improves the performance further.
- Scaling of the matrices, RHS and solution vector is now moved to separate kernels and called from core.
- A CMake option `GINKGO_CUDA_BATCH_SOLVERS_MAXREGCOUNT` was added to limit register usage.

Profile summary of original unoptimized code for :
batch-bcgs-before

And after all the above points (the default block size for batch BiCGStab was increase to 256):
batch-bcgs-after

Issues to be fixed in future PRs:
- IDR and CG are not yet optimized. This also means we currently keep code that can be removed after these are also optimized.
- Lot of code duplication outside the actual Cuda/Hip kernels.
- No attention has been paid to the OpenMP implementations.


Slaedr:
format!

benchmark/solver/batch_solver_profile.cpp
```diff
@@ -0,0 +1,39 @@
...
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+#include "benchmark/solver/batch_solver.hpp"
+
+
+int main(int argc, char *argv[])
+{
+    return read_data_and_launch_benchmark(argc, argv, false);
```
pratikvn, line 38: Do you need two separate files just different parameters ?

Slaedr: I basically need a separate executable that does not depend on stdin. I had trouble running it in some profiling tools otherwise; I think both nvvp and nsight had issues with that. It's also easier to use it with cuda-gdb this way.

pratikvn: I see, maybe you can add a comment, so that we remember the reason ?

common/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -285,127 +178,67 @@ __device__ __forceinline__ void update_x_middle(
...
 template <typename StopType, typename PrecType, typename LogType,
           typename BatchMatrixType, typename ValueType>
-__global__ void apply_kernel(
-    const int max_iter, const gko::remove_complex<ValueType> tol,
-    LogType logger, PrecType prec_shared, const BatchMatrixType a,
-    const gko::batch_dense::UniformBatch<const ValueType> left,
-    const gko::batch_dense::UniformBatch<const ValueType> right,
-    const gko::batch_dense::UniformBatch<ValueType> b,
-    const gko::batch_dense::UniformBatch<ValueType> x)
+__global__ void apply_kernel(const int shared_gap, const int max_iter,
```
pratikvn, line 181: Is there a particular reason to pass this shared_gap to the function ? And not just infer it from num_rows ?

common/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -416,180 +249,95 @@ __global__ void apply_kernel(
...
+            //__syncthreads();
 
+            if (threadIdx.x == blockDim.x - 1) {
+                rho_old_sh[0] = rho_new_sh[0];
+            }
             __syncthreads();
         }
 
-        if (left_global_entry.values) {
-            batch_scale(right_global_entry, x_shared_entry);
```
pratikvn, line 585: I guess now scaling is done via another kernel call ?

common/solver/batch_cg_kernels.hpp.inc
```diff
@@ -187,18 +187,16 @@ template <typename StopType, typename PrecType, typename LogType,
 __global__ void apply_kernel(
     const int max_iter, const gko::remove_complex<ValueType> tol,
     LogType logger, PrecType prec_shared, const BatchMatrixType a,
-    const gko::batch_dense::UniformBatch<const ValueType> left,
-    const gko::batch_dense::UniformBatch<const ValueType> right,
-    const gko::batch_dense::UniformBatch<ValueType> b,
+    const gko::batch_dense::UniformBatch<const ValueType> b,
```
pratikvn, line 190: Looks like CG has not been specialized to 1RHS ? Does the multiple RHS version efficient enough for this ?

common/solver/batch_gmres_kernels.hpp.inc
```diff
@@ -35,403 +35,210 @@ namespace {
...
-        const int c = li % H_shared_entry.num_rhs;
-
-        H_shared_entry.values[r * H_shared_entry.stride + c] =
-            zero<ValueType>();
+    if (threadIdx.x / config::warp_size == 0) {
+        single_compute_norm2(warp_grp, a_gl_entry.num_rows, b_global_entry,
+                             rhs_norms);
+    } else if (threadIdx.x / config::warp_size == 1) {
+        single_compute_norm2(warp_grp, a_gl_entry.num_rows, r_shared_entry,
+                             res_norms);
```
pratikvn, line 66: You would probably need another `__syncthreads()` after this ?

common/solver/batch_gmres_kernels.hpp.inc
```diff
@@ -35,403 +35,210 @@ namespace {
...
-    const gko::batch_dense::BatchEntry<real_type> &z_norms_shared_entry =
-        tmp_norms_shared_entry;
-
-    compute_norm2(z_shared_entry, z_norms_shared_entry, converged);
-
+    __shared__ real_type tmp_norm;
+    if (threadIdx.x / config::warp_size == 0) {
+        single_compute_norm2(warp_grp, num_rows, z_shared_entry, tmp_norm);
+    }
     __syncthreads();
```
pratikvn, line 100: I get the idea , but I guess this means that a lot of warps are stalling here ?

common/solver/batch_idr_kernels.hpp.inc
```diff
@@ -846,9 +846,7 @@ __global__ void apply_kernel(
     const int subspace_dim, const gko::remove_complex<ValueType> kappa,
     const bool smoothing, const bool deterministic, LogType logger,
     PrecType prec_shared, const BatchMatrixType a,
-    const gko::batch_dense::UniformBatch<const ValueType> left,
-    const gko::batch_dense::UniformBatch<const ValueType> right,
-    const gko::batch_dense::UniformBatch<ValueType> b,
+    const gko::batch_dense::UniformBatch<const ValueType> b,
```
pratikvn, line 849: I guess this also has not yet been adapted for single rhs ?

pratikvn (commented):
Some comments

common/matrix/batch_csr_kernels.hpp.inc
```diff
@@ -67,12 +67,14 @@ __device__ __forceinline__ void spmv_kernel(
...
-        // subwarp_grp level reduction
-        for (int i = min(gap_rounded_off, static_cast<int>(tile_size)) / 2;
-             i > 0; i /= 2) {
+        // const int gap_rounded_off =
+        //    1 << static_cast<int>(
+        //        ceil(log2(static_cast<float>(row_end - row_start))));
+        //// subwarp_grp level reduction
+        // for (int i = min(gap_rounded_off, static_cast<int>(tile_size)) / 2;
+        //     i > 0; i /= 2) {
+#pragma unroll
```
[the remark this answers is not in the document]
Slaedr: Yes, the difference was significant.

common/solver/batch_bicgstab_kernels.hpp.inc
Slaedr: Yes, it's now done in core using separate kernel calls. I think it's just more maintainable that way, and the overhead is not significant for this.

common/matrix/batch_dense_kernels.hpp.inc
```diff
@@ -207,7 +207,8 @@ __device__ __forceinline__ void one_norm2(
     }
 
     // subwarp_grp level reduction
-    for (int j = subwarp_grp.size() / 2; j > 0; j /= 2) {
+#pragma unroll
+    for (int j = config::warp_size / 2; j > 0; j /= 2) {
```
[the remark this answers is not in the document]
Slaedr: I'm not sure; I have not tested it for anything but `warp_size`.

common/solver/batch_bicgstab_kernels.hpp.inc
Slaedr: It is indeed inferred from num_rows, but I don't want to repeat that calculation logic, so I'm passing it to the kernel. It turns out to be slightly more performant with the shared memory allocation rounded to 32.

common/solver/batch_cg_kernels.hpp.inc
Slaedr: No, that's just pending. We won't be using CG immediately, so it can be done later. Once I do CG and IDR as well, a bunch of cuda code for multiple RHS can also be removed.

common/solver/batch_gmres_kernels.hpp.inc
Slaedr: We don't need it for correctness, I think. Would we want it for performance? At the moment I don't think that's the case either.

common/solver/batch_gmres_kernels.hpp.inc
Slaedr: Yes. According to Isha's tests, this should be faster than trying to use all warps, at least for the matrices we are working with right now.

common/solver/batch_idr_kernels.hpp.inc
Slaedr: Yes, similar to CG.

cuda/matrix/batch_dense_kernels.cu
```diff
@@ -60,11 +60,12 @@ namespace cuda {
 namespace batch_dense {
 
 
-constexpr auto default_block_size = 512;
+constexpr auto default_block_size = 256;
```
pratikvn, line 63: I guess you experimented and found that 256 was a better default block size that 512 ? Do you know why ?

cuda/solver/batch_cg_kernels.cu
```diff
@@ -74,70 +74,54 @@ namespace batch_cg {
...
+#if GKO_CUDA_BATCH_USE_DYNAMIC_SHARED_MEM
+        gko::kernels::batch_cg::local_memory_requirement<ValueType>(a.num_rows,
+                                                                    b.num_rhs);
+#else
+        0;
+#endif
 
     if (opts.preconditioner == gko::preconditioner::batch::type::none) {
-        const int shared_size =
+        shared_size +=
```
pratikvn, line 102: I think you are missing the `#else` here

cuda/solver/batch_cg_kernels.cu
```diff
@@ -74,70 +74,54 @@ namespace batch_cg {
...
-                <<<nbatch, default_block_size, shared_size>>>(
-                    opts.max_its, opts.residual_tol, logger,
-                    BatchIdentity<ValueType>(), a, left, right, b, x);
+            BATCH_CG_KERNEL_LAUNCH(AbsResidualMaxIter, BatchIdentity);
         }
 
     } else if (opts.preconditioner ==
                gko::preconditioner::batch::type::jacobi) {
-        const int shared_size =
+        shared_size +=
```
pratikvn, line 115: Missing `#else`

cuda/solver/batch_gmres_kernels.cu
```diff
@@ -74,68 +75,58 @@ namespace batch_gmres {
...
-                    opts.max_its, opts.residual_tol, opts.restart_num, logger,
-                    BatchIdentity<ValueType>(), a, left, right, b, x);
+            BATCH_GMRES_KERNEL_LAUNCH(SimpleRelResidual, BatchIdentity);
         }
-
-
     } else if (opts.preconditioner ==
                gko::preconditioner::batch::type::jacobi) {
-        const int shared_size =
+        shared_size +=
```
pratikvn, line 120: MIssing `#else`

reference/test/matrix/batch_dense_kernels.cpp
```diff
@@ -134,15 +134,15 @@ class BatchDense : public ::testing::Test {
...
-    template <typename MtxType>
-    std::unique_ptr<MtxType> gen_mtx(int num_rows, int num_cols)
-    {
-        return gko::test::generate_random_matrix<MtxType>(
-            num_rows, num_cols,
-            std::uniform_int_distribution<gko::size_type>(num_cols, num_cols),
-            std::normal_distribution<gko::remove_complex<value_type>>(0.0, 1.0),
-            rand_engine, exec);
-    }
+    // template <typename MtxType>
```
pratikvn, line 137: Debug ?

pratikvn (approved):
Some minor things. Otherwise LGTM!

cuda/matrix/batch_dense_kernels.cu
Slaedr: That's a good question. I don't remember why I changed this. I think it was while I was optimizing the scale-by-vector operation. In any case it matters little because most of the time is spent in the solver apply kernel.

cuda/solver/batch_cg_kernels.cu
Slaedr: Good catch, the `shared_size +=` should be inside the conditional.

PR #840, opened 2021-07-21
Ignore IDE generated files

gfcas:
According to the dealii/.gitignore file this adds some IDE generated files to the ``.gitignore`` file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 # Created by https://www.gitignore.io/api/
 
-###CMake
+### CMake
 CMakeCache.txt
 CMakeFiles
 CMakeScripts
@@ -12,21 +12,21 @@ compile_commands.json
 CTestTestfile.cmake
 build
 
-###IDE
-# clion
+### IDE
+# Clion
 .idea
 
-# eclipse
+# Eclipse
 .cproject
 .project
 
-# visual studio code
+# Visual studio code
 .vscode
 
 ### Generated header file
 include/config.hpp
 
-###C++
+### C++
 # Prerequisites
 *.d
 
@@ -60,10 +60,25 @@ include/config.hpp
 *.out
 *.app
 
-#other
+# Others
 diff.patch
 GRTAGS
 GSYMS
 GTAGS
 GPATH
 .gitignore
+
+### OS
+# temporary and backup files
+*~
+*#
+*.swp
+
+# MacOS self-generated files
+.DS_Store
+.DS_Store?
+._*
+.Spotlight-V100
+.Trashes
+ehthumbs.db
+Thumbs.db
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,17 @@ compile_commands.json
 CTestTestfile.cmake
 build
 
+###IDE
+# clion
+.idea
+
+# eclipse
+.cproject
+.project
+
+# visual studio code
+.vscode
+
 ### Generated header file
 include/config.hpp
 


upsj:
Generally this sounds like a good idea, you could also add `.idea` for CLion-specific stuff.

I personally work around this kind of issue by setting a global .gitignore and/or putting a .gitignore with content `*` into folders I want to ignore, like CMake build trees, which are not always named `build`

gfcas:
> 
> 
> Generally this sounds like a good idea, you could also add `.idea` for CLion-specific stuff.
> 
> I personally work around this kind of issue by setting a global .gitignore and/or putting a .gitignore with content `*` into folders I want to ignore, like CMake build trees, which are not always named `build`

Of course ... already done. I use VSC but just copied the part from the dealii/.gitignore file.

.gitignore
```diff
@@ -12,6 +12,18 @@ compile_commands.json
...
 build
 
+###IDE
+# clion
+.idea
+
+# eclipse
+.cproject
+.project
+qtcreator
```
upsj, line 22: This one is a separate IDE, though I am not sure that is the canonical settings path they use? Has been a while since I've last used it.

upsj (approved):
LGTM!

.gitignore
gfcas: What do you suggest? Removing some lines?

upsj: If somebody has QtCreator installed and can check quickly, we can look more closely at it, but for now, I would suggest removing it. There are other files that might be helpful (https://github.com/github/gitignore/blob/218a941be92679ce67d0484547e3e142b2f5f6f0/Qt.gitignore), but since nobody of us seems to be using it, and Ginkgo is not a Qt application, I guess we can ignore those.

gfcas: OK

pratikvn (approved):
You might also consider adding `.DS_Store` automatically created on Mac.

gfcas:
> 
> 
> You might also consider adding `.DS_Store` automatically created on Mac.

MacOS and some further OS related files are added. As already said, I copied the entries from the dealii/.gitignore file. Should we mention this somewhere?

thoasm (approved):
LGTM!

.gitignore
```diff
@@ -12,10 +12,21 @@ compile_commands.json
 CTestTestfile.cmake
 build
 
+### IDE
+# Clion
+.idea
+
+# Eclipse
```
yhmtsai, line 19: https://github.com/github/gitignore/blob/master/Global/Eclipse.gitignore

.gitignore
```diff
@@ -12,10 +12,21 @@ compile_commands.json
...
+.cproject
+.project
+
+# Visual studio code
```
yhmtsai, line 23: https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore

.gitignore
```diff
@@ -49,10 +60,25 @@ include/config.hpp
...
+# Others
 diff.patch
 GRTAGS
 GSYMS
 GTAGS
 GPATH
 .gitignore
+
+### OS
+# temporary and backup files
```
yhmtsai, line 72: Linux?
https://github.com/github/gitignore/blob/master/Global/Linux.gitignore

.gitignore
```diff
@@ -49,10 +60,25 @@ include/config.hpp
...
+*~
+*#
+*.swp
+
+# MacOS self-generated files
```
yhmtsai, line 77: https://github.com/github/gitignore/blob/master/Global/macOS.gitignore

yhmtsai (commented):
I do not really think it should be in ginkgo.
it is the editor/platform dependent.
according to you have MacOS, so maybe needs Windows https://github.com/github/gitignore/blob/master/Global/Windows.gitignore
https://github.com/github/gitignore gives many gitignore sample
they also suggest putting these into global gitignore

pratikvn:
In general I agree with @yhmtsai , with environment specific modifications (IDE's, OSes etc), it might be better to add it to `.git/info/exclude` rather than `.gitignore`.

upsj:
I am fine with the PR in its current stage. If somebody starts developing with Ginkgo on a new platform, we can always add additional ignores, but for now I think we can leave it as it is, and not try to catch every eventuality.

upsj:
rebase!

gfcas:
Sorry, I was offline for a couple of days. I was not aware of the github/gitignore repo. After reading the hint of @yhmtsai I would have agreed in not adding this PR. I don't know maybe you want to revert it? Nevertheless, thanks for your suggestions! It helps me a lot for my own project.

thoasm:
@gfcas I don't think it is necessary to revert the PR. These are quality of life improvements that don't hurt Ginkgo and may improve the development process for some people. If we need to commit any of the file formats that are now in the `.gitignore`, we can simply remove them from the file.
If anyone disagrees (for example @yhmtsai), I am happy to discuss. I am not categorically against reverting the PR, but I like both the changes done in this PR and think that there is basically no downside in keeping it in Ginkgo.

yhmtsai:
Yes, I do not have a strong opinion to revert this pr because it does not hurt anything.

PR #841, opened 2021-07-21
Fix enablelogging superclass

MarcelKoch:
Currently, the documentation of `EnableLogging` says that it is derived from the template parameter, but that is not the case. This PR fixes the issue, although it seems like it didn't cause any problems.

### DIFF
--- a/include/ginkgo/core/log/logger.hpp
+++ b/include/ginkgo/core/log/logger.hpp
@@ -584,7 +584,7 @@ public:
  *                          class.
  */
 template <typename ConcreteLoggable, typename PolymorphicBase = Loggable>
-class EnableLogging : public Loggable {
+class EnableLogging : public PolymorphicBase {
 public:
     void add_logger(std::shared_ptr<const Logger> logger) override
     {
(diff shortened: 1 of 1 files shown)


include/ginkgo/core/log/logger.hpp
```diff
@@ -594,7 +594,7 @@ class EnableLogging : public Loggable {
     void remove_logger(const Logger *logger) override
     {
         auto idx = find_if(begin(loggers_), end(loggers_),
-                           [&logger](std::shared_ptr<const Logger> l) {
+                           [&logger](std::shared_ptr<const Logger> &l) {
```
upsj, line 597: tiny nit:
```suggestion
                           [&logger](const std::shared_ptr<const Logger> &l) {
```
or `const auto& l` would make sure we don't modify the smart pointer.

upsj (approved):
LGTM! Great catch!

yhmtsai (approved):

yhmtsai:
I disable the ready_for_merge first to avoid this pr always execute the full pipeline.
when the MSVC cuda issue is solved, please feel free to add it back.

tcojean:
rebase!

PR #843, opened 2021-07-22
Dpcpp dense transpose, (cb)gmres impl and enable apply test of solver

yhmtsai:
This PR adds
- Dpcpp Dense transpose
- cbgmres, gmres 
- atomic_max bind
- enable the test of solver on dpcpp
- swap some dpcpp job. three jobs are in quick pipeline, and others are in full pipeline
- put the free_after_use in gtest directly not assert_no_exit, which gives issue on devcloud cpu.
if it fails it terminate whole executor test, so we still get the failed information


dpcpp/base/onemkl_bindings.hpp
```diff
@@ -90,7 +90,7 @@ struct is_supported<std::complex<float>> : std::true_type {};
 template <>
 struct is_supported<std::complex<double>> : std::true_type {};
 
-
+namespace {
```
yhmtsai, line 93: TODO: fix

dpcpp/stop/residual_norm_kernels.dp.cpp
```diff
@@ -52,7 +56,66 @@ namespace dpcpp {
 namespace residual_norm {
 
 
-constexpr int default_group_size = 512;
+constexpr int default_block_size = 256;
+
+
+template <typename ValueType>
+void residual_norm_kernel(size_type num_cols, ValueType rel_residual_goal,
```
yhmtsai, line 63: maybe implement it with sycl directly not porting

MarcelKoch:
Regarding the sonarcloud failure, it should be easily fixed by extracting the shared gmres/cb-gmres kernels similar to the cuda/hip implementation.

yhmtsai:
@MarcelKoch Thanks, I forgot to handle it after I copy it during porting

dpcpp/components/atomic.dp.hpp
```diff
@@ -114,6 +126,19 @@ struct atomic_helper {
...
 
+template <cl::sycl::access::address_space addressSpace, typename ValueType,
+          typename = void>
+struct atomic_max_helper {
+    __dpct_inline__ static ValueType atomic_max(ValueType *, ValueType)
+    {
+        static_assert(sizeof(ValueType) == 0,
+                      "This default function is not implemented, only the "
+                      "specializations are.");
+        // TODO: add proper implementation of generic atomic max
```
upsj, line 137: We have atomic_max_helper, is that not sufficiently generic?

dpcpp/components/atomic.dp.hpp
```diff
@@ -203,6 +279,14 @@ __dpct_inline__ T atomic_add(T *__restrict__ addr, T val)
 }
 
 
+template <cl::sycl::access::address_space addressSpace = atomic::global_space,
+          typename T>
+__dpct_inline__ T atomic_max(T *__restrict__ addr, T val)
```
upsj, line 284: I am not certain `__restrict__` makes any difference here, since the function takes only a single pointer parameter.

dpcpp/matrix/dense_kernels.dp.cpp
```diff
@@ -671,6 +671,111 @@ GKO_ENABLE_IMPLEMENTATION_CONFIG_SELECTION(reduce_total_cols,
 GKO_ENABLE_DEFAULT_CONFIG_CALL(reduce_total_cols_call, reduce_total_cols,
                                kcfg_1d_list)
 
+template <std::uint32_t sg_size, typename ValueType, typename Closure>
+void transpose(const size_type nrows, const size_type ncols,
+               const ValueType *__restrict__ in, const size_type in_stride,
+               ValueType *__restrict__ out, const size_type out_stride,
+               Closure op, sycl::nd_item<3> item_ct1,
+               UninitializedArray<ValueType, sg_size *(sg_size + 1)> *space)
```
upsj, line 679: You could also pass `&space[0]` as a parameter directly

dpcpp/stop/residual_norm_kernels.dp.cpp
```diff
@@ -52,7 +56,66 @@ namespace dpcpp {
...
+            sycl_nd_range(grid, block), [=](sycl::nd_item<3> item_ct1) {
+                residual_norm_kernel(num_cols, rel_residual_goal, tau, orig_tau,
+                                     stoppingId, setFinalized, stop_status,
+                                     device_storage, item_ct1);
+            });
+    });
+}
+
+
+void init_kernel(bool *__restrict__ device_storage)
```
upsj, line 104: `__restrict__` shouldn't have an impact on single-parameter functions

dpcpp/stop/residual_norm_kernels.dp.cpp
upsj: I agree here, this would be a pretty straightforward kernel.

dpcpp/solver/gmres_kernels.dp.cpp
```diff
@@ -54,53 +68,374 @@ namespace dpcpp {
...
-namespace {
-
-
-template <typename ValueType>
-void finish_arnoldi(size_type num_rows, matrix::Dense<ValueType> *krylov_bases,
-                    matrix::Dense<ValueType> *hessenberg_iter, size_type iter,
-                    const stopping_status *stop_status) GKO_NOT_IMPLEMENTED;
+constexpr int default_block_size = 256;
+// default_dot_dim can not be 64 in hip because 64 * 64 exceeds their max block
+// size limit.
```
upsj, line 73: nit: remove

dpcpp/CMakeLists.txt
```diff
@@ -75,7 +78,7 @@ else ()
     target_link_options(ginkgo_dpcpp PUBLIC -fsycl-device-code-split=per_kernel)
 endif()
 target_link_libraries(ginkgo_dpcpp PUBLIC ginkgo_device)
-target_link_libraries(ginkgo_dpcpp PRIVATE MKL::MKL_DPCPP)
+target_link_libraries(ginkgo_dpcpp PRIVATE MKL::MKL_DPCPP oneDPL)
```
upsj, line 81: Are we using this already? Or is this preparation for IDR? Maybe leave it for a later PR then?

upsj (approved):
LGTM!

dpcpp/base/onemkl_bindings.hpp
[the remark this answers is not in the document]
yhmtsai: I answer this together in the next one

dpcpp/base/onemkl_bindings.hpp
```diff
@@ -91,36 +94,34 @@ template <>
...
-
-// Bind the dot for x^T * y
-GKO_BIND_DOT(float, dot, oneapi::mkl::blas::row_major::dot);
-GKO_BIND_DOT(double, dot, oneapi::mkl::blas::row_major::dot);
-GKO_BIND_DOT(std::complex<float>, dot, oneapi::mkl::blas::row_major::dotu);
-GKO_BIND_DOT(std::complex<double>, dot, oneapi::mkl::blas::row_major::dotu);
+#define GKO_DECLARE_DOT(ValueType, Name)                                 \
+    void Name(::cl::sycl::queue &exec_queue, std::int64_t n,             \
+              const ValueType *x, std::int64_t incx, const ValueType *y, \
+              std::int64_t incy, ValueType *result)
```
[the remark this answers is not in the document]
yhmtsai: When I use it like cuda, compiler will complain they have multiple.
I use anonymous neamspace originally such that it is built for each file contains onemkl_binding header.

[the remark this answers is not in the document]
upsj: have you tried making them `inline`? That's also the reason why most functions in our cuBLAS wrappers are templates, since they are implicitly `inline`/weak symbols

dpcpp/CMakeLists.txt
yhmtsai: good catch. Yes, it is the IDR related thing. I will delete them

dpcpp/base/onemkl_bindings.hpp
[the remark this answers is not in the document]
yhmtsai: Thanks! it works, and I miss inline

dpcpp/components/atomic.dp.hpp
thoasm: I would keep it as it can also improve caching.

dpcpp/components/atomic.dp.hpp
yhmtsai: It is copy-paste from cuda

tcojean (commented):
Some small comments

dpcpp/CMakeLists.txt
```diff
@@ -8,6 +8,8 @@ set(GINKGO_DPCPP_VERSION ${GINKGO_DPCPP_VERSION} PARENT_SCOPE)
 
 find_package(MKL CONFIG REQUIRED HINTS "$ENV{MKLROOT}")
 set(GINKGO_MKL_ROOT "${MKL_ROOT}" PARENT_SCOPE)
+find_package(oneDPL REQUIRED HINTS "$ENV{DPL_ROOT}")
+set(GINKGO_DPL_ROOT "${DPL_ROOT}" PARENT_SCOPE)
```
tcojean, line 12: Do we need this in this PR? This seems unused so far, I would put it in the related PR instead.

PR #845, opened 2021-07-23
dpcpp porting the rest of matrix format

yhmtsai:
This PR ports the rest of matrix format

TODO:
- [x] ~~refine the history such that the unchanged part still be unchanged in the history~~
git history already looks good
- [x] csr dpcpp subgroup size (use 16 for now)
- [x] ~~Csr apply Unsorted Csr~~ delete the test, because cuda/we do not ensure it always work


dpcpp/matrix/hybrid_kernels.dp.cpp
```diff
@@ -57,6 +65,172 @@ namespace dpcpp {
...
+
+template <typename ValueType1, typename ValueType2>
+void add(dim3 grid, dim3 block, size_type dynamic_shared_memory,
+         sycl::queue *queue, size_type num, ValueType1 *val1,
+         const ValueType2 *val2)
+{
+    queue->submit([&](sycl::handler &cgh) {
+        cgh.parallel_for(
+            sycl_nd_range(grid, block),
+            [=](sycl::nd_item<3> item_ct1) { add(num, val1, val2, item_ct1); });
```
Slaedr, line 226: Nit: For small functions, you could probably just put the function body in the lambda here.

dpcpp/matrix/csr_kernels.dp.cpp
```diff
@@ -71,11 +78,1812 @@ namespace dpcpp {
...
+                     UninitializedArray<IndexType, spmv_block_size> *tmp_ind,
+                     UninitializedArray<ValueType, spmv_block_size> *tmp_val)
+{
+    merge_path_reduce(
+        nwarps, last_val, last_row, c, c_stride, [](ValueType &x) { return x; },
+        item_ct1, tmp_ind, tmp_val);
+}
+
+template <typename ValueType, typename IndexType>
+void abstract_reduce(dim3 grid, dim3 block, size_type dynamic_shared_memory,
```
Slaedr, line 608: Do you actually use this `dynamic_shared_memory` parameter anywhere? You could consider removing it wherever it's not used, which seems to be everywhere.

dpcpp/test/matrix/sellp_kernels.cpp
```diff
@@ -0,0 +1,412 @@
...
+
+    void set_up_apply_vector(
+        int slice_size = gko::matrix::default_slice_size,
+        int stride_factor = gko::matrix::default_stride_factor,
+        int total_cols = 0)
+    {
+        mtx = Mtx::create(ref);
+        mtx->copy_from(gen_mtx(532, 231));
+        expected = gen_mtx(532, 1);
+        y = gen_mtx(231, 1);
```
Slaedr, line 97:
```suggestion
        y = gen_mtx(231, total_cols);
```
You could remove the function below and just use this one with the `total_cols` parameter. I guess you would need to move `empty` here as well.

Slaedr (commented):
Nice work! I have a few comments and questions below. Most of them are around reducing the number of layers of functions for code size reasons.

In addition, have you thought whether the Dpcpp kernel implementations can be unified with those in `common`? Right now, it looks like most Dpcpp kernels are essentially
the common kernel + add `nd_item` parameter + find and replace.
Perhaps we could write a simple `nd_item` implementation for Cuda and make sure it's optimized away in release builds using constexpr functions and such. And there could a `GKO_KERNEL` defined to `__global__` for cuda and nothing for Dpcpp, etc. Since you already did the work of unifying Cuda and Dpcpp cooperative groups etc., this would be nice to reduce code duplication in these complicated kernels. I guess perhaps local shared memory is a significant barrier to doing this. Of course, even if it's possible, this will be a different PR.

cuda/test/matrix/csr_kernels.cpp
```diff
@@ -429,23 +429,6 @@ TEST_F(Csr, SimpleApplyToCsrMatrixIsEquivalentToRef)
 }
 
 
-TEST_F(Csr, SimpleApplyToCsrMatrixIsEquivalentToRefUnsorted)
```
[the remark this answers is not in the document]
upsj: The test tests something that is not officially supported by cuSPARSE and only works incidentally, so we should remove it.

dpcpp/matrix/diagonal_kernels.dp.cpp
```diff
@@ -51,11 +57,244 @@ namespace dpcpp {
...
+                    ValueType *__restrict__ result_values,
+                    sycl::nd_item<3> item_ct1)
+{
+    const auto tidx = thread::get_thread_id_flat(item_ct1);
+    const auto row = tidx / num_cols;
+    const auto col = tidx % num_cols;
+
+    if (row < num_rows) {
+        result_values[row * result_stride + col] =
+            source_values[row * source_stride + col] * diag[row];
```
upsj, line 80: common kernels?

dpcpp/matrix/sparsity_csr_kernels.dp.cpp
```diff
@@ -33,28 +33,14 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-#include "core/base/iterator_factory.hpp"
-#include "dpcpp/components/format_conversion.dp.hpp"
 
 
 namespace gko {
 namespace kernels {
 namespace dpcpp {
 /**
- * @brief The SparsityCsr pattern format namespace.
+ * @brief The Compressed sparse row matrix format namespace.
```
upsj, line 43: That doesn't seem entirely correct, since it is only the pattern, not the values.

dpcpp/components/intrinsics.dp.hpp
```diff
@@ -0,0 +1,92 @@
...
+namespace gko {
+namespace kernels {
+namespace dpcpp {
+
+
+/**
+ * @internal
+ * Returns the number of set bits in the given mask.
+ */
+__dpct_inline__ int popcnt(uint32 mask) { return sycl::popcount(mask); }
```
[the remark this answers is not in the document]
upsj: IMO bitmasks should always be `unsigned` and indices preferably `signed`, not sure I like what the SYCL standard is doing here - compare C++20's [popcount](https://en.cppreference.com/w/cpp/numeric/popcount). I think the implicit cast uint32 -> int for the return value here is fine, since the outputs range from 0 to 32/64 only.

dpcpp/components/merging.dp.hpp
```diff
@@ -0,0 +1,338 @@
...
+ *                  `*_val` and `*_idx` are the values resp. the indices of the
+ *                  values from a/b being compared at output index `c_index`.
+ *                  `valid` specifies if the current thread has to merge an
+ *                  element (this is necessary for shfl and ballot operations).
+ *                  It must return `false` on all threads of the group iff the
+ *                  merge shouldn't be continued.
+ */
+template <int group_size, typename ValueType, typename IndexType,
+          typename Group, typename Callback>
+__dpct_inline__ void group_merge(const ValueType *__restrict__ a,
```
upsj, line 159: TBH I'm not sure whether it makes sense to have these also for SYCL. Do random-access shuffles give good performance on Intel GPUs?

upsj (approved):
LGTM, great job! I think there are a few kernels for which the 3D indexing scheme doesn't necessarily make much sense (conversions especially, everything with flat indexing and no particular performance tuning), and a few kernels could be moved to `common`.

yhmtsai:
@upsj @Slaedr Sorry, I do not delete unused functions yet. The functions (merging and some instruction like popcnt etc) should not be in this pr. I keep using @upsj #799 , so there's no test to test these functions.
thanks for the comments on these. I will try replying to something but it may not be correct because I do not have experience with them. And we can reference them again when we have another pr for it

dpcpp/components/segment_scan.dp.hpp
```diff
@@ -0,0 +1,87 @@
...
+__dpct_inline__ bool segment_scan(
+    const group::thread_block_tile<subwarp_size> &group, const IndexType ind,
+    ValueType *__restrict__ val)
+{
+    bool head = true;
+#pragma unroll
+    for (int i = 1; i < subwarp_size; i <<= 1) {
+        const IndexType add_ind = group.shfl_up(ind, i);
+        ValueType add_val = zero<ValueType>();
+        if (add_ind == ind && group.thread_rank() >= i) {
```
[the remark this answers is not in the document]
yhmtsai: Using thread_rank should be better than sycl/cuda, which virtualizes the group operations.
sycl will be nd_item.get_local_id(2) if using porting

dpcpp/matrix/coo_kernels.dp.cpp
```diff
@@ -61,11 +69,437 @@ namespace dpcpp {
...
+            }
+        }
+        temp += val[idx] * b[col[idx] * b_stride + column_id];
+        atomic_add(&(c[curr_row * c_stride + column_id]), scale(temp));
+    }
+}
+
+
+template <typename ValueType, typename IndexType>
+void abstract_spmm(const size_type nnz, const size_type num_elems,
```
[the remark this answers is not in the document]
yhmtsai: you should be right. SYCL should be able to pass the lambda directly.
Should we go this way?

dpcpp/matrix/csr_kernels.dp.cpp
```diff
@@ -71,11 +78,1812 @@ namespace dpcpp {
...
+        if (tid_in_warp == 0) {
+            auto slice_length =
+                ceildiv(warp_result, stride_factor) * stride_factor;
+            slice_lengths[sliceid] = slice_length;
+            slice_sets[sliceid] = slice_length;
+        }
+    }
+}
+
+void calculate_slice_lengths(dim3 grid, dim3 block,
```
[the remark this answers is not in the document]
yhmtsai: we have a macro for this simple kernel wrapper. something like `hipLaunchKernelGGL` indeed can handle this situation.
but it can not be a general case because some kernels contain shared_memory.
when the kernel contains shared_memory (dynamic or static), there's the place we allocate shared_memory.
adding one more layer for handling it together should work but we already have many macro/layer in dpcpp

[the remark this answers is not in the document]
yhmtsai: the template is another issue

dpcpp/matrix/diagonal_kernels.dp.cpp
```diff
@@ -51,11 +57,244 @@ namespace dpcpp {
...
+
+    if (tidx >= size) {
+        return;
+    }
+
+    trans_values[tidx] = conj(orig_values[tidx]);
+}
+
+template <typename ValueType>
+void conj_transpose(dim3 grid, dim3 block, size_type dynamic_shared_memory,
```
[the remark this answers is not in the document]
yhmtsai: please see the above reply

dpcpp/matrix/csr_kernels.dp.cpp
yhmtsai: I can only remove the variable name to avoid the warning.
I can not remove the variable because it needs to provide a consistent kernel call.
`kernel(grid, block, dynamic_shared_memory, queue, ...)`

dpcpp/test/matrix/sellp_kernels.cpp
yhmtsai: good catch, we do not use them correctly in sellp test

dpcpp/matrix/csr_kernels.dp.cpp
```diff
@@ -71,11 +78,1820 @@ namespace dpcpp {
...
+constexpr int spmv_block_size = warps_in_block * config::warp_size;
+constexpr int wsize = config::warp_size;
+constexpr int classical_overweight = 32;
+
+
+/**
+ * A compile-time list of the number items per threads for which spmv kernel
+ * should be compiled.
+ */
+using compiled_kernels = syn::value_list<int, 6>;
```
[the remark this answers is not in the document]
yhmtsai: no, I use the smallest one from cuda for merge_spmv

dpcpp/components/merging.dp.hpp
yhmtsai: I don't know actually.

dpcpp/matrix/sparsity_csr_kernels.dp.cpp
yhmtsai: this gives the same name as csr?

thoasm (commented):
Part 1/2 of my review

dpcpp/test/matrix/CMakeLists.txt
```diff
@@ -1,3 +1,7 @@
+ginkgo_create_test(coo_kernels)
 ginkgo_create_test(csr_kernels)
 ginkgo_create_test(dense_kernels)
-ginkgo_create_test(ell_kernels)
```
tcojean, line 3: Shouldn't this stay?

cuda/test/matrix/csr_kernels.cpp
```diff
@@ -429,23 +429,6 @@ TEST_F(Csr, SimpleApplyToCsrMatrixIsEquivalentToRef)
...
-
-    mtx->apply(trans.get(), square_mtx.get());
-    dmtx->apply(d_trans.get(), square_dmtx.get());
-
-    GKO_ASSERT_MTX_NEAR(square_dmtx, square_mtx, 1e-14);
-    GKO_ASSERT_MTX_EQ_SPARSITY(square_dmtx, square_mtx);
-    ASSERT_TRUE(square_dmtx->is_sorted_by_column_index());
-}
-
-
```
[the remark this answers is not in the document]
yhmtsai: in cusparse, they denote the csr sorted by row and then by column
https://docs.nvidia.com/cuda/cusparse/index.html#csr-format
they may implement the spgemm with hash table, so that works in cuda.
but the current impl in dpcpp does not support unsorted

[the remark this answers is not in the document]
upsj: To quote from the Csr documentation:
```
 * Both the SpGEMM and SpGEAM operation require the input matrices to be sorted
 * by column index, otherwise the algorithms will produce incorrect results.
 ```

File: dpcpp/matrix/csr_kernels.dp.cpp [hunk 16]
```diff
@@ -71,11 +78,1820 @@ namespace dpcpp {
...
+{
+    auto tid = thread::get_thread_id_flat(item_ct1);
+    if (tid < num_nonzeros) {
+        out_cols[tid] = permutation[in_cols[tid]];
+        out_vals[tid] = in_vals[tid];
+    }
+    if (tid <= num_rows) {
+        out_row_ptrs[tid] = in_row_ptrs[tid];
+    }
+}
```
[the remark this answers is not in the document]
yhmtsai: when I write it in place, where can I write the __restrict__?
they are captured by lambda

dpcpp/matrix/csr_kernels.dp.cpp
```diff
@@ -71,11 +78,1820 @@ namespace dpcpp {
...
+int compute_items_per_thread(std::shared_ptr<const DpcppExecutor> exec)
+{
+    int num_item = 6;
+    // Ensure that the following is satisfied:
+    // sizeof(IndexType) + sizeof(ValueType)
+    // <= items_per_thread * sizeof(IndexType)
+    constexpr int minimal_num =
+        ceildiv(sizeof(IndexType) + sizeof(ValueType), sizeof(IndexType));
+    int items_per_thread = num_item * 4 / sizeof(IndexType);
+    return std::max(minimal_num, items_per_thread);
```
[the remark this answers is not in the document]
yhmtsai: I need to think about this part.
we have these in cuda because we use gko::max in device not in host. but in dpcpp, it may be not different

dpcpp/components/segment_scan.dp.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+
+/**
+ * @internal
+ *
+ * Compute a segement scan using add operation (+) of a subwarp. Each segment
+ * performs suffix sum. Works on the source array and returns whether the thread
+ * is the first element of its segment with same `ind`.
+ */
+template <unsigned subwarp_size, typename ValueType, typename IndexType>
```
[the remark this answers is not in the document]
yhmtsai: Yes, it is subgroup size

dpcpp/matrix/coo_kernels.dp.cpp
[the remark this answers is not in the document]
Slaedr: I think that would be good, mainly because otherwise all the intermediate functions unnecessarily increases source code size.

dpcpp/matrix/csr_kernels.dp.cpp
Slaedr: Is that because your script produces these functions, and it needs a common function signature?

yhmtsai: function signature is for calling.
when we have the kernel, user will not know the function contains grid, block, dynamic, queue
like hiplaunchkernel always has these four args

dpcpp/matrix/coo_kernels.dp.cpp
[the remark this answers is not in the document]
yhmtsai: With GKO_ENABLE_DEFAULT_HOST, it seems not to reduce the source code.
With GKO_ENABLE_DEFAULT_HOST, we do not have `cgh.parallel_for` but abstract_spmm for lambda.
Without it, we do not have abstract_spmm for lambda but we need to write `cgh.parallel_for` manually

dpcpp/matrix/csr_kernels.dp.cpp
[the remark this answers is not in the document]
thoasm: Is it not possible to add `__restrict__` to the kernel function? Is that not forwarded to the in-place lambda? I guess that would need testing, but I would much rather have that than the current additional lambda, which causes a lot of boiler-plate code for only the `__restrict__`. If that is not possible, we can keep the current code.

dpcpp/matrix/csr_kernels.dp.cpp
Slaedr: So you are planning to have something like a common `syclLaunchKernel`? Because if not, each of these functions can then have its own signature, right?

Slaedr (commented):
I see that you're using the `GKO_ENABLE_DEFAULT_HOST` macro to reduce the source code, which is better than before. But would it not be possible to have an actual function rather than a macro? If shared memory is needed, couldn't you have an `if` statement in there that checks if the requested size is greater than zero? If so, it should create the shared memory buffer and a different `cgh.parallel_for` launch accordingly, and if not, it should just use what you already have in the macro. I don't think there will be any performance implications. Such a function would then be called from the final "kernel" function, very similar to `hipLaunchKernelGGL`. If something can be done with reasonable ease within the language, we should prefer that over macros. If that turns out to be too complicated, then I think what you have right now is fine.

dpcpp/matrix/csr_kernels.dp.cpp
yhmtsai: Yes, each of these functions can have its own signature.
the signature what I mentioned is mainly for kernel syntax.
in cuda, the syntax is in `<<<` `>>>`
in hip, the first five args are syntax in `hipLaunchKernelGGL(kernel_name, grid, block, dynamic_shared_memory, stream, others args)`
in dpcpp, I would like have something similar like this. the first four arguments are kernel syntax

yhmtsai:
@Slaedr the shared memory contains both static and dynamic.
according to the dynamic_shared_memory, I can only know the dynamic one.
the function also needs to know whether passing the pointer to kernel.
dpct will not help it.
```
__global__ kernel(args) {
  __shared__ float a[];
  __shared__ double b[];
  extern __shared__ int s[];
}
kernel<<<grid, block, k>>>(args);
```
in dpcpp,
```
queue->submit({
allocate dynamic in s.
allocate static in a.
allocate static in b.
cgh.parallel_for(..., {
  [=](nd_item) {kernel(item_ct1, args, a, b, s)}
})
}
) 
```

Slaedr (approved):
@yhmtsai I see; I guess I need to read up a bit on this. Alright, if you think it's not worth going there for now, this is good to go from my side.

dpcpp/matrix/csr_kernels.dp.cpp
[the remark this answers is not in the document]
yhmtsai: another issue comes to my mind.
if I put the function inside  the lambda directly, I seems to need to take similar amount lines for it. (if it is mainly from #lines concern)
I need to take the plain pointer first and then can put it in lambda function.
Otherwise, the lambda will try to capture the whole class not just the pointer, which is not allowed.
It will be like following
```
const IndexType *__restrict__ permutation = permutation.get_values();
// .... for each pointer.
and then
cgh.parallel_for(, {...});
```
these may be suitable for common (generic) kernel

dpcpp/components/intrinsics.dp.hpp
```diff
@@ -0,0 +1,92 @@
...
+ * starting from the least significant bit.
+ */
+__dpct_inline__ int ffs(uint32 mask) { return __builtin_ffs(mask); }
+
+/** @copydoc ffs */
+__dpct_inline__ int ffs(uint64 mask)
+{
+    // the cast is necessary, as the overloads defined by HIP are ambiguous
+    return __builtin_ffsll(static_cast<unsigned long long int>(mask));
+}
```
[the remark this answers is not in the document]
yhmtsai: sorry for confusing, I delete it first because I do not use it.

dpcpp/test/matrix/sellp_kernels.cpp
```diff
@@ -0,0 +1,398 @@
...
+    GKO_ASSERT_MTX_NEAR(result, expected, r<vtype>::value);
+}
+
+
+TEST_F(Sellp, SimpleApplyMultipleRHSIsEquivalentToRef)
+{
+    set_up_apply_matrix(64);
+
+
+    mtx->apply(y.get(), expected.get());
```
thoasm, line 191: nit: spacing
```suggestion
    set_up_apply_matrix(64);

    mtx->apply(y.get(), expected.get());
```

dpcpp/test/matrix/sellp_kernels.cpp
```diff
@@ -0,0 +1,398 @@
...
+    GKO_ASSERT_MTX_NEAR(result, expected, r<vtype>::value);
+}
+
+
+TEST_F(Sellp, AdvancedApplyMultipleRHSIsEquivalentToRef)
+{
+    set_up_apply_matrix(64);
+
+
+    mtx->apply(alpha.get(), y.get(), beta.get(), expected.get());
```
thoasm, line 205: Same here (and a lot other lines in this file):
```suggestion
    set_up_apply_matrix(64);

    mtx->apply(alpha.get(), y.get(), beta.get(), expected.get());
```

dpcpp/test/matrix/sellp_kernels.cpp
```diff
@@ -0,0 +1,398 @@
...
+    result->copy_from(dresult.get());
+    GKO_ASSERT_MTX_NEAR(result, expected, r<vtype>::value);
+}
+
+
+TEST_F(Sellp, ApplyToComplexIsEquivalentToRef)
+{
+    set_up_apply_matrix(64);
+
+    auto complex_b = gen_mtx<ComplexVec>(231, 3);
```
thoasm, line 246: Here, `set_up_apply_matrix` is part of the Allocate part, so there should be no empty line between different Allocate stages:
```suggestion
    set_up_apply_matrix(64);
    auto complex_b = gen_mtx<ComplexVec>(231, 3);
```

dpcpp/test/matrix/sellp_kernels.cpp
```diff
@@ -0,0 +1,398 @@
...
+
+    GKO_ASSERT_MTX_NEAR(dcomplex_x, complex_x, r<vtype>::value);
+}
+
+
+TEST_F(Sellp, AdvancedApplyToComplexIsEquivalentToRef)
+{
+    set_up_apply_matrix(64);
+
+    auto complex_b = gen_mtx<ComplexVec>(231, 3);
```
thoasm, line 264: same here
```suggestion
    set_up_apply_matrix(64);
    auto complex_b = gen_mtx<ComplexVec>(231, 3);
```

dpcpp/matrix/coo_kernels.dp.cpp
```diff
@@ -61,11 +70,311 @@ namespace dpcpp {
 namespace coo {
 
 
+constexpr int default_block_size = 256;
+constexpr int warps_in_block = 4;
```
tcojean, line 74: Is this correct? Don't you need 7 or 8 warps/subgroups (number of EU per subslice), * 2 (two ALUs per EU) each of size 16 or whichever?
See https://software.intel.com/content/www/us/en/develop/download/architecture-overview-for-intel-processor-graphics-gen11.html

dpcpp/test/matrix/sellp_kernels.cpp
```diff
@@ -0,0 +1,398 @@
...
+    GKO_ASSERT_MTX_NEAR(dcomplex_x, complex_x, r<vtype>::value);
+}
+
+
+TEST_F(Sellp, ConvertToDenseIsEquivalentToRef)
+{
+    set_up_apply_matrix(64);
+
+
+    auto dense_mtx = gko::matrix::Dense<vtype>::create(ref);
```
thoasm, line 283: 2 empty lines too much (same as the next couple of test), as these are all parts of the Assert block:
```suggestion
    set_up_apply_matrix(64);
    auto dense_mtx = gko::matrix::Dense<vtype>::create(ref);
```

dpcpp/matrix/coo_kernels.dp.cpp
```diff
@@ -61,11 +70,311 @@ namespace dpcpp {
...
+ * @param b_stride  the stride of the input dense vector
+ * @param c  the output dense vector
+ * @param c_stride  the stride of the output dense vector
+ * @param scale  the function on the added value
+ *
+ * @tparam ValueType  type of values stored in the matrix
+ * @tparam IndexType  type of matrix indexes stored in the structure
+ * @tparam Closure  type of the function used to write the result
+ */
+template <int subwarp_size = config::warp_size, typename ValueType,
```
tcojean, line 99: For all of these, I think it would be better in terms of clarity if it was either subgroup_size or warp_size.

```suggestion
template <int subgroup_size = config::warp_size, typename ValueType,
```

cuda/test/matrix/sellp_kernels.cpp
```diff
@@ -207,7 +183,8 @@ TEST_F(Sellp, AdvancedApplyWithSliceSizeAndStrideFActorIsEquivalentToRef)
 
 TEST_F(Sellp, SimpleApplyMultipleRHSIsEquivalentToRef)
 {
-    set_up_apply_matrix();
+    set_up_apply_matrix(64);
+
```
thoasm, line 187: Why did you add an additional empty line? The additional one should be removed (same with all following additions of empty lines after `set_up_apply_matrix`).

cuda/test/matrix/sellp_kernels.cpp
```diff
@@ -261,7 +239,8 @@ TEST_F(Sellp,
 
 TEST_F(Sellp, ApplyToComplexIsEquivalentToRef)
 {
-    set_up_apply_matrix();
+    set_up_apply_matrix(64);
+
     auto complex_b = gen_mtx<ComplexVec>(231, 3);
```
thoasm, line 244: Including this empty line (and the one in the next test) since the setup step is supposed to be part of the Allocation block:
```suggestion
    set_up_apply_matrix(64);
    auto complex_b = gen_mtx<ComplexVec>(231, 3);
```

cuda/test/matrix/sellp_kernels.cpp
```diff
@@ -295,7 +275,7 @@ TEST_F(Sellp, AdvancedApplyToComplexIsEquivalentToRef)
 
 TEST_F(Sellp, ConvertToDenseIsEquivalentToRef)
 {
-    set_up_apply_matrix();
+    set_up_apply_matrix(64);
 
     auto dense_mtx = gko::matrix::Dense<>::create(ref);
```
thoasm, line 280: These empty lines should also be deleted (and in the following tests) since the setup and the following allocations are part of the same Allocation block:
```suggestion
    set_up_apply_matrix(64);
    auto dense_mtx = gko::matrix::Dense<>::create(ref);
```

hip/test/matrix/sellp_kernels.hip.cpp
```diff
@@ -207,7 +183,8 @@ TEST_F(Sellp, AdvancedApplyWithSliceSizeAndStrideFActorIsEquivalentToRef)
 
 TEST_F(Sellp, SimpleApplyMultipleRHSIsEquivalentToRef)
 {
-    set_up_apply_matrix();
+    set_up_apply_matrix(64);
+
```
thoasm, line 187: Same with these empty lines (and the following setup tests in this file):
```suggestion
    set_up_apply_matrix(64);
```

include/ginkgo/core/matrix/csr.hpp
```diff
@@ -493,6 +512,12 @@ class Csr : public EnableLinOp<Csr<ValueType, IndexType>>,
         /* Use imbalance strategy when the matrix has more more than 1e8 on AMD
          * hardware */
         const index_type amd_nnz_limit{static_cast<index_type>(1e8)};
+        /* Use imbalance strategy when the maximum number of nonzero per row is
+         * more than 25600 on Intel hardware. */
+        const index_type intel_row_len_limit = 25600;
```
thoasm, line 517: small nit: These should be `static` (since all the variables should have the same value) and potentially even `constexpr` (if you don't actually need the address of these variables).
Since this is in the public interface, I assume making it `constexpr` might not be ideal, but IMO making it `static` (to have only one `const` variable, and not one per object instance) should be possible.

dpcpp/matrix/hybrid_kernels.dp.cpp
```diff
@@ -57,6 +66,145 @@ namespace dpcpp {
 namespace hybrid {
 
 
+constexpr int default_block_size = 256;
+constexpr int warps_in_block = 4;
```
tcojean, line 70: Same question, is 4 a good number?

dpcpp/matrix/hybrid_kernels.dp.cpp
```diff
@@ -57,6 +66,145 @@ namespace dpcpp {
...
+ * It is almost like COO spmv routine.
+ * It performs is_nonzeros(Coo) times the vector whose values are one
+ *
+ * @param nnz  the number of nonzeros in the matrix
+ * @param num_line  the maximum round of each warp
+ * @param val  the value array of the matrix
+ * @param row  the row index array of the matrix
+ * @param nnz_per_row  the output nonzeros per row
+ */
+template <int subwarp_size = config::warp_size, typename ValueType,
```
tcojean, line 87: `subgroup_size` or `warp_size`

dpcpp/matrix/hybrid_kernels.dp.cpp
```diff
@@ -57,6 +66,145 @@ namespace dpcpp {
...
+
+        bool is_first_in_segment =
+            segment_scan<subwarp_size>(tile_block, curr_row, &temp_val);
+        if (is_first_in_segment) {
+            atomic_add(&(nnz_per_row[curr_row]), temp_val);
+        }
+    }
+}
+
+template <int subwarp_size = config::warp_size, typename ValueType,
```
tcojean, line 138: Also here

thoasm (approved):
I mainly have nits about formatting, otherwise, it looks good to me.

dpcpp/matrix/csr_kernels.dp.cpp
```diff
@@ -71,11 +79,1524 @@ namespace dpcpp {
 namespace csr {
 
 
+constexpr int default_block_size = 256;
+constexpr int warps_in_block = 4;
```
tcojean, line 83: Same comment.

dpcpp/matrix/csr_kernels.dp.cpp
```diff
@@ -71,11 +79,1524 @@ namespace dpcpp {
...
+        }
+
+    } else {
+        *row = num_rows - 1;
+        *row_end = data_size;
+    }
+}
+
+
+template <unsigned subwarp_size, typename ValueType, typename IndexType,
```
tcojean, line 157: Same comment as the other files, in multiple places here.

tcojean (approved):
LGTM.

dpcpp/matrix/csr_kernels.dp.cpp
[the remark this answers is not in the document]
thoasm: I would actually prefer this solution (for smaller kernels). Not just for reducing the number of lines, but also to make reading the kernel easier.

dpcpp/matrix/coo_kernels.dp.cpp
yhmtsai: I think 7 or 8 should be used in oversubscription?

dpcpp/matrix/sparsity_csr_kernels.dp.cpp
yhmtsai: Isn't it confusing between csr and SparsityCsr? they give the same brief

dpcpp/test/matrix/sellp_kernels.cpp
```diff
@@ -0,0 +1,398 @@
...
+TEST_F(Sellp, SimpleApplyIsEquivalentToRef)
+{
+    set_up_apply_matrix();
+
+    mtx->apply(y.get(), expected.get());
+    dmtx->apply(dy.get(), dresult.get());
+
+    auto result = Vec::create(ref);
+    result->copy_from(dresult.get());
+    GKO_ASSERT_MTX_NEAR(result, expected, r<vtype>::value);
```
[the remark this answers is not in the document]
yhmtsai: good catch!

include/ginkgo/core/matrix/csr.hpp
yhmtsai: I found it is in public. is it okay to change it?

include/ginkgo/core/matrix/csr.hpp
```diff
@@ -548,6 +585,10 @@ class Csr : public EnableLinOp<Csr<ValueType, IndexType>>,
             // <row_len_limit>, use load_balance otherwise use classical
             index_type nnz_limit = nvidia_nnz_limit;
             index_type row_len_limit = nvidia_row_len_limit;
+            if (!cuda_strategy_) {
+                nnz_limit = intel_nnz_limit;
+                row_len_limit = intel_row_len_limit;
+            }
 #if GINKGO_HIP_PLATFORM_HCC
             if (!cuda_strategy_) {
                 nnz_limit = amd_nnz_limit;
```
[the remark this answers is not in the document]
yhmtsai: Maybe use a enum?
is cuda_strategy flag already in the previous release?

include/ginkgo/core/matrix/csr.hpp
thoasm: Actually, if we made this `static const`, we would have to instantiate each variable once (so only one instance is created), which is what I forgot when commenting on this issue...
I guess, for now, it is fine keeping it as it is. However, if we don't use `intel_row_len_limit` outside, it might make sense to make it `private` and `static constexpr`.

dpcpp/matrix/csr_kernels.dp.cpp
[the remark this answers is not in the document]
yhmtsai: after revisiting it, if using DEFAULT_HOST, I think they are similar.
when you search col_permute_kernel, only one place has the code.
but this should be moved to common part

tcojean:
rebase!

PR #846, opened 2021-07-23
Work around ILU and IC use-after-free on CUDA 11.4

upsj:
This PR works around an issue in CUDA 11.4 probably caused by the ILU info destruction not waiting for the previous kernels.

### DIFF
--- a/cuda/factorization/ic_kernels.cu
+++ b/cuda/factorization/ic_kernels.cu
@@ -81,6 +81,11 @@ void compute(std::shared_ptr<const DefaultExecutor> exec,
                   m->get_const_row_ptrs(), m->get_const_col_idxs(), info,
                   CUSPARSE_SOLVE_POLICY_USE_LEVEL, buffer.get_data());
 
+    // CUDA 11.4 has a use-after-free bug on Turing
+#if (CUDA_VERSION >= 11040)
+    exec->synchronize();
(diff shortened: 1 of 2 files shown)


yhmtsai (approved):
LGTM. does error message have specific information to point here?
is use-after-free cases on info, desc, or temporary workspace?

upsj:
From the bug report I wrote:
> The following sequence of calls causes an illegal memory access
cusparseCreateMatDescr
cusparseCreateCsric02Info
cusparseDcsric02_bufferSize
cusparseDcsric02_analysis with CUSPARSE_SOLVE_POLICY_USE_LEVEL
cusparseDcsric02 with CUSPARSE_SOLVE_POLICY_USE_LEVEL
cusparseDestroyCsric02Info
cusparseDestroyMatDescr
> 
> If I put a cudaDeviceSynchronize call before cusparseDestroyCsric02Info, the issue goes away, so most likely cusparseDestroyCsric02Info does not wait on the previous kernel to finish before freeing its memory.
The corresponding code worked and produced correct results up to CUDA 11.3

hartwiganzt (approved):

upsj:
The bug was confirmed, though it seems to happen only on Turing, not on Volta

upsj:
rebase!

PR #847, opened 2021-07-27
CuBLAS batched direct solver interface for Ginkgo

Slaedr:
Adds a `BatchDirect` solver class to enable calling batched direct solvers from Ginkgo. Currently only the Cuda backend is enabled using a CuBLAS interface.
In addition,
- preprocessing kernels for the batched iterative solvers have been fused from 2 to 1.
- IDR factory parameter naming is now consistent with other batch solvers.

- [ ] Some of the preprocessing required for getting our data in the layout required by CuBLAS is currently in `core`. Maybe this should be moved to the Cuda backend function. But since all backends right now need the layout transposed anyway, perhaps this remains in core for now.


Slaedr:
format!

PR #848, opened 2021-07-28
Add CSR scaling

fritzgoebel:
This PR adds scaling CSR matrices with a scalar as this is needed by opencarp. The scalar has to be a 1x1 Dense matrix, the usage is the same as for dense matrices.


test/matrix/csr_kernels.cpp
```diff
@@ -0,0 +1,149 @@
...
+#include <ginkgo/core/matrix/dense.hpp>
+/*#include <ginkgo/core/base/array.hpp>
+#include <ginkgo/core/base/math.hpp>
+#include <ginkgo/core/matrix/coo.hpp>
+#include <ginkgo/core/matrix/csr.hpp>
+#include <ginkgo/core/matrix/diagonal.hpp>
+#include <ginkgo/core/matrix/ell.hpp>
+#include <ginkgo/core/matrix/hybrid.hpp>
+#include <ginkgo/core/matrix/sellp.hpp>
+#include <ginkgo/core/matrix/sparsity_csr.hpp>*/
```
upsj, line 54:
```suggestion
```

include/ginkgo/core/matrix/csr.hpp
```diff
@@ -1014,6 +1038,22 @@ class Csr : public EnableLinOp<Csr<ValueType, IndexType>>,
         strategy_->process(row_ptrs_, &srow_);
     }
 
+    /**
+     * @copydoc scale(const LinOp *)
+     *
+     * @note  Other implementations of dense should override this function
+     *        instead of scale(const LinOp *alpha).
```
upsj, line 1045: copypasta :)
```suggestion
     * @note  Other implementations of Csr should override this function
     *        instead of scale(const LinOp *alpha).
```

include/ginkgo/core/matrix/csr.hpp
```diff
@@ -1014,6 +1038,22 @@ class Csr : public EnableLinOp<Csr<ValueType, IndexType>>,
...
+     */
+    virtual void scale_impl(const LinOp *alpha);
+
+    /**
+     * @copydoc inv_scale(const LinOp *)
+     *
+     * @note  Other implementations of dense should override this function
+     *        instead of inv_scale(const LinOp *alpha).
```
upsj, line 1053:
```suggestion
     * @note  Other implementations of Csr should override this function
     *        instead of inv_scale(const LinOp *alpha).
```

include/ginkgo/core/matrix/csr.hpp
```diff
@@ -848,6 +848,30 @@ class Csr : public EnableLinOp<Csr<ValueType, IndexType>>,
...
 
+    /**
+     * Scales the matrix with a scalar.
+     *
+     * @param alpha  The entire matrix is scaled by alpha. alpha has to be a 1x1
+     * Dense matrix.
+     */
+    void scale(const LinOp *alpha)
+    {
+        auto exec = this->get_executor();
```
upsj, line 859: I think you can pull the 1x1 assertions out here as well, since this should hold also for all potential subclasses (even though deriving from Csr might not be the best idea, but still).

upsj (commented):
LGTM from the implementation standpoint.
On the design level, I noticed we have a few different ways to do things:
* Some operations have both member functions as well as LinOp apply representations (*_permute vs. Permutation->apply(...))
* Some others have only member functions representing certain operations (Dense::scale could be Scalar->apply(Dense, Dense) or Diagonal->apply(Dense, Dense))
* Finally some can only be represented via LinOp interfaces (SpGEMM, diagonal row scaling for Dense or any diagonal scaling for Csr)

So maybe we can investigate this design space some more before merging this interface? (Best after 1.4.0?)

common/matrix/csr_kernels.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+
+template <typename ValueType, typename IndexType>
+void inv_scale(std::shared_ptr<const DefaultExecutor> exec,
+               const matrix::Dense<ValueType> *alpha,
+               matrix::Csr<ValueType, IndexType> *x)
+{
+    run_kernel(
+        exec,
+        [] GKO_KERNEL(auto nnz, auto alpha, auto x) { x[nnz] /= alpha[0]; },
```
MarcelKoch, line 75: the inverse `inv=1/alpha[0]` should probably be precomputed and used as `x[nnz] *= inv`. The compiler might do that automatically, but I think here it is very easy to help the compiler with that.

core/matrix/csr.cpp
```diff
@@ -577,6 +579,28 @@ Csr<ValueType, IndexType>::compute_absolute() const
 }
 
 
+template <typename ValueType, typename IndexType>
+void Csr<ValueType, IndexType>::scale_impl(const LinOp *alpha)
+{
+    GKO_ASSERT_EQUAL_ROWS(alpha, dim<2>(1, 1));
+    GKO_ASSERT_EQUAL_COLS(alpha, dim<2>(1, 1));
```
MarcelKoch, line 586: you could use
```suggestion
    GKO_ASSERT_EQUAL_DIMENSIONS(alpha, dim<2>(1, 1));
```

core/matrix/csr.cpp
```diff
@@ -577,6 +579,28 @@ Csr<ValueType, IndexType>::compute_absolute() const
...
+    exec->run(csr::make_scale(make_temporary_conversion<ValueType>(alpha).get(),
+                              this));
+}
+
+
+template <typename ValueType, typename IndexType>
+void Csr<ValueType, IndexType>::inv_scale_impl(const LinOp *alpha)
+{
+    GKO_ASSERT_EQUAL_ROWS(alpha, dim<2>(1, 1));
+    GKO_ASSERT_EQUAL_COLS(alpha, dim<2>(1, 1));
```
MarcelKoch, line 597: also here
```suggestion
    GKO_ASSERT_EQUAL_DIMENSIONS(alpha, dim<2>(1, 1));
```

test/matrix/CMakeLists.txt
```diff
@@ -1 +1,2 @@
+ginkgo_create_common_test(csr_kernels)
 ginkgo_create_common_test(dense_kernels)
```
MarcelKoch, line 2: nit: I don't think the current contributing guidelines contain something on a new line at the end of a file, but I think it would be good to use it by default
```suggestion
ginkgo_create_common_test(dense_kernels)

```

test/matrix/csr_kernels.cpp
```diff
@@ -0,0 +1,149 @@
...
+    x->inv_scale(alpha.get());
+    dx->inv_scale(dalpha.get());
+
+    auto result = Mtx::create(ref);
+    result->copy_from(dx.get());
+    GKO_ASSERT_MTX_NEAR(result, x, r<vtype>::value);
+}
+
+
+}  //  namespace
```
MarcelKoch, line 149: nit: again perhaps a new line?
```suggestion
}  //  namespace

```

MarcelKoch (approved):
lgtm, with some minor suggestions.

BTW, in what way does opencarp require this? I'm guessing using the advanced apply with the included scaling is not enough for your case.

Also, if that is merged before #820, I will add an overload accepting a value_type later on, since this seems like a suitable place for that.

test/matrix/CMakeLists.txt
yhmtsai: I think it is also be handled by editorconfig

yhmtsai (commented):
LGTM. I also need to bring the question again.
What should we use the generic kernel for?

upsj:
@yhmtsai You bring up an important question. The crucial points for me were
* Pointwise and regular memory access
* 1D or 2D indexing
* simple local operations

IMO this fits the model quite well.

fritzgoebel:
> lgtm, with some minor suggestions.
> 
> BTW, in what way does opencarp require this? I'm guessing using the advanced apply with the included scaling is not enough for your case.
> 
> Also, if that is merged before #820, I will add an overload accepting a value_type later on, since this seems like a suitable place for that.

In some points in opencarp, an already assembled matrix is scaled by a constant and stored again. I guess in theory, it would be possible to use the advanced apply but this would mean restructuring this part of opencarp. I'd rather have this simple scaling available tbh.

core/matrix/csr_kernels.hpp
```diff
@@ -185,6 +185,16 @@ namespace kernels {
                           const matrix::Csr<ValueType, IndexType> *orig, \
                           matrix::Diagonal<ValueType> *diag)
 
+#define GKO_DECLARE_CSR_SCALE_KERNEL(ValueType, IndexType)  \
+    void scale(std::shared_ptr<const DefaultExecutor> exec, \
+               const matrix::Dense<ValueType> *alpha,       \
+               matrix::Csr<ValueType, IndexType> *to_scale)
+
+#define GKO_DECLARE_CSR_INV_SCALE_KERNEL(ValueType, IndexType)  \
+    void inv_scale(std::shared_ptr<const DefaultExecutor> exec, \
```
Slaedr, line 194: Do we really need a separate `inv_scale` operation? Isn't it easily done at the calling site by storing the inverse in another `LinOp` and calling `scale`? Or do you need both operations in the same code path? If that is the case, then this makes sense because we don't want to send another number over PCIe or something just for that. If you do not need both operations in the same run, I'd prefer removing `inv_scale` for now and only adding it when the need arises.

include/ginkgo/core/matrix/csr.hpp
Slaedr: Does it make sense to have a `Scalable` mix-in for this?

Slaedr (commented):
Looks good, mostly. My main issue is related to the need for `inv_scale` right now, please see below.

core/matrix/csr_kernels.hpp
yhmtsai: If consider dense functionalities, which contains the opposite operation, it might be okay at this moment.
it will depends on how the scalar generate, if it is always generated by user (`alpha = initialize ...`), it should be fine with only one scale function. otherwise, the scale will requires additional inverse kernel to behave the inv_scale

test/matrix/csr_kernels.cpp
```diff
@@ -0,0 +1,140 @@
...
+TEST_F(Csr, ScaleIsEquivalentToRef)
+{
+    set_up_apply_data();
+
+    x->scale(alpha.get());
+    dx->scale(dalpha.get());
+
+    auto result = Mtx::create(ref);
+    result->copy_from(dx.get());
+    GKO_ASSERT_MTX_NEAR(result, x, r<vtype>::value);
```
yhmtsai, line 123:
```suggestion
    GKO_ASSERT_MTX_NEAR(result, x, r<vtype>::value);
```
`GKO_ASSERT_MTX_NEAR` should be able to take care of the memory

test/matrix/csr_kernels.cpp
```diff
@@ -0,0 +1,140 @@
...
+TEST_F(Csr, InvScaleIsEquivalentToRef)
+{
+    set_up_apply_data();
+
+    x->inv_scale(alpha.get());
+    dx->inv_scale(dalpha.get());
+
+    auto result = Mtx::create(ref);
+    result->copy_from(dx.get());
+    GKO_ASSERT_MTX_NEAR(result, x, r<vtype>::value);
```
yhmtsai, line 136:
```suggestion
    GKO_ASSERT_MTX_NEAR(result, x, r<vtype>::value);
```

yhmtsai (approved):
LGTM. some nit

fritzgoebel:
format!

PR #849, opened 2021-07-28
Dpcpp ports IDR

yhmtsai:
This PR adds IDR support
Note. there is a weird barrier in IDR kernel. I do not think it need the barrier but it is always failed when I add barrier in other place without it. The issue is only happened in CPU

TODO:
- [x] use oneDPL random number generator to generate complex
- [x] ~~refine atomic_max such that be similar impl as cuda?~~ we need the memoryspace template, so we can not do similar with cuda

### DIFF
--- a/common/solver/idr_kernels.hpp.inc
+++ b/common/solver/idr_kernels.hpp.inc
@@ -59,9 +59,10 @@ __global__
     const auto tidx = thread::get_thread_id_flat();
 
     __shared__ UninitializedArray<ValueType, block_size> reduction_helper_array;
-    ValueType *__restrict__ reduction_helper = reduction_helper_array;
-
-    __shared__ remove_complex<ValueType> reduction_helper_real[block_size];
+    // they are not be used in the same time.
+    ValueType *reduction_helper = reduction_helper_array;
+    auto reduction_helper_real =
+        reinterpret_cast<remove_complex<ValueType> *>(reduction_helper);
 
     for (size_type row = 0; row < num_rows; row++) {
         for (size_type i = 0; i < row; i++) {
--- a/common/solver/idr_kernels.hpp.inc
+++ b/common/solver/idr_kernels.hpp.inc
@@ -71,9 +71,9 @@ __global__
                 dot += values[row * stride + j] * conj(values[i * stride + j]);
             }
 
-            reduction_helper[tidx] = dot;
-
+            // Ensure already finish reading this shared memory
             __syncthreads();
+            reduction_helper[tidx] = dot;
             reduce(
                 group::this_thread_block(), reduction_helper,
                 [](const ValueType &a, const ValueType &b) { return a + b; });
@@ -89,10 +89,9 @@ __global__
         for (size_type j = tidx; j < num_cols; j += block_size) {
             norm += squared_norm(values[row * stride + j]);
         }
-
-        reduction_helper_real[tidx] = norm;
-
+        // Ensure already finish reading this shared memory
         __syncthreads();
+        reduction_helper_real[tidx] = norm;
         reduce(group::this_thread_block(), reduction_helper_real,
                [](const remove_complex<ValueType> &a,
                   const remove_complex<ValueType> &b) { return a + b; });
--- a/dpcpp/CMakeLists.txt
+++ b/dpcpp/CMakeLists.txt
@@ -8,6 +8,8 @@ set(GINKGO_DPCPP_VERSION ${GINKGO_DPCPP_VERSION} PARENT_SCOPE)
 
 find_package(MKL CONFIG REQUIRED HINTS "$ENV{MKLROOT}")
 set(GINKGO_MKL_ROOT "${MKL_ROOT}" PARENT_SCOPE)
+find_package(oneDPL REQUIRED HINTS "$ENV{DPL_ROOT}")
+set(GINKGO_DPL_ROOT "${DPL_ROOT}" PARENT_SCOPE)
 
 add_library(ginkgo_dpcpp $<TARGET_OBJECTS:ginkgo_dpcpp_device> "")
 target_sources(ginkgo_dpcpp
@@ -75,7 +77,7 @@ else ()
     target_link_options(ginkgo_dpcpp PUBLIC -fsycl-device-code-split=per_kernel)
 endif()
 target_link_libraries(ginkgo_dpcpp PUBLIC ginkgo_device)
-target_link_libraries(ginkgo_dpcpp PRIVATE MKL::MKL_DPCPP)
+target_link_libraries(ginkgo_dpcpp PRIVATE MKL::MKL_DPCPP oneDPL)
 if (GINKGO_DPCPP_SINGLE_MODE)
     target_compile_definitions(ginkgo_dpcpp PRIVATE GINKGO_DPCPP_SINGLE_MODE=1)
 endif()
--- a/dpcpp/solver/idr_kernels.dp.cpp
+++ b/dpcpp/solver/idr_kernels.dp.cpp
@@ -130,7 +130,8 @@ void orthonormalize_subspace_vectors_kernel(
                 */
                 dot += values[row * stride + j] * conj(values[i * stride + j]);
             }
-
+            // TODO: check with intel why we need this here.
+            item_ct1.barrier();
             reduction_helper[tidx] = dot;
 
             /*
@@ -595,8 +596,8 @@ void compute_omega_kernel(
     if (!stop_status[global_id].has_stopped()) {
         auto thr = omega[global_id];
         omega[global_id] /= tht[global_id];
-        auto absrho = std::abs(thr / (std::sqrt((float)(real(tht[global_id]))) *
-                                      residual_norm[global_id]));
+        auto absrho = std::abs(
+            thr / (std::sqrt(real(tht[global_id])) * residual_norm[global_id]));
 
         if (absrho < kappa) {
             omega[global_id] *= kappa / absrho;
--- a/dpcpp/solver/idr_kernels.dp.cpp
+++ b/dpcpp/solver/idr_kernels.dp.cpp
@@ -33,15 +33,27 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "core/solver/idr_kernels.hpp"
 
 
-#include <memory>
+#include <ctime>
+#include <dpcpp/base/curand_bindings.hpp>
+#include <dpcpp/base/math.hpp>
+#include <random>
 
 
-#include <ginkgo/core/base/array.hpp>
+#include <CL/sycl.hpp>
+
+
 #include <ginkgo/core/base/exception_helpers.hpp>
-#include <ginkgo/core/base/executor.hpp>
 #include <ginkgo/core/base/math.hpp>
-#include <ginkgo/core/matrix/dense.hpp>
-#include <ginkgo/core/stop/stopping_status.hpp>
+
+
+#include "core/components/fill_array.hpp"
+#include "dpcpp/base/config.hpp"
+#include "dpcpp/base/dim3.dp.hpp"
+#include "dpcpp/base/onemkl_bindings.hpp"
+#include "dpcpp/components/atomic.dp.hpp"
+#include "dpcpp/components/cooperative_groups.dp.hpp"
+#include "dpcpp/components/reduction.dp.hpp"
+#include "dpcpp/components/thread_ids.dp.hpp"
 
 
 namespace gko {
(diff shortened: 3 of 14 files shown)


dpcpp/CMakeLists.txt
```diff
@@ -8,6 +8,8 @@ set(GINKGO_DPCPP_VERSION ${GINKGO_DPCPP_VERSION} PARENT_SCOPE)
 
 find_package(MKL CONFIG REQUIRED HINTS "$ENV{MKLROOT}")
 set(GINKGO_MKL_ROOT "${MKL_ROOT}" PARENT_SCOPE)
+find_package(oneDPL REQUIRED HINTS "$ENV{DPL_ROOT}")
```
tcojean, line 11: Should it be like MKL?
```suggestion
find_package(oneDPL CONFIG REQUIRED HINTS "$ENV{DPL_ROOT}")
```

dpcpp/test/solver/idr_kernels.cpp
```diff
@@ -0,0 +1,414 @@
...
+    int m = 123;
+    int n = 1;
+    auto ref_solver = ref_idr_factory->generate(mtx);
+    auto dpcpp_solver = dpcpp_idr_factory->generate(d_mtx);
+    auto b = gen_mtx(m, n);
+    auto x = gen_mtx(m, n);
+    auto d_b = Mtx::create(dpcpp);
+    auto d_x = Mtx::create(dpcpp);
+    d_b->copy_from(b.get());
+    d_x->copy_from(x.get());
```
tcojean, line 313: Minor: maybe we could make a function for this since it's used in many tests.

dpcpp/solver/idr_kernels.dp.cpp
```diff
@@ -55,11 +66,760 @@ namespace dpcpp {
...
+
+    for (size_type row = 0; row < num_rows; row++) {
+        for (size_type i = 0; i < row; i++) {
+            auto dot = zero<ValueType>();
+            for (size_type j = tidx; j < num_cols; j += block_size) {
+                /*
+                DPCT1007:5: Migration of this CUDA API is not supported by the
+                Intel(R) DPC++ Compatibility Tool.
+                */
+                dot += values[row * stride + j] * conj(values[i * stride + j]);
```
tcojean, line 131: DPCT issue

dpcpp/solver/idr_kernels.dp.cpp
```diff
@@ -55,11 +66,760 @@ namespace dpcpp {
...
+            */
+            item_ct1.barrier();
+            ::gko::kernels::dpcpp::reduce(
+                group::this_thread_block(item_ct1), reduction_helper,
+                [](const ValueType &a, const ValueType &b) { return a + b; });
+            /*
+            DPCT1065:4: Consider replacing sycl::nd_item::barrier() with
+            sycl::nd_item::barrier(sycl::access::fence_space::local_space) for
+            better performance, if there is no access to global memory.
+            */
```
tcojean, line 150: DPCT issue. Since we use the `local` accessor, I think the suggestion might make sense?

dpcpp/solver/idr_kernels.dp.cpp
```diff
@@ -55,11 +66,760 @@ namespace dpcpp {
...
+        ::gko::kernels::dpcpp::reduce(
+            group::this_thread_block(item_ct1), reduction_helper_real,
+            [](const remove_complex<ValueType> &a,
+               const remove_complex<ValueType> &b) { return a + b; });
+        /*
+        DPCT1065:2: Consider replacing sycl::nd_item::barrier() with
+        sycl::nd_item::barrier(sycl::access::fence_space::local_space) for
+        better performance, if there is no access to global memory.
+        */
+        item_ct1.barrier();
```
tcojean, line 181: Same thing.

dpcpp/solver/idr_kernels.dp.cpp
```diff
@@ -55,11 +66,760 @@ namespace dpcpp {
...
+    const auto rhs = item_ct1.get_group(2) * default_dot_dim + tidx;
+    const auto num = ceildiv(num_rows, item_ct1.get_group_range(1));
+    const auto start_row = item_ct1.get_group(1) * num;
+    const auto end_row = ((item_ct1.get_group(1) + 1) * num > num_rows)
+                             ? num_rows
+                             : (item_ct1.get_group(1) + 1) * num;
+    // Used that way to get around dynamic initialization warning and
+    // template error when using `reduction_helper_array` directly in `reduce`
+
+    ValueType *__restrict__ reduction_helper = (*reduction_helper_array);
```
tcojean, line 381: Formatting.
```suggestion

    // Used that way to get around dynamic initialization warning and
    // template error when using `reduction_helper_array` directly in `reduce`
    ValueType *__restrict__ reduction_helper = (*reduction_helper_array);
```

dpcpp/solver/idr_kernels.dp.cpp
```diff
@@ -55,11 +66,760 @@ namespace dpcpp {
...
+            local_res += p_i[i] * g_k[g_idx];
+        }
+    }
+    reduction_helper[tidx * (default_dot_dim + 1) + tidy] = local_res;
+    /*
+    DPCT1065:10: Consider replacing sycl::nd_item::barrier() with
+    sycl::nd_item::barrier(sycl::access::fence_space::local_space) for better
+    performance, if there is no access to global memory.
+    */
+    item_ct1.barrier();
```
tcojean, line 397: Same thing.

dpcpp/solver/idr_kernels.dp.cpp
```diff
@@ -55,11 +66,760 @@ namespace dpcpp {
...
+        if (nrhs > 1 || is_complex<ValueType>()) {
+            components::fill_array(exec, alpha->get_values(), nrhs,
+                                   zero<ValueType>());
+            multidot_kernel(grid_dim, block_dim, 0, exec->get_queue(), size,
+                            nrhs, p_i, g_k->get_values(), g_k->get_stride(),
+                            alpha->get_values(), stop_status->get_const_data());
+        } else {
+            onemkl::dot(*exec->get_queue(), size, p_i, 1, g_k->get_values(),
+                        g_k->get_stride(), alpha->get_values());
+        }
```
tcojean, line 735: Is the oneMKL dot better on one RHS as well like in CUDA? Maybe our implementation is better?

dpcpp/solver/idr_kernels.dp.cpp
```diff
@@ -55,11 +66,760 @@ namespace dpcpp {
...
+        if (nrhs > 1 || is_complex<ValueType>()) {
+            components::fill_array(exec, m_i, nrhs, zero<ValueType>());
+            multidot_kernel(grid_dim, block_dim, 0, exec->get_queue(), size,
+                            nrhs, p_i, g_k->get_const_values(),
+                            g_k->get_stride(), m_i,
+                            stop_status->get_const_data());
+        } else {
+            onemkl::dot(*exec->get_queue(), size, p_i, 1,
+                        g_k->get_const_values(), g_k->get_stride(), m_i);
+        }
```
tcojean, line 779: Same question.

tcojean (approved):
LGTM!

upsj (approved):
LGTM! I actually hadn't looked at the IDR kernels yet. Most of them look pretty well-suited for common kernels.

dpcpp/CMakeLists.txt
yhmtsai: I use the example from https://github.com/oneapi-src/oneDPL/tree/main/cmake#how-to-use-onedpl-package-from-cmake

dpcpp/test/solver/idr_kernels.cpp
yhmtsai: I guess the matrix setup like set_up_apply_mtx()?

dpcpp/solver/idr_kernels.dp.cpp
yhmtsai: I do not test it acutually

dpcpp/solver/idr_kernels.dp.cpp
```diff
@@ -55,11 +66,760 @@ namespace dpcpp {
...
+            auto dot = zero<ValueType>();
+            for (size_type j = tidx; j < num_cols; j += block_size) {
+                /*
+                DPCT1007:5: Migration of this CUDA API is not supported by the
+                Intel(R) DPC++ Compatibility Tool.
+                */
+                dot += values[row * stride + j] * conj(values[i * stride + j]);
+            }
+            // TODO: check with intel why we need this here.
+            item_ct1.barrier();
```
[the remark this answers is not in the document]
yhmtsai: I also think it should not need this one, but it is the place I can make it correct on cpu.
I do not have time to investigate it yet

dpcpp/solver/idr_kernels.dp.cpp
```diff
@@ -55,11 +66,760 @@ namespace dpcpp {
...
+
+
+template <typename ValueType>
+void multidot_kernel(
+    size_type num_rows, size_type nrhs, const ValueType *__restrict__ p_i,
+    const ValueType *__restrict__ g_k, size_type g_k_stride,
+    ValueType *__restrict__ alpha,
+    const stopping_status *__restrict__ stop_status, sycl::nd_item<3> item_ct1,
+    UninitializedArray<ValueType, default_dot_dim *(default_dot_dim + 1)>
+        *reduction_helper_array)
```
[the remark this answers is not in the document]
yhmtsai: I think it is the pointer not reference?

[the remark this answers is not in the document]
upsj: If you passed it by reference, you wouldn't need to dereference it later.

dpcpp/solver/idr_kernels.dp.cpp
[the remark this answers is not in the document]
yhmtsai: I am thinking it requires the global space sync because we update the value in global space later and use it in next loop.
Originally, I thought it always accessed by the same thread (with same tid), so it should be okay even it is not visible from global. However, it may not be true in CPU?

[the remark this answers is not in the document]
upsj: We have two other barriers before we access `values `again, and the column is being used exclusively by that thread. Does the code break if you remove this synchronization? If so, that sounds like a compiler issue.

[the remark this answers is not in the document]
yhmtsai: Yes, it will break if I remove the barrier.
I hope it's a compile issue. In my understanding, the value should be update because it is on the same thread.
It is only happened in cpu and we also need this weird kind barrier in coo or csr, too.

yhmtsai:
@upsj I will not move some of them into common in this pr because some kernel in the function can not be in common.
For example, `initialize` contain `initialize_m` and `initialize_subspace_vectors` (able to common), but `orthonormalize_subspace_vectors` not. I think I can not put individual kernel into common?

upsj:
@yhmtsai I see, that's fine by me then :)

dpcpp/CMakeLists.txt
upsj: They are not necessarily consistent inside Intel's documentation, and since there is no FindoneDPL.cmake module inside CMake (and I don't think there will be in the forseeable future), it doesn't really matter, except for slightly changing the error message.

dpcpp/solver/idr_kernels.dp.cpp
```diff
@@ -55,11 +66,760 @@ namespace dpcpp {
...
+{
+    stream->submit([&](sycl::handler &cgh) {
+        sycl::accessor<UninitializedArray<ValueType, block_size>, 0,
+                       sycl::access_mode::read_write,
+                       sycl::access::target::local>
+            reduction_helper_array_acc_ct1(cgh);
+        sycl::accessor<remove_complex<ValueType>, 1,
+                       sycl::access_mode::read_write,
+                       sycl::access::target::local>
+            reduction_helper_real_acc_ct1(sycl::range<1>(block_size), cgh);
```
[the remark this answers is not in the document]
yhmtsai: I guess I need to add sync before reinterpret-cast?
It may not be needed due to data dependency on values though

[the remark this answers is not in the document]
upsj: We only need a barrier between the dot and norm computation, or make sure that we don't accidentally claim they are non-aliasing with `__restrict__`

dpcpp/solver/idr_kernels.dp.cpp
```diff
@@ -55,11 +66,727 @@ namespace dpcpp {
...
+template <size_type block_size, typename ValueType>
+void orthonormalize_subspace_vectors_kernel(
+    size_type num_rows, size_type num_cols, ValueType *__restrict__ values,
+    size_type stride, sycl::nd_item<3> item_ct1,
+    UninitializedArray<ValueType, block_size> &reduction_helper_array,
+    remove_complex<ValueType> *reduction_helper_real)
+{
+    const auto tidx = thread::get_thread_id_flat(item_ct1);
+
+    ValueType *__restrict__ reduction_helper = reduction_helper_array;
```
yhmtsai, line 119: @upsj is it what you means?
also the calling, 
```
cgh.parallel_for(
            sycl_nd_range(grid, block), [=](sycl::nd_item<3> item_ct1) {
                orthonormalize_subspace_vectors_kernel<block_size>(
                    num_rows, num_cols, values, stride, item_ct1,
                    *reduction_helper_array_acc_ct1.get_pointer(),
                    reduction_helper_real_acc_ct1.get_pointer().get());
            });
```
but I am a little unsure.
Is there any reason such that dpct do not use this way?

upsj: If you pass only the complex reduction helper, you can use
```
ValueType *__restrict__ reduction_helper = reduction_helper_array;
remove_complex<ValueType>* real_reduction_helper = reinterpret_cast<remove_complex<ValueType>*>(reduction_helper);
```
That should be fine.

yhmtsai: I mean the reference part

upsj: yes

yhmtsai: does real_reduction_helper not lead issue with reduction_helper?
reduction_helper thinks no one overlap on it, but real_reduction_helper does.

upsj: That's why it's important to set the `__restrict__` on the first variable, but not the second! The compiler can't assume that pointers derived from the same pointer don't overlap.

yhmtsai: the second is overlapped on the first, so I thought it destroys the first `__restrict__` property?
Might the first be arranged after the second (if we do not consider other dependency) because the first is `__restrict__`?

tcojean:
rebase!

yhmtsai:
rebase!

common/solver/idr_kernels.hpp.inc
```diff
@@ -79,6 +80,8 @@ __global__
             __syncthreads();
 
             dot = reduction_helper[0];
+            // avoid the first thread writing it before every threads read it.
+            __syncthreads();
```
upsj, line 84: I think we can remove this and move the first __syncthreads() before `reduction_helper[tidx] = dot`, because `reduce` already synchronizes before accessing memory. Not sure whether the CUDA compiler removes duplicate barrier.sync instructions.

yhmtsai: I guess you mean removing the second one?
I will need to add another one before `reduction_helper_real[tidx] = norm;` to avoid `reduction_helper_real[tidx]` issue because they are on the same memory.

upsj: I mean
```cpp
for (...) {
__syncthreads();
reduction_helper[tidx] = dot;
reduce(...);
__syncthreads();
dot = reduction_helper[0];
}
...
__syncthreads();
reduction_helper_real[tidx] = norm;
reduce(...);
__syncthreads();
norm = reduction_helper_real[0];
```

yhmtsai: I see. I am fine with both version

PR #850, opened 2021-08-03
Fix the residual norm test baseline

yhmtsai:
This PR fixes the baseline usage of residual norm.
1. one uses `res_norm` to set `res_norm` with `rhs_norm` baseline
2. use the updated res to set. `first = 1.1 * r * baseline, second = 0.9 * r * first` . It gives `0.99 r^2 baseline`, so I guess it is not expected originally?

It should be mentioned in the self-res-computation PR, but I only put it on list and forget it.


upsj (approved):
LGTM! So this only worked accidentally before?

yhmtsai:
the first gives the same result because res_norm and rhs_norm are the identical in the beginning.
the second gives much smaller than the criterion setting

pratikvn (approved):
LGTM!

tcojean (approved):

yhmtsai:
rebase!

PR #851, opened 2021-08-04
Move internal constexpr symbols to internal linkage

upsj:
constexpr configuration variables in our GPU implementations are only used inside the same translation unit, but still emit symbols which might cause some issues if we changed the value for a single definition of this value (more precisely, ODR violations). Just to be safe, this PR sets all of them to internal linkage.
For more details: https://en.cppreference.com/w/cpp/language/storage_duration

### DIFF
--- a/common/factorization/par_ilut_select_kernels.hpp.inc
+++ b/common/factorization/par_ilut_select_kernels.hpp.inc
@@ -33,15 +33,17 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 namespace kernel {
 
 
-constexpr auto searchtree_width = 1 << sampleselect_searchtree_height;
-constexpr auto searchtree_inner_size = searchtree_width - 1;
-constexpr auto searchtree_size = searchtree_width + searchtree_inner_size;
+static constexpr auto searchtree_width = 1 << sampleselect_searchtree_height;
+static constexpr auto searchtree_inner_size = searchtree_width - 1;
+static constexpr auto searchtree_size =
+    searchtree_width + searchtree_inner_size;
 
-constexpr auto sample_size = searchtree_width * sampleselect_oversampling;
+static constexpr auto sample_size =
+    searchtree_width * sampleselect_oversampling;
 
-constexpr auto basecase_size = 1024;
-constexpr auto basecase_local_size = 4;
-constexpr auto basecase_block_size = basecase_size / basecase_local_size;
+static constexpr auto basecase_size = 1024;
+static constexpr auto basecase_local_size = 4;
+static constexpr auto basecase_block_size = basecase_size / basecase_local_size;
 
 
 // must be launched with one thread block and block size == searchtree_width
--- a/cuda/matrix/sellp_kernels.cu
+++ b/cuda/matrix/sellp_kernels.cu
@@ -59,7 +59,7 @@ namespace cuda {
 namespace sellp {
 
 
-constexpr auto default_block_size = 512;
+static constexpr auto default_block_size = 512;
 
 
 #include "common/matrix/sellp_kernels.hpp.inc"
--- a/dpcpp/matrix/dense_kernels.dp.cpp
+++ b/dpcpp/matrix/dense_kernels.dp.cpp
@@ -70,15 +70,15 @@ namespace dense {
 
 
 using KCFG_1D = ConfigSet<11, 7>;
-constexpr auto kcfg_1d_list =
+static constexpr auto kcfg_1d_list =
     syn::value_list<std::uint32_t, KCFG_1D::encode(512, 64),
                     KCFG_1D::encode(512, 32), KCFG_1D::encode(512, 16),
                     KCFG_1D::encode(256, 32), KCFG_1D::encode(256, 16),
                     KCFG_1D::encode(256, 8)>();
-constexpr auto subgroup_list =
+static constexpr auto subgroup_list =
     syn::value_list<std::uint32_t, 64, 32, 16, 8, 4>();
-constexpr auto kcfg_1d_array = syn::as_array(kcfg_1d_list);
-constexpr auto default_block_size = 256;
+static constexpr auto kcfg_1d_array = syn::as_array(kcfg_1d_list);
+static constexpr auto default_block_size = 256;
 
 
 namespace kernel {
--- a/hip/factorization/par_ic_kernels.hip.cpp
+++ b/hip/factorization/par_ic_kernels.hip.cpp
@@ -54,7 +54,7 @@ namespace hip {
 namespace par_ic_factorization {
 
 
-constexpr auto default_block_size = 512;
+static constexpr auto default_block_size = 512;
 
 
 // subwarp sizes for all warp-parallel kernels (sweep)
--- a/hip/factorization/par_ict_kernels.hip.cpp
+++ b/hip/factorization/par_ict_kernels.hip.cpp
@@ -68,7 +68,7 @@ namespace hip {
 namespace par_ict_factorization {
 
 
-constexpr auto default_block_size = 512;
+static constexpr auto default_block_size = 512;
 
 
 // subwarp sizes for all warp-parallel kernels (filter, add_candidates)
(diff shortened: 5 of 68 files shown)


common/factorization/par_ilut_select_kernels.hpp.inc
```diff
@@ -33,15 +33,17 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 namespace kernel {
 
 
-constexpr auto searchtree_width = 1 << sampleselect_searchtree_height;
-constexpr auto searchtree_inner_size = searchtree_width - 1;
-constexpr auto searchtree_size = searchtree_width + searchtree_inner_size;
+static constexpr auto searchtree_width = 1 << sampleselect_searchtree_height;
+static constexpr auto searchtree_inner_size = searchtree_width - 1;
```
yhmtsai, line 37: I also did some changes on these from dpcpp matrix review.
use `constexpr int` not `constexpr auto` if it is initialized by value.
Could you also do it here?

yhmtsai (approved):
LGTM. Could you describe more detail about the possible issue?
the original will have symbol in global and it might use the wrong one if some file define it?

upsj:
Sure, I'll integrate the changes, I guess your PR will be merged first anyways?

The issue is that when they are in the same namespace, all of these declarations describe the same variable, so if some of them have different values, and the constexpr values were not constant-folded inside some parts of the code, we could be mixing different values for the same constant. This is most likely not an issue, since constexpr values are usually inserted directly, but IMO it's still good practice to ensure they don't get exported if we have them in multiple places.

yhmtsai:
I think the matrix pr will be tomorrow, so I guess this can be merged quickly if it is reviewed.
I will also add the static in coming dpcpp part.

cuda/matrix/sellp_kernels.cu
```diff
@@ -59,7 +59,7 @@ namespace cuda {
 namespace sellp {
 
 
-constexpr auto default_block_size = 512;
+static constexpr auto default_block_size = 512;
```
thoasm, line 62:
```suggestion
static constexpr int default_block_size = 512;
```

dpcpp/matrix/dense_kernels.dp.cpp
```diff
@@ -70,15 +70,15 @@ namespace dense {
...
                     KCFG_1D::encode(512, 32), KCFG_1D::encode(512, 16),
                     KCFG_1D::encode(256, 32), KCFG_1D::encode(256, 16),
                     KCFG_1D::encode(256, 8)>();
-constexpr auto subgroup_list =
+static constexpr auto subgroup_list =
     syn::value_list<std::uint32_t, 64, 32, 16, 8, 4>();
-constexpr auto kcfg_1d_array = syn::as_array(kcfg_1d_list);
-constexpr auto default_block_size = 256;
+static constexpr auto kcfg_1d_array = syn::as_array(kcfg_1d_list);
+static constexpr auto default_block_size = 256;
```
thoasm, line 81:
```suggestion
static constexpr int default_block_size = 256;
```

hip/factorization/par_ic_kernels.hip.cpp
```diff
@@ -54,7 +54,7 @@ namespace hip {
 namespace par_ic_factorization {
 
 
-constexpr auto default_block_size = 512;
+static constexpr auto default_block_size = 512;
```
thoasm, line 57:
```suggestion
static constexpr int default_block_size = 512;
```

hip/factorization/par_ict_kernels.hip.cpp
```diff
@@ -68,7 +68,7 @@ namespace hip {
 namespace par_ict_factorization {
 
 
-constexpr auto default_block_size = 512;
+static constexpr auto default_block_size = 512;
```
thoasm, line 71:
```suggestion
static constexpr int default_block_size = 512;
```

hip/factorization/par_ilut_filter_kernel.hip.cpp
```diff
@@ -67,7 +67,7 @@ namespace hip {
 namespace par_ilut_factorization {
 
 
-constexpr auto default_block_size = 512;
+static constexpr auto default_block_size = 512;
```
thoasm, line 70:
```suggestion
static constexpr int default_block_size = 512;
```

hip/factorization/par_ilut_select_common.hip.hpp
```diff
@@ -45,8 +45,8 @@ namespace hip {
 namespace par_ilut_factorization {
 
 
-constexpr auto default_block_size = 512;
-constexpr auto items_per_thread = 16;
+static constexpr auto default_block_size = 512;
+static constexpr auto items_per_thread = 16;
```
thoasm, line 49:
```suggestion
static constexpr int default_block_size = 512;
static constexpr int items_per_thread = 16;
```

hip/factorization/par_ilut_spgeam_kernel.hip.cpp
```diff
@@ -68,7 +68,7 @@ namespace hip {
 namespace par_ilut_factorization {
 
 
-constexpr auto default_block_size = 512;
+static constexpr auto default_block_size = 512;
```
thoasm, line 71:
```suggestion
static constexpr int default_block_size = 512;
```

hip/factorization/par_ilut_sweep_kernel.hip.cpp
```diff
@@ -68,7 +68,7 @@ namespace hip {
 namespace par_ilut_factorization {
 
 
-constexpr auto default_block_size = 512;
+static constexpr auto default_block_size = 512;
```
thoasm, line 71:
```suggestion
static constexpr int default_block_size = 512;
```

hip/matrix/dense_kernels.hip.cpp
```diff
@@ -67,7 +67,7 @@ namespace hip {
 namespace dense {
 
 
-constexpr auto default_block_size = 512;
+static constexpr auto default_block_size = 512;
```
thoasm, line 70:
```suggestion
static constexpr int default_block_size = 512;
```

hip/matrix/diagonal_kernels.hip.cpp
```diff
@@ -57,7 +57,7 @@ namespace hip {
 namespace diagonal {
 
 
-constexpr auto default_block_size = 512;
+static constexpr auto default_block_size = 512;
```
thoasm, line 60:
```suggestion
static constexpr int default_block_size = 512;
```

hip/matrix/sellp_kernels.hip.cpp
```diff
@@ -62,7 +62,7 @@ namespace hip {
 namespace sellp {
 
 
-constexpr auto default_block_size = 512;
+static constexpr auto default_block_size = 512;
```
thoasm, line 65:
```suggestion
static constexpr int default_block_size = 512;
```

thoasm (approved):
LGTM, my only comments are `constexpr auto` -> `constexpr int` whenever we use integer literals (just like @yhmtsai also commented).

tcojean:
I currently don't understand the root issue. If I read the portion of cppreference properly, constexpr variables are by default internal linkage?
https://en.cppreference.com/w/cpp/language/storage_duration (see internal linkage portion)

> const-qualified variables (including constexpr) that aren't declared extern and aren't previously declared to have external linkage;

upsj:
@tcojean You are right, that is actually not an issue. I thought I remembered seeing default_block_size symbols throughout the libraries, but I seem to be mistaken.

tcojean:
What I'm not sure about is the `non exported (since C++20)` part, does it mean constexpr is actually external linkage before C++20 but it's not afterward?

upsj:
We had some issues with the Intel compiler related to symbols for logger mask values, but otherwise, this should not matter.
@tcojean non-exported sounds like it deals with Modules, so not relevant for us.

tcojean:
Ah that's module export not the plain `export` keyword. The wording is really confusing... Thanks!
https://en.cppreference.com/w/cpp/language/modules

PR #852, opened 2021-08-04
Fix Readme and disable MSVC-CUDA 10.2

tcojean:
+ Update to the new package status. Simplify the HIP-related INSTALL.md section.
+ Disable the MSVC-CUDA 10.2 job.

There is an issue with the CUDA implementation which prevents a proper
execution. See pytorch/pytorch#25393 and NVIDIA/thrust#1090. Tweaking
the compiler settings would allow getting fewer errors, but it seems
impossible to prevent the errors altogether.

Is there anything else that should be updated?

### DIFF
--- a/.github/workflows/windows-msvc-cuda.yml
+++ b/.github/workflows/windows-msvc-cuda.yml
@@ -15,7 +15,6 @@ jobs:
       fail-fast: false
       matrix:
         config:
-        - {version: "10.2.89.20191206", name: "cuda102/release/shared", "mixed": "ON"}
         - {version: "latest", name: "cuda-latest/release/shared", "mixed": "OFF"}
     name: msvc/${{ matrix.config.name }} (only compile)
     runs-on: [windows-latest]
--- a/.github/workflows/windows-msvc-cuda.yml
+++ b/.github/workflows/windows-msvc-cuda.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         config:
-        - {version: "latest", name: "cuda-latest/release/shared", "mixed": "OFF"}
+        - {version: "latest", name: "cuda-latest/release/shared", "mixed": "ON"}
     name: msvc/${{ matrix.config.name }} (only compile)
     runs-on: [windows-latest]
 
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Prerequisites
 
 For Ginkgo core library:
 
-*   _cmake 3.9+_
+*   _cmake 3.13+_
 *   C++14 compliant compiler, one of:
     *   _gcc 5.3+, 6.3+, 7.3+, all versions after 8.1+_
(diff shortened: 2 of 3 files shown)


.github/workflows/windows-msvc-cuda.yml
```diff
@@ -15,7 +15,6 @@ jobs:
       fail-fast: false
       matrix:
         config:
-        - {version: "10.2.89.20191206", name: "cuda102/release/shared", "mixed": "ON"}
         - {version: "latest", name: "cuda-latest/release/shared", "mixed": "OFF"}
```
upsj, line 18: just to make sure we test mixed-precision kernel compilation
```suggestion
        - {version: "latest", name: "cuda-latest/release/shared", "mixed": "ON"}
```

README.md
```diff
@@ -59,21 +73,14 @@ following:
...
-The Ginkgo HIP module has the following __additional__ requirements:
-
-* _ROCm 3.5+_
-*    the HIP, hipBLAS, hipSPARSE, hip/rocRAND packages compiled with either:
-    * _AMD_ backend (using the `clang` compiler)
-    * _CUDA 9.2+_ backend. When using CUDA 10+, _cmake 3.12.2+_ is required.
-
 ### Windows
 
 The prequirement needs to be verified
```
upsj, line 78:
```suggestion
```

README.md
```diff
@@ -36,7 +36,7 @@ Prerequisites
 
 For Ginkgo core library:
 
-*   _cmake 3.9+_
+*   _cmake 3.13+_
 *   C++14 compliant compiler, one of:
     *   _gcc 5.3+, 6.3+, 7.3+, all versions after 8.1+_
     *   _clang 3.9+_
```
upsj, line 42: Maybe mention explicitly which compilers we are running in CI, together with a sentence like `Compilation might succeed on older compilers, but we do not monitor their status actively`. The GCC list is very verbose compared to the rest, maybe we can shorten it to something like 5.3+?

yhmtsai:
the msvc cuda issue maybe also ref this one? https://github.com/NVIDIA/thrust/issues/1090
many issue also link to it

README.md
tcojean: The problem is the GCC release system, which isn't linear. e.g., 5.5 was released after 7.2 :)
https://gcc.gnu.org/releases.html

We tested and fixed so that the ones listed here work, but we have no idea for later ones. It's true though since in practice we only test 5.5 we might have lost 5.4 or 5.3 compatibilities somewhere down the line... In short, it's a mess and I'm also not sure if listing what we test would make this any clearer (particularly since it might change at every container update).

tcojean: I think I will go with 5.5+ which we are more sure about and are still testing.

pratikvn (approved):
LGTM!

tcojean:
rebase!

PR #854, opened 2021-08-05
Some fixes for Scalar Jacobi.

pratikvn:
This PR aims to fix some of the incomplete functionalities that were left during the specialization for Scalar Jacobi.

Fixes #838

PR #855, opened 2021-08-06
Split common kernels

tcojean:
Separate the two existing implementation types:
+ CUDA/HIP kernels into `cuda_hip`
+ The unified lambda-based interface into `unified`.

Depends on #854


tcojean:
format!

upsj (approved):
LGTM! We may need to adapt the include guards for the kernel_launch* files to reflect the correct path? Curious why the formatting script didn't pick them up.

pratikvn (approved):
LGTM!

yhmtsai (approved):
LGTM. `FORMAT_HEADER_REGEX` in bot-pr-format-base.sh does not consider common, so it only apply format not format_header

PR #856, opened 2021-08-08
IDR dpcpp test workaround

yhmtsai:
After some investigation, the random NaN in IDR test may be from oneMKL GEMM.
we already report it to Intel.
Before intel fix it or we figure there is another issue in IDR, make gitlab allow dpcpp/gpu only fails on dpcpp idr test.

when dpcpp is a queue on gpu, skip the IdrIterationOneRHSIsEquivalentToRef test

**the followings are cancelled**
The current branch is for demo.
In this demo, the gitlab job will ignore the `reference/test/solver/idr_kernels` failure if the job specify IGNORE_IDR="TRUE"
- Normal: [build/nocuda/gcc/omp/release/shared](https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/1488592006) gives the normal job still recognize the failure because the job does not use `IGNORE_IDR="TRUE"`
- Allow failure on specified job: [build/nocuda-nomixed/gcc/omp/release/shared](https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/1488592011) only fails in `reference/test/solver/idr_kernels` but it is passed due to `IGNORE_IDR="TRUE"`
- Only allowed ignored test fail: [build/cuda92/gcc/all/release/shared](https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/1488591982) has `IGNORE_IDR="TRUE"` but still fails because `cuda/test/solver/idr_kernels` also fails in this job.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -120,22 +120,7 @@ include:
     - ninja -j${NUM_CORES} -l${CI_LOAD_LIMIT} install
     - |
         (( $(ctest -N | tail -1 | sed 's/Total Tests: //') != 0 )) || exit 1
-    - |
-        if [[ "${IGNORE_IDR}" == "TRUE" ]]; then
-          echo "some error may be ignored"
-          ctest -V -O test.log || true
-          num_fail=$( grep '(Failed)' test.log | wc -l )
-          ignored_num_fail=$( grep 'reference/test/solver/idr_kernels (Failed)' test.log | wc -l )
-          if [[ ${num_fail} -eq ${ignored_num_fail} ]]; then
-            echo "the error is ignored"
-            exit 3
-          else
-            echo "the error is not ignored"
-            exit 1
-          fi
-        else
-          ctest -V
-        fi
+    - ctest -V
     - ninja test_install
     - pushd test/test_install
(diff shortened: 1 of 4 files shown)


upsj:
Why this complex workaround, if we could just disable the corresponding IDR test in DPCPP with GTEST_SKIP or a DISABLED_ prefix?

yhmtsai:
I thought we still keep the random error in the job not just ignore it?

upsj:
IMO, Flaky tests don't give us any useful information, but a lot of headaches. As long as we have tests for the individual kernels, we know that the error comes from oneMKL. Just like when we didn't have SpMVs, but the rest of the kernels worked fine.

yhmtsai:
I prefer keeping the error message but do not have a strong opinion on it.
I am also fine with skipping the test if we keep it in mind.
Do you know the gtest class construction is in every TEST_F or whole testing file?
if it is the first one, I can only skip the one rhs test with real subspace.
if it is the second one, I need to skip all the following tests.

tcojean (approved):
LGTM.

upsj (approved):

PR #858, opened 2021-08-10
Add more verbose output to array assertions

upsj:
This PR outputs the surrounding area in an array equality assertion on top of the mismatching index:
```
Array vals is different from dvals at index 6337
...
  1234,1234
  1234,1234
  1234,1234
  1234,1234
  1234,1234
> -1234,1234
  -1234,1234
  -1234,1234
  -1234,1234
  -1234,1234
  -1234,1234
...
```

### DIFF
--- a/core/test/utils/assertions.hpp
+++ b/core/test/utils/assertions.hpp
@@ -375,12 +375,33 @@ template <typename ValueType>
     auto exec = first.get_executor()->get_master();
     Array<ValueType> first_array(exec, first);
     Array<ValueType> second_array(exec, second);
-    for (decltype(first.get_num_elems()) i = 0; i < num_elems1; ++i) {
+    for (size_type i = 0; i < num_elems1; ++i) {
         if (!(first_array.get_const_data()[i] ==
               second_array.get_const_data()[i])) {
             auto fail = ::testing::AssertionFailure();
             fail << "Array " << first_expression << " is different from "
                  << second_expression << " at index " << i << "\n";
+            // how many surrounding lines to print?
+            constexpr size_type context_size = 5;
+            // find boundaries, avoid overflows
+            const size_type context_begin = i - std::min(context_size, i);
+            const size_type context_end =
+                std::min(i + context_size + 1, num_elems1);
+            if (i > context_size) {
+                fail << "...\n";
+            }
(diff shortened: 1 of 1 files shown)


core/test/utils/assertions.hpp
```diff
@@ -375,12 +375,33 @@ ::testing::AssertionResult array_equal_impl(
...
+            const size_type context_end =
+                std::min(i + context_size + 1, num_elems1);
+            if (i > context_size) {
+                fail << "...\n";
+            }
+            for (auto j = context_begin; j < context_end; j++) {
+                fail << (j == i ? "> " : "  ")
+                     << ::testing::PrintToString(
+                            first_array.get_const_data()[j])
+                     << ','
```
yhmtsai, line 397:
```suggestion
                     << ', '
```

yhmtsai (approved):
LGTM. It's pretty helpful.
It will output the first segment containing incorrect element and then output the full list of both array, right?

core/test/utils/assertions.hpp
upsj: It outputs 5 elements before and after the first mismatching element, so yes.

core/test/utils/assertions.hpp
```diff
@@ -375,12 +375,33 @@ ::testing::AssertionResult array_equal_impl(
...
     Array<ValueType> second_array(exec, second);
-    for (decltype(first.get_num_elems()) i = 0; i < num_elems1; ++i) {
+    for (size_type i = 0; i < num_elems1; ++i) {
         if (!(first_array.get_const_data()[i] ==
               second_array.get_const_data()[i])) {
             auto fail = ::testing::AssertionFailure();
             fail << "Array " << first_expression << " is different from "
                  << second_expression << " at index " << i << "\n";
+            // how many surrounding lines to print?
+            constexpr size_type context_size = 5;
```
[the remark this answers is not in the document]
upsj: Good point, the 5 was pretty specific to the case I was testing, where I had some space between relevant entries. I'll change it before merging
```suggestion
            constexpr size_type context_size = 3;
```

tcojean (approved):

upsj:
rebase!

PR #759, opened 2021-05-05
Add Sparse BLAS benchmark

upsj:
To simplify #626 a bit, I pulled out some changes from it:

This PR adds benchmarks for SpGEAM, SpGEAM and transpose as well as named lambda operations to Executors.
As input parameters we use
* SpGEAM: A + A' where A' is a distance-limited row-permutation of the input matrix
* SpGEMM: A * A' where A' is either A (square), A^T (rectangular), a generated random sparse or dense matrix of appropriate size.

~~Additionally it contains wrappers for nsparse, SpECK and KokkosKernels.~~

PR #859, opened 2021-08-17
Use lambda captures for GKO_REGISTER_OPERATION

### LINKED ISSUE #177: Improved kernel mapping
Currently we map the kernels to executors with [`GKO_REGISTER_OPERATION`](https://github.com/ginkgo-project/ginkgo/blob/develop/core/base/executor.hpp#L264). There might be a better way to implement this functionality by [_inheriting from lambdas_](https://www.youtube.com/watch?v=W-xTpqj31mI).

upsj:
This PR simplifies the GKO_REGISTER_OPERATION macros by using lambda captures of the arguments instead of storing them inside a `std::tuple`. This would make the code much easier to understand, and also make it more easy for indexers like in CLion or clangd, allowing them to infer the parameters that can be passed to the make_* function.

~~A notable downside to this is that it creates 5 copies of the kernel arguments instead of only one. It should be possible to work around this by creating the lambda that is being executed only on-the-fly, but I'm not sure that's even necessary. We are talking about the choice between 128 bytes and 640 bytes of stack space for running the operation for the largest kernels we are launching (BiCGStab initialize), which amounts to less than 0.1us difference per iteration (1000 iterations) in the `overhead` solver benchmark~~

I rewrote the dispatch to avoid the overhead of 5-fold storage of the operation arguments by putting everything into the same lambda.

Closes #177

### DIFF
--- a/core/matrix/dense.cpp
+++ b/core/matrix/dense.cpp
@@ -468,10 +468,9 @@ void Dense<ValueType>::move_to(Dense<next_precision<ValueType>> *result)
 template <typename ValueType>
 void Dense<ValueType>::convert_to(Coo<ValueType, int32> *result) const
 {
-    conversion_helper(
-        result, this,
-        dense::template make_convert_to_coo<const Dense<ValueType> *&,
-                                            decltype(result)>);
+    conversion_helper(result, this, [](const auto &in, const auto &out) {
+        return dense::make_convert_to_coo(in, out);
+    });
 }
 
 
@@ -485,10 +484,9 @@ void Dense<ValueType>::move_to(Coo<ValueType, int32> *result)
 template <typename ValueType>
 void Dense<ValueType>::convert_to(Coo<ValueType, int64> *result) const
 {
-    conversion_helper(
-        result, this,
-        dense::template make_convert_to_coo<const Dense<ValueType> *&,
-                                            decltype(result)>);
+    conversion_helper(result, this, [](const auto &in, const auto &out) {
+        return dense::make_convert_to_coo(in, out);
+    });
 }
 
 
@@ -502,10 +500,9 @@ void Dense<ValueType>::move_to(Coo<ValueType, int64> *result)
 template <typename ValueType>
 void Dense<ValueType>::convert_to(Csr<ValueType, int32> *result) const
 {
-    conversion_helper(
-        result, this,
-        dense::template make_convert_to_csr<const Dense<ValueType> *&,
-                                            decltype(result)>);
+    conversion_helper(result, this, [](const auto &in, const auto &out) {
+        return dense::make_convert_to_csr(in, out);
+    });
     result->make_srow();
 }
 
@@ -520,10 +517,9 @@ void Dense<ValueType>::move_to(Csr<ValueType, int32> *result)
 template <typename ValueType>
 void Dense<ValueType>::convert_to(Csr<ValueType, int64> *result) const
 {
-    conversion_helper(
-        result, this,
-        dense::template make_convert_to_csr<const Dense<ValueType> *&,
-                                            decltype(result)>);
+    conversion_helper(result, this, [](const auto &in, const auto &out) {
+        return dense::make_convert_to_csr(in, out);
+    });
     result->make_srow();
 }
 
@@ -538,10 +534,9 @@ void Dense<ValueType>::move_to(Csr<ValueType, int64> *result)
 template <typename ValueType>
 void Dense<ValueType>::convert_to(Ell<ValueType, int32> *result) const
 {
-    conversion_helper(
-        result, this,
-        dense::template make_convert_to_ell<const Dense<ValueType> *&,
-                                            decltype(result)>);
+    conversion_helper(result, this, [](const auto &in, const auto &out) {
+        return dense::make_convert_to_ell(in, out);
+    });
 }
 
 
@@ -555,10 +550,9 @@ void Dense<ValueType>::move_to(Ell<ValueType, int32> *result)
 template <typename ValueType>
 void Dense<ValueType>::convert_to(Ell<ValueType, int64> *result) const
 {
-    conversion_helper(
-        result, this,
-        dense::template make_convert_to_ell<const Dense<ValueType> *&,
-                                            decltype(result)>);
+    conversion_helper(result, this, [](const auto &in, const auto &out) {
+        return dense::make_convert_to_ell(in, out);
+    });
 }
 
 
@@ -572,10 +566,9 @@ void Dense<ValueType>::move_to(Ell<ValueType, int64> *result)
 template <typename ValueType>
 void Dense<ValueType>::convert_to(Hybrid<ValueType, int32> *result) const
 {
-    conversion_helper(
-        result, this,
-        dense::template make_convert_to_hybrid<const Dense<ValueType> *&,
-                                               decltype(result)>);
+    conversion_helper(result, this, [](const auto &in, const auto &out) {
+        return dense::make_convert_to_hybrid(in, out);
+    });
 }
 
 
@@ -589,10 +582,9 @@ void Dense<ValueType>::move_to(Hybrid<ValueType, int32> *result)
 template <typename ValueType>
 void Dense<ValueType>::convert_to(Hybrid<ValueType, int64> *result) const
 {
-    conversion_helper(
-        result, this,
-        dense::template make_convert_to_hybrid<const Dense<ValueType> *&,
-                                               decltype(result)>);
+    conversion_helper(result, this, [](const auto &in, const auto &out) {
+        return dense::make_convert_to_hybrid(in, out);
+    });
 }
 
 
@@ -606,10 +598,9 @@ void Dense<ValueType>::move_to(Hybrid<ValueType, int64> *result)
 template <typename ValueType>
 void Dense<ValueType>::convert_to(Sellp<ValueType, int32> *result) const
 {
-    conversion_helper(
-        result, this,
-        dense::template make_convert_to_sellp<const Dense<ValueType> *&,
-                                              decltype(result)>);
+    conversion_helper(result, this, [](const auto &in, const auto &out) {
+        return dense::make_convert_to_sellp(in, out);
+    });
 }
 
 
@@ -623,10 +614,9 @@ void Dense<ValueType>::move_to(Sellp<ValueType, int32> *result)
 template <typename ValueType>
 void Dense<ValueType>::convert_to(Sellp<ValueType, int64> *result) const
 {
-    conversion_helper(
-        result, this,
-        dense::template make_convert_to_sellp<const Dense<ValueType> *&,
-                                              decltype(result)>);
+    conversion_helper(result, this, [](const auto &in, const auto &out) {
+        return dense::make_convert_to_sellp(in, out);
+    });
 }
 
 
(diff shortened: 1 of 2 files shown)


core/device_hooks/cuda_hooks.cpp
```diff
@@ -110,8 +110,10 @@ void CudaExecutor::synchronize() const GKO_NOT_COMPILED(cuda);
 
 void CudaExecutor::run(const Operation &op) const
 {
+    this->template log<log::Logger::operation_launched>(this, &op);
```
upsj, line 113: These changes were left-over from cherry-picking the named lambda commit from #759, they are not really necessary here.

include/ginkgo/core/base/executor.hpp
```diff
@@ -276,43 +276,122 @@ class Operation {
...
+ * @tparam ClosureCuda  the type of the functor executed on CUDA
+ * @tparam ClosureHip  the type of the functor executed on HIP
+ * @tparam ClosureDpcpp  the type of the functor executed on DPC++
+ */
+template <typename ClosureReference, typename ClosureOmp, typename ClosureCuda,
+          typename ClosureHip, typename ClosureDpcpp>
+class RegisteredOperation : public Operation {
+public:
+    /**
+     * Creates a RegisteredOperation object from three functors and a
```
upsj, line 304: five

include/ginkgo/core/base/executor.hpp
```diff
@@ -276,43 +276,122 @@ class Operation {
...
+     * name.
+     *
+     * @param name  the name to be used for this operation
+     * @param op_omp  a functor object which will be called by OmpExecutor
```
upsj, line 308: reference

core/matrix/dense.cpp
```diff
@@ -119,7 +119,8 @@ inline void conversion_helper(Coo<ValueType, IndexType> *result,
     exec->run(dense::make_count_nonzeros(source, &num_stored_nonzeros));
     auto tmp = Coo<ValueType, IndexType>::create(exec, source->get_size(),
                                                  num_stored_nonzeros);
-    exec->run(op(source, tmp.get()));
+    auto tmp_ptr = tmp.get();
```
upsj, line 122: this change might be unnecessary if we make the parameters `const auto&`

core/matrix/dense.cpp
```diff
@@ -425,10 +431,9 @@ void Dense<ValueType>::move_to(Dense<next_precision<ValueType>> *result)
 template <typename ValueType>
 void Dense<ValueType>::convert_to(Coo<ValueType, int32> *result) const
 {
-    conversion_helper(
-        result, this,
-        dense::template make_convert_to_coo<const Dense<ValueType> *&,
-                                            decltype(result)>);
+    conversion_helper(result, this, [](auto &in, auto &out) {
```
upsj, line 434: This change was necessary since the compiler was for some reason not able to deduce which overload of make_convert_to_coo to use, even though we specified all template parameters.

yhmtsai: Could you also provide the error message?

core/matrix/dense.cpp
```diff
@@ -669,8 +661,6 @@ void Dense<ValueType>::read(const mat_data32 &data)
 
 
 namespace {
-
-
```
yhmtsai, line 673: also here

core/matrix/dense.cpp
```diff
@@ -1231,8 +1221,6 @@ void Dense<ValueType>::get_imag(
     exec->run(dense::make_get_imag(
         this, make_temporary_output_clone(exec, result).get()));
 }
-
-
```
yhmtsai, line 1235: and this

include/ginkgo/core/base/executor.hpp
```diff
@@ -386,45 +465,35 @@ private:                                                                     \
...
-                  "This assert is used to counter the false positive extra "  \
+#define GKO_REGISTER_OPERATION(_name, _kernel)                                 \
+    template <typename... Args>                                                \
+    auto make_##_name(Args &&...args)                                          \
+    {                                                                          \
+        return ::gko::detail::make_register_operation(                         \
+            #_name, sizeof...(Args),                                           \
+            [&args...](std::shared_ptr<const ::gko::ReferenceExecutor> exec) { \
+                ::gko::kernels::reference::_kernel(                            \
+                    std::move(exec), std::forward<Args>(args)...);             \
```
yhmtsai, line 476: do you mean the five copies of arguments here?

upsj: yes

core/matrix/dense.cpp
upsj:
```
I'm guessing this has to do that the return type of make_* is `auto`, which in turn is based on the return type of the templated function make_register_operation.
```

upsj:
@yhmtsai @tcojean Can you take a second look at my modifications? I avoided the `Args` overhead now by passing a second parameter to the lambda to distinguish between the different types (and avoid having to `dynamic_cast` to the individual types again)

core/matrix/dense.cpp
```diff
@@ -468,10 +468,9 @@ void Dense<ValueType>::move_to(Dense<next_precision<ValueType>> *result)
 template <typename ValueType>
 void Dense<ValueType>::convert_to(Coo<ValueType, int32> *result) const
 {
-    conversion_helper(
-        result, this,
-        dense::template make_convert_to_coo<const Dense<ValueType> *&,
-                                            decltype(result)>);
+    conversion_helper(result, this, [](const auto &in, const auto &out) {
```
upsj, line 471: Note: those are `const` references to a `const` and a non-`const` pointer.

include/ginkgo/core/base/executor.hpp
```diff
@@ -276,43 +276,95 @@ class Operation {
...
-                  "This assert is used to counter the false positive extra " \
-                  "semi-colon warnings")
+namespace detail {
+
+
+constexpr static int ref_exec_tag = 0;
+constexpr static int omp_exec_tag = 1;
+constexpr static int cuda_exec_tag = 2;
+constexpr static int hip_exec_tag = 3;
+constexpr static int dpcpp_exec_tag = 4;
```
yhmtsai, line 287: maybe use enum?
It is good to me because we do not need to consider the user-customized executor and it is not possible before, neither.

core/matrix/dense.cpp
yhmtsai: Do you try the original one with current switch implementation?

upsj:
Nice side-effect: We get *much* better error messages with incorrect parameters:
```
../core/matrix/coo_kernels.hpp:51:10: note:   template argument deduction/substitution failed:
     void spmv(std::shared_ptr<const DefaultExecutor> exec, \
          ^~~~
../core/matrix/coo_kernels.hpp:92:5: note: in expansion of macro ‘GKO_DECLARE_COO_SPMV_KERNEL’
     GKO_DECLARE_COO_SPMV_KERNEL(ValueType, IndexType);             \
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~
../core/matrix/coo_kernels.hpp:146:1: note: in expansion of macro ‘GKO_DECLARE_ALL_AS_TEMPLATES’
 GKO_DECLARE_ALL_AS_TEMPLATES;
 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from ../include/ginkgo/core/base/array.hpp:46,
                 from ../include/ginkgo/core/matrix/coo.hpp:37,
                 from ../core/matrix/coo.cpp:33:
../include/ginkgo/core/base/executor.hpp:474:51: note:   candidate expects 4 arguments, 3 provided
                     ::gko::kernels::dpcpp::_kernel(                            \
                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                         std::static_pointer_cast<const ::gko::DpcppExecutor>(  \
                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                             exec),                                             \
                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                         std::forward<Args>(args)...);                          \
                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

include/ginkgo/core/base/executor.hpp
```diff
@@ -386,45 +438,51 @@ private:                                                                     \
...
-                  "This assert is used to counter the false positive extra "  \
+#define GKO_REGISTER_OPERATION(_name, _kernel)                                 \
+    template <typename... Args>                                                \
+    auto make_##_name(Args &&... args)                                         \
+    {                                                                          \
+        return ::gko::detail::make_register_operation(                         \
+            #_name, sizeof...(Args),                                           \
+            [&args...](std::shared_ptr<const ::gko::Executor> exec, int tag) { \
+                switch (tag) {                                                 \
+                case ::gko::detail::ref_exec_tag:                              \
```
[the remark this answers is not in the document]
upsj: So you would prefer the version with individual lambdas, taking its memory overhead into account? We could also avoid the parameter by dynamic_cast-checking for the dynamic type of the Executor parameter, but that would again introduce unnecessary overhead of dynamic dispatching based on the same dynamic type twice (once via `Operation::run`, once via inner lambda)

include/ginkgo/core/base/executor.hpp
upsj: Yes, totally! For some reason I didn't think of using an `enum`

core/matrix/dense.cpp
upsj: It still doesn't work, since it can't deduce the `auto` return type, and thus the type of the parameter correctly.

include/ginkgo/core/base/executor.hpp
```diff
@@ -386,45 +431,59 @@ private:                                                                     \
...
+#define GKO_REGISTER_OPERATION(_name, _kernel)                                 \
+    template <typename... Args>                                                \
+    auto make_##_name(Args &&...args)                                          \
+    {                                                                          \
+        return ::gko::detail::make_register_operation(                         \
+            #_name, sizeof...(Args), [&args...](auto exec) {                   \
+                if (std::is_same<                                              \
+                        decltype(exec),                                        \
+                        std::shared_ptr<const ::gko::ReferenceExecutor>>::     \
+                        value) {                                               \
```
tcojean, line 443: I see, that works because we use exec->run(kernel) which already evaluates to the correct executor type, that's a good approach.

For simplicity, how about doing something like this instead?
```suggestion
                using exec_type = std::decay_t<decltype(exec)::element_type>;
                if (std::is_same<                                              \
                        exec_type,                                        \
                        ::gko::ReferenceExecutor>::value) { \
```

include/ginkgo/core/base/executor.hpp
```diff
@@ -276,43 +276,88 @@ class Operation {
...
+namespace detail {
+
+
+/**
+ * The RegisteredOperation class wraps a functor that will be called with a tag
+ * parameter based on the dynamic type of the executor that runs it.
+ *
+ * ReferenceExecutor uses ref_exec_tag, OmpExecutor uses omp_exec_tag,
+ * CudaExecutor uses cuda_exec_tag, HipExecutor uses hip_exec_tag and
+ * DpcppExecutor uses dpcpp_exec_tag.
```
tcojean, line 289: Needs to be reverted to the previous state

tcojean (approved):
LGTM!

PR #860, opened 2021-08-18
Fix Amgx_pgm match_edge issue on isolated agg

yhmtsai:
This PR fixes the match_edge does not set the agg when it is a isolated point, whose strongest neighbor is itself.
1st commit - reference issue
2nd commit - fix ref and raise device issue
3rd commit - fix all

Moreover, switching the order of RAP product to keep more rows in the internal result

### DIFF
--- a/core/multigrid/amgx_pgm.cpp
+++ b/core/multigrid/amgx_pgm.cpp
@@ -93,8 +93,8 @@ void AmgxPgm<ValueType, IndexType>::generate()
     // Initial agg = -1
     exec->run(amgx_pgm::make_fill_array(agg_.get_data(), agg_.get_num_elems(),
                                         -one<IndexType>()));
-    IndexType num_unagg{0};
-    IndexType num_unagg_prev{0};
+    IndexType num_unagg = num_rows;
+    IndexType num_unagg_prev = num_rows;
     // TODO: if mtx is a hermitian matrix, weight_mtx = abs(mtx)
     // compute weight_mtx = (abs(mtx) + abs(mtx'))/2;
     auto abs_mtx = amgxpgm_op->compute_absolute();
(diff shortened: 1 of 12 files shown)


hartwiganzt (approved):

core/multigrid/amgx_pgm.cpp
```diff
@@ -153,9 +155,9 @@ void AmgxPgm<ValueType, IndexType>::generate()
     // TODO: use less memory footprint to improve it
     auto coarse_matrix =
         share(matrix_type::create(exec, gko::dim<2>{coarse_dim, coarse_dim}));
-    auto tmp = matrix_type::create(exec, gko::dim<2>{coarse_dim, fine_dim});
-    restrict_op->apply(amgxpgm_op, tmp.get());
-    tmp->apply(prolong_op.get(), coarse_matrix.get());
+    auto tmp = matrix_type::create(exec, gko::dim<2>{fine_dim, coarse_dim});
+    amgxpgm_op->apply(prolong_op.get(), tmp.get());
+    restrict_op->apply(tmp.get(), coarse_matrix.get());
```
upsj, line 160: I'm curious about this change - it should compute the same result, does this have a performance impact?

upsj (approved):
LGTM!

core/multigrid/amgx_pgm.cpp
yhmtsai: Yes, they give the same result in theory.
I think the parallel way usually depends on the #rows, so keeping more rows might be better.
the amgx_pgm paper claims it reduces the increasing hash table possibility if there are more rows.

upsj: good point! This might improve parallelism for medium-sized problems. The rest of course depends on the SpGEMM implementation.

cuda/preconditioner/jacobi_kernels.cu
```diff
@@ -65,8 +65,6 @@ constexpr int default_num_warps = 32;
 // current GPUs have at most 84 SMs)
 constexpr int default_grid_size = 32 * 32 * 128;
 
-constexpr int default_block_size = 512;
```
pratikvn, line 68: Any reason this was removed ?

pratikvn (approved):
LGTM!

cuda/preconditioner/jacobi_kernels.cu
yhmtsai: it's unused because those kernels are moved to unified kernel

core/multigrid/amgx_pgm.cpp
yhmtsai: I play around the current example I have, which is 6525 x 6525. It does not show the noticeable changes from this

Slaedr (approved):
LGTM!

PR #863, opened 2021-08-19
Simplify copy to device in tests

upsj:
This PR replaces almost all
```
d_mtx = Mtx::create(exec);
d_mtx->copy_from(mtx.get());
```
calls by the simpler
```
d_mtx = gko::clone(exec, mtx);
```
Most of the replacements were done with pattern matching, so it makes sense to check that all of them are actually correct (and don't involve an unintended replacement of a conversion by a copy, but that would only be possible if we were storing a gko::LinOp instead of a concrete type somewhere)


omp/test/solver/lower_trs_kernels.cpp
```diff
@@ -93,25 +93,17 @@ class LowerTrs : public ::testing::Test {
...
-        d_b = Mtx::create(omp);
-        d_b->copy_from(b.get());
-        d_x = Mtx::create(omp);
-        d_x->copy_from(x.get());
-        dt_b = Mtx::create(omp);
-        dt_b->copy_from(b.get());
-        dt_x = Mtx::create(omp);
-        dt_x->copy_from(x.get());
+        t_b = b->clone();
+        t_x = x->clone();
```
yhmtsai, line 97: just curios, the cuda and hip do not use t_b and t_x?

omp/test/solver/upper_trs_kernels.cpp
```diff
@@ -93,25 +93,17 @@ class UpperTrs : public ::testing::Test {
...
-        d_b = Mtx::create(omp);
-        d_b->copy_from(b.get());
-        d_x = Mtx::create(omp);
-        d_x->copy_from(x.get());
-        dt_b = Mtx::create(omp);
-        dt_b->copy_from(b.get());
-        dt_x = Mtx::create(omp);
-        dt_x->copy_from(x.get());
+        t_b = b->clone();
+        t_x = x->clone();
```
yhmtsai, line 97: same curios

test/solver/bicg_kernels.cpp
```diff
@@ -131,32 +131,19 @@ class Bicg : public ::testing::Test {
...
+        d_z2 = gko::clone(exec, z2);
+        d_p2 = gko::clone(exec, p2);
+        d_q2 = gko::clone(exec, q2);
+        d_x = gko::clone(exec, x);
+        d_beta = gko::clone(exec, beta);
+        d_prev_rho = gko::clone(exec, prev_rho);
+        d_rho = gko::clone(exec, rho);
         d_stop_status = std::unique_ptr<gko::Array<gko::stopping_status>>(
             new gko::Array<gko::stopping_status>(exec, n));
         *d_stop_status = *stop_status;
```
yhmtsai, line 149: can also use `make_unique`

test/solver/cg_kernels.cpp
```diff
@@ -116,24 +116,15 @@ class Cg : public ::testing::Test {
...
+        d_z = gko::clone(exec, z);
+        d_p = gko::clone(exec, p);
+        d_q = gko::clone(exec, q);
+        d_x = gko::clone(exec, x);
+        d_beta = gko::clone(exec, beta);
+        d_prev_rho = gko::clone(exec, prev_rho);
+        d_rho = gko::clone(exec, rho);
         d_stop_status = std::unique_ptr<gko::Array<gko::stopping_status>>(
             new gko::Array<gko::stopping_status>(exec, n));
         *d_stop_status = *stop_status;
```
yhmtsai, line 130: also here

test/solver/cgs_kernels.cpp
```diff
@@ -143,36 +142,21 @@ class Cgs : public ::testing::Test {
...
+        d_u_hat = gko::clone(exec, u_hat);
+        d_v_hat = gko::clone(exec, v_hat);
+        d_t = gko::clone(exec, t);
+        d_x = gko::clone(exec, x);
+        d_alpha = gko::clone(exec, alpha);
+        d_beta = gko::clone(exec, beta);
+        d_gamma = gko::clone(exec, gamma);
+        d_rho_prev = gko::clone(exec, rho_prev);
+        d_rho = gko::clone(exec, rho);
         d_stop_status = std::unique_ptr<gko::Array<gko::stopping_status>>(
```
yhmtsai, line 160: also here

test/solver/fcg_kernels.cpp
```diff
@@ -118,28 +118,17 @@ class Fcg : public ::testing::Test {
...
+        d_p = gko::clone(exec, p);
+        d_q = gko::clone(exec, q);
+        d_x = gko::clone(exec, x);
+        d_beta = gko::clone(exec, beta);
+        d_prev_rho = gko::clone(exec, prev_rho);
+        d_rho_t = gko::clone(exec, rho_t);
+        d_rho = gko::clone(exec, rho);
         d_stop_status = std::unique_ptr<gko::Array<gko::stopping_status>>(
             new gko::Array<gko::stopping_status>(exec, n));
         *d_stop_status = *stop_status;
```
yhmtsai, line 134: also here

yhmtsai (approved):
LGTM. I try to make sure there's no conversion between different type. hope I do not miss anything.

omp/test/solver/lower_trs_kernels.cpp
upsj: Yes, they don't use the individual kernels, but the `core` interface.

cuda/test/solver/idr_kernels.cpp
```diff
@@ -122,41 +122,23 @@ class Idr : public ::testing::Test {
...
+        d_g = gko::clone(cuda, g);
+        d_u = gko::clone(cuda, u);
+        d_c = gko::clone(cuda, c);
+        d_v = gko::clone(cuda, v);
+        d_p = gko::clone(cuda, p);
+        d_alpha = gko::clone(cuda, alpha);
+        d_omega = gko::clone(cuda, omega);
+        d_tht = gko::clone(cuda, tht);
+        d_residual_norm = gko::clone(cuda, residual_norm);
+        d_stop_status = std::make_unique<gko::Array<gko::stopping_status>>(
```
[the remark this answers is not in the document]
Slaedr: Is there any actual advantage over the previous way? It was already written in a way that the raw pointer does not escape anywhere. I guess this just "looks more modern" or something.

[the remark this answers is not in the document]
upsj: For the most part, it's simpler, since you only need to specify the parameter once. Usually make_unique only shows its memory-safety advantages if you create multiple instances at the same time, e.g. to pass them as parameters: https://pvs-studio.com/en/docs/warnings/v824/

omp/test/matrix/csr_kernels.cpp
```diff
@@ -108,31 +108,20 @@ class Csr : public ::testing::Test {
...
     void set_up_apply_data(int num_vectors = 1)
     {
-        mtx = Mtx::create(ref);
-        mtx->copy_from(gen_mtx<Vec>(mtx_size[0], mtx_size[1], 1));
-        complex_mtx = ComplexMtx::create(ref);
-        complex_mtx->copy_from(
-            gen_mtx<ComplexVec>(mtx_size[0], mtx_size[1], 1));
-        square_mtx = Mtx::create(ref);
-        square_mtx->copy_from(gen_mtx<Vec>(mtx_size[0], mtx_size[0], 1));
+        mtx = gen_mtx<Mtx>(mtx_size[0], mtx_size[1], 1);
```
Slaedr, line 111: Ah yes, good catch. It can simply be moved-assigned.

Slaedr (approved):
LGTM. What kind of script did you write for this? Perl?

On a separate note, we should probably not mix copying and conversion *that* seemlessly.

cuda/test/utils/assertions_test.cpp
```diff
@@ -71,7 +71,6 @@ TEST_F(MatricesNear, CanPassCudaMatrix)
 {
     auto mtx = gko::initialize<gko::matrix::Dense<>>(
         {{1.0, 2.0, 3.0}, {0.0, 4.0, 0.0}}, ref);
-    // TODO: CUDA conversion Dense -> Csr not yet implemented
     auto csr_ref = gko::matrix::Csr<>::create(ref);
     csr_ref->copy_from(mtx.get());
```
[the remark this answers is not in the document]
upsj: That was a deliberate choice from the beginning, we should probably reconsider this for the future (I am inclined to agree with your comment), but we can't easily change this now.

upsj:
@Slaedr this was plain Regex replacements with backreferences, matching
```
d(something1) = (something2)::create((something3));
d(something1)->copy_from((something1).get());
```
and variations thereof.

upsj:
rebase!

examples/ir-ilu-preconditioned-solver/ir-ilu-preconditioned-solver.cpp
```diff
@@ -97,12 +97,9 @@ int main(int argc, char *argv[])
...
     }
-    auto x = vec::create(exec);
-    auto b = vec::create(exec);
-    x->copy_from(host_x.get());
-    b->copy_from(host_x.get());
-    auto clone_x = vec::create(exec);
-    clone_x->copy_from(lend(x));
+    auto x = gko::clone(exec, host_x);
+    auto b = gko::clone(exec, host_x);
+    auto clone_x = gko::clone(exec, x);
```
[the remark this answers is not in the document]
upsj: We use copy_from in other places in the examples as well. Do you think that would be sufficient, or should I revert this part?

PR #864, opened 2021-08-19
Add support for mixed real/complex BLAS operations

upsj:
This PR adds support for `complex->(inv_)scale(real)` and `complex->(add|sub)_scaled(real, complex)` operations. The former is being used to normalize vectors in #861.


common/unified/matrix/dense_kernels.cpp
```diff
@@ -181,6 +260,33 @@ void sub_scaled(std::shared_ptr<const DefaultExecutor> exec,
 GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE(GKO_DECLARE_DENSE_SUB_SCALED_KERNEL);
 
 
+template <typename ValueType>
+void sub_scaled_real_complex(std::shared_ptr<const DefaultExecutor> exec,
+                             const matrix::Dense<ValueType> *alpha,
+                             const matrix::Dense<std::complex<ValueType>> *x,
```
yhmtsai, line 266: or it is combined with the previous one by two templated ValueType

core/matrix/dense.cpp
```diff
@@ -266,8 +270,20 @@ void Dense<ValueType>::inv_scale_impl(const LinOp *alpha)
         GKO_ASSERT_EQUAL_COLS(this, alpha);
     }
     auto exec = this->get_executor();
-    exec->run(dense::make_inv_scale(
-        make_temporary_conversion<ValueType>(alpha).get(), this));
+    // if alpha is real (convertible to real) and ValueType complex
+    if (dynamic_cast<const ConvertibleTo<Dense<>> *>(alpha) &&
```
yhmtsai, line 274: Is checking converting to Dense<double> enough?

core/matrix/dense.cpp
```diff
@@ -266,8 +270,20 @@ void Dense<ValueType>::inv_scale_impl(const LinOp *alpha)
...
+        is_complex<ValueType>()) {
+        // use the real-complex kernel
+        exec->run(dense::make_inv_scale_real_complex(
+            make_temporary_conversion<remove_complex<ValueType>>(alpha).get(),
+            dynamic_cast<complex_type *>(this)));
```
yhmtsai, line 279: codcov reports it is not covered by test.
It should be checked by SingleVectorComplexRealScaleIsEquivalentToRef.

yhmtsai (approved):
LGTM.  Should we use two templated type for them?

core/matrix/dense.cpp
upsj: Yes, as all Dense formats are inter-convertible (double->float->double), this can be used to check whether something is a real dense vector.

common/unified/matrix/dense_kernels.cpp
upsj: Yes, that would be the alternative option. I wanted to avoid having to manually unroll the instantiation macros, so I went with this solution.

MarcelKoch: I would also favor a single implementation with two templates, since the new kernels are completely identical to the old ones. I'm guessing that adding a new instantiation macro for that wouldn't be too bad, but that might be wrong.

upsj: Agreed, I moved to a single kernel for both.

core/matrix/dense.cpp
```diff
@@ -266,8 +266,20 @@ void Dense<ValueType>::inv_scale_impl(const LinOp *alpha)
         GKO_ASSERT_EQUAL_COLS(this, alpha);
     }
     auto exec = this->get_executor();
-    exec->run(dense::make_inv_scale(
-        make_temporary_conversion<ValueType>(alpha).get(), this));
+    // if alpha is real (convertible to real) and ValueType complex
+    if (dynamic_cast<const ConvertibleTo<Dense<>> *>(alpha) &&
```
MarcelKoch, line 270: Are the two branches still necessary with the templated kernels? The only difference is choosing the correct conversion of alpha. Perhaps that could be moved before the kernel run, s.t. only one kernel call is necessary.

upsj: yes, since we use the same piece of code for real and complex ValueType, we need something that compiles for settings, even if it is never taken for real ValueType.

core/matrix/dense.cpp
upsj: Codecov reports the first line (277) as covered, only the last line (279) seems to be missing? I guess we can safely ignore this, maybe the line doesn't get deleted as dead code in the binary for real ValueType?

MarcelKoch (approved):

include/ginkgo/core/base/types.hpp
```diff
@@ -461,6 +461,58 @@ GKO_ATTRIBUTES constexpr bool operator!=(precision_reduction x,
...
+#else
+#define GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE(_macro)          \
+    GKO_INSTANTIATE_FOR_EACH_NON_COMPLEX_VALUE_TYPE(_macro); \
+    template _macro(std::complex<float>);                    \
+    template _macro(std::complex<double>)
+#endif
+
+
+/**
+ * Instantiates a template for each value type compiled by Ginkgo.
```
pratikvn, line 487:
```suggestion
 * Instantiates a template for each value and scalar type compiled by Ginkgo.
```

include/ginkgo/core/base/types.hpp
```diff
@@ -461,6 +461,58 @@ GKO_ATTRIBUTES constexpr bool operator!=(precision_reduction x,
...
+ *
+ * @param _macro  A macro which expands the template instantiation
+ *                (not including the leading `template` specifier).
+ *                Should take one argument, which is replaced by the
+ *                value type.
```
pratikvn, line 492:
```suggestion
 *                Should take two arguments, which are replaced by the
 *                value type and scalar type respectively.
```

include/ginkgo/core/base/types.hpp
```diff
@@ -461,6 +461,58 @@ GKO_ATTRIBUTES constexpr bool operator!=(precision_reduction x,
...
+    GKO_INSTANTIATE_FOR_EACH_NON_COMPLEX_VALUE_TYPE(_macro); \
+    template _macro(std::complex<float>);                    \
+    template <>                                              \
+    _macro(std::complex<double>) GKO_NOT_IMPLEMENTED
+#else
+#define GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE(_macro)          \
+    GKO_INSTANTIATE_FOR_EACH_NON_COMPLEX_VALUE_TYPE(_macro); \
+    template _macro(std::complex<float>);                    \
+    template _macro(std::complex<double>)
+#endif
```
pratikvn, line 483: I think this macro is duplicated from lines 450 to 461. It is interesting that this did not cause any issues.

pratikvn (changes requested):
A minor duplication issue. Otherwise looks good.

PR #865, opened 2021-08-20
Develop is now preparing v1.5.0

tcojean:
Bump develop to v1.5.0 to distinguish it from the next master.


upsj (approved):

yhmtsai (approved):
LGTM. Should we also update the version for example? or we only need to update them when we change the example

tcojean:
I think to remove headaches and maintenance burden, I would just systematically bump the examples as well. Otherwise forgetting to update the version when updating the example could create a bug. Still, the previous 1.4.0 examples will be compatible with 1.5.0 as they should be.

pratikvn (approved):

PR #868, opened 2021-08-24
Hide executor operation symbols in core

upsj:
This significantly reduces the number of symbols being exported in Windows builds, allowing both Debug and Release builds with CUDA on Windows without modifying the compilation flags.


pratikvn (approved):
LGTM!

fritzgoebel (approved):
lgtm

PR #871, opened 2021-08-27
Jacobi build parallel

upsj:
This PR parallelizes the Jacobi compilation for CUDA/HIP into separate files for each block size. That should significantly reduce the memory requirement and worst-case runtime for jobs.


cuda/CMakeLists.txt
```diff
@@ -128,6 +128,33 @@ target_sources(ginkgo_cuda
...
 endforeach(source_file)
+if(GINKGO_JACOBI_FULL_OPTIMIZATIONS)
+    # keep in sync with jacobi_common.hpp
+    set(GKO_CUDA_JACOBI_BLOCK_SIZES)
+    foreach(blocksize RANGE 1 32)
+        list(APPEND GKO_JACOBI_BLOCK_SIZES ${blocksize})
+    endforeach()
+else()
+    # keep in sync with jacobi_common.hpp
+    set(GKO_CUDA_JACOBI_BLOCK_SIZES 1 2 4 8 13 16 32)
```
tcojean, line 139: Maybe we should make a CMAKE define just for this so that these magical values are actually in sync? It can be common for HIP as well. You would need to add a definition with commas though

tcojean (approved):
LGTM

cuda/CMakeLists.txt
upsj: Good idea, I generate the jacobi_common.hpp automatically now

hip/preconditioner/jacobi_generate_kernel.hip.cpp
```diff
@@ -81,41 +78,11 @@ void generate(syn::value_list<int, max_block_size>,
...
-                                             warps_per_block>),
-            dim3(grid_size), dim3(block_size), 0, 0, mtx->get_size()[0],
-            mtx->get_const_row_ptrs(), mtx->get_const_col_idxs(),
-            as_hip_type(mtx->get_const_values()), as_hip_type(block_data),
-            storage_scheme, block_ptrs, num_blocks);
-    }
-}
+              const IndexType* block_ptrs, size_type num_blocks);
 
 GKO_ENABLE_IMPLEMENTATION_SELECTION(select_generate, generate);
```
yhmtsai, line 83: It's interesting. The original implementation selection also builds the definition for the practical kernel, but now it gives the usages of the declaration of the kernels.

cuda/CMakeLists.txt
```diff
@@ -129,6 +129,33 @@ target_sources(ginkgo_cuda
 foreach(source_file IN LISTS GKO_CUDA_COMMON_SOURCES)
     set_source_files_properties(${source_file} PROPERTIES LANGUAGE CUDA)
 endforeach(source_file)
+if(GINKGO_JACOBI_FULL_OPTIMIZATIONS)
+    set(GKO_CUDA_JACOBI_BLOCK_SIZES)
+    foreach(blocksize RANGE 1 32)
+        list(APPEND GKO_JACOBI_BLOCK_SIZES ${blocksize})
```
yhmtsai, line 135:
```suggestion
        list(APPEND GKO_CUDA_JACOBI_BLOCK_SIZES ${blocksize})
```

yhmtsai: but why is it not caught by the test?

yhmtsai (changes requested):

cuda/CMakeLists.txt
upsj: Oh, good catch! Probably because we don't build with full optimizations in any tests?

upsj:
@yhmtsai do you have an idea what the issue with the formatting is? It wants to reorder the generated header into a separate group, and put it into `<...>` instead of `"..."`

yhmtsai:
format_header will check the header is reachable or not in ginkgo (source header) to decide "..." (reachable) or <...> (not reachable)
it needs to add special cases or different way for those headers generated in the binary folder after cmake

upsj:
Can you maybe add a special comment or config option to the script? I'm not that familiar with it and don't want to break anything ;)

yhmtsai:
@upsj The following patch should work. I do not write a general solution for them.

```
index 7a24f50bef..1b57edc79f 100755
--- a/dev_tools/scripts/format_header.sh
+++ b/dev_tools/scripts/format_header.sh
@@ -4,6 +4,7 @@ CLANG_FORMAT=${CLANG_FORMAT:="clang-format"}
 
 convert_header () {
index 7a24f50bef..1b57edc79f 100755
--- a/dev_tools/scripts/format_header.sh
+++ b/dev_tools/scripts/format_header.sh
@@ -4,6 +4,7 @@ CLANG_FORMAT=${CLANG_FORMAT:="clang-format"}
 
 convert_header () {
     local regex="^(#include )(<|\")(.*)(\"|>)$"
+    local jacobi_regex="(cuda|hip)\/preconditioner\/jacobi_common(\.hip)?\.hpp"
     if [[ $@ =~ ${regex} ]]; then
         header_file="${BASH_REMATCH[3]}"
         if [ -f "${header_file}" ]; then
@@ -14,6 +15,8 @@ convert_header () {
             fi
         elif [ "${header_file}" = "matrices/config.hpp" ]; then
             echo "#include \"${header_file}\""
+        elif [[ "${header_file}" =~ ${jacobi_regex} ]]; then
+            echo "#include \"${header_file}\""
         else
             echo "#include <${header_file}>"
         fi

upsj:
This formatting message can now be ignored, since it will be fixed after merge (we always use the `develop` formatting script)

yhmtsai (approved):

PR #872, opened 2021-08-29
Change clang-format PointerAlignment to Left

upsj:
Not sure if this is controversial, but following a previous discussion in Slack, it sounded like this might be a sensible choice now that we pushed a release.

This avoids the clang-format inconsistencies related to variadic pointer and reference parameters with PointerAlignment: Right.
Older clang-format versions format `T&&... a` as `T &&... a`, but newer versions format it as `T &&...a`, while normal
variadic parameters are formatted as `T... a` any time. This inconsistency is not present for PointerAlignment: Left
Also, ignoring multiple variables being declared in the same expression (which we don't do anyways), `type* obj` makes it slightly more clear that the left part is the type and the right part is the variable being declared, while `type *obj` puts more emphasis on what `obj` is pointing on.

Related discussion of the behavior differences: https://stackoverflow.com/questions/65964384/how-to-make-clang-format-11-put-a-space-after-variadic-template-decleration
Discussion on attachment in general: https://www.stroustrup.com/bs_faq2.html#whitespace


yhmtsai:
I am fine with this change because we also have a rule to make the definition clear.
In short, we do not allow to define two variables in the same line. we must keep forcing it
Thus, we do not have the confusion introduced by pointer left alignment.
Others changes are introduced by `make format` with new `.clang-format`, right?

upsj:
I ran ` clang-format -i `  on all files matching ` *.cuh,*.cu,*.hpp.inc,*.cpp,*.hpp `  manually, since make format only picks up changed files. Anything I might have missed?

tcojean (approved):

accessor/accessor_references.hpp
```diff
@@ -364,50 +366,52 @@ class scaled_reduced_storage
...
+    operator=(const scaled_reduced_storage& ref) &&
     {
         std::move(*this) = static_cast<arithmetic_type>(ref);
         return static_cast<arithmetic_type>(*this);
     }
 
     constexpr GKO_ACC_ATTRIBUTES arithmetic_type
-    operator=(scaled_reduced_storage &&ref) &&noexcept
+        operator=(scaled_reduced_storage&& ref) &&
+        noexcept
```
yhmtsai, line 407: this will be two line

yhmtsai (approved):

Slaedr:
This might significantly complicate the `batch-develop` workflow now...

(Also: `const int* const x` looks weird compared to `const int *const x` mainly because the second `const` applies the the pointer, and the second one suits that better. 😅 )

PR #874, opened 2021-08-30
Fixes to the smoke tests

tcojean:
Improve smoke tests in two ways:
+ Skip smoke tests when no device is found
+ Don't use the `imbalance` strategy with CSR, which requires a device executor.


yhmtsai (approved):
LGTM

upsj (approved):
LGTM!

PR #876, opened 2021-09-01
Remove multiple RHS and optimize batch CG and IDR on Cuda

Slaedr:
This completes the optimization of batched CG and IDR to the level of the other batched solvers. There may be more opportunities for optimization in IDR; not all possible optimizations have actually been done.

In addition, the test utility function used for asserting batch matrix closeness has been modified slightly to fail when there are NaNs.


Slaedr:
format!

PR #878, opened 2021-09-03
Example showing how to use batch solvers with application codes

Slaedr:
The example has a function that mimics an "application", in that generates data in raw pointers already on the device. Ginkgo batched solvers are able to use that data in-place.

Limitations:
- If the application data is a const pointer, the current way will not work. We will have to copy the data in that case.


hartwiganzt (approved):

hartwiganzt:
LGTM. I think it shows how to use the technology. But I did not check technical aspects.

examples/batched-solver/batched-solver.cpp
```diff
@@ -108,46 +150,200 @@ int main(int argc, char *argv[])
...
+    auto res = vec_type::create(exec, batch_vec_size);
+    res->copy_from(lend(b));
+    A->apply(lend(one), lend(x), lend(neg_one), lend(res));
+    // allocate and compute residual norm on the device
+    auto res_norm =
+        gko::batch_initialize<real_vec_type>(num_systems, {0.0}, exec);
+    res->compute_norm2(lend(res_norm));
+    // transfer residual norm to the host
+    auto host_res_norm = real_vec_type::create(exec->get_master());
+    host_res_norm->copy_from(lend(res_norm));
```
pratikvn, line 242: Can be simplified, because internally, it should be able to use `make_temporary_clone`
```suggestion
    auto res_norm =
        gko::batch_initialize<real_vec_type>(num_systems, {0.0}, exec->get_master());
    res->compute_norm2(lend(res_norm));
```

pratikvn (approved):
LGTM!  Maybe we can also actually deploy the documentation from either this branch/ the batch-develop branch, so that this example and the batch-develop documentation appears nicely on the webpage.

examples/batched-solver/batched-solver.cpp
```diff
@@ -108,46 +150,200 @@ int main(int argc, char *argv[])
     // executor where Ginkgo will perform the computation
     const auto exec = exec_map.at(executor_string)();  // throws if not valid
 
-    auto A = share(gko::read<mtx>(std::ifstream("data/A.mtx"), exec));
-    // Create RHS and initial guess as 1
-    gko::size_type size = A->get_size()[0];
-    auto host_x = vec::create(exec->get_master(), gko::dim<2>(size, 1));
-    for (auto i = 0; i < size; i++) {
-        host_x->at(i, 0) = 1.;
+    const int num_rows = 35;  // per system
```
tcojean, line 153: In terms of comments, it might make sense to make use of those `@sect` or any other system to show where the phases of the example start/end (general setup, actually solving the problem, unbatch & data analysis, etc).

examples/batched-solver/doc/intro.dox
```diff
@@ -1,19 +1,10 @@
...
-    Ginkgo and its features better.
-    <li> <b>Results:</b> This section shows the results of the code when run. Though the
-    results may not be completely the same, you can expect the behaviour to be similar.
-    <li> <b>The plain program:</b> This is the complete code without any comments to have
-    an complete overview of the code.
-</ol>
+A `batch' here means a set of small linear systems that can be solved
+independently, but each system is too small to use an entire computing device.
+For now, a batch of sparse matrices is required to have a common sparsity
+pattern for all the matrices.
```
tcojean, line 10: I'm not sure if this is the best place, but it might be a good idea to mention any pitfalls or limitations, like size/shared memory concerns etc. That might be more relevant in the future than now though so it's fine like this for now.

tcojean (approved):
LGTM

examples/batched-solver/batched-solver.cpp
Slaedr: That does not actually work because I'm printing it out, not `write`ing it. That's because I found it annoying to display the matrix market header before each line just for looking at the residual norms.

examples/batched-solver/batched-solver.cpp
```diff
@@ -108,46 +147,205 @@ int main(int argc, char *argv[])
...
+    const real_type reduction_factor{1e-6};
+
+    // @sect3{Create the batch solver factory}
+    // Create a batched solver factory with relevant parameters.
+    auto solver_gen =
+        solver_type::build()
+            .with_max_iterations(500)
+            .with_residual_tol(reduction_factor)
+            .with_tolerance_type(gko::stop::batch::ToleranceType::relative)
+            .with_preconditioner(gko::preconditioner::batch::type::jacobi)
```
yhmtsai, line 205: I am not convinced with it.
At least for preconditioners, it gives different usage from usual ginkgo usage.
It also gives several configs from one of any preconditioner.
the issue also applies tolerance

examples/batched-solver/batched-solver.cpp
yhmtsai: does batch_initialize need?
I think it can be dense create and fill

examples/batched-solver/batched-solver.cpp
```diff
@@ -108,46 +147,205 @@ int main(int argc, char *argv[])
...
+    index_type* const col_idxs = exec->alloc<index_type>(nnz);
+    exec->copy_from(exec->get_master().get(), static_cast<size_type>(nnz),
+                    colidxs.data(), col_idxs);
+    value_type* const all_values = exec->alloc<value_type>(nsystems * nnz);
+    exec->copy_from(exec->get_master().get(), nsystems * nnz, allvalues.data(),
+                    all_values);
+    value_type* const all_b = exec->alloc<value_type>(nsystems * nrows);
+    exec->copy_from(exec->get_master().get(), nsystems * nrows, allb.data(),
+                    all_b);
+    return {nsystems, nrows, nnz, row_ptrs, col_idxs, all_values, all_b};
```
yhmtsai, line 342: Is there any reason not to use array?
it requires the manually free afterward. Also the clean up does not need it actually

examples/batched-solver/batched-solver.cpp
yhmtsai: It also changes how we handle the memory of preconditioners.
in a current way, you need to handle different generate methods in if-else and tons of pointer (void maybe).
Jacobi may not introduce too much different performance impact on generating in apply, but it should give the difference in ILU or some not light-way preconditioner

Slaedr: The idea here is that we'll only have a few (maybe 2 or 3) preconditioners for batched solvers, and they will be tightly integrated with the solvers. So I don't yet know what we can put in a "core" module for batched preconditioners. So there are no user-facing classes for them, and it makes sense to me to just provide them as factory parameters. We are open to other options though, and the interface can be changed.

examples/batched-solver/batched-solver.cpp
Slaedr: Here the idea is to show integration with applications. It's not unusual for applications to directly deal with pointers. Even if not, that is the lowest common denominator for communication between the application and Ginkgo. So I thought it would be a good idea to show interfacing using raw pointers.

examples/batched-solver/batched-solver.cpp
Slaedr: Never mind, I made a mistake while testing it out; it works.

examples/batched-solver/batched-solver.cpp
yhmtsai: I see, then maybe wrap a destructor of the structure should be better than using array

examples/batched-solver/batched-solver.cpp
yhmtsai: for example, jacobi contains max_blocks, parilu contains sweeps, or ilut contains threshold, etc...
you will need to add all configurations from possible preconditioners into all solvers.
The preconditioner is not only a preconditioner but also a linop. If it is a linop, it always needs core even if the structure is close to empty.

Slaedr: Yes, that may well turn out to be an issue once we want to integrate batch Par ILU or something. We'll need to think of an architecture for that - I'm sure something is possible. I will definitely keep this mind when thinking about the final code structure for this.

examples/batched-solver/batched-solver.cpp
Slaedr: I wanted to keep the struct and example simple... if I add a destructor, then it makes sense to add a constructor, and it'll start to get quite complex. What I had in mind was that if you have a C-like code, and you're willing to compile some files using C++, you can use GInkgo.

yhmtsai: the constructor and destructor are like what you did, I think.

PR #880, opened 2021-09-06
Example for using batched solvers with linear systems read from files

Slaedr:
An example to show how to read a batch of matrices from files and solve them using a batched solver.

Should merge #878 and probably also #879 before this one.


examples/batched-solver-from-files/batched-solver-from-files.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+    // Generate the batch solver from the batch matrix
+    auto solver = solver_gen->generate(A);
+
+    // add the logger to the solver
+    solver->add_logger(logger);
+    // Solve the batch system
+    solver->apply(lend(b), lend(x));
+    // This is not necessary, but one might want to remove the logger before
+    //  the next solve using the same solver object.
```
pratikvn, line 181: Can we maybe add a simple chrono timing here to make it easier to time this ?

examples/batched-solver-from-files/batched-solver-from-files.cpp
```diff
@@ -0,0 +1,223 @@
...
+    }
+    x->copy_from(host_x.get());
+
+    const real_type reduction_factor{1e-6};
+
+    // @sect3{Create the batch solver factory}
+    // Create a batched solver factory with relevant parameters.
+    auto solver_gen =
+        solver_type::build()
+            .with_max_iterations(500)
```
pratikvn, line 161: I would also add the `num_shared_vectors` parameter here, and a comment to say what it does.

pratikvn (approved):
LGTM! Thanks for adding this! Maybe we should just add some timing capabilities so that they can experiment with that as well.

examples/batched-solver-from-files/batched-solver-from-files.cpp
tcojean: Agree here. The issue is maybe since it's batched you want a GPU-side timer for better resolution? At least a CPU-side timer would be useful though.

examples/batched-solver-from-files/doc/kind
```diff
@@ -0,0 +1 @@
+basic
```
tcojean, line 1: That's more for later, but at some point, it will make sense to have a `batched` kind for these examples.

examples/batched-solver-from-files/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+cmake_minimum_required(VERSION 3.9)
```
tcojean, line 1: Officially, Ginkgo is now at `3.13` but it doesn't matter here.

examples/batched-solver-from-files/batched-solver-from-files.cpp
```diff
@@ -0,0 +1,223 @@
...
+             }},
+            {"reference", [] { return gko::ReferenceExecutor::create(); }}};
+
+    // executor where Ginkgo will perform the computation
+    const auto exec = exec_map.at(executor_string)();  // throws if not valid
+
+    const std::string problem_descr_str = argc >= 3 ? argv[2] : "pores_1";
+    const size_type num_systems = argc >= 4 ? std::atoi(argv[3]) : 2;
+    const size_type num_duplications = argc >= 5 ? std::atoi(argv[4]) : 2;
+    const std::string batch_scaling = argc >= 6 ? argv[5] : "none";
```
tcojean, line 106: I think it might make sense to describe these with a comment since they are tied to where you find the matrices, although it's somewhat easy to understand when looking few lines below.

tcojean: Another way to do it is to put this all together in the comment of `Read batch from files`.

examples/batched-solver-from-files/doc/kind
tcojean: Similarly, for clarity of user documentation, I think in the future it would be good to add the `builds-on` file to state this is built on top of the other example (or vice versa) in order to link the batched examples together.

tcojean (approved):
LGTM

examples/batched-solver-from-files/doc/kind
Slaedr: Maybe I just add a "batched" kind to both of them. Neither really builds on the other.

pratikvn:
Can you add a `builds-on` file as in the other example for this and the other merged batched solver example ?

Slaedr:
I don't think either of these build on the other, though. Maybe I misunderstood the concept, but what I mean is you don't need to be familiar with one example to understand the other.

pratikvn:
No, I mean those files are needed to compile the documentation for the examples. If they don't build on anything you can leave it empty as in the simple-solver example.

pratikvn:
I think we can go ahead and merge this as there should be no conflicts with `batch-develop`

PR #882, opened 2021-09-08
Automatically cancel earlier actions on the same branch

MarcelKoch:
This PR enables automatic cancellation of actions on outdated commits of a branch. This will also cancel already running actions. The feature is still in beta, so we might need to adjust it in the future. 

Github docs: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency

### DIFF
--- a/.github/workflows/bot-pr-updated.yml
+++ b/.github/workflows/bot-pr-updated.yml
@@ -3,7 +3,7 @@ on:
     types: [opened,synchronize]
 name: OnSyncPR
 concurrency:
-  group: ${{ github.workflow }}-${{ github.head_ref }}
+  group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 jobs:
   check-format:
--- a/.github/workflows/bot-pr-updated.yml
+++ b/.github/workflows/bot-pr-updated.yml
@@ -1,8 +1,10 @@
 on:
   pull_request_target:
     types: [opened,synchronize]
-
 name: OnSyncPR
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref }}
+  cancel-in-progress: true
 jobs:
   check-format:
     runs-on: ubuntu-18.04
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -9,6 +9,10 @@ on:
         required: false
         default: false
 
+concurrency:
+  group: ${{ github.workflow}}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   osx-clang-omp:
     strategy:
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -2,6 +2,13 @@ name: OSX-build
 
 on:
   push:
+    branches:
+      - 'master'
+      - 'develop'
+    tags:
+      - '**'
+  pull_request_target:
+    types: [opened,synchronize]
   workflow_dispatch:
     inputs:
       debug_enabled:
@@ -10,7 +17,7 @@ on:
         default: false
 
 concurrency:
-  group: ${{ github.workflow}}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ (github.head_ref && github.ref) || github.run_id }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -5,9 +5,10 @@ on:
     branches:
       - 'master'
       - 'develop'
+      - 'release/**'
     tags:
       - '**'
-  pull_request_target:
+  pull_request:
     types: [opened,synchronize]
   workflow_dispatch:
     inputs:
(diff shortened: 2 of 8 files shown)


.github/workflows/osx.yml
```diff
@@ -9,6 +9,10 @@ on:
         required: false
         default: false
 
+concurrency:
+  group: ${{ github.workflow}}-${{ github.ref }}
```
upsj, line 13: nit: 
```suggestion
  group: ${{ github.workflow }}-${{ github.ref }}
```

upsj (approved):
LGTM! Great idea!

yhmtsai:
It needs to distinguish the develop branch and other branches.
The develop branch always run full tests even if a new merge comes in.

yhmtsai:
Also, collect the previous discussion here.

- auto-cancel:
Pro: do not need to cancel the jobs manually, and release the limited resource automatically
Con: Need to wait or rerun the job when we need it. For example, add a fail test before fix it or develop branch, which the intermediate pipeline result is still important

- current behavior:
Pro: keep every test if we do not cancel it
Con: cancelling 5 or 6 workflow in github is painful

MarcelKoch:
@yhmtsai agreed on the distinction for the develop branch.
Could you elaborate on the Con point for this PR, I'm not sure if I can follow that? As far as I can see, at least the logs from the cancelled are still available.

yhmtsai:
the logs from the cancelled is visible only for the already run log, right?
It will stop log in the middle when it receive the termination.

add the failed test to commit <-- needs to wait the test finish
pushing the fix too quickly will not launch failed test. <-- potential issue
(need to rerun/wait the previous fail)
I am not sure the rerun will be cancelled or not when the new pipeline is issued again

I do not stop this pr if develop always finish its pipeline.
In my opinion, the debug staff frequency is lower than the new-feature staff.
the paying (rerun/wait for previous fail) is acceptable and that's why I think it is doable before

MarcelKoch:
There seems to be no good option to only turn off this feature for develop. Possible options I've encountered so far:
- add it under `job.<id>`, but this will lead to non-deterministic cancellation
- disable the `cancel-in-progress` option, but this will keep the job only if it's already queued or running
- create separate workflows for develop, but this will result in a lot of duplication
The 2nd option is the most promising one, but I might still happen that a job on develop is cancelled.

upsj:
Github added support for [fallback values](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-a-fallback-value), so we can distinguish between PRs and develop merges.

yhmtsai:
Maybe also gitlab too? AFAIK, it needs to change the setting of the gitlab

upsj:
Some explanation: Currently the actions are being run based on the `push` event, which means even branches without a PR will run actions. This changes the behavior to use `push` only for `master`, `develop` and release tags. The PR jobs are being launched by the `pull_request` event instead, which has the correct `head_ref` available to build a PR-unique concurrency group ID. If `head_ref` is not set, it uses the run ID instead, which is always unique, so no cancellation will happen for non-PR jobs.

pratikvn:
@upsj, Maybe you can disable the abidiff failing here ?

.github/workflows/osx.yml
```diff
@@ -10,7 +17,7 @@ on:
         default: false
 
 concurrency:
-  group: ${{ github.workflow}}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ (github.head_ref && github.ref) || github.run_id }}
```
yhmtsai, line 21: what's && here? are they trancated together

upsj: This is basically "if (head_ref) { github.ref } else { run_id }"

upsj:
@pratikvn good point! I added it

.github/workflows/osx.yml
yhmtsai: okay, it is still in the fallback behavior

upsj:
@yhmtsai I think due to the fact that Gitlab is being mirrored from Gitlab's side, there is no direct way to distinguish between PR jobs and regular jobs. A more appropriate approach would probably be pushing to Gitlab from a Github Action, this would also enable things like a `check-pr!` command.

upsj:
github.head_ref is the ID of the last commit in this PR, thus it is not suitable to identify the PR. It is only available if the action was triggered by a pull request-related event (in this case: pull_request synchronized, which is triggered if the PR branch was updated and can be merged into the base branch, `develop` in most cases). `github.ref` on the other hand is a textual description like `refs/pull/882` that is unique and fixed for each PR.

.github/workflows/bot-pr-updated.yml
```diff
@@ -1,8 +1,10 @@
 on:
   pull_request_target:
     types: [opened,synchronize]
-
 name: OnSyncPR
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref }}
```
yhmtsai, line 6: if the head_ref is the ID of last commit, any update or rebase will not use the same head_ref such that they never face the cancel situtation

upsj: oh right, I forgot to update this, that should be `head_ref`

yhmtsai (approved):
LGTM

PR #885, opened 2021-09-13
Create submatrix for sparse matrix formats

pratikvn:
This PR adds support for creating submatrices from spans for sparse matrix formats. Dense already has this functionality.

An additional interface is added to ensure the same function signatures for the different matrix formats. 

Only CSR is supported for now, but other formats could be easily added.

### TODO
+ [x] Merge #831


common/cuda_hip/matrix/csr_kernels.hpp.inc
```diff
@@ -1046,3 +1046,70 @@ __global__ __launch_bounds__(default_block_size) void inv_symm_permute_kernel(
...
+template <typename IndexType>
+__global__
+    __launch_bounds__(default_block_size) void calculate_nnz_per_row_in_span(
+        const span row_span, const span col_span,
+        const IndexType* __restrict__ row_ptrs,
+        const IndexType* __restrict__ col_idxs,
+        size_type* __restrict__ nnz_per_row)
+{
+    const auto tidx = thread::get_thread_id_flat();
+    if (tidx >= row_span.begin && tidx < row_span.end) {
```
upsj, line 1104: you could avoid some idle threads here for small submatrix extraction (need to adjust launch parameters, of course)
```suggestion
    const auto tidx = thread::get_thread_id_flat() + row_span.begin;
    if (tidx < row_span.end) {
```

core/matrix/csr.cpp
```diff
@@ -534,6 +539,28 @@ bool Csr<ValueType, IndexType>::is_sorted_by_column_index() const
...
+    exec->run(csr::make_fill_array(row_nnz.get_data(), row_nnz.get_num_elems(),
+                                   zero<size_type>()));
+    exec->run(csr::make_calculate_nonzeros_per_row_in_span(
+        this, row_span, column_span, &row_nnz));
+    auto sub_mat_nnz = reduce(row_nnz);
+    auto sub_mat =
+        Mat::create(exec, sub_mat_size, sub_mat_nnz, this->get_strategy());
+    exec->run(csr::make_compute_submatrix(this, &row_nnz, row_span, column_span,
+                                          sub_mat.get()));
+    return sub_mat;
```
upsj, line 560: are we building the strategy data somewhere?
```suggestion
    sub_mat->make_srow();
    return sub_mat;
```

core/matrix/csr.cpp
```diff
@@ -534,6 +539,28 @@ bool Csr<ValueType, IndexType>::is_sorted_by_column_index() const
...
+    Array<size_type> row_nnz(exec, row_span.length());
+    exec->run(csr::make_fill_array(row_nnz.get_data(), row_nnz.get_num_elems(),
+                                   zero<size_type>()));
+    exec->run(csr::make_calculate_nonzeros_per_row_in_span(
+        this, row_span, column_span, &row_nnz));
+    auto sub_mat_nnz = reduce(row_nnz);
+    auto sub_mat =
+        Mat::create(exec, sub_mat_size, sub_mat_nnz, this->get_strategy());
+    exec->run(csr::make_compute_submatrix(this, &row_nnz, row_span, column_span,
+                                          sub_mat.get()));
```
upsj, line 559: make_compute_submatrix calls components::prefix_sum, doing the same kind of reduction again.
If you want to avoid this duplicate work and storage, you could extend `row_nnz` by one element and pass `row_nnz` directly to the `Mat` constructor together with matching `col_idxs` and `values` arrays.

cuda/test/matrix/csr_kernels2.cpp
```diff
@@ -0,0 +1,139 @@
...
+    set_up_apply_data();
+
+    gko::span rspan{36, 98};
+    gko::span cspan{26, 104};
+    auto smat1 = this->mtx->create_submatrix(rspan, cspan);
+    auto sdmat1 = this->dmtx->create_submatrix(rspan, cspan);
+    auto hmat = Mtx::create(this->ref);
+    hmat->copy_from(sdmat1.get());
+
+    GKO_ASSERT_MTX_NEAR(hmat, smat1, 1e-14);
```
upsj, line 135: nit: we copy the values over directly, so this should match bitwise.
```suggestion
    GKO_ASSERT_MTX_NEAR(hmat, smat1, 0);
```

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -654,6 +654,20 @@ class Preconditionable {
...
+ * The row and column spans must be in range of the size of the matrix.
+ *
+ * @ingroup LinOp
+ */
+template <typename ConcreteType>
+class SubMatrixExtractable {
+public:
+    virtual std::unique_ptr<ConcreteType> create_submatrix(
+        const gko::span& row_span, const gko::span& column_span) const = 0;
+};
```
upsj, line 668: I'm not sure about this interface - being templated on ConcreteType, it doesn't give us much flexibility in generically creating a submatrix for any matrix type. Do we expect to need this for other matrix types as well? We could for now directly add it as a Csr member function.

omp/matrix/csr_kernels.cpp
```diff
@@ -713,6 +713,68 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
...
+    }
+    components::prefix_sum(exec, res_row_ptrs, num_rows + 1);
+#pragma omp parallel for
+    for (size_type row = 0; row < num_rows; ++row) {
+        size_type res_nnz = res_row_ptrs[row];
+        for (size_type nnz = row_ptrs[row_offset + row];
+             nnz < row_ptrs[row_offset + row + 1]; ++nnz) {
+            if ((col_idxs[nnz] < (col_offset + num_cols) &&
+                 col_idxs[nnz] >= col_offset)) {
+                result->get_col_idxs()[res_nnz] = col_idxs[nnz] - col_offset;
```
upsj, line 766: nit: you could simplify this condition a bit
```suggestion
            const auto local_col = col_idxs[nnz] - col_offset;
            if (local_col >= 0 && local_col < num_cols) {
                result->get_col_idxs()[res_nnz] = local_col;
```

omp/test/matrix/csr_kernels2.cpp
```diff
@@ -0,0 +1,138 @@
...
+TEST_F(Csr, CreateSubMatrixIsEquivalentToRef)
+{
+    set_up_apply_data();
+
+    gko::span rspan{36, 98};
+    gko::span cspan{26, 104};
+    auto smat1 = this->mtx->create_submatrix(rspan, cspan);
+    auto sdmat1 = this->dmtx->create_submatrix(rspan, cspan);
+
+    GKO_ASSERT_MTX_NEAR(sdmat1, smat1, 1e-14);
```
upsj, line 134:
```suggestion
    GKO_ASSERT_MTX_NEAR(sdmat1, smat1, 0);
```

reference/matrix/csr_kernels.cpp
```diff
@@ -633,6 +633,62 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
...
 
+template <typename ValueType, typename IndexType>
+void calculate_nonzeros_per_row_in_span(
+    std::shared_ptr<const DefaultExecutor> exec,
+    const matrix::Csr<ValueType, IndexType>* source, const span& row_span,
+    const span& col_span, Array<size_type>* row_nnz)
+{
+    size_type res_row = 0;
+    for (size_type row = row_span.begin; row < row_span.end; ++row) {
+        for (size_type col = source->get_const_row_ptrs()[row];
```
upsj, line 644: using these member functions can hinder readability a bit, how about putting them into variables `row_ptrs`, `col_idxs`?

cuda/test/matrix/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 ginkgo_create_test(coo_kernels)
 ginkgo_create_test(csr_kernels)
+ginkgo_create_test(csr_kernels2)
```
upsj, line 3: Why not extend csr_kernels?

upsj (commented):
Looks pretty useful! I'm guessing this is related to RAS overlaps? Does this degree of genericity suffice, i.e. are ranges enough to describe overlapping regions, or can there be gaps?

The code itself also looks good, only minor comments and questions (I will review the array reduce part in the other PR). Nonzero filtering is also part of the ParILUT implementation, not sure though whether it makes sense to try and merge those two.

include/ginkgo/core/base/lin_op.hpp
pratikvn: I think it might make sense to have this for other formats as well. And this interface would enforce the same signatures for all the formats.

cuda/test/matrix/CMakeLists.txt
pratikvn: The csr_kernels has waay too many tests and running them takes a long amount of time. It might make sense to split them across a few files ?

pratikvn:
Yes, ideally, I think for RAS you would need a superset of this functionality, which i think could probably take a set of ranges, for example an `IndexSet` and build the submatrix from that. But I think that is slightly more complicated to implement. I will look into that once we have the `IndexSet` merged. 

Nevertheless, I think this simple one with just a continuous range is also useful.

include/ginkgo/core/base/lin_op.hpp
pratikvn: I think if you want to be able to create submatrices in general for all matrix formats, then you either need to have this interface, or have separate functions in each matrix format. 

I am not sure we have currently strictly follow a rule where we say we only implement cheap operations for formats other than Csr and Dense. I think this would unnecessarily restrict our ability to provide functionality in other formats, and have to convert to CSR every time, which might not be reasonable.

If we are concrete matrix format agnostic (which in many solvers we are), I think we should try to maintain the same functionality and interface across all of them possibly throwing a runtime error when some functionality has not been implemented.

Regarding Dense functionality, I agree that the functionality is different there, but I think the function name, `create_submatrix` is incorrect there and we should probably change it to `create_submatrix_view` which would be more accurate for what it does. Maybe we should do this in 2.0.

common/cuda_hip/matrix/csr_kernels.hpp.inc
```diff
@@ -1046,3 +1046,70 @@ __global__ __launch_bounds__(default_block_size) void inv_symm_permute_kernel(
...
+__global__
+    __launch_bounds__(default_block_size) void calculate_nnz_per_row_in_span(
+        const span row_span, const span col_span,
+        const IndexType* __restrict__ row_ptrs,
+        const IndexType* __restrict__ col_idxs,
+        IndexType* __restrict__ nnz_per_row)
+{
+    const auto tidx = thread::get_thread_id_flat() + row_span.begin;
+    if (tidx < row_span.end) {
+        for (size_type col = row_ptrs[tidx]; col < row_ptrs[tidx + 1]; ++col) {
```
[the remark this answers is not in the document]
pratikvn: I will get back to this later, as this is internal implementation only and can be updated later. But thank you for the suggestions.

common/cuda_hip/matrix/csr_kernels.hpp.inc
```diff
@@ -1046,3 +1046,70 @@ __global__ __launch_bounds__(default_block_size) void inv_symm_permute_kernel(
...
+        ValueType* __restrict__ result_values)
+{
+    const auto tidx = thread::get_thread_id_flat();
+    if (tidx < num_rows) {
+        size_type res_nnz = result_row_ptrs[tidx];
+        for (size_type nnz = source_row_ptrs[row_offset + tidx];
+             nnz < source_row_ptrs[row_offset + tidx + 1]; ++nnz) {
+            if ((source_col_idxs[nnz] < (col_offset + num_cols) &&
+                 source_col_idxs[nnz] >= col_offset)) {
+                result_col_idxs[res_nnz] = source_col_idxs[nnz] - col_offset;
```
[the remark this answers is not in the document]
pratikvn: Same as above, I will get back to this as: 
1. I need to update this with the IndexSet to make it more general
2. This is used only in in the generate step, which should be a one time call at the preconditioner setup, so it is not very critical, although some more performance would not hurt.

include/ginkgo/core/base/lin_op.hpp
fritzgoebel: I kind of agree to both that with the one creating a few and the other extracting a submatrix, they should not be used interchangibly and that the dense function name could be more precise. I'd vote for keeping them apart for now.

common/cuda_hip/matrix/csr_kernels.hpp.inc
```diff
@@ -1046,3 +1046,59 @@ __global__ __launch_bounds__(default_block_size) void inv_symm_permute_kernel(
...
+        const size_type col_offset,
+        const IndexType* __restrict__ source_row_ptrs,
+        const IndexType* __restrict__ source_col_idxs,
+        const ValueType* __restrict__ source_values,
+        const IndexType* __restrict__ result_row_ptrs,
+        IndexType* __restrict__ result_col_idxs,
+        ValueType* __restrict__ result_values)
+{
+    const auto tidx = thread::get_thread_id_flat();
+    // for (int tidx = threadIdx.x; tidx < num_rows; tidx += blockDim.x) {
```
fritzgoebel, line 1068:
```suggestion
```

reference/matrix/csr_kernels.cpp
```diff
@@ -633,6 +633,63 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
...
+    const matrix::Csr<ValueType, IndexType>* source, const span& row_span,
+    const span& col_span, Array<IndexType>* row_nnz)
+{
+    size_type res_row = 0;
+    for (size_type row = row_span.begin; row < row_span.end; ++row) {
+        for (size_type col = source->get_const_row_ptrs()[row];
+             col < source->get_const_row_ptrs()[row + 1]; ++col) {
+            if (source->get_const_col_idxs()[col] < col_span.end &&
+                source->get_const_col_idxs()[col] >= col_span.begin) {
+                row_nnz->get_data()[res_row]++;
```
fritzgoebel, line 648:
```suggestion
        for (size_type nnz = source->get_const_row_ptrs()[row];
             nnz < source->get_const_row_ptrs()[row + 1]; ++nnz) {
            if (source->get_const_col_idxs()[nnz] < col_span.end &&
                source->get_const_col_idxs()[nnz] >= col_span.begin) {
                row_nnz->get_data()[res_row]++;
```
Maybe nit-picking here, but I'd maybe call this `nnz` as the `col` is what's in the `col_idxs`.

omp/matrix/csr_kernels.cpp
```diff
@@ -713,6 +713,62 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
...
+{
+    const auto row_ptrs = source->get_const_row_ptrs();
+    const auto col_idxs = source->get_const_col_idxs();
+#pragma omp parallel for
+    for (size_type row = row_span.begin; row < row_span.end; ++row) {
+        for (size_type col = row_ptrs[row]; col < row_ptrs[row + 1]; ++col) {
+            if (col_idxs[col] >= col_span.begin &&
+                col_idxs[col] < col_span.end) {
+                row_nnz->get_data()[row - row_span.begin]++;
+            }
```
fritzgoebel, line 730:
```suggestion
        for (size_type nnz = row_ptrs[row]; nnz < row_ptrs[row + 1]; ++nnz) {
            if (col_idxs[nnz] >= col_span.begin &&
                col_idxs[nnz] < col_span.end) {
                row_nnz->get_data()[row - row_span.begin]++;
            }
```
Same here

fritzgoebel (changes requested):
LGTM mostly, but there's a lot of comments in the cuda tests that probably should be removed again before merging this.

fritzgoebel (approved):

include/ginkgo/core/base/lin_op.hpp
upsj: I don't think that's true? The `ValuesArray` parameter in the `Dense` constructor is taken by forwarding reference, `create` forwards, the array parameter used in `create_submatrix` is an prvalue, and the `Array` constructor turns the move-construction into a move-assignment, which moves when the executors match.

include/ginkgo/core/matrix/csr.hpp
```diff
@@ -981,6 +985,30 @@ class Csr : public EnableLinOp<Csr<ValueType, IndexType>>,
...
+          row_ptrs_{exec, std::move(row_ptrs)},
+          srow_(exec),
+          strategy_(strategy->copy())
+    {
+        GKO_ASSERT(row_ptrs_.get_num_elems() == size[0] + 1);
+        auto num_nnz = exec->copy_val_to_host(row_ptrs_.get_data() + size[0]);
+        values_ = Array<ValueType>(exec, num_nnz);
+        col_idxs_ = Array<IndexType>(exec, num_nnz);
+        this->make_srow();
+    }
```
upsj, line 1010: I don't think this is a common enough operation that we need a separate constructor for it.
The same can be accomplished with `create(exec, size, row_ptrs, Array<ValueType>(exec, nnz), Array<IndexType>(exec, nnz), strategy)`.
More in detail: For each matrix type, we have a set of size-based constructor and a single constructor from existing data. I think we should leave it at that separation, to create a uniform interface.

dpcpp/matrix/csr_kernels.dp.cpp
```diff
@@ -1625,6 +1625,27 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
     GKO_DECLARE_CSR_ADVANCED_SPMV_KERNEL);
 
 
+template <typename ValueType, typename IndexType>
+void calculate_nonzeros_per_row_in_span(
+    std::shared_ptr<const DefaultExecutor> exec,
+    const matrix::Csr<ValueType, IndexType>* source, const span& row_span,
+    const span& col_span, Array<IndexType>* row_nnz) GKO_NOT_IMPLEMENTED;
```
upsj, line 1632: the row-parallel implementations from CUDA and HIP can be directly translated to implementations for DPCPP, can you add that here?

core/matrix/csr.cpp
```diff
@@ -49,6 +49,8 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include "core/components/absolute_array.hpp"
 #include "core/components/fill_array.hpp"
+#include "core/components/prefix_sum.hpp"
+#include "core/components/reduce_array.hpp"
```
upsj, line 53: no longer used
```suggestion
```

core/matrix/csr.cpp
```diff
@@ -85,6 +90,8 @@ GKO_REGISTER_OPERATION(is_sorted_by_column_index,
                        csr::is_sorted_by_column_index);
 GKO_REGISTER_OPERATION(extract_diagonal, csr::extract_diagonal);
 GKO_REGISTER_OPERATION(fill_array, components::fill_array);
+GKO_REGISTER_OPERATION(reduce_add_array, components::reduce_add_array);
```
upsj, line 93:
```suggestion
```

core/matrix/csr.cpp
```diff
@@ -536,6 +543,29 @@ bool Csr<ValueType, IndexType>::is_sorted_by_column_index() const
...
+std::unique_ptr<Csr<ValueType, IndexType>>
+Csr<ValueType, IndexType>::create_submatrix(const gko::span& row_span,
+                                            const gko::span& column_span) const
+{
+    using Mat = Csr<ValueType, IndexType>;
+    auto exec = this->get_executor();
+    auto sub_mat_size = gko::dim<2>(row_span.length(), column_span.length());
+    Array<IndexType> row_ptrs(exec, row_span.length() + 1);
+    exec->run(csr::make_fill_array(
+        row_ptrs.get_data(), row_ptrs.get_num_elems(), zero<IndexType>()));
```
upsj, line 556: Why is this necessary? The following step should write to each entry except the last one, which gets filled by the prefix_sum

cuda/test/matrix/csr_kernels.cpp
```diff
@@ -926,4 +938,75 @@ TEST_F(Csr, OutplaceAbsoluteComplexMatrixIsEquivalentToRef)
 }
 
 
+TEST_F(Csr, CalculateNnzPerRowIsEquivalentToRef)
```
upsj, line 941: nit: this should point to submatrix computation somehow
```suggestion
TEST_F(Csr, CalculateSubmatrixNnzPerRowIsEquivalentToRef)
```

hip/test/matrix/csr_kernels.hip.cpp
```diff
@@ -937,4 +948,75 @@ TEST_F(Csr, OutplaceAbsoluteComplexMatrixIsEquivalentToRef)
...
+
+    GKO_ASSERT_ARRAY_EQ(row_nnz, drow_nnz);
+}
+
+
+TEST_F(Csr, ComputeSubmatrixIsEquivalentToRef)
+{
+    using Mtx = gko::matrix::Csr<>;
+    set_up_apply_data();
+
```
upsj, line 976: nit: AAA
```suggestion
```

hip/test/matrix/csr_kernels.hip.cpp
```diff
@@ -937,4 +948,75 @@ TEST_F(Csr, OutplaceAbsoluteComplexMatrixIsEquivalentToRef)
...
+        Mtx::create(this->hip, gko::dim<2>(rspan.length(), cspan.length()),
+                    std::move(drow_nnz));
+
+
+    gko::kernels::reference::csr::compute_submatrix(this->ref, this->mtx2.get(),
+                                                    rspan, cspan, smat1.get());
+    gko::kernels::hip::csr::compute_submatrix(this->hip, this->dmtx2.get(),
+                                              rspan, cspan, sdmat1.get());
+    auto hmat = Mtx::create(this->ref);
+    hmat->copy_from(sdmat1.get());
```
upsj, line 1000: Why is this necessary? ASSERT_MTX_NEAR already deals with device data

hip/test/matrix/csr_kernels.hip.cpp
```diff
@@ -937,4 +948,75 @@ TEST_F(Csr, OutplaceAbsoluteComplexMatrixIsEquivalentToRef)
...
+{
+    using Mtx = gko::matrix::Csr<>;
+    set_up_apply_data();
+
+    gko::span rspan{47, 81};
+    gko::span cspan{2, 31};
+    auto smat1 = this->mtx2->create_submatrix(rspan, cspan);
+    auto sdmat1 = this->dmtx2->create_submatrix(rspan, cspan);
+    auto hmat = Mtx::create(this->ref);
+    hmat->copy_from(sdmat1.get());
```
upsj, line 1016: same here

upsj (dismissed):
LGTM in both implementation and tests, really nice job there. The warp-parallel implementation would be nice to have, but not necessary in this PR from my side.
I only don't think creating a separate class requiring a CRTP parameter gives any advantage over directly adding the function to Csr though. Also, I would like to avoid the addition of another constructor to Csr, since we already have a way to do this with little more core.

Slaedr:
I would still like to know what the use case is for this functionality. I think that context is important in deciding whether this is a useful feature or not. For example, I think this might be useful if the `create_submatrix` could work on more general `IndexSet`s rather than `span`s.

upsj:
On the domain decomposition algorithms, is it normally possible to represent blocks and overlaps as contiguous ranges? In a straightforward 5pt stencil implementation, you can only get slices that way, and if it was stored in a blocked fashion, the overlaps probably need to consist of many separate ranges except for a single contiguous one?

Slaedr:
@pratikvn I guess you have fine-grain domain decomposition in mind, where you would have many subdomains in a single process? In that case, I suppose you plan to reorder the DOFs such that those within a subdomain are always contiguous?

pratikvn:
In general, overlaps are just extensions of the diagonal block by a few elements along the diagonal in either direction and then seeing the entire matrix with the rows to the left/right and columns to the top/bottom of the overlap elements. You would like to try to get the entire matrix (including the overlap) as a single block as that improves performance due to just a single apply call. 

@Slaedr, yes, both fine-grained and coarse grained should be possible. For fine grained implmentations, you probably would not use the core side implementations, but call the kernels themselves in your algorithms. But to compose the algorithms for testing and prototyping purposes, this would be nevertheless useful.

Slaedr:
I missed @upsj 's point about overlaps. If you need to extract overlapping sub-blocks, it would not be possible to have the subdomains contiguous in the global ordering, I believe. I guess that significantly decreases the current interface's utility for domain decomposition?

You might also consider having `IndexSet` in the interface, but implement it only for a single contiguous range for now, and expand it later if need be.

pratikvn:
@Slaedr , yes, I plan to add some functions which take in a `IndexSet`, but will probably do it another PR, as that will take some more time.

reference/test/matrix/csr_kernels.cpp
```diff
@@ -1624,4 +1624,81 @@ TYPED_TEST(CsrComplex, OutplaceAbsolute)
...
+          -1.0, 3.5
+           0.0, 0.0
+        */
+        auto ref4 = gko::initialize<Mtx>(
+            {I<T>{4.5, 7.5}, I<T>{4.5, 0.0}, I<T>{2.5, 0.0}, I<T>{-1.0, 3.5},
+             I<T>{0.0, 0.0}},
+            this->exec);
+
+        GKO_EXPECT_MTX_NEAR(sub_mat4.get(), ref4.get(), 0.0);
+    }
```
Slaedr, line 1700: Could you also add tests for the edge cases of extracting
- 0 x 0 matrix,
- 1 x 1 matrix,
- the full 7 x 5 matrix?

Slaedr (commented):
Except for the question of IndexSet vs span, I think this looks good. I agree with you that this is relevant for all matrix formats.

In general, I think IndexSet makes sense for this. You could add a small convenience function to get a vector of spans from an index set, and possibly also to also return the first span of an index set.

include/ginkgo/core/matrix/csr.hpp
```diff
@@ -763,7 +763,10 @@ class Csr : public EnableLinOp<Csr<ValueType, IndexType>>,
 
     std::unique_ptr<Diagonal<ValueType>> extract_diagonal() const override;
 
-    std::unique_ptr<absolute_type> compute_absolute() const override;
+    std::unique_ptr<Csr<ValueType, IndexType>> create_submatrix(
```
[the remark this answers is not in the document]
pratikvn: Yeah, I think we should be able to add the mixin later without breaking interface.

reference/test/matrix/csr_kernels.cpp
pratikvn: Good point

reference/test/matrix/csr_kernels.cpp
```diff
@@ -1624,4 +1624,105 @@ TYPED_TEST(CsrComplex, OutplaceAbsolute)
...
+        {
+            // clang-format off
+            I<T>{1.0, 3.0, 4.5, 0.0, 2.0}, // 0
+            I<T>{1.0, 0.0, 4.5, 7.5, 3.0}, // 1
+            I<T>{0.0, 3.0, 4.5, 0.0, 2.0}, // 2
+            I<T>{0.0,-1.0, 2.5, 0.0, 2.0}, // 3
+            I<T>{1.0, 0.0,-1.0, 3.5, 1.0}, // 4
+            I<T>{0.0, 1.0, 0.0, 0.0, 2.0}, // 5
+            I<T>{0.0, 3.0, 0.0, 7.5, 1.0}  // 6
+                                           // clang-format on
```
greole, line 1660: Minor: I think the ` format on`  statement should have the same alignment as  ` format off`
```suggestion
             // clang-format off
```

reference/test/matrix/csr_kernels.cpp
```diff
@@ -1624,4 +1624,105 @@ TYPED_TEST(CsrComplex, OutplaceAbsolute)
...
+            {
+                // clang-format off
+            I<T>{1.0, 3.0, 4.5, 0.0, 2.0}, // 0
+            I<T>{1.0, 0.0, 4.5, 7.5, 3.0}, // 1
+            I<T>{0.0, 3.0, 4.5, 0.0, 2.0}, // 2
+            I<T>{0.0,-1.0, 2.5, 0.0, 2.0}, // 3
+            I<T>{1.0, 0.0,-1.0, 3.5, 1.0}, // 4
+            I<T>{0.0, 1.0, 0.0, 0.0, 2.0}, // 5
+            I<T>{0.0, 3.0, 0.0, 7.5, 1.0}  // 6
+                                               // clang-format on
```
greole, line 1713: Minor: Same as above, I think the` format on` statement should have the same alignment as `format off`

```suggestion
            // clang-format off
            I<T>{1.0, 3.0, 4.5, 0.0, 2.0}, // 0
            I<T>{1.0, 0.0, 4.5, 7.5, 3.0}, // 1
            I<T>{0.0, 3.0, 4.5, 0.0, 2.0}, // 2
            I<T>{0.0,-1.0, 2.5, 0.0, 2.0}, // 3
            I<T>{1.0, 0.0,-1.0, 3.5, 1.0}, // 4
            I<T>{0.0, 1.0, 0.0, 0.0, 2.0}, // 5
            I<T>{0.0, 3.0, 0.0, 7.5, 1.0}  // 6
            // clang-format on
```

dpcpp/matrix/csr_kernels.dp.cpp
```diff
@@ -1625,6 +1625,111 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
...
+            }
+        }
+    }
+}
+
+GKO_ENABLE_DEFAULT_HOST(calc_nnz_in_span, calc_nnz_in_span);
+
+
+template <typename ValueType, typename IndexType>
+void compute_submat(size_type num_rows, size_type num_cols, size_type num_nnz,
```
greole, line 1654: Maybe here consistent names for the kernel implementation can be used, cf. with `compute_submatrix_idxs_and_vals` in the cuda version.

greole (commented):
Some very minor things from my side.

reference/test/matrix/csr_kernels.cpp
pratikvn: Yeah, my clang-format seems to do that. I am not sure why.

reference/test/matrix/csr_kernels.cpp
greole: Probably not, but the alignment of the commas and the sign is much nicer after turning clang-format off.

common/cuda_hip/matrix/csr_kernels.hpp.inc
```diff
@@ -1046,3 +1046,59 @@ __global__ __launch_bounds__(default_block_size) void inv_symm_permute_kernel(
...
+        ValueType* __restrict__ result_values)
+{
+    const auto tidx = thread::get_thread_id_flat();
+    if (tidx < num_rows) {
+        size_type res_nnz = result_row_ptrs[tidx];
+        for (size_type nnz = source_row_ptrs[row_offset + tidx];
+             nnz < source_row_ptrs[row_offset + tidx + 1]; ++nnz) {
+            if ((source_col_idxs[nnz] < (col_offset + num_cols) &&
+                 source_col_idxs[nnz] >= col_offset)) {
+                result_col_idxs[res_nnz] = source_col_idxs[nnz] - col_offset;
```
upsj, line 1074: A few naming nits to make this more easy to read:
```suggestion
    const auto res_row = thread::get_thread_id_flat();
    if (res_row < num_rows) {
        const auto src_row = res_row + row_offset;
        auto res_nnz = result_row_ptrs[res_row];
        for (auto nnz = source_row_ptrs[src_row]; nnz < source_row_ptrs[src_row + 1]; ++nnz) {
            const auto res_col = source_col_idxs[nnz] - col_offset;
            if (res_col < num_cols && res_col >= 0) {
                result_col_idxs[res_nnz] = res_col;
```

common/cuda_hip/matrix/csr_kernels.hpp.inc
```diff
@@ -1046,3 +1046,59 @@ __global__ __launch_bounds__(default_block_size) void inv_symm_permute_kernel(
...
+{
+    const auto tidx = thread::get_thread_id_flat() + row_span.begin;
+    if (tidx < row_span.end) {
+        nnz_per_row[tidx - row_span.begin] = zero<IndexType>();
+        for (size_type col = row_ptrs[tidx]; col < row_ptrs[tidx + 1]; ++col) {
+            if (col_idxs[col] >= col_span.begin &&
+                col_idxs[col] < col_span.end) {
+                nnz_per_row[tidx - row_span.begin]++;
+            }
+        }
```
upsj, line 1099: nit: clearer naming
```suggestion
    const auto src_row = thread::get_thread_id_flat() + row_span.begin;
    if (src_row < row_span.end) {
        IndexType nnz{};
        for (auto i = row_ptrs[src_row]; i < row_ptrs[src_row + 1]; ++i) {
            if (col_idxs[i] >= col_span.begin &&
                col_idxs[i] < col_span.end) {
                nnz++;
            }
        }
        nnz_per_row[src_row - row_span.begin] = nnz;
```

include/ginkgo/core/matrix/csr.hpp
```diff
@@ -763,7 +763,10 @@ class Csr : public EnableLinOp<Csr<ValueType, IndexType>>,
...
+        const gko::span& row_span, const gko::span& column_span) const;
+
+    std::unique_ptr<absolute_type> compute_absolute() const;
```
upsj, line 769: This seems to have gotten lost
```suggestion
    std::unique_ptr<absolute_type> compute_absolute() const override;
```

upsj (approved):
LGTM! I have only a few naming/readability/type nits for all implementations.

Slaedr (approved):
LGTM. I still don't quite know good uses for this, but I guess that's okay. It might at least come in handy for playing around (initial experiments) with stuff, and maybe some testing, as you said.

What would be interesting is creating sparse submatrix views. That could actually be used in production code for things like multi-physics problems, but that's a whole other story. We could have 4-array (-view) version of CSR format for that, for example.

PR #887, opened 2021-09-14
Remove CUDA < 9.2, add CUDA 11.4

upsj:
pretty self-explanatory :)

### DIFF
--- a/cuda/base/cusparse_bindings.hpp
+++ b/cuda/base/cusparse_bindings.hpp
@@ -850,10 +850,6 @@ inline void destroy(cusparseSpMatDescr_t descr)
 #endif
 
 
-// CUDA versions 9.2 and above have csrsm2.
-#if (defined(CUDA_VERSION) && (CUDA_VERSION >= 9020))
-
-
 inline csrsm2Info_t create_solve_info()
 {
     csrsm2Info_t info{};
@@ -866,27 +862,6 @@ inline void destroy(csrsm2Info_t info)
 {
     GKO_ASSERT_NO_CUSPARSE_ERRORS(cusparseDestroyCsrsm2Info(info));
 }
-
-
-// CUDA_VERSION<=9.1 do not support csrsm2.
-#elif (defined(CUDA_VERSION) && (CUDA_VERSION < 9020))
-
-
-inline cusparseSolveAnalysisInfo_t create_solve_info()
-{
-    cusparseSolveAnalysisInfo_t info{};
-    GKO_ASSERT_NO_CUSPARSE_ERRORS(cusparseCreateSolveAnalysisInfo(&info));
-    return info;
-}
-
-
-inline void destroy(cusparseSolveAnalysisInfo_t info)
-{
-    GKO_ASSERT_NO_CUSPARSE_ERRORS(cusparseDestroySolveAnalysisInfo(info));
-}
-
-
-#endif
 
 
 inline csrilu02Info_t create_ilu0_info()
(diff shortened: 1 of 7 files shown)


tcojean:
Maybe we can also skip 10.1 or some other intermediate version and instead add 11.2 as well? In order to spread out somewhat evenly the CUDA versions which we actually test.

upsj:
Looks like `gtest` is broken with GCC 11 (https://github.com/google/googletest/issues/3514), nvcc doesn't play nice with clang-12 as claimed in the documentation, and let's not even talk about `icpc` :)

upsj:
As @tcojean and I discussed, 16 GB for the `intel-basekit` isn't feasible, so we will most likely leave `icpx` out of this, since `icpc` has some issues again (https://community.intel.com/t5/Intel-C-Compiler/undefined-reference-to-std-failed-assertion-with-icpc-on-Fedora/td-p/1278857)

Slaedr:
Is the server running out of storage? Perhaps we could even consider leaving icpc out and only testing with icpx, since the former now officially has a sort of maintenance-only status anyway.

tcojean:
It's not only for the server but also that our containers are uploaded to dockerhub.

Other arguments for not bothering with this for now:
+ We have 8 CUDA containers currently and 2 CPUs, which would potentially be 160 GB of Intel software. I think at most we should include Intel in one or two containers overall (since we already have the oneAPI container anyway).
+ The newer Intel compilers seem to be fairly fragile at the moment, and we are not in a hurry to combine DPC++ and CUDA backend compilation.

I'm personally fine if Intel isn't added here, it can be a separate pull request whenever things are more stable. Another option, is to add the Intel support to an older container if that fixes some issues.

upsj:
Are you freaking kidding me? Even nvcc 11.4 + clang 9 + libstdc++ gcc-9 is broken. I give up, no clang for you. https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/1609871929

tcojean (approved):
LGTM

cuda/solver/common_trs_kernels.cuh
```diff
@@ -59,16 +59,13 @@ namespace solver {
 
 
 struct SolveStruct {
-    virtual void dummy() {}
+    virtual ~SolveStruct() {}
 };
```
yhmtsai, line 63: might not need this anymore?

yhmtsai (approved):

cuda/solver/common_trs_kernels.cuh
upsj: Right, I think in this case we don't even need a vtable, but I wanted to use a more canonical way to make sure we still emit one.

cuda/base/cusparse_bindings.hpp
```diff
@@ -850,10 +850,6 @@ inline void destroy(cusparseSpMatDescr_t descr)
 #endif
 
 
-// CUDA versions 9.2 and above have csrsm2.
```
pratikvn, line 853: So, we don't support CUDA 9.2 at all ? I thought this PR would stop using a CI job for 9.2 but we would still support 9.2 ? Maybe we can leave these #if defines, so if someone wants to compile with CUDA <9.2, they still can ?

upsj: That would very much complicate #764 (due to the support for multiple rhs/need for tranposition), so I would really prefer to just deprecate support for older versions. Compute capabilities down to 3.0 are still officially supported in CUDA 10, so we are not losing any hardware support by this.

pratikvn: Okay, then that is fine with me. Please also update the README.md, INSTALL.md and other pages where we currently say we support CUDA 9.0+. I guess now it would be CUDA 10.0+

upsj: 9.2 still works fine :)

PR #889, opened 2021-09-15
Always enable GKO_BUILD_REFERENCE for tests

upsj:
This PR enables always the reference executor for any tests.

As more and more functionality (`Dense::copy`, `Matrix::read`, ...) uses kernels for their executions, more and more `core` tests start to fail with all executors disabled. Since many of them were only working because some executor and Array functionality is available even with disabled executor, I think it makes sense to always enable the Reference executor for them.


tcojean:
While I like this, I think this is more related to testing requirements and some applications could want to build only CUDA/HIP/DPC++ (what they actually need to use) without reference also for library size optimizations.

Maybe we should simply force reference ON when compiling any form of test (and maybe benchmarks?).

yhmtsai:
Should we move the test to reference if they need reference executor?
When build other executor test, ginkgo will turn BUILD_REFERENCE on.

pratikvn (approved):
LGTM!

tcojean (approved):
LGTM

tcojean:
Maybe also mention in `INSTALL.md` that REFENCE is forced on whenever tests are enabled.

Slaedr (approved):
Makes sense.

upsj:
rebase!

yhmtsai:
can these tests not moved to reference?
I thought we use this way to handle this situation before

upsj:
This was prompted by changes in #886, which make all matrix format tests essentially useless in `core`, since they require kernel calls. I agree that the test framework itself also needs to adapt to these changes, but moving more and more tests to `reference` doesn't seem like the right thing to me.

yhmtsai (approved):
okay. If consider the compiling path, that core test requires reference also makes sense

PR #890, opened 2021-09-18
Create matrices from const data

upsj:
This enables the use of Array<const value_type> and uses it to create const LinOp pointers for matrices from existing const data.

TODO:
- [ ] Add tests
- [ ] Add example wrapping existing data (both const and non-const)


Slaedr (commented):
Great idea to "convert" const pointers into a const matrix object. I like `create_const` which ensures that const Array views can be used to create only const matrices. Even though we are casting away the constness, we are at least enforcing a different but very similar contract.

Perhaps some core tests would be good.

include/ginkgo/core/base/executor.hpp
```diff
@@ -900,7 +900,7 @@ class Executor : public log::EnableLogging<Executor> {
      *
      * @param ptr  pointer to the allocated memory block
      */
-    virtual void raw_free(void* ptr) const noexcept = 0;
+    virtual void raw_free(const void* ptr) const noexcept = 0;
```
pratikvn, line 903: I guess technically interface breaking ? But we can probably ignore it if expect no outside application to have derived from the `Executor` class yet ?

include/ginkgo/core/base/array.hpp
```diff
@@ -627,6 +655,41 @@ class copy_back_deleter<Array<T>> {
...
+            typename Array<const ValueType>::default_deleter{array.exec_};
+    } else {
+        new_deleter = [deleter](auto ptr) { deleter(ptr); };
+    }
+    Array<ValueType> result{array.exec_, array.num_elems_,
+                            const_cast<ValueType*>(array.data_.release()),
+                            new_deleter};
+    array.num_elems_ = 0;
+    array.data_ = typename Array<const ValueType>::data_manager{
+        nullptr, typename Array<const ValueType>::default_deleter{array.exec_}};
```
pratikvn, line 687: I think you can use `array.clear()` ? Then you can get rid of making this function a friend of `Array` as well

pratikvn (approved):
Great idea! Thanks for adding this. Can we maybe prioritize this ? Would clean up some of my code.

LGTM! Only missing a few tests.

include/ginkgo/core/base/executor.hpp
upsj: Now that I think about this again, I believe maybe a more appropriate solution would be to solve this mismatch on the `deleter` or `Array<const ...>` level, maybe with a special deleter for `const` data that casts `const`-ness away?

include/ginkgo/core/base/array.hpp
upsj: In general yes, that's the same code, but I'm still not sure about this one, as `array.clear()` should assume an intact Array, while at this point with `data_.release()` we left the array in an incomplete state. Is that thinking too strictly in terms of invariants?

upsj: The `friend` declaration will still be necessary to be able to steal the pointer via `release()`.

include/ginkgo/core/base/executor.hpp
pratikvn: Yes, I think if that works, that might be better solution.

include/ginkgo/core/base/array.hpp
pratikvn: I think with `data_.reset(nullptr)` in `clear` you would do the same that you do here ? But I agree you need the `friend` for the `release()`

include/ginkgo/core/base/executor.hpp
Slaedr: I think this much interface break should be okay. We should just document this in the release notes whenever this is released. But I agree that dealing with this in Array's internal deleter might be a bit better.

include/ginkgo/core/base/array.hpp
upsj: The main difference is that `data.reset(nullptr)` calls the deleter, while `release()` doesn't do anything - we need the latter behavior.

pratikvn (approved):
LGTM!

core/test/matrix/coo.cpp
```diff
@@ -152,6 +152,26 @@ TYPED_TEST(Coo, CanBeCreatedFromExistingData)
...
 
+TYPED_TEST(Coo, CanBeCreatedFromExistingConstData)
+{
+    using value_type = typename TestFixture::value_type;
+    using index_type = typename TestFixture::index_type;
+    value_type values[] = {1.0, 2.0, 3.0, 4.0};
+    index_type col_idxs[] = {0, 1, 1, 0};
+    index_type row_idxs[] = {0, 0, 1, 2};
+
+    auto mtx = gko::matrix::Coo<value_type, index_type>::create_const(
```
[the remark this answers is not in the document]
upsj: It could potentially be handled, but I would prefer to be more explicit, since `create` is a variadic template, which makes for unpleasant error messages if something doesn't fit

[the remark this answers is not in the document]
pratikvn: I think the overload here would make it easier to switch between const and non const arrays without having to "re-write" the code. So, maybe have both ?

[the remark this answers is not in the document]
upsj: Good point, I need to think this through a bit more.

include/ginkgo/core/base/array.hpp
```diff
@@ -90,14 +101,25 @@ class Array {
...
      * The default deleter type used by Array.
+     * For const values, we need to use a deleter that takes a const array.
+     * This does not make much sense (read-only uninitialized memory), but helps
+     * keep the interface simple.
      */
-    using default_deleter = executor_deleter<value_type[]>;
+    using default_deleter = std::conditional_t<
+        std::is_const<value_type>::value,
+        const_executor_deleter<std::remove_const_t<value_type>[]>,
+        executor_deleter<value_type[]>>;
```
MarcelKoch, line 111:
```suggestion
    using default_deleter = executor_deleter<value_type[]>;
```

include/ginkgo/core/base/array.hpp
```diff
@@ -90,14 +101,25 @@ class Array {
 
     /**
      * The default deleter type used by Array.
+     * For const values, we need to use a deleter that takes a const array.
+     * This does not make much sense (read-only uninitialized memory), but helps
+     * keep the interface simple.
```
MarcelKoch, line 106: If we assume that an const array can't own its data, this would not be necessary.
```suggestion
```

include/ginkgo/core/base/array.hpp
```diff
@@ -627,6 +710,40 @@ class copy_back_deleter<Array<T>> {
...
+ */
+template <typename ValueType>
+Array<ValueType> array_const_cast(Array<const ValueType>&& array)
+{
+    auto deleter = array.data_.get_deleter();
+    std::function<void(ValueType[])> new_deleter;
+    if (array.is_owning()) {
+        new_deleter = typename Array<ValueType>::default_deleter{array.exec_};
+    } else {
+        new_deleter = [deleter](auto ptr) { deleter(ptr); };
```
MarcelKoch, line 734: Perhaps
```suggestion
        new_deleter = typename Array<ValueType>::view_deleter{};
```

include/ginkgo/core/base/array.hpp
```diff
@@ -627,6 +710,40 @@ class copy_back_deleter<Array<T>> {
...
+ *
+ * @param array  the array to be cast. It will be empty with default deleter
+ *               afterwards.
+ * @returns a non-const array wrapping the same pointer and deleter on the same
+ *          executor as `array`.
+ */
+template <typename ValueType>
+Array<ValueType> array_const_cast(Array<const ValueType>&& array)
+{
+    auto deleter = array.data_.get_deleter();
```
MarcelKoch, line 729: See below
```suggestion
```

include/ginkgo/core/base/array.hpp
```diff
@@ -627,6 +710,40 @@ class copy_back_deleter<Array<T>> {
...
+        new_deleter = typename Array<ValueType>::default_deleter{array.exec_};
+    } else {
+        new_deleter = [deleter](auto ptr) { deleter(ptr); };
+    }
+    Array<ValueType> result{array.exec_, array.num_elems_,
+                            const_cast<ValueType*>(array.data_.release()),
+                            new_deleter};
+    array.num_elems_ = 0;
+    array.data_ = typename Array<const ValueType>::data_manager{
+        nullptr, typename Array<const ValueType>::default_deleter{array.exec_}};
```
MarcelKoch, line 741: If a const array doesn't own the data, we could use the view deleter directly. This prevents compiler errors from trying to use `executor_deleter<const T>`.
```suggestion
        nullptr, typename Array<const ValueType>::const_view_deleter{}};
```

include/ginkgo/core/base/executor.hpp
```diff
@@ -1101,6 +1101,66 @@ class executor_deleter<T[]> {
 };
 
 
+/**
+ * This is a deleter that uses an executor's `free` method to deallocate const
+ * data. This is mainly necessary to make Array<const ValueType> work.
+ *
+ * @tparam T  the type of object being deleted
```
MarcelKoch, line 1108: Perhaps explicitly state, that `T` can't have a `const` qualifier?
```suggestion
 * @tparam T  the type of object being deleted, without const qualifier
```

include/ginkgo/core/matrix/coo.hpp
```diff
@@ -258,6 +258,31 @@ class Coo : public EnableLinOp<Coo<ValueType, IndexType>>,
...
+     * @param exec  the executor to create the matrix on
+     * @param size  the dimensions of the matrix
+     * @param values  the value array of the matrix
+     * @param col_idxs  the column index array of the matrix
+     * @param row_ptrs  the row index array of the matrix
+     * @returns A smart pointer to the constant matrix wrapping the input arrays
+     *          (if they reside on the same executor as the matrix) or a copy of
+     *          these arrays on the correct executor.
+     */
+    static std::unique_ptr<const Coo> create_const(
```
MarcelKoch, line 273: Should `create_const` be abstracted into an interface, the same way it's done for `create`? (Same goes for all other matrix types)

core/test/base/array.cpp
```diff
@@ -591,4 +592,106 @@ TYPED_TEST(Array, MoveArrayToView)
...
+TYPED_TEST(Array, ArrayConstCastWrapsDefaultDeleter)
+{
+    auto data = this->exec->template alloc<TypeParam>(10);
+    gko::Array<const TypeParam> array(this->exec, 10, data);
+    ASSERT_TRUE(array.is_owning());
+
+    auto mutable_array = gko::detail::array_const_cast(std::move(array));
+
+    ASSERT_TRUE(mutable_array.is_owning());
+}
```
MarcelKoch, line 694: If we assume an const array doesn't own data, this test needs to be removed.
(See below)
```suggestion
```

core/test/matrix/coo.cpp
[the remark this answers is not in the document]
MarcelKoch: I think the problem would be the return type. The current `create_const` method returns a `uniqe_ptr<const Coo>`, which the normal `create` doesn't. I don't think it would be possible to get the normal `create` to return both types. So, I'm fine with being a bit more explicit.

core/test/matrix/diagonal.cpp
```diff
@@ -114,6 +114,20 @@ TYPED_TEST(Diagonal, CanBeCreatedFromExistingData)
 }
 
 
+TYPED_TEST(Diagonal, CanBeCreatedFromExistingConstData)
+{
+    using Diag = typename TestFixture::Diag;
```
MarcelKoch, line 119:
```suggestion
```

core/test/matrix/fbcsr.cpp
```diff
@@ -404,6 +404,33 @@ TYPED_TEST(Fbcsr, CanBeCreatedFromExistingData)
 }
 
 
+TYPED_TEST(Fbcsr, CanBeCreatedFromExistingConstData)
+{
+    using Mtx = typename TestFixture::Mtx;
```
MarcelKoch, line 409:
```suggestion
```

core/test/matrix/fbcsr.cpp
```diff
@@ -404,6 +404,33 @@ TYPED_TEST(Fbcsr, CanBeCreatedFromExistingData)
...
+    using value_type = typename TestFixture::value_type;
+    using index_type = typename TestFixture::index_type;
+    using size_type = gko::size_type;
+    const int bs = this->fbsample.bs;
+    const size_type nbrows = this->fbsample.nbrows;
+    const size_type nbcols = this->fbsample.nbcols;
+    const size_type bnnz = this->fbsample.nbnz;
+    std::unique_ptr<Mtx> refmat = this->fbsample.generate_fbcsr();
```
MarcelKoch, line 417:
```suggestion
    auto refmat = this->fbsample.generate_fbcsr();
```

MarcelKoch (changes requested):
The current code seems to assume that an array of `const T` might be owning its data. I don't think that is particular useful. IMO, an array of type `Array<const T>` should always be a view. 

Other than that, I think this is a really nice addition.

include/ginkgo/core/matrix/coo.hpp
upsj: This might require quite a lot of plumbing code due to the different sets of parameters the different matrix types require - variadic parameter set, conditionally transforming parameters via `array_const_cast`. Do you think that would be worth the effort?

MarcelKoch: I'm not sure if I follow here. I had the exact same setup as with `EnableCreateMethod` in mind. For example, the interface could be just
```c++
template <typename ConcreteType>
class EnableCreateConstMethod {
public:
    template <typename... Args>
    static std::unique_ptr<const ConcreteType> create_const(Args&&... args)
    {
        return std::unique_ptr<const ConcreteType>(
            new ConcreteType(std::forward<Args>(args)...));
    }
};
```
and then the same implementation for the `create_const` would be moved to the constructor, e.g.
```c++
Coo(std::shared_ptr<const Executor> exec, const dim<2>& size,
    Array<const ValueType>&& values, Array<const IndexType>&& col_idxs,
    Array<const IndexType>&& row_idxs)
: Coo(std::move(exec), size, gko::detail::array_const_cast(std::move(values)),
        gko::detail::array_const_cast(std::move(col_idxs)),
        gko::detail::array_const_cast(std::move(row_idxs)))
{}
```
I don't think that would be much of an effort, except that this would need to be changed for every matrix format.

MarcelKoch: Ok, I get the point. Before this addition, the templated constructor would not work with const arrays, so you would get a compiler error if you tried that before (at least that's what I hope). I haven't tested it, but with the changes to `Array` would the templated constructor now compile with const arrays? If so, then we already have that problem.

MarcelKoch: Just checked, and even with the changes to `Array` you can't use the templated constructor with const arrays. So then we don't add the interface, at least not until we have a neat way to introduce it.

upsj: This brings up some good points - maybe we should instead introduce a `ConstArrayView`-like type that doesn't have any ownership semantics or assignment operators, and use that instead of `Array<const>`? It would look the same for users, but avoid any workarounds to make Array<const> behave consistently?

include/ginkgo/core/base/array.hpp
```diff
@@ -65,6 +65,86 @@ void convert_data(std::shared_ptr<const Executor> exec, size_type size,
...
+                   const ValueType* data)
+        : exec_{std::move(exec)}, num_elems_{num_elems}, data_{data}
+    {}
+
+    /*
+     * To avoid any collisions with the value semantics of normal arrays,
+     * disable assignment altogether.
+     */
+    ConstArrayView& operator=(const ConstArrayView&) = delete;
+    ConstArrayView& operator=(ConstArrayView&&) = delete;
```
Slaedr, line 102: We could still allow move-assignment and nullify the internal pointer of the passed argument.

include/ginkgo/core/base/array.hpp
```diff
@@ -65,6 +65,86 @@ void convert_data(std::shared_ptr<const Executor> exec, size_type size,
...
+
+    ConstArrayView(const ConstArrayView&) = default;
+    /* Move-construction uses copy-construction to preserve executors. */
+    ConstArrayView(ConstArrayView&& other) : ConstArrayView{other} {}
```
Slaedr, line 106: I think we should preserve the normal meaning of move constructors and nullify the argument's internal pointer.

include/ginkgo/core/base/array.hpp
```diff
@@ -627,6 +747,27 @@ class copy_back_deleter<Array<T>> {
...
+ * @param view  the array view to be cast.
+ * @returns a non-const non-owning array wrapping the same pointer on the same
+ *          executor as `view`.
+ */
+template <typename ValueType>
+Array<ValueType> array_const_cast(ConstArrayView<ValueType> view)
+{
+    return Array<ValueType>::view(
+        view.get_executor(), view.get_num_elems(),
+        const_cast<ValueType*>(view.get_const_data()));
```
Slaedr, line 767: While it looks sacrilegious, I agree that you've designed the whole thing well. The only I concern I have is - when exactly this is perfectly legal? For example, if I pass a statically allocated array to a view, and then call this function, I believe that's UB? IIRC, this is only well-defined if the pointer-to-const points to a block of dynamically allocated memory.

Slaedr (commented):
Excellent work! I have one major concern about the `array_const_cast` below, apart from which this LGTM.

include/ginkgo/core/base/array.hpp
upsj: `https://en.cppreference.com/w/cpp/language/const_cast
> Modifying a const object through a non-const access path and referring to a volatile object through a non-volatile glvalue results in undefined behavior.

It's not the `const_cast` that results in undefined behavior, but a subsequent write. Imagine writing to a constant stored in read-only memory, that would give you some kind of memory access violation. Basically from my understanding, for dynamically allocated memory, this doesn't matter at all, and for our use case (returning const matrices), this doesn't even occur.

PR #891, opened 2021-09-18
Enable Hip kernels for batch solvers

Slaedr:
Enables Hip kernels and adds tests for batch BiCGStab, GMRES, CG and Richardson. Batch IDR will be ported in a separate PR as it needs a bit more work to abstract the device-side random number generation.

In addition,
- tests that compare with reference for a partly randomized matrix have been added to Cuda tests (and the new Hip tests)
- some batch dense kernels for Cuda that depend on a 'converged' bitset for multiple RHS have been removed.

Currently, no performance analysis has been done.


Slaedr:
format!

Slaedr:
format!

PR #894, opened 2021-09-30
Scaling example

fritzgoebel:
This PR adds an example enabling users to perform simple strong and weak scaling studies for the distributed SpMV.
The example can be run with 2D (5-pt or 9-pt) and 3D (7-pt or 27-pt) stencil matrices. All matrix values are 1 for now as the result values don't really matter for this example.

### DIFF
--- a/examples/distributed-spmv-scaling/distributed-spmv-scaling.cpp
+++ b/examples/distributed-spmv-scaling/distributed-spmv-scaling.cpp
@@ -59,13 +59,13 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 template <typename ValueType, typename IndexType>
 gko::matrix_data<ValueType, IndexType> generate_2d_stencil(
-    const IndexType dp, const gko::distributed::communicator &comm,
+    const IndexType dp, std::shared_ptr<gko::mpi::communicator> comm,
     bool restricted, bool strong_scaling)
 {
-    const auto mat_size = strong_scaling ? dp * dp : dp * dp * comm.size();
-    const auto rows_per_rank = gko::ceildiv(mat_size, comm.size());
-    const auto start = rows_per_rank * comm.rank();
-    const auto end = gko::min(rows_per_rank * (comm.rank() + 1), mat_size);
+    const auto mat_size = strong_scaling ? dp * dp : dp * dp * comm->size();
+    const auto rows_per_rank = gko::ceildiv(mat_size, comm->size());
+    const auto start = rows_per_rank * comm->rank();
+    const auto end = gko::min(rows_per_rank * (comm->rank() + 1), mat_size);
 
     auto A_data =
         gko::matrix_data<ValueType, IndexType>(gko::dim<2>{mat_size, mat_size});
@@ -99,14 +99,14 @@ gko::matrix_data<ValueType, IndexType> generate_2d_stencil(
  */
 template <typename ValueType, typename IndexType>
 gko::matrix_data<ValueType, IndexType> generate_3d_stencil(
-    const IndexType dp, const gko::distributed::communicator &comm,
+    const IndexType dp, std::shared_ptr<gko::mpi::communicator> comm,
     bool restricted, bool strong_scaling)
 {
     const auto mat_size =
-        strong_scaling ? dp * dp * dp : dp * dp * dp * comm.size();
-    const auto rows_per_rank = gko::ceildiv(mat_size, comm.size());
-    const auto start = rows_per_rank * comm.rank();
-    const auto end = gko::min(rows_per_rank * (comm.rank() + 1), mat_size);
+        strong_scaling ? dp * dp * dp : dp * dp * dp * comm->size();
+    const auto rows_per_rank = gko::ceildiv(mat_size, comm->size());
+    const auto start = rows_per_rank * comm->rank();
+    const auto end = gko::min(rows_per_rank * (comm->rank() + 1), mat_size);
 
     auto A_data =
         gko::matrix_data<ValueType, IndexType>(gko::dim<2>{mat_size, mat_size});
@@ -137,9 +137,9 @@ gko::matrix_data<ValueType, IndexType> generate_3d_stencil(
 }
 
 
-int main(int argc, char *argv[])
+int main(int argc, char* argv[])
 {
-    MPI_Init(&argc, &argv);
+    const auto fin = gko::mpi::init_finalize(argc, argv);
     // Use some shortcuts. In Ginkgo, vectors are seen as a gko::matrix::Dense
     // with one column/one row. The advantage of this concept is that using
     // multiple vectors is a now a natural extension of adding columns/rows are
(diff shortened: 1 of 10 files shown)


fritzgoebel:
format!

tcojean:
That's a really useful addition!

I have a few quick comments/questions, not only to you but also to people who know more about the MPI backend
+ That only works on a single node right, since you directly pass the rank to the executor?
   - With MPI, isn't there a mode where you use `CUDA_VISIBLE_DEVICES` or similar variables so that each MPI process only see one executor, which would always be number 0, but different between every MPI process?
   - Should OpenMP be disabled (until we can bind threads or we need to mention somewhere the user needs to take care of that through MPI/OpenMP variables
+ What safety additions are required so that exceptions don't get thrown all the time, device allocation, probably some MPI interaction calls would benefit from `try/catch`, or is that managed at a lower level?

pratikvn:
@fritzgoebel, I think it might make sense to add this to an updated version of this branch, `mpi-base-dist-mat`, which also runs CI for MPI ? 

@tcojean ,

1. Yes, I think you can have a local_rank getter, which should be sufficient for 1MPI rank to 1GPU association. See for example: https://github.com/ginkgo-project/ginkgo/blob/mpi-base-dist-mat/examples/distributed-solver/distributed-solver.cpp#L95

2. Disabling OpenMP is a good idea and I think we should do that until we fix/update the thread binding.

3. I guess you mean each rank throwing an exception ? MPI calls are wrapped in `GKO_ASSERT_NO_MPI_ERRORS` to those should be captured and thrown properly.

tcojean:
> 1. Yes, I think you can have a local_rank getter, which should be sufficient for 1MPI rank to 1GPU association. See for example: https://github.com/ginkgo-project/ginkgo/blob/mpi-base-dist-mat/examples/distributed-solver/distributed-solver.cpp#L95
Neat, using this would fix the issue in that case then

> 2. Disabling OpenMP is a good idea and I think we should do that until we fix/update the thread binding.

It's maybe also fine as long as people are told to only use one process per machine, but it's indeed not going to be as powerful as people could expect it to be (only that mode should be able to work for now I think).

> 3. I guess you mean each rank throwing an exception ? MPI calls are wrapped in `GKO_ASSERT_NO_MPI_ERRORS` to those should be captured and thrown properly.

Yes, they are thrown but then that means most Ginkgo API calls need to be surrounded by the user in `try/catch` blocks so that the exceptions are caught properly by the main function instead of thrown into the wild? https://github.com/ginkgo-project/ginkgo/blob/mpi-base-dist-mat/include/ginkgo/core/base/exception_helpers.hpp#L335

We have the same issue with memory allocations for example

Quick mention, it looks like in some places the check are missing: https://github.com/ginkgo-project/ginkgo/blob/mpi-base-dist-mat/mpi/base/bindings.hpp#L87

fritzgoebel:
format!

examples/distributed-spmv-scaling/distributed-spmv-scaling.cpp
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    const auto executor_string = argc >= 2 ? argv[1] : "reference";
+    std::map<std::string, std::function<std::shared_ptr<gko::Executor>()>>
+        exec_map{
+            {"omp", [] { return gko::OmpExecutor::create(); }},
+            {"cuda",
+             [local_rank] {
+                 return gko::CudaExecutor::create(
+                     local_rank, gko::ReferenceExecutor::create(), true);
```
Slaedr, line 187: Continuing along @tcojean 's questions, right now this might not work when there's only device 0 on each process, right? Even if the device ID is global and independent of which process is assigned what resource by the scheduler, there's no guarantee device ID 3 would be accessible to rank 3.

Would it not be better to either (1) always assume device 0, assuming the user maps one GPU to one process in the job scheduler (`--gpus-per-task=1` on slurm, I think), or (2) use hwloc or something to access the number of GPUs available to this task and then decide `device_id`. Option 2 becomes useful only when we want to support multiple GPUs per process, so maybe we ignore that for now.

examples/distributed-spmv-scaling/distributed-spmv-scaling.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+    // build partition: uniform number of rows per rank
+    gko::Array<gko::int64> ranges_array{
+        exec->get_master(), static_cast<gko::size_type>(comm->size() + 1)};
+    for (int i = 0; i < comm->size(); i++) {
+        ranges_array.get_data()[i] = i * rows_per_rank;
+    }
+    ranges_array.get_data()[comm->size()] = mat_size;
+    auto partition = gko::share(
+        part_type::build_from_contiguous(exec->get_master(), ranges_array));
```
Slaedr, line 228: We should probably make it explicit in the naming that this is a row partition. Either `gko::distributed::RowPartition` for the class name, or `part_type::build_from_contiguous_rows`, something like that.

Slaedr (commented):
Looks good. I have a question about the device IDs and ranks though.

examples/distributed-spmv-scaling/distributed-spmv-scaling.cpp
tcojean: I'm not sure what is best to take care of that right now, but I think that's a fairly important point and a fairly annoying one. AFAIK, both situations can happen where only one device (number 0) is available for each MPI process (as you say using some job scheduler option), or each process sees the full machine and thus GPUs need to be split like now.

We probably should work in both situations? But what would be a convenient way? Should we have a simple function somewhere doing `max(available_devices, local_rank)`?

Slaedr: According to [this](https://docs.olcf.ornl.gov/systems/summit_user_guide.html#gpu-compute-modes), on Summit, the default setting is to have one GPU per process. On Spock, the default setting is all GPUs see all processes. I guess we need to support both, or document strongly that we need users to set the correct scheduler settings.

pratikvn: I think that is regarding assignment rather than what the process can view. With Summit, AFAIK, it is a bit different with resource sets. I think you always need to specify how many GPUs you need. And the numbering depends on how many resource sets you ask for. By default, I think it is set to 1, so the GPUs will be numbered from 0 to 5.

examples/distributed-spmv-scaling/distributed-spmv-scaling.cpp
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    return A_data;
+}
+
+
+int main(int argc, char* argv[])
+{
+    const auto fin = gko::mpi::init_finalize(argc, argv);
+    // Use some shortcuts. In Ginkgo, vectors are seen as a gko::matrix::Dense
```
greole, line 143: This comment, starting at "In Gingko ..." is a bit convoluted, may be leave it out all together?

greole (commented):
Here are some very minor typos I found.

MarcelKoch:
Superseded by #1204

PR #895, opened 2021-10-01
Add transposition Linop

yhmtsai:
This PR adds transposition LinOp such that build the transposition in lazy way or now.
When the behavior is lazy, building transposition will happen in first apply, get_transposition or prepare_transposition.

TODO:
- [ ] is ConjTranspose needed? use template or parameter to control?


include/ginkgo/core/base/transposition.hpp
```diff
@@ -0,0 +1,164 @@
...
+     *
+     * @return the operator
+     */
+    std::shared_ptr<const LinOp> get_operator() const noexcept
+    {
+        return operator_;
+    }
+
+    /**
+     * Returns transposition of operator
```
Slaedr, line 84:
```suggestion
     * Returns transpose of operator
```

include/ginkgo/core/base/transposition.hpp
```diff
@@ -0,0 +1,164 @@
...
+     *
+     * @return transposition of operator
+     */
+    std::shared_ptr<const LinOp> get_transposition() const noexcept
```
Slaedr, line 88:
```suggestion
    std::shared_ptr<const LinOp> get_transpose() const noexcept
```

Slaedr (commented):
This might be quite useful; nice work. But as usual, I want to be careful about introducing a new interface. Do you have a specific use case for this?

yhmtsai:
@Slaedr the original idea is to make IC always return two LinOp.
Currently, IC will return one or two according to `both_factor` setting
https://github.com/ginkgo-project/ginkgo/discussions/681

PR #896, opened 2021-10-04
use sycl-std 2017 after dpcpp2021.4 to propagate subgroup

yhmtsai:
dpcpp releases 2021.4 and they use sycl2020 as default standard recently.
However, ginkgo sets the subgroup size in the function and the old dpcpp will propagate them, which is not allowed in sycl2020 anymore.
Before we change the code to fit the new standard, we use -sycl-std=2017 as workaround.

TODO
[ ] make sure it passes the dpcpp test in ci


upsj (approved):

yhmtsai:
rebase!

dpcpp/CMakeLists.txt
```diff
@@ -66,8 +66,11 @@ target_sources(ginkgo_dpcpp
 ginkgo_compile_features(ginkgo_dpcpp)
 target_compile_definitions(ginkgo_dpcpp PRIVATE GKO_COMPILING_DPCPP)
 
+# -sycl-std=2017 is temporary workaround after 2021.4 to propagate subgroup setting correctly
+set(GINKGO_DPCPP_FLAGS "-sycl-std=2017")
```
tcojean, line 70: Wouldn't it be better to put this into CMakeLists.txt:70, like what we do for the fp-model.

yhmtsai: After discussion, the sycl-std also affects some USM/buffer usage, so it is global.
Move to `GINKGO_COMPILER_FLAGS`

tcojean (approved):

PR #897, opened 2021-10-05
CUDA >= 11.0 support for int64 indices for spmv and spgemm

fritzgoebel:
For the cusparse spmv and spgemm, moves the is_supported calls to cuda versions < 11.0

### DIFF
--- a/cuda/matrix/csr_kernels.cu
+++ b/cuda/matrix/csr_kernels.cu
@@ -314,8 +314,9 @@ void spmv(std::shared_ptr<const CudaExecutor> exec,
 
 #if defined(CUDA_VERSION) && (CUDA_VERSION < 11000)
             // TODO: add implementation for int64 and multiple RHS
-            if (cusparse::is_supported<ValueType, IndexType>::value)
+            if (!cusparse::is_supported<ValueType, IndexType>::value) {
                 GKO_NOT_IMPLEMENTED;
+            }
 
             auto descr = cusparse::create_mat_descr();
             auto row_ptrs = a->get_const_row_ptrs();
@@ -399,8 +400,9 @@ void advanced_spmv(std::shared_ptr<const CudaExecutor> exec,
         if (b->get_stride() != 1 || c->get_stride() != 1) GKO_NOT_IMPLEMENTED;
 
 #if defined(CUDA_VERSION) && (CUDA_VERSION < 11000)
-        if (cusparse::is_supported<ValueType, IndexType>::value)
+        if (!cusparse::is_supported<ValueType, IndexType>::value) {
             GKO_NOT_IMPLEMENTED;
+        }
 
         auto descr = cusparse::create_mat_descr();
         auto row_ptrs = a->get_const_row_ptrs();
@@ -509,8 +511,9 @@ void spgemm(std::shared_ptr<const CudaExecutor> exec,
     auto& c_vals_array = c_builder.get_value_array();
 
 #if defined(CUDA_VERSION) && (CUDA_VERSION < 11000)
-    if (cusparse::is_supported<ValueType, IndexType>::value)
+    if (!cusparse::is_supported<ValueType, IndexType>::value) {
         GKO_NOT_IMPLEMENTED;
+    }
 
     auto a_descr = cusparse::create_mat_descr();
     auto b_descr = cusparse::create_mat_descr();
@@ -675,8 +678,9 @@ void advanced_spgemm(std::shared_ptr<const CudaExecutor> exec,
     auto c_row_ptrs = c->get_row_ptrs();
 
 #if defined(CUDA_VERSION) && (CUDA_VERSION < 11000)
-    if (cusparse::is_supported<ValueType, IndexType>::value)
+    if (!cusparse::is_supported<ValueType, IndexType>::value) {
         GKO_NOT_IMPLEMENTED;
+    }
 
     matrix::CsrBuilder<ValueType, IndexType> c_builder{c};
     auto& c_col_idxs_array = c_builder.get_col_idx_array();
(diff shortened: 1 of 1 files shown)


cuda/matrix/csr_kernels.cu
```diff
@@ -303,59 +303,58 @@ void spmv(std::shared_ptr<const CudaExecutor> exec,
             syn::value_list<int>(), syn::type_list<>(), exec, a, b, c);
     } else if (a->get_strategy()->get_name() == "sparselib" ||
                a->get_strategy()->get_name() == "cusparse") {
-        if (cusparse::is_supported<ValueType, IndexType>::value) {
```
yhmtsai, line 306: Does cusparse support int64 in those routines we use?
If it is, another way is to update cusparse::is_supported according to the cuda version.

upsj: I guess the difference here is that cuSPARSE itself will throw an exception of a combination is not supported with the Generic API, while in older versions, you had differently named functions for different types, and no 64 bit support at all. As long as we have both types of routines in Ginkgo, we need to differentiate between them.

cuda/matrix/csr_kernels.cu
```diff
@@ -396,55 +395,50 @@ void advanced_spmv(std::shared_ptr<const CudaExecutor> exec,
         }
     } else if (a->get_strategy()->get_name() == "sparselib" ||
                a->get_strategy()->get_name() == "cusparse") {
-        if (cusparse::is_supported<ValueType, IndexType>::value) {
-            // TODO: add implementation for int64 and multiple RHS
-            if (b->get_stride() != 1 || c->get_stride() != 1)
-                GKO_NOT_IMPLEMENTED;
+        // TODO: add implementation for int64 and multiple RHS
```
upsj, line 398:
```suggestion
        // TODO: add implementation for multiple RHS
```
The rest should be addressed with #905

cuda/matrix/csr_kernels.cu
```diff
@@ -499,113 +492,111 @@ void spgemm(std::shared_ptr<const CudaExecutor> exec,
...
-        auto info = cusparse::create_spgemm_info();
-        // allocate buffer
-        size_type buffer_size{};
-        cusparse::spgemm_buffer_size(
-            handle, m, n, k, &alpha, a_descr, a_nnz, a_row_ptrs, a_col_idxs,
-            b_descr, b_nnz, b_row_ptrs, b_col_idxs, null_value, d_descr,
-            zero_nnz, null_index, null_index, info, buffer_size);
-        Array<char> buffer_array(exec, buffer_size);
-        auto buffer = buffer_array.get_data();
+    if (cusparse::is_supported<ValueType, IndexType>::value)
```
upsj, line 512:
```suggestion
    if (!cusparse::is_supported<ValueType, IndexType>::value)
```

cuda/matrix/csr_kernels.cu
```diff
@@ -396,55 +395,50 @@ void advanced_spmv(std::shared_ptr<const CudaExecutor> exec,
...
-            auto row_ptrs = a->get_const_row_ptrs();
-            auto col_idxs = a->get_const_col_idxs();
-            cusparse::spmv(exec->get_cusparse_handle(),
-                           CUSPARSE_OPERATION_NON_TRANSPOSE, a->get_size()[0],
-                           a->get_size()[1], a->get_num_stored_elements(),
-                           alpha->get_const_values(), descr,
-                           a->get_const_values(), row_ptrs, col_idxs,
-                           b->get_const_values(), beta->get_const_values(),
-                           c->get_values());
+        if (cusparse::is_supported<ValueType, IndexType>::value)
```
upsj, line 402:
```suggestion
        if (!cusparse::is_supported<ValueType, IndexType>::value)
```

cuda/matrix/csr_kernels.cu
```diff
@@ -303,59 +303,58 @@ void spmv(std::shared_ptr<const CudaExecutor> exec,
...
-                auto col_idxs = a->get_const_col_idxs();
-                cusparse::spmv(handle, CUSPARSE_OPERATION_NON_TRANSPOSE,
-                               a->get_size()[0], a->get_size()[1],
-                               a->get_num_stored_elements(), &alpha, descr,
-                               a->get_const_values(), row_ptrs, col_idxs,
-                               b->get_const_values(), &beta, c->get_values());
-
-                cusparse::destroy(descr);
+            // TODO: add implementation for int64 and multiple RHS
+            if (cusparse::is_supported<ValueType, IndexType>::value)
```
upsj, line 317:
```suggestion
            if (!cusparse::is_supported<ValueType, IndexType>::value)
```

upsj (approved):
LGTM! Only minor nits

cuda/matrix/csr_kernels.cu
```diff
@@ -303,59 +303,59 @@ void spmv(std::shared_ptr<const CudaExecutor> exec,
...
-                               a->get_const_values(), row_ptrs, col_idxs,
-                               b->get_const_values(), &beta, c->get_values());
-
-                cusparse::destroy(descr);
+            // TODO: add implementation for int64 and multiple RHS
+            if (!cusparse::is_supported<ValueType, IndexType>::value) {
+                GKO_NOT_IMPLEMENTED;
+            }
+
+            auto descr = cusparse::create_mat_descr();
```
yhmtsai, line 321: Could you add the structure to handle the create/destroy like cublas/cusparse handle?
also dn_vec and sp_mat.
I did not add it properly when I push it :sweat_smile:

upsj: Maybe we could make that part of another PR? I already started updating the SpMV in #905, maybe I can include that there? This is a rather small fix. What do you mean more precisely? RAII for handles?

upsj:
BTW I can confirm this works on CUDA 10.2 and 11.0

yhmtsai (approved):

cuda/matrix/csr_kernels.cu
yhmtsai: I see, yes RAII for these mat_descr

upsj:
rebase!

PR #898, opened 2021-10-06
Add a BatchEll matrix format

pratikvn:
This PR adds a BatchEll matrix to handle batch matrices that have a uniform number of nonzeros per row. The performance gains are mainly from better coalesced accesses. The shared memory usage remains mostly the same. 

A simple kernel for SpMV has been added, that uses 1 thread per row and one thread block handles one batch entry of the matrix. In case of higher number of non-zeros per row, it might make more sense to have multiple threads work per row and have atomic adds between these threads. 

The interface more or less follows from BatchCsr, and adds a `batch_num_stored_elems_per_row` struct that follows similar ideas from `batch_stride`. `batch_stride` has also been moved to `BatchLinOp` from `BatchDense`. 

The plot below shows performance benefits of using a BatchEll format instead of a BatchCsr format for a nine point (2D) stencil matrix, but within the BatchBicgstab solver. 
csr_ell_times


include/ginkgo/core/matrix/batch_ell.hpp
```diff
@@ -0,0 +1,618 @@
...
+
+
+template <typename ValueType>
+class BatchDense;
+
+template <typename ValueType, typename IndexType>
+class BatchEll;
+
+
+class batch_num_stored_elems_per_row {
```
Slaedr, line 54: Do we really need this? IMO the number of nonzero elements per row is also a kind of stride. Maybe just
```
typedef batch_stride batch_num_stored_elems_per_row;
```

reference/test/matrix/batch_ell_kernels.cpp
```diff
@@ -0,0 +1,433 @@
...
+    EXPECT_EQ(y->at(0, 0, 1), uy0->at(0, 1));
+    EXPECT_EQ(y->at(0, 1, 1), uy0->at(1, 1));
+    EXPECT_EQ(y->at(1, 0, 0), uy1->at(0, 0));
+    EXPECT_EQ(y->at(1, 1, 0), uy1->at(1, 0));
+    EXPECT_EQ(y->at(1, 0, 1), uy1->at(0, 1));
+    EXPECT_EQ(y->at(1, 1, 1), uy1->at(1, 1));
+}
+
+
+TYPED_TEST(BatchEll, ApplyFailsOnWrongInnerDimension)
```
Slaedr, line 346: I guess these kinds of tests should also be in `core/test/matrix/batch_ell.cpp`?

Slaedr (changes requested):
The batch ELL kernel implementations themselves look good, but I have some concerns I point out below.

Batch scaling is not supported for `BatchEll` yet, but I suppose that's fine for now.

common/cuda_hip/matrix/batch_ell_kernels.hpp.inc
```diff
@@ -0,0 +1,241 @@
...
+        const auto beta_b = gko::batch::batch_entry(beta, ibatch);
+        const ValueType alphav = alpha_b.values[0];
+        const ValueType betav = beta_b.values[0];
+        advanced_spmv_kernel(alphav, a_b, b_b, betav, c_b);
+    }
+}
+
+
+template <typename ValueType>
+__global__ __launch_bounds__(config::max_block_size) void initialize_zero_dense(
```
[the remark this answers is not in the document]
pratikvn: I am not sure here. Because the idea was to use these only in BatchEll. At least currently, we don't use these in batch_dense. At some point if we do, then we can unify them.

common/cuda_hip/preconditioner/batch_jacobi.hpp.inc
```diff
@@ -77,7 +112,9 @@ public:
                  j < mat.row_ptrs[irow + 1]; j += warp_size) {
                 const int found = (mat.col_idxs[j] == irow);
                 if (found) {
-                    work_[irow] = one<ValueType>() / mat.values[j];
+                    work_[irow] = (mat.values[j] != zero<ValueType>())
```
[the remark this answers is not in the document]
pratikvn: I think it is mainly a verification to make sure we don't get any inf's. Do you think there is any specific advantage of your approach ? I doubt we will have entries that are < numeric_limits and not zero because we kind of sanitize in read and so on ?

omp/test/matrix/batch_ell_kernels.cpp
```diff
@@ -0,0 +1,394 @@
...
+};
+
+using valuetypes =
+    ::testing::Types<std::tuple<float, int>, std::tuple<double, gko::int32>,
+                     std::tuple<std::complex<float>, gko::int32>,
+                     std::tuple<std::complex<double>, gko::int32>>;
+TYPED_TEST_SUITE(BatchEll, valuetypes);
+
+
+TYPED_TEST(BatchEll, CanBeUnbatchedIntoEllMatrices)
```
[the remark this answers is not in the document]
pratikvn: Yes, but I wanted to make sure it worked on all backends.

include/ginkgo/core/matrix/batch_ell.hpp
pratikvn: I guess it is nice to be able to have separate member function names. But that is a good point. We can probably streamline this for `batch_stride` and this struct with a macro and create both classes and any further ones. I will leave it as it is for now and maybe do that later.

reference/test/matrix/batch_ell_kernels.cpp
pratikvn: I guess technically yes, but we do keep it in reference in other non-batched kernels, as apply is not technically a core function, in the sense that it needs to call a kernel.

common/cuda_hip/matrix/batch_ell_kernels.hpp.inc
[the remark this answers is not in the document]
Slaedr: If it's not too annoying you could just move it; I can see this being useful in general.

common/cuda_hip/preconditioner/batch_jacobi.hpp.inc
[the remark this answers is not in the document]
Slaedr: Basically I thought this was related to out recent discussion about numerical issues with Jacobi.

include/ginkgo/core/matrix/batch_ell.hpp
Slaedr: The most useful member function names are already neutral, like `at`. Maybe you could just rename a couple of functions in `batch_stride` itself, though that's not necessary. The num_stored_elems_per_row is exactly the column-stride of the storage.

reference/test/matrix/batch_ell_kernels.cpp
Slaedr: Now that reference is always built, and the actual thing being tested here is in core, I think it should go to core. But if you want to leave it here, that's fine.

PR #899, opened 2021-10-06
Update trigger labels

tcojean:
I changed the status labels to now be in the form `1:ST:ready-to-merge` so that they appear first in the label order, as was the case before Github updated their ordering policy some time ago.

To stay consistent, I also update it in the trigger job.


thoasm:
The purpose of the `1` is that it is sorted as the first label, especially when going through the Pull Requests. The status (`WIP` / `Ready for Review`) are the most important labels, so they should be first.
Previously, capitalized was sorted in front of lower case, now, character case does not matter. Currently, it looks like numbers are sorted in front of letters, so the `1` prefix was added.

PR #901, opened 2021-10-09
Row gather linop

yhmtsai:
This PR adds the row gather linop (Rowgatherer) interface to use dense row gather operation.
Also add mixed operation and the advanced row gather operation


include/ginkgo/core/matrix/row_gatherer.hpp
```diff
@@ -0,0 +1,181 @@
...
+#include <ginkgo/core/base/utils.hpp>
+
+
+namespace gko {
+namespace matrix {
+
+
+/**
+ * RowGatherer is a matrix "format" which stores the row rowgatherer
+ * arrays which can be used to perform rowgatherer rows to another matrix.
```
thoasm, line 58: From what I can tell, it does not store "row rowgatherer arrays" (also, it just does not sound right).
This format is storing row indices which can/will be used to gather rows.

include/ginkgo/core/matrix/row_gatherer.hpp
```diff
@@ -0,0 +1,181 @@
...
+
+public:
+    using index_type = IndexType;
+
+    /**
+     * Returns a pointer to the array of row_gather_index.
+     *
+     * @return the pointer to the row_gather_index array.
+     */
+    index_type* get_row_gather_index() noexcept
```
thoasm, line 86: Shouldn't this be plural: `get_row_gatherer_indices`? Same for the function below.

include/ginkgo/core/matrix/row_gatherer.hpp
```diff
@@ -0,0 +1,181 @@
...
+     * array.
+     */
+    size_type get_row_gather_index_size() const noexcept
+    {
+        return row_gather_index_.get_num_elems();
+    }
+
+protected:
+    /**
+     * Creates an uninitialized RowGatherer arrays on the specified executor..
```
thoasm, line 117:
```suggestion
     * Creates an uninitialized RowGatherer arrays on the specified executor.
```

include/ginkgo/core/matrix/row_gatherer.hpp
```diff
@@ -0,0 +1,181 @@
...
+     * @param exec  Executor associated to the matrix
+     * @param size  size of the RowGatherable matrix
+     */
+    RowGatherer(std::shared_ptr<const Executor> exec, const dim<2>& size)
+        : EnableLinOp<RowGatherer>(exec, size), row_gather_index_(exec, size[0])
+    {}
+
+    /**
+     * Creates a RowGatherer matrix from an already allocated (and initialized)
+     * row and column rowgatherer arrays.
```
thoasm, line 137: Row and column? I thought only row

include/ginkgo/core/matrix/row_gatherer.hpp
```diff
@@ -0,0 +1,181 @@
...
+     *
+     * @tparam IndicesArray  type of array of indices
+     *
+     * @param exec  Executor associated to the matrix
+     * @param size  size of the rowgatherer array.
+     * @param row_gather_indexindices array of rowgatherer array
```
thoasm, line 143: If you want to stick with singular, then
```suggestion
     * @param row_gather_index array of rowgatherer array
```

reference/test/matrix/dense_kernels.cpp
```diff
@@ -2154,13 +2154,95 @@ TYPED_TEST(Dense, SquareSubmatrixCanGatherRowsIntoDense)
...
+}
+
+
+TYPED_TEST(Dense, NonSquareSubmatrixCanGatherRowsWithRowGatherable)
+{
+    // clang-format off
+    // {1.0, 3.0, 2.0},
+    // {0.0, 5.0, 0.0}
+    // clang-format on
+
```
thoasm, line 2173: Do you need the empty line?
Also, please add that this is `mtx4` in the comments. It took me some time to realize that.
I also don't think you need to turn off clang-format for this:
```suggestion
    // mtx4: {{1.0, 3.0, 2.0}, {0.0, 5.0, 0.0}}
```

The same comment applies for the tests below with the same matrix as a comment.

test/matrix/dense_kernels.cpp
```diff
@@ -70,14 +70,15 @@ class Dense : public ::testing::Test {
 #else
     using vtype = double;
 #endif
+    // in single mode, mixedtype will be the same as vtype
+    using mixedtype = float;
```
thoasm, line 74: nit:
```suggestion
    // in single mode, mixed_type will be the same as vtype
    using mixed_type = float;
```

thoasm (commented):
Mostly formatting and documentation comments.

core/matrix/dense.cpp
```diff
@@ -1010,7 +1029,6 @@ std::unique_ptr<Dense<ValueType>> Dense<ValueType>::row_gather(
     return result;
 }
 
-
```
greole, line 1013: I believe this empty line should stay

core/matrix/dense.cpp
```diff
@@ -1022,20 +1040,125 @@ std::unique_ptr<Dense<ValueType>> Dense<ValueType>::row_gather(
     return result;
 }
```
greole, line 1042: And here is one empty line missing
```suggestion


```

core/matrix/dense.cpp
```diff
@@ -1022,20 +1040,125 @@ std::unique_ptr<Dense<ValueType>> Dense<ValueType>::row_gather(
...
+          typename... Args>
+void run(T obj, func f, Args... args)
+{
+    if (auto dobj = dynamic_cast<K>(obj)) {
+        f(dobj, args...);
+    } else {
+        run<Types...>(obj, f, args...);
+    }
+}
+
```
greole, line 1093: Minor: here is one empty line missing
```suggestion


```

core/test/matrix/row_gatherer.cpp
```diff
@@ -0,0 +1,239 @@
...
+#include <ginkgo/core/matrix/csr.hpp>
+#include <ginkgo/core/matrix/dense.hpp>
+
+
+#include "core/test/utils.hpp"
+#include "core/test/utils/assertions.hpp"
+
+
+namespace {
+
```
greole, line 50: Minor: here is one empty line missing
```suggestion


```

greole (commented):
Not a full review, but some minor comments ...

core/multigrid/amgx_pgm.cpp
```diff
@@ -141,27 +142,34 @@ void AmgxPgm<ValueType, IndexType>::generate()
...
     exec->copy_from(exec.get(), agg_.get_num_elems(), agg_.get_const_data(),
-                    prolong_op->get_col_idxs());
-    exec->run(amgx_pgm::make_fill_seq_array(prolong_op->get_row_ptrs(),
+                    prolong_csr->get_col_idxs());
+    exec->run(amgx_pgm::make_fill_seq_array(prolong_csr->get_row_ptrs(),
                                             fine_dim + 1));
-    exec->run(amgx_pgm::make_fill_array(prolong_op->get_values(), fine_dim,
+    exec->run(amgx_pgm::make_fill_array(prolong_csr->get_values(), fine_dim,
                                         one<ValueType>()));
     // TODO: implement the restrict_op from aggregation.
```
pratikvn, line 161: I guess because of this transpose operation, this generation is slower than AMGX ?

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -360,42 +362,29 @@ class Dense
...
-     * Copies the given rows from this matrix into `row_gathered`
-     *
-     * @param gather_indices  pointer to an array containing row indices
-     *                        from this matrix. It may contain duplicates.
-     * @param row_gathered  pointer to a Dense matrix that will store the
-     *                      gathered rows:
-     *                      `output(i,j) = input(gather_indices(i), j)`
-     *                      It must have the same number of columns as this
-     *                      matrix and `gather_indices->get_num_elems()` rows.
-     */
```
pratikvn, line 390: Unfortunately, this breaks interface. [See 1.4.0 dense.hpp](https://github.com/ginkgo-project/ginkgo/blob/v1.4.0/include/ginkgo/core/matrix/dense.hpp#L391)

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -360,42 +362,29 @@ class Dense
...
     void row_gather(const Array<int32>* gather_indices,
-                    Dense* row_gathered) const;
+                    LinOp* row_collection) const override;
 
-    /**
-     * @copydoc row_gather(const Array<int32>*, Dense*) const
-     */
```
pratikvn, line 396: Breaks interface

pratikvn (commented):
Some comments

include/ginkgo/core/matrix/dense.hpp
yhmtsai: this interface is inherited from the EnableRowGather, which is also the reason why I need to create both for LinOp and Dense<VT> here

thoasm: I don't think it does, see the Line 377+ of this file. There are the `row_gather` functions that take a `Dense` input instead of a `LinOp`.

include/ginkgo/ginkgo.hpp
```diff
@@ -89,6 +89,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/matrix/hybrid.hpp>
 #include <ginkgo/core/matrix/identity.hpp>
 #include <ginkgo/core/matrix/permutation.hpp>
+#include <ginkgo/core/matrix/row_gatherer.hpp>
```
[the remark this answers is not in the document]
yhmtsai: thanks! I totally forgot it

include/ginkgo/core/base/array.hpp
```diff
@@ -803,6 +809,17 @@ Array<ValueType> array_const_cast(ConstArrayView<ValueType> view)
...
+Array<ValueType> ConstArrayView<ValueType>::copy_to_array() const
+{
+    Array<ValueType> result(this->get_executor(), this->get_num_elems());
+    result.get_executor()->copy_from(this->get_executor().get(),
+                                     this->get_num_elems(),
+                                     this->get_const_data(), result.get_data());
+    return result;
+}
+
+
```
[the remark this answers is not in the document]
yhmtsai: do you mean implement it inside the class?
I thought it is impossible because ConstArrayView is before Array such that we can not use Array there.

include/ginkgo/core/matrix/dense.hpp
pratikvn: Oh, yes, I missed that. A took the git diff at its face value. Thanks. Then, I don't think that should be an issue.

include/ginkgo/core/base/array.hpp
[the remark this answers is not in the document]
thoasm: Yes, sorry, you are correct, since you actually create an `Array`, the type must be completely defined.

core/multigrid/amgx_pgm.cpp
```diff
@@ -141,27 +142,34 @@ void AmgxPgm<ValueType, IndexType>::generate()
...
+    // prolong_row_gather is the lightway implementation for prolongation
+    // TODO: However, we still create the csr to process coarse/restrict matrix
+    // generation. It may be changed when we have the direct triple product from
+    // agg index.
+    auto prolong_row_gather = share(matrix::RowGatherer<IndexType>::create(
+        exec, gko::dim<2>{fine_dim, coarse_dim}));
+    exec->copy_from(exec.get(), agg_.get_num_elems(), agg_.get_const_data(),
+                    prolong_row_gather->get_row_gather_index());
+    auto prolong_csr = share(
         matrix_type::create(exec, gko::dim<2>{fine_dim, coarse_dim}, fine_dim));
```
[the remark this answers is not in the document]
yhmtsai: I go changing matrix_type to csr_type

reference/test/matrix/dense_kernels.cpp
yhmtsai: I delete them now.
I think they are shown to help checking test result, but it requires update when we change the initialization of mtx*

core/matrix/dense.cpp
yhmtsai: It is overloaded, so one empty is enough

core/matrix/dense.cpp
yhmtsai: It is overloaded, so one empty is enough

core/multigrid/amgx_pgm.cpp
yhmtsai: yes, and it uses two csr spgemm to generate coarse

core/test/utils/assertions.hpp
```diff
@@ -624,6 +625,55 @@ ::testing::AssertionResult array_equal(const std::string& first_expression,
                                     second);
 }
 
+/**
+ * array_equal overload: first is ConstArrayView.
```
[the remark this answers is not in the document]
yhmtsai: it requires us to set SORT_MEMBER_DOCS = NO, to avoid the reordering issue

thoasm (approved):
LGTM!

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -562,6 +562,112 @@ class Permutable {
...
+ *                     [CRTP parameter]
+ * @tparam IndexType  the type of the indices used to extract/gather the rows
+ *
+ * @ingroup LinOp
+ */
+template <typename OutputType, typename IndexType>
+class EnableRowGather : public RowGatherable<IndexType> {
+public:
+    using output_type = OutputType;
+    // void row_gather(const Array<IndexType>*, LinOp*) const;
```
pratikvn, line 636: Commented code

include/ginkgo/core/matrix/row_gatherer.hpp
```diff
@@ -0,0 +1,183 @@
...
+    {
+        GKO_ASSERT_EQ(size[0], row_gatherer_indices_.get_num_elems());
+    }
+
+    void apply_impl(const LinOp* in, LinOp* out) const
+    {
+        auto gather = gko::as<RowGatherable<index_type>>(in);
+        gather->row_gather(&row_gatherer_indices_, out);
+    }
+
```
pratikvn, line 165: extra line

core/test/matrix/row_gatherer.cpp
```diff
@@ -0,0 +1,241 @@
...
+    ASSERT_THROW(gko::matrix::RowGatherer<i_type>::create(
+                     this->exec, gko::dim<2>{4, 2},
+                     gko::Array<i_type>::view(this->exec, 3, data)),
+                 gko::ValueMismatch);
+}
+
+
+TYPED_TEST(RowGatherer, KnowsItsSizeAndValues)
+{
+    this->assert_equal_to_original_mtx(this->mtx.get());
```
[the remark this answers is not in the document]
yhmtsai: no, because the row gatherer does not have the write_to(matrix_data), can not use GKO_ASSERT_MTX_NEAR

core/test/matrix/row_gatherer.cpp
```diff
@@ -0,0 +1,241 @@
...
+TYPED_TEST_SUITE(RowGatherer, TestType);
+
+
+TYPED_TEST(RowGatherer, CanBeEmpty)
+{
+    using i_type = typename TestFixture::i_type;
+    auto empty = gko::matrix::RowGatherer<i_type>::create(this->exec);
+
+    this->assert_empty(empty.get());
+    ASSERT_EQ(empty->get_const_row_gatherer_indices(), nullptr);
```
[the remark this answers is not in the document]
yhmtsai: good idea

include/ginkgo/core/base/lin_op.hpp
yhmtsai: no, it describes the following `using ..` brings which code to avoid compilation issue.
I add more sentences to describe it

core/test/matrix/row_gatherer.cpp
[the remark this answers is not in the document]
upsj: Wouldn't it be useful to add that feature, then?

core/matrix/dense.cpp
```diff
@@ -1032,20 +1050,134 @@ std::unique_ptr<Dense<ValueType>> Dense<ValueType>::row_gather(
...
+template <typename K, typename... Types, typename T, typename Func,
+          typename... Args>
+void run(T obj, Func f, Args... args)
+{
+    if (auto dobj = dynamic_cast<K>(obj)) {
+        f(dobj, args...);
+    } else {
+        run<Types...>(obj, f, args...);
+    }
+}
```
upsj, line 1110: This should be moved to some utility header, we will most likely need it elsewhere again.

core/solver/multigrid.cpp
```diff
@@ -72,30 +72,38 @@ namespace {
...
+ * @tparam ...Types  the other types will be tried in the conversion if K fails
+ * @tparam T  the type of input object waiting converted
+ * @tparam Func  the function will run if the object can be converted to pointer
+ *               of const Base<K>
+ * @tparam ...Args  the additional arguments for the Func
+ *
+ * @param obj  the input object waiting converted
+ * @param f  the function will run if obj can be converted successfully
+ * @param args  the additional arguments for the function
  */
```
upsj, line 107: same here

core/test/utils/assertions.hpp
```diff
@@ -624,6 +625,58 @@ ::testing::AssertionResult array_equal(const std::string& first_expression,
...
+ */
+template <typename ValueType>
+::testing::AssertionResult array_equal(
+    const std::string& first_expression, const std::string& second_expression,
+    const gko::detail::ConstArrayView<ValueType>& first,
+    const Array<ValueType>& second)
+{
+    return detail::array_equal_impl(first_expression, second_expression,
+                                    first.copy_to_array(), second);
+}
```
upsj, line 643: nice idea :)

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -562,6 +562,112 @@ class Permutable {
...
 
 
+/**
+ * The RowGatherable is an interface that allows to get the row gather result of
+ * a LinOp.
+ *
+ * @tparam IndexType  the type of the indices used to extract/gather the rows
+ */
+template <typename IndexType>
+class RowGatherable {
```
upsj, line 572: Do we need a new Mixin here? row gather on CSR and Dense have two quite separate functions, and don't need to be made generic? Also the fact that std::unique_ptr is not covariant as a return type makes for a weird interface for the _linop overload.

include/ginkgo/core/matrix/row_gatherer.hpp
```diff
@@ -0,0 +1,183 @@
...
+     * Returns the number of elements explicitly stored in the
+     * row_gatherer_indices array.
+     *
+     * @return the number of elements explicitly stored in the
+     * row_gatherer_indices array.
+     */
+    size_type get_row_gatherer_indices_size() const noexcept
+    {
+        return row_gatherer_indices_.get_num_elems();
+    }
```
upsj, line 113: That's equal to get_size()[0], right? Why do we need a separate function for that?

upsj (changes requested):
LGTM for the most part, just a few points on the interface I would like to discuss

include/ginkgo/core/base/lin_op.hpp
upsj: Or do you need this for some mixed-precision application?

yhmtsai: Yes, I think so? because the RowGatherer does not have value type information, I need this interface to call row_gather.
I thought CSR and Dense have the same interface for row_gather, doesn't it?
I do not understand the last sentence. could you describe it more detail?

include/ginkgo/core/matrix/row_gatherer.hpp
yhmtsai: that's true. I generate this from permutation, so just rename it to fit row_gather.
I will remove it

include/ginkgo/core/base/lin_op.hpp
upsj: you can't override a std::unique_ptr<LinOp> function() with a std::unique_ptr<Dense> function(). Covariance means that if your class is more specialized, your return type can also be more specialized. This is true for pointers, but not for smart pointers.

yhmtsai: Good to know this! does the compiler show the override only partially message for this kind on pointers?
I think the return type is not considered in the function overload, so I still need to use two version for them?

core/test/utils.hpp
```diff
@@ -200,7 +228,8 @@ struct TypenameNameGenerator {
 
 
 struct PairTypenameNameGenerator {
-    template <typename T>
+    template <typename T,
+              std::enable_if_t<std::tuple_size<T>::value == 2, bool> = true>
```
upsj, line 232: Nit: That would be a TupleTypenameNameGenerator, not Pair. No need to provide two overloads

include/ginkgo/core/matrix/row_gatherer.hpp
```diff
@@ -0,0 +1,170 @@
...
+
+    /**
+     * Returns a pointer to the array of row_gatherer_indices.
+     *
+     * @return the pointer to the row_gatherer_indices array.
+     */
+    index_type* get_row_gatherer_indices() noexcept
+    {
+        return row_gatherer_indices_.get_data();
+    }
```
upsj, line 89:
```suggestion
    /**
     * Returns a pointer to the row index array.
     *
     * @return the pointer to the row index array.
     */
    index_type* get_row_indices() noexcept
    {
        return row_gatherer_indices_.get_data();
    }
```

include/ginkgo/core/matrix/row_gatherer.hpp
```diff
@@ -0,0 +1,170 @@
...
+     * @copydoc get_row_gatherer_indices()
+     *
+     * @note This is the constant version of the function, which can be
+     *       significantly more memory efficient than the non-constant version,
+     *       so always prefer this version.
+     */
+    const index_type* get_const_row_gatherer_indices() const noexcept
+    {
+        return row_gatherer_indices_.get_const_data();
+    }
```
upsj, line 101: same here

include/ginkgo/core/matrix/row_gatherer.hpp
```diff
@@ -0,0 +1,170 @@
...
+     */
+    template <typename IndicesArray>
+    RowGatherer(std::shared_ptr<const Executor> exec, const dim<2>& size,
+                IndicesArray&& row_gatherer_indices)
+        : EnableLinOp<RowGatherer>(exec, size),
+          row_gatherer_indices_{
+              exec, std::forward<IndicesArray>(row_gatherer_indices)}
+    {
+        GKO_ASSERT_EQ(size[0], row_gatherer_indices_.get_num_elems());
+    }
```
upsj, line 146: I know we do it like this in other places, but this makes the interface unnecessarily generic and forces it to be in the header.
```suggestion
    RowGatherer(std::shared_ptr<const Executor> exec, const dim<2>& size,
                Array<IndexType> row_gatherer_indices)
        : EnableLinOp<RowGatherer>(exec, size),
          row_gatherer_indices_{
              exec, std::move(row_gatherer_indices)}
    {
        GKO_ASSERT_EQ(size[0], row_gatherer_indices_.get_num_elems());
    }
```

include/ginkgo/core/matrix/row_gatherer.hpp
```diff
@@ -0,0 +1,170 @@
...
+ * the rowgatherer indices.
+ *
+ * @ingroup rowgatherer
+ * @ingroup matrix
+ * @ingroup LinOp
+ */
+template <typename IndexType = int32>
+class RowGatherer : public EnableLinOp<RowGatherer<IndexType>>,
+                    public EnableCreateMethod<RowGatherer<IndexType>> {
+    friend class EnableCreateMethod<RowGatherer>;
```
upsj, line 75: Can also use a create_const method

include/ginkgo/core/matrix/row_gatherer.hpp
```diff
@@ -0,0 +1,170 @@
...
+        return row_gatherer_indices_.get_const_data();
+    }
+
+protected:
+    /**
+     * Creates an uninitialized RowGatherer arrays on the specified executor.
+     *
+     * @param exec  Executor associated to the LinOp
+     */
+    RowGatherer(std::shared_ptr<const Executor> exec)
```
upsj, line 109: All these member functions can be moved to the .cpp file

upsj (commented):
I forgot that I was blocking this, sorry. A few small nits, most of my comments have been adressed.
Still, I don't think we need a separate interface for behavior that is only implemented in Dense.
The advanced row_gather might also use a separate name like row_gather_add or similar, though I'm not 100% sure about that.
Also, a RowGatherer::write function might simplify things.

yhmtsai:
@upsj 
> How about you call the dense kernels inside RowGatherer, which means you can use mixed_precision_dispatch to figure out their dimensions? That eliminates the need for an interface.
Do you mean call the dense kernels in the row gatherer apply function? I do not understand what dimensions to figure out. Could you give more detail? The interface can be used by Csr, too. Csr->row_gather(index) = Csr
> Also, a RowGatherer::write function might simplify things in the tests.
Yes, but I am a little scared about setting the type which is not used by the class-self.

upsj:
@yhmtsai
Sorry, I typed the wrong word, I meant precisions, not dimensions, i.e. the value type.

Since this PR only involves Dense, I don't see the advantage of this approach. Adding Csr would introduce another feature hidden behind an apply(...) function like #942, which I would like to avoid as much as possible.

upsj:
About write: Fft also declares a value_type, even though it doesn't have values strictly speaking.

yhmtsai:
the precision is also dispatched in the row_gather of dense.
Yes, if we only restrict the apply in dense, do not need the interface.
I stand at the opposite side. I think it should be allowed for any kind of LinOp when we have the implementation.
The complex interface (dense input/output additionally) is to fit the original dense interface.
about the row gather, if we only allow dense case, what should we have the interface for Csr case when we need it?
dense can use `row_gather->apply(dense, dense)` or `dense->row_gather(index)`
but csr can only use `csr->row_gather(index)`

include/ginkgo/core/matrix/row_gatherer.hpp
upsj: Ah nevermind, using this interface allows us to avoid a single copy if the array is on another executor than the RowGatherer

yhmtsai:
@upsj row gather only consider dense now, so apply will go over all possible dense type to call row_gather

upsj (approved):

PR #902, opened 2021-10-11
Move BwUni-CI to nla-gpu1 during maintenance

tcojean:
Since BwUniCluster is currently under maintenance, its jobs are moved to our new AMD systems for now.

### DIFF
--- a/.gitlab/image.yml
+++ b/.gitlab/image.yml
@@ -10,21 +10,24 @@
   tags:
     - private_ci
     - cpu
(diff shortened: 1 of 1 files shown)


upsj:
Would it be possible to register a different runner under the same tag temporarily? Or is this change strictly necessary?

upsj:
I think doing this in the background would make it easier to change back in the future, without having to create another PR.

PR #903, opened 2021-10-11
Improve test naming in typed tests

upsj:
This passes a suitable NameGenerator to all templated tests, which results in a more descriptive test name:
Before:
`Csr/7.GeneratesCorrectMatrixData`
After
`Csr/<std::complex<double>, long>.GeneratesCorrectMatrixData`


tcojean:
That's another issue, I'm looking into it.

PR #904, opened 2021-10-11
Add generic matrix format test

### LINKED ISSUE #590: Test operations on empty (dimensions or #nnz) matrices
Currently, we don't really guard against executing operations on empty matrices. Similar to #380, there are a few places where this might cause issues, e.g. empty kernel launches (cause cuda-memcheck failures).

upsj:
This tests a few generic behaviors for all sparse matrix formats:
* SpMV with different strides, numbers of rhs, real-complex behavior and mixed precision
* Conversions from and to Csr and Dense
* write -> read round-trips
* Handling of empty matrices

TODO:
- [x] Fix found issues
- [x] ~~Fix mixed-precision ELL intermediate formats~~
- [x] ~~Add more specific mixed-precision tests (stricter tolerances)~~
- [x] ~~Make GKO_ASSERT_MTX_EQ_SPARSITY more strict w.r.t numeric zeros~~
- [x] ~~Simplify handling of ref/device objects? (remove test_pair)~~
- [x] ~~Remove fine-grained exception reporting?~~
- [x] ~~Maybe remove corresponding tests from device-specific kernel tests~~

Fixes #590


test/matrix/matrix.cpp
```diff
@@ -0,0 +1,782 @@
...
+    HybridImbalanceBoundedLimitStrategy, HybridMinStorageStrategy,
+    HybridAutomaticStrategy>;
+
+TYPED_TEST_SUITE(Matrix, MatrixTypes, TypenameNameGenerator);
+
+
+TYPED_TEST(Matrix, SpMVIsEquivalentToRef)
+{
+    this->forall_matrix_scenarios([&](auto mtx) {
+        this->forall_vector_scenarios(mtx, [&](auto b, auto x) {
```
pratikvn, line 621: Really nice!

test/matrix/matrix.cpp
```diff
@@ -0,0 +1,782 @@
...
+                mtx.dev->apply(alpha.dev.get(), b.dev.get(), alpha.dev.get(),
+                               x.dev.get());
+
+                GKO_ASSERT_MTX_NEAR(x.ref, x.dev, this->mixed_tol());
+            });
+    });
+}
+
+
+TYPED_TEST(Matrix, ConvertToCsrIsEquivalentToRef)
```
pratikvn, line 684: Should we be testing conversions for other matrix formats as well ?

pratikvn (approved):
Looks good to me! Really nice work! Should make our testing more robust.

Slaedr (approved):
Excellent work! This is very cool.

test/matrix/matrix.cpp
```diff
@@ -0,0 +1,782 @@
...
+#include <ginkgo/core/matrix/csr.hpp>
+#include <ginkgo/core/matrix/dense.hpp>
+#include <ginkgo/core/matrix/ell.hpp>
+#include <ginkgo/core/matrix/fbcsr.hpp>
+#include <ginkgo/core/matrix/hybrid.hpp>
+#include <ginkgo/core/matrix/sellp.hpp>
+
+
+#include "core/test/utils.hpp"
+#include "ginkgo/core/base/matrix_data.hpp"
```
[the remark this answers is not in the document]
upsj: clangd adding includes automatically without asking :-D

test/matrix/matrix.cpp
```diff
@@ -0,0 +1,782 @@
...
+    static std::unique_ptr<matrix_type> create(
+        std::shared_ptr<gko::Executor> exec, gko::dim<2> size)
+    {
+        return matrix_type::create(exec->get_master(), size);
+    }
+
+    static void check_property(const std::unique_ptr<matrix_type>&) {}
+};
+
+struct DenseWithDefaultStride : SimpleMatrixTest<gko::matrix::Dense<double>> {
```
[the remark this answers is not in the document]
upsj: no, that's definitely possible! I just started with the easy ones

test/matrix/matrix.cpp
upsj: Which formats do you mean? We only have conversions from all formats to Csr/Dense and from Csr/Dense to all formats, with minor exceptions

test/matrix/matrix.cpp
```diff
@@ -0,0 +1,782 @@
...
+        }
+        {
+            SCOPED_TRACE("Sparse Matrix with variable row nnz (200x100)");
+            guarded_fn(gen_mtx_data(200, 100, 10, 50));
+        }
+        {
+            SCOPED_TRACE(
+                "Sparse Matrix with heavily imbalanced row nnz (200x100)");
+            guarded_fn(gen_mtx_data(200, 100,
+                                    std::binomial_distribution<>{100, 0.05}));
```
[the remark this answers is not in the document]
upsj: No, this is a random distribution, that means that the average row has something like 5 entries, with 100 being extremely unlikely. But now that I think of it, I should probably use another, more obviously skewed distribution and guard agains overflows inside gen_mtx_data

test/matrix/matrix.cpp
pratikvn: Ah, yes. That is true. I forgot that you only test if all the permutations for a certain format have been implemented.

test/matrix/matrix.cpp
```diff
@@ -0,0 +1,782 @@
...
+
+    template <typename VecType = Vec>
+    test_pair<VecType> gen_out_vec(const test_pair<Mtx>& mtx, int nrhs,
+                                   int stride)
+    {
+        auto size = gko::dim<2>{mtx.ref->get_size()[0],
+                                static_cast<gko::size_type>(nrhs)};
+        auto result = VecType::create(ref, size, stride);
+        result->read(gen_dense_data<typename VecType::value_type,
+                                    typename Mtx::index_type>(size));
```
yhmtsai, line 436: how about to add the NaN in output vector when they are used in apply(b, x)?
It may give the issue on dpcpp sparsely csr apply which I do not know they fix or not, or some other formats. But it is easy to fix on our side.

test/matrix/matrix.cpp
```diff
@@ -0,0 +1,782 @@
...
+            mtx->get_strategy().get()));
+    }
+};
+
+
+#endif
+
+
+struct Ell : SimpleMatrixTest<gko::matrix::Ell<double, int>> {};
+
```
yhmtsai, line 186:
```suggestion


```

test/matrix/matrix.cpp
```diff
@@ -0,0 +1,782 @@
...
+    {
+        return matrix_type::create(exec, size, 0, 2);
+    }
+
+    static void check_property(const std::unique_ptr<matrix_type>& mtx)
+    {
+        ASSERT_EQ(mtx->get_block_size(), 2);
+    }
+};
+
```
yhmtsai, line 216:
```suggestion


```
I guess you split different format with 2 empty line but different setting with 1 empty line?

test/matrix/matrix.cpp
```diff
@@ -0,0 +1,782 @@
...
+        return matrix_type::create(exec, size, 32, 2, 0);
+    }
+
+    static void check_property(const std::unique_ptr<matrix_type>& mtx)
+    {
+        ASSERT_EQ(mtx->get_stride_factor(), 2);
+        ASSERT_EQ(mtx->get_slice_size(), 32);
+    }
+};
+
```
yhmtsai, line 239:
```suggestion


```

test/matrix/matrix.cpp
```diff
@@ -0,0 +1,782 @@
...
+    using value_type = typename TestFixture::value_type;
+    using index_type = typename TestFixture::index_type;
+    this->forall_matrix_data_scenarios([&](auto data) {
+        auto new_mtx = TestConfig::create(this->exec, data.size);
+        gko::matrix_data<value_type, index_type> out_data;
+
+        new_mtx->read(data);
+        new_mtx->write(out_data);
+
+        if (!TestConfig::preserves_zeros()) {
```
yhmtsai, line 775: the preserve_zeros depends on the matrix format now. Should we move preserve_zeros is an option of the matrix?

test/matrix/matrix.cpp
yhmtsai: Do we also make the minor exception test shown here?

yhmtsai (approved):
LGTM. Thanks for implementing it! some nit and question in review comments.

test/matrix/matrix.cpp
upsj: In the future yes

test/matrix/matrix.cpp
upsj: yes

test/matrix/matrix.cpp
upsj: The tests are already handled inside the format-specific test files, so I will leave them out for now, but we can definitely add such tests later.

PR #909, opened 2021-10-22
Add a class for partitions of intervals

### LINKED ISSUE #907: Create a clean distributed-ginkgo branch
To ease the development of distributed functionality for ginkgo, there should exist a branch `distributed-ginkgo` that contains the current and agreed-upon status.

For this, I want to incrementally add the functionality that is already in the branch `mpi-base-dist-mat` into the new branch. Currently, I think this will require PR for the following functionality:
- [x] MPI bindings #908  (1) 
- [x] a partition of global indices #909  (2)
- [x] distributed vectors (3) #961 
- [x] distributed matrices (4) #971
- [x] distributed solvers (5) #976 
- [ ] enable logging of MPI communications (6)
- [ ] add distributed domain decomposition predonditioner (7)
- [ ] figure out mutable access to local vectors and matrices (8) #1159 

There are the following dependencies between these tasks:
- (5) → (4) & (3)
- (4) & (3) → (2) & (1)
- (6) → (1)
- (7) → (4) & (3)
- (8) → (4) 

At the moment, there are no dependencies between (1) and (2).

MarcelKoch:
This PR adds a partition class, which handles an interval that is distributed across multiple ranks. 

Partially closes #907.

Main contributions are from @upsj.


Slaedr:
@upsj Okay. In that case, where will the Partition class be used? Only for setting up the initial communication patterns for SpMVs etc.? Can the Partition object be deleted after that?

upsj:
@Slaedr yes, in my [original design](https://github.com/ginkgo-project/ginkgo/blob/distributed-matrix/include/ginkgo/core/distributed/matrix.hpp) it is only used inside read_distributed, but I think @MarcelKoch needed to modify it a bit to make repartitioning work?

cuda/distributed/partition_kernels.cu
```diff
@@ -0,0 +1,125 @@
...
+namespace gko {
+namespace kernels {
+namespace cuda {
+namespace partition {
+
+
+template <typename LocalIndexType, typename GlobalIndexType>
+void build_starting_indices(std::shared_ptr<const DefaultExecutor> exec,
+                            const GlobalIndexType* range_offsets,
+                            const int* range_parts, size_type num_ranges,
```
Slaedr, line 57:
```suggestion
                            const comm_index_type* range_parts, size_type num_ranges,
```

cuda/distributed/partition_kernels.cu
```diff
@@ -0,0 +1,125 @@
...
+                            int num_parts, int& num_empty_parts,
+                            LocalIndexType* ranks, LocalIndexType* sizes)
```
Slaedr, line 59: To make it clearer:
```suggestion
                            LocalIndexType* local_start_indices, LocalIndexType* part_sizes)
```

cuda/distributed/partition_kernels.cu
```diff
@@ -0,0 +1,125 @@
...
+                               tmp_part_id_ptr + num_ranges, value_it);
+    // compute inclusive prefix sum for each part
+    thrust::inclusive_scan_by_key(thrust::device, tmp_part_id_ptr,
+                                  tmp_part_id_ptr + num_ranges, range_sizes_ptr,
+                                  range_sizes_ptr);
+    // write back the results
+    run_kernel(
+        exec,
+        [] GKO_KERNEL(auto i, auto grouped_range_ranks, auto grouped_part_ids,
+                      auto orig_idxs, auto ranks, auto sizes) {
```
Slaedr, line 100: Again, it would be nice to have the formal parameters with the same names as the arrays in the surrounding code.

dpcpp/distributed/partition_kernels.dp.cpp
```diff
@@ -0,0 +1,108 @@
...
+    // TODO compute "row_ptrs" for tmp_part_ids
+    // TODO compute prefix_sum over range_sizes
+    // TODO compute adjacent differences, set -part_size at part boundaries
+    // TODO compute prefix_sum again
+    // write back the results
+    // TODO this needs to be adapted to the output of the algorithm above
+    // TODO count number of zeros in size and store in num_empty_parts
+    run_kernel(
+        exec,
+        [] GKO_KERNEL(auto i, auto grouped_range_ranks, auto grouped_part_ids,
```
Slaedr, line 84: Since this code is not particularly trivial, I suggest you remove this kernel and put it in a common header file; perhaps `common/unified/partition_kernels.hpp`? Or any other place you think is appropriate. This could apply to the kernel above as well.

dpcpp/distributed/partition_kernels.dp.cpp
```diff
@@ -0,0 +1,108 @@
...
+namespace partition {
+
+
+template <typename LocalIndexType, typename GlobalIndexType>
+void build_starting_indices(std::shared_ptr<const DefaultExecutor> exec,
+                            const GlobalIndexType* range_offsets,
+                            const int* range_parts, size_type num_ranges,
+                            int num_parts, int& num_empty_parts,
+                            LocalIndexType* ranks, LocalIndexType* sizes)
+{
```
Slaedr, line 48: Since this kernel is not yet supposed to work.
```suggestion
{
    GKO_NOT_IMPLEMENTED;
```

hip/distributed/partition_kernels.hip.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+
+template <typename LocalIndexType>
+void build_starting_indices(std::shared_ptr<const DefaultExecutor> exec,
+                            const global_index_type* range_offsets,
+                            const int* range_parts, size_type num_ranges,
+                            int num_parts, LocalIndexType* ranks,
+                            LocalIndexType* sizes)
+{
+    Array<LocalIndexType> range_sizes{exec, num_ranges};
```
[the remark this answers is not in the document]
Slaedr: Instead, I would suggest moving the common kernels below to `common/unified` so that they can be used in DPC++ as well. The rest (thrust calls etc.) could be left here.

omp/distributed/partition_kernels.cpp
```diff
@@ -0,0 +1,102 @@
...
+                            const int* range_parts, size_type num_ranges,
+                            int num_parts, int& num_empty_parts,
+                            LocalIndexType* ranks, LocalIndexType* sizes)
+{
+    std::fill_n(sizes, num_parts, 0);
+    auto num_threads = static_cast<size_type>(omp_get_max_threads());
+    auto size_per_thread = (num_ranges + num_threads - 1) / num_threads;
+    vector<LocalIndexType> local_sizes(num_parts * num_threads, 0, {exec});
+#pragma omp parallel
+    {
```
Slaedr, line 60: This kernel could be replaced by the same code as for cuda/hip. Thrust can, in fact, be used on the host with an openmp backend as well. You'll just have to look carefully at what Thrust macro to define where so that it uses OpenMP in this file.

I guess one issue is we don't look for Thrust when we're not building the cuda or hip backends. If we don't want to add a thrust dependency for that, perhaps it makes sense to keep this.

test/distributed/partition_kernels.cpp
```diff
@@ -0,0 +1,463 @@
...
+                                                                    ranges);
+    auto dpart = gko::distributed::Partition<
+        local_index_type, global_index_type>::build_from_contiguous(this->exec,
+                                                                    dranges);
+
+    this->assert_equal(part, dpart);
+}
+
+
+TYPED_TEST(Partition, BuildsFromContiguousWithSomeMostlyEmptyParts)
```
Slaedr, line 288:
```suggestion
TYPED_TEST(Partition, BuildsFromContiguousWithMostlyEmptyParts)
```

Slaedr (commented):
Great work on the algorithms. I guess there are a couple of things left:
- core tests
- Openmp tests if we decide to keep the separate implementation.

Apart from that, I have some comments below.

common/unified/distributed/partition_kernels.cpp
```diff
@@ -0,0 +1,174 @@
...
+            output[i] = cur_part != prev_part ? 1 : 0;
+        },
+        mapping.get_num_elems(), mapping, range_index_ranks);
+    components::prefix_sum(exec, range_index_ranks.get_data(),
+                           mapping.get_num_elems() + 1);
+    run_kernel(
+        exec,
+        [] GKO_KERNEL(auto i, auto size, auto mapping, auto prefix_sum,
+                      auto ranges, auto range_parts) {
+            const auto prev_part = i > 0 ? mapping[i - 1] : comm_index_type{-1};
```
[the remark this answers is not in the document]
MarcelKoch: The PR #676 has a safer sentinel value, which does basically the same, so we will use that one after that PR is merged. I don't think it is necessary to add an assert until then.

common/unified/distributed/partition_kernels.cpp
```diff
@@ -0,0 +1,174 @@
...
+void build_from_contiguous(std::shared_ptr<const DefaultExecutor> exec,
+                           const Array<GlobalIndexType>& ranges,
+                           GlobalIndexType* range_bounds,
+                           comm_index_type* part_ids)
+{
+    run_kernel(
+        exec,
+        [] GKO_KERNEL(auto i, auto ranges, auto bounds, auto ids) {
+            if (i == 0) {
+                bounds[0] = 0;
```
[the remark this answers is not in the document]
MarcelKoch: All array of the partition class are filled with 0, so `ranges[0]` doesn't need to be changed. I will check if setting `bounds[0]` is even necessary.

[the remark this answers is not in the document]
upsj: That is mainly a simplification like row pointers, where we could theoretically omit the initial 0, but it makes implementations much easier at little overhead.

common/unified/distributed/partition_kernels.cpp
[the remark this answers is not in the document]
Slaedr: What is the safer sentinel value?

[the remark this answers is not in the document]
MarcelKoch: See https://github.com/ginkgo-project/ginkgo/blob/e18f33a663b3fe29263c4361e02553b6adc82432/include/ginkgo/core/base/types.hpp#L637
It basically does what you suggest.

omp/distributed/partition_kernels.cpp
MarcelKoch: Yes, I don't think the benefit from unifying it is bigger than the hassle from using/not using thrust, so I will keep it as it is.

test/distributed/partition_kernels.cpp
```diff
@@ -0,0 +1,463 @@
...
+TYPED_TEST(Partition, IsOrderedRandom)
+{
+    using local_index_type = typename TestFixture::local_index_type;
+    using global_index_type = typename TestFixture::global_index_type;
+    comm_index_type num_parts = 7;
+    std::uniform_int_distribution<comm_index_type> part_dist{0, num_parts - 1};
+    gko::Array<comm_index_type> mapping{this->ref, 10000};
+    for (gko::size_type i = 0; i < mapping.get_num_elems(); i++) {
+        mapping.get_data()[i] = part_dist(this->rand_engine);
+    }
```
yhmtsai, line 449: can also use generate_random_array

common/unified/distributed/partition_kernels.cpp
```diff
@@ -0,0 +1,174 @@
...
+                           const Array<GlobalIndexType>& ranges,
+                           GlobalIndexType* range_bounds,
+                           comm_index_type* part_ids)
+{
+    run_kernel(
+        exec,
+        [] GKO_KERNEL(auto i, auto ranges, auto bounds, auto ids) {
+            bounds[i + 1] = ranges[i + 1];
+            ids[i] = i;
+        },
```
yhmtsai, line 78: range_bounds[0]?

common/unified/distributed/partition_kernels.hpp.inc
```diff
@@ -0,0 +1,97 @@
...
+{
+    run_kernel(
+        exec,
+        [] GKO_KERNEL(auto i, auto grouped_starting_indices,
+                      auto grouped_part_ids, auto orig_idxs,
+                      auto starting_indices, auto part_sizes) {
+            auto prev_part =
+                i > 0 ? grouped_part_ids[i - 1] : comm_index_type{-1};
+            auto cur_part = grouped_part_ids[i];
+            auto next_part = grouped_part_ids[i + 1];  // safe due to sentinel
```
yhmtsai, line 81: Is it due to sentinel? I thought it due to the size and index?

yhmtsai (changes requested):
the public interface in partition.hpp needs `@return` or `@param`
common/unified/distributed/partition_kernels.hpp.inc should be in cuda_hip if it only be used for cuda_hip.
`common/cuda_hip/distributed/partition_kernels.hpp.inc` should be in cuda and hip separately if we stay the original design.

common/unified/distributed/partition_kernels.cpp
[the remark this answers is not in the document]
Slaedr: Now I guess we can use `invalid_index` here?

common/unified/distributed/partition_kernels.hpp.inc
```diff
@@ -0,0 +1,97 @@
...
+HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+#ifndef GINKGO_PARTITION_KERNELS_HPP
```
Slaedr, line 33: If you want to keep the guard, maybe
```suggestion
#ifndef GKO_COMMON_UNIFIED_PARTITION_KERNELS_HPP
```

common/unified/distributed/partition_kernels.hpp.inc
```diff
@@ -0,0 +1,97 @@
...
+    const Array<GlobalIndexType>& permutation, LocalIndexType* starting_indices,
+    LocalIndexType* part_sizes)
+{
+    run_kernel(
+        exec,
+        [] GKO_KERNEL(auto i, auto grouped_starting_indices,
+                      auto grouped_part_ids, auto orig_idxs,
+                      auto starting_indices, auto part_sizes) {
+            auto prev_part =
+                i > 0 ? grouped_part_ids[i - 1] : comm_index_type{-1};
```
Slaedr, line 79:
```suggestion
                i > 0 ? grouped_part_ids[i - 1] : invalid_index<comm_index_type>();
```

common/unified/distributed/partition_kernels.cpp
```diff
@@ -0,0 +1,174 @@
...
+        mapping.get_num_elems(), mapping, range_starting_index);
+    components::prefix_sum(exec, range_starting_index.get_data(),
+                           mapping.get_num_elems() + 1);
+    run_kernel(
+        exec,
+        [] GKO_KERNEL(auto i, auto size, auto mapping,
+                      auto range_starting_index, auto ranges,
+                      auto range_parts) {
+            const auto prev_part = i > 0 ? mapping[i - 1] : comm_index_type{-1};
```
Slaedr, line 107:
```suggestion
            const auto prev_part = i > 0 ? mapping[i - 1] : invalid_index<comm_index_type>();
```

core/distributed/partition.cpp
```diff
@@ -0,0 +1,151 @@
...
+std::unique_ptr<Partition<LocalIndexType, GlobalIndexType>>
+Partition<LocalIndexType, GlobalIndexType>::build_from_contiguous(
+    std::shared_ptr<const Executor> exec, const Array<GlobalIndexType>& ranges)
+{
+    auto local_ranges = make_temporary_clone(exec, &ranges);
+    if (ranges.get_num_elems() > 0) {
+        auto result = Partition::create(
+            exec, static_cast<comm_index_type>(ranges.get_num_elems() - 1),
+            ranges.get_num_elems() - 1);
+        exec->run(partition::make_build_from_contiguous(
```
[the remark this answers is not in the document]
Slaedr: It might be a good idea to have a check only in debug builds.

core/distributed/partition.cpp
```diff
@@ -0,0 +1,151 @@
...
+}
+
+
+template <typename LocalIndexType, typename GlobalIndexType>
+std::unique_ptr<Partition<LocalIndexType, GlobalIndexType>>
+Partition<LocalIndexType, GlobalIndexType>::build_from_contiguous(
+    std::shared_ptr<const Executor> exec, const Array<GlobalIndexType>& ranges)
+{
+    auto local_ranges = make_temporary_clone(exec, &ranges);
+    if (ranges.get_num_elems() > 0) {
```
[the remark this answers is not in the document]
Slaedr: If range `i` contains one element, say 3, it would be [3,4), so `ranges[i] == 3` and `ranges[i+1] == 4`. So I guess, indeed, a valid `ranges` must have at least two entries.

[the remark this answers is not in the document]
MarcelKoch: Sorry about this check, that is some WIP that I forgot to leave out. I will remove that again, and add some validation support when that becomes possible.

common/unified/distributed/partition_kernels.cpp
```diff
@@ -0,0 +1,174 @@
...
+    const auto num_ranges = partition->get_num_ranges();
+    // it is necessary to use uint32 as a temporary result, since
+    // bool can't be used with suffles
+    Array<uint32> result_uint32{exec, 1};
+    run_kernel_reduction(
+        exec,
+        [] GKO_KERNEL(auto i, const auto part_ids) {
+            return static_cast<uint32>(part_ids[i] < part_ids[i + 1]);
+        },
+        [] GKO_KERNEL(const auto a, const auto b) { return a && b; },
```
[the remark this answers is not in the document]
Slaedr: I think `&&` might be better here, because we really mean a full AND here, not a bitwise AND. The bitwise AND should also work, but then we are relying on the cast to always put the digit 1 in the same position while casting from `true`. I'm not sure what the standard says, though that's probably always the case. But still, I doubt it matters for performance and `&&` is clearer here.

common/unified/distributed/partition_kernels.cpp
Slaedr: I think that's taken care of because it's initialized to 0 in the constructor. But, for robustness w.r.t. future changes in the constructor, it might be better to explicitly set it here.

Slaedr (approved):
I agree with some of @yhmtsai's points. Once those are addressed, I think this is good to go. 

(I guess we're all okay with storing the entire partition everywhere. 😅 )

core/distributed/partition.cpp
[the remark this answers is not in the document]
MarcelKoch: At the moment it just garbage in, garbage out. A check could be added as part of the validation overhaul.

omp/distributed/partition_kernels.cpp
```diff
@@ -0,0 +1,102 @@
...
+            for (size_type thread = 0; thread < num_threads; ++thread) {
+                auto idx = num_parts * thread + part;
+                auto local_size = local_sizes[idx];
+                local_sizes[idx] = size;
+                size += local_size;
+            }
+            sizes[part] = size;
+            num_empty_parts += size == 0 ? 1 : 0;
+        }
+        // add global baselines to local ranks
```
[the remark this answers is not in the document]
MarcelKoch: I don't think that is necessary, since we don't use the `nowait` clause

reference/test/distributed/partition_kernels.cpp
```diff
@@ -0,0 +1,321 @@
...
+    assert_equal_data(partition->get_range_starting_indices(), {0, 0, 0, 0, 0});
+    assert_equal_data(partition->get_part_sizes(), {0, 0, 0, 0, 0});
+}
+
+
+TYPED_TEST(Partition, BuildsFromGlobalSizeWithEmptyParts)
+{
+    using local_index_type = typename TestFixture::local_index_type;
+    using global_index_type = typename TestFixture::global_index_type;
+
```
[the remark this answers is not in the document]
MarcelKoch: I think it's necessary for our AAA pattern convention.

omp/distributed/partition_kernels.cpp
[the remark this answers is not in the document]
upsj: Might make sense to double-check that, I've been bitten by such an assumption before (which is also why we have a barrier above)

common/unified/distributed/partition_kernels.cpp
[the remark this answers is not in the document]
yhmtsai: I guess you mean from logical point?
each entry is true/false -> 1/0 and we reduce them here. They are same meaning if they always 1/0.
If to keep the logical point, I would suggest we have explicit cast here.

omp/distributed/partition_kernels.cpp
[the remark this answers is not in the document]
yhmtsai: @MarcelKoch Could you point me where I can get the `omp for` has sync in the end?
the previous for is normal for in each thread not `#omp for` so it needs the sync, I think. Or the `#omp for` also has a sync in the beginning

reference/test/distributed/partition_kernels.cpp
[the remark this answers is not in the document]
yhmtsai: I see, it's also okay.
we do not consider the `using...` in any AAA section, so it always go to the nearest AAA section.
If you consider it is belong to Assign section, I think it should make sense, too.

common/unified/distributed/partition_kernels.hpp.inc
yhmtsai: GKO_COMMON... is more proper.
the format_header only consider the first guard with `#ifndef GKO_...`, so it does not catch this and format it.

omp/distributed/partition_kernels.cpp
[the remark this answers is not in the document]
MarcelKoch: I've looked at the 4.5 standard specification https://www.openmp.org/wp-content/uploads/openmp-4.5.pdf, and I interpret the section on p.205 l.17-22 on the `reduction` clause as if it will synchronize the result at the end of the reduction.
But perhaps that is not true if we rely on an older standard.

test/distributed/partition_kernels.cpp
MarcelKoch: thanks, didn't know about that function.

common/unified/distributed/partition_kernels.hpp.inc
MarcelKoch: I've removed the guard, since we don't use it in these cases.

common/unified/distributed/partition_kernels.cpp
MarcelKoch: I set it now explicitly in the kernel.

common/unified/distributed/partition_kernels.hpp.inc
MarcelKoch: The sentinel value is set in the kernel above, so it is correct, but also confusing, since I've extracted both kernels into their own functions. I've adjusted the comment to reflect that.

common/unified/distributed/partition_kernels.cpp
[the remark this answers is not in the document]
MarcelKoch: In which direction do mean the cast? For me, the only reasonable is casting the result to `uint32`. Also, I would like to stay with `&&` since it's clearer, as @Slaedr mentioned.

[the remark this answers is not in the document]
upsj: Even though it's not explicitly specified (I will take care of that though!), run_kernel_reduction should internally assign every result to the ValueType deduced for the reduction, i.e. the identity element parameter. So we don't need to worry about returning a different type here.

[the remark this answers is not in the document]
MarcelKoch: I've added the cast to `unit32`, although I don't think it should matter too much.

omp/distributed/partition_kernels.cpp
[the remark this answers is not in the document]
yhmtsai: from 3.1, https://www.openmp.org/wp-content/uploads/OpenMP3.1.pdf
p105 l.13 - 19 gives the similar behavior

yhmtsai:
LGTM for kernel side. but the unified stuff looks weird to me.
`common/unified/distributed/partition_kernels.hpp.inc` is used in `common/cuda_hip/distributed/partition_kernels.hpp.inc`

`common/cuda_hip/distributed/partition_kernels.hpp.inc` should avoid including the file because we do not have guard there.
unified is supposed to work on all executor except for reference, so removing dpcpp from test/kernel is weird to me.
which part of the unified code doesn't work on dpcpp? they should work on dpcpp, I think

MarcelKoch:
@yhmtsai The work for dpcpp was not finished and it's not a priority for me. At some point we discussed, that we should rather proceed without the dpcpp kernels than wait until that is fixed. But in the end, it would also use the kernels from `common/unified/distributed/partition_kernels.hpp.inc`, that's why I left it there instead of putting it into the `cuda_hip` file. I agree that using a `*.hpp.inc` file in `common/unified` looks a bit weird, but since the usage is more similar to the `cuda_hip` files, I went this way.

I don't think an include guard is necessary, since `partition_kernel.hpp.inc` won't be included twice somewhere, and leaving out the include guard is more consistent with the rest of the `common` files.

Regarding the test, the alternative would be to write the same tests three times, which seems unnecessary to me. Also, the dpcpp will become eventually available, so I think is justified to have the common tests, with the one exception.

yhmtsai:
Yes, we do not put the guard in `cuda_hip/*.inc` file. I am not saying the guard is needed when we use cuda_hip like a macro way.
We do not include any inc file in inc and it should explicitly be include from the file such that we know the file are copied to here. include inc in inc will make us ignore that. we may include two inc and they have the duplicated item.
That's a reason why I think we should avoid include a file inside inc when we do not have guard there.

if the unified/*.inc if only used in cuda_hip, can it not be in cuda_hip?
I agree with processing the pr without dpcpp support, but it might be considered moved out of unified.
If I understand it correctly, the kernels in unified should also work on dpcpp. you can test the kernel directly.

test can test them the directly not from the constructor so you will not face the build_start_indices not implemented in dpcpp. we also have the duplicated test in hip/cuda/omp actually.
Using disable_executor makes sense in duplication aspect although it changes my first thought on the test folder.
I draw back my comment on moving test. the changes on duplication is more painful.

Could you also add the TODO inside the code? for example the num_range == 1 check situation, or something else is mentioned in the review.

upsj:
The code will be used in DPC++ as well once https://github.com/oneapi-src/oneDPL/pull/388 is merged and released. This issue is the only reason we can't merge the DPC++ code right away.

MarcelKoch:
I will put the kernels from `partition_kernels.hpp.inc` into the `common/cuda_hip` file directly. We can refactor that again, when the dpcpp stuff becomes ready. I hope that will resolve all these issues.

I will add a todo for the dpcpp implementation, but I think the num_range == 1 check I had previously is now addressed in the documentation.

However, I will wait on pushing these changes, until I've fixed the cuda 9.2 issue.

reference/test/distributed/partition_kernels.cpp
```diff
@@ -0,0 +1,339 @@
...
+    assert_equal_data(partition->get_range_starting_indices(), {0, 0, 0, 0, 0});
+    assert_equal_data(partition->get_part_sizes(), {5, 0, 2, 2, 1});
+}
+
+
+TYPED_TEST(Partition, BuildsFromRangeWithSingleElement)
+{
+    using local_index_type = typename TestFixture::local_index_type;
+    using global_index_type = typename TestFixture::global_index_type;
+    gko::Array<global_index_type> ranges{this->ref, {0}};
```
yhmtsai, line 166: what happen when it is not zero?

include/ginkgo/core/distributed/partition.hpp
```diff
@@ -0,0 +1,321 @@
...
+        std::shared_ptr<const Executor> exec,
+        const Array<comm_index_type>& mapping, comm_index_type num_parts);
+
+    /**
+     * Builds a partition consisting of contiguous ranges, one for each part.
+     *
+     * @param exec  the Executor on which the partition should be built
+     * @param ranges  the boundaries of the ranges representing each part.
+                      Part i contains the indices [ranges[i], ranges[i + 1]).
+                      Has to contain at least one element.
```
yhmtsai, line 255: I guess here? I can not get the part how to handle one element. Do I miss anything?

test/distributed/partition_kernels.cpp
```diff
@@ -0,0 +1,474 @@
...
+    using local_index_type = typename TestFixture::local_index_type;
+    using global_index_type = typename TestFixture::global_index_type;
+    comm_index_type num_parts = 7;
+    std::uniform_int_distribution<comm_index_type> part_dist{0, num_parts - 1};
+    auto mapping = gko::test::generate_random_array<comm_index_type>(
+        10000, part_dist, this->rand_engine, this->ref);
+    gko::Array<comm_index_type> dmapping{this->exec, mapping};
+
+    auto part = gko::distributed::Partition<
+        local_index_type, global_index_type>::build_from_mapping(this->ref,
```
yhmtsai, line 142: because use many gko::distributed::Partition<local_index_type, global_index_type>, maybe use alias for it

include/ginkgo/core/distributed/partition.hpp
MarcelKoch: You're right, I was thinking about something else. To clarify, I added the comment to prevent issues from using `size(ranges) - 1`. If you pass just one element you will still get an empty range.

reference/test/distributed/partition_kernels.cpp
MarcelKoch: I've clarified the documentation, that the first element of the ranges needs always to be 0. Still, it wouldn't matter, you would still get an empty partition, but with `range_bounds = {0}`, regardless of what element was passed.

include/ginkgo/core/distributed/partition.hpp
yhmtsai: I see. that makes sense

reference/test/distributed/partition_kernels.cpp
yhmtsai: Thanks for clarification

yhmtsai (approved):
LGTM. Is the cuda9.2 issue also from executing some kernel with zero grid size?

upsj:
@yhmtsai Yes, as far as I can tell, Thrust internally checks whether a previous kernel launch failed, and even though this kind of failure never cascades, it means that any Thrust kernel call after an empty CUDA kernel call causes an exception to be thrown.

yhmtsai:
I see. There were some kernels with zero grid size but do not give the error in gtest because no thrust kernel after that to raise it.

PR #913, opened 2021-10-26
Build smoother on current level fine not previous level coarse

yhmtsai:
It only changes the behavior when building mixed precision multigrid.
Originally, the smoother (float) uses coarse matrix from previous level (double), so it will lead many precision conversion inside the smoother.
MultigridLevel will set the fine_op after conversion, so it will give the float matrix.
Create smoother on the fine matrix such that the smoother uses the precision of the current level


tcojean (approved):
LGTM

yhmtsai:
format-rebase!

pratikvn (approved):
LGTM!

PR #917, opened 2021-11-02
Remove multiple RHS functionality from reference and OpenMP batch code

Slaedr:
To have a uniform interface for batch stopping criteria and loggers across all backends, it is necessary to remove multi-RHS features from the backends that currently support them  - reference and omp. As background, multiple RHS capability was removed from Cuda because it made it difficult to get good performance, and there are no known use cases for this functionality.


reference/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -345,28 +283,27 @@ inline void batch_entry_bicgstab_impl(
...
-    StopType stop(nrhs, opts.max_its, opts.residual_tol, rhs_norms_entry.values,
-                  converged);
+    StopType stop(opts.residual_tol, rhs_norms_entry.values);
 
     int iter = -1;
 
     while (1) {
         ++iter;
 
+        // bool all_converged = stop.check_converged(iter,
```
pratikvn, line 293: commented code, here and in a few other places

pratikvn (approved):
LGTM! Some minor comments are left out, but otherwise nothing major.

Sorry that you had to go to this trouble of writing all the multiple rhs code in the first place. It was actually some pretty nice functionality. Maybe it would be nice to archive it somewhere, in case we need to have a look at it later.

reference/solver/batch_bicgstab_kernels.hpp.inc
Slaedr: Ah yes, I've removed them in the next PR which is coming up shortly.

PR #918, opened 2021-11-03
Unified dispatch for all backends and batched solvers

Slaedr:
This introduces a "batch dispatcher" to handle dispatching to the correct template instantiation of batched solvers from runtime solvers.

Each solver/backend provides a "kernel launcher" class with a specific interface, into which all solver-specific and backend-specific setup logic goes. This launcher is templated on all the relevant parameters: batch matrix type, precondition type, stopping criterion type and logger type (in addition to the value type) This kernel launcher is then used by the dispatcher, the latter providing the template parameters.

Additionally, matrix-dependent block size is now added to hip batch BiCGStab.

TODO:
- [x] HIP backend kernel dispatch


core/solver/batch_dispatch.hpp.inc
```diff
@@ -0,0 +1,181 @@
...
+ *   kernel.
+ * @tparam ExecType  Executor to run the kernel on.
+ * @tparam OptsType  Structure type of options for the particular solver to be
+ *   used.
+ * @tparam ValueType  The user-facing value type.
+ * @tparam DeviceValueType  The backend-specific value type corresponding to
+ *   ValueType.
+ */
+template <typename KernelCaller, typename ExecType, typename OptsType,
+          typename ValueType, typename DeviceValueType>
```
pratikvn, line 62: Why have both ValueType and DeviceValueType, I think there is a one to one correspondence, so you should be able to cast one to the other and there should be no need for templating ? 

I think a single ValueType makes it easy for mixed precision inference later as well.

cuda/solver/batch_idr_kernels.cu
```diff
@@ -83,64 +84,46 @@ namespace batch_idr {
...
-
-        if (opts.tol_type == gko::stop::batch::ToleranceType::absolute) {
-            BATCH_IDR_KERNEL_LAUNCH(SimpleAbsResidual, BatchJacobi);
-        } else {
-            BATCH_IDR_KERNEL_LAUNCH(SimpleRelResidual, BatchJacobi);
-        }
-    } else {
-        GKO_NOT_IMPLEMENTED;
+template <typename CuValueType>
+class KernelCaller {
```
pratikvn, line 88: Instead of declaring this class in every solver file, you could declare it in a common (across solvers and back-ends) file with some default implementations if possible and override the method in each solver and back-end.

Maybe some Macro magic (replacing Solver names) and variadic macros will help reducing a lot of this code duplication,

pratikvn (commented):
Really nice concept! Some initial comments.

core/solver/batch_dispatch.hpp.inc
Slaedr: I'll try to think about how that can be done. The issue is this same code is used for all backends, so that correspondence depends on the backend.

cuda/solver/batch_idr_kernels.cu
Slaedr: In general, different solvers might also need different constructors for the `KernelCaller`. See the cuda batch IDR, for example. Different solvers would have different shared shared memory logic and stuff too. So we would not be saving much, and it's not worth introducing macros for this. In the future, we can see if more things can be abstracted into some kind of parent class, at least for each solver across all backends.

Slaedr:
format!

core/solver/batch_dispatch.hpp.inc
Slaedr: This is now addressed.

PR #922, opened 2021-11-11
Add abidiff action

### LINKED ISSUE #776: Add ABI stability check
Add an [`abidiff`](https://sourceware.org/libabigail/manual/) against `develop` `libginkgo.so` run at some point, notifying us whether a PR changes the ABI of our shared library. Might be useful when tackling #715 and evaluating whether something is part of the public binary interface

upsj:
Add a Github Action that runs abidiff on the old and new version of the library in a PR and reports the difference if they differ.

You can see this in action at https://github.com/upsj/ginkgo-bot-playground/pull/14, where it reports the change to Sellp and all ripple effects.

Fixes #776

### DIFF
--- a/.github/workflows/bot-pr-updated.yml
+++ b/.github/workflows/bot-pr-updated.yml
@@ -22,3 +22,39 @@ jobs:
         with:
           name: patch
           path: format.patch
+  abidiff:
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.author_association == 'COLLABORATOR' || github.event.pull_request.author_association == 'MEMBER' || github.event.pull_request.author_association == 'OWNER'
+    env:
+      CMAKE_FLAGS: -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=DEBUG -DGINKGO_BUILD_TESTS=OFF -DGINKGO_BUILD_EXAMPLES=OFF -DGINKGO_BUILD_BENCHMARKS=OFF -DGINKGO_BUILD_HWLOC=OFF -DGINKGO_BUILD_REFERENCE=OFF -DGINKGO_BUILD_OMP=OFF -DGINKGO_BUILD_CUDA=OFF -DGINKGO_BUILD_HIP=OFF -DGINKGO_BUILD_DPCPP=OFF
+    steps:
+      - name: Checkout the new code (shallow clone)
+        uses: actions/checkout@v2
+        with:
+          path: new
+          ref: ${{ github.event.pull_request.head.ref }}
+      - name: Checkout the old code (shallow clone)
+        uses: actions/checkout@v2
+        with:
+          path: old
+          ref: develop
+      - name: Install abidiff
+        run: sudo apt-get install abigail-tools
+      - name: Build both libraries
+        run: |
+          mkdir build-new
+          mkdir build-old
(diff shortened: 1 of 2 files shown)


upsj:
Maybe for some background: Any time we change anything that is part of the exported symbols and debug information, this action would compare the old and new Ginkgo library and report any impact these changes have (if any). That might also include internal symbols, but once we deal with #715, will be limited to those exposed in the public interface, i.e. `include/ginkgo`.

We don't necessarily need this, though ABI stability might be an interesting concern at some point (and I always wanted to play around with `abidiff` :smile: )

yhmtsai:
Can we merge the pr with 2 approve even there's a pipeline is failed?
I am not sure whether the job with `if: failure()` will cancel the jobs in the same workflow or not.
If it will cancel, need `fail-fast: false` in jobs

upsj:
Yes, Sonarqube fails regularly for example

tcojean:
Would it be possible to use this action instead to post the file as a comment? So that the size doesn't blow up, the result of the `--stat` option of `abidiff` could be posted as a comment instead with the detail uploaded as an artifact.
https://github.com/marketplace/actions/pr-comment

.github/workflows/bot-pr-updated.yml
```diff
@@ -22,3 +22,37 @@ jobs:
...
+    - name: Checkout the new code (shallow clone)
+      uses: actions/checkout@v2
+      with:
+        path: new
+        ref: ${{ github.event.pull_request.head.ref }}
+    - name: Checkout the old code (shallow clone)
+      uses: actions/checkout@v2
+      with:
+        path: old
+        ref: develop
```
tcojean, line 40: This might produce clutter when creating a PR to merge to master. I think it's best to filter by PR target to `develop`? Otherwise if possible use the PR target instead here.

tcojean (approved):
Nice, I like it.

.github/workflows/bot-pr-updated.yml
pratikvn: Maybe it is better also to always use master as the old ref ? Maybe it is more important that we are ABI compatible with the current release tag (on master) rather than the target branch ? 

We could then also additionally cache the master build in a docker container and speed up this job by two times. Maybe this is not that important as we dont really build any tests or backend kernels.

yhmtsai: I guess you mean `${{ github.event.pull_request.base.ref }}`?

yhmtsai (approved):
Not sure how we take care of Abi-diff during reviewing

.github/workflows/bot-pr-updated.yml
yhmtsai: master is a good point because we need to keep public interface during the minor version update
However, I also wonder there are too many changes such that we can not get good information from the current branch

tcojean: I don't think we should make the master the base, for every PR, otherwise, with every new PR we will get more and more lines of `ABIDIFF` change. If this was enforcing ABI stability (which is hard, IHMO), then master would be a good base point, indeed.

upsj:
I'll merge this directly, since it doesn't change any code/build stuff

upsj:
rebase!

PR #923, opened 2021-11-12
Make ginkgo.pc libraries be found recursively

tcojean:
Small improvement for packages relying on our `ginkgo.pc`

This makes the libraries be found recursively so that `-lhwloc` which is also a dependency of `ginkgo_device` is found automatically. The aim is that this would also adapt to any other library structure change. This solution is obviously not bulletproof, but it should work in most cases.

Note that CMake on their side is working on pkg-config file creating support, but it's still ongoing AFAIK.
https://gitlab.kitware.com/cmake/cmake/-/issues/22621

This is required for OpenCARP and I already use it there: 
https://github.com/ginkgo-project/ginkgo/commit/1dd2dc73302e915217ca95f47f3084d25c78febd


cmake/information_helpers.cmake
```diff
@@ -1,11 +1,24 @@
...
+            get_target_property(GINKGO_LIBS_INTERFACE_LIBS "${_libs}" INTERFACE_LINK_LIBRARIES)
+            ginkgo_interface_libraries_recursively("${GINKGO_LIBS_INTERFACE_LIBS}")
+        endif()
+    endforeach()
+endmacro()
+
 macro(ginkgo_interface_information)
     get_target_property(GINKGO_INTERFACE_LINK_LIBRARIES ginkgo INTERFACE_LINK_LIBRARIES)
     set(GINKGO_INTERFACE_LINK_FLAGS "-L${CMAKE_INSTALL_PREFIX}/${GINKGO_INSTALL_LIBRARY_DIR} -lginkgo")
     set(GINKGO_INTERFACE_CXX_FLAGS "-I${CMAKE_INSTALL_PREFIX}/${GINKGO_INSTALL_INCLUDE_DIR}")
```
upsj, line 14: We may also need to propagate include paths here (hwloc, MPI, ...) to be future-proof?

thoasm (approved):
LGTM!

upsj (approved):
LGTM! Seems to work like it is supposed to, I guess the order of the recursion takes care of the correct link order?

cmake/information_helpers.cmake
```diff
@@ -1,11 +1,24 @@
+macro(ginkgo_interface_libraries_recursively INTERFACE_LIBS)
+    foreach(_libs ${INTERFACE_LIBS})
+        if (TARGET ${_libs})
+            list(APPEND GINKGO_INTERFACE_LIBS_FOUND "-l${_libs}")
+            get_target_property(GINKGO_LIBS_INTERFACE_LIBS "${_libs}" INTERFACE_LINK_LIBRARIES)
+            ginkgo_interface_libraries_recursively("${GINKGO_LIBS_INTERFACE_LIBS}")
+        endif()
```
yhmtsai, line 7: infinite loop? if -lliba -llibb -lliba for some specific cases

yhmtsai (commented):
why do we need to find the hwloc of device? Isn't it required automatically when we link lib ginkgo_device?

cmake/information_helpers.cmake
tcojean: Technically it's, for now, the same directory as we install hwloc if it wasn't present in the user environment/config settings. I guess that can be done as well?

About the linking order, for now I have to say it's fairly random. I believe the recursion order should indeed take care of that, but I wouldn't say there is any guarantee here.

tcojean:
This has been updated and complexified by quite a margin, in particular, to deal properly with MPI and include more link options.

This is what is now generated in a simple hwloc+MPI setting:
``` CMake

prefix=/home/user/0.INSTALL/ginkgo
libdir=${prefix}/lib64
includedir=${prefix}/include

Name:
Description: A numerical linear algebra library targeting many-core architectures
Version: 1.5.0
URL: https://ginkgo-project.github.io/

Requires:
Libs:  -L/home/user/0.INSTALL/ginkgo/lib64 -lginkgo -lginkgo_device /usr/lib64/libhwloc.so -lginkgo_omp -lginkgo_cuda -lginkgo_reference -lginkgo_hip -lginkgo_dpcpp -Wl,-rpath -Wl,/usr/lib64 -Wl,-rpath -Wl,/usr/lib64 -Wl,--enable-new-dtags -pthread /usr/lib64/libmpi.so
Cflags: -I/home/user/0.INSTALL/ginkgo/include -I/usr/include -pthread
```

CMakeLists.txt
```diff
@@ -335,6 +335,8 @@ endif()
 
 configure_file(${Ginkgo_SOURCE_DIR}/cmake/ginkgo.pc.in
     ${Ginkgo_BINARY_DIR}/ginkgo.pc @ONLY)
+file(GENERATE OUTPUT ${Ginkgo_BINARY_DIR}/ginkgo.pc
+     INPUT ${Ginkgo_BINARY_DIR}/ginkgo.pc)
```
upsj, line 339: This looks like it would cause issues for multi-config generators, and if the two versions of ginkgo.pc don't match. CMake actually complains about that in my case.
```suggestion
configure_file(${Ginkgo_SOURCE_DIR}/cmake/ginkgo.pc.in
    ${Ginkgo_BINARY_DIR}/ginkgo.pc.in @ONLY)
file(GENERATE OUTPUT ${Ginkgo_BINARY_DIR}/$<CONFIG>/ginkgo.pc
     INPUT ${Ginkgo_BINARY_DIR}/ginkgo.pc.in)
```

cmake/information_helpers.cmake
```diff
@@ -1,11 +1,98 @@
...
+            if (upper_CMAKE_BUILD_TYPE STREQUAL "DEBUG" AND "${_libs}" MATCHES "ginkgo.*")
+                set(GINKGO_INTERFACE_LIB_NAME "-l${_libs}${CMAKE_DEBUG_POSTFIX}")
+            elseif("${_libs}" MATCHES "ginkgo.*") # Ginkgo libs are appended in the form -l
+                set(GINKGO_INTERFACE_LIB_NAME "-l${_libs}")
+            endif()
+
+            # Get the link flags and treat them
+            get_target_property(GINKGO_INTERFACE_LIBS_LINK_FLAGS "${_libs}"
+                INTERFACE_LINK_OPTIONS)
+            if (GINKGO_INTERFACE_LIBS_LINK_FLAGS)
```
upsj, line 22: need to treat [SHELL:](https://cmake.org/cmake/help/latest/prop_tgt/LINK_OPTIONS.html?highlight=shell#option-de-duplication) correctly. I see `SHELL:-pthread` in my output

cmake/information_helpers.cmake
```diff
@@ -1,11 +1,98 @@
+function(filter_generator_expressions INPUT OUTPUT)
+    string(REGEX REPLACE ".+LINK:.+:(.+)>" "\\1"
+                    TMP "${INPUT}")
+    string(REGEX REPLACE ".+\\$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>:.+>(.+)" "\\1"
+                    TMP "${TMP}")
+    string(REGEX REPLACE ".+INTERFACE:.+>" "" TMP "${TMP}")
```
upsj, line 6: Where do all of these occur? I don't see them in my output.

upsj (changes requested):
There seem to be a few issues when running CMake on my system

cmake/information_helpers.cmake
tcojean: In my version, 3.20.5, they appear. https://gitlab.kitware.com/cmake/cmake/-/blob/v3.20.5/Modules/FindMPI.cmake#L1175

Which CMake version do you use so that I can test the other problems? I should also browse the different versions to see if there are any more of those annoying generators which they wrap everything within the FindMPI.

upsj: I am using 3.16.3

tcojean: This should work like expected. I tried a local cmake-3.16.3 install and the extra `SHELL:` should be taken care of. The only thing I couldn't try are the two concurrent builds, but since everything uses generator expressions now I don't think you should see the previous issues anymore?

cmake/install_helpers.cmake
```diff
@@ -71,7 +71,8 @@ endfunction()
 
 function(ginkgo_install)
     # pkg-config file
-    install(FILES "${Ginkgo_BINARY_DIR}/ginkgo.pc" DESTINATION "${GINKGO_INSTALL_PKGCONFIG_DIR}")
+    install(FILES ${Ginkgo_BINARY_DIR}/ginkgo_$<CONFIG>.pc
+        DESTINATION "${GINKGO_INSTALL_PKGCONFIG_DIR}")
```
upsj, line 75:
```suggestion
        DESTINATION "${GINKGO_INSTALL_PKGCONFIG_DIR}"
        RENAME ginkgo.pc)
```

upsj (changes requested):
I tested with all major CMake versions since 3.13 and different configs (with MVAPICH, OpenMPI or no MPI, with vendored and installed HWLOC).

There are still two issues with the implementation
* When using vendored `hwloc`, the pkg-config file contains the the build location of hwloc, not the installed location
* The installed file probably needs to be renamed

cmake/information_helpers.cmake
```diff
@@ -1,11 +1,104 @@
...
+    # Prepare recursively populated library list
+    string(TOUPPER "${CMAKE_BUILD_TYPE}" upper_CMAKE_BUILD_TYPE)
+    if (upper_CMAKE_BUILD_TYPE STREQUAL "DEBUG")
+        list(APPEND GINKGO_INTERFACE_LIBS_FOUND "-lginkgo${CMAKE_DEBUG_POSTFIX}")
+    else()
+        list(APPEND GINKGO_INTERFACE_LIBS_FOUND "-lginkgo")
+    endif()
+    # Prepare recursively populated include directory list
+    list(APPEND GINKGO_INTERFACE_CFLAGS_FOUND
+        "-I${CMAKE_INSTALL_PREFIX}/${GINKGO_INSTALL_INCLUDE_DIR}")
```
yhmtsai, line 83:
```suggestion
        "-I\${includedir}")
```
should it reuse the variable in pc like includedir or something else?

cmake/information_helpers.cmake
```diff
@@ -1,11 +1,104 @@
...
+                INTERFACE_COMPILE_OPTIONS)
+            filter_generator_expressions("${GINKGO_LIBS_INTERFACE_OPTS}" GINKGO_LIBS_INTERFACE_OPTS_FILTERED)
+            if (GINKGO_LIBS_INTERFACE_OPTS)
+                list(APPEND GINKGO_INTERFACE_CFLAGS_FOUND "${GINKGO_LIBS_INTERFACE_OPTS_FILTERED}")
+            endif()
+
+            # Keep recursing through the libraries
+            get_target_property(GINKGO_LIBS_INTERFACE_LIBS "${_libs}"
+                INTERFACE_LINK_LIBRARIES)
+            ginkgo_interface_libraries_recursively("${GINKGO_LIBS_INTERFACE_LIBS}")
```
yhmtsai, line 63: I think it will be a problem when ginkgo contain circular link? but we should not have this kind of issue because we always check the circular deps?

yhmtsai: when _libs is already in the list, it does not need to go recursively again.
for example, many ginkgo link ginkgo_device and ginkgo_device try to find hwloc

yhmtsai (commented):
hwloc is a target and then it contains the hwloc.so file.
use it directly because the first condition only handles ginkgo lib.
recursive_search(hwloc.so) existed condition path, add it directly into link flag.
Is it the expected behavior?

cmake/information_helpers.cmake
```diff
@@ -1,11 +1,113 @@
+function(filter_generator_expressions INPUT OUTPUT)
+    # See https://gitlab.kitware.com/cmake/cmake/-/blob/v3.22.2/Modules/FindMPI.cmake#L1218
+    # and other versions of this file for what we are removing here.
+    string(REGEX REPLACE "[$<]+COMPILE.*>:.+[>]+(.+)" "\\1" TMP "${INPUT}")
+    # There can be at least two type of SHELL, one with generator $<> form, one
+    # without (v3.16.x). Sometimes, it also has extra arguments. We need to do
+    # at least a greedy regex also consuming the final `>` if present before
+    # doing a non greedy one for the leftovers.
```
yhmtsai, line 8: do you have some example input here?

yhmtsai (approved):
LGTM. some nit. If you have examples about the SHELL, that will be great. @upsj already check all cmake output, so I do not wonder about the correctness.

thoasm (approved):
LGTM

tcojean:
If there is no more issues, I would merge this once the pipeline succeeds.

upsj:
Multi-Config generators currently don't generate the correct library name for debug builds due to `CMAKE_DEBUG_SUFFIX`. A fix for that would be using [`TARGET_LINKER_FILE_BASE_NAME`](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:TARGET_LINKER_FILE_BASE_NAME), but that requires CMake 3.15, so we will update it in a later version. Corresponding diff:
```diff
diff --git a/cmake/information_helpers.cmake b/cmake/information_helpers.cmake
index 5cf80368c8..037f39e4a7 100644
--- a/cmake/information_helpers.cmake
+++ b/cmake/information_helpers.cmake
@@ -20,8 +20,8 @@ macro(ginkgo_interface_libraries_recursively INTERFACE_LIBS)
         if (NOT "${_libs}" IN_LIST GINKGO_INTERFACE_LIBS_FOUND
                 AND NOT "-l${_libs}" IN_LIST GINKGO_INTERFACE_LIBS_FOUND)
             if (TARGET ${_libs})
-                if (upper_CMAKE_BUILD_TYPE STREQUAL "DEBUG" AND "${_libs}" MATCHES "ginkgo.*")
-                    set(GINKGO_INTERFACE_LIB_NAME "-l${_libs}${CMAKE_DEBUG_POSTFIX}")
+                if ("${_libs}" MATCHES "ginkgo.*")
+                    set(GINKGO_INTERFACE_LIB_NAME "-l$<TARGET_LINKER_FILE_BASE_NAME:${_libs}>")
                 elseif("${_libs}" MATCHES "ginkgo.*") # Ginkgo libs are appended in the form -l
                     set(GINKGO_INTERFACE_LIB_NAME "-l${_libs}")
                 endif()
@@ -81,12 +81,7 @@ macro(ginkgo_interface_information)
     unset(GINKGO_INTERFACE_LIBS_FOUND)
     unset(GINKGO_INTERFACE_CFLAGS_FOUND)
     # Prepare recursively populated library list
-    string(TOUPPER "${CMAKE_BUILD_TYPE}" upper_CMAKE_BUILD_TYPE)
-    if (upper_CMAKE_BUILD_TYPE STREQUAL "DEBUG")
-        list(APPEND GINKGO_INTERFACE_LIBS_FOUND "-lginkgo${CMAKE_DEBUG_POSTFIX}")
-    else()
-        list(APPEND GINKGO_INTERFACE_LIBS_FOUND "-lginkgo")
-    endif()
+    list(APPEND GINKGO_INTERFACE_LIBS_FOUND "-l$<TARGET_LINKER_FILE_BASE_NAME:ginkgo>")
     # Prepare recursively populated include directory list
     list(APPEND GINKGO_INTERFACE_CFLAGS_FOUND
         "-I${CMAKE_INSTALL_PREFIX}/${GINKGO_INSTALL_INCLUDE_DIR}")
```

PR #924, opened 2021-11-19
dpcpp port isai (row_limit=32)

yhmtsai:
It ports the isai to dpcpp module.

TODO:
- [x] ~~move the device_numeric_limit back to math~~ use std directly.
`device_numeric_limits<IndexType>::max` -> `std::numeric_limits<IndexType>::max()` needs () in the end
- [x] clean some dpct warning, which are fixed
- [ ] use subgroup size for row_limit. it might be in different pr


upsj (approved):
LGTM! Just needs tests for the dpcpp searching/merging kernels. We may need to evaluate the performance of the merging kernels on SYCL, since they use random access shuffles, which may perform really poorly on architectures with insufficient HW support?

I guess I kind of dislike mixing get_thread_id_flat with 3D indexing via dim3, but of course that is the straightforward way to port the code.

dpcpp/preconditioner/isai_kernels.dp.cpp
```diff
@@ -33,54 +33,621 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+    } else {
+        // handle short rows directly: no excess
+        if (local_id == 0) {
+            excess_rhs_sizes[row] = 0;
+            excess_nnz[row] = 0;
+        }
+
+        // subwarp_size^2 storage per subwarp
+
+
```
fritzgoebel, line 143:
```suggestion
```

fritzgoebel (approved):
LGTM, only a few nits

yhmtsai:
format!

PR #926, opened 2021-11-23
Small fix for unified kernel reduction

pratikvn:
Capturing by reference seems to cause issues on Summit, but capturing by value seems to work. This should not have any performance effects, due to it being a simple scalar value.

### DIFF
--- a/cuda/base/kernel_launch_reduction.cuh
+++ b/cuda/base/kernel_launch_reduction.cuh
@@ -458,7 +458,7 @@ void run_kernel_row_reduction(std::shared_ptr<const CudaExecutor> exec,
     } else {
         select_run_generic_kernel_row_reduction(
             subwarp_sizes(),
-            [&](int compiled_subwarp_size) {
+            [=](int compiled_subwarp_size) {
                 return compiled_subwarp_size >= cols ||
                        compiled_subwarp_size == config::warp_size;
             },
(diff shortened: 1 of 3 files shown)


cuda/base/kernel_launch_reduction.cuh
```diff
@@ -458,7 +458,7 @@ void run_kernel_row_reduction(std::shared_ptr<const CudaExecutor> exec,
     } else {
         select_run_generic_kernel_row_reduction(
             subwarp_sizes(),
-            [&](int compiled_subwarp_size) {
+            [=](int compiled_subwarp_size) {
```
upsj, line 461: I think an explicit capture would be safest here
```suggestion
            [cols](int compiled_subwarp_size) {
```

upsj (approved):
LGTM! Can you add the same changes for HIP (and maybe DPC++ if appropriate?)

yhmtsai (approved):
I agree with that it should not change anything about the performance and it only takes the value on the cpu side

dpcpp/base/kernel_launch_reduction.dp.hpp
```diff
@@ -612,7 +612,7 @@ void run_kernel_col_reduction_stage1(std::shared_ptr<const DpcppExecutor> exec,
     if (cols <= sg_size) {
         select_generic_col_reduction_small(
             subsubgroup_sizes(),
-            [&](int compiled_ssg_size) {
+            [cols](int compiled_ssg_size) {
                 return compiled_ssg_size >= cols ||
                        compiled_ssg_size == sg_size;
```
tcojean, line 617: I guess `sg_size` is always a compile value so it's fine like this?

tcojean (approved):
LGTM.

Since it's only integers this should actually be better this way.

dpcpp/base/kernel_launch_reduction.dp.hpp
pratikvn: Yeah, I think so. I will see if it compiles. I was too lazy to try and compile it myself with dpcpp :P

upsj:
rebase!

PR #927, opened 2021-11-24
Enable batch dense to be used as matrix type for batch solvers

Slaedr:
- Adds missing batch dense kernels.
- Enables testing of solvers with different batch matrix types.
- Enables benchmarking of solvers with batch dense as well.

### DIFF
--- a/benchmark/utils/formats.hpp
+++ b/benchmark/utils/formats.hpp
@@ -172,6 +172,7 @@ using hybrid = gko::matrix::Hybrid<etype, itype>;
 using csr = gko::matrix::Csr<etype, itype>;
 using batch_csr = gko::matrix::BatchCsr<etype>;
 using batch_ell = gko::matrix::BatchEll<etype>;
+using batch_dense = gko::matrix::BatchDense<etype>;
 
 /**
  * Creates a Ginkgo matrix from the intermediate data representation format
@@ -319,7 +320,9 @@ const std::map<std::string, std::function<std::unique_ptr<gko::BatchLinOp>(
         {"batch_ell",
          read_batch_matrix_from_data<gko::matrix::BatchEll<etype>>},
         {"batch_csr",
-         read_batch_matrix_from_data<gko::matrix::BatchCsr<etype>>}};
+         read_batch_matrix_from_data<gko::matrix::BatchCsr<etype>>},
+        {"batch_dense",
+         read_batch_matrix_from_data<gko::matrix::BatchDense<etype>>}};
 
 
 // clang-format off
--- a/benchmark/utils/formats.hpp
+++ b/benchmark/utils/formats.hpp
@@ -170,9 +170,6 @@ namespace formats {
 // some shortcuts
 using hybrid = gko::matrix::Hybrid<etype, itype>;
 using csr = gko::matrix::Csr<etype, itype>;
-using batch_csr = gko::matrix::BatchCsr<etype>;
-using batch_ell = gko::matrix::BatchEll<etype>;
-using batch_dense = gko::matrix::BatchDense<etype>;
 
 /**
  * Creates a Ginkgo matrix from the intermediate data representation format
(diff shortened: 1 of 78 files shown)


benchmark/utils/formats.hpp
```diff
@@ -172,6 +172,7 @@ using hybrid = gko::matrix::Hybrid<etype, itype>;
 using csr = gko::matrix::Csr<etype, itype>;
 using batch_csr = gko::matrix::BatchCsr<etype>;
 using batch_ell = gko::matrix::BatchEll<etype>;
+using batch_dense = gko::matrix::BatchDense<etype>;
```
Slaedr, line 175: I guess this is not actually needed; I'll remove it next time.

core/test/utils/batch.hpp
```diff
@@ -268,10 +278,58 @@ BatchSystem<ValueType> generate_solvable_batch_system(
...
+        mtx->write(mdata2);
+        if (mdata.size() != mdata2.size()) {
+            throw std::runtime_error("sizes not same");
+        }
+        for (size_t ib = 0; ib < mdata.size(); ib++) {
+            if (mdata[ib].nonzeros != mdata2[ib].nonzeros) {
+                throw std::runtime_error("Not same");
+            }
+        }
+    }
```
Slaedr, line 299: Remove

cuda/solver/batch_bicgstab_kernels.cu
```diff
@@ -154,7 +154,8 @@ public:
         const size_type nbatch = a.num_batch;
         const int shared_gap = ((a.num_rows - 1) / 8 + 1) * 8;
 
-        // TODO: Add function to BatchCSR to return storage needed per matrix
+        // TODO: Add function to batch matrix types to return storage needed per
```
Slaedr, line 157: Fix

Slaedr:
For getting the matrix storage requirements, I added small functions to the uniform batch structs. It's a bit delicate in terms of performance because these structs are passed to the device kernels, but because it's an inline function it should not matter. I could also add a variable, and that would also probably be fine, but it would at least increase the constant memory usage for holding the arguments. I'm guessing the device compiler should be able to throw away unused inline functions. I also checked the performance on summit for gri30 + csr + bicgstab, and there's no regression.

core/solver/batch_bicgstab.cpp
```diff
@@ -146,13 +145,10 @@ void BatchBicgstab<ValueType>::apply_impl(const BatchLinOp* b,
     if (acsr) {
```
pratikvn, line 145: Maybe this if can also be removed now ?

common/cuda_hip/matrix/batch_dense_kernels.hpp.inc
```diff
@@ -35,6 +35,109 @@ template <typename ValueType>
 using BatchEntry = gko::batch_dense::BatchEntry<ValueType>;
 
 
+template <typename ValueType>
+__device__ __forceinline__ void single_spmv_kernel(
```
pratikvn, line 39: Maybe rename to `singke_apply_kernel` ? It is not sparse in this case :)

pratikvn (approved):
Really good work! There is some commented code, but otherwise looks good to me.

common/cuda_hip/matrix/batch_dense_kernels.hpp.inc
Slaedr: Yeah, I noticed that, but the name needs to be the same across all matrix formats for the batch solvers to work 😅 Maybe we change it to `single_apply_kernel` everywhere, though for free functions that's not as descriptive. Maybe `single_matvec_kernel`?

pratikvn: Ah, yeah. True. Maybe we can rename it to `single_apply_kernel` everywhere.

PR #928, opened 2021-11-25
Dpcpp port ParILU(T)/IC(T)

yhmtsai:
This PR ports the ParILU(T)/IC(T).

TODO:
- [x] Merge #924 
- [x] Move the ffs fix to #924
- [ ] Check the residual changes on float
- [x] Merge the kernel file on dpcpp because it duplicates the same kernel.inc into three file. Or put it as inc in the same folder.

PR #929, opened 2021-11-25
Dpcpp port jacobi

yhmtsai:
This PR ports the Jacobi to dpcpp including scaler and adaptive-block.

Currently, dpcpp can not use different subgroup size inside the kernel, so we do not have the concept, which handles several groups in warp and they share the same precision(that is, all blocks need to be isolated).
there are two part affected by this.
- the constructor from max_block_size: the constructor takes MAX_SIZE/max_block_size as the group size. we need to set the max_block_size as config::warp_size such that group_size = 1 and it will give the same result as reference.
- only generate the 32 kernel
It should only affect the block-jacobi generation with auto-detect (TODO: check)

TODO
- [x] merge #928 #924 
- [ ] ~~check the adaptive-block can be done without changing subgroup size~~
- [x] CPU resource issue
- [x] fix format_header for handling the generated header from cmake (partially not automatically)


dpcpp/preconditioner/jacobi_kernels.dp.cpp
```diff
@@ -33,212 +33,437 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-    const IndexType* prev_row_ptr, const IndexType* curr_row_ptr,
-    const IndexType* next_row_ptr) GKO_NOT_IMPLEMENTED;
-
-
-template <typename ValueType, typename IndexType>
-size_type find_natural_blocks(const matrix::Csr<ValueType, IndexType>* mtx,
-                              uint32 max_block_size,
-                              IndexType* block_ptrs) GKO_NOT_IMPLEMENTED;
+// a total of 8 32-subgroup (256 threads)
+constexpr int default_num_warps = 8;
```
tcojean, line 67: Maybe base this off `GKO_DPCPP_JACOBI_BLOCK_SIZES` (or the max of it), to reduce the magic numbers a bit? We also have `config:::max_block_size` which could be used instead of the `256`?

dpcpp/preconditioner/jacobi_kernels.dp.cpp
```diff
@@ -33,212 +33,437 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+// TODO: get a default_grid_size for dpcpp
+// with current architectures, at most 32 warps can be scheduled per SM (and
+// current GPUs have at most 84 SMs)
+constexpr int default_grid_size = 32 * 32 * 128;
+
+
+template <int warps_per_block>
```
tcojean, line 74: Isn't this unused in the end? There's no launch bounds here which is what this was used for.

tcojean (approved):
LGTM.

dpcpp/CMakeLists.txt
```diff
@@ -76,6 +107,9 @@ if (BUILD_SHARED_LIBS)
 else ()
     target_link_options(ginkgo_dpcpp PUBLIC -fsycl-device-code-split=per_kernel)
 endif()
+# include path for generated headers like jacobi_common.hpp
+target_include_directories(ginkgo_dpcpp
+    PRIVATE ${CMAKE_CURRENT_BINARY_DIR}/..)
```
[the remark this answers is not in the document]
yhmtsai: yes, I think so. we still use `#include <dpcpp/preconditioner/...>` from the total binary folder, which is the parent folder of dpcpp (current binary folder)

dpcpp/test/preconditioner/jacobi_kernels.cpp
```diff
@@ -0,0 +1,937 @@
...
+#include "core/utils/matrix_utils.hpp"
+#include "dpcpp/test/utils.hpp"
+
+
+namespace {
+
+
+class Jacobi : public ::testing::Test {
+protected:
+    using index_type = int32_t;
```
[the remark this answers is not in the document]
yhmtsai: it's used by the `generate_random_matrix_data`

dpcpp/components/diagonal_block_manipulation.dp.hpp
```diff
@@ -0,0 +1,128 @@
...
+#pragma unroll
+    for (int b = 0; b < processed_blocks; ++b, ++bid) {
+        if (bid < num_blocks) {
+            bstart += bsize;
+            bsize = block_ptrs[bid + 1] - bstart;
+#pragma unroll
+            for (int i = 0; i < max_block_size; ++i) {
+                if (i < bsize) {
+                    if (item_ct1.get_local_id(1) == b &&
+                        item_ct1.get_local_id(2) < max_block_size) {
```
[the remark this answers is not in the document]
yhmtsai: I thought about it within porting process. It needs to add MACRO to define the variable like threadIdx, blockIdx.
I also wonder it will take more registers on the kernels, which should be avoid.

dpcpp/test/preconditioner/jacobi_kernels.cpp
```diff
@@ -0,0 +1,937 @@
...
+        if (dpcpp != nullptr) {
+            ASSERT_NO_THROW(dpcpp->synchronize());
+        }
+    }
+
+    void initialize_data(
+        std::initializer_list<gko::int32> block_pointers,
+        std::initializer_list<gko::precision_reduction> block_precisions,
+        std::initializer_list<value_type> condition_numbers,
+        gko::uint32 max_block_size, int min_nnz, int max_nnz, int num_rhs = 1,
```
thoasm, line 86: nit: Requires the header `initializer_list` and `ginkgo/core/base/types.hpp`.

dpcpp/test/preconditioner/jacobi_kernels.cpp
```diff
@@ -0,0 +1,937 @@
...
+        value_type accuracy = 0.1, bool skip_sorting = true)
+    {
+        std::ranlux48 engine(42);
+        const auto dim = *(end(block_pointers) - 1);
+        if (condition_numbers.size() == 0) {
+            mtx = gko::test::generate_random_matrix<Mtx>(
+                dim, dim, std::uniform_int_distribution<>(min_nnz, max_nnz),
+                std::normal_distribution<value_type>(0.0, 1.0), engine, ref);
+        } else {
+            std::vector<mtx_data> blocks;
```
thoasm, line 96: nit: requires header `vector`

dpcpp/test/preconditioner/jacobi_kernels.cpp
```diff
@@ -0,0 +1,937 @@
...
+                    begin(block_pointers)[i + 1] - begin(block_pointers)[i];
+                const auto cond = begin(condition_numbers)[i];
+                blocks.push_back(mtx_data::cond(
+                    size, cond, std::normal_distribution<value_type>(-1, 1),
+                    engine));
+            }
+            mtx = Mtx::create(ref);
+            mtx->read(mtx_data::diag(begin(blocks), end(blocks)));
+        }
+        gko::array<gko::int32> block_ptrs(ref, block_pointers);
```
thoasm, line 108: nit: requires header `<ginkgo/core/base/array.hpp>`

thoasm (approved):
LGTM!

dpcpp/preconditioner/jacobi_advanced_apply_instantiate.inc.dp.cpp
```diff
@@ -0,0 +1,239 @@
...
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+#include "core/preconditioner/jacobi_kernels.hpp"
+
+
+#include <dpcpp/preconditioner/jacobi_common.hpp>
```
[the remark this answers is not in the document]
yhmtsai: yes. it is the issue what I want to fix for format_header.
Unfortunately, I do not find time to finish the fix.
I will manually fix them first

dpcpp/test/preconditioner/jacobi_kernels.cpp
```diff
@@ -0,0 +1,941 @@
...
+#include "core/test/utils.hpp"
+#include "core/test/utils/unsort_matrix.hpp"
+#include "core/utils/matrix_utils.hpp"
+#include "dpcpp/test/utils.hpp"
+
+
+namespace {
+
+
+class Jacobi : public ::testing::Test {
```
upsj, line 59: Post-merge review comment: Could this be merged with test/preconditioner/jacobi_kernels.cpp?

PR #933, opened 2021-11-29
dpcpp ports amgx_pgm

yhmtsai:
This PR ports the amgx_pgm to dpcpp.
the current `thrust::tie` can not be translated to dpcpp by dpct.

@upsj Do you know something like `thrust::tie` inside oneDPL?


upsj:
`std::tie` from `<tuple>`

Slaedr (approved):
LGTM! Looks completely consistent with the cuda implementation. Just one comment below.

upsj (approved):
LGTM as well! I agree with @Slaedr, and would even go one step further: All of these kernels use one thread per row. So you could simplify them heavily by putting them directly into a 1D parallel_for

dpcpp/multigrid/amgx_pgm_kernels.dp.cpp
```diff
@@ -44,6 +45,13 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/multigrid/amgx_pgm.hpp>
 
 
+#include "core/components/fill_array_kernels.hpp"
+#include "core/components/prefix_sum_kernels.hpp"
+#include "dpcpp/base/dim3.dp.hpp"
+#include "dpcpp/components/reduction.dp.hpp"
```
upsj, line 51: seems unused?

yhmtsai:
no, the extra is our own effort.
it gives the same structure as the cuda and hip call including the share memory usage.
when the kernel call many times, the function in dpcpp is hard to align with other executor

yhmtsai:
@upsj @Slaedr I move those kernels to unified and add `device_std` as thrust for cuda/hip and as std for others

dpcpp/test/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,348 @@
...
+#include "core/multigrid/amgx_pgm_kernels.hpp"
+#include "core/test/utils.hpp"
+#include "core/test/utils/matrix_generator.hpp"
+#include "core/test/utils/unsort_matrix.hpp"
+
+
+namespace {
+
+
+class AmgxPgm : public ::testing::Test {
```
Slaedr, line 62: All these tests can now be move to `tests/multigrid`, I guess? Then you could remove `hip/test/multigrid/amgx_pgm_kernels.cpp` etc.

Slaedr (approved):
Excellent work!

Did you try benchmarking your previous implementations against these ones? Since we seem to be discussing multigrid performance quite often, it would be good to document the performance difference between the separate kernels and the unified ones. I guess it's unlikely there's a big difference, but it will be good to document it.

It might make sense to develop more fine-grain parallel kernels for the aggregation kernels, primarily to get good coalescing access, though I understand that is far from trivial. That can be left to another PR.

common/unified/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -0,0 +1,279 @@
...
+        [] GKO_KERNEL(auto row, auto row_ptrs, auto col_idxs, auto weight_vals,
+                      auto diag, auto agg, auto strongest_neighbor) {
+            auto max_weight_unagg = zero<ValueType>();
+            auto max_weight_agg = zero<ValueType>();
+            IndexType strongest_unagg = -1;
+            IndexType strongest_agg = -1;
+            if (agg[row] != -1) {
+                return;
+            }
+            for (auto idx = row_ptrs[row]; idx < row_ptrs[row + 1]; idx++) {
```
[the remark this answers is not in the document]
upsj: I tend to agree here - there is a clear path to using warp-parallelism here in the future. It is a cheap way to implement it in DPC++. There may be a few abstract primitives we could use for the GPU kernels: Csr row-wise reduction + counting, Csr row-wise filtering, ... Might make sense to put that into a file in `components`? But for now I wouldn't block the PR for this.
Probably should take a look at specializing it before the release?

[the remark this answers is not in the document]
yhmtsai: yes it should be parallelized by warp per row.
we do not yet test the performance of the generation step.
at least, the R*A*P also need to improve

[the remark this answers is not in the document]
Slaedr: My only concern is we might have to go back to backend-specific kernels if we need more advanced algorithms. But for now, I think we should move ahead with this.

yhmtsai:
unified ~107.86ms
```
4.81%  3.1193ms         4  779.82us  129.02us  1.8408ms find_strongest_neighbor
0.05%  33.488us         4  8.3720us  6.4610us  10.141us match_edge
```

original ~107.108 (skip_sorting 62.4452)
```
4.82%  3.1674ms         4  791.86us  146.08us  1.8393ms find_strongest_neighbor
0.06%  36.687us         4  9.1710us  6.8120us  10.939us match_edge
```
only check mfem l-shape 1 level generation on V100
it does not give noticeable difference between these.
the main bottleneck is spgemm and sorting.
sorting takes ~ 40 ~ 45% of generation and spgemm takes ~10%
without sorting. spgemm (sum of RadixSort and spgemm_core) takes ~20% and find_strongest_neighbor takes ~10%

test/multigrid/amgx_pgm_kernels.cpp
```diff
@@ -164,9 +168,9 @@ class AmgxPgm : public ::testing::Test {
     }
 
     std::shared_ptr<gko::ReferenceExecutor> ref;
-    std::shared_ptr<const gko::CudaExecutor> cuda;
+    std::shared_ptr<gko::EXEC_TYPE> exec;
 
-    std::default_random_engine rand_engine;
+    std::ranlux48 rand_engine;
```
upsj, line 173: I guess this should be reverted?
```suggestion
    std::default_random_engine rand_engine;
```

yhmtsai: good catch!

PR #936, opened 2021-12-02
Fix the permission of format-rebase!

yhmtsai:
This pr adds the permission of the format-rebase! and add the execution permission inside the comment-bot.
The issue is found at https://github.com/ginkgo-project/ginkgo/runs/4389847667?check_suite_focus=true


upsj (changes requested):
`cp` preserves permissions, so the only necessary fix is the `format-rebase.sh` chmod

.github/workflows/bot-pr-comment.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     - name: Add appropriate labels
       env:
         GITHUB_TOKEN: ${{ secrets.BOT_TOKEN }}
-      run: cp .github/label.sh /tmp && /tmp/label.sh
+      run: cp .github/label.sh /tmp && chmod +x /tmp/label.sh && /tmp/label.sh
```
thoasm, line 17: This applies to all changes below as well:
If you want to make absolutely sure that the mode is preserved while copying, you can use `cp --preserve` or `cp --preserve=mode` (removing the need for `chmod`):
```suggestion
      run: cp --preserve .github/label.sh /tmp && /tmp/label.sh
```

thoasm (approved):
Thanks for looking into and fixing this!

upsj (approved):

thoasm:
@upsj I know that it was missing execution mode. the reason for my comment was that I did not see it actually preserved the mode in the documentation of `cp`. Usually, it does, but since there is the `--preserve` option, I thought it made sense to specify that the mode is preserved for sure.

PR #937, opened 2021-12-07
Update container names

pratikvn:
This PR updates the jobs with the appropriate container names that were updated due to the addition of MPI to the containers. See [ginkgo-container repository MR#32](https://gitlab.com/ginkgo-project/ginkgo-containers/-/merge_requests/32).

Also cleanup and re-organize existing MPI jobs.

The clang-cuda container was also updated to llvm-11, which was missing, so that also has been updated.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -269,7 +269,7 @@ trigger_pipeline:
 # Build jobs
 # Job with example runs.
 # cuda 9.2 and friends
-build/cuda92/gcc/all/release/shared:
+build/cuda92/nompi/gcc/all/release/shared:
   <<: *default_build_with_test
   extends:
     - .quick_test_condition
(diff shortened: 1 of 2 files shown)


.gitlab-ci.yml
```diff
@@ -278,13 +286,14 @@ build/cuda92/gcc/all/release/shared:
 # cuda 10.0 and friends
 # Make sure that our jobs run when using self-installed
 # third-party HWLOC.
-build/cuda100/gcc/all/debug/shared:
+build/cuda100/mvapich2/gcc/all/debug/shared:
```
tcojean, line 289: I think you should have a consistent scheme, either you update the names everywhere or nowhere? I don't mind whichever way as long as it's consistent.

tcojean (approved):
LGTM. Only a minor comment due to one name update.

.gitlab-ci.yml
pratikvn: My reasoning was that these were the job names and hence they reflect what the job is doing. If there is nompi, I left the name out. I guess I could rename jobs that don't build mpi to `build.../nompi/...`. Would that be better ?

MarcelKoch (approved):
LGTM. I would second that you could make the naming scheme better by adding `nompi` if they don't use mpi.

PR #942, opened 2021-12-20
Scaling + addition by scaled identity (`M <- aI + bM`) for Csr and Dense matrices

Slaedr:
Enables `M <- a I + b M` operations when M is Dense or Csr. These are optimized kernels that can be used for:
- assembling the LHS for a backward Euler type time integration of a PDE
- shifting a matrix before precondition construction, eg. for nearly singular problems.

In addition, there's a small bugfix to the 2D common kernel run function. Now, the `const`ness of the pointer to the matrix does not determine whether the entries can be modified; earlier something like `Mtx* const mtx` would not allow the matrix to be modified in the kernel.


reference/test/matrix/identity.cpp
```diff
@@ -221,4 +222,46 @@ TYPED_TEST(Identity, AppliesLinearCombinationToMixedComplex)
...
+    using T = typename TestFixture::value_type;
+    using Csr = typename gko::matrix::Csr<T, int>;
+    auto alpha = gko::initialize<Vec>({2.0}, this->exec);
+    auto beta = gko::initialize<Vec>({-1.0}, this->exec);
+    auto identity = Id::create(this->exec, 3);
+    auto bd = gko::initialize<Vec>(
+        3, {I<T>{2.0, 0.0, -1.0}, I<T>{1.0, 0.0, 2.5}, I<T>{0.0, -1.0, 4.0}},
+        this->exec);
+    auto b = Csr::create(this->exec);
+    bd->convert_to(b.get());
```
pratikvn, line 258: You can directly use `gko::initialize<Csr>`, which automatically converts the Dense matrix into a Csr and gives you back a Csr matrix.

test/matrix/identity_kernels.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+
+TEST_F(Identity, ScaleDenseAddIdentity)
+{
+    set_up_apply_data();
+
+    id->apply(alpha.get(), id.get(), beta.get(), x.get());
+    did->apply(dalpha.get(), did.get(), dbeta.get(), dx.get());
+
+    GKO_ASSERT_MTX_NEAR(x, dx, std::numeric_limits<vtype>::epsilon());
```
pratikvn, line 158: use `r<vtype>::value`

test/matrix/identity_kernels.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+    template <typename MtxType>
+    std::unique_ptr<MtxType> gen_mtx(int num_rows, int num_cols,
+                                     int min_nnz_row)
+    {
+        return gko::test::generate_random_matrix<MtxType>(
+            num_rows, num_cols,
+            std::uniform_int_distribution<>(min_nnz_row, num_cols),
+            std::normal_distribution<>(-1.0, 1.0), rand_engine, ref);
+    }
```
pratikvn, line 105: unused

test/matrix/identity_kernels.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+    std::shared_ptr<gko::ReferenceExecutor> ref;
+    std::shared_ptr<gko::EXEC_TYPE> exec;
+
+    std::ranlux48 rand_engine;
+
+    std::unique_ptr<Id> id;
+    std::unique_ptr<Mtx> x;
+    std::unique_ptr<ComplexMtx> c_x;
+    std::unique_ptr<ComplexMtx> c_alpha;
```
pratikvn, line 137: unused

test/matrix/identity_kernels.cpp
```diff
@@ -0,0 +1,162 @@
...
+    std::unique_ptr<Mtx> alpha;
+    std::unique_ptr<Mtx> beta;
+    std::unique_ptr<Mtx> square;
+    std::unique_ptr<Id> did;
+    std::unique_ptr<Mtx> dx;
+    std::unique_ptr<ComplexMtx> dc_x;
+    std::unique_ptr<ComplexMtx> dc_alpha;
```
pratikvn, line 144: unused

pratikvn (approved):
Nice work. Thanks for adding this so quickly! Mostly minor things. But otherwise looks good!

core/matrix/identity.cpp
```diff
@@ -54,12 +71,38 @@ template <typename ValueType>
...
 {
-    precision_dispatch_real_complex<ValueType>(
-        [this](auto dense_alpha, auto dense_b, auto dense_beta, auto dense_x) {
-            dense_x->scale(dense_beta);
-            dense_x->add_scaled(dense_alpha, dense_b);
-        },
-        alpha, b, beta, x);
+    if (auto bI = dynamic_cast<const Identity<ValueType>*>(b)) {
+        GKO_ASSERT_IS_SQUARE_MATRIX(x);
+        if (auto xd = dynamic_cast<Dense<ValueType>*>(x)) {
```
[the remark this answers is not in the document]
Slaedr: But each time needs a different kernel to be called, so I don't see a simple way around this.

cuda/test/matrix/identity.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+
+TEST_F(Identity, ScaleDenseAddIdentity)
+{
+    set_up_apply_data();
+
+    id->apply(alpha.get(), id.get(), beta.get(), x.get());
+    did->apply(dalpha.get(), did.get(), dbeta.get(), dx.get());
+
+    GKO_ASSERT_MTX_NEAR(x, dx, std::numeric_limits<vtype>::epsilon());
```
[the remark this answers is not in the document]
Slaedr: I think the `r` is 10 times this value, IIRC. This is also more readable, I thought.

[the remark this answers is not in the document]
pratikvn: I think it is better for uniformity to use `r<vtype>::value`. But it is not something major.

pratikvn (dismissed):
I think the tests should be moved to `backend/test/matrix/'csr/dense'_kernels.cpp` instead of Identity as this currently does not do anything with the Identity interface. Then you can delete all the identity test files as well.

Otherwise looks good to me.

Slaedr:
@pratikvn Since we are making the operation available via `Identity`'s advanced apply as well, I think we should have a test for that. That would then automatically also test the mixin interface.

pratikvn:
The reason I would prefer the tests in the kernel tests is because the kernels themselves are in the Dense and CSR classes. Identity is just using those kernels. So IMO, they should be tested where they are implemented, which here is the classes themselves. Additionally, then you can just call the actual kernels, `add_scaled_identity`, which I feel is more clear than just directly calling the apply in the Identity class.

Slaedr:
@pratikvn For now, rectangular matrices are being tested for the respective matrix types using `add_scaled_identity` only in reference, not the other backends. Square matrices are being tested for all backends under `identity`. I think I will leave the identity tests as they are, and add rectangular tests for all backends under the respective matrix types using `add_scaled_identity`. That should test both interfaces adequately.

pratikvn:
Okay, I guess that is fine.

pratikvn (approved):
LGTM!

core/matrix/identity.cpp
```diff
@@ -54,12 +54,21 @@ template <typename ValueType>
...
+        }
+    } else {
+        precision_dispatch_real_complex<ValueType>(
+            [this](auto dense_alpha, auto dense_b, auto dense_beta,
+                   auto dense_x) {
+                dense_x->scale(dense_beta);
+                dense_x->add_scaled(dense_alpha, dense_b);
+            },
+            alpha, b, beta, x);
+    }
```
upsj, line 71: From our internal poll, it seems like this interface option was not really popular. Can we remove it for now and maybe come back to it later?

hip/matrix/csr_kernels.hip.cpp
```diff
@@ -1291,6 +1291,27 @@ void extract_diagonal(std::shared_ptr<const HipExecutor> exec,
...
+{
+    const auto nrows = mtx->get_size()[0];
+    const auto nthreads = nrows * config::warp_size;
+    const auto nblocks = ceildiv(nthreads, default_block_size);
+    hipLaunchKernelGGL(kernel::add_scaled_identity, nblocks, default_block_size,
+                       0, 0, as_hip_type(alpha->get_const_values()),
+                       as_hip_type(beta->get_const_values()),
+                       static_cast<IndexType>(nrows), mtx->get_const_row_ptrs(),
+                       mtx->get_const_col_idxs(),
+                       as_hip_type(mtx->get_values()));
```
upsj, line 1308: If you don't mind being "clever", this (and anything dealing with updating diagonals) can actually be implemented with the common kernels (at the cost of random access to the row pointers)

```cpp
if (col < num_rows && idx >= row_ptrs[col] && idx < row_ptrs[col + 1]) {
    // we have a diagonal entry
} else {
    // we have a normal entry
}
```

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -767,6 +767,31 @@ class EnableAbsoluteComputation : public AbsoluteComputable {
...
+    void add_scaled_identity(const LinOp* const a, const LinOp* const b)
+    {
+        GKO_ASSERT_IS_SCALAR(a);
+        GKO_ASSERT_IS_SCALAR(b);
+        add_scaled_identity_impl(a, b);
+    }
+
+private:
+    virtual void add_scaled_identity_impl(const LinOp* a, const LinOp* b) = 0;
+};
```
upsj, line 792: Also in light of #796 I am a bit wary about adding functionality to both Dense and Csr behind a common interface. Can we instead implement the functions separately? Independent of that, you should probably deal with Executor mismatch if you provide this implementation (which necessitates a Mixin and makes the whole interface more difficult to use anyways)

core/matrix/csr.cpp
```diff
@@ -619,6 +623,16 @@ void Csr<ValueType, IndexType>::inv_scale_impl(const LinOp* alpha)
 }
 
 
+template <typename ValueType, typename IndexType>
+void Csr<ValueType, IndexType>::add_scaled_identity_impl(const LinOp* const a,
+                                                         const LinOp* const b)
+{
+    this->get_executor()->run(csr::make_add_diagonal_elems(this, false));
```
upsj, line 630: Not 100% happy about this one, as it may cause issues e.g. if we are using views to application data.

upsj (dismissed):
LGTM from the implementation standpoint, a few interface considerations:

core/matrix/identity.cpp
Slaedr: Out of the 3 votes, two were for including this. I think you wanted purely `add_scaled_identity` only, while @yhmtsai and @tcojean thought having both would be good.

core/matrix/csr.cpp
Slaedr: Yes, that's a good point. Maybe instead we should just check if all diagonals exist and throw an exception if a diagonal entry is missing?

upsj: That's what I had in mind, yes

include/ginkgo/core/base/lin_op.hpp
Slaedr: I'm treating Dense as a matrix here, yes. But this functionality is needed by Sundials in Dense (it's using gko Dense as a matrix type in addition to using it as a vector type). What would the advantage of having a separate function for Dense be? I'd rather enforce the same interface for the same operation on different formats. Ultimately we have to work with how the code currently is - Dense is both matrix and vector.

Slaedr: As for handling executors, that's a good point. I'll add that in the `add_scaled_identity` function.

upsj: For this to be possible, you need to query the executor of the object, i.e. you need to turn it into a mixin, so for the Identity interface to work, you need two classes "ScaledIdentityAddable" and "EnableScaledIdentityAddition", the former without template paramters for the feature check, the latter as a mixin.

Slaedr: Yeah, I tried a couple of things and that seems to be the best solution.

core/matrix/csr_kernels.hpp
```diff
@@ -216,6 +216,17 @@ namespace kernels {
                    const matrix::Dense<ValueType>* alpha,       \
                    matrix::Csr<ValueType, IndexType>* to_scale)
 
+#define GKO_DECLARE_CSR_CHECK_DIAGONAL_ENTRIES_EXIST(ValueType, IndexType) \
+    void check_diagonal_entries_exist(                                     \
+        std::shared_ptr<const DefaultExecutor> exec,                       \
+        const matrix::Csr<ValueType, IndexType>* mtx, bool* has_all_diags)
```
upsj, line 222: nit: using a reference makes it clearer that this is a host value, compared to device data that is usually passed via raw pointers.
```suggestion
        const matrix::Csr<ValueType, IndexType>* mtx, bool& has_all_diags)
```

hip/matrix/csr_kernels.hip.cpp
```diff
@@ -1291,6 +1291,51 @@ void extract_diagonal(std::shared_ptr<const HipExecutor> exec,
...
+template <typename ValueType, typename IndexType>
+void add_scaled_identity(std::shared_ptr<const HipExecutor> exec,
+                         const matrix::Dense<ValueType>* const alpha,
+                         const matrix::Dense<ValueType>* const beta,
+                         matrix::Csr<ValueType, IndexType>* const mtx)
+{
+    const auto nrows = mtx->get_size()[0];
+    const auto nthreads = nrows * config::warp_size;
+    const auto nblocks = ceildiv(nthreads, default_block_size);
+    hipLaunchKernelGGL(kernel::add_scaled_identity, nblocks, default_block_size,
```
upsj, line 1327: launching empty kernels in CUDA sets an error flag
```suggestion
if (nblocks > 0)
    hipLaunchKernelGGL(kernel::add_scaled_identity, nblocks, default_block_size,
```

cuda/matrix/csr_kernels.cu
```diff
@@ -1447,6 +1447,49 @@ void extract_diagonal(std::shared_ptr<const CudaExecutor> exec,
...
 
+template <typename ValueType, typename IndexType>
+void check_diagonal_entries_exist(
+    std::shared_ptr<const CudaExecutor> exec,
+    const matrix::Csr<ValueType, IndexType>* const mtx,
+    bool* const has_all_diags)
+{
+    const size_type num_blocks = exec->get_num_multiprocessor() * 4;
+    Array<int> block_has_diags(exec, num_blocks);
+    kernel::check_diagonal_entries<<<num_blocks, default_block_size>>>(
```
upsj, line 1458:
```suggestion
if (num_blocks > 0) 
    kernel::check_diagonal_entries<<<num_blocks, default_block_size>>>(
```
same for the other kernel launches with variabel number of blocks

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -767,6 +767,58 @@ class EnableAbsoluteComputation : public AbsoluteComputable {
...
+     * @param b  Scalar to multiply this before adding the scaled identity to
+     *   it.
+     */
+    void add_scaled_identity(const LinOp* const a,
+                             const LinOp* const b) override
+    {
+        GKO_ASSERT_IS_SCALAR(a);
+        GKO_ASSERT_IS_SCALAR(b);
+        auto ae =
+            make_temporary_clone(as<ConcreteType>(this)->get_executor(), a);
```
upsj, line 811: nit: usually we use a `self` function that `static_cast`s to ConcreteType*

upsj (commented):
Halfway done with my review, still need to check the GPU kernels and tests. Looks good, I like the error handling, even though we could probably use a function that acts without error handling.

core/matrix/csr_kernels.hpp
Slaedr: This is consistent with `is_sorted_by_column_index`, `calculate_total_cols` etc. though.

cuda/matrix/csr_kernels.cu
Slaedr: We never seem to have done this before. Was the assumption that in practice the number of blocks would not be zero because of the use of `ceildiv`? In this particular location, we would always have a nonzero number of SMs, so it should be fine.

include/ginkgo/core/base/lin_op.hpp
Slaedr: Perhaps it's best to just replace `as` with `static_cast` instead of adding two more identifiers in the interface?

cuda/matrix/csr_kernels.cu
upsj: ceildiv(0,...) is still 0, and empty kernel launches cause CUDA to return an error code from cudaGetLastError. I added guards like these to all kernel launches in #905, and should probably establish this elsewhere as well.

core/matrix/csr_kernels.hpp
upsj: passing a pointer for a single out value is a C-ism that we don't need to continue doing. If we use pointers for device data and references for host data, the interface becomes more consistent.

include/ginkgo/core/base/lin_op.hpp
upsj: True, if we only use it in a single place, we don't need to add another function.

upsj (commented):
LGTM from the implementation and test standpoint.
I have a few suggestions for simplification: 
* The GPU kernels are a bit overengineered (still a nice job with the manual reduction :) ), they could get away with setting a single flag on failure. This idea might even allow us to eliminate the check kernel entirely, if we are okay leaving the matrix in a partially modified state in the error case?
* The test helpers are simpler if they operate on matrix_data
* Do you need to use add_scaled_identity in a polymorphic context? Or is it sufficient if the functions have the same interface between Dense and Csr? I.e. do you ever have somthing like
```
LinOp* a;
as<ScaledIdentityAddable>(a)->add_scaled_identity(...)
```
or can you store `a` with its concrete type?
Adding a feature class for this is a one-way street, we can't remove it later without breaking interface, but we can always add it in case it turns out we need it.

common/cuda_hip/matrix/csr_kernels.hpp.inc
```diff
@@ -842,4 +842,125 @@ __global__
...
+        // TODO: round up warps_per_block to a power of 2 below
+        // (by default it already is)
+        for (uint32 k = static_cast<uint32>(warps_per_block / 2); k >= 1;
+             k >>= 1) {
+            val = static_cast<int>(val && warp_tile.shfl_down(val, k));
+        }
+        if (warp_tile.thread_rank() == 0) {
+            block_reduced[blockIdx.x] = val;
+        }
+    }
```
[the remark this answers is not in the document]
Slaedr: I don't see an `is_sorted_by_row`. Which one are you referring to? The algorithm for `is_sorted_by_col` does not have coalesced access; this might actually be better. Eventually, I think we will move to provide the user with an option of disabling this check, as Pratik mentioned somewhere above. That could be done either through user-settable matrix properties (have a property called `has_all_diagonal_entries`), or using the checking infrastructure that Gregor had started working on.

Slaedr:
@upsj Right now, the application wants to use either `Csr` or `Dense` for the same matrix, depending on the further downstream application. Eventually, perhaps also `Ell`. I think we should make it easier for the application to use different matrix formats depending on some runtime information by adding the feature class.

common/cuda_hip/matrix/csr_kernels.hpp.inc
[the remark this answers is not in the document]
upsj: ah yeah, that's the one I mean. is_sorted_by_col_idxs provides fast-fail semantics due to the fact that you can short-circuit the && operator. Also as the result is either `true` or `false`, and if it is `false` in one thread, that already determines the global result, you can simplify that a lot like `if (warp.any(!has_diagonal)) { if (lane == 0) { flag = false; } return; }`

common/cuda_hip/matrix/csr_kernels.hpp.inc
```diff
@@ -842,4 +842,64 @@ __global__
...
+        const IndexType row_start = row_ptrs[row];
+        const IndexType num_nz = row_ptrs[row + 1] - row_start;
+        int row_has_diag_local = 0;
+        for (IndexType iz = tid_in_warp; iz < num_nz; iz += warp_size) {
+            if (col_idxs[iz + row_start] == row) {
+                row_has_diag_local = 1;
+                break;
+            }
+        }
+        int row_has_diag = tile_grp.any(row_has_diag_local == 1);
```
upsj, line 894: minor type suggestion: predicate can be bool, the result of `any` is config::lane_mask_type
```suggestion
        bool row_has_diag_local = false;
        for (IndexType iz = tid_in_warp; iz < num_nz; iz += warp_size) {
            if (col_idxs[iz + row_start] == row) {
                row_has_diag_local = true;
                break;
            }
        }
        auto row_has_diag = static_cast<bool>(tile_grp.any(row_has_diag_local));
```

core/matrix/csr.cpp
```diff
@@ -672,6 +675,23 @@ void Csr<ValueType, IndexType>::inv_scale_impl(const LinOp* alpha)
...
+template <typename ValueType, typename IndexType>
+void Csr<ValueType, IndexType>::add_scaled_identity_impl(const LinOp* const a,
+                                                         const LinOp* const b)
+{
+    bool has_diags{false};
+    this->get_executor()->run(
+        csr::make_check_diagonal_entries(this, has_diags));
+    if (!has_diags) {
+        GKO_UNSUPPORTED_MATRIX_PROPERTY(
+            "The matrix has one or more structurally zero diagonals!");
```
upsj, line 687:
```suggestion
            "The matrix has one or more structurally zero diagonal entries!");
```

core/matrix/csr.cpp
```diff
@@ -672,6 +675,23 @@ void Csr<ValueType, IndexType>::inv_scale_impl(const LinOp* alpha)
 }
 
 
+template <typename ValueType, typename IndexType>
+void Csr<ValueType, IndexType>::add_scaled_identity_impl(const LinOp* const a,
+                                                         const LinOp* const b)
+{
+    bool has_diags{false};
```
pratikvn, line 682: Can this be made a parameter to the function and defaulted to `false` ?

core/test/utils/matrix_utils_test.cpp
```diff
@@ -198,4 +201,77 @@ TYPED_TEST(MatrixUtils, MakeHpdMatrixWithRatioCorrectly)
...
+        }
+    }
+    ASSERT_TRUE(all_diags);
+    bool has_diag = false;
+    for (int j = rowptrs[row_to_remove]; j < rowptrs[row_to_remove + 1]; j++) {
+        if (colidxs[j] == row_to_remove) {
+            has_diag = true;
+        }
+    }
+    ASSERT_FALSE(has_diag);
```
upsj, line 242: nit: this can be simplified a bit
```suggestion
    for (int i = 0; i < 3; i++) {
        bool has_diag = false;
        for (int j = rowptrs[i]; j < rowptrs[i + 1]; j++) {
            if (colidxs[j] == i) {
                has_diag = true;
            }
        }
        ASSERT_EQ(has_diag, i == row_to_remove);
    }
```
if you like std:: algorithms, you can also replace the inner loop by
```cpp
col_begin = colidxs + rowptrs[i];
col_end = colidxs + rowptrs[i + 1]
bool has_diag = std::find(col_begin, col_end, i) != col_end;
```

core/test/utils/matrix_utils_test.cpp
```diff
@@ -198,4 +201,77 @@ TYPED_TEST(MatrixUtils, MakeHpdMatrixWithRatioCorrectly)
...
+            if (colidxs[j] == i) {
+                has_diag = true;
+            }
+        }
+        if (!has_diag) {
+            all_diags = false;
+            break;
+        }
+    }
+    ASSERT_TRUE(all_diags);
```
upsj, line 273:
```suggestion
        ASSERT_TRUE(has_diag);
    }    
```

cuda/matrix/csr_kernels.cu
upsj: This still needs to be addressed

cuda/matrix/csr_kernels.cu
```diff
@@ -1262,6 +1262,54 @@ void extract_diagonal(std::shared_ptr<const CudaExecutor> exec,
...
+    std::shared_ptr<const CudaExecutor> exec,
+    const matrix::Csr<ValueType, IndexType>* const mtx, bool& has_all_diags)
+{
+    const size_type num_warps = mtx->get_size()[0];
+    const size_type num_blocks =
+        num_warps / (default_block_size / config::warp_size);
+    Array<bool> h_has_diags(exec->get_master(), 1);
+    h_has_diags.get_data()[0] = true;
+    Array<bool> has_diags(exec, 1);
+    has_diags = h_has_diags;
```
upsj, line 1276:
```suggestion
    Array<bool> has_diags(exec, {true});
```

cuda/matrix/csr_kernels.cu
```diff
@@ -1262,6 +1262,54 @@ void extract_diagonal(std::shared_ptr<const CudaExecutor> exec,
...
+    kernel::check_diagonal_entries<<<num_blocks, default_block_size>>>(
+        static_cast<IndexType>(
+            std::min(mtx->get_size()[0], mtx->get_size()[1])),
+        mtx->get_const_row_ptrs(), mtx->get_const_col_idxs(),
+        has_diags.get_data());
+    h_has_diags = has_diags;
+    has_all_diags = h_has_diags.get_data()[0];
```
upsj, line 1283:
```suggestion
    has_all_diags = exec->copy_val_to_host(has_diags.get_const_data());
```

omp/matrix/csr_kernels.cpp
```diff
@@ -966,6 +966,58 @@ void extract_diagonal(std::shared_ptr<const OmpExecutor> exec,
...
+template <typename ValueType, typename IndexType>
+void check_diagonal_entries_exist(
+    std::shared_ptr<const OmpExecutor> exec,
+    const matrix::Csr<ValueType, IndexType>* const mtx, bool& has_all_diags)
+{
+    bool l_has_all_diags = true;
+    const size_type minsize = std::min(mtx->get_size()[0], mtx->get_size()[1]);
+    const auto row_ptrs = mtx->get_const_row_ptrs();
+    const auto col_idxs = mtx->get_const_col_idxs();
+#pragma omp parallel for reduction(&& : l_has_all_diags)
```
pratikvn, line 978: Not sure if this is faster, or just a sequential with early exit is faster.

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -779,6 +779,58 @@ class EnableAbsoluteComputation : public AbsoluteComputable {
...
+class EnableScaledIdentityAddition : public ScaledIdentityAddable {
+public:
+    /**
+     * Scales this and adds another scalar times the identity to it.
+     *
+     * @param a  Scalar to multiply the identity operator by before adding.
+     * @param b  Scalar to multiply this before adding the scaled identity to
+     *   it.
+     */
+    void add_scaled_identity(const LinOp* const a,
```
upsj, line 817: Not a blocker or anything, just a comment: I can't exactly say why, but this parameter order seems the wrong way round for me.

pratikvn (approved):
LGTM!

upsj (approved):
LGTM - Nice job! Two things on the feature class interface
* The current approach doesn't give a canonical place to add documentation to add_scaled_identity, especially noting the preconditions of the function for Csr.
* Removing a feature class is hard, adding it later is simple, in case it truly becomes necessary in a polymorphic context.

core/matrix/csr.cpp
```diff
@@ -672,6 +675,23 @@ void Csr<ValueType, IndexType>::inv_scale_impl(const LinOp* alpha)
...
+    this->get_executor()->run(
+        csr::make_check_diagonal_entries(this, has_diags));
```
yhmtsai, line 684: should it always check?

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -779,6 +779,58 @@ class EnableAbsoluteComputation : public AbsoluteComputable {
...
+                             const LinOp* const b) override
+    {
+        GKO_ASSERT_IS_SCALAR(a);
+        GKO_ASSERT_IS_SCALAR(b);
+        auto ae = make_temporary_clone(
+            static_cast<ConcreteType*>(this)->get_executor(), a);
+        auto be = make_temporary_clone(
+            static_cast<ConcreteType*>(this)->get_executor(), b);
```
yhmtsai, line 825: does `static_cast<LinOp*>` work for it? If work, you can use normal feature class not mixin

yhmtsai (changes requested):
std::complex<float> and std::complex<double> can not operate together.
Thus, the dense part will be broken

core/matrix/csr.cpp
Slaedr: Right now, we don't have a mechanism for the user to tell us whether it's needed. If we add a property, I'm not sure where it should go right now. So for now, I'm thinking of always carrying out the check, until we have a good general way to handle matrix properties or pre-condition checks. Or until someone says it's too slow with the check.

include/ginkgo/core/base/lin_op.hpp
Slaedr: That's a good idea! It appears to work. We might prefer `as<LinOp>` to make it give a better error message in case the user does something strange. I don't think that would have a significant performance impact.

reference/matrix/dense_kernels.cpp
```diff
@@ -1035,6 +1035,27 @@ void get_imag(std::shared_ptr<const ReferenceExecutor> exec,
...
+                         const matrix::Dense<ScalarType>* const alpha,
+                         const matrix::Dense<ScalarType>* const beta,
+                         matrix::Dense<ValueType>* const mtx)
+{
+    const auto dim = mtx->get_size();
+    for (size_type row = 0; row < dim[0]; row++) {
+        for (size_type col = 0; col < dim[1]; col++) {
+            mtx->at(row, col) = beta->get_const_values()[0] * mtx->at(row, col);
+            if (row == col) {
+                mtx->at(row, row) += alpha->get_const_values()[0];
```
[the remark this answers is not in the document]
Slaedr: Do you mean when scalars are complex but matrix values are real? I think that should be handled by `precision_conversion_real_complex`?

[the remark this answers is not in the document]
yhmtsai: no, they are `complex<float>` and `complex<double>`
sorry, it is disappear when I do not use ` to wrap it

core/matrix/csr.cpp
pratikvn: Okay, I guess we can wait for matrix properties to have a fast path for this.

reference/matrix/dense_kernels.cpp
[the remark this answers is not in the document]
Slaedr: Good catch. I added `make_temporary_conversion`; that should take care of it.

reference/test/matrix/dense_kernels.cpp
```diff
@@ -4251,4 +4266,22 @@ TYPED_TEST(DenseComplex, ConjDot)
 }
 
 
+TEST(DenseComplex, ScaleAddIdentityRectangularMixed)
```
yhmtsai, line 4269: maybe not neccessary because Scalar and ValueType will be the same

core/matrix/dense.cpp
```diff
@@ -1427,6 +1428,21 @@ void Dense<ValueType>::get_imag(
 }
 
 
+template <typename ValueType>
+void Dense<ValueType>::add_scaled_identity_impl(const LinOp* const a,
+                                                const LinOp* const b)
+{
+    precision_dispatch_real_complex<ValueType>(
```
yhmtsai, line 1435: it already did the conversion, I think

yhmtsai (approved):
LGTM. only one nit: the Dense::add_scaled_identity only support ValueType == ScalarType, so maybe do not need additional template parameter. Also the preccision_dispatch_real_complex already take care of dense_alpha, dense_beta, I think

reference/test/matrix/dense_kernels.cpp
Slaedr: What do you mean? In the test, the scalar type is `complex<double>` but the matrix value type is `complex<float>`. I guess this test is to make sure that if the user does pass a different scalar type, it compiles and works.

I agree that the extra template parameter on the kernels is useless right now, but having the extra generality would not hurt, I guess; at some point, we might need to enable ScalarType != ValueType.

yhmtsai: I read the `GKO_INSTANTIATE_FOR_EACH_VALUE_AND_SCALAR_TYPE` again and realize my assumption is wrong.
`GKO_INSTANTIATE_FOR_EACH_VALUE_AND_SCALAR_TYPE` only generates additional type pair for real scalar but complex data. not for all possible precision combination.
By the way, are precision_dispatch_real_complex a and b always the same type?

core/matrix/dense.cpp
Slaedr: Right, the new test still works if I remove the make_temporary_conversion. I guess it would only be needed if I added the possibility for scalar type to be something other than value type. I think I'll delete the last commit.

PR #943, opened 2021-12-21
Compatibility to SYCL 2020

upsj:
This PR uses sycl::reqd_sub_group_size in all kernels that use cooperative groups for communication. I found them by searching for tiled_partition and reduce(...)


dpcpp/matrix/dense_kernels.dp.cpp
```diff
@@ -205,7 +205,7 @@ void calculate_slice_lengths(size_type num_rows, size_type slice_size,
                              size_type* __restrict__ slice_sets,
                              sycl::nd_item<3> item_ct1)
 {
-    constexpr auto sg_size = cfg;
+    constexpr auto sg_size = KCFG_1D::decode<1>(cfg);
```
upsj, line 208: This looks like it was a bug before?

yhmtsai: No, the cfg is subgroup list only. it fixes the workgroup size

dpcpp/test/components/cooperative_groups.dp.cpp
```diff
@@ -117,10 +117,10 @@ class CooperativeGroups : public testing::TestWithParam<unsigned int> {
 
 // kernel implementation
 template <std::uint32_t config>
-__WG_BOUND__(KCfg::decode<0>(config))
+__WG_BOUND__(KCFG_1D::decode<0>(config))
```
yhmtsai, line 120: this will not have effect

dpcpp/components/cooperative_groups.dp.hpp
```diff
@@ -353,8 +353,7 @@ template <unsigned Size, typename Group>
 __dpct_inline__
     std::enable_if_t<(Size > 1) && Size <= 64 && !(Size & (Size - 1)),
                      detail::thread_block_tile<Size>>
-        tiled_partition
-    [[intel::reqd_sub_group_size(Size)]] (const Group& group)
```
yhmtsai, line 357: we should keep this.
 we can use old compiler with sycl1.2.1 to check whether the subgroup size is changed or not.
 It's helpful when porting the block Jacobi and threshold selection. Otherwise, SYCL will just ignore the setting without any compile error

dpcpp/base/helper.hpp
```diff
@@ -84,10 +86,13 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
     {                                                                 \
         queue->submit([&](sycl::handler& cgh) {                       \
-            cgh.parallel_for(sycl_nd_range(grid, block),              \
-                             [=](sycl::nd_item<3> item_ct1) {         \
-                                 kernel_<encoded>(args..., item_ct1); \
-                             });                                      \
+            cgh.parallel_for(                                         \
+                sycl_nd_range(grid, block), [=                        \
+            ](sycl::nd_item<3> item_ct1) [[sycl::reqd_sub_group_size( \
+                                                KCFG_1D::decode<1>(   \
```
yhmtsai, line 92: this is not always KCFG_1D.
and it will give error when the subgroup size is equal to 1, I think

dpcpp/base/helper.hpp
```diff
@@ -62,10 +62,12 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
     {                                                                     \
         queue->submit([&](sycl::handler& cgh) {                           \
-            cgh.parallel_for(sycl_nd_range(grid, block),                  \
-                             [=](sycl::nd_item<3> item_ct1) {             \
-                                 kernel_(args..., item_ct1);              \
-                             });                                          \
+            cgh.parallel_for(                                             \
+                sycl_nd_range(grid, block), [=                            \
+            ](sycl::nd_item<3> item_ct1) [[sycl::reqd_sub_group_size(     \
+                                                config::warp_size)]] {    \
```
yhmtsai, line 68: should we leave as the default one not config::warp_size?

dpcpp/components/cooperative_groups.dp.hpp
yhmtsai: Because dpcpp 2022 also remove the propagation even if it is on old standard, I withdraw the comment here

dpcpp/base/helper.hpp
yhmtsai: Could you add a int map for it? only specialize the 1 to 16(the default of gpu). others just return themselves

dpcpp/test/components/cooperative_groups.dp.cpp
yhmtsai: I will move it after this pr. It should not give any error

dpcpp/solver/cb_gmres_kernels.dp.cpp
```diff
@@ -453,7 +457,9 @@ void multinorm2_inf_kernel(
             reduction_helper_array_acc_ct1(cgh);
 
         cgh.parallel_for(
-            sycl_nd_range(grid, block), [=](sycl::nd_item<3> item_ct1) {
+            sycl_nd_range(grid, block), [=
+        ](sycl::nd_item<3> item_ct1) [[sycl::reqd_sub_group_size(
+                                            default_dot_dim)]] {
```
tcojean, line 462: Shouldn't we use `config::warp_size` here? It's the same size for now. Are there performance issues for it to be exactly 16?
```suggestion
                                            config::warp_size)]] {
```

dpcpp/solver/cb_gmres_kernels.dp.cpp
```diff
@@ -352,7 +354,9 @@ void multinorminf_without_stop_kernel(
...
```
tcojean, line 359:
```suggestion
                                            config::warp_size)]] {
```

dpcpp/solver/gmres_kernels.dp.cpp
```diff
@@ -204,7 +204,9 @@ void multidot_kernel(dim3 grid, dim3 block, size_type dynamic_shared_memory,
             reduction_helper_array_acc_ct1(cgh);
 
         cgh.parallel_for(
-            sycl_nd_range(grid, block), [=](sycl::nd_item<3> item_ct1) {
+            sycl_nd_range(grid, block), [=
+        ](sycl::nd_item<3> item_ct1) [[sycl::reqd_sub_group_size(
+                                            default_dot_dim)]] {
```
tcojean, line 209:
```suggestion
                                            config::warp_size)]] {
```

dpcpp/solver/idr_kernels.dp.cpp
```diff
@@ -384,7 +386,9 @@ void multidot_kernel(dim3 grid, dim3 block, size_t dynamic_shared_memory,
             reduction_helper_array_acc_ct1(cgh);
 
         cgh.parallel_for(
-            sycl_nd_range(grid, block), [=](sycl::nd_item<3> item_ct1) {
+            sycl_nd_range(grid, block), [=
+        ](sycl::nd_item<3> item_ct1) [[sycl::reqd_sub_group_size(
+                                            default_dot_dim)]] {
```
tcojean, line 391:
```suggestion
                                            default_dot_dim)]] {
```
```suggestion
                                            config::warp_size)]] {
```

dpcpp/base/helper.hpp
```diff
@@ -119,9 +124,9 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
     }
 
 // __WG_BOUND__ gives the cuda-like launch bound in cuda ordering
-#define __WG_BOUND_1D__(x) [[intel::reqd_work_group_size(1, 1, x)]]
-#define __WG_BOUND_2D__(x, y) [[intel::reqd_work_group_size(1, y, x)]]
-#define __WG_BOUND_3D__(x, y, z) [[intel::reqd_work_group_size(z, y, x)]]
+#define __WG_BOUND_1D__(x) [[sycl::reqd_work_group_size(1, 1, x)]]
+#define __WG_BOUND_2D__(x, y) [[sycl::reqd_work_group_size(1, y, x)]]
```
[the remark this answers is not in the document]
tcojean: These also need to be used only in the first lambda to have any effect, right?

tcojean (commented):
Some comments. There were also good comments by Mike to be addressed.

dpcpp/base/helper.hpp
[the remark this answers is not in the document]
yhmtsai: I think it need 1, 1 if from cuda: dim0 -> z dim1 -> y dim2 -> x
Yes, it needs to be in the lambda of parallel for

dpcpp/matrix/dense_kernels.dp.cpp
upsj: The corresponding code was removed, so this is no longer relevant

dpcpp/solver/cb_gmres_kernels.dp.cpp
upsj: The correctness of these kernels internally depends on the value of default_dot_dim, so I don't want to touch this right now. But in general, I was planning to move all the reductions in #861 anyways, so this code may not be that relevant in the long run?

yhmtsai (approved):
LGTM. only one minor question, should we set the subgroup size when we do not use it (like `GKO_ENABLE_DEFAULT_HOST`)? I think it is good for porting in the beginning because we can always use config::warp_size as default setting without concerning vary subgroup in dpcpp.
possible downside: we potentially set the parallelization to SIMD?

dpcpp/base/helper.hpp
upsj: I would leave it as warp_size, since that's als what we are using inside a few of these kernels (mostly SpMV)

dpcpp/test/components/cooperative_groups.dp.cpp
```diff
@@ -184,39 +185,65 @@ void cg_all(bool* s, sycl::nd_item<3> item_ct1)
...
+template <std::uint32_t encoded, typename... InferredArgs>
+inline void cg_all(dim3 grid, dim3 block, gko::size_type, sycl::queue* queue,
+                   InferredArgs... args)
+{
+    queue->submit([&](sycl::handler& cgh) {
+        cgh.parallel_for(
+            sycl_nd_range(grid, block),
+            [=](sycl::nd_item<3> item_ct1)
+                [[sycl::reqd_sub_group_size(KCFG_1D::decode<1>(
+                    encoded))]] __WG_BOUND__(KCFG_1D::decode<0>(config)) {
```
tcojean, line 197:
```suggestion
                    encoded))]] __WG_BOUND__(KCFG_1D::decode<0>(encoded)) {
```

dpcpp/test/components/cooperative_groups.dp.cpp
```diff
@@ -184,39 +185,65 @@ void cg_all(bool* s, sycl::nd_item<3> item_ct1)
...
+template <std::uint32_t encoded, typename... InferredArgs>
+inline void cg_any(dim3 grid, dim3 block, gko::size_type, sycl::queue* queue,
+                   InferredArgs... args)
+{
+    queue->submit([&](sycl::handler& cgh) {
+        cgh.parallel_for(
+            sycl_nd_range(grid, block),
+            [=](sycl::nd_item<3> item_ct1)
+                [[sycl::reqd_sub_group_size(KCFG_1D::decode<1>(
+                    encoded))]] __WG_BOUND__(KCFG_1D::decode<0>(config)) {
```
tcojean, line 231:
```suggestion
                    encoded))]] __WG_BOUND__(KCFG_1D::decode<0>(encoded)) {
```

dpcpp/test/components/cooperative_groups.dp.cpp
```diff
@@ -227,7 +254,21 @@ void cg_ballot(bool* s, sycl::nd_item<3> item_ct1)
...
+template <std::uint32_t encoded, typename... InferredArgs>
+inline void cg_ballot(dim3 grid, dim3 block, gko::size_type, sycl::queue* queue,
+                      InferredArgs... args)
+{
+    queue->submit([&](sycl::handler& cgh) {
+        cgh.parallel_for(
+            sycl_nd_range(grid, block),
+            [=](sycl::nd_item<3> item_ct1)
+                [[sycl::reqd_sub_group_size(KCFG_1D::decode<1>(
+                    encoded))]] __WG_BOUND__(KCFG_1D::decode<0>(config)) {
```
tcojean, line 266:
```suggestion
                    encoded))]] __WG_BOUND__(KCFG_1D::decode<0>(encoded)) {
```

tcojean (approved):
LGTM!

dpcpp/test/components/cooperative_groups.dp.cpp
upsj: good catch, thanks!

PR #951, opened 2022-01-05
MPI updates and fixes

pratikvn:
This PR fixes some GTest issues, and adds more comprehensive tests for bindings.

+ Move gtest compilation to the `core/test/mpi` unit.
+ Add non-blocking equivalents for the frequently used MPI calls.
+ Some doc fixes. 
+ Update tests and add tests for all bindings.

TODO:
+ [x] Update CI to be more comprehensive.


pratikvn:
format!

core/test/mpi/base/bindings.cpp
```diff
@@ -170,23 +211,26 @@ TYPED_TEST(MpiBindings, CanPutValuesWithExclusiveLock)
...
-                win.put(data.data(), 4, rank, 0, 4);
-                win.flush(0);
-                win.unlock(rank);
+
+    {
+        auto win = window(data.data(), 4, comm);
+        if (my_rank == 0) {
+            for (auto rank = 0; rank < num_ranks; ++rank) {
+                if (rank != my_rank) {
+                    win.lock(rank, 0, window::lock_type::exclusive);
```
Slaedr, line 226: If the most used value of `assert` is `0`, perhaps that could be the last argument to `lock` so that we normally don't have to bother with it while creating an exclusive lock.

core/test/mpi/base/bindings.cpp
```diff
@@ -146,17 +152,52 @@ TYPED_TEST(MpiBindings, CanPutValuesWithLockAll)
...
+    {
+        auto win = window(data.data(), 4, comm);
+        if (my_rank == 0) {
+            win.lock_all();
+            for (auto rank = 0; rank < num_ranks; ++rank) {
+                if (rank != my_rank) {
+                    win.put(data.data(), 4, rank, 0, 4);
+                }
+            }
+            win.unlock_all();
```
Slaedr, line 165: You should probably also add tests that actually test the lock. Eg. before this statement, try to access the window on another rank (perhaps in both cases: with and without a separate lock) and test that the expected thing happens. I need to refresh myself on how exactly that's supposed to behave, but it would be nice to have such tests.

Slaedr (approved):
LGTM. Couple of suggestions below, mainly about RMA tests and the function signature for lock functions.

MarcelKoch (commented):
LGTM, only two minor nits.

.gitlab-ci.yml
```diff
@@ -552,6 +555,8 @@ build/amd/nompi/clang/hip/release/static:
     CXX_COMPILER: "clang++"
     BUILD_OMP: "ON"
     BUILD_HIP: "ON"
+    MPI_AS_ROOT: "ON"
+    BUILD_MPI: "ON"
```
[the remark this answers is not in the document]
pratikvn: Yes, OpenMPI developers strictly warn against running mpi as a root user. It seems that if there is a bug or something, then mpirun as root might leave the system itself in a unusable state. But if someone wants to do that nonetheless, then they provide a couple of ways to do so.

For us, it is not an issue because we mainly run inside containers, which is fully isolated, so any corruptions/crashes in the container should not affect the outer system.

core/test/mpi/base/bindings.cpp
pratikvn: Yes, I agree that it would be great to have a nice way to test the locks separately. But the problem is that it is difficult to find a consistent and correct way of doing so. The failure is also usually a seg-fault, which makes it difficult to capture it.

MarcelKoch (approved):

core/test/mpi/base/bindings.cpp
yhmtsai: maybe you can try ASSERT_EXIT like cuda/test/base/cuda_executor_reset if the segmentation fault always return 11 error code. It create another thread to do the test, so I am not sure whether it is suitable for MPI test.

.gitlab-ci.yml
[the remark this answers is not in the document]
Slaedr: But you could just run as a regular user, rather than as root, right?

core/test/mpi/base/bindings.cpp
pratikvn: Yes, but that is only when an access to the data within the epoch fails. It does not always result in a seg-fault, so you cannot always rely on ASSERT_EXIT.

.gitlab-ci.yml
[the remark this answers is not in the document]
pratikvn: No, docker always runs as root.

pratikvn:
Sorry, I ended adding a few more wrapper functions I needed, namely `MPI_'Get'Accumulate` and `MPI_Fetch_and_op` operations. Maybe @Slaedr and @MarcelKoch can have another look if everything is right with the CI.

MarcelKoch (approved):

include/ginkgo/core/base/mpi.hpp
```diff
@@ -1166,7 +1320,54 @@ class window {
...
+     *
+     * @param origin_buffer  the buffer to send
+     * @param origin_count  the number of elements to put
+     * @param target_rank  the rank to put the data to
+     * @param target_disp  the displacement at the target window
+     * @param target_count  the request handle for the send call
+     * @param operation  the reduce operation. See @MPI_Op
+     */
+    template <typename PutType>
+    void accumulate(const PutType* origin_buffer, const int origin_count,
```
Slaedr, line 1338: I guess this has no test?

include/ginkgo/core/base/mpi.hpp
```diff
@@ -1166,7 +1320,54 @@ class window {
...
+     * @param origin_count  the number of elements to put
+     * @param target_rank  the rank to put the data to
+     * @param target_disp  the displacement at the target window
+     * @param target_count  the request handle for the send call
+     * @param operation  the reduce operation. See @MPI_Op
+     *
+     * @return  the request handle for the send call
+     */
+    template <typename PutType>
+    request r_accumulate(const PutType* origin_buffer, const int origin_count,
```
Slaedr, line 1361: Test for this one?

core/test/mpi/base/bindings.cpp
```diff
@@ -290,174 +408,714 @@ TYPED_TEST(MpiBindings, CanGetValuesWithFence)
...
-        ASSERT_EQ(comp_data[1], TypeParam{3});
 
-    } else if (my_rank == 1) {
-        ASSERT_EQ(comp_data[0], TypeParam{1});
-        ASSERT_EQ(comp_data[1], TypeParam{3});
+    {
+        auto win = window(target.data(), 4, comm);
+
+        if (my_rank == 2) {
+            win.lock_all();
```
Slaedr, line 552: Do we need locks here? Can we do this test without locks? It might be better in that case - to make this a proper *unit* test.

Slaedr (approved):
Mostly LGTM.

include/ginkgo/core/base/mpi.hpp
pratikvn: It should be there. See [line 302](https://github.com/ginkgo-project/ginkgo/blob/mpi-fixes/core/test/mpi/base/bindings.cpp#L302)

include/ginkgo/core/base/mpi.hpp
pratikvn: It should be there. See [Line 349](https://github.com/ginkgo-project/ginkgo/blob/mpi-fixes/core/test/mpi/base/bindings.cpp#L349)

core/test/mpi/base/bindings.cpp
pratikvn: Yes, as it is a one-sided communication, you need either a passive target sync or an active synchronization (fence type) to actually allow the communication.

include/ginkgo/core/base/mpi.hpp
Slaedr: Oh yeah, somehow I missed them.

PR #953, opened 2022-01-12
Exclude CUDA 11.3.0 from the triangular solver

thoasm:
Only CUDA 11.3.1 starts to support `cusparseSpSM` (see the [CUDA documentation for 11.3.1](https://docs.nvidia.com/cuda/archive/11.3.1/cusparse/index.html#cusparse-generic-function-spsm)), version 11.3.0 does not support it yet (`cusparseSpSM` is not present in [CUDA documentation for 11.3.0](https://docs.nvidia.com/cuda/archive/11.3.0/cusparse/index.html)).

### DIFF
--- a/cuda/base/cusparse_bindings.hpp
+++ b/cuda/base/cusparse_bindings.hpp
@@ -879,7 +879,7 @@ inline void destroy(cusparseSpMatDescr_t descr)
 }
 
 
-#if (CUDA_VERSION >= 11030)
+#if (CUDA_VERSION >= 11031)
 
 
 template <typename AttribType>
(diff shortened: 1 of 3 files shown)


yhmtsai (approved):

pratikvn (approved):

cuda/base/cusparse_bindings.hpp
```diff
@@ -879,7 +879,7 @@ inline void destroy(cusparseSpMatDescr_t descr)
 }
 
 
-#if (CUDA_VERSION >= 11030)
+#if (CUDA_VERSION > 11030)
```
upsj, line 882: nit: List the version it was introduced in, not the last one that didn't have it
```suggestion
#if (CUDA_VERSION >= 11031)
```

upsj (approved):
LGTM!

cuda/base/cusparse_bindings.hpp
thoasm: Wasn't sure in which way I should do it, thanks! Now it is the proper way.

PR #956, opened 2022-01-26
Remove unnecessary device_guard usage

upsj:
pulled out of #938, we don't need to set the device ID kernels, because it already gets set by the executor->run function.

### DIFF
--- a/cuda/base/device_guard.hpp
+++ b/cuda/base/device_guard.hpp
@@ -57,11 +57,12 @@ namespace cuda {
  */
 class device_guard {
 public:
-    device_guard(int device_id)
+    device_guard(int device_id) : original_device_id{}, need_reset{}
     {
         GKO_ASSERT_NO_CUDA_ERRORS(cudaGetDevice(&original_device_id));
         if (original_device_id != device_id) {
             GKO_ASSERT_NO_CUDA_ERRORS(cudaSetDevice(device_id));
+            need_reset = true;
         }
     }
 
@@ -75,16 +76,19 @@ public:
 
     ~device_guard() noexcept(false)
     {
-        /* Ignore the error during stack unwinding for this call */
-        if (std::uncaught_exception()) {
-            cudaSetDevice(original_device_id);
-        } else {
-            GKO_ASSERT_NO_CUDA_ERRORS(cudaSetDevice(original_device_id));
+        if (need_reset) {
+            /* Ignore the error during stack unwinding for this call */
+            if (std::uncaught_exception()) {
+                cudaSetDevice(original_device_id);
+            } else {
+                GKO_ASSERT_NO_CUDA_ERRORS(cudaSetDevice(original_device_id));
+            }
         }
     }
 
 private:
-    int original_device_id{};
+    int original_device_id;
+    bool need_reset;
 };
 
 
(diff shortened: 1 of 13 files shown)


yhmtsai (approved):

cuda/factorization/ilu_kernels.cu
```diff
@@ -57,7 +56,6 @@ void compute_lu(std::shared_ptr<const DefaultExecutor> exec,
 {
     const auto id = exec->get_device_id();
     auto handle = exec->get_cusparse_handle();
-    gko::cuda::device_guard g{id};
```
Slaedr, line 60: Do you remember why this was only added for ic and ilu, but not other preconditioners?

Slaedr (approved):
LGTM! Thanks for the quick PR. However, why do we need to set the device at every kernel run? Can't we just set it once when the executor is created? I think we don't currently set it in executor constructors, though the device ID is already available at that point.

cuda/factorization/ilu_kernels.cu
upsj: no, I must have not remembered that Executor already sets them.

upsj:
That's the issue with global state - we have no way of knowing whether it was modified somewhere, so in the worst case users change the device ID inside a custom LinOp::apply and we run solver kernels on GPU 1 while the data resides on GPU 0, which gives a hard to identify access violation.

Slaedr:
I see. Okay, in general, that makes sense.

cuda/base/device_guard.hpp
```diff
@@ -60,7 +60,9 @@ class device_guard {
     device_guard(int device_id)
     {
         GKO_ASSERT_NO_CUDA_ERRORS(cudaGetDevice(&original_device_id));
-        GKO_ASSERT_NO_CUDA_ERRORS(cudaSetDevice(device_id));
+        if (original_device_id != device_id) {
```
pratikvn, line 63: :+1:

pratikvn (approved):
LGTM!

cuda/base/device_guard.hpp
Slaedr, line 63: Nice.

Slaedr (approved):
LGTM. That's even better!

tcojean (approved):

cuda/base/device_guard.hpp
```diff
@@ -57,10 +57,13 @@ namespace cuda {
  */
 class device_guard {
 public:
-    device_guard(int device_id)
+    device_guard(int device_id) : original_device_id{}, need_reset{}
```
pratikvn, line 60: For readability, might be better to be explicit and initialize it to `false` ? 
```suggestion
    device_guard(int device_id) : original_device_id{}, need_reset{false}
```

hip/base/device_guard.hip.hpp
```diff
@@ -57,10 +57,13 @@ namespace hip {
  */
 class device_guard {
 public:
-    device_guard(int device_id)
+    device_guard(int device_id) : original_device_id{}, need_reset{}
```
upsj, line 60:
```suggestion
    device_guard(int device_id) : original_device_id{}, need_reset{false}
```

PR #958, opened 2022-01-27
fix uninitialized data, add check in dense stride test

yhmtsai:
This PR initializes the value in stride part of dense, and add the check of value in stride of scale-series test


upsj (approved):
LGTM! There are other strided matrices in that test, maybe check the same behavior for them?

yhmtsai:
I also add the check of value in the stride for apply result.
If there's anything missing, please let me know

pratikvn (approved):
LGTM!

PR #960, opened 2022-01-30
Two level and use type as config

yhmtsai:
This PR adds some macro helper and the kernel will use the type for device config.
Moreover, it adds two level selection, i.e. first for device attributes and second for kernel own attributes.
Macro detail:
There are two kinds of selecting config type.
1. Use ConfigSet to select the kernel. the selection will change the ConfigSet to device_config type.
pro: keep the kernel call like cuda/hip. only the additional first arg for device config. Leave the kernel only do kernel things. reuse or keep the same config uint.
con: uint to type is not a direct usage
2. Use type selection
pro: only handle type.
con: kernel call will be function containing a kernel or multiple kernel if they need the same config.

In my opinion, when porting cuda to dpcpp, I will definitely go for option 1 to reduce the difference between cuda and dpcpp (on device core view). However, for self-implemented kernel or later change, I think both should be okay If the type selection is easier to use or the uint selection does not reduce the effort.


thoasm (commented):
Review Part 1/2

dpcpp/components/prefix_sum_kernels.dp.cpp
```diff
@@ -50,18 +51,15 @@ namespace dpcpp {
 namespace components {
 
 
-using BlockCfg = ConfigSet<11>;
+constexpr auto block_cfg_list = block_cfg_list_t();
```
upsj, line 54: no need to emit a symbol for this (also might cause ODR issues if we have different values for the same variable in different files)
```suggestion
static constexpr auto block_cfg_list = block_cfg_list_t();
```

upsj (approved):
LGTM! I would be happiest if we had only a single solution (int or config, not both), since that blows up implementation_selection.hpp quite a bit. Maybe also add tests for the config selection via type?

dpcpp/base/kernel_launch_reduction.dp.hpp
```diff
@@ -46,22 +46,18 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-constexpr auto kcfg_1d_list_simple_reduction =
-    syn::value_list<std::uint32_t, KCFG_1D::encode(512, 64),
-                    KCFG_1D::encode(512, 32), KCFG_1D::encode(512, 16),
-                    KCFG_1D::encode(256, 32), KCFG_1D::encode(256, 16),
-                    KCFG_1D::encode(256, 8)>();
+constexpr auto kcfg_1d_list_simple_reduction = kcfg_1d_list_t();
 
 
-template <std::uint32_t cfg, typename ValueType, typename KernelFunction,
+template <typename cfg, typename ValueType, typename KernelFunction,
```
thoasm, line 60: Same in this file. IMO, it should be capitalized at least, so it is clear that it is a type (`DeviceConfig` would be best IMO).

dpcpp/components/prefix_sum.dp.hpp
```diff
@@ -129,14 +129,15 @@ __dpct_inline__ void subwarp_prefix_sum(ValueType element,
  * @note To calculate the prefix sum over an array of size bigger than
  *       `block_size`, `finalize_prefix_sum` has to be used as well.
  */
-template <std::uint32_t block_size, typename ValueType>
-void start_prefix_sum(size_type num_elements, ValueType* __restrict__ elements,
-                      ValueType* __restrict__ block_sum,
-                      sycl::nd_item<3> item_ct1,
-                      UninitializedArray<ValueType, block_size>& prefix_helper)
+template <typename cfg, typename ValueType>
```
thoasm, line 132: The same in this file.

dpcpp/components/reduction.dp.hpp
```diff
@@ -202,13 +198,12 @@ void reduce_array(size_type size, const ValueType* __restrict__ source,
  * `source` of any size. Has to be called a second time on `result` to reduce
  * an array larger than `block_size`.
  */
-template <std::uint32_t cfg, typename ValueType>
-void reduce_add_array(
-    size_type size, const ValueType* __restrict__ source,
-    ValueType* __restrict__ result, sycl::nd_item<3> item_ct1,
-    UninitializedArray<ValueType, KCFG_1D::decode<0>(cfg)>& block_sum)
+template <typename cfg, typename ValueType>
```
thoasm, line 201: Same in this file.

dpcpp/matrix/dense_kernels.dp.cpp
```diff
@@ -108,62 +106,74 @@ void transpose(const size_type nrows, const size_type ncols,
     }
 }
 
-template <std::uint32_t sg_size, typename ValueType>
-void transpose(const size_type nrows, const size_type ncols,
-               const ValueType* __restrict__ in, const size_type in_stride,
-               ValueType* __restrict__ out, const size_type out_stride,
-               sycl::nd_item<3> item_ct1,
-               UninitializedArray<ValueType, sg_size*(sg_size + 1)>& space)
+template <typename cfg, typename ValueType>
```
thoasm, line 109: And here.

dpcpp/matrix/ell_kernels.dp.cpp
```diff
@@ -304,8 +304,9 @@ void spmv(dim3 grid, dim3 block, size_type dynamic_shared_memory,
 namespace {
 
 
-template <int info, typename InputValueType, typename MatrixValueType,
-          typename OutputValueType, typename IndexType>
+template <int info, typename cfg, typename InputValueType,
```
thoasm, line 307: Same in this file.

dpcpp/matrix/ell_kernels.dp.cpp
```diff
@@ -315,6 +316,7 @@ void abstract_spmv(syn::value_list<int, info>,
                    const matrix::Dense<MatrixValueType>* alpha = nullptr,
                    const matrix::Dense<OutputValueType>* beta = nullptr)
 {
+    std::cout << "cfg " << cfg::block_size << std::endl;
```
thoasm, line 319: Debug output?
Should be removed.

dpcpp/test/components/cooperative_groups.dp.cpp
```diff
@@ -116,10 +113,10 @@ class CooperativeGroups : public testing::TestWithParam<unsigned int> {
 
 
 // kernel implementation
-template <std::uint32_t config>
+template <typename config>
```
thoasm, line 116: This should have a unified name like the others (applies to this whole file).

dpcpp/synthesizer/implementation_selection.hpp
```diff
@@ -0,0 +1,278 @@
...
+            _name(::gko::syn::value_list<std::uint32_t, Rest...>(),          \
+                  is_eligible, bool_args, int_args, size_args, type_args,    \
+                  std::forward<InferredArgs>(args)...);                      \
+        }                                                                    \
+    }
+
+
+#define GKO_ENABLE_IMPLEMENTATION_TWO_SELECTION_KERNEL(_name, _callable)      \
+    template <typename Predicate, bool... BoolArgs, int... IntArgs,           \
+              gko::size_type... SizeTArgs, typename... TArgs, typename CFG,   \
```
thoasm, line 154: `CFG` -> `DeviceConfig` or similar. It should be the same name as the others.

thoasm (dismissed):
Mostly comments about the `device_config` type naming.

dpcpp/base/config.hpp
```diff
@@ -72,6 +75,76 @@ struct config {
...
+    return KCFG_1D::encode(T::block_size, T::subgroup_size);
+};
+
+/**
+ * encode_list base type
+ *
+ * @tparam T  the input template
+ */
+template <typename T, typename = void>
+struct encode_list {};
```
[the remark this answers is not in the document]
yhmtsai: I use specialization to extract types... from the type_list, so I think it's needed?
but I can remove the second typename = void

dpcpp/base/helper.hpp
```diff
@@ -103,6 +103,21 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
         });                                                               \
     }
 
+#define GKO_ENABLE_DEFAULT_HOST_CONFIG_TYPE(name_, kernel_)                 \
+    template <typename KCFG, typename... InferredArgs>                      \
+    inline void name_(dim3 grid, dim3 block, gko::size_type,                \
```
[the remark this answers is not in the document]
yhmtsai: it's required for gko:: because macro can be used out of gko namespace like test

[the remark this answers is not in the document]
thoasm: my bad, this is inside a macro which I apparently did not see...

dpcpp/base/config.hpp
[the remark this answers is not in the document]
thoasm: For what do you need this specialization? To me, the one below should also work without the specialization part.

dpcpp/base/config.hpp
```diff
@@ -72,6 +76,68 @@ struct config {
...
+/**
+ * encode_list specializes for the type_list. It will convert the each type to
+ * encoded information.
+ *
+ * @tparam T  the input template
+ */
+template <typename... Types>
+struct encode_list<syn::type_list<Types...>> {
+    using type = syn::value_list<uint32, Types::encode...>;
+};
```
thoasm, line 112: This should work, so you don't need the specialization:
```suggestion
template <typename... Types>
struct encode_list {
    using type = syn::value_list<uint32, Types::encode...>;
};
```

yhmtsai: ah, you are right. I somehow think I need to use `type_list` to wrap the type list always.

yhmtsai: I need to use `type_list<type...>` as input.
I create a `type_list` type `using dcfg_type_list_t =  type_list<device_config<.>...>;` and then use it to create its value_list varient `using dcfg_list_t = encode<dcfg_type_list>::type;`

thoasm: OK, makes sense.

dpcpp/base/kernel_launch_reduction.dp.hpp
```diff
@@ -302,7 +299,7 @@ void run_kernel_reduction_cached(std::shared_ptr<const DpcppExecutor> exec,
 namespace {
 
 
-template <std::uint32_t cfg, int ssg_size, typename ValueType,
+template <typename cfg, int ssg_size, typename ValueType,
```
thoasm, line 302:
```suggestion
template <typename DeviceConfig, int ssg_size, typename ValueType,
```

dpcpp/base/kernel_launch_reduction.dp.hpp
```diff
@@ -358,16 +355,16 @@ GKO_ENABLE_IMPLEMENTATION_SELECTION(select_generic_kernel_row_reduction_2d,
                                     generic_kernel_row_reduction_2d);
 
 
-template <std::uint32_t cfg, int ssg_size, typename ValueType,
+template <typename cfg, int ssg_size, typename ValueType,
```
thoasm, line 358:
```suggestion
template <typename DeviceConfig, int ssg_size, typename ValueType,
```

dpcpp/base/kernel_launch_reduction.dp.hpp
```diff
@@ -433,16 +430,16 @@ void generic_kernel_col_reduction_2d_small(
 }
 
 
-template <std::uint32_t cfg, typename ValueType, typename KernelFunction,
+template <typename cfg, typename ValueType, typename KernelFunction,
```
thoasm, line 433:
```suggestion
template <typename DeviceConfig, typename ValueType, typename KernelFunction,
```

dpcpp/base/kernel_launch_reduction.dp.hpp
```diff
@@ -502,7 +499,7 @@ void generic_kernel_reduction_finalize_2d(
 }
 
 
-template <std::uint32_t cfg, int ssg_size, typename ValueType,
+template <typename cfg, int ssg_size, typename ValueType,
```
thoasm, line 502:
```suggestion
template <typename DeviceConfig, int ssg_size, typename ValueType,
```

dpcpp/base/kernel_launch_reduction.dp.hpp
```diff
@@ -551,7 +548,7 @@ GKO_ENABLE_IMPLEMENTATION_SELECTION(select_generic_col_reduction_small,
                                     run_generic_col_reduction_small);
 
 
-template <std::uint32_t cfg, typename ValueType, typename KernelFunction,
+template <typename cfg, typename ValueType, typename KernelFunction,
```
thoasm, line 551:
```suggestion
template <typename DeviceConfig, typename ValueType, typename KernelFunction,
```

dpcpp/base/kernel_launch_reduction.dp.hpp
```diff
@@ -595,17 +592,18 @@ void run_kernel_row_reduction_stage1(std::shared_ptr<const DpcppExecutor> exec,
...
 
-GKO_ENABLE_IMPLEMENTATION_CONFIG_SELECTION(select_kernel_row_reduction_stage1,
-                                           run_kernel_row_reduction_stage1);
+GKO_ENABLE_IMPLEMENTATION_CONFIG_SELECTION_TOTYPE(
+    select_kernel_row_reduction_stage1, run_kernel_row_reduction_stage1,
+    DCFG_1D);
 
 
-template <std::uint32_t cfg, typename ValueType, typename KernelFunction,
+template <typename cfg, typename ValueType, typename KernelFunction,
```
thoasm, line 606:
```suggestion
template <typename DeviceConfig, typename ValueType, typename KernelFunction,
```

dpcpp/test/components/cooperative_groups.dp.cpp
```diff
@@ -219,31 +203,17 @@ void cg_any(bool* s, sycl::nd_item<3> item_ct1)
...
-GKO_ENABLE_IMPLEMENTATION_CONFIG_SELECTION(cg_any, cg_any)
+GKO_ENABLE_DEFAULT_HOST_CONFIG_TYPE(cg_any, cg_any)
+GKO_ENABLE_IMPLEMENTATION_CONFIG_SELECTION_TOTYPE(cg_any, cg_any, DCFG_1D)
 GKO_ENABLE_DEFAULT_CONFIG_CALL(cg_any_call, cg_any, default_config_list)
 
 TEST_P(CooperativeGroups, Any) { test_all_subgroup(cg_any_call<bool*>); }
 
 
-template <std::uint32_t config>
+template <typename cfg>
```
thoasm, line 213:
```suggestion
template <typename DeviceConfig>
```

thoasm (approved):
Some `cfg` -> `DeviceConfig` changes were missed.
Other than that, LGTM.

PR #961, opened 2022-01-31
Distributed vector

### LINKED ISSUE #907: Create a clean distributed-ginkgo branch

MarcelKoch:
This PR adds a distributed (multi-)vector class. I've opted into using mostly multivector based descriptions instead of using dense for that. This might clash a bit with the rest, until we separate dense into multivector and dense matrix.

Regarding the tests, I've put the tests that require MPI into `/core/test/mpi/distributed`, not sure if that is the best place.

Partially closes #907 

Note: should we use `distributed-develop` as target branch again? That way we could make sure that we have all basic things (vectors, matrices, solvers) together for the next release.

Main contributions are from @upsj  and @pratikvn.


MarcelKoch:
format-rebase!

MarcelKoch:
format!

upsj (commented):
A first look over the code. The class structure and implementation looks good to me, haven't checked the tests yet.

A fundamental discussion point I would like to raise is whether Vector should store Partition. In my original design, both Matrix and Vector exclusively used local indexing after they have been filled, and I believe this is a property we could use generally. If users need to translate global indices and local indices, they have the original partition available from their setup code. Using the vector and matrix as a convenient vehicle to carry them around seems like it might overload the functionality a bit.

core/distributed/vector.cpp
```diff
@@ -0,0 +1,392 @@
...
+void Vector<ValueType, LocalIndexType, GlobalIndexType>::apply_impl(
+    const LinOp* alpha, const LinOp* b, const LinOp* beta, LinOp* x) const
+{
+    GKO_NOT_SUPPORTED(this);
+}
+
+
+template <typename ValueType, typename LocalIndexType, typename GlobalIndexType>
+Vector<ValueType, LocalIndexType, GlobalIndexType>::Vector(
+    std::shared_ptr<const Executor> exec, mpi::communicator comm,
```
pratikvn, line 72: Do we want to take the `comm` object by value ?

core/distributed/vector.cpp
```diff
@@ -0,0 +1,392 @@
...
+    dim<2> global_size, dim<2> local_size, size_type stride)
+    : EnableLinOp<
+          Vector<ValueType, LocalIndexType, GlobalIndexType>>{exec,
+                                                              global_size},
+      DistributedBase{comm},
+      partition_{
+          partition
+              ? std::move(partition)
+              : gko::share(
+                    Partition<LocalIndexType, GlobalIndexType>::create(exec))},
```
pratikvn, line 83: I am not sure if this is what we want. Storing an empty partition gives us no benefits and is probably more error-prone as users and devs might assume it is a valid object. I would probably prefer to throw if the partition is invalid.

core/distributed/vector.cpp
```diff
@@ -0,0 +1,392 @@
...
+      local_{exec, local_size,
+             stride != invalid_index<size_type>() ? stride : local_size[1]}
```
pratikvn, line 85: I am not sure I understand this. Why would we want the stride to be `local_size[1]`, if an invalid_index was somehow passed in. Maybe you meant to make that parameter optional with `size_type stride=local_size[1]` in line 76 ?

core/distributed/vector_kernels.hpp
```diff
@@ -0,0 +1,82 @@
...
+        const Array<matrix_data_entry<ValueType, GlobalIndexType>>& input,    \
+        const distributed::Partition<LocalIndexType, GlobalIndexType>*        \
+            partition,                                                        \
+        comm_index_type local_part,                                           \
+        Array<matrix_data_entry<ValueType, LocalIndexType>>& local_data,      \
+        ValueType deduction_help)
+
+
+#define GKO_DECLARE_ALL_AS_TEMPLATES                                      \
+    using comm_index_type = distributed::comm_index_type;                 \
```
pratikvn, line 64: This is probably better defined in one of the headers in the `distributed` namespace

core/test/mpi/distributed/vector.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+TYPED_TEST(Vector, CanReadLocalMatrixData)
+{
+    using md_type = typename TestFixture::md_type;
+    using part_type = typename TestFixture::part_type;
+    using value_type = typename TestFixture::value_type;
+    md_type md[3] = {
+        {gko::dim<2>{6, 2}, {{0, 0, 0}, {0, 1, 1}, {1, 0, 2}, {1, 1, 3}}},
+        {gko::dim<2>{6, 2}, {{2, 0, 4}, {2, 1, 5}, {3, 0, 6}, {3, 1, 7}}},
+        {gko::dim<2>{6, 2}, {{4, 0, 8}, {4, 1, 9}, {5, 0, 10}, {5, 1, 11}}}};
```
pratikvn, line 215: This is still visible to all ranks. I would maybe also add another test where data is only locally visible.

include/ginkgo/core/base/types.hpp
```diff
@@ -693,7 +758,7 @@ GKO_ATTRIBUTES constexpr bool operator!=(precision_reduction x,
 template <typename IndexType>
 inline constexpr GKO_ATTRIBUTES IndexType invalid_index()
 {
-    static_assert(std::is_signed<IndexType>::value,
+    static_assert(std::is_integral<IndexType>::value,
```
pratikvn, line 761: Here you are changing the meaning of this function, so if you need an integral check, then you should probably have another function specific for that ?

MarcelKoch:
Regarding the partition, my use case for that was repartitioning, which we should support in the long term. This could be achieved without storing the partition in the vector class, since the repartition would need to store it anyway. However, I don't see how we could make sure that the vector we are repartitioning is valid for that operation, if the vector does not store its partition. I think the additional safety we get from storing it is worth the downside. (I'm not entirely sure what you see as a downside, except that it is not necessary, perhaps you could elaborate that?)

Also, I think there is a strong connection between vector and partition. The vector is only valid in combination with this one partition, which is expressed in that way.

include/ginkgo/core/distributed/vector.hpp
```diff
@@ -0,0 +1,340 @@
...
+     * @param stride  Stride of the local vector. If not specified, it defaults
+     *                to local_size[1]
+     */
+    explicit Vector(
+        std::shared_ptr<const Executor> exec,
+        mpi::communicator comm = mpi::communicator(MPI_COMM_WORLD),
+        std::shared_ptr<const Partition<LocalIndexType, GlobalIndexType>>
+            partition = nullptr,
+        dim<2> global_size = {}, dim<2> local_size = {},
+        size_type stride = invalid_index<size_type>());
```
pratikvn, line 319: I think `{}` is a better default here ?

common/unified/matrix/dense_kernels.cpp
```diff
@@ -377,6 +377,37 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE(
...
+template <typename ValueType>
+void compute_norm2_sqr(std::shared_ptr<const DefaultExecutor> exec,
+                       const matrix::Dense<ValueType>* x,
+                       matrix::Dense<remove_complex<ValueType>>* result)
+{
+    run_kernel_col_reduction(
+        exec,
+        [] GKO_KERNEL(auto i, auto j, auto x) { return squared_norm(x(i, j)); },
+        [] GKO_KERNEL(auto a, auto b) { return a + b; },
+        [] GKO_KERNEL(auto a) { return a; }, remove_complex<ValueType>{},
```
pratikvn, line 389: I think you can use the shirtcut `GKO_KERNEL_REDUCE_SUM`

pratikvn (commented):
Mostly looks good. Some minor questions on function interfaces.

core/distributed/vector.cpp
MarcelKoch: Yes, that is exactly what I want, but you can't use a previous parameter as default argument ...

pratikvn:
Regarding the base class to merge this into, `distributed-develop` might be a good choice, unless we decide that the interface more or less looks good, then we can try to merge this directly into `develop`

I think storing a partition in Vector also allows us to only Vector operations on data (with a different partition), without access to matrix or other partition storing classes. Therefore, I also think it would be useful to store the partition in the vector class.

include/ginkgo/core/base/types.hpp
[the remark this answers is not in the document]
MarcelKoch: In the constructor of vector, to get around the fact that I can't use a previous parameter as the default value for the next parameter. Perhaps I could use a `struct unspecified_parameter{}` for that instead.

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -130,6 +140,9 @@ class Dense
     friend class SparsityCsr<ValueType, int32>;
     friend class SparsityCsr<ValueType, int64>;
     friend class Dense<to_complex<ValueType>>;
+    friend class distributed::Vector<ValueType, int32, int32>;
+    friend class distributed::Vector<ValueType, int32, int64>;
+    friend class distributed::Vector<ValueType, int64, int64>;
```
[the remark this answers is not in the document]
MarcelKoch: We are storing the local vector by value in vector, so I needed that. But I can do without, although the initialization of the local vector now looks a bit odd.

core/distributed/vector.cpp
```diff
@@ -0,0 +1,392 @@
...
+{
+    auto rank = comm.rank();
+
+    Array<matrix_data_entry<ValueType, LocalIndexType>> local_data{exec};
+    exec->run(vector::make_build_local(global_data, partition, rank, local_data,
+                                       ValueType{}));
+
+    auto local_rows = static_cast<size_type>(partition->get_part_size(rank));
+    dim<2> local_size{local_rows, num_cols};
+    local_mtx->read({local_size, local_data});
```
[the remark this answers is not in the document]
MarcelKoch: I think that should create a `device_matrix_data`, so that should be the same as writing directly to dense. I should use `move` to remove the unnecessary copy of the array though.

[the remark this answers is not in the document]
upsj: using device_matrix_data is at least a 2x overhead for double/int, and the complexity of the algorithm is really not necessary, because we can skip storing anything in a variable-size array

include/ginkgo/core/base/types.hpp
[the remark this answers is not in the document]
upsj: I think the most straightforward solution to that would be adding another overload of the constructor with only non-defaulted parameters

include/ginkgo/core/matrix/dense.hpp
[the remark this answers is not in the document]
upsj: Right sorry, I remember writing that now :)

core/distributed/vector.cpp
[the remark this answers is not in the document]
MarcelKoch: Ok, I think I get it now, you mean passing `local_mtx->get_values()` into `build_local` right?

include/ginkgo/core/distributed/vector.hpp
```diff
@@ -0,0 +1,340 @@
...
+
+
+#include <ginkgo/config.hpp>
+#include <ginkgo/core/base/mpi.hpp>
+#include <ginkgo/core/distributed/base.hpp>
+#include <ginkgo/core/distributed/partition.hpp>
+#include <ginkgo/core/matrix/dense.hpp>
+
+
+#if GINKGO_BUILD_MPI
```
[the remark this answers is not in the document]
MarcelKoch: I think `config.hpp` should be included, since it defines `GINKGO_BUILD_MPI`. Is it okay, formatting-wise, to split up the includes in that way?

core/distributed/vector.cpp
```diff
@@ -0,0 +1,392 @@
...
+    auto dense_res =
+        make_temporary_clone(exec, as<matrix::Dense<ValueType>>(result));
+    this->get_const_local()->compute_dot(as<Vector>(b)->get_const_local(),
+                                         dense_res.get());
+    exec->synchronize();
+    auto dense_res_host =
+        make_temporary_clone(exec->get_master(), dense_res.get());
+    this->get_communicator().all_reduce(dense_res_host->get_values(),
+                                        static_cast<int>(this->get_size()[1]),
+                                        MPI_SUM);
```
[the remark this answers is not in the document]
MarcelKoch: Right, that needs to be investigated now...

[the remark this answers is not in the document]
pratikvn: I think the way it is done in [`rebase-dist-ras-precond` branch](https://github.com/ginkgo-project/ginkgo/blob/rebase-dist-ras-precond/core/distributed/vector.cpp#L208) is a reasonable approach. At least for now, a compile time decision should be sufficient.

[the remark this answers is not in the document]
MarcelKoch: That seems reasonable to me. I will add that.

core/distributed/vector.cpp
MarcelKoch: I think the design of `mpi::communicator`` was to make it cheap to copy (among other things), so I think it should be fine. Also, this allows users to pass in `MPI_Comm`. 
Btw, what would you suggest?

core/distributed/vector.cpp
MarcelKoch: I would put validation checking off, until the PR for that is done. Until then, I would just assume that we have a valid partition. Still, I will change it to just use `std::move(partition)`.

core/test/mpi/distributed/vector.cpp
MarcelKoch: I'm only accessing `md[rank]` so I don't think it is that much different from what you describe.

include/ginkgo/core/base/types.hpp
MarcelKoch: I'm reverting that, since it is not necessary anymore.

core/distributed/vector.cpp
pratikvn: Hm, I was thinking of maybe a shared_ptr object, or maybe taking it by reference. But indeed it is true that copying it is cheap as internally it stores a shared_ptr to MPI_Comm. So I guess this should be okay.

core/distributed/vector.cpp
pratikvn: Yeah, I understand that initially you want to create the vector with a Partition object, but when going through the `read` path of filling in the vector, the partition object there takes preference. 

Another issue, which also probably the validation should take into account is that the local_size should match what the partition gives. We should also probably discuss what parameter ordering we want in this constructor and which ones should have default values and which not.

core/distributed/vector.cpp
pratikvn: You can also initialize the object in the constructor body. But I guess that is not completely ideal.

core/test/mpi/distributed/vector.cpp
pratikvn: Yes, that is true. But all the data is still visible to all ranks. I feel like it would be more fool-proof to have only local data. But this is only a nit and not something absolutely necessary to add.

core/distributed/vector.cpp
MarcelKoch: I'm open to that discussion. I think `comm` since that has always needs to be defined for a distributed object. For me the partition is the next most important thing (if we keep it), since it describes how the object is distributed. The rest is sorted as usual. The only part that I could see changing, is the position of the partition, perhaps it makes sense to switch it with the global (and local?) size.

include/ginkgo/core/matrix/dense.hpp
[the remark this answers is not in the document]
MarcelKoch: This gives me a bit more trouble than expected. I tried to do `local_{*local_vector_type::create(...)` in the constructor, but that only partially works. Somehow, the executor of the intermediate vector is not copied. Is that a bug or expected?

Also, I don't think that I can switch to using `local_` as a pointer type, since both unique and shared will mess up cloning.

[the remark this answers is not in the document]
upsj: as a friend class, you should be able to use the constructor directly, what do you need create(...) for?

core/distributed/vector.cpp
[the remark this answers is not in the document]
upsj: or passing local_mtx directly, yes

include/ginkgo/core/distributed/vector.hpp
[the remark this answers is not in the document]
upsj: right, forgot that, thanks!

include/ginkgo/core/matrix/dense.hpp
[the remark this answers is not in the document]
MarcelKoch: Yes, sorry I was thinking about ways how to do it without the friend declaration, but I think there is no way around it.

core/test/mpi/distributed/vector.cpp
MarcelKoch: I will leave it like that. I think this is simpler than splitting everything up per rank.

core/distributed/vector_kernels.hpp
MarcelKoch: This is mostly to shorten the kernel definition. I wouldn't put that in another header, since that shouldn't be publicly available. I could also just remove it.

include/ginkgo/core/base/types.hpp
```diff
@@ -559,6 +559,73 @@ GKO_ATTRIBUTES constexpr bool operator!=(precision_reduction x,
...
+ * Instantiates a template for each non-complex value, local and global index
+ * type compiled by Ginkgo.
+ *
+ * @param _macro  A macro which expands the template instantiation
+ *                (not including the leading `template` specifier).
+ *                Should take three arguments, which are replaced by the
+ *                value, the local and the global index types.
+ */
+#if GINKGO_DPCPP_SINGLE_MODE
+#define GKO_INSTANTIATE_FOR_EACH_NON_COMPLEX_VALUE_AND_LOCAL_GLOBAL_INDEX_TYPE( \
```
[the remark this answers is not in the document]
MarcelKoch: Honestly, I'm not sure how that would work. I guess it would be somewhat easy, all possible combination of local and global index type, but I wanted to not instantiate the case `local_idx_type = int64` and `global_idx_type = int32`. Do you already have something in mind for this?

MarcelKoch:
I've worked in most of the reviews. There are two open questions so far (1 minor, 1 major):
1. Do agree on the current order of the constructor parameters and their default values?
2. Do we keep the partition in the vector?

For 2. the arguments against (as I understand it) are:
- There is no use case where the partition information is required, but won't be provided elsewhere. This especially holds for the SpMV

The arguments in favor are:
- Can ensure that we always operate on consistent data, e.g. SpMV. For example, we could have cases, where the local and global sizes all match up, but the partitions are not the same. (Note: this would require testing partitions for equality in constant time)
- Simple access to the partition, without relying on other classes that might store it.

include/ginkgo/core/base/types.hpp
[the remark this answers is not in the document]
upsj: Makes sense, then nevermind :)

MarcelKoch:
I've added the `DenseCache` that is used in the distributed matrix also to this PR. If wanted, I could also extract that into its own PR, since the changes for that are quite small.

pratikvn (commented):
Mostly looks good. Some comments on the whether we want to make DenseCache public. Regarding the discussion points:

1. I think the constructor order is okay. 
2. I think a distributed object should have always have partition member. 
     - It should be something like a Concept `require` for distributed objects. 
     - There are cases for vector operations, where the partition makes sense. For example dot products between two vectors. 
     - Additionally, the read function already takes a partition. In my opinion, it might make more sense to not have the read function take a partition, but create the object with a partition and the read function use the object's partition.

MarcelKoch:
@pratikvn you raise a very important point. The current implentations are only correct, if both vectors (think axpy, dot, ...) use the same partition. It is not enough that both local vectors have the same size. This is something that we should check for at some point.

include/ginkgo/core/base/cache.hpp
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * Manages a Dense vector that is buffered and reused internally to avoid
+ * repeated allocations. Copying an instance will only yield an empty object
+ * since copying the cached vector would not make sense.
+ *
+ * @internal  The struct is present to wrap cache-like buffer storage that will
+ *            not be copied when the outer object gets copied.
+ */
+template <typename ValueType>
+struct DenseCache {
```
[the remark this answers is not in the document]
pratikvn: I mean you could put it in "core/distributed/" instead of "include/ginkgo/core/distributed"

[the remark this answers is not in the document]
pratikvn: Ah, I see. Thanks.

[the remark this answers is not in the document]
MarcelKoch: I will put it in the `detail` namespace to make it clear that this is internal.

upsj:
I tend to disagree with the premise that a vector needs to contain information that allows you to understand what its row indices mean. Take Dense for example, without context (e.g. a FEM mesh), you don't know how to interpret the different entries. That context (mapping mesh entries to DOFs) is provided by the application and not stored in the vector. I see the distributed vector in the same way: The application provides a way to interpret what the local indices mean based on its FEM mesh and partition.

Also a side note we haven't talked about: For the most efficient way to store a partition (i.e. one contiguous range for each rank), the local size is entirely sufficient to check for consistency (assuming the default MPI configuration to quit if one rank aborts) Instead of checking the global property "matching partition" everywhere, you check the local property "matching size" everywhere, which is equivalent, because in this case, the `read_distributed` function doesn't do any (implicit) global reordering.

Finally, I may not have made the fine distinction between our current approach and L/T vectors clear enough:
* Our approach partitions the matrix by row only, storing the off-diagonal block separately and communicating only via a sparse gather before applying the off-diagonal matrix.
* The CEED approach symmetrically stores a slightly larger matrix (local row + overlap) in that it extends both the rows and columns by its set of off-diagonal columns (that of course only works for pattern-symmetric matrices) and communicates both via a sparse gather before the (in this case) single SpMV and via a sparse scatter-add afterwards (to collect contributions from neighboring elements on the boundary). That means more communication, but also makes the local handling much nicer, because you basically get a local domain made out of complete elements (I am not 100% sure about the details of it, but I guess with your experience in RAS and domain composition, this might sound understandable/familiar?)
The main difference is that there is no longer a separation between diagonal and off-diagonal matrix, and you can imagine the P and P^T matrices as communication operators that do the gather/scatter operations.

MarcelKoch:
Since in the current state, the vector does not require the partition (except for consistency checking), I will remove the partition for now. I will also remove it from the subsequent matrix pull request. I'm quite sure that I will have to add it back in later, but I will see how far we can go without. As a side note, since I'm planning to keep everything in the `distributed-develop` branch until we have vector+matrix+solver+preconditioner together, we can be quite liberal with interface changes.

@upsj @pratikvn can we agree on that for the moment?

Regarding the CEED L-vector: this is also how Dune (mostly) considers its distributed data. Regardless, that is pretty far removed from our distributed model (and as @upsj mentioned probably less general). But since the L-vector style is somewhat common for FEM applications, we could consider adding either a different distributed matrix type, or a simple way to construct our distributed matrix from an L-vector matrix (preferred). Still, that should be considered later on.

pratikvn:
Considering Dense as an example, you do store its information regarding its mapping in the class. While the storage is a simple contiguous array of values (possibly padded), you store, strides, which allows you to infer meaning from that data structure. We could technically look at it just as a contiguous array and then read would have to take in all the meta information. `read(data, size, stride)`, but that is cumbersome.

I am not sure I completely agree that local sizes are sufficient to check the consistency for all operations. Consider two vectors, one stored in the same set of ranks, but with different local sizes (different partitions) and the global size being the same. Because the global size is the same, they can perform operations such as a dot product, but if you just check their local sizes, you end up concluding that they are not conformant. 

Yes, I think the CEED case is probably not general enough for us. Symmetric patterns and being that the structure has been designed with FEM specifically in mind, but I have not looked at that in detail, so I don't want to comment on that now. I will have a look and get back to you.

I think even without repartitioning, you need to have a partition object in the class when operating between objects that have different partitions. I think here you have two options: 1. Have the partition object as a class member or 2. Have free functions which take in both partitions and do the communication and setup. I prefer 1. as that is more compact and uniform for users.

Yes, we would need to propagate the partitions to all the intermediate vectors and distributed preconditioners. But I see that as something advantageous as then you can possibly have different intermediate objects with different partitions and operations between them can be expressed in a uniform fashion without having to pass partition objects everywhere.

MarcelKoch:
@pratikvn For now, we don't support redistributions. When we add that at a later point we will reevaluate, if we should store the partition in the vector class. Until then, I would like to push this forward without, as it is not necessary right now. Also, I'm a bit hesitant to allow implicit redistribution in the future, as that can be quite expensive.

Although I still I agree that at least conceptually the partition belongs to the vector class.

pratikvn:
My point is that I don't think there is any other way than adding partition to the class itself when we do support re-distributions. But I guess I am okay with looking into that when we plan to add re-distributions and we can move this forward now.

MarcelKoch:
Regarding the re-distribution, I would envision to have a class that is storing the two partitions and handles the data transfers. With that, you would not necessarily need the partition as a class member in the vector, only for consistency checking again. However, if we want implicit re-distribution, then that could not be achieved with this class.

MarcelKoch:
format!

MarcelKoch:
format!

pratikvn (commented):
Nice work. Good idea using dense cache for the reduction operations. Missing some tests, but otherwise looks good!

CMakeLists.txt
```diff
@@ -98,6 +98,8 @@ option(GINKGO_DPCPP_SINGLE_MODE "Do not compile double kernels for the DPC++ bac
 option(GINKGO_INSTALL_RPATH "Set the RPATH when installing its libraries." ON)
 option(GINKGO_INSTALL_RPATH_ORIGIN "Add $ORIGIN (Linux) or @loader_path (MacOS) to the installation RPATH." ON)
 option(GINKGO_INSTALL_RPATH_DEPENDENCIES "Add dependencies to the installation RPATH." OFF)
+option(GINKGO_FORCE_GPU_AWARE_MPI "Assert that the MPI library is GPU aware. This forces Ginkgo to assume that GPU aware functionality is available (OFF (default) or ON), but may fail
```
upsj, line 101: This is something we should address in a more robust fashion before the next release, but it shouldn't hold up this PR.

core/distributed/vector.cpp
```diff
@@ -0,0 +1,394 @@
...
+    const Partition<LocalIndexType, GlobalIndexType>* partition,
+    Vector<ValueType>* result)
+{
+    auto exec = result->get_executor();
+
+    GKO_ASSERT(partition->get_executor() == exec);
+
+    auto global_rows = static_cast<size_type>(partition->get_size());
+    auto global_cols = data.get_size()[1];
+    auto tmp = Vector<ValueType>::create(exec, result->get_communicator(),
```
upsj, line 71: Why create a temporary here, instead of resizing the existing storage? This means that even if we already have sufficient storage locally, you still discard and replace it.

core/distributed/vector.cpp
```diff
@@ -0,0 +1,394 @@
...
+
+template <typename ValueType, typename LocalIndexType, typename GlobalIndexType>
+void read_distributed_impl(
+    const device_matrix_data<ValueType, GlobalIndexType>& data,
+    const Partition<LocalIndexType, GlobalIndexType>* partition,
+    Vector<ValueType>* result)
+{
+    auto exec = result->get_executor();
+
+    GKO_ASSERT(partition->get_executor() == exec);
```
upsj, line 67: this should use make_temporary_clone instead, adding the necessary functions to Partition

core/distributed/vector.cpp
```diff
@@ -0,0 +1,394 @@
...
+    this->get_local()->fill(value);
+}
+
+
+template <typename ValueType>
+void Vector<ValueType>::convert_to(
+    Vector<next_precision<ValueType>>* result) const
+{
+    result->set_size(this->get_size());
+    result->set_communicator(this->get_communicator());
```
upsj, line 146: Are communicators immutable properties like executors or mutable? I would prefer the former, and fail if the communicators have different sizes.

core/distributed/vector.cpp
```diff
@@ -0,0 +1,394 @@
...
+template <typename ValueType>
+void Vector<ValueType>::compute_dot(const LinOp* b, LinOp* result) const
+{
+    auto exec = this->get_executor();
+    const auto comm = this->get_communicator();
+    auto dense_res =
+        make_temporary_clone(exec, as<matrix::Dense<ValueType>>(result));
+    this->get_const_local()->compute_dot(as<Vector>(b)->get_const_local(),
+                                         dense_res.get());
+    exec->synchronize();
```
upsj, line 296: Is this necessary with non GPU-aware MPI?

core/distributed/vector.cpp
```diff
@@ -0,0 +1,394 @@
...
+template <typename ValueType>
+void Vector<ValueType>::compute_conj_dot(const LinOp* b, LinOp* result) const
+{
+    auto exec = this->get_executor();
+    const auto comm = this->get_communicator();
+    auto dense_res =
+        make_temporary_clone(exec, as<matrix::Dense<ValueType>>(result));
+    this->get_const_local()->compute_conj_dot(as<Vector>(b)->get_const_local(),
+                                              dense_res.get());
+    exec->synchronize();
```
upsj, line 321: same here

core/distributed/vector.cpp
```diff
@@ -0,0 +1,394 @@
...
+void Vector<ValueType>::compute_norm2(LinOp* result) const
+{
+    using NormVector = typename local_vector_type::absolute_type;
+    GKO_ASSERT_EQUAL_DIMENSIONS(result, dim<2>(1, this->get_size()[1]));
+    auto exec = this->get_executor();
+    const auto comm = this->get_communicator();
+    auto dense_res = make_temporary_clone(exec, as<NormVector>(result));
+    exec->run(vector::make_compute_norm2_sqr(this->get_const_local(),
+                                             dense_res.get()));
+    exec->synchronize();
```
upsj, line 347: same here

core/distributed/vector.cpp
```diff
@@ -0,0 +1,394 @@
...
+template <typename ValueType>
+void Vector<ValueType>::compute_norm1(LinOp* result) const
+{
+    using NormVector = typename local_vector_type::absolute_type;
+    GKO_ASSERT_EQUAL_DIMENSIONS(result, dim<2>(1, this->get_size()[1]));
+    auto exec = this->get_executor();
+    const auto comm = this->get_communicator();
+    auto dense_res = make_temporary_clone(exec, as<NormVector>(result));
+    this->get_const_local()->compute_norm1(dense_res.get());
+    exec->synchronize();
```
upsj, line 373: same here

include/ginkgo/core/base/mpi.hpp
```diff
@@ -58,6 +58,19 @@ namespace gko {
 namespace mpi {
 
 
+/**
+ * Return if GPU aware functionality is available
+ */
+static constexpr bool is_gpu_aware()
```
upsj, line 64: This will probably depend on the specific executor

include/ginkgo/core/distributed/base.hpp
```diff
@@ -0,0 +1,80 @@
...
+class DistributedBase {
+public:
+    virtual ~DistributedBase() = default;
+
+    mpi::communicator get_communicator() const { return comm_; }
+
+    explicit DistributedBase(mpi::communicator comm) : comm_{std::move(comm)} {}
+
+protected:
+    void set_communicator(mpi::communicator comm) { comm_ = std::move(comm); }
```
upsj, line 66: Maybe consider modelling this after PolymorphicObject, which leaves `exec` unmodified on assignment

include/ginkgo/core/distributed/vector.hpp
```diff
@@ -0,0 +1,348 @@
...
+
+    void apply_impl(const LinOp*, LinOp*) const override;
+
+    void apply_impl(const LinOp*, const LinOp*, const LinOp*,
+                    LinOp*) const override;
+
+private:
+    local_vector_type local_;
+    mutable ::gko::detail::DenseCache<ValueType> host_reduction_buffer_;
+    mutable ::gko::detail::DenseCache<remove_complex<ValueType>>
```
upsj, line 336: I think to be safer, we should move the mutability into DenseCache

upsj (changes requested):
Looks good for the most part, nice job keeping everything updated! A few comments on DenseCache and DistributedBase related to (im-)mutability

include/ginkgo/core/base/mpi.hpp
MarcelKoch: Perhaps we can also postpone that to later when we do something cleverer with our cmake setup for that.

include/ginkgo/core/distributed/vector.hpp
```diff
@@ -0,0 +1,359 @@
...
+     * @param exec  Executor associated with vector
+     * @param comm  Communicator associated with vector, the default is
+     *              MPI_COMM_WORLD
+     * @param partition  Partition of global rows
+     * @param global_size  Global size of the vector
+     * @param local_size  Processor-local size of the vector
+     * @param stride  Stride of the local vector.
+     */
+    Vector(std::shared_ptr<const Executor> exec, mpi::communicator comm,
+           dim<2> global_size, dim<2> local_size, size_type stride);
```
upsj, line 322: One constructor I am missing is one that takes an existing Dense matrix and wraps it into a distributed vector by computing the global size via allreduce (wrapping it here would mean move-assignment)

core/distributed/vector.cpp
```diff
@@ -0,0 +1,389 @@
...
+{}
+
+template <typename ValueType>
+Vector<ValueType>::Vector(std::shared_ptr<const Executor> exec,
+                          mpi::communicator comm, dim<2> global_size,
+                          dim<2> local_size, size_type stride)
+    : EnableLinOp<Vector<ValueType>>{exec, global_size},
+      DistributedBase{comm},
+      local_{exec, local_size, stride}
+{}
```
upsj, line 115: We should probably check that the number of rhs matches for global size and local size?

test/mpi/distributed/vector.cpp
```diff
@@ -0,0 +1,545 @@
...
+namespace {
+
+
+bool needs_transfers(std::shared_ptr<const gko::Executor> exec)
+{
+    return exec->get_master() != exec && !gko::mpi::is_gpu_aware();
+}
+
+
+class HostToDeviceLogger : public gko::log::Logger {
```
upsj, line 57: Really nice idea :)

test/mpi/distributed/vector.cpp
```diff
@@ -0,0 +1,545 @@
...
+        dy = gko::clone(exec, y);
+    }
+
+    void SetUp() override { ASSERT_EQ(comm.size(), 3); }
+
+    void init_result()
+    {
+        res = dense_type::create(ref, gko::dim<2>{1, size[1]});
+        dres = dense_type::create(exec, gko::dim<2>{1, size[1]});
+        res->fill(0.);
```
upsj, line 150: nit: we usually write them out
```suggestion
        res->fill(0.0);
```

upsj (approved):
LGTM! Really nice job!

One thing I am not 100% sure on: Is it a good idea to have a mutable get_local() function? That allows users to accidentally clear out the internal vector. I guess a better solution could be to only return it as const, and provide get_local_values(), at_local() etc. wrappers? If people actually need to move the local data out of the distributed vector, we could provide something like the device_matrix_data::empty_out function.

include/ginkgo/core/distributed/vector.hpp
```diff
@@ -0,0 +1,359 @@
...
+     * columns is relevant. The number of rows is inferred from the partition.
+     *
+     * @note The matrix data can contain entries for rows other than those owned
+     *        by the process. Entries for those rows are discarded.
+     *
+     * @param data  The device_matrix_data structure
+     * @param partition  The global row partition
+     */
+    template <typename LocalIndexType, typename GlobalIndexType>
+    void read_distributed(
```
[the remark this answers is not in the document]
MarcelKoch: Yeah, I got probably a bit confused about the instantiation syntax for that, and thought that would not be possible.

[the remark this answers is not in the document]
upsj: you need two template <typename ...> statements, I can just never remember which comes first, class or function template parameters :laughing:

include/ginkgo/core/distributed/vector.hpp
MarcelKoch: So `Vector(exec, comm, std::move(local_vec))` without any further parameters right? I also will add a constructor `Vector(exec, comm, global_size, local_size, ValueArray&&, stride)` similar to the dense one. (I need that one for the `create_real_view` which is needed for the dispatching...)

upsj: I don't think you need the second one, because you can use the corresponding Dense constructor and pass the matrix to the first constructor. We probably need a version with and without global_size, and only compute it (via allreduce) if necessary.

I mean matrix::Dense* local_mtx and local_mtx->move_to(&member_local_matrix), because users will most likely not use references to dense matrices, but pointers.

MarcelKoch: Yes, of course, don't know what I was thinking there. But should it perhaps be `std::unique_ptr<local_type>` to make it clear that we will move from that?

core/distributed/vector.cpp
pratikvn: I dont think this is necessary even with GPU-aware MPI. All our kernel calls are currently launched on the default stream. So, everything is completely synchronous.

core/test/mpi/distributed/vector.cpp
```diff
@@ -0,0 +1,752 @@
...
+
+    VectorRead()
+        : ref(gko::ReferenceExecutor::create()),
+          comm(MPI_COMM_WORLD),
+          part(gko::share(part_type::build_from_contiguous(
+              this->ref, {ref, {0, 2, 4, 6}}))),
+          md{{0, 1}, {2, 3}, {4, 5}, {6, 7}, {8, 9}, {10, 11}}
+    {}
+
+    void SetUp() override { ASSERT_EQ(this->comm.size(), 3); }
```
pratikvn, line 77: I think we should not need to limit ourselves to 3 processes. We could use oversubscription to allow us to test for larger number of ranks. Most implementations allow it by default and just give you degraded performance (which is okay for our tests), some implementations like OpenMPI allow oversubscription with a flag (`--oversubscribe`)

core/test/mpi/distributed/vector.cpp
```diff
@@ -0,0 +1,752 @@
...
+    auto res = dense_type::absolute_type::create(this->ref, gko::dim<2>{1, 2});
+    auto ref_res = gko::initialize<typename dense_type::absolute_type>(
+        {{static_cast<gko::remove_complex<value_type>>(std::sqrt(220)),
+          static_cast<gko::remove_complex<value_type>>(std::sqrt(286))}},
+        this->ref);
+
+    this->vec_a->compute_norm2(res.get());
+
+    GKO_ASSERT_MTX_NEAR(res, ref_res, r<value_type>::value);
+}
```
pratikvn, line 358: Missing compute_norm1 ?

core/test/mpi/distributed/vector.cpp
```diff
@@ -0,0 +1,752 @@
...
+    auto b = dist_vec_type::create(this->ref, this->comm, gko::dim<2>{1, 1},
+                                   gko::dim<2>{});
+    auto c = dist_vec_type::create(this->ref, this->comm, gko::dim<2>{2, 2},
+                                   gko::dim<2>{});
+    auto d = dist_vec_type::create(this->ref, this->comm, gko::dim<2>{1, 1},
+                                   gko::dim<2>{});
+    auto e = dist_vec_type::create(this->ref, this->comm, gko::dim<2>{2, 2},
+                                   gko::dim<2>{});
+    ASSERT_THROW(a->apply(b.get(), c.get(), d.get(), e.get()),
+                 gko::NotSupported);
```
pratikvn, line 450:
```suggestion
                                   gko::dim<2>{});
                                   
    ASSERT_THROW(a->apply(b.get(), c.get(), d.get(), e.get()),
                 gko::NotSupported);
```

include/ginkgo/core/distributed/base.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+
+namespace gko {
+namespace distributed {
+
+
+/**
+ * A base class for distributed objects.
+ *
+ * This class gives access to the used mpi::communicator object.
```
pratikvn, line 53:
```suggestion
 * This class gives access to and stores the used mpi::communicator object.
```

pratikvn (approved):
Really nice work on the comprehensive tests! LGTM!

include/ginkgo/core/distributed/vector.hpp
upsj: That might unnecessarily restrict from the user's perspective: If they do their own memory management, they first need to release the pointer from their own structures and then wrap it in unique_ptr. This might be a data point to keep in mind if we decide to tackle #179 at some point.

core/distributed/vector.cpp
upsj: As far as I understood a few discussions on OpenMPI's design in a related issue, you cannot configure which stream (if any) MPI uses to execute the transfers. Might even be DMA without involvement of the SMs? So I think this will always be necessary.

core/test/mpi/distributed/vector.cpp
```diff
@@ -0,0 +1,864 @@
...
+
+    std::shared_ptr<gko::Executor> ref;
+    gko::mpi::communicator comm;
+    std::shared_ptr<part_type> part;
+
+    md_type md;
+    md_type md_localized[3];
+};
+
+
```
yhmtsai, line 87:
```suggestion
```
nit

core/test/mpi/distributed/vector.cpp
```diff
@@ -0,0 +1,864 @@
...
+    vec->read_distributed(this->md, this->part.get());
+
+    GKO_ASSERT_EQUAL_DIMENSIONS(vec->get_size(), gko::dim<2>(6, 2));
+    GKO_ASSERT_EQUAL_DIMENSIONS(vec->get_local()->get_size(),
+                                gko::dim<2>(2, 2));
+    I<I<value_type>> ref_data[3] = {
+        {{0, 1}, {2, 3}},
+        {{4, 5}, {6, 7}},
+        {{8, 9}, {10, 11}},
+    };
```
yhmtsai, line 107: this part can be moved to the position after `auto rank = ...`

core/test/mpi/distributed/vector.cpp
```diff
@@ -0,0 +1,864 @@
...
+    GKO_ASSERT_MTX_NEAR(vec->get_local(), ref_data[rank], r<value_type>::value);
```
yhmtsai, line 108:
```suggestion
    GKO_ASSERT_MTX_NEAR(vec->get_local(), ref_data[rank], 0.0);
```

core/test/mpi/distributed/vector.cpp
```diff
@@ -0,0 +1,864 @@
...
+        {{4, 5}, {6, 7}},
+        {{8, 9}, {10, 11}},
+    };
+
+    vec->read_distributed(md, part.get());
+
+    GKO_ASSERT_EQUAL_DIMENSIONS(vec->get_size(), gko::dim<2>(6, 2));
+    GKO_ASSERT_EQUAL_DIMENSIONS(vec->get_local()->get_size(),
+                                gko::dim<2>(2, 2));
+    GKO_ASSERT_MTX_NEAR(vec->get_local(), ref_data[rank], r<vt>::value);
```
yhmtsai, line 164:
```suggestion
    GKO_ASSERT_MTX_NEAR(vec->get_local(), ref_data[rank], 0.0);
```

core/test/mpi/distributed/vector.cpp
```diff
@@ -0,0 +1,864 @@
...
+        {{0, 1}, {6, 7}, {10, 11}},
+        {{2, 3}},
+        {{4, 5}, {8, 9}},
+    };
+
+    vec->read_distributed(md, part.get());
+
+    GKO_ASSERT_EQUAL_DIMENSIONS(vec->get_size(), gko::dim<2>(6, 2));
+    GKO_ASSERT_EQUAL_DIMENSIONS(vec->get_local()->get_size(), ref_size[rank]);
+    GKO_ASSERT_MTX_NEAR(vec->get_local(), ref_data[rank], r<value_type>::value);
```
yhmtsai, line 189:
```suggestion
    GKO_ASSERT_MTX_NEAR(vec->get_local(), ref_data[rank], 0.0);
```
also for other test which only do data movement/assignment

core/test/mpi/distributed/vector.cpp
```diff
@@ -0,0 +1,864 @@
...
+        gko::Array<vt>{this->ref, I<vt>{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11}}};
+    auto part = gko::share(
+        part_type::build_from_contiguous(this->ref, {this->ref, {0, 2, 4, 6}}));
+    auto vec = TestFixture::dist_vec_type::create(this->ref, this->comm);
+    auto rank = this->comm.rank();
+    I<I<vt>> ref_data[3] = {
+        {{0, 1}, {2, 3}},
+        {{4, 5}, {6, 7}},
+        {{8, 9}, {10, 11}},
+    };
```
yhmtsai, line 157: you can also put it in the TestFixture as the answer of md_localized

core/test/mpi/distributed/vector.cpp
```diff
@@ -0,0 +1,864 @@
...
+
+    std::unique_ptr<dist_vec_type> vec_a;
+    std::unique_ptr<dist_vec_type> vec_b;
+};
+
+
+TYPED_TEST_SUITE(VectorReductions, gko::test::ValueTypes);
+
+
+TYPED_TEST(VectorReductions, ComputesDotProduct)
```
yhmtsai, line 332: I think these tests are not belong to core test

core/distributed/vector.cpp
```diff
@@ -0,0 +1,503 @@
...
+    Vector<ValueType>* result)
+{
+    auto exec = result->get_executor();
+
+    auto rank = result->get_communicator().rank();
+    result->get_local()->fill(zero<ValueType>());
+    exec->run(vector::make_build_local(
+        data, make_temporary_clone(exec, partition).get(), rank,
+        result->get_local()));
+}
```
yhmtsai, line 72: Is this unused?

yhmtsai (changes requested):
First part of my review

core/distributed/vector.cpp
MarcelKoch: Yes, I think that is a rebase mistake.

core/matrix/dense_kernels.hpp
```diff
@@ -148,6 +148,15 @@ namespace kernels {
                              const device_matrix_data<_type, _prec>& data, \
                              matrix::Dense<_type>* output)
 
+#define GKO_DECLARE_DENSE_COMPUTE_NORM2_SQR_KERNEL(_type)               \
+    void compute_norm2_sqr(std::shared_ptr<const DefaultExecutor> exec, \
```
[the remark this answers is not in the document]
MarcelKoch: Yes, it's too close, I will change it.

core/test/mpi/distributed/vector.cpp
upsj: This question of where to place tests has a bit of a long-winded answer. Originally, everything in `core` tested only code that called no kernels anywhere. Then this became infeasible because even `read` became a kernel call, so now we need to build `reference` kernels every time. In total, I think this would be best suited for `test`, because it also involves the interaction with device memory for different executors.

putting them in `reference` is not an issue, since all reference tests are linked against `ginkgo`, not `ginkgo_reference` alone.

core/matrix/dense.cpp
```diff
@@ -343,7 +345,11 @@ void Dense<ValueType>::convert_to(Dense<ValueType>* result) const
 template <typename ValueType>
 void Dense<ValueType>::move_to(Dense<ValueType>* result)
 {
-    this->convert_to(result);
+    if (this != result) {
```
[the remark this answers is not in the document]
upsj: No, it should not. But at the same time, this should be handled in operator= via EnablePolymorphicAssignment, not in move_to. See #753 for an implementation that keeps the moved-from state valid.

include/ginkgo/core/distributed/vector.hpp
```diff
@@ -0,0 +1,375 @@
...
+ * if both vectors where created using the same partition.
+ *
+ * @tparam ValueType  The precision of vector elements.
+ *
+ * @ingroup dist_vector
+ * @ingroup distributed
+ */
+template <typename ValueType = double>
+class Vector
+    : public EnableLinOp<Vector<ValueType>>,
```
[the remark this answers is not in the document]
yhmtsai: I see

core/test/mpi/distributed/vector.cpp
yhmtsai: I think putting it into test is good because the test also need checked on gpu with gpu-aware mpi

test/utils/executor.hpp
```diff
@@ -33,9 +33,15 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #ifndef GKO_TEST_UTILS_EXECUTOR_HPP_
 #define GKO_TEST_UTILS_EXECUTOR_HPP_
 
+
```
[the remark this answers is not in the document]
MarcelKoch: I thought there should be always 2 lines separating each include block? Or does it not count as an independent block due to the if guard?

reference/test/matrix/dense_kernels.cpp
```diff
@@ -630,6 +630,59 @@ TYPED_TEST(Dense, ComputesNorm2Mixed)
...
+    using MixedT = typename MixedMtx::value_type;
+    using MixedT_nc = gko::remove_complex<MixedT>;
+    using MixedNormVector = gko::matrix::Dense<MixedT_nc>;
+    auto mtx(gko::initialize<MixedMtx>(
+        {I<MixedT>{1.0, 0.0}, I<MixedT>{2.0, 3.0}, I<MixedT>{2.0, 4.0}},
+        this->exec));
+    auto result = MixedNormVector::create(this->exec, gko::dim<2>{1, 2});
+
+    gko::kernels::reference::dense::compute_norm2_sqr(
+        gko::as<gko::ReferenceExecutor>(this->exec), mtx.get(), result.get());
```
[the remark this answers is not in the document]
MarcelKoch: Good catch, the kernel itself actually does not support mixed types, so I've removed the test. But I've adjusted the existing `compute_norm2` test for the mixed case.

core/matrix/dense.cpp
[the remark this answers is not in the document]
MarcelKoch: I would hope that #753 gets merged before this gets merged into develop, so I would like to keep this at the moment. Also I will adapt the convert/move for the distributed vector to use the operator=.

core/test/mpi/distributed/vector.cpp
MarcelKoch: That gives me some segfaults, so I will keep them in each test at the moments.

test/utils/executor.hpp
```diff
@@ -85,4 +91,62 @@ void init_executor(std::shared_ptr<gko::ReferenceExecutor> ref,
...
+    exec = gko::CudaExecutor::create(device_id, ref);
+}
+
+
+void init_executor(std::shared_ptr<gko::ReferenceExecutor> ref,
+                   std::shared_ptr<gko::HipExecutor>& exec,
+                   gko::mpi::communicator comm)
+{
+    ASSERT_GT(gko::HipExecutor::get_num_devices(), 0);
+    auto device_id = comm.node_local_rank() % gko::HipExecutor::get_num_devices();
```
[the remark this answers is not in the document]
MarcelKoch: At least on horeka that works, if I don't use gpu-aware mpi. If that is enabled, this part still works, but the reduction test fail later on.

pratikvn:
@MarcelKoch, can you have a look at the [code smells reported by Sonarcloud](https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&pullRequest=961&id=ginkgo-project_ginkgo&open=AX9A-Vj1fHjDZn1uKbFT) ? I think some of them are quite relevant.

MarcelKoch:
@pratikvn could you point me to some that you find particularly important? Most seem to concern things we usually (have to) do, like the const_cast, which I think is the most major one, capturing everything in a lambda, shadow variable in lambda, etc...

reference/test/distributed/vector_kernels.cpp
```diff
@@ -0,0 +1,162 @@
...
+};
+
+TYPED_TEST_SUITE(Vector, gko::test::ValueLocalGlobalIndexTypes);
+
+
+TYPED_TEST(Vector, BuildsLocalEmpty)
+{
+    using local_index_type = typename TestFixture::local_index_type;
+    using global_index_type = typename TestFixture::global_index_type;
+    this->mapping = {this->ref, {1, 0, 2, 2, 0, 1, 1, 2}};
```
yhmtsai, line 115: this mapping is recreated in each test, so it does not need to be in TestFixture

omp/test/distributed/vector_kernels.cpp
```diff
@@ -0,0 +1,225 @@
...
+    auto md = gko::test::generate_random_matrix_data<ValueType, IndexType>(
+        num_rows, num_cols, std::forward<NonzeroDistribution>(nonzero_dist),
+        std::forward<ValueDistribution>(value_dist),
+        std::forward<Engine>(engine));
+    md.ensure_row_major_order();
+    return gko::device_matrix_data<ValueType, IndexType>::create_from_host(exec,
+                                                                           md);
+}
+
+
```
yhmtsai, line 124:
```suggestion

```

yhmtsai (commented):
the second part of my review. some question and nit

core/base/dense_cache.cpp
```diff
@@ -0,0 +1,68 @@
...
+}
+
+
+template <typename ValueType>
+void DenseCache<ValueType>::init_from(
+    const matrix::Dense<ValueType>* template_vec) const
+{
+    if (!vec || vec->get_size() != template_vec->get_size() ||
+        vec->get_executor() != template_vec->get_executor()) {
+        vec = matrix::Dense<ValueType>::create_with_config_of(template_vec);
```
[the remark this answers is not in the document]
MarcelKoch: yes and that is the intended behavior (I also mention that in the documentation)

core/distributed/vector.cpp
```diff
@@ -0,0 +1,472 @@
...
+template <typename ValueType>
+Vector<ValueType>::Vector(std::shared_ptr<const Executor> exec,
+                          mpi::communicator comm,
+                          local_vector_type* local_vector)
+    : EnableLinOp<Vector<ValueType>>{exec, {}},
+      DistributedBase{comm},
+      local_{exec}
+{
+    this->set_size(compute_global_size(comm, local_vector->get_size()));
+    local_vector->move_to(&local_);
```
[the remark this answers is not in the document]
MarcelKoch: yes, the idea is that the distributed vector takes ownership of the local vector. There is no need to have it twice, since you can get read/write access from the distributed vector.

core/distributed/vector.cpp
```diff
@@ -0,0 +1,472 @@
...
+
+template <typename ValueType>
+std::unique_ptr<typename Vector<ValueType>::real_type>
+Vector<ValueType>::create_real_view()
+{
+    const auto num_global_rows = this->get_size()[0];
+    const auto num_cols =
+        is_complex<ValueType>() ? 2 * this->get_size()[1] : this->get_size()[1];
+
+    return real_type ::create(this->get_executor(), this->get_communicator(),
```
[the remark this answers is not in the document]
MarcelKoch: curious, that is something that is not covered by clang format, there isn't even an option for that...

core/distributed/vector_kernels.hpp
```diff
@@ -0,0 +1,81 @@
...
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+#ifndef GKO_CORE_DISTRIBUTED_VECTOR_KERNELS_HPP_
+#define GKO_CORE_DISTRIBUTED_VECTOR_KERNELS_HPP_
+
+
+// can't include ginkgo/core/distributed/vector.hpp since that requires linking
+// against MPI
```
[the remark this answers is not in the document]
MarcelKoch: yes, the kernels are always compiled, as they don't need mpi. But I will check if its feasible to only compile them conditionally.

core/distributed/vector.cpp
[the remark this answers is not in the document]
upsj: related: This is the type of output you get when unpacking a macro containing namespaces, IIRC.

test/mpi/distributed/vector.cpp
```diff
@@ -0,0 +1,917 @@
...
+    md_type md;
+    md_type md_localized[3];
+
+    std::default_random_engine engine;
+};
+
+TYPED_TEST_SUITE(VectorCreation, gko::test::ValueLocalGlobalIndexTypes);
+
+
+#ifdef GKO_COMPILING_REFERENCE
```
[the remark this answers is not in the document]
MarcelKoch: This is a macro specifically from the `ginkgo_create_common_*_test` function. It sets the `GKO_COMPILIN_*` macro for each device that it's compiling.

test/mpi/distributed/vector.cpp
```diff
@@ -0,0 +1,917 @@
...
+              this->ref, {ref, {0, 2, 4, 6}}))),
+          local_size{4, 11},
+          size{local_size[1] * comm.size(), 11},
+          md{{0, 1}, {2, 3}, {4, 5}, {6, 7}, {8, 9}, {10, 11}},
+          md_localized{{{0, 1}, {2, 3}}, {{4, 5}, {6, 7}}, {{8, 9}, {10, 11}}}
+    {}
+
+    void SetUp() override
+    {
+        ASSERT_GE(this->comm.size(), 3);
```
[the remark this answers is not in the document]
MarcelKoch: yes, you're right, some test won't work otherwise. Thanks for bringing that up.

reference/distributed/vector_kernels.cpp
```diff
@@ -0,0 +1,95 @@
...
+    auto range_bounds = partition->get_range_bounds();
+    auto range_parts = partition->get_part_ids();
+    auto range_starting_indices = partition->get_range_starting_indices();
+    auto num_ranges = partition->get_num_ranges();
+
+    auto find_range = [&](GlobalIndexType idx, size_type hint) {
+        if (range_bounds[hint] <= idx && idx < range_bounds[hint + 1]) {
+            return hint;
+        } else {
+            auto it = std::upper_bound(range_bounds + 1,
```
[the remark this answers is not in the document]
MarcelKoch: Because we are checking here which range `[a, b)` contains a current index, i.e. for which range `a<=i<b` holds. And we do that by finding the range with the largest index s.t. its end is smaller than the index. So we only want to iterate over the end points of all ranges, which is why we start with `range_bounds[1]` the end point of the first range.

include/ginkgo/core/distributed/vector.hpp
```diff
@@ -0,0 +1,375 @@
...
+ * auto vector = Vector<...>::create(exec, comm);
+ * vector->read_distributed(matrix_data, part);
+ * ```
+ * Using this approach the size of the global vectors, as well as the size of
+ * the local vectors, will be automatically inferred. It is possible to create a
+ * vector with specified global and local sizes and fill the local vectors using
+ * the accessor get_local.
+ *
+ * @note Operations between two vectors (axpy, dot product, etc.) are only valid
+ * if both vectors where created using the same partition.
```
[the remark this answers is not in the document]
MarcelKoch: The restriction is only conceptual. All operations where the local sizes are the same will work. But these local vectors of the same size might be interpreted differently if they were create (read) using different partitions. Currently, we don't have a way to check if two vectors were created with the same partition, so I could only add this note to the documentation.

core/distributed/vector.cpp
```diff
@@ -0,0 +1,472 @@
...
+{}
+
+
+template <typename ValueType>
+Vector<ValueType>::Vector(std::shared_ptr<const Executor> exec,
+                          mpi::communicator comm, dim<2> global_size,
+                          dim<2> local_size, size_type stride)
+    : EnableLinOp<Vector<ValueType>>{exec, global_size},
+      DistributedBase{comm},
+      local_{exec, local_size, stride}
```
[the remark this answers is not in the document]
MarcelKoch: I didn't put that in here, since that would be more suited for the validation stuff. Also checking against `compute_global_size` requires a global communication, which you might not want.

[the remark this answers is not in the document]
yhmtsai: Do it need the corresponding constructor like(exec, local_size, stride) as the local_vector cases?

include/ginkgo/core/distributed/vector.hpp
```diff
@@ -0,0 +1,416 @@
...
+
+    void resize(dim<2> global_size, dim<2> local_size);
+
+    void apply_impl(const LinOp*, LinOp*) const override;
+
+    void apply_impl(const LinOp*, const LinOp*, const LinOp*,
+                    LinOp*) const override;
+
+private:
+    local_vector_type local_;
```
yhmtsai, line 403: use Dense directly not with shared pointer?

yhmtsai (commented):
I think storing the partition in distributed should be common? I need to go over the above discussion. I think we can not move user given data, shouldn't just store the shared pointer?

include/ginkgo/core/distributed/vector.hpp
upsj: value semantics ;)

core/distributed/vector.cpp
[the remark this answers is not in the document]
upsj: moving in this case means just swapping a few pointers, no actual data movement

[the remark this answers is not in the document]
yhmtsai: Yes, but original object can not be used anymore.
If user creates the dense view, then the copy only copies the view pointer, the data is not copied.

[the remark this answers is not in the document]
upsj: yes, and that is intended (if I understand you correctly). That allows us to wrap user data also into distributed objects. A copy/clone of a non-composite object (like Dense) always performs a deep copy, i.e. it copies the values into a new array. If you want to reuse the storage, you need to move from it.

yhmtsai:
TBH, I still feel the pointer usage inconsistent. there are documentation on the constructor, so I do not strictly stop this pull request

pratikvn:
@MarcelKoch, none are major. A couple of variable updates with shadowing local variables, The DistributedBase to follow rule of 5 instead of rule of zero, because I think you need that copy assignment there. Some minor things in `test/mpi/distributed/vector.cpp`: virtual destructor for HostToDeviceLogger, usage of `make_unique`.

PR #962, opened 2022-02-01
Batch diagonal matrix type for 'batch scaling'

Slaedr:
Adds a new batch matrix type for batches of diagonal matrices. The 'batch diagonal scaling' operations are now reformulated in terms of the new `BatchDiagonal`.


cuda/matrix/batch_diagonal_kernels.cu
```diff
@@ -0,0 +1,103 @@
...
+    GKO_DECLARE_BATCH_DIAGONAL_SIMPLE_APPLY_KERNEL);
+
+
+template <typename ValueType>
+void conj_transpose(std::shared_ptr<const CudaExecutor> exec,
+                    const matrix::BatchDiagonal<ValueType>* orig,
+                    matrix::BatchDiagonal<ValueType>* trans)
+{
+    GKO_NOT_IMPLEMENTED;
+    // use batch dense conjugate transpose?
```
pratikvn, line 93: Probably can be easily implemented with the unified kernel interface ?

include/ginkgo/core/matrix/batch_diagonal.hpp
```diff
@@ -0,0 +1,626 @@
...
+    auto mtx = batch_diag::create(exec);
+    tmp->move_to(mtx.get());
+    return mtx;
+}
+
+
+}  // namespace gko
+
+
+#endif  // GKO_PUBLIC_CORE_MATRIX_BATCH_DENSE_HPP_
```
pratikvn, line 626:
```suggestion
#endif  // GKO_PUBLIC_CORE_MATRIX_BATCH_DIAGONAL_HPP_
```

pratikvn (commented):
Nice work! We could think about moving some of the batch kernels to the unified kernel interface. Especially the scaling would quite nicely there as the kernels are quite simple. 

Only a couple of other nits. Otherwise, looks good.

Slaedr:
format!

core/matrix/batch_diagonal.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+
+#define GKO_DECLARE_TWO_SIDED_BATCH_SYSTEM_TRANSFORM(_type)   \
+    void two_sided_batch_system_transform(                    \
+        std::shared_ptr<const Executor> exec,                 \
+        const BatchDiagonal<_type>* left_op,                  \
+        const BatchDiagonal<_type>* rght_op, BatchLinOp* mtx, \
+        BatchDense<_type>* rhs)
+GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE(
+    GKO_DECLARE_TWO_SIDED_BATCH_SYSTEM_TRANSFORM);
```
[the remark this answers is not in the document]
Slaedr: It depends on a kernel call. That would bring non-public stuff into a public header.

core/matrix/batch_diagonal.cpp
```diff
@@ -0,0 +1,255 @@
...
+template <typename ValueType>
+void BatchDiagonal<ValueType>::apply_impl(const BatchLinOp* b,
+                                          BatchLinOp* x) const
+{
+    if (!this->get_size().stores_equal_sizes()) {
+        GKO_NOT_IMPLEMENTED;
+    }
+    x->copy_from(b);
+    if (auto xd = dynamic_cast<BatchDense<ValueType>*>(x)) {
+        this->get_executor()->run(batch_diagonal::make_simple_apply(this, xd));
```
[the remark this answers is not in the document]
Slaedr: Yep, this is now rewritten.

core/matrix/batch_diagonal.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+template <typename ValueType>
+std::unique_ptr<BatchLinOp> BatchDiagonal<ValueType>::transpose() const
+{
+    auto exec = this->get_executor();
+    auto trans_cpy =
+        BatchDiagonal::create(exec, gko::transpose(this->get_size()));
+
+    exec->copy(this->get_num_stored_elements(), this->get_const_values(),
+               trans_cpy->get_values());
```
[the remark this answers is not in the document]
Slaedr: How? The size needs to be transposed.

Slaedr:
@pratikvn Thanks for taking a look so soon, but it was a bit too soon. If you could take a look at the changes now, that would be great.

core/test/utils/batch.hpp
```diff
@@ -298,27 +301,72 @@ void check_relative_diff(const matrix::BatchCsr<ValueType>* const a_ref,
...
+                refnorm += gko::squared_norm(h_a->get_const_values()[iz]);
             }
         }
         ASSERT_LE(relerr / refnorm, tolerance);
     }
 }
 
+
+template <typename Mtx>
+std::pair<bool, size_type> check_relative_difference(const Mtx* const a_ref,
```
Slaedr, line 333: I will remove the redundancy with the above function before merging this.

core/matrix/batch_diagonal.cpp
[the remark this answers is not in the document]
pratikvn: Isn't diagonal always square ?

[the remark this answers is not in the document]
Slaedr: No, I thought we can allow non-square diagonal matrices, since it's not too much more work right now to enable it. And mathematically they are sometimes useful.

include/ginkgo/core/base/batch_lin_op.hpp
```diff
@@ -697,7 +694,8 @@ class EnableBatchScaling : public BatchScalable {
          * BatchLinOp so that the following dynamic_cast is not needed.
```
pratikvn, line 694: Can we have a `constexpr` type static cast here, which tries to check if the type is a `BatchLinOp` at compile time, so that this dynamic_cast is not needed anymore ?

include/ginkgo/core/matrix/batch_diagonal.hpp
```diff
@@ -0,0 +1,631 @@
...
+    {
+        return values_.get_data()[linearize_index(batch, idx)];
+    }
+
+    /**
+     * @copydoc BatchDiagonal::at(size_type, size_type, size_type)
+     */
+    ValueType at(size_type batch, size_type idx) const noexcept
+    {
+        return values_.get_const_data()[linearize_index(batch, idx)];
```
pratikvn, line 288: We should probably add a note this this might not be the most efficient call, though it simplifies the usage and access to data. Unless it is always inlined, the compiler might not be able to optimize kernels using this call.

include/ginkgo/core/matrix/batch_diagonal.hpp
```diff
@@ -0,0 +1,631 @@
...
+ * @tparam Matrix  matrix type to initialize
+ *
+ * @param vals  values used to initialize the batch vector
+ * @param exec  Executor associated to the vector
+ *
+ * @ingroup BatchLinOp
+ * @ingroup mat_formats
+ */
+template <typename ValueType>
+std::unique_ptr<matrix::BatchDiagonal<ValueType>> batch_diagonal_initialize(
```
pratikvn, line 553: Do we need to name the function as such ? Can we not just put it in the `batch_diagonal` namespace and name the function as `initialize` ?

reference/test/matrix/batch_diagonal_kernels.cpp
```diff
@@ -0,0 +1,502 @@
...
+
+    GKO_ASSERT_BATCH_MTX_NEAR(trans_as_batch_diag, check, 0.0);
+}
+
+
+using namespace std::complex_literals;
+
+
+template <typename T>
+class ComplexBatchDiagonal : public ::testing::Test {
```
pratikvn, line 394: Nice that you test with non-trivial complex values, but I think without loss of generality, we could also directly test it in the above framework without a specific complex test suite ?

core/matrix/batch_diagonal_kernels.hpp
```diff
@@ -0,0 +1,128 @@
...
+#define GKO_DECLARE_BATCH_DIAGONAL_APPLY_KERNEL(_type)      \
+    void apply(std::shared_ptr<const DefaultExecutor> exec, \
+               const matrix::BatchDiagonal<_type>* a,       \
+               const matrix::BatchDense<_type>* b,          \
+               matrix::BatchDense<_type>* x)
+
+#define GKO_DECLARE_BATCH_DIAGONAL_SIMPLE_APPLY_KERNEL(_type)      \
+    void simple_apply(std::shared_ptr<const DefaultExecutor> exec, \
+                      const matrix::BatchDiagonal<_type>* a,       \
+                      matrix::BatchDense<_type>* b)
```
pratikvn, line 59: Is this being used anywhere ? I think this is an in-place apply ? I think you should name it that way.

pratikvn (approved):
Looks good to me!

include/ginkgo/core/base/batch_lin_op.hpp
Slaedr: I don't think I know about this. Is that even possible? I thought the only way would be to make this a CRTP class, but that seems a bit overkill.

include/ginkgo/core/matrix/batch_diagonal.hpp
Slaedr: At higher opimization setting, I guess `linearize_index` will be inlined. But still, I'll add a note because in spite of inlining, it might still be slower than manually indexing under specific assumptions known to the application.

include/ginkgo/core/matrix/batch_diagonal.hpp
Slaedr: We don't normally have type-specific namespaces in our public interface. What I wanted to do was add this as a specialization of `batch_initialize`, but it led to some annoying substitution debugging, IIRC. I think it can probably be done, but perhaps I could do it in a separate PR later.

core/matrix/batch_diagonal_kernels.hpp
Slaedr: No, for now it's not being used. Would you say we should remove it? I kept it because at some point we may need to offer an interface to this too, something like `apply_in_place`.

reference/test/matrix/batch_diagonal_kernels.cpp
Slaedr: I would then be creating the complex-valued matrices within the individual test functions, and not using the class member matrices. Or I would add the complex-valued matrices to the larger test suite class. But I do want to test at least conjugate transpose, and preferably transpose, with actual complex values.

Slaedr:
format!

PR #963, opened 2022-02-04
Move `device_matrix_data` to SoA storage

upsj:
Instead of storing tuples (row, column, value), store three separate vectors row_idxs, col_idxs, values. This also makes it possible to move data directly into types like Csr and Coo, so I added an rvalue overload to `read` that avoids a copy in this case.


omp/components/device_matrix_data_kernels.cpp
```diff
@@ -50,43 +50,51 @@ namespace components {
 
 template <typename ValueType, typename IndexType>
 void remove_zeros(std::shared_ptr<const DefaultExecutor> exec,
-                  Array<matrix_data_entry<ValueType, IndexType>>& data)
+                  Array<ValueType>& values, Array<IndexType>& row_idxs,
+                  Array<IndexType>& col_idxs)
```
pratikvn, line 54: I think this might be a candidate for using plain pointers + size instead of wrapping in arrays ?

omp/components/device_matrix_data_kernels.cpp
```diff
@@ -96,9 +104,13 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
...
 void sort_row_major(std::shared_ptr<const DefaultExecutor> exec,
-                    Array<matrix_data_entry<ValueType, IndexType>>& data)
+                    device_matrix_data<ValueType, IndexType>& data)
 {
-    std::sort(data.get_data(), data.get_data() + data.get_num_elems());
+    Array<matrix_data_entry<ValueType, IndexType>> tmp{exec,
+                                                       data.get_num_elems()};
+    soa_to_aos(exec, data, tmp);
+    std::sort(tmp.get_data(), tmp.get_data() + tmp.get_num_elems());
+    aos_to_soa(exec, tmp, data);
```
pratikvn, line 113: Its unfortunate that the stdlib does not have a zip_iterator

pratikvn (approved):
LGTM!

core/base/device_matrix_data.cpp
```diff
@@ -54,57 +58,162 @@ GKO_REGISTER_OPERATION(sort_row_major, components::sort_row_major);
...
+    const
+{
+    return col_idxs_.get_const_data();
+}
+
+
+template <typename ValueType, typename IndexType>
+IndexType* device_matrix_data<ValueType, IndexType>::get_row_idxs()
+{
+    return row_idxs_.get_data();
```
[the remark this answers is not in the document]
upsj: I was trying to keep the headers small, but that actually came back to bite me, since it creates a circular dependency. I'll move them to the header.

dpcpp/components/device_matrix_data_kernels.dp.cpp
```diff
@@ -48,22 +48,34 @@ namespace components {
 
 template <typename ValueType, typename IndexType>
 void remove_zeros(std::shared_ptr<const DefaultExecutor> exec,
-                  Array<matrix_data_entry<ValueType, IndexType>>& data)
+                  Array<ValueType>& values, Array<IndexType>& row_idxs,
```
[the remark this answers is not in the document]
upsj: I think that's not worth it, the details (thrust needs device_ptr, onedpl doesn't, algorithms and iterator types are named differently etc.)

omp/components/device_matrix_data_kernels.cpp
upsj: Here I need to resize the arrays, so that's not possible unfortunately.

omp/components/device_matrix_data_kernels.cpp
upsj: yes, we have IteratorFactory and the whole shebang, but it only works with pairs, not arbitrary tuples. Also a parallel std::sort would be really helpful

upsj: #966 (:

omp/components/device_matrix_data_kernels.cpp
pratikvn: Ah, I overlooked the resize.

MarcelKoch (changes requested):
Some small remarks, otherwise looks good. I would like to see the tests extended a bit as mentioned below.

- Why are the kernel implementations under `/device/components/` instead of `/device/base/`, since that's where the header + core implementation is located?

- I think we should add tests for copy constructor and `resize_and_reset`? Also, do we have tests for `read(device_matrix_data)`? At least, I couldn't find them (that would explain the missing coverage).

- Also, it seems really annoying that in the openMP case we don't have a standard algorithms library. Someone mentioned that thrust might also work in that case. Perhaps we should look into having parallelized version of at least some stl algorthms.

include/ginkgo/core/base/device_matrix_data.hpp
```diff
@@ -99,8 +129,8 @@ struct device_matrix_data {
      *          host executor or copied into an owning Array if `exec` is a
      *          device executor.
```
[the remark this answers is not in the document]
upsj: Right, I forgot to update the documentation, thanks!

include/ginkgo/core/base/device_matrix_data.hpp
```diff
@@ -116,21 +146,174 @@ struct device_matrix_data {
...
+     *
+     * @return a struct containing the internal arrays.
+     */
+    arrays empty_out();
+
+private:
+    dim<2> size_;
+    Array<value_type> values_;
+    Array<index_type> col_idxs_;
+    Array<index_type> row_idxs_;
```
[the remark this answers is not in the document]
upsj: That would be possible, but it makes the constructor more unclear, and it's an internal detail, so I would probably prefer staying more explicit here.

upsj:
@MarcelKoch I totally agree on the missing parallel std::sort. That is pretty non-trivial, so maybe we can manage to do that in another PR. The kernels are in `components` because we don't have any kernels in `base` right now and I didn't want to change that yet - but that is a good point. I added tests for matrix::read and the missing device_matrix_data functionality.

upsj:
@MarcelKoch I also moved everything to `base`, we do have some kernels in there after all.

include/ginkgo/core/base/device_matrix_data.hpp
```diff
@@ -94,13 +125,11 @@ struct device_matrix_data {
...
-     * @return  a device_matrix_data object with the same size as `data` and the
-     *          same entries, either wrapped as a non-owning view if `exec` is a
-     *          host executor or copied into an owning Array if `exec` is a
-     *          device executor.
+     * @return  a device_matrix_data object with the same size and entries as
+     *          `data` copied to the device executor.
      */
-    static device_matrix_data create_view_from_host(
-        std::shared_ptr<const Executor> exec, host_type& data);
+    static device_matrix_data create_from_host(
```
MarcelKoch, line 131: I can't put in my suggestion directly, since the comment I would like to change is outside your changes. Still, I would change the following line above:
```
Creates a view of the given host data on the given executor.
```
to something like
```
Creates a device_matrix_data from the given host data on the given executor.
```

MarcelKoch (approved):
I have still some small nits, but these won't hold up the PR.
- Should the device_matrix_data common unit tests stay under `/test/components` or should it be moved to `/test/base/?
- Should the kernels stay in the `components` namespace? 

In both cases I'm just checking if that was considered.

PR #964, opened 2022-02-04
Create submatrix from Index sets

pratikvn:
This PR adds functionality to create submatrices from IndexSet objects, allowing one to create submatrices not only from contiguous spans, but also with dis-contiguous sets of indices. 

+ Currently, only CSR is supported

Some index set related changes were also made:
+ Make index set a non-polymorphic class.
+ Rename IndexSet to index_set

### TODO
+ [x] Remove the workaround and pre-compute the local index array to use it in the computations


pratikvn:
format!

reference/test/matrix/csr_kernels.cpp
```diff
@@ -1695,4 +1695,69 @@ TYPED_TEST(Csr, CanGetSubmatrix2)
...
+            I<T>{0.0, 3.0, 4.5, 0.0, 2.0},   // 2
+            I<T>{0.0, -1.0, 2.5, 0.0, 2.0},  // 3
+            I<T>{1.0, 0.0, -1.0, 3.5, 1.0},  // 4
+            I<T>{0.0, 1.0, 0.0, 0.0, 2.0},   // 5
+            I<T>{0.0, 3.0, 0.0, 7.5, 1.0}    // 6
+        },
+        this->exec);
+    ASSERT_EQ(mat->get_num_stored_elements(), 23);
+    {
+        auto row_set = gko::IndexSet<index_type>(this->exec, {0, 1});
```
yhmtsai, line 1717:
```suggestion
        SCOPED_TRACE("Sub Test Name")
        auto row_set = gko::IndexSet<index_type>(this->exec, {0, 1});
```
but you can also put the mat to the class, and separate these subtests to tests?

.github/abidiff.sh
```diff
@@ -6,4 +6,4 @@ popd
 
 bot_delete_comments_matching "Note: This PR changes the Ginkgo ABI"
 
-abidiff build-old/lib/libginkgod.so build-new/lib/libginkgod.so &> abi.diff || (bot_comment "Note: This PR changes the Ginkgo ABI:\n\`\`\`\n$(head -n2 abi.diff | tr '\n' ';' | sed 's/;/\\n/g')\`\`\`\nFor details check the full ABI diff under **Artifacts** [here]($JOB_URL)" && false)
+abidiff build-old/lib/libginkgod.so build-new/lib/libginkgod.so &> abi.diff || (bot_comment "Note: This PR changes the Ginkgo ABI:\n\`\`\`\n$(head -n2 abi.diff | tr '\n' ';' | sed 's/;/\\n/g')\`\`\`\nFor details check the full ABI diff under **Artifacts** [here]($JOB_URL)")
```
yhmtsai, line 9: is the abidiff check always mark passed now?

yhmtsai (changes requested):
some nit and unused parameter needs to be deleted

core/base/index_set.cpp
```diff
@@ -105,6 +105,20 @@ IndexType IndexSet<IndexType>::get_local_index(const IndexType index) const
 }
 
 
+template <typename IndexType>
+IndexType IndexSet<IndexType>::get_subset_id(const IndexType index) const
+{
+    auto exec = this->get_executor();
+    auto ss_end_host = Array<IndexType>(exec->get_master(), this->subsets_end_);
```
upsj, line 112: make_temporary_clone

include/ginkgo/core/base/index_set.hpp
```diff
@@ -196,6 +218,28 @@ class IndexSet : public EnablePolymorphicObject<IndexSet<IndexType>> {
...
+     *       throughput.
+     *
+     * @param global_index  the global index.
+     *
+     * @return  the local index of the element in the index set.
+     *
+     * @warning This single entry query can have significant kernel lauch
+     *          overheads and should be avoided if possible.
+     */
+    index_type get_subset_id(index_type global_index) const;
```
upsj, line 241: Do we need this functionality? Its performance is even worse than get_local_index and it's not used anywhere outside tests

omp/base/index_set_kernels.cpp
```diff
@@ -148,29 +148,29 @@ GKO_INSTANTIATE_FOR_EACH_INDEX_TYPE(GKO_DECLARE_INDEX_SET_POPULATE_KERNEL);
...
-        GKO_ASSERT(index < index_space_size);
-        auto bucket =
-            std::distance(subset_begin->get_const_data(),
-                          std::upper_bound(subset_begin->get_const_data(),
-                                           subset_begin->get_const_data() +
-                                               subset_begin->get_num_elems(),
-                                           index));
+    for (size_type i = 0; i < num_indices; ++i) {
+        auto index = global_indices[i];
+        if (index > index_space_size) {
```
upsj, line 161: you need to be careful with your comparison operators! This kind of issue is unlikely to be found via tests and can cause hard-to-find bugs later
```suggestion
        if (index >= index_space_size) {
```

omp/base/index_set_kernels.cpp
```diff
@@ -148,29 +148,29 @@ GKO_INSTANTIATE_FOR_EACH_INDEX_TYPE(GKO_DECLARE_INDEX_SET_POPULATE_KERNEL);
...
+            local_indices[i] = invalid_index<IndexType>();
+            continue;
+        }
+        const auto bucket = std::distance(
+            subset_begin,
+            std::upper_bound(subset_begin, subset_begin + num_subsets, index));
```
upsj, line 167: shouldn't this be using subset_end? Then you also won't need the following line

omp/base/index_set_kernels.cpp
```diff
@@ -182,29 +182,24 @@ GKO_INSTANTIATE_FOR_EACH_INDEX_TYPE(
...
-                 ->get_const_data()[superset_indices->get_num_elems() - 1]));
-        auto bucket = std::distance(
-            superset_indices->get_const_data(),
-            std::upper_bound(superset_indices->get_const_data(),
-                             superset_indices->get_const_data() +
-                                 superset_indices->get_num_elems(),
-                             index));
+    for (size_type i = 0; i < num_indices; ++i) {
+        auto index = local_indices[i];
+        GKO_ASSERT(index <= (superset_indices[num_subsets]));
```
upsj, line 195:
```suggestion
        GKO_ASSERT(index < superset_indices[num_subsets]);
```

omp/matrix/csr_kernels.cpp
```diff
@@ -742,6 +744,49 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
...
+                exec,
+                static_cast<size_type>(src_ptrs[row + 1] - src_ptrs[row]));
+            gko::kernels::omp::index_set::global_to_local(
+                exec, col_index_set.get_size(), col_index_set.get_num_subsets(),
+                col_index_set.get_subsets_begin(),
+                col_index_set.get_subsets_end(),
+                col_index_set.get_superset_indices(),
+                static_cast<IndexType>(l_idxs.get_num_elems()),
+                source->get_const_col_idxs() + src_ptrs[row], l_idxs.get_data(),
+                false);
```
upsj, line 774: This seems really inefficient to me - it needs a lot of space even if extracting a small submatrix, and the information isn't reused between the two kernel calls. If we force the index set to be sorted (which IMO is a good invariant to have independent of this kernel), we can use upper_bound and cache the previous result, like we do in some Partition kernels.

upsj (dismissed):
There are some subtle issues with the algorithms that should be addressed.

include/ginkgo/core/base/index_set.hpp
pratikvn: Yes, it is useful to implement some quick checks and prototyping and I agree that this is inefficient and should not be used in performance critical areas, and hence the warning.

omp/matrix/csr_kernels.cpp
pratikvn: I am not sure you can re-use the information between two calls. The sets of global indices can be completely different. Unless you cache the local indices of the entire col_idxs of the parent matrix, I think it might be better to look-up the indices in every row that is relevant to the submatrix ? 

And the index_set itself is always sorted, but I am not sure we can guarantee that the indices themselves that we need to convert are sorted.

upsj: Ah sorry, I mixed up the loops. I thought you were computing the local indices for the entire slice. Then my comment is more specific: Allocating inside such a hot loop will not give you good performance, and you read each output entry of the global_to_local kernel only once. A solution without allocation would be to use the inner kernel from global_to_local directly.

reference/base/index_set_kernels.cpp
```diff
@@ -207,37 +207,31 @@ GKO_INSTANTIATE_FOR_EACH_INDEX_TYPE(
...
         }
-        auto index = local_indices->get_const_data()[i];
-        GKO_ASSERT(
-            index <=
-            (superset_indices
-                 ->get_const_data()[superset_indices->get_num_elems() - 1]));
-        auto shifted_superset =
-            &superset_indices->get_const_data()[shifted_bucket];
+        auto index = local_indices[i];
+        GKO_ASSERT(index <= (superset_indices[num_subsets]));
```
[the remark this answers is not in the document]
pratikvn: No, it was an oversight. Fixed now

omp/matrix/csr_kernels.cpp
MarcelKoch: You could also allocate an array with the maximal number of nnz per row outside these loops and then use a view of that, if you want to keep the kernel call.

pratikvn: Yes, that is what I went with now for the reference version.

omp/base/index_set_kernels.cpp
pratikvn: I think that is more or less equivalent ? I think you need to have a shifted index anyway to access the superset_indices later.

reference/test/matrix/csr_kernels.cpp
pratikvn: Yeah, I added the SCOPED_TRACE, but when it fails, it already gives the line of failure, so otherwise also, it is okay I think.

yhmtsai: ah right, SCOPED_TRACE is more helpful for the additional function call, not this case.
Sorry about that

omp/matrix/csr_kernels.cpp
yhmtsai: It only computes the number of valid elements, so it shoule not need to convert global to local with allocation.
going through the intexset and the col index is enough to get the number.

yhmtsai: you can also use the submatrix row pointer information to create the compete `l_idx` out of the for-loop in `compute_submatrix_from_index_set` if you do not want to compute the maximum per each set.

yhmtsai (changes requested):
there are some unused parameters. compute_submatrix_from_index_set is directly tested in omp but not reference. also some performance comments

core/matrix/csr.cpp
```diff
@@ -612,6 +617,49 @@ Csr<ValueType, IndexType>::create_submatrix(const gko::span& row_span,
...
+            row_index_set.get_subsets_begin());
+        auto row_end = row_index_set.get_executor()->copy_val_to_host(
+            row_index_set.get_subsets_end());
+        auto col_st = col_index_set.get_executor()->copy_val_to_host(
+            col_index_set.get_subsets_begin());
+        auto col_end = col_index_set.get_executor()->copy_val_to_host(
+            col_index_set.get_subsets_end());
+
+        return this->create_submatrix(span(row_st, row_end),
+                                      span(col_st, col_end));
```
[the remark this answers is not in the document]
pratikvn: Yes, that has already been added to #402

reference/test/matrix/csr_kernels.cpp
```diff
@@ -1760,4 +1764,73 @@ TYPED_TEST(Csr, CanGetSubmatrix2)
 }
 
 
+TYPED_TEST(Csr, CanGetSubmatrixWithIndexSet)
```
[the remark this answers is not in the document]
pratikvn: This test calls that kernel.

omp/matrix/csr_kernels.cpp
pratikvn: Going through the index_set/querying if an index is in the index set requires conversion from global_to_local. I moved that out of the loop with max_nnz_per_row for both ref and omp now. 

Yes, I can use to compute the entire l_idx outside the for loop, but I feel that is unnecessary and requires more memory ?

yhmtsai: the first part of global_to_local without allocation checks the validation and it does not transfer the data to local yet
```
for (size_type i = 0; i < num_indices; ++i) {
    auto index = global_indices[i];
    if (index >= index_space_size) {
        continue;
    }
    const auto bucket = std::distance(
        subset_begin,
        std::upper_bound(subset_begin, subset_begin + num_subsets, index));
    auto shifted_bucket = bucket == 0 ? 0 : (bucket - 1);
    if (subset_end[shifted_bucket] <= index) {
        continue;
    }
    nnz[res_row]++;
}
```

omp/test/matrix/csr_kernels.cpp
```diff
@@ -776,7 +776,6 @@ TEST_F(Csr, ComputeSubmatrixFromIndexSetIsEquivalentToRef)
         this->ref, row_nnz.get_data(), row_nnz.get_num_elems());
     auto num_nnz = row_nnz.get_data()[rset.get_num_elems()];
     auto drow_nnz = gko::Array<int>(this->omp, row_nnz);
-    drow_nnz.fill(gko::one<int>());
```
yhmtsai: does it give error?

pratikvn: No, I was doing it incorrectly. I am already copying from row_nnz, so there is no need to fill it after that.

omp/matrix/csr_kernels.cpp
```diff
@@ -775,6 +826,64 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
...
+    auto num_col_subsets = col_index_set.get_num_subsets();
+    auto col_subset_begin = col_index_set.get_subsets_begin();
+    auto col_subset_end = col_index_set.get_subsets_end();
+    auto col_superset_indices = col_index_set.get_superset_indices();
+    const auto src_ptrs = source->get_const_row_ptrs();
+    const auto src_col_idxs = source->get_const_col_idxs();
+    const auto src_values = source->get_const_values();
+
+    size_type res_nnz = 0;
+    for (size_type set = 0; set < num_row_subsets; ++set) {
```
yhmtsai, line 854: #pragma unroll ?
you build the `row_ptr` from `calculate_nonzeros_per_row_in_index_set`
the `res_nnz` can be initial from `res_row_ptr[row]` and inside the loop, I think?

omp/test/matrix/csr_kernels.cpp
```diff
@@ -755,6 +755,47 @@ TEST_F(Csr, ComputeSubmatrixIsEquivalentToRef)
...
+    gko::IndexSet<IndexType> dcset(this->omp, cset);
+    auto size = this->mtx2->get_size();
+    auto row_nnz = gko::Array<int>(this->ref, rset.get_num_elems() + 1);
+    row_nnz.fill(gko::zero<int>());
+    gko::kernels::reference::csr::calculate_nonzeros_per_row_in_index_set(
+        this->ref, this->mtx2.get(), rset, cset, &row_nnz);
+    gko::kernels::reference::components::prefix_sum(
+        this->ref, row_nnz.get_data(), row_nnz.get_num_elems());
+    auto num_nnz = row_nnz.get_data()[rset.get_num_elems()];
+    auto drow_nnz = gko::Array<int>(this->omp, row_nnz);
```
yhmtsai, line 778: ah, here. fill different value to ensure they are all modified by function

reference/test/matrix/csr_kernels.cpp
```diff
@@ -1760,4 +1764,85 @@ TYPED_TEST(Csr, CanGetSubmatrix2)
...
+            I<T>{1.0, 0.0, 4.5, 7.5, 3.0},   // 1
+            I<T>{0.0, 3.0, 4.5, 0.0, 2.0},   // 2
+            I<T>{0.0, -1.0, 2.5, 0.0, 2.0},  // 3
+            I<T>{1.0, 0.0, -1.0, 3.5, 1.0},  // 4
+            I<T>{0.0, 1.0, 0.0, 0.0, 2.0},   // 5
+            I<T>{0.0, 3.0, 0.0, 7.5, 1.0}    // 6
+        },
+        this->exec);
+    ASSERT_EQ(mat->get_num_stored_elements(), 23);
+    {
```
yhmtsai, line 1785:
```suggestion
        this->exec);

    ASSERT_EQ(mat->get_num_stored_elements(), 23);

    {
```

reference/test/matrix/csr_kernels.cpp
```diff
@@ -1760,4 +1764,85 @@ TYPED_TEST(Csr, CanGetSubmatrix2)
...
+                                          I<T>{0.0, 3.0, 7.5, 1.0}},  // 6
+                                         this->exec);
+
+        GKO_EXPECT_MTX_NEAR(sub_mat1.get(), ref1.get(), 0.0);
+    }
+
+    {
+        SCOPED_TRACE("Non Square 2x4");
+        auto row_set = gko::IndexSet<index_type>(this->exec, {5, 6});
+        auto col_set = gko::IndexSet<index_type>(this->exec, {4, 3, 0, 1});
```
yhmtsai, line 1837: if we set it is sorted, we will get the permutated one?
sub_col 0 <-> src_col 4
sub_col 1 <-> src_col 3
sub_col 2 <-> src_col 0
sub_col 3 <-> src_col 1

yhmtsai (approved):
LGTM in the sorted IndexSet. some performance nit on omp.
Could we use it to get the specified order submatrix?
sub row 0 <-> src row 3
sub row 1 <-> src row 0

reference/test/matrix/csr_kernels.cpp
pratikvn: No, the index_set always sorts its input and then stores them. I wanted to make sure that with unsorted inputs to index_set, this algorithm still works

omp/test/matrix/csr_kernels.cpp
pratikvn: I added another test for `calculate_nnz_from_from_index_set` to test that kernel separetely.

omp/matrix/csr_kernels.cpp
pratikvn: `row` here is the global row, so I would need to convert it to the local submatrix row first, I think. But that should work now. Thanks.

core/base/index_set.cpp
```diff
@@ -105,6 +105,23 @@ IndexType IndexSet<IndexType>::get_local_index(const IndexType index) const
...
+    auto ss_end_host = make_temporary_clone<const Array<IndexType>>(
+        this->get_executor()->get_master(), &this->subsets_end_);
+    auto ss_begin_host = make_temporary_clone<const Array<IndexType>>(
+        this->get_executor()->get_master(), &this->subsets_begin_);
+    for (size_type id = 0; id < this->get_num_subsets(); ++id) {
+        if (index < ss_end_host->get_const_data()[id] &&
+            index >= ss_begin_host->get_const_data()[id]) {
+            return id;
+        }
+    }
```
upsj, line 120: I'd prefer this being handled by a kernel call like get_local_index, if at all

include/ginkgo/core/base/index_set.hpp
```diff
@@ -100,6 +100,27 @@ class IndexSet : public EnablePolymorphicObject<IndexSet<IndexType>> {
...
+     * @param is_sorted  a parameter that specifies if the indices array is
+     *                   sorted or not. `true` if sorted.
+     */
+    IndexSet(std::shared_ptr<const gko::Executor> executor,
+             std::initializer_list<IndexType> init_list,
+             const bool is_sorted = false)
+        : EnablePolymorphicObject<IndexSet>(std::move(executor)),
+          index_space_size_(
+              *(std::max_element(std::begin(init_list), std::end(init_list))) +
+              1)
```
upsj, line 118: this would likely segfault on an empty initializer list

include/ginkgo/core/base/index_set.hpp
```diff
@@ -169,7 +190,7 @@ class IndexSet : public EnablePolymorphicObject<IndexSet<IndexType>> {
      * @param local_index  the local index.
      * @return  the global index from the index set.
      *
-     * @warning This single entry query can have significant kernel lauch
```
upsj, line 172: orihime-inoue-bleach

include/ginkgo/core/matrix/csr.hpp
```diff
@@ -792,6 +793,10 @@ class Csr : public EnableLinOp<Csr<ValueType, IndexType>>,
 
     std::unique_ptr<Diagonal<ValueType>> extract_diagonal() const override;
 
+    std::unique_ptr<Csr<ValueType, IndexType>> create_submatrix(
```
upsj, line 796: I think this could also use some documentation?
```suggestion
    std::unique_ptr<Csr> create_submatrix(
```

reference/test/matrix/csr_kernels.cpp
upsj: maybe add a comment here explaining that detail?

reference/test/matrix/csr_kernels.cpp
```diff
@@ -1760,4 +1764,87 @@ TYPED_TEST(Csr, CanGetSubmatrix2)
...
+            I<T>{0.0, -1.0, 2.5, 0.0, 2.0},  // 3
+            I<T>{1.0, 0.0, -1.0, 3.5, 1.0},  // 4
+            I<T>{0.0, 1.0, 0.0, 0.0, 2.0},   // 5
+            I<T>{0.0, 3.0, 0.0, 7.5, 1.0}    // 6
+        },
+        this->exec);
+
+    ASSERT_EQ(mat->get_num_stored_elements(), 23);
+
+    {
```
upsj, line 1787: Test for empty IndexSet and full IndexSet maybe?

omp/base/index_set_kernels.cpp
```diff
@@ -63,21 +65,18 @@ namespace index_set {
...
     for (size_type subset = 0; subset < num_subsets; ++subset) {
-        for (size_type i = 0; i < ss_indices[subset + 1] - ss_indices[subset];
-             ++i) {
-            indices[ss_indices[subset] + i] =
-                subset_begin->get_const_data()[subset] + i;
+        for (size_type i = 0;
+             i < superset_indices[subset + 1] - superset_indices[subset]; ++i) {
+            decomp_indices[superset_indices[subset] + i] =
+                subset_begin[subset] + i;
         }
```
upsj, line 80: could be turned into the more common form
```suggestion
        IndexType local_i{};
        for (auto i = superset_indices[subset]; i < superset_indices[subset + 1]; ++i) {
            decomp_indices[i] = local_i + subset_begin[subset];
            local_i++;
        }
```

omp/base/index_set_kernels.cpp
upsj: yes, I guess so. If I read upper_bound, I am thinking: you are looking for something that is larger than the entry, and subset_end would be less surprising

omp/base/index_set_kernels.cpp
```diff
@@ -148,29 +147,30 @@ GKO_INSTANTIATE_FOR_EACH_INDEX_TYPE(GKO_DECLARE_INDEX_SET_POPULATE_KERNEL);
...
+            continue;
+        }
+        const auto bucket = std::distance(
+            subset_begin,
+            std::upper_bound(subset_begin, subset_begin + num_subsets, index));
         auto shifted_bucket = bucket == 0 ? 0 : (bucket - 1);
-        if (subset_end->get_const_data()[shifted_bucket] <= index) {
-            local_indices->get_data()[i] = invalid_index<IndexType>();
+        if (subset_end[shifted_bucket] <= index ||
+            index < subset_begin[shifted_bucket]) {
```
upsj, line 169:
```suggestion
        if (index >= subset_end[shifted_bucket] ||
            index < subset_begin[shifted_bucket]) {
```

omp/base/index_set_kernels.cpp
```diff
@@ -181,30 +181,26 @@ GKO_INSTANTIATE_FOR_EACH_INDEX_TYPE(
...
-                 ->get_const_data()[superset_indices->get_num_elems() - 1]));
-        auto bucket = std::distance(
-            superset_indices->get_const_data(),
-            std::upper_bound(superset_indices->get_const_data(),
-                             superset_indices->get_const_data() +
-                                 superset_indices->get_num_elems(),
-                             index));
+    for (size_type i = 0; i < num_indices; ++i) {
+        auto index = local_indices[i];
+        if (index >= superset_indices[num_subsets]) {
```
upsj, line 193: `index` could also be negative

omp/base/index_set_kernels.cpp
```diff
@@ -181,30 +181,26 @@ GKO_INSTANTIATE_FOR_EACH_INDEX_TYPE(
...
+            global_indices[i] = invalid_index<IndexType>();
+            continue;
+        }
+        const auto bucket = std::distance(
+            superset_indices,
+            std::upper_bound(superset_indices,
+                             superset_indices + num_subsets + 1, index));
         auto shifted_bucket = bucket == 0 ? 0 : (bucket - 1);
```
upsj, line 201: No need for the shift if you already use shifted inputs
```suggestion
        const auto bucket = std::distance(
            superset_indices + 1,
            std::upper_bound(superset_indices + 1,
                             superset_indices + num_subsets + 1, index));
```

omp/matrix/csr_kernels.cpp
```diff
@@ -742,6 +744,55 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
...
+                    std::upper_bound(col_subset_begin,
+                                     col_subset_begin + num_col_subsets,
+                                     index));
+                auto shifted_bucket = bucket == 0 ? 0 : (bucket - 1);
+                if (col_subset_end[shifted_bucket] <= index ||
+                    (index < col_subset_begin[shifted_bucket])) {
+                    continue;
+                } else {
+                    row_nnz->get_data()[res_row]++;
+                }
```
upsj, line 785:
```suggestion
                if (index < col_subset_end[shifted_bucket] <= index &&
                    index >= col_subset_begin[shifted_bucket]) {
                    row_nnz->get_data()[res_row]++;
                }
```

omp/matrix/csr_kernels.cpp
```diff
@@ -742,6 +744,55 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
     GKO_DECLARE_CSR_CALC_NNZ_PER_ROW_IN_SPAN_KERNEL);
 
 
+template <typename ValueType, typename IndexType>
+void calculate_nonzeros_per_row_in_index_set(
+    std::shared_ptr<const DefaultExecutor> exec,
+    const matrix::Csr<ValueType, IndexType>* source,
+    const IndexSet<IndexType>& row_index_set,
+    const IndexSet<IndexType>& col_index_set, Array<IndexType>* row_nnz)
```
upsj, line 752: Is there a reason for passing an array here?
```suggestion
    const IndexSet<IndexType>& col_index_set, IndexType* row_nnz)
```

omp/matrix/csr_kernels.cpp
```diff
@@ -775,6 +826,68 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
...
+                const auto bucket = std::distance(
+                    col_subset_begin,
+                    std::upper_bound(col_subset_begin,
+                                     col_subset_begin + num_col_subsets,
+                                     index));
+                auto shifted_bucket = bucket == 0 ? 0 : (bucket - 1);
+                if (col_subset_end[shifted_bucket] <= index ||
+                    (index < col_subset_begin[shifted_bucket])) {
+                    continue;
+                } else {
```
upsj, line 875: same here: invert the condition to avoid the `continue`

include/ginkgo/core/base/index_set.hpp
```diff
@@ -100,6 +100,27 @@ class IndexSet : public EnablePolymorphicObject<IndexSet<IndexType>> {
...
+    /**
+     * Creates an index set on the specified executor from the initializer list.
+     *
+     * @param exec  the Executor where the index set data will be allocated
+     * @param init_list  the indices that the index set should hold in an
+     *                   initializer_list.
+     * @param is_sorted  a parameter that specifies if the indices array is
+     *                   sorted or not. `true` if sorted.
+     */
+    IndexSet(std::shared_ptr<const gko::Executor> executor,
```
upsj, line 112:
```suggestion
    explicit IndexSet(std::shared_ptr<const gko::Executor> executor,
```
same for the other constructors

upsj (commented):
A few details I would like to see changed, otherwise nice work! I think IndexSet needs some more attention also from the value semantics standpoint, because it is a polymorphic object, but does not provide a `create` method. Does it need to be polymorphic? Doesn't look like it has virtual methods. Maybe we should treat it like Partition? (and maybe that one shouldn't be a polymorphic type either?)

include/ginkgo/core/base/index_set.hpp
pratikvn: I had completely forgotten about this. All the weird dubs that people made over this was hilarious.

core/base/index_set.cpp
```diff
@@ -131,9 +149,10 @@ Array<IndexType> IndexSet<IndexType>::map_local_to_global(
 
     GKO_ASSERT(this->get_num_subsets() >= 1);
     exec->run(index_set::make_local_to_global(
-        this->index_space_size_, &this->subsets_begin_, &this->subsets_end_,
-        &this->superset_cumulative_indices_, &local_indices, &global_indices,
-        is_sorted));
+        this->get_num_subsets(), this->get_subsets_begin(),
+        this->get_superset_indices(),
+        static_cast<IndexType>(local_indices.get_num_elems()),
```
MarcelKoch, line 154: perhaps it would be better to change the kernel parameter type to `size_type`? (Also for other similar kernels)

core/base/index_set_kernels.hpp
```diff
@@ -67,25 +66,22 @@ namespace kernels {
...
+        const IndexType* subset_begin, const IndexType* subset_end,     \
+        const IndexType* superset_indices, const IndexType num_indices, \
+        const IndexType* global_indices, IndexType* local_indices,      \
+        const bool is_sorted)
+
+#define GKO_DECLARE_INDEX_SET_LOCAL_TO_GLOBAL_KERNEL(IndexType)         \
+    void local_to_global(                                               \
+        std::shared_ptr<const DefaultExecutor> exec,                    \
+        const IndexType num_subsets, const IndexType* subset_begin,     \
+        const IndexType* superset_indices, const IndexType num_indices, \
```
MarcelKoch, line 82:
```suggestion
        const IndexType* superset_indices, const size_type num_indices, \
```

reference/test/matrix/csr_kernels.cpp
MarcelKoch: Also perhaps with only one empty set?

MarcelKoch (approved):
LGTM, not much else to remark here.

As a side node, I'm a bit unhappy that `IndexSet::get_num_elems` and `IndexSet::get_num_subsets` return `index_type` and not `gko::size_type`. I think the latter one is more appropriate, since we use that one in nearly all other similar contexts. But I guess that is not completely relevant to this PR, and it will have to wait for 2.0 either way.

include/ginkgo/core/matrix/csr.hpp
```diff
@@ -35,6 +35,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
 #include <ginkgo/core/base/array.hpp>
+#include <ginkgo/core/base/index_set.hpp>
```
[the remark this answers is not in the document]
pratikvn: Yeah, I feel an include where possible is a much cleaner way to go rather than having incomplete types, especially when we know that the class/file we are including is a basic data structure and is guaranteed to have no dependence on the class it is being included in.

core/base/index_set.cpp
pratikvn: I wanted to be a bit more explicit, so I prefer having `IndexType` as the type for the `num_indices` parameter. In general, we also wanted to move away from using `size_type` and using `IndexType` everywhere.

include/ginkgo/core/base/index_set.hpp
```diff
@@ -79,27 +79,50 @@ namespace gko {
...
  * @tparam index_type  type of the indices being stored in the index set.
  *
- * @ingroup IndexSet
+ * @ingroup index_set
  */
 template <typename IndexType = int32>
-class IndexSet : public EnablePolymorphicObject<IndexSet<IndexType>> {
-    friend class EnablePolymorphicObject<IndexSet>;
-
+class index_set {
```
upsj, line 85: Probably needs rule-of-five overloads, otherwise you can have an `index_set` that claims to contain data, but its arrays are empty.

pratikvn: I think the `default` ones do the right thing here ? So rule of zero should be enough ?

pratikvn: Ah, right. I see what you mean.

include/ginkgo/core/base/index_set.hpp
```diff
@@ -89,13 +89,25 @@ class index_set {
      */
     using index_type = IndexType;
 
+    /**
+     * Creates an empty Array not tied to any executor.
+     */
+    index_set() noexcept
+        : exec_(nullptr), index_space_size_{0}, num_stored_indices_{0}
+    {}
+
```
upsj, line 98: I don't think you need this!

include/ginkgo/core/base/index_set.hpp
```diff
@@ -144,20 +156,125 @@ class index_set {
...
+        if (&other == this) {
+            return *this;
+        }
+        if (other.get_executor() == nullptr) {
+            this->clear();
+            return *this;
+        }
+        if (exec_ == nullptr) {
+            this->exec_ = other.get_executor();
+        }
```
upsj, line 220: These can then be removed
```suggestion
```

include/ginkgo/core/base/index_set.hpp
```diff
@@ -144,20 +156,125 @@ class index_set {
...
         this->index_space_size_ = other.index_space_size_;
         this->num_stored_indices_ = other.num_stored_indices_;
-        subsets_begin_ = gko::Array<index_type>(exec, other.subsets_begin_);
-        subsets_end_ = gko::Array<index_type>(exec, other.subsets_end_);
-        superset_cumulative_indices_ =
-            gko::Array<index_type>(exec, other.superset_cumulative_indices_);
+        other.index_space_size_ = 0;
+        other.num_stored_indices_ = 0;
```
upsj, line 255: a more concise pattern for this kind of operation is
```cpp
this->index_space_size_ = std::exchange(other.index_space_size_, 0);
```

include/ginkgo/core/base/index_set.hpp
```diff
@@ -345,11 +462,11 @@ class index_set {
...
-    index_type index_space_size_;
-    index_type num_stored_indices_;
-    gko::Array<index_type> subsets_begin_;
-    gko::Array<index_type> subsets_end_;
-    gko::Array<index_type> superset_cumulative_indices_;
+    index_type index_space_size_{};
+    index_type num_stored_indices_{};
+    gko::Array<index_type> subsets_begin_{};
+    gko::Array<index_type> subsets_end_{};
+    gko::Array<index_type> superset_cumulative_indices_{};
```
upsj, line 469: I don't think we need this, since all remaining constructors initialize them already!

include/ginkgo/core/base/index_set.hpp
```diff
@@ -144,20 +156,125 @@ class index_set {
...
+        if (&other == this) {
+            return *this;
+        }
+        if (other.get_executor() == nullptr) {
+            this->clear();
+            return *this;
+        }
+        if (exec_ == nullptr) {
+            this->exec_ = other.get_executor();
+        }
```
upsj, line 251: same here: can be removed

upsj (approved):
LGTM! Only I don't think we need an executor-less `index_set`

PR #967, opened 2022-02-08
Use vendor implementations for some Dense kernels in Ginkgo

pratikvn:
~~This PR adds a strategy enum to allow runtime switching between vendor and Ginkgo provided kernels.~~

~~+ Is not kernel specific, but Dense object specific.~~
~~+ `create_with_config_of` and `create_with_type_of` set the same strategy as their parent.~~
~~+ Strategies of objects created within Ginkgo cannot be controlled, but set to default, which is currently using Ginkgo kernels.~~

~~This is still a WIP. Only `dot` has been implemented as a prototype. You can see how it works in the cuda/hip dense kernel tests.~~

The implementation switch is now inside the kernel and the single vector implementation is the vendor implementation (where available) and multi-vector always uses Ginkgo's implementation. Only the following routines are replaced:

1. Dot
2. Conjugate dot
3. Norm2


include/ginkgo/core/matrix/dense.hpp
```diff
@@ -1107,6 +1129,7 @@ class Dense
 private:
     Array<value_type> values_;
     size_type stride_;
+    strategy_type strategy_{strategy_type::gko};
```
Slaedr, line 1132: Are you sure we want to have only one strategy type for the entire class `Dense`? What if someone wants gko norm2 and vendor dot? Or we might add the option of vendor apply (once Vector is separated from Dense), and then that leads to even more options. Unless there's a better solution, I'd suggest
```suggestion
    strategy_type dot_strategy_{strategy_type::gko};
    strategy_type norm2_strategy_{strategy_type::gko};
```
and so on.

reference/matrix/dense_kernels.cpp
```diff
@@ -296,6 +296,26 @@ void compute_dot(std::shared_ptr<const ReferenceExecutor> exec,
 GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE(GKO_DECLARE_DENSE_COMPUTE_DOT_KERNEL);
 
 
+template <typename ValueType>
+void compute_dot_vendor(std::shared_ptr<const ReferenceExecutor> exec,
+                        const matrix::Dense<ValueType>* x,
+                        const matrix::Dense<ValueType>* y,
+                        matrix::Dense<ValueType>* result)
+{
+    for (size_type j = 0; j < x->get_size()[1]; ++j) {
```
Slaedr, line 305: Haha, this is funny. Can we just optionally link to Netlib blas and call that routine here instead? If we don't find a system Blas, we could leave this unimplemented. For testing, we already have `compute_dot` to compare against.

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -903,10 +922,12 @@ class Dense
...
     template <typename ValuesArray>
     Dense(std::shared_ptr<const Executor> exec, const dim<2>& size,
-          ValuesArray&& values, size_type stride)
+          ValuesArray&& values, size_type stride,
+          const strategy_type strategy = strategy_type::gko)
         : EnableLinOp<Dense>(exec, size),
           values_{exec, std::forward<ValuesArray>(values)},
-          stride_{stride}
+          stride_{stride},
+          strategy_(strategy)
```
Slaedr, line 930: If we accept my suggestion on multiple strategies, we could copy this argument to all strategies, and provide another constructor that takes all the individual strategies.

upsj:
What is the purpose of this strategy approach? From the user side, I mainly care about performance and correctness. Opposite to Csr, both of those are not really data-dependent, as the only data with performance impact are the matrix dimensions/stride.
If we want to use the vendor libraries for #rhs = 1, why don't we do it on the kernel side?
If this is about benchmarking, we can add hipBLAS/cuBLAS wrappers to the BLAS benchmark like we did for Csr.

Slaedr (changes requested):
The general idea makes sense. However, I think we, and users, may want to use different strategies (I guess they're really blas-backends or something) for different operations. See my comments below.

Slaedr:
@upsj You make a good point; we could go that way too. But if the idea is to give the user choice of blas backend, I'm for that as well. From time to time, we or the vendor might introduce improvements that might change the calculus for the user. We may not be able to or want to react to changes from the vendor side quickly enough.

cuda/matrix/dense_kernels.cu
```diff
@@ -104,7 +129,7 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE(GKO_DECLARE_DENSE_SIMPLE_APPLY_KERNEL);
 
 
 template <typename ValueType>
-void apply(std::shared_ptr<const CudaExecutor> exec,
+void apply(std::shared_ptr<const DefaultExecutor> exec,
```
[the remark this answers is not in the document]
pratikvn: `DefaultExecutor` makes it easier to copy the code when necessary  :)

hip/matrix/dense_kernels.hip.cpp
```diff
@@ -77,7 +77,80 @@ constexpr int default_block_size = 512;
...
+{
+    if (x->get_size()[1] == 1 && y->get_size()[1] == 1) {
+        if (hipblas::is_supported<ValueType>::value) {
+            auto handle = exec->get_hipblas_handle();
+            hipblas::pointer_mode_guard pm_guard(handle);
+            hipblas::dot(handle, x->get_size()[0], x->get_const_values(),
+                         x->get_stride(), y->get_const_values(),
+                         y->get_stride(), result->get_values());
+        } else {
+            GKO_NOT_IMPLEMENTED;
```
upsj, line 93: I don't think that's a good alternative, how about deferring to compute_dot here as well?

upsj (approved):
LGTM! Can you add your tuning results for the generic reductions here, as well?

pratikvn:
I only ran the tuning results with single vector and with dot product in mind. I did not really verify how that affects the other  kernels using this generic reduction, so I am a bit wary of changing the oversubscription directly in the code as of now. I will try to do a more comprehensive run and update that in another PR.

cuda/matrix/dense_kernels.cu
```diff
@@ -131,9 +129,8 @@ void compute_norm2_dispatch(std::shared_ptr<const DefaultExecutor> exec,
     if (x->get_size()[1] == 1) {
         if (cublas::is_supported<ValueType>::value) {
             auto handle = exec->get_cublas_handle();
-            cublas::pointer_mode_guard pm_guard(handle);
             cublas::norm2(handle, x->get_size()[0], x->get_const_values(),
-                          x->get_stride(), result->get_values());
+                          x->get_stride, result->get_values());
```
upsj, line 133: This is most likely wrong

pratikvn: Yikes! Sorry :)

dpcpp/matrix/dense_kernels.dp.cpp
```diff
@@ -190,7 +190,46 @@ GKO_ENABLE_DEFAULT_CONFIG_CALL(conj_transpose_call, conj_transpose,
 
 
 template <typename ValueType>
-void simple_apply(std::shared_ptr<const DpcppExecutor> exec,
+void compute_dot_dispatch(std::shared_ptr<const DefaultExecutor> exec,
+                          const matrix::Dense<ValueType>* x,
+                          const matrix::Dense<ValueType>* y,
+                          matrix::Dense<ValueType>* result)
+{
+    // TODO Add onemkl for single column ?
```
yhmtsai, line 198: could you also copy this TODO to another two?

yhmtsai (approved):
LGTM. could you also add the norm2 test for one vector?
There are 1/20 vectors test for dot and conj_dot but only 20 vectors test for norm2

PR #968, opened 2022-02-08
Batched add_scaled_identity operation

Slaedr:
Extends `M <- aI + bM` operation to the batched setting. Currently, `BatchDense` and `BatchCsr` are `BatchScaledIdentityAddable`.

Also fixes a bug with two-sided diagonal scaling in the case of rectangular matrices.


core/matrix/batch_csr.cpp
```diff
@@ -225,6 +228,24 @@ bool BatchCsr<ValueType, IndexType>::is_sorted_by_column_index() const
...
+    const BatchLinOp* const a, const BatchLinOp* const b)
+{
+    bool has_all_diags = false;
+    this->get_executor()->run(
+        batch_csr::make_check_diagonal_entries_exist(this, has_all_diags));
+    if (!has_all_diags) {
+        // TODO: Replace this with proper exception helper after merging
+        // non-batched add_scaled_identity PR
+        throw std::runtime_error("Matrix does not have all diagonal entries!");
+    }
```
pratikvn, line 242: Why do we need to check if the CSR matrix (that is being scaled) has all diagonals ? If some of its diagonals are zero, it is still a valid CSR matrix, right ?

pratikvn (commented):
A question about diagonal checking

core/matrix/batch_csr.cpp
Slaedr: Do you say it would be better to just introduce diagonal entries wherever they don't exist?

pratikvn: A matrix with some diagonal elements zero is still a valid matrix, which might also have a valid solution with some rhs. I think we should not throw an error if the matrix does not have some diagonal entries. 

From a scaling point of view as well, you should be able to still scale a matrix if it has some zero entries in the diagonal, right ?

upsj: If a diagonal entry is structurally zero in this add_scaled_**identity** operation, you either a) need to modify the sparsity pattern or b) leave a diagonal entry unmodified. a) is impossible if wrapping user data and b) leads to incorrect results

pratikvn: I would say a) could be a way to go. You are already over-writing the values array of the matrix, so the data (even if wrapped from the user), can have a modified sparsity pattern. But I guess modifying the sparsity pattern might not be suitable or expected by the user. 

In that case, I would maybe suggest have a flag if possible similar to our `is_sorted` flag, which the user can set if they are sure that the matrix has all its diagonals and then we can skip this check ? 

In any case, we should update the documentation to say that all diagonal entries are necessary for this operation and in case the matrix has a zero diagonal, an explicitly stored numerical zero is necessary for this operation to work.

Slaedr: The problem is we would need to reallocate to include new diagonal entries. That would break wrapping of user data, as Tobias pointed out. I think this comes down to having good matrix properties, or checks that can be disabled easily. For now, perhaps we skip this check and just document that we assume all diagonal entries exist?

Slaedr:
format!

PR #969, opened 2022-02-08
CSR: better strategy defaults.

tcojean:
Provide better CSR strategy defaults.

This duplicates two constructors to select a strategy based on the
executor type instead of defaulting to `automatical(CUDA)`. This should
not break any existing behavior (the `CUDA` case still defaults to
`automatical`), but instead, fix the currently broken combinations.

Background:
Some interesting situation just happened. Due to a seemingly broken
CUDA installation on a cluster, the CUDA API call in `get_num_devices()`
can fail, for example, due to an insufficient driver version or something
else:
```
terminate called after throwing an instance of 'gko::CudaError'
  what():  ../cuda/base/executor.cpp:223: get_num_devices: cudaErrorInsufficientDriver: CUDA driver version is insufficient for CUDA runtime version
```

Since the current CSR strategy is to instantiate a `CudaExecutor`, this
means that in that case all of Ginkgo is broken at runtime, even when
not using an actual `CudaExecutor`. The aim of this PR is to change this
long-time problem with the default strategy type.

### DIFF
--- a/include/ginkgo/core/matrix/csr.hpp
+++ b/include/ginkgo/core/matrix/csr.hpp
@@ -963,7 +963,7 @@ public:
         gko::detail::ConstArrayView<ValueType>&& values,
         gko::detail::ConstArrayView<IndexType>&& col_idxs,
         gko::detail::ConstArrayView<IndexType>&& row_ptrs,
-        std::shared_ptr<strategy_type> strategy = std::make_shared<sparselib>())
+        std::shared_ptr<strategy_type> strategy)
     {
         // cast const-ness away, but return a const object afterwards,
         // so we can ensure that no modifications take place.
@@ -971,6 +971,23 @@ public:
             exec, size, gko::detail::array_const_cast(std::move(values)),
             gko::detail::array_const_cast(std::move(col_idxs)),
             gko::detail::array_const_cast(std::move(row_ptrs)), strategy});
+    }
+
+    /*
+     * This is version of create_const with a default strategy.
+     */
+    static std::unique_ptr<const Csr> create_const(
+        std::shared_ptr<const Executor> exec, const dim<2>& size,
+        gko::detail::ConstArrayView<ValueType>&& values,
+        gko::detail::ConstArrayView<IndexType>&& col_idxs,
+        gko::detail::ConstArrayView<IndexType>&& row_ptrs)
+    {
+        // cast const-ness away, but return a const object afterwards,
+        // so we can ensure that no modifications take place.
+        return std::unique_ptr<const Csr>(new Csr{
+            exec, size, gko::detail::array_const_cast(std::move(values)),
+            gko::detail::array_const_cast(std::move(col_idxs)),
+            gko::detail::array_const_cast(std::move(row_ptrs))});
     }
 
 protected:
@@ -986,16 +1003,16 @@ protected:
     {}
 
     /**
-     * Creates an uninitialized CSR matrix of the specified size.
+     * Creates an uninitialized CSR matrix of the specified size with a user
+     * chosen strategy.
      *
      * @param exec  Executor associated to the matrix
      * @param size  size of the matrix
      * @param num_nonzeros  number of nonzeros
      * @param strategy  the strategy of CSR
      */
-    Csr(std::shared_ptr<const Executor> exec, const dim<2>& size = dim<2>{},
-        size_type num_nonzeros = {},
-        std::shared_ptr<strategy_type> strategy = std::make_shared<sparselib>())
+    Csr(std::shared_ptr<const Executor> exec, const dim<2>& size,
+        size_type num_nonzeros, std::shared_ptr<strategy_type> strategy)
         : EnableLinOp<Csr>(exec, size),
           values_(exec, num_nonzeros),
           col_idxs_(exec, num_nonzeros),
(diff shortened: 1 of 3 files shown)


tcojean:
Looks like this wouldn't be an issue with a real develop codebase, which has `sparselib` as a strategy everywhere. Due to relying on many commits, the OpenCARP branch still reintroduced `automatical` as a strategy.

I guess this can still be a suggestion for dynamically adapting the strategy to executor type.

upsj:
Yes, this kind of thing has commonly lead to issues in Csr automatic tests if you forgot to pass the HIP parameter and relied on the default instead.

include/ginkgo/core/matrix/csr.hpp
```diff
@@ -1042,11 +1079,50 @@ class Csr : public EnableLinOp<Csr<ValueType, IndexType>>,
...
+        : EnableLinOp<Csr>(exec, size),
+          values_{exec, std::forward<ValuesArray>(values)},
+          col_idxs_{exec, std::forward<ColIdxsArray>(col_idxs)},
+          row_ptrs_{exec, std::forward<RowPtrsArray>(row_ptrs)}
+    {
+        GKO_ASSERT_EQ(values_.get_num_elems(), col_idxs_.get_num_elems());
+        GKO_ASSERT_EQ(this->get_size()[0] + 1, row_ptrs_.get_num_elems());
+        initialize_default_strategy(row_ptrs_.get_num_elems());
+        this->make_srow();
+    }
```
upsj, line 1102: how about this?
You only need the number of nonzeros to call clac_size, but that is only necessary if you initialize the array yourself, otherwise that would be handled by make_srow calling srow_.resize_and_reset
```suggestion
    : Csr{exec, size, std::forward<ValuesArray>(values), std::forward<ColIdxsArray>(col_idxs), std::forward<RowPtrsArray>(row_ptrs), make_default_strategy(exec)) {}
```

include/ginkgo/core/matrix/csr.hpp
```diff
@@ -1005,6 +1022,26 @@ class Csr : public EnableLinOp<Csr<ValueType, IndexType>>,
...
+     * @param exec  Executor associated to the matrix
+     * @param size  size of the matrix
+     * @param num_nonzeros  number of nonzeros
+     */
+    Csr(std::shared_ptr<const Executor> exec, const dim<2>& size = dim<2>{},
+        size_type num_nonzeros = {})
+        : EnableLinOp<Csr>(exec, size),
+          values_(exec, num_nonzeros),
+          col_idxs_(exec, num_nonzeros),
+          row_ptrs_(exec, size[0] + 1)
```
upsj, line 1038: same here: this could be handled by a call to the constructor with strategy

include/ginkgo/core/matrix/csr.hpp
```diff
@@ -971,6 +971,23 @@ class Csr : public EnableLinOp<Csr<ValueType, IndexType>>,
...
+        gko::detail::ConstArrayView<IndexType>&& col_idxs,
+        gko::detail::ConstArrayView<IndexType>&& row_ptrs)
+    {
+        // cast const-ness away, but return a const object afterwards,
+        // so we can ensure that no modifications take place.
+        return std::unique_ptr<const Csr>(new Csr{
+            exec, size, gko::detail::array_const_cast(std::move(values)),
+            gko::detail::array_const_cast(std::move(col_idxs)),
+            gko::detail::array_const_cast(std::move(row_ptrs))});
+    }
```
upsj, line 989: same here, can be represented as a call to create_const with strategy parameter make_default_strategy(exec). AFAIK nnz doesn't need to be known to initialize the strategy.

upsj (approved):
LGTM! Only minor suggestions

tcojean:
Pinging @yhmtsai I see that this was reintroduced (on my side) in the `improve_multigrid` branch. It makes sense to default to `automatical` in the correct settings for better performance. Does this way of doing it look good? Any different default settings?

pratikvn (approved):
LGTM!

yhmtsai (approved):
default setting as automatic is okay to me.
do you think we should change the load_balance/automatic default constructor to throw error not use cuda executor?

include/ginkgo/core/matrix/csr.hpp
```diff
@@ -1042,11 +1070,47 @@ class Csr : public EnableLinOp<Csr<ValueType, IndexType>>,
...
 
     void apply_impl(const LinOp* alpha, const LinOp* b, const LinOp* beta,
                     LinOp* x) const override;
 
+    // TODO: This provides some more sane settings. Please fix this!
+    static std::shared_ptr<strategy_type> make_default_strategy(
+        std::shared_ptr<const Executor> exec)
+    {
+        auto cuda_exec = std::dynamic_pointer_cast<const CudaExecutor>(exec);
+        auto hip_exec = std::dynamic_pointer_cast<const HipExecutor>(exec);
```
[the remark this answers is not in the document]
tcojean: I thought of not doing the dpcpp cast since it would default to classical. I guess I can add it.

I'm not sure if we should throw in the load_balance/automatic default executor. I think with this version indeed it should basically never be used anymore, so we can start depreciating it? It only depends if some user somewhere had a `std::make_shared<automatical>()`.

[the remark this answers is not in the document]
yhmtsai: depreciating is also good.

upsj (approved):

include/ginkgo/core/matrix/csr.hpp
```diff
@@ -1042,11 +1076,50 @@ class Csr : public EnableLinOp<Csr<ValueType, IndexType>>,
...
+        auto dpcpp_exec = std::dynamic_pointer_cast<const DpcppExecutor>(exec);
+        std::shared_ptr<strategy_type> new_strategy;
+        if (cuda_exec) {
+            new_strategy = std::make_shared<automatical>(cuda_exec);
+        } else if (hip_exec) {
+            new_strategy = std::make_shared<automatical>(hip_exec);
+        } else if (dpcpp_exec) {
+            new_strategy = std::make_shared<classical>();
```
yhmtsai, line 1116: should it be automatical or classical?

tcojean: I think it depends on your experience. If I remember, you said classical was almost always faster? Or are there some cases where load_balance was useful as well?

yhmtsai: load_balance has pretty limited benefit on intel DG1.
automatical does the same thing, which almost always chooses classical from the condition.
when it is on device executor, we use automatical by default. (dpcpp tend to choose classical in automatical strategy)

tcojean: So you want the decision to be left to automatical. I will use it then. As a side comment, I think it's fine to change the default strategies now that we have this, if we think it provides better performance then there shouldn't be a problem (e.g. switching to sparse lib or another).

PR #970, opened 2022-02-11
Sparsity csr spmv

yhmtsai:
This PR copies the classical csr spmv almost to sparsity csr
only limited subwarp/subgroup support because amgx_pgm with aggregation 2 usually generate 1 ~ 3 nnz per row

TODO:
- [ ] add different subwarp/subgroup support? (not in this pr. when we have more general usage on it, we can add the selection again)
- [x] move the test to matrix test
- [x] fix multigrid test


hip/test/matrix/sparsity_csr_kernels.hip.cpp
```diff
@@ -0,0 +1,167 @@
...
+        dalpha = gko::clone(hip, alpha);
+        dbeta = gko::clone(hip, beta);
+    }
+
+
+    std::shared_ptr<gko::ReferenceExecutor> ref;
+    std::shared_ptr<const gko::HipExecutor> hip;
+
+    const gko::dim<2> mtx_size;
+    std::ranlux48 rand_engine;
```
upsj, line 107: default_random_engine :)

yhmtsai:
I agree with you. we can use arithmetic_type for the operations.
Also, we have different way to handle the mixed precision and it's might be hard or expensive to keep the same precision with GINKGO_MIXED_PRECISION=OFF

test/matrix/matrix.cpp
```diff
@@ -348,6 +354,19 @@ struct HybridAutomaticStrategy
...
 
 
+struct SparsityCsr
+    : SimpleMatrixTest<gko::matrix::SparsityCsr<matrix_value_type, int>> {
+    static void modify_data(gko::matrix_data<matrix_value_type, int>& data)
+    {
+        using entry_type =
+            gko::matrix_data<matrix_value_type, int>::nonzero_type;
+        for (auto& entry : data.nonzeros) {
+            entry = entry_type{entry.row, entry.column, matrix_value_type{1}};
```
upsj, line 364: entry.value = gko::one<matrix_value_type>();

cuda/matrix/sparsity_csr_kernels.cu
```diff
@@ -113,19 +113,30 @@ void classical_spmv(syn::value_list<int, subwarp_size>,
         c->get_values(),
         std::array<acc::size_type, 1>{
             {static_cast<acc::size_type>(c->get_stride())}});
-
+    if (c->get_size()[0] == 0 || c->get_size()[1] == 0) {
+        // empty output: nothing to do
+        return;
+    }
```
upsj, line 119: Is this necessary with the guards below?

reference/matrix/sparsity_csr_kernels.cpp
```diff
@@ -59,46 +61,55 @@ namespace reference {
...
 {
+    using arithmetic_type =
+        highest_precision<InputValueType, OutputValueType, MatrixValueType>;
     auto row_ptrs = a->get_const_row_ptrs();
     auto col_idxs = a->get_const_col_idxs();
-    auto val = a->get_const_value()[0];
+    const auto val = static_cast<arithmetic_type>(a->get_const_value()[0]);
 
     for (size_type row = 0; row < a->get_size()[0]; ++row) {
         for (size_type j = 0; j < c->get_size()[1]; ++j) {
```
upsj, line 78: I would suggest swapping the outer and inner loop here, this enables keeping the partial sum in an `arithmetic_type` variable, giving better precision

omp/matrix/sparsity_csr_kernels.cpp
```diff
@@ -62,47 +64,51 @@ namespace omp {
...
+          matrix::Dense<OutputValueType>* c)
 {
     auto row_ptrs = a->get_const_row_ptrs();
     auto col_idxs = a->get_const_col_idxs();
-    auto val = a->get_const_value()[0];
+    const auto val = static_cast<OutputValueType>(a->get_const_value()[0]);
 
 #pragma omp parallel for
     for (size_type row = 0; row < a->get_size()[0]; ++row) {
         for (size_type j = 0; j < c->get_size()[1]; ++j) {
```
upsj, line 80: Same here, maybe consider block-unrolling like with ELL etc.

upsj (approved):
LGTM! Nice job keeping the higher precision consistent on the GPU. I think we only need a bit more scrutiny on the CPU side

omp/matrix/sparsity_csr_kernels.cpp
upsj: Maybe that is something for another PR though, together with Csr SpMV

cuda/matrix/sparsity_csr_kernels.cu
yhmtsai: right, they are the same thing. I will still keep the check in classical csr spmv because the checks are in different function and it's more complex

pratikvn (approved):
Nice work! LGTM!

PR #971, opened 2022-02-14
Adds support for MPI distributed matrix

MarcelKoch:
This PR adds a row-wise distributed matrix. Each rank owns a number of rows of the global matrix, as defined in a partition. Locally, the rows are split into a diagonal part which contains only entries with column indices that belong to the local row range.
The remaining entries are gathered into an offdiagonal part. The offdiagonal columns are renumbered such that empty columns are removed. This renumbering is stored in the matrix, although there is no access to it at the moment.

Current limitations:
- only reference + openMP read kernels
- only square global matrices are allowed
- can't apply to vectors with multiple columns

Partially addresses #907


include/ginkgo/core/base/mpi.hpp
```diff
@@ -88,13 +88,53 @@ GKO_REGISTER_MPI_TYPE(int, MPI_INT);
...
 GKO_REGISTER_MPI_TYPE(long, MPI_LONG);
+GKO_REGISTER_MPI_TYPE(long long, MPI_LONG_LONG_INT);
 GKO_REGISTER_MPI_TYPE(float, MPI_FLOAT);
 GKO_REGISTER_MPI_TYPE(double, MPI_DOUBLE);
 GKO_REGISTER_MPI_TYPE(long double, MPI_LONG_DOUBLE);
 GKO_REGISTER_MPI_TYPE(std::complex<float>, MPI_C_COMPLEX);
 GKO_REGISTER_MPI_TYPE(std::complex<double>, MPI_C_DOUBLE_COMPLEX);
 
 
+class contiguous_type {
```
upsj, line 99: nice :)

upsj (commented):
First half of my review, mostly interface comments

omp/distributed/matrix_kernels.cpp
```diff
@@ -0,0 +1,248 @@
...
+void build_diag_offdiag(
+    std::shared_ptr<const DefaultExecutor> exec,
+    const device_matrix_data<ValueType, GlobalIndexType>& input,
+    const distributed::Partition<LocalIndexType, GlobalIndexType>* partition,
+    comm_index_type local_part,
+    device_matrix_data<ValueType, LocalIndexType>& diag_data,
+    device_matrix_data<ValueType, LocalIndexType>& offdiag_data,
+    Array<LocalIndexType>& local_gather_idxs, comm_index_type* recv_offsets,
+    Array<GlobalIndexType>& local_to_global_ghost)
+{
```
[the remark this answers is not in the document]
MarcelKoch: Yes, I've fixed the issues.

include/ginkgo/core/base/device_matrix_data.hpp
```diff
@@ -234,7 +234,12 @@ class device_matrix_data {
      *
      * @param new_num_entries  the new number of stored matrix entries.
      */
-    void resize_and_reset(size_type new_num_entries);
+    void resize_and_reset(size_type new_num_entries)
+    {
+        row_idxs_.resize_and_reset(new_num_entries);
+        col_idxs_.resize_and_reset(new_num_entries);
+        values_.resize_and_reset(new_num_entries);
+    }
```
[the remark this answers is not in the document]
MarcelKoch: So the issue is that I also need to set the dimension correctly, esp. for the offdiag. I think I can do that without calling the `resize_and_reset`, but it might be a bit cumbersome.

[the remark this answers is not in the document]
upsj: The issue here is one of circular dependencies: core depends on device-specific libraries, so device-specific libraries can't call core functionality. This should be handled on the core level IMO. Pulling more functions into headers makes things difficult to untangle, also w.r.t. #715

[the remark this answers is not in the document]
MarcelKoch: I generally agree with having a cleaner separation of .ccp/.hpp. Also my previous comment was wrong, and it's actually pretty straight forward to remove that call.

reference/distributed/matrix_kernels.cpp
```diff
@@ -0,0 +1,194 @@
...
+                   });
+
+    // build recv_offsets
+    components::prefix_sum(exec, recv_offsets, num_parts + 1);
+    const auto num_ghost_columns =
+        static_cast<size_type>(recv_offsets[num_parts]);
+    local_gather_idxs.resize_and_reset(num_ghost_columns);
+    std::unordered_map<GlobalIndexType, LocalIndexType> offdiag_global_to_local;
+    // collect and renumber offdiagonal columns
+    for (auto entry : offdiag_cols) {
```
pratikvn, line 146: nit:
```suggestion
    for (const auto& entry : offdiag_cols) {
```

pratikvn (commented):
First part of the review

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -0,0 +1,197 @@
...
+    using EnableLinOp<Matrix>::move_to;
+
+    void convert_to(Matrix<next_precision<value_type>, local_index_type,
+                           global_index_type>* result) const override;
+
+    void move_to(Matrix<next_precision<value_type>, local_index_type,
+                        global_index_type>* result) override;
+
+    /**
+     * Reads a square matrix from the matrix_data structure and a global row
```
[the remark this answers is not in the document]
upsj: The matrix storage only works under the assumption that input and output vector have the same partition, otherwise you need to use two different partition objects for row and column indices. This should *theoretically* be possible to handle, but requires quite a few modifications and a separate read_distributed(data, row_partition, col_partition), so I think we should not add that yet. But based on that, maybe `diag` and `offdiag` should instead be named something like `local` and `non_local` (there are probably better names for that :D)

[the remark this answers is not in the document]
MarcelKoch: TBH, I think adding that will be straight forward. The changes to read_distributed and its kernel are relatively minor, so I will add it in this PR.

reference/test/distributed/matrix_kernels.cpp
```diff
@@ -0,0 +1,370 @@
...
+    gko::Array<global_index_type> result[num_parts] = {
+        {this->ref, {3}}, {this->ref, {0, 6}}, {this->ref, {4}}};
+
+    for (int local_id = 0; local_id < num_parts; ++local_id) {
+        gko::kernels::reference::distributed_matrix::build_diag_offdiag(
+            this->ref, this->create_input_full_rank(), partition.get(),
+            local_id, this->diag, this->offdiag, this->gather_idxs,
+            this->recv_offsets.get_data(), this->local_to_global_ghost);
+
+        GKO_ASSERT_ARRAY_EQ(result[local_id], this->local_to_global_ghost);
```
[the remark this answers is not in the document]
MarcelKoch: The other outputs are already tested in the other tests. I was just to lazy to add `local_to_global_ghost` to the other tests.

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -0,0 +1,232 @@
...
+     *
+     * Both the number of rows and columns of the device_matrix_data is ignored.
+     * The global size of the final matrix is inferred only from the size row
+     * partition.
+     *
+     * @note The matrix data can contain entries for rows other than those owned
+     *        by the process. Entries for those rows are discarded.
+     *
+     * @param data  The device_matrix_data structure.
+     * @param partition  The global row and column partition.
```
upsj, line 105: maybe we should get the term `symmetric` in there somehow?

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -0,0 +1,232 @@
...
+namespace distributed {
+
+
+template <typename ValueType>
+class Vector;
+
+
+template <typename ValueType = double, typename LocalIndexType = int32,
+          typename GlobalIndexType = int64>
+class Matrix
```
upsj, line 61: Needs documentation

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -0,0 +1,232 @@
...
+                        GlobalIndexType>;
+
+public:
+    using value_type = ValueType;
+    using index_type = GlobalIndexType;
+    using local_index_type = LocalIndexType;
+    using global_index_type = GlobalIndexType;
+    using global_vector_type = gko::distributed::Vector<ValueType>;
+    using local_vector_type = typename global_vector_type::local_vector_type;
+    using local_matrix_type = gko::matrix::Csr<value_type, local_index_type>;
```
upsj, line 82: one thing I am not sure about is how to deal with this parameter. In theory, the entire setup is flexible enough to allow any matrix format to be used, we only need to figure out how to let users specify the matrix type, maybe by passing in a prototype?

omp/test/distributed/matrix_kernels.cpp
```diff
@@ -0,0 +1,368 @@
...
+
+
+TYPED_TEST(Matrix, BuildsDiagOffdiagEmptyIsSameAsRef)
+{
+    using value_type = typename TestFixture::value_type;
+    using local_index_type = typename TestFixture::local_index_type;
+    using global_index_type = typename TestFixture::global_index_type;
+    gko::Array<comm_index_type> mapping{this->ref, {1, 0, 2, 2, 0, 1, 1, 2}};
+    comm_index_type num_parts = 3;
+    auto partition = gko::distributed::Partition<
```
upsj, line 154: nit: Arrange, Act, Assert in three separate blocks
```suggestion

    auto partition = gko::distributed::Partition<
```

reference/test/distributed/matrix_kernels.cpp
```diff
@@ -0,0 +1,553 @@
...
+    this->mapping = {this->ref, {1, 2, 0, 0, 2, 1}};
+    comm_index_type num_parts = 3;
+    auto partition = gko::distributed::Partition<lit, git>::build_from_mapping(
+        this->ref, this->mapping, num_parts);
+
+    this->validate(
+        gko::dim<2>{6, 6}, partition.get(), partition.get(),
+        {0, 0, 1, 1, 2, 3, 4, 5}, {0, 5, 1, 4, 3, 2, 4, 0},
+        {1, 2, 3, 4, 5, 6, 7, 8},
+        {std::make_tuple(gko::dim<2>{2, 2}, I<git>{0, 1}, I<git>{1, 0},
```
upsj, line 252: the German in me had to chuckle at `I<git>` 😄

upsj (commented):
Really nice, I like how it's come along. I would probably give this a second pass after updates before approving it, but most things are minor

core/distributed/matrix_kernels.hpp
```diff
@@ -0,0 +1,86 @@
...
+    void build_diag_offdiag(                                           \
+        std::shared_ptr<const DefaultExecutor> exec,                   \
+        const device_matrix_data<ValueType, GlobalIndexType>& input,   \
+        const distributed::Partition<LocalIndexType, GlobalIndexType>* \
+            row_partition,                                             \
+        const distributed::Partition<LocalIndexType, GlobalIndexType>* \
+            col_partition,                                             \
+        comm_index_type local_part,                                    \
+        device_matrix_data<ValueType, LocalIndexType>& diag_data,      \
+        device_matrix_data<ValueType, LocalIndexType>& offdiag_data,   \
```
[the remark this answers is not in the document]
MarcelKoch: I need to set the size/dim of the data, which at least for the offdiag case is only known during the kernel.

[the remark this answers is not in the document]
upsj: good point. I am really not happy with mixing `core` and kernel code in that fashion, an alternative would be a single out-parameter for the number of offdiag columns?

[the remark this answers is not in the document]
MarcelKoch: So I guess the only core code that is acceptable is array code, right? I would still need either the resize or the assign operator.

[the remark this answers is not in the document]
upsj: yes. At some point in the future (probably 2.0), we will need to split everything into a core-core (probably executor-only stuff + header-only types like Array that depend directly on it) and regular core (all other Ginkgo types) so that core and kernels depend on core-core, and core depends on kernels.

omp/distributed/matrix_kernels.cpp
```diff
@@ -0,0 +1,283 @@
...
+
+
+namespace gko {
+namespace kernels {
+namespace omp {
+namespace distributed_matrix {
+
+
+template <typename ValueType, typename LocalIndexType, typename GlobalIndexType>
+void build_diag_offdiag(
```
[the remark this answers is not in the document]
MarcelKoch: Yes, but perhaps don't spend too much time on that. I guess I can simplify it after I worked on the device kernels, since they also need to be simpler.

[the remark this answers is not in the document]
upsj: fair point. I guess I would like to see all allocations being tracked here, but we've had some bad experiences with OpenMP and executors before (https://github.com/ginkgo-project/ginkgo/pull/740, https://github.com/ginkgo-project/ginkgo/pull/565), so I was looking for a way to do that more cleanly. But it's probably not worth it.

core/distributed/matrix_kernels.hpp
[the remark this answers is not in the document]
MarcelKoch: perhaps to clarify what I'm doing, I create a new device matrix data with the correct size and move that into the input data. So the only implementation that I need here is the device_matrix_data constructor that takes the arrays as parameter. Do you expect that we also move that constructor into core?

include/ginkgo/core/distributed/matrix.hpp
MarcelKoch: The matrix itself is not necessarily symmetric, so I don't think we should add it here. It would probably be more confusing.

core/distributed/matrix_kernels.hpp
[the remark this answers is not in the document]
upsj: hmm, that's difficult to judge. I would prefer if we kept them separate (especially because in Windows, all members of a class need to be part of the same DLL unless you tag each one individually with different __dllspec attributes, which misses some special member functions etc.), but that would mean that the constructor can't be called from the kernel.

include/ginkgo/core/distributed/matrix.hpp
upsj: I was talking about the partition being symmetric, which is the term I recall seeing most of the time associated with this kind of partition.

pratikvn:
One more thing I would like to bring up and something we might want to address is different ways of creating a distributed matrix. To be specific, the auxiliary vectors such as gather_idxs, recv_idxs etc. Currently, these are setup only during one of the read_distributed calls. 

I think it would be very useful to either have kernels or member functions to `distributed::Matrix` which setup this data from the locally stored matrix (row_ptrs, col_idxs, values).

upsj:
@pratikvn I fully agree, though this could (maybe even should) be done in a subsequent PR. We definitely need to take good care to clearly describe how this data can be set up etc, so this should only be used by *very* advanced users

pratikvn:
@upsj, Yes, I agree, we should probably tackle that in another PR, but I wanted to make a note of it, because it would be very useful, more so for our own Ginkgo internal projects than probably for external users. If anyone wants to interface distributed-ginkgo though, I think that will be essential.

MarcelKoch:
format!

core/distributed/matrix.cpp
```diff
@@ -0,0 +1,431 @@
...
+        recv_gather_idxs.set_executor(exec->get_master());
+        gather_idxs_.clear();
+        gather_idxs_.set_executor(exec->get_master());
+    }
+    gather_idxs_.resize_and_reset(send_offsets_.back());
+    comm.all_to_all_v(recv_gather_idxs.get_const_data(), recv_sizes_.data(),
+                      recv_offsets_.data(), gather_idxs_.get_data(),
+                      send_sizes_.data(), send_offsets_.data());
+    if (needs_host_buffer) {
+        gather_idxs_.set_executor(exec);
```
pratikvn, line 209: Can we extract this into an function ? I think it would be very useful to setup these communication structures in cases where we setup the matrix not from a read_distributed interface, for example directly from one of the `Matrix` constructors

MarcelKoch: I'm not sure I would take this step at the moment. When we are revisiting different modes to create a distributed matrix, it would make sense to refactor this. But at the moment I'm don't know how the refactor would look like. So I would leave it as it is.

core/distributed/matrix.cpp
```diff
@@ -0,0 +1,429 @@
...
+    // exchange step 2: exchange gather_idxs from receivers to senders
+    auto needs_host_buffer =
+        exec->get_master() != exec && !gko::mpi::is_gpu_aware();
+    if (needs_host_buffer) {
+        recv_gather_idxs.set_executor(exec->get_master());
+        gather_idxs_.clear();
+        gather_idxs_.set_executor(exec->get_master());
+    }
+    gather_idxs_.resize_and_reset(send_offsets_.back());
+    comm.all_to_all_v(recv_gather_idxs.get_const_data(), recv_sizes_.data(),
```
yhmtsai, line 205: maybe let all_to_all_v to handle the gpu_aware?

MarcelKoch: This would require that we pass in an executor to the call, or the constructor of the communicator. At the moment, I wouldn't go that way, especially since that would complicate handling non-blocking calls quite a lot. Perhaps if we add the device guard to the mpi calls, then it makes sense to revisit this.

core/distributed/matrix.cpp
```diff
@@ -0,0 +1,429 @@
...
+                      recv_offsets_.data(), gather_idxs_.get_data(),
+                      send_sizes_.data(), send_offsets_.data());
+    if (needs_host_buffer) {
+        gather_idxs_.set_executor(exec);
```
tcojean, line 209: What about `recv_gather_idxs`? Why isn't it moved back as well?

tcojean (approved):
LGTM

include/ginkgo/core/base/mpi.hpp
```diff
@@ -88,13 +94,53 @@ GKO_REGISTER_MPI_TYPE(int, MPI_INT);
...
+    }
+
+    MPI_Datatype get() { return type_; }
+
+private:
+    MPI_Datatype type_;
+};
+
+
 template <typename T>
```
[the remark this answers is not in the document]
MarcelKoch: I've replaced this function with just using `MPI_IN_PLACE` directly. I don't know why we've introduced this in the first place.

core/distributed/matrix.cpp
```diff
@@ -0,0 +1,429 @@
...
+    std::shared_ptr<const Executor> exec, mpi::communicator comm)
+    : Matrix(exec, comm, with_matrix_type<gko::matrix::Csr>())
+{}
+
+
+template <typename ValueType, typename LocalIndexType, typename GlobalIndexType>
+Matrix<ValueType, LocalIndexType, GlobalIndexType>::Matrix(
+    std::shared_ptr<const Executor> exec, mpi::communicator comm,
+    const LinOp* inner_matrix_type)
+    : Matrix(exec, comm, inner_matrix_type, inner_matrix_type)
```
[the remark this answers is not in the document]
MarcelKoch: I think this is related to #715, and I've briefly discussed at other points with @upsj what should be in .hpp and what not. For this type, I think it makes sense to move everything into the .cpp. Constructing a matrix (even copying) seems not like a common use-case for the device modules. I will also move the implementations that are currently in the .hpp into the .cpp.

core/distributed/matrix.cpp
MarcelKoch: `recv_gather_idxs` is just a temporary to compute `gather_idxs` and are not used afterwards.

omp/distributed/matrix_kernels.cpp
```diff
@@ -0,0 +1,266 @@
...
+    using global_nonzero = matrix_data_entry<ValueType, GlobalIndexType>;
+    using local_nonzero = matrix_data_entry<ValueType, LocalIndexType>;
+    auto input_row_idxs = input.get_const_row_idxs();
+    auto input_col_idxs = input.get_const_col_idxs();
+    auto input_vals = input.get_const_values();
+    auto row_part_ids = row_partition->get_part_ids();
+    auto col_part_ids = col_partition->get_part_ids();
+    auto num_parts = row_partition->get_num_parts();
+    size_type row_range_id_hint = 0;
+    size_type col_range_id_hint = 0;
```
upsj, line 80: I think just to be sure, these should be firstprivate

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -0,0 +1,522 @@
...
+struct is_matrix_builder : std::false_type {};
+
+template <typename T>
+struct is_matrix_builder<
+    T, xstd::void_t<decltype(std::declval<T>().template create<double, int>(
+           std::declval<std::shared_ptr<const Executor>>()))>>
+    : std::true_type {};
+
+
+}  // namespace detail
```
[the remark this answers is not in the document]
MarcelKoch: Since the latter one does not repeat the value and index type, I think it is worth it. Also, I will remove the `is_matrix_builder` stuff, since I didn't end up using it.

core/distributed/matrix.cpp
```diff
@@ -0,0 +1,440 @@
...
+        offdiag_mtx_->apply(one_scalar_.get(), recv_buffer_.get(),
+                            one_scalar_.get(), local_x.get());
+    } else {
+        diag_mtx_->apply(dense_b->get_local_vector(), local_x.get());
+    }
+}
+
+
+template <typename ValueType, typename LocalIndexType, typename GlobalIndexType>
+void Matrix<ValueType, LocalIndexType, GlobalIndexType>::apply_impl(
```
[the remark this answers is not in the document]
MarcelKoch: Yes, it's added in the solver PR, but I can also move that here.

core/distributed/matrix.cpp
```diff
@@ -0,0 +1,440 @@
...
+      DistributedBase{comm},
+      send_offsets_(comm.size() + 1),
+      send_sizes_(comm.size()),
+      recv_offsets_(comm.size() + 1),
+      recv_sizes_(comm.size()),
+      gather_idxs_{exec},
+      local_to_global_ghost_{exec},
+      one_scalar_{},
+      diag_mtx_{inner_matrix_type->clone(exec)},
+      offdiag_mtx_{ghost_matrix_type->clone(exec)}
```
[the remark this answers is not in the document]
MarcelKoch: Perhaps local/neighbor? But I don't think there is a nice application agnostic choice for that out there.

PS: I like inner/ghost because both are the same length.

include/ginkgo/core/distributed/matrix.hpp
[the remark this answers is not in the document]
upsj: The main issue I see with it is that it introduces another unfamiliar idiom and an associated free function that is only used in this one constructor. The difference in amount and complexity of code you need to write is marginal. But as I said above, I don't want to hold the PR up over this, especially since the effort was already spent :)

include/ginkgo/core/base/mpi.hpp
```diff
@@ -88,18 +102,93 @@ GKO_REGISTER_MPI_TYPE(int, MPI_INT);
...
+
+    /**
+     * Disallow copying of wrapper type.
+     */
+    contiguous_type& operator=(const contiguous_type&) = delete;
+
+    /**
+     * Move constructor, leaves other with MPI_DATATYPE_NULL.
+     *
+     * @param other  the to be moved from object.
```
yhmtsai, line 151:
```suggestion
     * @param other  to be moved from object `other`.
```

include/ginkgo/core/base/mpi.hpp
```diff
@@ -88,18 +102,93 @@ GKO_REGISTER_MPI_TYPE(int, MPI_INT);
...
+     */
+    contiguous_type(contiguous_type&& other) noexcept : type_(MPI_DATATYPE_NULL)
+    {
+        *this = std::move(other);
+    }
+
+    /**
+     * Move assignment, leaves other with MPI_DATATYPE_NULL.
+     *
+     * @param other  the to be moved from object.
```
yhmtsai, line 161: same

include/ginkgo/core/base/mpi.hpp
```diff
@@ -446,6 +543,10 @@ class communicator {
      * @param source_rank  the rank to receive the data from
      * @param recv_tag  the tag for the recv call
      *
+     * @tparam RecvType  the type of the data to send. Has to be a type which
```
yhmtsai, line 546:
```suggestion
     * @tparam RecvType  the type of the data to receive. Has to be a type which
```

include/ginkgo/core/base/mpi.hpp
```diff
@@ -467,6 +568,10 @@ class communicator {
...
```
yhmtsai, line 571:
```suggestion
     * @tparam RecvType  the type of the data to receive. Has to be a type which
```

include/ginkgo/core/base/mpi.hpp
```diff
@@ -486,6 +591,10 @@ class communicator {
      * @param buffer  the buffer to broadcsat
      * @param count  the number of elements to broadcast
      * @param root_rank  the rank to broadcast from
+     *
+     * @tparam BroadcastType  the type of the data to send. Has to be a type
+     *                        which has a specialization of type_impl that
+     * defines its MPI_Datatype.
```
yhmtsai, line 597:
```suggestion
     *                        which has a specialization of type_impl that
     *                        defines its MPI_Datatype.
```

include/ginkgo/core/base/mpi.hpp
```diff
@@ -503,6 +612,10 @@ class communicator {
      * @param count  the number of elements to broadcast
      * @param root_rank  the rank to broadcast from
      *
+     * @tparam BroadcastType  the type of the data to send. Has to be a type
+     *                        which has a specialization of type_impl that
+     * defines its MPI_Datatype.
```
yhmtsai, line 617:
```suggestion
     *                        which has a specialization of type_impl that
     *                        defines its MPI_Datatype.
```

yhmtsai (commented):
I will finish reviewing the left part later

yhmtsai (changes requested):
there's undefined behavior in uniform_int_distribution(a, b) when a > b

omp/test/distributed/matrix_kernels.cpp
```diff
@@ -0,0 +1,384 @@
...
+        gko::test::generate_random_array<gko::distributed::comm_index_type>(
+            num_rows,
+            std::uniform_int_distribution<gko::distributed::comm_index_type>(
+                0, num_parts - 1),
+            this->engine, this->ref);
+    auto input = gko::test::generate_random_device_matrix_data<
+        value_type, global_index_type>(
+        num_rows, num_cols,
+        std::uniform_int_distribution<int>(static_cast<int>(num_cols),
+                                           static_cast<int>(num_cols - 1)),
```
upsj, line 233: what Mike says: This doesn't look right

include/ginkgo/core/base/mpi.hpp
MarcelKoch: would the following be acceptable?
```suggestion
     * @param other  to be moved from object.
```

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -0,0 +1,533 @@
...
+
+    void convert_to(Matrix<next_precision<value_type>, local_index_type,
+                           global_index_type>* result) const override;
+
+    void move_to(Matrix<next_precision<value_type>, local_index_type,
+                        global_index_type>* result) override;
+
+    /**
+     * Reads a square matrix from the device_matrix_data structure and a global
+     * partition.
```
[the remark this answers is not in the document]
MarcelKoch: yes, otherwise it's not clear how the columns are distributed.

core/distributed/matrix.cpp
```diff
@@ -0,0 +1,440 @@
...
+    auto recv_size = recv_offsets_.back();
+    auto send_dim = dim<2>{static_cast<size_type>(send_size), num_cols};
+    auto recv_dim = dim<2>{static_cast<size_type>(recv_size), num_cols};
+    recv_buffer_.init(exec, recv_dim);
+    send_buffer_.init(exec, send_dim);
+
+    local_b->row_gather(&gather_idxs_, send_buffer_.get());
+
+    auto needs_host_buffer =
+        exec->get_master() != exec && !gko::mpi::is_gpu_aware();
```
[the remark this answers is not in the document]
MarcelKoch: Currently the gpu-aware stuff is only controlled by cmake, so I think for now it makes sense to keep it in a function. But we can think about moving it to the executor at a later time.

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -0,0 +1,544 @@
...
+
+namespace detail {
+
+
+template <template <typename, typename> class MatrixType,
+          typename... CreateArgs>
+struct MatrixTypeBuilderFromValueAndIndex {
+    template <typename ValueType, typename IndexType, std::size_t... I>
+    auto create_impl(std::shared_ptr<const Executor> exec,
+                     std::index_sequence<I...>)
```
pratikvn, line 68: Nice!

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -0,0 +1,544 @@
...
+    void move_to(Matrix<next_precision<value_type>, local_index_type,
+                        global_index_type>* result) override;
+
+    /**
+     * Reads a square matrix from the device_matrix_data structure and a global
+     * partition.
+     *
+     * Both the number of rows and columns of the device_matrix_data is ignored.
+     * The global size of the final matrix is inferred only from the size row
+     * partition.
```
pratikvn, line 283:
```suggestion
     * The global size of the final matrix is inferred only from the size of the row
     * partition.
```

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -0,0 +1,544 @@
...
+    std::shared_ptr<const LinOp> get_local_matrix() const { return local_mtx_; }
+
+    /**
+     * Get read access to the stored non-local matrix
+     *
+     * @return  Shared pointer to the stored non-local matrix
+     */
+    std::shared_ptr<const LinOp> get_non_local_matrix() const
+    {
+        return non_local_mtx_;
```
pratikvn, line 356:
```suggestion
        return this->non_local_mtx_;
```

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -0,0 +1,544 @@
...
+     */
+    Matrix& operator=(Matrix&& other);
+
+protected:
+    /**
+     * Creates an empty distributed matrix.
+     *
+     * @param exec  Executor associated with this matrix.
+     * @param comm  Communicator associated with this matrix.
+     *              The default is the MPI_COMM_WORLD.
```
pratikvn, line 399:
```suggestion
     *              The default is MPI_COMM_WORLD.
```

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -0,0 +1,544 @@
...
+     * Creates an empty distributed matrix with specified type
+     * for local matricies.
+     *
+     * @note It internally clones the passed in matrix_template. Therefore, the
+     *       LinOp should be empty.
+     *
+     * @param exec  Executor associated with this matrix.
+     * @param comm  Communicator associated with this matrix.
+     * @param matrix_template  the local matrices will be constructed with the
+     *                         same runtime type.
```
pratikvn, line 483: Maybe also mention that the non-local matrices will also have the same type.

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -0,0 +1,544 @@
...
+    std::vector<comm_index_type> send_sizes_;
+    std::vector<comm_index_type> recv_offsets_;
+    std::vector<comm_index_type> recv_sizes_;
+    array<local_index_type> gather_idxs_;
+    array<global_index_type> non_local_to_global_;
+    gko::detail::DenseCache<value_type> one_scalar_;
+    gko::detail::DenseCache<value_type> host_send_buffer_;
+    gko::detail::DenseCache<value_type> host_recv_buffer_;
+    gko::detail::DenseCache<value_type> send_buffer_;
+    gko::detail::DenseCache<value_type> recv_buffer_;
```
pratikvn, line 531: Can we maybe add these into structs to simplify readability and that also maybe reduces the number of parameters we need to pass through between functions.

core/distributed/matrix.cpp
```diff
@@ -0,0 +1,440 @@
...
+    array<local_index_type> recv_gather_idxs{exec};
+    array<comm_index_type> recv_offsets_array{exec, num_parts + 1};
+
+    // build diagonal, off-diagonal matrix and communication structures
+    exec->run(matrix::make_build_diag_offdiag(
+        data, make_temporary_clone(exec, row_partition).get(),
+        make_temporary_clone(exec, col_partition).get(), local_part,
+        diag_row_idxs, diag_col_idxs, diag_values, offdiag_row_idxs,
+        offdiag_col_idxs, offdiag_values, recv_gather_idxs,
+        recv_offsets_array.get_data(), non_local_to_global_));
```
pratikvn, line 176: I think it is better to be uniform here with either all parameters as pointers or all as `array`s

core/distributed/matrix.cpp
```diff
@@ -0,0 +1,440 @@
...
+        exec, dim<2>{num_diag_rows, num_diag_cols}, std::move(diag_row_idxs),
+        std::move(diag_col_idxs), std::move(diag_values)};
+    device_matrix_data<value_type, local_index_type> offdiag_data{
+        exec, dim<2>{num_diag_rows, num_ghost_cols},
+        std::move(offdiag_row_idxs), std::move(offdiag_col_idxs),
+        std::move(offdiag_values)};
+    as<ReadableFromMatrixData<ValueType, LocalIndexType>>(this->local_mtx_)
+        ->read(diag_data);
+    as<ReadableFromMatrixData<ValueType, LocalIndexType>>(this->non_local_mtx_)
+        ->read(offdiag_data);
```
pratikvn, line 194: If the requested type is COO, then you can just move the arrays into the COO object, right ?

core/distributed/matrix.cpp
```diff
@@ -0,0 +1,440 @@
...
+
+    // exchange step 1: determine recv_sizes, send_sizes, send_offsets
+    exec->get_master()->copy_from(exec.get(), num_parts + 1,
+                                  recv_offsets_array.get_data(),
```
pratikvn, line 198: nit: 
```suggestion
                                  recv_offsets_array.get_const_data(),
```

core/distributed/matrix.cpp
```diff
@@ -0,0 +1,440 @@
...
+}
+
+
+template <typename ValueType, typename LocalIndexType, typename GlobalIndexType>
+void Matrix<ValueType, LocalIndexType, GlobalIndexType>::convert_to(
+    Matrix<next_precision<value_type>, local_index_type, global_index_type>*
+        result) const
+{
+    GKO_ASSERT(this->get_communicator().size() ==
+               result->get_communicator().size());
```
pratikvn, line 106: I think we should maybe copy the communicator object as well ? Currently, if the matrix being moved into has not been created with a communicator, then these `copy_from` and `move_to` operations throw. The default constructor creates it with MPI_COMM_WORLD, so I am also not completely sure about that.

core/distributed/matrix.cpp
```diff
@@ -0,0 +1,440 @@
...
+        non_local_mtx_->move_from(other.non_local_mtx_.get());
+        gather_idxs_ = std::move(other.gather_idxs_);
+        send_offsets_ = std::move(other.send_offsets_);
+        recv_offsets_ = std::move(other.recv_offsets_);
+        recv_sizes_ = std::move(other.recv_sizes_);
+        send_sizes_ = std::move(other.send_sizes_);
+        recv_sizes_ = std::move(other.recv_sizes_);
+        non_local_to_global_ = std::move(other.non_local_to_global_);
+        one_scalar_.init(this->get_executor(), dim<2>{1, 1});
+        one_scalar_->fill(one<value_type>());
```
pratikvn, line 426: You can also probably just move from `other.one_scalar_` ?

core/test/mpi/distributed/matrix.cpp
```diff
@@ -0,0 +1,630 @@
...
+        ASSERT_NO_THROW(gko::as<expected_type>(mat->get_local_matrix()));
+        additional_test(mat->get_local_matrix().get());
+        additional_test(mat->get_non_local_matrix().get());
+        this->expected_interface_no_throw(mat.get(), with_matrix_type,
+                                          with_matrix_type);
+    });
+}
+
+
+TYPED_TEST(MatrixBuilder, BuildWithInnerAndGhost)
```
pratikvn, line 233:
```suggestion
TYPED_TEST(MatrixBuilder, BuildWithLocalAndNonLocal)
```

core/test/mpi/distributed/matrix.cpp
```diff
@@ -0,0 +1,630 @@
...
+
+
+    std::shared_ptr<const gko::ReferenceExecutor> ref;
+    gko::mpi::communicator comm;
+};
+
+TYPED_TEST_SUITE(MatrixBuilder, gko::test::ValueLocalGlobalIndexTypes);
+
+
+TYPED_TEST(MatrixBuilder, BuildWithInner)
```
pratikvn, line 210:
```suggestion
TYPED_TEST(MatrixBuilder, BuildWithLocal)
```

core/test/mpi/distributed/matrix.cpp
```diff
@@ -0,0 +1,630 @@
...
+    this->template forall_matrix_types([this](auto with_inner_matrix_type,
+                                              auto expected_inner_type_ptr,
+                                              auto additional_inner_test) {
+        using expected_inner_type = typename std::remove_pointer<decltype(
+            expected_inner_type_ptr.get())>::type;
+        this->forall_matrix_types([=](auto with_ghost_matrix_type,
+                                      auto expected_ghost_type_ptr,
+                                      auto additional_ghost_test) {
+            using expected_ghost_type = typename std::remove_pointer<decltype(
+                expected_ghost_type_ptr.get())>::type;
```
pratikvn, line 247: nit: Maybe for this whole file do a `s/inner/local` and `s/ghost/non_local` ?

pratikvn (approved):
Really nice work. The framework and tests look comprehensive. 

A minor nit: I think currently there is a mix of all three terms: `diag/off_diag`, `local/non_local` and `inner/ghost`. I think you should uniformize everything to `local/non_local`, and not only the user facing side. For example the kernel names are still `build_diag_offdiag` and to make things easier `build_local_nonlocal` would be better and similar in other places as well.

core/distributed/matrix.cpp
MarcelKoch: for recv_offsets it's not necessary to use array, because the array size is known beforehand, but I will still switch it to array.

core/distributed/matrix.cpp
MarcelKoch: The idea was to handle the communicator similar to how we handle the executor, i.e. it doesn't change during assignments. I don't think it's possible to create a matrix without a communicator (unless we talk about MPI_COMM_NULL), so that shouldn't cause issues. But if the sizes differ than I throw here, because the whole send_sizes_, etc are not compatible.

core/distributed/matrix.cpp
MarcelKoch: I think filling it explicitly here is the safer choice.

core/distributed/matrix.cpp
upsj: I've been somewhat consistently using the rule: If it needs to be resized, use array&, otherwise use pointer. Not sure if that makes sense, but I'll just mention it here

greole (approved):
Great work! Just some minor comments from my side.

core/distributed/matrix.cpp
```diff
@@ -0,0 +1,431 @@
...
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+#include <ginkgo/core/distributed/matrix.hpp>
+
+
```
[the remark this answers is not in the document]
MarcelKoch: yes, that is our coding convention

core/distributed/matrix.cpp
```diff
@@ -0,0 +1,440 @@
...
+    // exchange step 1: determine recv_sizes, send_sizes, send_offsets
+    exec->get_master()->copy_from(exec.get(), num_parts + 1,
+                                  recv_offsets_array.get_const_data(),
+                                  recv_offsets_.data());
+    std::adjacent_difference(recv_offsets_.begin() + 1, recv_offsets_.end(),
+                             recv_sizes_.begin());
+    comm.all_to_all(recv_sizes_.data(), 1, send_sizes_.data(), 1);
+    std::partial_sum(send_sizes_.begin(), send_sizes_.end(),
+                     send_offsets_.begin() + 1);
+    send_offsets_[0] = 0;
```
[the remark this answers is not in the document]
MarcelKoch: yes, that is only part of the other branch

MarcelKoch:
format!

yhmtsai (approved):
I would like to ensure the testing matrix's property is what you want to test before merging it.
Others comments are for format and I don't mind you do that in other pr

omp/test/distributed/matrix_kernels.cpp
```diff
@@ -0,0 +1,384 @@
...
+        gko::test::generate_random_array<gko::distributed::comm_index_type>(
+            num_rows,
+            std::uniform_int_distribution<gko::distributed::comm_index_type>(
+                0, num_parts - 1),
+            this->engine, this->ref);
+    auto input = gko::test::generate_random_device_matrix_data<
+        value_type, global_index_type>(
+        num_rows, num_cols,
+        std::uniform_int_distribution<int>(static_cast<int>(num_cols),
+                                           static_cast<int>(num_cols)),
```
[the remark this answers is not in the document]
MarcelKoch: I might have wanted to ensure that no parts are empty, so I used a dense matrix here. I don't think it would make any difference though. I will keep it as it is.

PR #972, opened 2022-02-16
DPC++ CMake configuration

### LINKED ISSUE #919: Dpcpp debug build issue
When we add more function into dpcpp module, the debug build will give some compilation issue.
It will throw `relocation truncated to fit: R_X86_64_GOTPCREL....` and `PC-relative offset overflows in PLT entry ...`

The following are some observation:
- It depends on the system gcc, so different environments do not throw the error at the same commit.
- It's does not help when we add -mc-model=medium/large
- the total object file size are around 220 M but the library takes 1.9 G (when it is still compilable)
- With [bloaty](https://github.com/google/bloaty),  more than 90% size is for `__CLANG_OFFLOAD_BUNDLE__sycl-spir64`
- `-fsycl-device-code-split=per_source` only needs 216M for library, not 1.9G when `-fsycl-device-code-split=per_kernel`

The size seems to introduced by inserting the debug information to each kernel.

possible workaround and decision currently from our side
- split the dpcpp to several sub-libraries.
It works but requires -lginkgo_dpcpp_1 -lginkgo_dpcpp_2 ... when not using CMake.
Thus, we do not prefer this way
- Collecting all subgroup/workgroup list into config and it can be changed in CMake.
By doing so, we can limit ginkgo only compiles the supported parameters kernel on specific device and use `-fsycl-device-code-split=per_source` to reduce the library size.

Before pushing these fixes, we disable the debug build in #676

tcojean:
Gives the ability to select DPC++ hardware parameters from CMake.

This is old code and needs to be retested.

- [x] Fix ELL deadlock on DevEP
- [ ] Fix `per_source` on CPU through generic file splitting. 

Fixes #919


dpcpp/CMakeLists.txt
```diff
@@ -51,25 +51,67 @@ target_sources(ginkgo_dpcpp
...
+    ${CMAKE_CURRENT_SOURCE_DIR}/base/config.hpp @ONLY)
+
+
 ginkgo_compile_features(ginkgo_dpcpp)
 target_compile_definitions(ginkgo_dpcpp PRIVATE GKO_COMPILING_DPCPP _ONEDPL_COMPILE_KERNEL=0)
 
 set(GINKGO_DPCPP_FLAGS ${GINKGO_DPCPP_FLAGS} PARENT_SCOPE)
 target_compile_options(ginkgo_dpcpp PRIVATE "${GINKGO_DPCPP_FLAGS}")
 target_compile_options(ginkgo_dpcpp PRIVATE "${GINKGO_COMPILER_FLAGS}")
+target_compile_features(ginkgo_dpcpp PRIVATE cxx_std_17)
```
upsj, line 93: Maybe this should be PUBLIC to avoid potential differences in code generation for header files leading to issues with ODR?

cmake/build_helpers.cmake
```diff
@@ -152,3 +152,46 @@ function(ginkgo_extract_dpcpp_version DPCPP_COMPILER GINKGO_DPCPP_VERSION)
...
+    list(GET _sort_list ${sublist_counter} counter_element)
+  endwhile()
+
+  # Place the last element at the correct position.
+  math(EXPR counter_plus_one "${sublist_counter} + 1")
+  list(REMOVE_AT _sort_list ${counter_plus_one})
+  list(INSERT _sort_list ${counter_plus_one} ${last_element})
+  # Send the modified list back up to parent scope.
+  set(sort_list ${_sort_list} PARENT_SCOPE)
+endfunction()
```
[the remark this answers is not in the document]
tcojean: Well in CMake 3.18 this whole thing disappears as there is native support for this. But I understand your point.

dpcpp/CMakeLists.txt
```diff
@@ -79,7 +79,8 @@ target_link_options(ginkgo_dpcpp PRIVATE -fsycl-device-lib=all)
 
 if (NOT GINKGO_DPCPP_AOT_TARGETS STREQUAL "")
     string(REPLACE ";" "," GINKGO_DPCPP_AOT_TARGETS_FORMATTED "${GINKGO_DPCPP_AOT_TARGETS}")
-    target_compile_options(ginkgo_dpcpp PRIVATE -fsycl-targets=${GINKGO_DPCPP_AOT_TARGETS_FORMATTED})
+    target_compile_options(ginkgo_dpcpp PUBLIC -fsycl-targets=${GINKGO_DPCPP_AOT_TARGETS_FORMATTED})
+    target_link_options(ginkgo_dpcpp PUBLIC -fsycl-targets=${GINKGO_DPCPP_AOT_TARGETS_FORMATTED})
```
tcojean, line 83: Good catch on adding `link_options`. But does this need to be public? This would probably fail if some program links against Ginkgo but uses a non DPC++ compiler which does not have the option (like gcc).

yhmtsai: it's mainly from cpu side of https://www.intel.com/content/www/us/en/develop/documentation/oneapi-dpcpp-cpp-compiler-dev-guide-and-reference/top/compilation/ahead-of-time-compilation.html. If do not add it with the final execution, it can only use host queue (no subgroup or others support).
I am not pretty sure the compile or linking

MarcelKoch:
We need to reevaluate this approach to fix #919. I will close this PR for now.

Issue #805, opened 2021-06-28
Adapt solvers to use work vectors to minimize re-allocations


pratikvn:
Currently, in the core of all our solver implementations, in every apply call, we allocate the required vectors and de-allocate them at the end of function call. It might make sense to have work vectors, allocate them in the generate step, only de-allocate them when the solver factory goes out of scope. This can be particularly advantageous when the same solver being applied multiple times without re-generation, as should be the case in many applications. Some things to consider:

1. Currently, at generate time, we will not know the number of right hand-sides being used, but I think it makes sense to have a factory parameter setting this to allow for allocation of work vectors.
2. The performance of this approach needs to be validated to make sure it is worth the effort.

This is in some way related to having a common solver base class to ensure common parameters and functions can be derived for specific concrete solvers from there.

upsj:
In case allocation overheads become a measurable issue, there are three places we could try to tackle this:
* In Executor `raw_alloc/free`
* In IterativeSolverBase, basically providing protected functions *create_internal_vector(...)/create_stopping_criterion(...)* that use preexisting data if the dimensions match
* Explicitly in the solvers

I think I really like the idea of storing this data in the solvers, we just need to make sure it doesn't interfere with polymorphic vector setup in MFEM. But maybe an explicit `typeid(a) == typeid(b)` is enough?

yhmtsai:
I think it is from the MFEM discussion? to avoid the create_with_config_of issue.
Isn't it a part of MemorySpace? and maybe related to the global constant.
For the solver base, we need to mark the internal_vector mutable such that we use it in const LinOp.
Another way is `apply_with_workspace(x, b, workspace)`

upsj:
Doing this kind of optimization in the MemorySpace or Executor may be difficult, since we lack some of the context (input dimensions, stride).

Just to clarify my comment a bit, I would not pre-allocate data in `generate`, but instead on the first `apply` similar to `Composition`, `Combination`, `Ilu` and `Ic`. This avoids the need for a RHS factory parameter, and also limits the lifetime of these objects to the lifetime of the generated solver.

The storage class will definitely need to be `mutable`, but the objects themselves don't need that.

[...]

PR #770, opened 2021-05-20
Data Validation

greole:
This PR implements data validation strategies as outlined in #746. The aim is to provide a set of validation functions, to evaluate the validity of matrices for debug builds.

The implementation is done in several stages: 

- [x] Free functions checking:
  - [x] If a given matrix is symmetric.
  - [x] If a given matrix has a non-zero diagonal.
  - [x] If given indices are ordered.
  - [x] If given indices are within given bounds.
  - [x] If given indices are unique.
  - [x] If given values are finite.
  - [x] If differences in indices are below given value.
  - [x] If matrix is triangular.

- [ ] Matrices:
   - [x] Coo:  Values are finite, bounded indices (row,column), row pointers are ordered.
   - [x] Csr: Values are finite,  indices are within bounds, row_ptr are sorted and unique.
   - [x] Ell:  Values are  finite, column index is within bounds
   - [x] Diagonal: Values are  finite.
   - [x] Dense: Values are  finite.
   - [x] Fbcsr:  Values are finite,  indices are within bounds, row_ptr are sorted and unique.
   - [x] Hybrid: The underlying Coo and Ell matrix are valid.
   - [ ] Permutation: indices are unique wrt the permutation.
   - [ ] Sellp:  Slice offsets are non-descending, slice lengths are consistent with slice offsets.

   
- [x] Factorizations: 
   - [x] Check if output is valid and has non-zero diagonal
   - [ ] IC: check if diagonal entries are >0 and if input matrix is symmetric
   
- [ ] Preconditioners:
    - [ ] Jacobi: The block pointers need to be ascending with gaps smaller than max_block_size
    - [ ] Jacobi: The blocks must be invertible/inverted blocks must contain finite values
    - [ ] Ic: The system matrix must be symmetric
    - [ ] Ilu:
    - [ ] Isai: The output must be a valid Csr matrix

- [ ] Solvers:
   - [ ] All solvers: system matrix is valid
   - [ ] Iterative solvers: preconditioner is valid
   - [ ] CG: system matrix is symmetric.
   - [ ] Upper/LowerTrs: system matrix needs to be triangular with non-zero diagonal entries.

PR #973, opened 2022-02-18
Add generic solver test

upsj:
This adds a generic test testing the behavior of different solvers in different cases on a small input matrix, and fixes a few found issues.
The test will later be used in #753 to test the cross-executor behavior.

TODO
- [x] ~~fix IDR issues with strided vectors~~


MarcelKoch:
Perhaps you would want to add a test where the initial guess is already the solution. I think gmres and cb_gmres don't work correctly in that case. I have a patch for that, I just didn't get to push that yet.

upsj:
That a great idea! I'll get to that right away

test/solver/solver.cpp
```diff
@@ -0,0 +1,742 @@
...
+
+    test_pair<Mtx> gen_mtx(int num_rows, int num_cols, int min_cols,
+                           int max_cols)
+    {
+        auto data =
+            gko::test::generate_random_matrix_data<value_type, index_type>(
+                num_rows, num_cols,
+                std::uniform_int_distribution<>(min_cols, max_cols),
+                std::normal_distribution<>(0.0, 1.0), rand_engine);
+        Config::preprocess(data);
```
Slaedr, line 457: You could just use `generate_random_matrix<Mtx>`, once the matrix_utils are templated on the matrix type. The functions below don't actually need the intermediate step of generating a `matrix_data` outside the `generate_random_matrix` function.

test/solver/solver.cpp
```diff
@@ -0,0 +1,742 @@
...
+            }
+            if (Config::is_preconditionable()) {
+                SCOPED_TRACE("Preconditioned solver with 4 iterations");
+                guarded_fn(test_pair<SolverType>{
+                    Config::build_preconditioned(ref, 4).on(ref)->generate(
+                        mtx.ref),
+                    Config::build_preconditioned(exec, 4).on(exec)->generate(
+                        mtx.dev)});
+            }
+        }
```
Slaedr, line 582: You could also check for a `is_restartable` and test with 0 restart, 1 restart and some normal restart count. This could also include subspace dimension for IDR.

Slaedr (commented):
Nice job! I have a couple of suggestions below - one related to simplifying the use of matrix_utils and the other related to testing restart (or subspace dim) related stuff.

include/ginkgo/core/base/range.hpp
```diff
@@ -97,9 +97,9 @@ struct span {
     /**
      * Checks if a span is valid.
      *
-     * @return true if and only if `this->begin < this->end`
+     * @return true if and only if `this->begin <= this->end`
      */
-    GKO_ATTRIBUTES constexpr bool is_valid() const { return begin < end; }
+    GKO_ATTRIBUTES constexpr bool is_valid() const { return begin <= end; }
```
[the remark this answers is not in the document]
upsj: IMO that's quite natural: [a,b) contains all integers >= a and < b, so [1,1) is the empty set, yes. Having definitions that break down in empty cases is error-prone, as we can see when trying to extract an empty submatrix.

core/test/utils/matrix_utils.hpp
```diff
@@ -47,108 +47,221 @@ namespace test {
...
- * @tparam ValueType  valuetype of Dense matrix to process
+ * @param data  the matrix data
  *
- * @param mtx  the dense matrix
+ * @tparam ValueType  the value type
+ * @tparam IndexType  the index type
  */
-template <typename ValueType>
-void make_symmetric(matrix::Dense<ValueType>* mtx)
+template <typename ValueType, typename IndexType>
```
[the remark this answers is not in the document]
upsj: The old one is really limited in usability - it only works on Dense matrices, so there will be an intermediate step every time you want to use another matrix type (in my case: solver tests are much faster with Csr than with Dense)

omp/test/preconditioner/jacobi_kernels.cpp
```diff
@@ -399,10 +401,13 @@ TEST_F(Jacobi, OmpScalarApplyEquivalentToRef)
 {
     gko::size_type dim = 313;
     std::default_random_engine engine(42);
-    auto dense_smtx = gko::share(gko::test::generate_random_matrix<Vec>(
-        dim, dim, std::uniform_int_distribution<>(1, dim),
-        std::normal_distribution<>(1.0, 2.0), engine, ref));
-    gko::test::make_diag_dominant(dense_smtx.get());
```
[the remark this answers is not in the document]
upsj: this is really hard to do generically though. Doing matrix->write + make_diag_dominant + matrix->read has much more overhead, so we need a matrix_data involved either way

test/solver/solver.cpp
upsj: I tried incorporating this by using two GMRES configurations: One that restarts when doing 4 iterations, and one that doesn't. How would you behave differently if `is_restartable` is available or not?

core/test/utils/matrix_utils.hpp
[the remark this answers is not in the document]
Slaedr: I mean to say keep your current implementation, but move the `mtx->read(data)` and `mtx->write(data)` inside the respective function such as `make_symmetric`. Template them on the matrix type. Right now, there are always read and write functions surrounding your calls to these functions; you might as well move them inside.

core/test/utils/matrix_utils_test.cpp
```diff
@@ -54,70 +54,128 @@ class MatrixUtils : public ::testing::Test {
...
+
+
+TYPED_TEST(MatrixUtils, MakeUpperTriangularCorrectly)
+{
+    gko::test::make_upper_triangular(this->data);
+
+    auto mtx = TestFixture::mtx_type::create(this->exec);
+    mtx->read(this->data);
+    for (gko::size_type i = 0; i < mtx->get_size()[0]; i++) {
+        for (gko::size_type j = 0; j < i; j++) {
```
Slaedr, line 123: I would also verify that the data in the upper triangular part is the same as before. Similar for the other operations.

test/solver/solver.cpp
Slaedr: Indeed, ideally, there should be no difference. But it would be nice just as a check for how the restart parameter is treated, since it and the max iterations are two separate variables in the implementation.

omp/test/preconditioner/jacobi_kernels.cpp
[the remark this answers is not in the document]
Slaedr: I agree that matrix_data should be involved. What I'm referring to is a very minor change where the matrix->write and matrix->read are inside the utils functions, to spare the need to do that manually every time.

cuda/test/preconditioner/jacobi_kernels.cpp
```diff
@@ -414,10 +416,13 @@ TEST_F(Jacobi, CudaScalarApplyEquivalentToRef)
 {
     gko::size_type dim = 313;
     std::default_random_engine engine(42);
-    auto dense_smtx = gko::share(gko::test::generate_random_matrix<Vec>(
-        dim, dim, std::uniform_int_distribution<>(1, dim),
-        std::normal_distribution<>(1.0, 2.0), engine, ref));
-    gko::test::make_diag_dominant(dense_smtx.get());
```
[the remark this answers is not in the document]
upsj: see https://github.com/ginkgo-project/ginkgo/pull/973#discussion_r810464108

omp/test/preconditioner/jacobi_kernels.cpp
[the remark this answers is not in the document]
upsj: see https://github.com/ginkgo-project/ginkgo/pull/973#discussion_r810464108

test/solver/solver.cpp
upsj: see https://github.com/ginkgo-project/ginkgo/pull/973#discussion_r810464108

core/test/utils/matrix_utils.hpp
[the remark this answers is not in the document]
upsj: Where is the `write` happening here? Taking a matrix parameter, we would have `generate_random_matrix_data -> read -> write -> make_* -> read`, this approach prevents omits a `read -> write` pair.

To make things more easily navigatable, I'll link the other comments here and resolve them: https://github.com/ginkgo-project/ginkgo/pull/973#discussion_r810464331 https://github.com/ginkgo-project/ginkgo/pull/973#discussion_r810525909 https://github.com/ginkgo-project/ginkgo/pull/973#discussion_r810467617

test/solver/solver.cpp
```diff
@@ -0,0 +1,742 @@
...
+        this->forall_solver_scenarios(mtx, [&](auto solver) {
+            this->template forall_vector_scenarios<MixedVec>(
+                solver, [&](auto b, auto x) {
+                    auto alpha = this->template gen_scalar<MixedVec>();
+                    auto beta = this->template gen_scalar<MixedVec>();
+
+                    solver.ref->apply(alpha.ref.get(), b.ref.get(),
+                                      alpha.ref.get(), x.ref.get());
+                    solver.dev->apply(alpha.dev.get(), b.dev.get(),
+                                      alpha.dev.get(), x.dev.get());
```
MarcelKoch, line 735:
```suggestion
                    solver.ref->apply(alpha.ref.get(), b.ref.get(),
                                      beta.ref.get(), x.ref.get());
                    solver.dev->apply(alpha.dev.get(), b.dev.get(),
                                      beta.dev.get(), x.dev.get());
```

test/solver/solver.cpp
```diff
@@ -0,0 +1,742 @@
...
+    this->forall_matrix_scenarios([&](auto mtx) {
+        this->forall_solver_scenarios(mtx, [&](auto solver) {
+            this->forall_vector_scenarios(solver, [&](auto b, auto x) {
+                auto alpha = this->gen_scalar();
+                auto beta = this->gen_scalar();
+
+                solver.ref->apply(alpha.ref.get(), b.ref.get(), alpha.ref.get(),
+                                  x.ref.get());
+                solver.dev->apply(alpha.dev.get(), b.dev.get(), alpha.dev.get(),
+                                  x.dev.get());
```
MarcelKoch, line 695:
```suggestion
                solver.ref->apply(alpha.ref.get(), b.ref.get(), beta.ref.get(),
                                  x.ref.get());
                solver.dev->apply(alpha.dev.get(), b.dev.get(), beta.dev.get(),
                                  x.dev.get());
```

test/solver/solver.cpp
```diff
@@ -0,0 +1,742 @@
...
+        ASSERT_EQ(mtx->get_stopping_criterion_factory(), nullptr);
+    }
+};
+
+
+struct Cg : SimpleSolverTest<gko::solver::Cg<solver_value_type>> {};
+
+
+struct Cgs : SimpleSolverTest<gko::solver::Cgs<solver_value_type>> {
+    static double tolerance() { return 1e5 * r<value_type>::value; }
```
MarcelKoch, line 129: Can you explain, why you are using such large error tolerances here? Since these will be used as a relative error bound, these values seem way too big. E.g. when using `float` the tolerances below become nearly 1 so, they are useless in a relative comparison. I think one issue might be that we don't solve until convergence in some cases, so it's not clear what to compare the results against.

MarcelKoch (commented):
I have just a few remarks on the new test itself. I will also check out the rest of this PR later.

I have a question regarding the tolerance used below. Besides, we should perhaps add an iterative solver test, where the test runs until convergence. We could then check, if the residual actually achieves (nearly) the prescribed accuracy.

test/solver/solver.cpp
MarcelKoch: I guess that this could become a more serious issue. With a tolerance that large, you could probably introduce errors into the device kernels and it would still pass. TBH I also don't know what the best way is to test these cases (0&4 iterations).

cuda/solver/cb_gmres_kernels.cu
```diff
@@ -479,6 +482,9 @@ void step_2(std::shared_ptr<const CudaExecutor> exec,
             matrix::Dense<ValueType>* before_preconditioner,
             const Array<size_type>* final_iter_nums)
 {
+    if (before_preconditioner->get_size()[1] == 0) {
```
MarcelKoch, line 485: Just a comment, not holding anything up, but would it make sense to check for that (i.e. `num_rhs > 0`) right at the beginning of the solver application? (That also holds for the other solvers, and it should reduce the repetition across solver steps and devices)

dpcpp/solver/gmres_kernels.dp.cpp
```diff
@@ -252,6 +255,9 @@ void update_next_krylov_kernel(
     size_type stride_krylov, const ValueType* hessenberg_iter,
     size_type stride_hessenberg, const stopping_status* stop_status)
 {
+    if (stride_krylov == 0) {
```
MarcelKoch, line 258: Just as a side note from seeing this, should we make sure that is a user provides a stride that it is always >= the number of columns?

include/ginkgo/core/base/range.hpp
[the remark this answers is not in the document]
MarcelKoch: I think that will introduce other issues. E.g. `span(1, 1)` would be valid, but it has zero length. That will probably lead to other issues elsewhere. Perhaps we could use a (debug-only) check in the two parameters constructor to check that the span is valid.

MarcelKoch (commented):
generally lgtm, but I would like to wait until the issue with the large tolerance is resolved.

Below are some smaller remarks, mainly about the `num_rhs == 0` checking.

include/ginkgo/core/base/range.hpp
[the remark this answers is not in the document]
MarcelKoch: Perhaps, my view is a bit too strict and it makes sense to have a valid span of zero length. I think for our current use cases, it should be ok. The only issue could be if someone wants to divide by the length, but I guess that should not be a common use case.

cuda/solver/cb_gmres_kernels.cu
upsj: Most of this code will change with #861, so I didn't want to spend too much time working on it.

dpcpp/solver/gmres_kernels.dp.cpp
upsj: stride >= num columns is an (implicit) invariant of the matrix format, so that would be caught with a `validate_data` #770

core/test/utils/assertions.hpp
```diff
@@ -259,7 +259,7 @@ ::testing::AssertionResult matrices_near_impl(
              << second_expression << " is " << err << "\n"
              << "\twhich is larger than " << tolerance_expression
              << " (which is " << tolerance << ")\n";
-        if (num_rows * num_cols <= 1000) {
+        if (num_rows * num_cols <= 25) {
```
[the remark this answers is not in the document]
upsj: I guess what I actually wanted is the more specific `num_rows < 10 && num_cols < 10`, as I was mostly looking at long vectors.

upsj:
rebase!

MarcelKoch:
I think we can't progress with this PR until we figured out the tolerance issue.

If we use `float` as value type (like we need for dpcpp single mode), then the test is completely useless. In that case, the tolerance for some solvers is ~1, so even if the code contains errors, the test will still succeed. I actually tested that with Fcg, where I introduced a bug, and the test succeeded (if I used float). If we use `double`, we should catch the most egregious errors, but I think it still leaves a bad feeling. 

I would suggest instead of comparing the device result with the reference result after a fixed number of iterations, to just check if the device solver converges to a specific tolerance. TBH, I think that is the only property that is worth checking.

upsj:
I tend to agree, I will probably keep this blocked and use it only to test #753. With the changes suggested by Pratik in #805, we should then be able to make logging and comparing these results much easier. Note that the errors mainly pop up with a large number of rhs, so with such a small input matrix, it may well be that one of the vectors converges very quickly, leading to instabilities.

MarcelKoch (approved):
I'm not holding this up anymore. The important issues have been raised, but I guess they will require more work than what would be fitting here.

I do have one request, could you disable the test for the `GINKGO_DPCPP_SINGLE_MODE`? At least one test will give you false positives, which I think we should prevent.

upsj:
format!

core/test/utils/matrix_utils.hpp
```diff
@@ -47,108 +47,221 @@ namespace test {
...
-    auto mtx_host =
-        make_temporary_clone(mtx->get_executor()->get_master(), mtx);
-
-    using std::abs;
-    for (size_type i = 0; i < mtx_host->get_size()[0]; ++i) {
-        auto sum = gko::zero<ValueType>();
-        for (size_type j = 0; j < mtx_host->get_size()[1]; ++j) {
-            sum += abs(mtx_host->at(i, j));
+    GKO_ASSERT_IS_SQUARE_MATRIX(data.size);
+    std::vector<remove_complex<ValueType>> norms(data.size[0]);
```
tcojean, line 195: What happens where there is an empty row? Does anything specific need to be done (adding a diagonal entry?) or is keeping the row empty fine?

core/test/utils/matrix_utils.hpp
```diff
@@ -47,108 +47,221 @@ namespace test {
...
+        if (entry.row == entry.column) {
+            diag_positions[entry.row] = i;
+        } else {
+            norms[entry.row] += gko::abs(entry.value);
+        }
+        i++;
+    }
+    for (i = 0; i < data.size[0]; i++) {
+        if (diag_positions[i] < 0) {
+            data.nonzeros.emplace_back(i, i, norms[i] * ratio);
```
tcojean, line 208: Wouldn't we have `0 * ratio` with an empty row?

core/test/utils/matrix_utils_test.cpp
```diff
@@ -57,70 +57,138 @@ class MatrixUtils : public ::testing::Test {
...
+    gko::test::make_lower_triangular(this->data);
+
+    auto mtx = TestFixture::mtx_type::create(this->exec);
+    mtx->read(this->data);
+    for (gko::size_type i = 0; i < mtx->get_size()[0]; i++) {
+        for (gko::size_type j = 0; j <= i; j++) {
+            ASSERT_EQ(mtx->at(i, j), orig_mtx->at(i, j));
+        }
+        for (gko::size_type j = i + 1; j < mtx->get_size()[1]; j++) {
+            ASSERT_EQ(mtx->at(i, j), gko::zero<TypeParam>());
```
tcojean, line 118: Should we stop at the first different number? Or use an EXPECT to go through all of them?
```suggestion
            EXPECT_EQ(mtx->at(i, j), orig_mtx->at(i, j));
        }
        for (gko::size_type j = i + 1; j < mtx->get_size()[1]; j++) {
            EXPECT_EQ(mtx->at(i, j), gko::zero<TypeParam>());
```

core/test/utils/matrix_utils_test.cpp
```diff
@@ -57,70 +57,138 @@ class MatrixUtils : public ::testing::Test {
...
+    gko::test::make_upper_triangular(this->data);
+
+    auto mtx = TestFixture::mtx_type::create(this->exec);
+    mtx->read(this->data);
+    for (gko::size_type i = 0; i < mtx->get_size()[0]; i++) {
+        for (gko::size_type j = 0; j < i; j++) {
+            ASSERT_EQ(mtx->at(i, j), gko::zero<TypeParam>());
+        }
+        for (gko::size_type j = i; j < mtx->get_size()[1]; j++) {
+            ASSERT_EQ(mtx->at(i, j), orig_mtx->at(i, j));
```
tcojean, line 137:
```suggestion
            EXPECT_EQ(mtx->at(i, j), gko::zero<TypeParam>());
        }
        for (gko::size_type j = i; j < mtx->get_size()[1]; j++) {
            EXPECT_EQ(mtx->at(i, j), orig_mtx->at(i, j));
```

core/test/utils/matrix_utils_test.cpp
```diff
@@ -57,70 +57,138 @@ class MatrixUtils : public ::testing::Test {
...
+
+TYPED_TEST(MatrixUtils, MakeUnitDiagonalCorrectly)
+{
+    gko::test::make_unit_diagonal(this->data);
+
+    auto mtx = TestFixture::mtx_type::create(this->exec);
+    mtx->read(this->data);
+    for (gko::size_type i = 0;
+         i < std::min(mtx->get_size()[0], mtx->get_size()[1]); i++) {
+        ASSERT_EQ(mtx->at(i, i), gko::one<TypeParam>());
```
tcojean, line 164:
```suggestion
        EXPECT_EQ(mtx->at(i, i), gko::one<TypeParam>());
```

core/test/utils/matrix_utils_test.cpp
```diff
@@ -57,70 +57,138 @@ class MatrixUtils : public ::testing::Test {
...
-    gko::test::make_symmetric(gko::lend(this->mtx));
+    gko::test::make_symmetric(this->data);
 
-    for (gko::size_type i = 0; i < this->mtx->get_size()[0]; i++) {
+    auto mtx = TestFixture::mtx_type::create(this->exec);
+    mtx->read(this->data);
+    for (gko::size_type i = 0; i < mtx->get_size()[0]; i++) {
         for (gko::size_type j = 0; j <= i; j++) {
-            ASSERT_EQ(this->mtx->at(i, j), this->mtx->at(j, i));
+            ASSERT_EQ(mtx->at(i, j), mtx->at(j, i));
```
tcojean, line 177:
```suggestion
            EXPECT_EQ(mtx->at(i, j), mtx->at(j, i));
```

core/test/utils/matrix_utils_test.cpp
```diff
@@ -57,70 +57,138 @@ class MatrixUtils : public ::testing::Test {
...
-    gko::test::make_hermitian(gko::lend(this->mtx));
+    gko::test::make_hermitian(this->data);
 
-    for (gko::size_type i = 0; i < this->mtx->get_size()[0]; i++) {
+    auto mtx = TestFixture::mtx_type::create(this->exec);
+    mtx->read(this->data);
+    for (gko::size_type i = 0; i < mtx->get_size()[0]; i++) {
         for (gko::size_type j = 0; j <= i; j++) {
-            ASSERT_EQ(this->mtx->at(i, j), gko::conj(this->mtx->at(j, i)));
+            ASSERT_EQ(mtx->at(i, j), gko::conj(mtx->at(j, i)));
```
tcojean, line 191:
```suggestion
            EXPECT_EQ(mtx->at(i, j), gko::conj(mtx->at(j, i)));
```

core/test/utils/matrix_utils_test.cpp
```diff
@@ -129,22 +197,19 @@ TYPED_TEST(MatrixUtils, MakeHermitianCorrectly)
...
         gko::remove_complex<T> off_diag_abs = 0;
-        for (gko::size_type j = 0; j < this->mtx->get_size()[1]; j++) {
+        for (gko::size_type j = 0; j < mtx->get_size()[1]; j++) {
             if (j != i) {
-                off_diag_abs += std::abs(this->mtx->at(i, j));
+                off_diag_abs += gko::abs(mtx->at(i, j));
             }
         }
-        ASSERT_NEAR(gko::real(this->mtx->at(i, i)), off_diag_abs, r<T>::value);
+        ASSERT_GT(gko::abs(mtx->at(i, i)) * (1 + r<T>::value), off_diag_abs);
```
tcojean, line 212:
```suggestion
        EXPECT_GT(gko::abs(mtx->at(i, i)) * (1 + r<T>::value), off_diag_abs);
```

core/test/utils/matrix_utils_test.cpp
```diff
@@ -153,51 +218,91 @@ TYPED_TEST(MatrixUtils, MakeDiagDominantWithRatioCorrectly)
...
-        for (gko::size_type j = 0; j < this->mtx->get_size()[1]; j++) {
+        for (gko::size_type j = 0; j < mtx->get_size()[1]; j++) {
             if (j != i) {
-                off_diag_abs += std::abs(this->mtx->at(i, j));
+                off_diag_abs += gko::abs(mtx->at(i, j));
             }
         }
-        ASSERT_NEAR(gko::real(this->mtx->at(i, i)), off_diag_abs * ratio,
-                    r<T>::value);
+        ASSERT_GT(gko::abs(mtx->at(i, i)) * (1 + r<T>::value),
```
tcojean, line 233:
```suggestion
        EXPECT_GT(gko::abs(mtx->at(i, i)) * (1 + r<T>::value),
```

test/solver/solver.cpp
```diff
@@ -0,0 +1,675 @@
...
+    using solver_type = SolverType;
+    using value_type = typename solver_type::value_type;
+    using index_type = gko::int32;
+    using matrix_type = gko::matrix::Csr<value_type, index_type>;
+
+    static bool is_iterative() { return true; }
+
+    static bool is_preconditionable() { return true; }
+
+    static double tolerance() { return 1e4 * r<value_type>::value; }
```
tcojean, line 84: If we don't have an issue open on the topic, please create one for checking the solver quality. All these tolerances look really high at first glance, we need to check if that's somehow expected in corner cases due to using random data, or if there are problems somewhere.

test/solver/solver.cpp
```diff
@@ -0,0 +1,675 @@
...
+            SCOPED_TRACE("Multivector with 0 columns");
+            guarded_fn(gen_in_vec<VecType>(solver, 0, 0),
+                       gen_out_vec<VecType>(solver, 0, 0));
+        }
+        {
+            SCOPED_TRACE("Single vector");
+            guarded_fn(gen_in_vec<VecType>(solver, 1, 1),
+                       gen_out_vec<VecType>(solver, 1, 1));
+        }
+        // GMRES and others are currently broken w.r.t lucky breakdown
```
tcojean, line 534: Also an issue, except if fixed somewhere.

tcojean (approved):
LGTM! That looks pretty good. I only have minor comments.

You should probably rebase because some of the `gko::share` issues you fixed were also fixed by Thomas I think.

core/test/utils/matrix_utils_test.cpp
upsj: I guess the only reason I used ASSERT is because the output would be huge in the failure case otherwise.

test/solver/solver.cpp
upsj: Not yet, so I'll have to add a issue/fix for that as well

core/test/utils/matrix_utils.hpp
upsj: good point. I guess in that case, we should just set the diagonal to something non-zero to make the matrix non-singular

tcojean (approved):

PR #979, opened 2022-03-04
Add a uniform coarsening algorithm for coarse grid generation.

pratikvn:
This PR adds a new class to generate coarse matrices which can be useful for multigrid and other multi-level methods in general.

Two main methods to generate coarse matrices are provided:

1. A naive one with user providing a constant jump between the rows.
2. ~~A user specified coarse row array.~~ (Separate class and separate PR)

#### TODO:
+ [x] Add omp, ref, cuda and HIP tests.
+ [x] Merge #986 

Due to missing device kernels for creation of submatrix with IndexSet, the creation of the submatrix from the index set will be in a separate PR.


upsj:
I mentioned this in the meeting before: When I hear the description of this functionality, I think "uniform". In my view, selection usually refers to finding a single element, or a small number of elements. The operations you are doing is a coarsening, so how about `UniformCoarsening`?

pratikvn:
Naming it `Coarsening` is a good idea. How about `SimpleCoarsening` ? Because you have the ability to specify an array for the coarsening indices, `Uniform` might not be the best name.

upsj:
These two might be different enough that we could put them into different types? FixedCoarsening and UniformCoarsening?
Do we have any generic strategies for describing or picking the averaging operation? That could be just picking one, (un-)weighted average, ...

pratikvn:
Yes, we could technically put them in two separate classes. I guess at some point we will have a lot of separate classes. But I guess that might be okay if they have different algorithms. 

Currently, this is just picking some nodes, but we could technically have different averaging strategies. This could be as simple as having a sum-normalized per row for the restriction and prolongation matrices to more involved averaging strategies based on diagonal and off-diagonal fine matrix element weights

upsj:
I was thinking that our AMGX implementation does two things: a heavy edge matching to compute the fine-to-coarse mapping, and a uniform averaging scheme. Maybe it might make sense to look at what hypre is doing here? I think they have some averaging schemes specific to certain coarsening algorithms, but also generic ones?

pratikvn:
Yes, I think generation of the restriction and prolongation matrices can be de-coupled from the coarsening strategy itself in some cases. I don't have a clear solution for that now (interface wise), but it might be something we should look at in the future.

yhmtsai:
the multigridLevel is the general thing. I do not put it as the coarsening method and the interpolator methods directly. The coarsening method might generate different information such that not all interpolator can rely on it. The MultigridLevel Factory can has option for intepolator but it should be decided by the coarsening method.
For example, Agg method might give the aggregation group in some sense (the agg group def might be different in different alg), the coarsening method gives the C-F. the information is not direct. Leave the intepolator decided by the the coarsening method to give full flexibilty for it.
The interpolator might not fit in the LinOpFactory directly. in Pgm, the intepolator uses the agg information and the size only.

core/multigrid/uniform_coarsening.cpp
```diff
@@ -0,0 +1,136 @@
...
+    exec->run(uniform_coarsening::make_fill_restrict_op(&coarse_rows_,
+                                                        restrict_op.get()));
+    exec->run(uniform_coarsening::make_fill_array(
+        restrict_op->get_values(), coarse_dim, one<ValueType>()));
+    exec->run(uniform_coarsening::make_fill_seq_array(
+        restrict_op->get_row_ptrs(), coarse_dim + 1));
+
+    auto prolong_op = gko::as<csr_type>(share(restrict_op->transpose()));
+
+    // TODO: Can be done with submatrix index_set.
```
upsj, line 117: I don't think that's true in the general case - the operation you need is more akin to a row_gather with arbitrary permutation than submatrix. That might also relate to your TODO on ordering?

pratikvn:
Closing stale PR

PR #981, opened 2022-03-04
Add ApplyWithInitialGuess, Split criterion in IR

yhmtsai:
~~IR use the mutable workspace to reuse the memory when the input #rows of rhs does not change.~~
Also, splitting the crierion in IR to eliminate one reduction computation in the end if the stopping criterion only checks the iteration count. The IR will not report the convegence reason from iteration count and residual norm check together.
There are other stuff added but they migh be not used now or moved to different pr
~~- ResidualCacheable: allow give the residual workspace from outside, which is useful in multigrid when the solver can not split the iteration check and residual norm check~~
because we split criterion in IR, there is no urgent need for it.
~~- ZeroInput: the first residual of the solver is the right hand side if the input initial guess is zero. this might be different structure or another pull request~~
it is replaced by ApplyWithInitialGuess. use the interface not the mutable variable to pass the initial_guess_mode.
~~Moreover, I can make EnableApplyHint to be feature class (not return the self pointer) or use apply
If make the member function as `apply`, it will require the class inherit from them to use `using LinOp::apply; using ApplyHint::apply` to avoid the ambiguous, I think.~~

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203097352075328


upsj:
Maybe we can get #753 merged first, then we can provide this functionality for temporary vectors more universally?

upsj:
`ResidualCachable` seems pretty specific to solvers, maybe we should put it into SolverBase and return a workspace ID (or invalid_index) for the residual, then we don't need a new interface?

For `ZeroInput`, IMO mutable state like this is really dangerous, and we should try to avoid it for anything but caching. How costly is the initial SpMV? Can we find another way to signal that the vector is zero? `Ir::build().with_initial_guess(initial_guess::zero)`?

yhmtsai:
if putting into workspace structre, it will need to have the `set_workspace(id, std::shared_ptr<Vector>)`. the object life is also extended?
if taking out the workspace outside, it requires doing dummy ir apply to generate the residual vector (or making the create_or_get_op operation public.). This looks weird to me.
making `ir.with_initial_guess(zero);` will lead also reset the initial guess in the middle of W-cycle.
they do not set the initial guess to zero again. Users needs to set them separately

upsj:
Where do you need the lifetime of the residual vector to be extended past its solver's lifetime? It sounds to me like you only want to avoid recomputing the residual by retrieving it from the smoother? Why do you need a shared_ptr at all? If the answer is "stopping criterion", that is just a design issue we need to solve on the criterion side, there should be no ownership involved.
For ZeroInput, there must be a better solution interface-wise (creating two smoothers for the W cycle - the first one with zero input, the second one without?), because IMO IR's behavior should not be special-cased just for this single use-case.
In general, to be feasible, the current approach would at least need to deal correctly with move-assignment etc, but I don't think it's a good starting point.

yhmtsai:
> Where do you need the lifetime of the residual vector to be extended past its solver's lifetime? It sounds to me like you only want to avoid recomputing the residual by retrieving it from the smoother? Why do you need a shared_ptr at all? If the answer is "stopping criterion", that is just a design issue we need to solve on the criterion side, there should be no ownership involved.

yes, but it should be from the multigrid. it can be vector* as the current status. there are three things in this pull request and they can be considered separately and they should be possible to extend to other Solver. The multigrid solver can use any kind of solver as smoother.

> For ZeroInput, there must be a better solution interface-wise (creating two smoothers for the W cycle - the first one with zero input, the second one without?), because IMO IR's behavior should not be special-cased just for this single use-case.
> In general, to be feasible, the current approach would at least need to deal correctly with assignment etc.

Yes, that's what I mean. but it should be controlled by user when we put it as a factory option.
Actually, I prefer the matrix property to do that.
The solver gets the hint from the input and uses it to reduce some unnecessary computation.
The zeroinput situation depends on the input not the solver self.
However, I am thinking about whether using the multigrid as a preconditioner is the same case as these situation

upsj:
I like this approach much better, thanks for the update! I would still like to consolidate this with a potential factory parameter `with_initial_guess` that does the same thing, but I think we should include more people in this discussion.

yhmtsai:
I would say with_initial_guess(zero) is changing the LinOp property like Multigrid usage (apply_use_initial_guess = false when enable)
another quite minor benefit: we can eliminate `fill(zero)` by apply_hint although the benefit is quite minor, I think.
I do not check this one yet.

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -50,6 +51,136 @@ namespace gko {
 namespace solver {
 
 
+/**
+ * give the hint of input to apply
```
tcojean, line 55:
```suggestion
 * Give a hint about the input of the apply method.
```

tcojean (commented):
Some simple comments.

While the current form is fine, I believe that with this we really need solver properties, because if we need more such parameters the amount of specialized `apply` function will not scale. We already have at least one property that is constant for each solver type `apply_uses_initial_guess`, now this property should be controlled by the user.

I agree that this should also be a factory parameter for consistency/convenience. The setting could be changed manually for each apply or be a default of the generated solver itself.

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -50,6 +51,141 @@ namespace gko {
...
+    {
+        if (hint == input_hint::zero) {
+            fill_zero(x);
+        } else if (hint == input_hint::rhs) {
+            x->copy_from(b);
+        }
+        this->apply_impl(b, x);
+    }
+
+    // override at the same time when overriden
```
fritzgoebel, line 148:
```suggestion
    // override at the same time when overridden
```

fritzgoebel (commented):
I would be fine with this approach but I would prefer the suggestion to make this a factory parameter rather than having a special apply function.

yhmtsai:
Rely on user setting the factory parameter.
```
Multigrid::build().with_presmoother(
  IR::build().with_apply_hint(zero).with_inner_solver(Jacobi::build()),
  IR::build().with_apply_hint(zero).with_inner_solver(***::build()))
```
with_apply_hint(zero) is required from the user.
Because the smoother will be `apply_initial_guess = false`, Multigrid no longer to wrap a IR for these operations by the apply_initial_guess.

yhmtsai:
https://github.com/ginkgo-project/ginkgo/commit/c440cb9622443d5281436987bd558e41d75577bd
I have tried the property approach at the above link.
It's not hard/complex as I thought. If the solver does not provide the apply hint, it still performs like before.
apply will use the property as the hint.
using apply_hint overrides the hint from the solver such that I can still use it in multigrid.

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -50,6 +51,198 @@ namespace gko {
...
+    } else {
+        GKO_NOT_IMPLEMENTED;
+    }
+}
+
+
+}  // namespace
+
+
+class ApplyHint {
```
pratikvn, line 97: MIssing doc for class.

pratikvn (commented):
A few comments and a general question:

I think our current approach to this with options inside the class parameters increases the complexity of the code quite a bit. Maybe having more apply methods which use additional parameters is a better way to do this ? For example:
```c++
// The default apply which does not use the data in x.
LinOp* apply(LinOp* b, LinOp* x) {}

// Apply with some initial guess
LinOp* apply(LinOp* b, LinOp* x, LinOp* initial_guess) {}
```

I think being explicit in this fashion avoids confusion and especially all the different parameters that need to be set in the class/LinOp you are trying to apply makes the class pretty convoluted.

yhmtsai:
if we would like to have the property in solver, we need to use it in apply.
apply_hint(b, x, hint) is already supported.
Giving the initial guess is different from the zero input. 
When give the zero initial guess, solver still does not the initial guess is zero.
It requires the property for it as I said in the beginning.
I only need zero input. given is for the original purpose, and rhs is from other request.

pratikvn:
Yes, I think for zero input, this makes sense. But I wanted to bring this general question up about having one single apply and having special cases inside controlled through the factory parameters and having different different apply functions, which I feel might be more clean ?

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -50,6 +51,198 @@ namespace gko {
...
+    // override this at the same time when overridden
+    virtual void apply_impl(const LinOp* alpha, const LinOp* b,
+                            const LinOp* beta, LinOp* x) const = 0;
+
+private:
+    input_hint hint_;
+};
+
+template <typename DerivedType>
+class EnableApplyHint : public ApplyHint {
```
tcojean, line 212: Also, despite it being just the `Enable` (mixin) version, there needs to be some documentation
```suggestion


template <typename DerivedType>
class EnableApplyHint : public ApplyHint {
```

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -50,6 +51,198 @@ namespace gko {
...
+     * Performs the operation x = op(b) with a hint, where op is this linear
+     * operator.
+     *
+     * @param b  the input vector(s) on which the operator is applied
+     * @param x  the output vector(s) where the result is stored
+     * @param hint  the input hint to handle the input vector(s)
+     *
+     * @return this
+     */
+    ApplyHint* apply_hint(const LinOp* b, LinOp* x, input_hint hint)
```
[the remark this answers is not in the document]
tcojean: On the other hand, do we need another name? Shouldn't we just use `apply()`? That probably makes interfacing simpler for the user, if they would ever use this function.

Another topic is, do we want to expose this to the user, actually? As far as I understand, this is mostly required for IR and other solvers when composing objects (solvers). Users probably only need the factory approach.

test/solver/ir_kernels.cpp
```diff
@@ -243,3 +243,37 @@ TEST_F(Ir, RichardsonApplyWithIterativeInnerSolverIsEquivalentToRef)
...
+
+
+TEST_F(Ir, ApplyWithGivenHintIsEquivalentToRef)
+{
+    using input_hint = gko::solver::input_hint;
+    auto mtx = gko::share(gen_mtx(50, 50, 52));
+    auto b = gen_mtx(50, 3, 7);
+    auto d_mtx = gko::share(clone(exec, mtx));
+    auto d_b = clone(exec, b);
+    for (auto hint : {input_hint::given, input_hint::rhs, input_hint::zero}) {
```
tcojean, line 255: Maybe, create a `std::initializer_list<gko::solver::input_hint>` in the `solver` header, so this is not repeated in every single place and the enum is iterable through that?

tcojean (commented):
Here are some comments. In general, I think I go in the direction of Pratik but I'm not sure which way is the best exactly. I think this is somewhat convoluted and I'm not too happy with having a million apply for every single option living in our code. Do we even need to expose this `apply` to the user? That would simplify things a little, as we can maybe put these functions as `protected` at least then (although that's still somewhat interface level if users use inheritance...).

test/solver/ir_kernels.cpp
yhmtsai: I think the iteration is only for test purposes, so I wouldn't like to introduce in public header

reference/test/solver/ir_kernels.cpp
```diff
@@ -450,4 +450,43 @@ TYPED_TEST(Ir, RichardsonConjTransposedSolvesTriangularSystem)
...
 
+TYPED_TEST(Ir, ApplyWithGivenHintIsEquivalentToRef)
+{
+    using Mtx = typename TestFixture::Mtx;
+    using value_type = typename TestFixture::value_type;
+    using input_hint = gko::solver::input_hint;
+    auto ref_solver =
+        gko::solver::Ir<value_type>::build()
+            .with_criteria(
+                gko::stop::Iteration::build().with_max_iters(1u).on(this->exec))
```
thoasm, line 461: Shouldn't 1 iteration already perform one iteration, meaning `x` should be modified slightly?

core/solver/multigrid.cpp
```diff
@@ -225,7 +228,8 @@ struct MultigridState {
 
     void run_cycle(multigrid::cycle cycle, size_type level,
                    const std::shared_ptr<const LinOp>& matrix, const LinOp* b,
-                   LinOp* x, bool is_first = true, bool is_end = true)
+                   LinOp* x, bool x_is_zero = false, bool is_first = true,
+                   bool is_end = true)
```
thoasm, line 232: This seems quite hard to use with so many boolean options (and easy to use it the wrong way because you remembered a different order).
Maybe an enum could help here (combining options with binary or)?

thoasm (commented):
For the most part, I am fine with this PR.
However, I dislike the word `hint` as it does not seem to fit its purpose. I would recommend using `guess` instead, and update both the code and the documentation.

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -50,6 +51,246 @@ namespace gko {
...
+     * Performs the operation x = op(b) with a hint, where op is this linear
+     * operator.
+     *
+     * @param b  the input vector(s) on which the operator is applied
+     * @param x  the output vector(s) where the result is stored
+     * @param hint  the input hint to handle the input vector(s)
+     *
+     * @return this
+     */
+    ApplyHint* apply_with_hint(const LinOp* b, LinOp* x, input_hint hint)
```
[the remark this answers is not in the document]
yhmtsai: apply will take the factory parameter as default.
apply_with_hint overrides the original behavior

reference/test/solver/ir_kernels.cpp
yhmtsai: I do not get it.
ref_solver and solver will perform 1 iteration update.
ref_solver takes x as initial guess like the original behavior.
but solver will add the hint (or initial_guess).
For example, ref_solver will take zero vector ref_x but solver will take nonzero vector x with zero guess.
Compare the result after one iteration

include/ginkgo/core/solver/solver_base.hpp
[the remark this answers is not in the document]
thoasm: But the `apply_impl` pretty much does exactly what `apply_with_hint` does, just that the first is `virtual`, and the second one a regular function that is overshadowed in the `EnableApplyHint`.
Since both are protected, I just don't see a point in having both.

reference/test/solver/ir_kernels.cpp
thoasm: You are correct, I thought you simply compared the `x` after running not a single iteration.
I somehow missed that you created another solver and checked the results against that.

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -50,6 +51,246 @@ namespace gko {
...
+     */
+    rhs,
+    /**
+     * the input is given
+     */
+    given
+};
+
+
+namespace {
```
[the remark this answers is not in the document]
yhmtsai: no, I do not want to export it to user. I will add the detail namespace

include/ginkgo/core/solver/solver_base.hpp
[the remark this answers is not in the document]
yhmtsai: do you mean the apply_impl virtual here and then apply_with_hint is only shown in EnableApplyHint?

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -50,6 +51,246 @@ namespace gko {
...
+    const DerivedType* apply_with_hint(const LinOp* alpha, const LinOp* b,
+                                       const LinOp* beta, LinOp* x,
+                                       input_hint hint) const
+    {
+        ApplyHint::apply_with_hint(alpha, b, beta, x, hint);
+        return self();
+    }
+
+protected:
+    GKO_ENABLE_SELF(DerivedType);
```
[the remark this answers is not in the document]
yhmtsai: you mean the `protected:` not `ENABLE_SELF()`, right?

core/solver/ir.cpp
```diff
@@ -160,21 +160,31 @@ std::unique_ptr<LinOp> Ir<ValueType>::conj_transpose() const
 
 template <typename ValueType>
 void Ir<ValueType>::apply_impl(const LinOp* b, LinOp* x) const
+{
+    this->apply_with_guess(b, x, this->get_apply_guess());
+}
+
+
+template <typename ValueType>
+void Ir<ValueType>::apply_with_guess(const LinOp* b, LinOp* x,
```
upsj, line 169: I don't think we need to save on characters here: `guess` doesn't really have a consistent meaning for solvers, `initial_guess` does
```suggestion
void Ir<ValueType>::apply_with_initial_guess(const LinOp* b, LinOp* x,
```

core/solver/multigrid.cpp
upsj: agree, though it is simpler to maintain if you have separate enum parameters for the different options.

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -50,6 +51,142 @@ namespace gko {
...
+        modify_input(b, x, guess);
+        this->apply_impl(alpha, b, beta, x);
+    }
+
+    /**
+     * Get the input guess
+     *
+     * @return initial_guess
+     */
+    initial_guess get_apply_guess() const { return guess_; }
```
upsj, line 162: I think we can be more specific in this naming. Since we only have initial guesses for `apply`, we don't need to prefix it here.
```suggestion
    initial_guess get_default_initial_guess() const { return guess_; }
```

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -50,6 +51,142 @@ namespace gko {
...
+
+class MultigridState;
+
+
+/**
+ * ApplyGuess provides a way to give the input guess for apply function.
+ * All functionalities have the protected access specifier. It should only be
+ * used internally.
+ */
+class ApplyGuess {
```
upsj, line 81:
```suggestion
class ApplyWithInitialGuess {
```

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -50,6 +51,142 @@ namespace gko {
...
+    friend class MultigridState;
+
+    /**
+     * Modify the input vector x by the guess
+     *
+     * @param b  the right hand side vectors
+     * @param x  the input vectors
+     * @param guess  the input guess
+     */
+    static void modify_input(const LinOp* b, LinOp* x, initial_guess guess)
```
upsj, line 92: Move to source? This is not a class template. Also naming suggestion:
```suggestion
    static void prepare_initial_guess(const LinOp* b, LinOp* x, initial_guess guess)
```

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -50,6 +51,142 @@ namespace gko {
...
+     */
+    zero,
+    /**
+     * the input is right hand side
+     */
+    rhs,
+    /**
+     * the input is given
+     */
+    given
```
upsj, line 69: nit: maybe `provided`?

core/solver/ir.cpp
```diff
@@ -192,52 +202,72 @@ void Ir<ValueType>::apply_dense_impl(const matrix::Dense<ValueType>* dense_b,
...
+            this->get_system_matrix()->apply(lend(neg_one_op), dense_x,
+                                             lend(one_op), lend(residual));
+            if (stop_criterion->update()
+                    .num_iterations(iter)
+                    .residual(residual_ptr)
+                    .solution(dense_x)
+                    .check(relative_stopping_id, true, &stop_status,
+                           &one_changed)) {
+                break;
+            }
```
upsj, line 255: I guess this kind of pattern is also interesting for the other solvers?

core/solver/ir.cpp
```diff
@@ -160,21 +160,31 @@ std::unique_ptr<LinOp> Ir<ValueType>::conj_transpose() const
...
+
+
+template <typename ValueType>
+void Ir<ValueType>::apply_with_guess(const LinOp* b, LinOp* x,
+                                     initial_guess guess) const
 {
     if (!this->get_system_matrix()) {
         return;
     }
+    this->modify_input(b, x, guess);
```
upsj, line 175: Why is this necessary? Doesn't apply_dense_impl already take care of that?

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -50,6 +51,142 @@ namespace gko {
...
+     * @return initial_guess
+     */
+    initial_guess get_apply_guess() const { return guess_; }
+
+    /**
+     * ApplyGuess constructor.
+     *
+     * @param guess  the input guess whose default is initial_guess::given
+     */
+    ApplyGuess(initial_guess guess = initial_guess::given) : guess_(guess) {}
```
upsj, line 169: `explicit`? Otherwise this can be constructed from an `initial_guess`

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -50,6 +51,142 @@ namespace gko {
 namespace solver {
 
 
+/**
+ * Give a initial guess about the input of the apply method.
+ */
+enum class initial_guess {
```
upsj, line 57: nit: This enum doesn't describe an initial guess, it describes how an initial guess is supposed to be chosen
```suggestion
enum class initial_guess_mode {
```

include/ginkgo/core/solver/solver_base.hpp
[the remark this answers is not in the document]
thoasm: Yes, I simply marked the other line as well to be sure it is also shown as part of my comment.

thoasm (approved):
I like the suggestions of @upsj, but the current implementation is already a huge improvement.

core/solver/ir.cpp
yhmtsai: no, apply_dense_impl only takes the information to eliminate the spmv computation.
but we still need to prepare the x with the correct value.
I modify the value before advanced_apply because I would like to make it give the same behavior as the default implementation.

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -50,6 +51,126 @@ namespace gko {
...
+    static void prepare_initial_guess(const LinOp* b, LinOp* x,
+                                      initial_guess_mode guess);
+
+    /**
+     * Applies a linear operator to a vector (or a sequence of vectors) with
+     * initial guess statement.
+     *
+     * Performs the operation x = op(b) with a initial guess statement, where op
+     * is this linear operator and initial guess statement will change the input
+     * vectors to fit the initial guess.
```
pratikvn, line 101:
```suggestion
     * is this linear operator and the initial guess parameter will modify the input
     * vector to the requested the initial guess mode (See @enum initial_guess_mode) .
```

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -50,6 +51,126 @@ namespace gko {
...
+    friend class MultigridState;
+
+    /**
+     * Modify the input vector x by the guess
+     *
+     * @param b  the right hand side vectors
+     * @param x  the input vectors
+     * @param guess  the input guess
+     */
+    static void prepare_initial_guess(const LinOp* b, LinOp* x,
```
pratikvn, line 92: Do you think there might be future cases where some solvers might want to do something specific to prepare the initial guesses and hence maybe this function could be virtual ?

pratikvn (approved):
LGTM!

core/solver/ir.cpp
upsj: That makes sense, thanks! I think in the long run, we should aim to eliminate the need for this, since we can fuse the initialization of `x` with other operations, but for now, that looks good.

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -50,6 +51,126 @@ namespace gko {
...
+    {}
+
+    /**
+     * set the default initial guess
+     *
+     * @param guess  the initial guess
+     */
+    void set_default_initial_guess(initial_guess_mode guess) { guess_ = guess; }
+
+    // override this at the same time when overridden
```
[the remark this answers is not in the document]
upsj: Does this actually interact with the `LinOp::apply_impl` or does it just lead to ambiguity? Without any subclass relation to `LinOp`, I'm not sure this does what you intend it to do

include/ginkgo/core/solver/solver_base.hpp
upsj: Since this is protected and needs to be called by the solver expliclty, I think we can handle this on the `apply_with_initial_guess` side, and theoretically don't need to keep this function.

core/solver/ir.cpp
```diff
@@ -160,21 +160,31 @@ std::unique_ptr<LinOp> Ir<ValueType>::conj_transpose() const
 
 template <typename ValueType>
 void Ir<ValueType>::apply_impl(const LinOp* b, LinOp* x) const
+{
+    this->apply_with_initial_guess(b, x, this->get_default_initial_guess());
+}
+
+
+template <typename ValueType>
+void Ir<ValueType>::apply_with_initial_guess(const LinOp* b, LinOp* x,
```
upsj, line 169: This matches the interface of `apply`, but doesn't validate the parameters or call `make_temporary_clone`

core/solver/ir.cpp
```diff
@@ -160,21 +160,31 @@ std::unique_ptr<LinOp> Ir<ValueType>::conj_transpose() const
...
+                                             initial_guess_mode guess) const
 {
     if (!this->get_system_matrix()) {
         return;
     }
+    this->prepare_initial_guess(b, x, guess);
```
upsj, line 175: By pulling this into the Lambda, we would no longer need to call `run` on it, since we already know its value type. In that case, we could even eliminate the function entirely by pulling its functionality into `apply_dense_impl` as I suggested above.

upsj (changes requested):
I thought about the dispatch mechanism a bit more, and would like to suggest some changes

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -50,6 +51,126 @@ namespace gko {
...
+     * @param guess  the input guess to handle the input vector(s)
+     *
+     * @return this
+     */
+    virtual void apply_with_initial_guess(const LinOp* b, LinOp* x,
+                                          initial_guess_mode guess) const
+    {
+        prepare_initial_guess(b, x, guess);
+        this->apply_impl(b, x);
+    }
```
upsj, line 114: This can go into the source, same with the other overload

include/ginkgo/core/solver/solver_base.hpp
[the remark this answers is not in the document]
yhmtsai: because all solver inherit this class, it provide the default behavior.
change the input x according to the initial_guess_mode.
By doing so, we do not need to reimplement all apply_with_initial_guess when there's no benefit, or we have no time to doing that. I do not want to do that in this pr. Otherwise, I would change the whole code interface three/four times before.

include/ginkgo/core/solver/solver_base.hpp
yhmtsai: I wouldn't implement it in each time and it is required for default implementation

upsj: My suggestion is: Move the code to `apply_with_initial_guess`, and maybe provide ValueType templated helpers for implementing this inside apply_dense_impl, but don't use it in solvers that already provide custom support for apply_with_initial_guess, because there you don't need the runtime dispatch.

include/ginkgo/core/solver/solver_base.hpp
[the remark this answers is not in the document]
yhmtsai: I think it is. https://godbolt.org/z/6YrhP6nKz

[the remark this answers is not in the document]
upsj: In any case, I think this is a rather confusing design, seeing as there is no relation between ApplyWithInitialGuess and LinOp - we are somewhere between a pure interface and an implementation here, and I'd rather see it only be one of those two things. So my suggestion is: Make all virtual functions in here pure, and instead implement the default behavior in a mixin, which can then rely on the base class being derived from LinOp.

include/ginkgo/core/solver/solver_base.hpp
upsj: If we move things to a mixin, this comment no longer applies

yhmtsai:
@upsj I add the mixin back. Besides temporary clone and validation, I also add the logger functionality (need add friend class in the DerivedType or EnableLogger). it's not in IterativeBase, each class need to inherit EnableApplyWith... and override the impl. I do not provide the default option because it will require eight if-else dynamic case for Dense and distributed::Dense.

tcojean (approved):

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -89,10 +89,16 @@ class ApplyWithInitialGuess {
...
+    template <typename VectorType>
+    static void prepare_initial_guess(const VectorType* b, VectorType* x,
+                                      initial_guess_mode guess)
+    {
+        if (guess == initial_guess_mode::zero) {
+            x->fill(zero<typename VectorType::value_type>());
+        } else if (guess == initial_guess_mode::rhs) {
+            x->copy_from(b);
+        }
+    }
```
upsj, line 101: The interface should only provide publically available operations. This looks more like a free function, which could probably move to a private header to reduce the amount of functions we freeze into the interface.

upsj (approved):

include/ginkgo/core/solver/solver_base.hpp
yhmtsai: your are right

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -155,18 +157,95 @@ class ApplyWithInitialGuess {
...
+ * object this mixin is used in, creating a clone on the correct executor if
+ * necessary.
+ *
+ * @tparam DerivedType  The type that this Mixin is used in. It must provide
+ *                      get_size() and get_executor() functions that return
+ *                      correctly initialized values and the logger
+ *                      functionality.
+ */
+template <typename DerivedType>
+class EnableApplyWithInitialGuess : public ApplyWithInitialGuess {
```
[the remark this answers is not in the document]
yhmtsai: I add it as TODO. it can do it now, but I don't like the way adding eight if-condition for dense and distributed_vector. Thus, I just put it as virtual now.

PR #986, opened 2022-03-12
Add a fixed coarsening class usable in multigrid and multi-level methods

pratikvn:
This PR adds a coarsening class, which allows generation of coarse matrices from user-defined coarsening arrays. The user needs to provide an array with the coarse indices (global indices).

### Features
+ Generation is very cheap compared to other multigrid methods, because the restriction column indices are the coarse indices provided. 
+ Additionally, the coarse matrix itself can be generated without SpGEMM by just creating them as submatrices from the provided coarse indices. This will be added in a future PR after the submatrix from IndexSet feature is complete on all executors.

### TODO
+ [ ] Discussion on expectation for users when providing sorted or un-sorted inputs for coarse_row vector.


test/multigrid/fixed_coarsening_kernels.cpp
```diff
@@ -0,0 +1,216 @@
...
+            .with_coarse_rows(d_coarse_rows)
+            .with_skip_sorting(true)
+            .on(exec);
+
+    auto mg_level = mg_level_factory->generate(system_mtx);
+    auto d_mg_level = d_mg_level_factory->generate(d_system_mtx);
+
+    GKO_ASSERT_MTX_NEAR(gko::as<Csr>(d_mg_level->get_prolong_op()),
+                        gko::as<Csr>(mg_level->get_prolong_op()),
+                        r<value_type>::value);
```
upsj, line 175:
```suggestion
                        0.0);
```

test/multigrid/fixed_coarsening_kernels.cpp
```diff
@@ -0,0 +1,216 @@
...
+    GKO_ASSERT_MTX_NEAR(gko::as<Csr>(d_mg_level->get_restrict_op()),
+                        gko::as<Csr>(mg_level->get_restrict_op()),
+                        r<value_type>::value);
```
upsj, line 178:
```suggestion
                        0.0);
```

test/multigrid/fixed_coarsening_kernels.cpp
```diff
@@ -0,0 +1,216 @@
...
+    GKO_ASSERT_MTX_NEAR(gko::as<Csr>(d_mg_level->get_coarse_op()),
+                        gko::as<Csr>(mg_level->get_coarse_op()),
+                        r<value_type>::value);
+}
+
+
+TEST_F(FixedCoarsening, GenerateMgLevelIsEquivalentToRefOnUnsortedMatrix)
```
upsj, line 185: Note: SpGEMM generally requires the second matrix to be sorted, so this only works "accidentally" because you compute it as R*(A*P) and not (R*A)*P

upsj (approved):
LGTM! Nice job building on top of what is already there :)

MarcelKoch:
Just a small note, you say that your implementation would be simpler, if you could use submatrix from index set. I think it would help in that case if you set the target branch of this PR to the submatrix from index set one. That way you can already use that here and won't need a second PR or update this one when the submatrix PR is merged.

core/multigrid/fixed_coarsening.cpp
```diff
@@ -0,0 +1,127 @@
...
+                    parameters_.coarse_rows.get_const_data(),
+                    restrict_op->get_col_idxs());
+    exec->run(fixed_coarsening::make_fill_array(restrict_op->get_values(),
+                                                coarse_dim, one<ValueType>()));
+    exec->run(fixed_coarsening::make_fill_seq_array(restrict_op->get_row_ptrs(),
+                                                    coarse_dim + 1));
+
+    auto prolong_op = gko::as<csr_type>(share(restrict_op->transpose()));
+
+    // TODO: Can be done with submatrix index_set.
```
upsj, line 108: Looks like my initial comment was swallowed by ~~the cookie monster~~ Github's servers. I don't think that is true in the general setting, since index_set has no ordering. The current implementation seems fine to me, does it need to be changed?

include/ginkgo/core/multigrid/fixed_coarsening.hpp
```diff
@@ -0,0 +1,149 @@
...
+                                       system_matrix->get_size()),
+          EnableMultigridLevel<ValueType>(system_matrix),
+          parameters_{factory->get_parameters()},
+          system_matrix_{system_matrix}
+    {
+        if (system_matrix_->get_size()[0] != 0) {
+            // generate on the existing matrix
+            this->generate();
+        }
+    }
```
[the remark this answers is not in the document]
pratikvn: That is true. But do we want to duplicate the declarations here and in the source file ? I guess there is a case to be made to keep the implementations in the source files to reduce the compilation time, but is there another reason ?

core/multigrid/fixed_coarsening.cpp
pratikvn: Yes, that is one thing I wanted to discuss. Do we want the possibility to have implicit re-ordering when specifying the coarse row indices ?

test/multigrid/fixed_coarsening_kernels.cpp
pratikvn: Good catch! Looks like I forgot to set the flag to false.

pratikvn:
@MarcelKoch, I thought it would be better to merge this into first and then work on the submatrix device kernels, as #964 does not yet have device kernels for submatrix creation.

include/ginkgo/core/multigrid/fixed_coarsening.hpp
```diff
@@ -0,0 +1,150 @@
...
+#include <ginkgo/core/multigrid/multigrid_level.hpp>
+
+namespace gko {
+namespace multigrid {
+
+
+/**
+ * FixedCoarsening is a very simple coarse grid generation algorithm. It selects
+ * the coarse matrix from the fine matrix by either constant jumps or with a
+ * user-specified index_set of rows.
```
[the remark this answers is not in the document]
pratikvn: Thanks, I forgot to update the documentation. This Coarsening class just takes in the indices from the user to construct a coarse matrix from that. The option of having jumps, will be a different class, as you can see in PR #979. But that is still a WIP.

PR #989, opened 2022-03-21
Add CI test, adjust the compiler version to check the lowest requirement, fix the corresponding error

yhmtsai:
This pr adds the dummy implementation in fft_kernels_stub when hipfft is not existed which is hip < 4.1
adds more ci tests with different hip versions to keep track of our hip support.
Moreover, it disables cmake_hip_architecture in cmake 3.21

Fix note:
- Use constructor not uninitialized list in zip_iterator. 
When gcc 5 can not use uninitialized list to construct tuple with different type. After 5, compiler might directly use the constructor. small example: https://godbolt.org/z/G4TrcKs6v the commented part is not allowed in gnu 5
- Relax error bound in sum_duplication of device_matrix_data
When Hip with gnu < 7, the error is larger than other combination.
I test it in HIP 3.5 with gnu 5, 6, 7, 8 and HIP 4.5 with gnu 6, 8.
- HipFFT 5.1 issue on 2d construction: detail in https://github.com/ROCmSoftwarePlatform/rocFFT/issues/353
TODO:
- [x] merge the ginkgo-container pr
- [x] choose combination of setting


upsj (approved):
LGTM! Thanks for fixing this. Do we really need to go down to HIP 3.5?

yhmtsai:
Our current ReadME puts the support Rocm-3.5+, so I prefer one test for ensuring the oldest version work.
I do not know how much divergence among these versions.

.gitlab-ci.yml
```diff
@@ -560,6 +561,38 @@ build/amd/openmpi/clang/hip/release/static:
...
+    - gpu
+  image: localhost:5000/hpccm_hip_d93a513e_gko-hip502-openmpi-gnu8-llvm11
+  variables:
+    <<: *default_variables
+    BUILD_OMP: "ON"
+    BUILD_HIP: "ON"
+    RUN_EXAMPLES: "ON"
+    BUILD_TYPE: "Debug"
+
+build/amd/openmpi/gcc/hip35/release/shared:
```
tcojean, line 580: Can you make this static? Or the other? To increase variety.

.gitlab-ci.yml
```diff
@@ -560,6 +561,38 @@ build/amd/openmpi/clang/hip/release/static:
...
 
+build/amd/openmpi/gcc/hip502/debug/shared:
+  <<: *default_build_with_test
+  extends:
+    - .quick_test_condition
+  tags:
+    - private_ci
+    - amdci
+    - gpu
+  image: localhost:5000/hpccm_hip_d93a513e_gko-hip502-openmpi-gnu8-llvm11
```
tcojean, line 572: Isn't 5.0 already tested with the other HIP job (which would be bleeding edge?). Also, maybe add the versioning to line 547. If that's the case, maybe testing 4.5 here instead makes more sense than having two 5.x?

.gitlab-ci.yml
```diff
@@ -560,6 +561,38 @@ build/amd/openmpi/clang/hip/release/static:
     BUILD_TYPE: "Release"
     BUILD_SHARED_LIBS: "OFF"
 
+build/amd/openmpi/gcc/hip502/debug/shared:
+  <<: *default_build_with_test
+  extends:
+    - .quick_test_condition
```
tcojean, line 567: For production, should this stay quick test? Maybe extended instead?

tcojean (approved):
LGTM.

Honestly, with the HIP ecosystem moving so much I don't think we need to keep supporting up to 3.5. Since it seems to work well for now we can keep this, but if we ever get any issue or missing feature we need from 3.5, we should definitely drop it.

test/test_install/CMakeLists.txt
```diff
@@ -58,6 +58,9 @@ if(GINKGO_BUILD_HIP)
         set (GINKGO_PIC_OPTION "-fPIC")
         set (GINKGO_CUDA_PIC_OPTION "-Xcompiler '-fPIC'")
     endif()
+    if (NOT GINKGO_BUILD_SHARED_LIBS AND CMAKE_BUILD_TYPE MATCHES [Dd][Ee][Bb][Uu][Gg])
+        set (GINKGO_PIC_OPTION "-fPIC")
+    endif()
```
upsj, line 63: This does not play nice with multi-config generators. Using generator expressions would be the correct thing here, but that probably doesn't work with HIP (I tried configuring HIP with `-G"Ninja Multi-Config"` before, and that didn't work)

upsj (approved):
LGTM! We could also consider adding a CUDA/ROCm ensemble build. I've been developing with this setup for a while, and it works quite well :)

.gitlab-ci.yml
```diff
@@ -675,20 +746,6 @@ build/nocuda-nomixed/nompi/clang/omp/debug/static:
     BUILD_SHARED_LIBS: "OFF"
     MIXED_PRECISION: "OFF"
 
-build/nocuda-nomixed/nompi/intel/omp/release/static:
```
tcojean, line 678: Intel was replaced with clang? Is there a specific reason?

.gitlab-ci.yml
```diff
@@ -635,19 +679,46 @@ build/nocuda/openmpi/clang/omp/debug/static:
...
+  extends:
+    - .quick_test_condition
+    - .use_gko-nocuda-mvapich2-gnu5-llvm39-intel2018
+  variables:
+    <<: *default_variables
     BUILD_OMP: "ON"
     BUILD_TYPE: "Release"
     BUILD_SHARED_LIBS: "OFF"
 
+build/nocuda-nomixed/nompi/clang/omp/release/static:
```
tcojean, line 708: see this

tcojean (approved):
LGTM

test/test_install/CMakeLists.txt
yhmtsai: I update it with $<$<CONFIG:Debug>:-fPIC>

.gitlab-ci.yml
yhmtsai: because we don't have test for llvm3.9 and we test intel in `build/nocuda/nompi/intel/core/debug/shared`, I change it to clang

PR #990, opened 2022-03-21
Add Dense reduction kernels using temporary Array

upsj:
This adds another version of run_kernel_reduction that reuses an existing array for temporary storage to avoid reallocations. The same parameter is being exposed via Dense reduction member functions. The PR also unifies the run_kernel_* implementations between CUDA and HIP.

~~I removed the sparselib kernels for simplicity, I am currently testing whether we can improve the tuning parameters of the run_kernel_reduction kernels to get rid of them entirely.~~


pratikvn:
Is the plan to remove the sparselib kernel here and merge the tuning PR first ?  Maybe we keep it until the tuning has been completed, because it looks like that PR is in an incomplete state.

upsj:
Yes, I was kind of lazy there and hoped I would manage to easily fit the tuning into this PR. I'll revert the changes

hip/test/base/kernel_launch.hip.cpp
```diff
@@ -350,67 +412,99 @@ TEST_F(KernelLaunch, Reduction2D) { run2d_reduction(exec); }
...
-            return (i + 1) * (j + 1);
-        },
-        [] GKO_KERNEL(auto i, auto j) { return i + j; },
-        [] GKO_KERNEL(auto j) { return j * 2; }, int64{}, output.get_data(),
-        gko::dim<2>{static_cast<size_type>(num_rows),
-                    static_cast<size_type>(num_cols)},
-        output);
+    // empty, most threads idle, most threads busy, multiple blocks
+    for (auto num_rows : {0, 10, 100, 1000, 10000}) {
+        // check different edge cases: subwarp sizes, blocked mode
```
pratikvn, line 466: Nice!

test/matrix/dense_kernels.cpp
```diff
@@ -998,18 +998,135 @@ TEST_F(Dense, ExtractDiagonalOnShortFatIntoDenseCrossExecutor)
...
+
+
+TEST_F(Dense, ComputeDotWithTmpIsEquivalentToRef)
+{
+    set_up_vector_data(40);
+
+    auto dot_size = gko::dim<2>{1, x->get_size()[1]};
+    auto dot_expected = Mtx::create(ref, dot_size);
+    auto ddot = Mtx::create(ref, dot_size);
+    gko::Array<char> tmp{ref};
```
pratikvn, line 1024: Maybe have a test with properly allocated `tmp` ?

pratikvn (commented):
Nice work! Looks mostly good. Some minor comments.

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -1012,64 +1012,64 @@ class Dense
     /**
      * @copydoc scale(const LinOp *)
      *
-     * @note  Other implementations of dense should override this function
-     *        instead of scale(const LinOp *alpha).
+     * @deprecated  This function will be removed in the future,
```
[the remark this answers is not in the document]
upsj: Yeah, I was also a bit unsure about this. My argument here is: Even if you derive from dense, every class taking Dense inputs will use `values_` and `stride_` to determine how the data will be accessed, so there is nothing sensible behavior-wise that a user could add here. It might make sense to add a logger event to member functions like these (not individual events but a generic linop_member_called (the name is bad, I know 😆 ) with a string parameter for the actual function name.

core/matrix/dense.cpp
```diff
@@ -269,6 +334,22 @@ void Dense<ValueType>::sub_scaled_impl(const LinOp* alpha, const LinOp* b)
...
+    auto local_res = make_temporary_clone(exec, result);
+    auto dense_b = make_temporary_conversion<ValueType>(local_b.get());
+    auto dense_res = make_temporary_conversion<ValueType>(local_res.get());
+    exec->run(
+        dense::make_compute_dot(this, dense_b.get(), dense_res.get(), tmp));
+}
+
+
 template <typename ValueType>
 void Dense<ValueType>::compute_dot_impl(const LinOp* b, LinOp* result) const
```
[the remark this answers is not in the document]
upsj: yes, it's a virtual function, so we can't simply remove it for interface reasons

common/cuda_hip/base/kernel_launch_reduction.hpp.inc
```diff
@@ -0,0 +1,534 @@
...
+                                 ValueType* result, size_type size,
+                                 Array<char>& tmp, KernelArgs&&... args)
+{
+    constexpr int oversubscription = 16;
+    constexpr auto block_size = default_block_size;
+    const auto num_blocks = std::min<int64>(
+        ceildiv(size, block_size), exec->get_num_warps() * oversubscription);
+    if (num_blocks > 1) {
+        const auto required_storage = sizeof(ValueType) * num_blocks;
+        tmp.set_executor(exec);
```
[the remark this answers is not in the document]
upsj: I could do this either on the `core` side or the kernel side. If users pass in an Array on the wrong executor, I need to somehow turn it into a correct one. Either by checking on the host side and creating a different array if necessary, or by changing the executor of the input array. Also by set_executor we might copy over data, so `tmp.clear(); tmp.set_executor(exec);` might be more appropriate?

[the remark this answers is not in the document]
pratikvn: I mean this kind of transparently transfers the data to the executor, with allocations, which the user probably did not intend to do, if they actually used the advanced version of the `_impl` function where they explicitly passed in a temp array. So, I am suggesting we throw instead of continue if there is an executor mismatch.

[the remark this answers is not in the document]
upsj: I think this is probably not the right point to change the executor, so I'll move the operation to `core`. If `tmp` was initialized without executor, we would still need to add some special-case treatment code. How about `if (exec != array.get_executor()) { array.clear(); array.set_executor(exec); }`?

common/cuda_hip/base/kernel_launch_reduction.hpp.inc
```diff
@@ -0,0 +1,534 @@
...
+        if (tmp.get_num_elems() < required_storage) {
+            tmp.resize_and_reset(required_storage);
```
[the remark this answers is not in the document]
pratikvn: Yeah, I guess that is true. That is probably enough.

core/matrix/dense.cpp
```diff
@@ -269,6 +334,26 @@ void Dense<ValueType>::sub_scaled_impl(const LinOp* alpha, const LinOp* b)
...
+    if (tmp.get_executor() != exec) {
+        tmp.clear();
+        tmp.set_executor(exec);
+    }
+    auto local_b = make_temporary_clone(exec, b);
+    auto local_res = make_temporary_clone(exec, result);
+    auto dense_b = make_temporary_conversion<ValueType>(local_b.get());
+    auto dense_res = make_temporary_conversion<ValueType>(local_res.get());
+    exec->run(
+        dense::make_compute_dot(this, dense_b.get(), dense_res.get(), tmp));
```
pratikvn, line 353: `make_compute_dot_dispatch` ? Same for other functions.

cuda/test/base/kernel_launch.cu
```diff
@@ -51,15 +51,42 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "core/test/utils.hpp"
 
 
-namespace {
-
-
```
yhmtsai, line 56: the namespace is deleted

dpcpp/solver/cb_gmres_kernels.dp.cpp
```diff
@@ -1006,13 +1006,14 @@ void initialize_2(std::shared_ptr<const DpcppExecutor> exec,
...
     const auto stride_arnoldi = arnoldi_norm->get_stride();
+    Array<char> tmp{exec};
 
     initialize_2_1_kernel<block_size>(
         grid_dim_1, block_dim, 0, exec->get_queue(), residual->get_size()[0],
         residual->get_size()[1], krylov_dim, krylov_bases,
         residual_norm_collection->get_values(),
         residual_norm_collection->get_stride());
-    kernels::dpcpp::dense::compute_norm2(exec, residual, residual_norm);
+    kernels::dpcpp::dense::compute_norm2(exec, residual, residual_norm, tmp);
```
yhmtsai, line 1016: same here

dpcpp/test/base/kernel_launch.dp.cpp
```diff
@@ -51,15 +51,42 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "core/test/utils.hpp"
 
 
-namespace {
-
-
```
yhmtsai, line 56: namespace

hip/test/base/kernel_launch.hip.cpp
```diff
@@ -51,15 +51,42 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "core/test/utils.hpp"
 
 
-namespace {
-
-
```
yhmtsai, line 56: namespace

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -786,12 +784,19 @@ class Dense
      *                (the number of columns in the vector must match the number
      *                of columns of this)
      */
-    void compute_norm2(LinOp* result) const
-    {
-        auto exec = this->get_executor();
-        this->compute_norm2_impl(
-            make_temporary_output_clone(exec, result).get());
-    }
+    void compute_norm2(LinOp* result) const;
```
yhmtsai, line 787: this is not virtual, so you can make the tmp = {} with default?

omp/base/kernel_launch_reduction.hpp
```diff
@@ -197,7 +204,8 @@ void run_kernel_row_reduction_impl(std::shared_ptr<const OmpExecutor> exec,
                                    KernelFunction fn, ReductionOp op,
                                    FinalizeOp finalize, ValueType identity,
                                    ValueType* result, size_type result_stride,
-                                   dim<2> size, MappedKernelArgs... args)
+                                   dim<2> size, Array<char>& tmp,
+                                   MappedKernelArgs... args)
```
yhmtsai, line 208: also &&?

omp/test/base/kernel_launch.cpp
```diff
@@ -51,15 +51,42 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "core/test/utils.hpp"
 
 
-namespace {
-
-
```
yhmtsai, line 56: namespace

test/matrix/dense_kernels.cpp
```diff
@@ -998,18 +998,203 @@ TEST_F(Dense, ExtractDiagonalOnShortFatIntoDenseCrossExecutor)
...
+    dx->compute_conj_dot(y.get(), ddot.get(), tmp);
+
+    GKO_ASSERT_MTX_NEAR(ddot, dot_expected, r<vtype>::value);
+}
+
+
 TEST_F(Dense, ComputeNorm1IsEquivalentToRef)
 {
-    set_up_apply_data();
+    set_up_vector_data(2);
```
yhmtsai, line 1103: it should keep () to test one vector

test/matrix/dense_kernels.cpp
```diff
@@ -998,18 +998,203 @@ TEST_F(Dense, ExtractDiagonalOnShortFatIntoDenseCrossExecutor)
 }
 
 
+TEST_F(Dense, ComputeDotIsEquivalentToRef)
+{
+    set_up_vector_data(2);
```
yhmtsai, line 1003: need () to test one vector

test/matrix/dense_kernels.cpp
```diff
@@ -998,18 +998,203 @@ TEST_F(Dense, ExtractDiagonalOnShortFatIntoDenseCrossExecutor)
...
+    x->compute_norm1(norm_expected.get());
+    dx->compute_norm1(dnorm.get(), tmp);
+
+    GKO_ASSERT_MTX_NEAR(norm_expected, dnorm, r<vtype>::value);
+}
+
+
+TEST_F(Dense, ComputeNorm2IsEquivalentToRef)
+{
+    set_up_vector_data(5);
```
yhmtsai, line 1153: same here

omp/base/kernel_launch_reduction.hpp
```diff
@@ -290,7 +298,7 @@ void run_kernel_col_reduction_sized_impl(
     syn::value_list<int, remainder_cols>,
     std::shared_ptr<const OmpExecutor> exec, KernelFunction fn, ReductionOp op,
     FinalizeOp finalize, ValueType identity, ValueType* result, dim<2> size,
-    MappedKernelArgs... args)
+    Array<char>& tmp, MappedKernelArgs... args)
```
yhmtsai, line 301:
```suggestion
    Array<char>& tmp, MappedKernelArgs&&... args)
```

upsj: this is deliberate, as MappedKernelArgs need to be value types, not references

cuda/test/base/kernel_launch.cu
upsj: this is an executable, not a library, so the unnamed namespace has no effect on the execution of the program

include/ginkgo/core/matrix/dense.hpp
upsj: Keeping them separate makes it easier to write separate documentation for them. Also I'm not sure you can have a non-const lvalue reference parameter with default value anyways.

test/matrix/dense_kernels.cpp
upsj: good point, at least for (conj_)dot and norm2 we should test 1 rhs with stride == 1

test/matrix/dense_kernels.cpp
upsj: this uses common kernels, so its behavior for 1 rhs is already being tested elsewhere.

upsj:
@yhmtsai Different executors use different amounts of temporary storage, so I don't think it makes sense to reuse the previous size. It is basically intended as a way for the user to provide an opaque array that they themselves will not touch.

cuda/test/base/kernel_launch.cu
yhmtsai: I know. It is from gtest example and we have it everywhere in ginkgo. but it is not important.

yhmtsai:
@upsj I think cuda/hip and maybe dpcpp have the same amount of workspace?
omp and reference might not need it.

core/matrix/dense.cpp
```diff
@@ -291,8 +397,25 @@ void Dense<ValueType>::compute_conj_dot_impl(const LinOp* b,
     auto exec = this->get_executor();
     auto dense_b = make_temporary_conversion<ValueType>(b);
     auto dense_res = make_temporary_conversion<ValueType>(result);
-    exec->run(dense::make_compute_conj_dot_dispatch(this, dense_b.get(),
-                                                    dense_res.get()));
+    Array<char> tmp{exec};
+    exec->run(dense::make_compute_conj_dot(this, dense_b.get(), dense_res.get(),
```
yhmtsai, line 401: dispatch?

core/matrix/dense.cpp
yhmtsai: is it missing dispatch?

yhmtsai (approved):
some dispatch are still missing?

core/matrix/dense.cpp
upsj: no, the corresponding operation calls _dispatch, see the beginning of the file. Do you think I should change it here as well? I wanted to keep the diff small, but that of course may create some confusion. I'd assume once #991 is executed and merged, we could probably reverse this?

upsj:
@yhmtsai The amount of storage depends on device characteristics, so it is most likely different even between different CudaExecutors, let alone between CUDA and HIP/SYCL/OpenMP. Only Reference uses no temporary storage at all.

cuda/test/base/kernel_launch.cu
upsj: If there is no reason for adding it, I don't think we need to keep it. [This SO question](https://stackoverflow.com/questions/47861534/why-does-google-test-sample-put-tests-in-an-anonymous-namespace) discusses the issue, but I don't really think the answer applies to us at all. Symbol visibility is a problem that we will deal with in the future #715, but it does not have any impact here (except for maybe a miniscule difference in link time because the symbol table of the object files is bigger)

core/matrix/dense.cpp
yhmtsai: I see. I do not notice the function is point to the dispatch one.
it should be okay

PR #997, opened 2022-03-24
Adds move_from to PolymorphicObject

MarcelKoch:
This PR adds `move_from(PolymorphicObject*)` to PolymorphicObject. It enables us to use the generic `copy_from` mechanisms for moves even if the move-from object is not available as a unique_ptr. That situation can appear if one tries to move from an object that is stored as a shared_ptr. (I know that moving from an object inside a shared_ptr is generally not a good idea, but I currently have no other option)

### DIFF
--- a/include/ginkgo/core/base/polymorphic_object.hpp
+++ b/include/ginkgo/core/base/polymorphic_object.hpp
@@ -208,6 +208,27 @@ public:
     }
 
     /**
+     * Moves another object into this object.
+     *
+     * This is the polymorphic equivalent of the move assignment operator.
+     *
+     * @see move_from_impl(std::unique_ptr<PolymorphicObject>)
+     *
+     * @param other  the object to move from
+     *
+     * @return this
+     */
+    PolymorphicObject* move_from(std::unique_ptr<PolymorphicObject> other)
+    {
+        this->template log<log::Logger::polymorphic_object_move_started>(
+            exec_.get(), other.get(), this);
+        auto copied = this->copy_from_impl(std::move(other));
+        this->template log<log::Logger::polymorphic_object_move_completed>(
+            exec_.get(), other.get(), this);
+        return copied;
+    }
+
+    /**
      * Transforms the object into its default state.
      *
      * Equivalent to `this->copy_from(this->create_default())`.
@@ -293,6 +314,17 @@ protected:
 
     /**
      * Implementers of PolymorphicObject should implement this function instead
+     * of move_from(std::unique_ptr<PolymorphicObject>).
+     *
+     * @param other  the object to move from
+     *
+     * @return this
+     */
+    virtual PolymorphicObject* move_from_impl(
+        std::unique_ptr<PolymorphicObject> other) = 0;
+
+    /**
+     * Implementers of PolymorphicObject should implement this function instead
      * of clear().
      *
      * @return this
@@ -366,6 +398,12 @@ public:
     AbstractObject* move_from(PolymorphicObject* other)
     {
         return static_cast<AbstractObject*>(this->move_from_impl(other));
+    }
+
+    AbstractObject* move_from(std::unique_ptr<PolymorphicObject> other)
+    {
+        return static_cast<AbstractObject*>(
+            this->move_from_impl(std::move(other)));
     }
 
     AbstractObject* clear()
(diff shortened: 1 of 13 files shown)


MarcelKoch:
Perhaps to explain a bit, I'm trying to move from an object that is stored as a shared_ptr. Of course, there is no direct conversion from shared_ptr to unique_ptr, so I can't use the existing `copy_from` overload easily. I also can't just create a unique_ptr from the shared_ptr like `std::unique_ptr<LinOp>(shared.get())`, since then both objects would try to delete the underlying pointer. I've now realized, that I could get the correct behavior also with the overload
```
void copy_from(std::unique_ptr<PolymorphicObject, null_deleter<PolymorphicObject> other)
```
Perhaps that is a better choice, since you can't easily call that by accident.

include/ginkgo/core/base/polymorphic_object.hpp
```diff
@@ -332,6 +363,11 @@ class EnableAbstractPolymorphicObject : public PolymorphicBase {
             this->copy_from_impl(std::move(other)));
     }
 
+    AbstractObject* move_from(PolymorphicObject* other)
+    {
+        return static_cast<AbstractObject*>(this->move_from_impl(other));
```
upsj, line 368: Is the logger being called if you call move_from/copy_from on a concrete object? Maybe we should test that behavior?

upsj (approved):
LGTM!

include/ginkgo/core/base/polymorphic_object.hpp
```diff
@@ -186,6 +186,27 @@ class PolymorphicObject : public log::EnableLogging<PolymorphicObject> {
...
+     *
+     * @see move_from_impl(PolymorphicObject *)
+     *
+     * @param other  the object to move from
+     *
+     * @return this
+     */
+    PolymorphicObject* move_from(PolymorphicObject* other)
+    {
+        this->template log<log::Logger::polymorphic_object_copy_started>(
```
pratikvn, line 202: Should we still be using the `*_copy_started` masks for move as well ? As it is not a deep copy, I think it might make sense to add different masks like `*_move_started` and *_move_completed`.

pratikvn (approved):
LGTM!

yhmtsai:
Could you also add the more desciption when you use this move_from the shared ptr?

MarcelKoch:
@yhmtsai this appeared in handling the convert_to/move_to for the distributed matrix with runtime local matrix types. Currently we still store the local matrices as shared pointers, but perhaps that is something to reevaluate. Still, I think having an additional (and clear) move_from would be helpful.

upsj:
Putting them into a `unique_ptr` and taking a `unique_ptr` parameter would force you to write conforming code directly. But I think this code should be useful anyways.

MarcelKoch:
Should I also add an overload `move_from(std::unique_ptr<PolymorphicObject>)`? I think that might be reasonable, because I personally didn't even consider `copy_from(std::unique_ptr)` to do a move due to its name. Of course the documentation says otherwise, but the name made me not consider it.

include/ginkgo/core/log/logger.hpp
```diff
@@ -458,6 +458,30 @@ public:                                                              \
...
+                              const Executor* exec,
+                              const PolymorphicObject* input,
+                              const PolymorphicObject* output)
+
+    /**
+     * PolymorphicObject's move completed event.
+     *
+     * @param exec  the executor used
+     * @param input  the PolymorphicObject to be move from
+     * @param output  the PolymorphicObject to be move to
```
pratikvn, line 478: nit:
```suggestion
     * @param input  the PolymorphicObject to be moved from
     * @param output  the PolymorphicObject to be moved into
```

pratikvn:
Just an additional note: You need to add the event to the [`logger.cpp` file as well.](https://github.com/ginkgo-project/ginkgo/blob/develop/core/log/logger.cpp#L52=)

include/ginkgo/core/base/polymorphic_object.hpp
MarcelKoch: No, the logger is not called here. I think this is a general pattern in our `Enable*` mixins. I'm not fixing this here, instead I will open a new pr which addresses all instances.

tcojean (approved):

PR #1007, opened 2022-04-01
Distributed matrix kernels

MarcelKoch:
This PR adds device kernels for the distributed matrix `read_distributed`. Currently, the implementation only supports cuda/hip using thrust.

Todo:
- [x] ~dpcpp implementation~ we will postpone that
- [x] ~perhaps update openMP~ would need STL with OMP support for that


MarcelKoch:
format!

common/cuda_hip/distributed/matrix_kernels.hpp.inc
```diff
@@ -0,0 +1,271 @@
...
+        };
+    auto map_to_local_col =
+        [col_range_bounds, col_range_starting_indices] __host__ __device__(
+            const GlobalIndexType col, const size_type range_id) {
+            return static_cast<LocalIndexType>(col -
+                                               col_range_bounds[range_id]) +
+                   col_range_starting_indices[range_id];
+        };
+
+    using input_type = thrust::tuple<GlobalIndexType, GlobalIndexType,
```
upsj, line 121: The code might become easier to read if you use structs insted?

omp/distributed/matrix_kernels.cpp
```diff
@@ -236,10 +236,10 @@ void build_diag_offdiag(
         local_to_global_ghost.get_data()[local_idx] = global_idx;
     }
 
-    // shift recv_offsets to the back, insert 0 in front again
-    LocalIndexType local_prev{};
-    for (size_type i = 0; i <= num_parts; i++) {
-        recv_offsets[i] = std::exchange(local_prev, recv_offsets[i]);
+    // shift recv_sizes to the back, insert 0 in front again
```
upsj, line 239: that doesn't seem to entirely match what is happening any more

upsj (approved):
LGTM! Really nice job! I will probably try get rid of the lower_bounds and upper_bounds calls with some precomputation, but that can be done independently.

common/cuda_hip/distributed/matrix_kernels.hpp.inc
MarcelKoch: Honestly, I'm not sure how to incorporate such a struct in thrust. Although better readability would be nice, I don't think the effort of figuring out how to use it is worth it.

upsj: something like
```cpp
struct input {
    GlobalIndexType row;
    GlobalIndexType col;
    ValueType val;
    size_type row_range;
    size_type col_range;

    __forceinline__ __device__ __host__ input(thrust::tuple<GlobalIndexType, GlobalIndexType, ...>);
};
```

upsj: it should probably be at file scope though, thrust doesn't like function-scoped structs.

MarcelKoch: I think this is the only instance where this makes sense, right? The other cases use smaller tuples, so it should be easier to decipher.

tcojean (commented):
Some simple comments

MarcelKoch:
format!

reference/distributed/matrix_kernels.cpp
```diff
@@ -100,108 +92,96 @@ void build_local_nonlocal(
...
+
+    // 2. remove duplicate columns, now the new column i has global index
+    // unique_columns[i]
+    unique_columns.erase(
+        std::unique(unique_columns.begin(), unique_columns.end()),
+        unique_columns.end());
+
+    // 3. create mapping from unique_columns
+    unordered_map<GlobalIndexType, LocalIndexType> non_local_column_map(exec);
+    for (std::size_t i = 0; i < unique_columns.size(); ++i) {
```
thoasm, line 155: nit:
```suggestion
    for (size_type i = 0; i < unique_columns.size(); ++i) {
```

test/mpi/distributed/matrix.cpp
```diff
@@ -51,131 +51,274 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+    gko::mpi::communicator comm;
+    std::shared_ptr<Partition> row_part;
+    std::shared_ptr<Partition> col_part;
+
+    gko::matrix_data<value_type, global_index_type> mat_input;
+    std::array<matrix_data, 3> dist_input;
+
+    std::unique_ptr<dist_mtx_type> dist_mat;
+
+    std::default_random_engine engine;
```
thoasm, line 127: nit: you are missing a couple of header includes in this file (which are all transitively included, but some of them would be better to include explicitly):
- `array`
- `memory`
- `random`
- `ginkgo/core/base/array.hpp`
- `ginkgo/core/base/matrtix_data.hpp`

thoasm (approved):
LGTM!

common/cuda_hip/distributed/matrix_kernels.hpp.inc
```diff
@@ -0,0 +1,290 @@
...
+        non_local_global_col_idxs_begin + num_non_local_cols,
+        thrust::make_zip_iterator(thrust::make_tuple(
+            non_local_col_part_ids.get_data(), permutation.get_data())));
+
+    // 4. map column index of non-local entries to new columns
+    non_local_col_idxs.resize_and_reset(num_non_local_elements);
+    array<size_type> lower_bounds{exec, num_non_local_elements};
+    // I have to precompute the lower bounds because the calling binary
+    // searches from the device does not work:
+    // https://github.com/NVIDIA/thrust/issues/1415
```
pratikvn, line 251: Maybe add a TODO here so that we know to fix it once it has been fixed from thrust's side ?

pratikvn (approved):
LGTM!

common/cuda_hip/distributed/matrix_kernels.hpp.inc
```diff
@@ -0,0 +1,287 @@
...
+                        col_range_bounds + num_col_ranges + 1,
+                        input.get_const_col_idxs(),
+                        input.get_const_col_idxs() + num_input_elements,
+                        col_range_ids.get_data());
+
+    // count number of diag<0> and offdiag<1> elements
+    auto range_ids_it = thrust::make_zip_iterator(thrust::make_tuple(
+        row_range_ids.get_const_data(), col_range_ids.get_const_data()));
+    auto num_elements_pair = thrust::transform_reduce(
+        thrust::device, range_ids_it, range_ids_it + num_input_elements,
```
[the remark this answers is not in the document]
upsj: that would mean adding a Thrust dependency for non-CUDA builds. Their OpenMP backend is also not particularly well tuned/maintained.

common/cuda_hip/distributed/matrix_kernels.hpp.inc
pratikvn: Yeah, I think that is okay. A TODO tag just allows us to keep track of possible improvements. :)

upsj: yes, that would make us very future-proof ;)

MarcelKoch: As far as I understood the issue I've linked to, they plan to deprecate the device-side launch of lower bounds anyway, so it probably will never be possible to do. But it seems very unclear how they actually want to proceed with it, so I will add the TODO.

upsj: you can also use our own binary search kernels in searching.cuh/hpp.inc. But I would like to do some preprocessing for this operation anyways :)

tcojean (approved):

MarcelKoch:
format!

PR #1008, opened 2022-04-01
Add functions to wrap `const` user arrays into Ginkgo batched matrices

Slaedr:
`create_const` functions are added to `BatchCsr`, `BatchEll` and `BatchDense` for the purpose of wrapping user data pointed to by `const` pointers.

In addition, the `BatchEll` interface is made more agnostic of whether the batch is uniform or non-uniform.

### DIFF
--- a/include/ginkgo/core/matrix/batch_ell.hpp
+++ b/include/ginkgo/core/matrix/batch_ell.hpp
@@ -139,7 +139,8 @@ public:
         GKO_ASSERT_EQ(size[1] + 1, col_ptrs.get_num_elems());
 
         auto batch_ell_mat = BatchEll<ValueType, IndexType>::create(
-            exec, num_batch_entries, size, num_elems_per_row);
+            exec, batch_dim<2>{num_batch_entries, size},
+            batch_stride{num_batch_entries, num_elems_per_row});
 
         batch_ell_mat->create_from_batch_csc_impl(values, row_idxs, col_ptrs);
         return batch_ell_mat;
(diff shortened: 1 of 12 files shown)


include/ginkgo/core/matrix/batch_ell.hpp
```diff
@@ -289,6 +290,34 @@ class BatchEll
...
+        std::shared_ptr<const Executor> exec, const batch_dim<2>& size,
+        const batch_stride& num_stored_elems_per_row,
+        const batch_stride& stride,
+        gko::detail::ConstArrayView<ValueType>&& values,
+        gko::detail::ConstArrayView<IndexType>&& col_idxs)
+    {
+        // cast const-ness away, but return a const object afterwards,
+        // so we can ensure that no modifications take place.
+        return std::unique_ptr<const BatchEll>(new BatchEll{
+            exec, size, num_stored_elems_per_row.at(0), stride.at(0),
```
pratikvn, line 316: Do you need to get the value at 0 ? I think passing in the batch_dim and batch_stride should also work ?

omp/test/matrix/batch_ell_kernels.cpp
```diff
@@ -68,8 +68,10 @@ class BatchEll : public ::testing::Test {
           mtx_size(10, gko::dim<2>(12, 7)),
           rand_engine(42)
     {
-        mtx = Mtx::create(exec, 2, gko::dim<2>{2, 3}, 2);
-        mtx2 = Mtx::create(exec, 2, gko::dim<2>{3, 3}, 2);
+        mtx = Mtx::create(exec, gko::batch_dim<2>{2, gko::dim<2>{2, 3}},
+                          gko::batch_stride{2, 2});
```
pratikvn, line 72: This is more verbose. Maybe we should have an implicit conversion from the previous version ? Not sure how error-proof that would be.

pratikvn (approved):
LGTM!

PR #1010, opened 2022-04-07
Fix (CB-)GMRES failing if initial guess is solution

fritzgoebel:
This PR adds a check to (CB-)GMRES to make sure the `restart_iter` is larger than zero after finishing. If it is zero, the initial guess was already a solution and the creation of `krylov_bases_small` breaks down because range accessors don't allow for a `range{0,0}`.

### DIFF
--- a/core/solver/cb_gmres.cpp
+++ b/core/solver/cb_gmres.cpp
@@ -480,22 +480,24 @@ void CbGmres<ValueType>::apply_dense_impl(
         }  // closes while(true)
         // Solve x
 
-        auto hessenberg_small = hessenberg->create_submatrix(
-            span{0, restart_iter},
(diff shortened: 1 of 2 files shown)


upsj:
I seem to recall fixing that somewhere else already. Maybe #861? I introduced a finalized flag into the stopping criterion just like in BiCGStab, which fixed the issues.

upsj:
A different case we need to fix is if the stopping criterion is not residual-based, but the solution already converged: Then you run into lucky breakdown which we need to catch. @MarcelKoch already suggested that for #973.

fritzgoebel:
Nice, I was aware of the second point, not the first one though. Then I can close this again.

PR #1011, opened 2022-04-07
add `test` to formatting regex for ginkgo-bot

upsj:
description :)

### DIFF
--- a/.github/bot-pr-base.sh
+++ b/.github/bot-pr-base.sh
@@ -3,8 +3,8 @@
 source .github/bot-base.sh
 
 EXTENSION_REGEX='\.(cuh?|hpp|hpp\.inc?|cpp)$'
-FORMAT_HEADER_REGEX='^(benchmark|core|cuda|hip|include/ginkgo/core|omp|reference|dpcpp|common/unified|test/[^t].*|test/t[^e].*)/'
-FORMAT_REGEX='^(common|examples|test/test_.*)/'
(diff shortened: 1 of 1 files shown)


.github/bot-pr-base.sh
```diff
@@ -3,8 +3,8 @@
 source .github/bot-base.sh
 
 EXTENSION_REGEX='\.(cuh?|hpp|hpp\.inc?|cpp)$'
-FORMAT_HEADER_REGEX='^(benchmark|core|cuda|hip|include/ginkgo/core|omp|reference|dpcpp|common/unified)/'
-FORMAT_REGEX='^(common|examples|test_install)/'
+FORMAT_HEADER_REGEX='^(benchmark|core|cuda|hip|include/ginkgo/core|omp|reference|dpcpp|common/unified|test/[^t].*|test/t[^e].*)/'
```
tcojean, line 6: Why do you specifically exclude `test/t.*` and `test/e.*`? I guess the first one is test/test_install?

upsj: yes, I want to format the header order for everything in `test`, except for `test/test_*`. Since it's unlikely that we will have another subdirectory with a `test/te.*` form, but something with `test/t.*` might be added in the future, I wanted to avoid this kind of issue for the future.

Slaedr (approved):
LGTM.

PR #1012, opened 2022-04-07
Add-scale (M <- a A + b M) operations

Slaedr:
Adds `add_scale` functions to `Dense` and `BatchDense`.

Also adds ways to use this functionality from the respective 'advanced' apply functions; these may be removed in the future. The final interface to this operation is yet to be determined.


common/cuda_hip/matrix/batch_dense_kernels.hpp.inc
```diff
@@ -211,6 +211,48 @@ __global__ __launch_bounds__(default_block_size, sm_multiplier) void add_scaled(
...
+    const gko::batch_dense::BatchEntry<const ValueType>& x,
+    const gko::batch_dense::BatchEntry<const ValueType>& beta,
+    const gko::batch_dense::BatchEntry<ValueType>& y)
+{
+    const int max_li = x.num_rows * x.num_rhs;
+    for (int li = threadIdx.x; li < max_li; li += blockDim.x) {
+        const int row = li / x.num_rhs;
+        const int col = li % x.num_rhs;
+
+        if (alpha.num_rhs == 1) {
```
pratikvn, line 226:
```suggestion
        if (alpha.num_rhs == 1 && beta.num_rhs==1) {
```

common/unified/matrix/dense_kernels.cpp
```diff
@@ -180,6 +180,39 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_SCALAR_TYPE(
...
 
 
+template <typename ValueType, typename ScalarType>
+void add_scale(std::shared_ptr<const DefaultExecutor> exec,
+               const matrix::Dense<ScalarType>* alpha,
+               const matrix::Dense<ValueType>* x,
+               const matrix::Dense<ScalarType>* beta,
+               matrix::Dense<ValueType>* y)
+{
+    if (alpha->get_size()[1] > 1) {
```
pratikvn, line 190:
```suggestion
    if (alpha->get_size()[1] > 1 && beta->get_size()[1] > 1) {
```

common/unified/matrix/dense_kernels.cpp
```diff
@@ -180,6 +180,39 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_SCALAR_TYPE(
     GKO_DECLARE_DENSE_ADD_SCALED_KERNEL);
 
 
+template <typename ValueType, typename ScalarType>
+void add_scale(std::shared_ptr<const DefaultExecutor> exec,
+               const matrix::Dense<ScalarType>* alpha,
+               const matrix::Dense<ValueType>* x,
+               const matrix::Dense<ScalarType>* beta,
+               matrix::Dense<ValueType>* y)
+{
```
pratikvn, line 189: Also maybe make sure that y and x both have non-zero sizes ? If input has zero size, then you can fill the output with zero ?

core/matrix/batch_dense.cpp
```diff
@@ -155,6 +156,31 @@ void BatchDense<ValueType>::add_scaled_impl(const BatchLinOp* alpha,
...
+            // different alpha for each column
+            GKO_ASSERT_BATCH_EQUAL_COLS(this, batch_alpha);
+        }
+    }
+    GKO_ASSERT_BATCH_EQUAL_DIMENSIONS(this, batch_a);
+    GKO_ASSERT_BATCH_EQUAL_DIMENSIONS(batch_alpha, batch_beta);
+    auto exec = this->get_executor();
+
+    exec->run(
+        batch_dense::make_add_scale(batch_alpha, batch_a, batch_beta, this));
```
pratikvn, line 180: nit: as it is being used only once,
```suggestion
    this->get_executor()->run(
        batch_dense::make_add_scale(batch_alpha, batch_a, batch_beta, this));
```

include/ginkgo/core/matrix/batch_dense.hpp
```diff
@@ -337,6 +337,20 @@ class BatchDense : public EnableBatchLinOp<BatchDense<ValueType>>,
                               make_temporary_clone(exec, b).get());
     }
 
+    /**
+     * Adds `a` scaled by `alpha` to the matrix scaled by `beta`.
+     *
```
pratikvn, line 342: Maybe add the expression,
```suggestion
     *
     * this = alpha * a + beta * this
     *
```

reference/matrix/batch_dense_kernels.hpp.inc
```diff
@@ -131,6 +131,33 @@ inline void add_scaled(
...
 
 
+template <typename ValueType>
+inline void add_scale(
+    const gko::batch_dense::BatchEntry<const ValueType>& alpha,
+    const gko::batch_dense::BatchEntry<const ValueType>& x,
+    const gko::batch_dense::BatchEntry<const ValueType>& beta,
+    const gko::batch_dense::BatchEntry<ValueType>& y)
+{
+    if (alpha.num_rhs == 1) {
```
pratikvn, line 141:
```suggestion
    if (alpha.num_rhs == 1 && beta.num_rhs == 1) {
```

pratikvn (approved):
Only small nits. LGTM!

common/cuda_hip/matrix/batch_dense_kernels.hpp.inc
Slaedr: This is already being checked in core.

common/unified/matrix/dense_kernels.cpp
Slaedr: I guess it should be fine because in core, we assert that x and y should have the same size. If the output has some finite size and the input is 0x0, that would be an error.

core/matrix/batch_dense.cpp
```diff
@@ -105,15 +107,22 @@ void BatchDense<ValueType>::apply_impl(const BatchLinOp* alpha,
...
     if (!this->get_size().stores_equal_sizes() ||
         !this->get_stride().stores_equal_strides()) {
         GKO_NOT_IMPLEMENTED;
     }
-    this->get_executor()->run(batch_dense::make_apply(
-        as<BatchDense<ValueType>>(alpha), this, as<BatchDense<ValueType>>(b),
-        as<BatchDense<ValueType>>(beta), as<BatchDense<ValueType>>(x)));
+    if (auto bid = dynamic_cast<const BatchIdentity<ValueType>*>(b)) {
+        if (auto xdense = dynamic_cast<BatchDense<ValueType>*>(x)) {
+            xdense->add_scale(alpha, this, beta);
```
pratikvn, line 116: Technically we could use `convert_and_copy_to` if we wanted to be flexible, but I think this is fine for now.

core/matrix/dense.cpp
```diff
@@ -238,6 +249,40 @@ void Dense<ValueType>::add_scaled_impl(const LinOp* alpha, const LinOp* b)
...
+    }
+    GKO_ASSERT_EQUAL_DIMENSIONS(this, a);
+    auto exec = this->get_executor();
+    auto alphac = make_temporary_clone(exec, alpha);
+    auto betac = make_temporary_clone(exec, beta);
+    auto ac = make_temporary_clone(exec, a);
+    if (dynamic_cast<const ConvertibleTo<Dense<>>*>(alpha) &&
+        is_complex<ValueType>()) {
+        exec->run(dense::make_add_scale(
+            make_temporary_conversion<remove_complex<ValueType>>(alphac.get())
```
pratikvn, line 271: Why do we need to have `alpha` and `beta` of non-complex type ?

include/ginkgo/core/matrix/batch_identity.hpp
```diff
@@ -0,0 +1,210 @@
...
+        return this->clone();
+    }
+
+    std::unique_ptr<BatchLinOp> conj_transpose() const override
+    {
+        return this->clone();
+    }
+
+    /**
+     * Unbatches the batched dense and creates a std::vector of Identity
```
pratikvn, line 98: I think this does not really unbatch the batch dense, but just creates a vector of Identity matrices from the given BatchIdentity matrix sizes

include/ginkgo/core/matrix/batch_identity.hpp
```diff
@@ -0,0 +1,210 @@
...
+     *
+     * @return  a std::vector containing the Identity matrices.
+     */
+    std::vector<std::unique_ptr<unbatch_type>> unbatch() const
+    {
+        GKO_ASSERT_BATCH_HAS_SQUARE_MATRICES(this);
+        auto exec = this->get_executor();
+        auto unbatch_mats = std::vector<std::unique_ptr<unbatch_type>>{};
+        const auto c_entry_size =
+            std::min(this->get_size().at(0)[0], this->get_size().at(0)[1]);
```
pratikvn, line 109: Unused

include/ginkgo/core/matrix/batch_identity.hpp
```diff
@@ -0,0 +1,210 @@
...
+    /**
+     * Creates a BatchIdentity matrix by duplicating Identity matrix
+     *
+     * @param exec  Executor associated to the matrix
+     * @param num_duplications  The number of times to duplicate
+     * @param input  The matrix to be duplicated.
+     */
+    BatchIdentity(std::shared_ptr<const Executor> exec,
+                  const size_type num_duplications,
+                  const Identity<value_type>* const input)
```
pratikvn, line 190: Do we really need to take an Identity matrix ? Just a size or a batch_size should be enough right ? Otherwise, the use has to go through the trouble of creating an Identity matrix first.

pratikvn (approved):
Nice work on the interface and the tests. LGTM!

common/cuda_hip/matrix/batch_dense_kernels.hpp.inc
pratikvn: IMO, this makes the code more readable and clear, but it is not that important.

core/matrix/dense.cpp
Slaedr: I guess we don't need that functionality. But I thought I'll keep it similar to the `add_scaled` function in `Dense`.

common/unified/matrix/dense_kernels.cpp
pratikvn: Yes, with the changes/semantics that Tobias implemented, using a 0x0 matrix is not necessarily an error. For example [see dense apply](https://github.com/ginkgo-project/ginkgo/blob/develop/cuda/matrix/dense_kernels.cu#L164=)

include/ginkgo/core/matrix/batch_identity.hpp
Slaedr: Oh yes, I need to remove "dense" and replace it with "identity". I think there are many places in the docstrings that need fixing; I'll do that later.

include/ginkgo/core/matrix/batch_identity.hpp
Slaedr: That can already be done by the first constructor. This is there just in case the user already has an identity matrix that they want to duplicate. I guess this can be removed.

PR #1013, opened 2022-04-08
Use reductions with reused tmp storage in solver

upsj:
This PR uses the new interface for Dense reduction kernels from #990 in all Krylov solvers that use them

### DIFF
--- a/core/solver/bicg.cpp
+++ b/core/solver/bicg.cpp
@@ -129,6 +129,8 @@ void Bicg<ValueType>::apply_dense_impl(const matrix::Dense<ValueType>* dense_b,
 
     auto exec = this->get_executor();
 
+    Array<char> reduction_tmp{exec};
+
     auto one_op = initialize<Vector>({one<ValueType>()}, exec);
     auto neg_one_op = initialize<Vector>({-one<ValueType>()}, exec);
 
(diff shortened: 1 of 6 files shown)


core/solver/bicg.cpp
```diff
@@ -129,6 +129,8 @@ void Bicg<ValueType>::apply_dense_impl(const matrix::Dense<ValueType>* dense_b,
 
     auto exec = this->get_executor();
 
+    Array<char> reduction_tmp{exec};
```
pratikvn, line 132: I guess the idea is to move this to one of the base classes in the value semantics (solver base class) PRs ?

pratikvn (approved):
LGTM!

core/solver/bicgstab.cpp
```diff
@@ -111,6 +111,8 @@ void Bicgstab<ValueType>::apply_dense_impl(
 
     auto exec = this->get_executor();
 
+    Array<char> reduction_tmp{exec};
```
yhmtsai, line 114: move it to mutable class member? in case we need reusing these in the multigrid or something else.

upsj: That should be handled later IMO, together with #805.

yhmtsai (approved):

upsj:
rebase!

PR #1014, opened 2022-04-08
Add OpenMP parallel sync-free triangular solvers

upsj:
This adds a simple sync-free parallel triangular solver to OpenMP.

cc @lksriemer

### DIFF
--- a/omp/CMakeLists.txt
+++ b/omp/CMakeLists.txt
@@ -1,4 +1,4 @@
-find_package(OpenMP 3.0 REQUIRED)
+find_package(OpenMP 4.0 REQUIRED)
 
 add_library(ginkgo_omp $<TARGET_OBJECTS:ginkgo_omp_device> "")
 target_sources(ginkgo_omp
--- a/omp/components/atomic.hpp
+++ b/omp/components/atomic.hpp
@@ -46,11 +46,21 @@ namespace kernels {
 namespace omp {
 
 
+template <typename IndexType>
+IndexType atomic_inc(IndexType& val)
+{
+    IndexType result{};
+#pragma omp atomic capture seq_cst
+    result = val++;
+    return result;
+}
+
+
 template <typename ValueType,
           std::enable_if_t<!is_complex<ValueType>()>* = nullptr>
 void atomic_add(ValueType& out, ValueType val)
 {
-#pragma omp atomic
+#pragma omp atomic update seq_cst
     out += val;
 }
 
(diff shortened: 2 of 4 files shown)


omp/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-find_package(OpenMP 3.0 REQUIRED)
+find_package(OpenMP 4.0 REQUIRED)
```
thoasm, line 1: Isn't this quite a significant change? Is it considered interface breaking when we move to a newer OpenMP version?

upsj: not really in terms of compiler support: All of our compilers support it AFAIK, except for MSVC, but they are behind anyways

upsj: Okay, almost, I didn't know we still test for [Clang 7.0](https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/2311759914). I can also do 3.1, but that means that we can't require sequential consistency semantics from the atomics, which I would be more comfortable with than just using `#pragma omp atomic *`

lksriemer:
In case we want to seriously get into the topic of shared memory, multithreaded sptrsv, I'll just leave some notes below.

Level based methods seem to enjoy some more popularity on the CPU. This might be due to "cheaply" available global synchronization, along with the smaller required minimum level size for saturation. Synchronization then either happens via barriers, or point-to-point. Different barriers are analyzed in [5]. In the point-to-point case dependency pruning might be of interest [1]. Graph coloring methods to balance parallelism may achieve good results. Hard to say how performance is in relation to the nan-spinning method implemented here.

2D blocking methods are also an option, different schemes are explored in [4], though it is hard to qualify how they compare on today's hardware.

Similar to the GPU, partitioned inverses can be employed reduce the length of the critical path [6, p.86], but stability needs careful consideration [7].

Reordering is as important as it is on the GPU, and also as difficult to predict. Locality is the defining factor. In case a bfs-style reordering is applied, a hybrid solver handling the "dense triangle" is a very good idea [2].

In the broader scheme, storage order is another thing to consider, possibly L and U after each other, U in inverted row order [3], or in a more sophisticated order mirroring blocked execution at cores.

[1] [Sparsifying Synchronization for High-Performance Shared-Memory Sparse Triangular Solver](https://link.springer.com/chapter/10.1007/978-3-319-07518-1_8)
[2] [A Hybrid Multithreaded Direct Sparse Triangular Solver](https://www.osti.gov/biblio/1399343)
[3] [Sparse triangular solves for ILU revisited: data layout crucial to better performance](https://journals.sagepub.com/doi/pdf/10.1177/1094342010389857)
[4] [Parallel algorithms for solving linear systems with sparse triangular matrices](https://link.springer.com/article/10.1007/s00607-009-0066-3)
[5] [Factors Impacting Performance of Multithreaded Sparse Triangular Solve](https://link.springer.com/chapter/10.1007/978-3-642-19328-6_6)
[6] [Parallel Algorithms for Dense Linear Algebra Computations](https://epubs.siam.org/doi/pdf/10.1137/1032002)
[7] [Stability of the Partitioned Inverse Method for Parallel Solution of Sparse Triangular Systems](https://epubs.siam.org/doi/epdf/10.1137/0915009)

upsj:
It seems we need to do some more work, we are not getting any speedup so far.

PR #1015, opened 2022-04-08
fix pull_request_target actions with cancel-in-progress

upsj:
With pull_request_target, the baseline ref is always the base branch, so we need a different way to figure out a unique PR identifier.


yhmtsai:
does it need to change the group `${{ github.workflow }}-${{ (github.head_ref && github.ref) || github.run_id }}`?
when the head_ref is defined, it also use github.ref as the identifer.
However, we do not see this behavior?

upsj:
that does not help, because `github.ref` is `develop` for `pull_request_target` independent of the PR, which is the issue that triggered this PR.

yhmtsai (approved):
I see. Thanks!
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request
and 
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

MarcelKoch (approved):

upsj:
rebase!

PR #1016, opened 2022-04-11
Adding globbing and wordsplitting prevention.

keldu:
Just set some quotes around some variables even though most wouldn't provide attack surfaces.
Defense in depth still is a good concept to follow.

### DIFF
--- a/.github/format.sh
+++ b/.github/format.sh
@@ -11,5 +11,5 @@ if [[ "$LIST_FILES" != "" ]]; then
   git commit -a -m "Format files
 
 Co-authored-by: $USER_COMBINED"
-  git push fork "$LOCAL_BRANCH":"$HEAD_BRANCH" 2>&1 || bot_error "Cannot push formatted branch, are edits for maintainers allowed?"
+  git push fork "$LOCAL_BRANCH:$HEAD_BRANCH" 2>&1 || bot_error "Cannot push formatted branch, are edits for maintainers allowed?"
 fi
--- a/.github/format.sh
+++ b/.github/format.sh
@@ -11,5 +11,5 @@ if [[ "$LIST_FILES" != "" ]]; then
   git commit -a -m "Format files
 
 Co-authored-by: $USER_COMBINED"
-  git push fork $LOCAL_BRANCH:$HEAD_BRANCH 2>&1 || bot_error "Cannot push formatted branch, are edits for maintainers allowed?"
+  git push fork "$LOCAL_BRANCH":"$HEAD_BRANCH" 2>&1 || bot_error "Cannot push formatted branch, are edits for maintainers allowed?"
 fi
--- a/.github/rebase.sh
+++ b/.github/rebase.sh
@@ -5,19 +5,19 @@ source .github/bot-pr-base.sh
 git remote add base "$BASE_URL"
 git remote add fork "$HEAD_URL"
 
-git fetch base $BASE_BRANCH
-git fetch fork $HEAD_BRANCH
+git fetch base "$BASE_BRANCH"
+git fetch fork "$HEAD_BRANCH"
 
 git config user.email "$USER_EMAIL"
 git config user.name "$USER_NAME"
 
-LOCAL_BRANCH=rebase-tmp-$HEAD_BRANCH
-git checkout -b $LOCAL_BRANCH fork/$HEAD_BRANCH
+LOCAL_BRANCH=rebase-tmp-"$HEAD_BRANCH"
+git checkout -b "$LOCAL_BRANCH" fork/"$HEAD_BRANCH"
 
 bot_delete_comments_matching "Error: Rebase failed"
 
 # do the rebase
-git rebase base/$BASE_BRANCH 2>&1 || bot_error "Rebase failed, see the related [Action]($JOB_URL) for details"
+git rebase base/"$BASE_BRANCH" 2>&1 || bot_error "Rebase failed, see the related [Action]($JOB_URL) for details"
 
 # push back
-git push --force-with-lease fork $LOCAL_BRANCH:$HEAD_BRANCH 2>&1 || bot_error "Cannot push rebased branch, are edits for maintainers allowed, or were changes pushed while the rebase was running?"
+git push --force-with-lease fork "$LOCAL_BRANCH":"$HEAD_BRANCH" 2>&1 || bot_error "Cannot push rebased branch, are edits for maintainers allowed, or were changes pushed while the rebase was running?"
--- a/.github/rebase.sh
+++ b/.github/rebase.sh
@@ -11,13 +11,13 @@ git fetch fork "$HEAD_BRANCH"
 git config user.email "$USER_EMAIL"
 git config user.name "$USER_NAME"
 
-LOCAL_BRANCH=rebase-tmp-"$HEAD_BRANCH"
-git checkout -b "$LOCAL_BRANCH" fork/"$HEAD_BRANCH"
+LOCAL_BRANCH="rebase-tmp-$HEAD_BRANCH"
+git checkout -b "$LOCAL_BRANCH" "fork/$HEAD_BRANCH"
 
 bot_delete_comments_matching "Error: Rebase failed"
 
 # do the rebase
-git rebase base/"$BASE_BRANCH" 2>&1 || bot_error "Rebase failed, see the related [Action]($JOB_URL) for details"
+git rebase "base/$BASE_BRANCH" 2>&1 || bot_error "Rebase failed, see the related [Action]($JOB_URL) for details"
 
 # push back
-git push --force-with-lease fork "$LOCAL_BRANCH":"$HEAD_BRANCH" 2>&1 || bot_error "Cannot push rebased branch, are edits for maintainers allowed, or were changes pushed while the rebase was running?"
+git push --force-with-lease fork "$LOCAL_BRANCH:$HEAD_BRANCH" 2>&1 || bot_error "Cannot push rebased branch, are edits for maintainers allowed, or were changes pushed while the rebase was running?"
--- a/dev_tools/oneapi/convert_source.sh
+++ b/dev_tools/oneapi/convert_source.sh
@@ -193,7 +193,7 @@ done < "${UNFORMAT_FILE}"
 ${CLANG_FORMAT} -style=file "${EMBED_FILE}" > "${FORMAT_FILE}"
 
 # Add an extra host function so that the converted DPC++ code will look like CUDA.
-"${SCRIPT_DIR}"/add_host_function.sh "${FORMAT_FILE}" > "${EMBED_HOST_FILE}"
+"${SCRIPT_DIR}/add_host_function.sh" "${FORMAT_FILE}" > "${EMBED_HOST_FILE}"
 
 while IFS='' read -r line; do
     if [[ "${line}" =~ ${KERNEL_SYNTAX_START} ]] || [[ "${IN_SYNTAX}" = "true" ]]; then
--- a/dev_tools/oneapi/convert_source.sh
+++ b/dev_tools/oneapi/convert_source.sh
@@ -365,7 +365,7 @@ if [ "${EXTRACT_KERNEL}" = "true" ]; then
         dpct_device_path=$(echo "${variable}" | sed 's/common/dpcpp_code/g')
         dpct_device_file=$(echo "${dpct_device_path}" | sed 's|/|@|g')
         dpct_device_file="output/${dpct_device_file}"
-        cat "${dpct_file}" | sed -n "/${device_regex} - start/,/${device_regex} - end/p" | sed "1d;\$d" > ${dpct_device_file}
+        cat "${dpct_file}" | sed -n "/${device_regex} - start/,/${device_regex} - end/p" | sed "1d;\$d" > "${dpct_device_file}"
         sed -i "/${device_regex} - start/,/${device_regex} - end/d;s~// *#include \"${device_regex}\"~#include \"${dpct_device_path}\"~g" ${dpct_file}
         dpct_dir=$(dirname "${dpct_device_path}")
         mkdir -p "${ROOT_DIR}/${dpct_dir}"
--- a/dev_tools/oneapi/convert_source.sh
+++ b/dev_tools/oneapi/convert_source.sh
@@ -193,7 +193,7 @@ done < "${UNFORMAT_FILE}"
 ${CLANG_FORMAT} -style=file "${EMBED_FILE}" > "${FORMAT_FILE}"
 
 # Add an extra host function so that the converted DPC++ code will look like CUDA.
-${SCRIPT_DIR}/add_host_function.sh "${FORMAT_FILE}" > "${EMBED_HOST_FILE}"
+"${SCRIPT_DIR}"/add_host_function.sh "${FORMAT_FILE}" > "${EMBED_HOST_FILE}"
 
 while IFS='' read -r line; do
     if [[ "${line}" =~ ${KERNEL_SYNTAX_START} ]] || [[ "${IN_SYNTAX}" = "true" ]]; then
@@ -365,7 +365,7 @@ if [ "${EXTRACT_KERNEL}" = "true" ]; then
         dpct_device_path=$(echo "${variable}" | sed 's/common/dpcpp_code/g')
         dpct_device_file=$(echo "${dpct_device_path}" | sed 's|/|@|g')
         dpct_device_file="output/${dpct_device_file}"
-        cat ${dpct_file} | sed -n "/${device_regex} - start/,/${device_regex} - end/p" | sed "1d;\$d" > ${dpct_device_file}
+        cat "${dpct_file}" | sed -n "/${device_regex} - start/,/${device_regex} - end/p" | sed "1d;\$d" > ${dpct_device_file}
         sed -i "/${device_regex} - start/,/${device_regex} - end/d;s~// *#include \"${device_regex}\"~#include \"${dpct_device_path}\"~g" ${dpct_file}
         dpct_dir=$(dirname "${dpct_device_path}")
         mkdir -p "${ROOT_DIR}/${dpct_dir}"
--- a/dev_tools/scripts/add_license.sh
+++ b/dev_tools/scripts/add_license.sh
@@ -17,33 +17,27 @@ COMMENTED_LICENSE_FILE="${THIS_DIR}/commented_license.tmp"
 DIFF_FILE="${THIS_DIR}/diff.patch.tmp"
 
 # Test if required commands are present on the system:
-command -v find &> /dev/null
-if [ ${?} -ne 0 ]; then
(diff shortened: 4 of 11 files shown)


.github/rebase.sh
```diff
@@ -5,19 +5,19 @@ source .github/bot-pr-base.sh
...
-LOCAL_BRANCH=rebase-tmp-$HEAD_BRANCH
-git checkout -b $LOCAL_BRANCH fork/$HEAD_BRANCH
+LOCAL_BRANCH=rebase-tmp-"$HEAD_BRANCH"
+git checkout -b "$LOCAL_BRANCH" fork/"$HEAD_BRANCH"
 
 bot_delete_comments_matching "Error: Rebase failed"
 
 # do the rebase
-git rebase base/$BASE_BRANCH 2>&1 || bot_error "Rebase failed, see the related [Action]($JOB_URL) for details"
+git rebase base/"$BASE_BRANCH" 2>&1 || bot_error "Rebase failed, see the related [Action]($JOB_URL) for details"
```
upsj, line 20:
```suggestion
git rebase "base/$BASE_BRANCH" 2>&1 || bot_error "Rebase failed, see the related [Action]($JOB_URL) for details"
```

.github/rebase.sh
```diff
@@ -5,19 +5,19 @@ source .github/bot-pr-base.sh
...
 
 # push back
-git push --force-with-lease fork $LOCAL_BRANCH:$HEAD_BRANCH 2>&1 || bot_error "Cannot push rebased branch, are edits for maintainers allowed, or were changes pushed while the rebase was running?"
+git push --force-with-lease fork "$LOCAL_BRANCH":"$HEAD_BRANCH" 2>&1 || bot_error "Cannot push rebased branch, are edits for maintainers allowed, or were changes pushed while the rebase was running?"
```
upsj, line 23:
```suggestion
git push --force-with-lease fork "$LOCAL_BRANCH:$HEAD_BRANCH" 2>&1 || bot_error "Cannot push rebased branch, are edits for maintainers allowed, or were changes pushed while the rebase was running?"
```

.github/format.sh
```diff
@@ -11,5 +11,5 @@ if [[ "$LIST_FILES" != "" ]]; then
   git commit -a -m "Format files
 
 Co-authored-by: $USER_COMBINED"
-  git push fork $LOCAL_BRANCH:$HEAD_BRANCH 2>&1 || bot_error "Cannot push formatted branch, are edits for maintainers allowed?"
+  git push fork "$LOCAL_BRANCH":"$HEAD_BRANCH" 2>&1 || bot_error "Cannot push formatted branch, are edits for maintainers allowed?"
```
upsj, line 14:
```suggestion
  git push fork "$LOCAL_BRANCH:$HEAD_BRANCH" 2>&1 || bot_error "Cannot push formatted branch, are edits for maintainers allowed?"
```

upsj (approved):
LGTM!

yhmtsai (approved):
LGTM although I think git always generates no-space branch, which replaces space by - when creating branch

tcojean (approved):
LGTM.

Did you try the scripts in `dev_tools/scripts`? I'm sure there we might have more issues.

Also, maybe we should add a shellkcheck job for the few bash scripts we have? See https://github.com/ginkgo-project/gitlab-hpc-ci-cb/blob/master/.github/workflows/shell_check.yml

keldu:
> LGTM.
> 
> Did you try the scripts in `dev_tools/scripts`? I'm sure there we might have more issues.
> 

I didn't.
Currently checking it out.

For the starter it's a lot of echo | sed calls which shellcheck complains about and suggests using ```${variable//search/replace}```. Though I am bit unfamiliar with sed syntax. I'll check it out and see if I can reproduce functionality. I think it's a bit too risky though to change these kind of calls.

Also some ```${cat file | sed}``` could be replaced by ```${sed ... < file}```. That's easier to do.

Will rework that folder today.

> Also, maybe we should add a shellkcheck job for the few bash scripts we have? See https://github.com/ginkgo-project/gitlab-hpc-ci-cb/blob/master/.github/workflows/shell_check.yml

I'm not sure what the implications would be, because there are some things which I wouldn't be comfortable to replace.
Mostly because those are styling/clean bash hints.

If shellcheck can provide this level of detail then it should be possible.

tcojean:
> I'm not sure what the implications would be, because there are some things which I wouldn't be comfortable to replace. Mostly because those are styling/clean bash hints.

> If shellcheck can provide this level of detail then it should be possible. The checking job shouldn't fail because of echo | sed instead of the bash replace syntax.

You can tell shellcheck to ignore some things, like this: 
https://github.com/ginkgo-project/gitlab-hpc-ci-cb/blob/master/run.sh#L35-L36

yhmtsai (dismissed):
I think bash can not distinguish the nested quote
"AAA"BB"CC" -> ("AAA")(BB)("CCC") not ("AAA(BB)CC")

yhmtsai:
Sorry for confusion. `"` works when there is `()`.
if "AAA$(BB"CCC")", it should work as expected. `(`) helps separation

dev_tools/scripts/create_new_algorithm.sh
```diff
@@ -304,66 +304,66 @@ then
         ## Automatically duplicate the entry of common_kernels.inc.cpp
         common_kernels_file="${GINKGO_ROOT_DIR}/core/device_hooks/common_kernels.inc.cpp"
         # add the new kernel file to headers
-        headers=( "$(grep '#include \"' $common_kernels_file)" )
+        headers=( "$(grep '#include \"' "$common_kernels_file")" )
```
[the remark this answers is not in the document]
keldu: That should be correct. You're right. I somehow missed the review.

dev_tools/scripts/create_new_algorithm.sh
```diff
@@ -24,9 +24,9 @@ function print_help {
 function list_sources {
     for type in solver preconditioner matrix factorization
     do
-        for i in $(ls $GINKGO_ROOT_DIR/core/$type/*.cpp)
+	for i in "${GINKGO_ROOT_DIR}/core/${type}/"*.cpp;
```
upsj, line 27: globs should also be evaluated in quotes, I think?
```suggestion
        for i in "${GINKGO_ROOT_DIR}/core/${type}/*.cpp";
```

keldu:
> Sorry for confusion. `"` works when there is `()`. if "AAA$(BB"CCC")", it should work as expected. `()` helps separation

To be honest, I also didn't know that it would work before I changed the scripts here.
It's a bit confusing, but `shellcheck` complains if you don't do it.

dev_tools/scripts/create_new_algorithm.sh
```diff
@@ -172,44 +172,44 @@ TEMPLATE_FILES_DESCRIPTIONS=(
...
+    sourcename=$(echo "${TEMPLATE_FILES[$i-1]}" | sed "s/${name}/${source_name}/" )
     sourcepath=${TEMPLATE_FILES_LOCATIONS[$i-1]}/${sourcename}
 
     # create folder for temporary files
-    mkdir -p ${TMPDIR}/${TEMPLATE_FILES_LOCATIONS[$i-1]}
+    mkdir -p "${TMPDIR}/${TEMPLATE_FILES_LOCATIONS[$i-1]}"
 
     # Evaluate the extension and try to find the matching files
-    for j in $(ls ${GINKGO_ROOT_DIR}/${sourcepath})
+    for j in "${GINKGO_ROOT_DIR}/${sourcepath}/"*;
```
upsj, line 186: same here

dev_tools/scripts/create_new_algorithm.sh
```diff
@@ -265,37 +265,37 @@ then
...
-                        insert_to=$(grep -n -m 1 "target_sources" $cmake_file | sed 's/:.*//')
+                        insert_to=$(grep -n -m 1 "target_sources" "$cmake_file" | sed 's/:.*//')
                         insert_to=$((insert_to + 1)) # account for the "PRIVATE"
-                        awk '/^target_sources/,/    .*\)/ {if (match($0, "target_sources") != 0 || match($0, "PRIVATE") != 0){ print $0 }; next}1'  $cmake_file > tmp
+                        awk '/^target_sources/,/    .*\)/ {if (match($0, "target_sources") != 0 || match($0, "PRIVATE") != 0){ print $0 }; next}1'  "$cmake_file" > tmp
                     fi
 
-                    mytmp=`mktemp`
-                    head -n$insert_to tmp > $mytmp
+		    mytmp=$(mktemp)
```
upsj, line 290: indentation?

upsj (approved):
LGTM! Mostly nits

keldu:
Rebased and force pushed because of a useless merge

dev_tools/oneapi/convert_source.sh
```diff
@@ -364,7 +364,7 @@ if [ "${EXTRACT_KERNEL}" = "true" ]; then
         dpct_device_path=$(echo "${variable}" | sed 's/common/dpcpp_code/g')
         dpct_device_file=$(echo "${dpct_device_path}" | sed 's|/|@|g')
         dpct_device_file="output/${dpct_device_file}"
-        cat ${dpct_file} | sed -n "/${device_regex} - start/,/${device_regex} - end/p" | sed "1d;\$d" > ${dpct_device_file}
+        cat "${dpct_file}" | sed -n "/${device_regex} - start/,/${device_regex} - end/p" | sed "1d;\$d" > ${dpct_device_file}
```
tcojean, line 367: Minor, was this one creating a warning
```suggestion
        cat "${dpct_file}" | sed -n "/${device_regex} - start/,/${device_regex} - end/p" | sed "1d;\$d" > "${dpct_device_file}"
```

dev_tools/scripts/add_license.sh
```diff
@@ -17,33 +17,27 @@ COMMENTED_LICENSE_FILE="${THIS_DIR}/commented_license.tmp"
...
 fi
-command -v sed &> /dev/null
-if [ ${?} -ne 0 ]; then
+if ! command -v sed &> /dev/null; then
     echo 'The command `sed` is required for this script to work, but not supported by your system.' 1>&2
     exit 1
 fi
-command -v cut &> /dev/null
-if [ ${?} -ne 0 ]; then
+if command -v cut &> /dev/null; then
```
tcojean, line 40: Missing `!` ?
```suggestion
if ! command -v cut &> /dev/null; then
```

dev_tools/scripts/update_ginkgo_header.sh
```diff
@@ -25,29 +25,24 @@ GINKGO_HEADER_TEMPLATE_FILE="${GINKGO_HEADER_FILE}.in"
...
 fi
-command -v sort &> /dev/null
-if [ ${?} -ne 0 ]; then
+if ! command -v sort &> /dev/null; then
     echo "${WARNING_PREFIX} "'The command `sort` is required for this script to work, but not supported by your system.' 1>&2
     exit 0
 fi
-command -v cmp &> /dev/null
-if [ ${?} -ne 0 ]; then
+if command -v cmp &> /dev/null; then
```
tcojean, line 36:
```suggestion
if ! command -v cmp &> /dev/null; then
```

tcojean (dismissed):
LGTM. Some minor issues

tcojean:
Random question since it's maybe a good time, do you need to update your mail in the [contributors.txt](https://github.com/ginkgo-project/ginkgo/blob/develop/contributors.txt#L17) ?

dev_tools/scripts/create_new_algorithm.sh
```diff
@@ -24,9 +24,9 @@ function print_help {
 function list_sources {
     for type in solver preconditioner matrix factorization
     do
-        for i in $(ls $GINKGO_ROOT_DIR/core/$type/*.cpp)
+        for i in "${GINKGO_ROOT_DIR}/core/${type}/*.cpp";
```
tcojean, line 27: This doesn't work with my bash on my computer because the `*` is quoted
```suggestion
        for i in "${GINKGO_ROOT_DIR}/core/${type}/"*.cpp;
```

dev_tools/scripts/create_new_algorithm.sh
```diff
@@ -24,9 +24,9 @@ function print_help {
...
         do
-            echo "$type "$(basename "$i" | cut -d"." -f1)
+            echo "$type $(basename "$i" | cut -d"." -f1)"
```
tcojean, line 29:
```suggestion
            echo "$type $(basename \"$i\" | cut -d'.' -f1)"
```

dev_tools/scripts/create_new_algorithm.sh
```diff
@@ -172,44 +172,44 @@ TEMPLATE_FILES_DESCRIPTIONS=(
...
+    sourcename=$(echo "${TEMPLATE_FILES[$i-1]}" | sed "s/${name}/${source_name}/" )
     sourcepath=${TEMPLATE_FILES_LOCATIONS[$i-1]}/${sourcename}
 
     # create folder for temporary files
-    mkdir -p ${TMPDIR}/${TEMPLATE_FILES_LOCATIONS[$i-1]}
+    mkdir -p "${TMPDIR}/${TEMPLATE_FILES_LOCATIONS[$i-1]}"
 
     # Evaluate the extension and try to find the matching files
-    for j in $(ls ${GINKGO_ROOT_DIR}/${sourcepath})
+    for j in "${GINKGO_ROOT_DIR}/${sourcepath}/*";
```
tcojean, line 186: The `ls` version is really required here, because we can have glob in the `sourcepath` variables (see the templates).
```suggestion
    for j in $(ls ${GINKGO_ROOT_DIR}/${sourcepath});
```

dev_tools/scripts/create_new_algorithm.sh
```diff
@@ -231,11 +231,11 @@ then
         for ((i=1; i<=${#CMAKE_FILES[@]}; i++))
         do
             sourcepath=${TEMPLATE_FILES_LOCATIONS[$i-1]}/${TEMPLATE_FILES[$i-1]}
-            for j in $(ls ${GINKGO_ROOT_DIR}/${sourcepath})
+        for j in "${GINKGO_ROOT_DIR}/${sourcepath}/"*;
```
tcojean, line 234: Same here, the `ls` is really needed.

tcojean (changes requested):
Some changes are required to get back the functionality of `create_new_algorithm.sh` script.

dev_tools/scripts/create_new_algorithm.sh
upsj: weird, I suggested this change because it works on my system, I thought globs are expanded inside `""` environments, just not in `''`

upsj:
General question: are we still using `create_new_algorithm.sh`? I was never happy with the results, and created the files myself instead, and the directory structure changed somewhat since then, we don't really seem to be updating it.

tcojean:
With the common kernels and general complexification of the Ginkgo structure, it becomes hard to get it to work, as different LinOp can live in many different places.

Overall, I think it was a good way to jump into Ginkgo and get a feeling for its structure.

dev_tools/scripts/create_new_algorithm.sh
tcojean: I have `GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu)` and it somehow definitely doesn't work here.

upsj: Ah, I see what is going wrong: The glob is being expanded, but into a single string:
`for f in "*"; do echo $f; done` lists all files in a single line, while removing the quotes gets them in separate lines. Then this needs to be changed, yes.

dev_tools/scripts/create_new_algorithm.sh
yhmtsai: does it need escape?

yhmtsai (approved):

dev_tools/scripts/create_new_algorithm.sh
keldu: But then we have to be really careful not to have any whitespace/IFS characters in there.

I'll check it out.

keldu: Hmm, if you don't quote `${sourcepath}`, then it shouldn't be a problem.

```
for j in "${GINKGO_ROOT_DIR}/"${sourcepath}/*;
```
Produces a newline separated list of the files.

```
for j in "${GINKGO_ROOT_DIR}/${sourcepath}/"*;
```

Produces a space separated list of the files. Weird.

dev_tools/scripts/create_new_algorithm.sh
keldu: Does it work here without the quoted `${sourcepath}`

dev_tools/oneapi/convert_source.sh
keldu: Yes, I overlooked it. Generally `cat "{dpct_file}" | cmd` is also a bit useless, which overshadowed the globbing catch a bit.

Alternative would be `cmd < ${dpct_file}`, but I didn't think that would be critical at all.

dev_tools/scripts/update_ginkgo_header.sh
keldu: Yes, That would be correct

dev_tools/scripts/create_new_algorithm.sh
keldu: Weird. at least `for j in "${GINKGO_ROOT_DIR}/"${sourcepath}/*;` worked for me and tobias.
There should be an alternative, because ls has some cases in which it is unsafe.

keldu: We may keep the ls for now. But should keep the problem with ls in consideration. This could break when some things like spaces occur in files.

I'll leave the ls in there then

keldu: And the indentation was wrong. yes.
Luckily that doesn't matter in bash. Still needs to be changed though

keldu:
> I think something weird happened during your update of this PR/branch. You have many unrelated commits. You rebase and only pick the commits that are relevant here, i.e., only authored by you.

damn. I wanted to rebase with the current develop ._.

I never cared about rebasing before tbh and just merged, so I'm still a bit unexperienced with that.

keldu:
Ok, the problem was that I forgot to force push after I rebased.
Instead I merged with the rebase strategy afterwards (: 
At least that was easy to fix.

dev_tools/scripts/create_new_algorithm.sh
```diff
@@ -24,9 +24,9 @@ function print_help {
 function list_sources {
     for type in solver preconditioner matrix factorization
     do
-        for i in $(ls $GINKGO_ROOT_DIR/core/$type/*.cpp)
+        for i in "${GINKGO_ROOT_DIR}/core/${type}/"*.cpp;
```
tcojean, line 27: This one also needs the `ls` form

PR #1020, opened 2022-04-12
Only allow gko::share() to work with rvalue references

thoasm:
Previously, `gko::share()` is allowed to act on regular references, which leads some users to incorrectly use `gko::share`.
By restricting it to rvalue references, it is harder to misuse this function while still allowing the same functionality with `gko::share(gko::give(val))` or `gko::share(std::move(val))` (both do the exact same thing).

Note: This does break the public interface as it is no longer possible to call `gko::share` with lvalues of any sort. However, I do consider it a bug that it was possible previously.
In its current form, writing errors accidentally is too easy, especially when examples are extended / modified.

If we don't adopt the `static_assert` for the xvalue check (rvalue reference for the caller), I would still suggest to have all other changes merged (especially the examples) in order to prevent new users or developers to reuse moved-from variables.

### DIFF
--- a/core/test/base/utils.cpp
+++ b/core/test/base/utils.cpp
@@ -163,7 +163,7 @@ TEST(Share, SharesSharedPointer)
     std::shared_ptr<Derived> p(new Derived());
     auto plain = p.get();
 
-    auto shared = gko::share(p);
+    auto shared = gko::share(std::move(p));
 
     ::testing::StaticAssertTypeEq<decltype(shared), std::shared_ptr<Derived>>();
     ASSERT_EQ(plain, shared.get());
@@ -175,7 +175,7 @@ TEST(Share, SharesUniquePointer)
     std::unique_ptr<Derived> p(new Derived());
     auto plain = p.get();
 
-    auto shared = gko::share(p);
+    auto shared = gko::share(std::move(p));
 
     ::testing::StaticAssertTypeEq<decltype(shared), std::shared_ptr<Derived>>();
     ASSERT_EQ(plain, shared.get());
--- a/core/test/base/utils.cpp
+++ b/core/test/base/utils.cpp
@@ -170,6 +170,14 @@ TEST(Share, SharesSharedPointer)
 }
 
 
+TEST(Share, SharesTemporarySharedPointer)
+{
+    auto shared = gko::share(std::make_shared<Derived>());
+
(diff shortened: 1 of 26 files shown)


include/ginkgo/core/base/utils_helper.hpp
```diff
@@ -202,7 +202,8 @@ inline detail::cloned_type<Pointer> clone(std::shared_ptr<const Executor> exec,
  * @tparam OwningPointer  type of pointer with ownership to the object
  *                        (has to be a smart pointer)
  *
- * @param p  a pointer to the object
+ * @param p  a pointer to the object. It must be movable, meaning it must be an
+ *           rvalue reference.
```
upsj, line 206:
```suggestion
 * @param p  a pointer to the object. It must be a temporary or explicitly marked moveable (rvalue reference).
```

include/ginkgo/core/base/utils_helper.hpp
```diff
@@ -211,6 +212,8 @@ inline detail::shared_type<OwningPointer> share(OwningPointer&& p)
 {
     static_assert(detail::have_ownership<OwningPointer>(),
                   "OwningPointer does not have ownership of the object");
+    static_assert(std::is_rvalue_reference<decltype(p)>::value,
+                  "p must be an xvalue for this function to work");
```
upsj, line 216: coming from the caller perspective, p can be either a prvalue or an xvalue, so I would suggest adapting the error message to the caller instead of the callee.
```suggestion
    static_assert(std::is_rvalue_reference<decltype(p)>::value,
                  "p must be an rvalue for this function to work");
```

upsj (approved):
LGTM, great observation!

thoasm:
@upsj yes, that is how I found most of the places I changed (some I already knew).

core/test/base/utils.cpp
```diff
@@ -175,7 +175,7 @@ TEST(Share, SharesUniquePointer)
     std::unique_ptr<Derived> p(new Derived());
```
yhmtsai, line 175: could you add a test for `share(make_unique<Derived>())` check?

yhmtsai (approved):
LGTM. adding one more test for checking intermediate data

tcojean (approved):
LGTM

PR #1027, opened 2022-04-20
Distinguish between structural and numerical zeros in Ell/Sellp

### LINKED ISSUE #1018: Structural Zeros in ELL/SELL-P
After our conversion kernels have been unified, it should now be much easier to change the padding entries in ELL/SELL-P from
`col = 0, value = 0` to `col = invalid_index<IndexType>(), value = 0` to make it possible to distinguish between structural and numerical zeros. Our tests are comprehensive enough that we should be able to check whether all changes have been done consistently.

upsj:
This changes the column index for structural zeros/padding to `invalid_index<...>` to enable us to distinguish between numerical and structural zeros. It also leads to the important property that each column index occurs only once in each row, which  is necessary to build reusable generic matrix assembly structures for all sparse matrix formats.

Closes #1018

### DIFF
--- a/common/unified/matrix/ell_kernels.cpp
+++ b/common/unified/matrix/ell_kernels.cpp
@@ -84,7 +84,8 @@ void fill_in_matrix_data(std::shared_ptr<const DefaultExecutor> exec,
             const auto end = row_ptrs[row + 1];
             auto out_idx = row;
             for (auto i = begin; i < begin + num_cols; i++) {
-                cols[out_idx] = i < end ? in_cols[i] : 0;
+                cols[out_idx] =
+                    i < end ? in_cols[i] : invalid_index<IndexType>();
                 values[out_idx] = i < end ? in_vals[i] : zero(values[out_idx]);
                 out_idx += stride;
             }
@@ -104,20 +105,20 @@ void fill_in_dense(std::shared_ptr<const DefaultExecutor> exec,
                    const matrix::Ell<ValueType, IndexType>* source,
                    matrix::Dense<ValueType>* result)
 {
-    // TODO simplify once we can guarantee unique column indices outside padding
+    // ELL is stored in column-major, so we swap row and column parameters
     run_kernel(
         exec,
-        [] GKO_KERNEL(auto row, auto cols, auto ell_cols, auto ell_stride,
-                      auto in_cols, auto in_vals, auto out) {
-            for (int64 ell_col = 0; ell_col < ell_cols; ell_col++) {
-                const auto ell_idx = ell_col * ell_stride + row;
-                const auto col = in_cols[ell_idx];
-                const auto val = in_vals[ell_idx];
-                out(row, col) += val;
+        [] GKO_KERNEL(auto ell_col, auto row, auto ell_stride, auto in_cols,
+                      auto in_vals, auto out) {
+            const auto ell_idx = ell_col * ell_stride + row;
+            const auto col = in_cols[ell_idx];
+            const auto val = in_vals[ell_idx];
+            if (col != invalid_index<IndexType>()) {
+                out(row, col) = val;
             }
         },
-        source->get_size()[0], source->get_size()[1],
-        source->get_num_stored_elements_per_row(),
+        dim<2>{source->get_num_stored_elements_per_row(),
+               source->get_size()[0]},
         static_cast<int64>(source->get_stride()), source->get_const_col_idxs(),
         source->get_const_values(), result);
 }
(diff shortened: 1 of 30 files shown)


reference/matrix/csr_kernels.cpp
```diff
@@ -783,14 +783,9 @@ void convert_to_hybrid(std::shared_ptr<const ReferenceExecutor> exec,
          i++) {
         for (size_type j = 0; j < result->get_ell_stride(); j++) {
             result->ell_val_at(j, i) = zero<ValueType>();
-            result->ell_col_at(j, i) = 0;
+            result->ell_col_at(j, i) = invalid_index<IndexType>();
         }
     }
-    for (size_type i = 0; i < result->get_coo_num_stored_elements(); i++) {
```
yhmtsai, line 789: coo may have more space than the input

reference/matrix/dense_kernels.cpp
```diff
@@ -571,36 +571,30 @@ void convert_to_hybrid(std::shared_ptr<const ReferenceExecutor> exec,
          i++) {
         for (size_type j = 0; j < result->get_ell_stride(); j++) {
             result->ell_val_at(j, i) = zero<ValueType>();
-            result->ell_col_at(j, i) = 0;
+            result->ell_col_at(j, i) = invalid_index<IndexType>();
         }
     }
-    for (size_type i = 0; i < result->get_coo_num_stored_elements(); i++) {
```
yhmtsai, line 577: same

reference/matrix/ell_kernels.cpp
```diff
@@ -203,7 +207,10 @@ void fill_in_dense(std::shared_ptr<const ReferenceExecutor> exec,
 
     for (size_type row = 0; row < num_rows; row++) {
         for (size_type i = 0; i < num_stored_elements_per_row; i++) {
-            result->at(row, source->col_at(row, i)) += source->val_at(row, i);
+            const auto col = source->col_at(row, i);
+            if (col != invalid_index<IndexType>()) {
+                result->at(row, col) = source->val_at(row, i);
```
yhmtsai, line 212: more than 1 entry share the same col index

common/unified/matrix/ell_kernels.cpp
```diff
@@ -104,20 +105,20 @@ void fill_in_dense(std::shared_ptr<const DefaultExecutor> exec,
...
-                const auto col = in_cols[ell_idx];
-                const auto val = in_vals[ell_idx];
-                out(row, col) += val;
+        [] GKO_KERNEL(auto ell_col, auto row, auto ell_stride, auto in_cols,
+                      auto in_vals, auto out) {
+            const auto ell_idx = ell_col * ell_stride + row;
+            const auto col = in_cols[ell_idx];
+            const auto val = in_vals[ell_idx];
+            if (col != invalid_index<IndexType>()) {
+                out(row, col) = val;
```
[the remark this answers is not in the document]
upsj: I think that's a property we should avoid, to make matrix assembly/update easier. This would also lead to correctness issues in preconditioners.

reference/matrix/csr_kernels.cpp
upsj: why is that?

reference/matrix/sellp_kernels.cpp
```diff
@@ -215,8 +219,10 @@ void fill_in_dense(std::shared_ptr<const ReferenceExecutor> exec,
             }
             for (size_type i = slice_sets[slice]; i < slice_sets[slice + 1];
                  i++) {
-                result->at(global_row, col_idxs[row + i * slice_size]) +=
-                    vals[row + i * slice_size];
+                const auto col = col_idxs[row + i * slice_size];
+                if (col != invalid_index<IndexType>()) {
+                    result->at(global_row, col) += vals[row + i * slice_size];
```
fritzgoebel, line 224: I guess if we don't allow ell to have multiple entries with the same col idx, we shouldn't for sellp as well?
```suggestion
                    result->at(global_row, col) = vals[row + i * slice_size];
```

fritzgoebel (approved):
LGTM. Nice work, simplifies Ell/Sellp quite a bit.
I agree that we should require unique entries rather than allow for multiple entries with the same column index in the same row.

reference/matrix/csr_kernels.cpp
yhmtsai: when converting to hybrid, it always recompute the config not preserve any higher allowance. Thus, it's not a issue currently.

yhmtsai (approved):
LGTM. For matrix assembly, does it make sense that user allocate more space in the beginning such that the assembly phase does not need reallocate when adding a new entry?
However, I do not know where application change the matrix data or add the new entry.
My above question may only make sense when they need to add limited entries in some iterative process

reference/matrix/csr_kernels.cpp
upsj: I see! IMO, Coo should never have padding entries, so this didn't really occur to me.

PR #1029, opened 2022-04-20
preconditioner Ic without the solver type

yhmtsai:
This PR makes the preconditioner IC wrapper without SolverType. It moves the solver type into the factory operator.
When we provide the solver, the SolverType is useless but we still need to put it into template.
the type requirement is the same as original.
`Ic<SolverType>::build().on(exec);` -> `IcWrapper::build().with_l_solver_factory(IcWrapper<SolverType>(exec)).on(exec)`
or any `IcWrapper::build().with_l_solver_factory(linop_factory).on(exec);`
It's also good for ResourceManager.

TODO:
- [ ] change the explicit type as when merging SolverBase
- [ ] add IluWrapper

### DIFF
--- a/include/ginkgo/core/base/composition.hpp
+++ b/include/ginkgo/core/base/composition.hpp
@@ -45,6 +45,21 @@ namespace gko {
 
 
 /**
+ * CompositeionBase provides the interface
+ */
+class CompositionBase {
+public:
+    /**
+     * Returns a list of operators of the composition.
+     *
+     * @return a list of operators
+     */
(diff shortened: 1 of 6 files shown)


upsj:
The whole factorization preconditioner setup needs to be reworked (https://github.com/ginkgo-project/ginkgo/issues/472), do you need this urgently? I am not really happy with adding a second *Wrapper type that does the same thing and likely removing it with 2.0 again.

yhmtsai:
close this in favor of #1811 and #1828 which use the same class as the current workaround

Issue #1006, opened 2022-03-30
Should gko::Array be gko::array?


yhmtsai:
According to our contribuiting guideline [naming scheme - structures and classes](https://github.com/ginkgo-project/ginkgo/wiki/Contributing-guidelines#structures-and-classes), gko::array should be the correct term not gko::Array.
If we keep following the guideline, we can create the alias in release 1.x and annouce a interface change/breaking in release 2.
if not, we need to adjust the guideline.

pratikvn:
+1. I think to make it consistent, we should change it to `gko::array`. I dont think we can really update the rule to make it consistent. This would also mean updating `ConstArrayView` to `const_array_view` and so on.

PR #1031, opened 2022-04-21
Deprecate Array/ConstArrayView and use array/const_array_view

yhmtsai:
According to https://github.com/ginkgo-project/ginkgo/issues/1006, we should stay with the contributed guideline.

It moves the actual class definition to array/const_array_view but adds the alias Array/ConstArrayView to avoid interface breaking.
The forward declaration needs to be snake_case, not CamelCase.

TODO:
- [ ] change all occurrences to snake_case in ginkgo
- [ ] mark the alias deprecated https://en.cppreference.com/w/cpp/language/attributes/deprecated to notify ginkgo users.


upsj (approved):
LGTM!

pratikvn (approved):
LGTM!

test/test_install/test_install.cpp
```diff
@@ -203,8 +203,8 @@ int main()
     // core/base/array.hpp
     {
         using type1 = int;
-        using ArrayType = gko::Array<type1>;
-        ArrayType test;
+        using arrayType = gko::array<type1>;
+        arrayType test;
```
yhmtsai, line 207:
```suggestion
        using array_type = gko::array<type1>;
        array_type test;
```

yhmtsai (commented):
some minor changes

PR #1032, opened 2022-04-26
Optimize matrix generation

### LINKED ISSUE #862: Improve random matrix generation
Currently, our random matrix generation is slow for even relatively medium sizes because we allocate a fully dense matrix data object and then fill values randomly into it. 

It might be useful to allocate and fill only sparse matrix data when generating the sparse matrix formats.

upsj:
Our current matrix generation shuffles the vector of all column indices to avoid duplicate column indices in generation. Instead, this PR uses a bitvector to keep track of which columns are already present (or absent, for very dense rows). This speeds up the matrix_cuda test from 10s to 8s in my experiments.

Closes #862

### DIFF
--- a/core/test/utils/matrix_generator.hpp
+++ b/core/test/utils/matrix_generator.hpp
@@ -214,7 +214,7 @@ matrix_data<ValueType, IndexType> generate_random_triangular_matrix_data(
             present_cols.assign(size, true);
             // remove max_row_nnz - nnz_in_row entries from present_cols
             size_type count = max_row_nnz;
-            while (nnz_in_row > count) {
+            while (count > nnz_in_row) {
                 const auto new_col = col_dist(engine);
                 if (present_cols[new_col]) {
                     present_cols[new_col] = false;
--- a/core/test/utils/matrix_generator.hpp
+++ b/core/test/utils/matrix_generator.hpp
@@ -37,7 +37,9 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <algorithm>
 #include <iterator>
 #include <numeric>
+#include <random>
 #include <type_traits>
+#include <unordered_set>
 #include <vector>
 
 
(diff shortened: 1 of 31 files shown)


core/test/utils/matrix_generator.hpp
```diff
@@ -37,7 +37,9 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <algorithm>
 #include <iterator>
 #include <numeric>
+#include <random>
 #include <type_traits>
+#include <unordered_set>
```
upsj, line 42:
```suggestion
```

pratikvn (approved):
Thanks for adding this. LGTM!

core/test/utils/matrix_generator.hpp
```diff
@@ -234,16 +282,15 @@ matrix_data<ValueType, IndexType> generate_random_triangular_matrix_data(
 template <typename MatrixType = matrix::Dense<>, typename NonzeroDistribution,
           typename ValueDistribution, typename Engine, typename... MatrixArgs>
 std::unique_ptr<MatrixType> generate_random_triangular_matrix(
-    size_type num_rows, size_type num_cols, bool ones_on_diagonal,
-    bool lower_triangular, NonzeroDistribution&& nonzero_dist,
-    ValueDistribution&& value_dist, Engine&& engine,
-    std::shared_ptr<const Executor> exec, MatrixArgs&&... args)
+    size_type size, bool ones_on_diagonal, bool lower_triangular,
```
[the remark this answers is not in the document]
upsj: if you need this explicitly, you can modify data.size after creating a square matrix

core/test/utils/matrix_generator.hpp
```diff
@@ -82,22 +84,48 @@ matrix_data<ValueType, IndexType> generate_random_matrix_data(
...
+            }
+            for (IndexType col = 0; col < num_cols; col++) {
+                if (present_cols[col]) {
+                    data.nonzeros.emplace_back(
+                        row, col,
+                        detail::get_rand_value<ValueType>(value_dist, engine));
+                }
+            }
+        } else {
+            // add nnz_in_row entries to present_cols
```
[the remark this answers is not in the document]
keldu: If I interpret the constructor right, then no reservation has been done.
Since we do know that we will have exactly `nnz_in_row` elements, then we can safely reserve that amount

core/test/utils/matrix_generator.hpp
```diff
@@ -82,22 +84,48 @@ matrix_data<ValueType, IndexType> generate_random_matrix_data(
...
+            present_cols.assign(num_cols, true);
+            // remove num_cols - nnz_in_row entries from present_cols
+            size_type count = num_cols;
+            while (count > nnz_in_row) {
+                const auto new_col = col_dist(engine);
+                if (present_cols[new_col]) {
+                    present_cols[new_col] = false;
+                    count--;
+                }
+            }
```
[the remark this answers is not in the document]
keldu: This overshoots the necessary allocation a bit.

core/test/utils/matrix_generator.hpp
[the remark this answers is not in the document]
keldu: I somehow hyperfocused on the inner loop and forgot the outer loop. Either an average would be required, but it's generally hard to guess I think.

core/test/utils/matrix_generator.hpp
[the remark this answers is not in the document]
keldu: Same as above. Missed the outer loop.

keldu (approved):

thoasm (approved):
LGTM, well done!

include/ginkgo/core/base/matrix_data.hpp
```diff
@@ -190,6 +191,7 @@ struct matrix_data {
     matrix_data(dim<2> size_, RandomDistribution&& dist, RandomEngine&& engine)
         : size{size_}
     {
+        nonzeros.reserve(size[0] * size[1]);
```
yhmtsai, line 194: reserving the dense size is quite easy to exceed the memory limit

core/test/utils/matrix_generator.hpp
```diff
@@ -163,52 +189,72 @@ std::unique_ptr<MatrixType> generate_random_matrix(
...
-        // select a subset of `nnz_in_row` column indexes, and fill these
-        // locations with random values
-        std::shuffle(begin(col_idx), end(col_idx), engine);
-        // add non-zeros
-        bool has_diagonal{};
-        for (size_type nz = 0; nz < nnz_in_row; ++nz) {
-            auto col = col_idx[nz];
-            // skip non-zeros outside triangle
-            if ((col > row && lower_triangular) ||
-                (col < row && !lower_triangular)) {
```
yhmtsai, line 191: Is it not allowed to generate the rectangular matrix filling upper/lower part?

include/ginkgo/core/base/matrix_data.hpp
upsj: this is a constructor for a dense random matrix, so that is intended.

core/test/utils/matrix_generator.hpp
upsj: if you want something like that, you can call the function and adjust the matrix_data::size afterwards. This doesn't seem worth the effort to me, since we don't use rectangular triangular matrices anywhere.

yhmtsai: no, you can not just adjust the matrix size. if generate the lower part in the tall and skinny matrix, it will also fill the value in the nonsquare part.

upsj: sure, you only need to start with dimension `min(rows, cols)` and change it to `{rows, cols}` afterwards

upsj: sorry, my mistake, I was not thinking of the trailing part. However, I still don't see a place where we need this.

upsj:
rebase!

PR #1033, opened 2022-04-26
Add an option for warmup solver iterations config

tcojean:
Small convenience PR to configure the iterations for a solver's warmup

### DIFF
--- a/benchmark/run_all_benchmarks.sh
+++ b/benchmark/run_all_benchmarks.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 ################################################################################
 # Environment variable detection
 #
--- a/benchmark/run_all_benchmarks.sh
+++ b/benchmark/run_all_benchmarks.sh
@@ -1,33 +1,40 @@
+#!/bin/bash
 ################################################################################
 # Environment variable detection
+#
+
+print_default() {
+    local var=$1
+    echo "$var  environment variable not set - assuming \"${!var}\"" 1>&2
+}
 
 if [ ! "${BENCHMARK}" ]; then
     BENCHMARK="spmv"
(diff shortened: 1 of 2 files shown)


benchmark/run_all_benchmarks.sh
```diff
@@ -1,33 +1,40 @@
+#!/bin/bash
 ################################################################################
 # Environment variable detection
+#
+
+print_default() {
```
pratikvn, line 6: :+1:

benchmark/run_all_benchmarks.sh
```diff
@@ -1,33 +1,40 @@
+#!/bin/bash
```
pratikvn, line 1: Maybe 
```suggestion
#!/usr/bin/env bash
```

pratikvn (approved):
LGTM!

benchmark/solver/solver.cpp
```diff
@@ -406,7 +409,8 @@ void solve_system(const std::string& solver_name,
         for (auto _ : ic.warmup_run()) {
             auto x_clone = clone(x);
             auto precond = precond_factory.at(precond_name)(exec);
-            auto solver = generate_solver(exec, give(precond), solver_name)
+            auto solver = generate_solver(exec, give(precond), solver_name,
+                                          FLAGS_warmup_max_iters)
```
upsj, line 413: We are writing the number of iterations out from it_logger, this may give us too small iteration counts. Maybe we should combine this with the ResidualLogger run below?

upsj (changes requested):
LGTM in general!

yhmtsai (approved):
agree with @upsj comments on it_logger. others lgtm

benchmark/solver/solver.cpp
upsj: I actually encountered a related issue in #750 where the iteration count is not available if we don't do warmup iterations.

upsj (approved):

PR #1035, opened 2022-04-27
Add logger highlighting inefficient allocation and copy patterns

### LINKED ISSUE #804: Make executor mismatch computations explicitly visible
Due to our first focus on functionality rather than performance, which is facilitated through the usage of `temporary_clone` functionality, we can transparently make executor copies, which is not ideal for performance. While these can be "debugged" through loggers, it might be useful for users to prevent these copies for the sake of performance and re-write their code. 

To resolve this, we should decide which transparent copies are allowable (host<-> host) and which are not (device<->host and host<->device), set a parent executing executor, and then throw in `temporary_clone`, if we are indeed copying to and from the non-allowed paths.

upsj:
Above a certain size threshold, we log

* Repeated allocations/deallocations of the same size
* Repeated cross-executor copies from/to the same data

We count how often these events take place and output messages at counts equal to powers of 10.

Closes #804


include/ginkgo/core/log/performance_hint.hpp
```diff
@@ -0,0 +1,133 @@
...
+#include <ginkgo/core/log/logger.hpp>
+
+
+namespace gko {
+namespace log {
+
+
+/**
+ * PerformanceHint is a Logger which analyzes the performance of the application
+ * and outputs hints for unnecessary copies and allocations.
```
MarcelKoch, line 51: I think we should have a list of patterns that we check for here.

include/ginkgo/core/log/stream.hpp
```diff
@@ -206,16 +206,15 @@ class Stream : public Logger {
...
+    explicit Stream(std::shared_ptr<const gko::Executor> exec,
+                    const Logger::mask_type& enabled_events, std::ostream& os,
+                    bool verbose)
+        : Logger(exec, enabled_events), os_(&os), verbose_(verbose)
     {}
 
 
 private:
-    std::ostream& os_;
+    std::ostream* os_;
```
yhmtsai, line 217: is any consideration here?

upsj: yes, reference members are a bad idea in general, because every assignment to the object gets passed through to the member, since the reference can't be reassigned.

core/log/performance_hint.cpp
```diff
@@ -0,0 +1,190 @@
...
+                                        const Executor* to,
+                                        const uintptr& location_from,
+                                        const uintptr& location_to,
+                                        const size_type& num_bytes) const
+{
+    if (num_bytes > copy_size_limit_ && from != to) {
+        const auto count1 = ++copy_src_histogram_[location_from];
+        const auto count2 = ++copy_dst_histogram_[location_to];
+        if (should_log_count(count1)) {
+            print_copy_from_message(log(), location_from, count1);
```
pratikvn, line 149: I think it might be useful to also print the executor type ? Might be a bit ugly to do dynamic casts to all the concrete executor types, but I think that might be useful information.

This might also be a time to add a get executor string method to the base Executor class and have the concrete classes store and return the names.

include/ginkgo/core/log/performance_hint.hpp
```diff
@@ -0,0 +1,133 @@
...
+    std::ostream& log() const;
+
+    std::ostream* os_;
+    mutable std::unordered_map<uintptr_t, size_type> allocation_sizes_;
+    mutable std::unordered_map<size_type, int> allocation_histogram_;
+    mutable std::unordered_map<uintptr_t, int> copy_src_histogram_;
+    mutable std::unordered_map<uintptr_t, int> copy_dst_histogram_;
+    size_type allocation_size_limit_ = 16;
+    size_type copy_size_limit_ = 16;
+    size_type histogram_max_size = 1024;
```
pratikvn, line 121: Probably the defaults are not necessary since you already specify them in the create function.

include/ginkgo/core/log/stream.hpp
```diff
@@ -188,7 +188,7 @@ class Stream : public Logger {
     static std::unique_ptr<Stream> create(
         std::shared_ptr<const Executor> exec,
         const Logger::mask_type& enabled_events = Logger::all_events_mask,
-        std::ostream& os = std::cout, bool verbose = false)
+        std::ostream& os = std::cerr, bool verbose = false)
```
pratikvn, line 191: Probably better to have `clog`, so that it is properly buffered and in order ?

pratikvn (commented):
A minor comment on having the type of executor info displayed as well. Otherwise looks really good!

fritzgoebel (approved):
LGTM

yhmtsai (approved):
LGTM in general. With record logger, we should be able to notice where to give the message.
one nit: the count based on size may be from different sources, but I have no idea how to deal with it by the current interface. It may be good enough in the current case thuogh.

include/ginkgo/core/log/stream.hpp
upsj: how would that improve results? using clog means that log messages may be delayed an arbitrary amount of time, and buffering has no impact on the ordering of messages.

core/log/performance_hint.cpp
```diff
@@ -0,0 +1,190 @@
...
+           << " that may point to unnecessary data transfers.\n";
+}
+
+
+void print_copy_to_message(std::ostream& stream, gko::uintptr ptr, int count)
+{
+    stream << "Observed " << count << " cross-executor copies to "
+           << location_name(ptr)
+           << " that may point to unnecessary data transfers.\n";
+}
```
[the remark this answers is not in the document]
upsj: we record both, since the logger is likely only attached to one of the two executors involved

core/log/performance_hint.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+
+void PerformanceHint::on_free_completed(const Executor* exec,
+                                        const uintptr& location) const
+{
+    const auto it = allocation_sizes_.find(location);
+    if (it != allocation_sizes_.end()) {
+        const auto size = it->second;
+        allocation_sizes_.erase(it);
+        const auto count = ++allocation_histogram_[size];
```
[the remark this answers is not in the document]
upsj: yes, that is the best we can do with the available information, unless we start inspecting backtraces (for which there is no portable API yet)

core/test/log/performance_hint.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+    std::vector<gko::array<char>> data;
+    data.reserve(10);
+    for (int i = 0; i < 10; i++) {
+        data.emplace_back(exec, 32);
+    }
+    // by now we have forgotten the first allocation.
+    while (!data.empty()) {
+        data.pop_back();
+    }
```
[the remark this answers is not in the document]
upsj: yes, that is intended behavior: We first create 10 entries, and then destroy them in the reverse order, meaning the first element to be created (last to be destroyed) will no longer be part of the logger's allocation_sizes_ history.

core/log/performance_hint.cpp
upsj: name_demangling should work there :)

upsj: actually it doesn't work too well, since I don't keep the information which allocator a pointer is associated with, and this doesn't work with `print_status`. Storing the executor would work for copies, but not for allocations.

pratikvn: But you do have the `from` and `to` executors available in the function, that you can propagate to the print functions ?

core/log/performance_hint.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+void PerformanceHint::print_status() const
+{
+    for (auto entry : allocation_histogram_) {
+        if (entry.second >= 10) {
+            print_allocation_message(log(), entry.first, entry.second);
+        }
+    }
+    for (auto entry : copy_src_histogram_) {
+        if (entry.second >= 10) {
```
pratikvn, line 170: Maybe it makes sense to print only for the largest count, instead of printing for every count that we track ?

pratikvn (approved):
Should be very useful! Thanks for adding this. LGTM!

core/log/performance_hint.cpp
upsj: they are no longer available in the print_status function though, which is the only issue here

core/log/performance_hint.cpp
upsj: that may hide some issues, with people having to discover them step by step, instead of seeing all of them in a single spot?

upsj: I don't follow. The histogram contains only a single count for each size, so the output of `print_status()` would be
```
[PERFORMANCE] >>> Observed 110 allocate-free pairs of size x that may point to unnecessary allocations.
```

pratikvn: Okay, maybe I misunderstood something. What I mean is [something like here](https://github.com/ginkgo-project/ginkgo/pull/1035/files#diff-28c174a9777e93aa8fcb0167c2cadb033990195e8bb0edb72c98fb0d208f7ea0R62)

upsj: That is independent of `print_status()`: The output you're seeing is immediate after the allocation took place, thus if the application has its own status output, the interspersed logger messages will allow users to see where this unnecessary allocation took place. print_status() is meant more as a tool to print a summary at the end.

PR #1037, opened 2022-05-04
Fix Logger in Mixin Usage

MarcelKoch:
This PR fixes logging used in the following mixins:
- EnablePolymorphicObject
- EnableDefaultLinOpFactory
Before, any class derived from EnablePolymorphicObject would not log the PolymorphicObject events. In the EnableDefaultLinOpFactory case, the generate function with logging is implemented in LinOpFactory, but in the current version this function does not get called. Instead, the default generate function is called (this has also been noted by @upsj).

This removes the adding of loggers from the AbstractFactory, so now only EnableDefaultFactory adds logger, if the concrete type supports it. And additionally, some implementation is simplified. Some tests are added to make sure that we actually log all events.

### DIFF
--- a/core/test/base/executor.cpp
+++ b/core/test/base/executor.cpp
@@ -651,4 +651,131 @@ TEST(ExecutorDeleter, AvoidsDeletionForNullExecutor)
 }
 
 
+struct DummyLogger : public gko::log::Logger {
+    DummyLogger(std::shared_ptr<const gko::Executor> exec)
+        : gko::log::Logger(std::move(exec),
+                           gko::log::Logger::executor_events_mask |
+                               gko::log::Logger::operation_events_mask)
+    {}
+
+    void on_allocation_started(const gko::Executor* exec,
+                               const gko::size_type& num_bytes) const override
+    {
+        allocation_started++;
+    }
+    void on_allocation_completed(const gko::Executor* exec,
+                                 const gko::size_type& num_bytes,
+                                 const gko::uintptr& location) const override
+    {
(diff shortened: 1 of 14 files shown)


upsj:
Seems like we have a bit of parallel work going on, I'm working on the same thing, though with a slightly different approach

core/test/base/executor.cpp
```diff
@@ -651,4 +651,131 @@ TEST(ExecutorDeleter, AvoidsDeletionForNullExecutor)
...
+    }
+
+
+    void on_operation_completed(const gko::Executor* exec,
+                                const gko::Operation* op) const override
+    {
+        operation_completed++;
+    }
+
+    int mutable allocation_started = 0;
```
upsj, line 717: nit: const west

core/test/base/executor.cpp
```diff
@@ -651,4 +651,131 @@ TEST(ExecutorDeleter, AvoidsDeletionForNullExecutor)
...
+    int mutable alloctaion_completed = 0;
```
upsj, line 718:
```suggestion
    int mutable allocation_completed = 0;
```

upsj (approved):
LGTM! Real nice job with the tests. I think it doesn't make sense to have a factory for a non-polymorphic type, so maybe we should keep logger transfer at the AbstractFactory level? Then we can also get rid of the propagate_logger setup.

MarcelKoch:
This now contains the changes from #1038

keldu (approved):
LGTM

PR #1038, opened 2022-05-04
Fix factory executor

### LINKED ISSUE #301: `LinOpFactory::generate` not copying the data to the correct executor
Currently, the `LinOpFactory::generate(std::shared_ptr<const LinOp> input) const` is not copying the data to the executor of the factory.
This means that all `apply` calls (and similar) from the matrix will happen on the executor of the matrix (and copying all data temporarily to that executor), and not on the executor of the factory, which could slow down a solver significantly.

To solve this problem, we can use `make_temporary_clone`, which copies the data if it is on a different executor, or uses the same pointer is actually is on the correct executor.
The following code (to be inserted into `LinOpFactory::generate`, file: `include/ginkgo/core/base/lin_op.hpp`) should fix that, but it is not tested:
```c++
        auto generated = AbstractFactory::generate(
            std::shared_ptr<const LinOp>{make_temporary_clone(this->get_executor(), input.get()).get()});
```
I am not sure if the `temporary_clone` (given from `make_temporary_clone`) survives until after the call to `AbstractFactory::generate`, or not (since there is now a `std::shared_ptr` in between, which was not the case in the other use cases). This needs to be checked.
Also, we should add a test which checks if it actually works.

upsj:
Kind of a generalized version of #1037, could probably use a few of its tests.

Fixes #301


upsj:
integrated into #1037

PR #1042, opened 2022-05-13
Update CI, add HoreKa jobs and CB

tcojean:
Various CI improvements and fixes, add tests and benchmarks on HoreKa.

In terms of CI:
* Move the `except`, `only` to `rules` definitions (they cannot be used simultaneously).
+ Try to have a green pipeline/status by changing the synchronization jobs.

In terms of HoreKa usage:
+ Add a few jobs that run a new `test` step on HoreKa to show how it can be done. See a working example: https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/pipelines/537713203
+ Add an SpMV benchmark for CUDA on A100 using the representative matrices. When detecting a change to relevant files, it creates clickable jobs, one for building and one for testing.
  + Here is an example of the benchmarking job running through: https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/pipelines/540934042
  + And the commit in the ginkgo-data repository: https://github.com/ginkgo-project/ginkgo-data/commit/b90828953bd0c235cc418ab2d3f8776dfd4e4cc9

TODO:
* [x] Ensure benchmarking job succeeds and commits to our repo
* [x] Restore the entire pipeline
* [x] Move DPCPP CPU job to HoreKa if it works


.gitlab/rules.yml
```diff
@@ -0,0 +1,34 @@
+.pr_condition:
+  rules:
+    - if: $CI_COMMIT_BRANCH == "develop" || $CI_COMMIT_BRANCH == "master" || $CI_COMMIT_TAG
```
upsj, line 3: Can you add a comment explaining those conditions? Some of them are not 100% obvious to me

.gitlab-ci.yml
```diff
@@ -3,225 +3,78 @@ stages:
...
+    - .use_status-job-settings
   variables:
     STATUS_CONTEXT: "quick"
   # we always exit with the code 3 such that it will process when retrying
   script: |
-    curl -X POST -H "Accept: application/vnd.github.v3+json" -H "Authorization: token ${BOT_STATUS_TOKEN}" \
+    curl -s -X POST -H "Accept: application/vnd.github.v3+json" -H "Authorization: token ${BOT_STATUS_TOKEN}" \
     https://api.github.com/repos/ginkgo-project/ginkgo/statuses/${CI_COMMIT_SHA} \
     -d "{\"state\":\"success\",\"context\":\"ci/gitlab/${STATUS_CONTEXT}\",\"target_url\":\"${CI_PIPELINE_URL}\"}"
-    exit 3
```
upsj, line 194: This was put in here deliberately by Mike, since it allows us to rerun the entire pipeline after it finished, by just clicking "rerun".

upsj (approved):
LGTM! Really nice to see this go forward :)

.gitlab/scripts.yml
```diff
@@ -0,0 +1,203 @@
...
+  - if [[ ! "${SLURM_GRES}" =~ "^gpu*"  ]]; then export NVIDIA_VISIBLE_DEVICES=void; fi
+  - if [[ -z "${EXECUTOR}" ]]; then exit 233; fi
+  - if [[ -z "${SYSTEM_NAME}" ]]; then exit 233; fi
+  - if [[ -z "${BENCHMARK}" ]]; then exit 233; fi
+  # Set environment variables
+  - eval $(ssh-agent -s)
+  - echo "${BOT_KEY}" | tr -d '\r' | ssh-add - >/dev/null
+  - mkdir -p ~/.ssh
+  - chmod 700 ~/.ssh
+  - ssh-keyscan -t rsa github.com >>~/.ssh/known_hosts
```
yhmtsai, line 160:
```suggestion
  - ssh-keyscan -t rsa github.com >> ~/.ssh/known_hosts
```
nit

.gitlab/scripts.yml
```diff
@@ -0,0 +1,203 @@
...
+    - cd benchmark
+    - git clone ${BENCHMARK_REPO} data-repo
+    # Use the representative matrix list and launch benchmark
+    - export MATRIX_LIST_FILE=$PWD/data-repo/data/represent.list
+    - ./run_all_benchmarks.sh
+    # Publish the results
+    - rsync -av results/ data-repo/data/
+    - cd data-repo/data/
+    - ./build-list . > list.json
+    - ./agregate < list.json > agregate.json
```
pratikvn, line 198: nit: But the executable name might also have to be updated
```suggestion
    - ./aggregate < list.json > aggregate.json
```

.gitlab-ci.yml
```diff
@@ -480,60 +333,111 @@ build/cuda102/nompi/intel/cuda/debug/static:
...
+test/cuda110/mvapich2/gcc/cuda/debug/shared:
+  extends:
+    - .horeka_test_template
+    - .default_variables
+    - .full_test_condition
+    - .use_gko-cuda110-mvapich2-gnu9-llvm9-intel2020
+  variables:
+    USE_NAME: "cuda110-mvapich2-gcc-${CI_PIPELINE_ID}"
+    SLURM_PARTITION: "accelerated"
+    SLURM_GRES: "gpu:1"
```
pratikvn, line 363: Could we already run multi-GPU MPI CI by increasing the number of GPUs here ?

pratikvn (approved):
LGTM!

.gitlab-ci.yml
tcojean: In terms of the infrastructure, definitely. What I don't know is how well MPI runs through containers, specifically enroot. I believe the [pyxis](https://github.com/NVIDIA/pyxis) module is required, but I remember hearing that some extra change might be required somewhere in enroot or the slurm driver (maybe for the container to be able to run some slurm commands and talk to the slurm daemon even without slurm installed?).

.gitlab/scripts.yml
```diff
@@ -0,0 +1,203 @@
...
+  variables:
+    GIT_CLEAN_FLAGS: none
+  script:
+    # We should be reusing the previous image and build directory.
+    # Speedup the tests by limiting the number of OMP threads to CI parallelism.
+    - export OMP_NUM_THREADS=${CI_PARALLELISM}
+    # The test job should have the name name as the image job, only changing
+    # build to test
+    - cd ${CI_PROJECT_DIR}${CI_PROJECT_DIR_SUFFIX}
+    - cd ${CI_JOB_NAME/test/build}
```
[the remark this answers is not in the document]
tcojean: It depends on the point of view. Here, I'm saying that we suppose the `test` job has exactly the same as the `build` one, only that `build` is changed to `test` (so that we can do this simple string replace in the other direction to find the build directory).

.gitlab/scripts.yml
tcojean: Good point, but since this requires changing everything, including the `GPE`, I would rather not do this right now.

.gitlab/rules.yml
tcojean: I added some comments, tell me if that helps

CMakeLists.txt
```diff
@@ -428,4 +428,4 @@ if(GINKGO_CONFIG_LOG_DETAILED)
 else()
     FILE(READ ${PROJECT_BINARY_DIR}/minimal.log GINKGO_LOG_SUMMARY)
 endif()
-MESSAGE("${GINKGO_LOG_SUMMARY}")
+MESSAGE(STATUS "${GINKGO_LOG_SUMMARY}")
```
upsj, line 431: just checking: is this the right thing? it adds -- to the summary line, was it missing?

tcojean: It does this in one line, otherwise, it's not too bad. The problem is, without this the message is printed to `stderr`, and the custom runner interprets that as an error. In general, I think we should minimize the amount of `stderr` traffic for non-errors.

See (none) or NOTICE here: https://cmake.org/cmake/help/latest/command/message.html

.gitlab-ci.yml
```diff
@@ -276,20 +136,19 @@ trigger_pipeline:
         echo "Can not find the corresponding Pull Request"
       fi
   # Override variables condition from .pr_condition
-  only:
-    variables:
-      - $RUN_CI_TAG && $STATUS_CONTEXT == "quick"
+  rules:
+    - if: $RUN_CI_TAG && $STATUS_CONTEXT == "quick"
```
yhmtsai, line 140: the rules override the pr condition rules. https://docs.gitlab.com/ee/ci/yaml/yaml_optimization.html#merge-details
it will run the trigger_pipline in master, develop, tags now

yhmtsai (changes requested):

.gitlab-ci.yml
tcojean: good point! should be fixed

yhmtsai (approved):

Issue #51, opened 2018-05-11
Class naming convention


tcojean:
This is a summary of a small discussion which arose during MR #46 but is more general in Ginkgo. The purpose of this Issue is simply to discuss it and collect everyone's opinion on the subject.

The current class naming convention for Ginkgo is detailed [here](https://github.com/ginkgo-project/ginkgo/wiki/Contributing-guidelines#structures-and-classes) and can be summed up as the following (correct me if I'm wrong in any of this post):
+ Usual classes follow the function, variable, ..., naming convention, which is `class_name`.
+ Classes with some sort of polymorphic behavior use `ClassName`

The reason for the distinction is that polymorphic classes should be used with care and are "heavier" in a way (and capital case can be seen as heavy).

There are some problems though to distinguish what we consider to be polymorphic behavior. We for example have called the array class `Array` which have executors so it can be seen as polymorphic behavior, but there are no virtual methods and all of them are quite cheap.

I personally think this naming convention can tend to be confusing and at the minimum, the convention should be made very clear somewhere in the code. The reason for that is that the convention itself serves as a form of documentation: `ClassName` documents that this class is polymorphic and could imply significant overhead depending on what you do with it or which function you use.

I also personally like to be able to distinguish classes from functions, variable names, ..., in all cases so I'm in favor of a uniform `ClassName` convention. In addition, it's true some classes (`LinOp`s for example) are heavily polymorphic and should be used we care. I think we should have a standard way of adding a `@warning` in the documentation for those classes.

What do you think @ginkgo-project/developers.

tcojean:
I'm OK with using `snake_case`. If I proposed `CamelCase` it is mostly because this is what the Google C++ Style Guide proposes:
https://google.github.io/styleguide/cppguide.html#Type_Names

hartwiganzt:
So the Google C++ Style Guide differs from the Standard C++ types ?

gflegar:
Yes, and this is not the only difference (references vs pointers, and using exceptions are some more examples). Google has their own take on C++, and while I do agree with a lot of their decisions there are some that don't make much sense to me.

PR #1052, opened 2022-06-09
Rename non-polymorphic types to snake_case

upsj:
This means replacing Partition by partition and making it non-polymorphic (to match index_set) as well as renaming UninitializedArray to uninitialized_array.

Also fixes a small warning on MSVC

Related to #51


fritzgoebel (approved):
LGTM

PR #1053, opened 2022-06-11
Enable batch scaled solve with BatchDense system matrix

Slaedr:
Removes an outdated restriction in EnableBatchSolver core, enabling BatchDense to be used with pre-scaling of the batch system. Adds a test for this.


pratikvn (approved):
LGTM!

PR #1054, opened 2022-06-14
Explicitly set device id for each MPI call

MarcelKoch:
This PR abstracts the device guard mechanic from the cuda/hip module and makes it available in other modules. The `Executor` gets a new abstract method `get_scoped_device_id` (name not finalized) which returns an object that has the same semantics as the older cuda/hip device guards. This means that during the lifetime of the returned object, the device id is set to the one stored in the executor. If that concept does not apply (reference/omp/dpcpp?) then the object has no effect.

In the mpi wrapper, the `communicator` now stores an executor. As a result, it is not possible anymore to implicitly convert from `MPI_Comm` to `communicator`. The `window` and `request` classes also stores a `executor`. Using the stored executor, all MPI calls are guarded to use the correct device id, if necessary.

Note: Currently this is based on `distributed-solvers`, but I would rebase it onto `distributed-develop` after it has been reviewd to push forward merging the distributed stuff into develop.


MarcelKoch:
format!

MarcelKoch:
format!

include/ginkgo/core/base/scoped_device_id.hpp
```diff
@@ -0,0 +1,227 @@
...
+    scoped_device_id(const scoped_device_id&) = delete;
+
+    // Allow move construction.
+    scoped_device_id(scoped_device_id&&) = default;
+
+    // Prohibit copy assignment.
+    scoped_device_id& operator=(const scoped_device_id&) = delete;
+
+    // Allow move construction.
+    scoped_device_id& operator=(scoped_device_id&&) = default;
```
upsj, line 216: does a scope guard need to be movable?

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -399,7 +408,7 @@ class Matrix
      *              The default is the MPI_COMM_WORLD.
      */
     explicit Matrix(std::shared_ptr<const Executor> exec,
-                    mpi::communicator comm = mpi::communicator(MPI_COMM_WORLD));
+                    mpi::communicator comm);
```
upsj, line 411: these are redundant and may lead to people accidentally passing the wrong executor in the communicator?

include/ginkgo/core/distributed/vector.hpp
```diff
@@ -438,16 +447,15 @@ class Vector
...
+     * @param comm  Communicator associated with vector
      * @param partition  Partition of global rows
      * @param global_size  Global size of the vector
      * @param local_size  Processor-local size of the vector, uses local_size[1]
      *                    as the stride
      */
     explicit Vector(std::shared_ptr<const Executor> exec,
-                    mpi::communicator comm = mpi::communicator(MPI_COMM_WORLD),
-                    dim<2> global_size = {}, dim<2> local_size = {});
+                    mpi::communicator comm, dim<2> global_size = {},
```
upsj, line 457: same here

test/utils/executor.hpp
```diff
@@ -66,92 +64,41 @@ void init_executor(std::shared_ptr<gko::ReferenceExecutor> ref,
                    std::shared_ptr<gko::CudaExecutor>& exec)
 {
     ASSERT_GT(gko::CudaExecutor::get_num_devices(), 0);
-    exec = gko::CudaExecutor::create(0, ref);
+    exec = gko::CudaExecutor::create(
```
upsj, line 67: I think it would be better if we had MPI-specific test utils, otherwise running the tests manually might have unexpected side-effects

include/ginkgo/core/base/mpi.hpp
```diff
@@ -1502,9 +1631,11 @@ class window {
     void lock(int rank, lock_type lock_t = lock_type::shared, int assert = 0)
     {
         if (lock_t == lock_type::shared) {
+            auto guard = this->exec_->get_scoped_device_id();
             GKO_ASSERT_NO_MPI_ERRORS(
                 MPI_Win_lock(MPI_LOCK_SHARED, rank, assert, this->window_));
         } else if (lock_t == lock_type::exclusive) {
+            auto guard = this->exec_->get_scoped_device_id();
```
upsj, line 1638:
```suggestion
        auto guard = this->exec_->get_scoped_device_id();
        if (lock_t == lock_type::shared) {
            GKO_ASSERT_NO_MPI_ERRORS(
                MPI_Win_lock(MPI_LOCK_SHARED, rank, assert, this->window_));
        } else if (lock_t == lock_type::exclusive) {
```

upsj (approved):
LGTM! Some minor comments, but they shouldn't hold things up

cuda/test/base/scoped_device_id.cu
```diff
@@ -0,0 +1,84 @@
...
+    {}
+
+    std::shared_ptr<gko::ReferenceExecutor> ref;
+    std::shared_ptr<gko::CudaExecutor> hip;
+};
+
+
+TEST_F(ScopedDeviceId, SetsId)
+{
+    auto new_device_id = std::max(hip->get_num_devices() - 1, 0);
```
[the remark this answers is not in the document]
MarcelKoch: I will add a CI job on horeka using multiple gpus, since that is also interesting for our gpu-aware mpi stuff. Then these tests would make more sense.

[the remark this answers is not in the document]
upsj: Sounds great! Could probably do the same thing on our AMD machine, since we already had gitlab-runner set up there before. cc @tcojean

include/ginkgo/core/base/scoped_device_id.hpp
MarcelKoch: I thought I need it, because it is returned from a function and we can't rely on copy elision in that case. But I might be wrong. I will try again to use only copy elision.

upsj: ah right, this is only guaranteed in C++17, so we probably need it for some older compilers

include/ginkgo/core/distributed/matrix.hpp
MarcelKoch: I guess you mean that we should use `MPI_Comm` instead? Because otherwise a user could not use a specific communicator.

upsj: or just take an mpi::communicator. Though I'm not 100% happy with the ergonomics of that either

hip/base/scoped_device_id.hip.cpp
```diff
@@ -41,59 +41,57 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-        }
-    }
+namespace detail {
 
-    device_guard(device_guard& other) = delete;
 
-    device_guard& operator=(const device_guard& other) = delete;
-
-    device_guard(device_guard&& other) = delete;
+hip_scoped_device_id::hip_scoped_device_id(int device_id)
```
pratikvn, line 51: I guess this is not a interface break, because this is not a user-facing header file. But this would break ABI though.

pratikvn (approved):
Some minor comments, but otherwise looks good.

include/ginkgo/core/base/mpi.hpp
```diff
@@ -78,6 +78,63 @@ static constexpr bool is_gpu_aware()
...
+                local_rank = std::stoi(str);
+            } else if (auto str = std::getenv("OMPI_COMM_WORLD_LOCAL_RANK")) {
+                local_rank = std::stoi(str);
+            } else if (auto str = std::getenv("MPI_LOCALRANKID")) {
+                local_rank = std::stoi(str);
+            } else if (auto str = std::getenv("SLURM_LOCALID")) {
+                local_rank = std::stoi(str);
+            } else {
+                local_rank = mpi_node_local_rank(comm);
+            }
```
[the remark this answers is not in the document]
MarcelKoch: Yes, that's true, and that's why I've added the default case. So if we can't determine the local rank from environment variables (which is think is preferable) then we use the open mpi construct for that.

include/ginkgo/core/base/mpi.hpp
```diff
@@ -1803,20 +1954,48 @@ class window {
...
 }  // namespace gko
 
 
+#else
+
+
+struct MPI_Comm {};
+
+
+constexpr MPI_Comm MPI_COMM_WORLD{};
```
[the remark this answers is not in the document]
MarcelKoch: I will remove this. It was only necessary for the tests, to have the same setup for the mpi/non-mpi tests, but I've split that up now.

include/ginkgo/core/base/mpi.hpp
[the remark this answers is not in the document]
pratikvn: I feel this this might create behaviour that the user not intend or want. For example, Summit uses variables from BSUB i think. So, it user sets it using those variables, then we just disregard them. In general, I think taking into account user settings from different places is a hard problem, and I dont have a solution right now. But that might be something we have to think about when interfacing to other applications.

MarcelKoch:
format!

include/ginkgo/core/base/mpi.hpp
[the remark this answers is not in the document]
MarcelKoch: In your Summit example, we would not be able to handle it anyway, so I guess nothing is lost.
For interfacing external libraries, if they have MPI already set up, they can just pass us in the local rank. I guess there needs to be a nice way to insert that information.

test/mpi/distributed/matrix.cpp
```diff
@@ -79,12 +79,13 @@ class MatrixCreation : public ::testing::Test {
...
         gko::distributed::Partition<local_index_type, global_index_type>;
     using matrix_data = gko::matrix_data<value_type, global_index_type>;
+    using local_matrix_type = gko::matrix::Csr<value_type, local_index_type>;
 
 
     MatrixCreation()
         : ref(gko::ReferenceExecutor::create()),
           size{5, 5},
-          comm(gko::mpi::communicator(MPI_COMM_WORLD)),
+          comm(gko::mpi::communicator(MPI_COMM_WORLD, exec)),
```
tcojean, line 88: Again for these, wouldn't the following work?
```suggestion
          comm{MPI_COMM_WORLD, exec},
```

include/ginkgo/core/base/mpi.hpp
```diff
@@ -413,7 +426,9 @@ class communicator {
      * @param color The color to split the original comm object
      * @param key  The key to split the comm object
      */
-    communicator(const MPI_Comm& comm, int color, int key)
+    communicator(const MPI_Comm& comm, int color, int key,
+                 std::shared_ptr<const Executor> exec)
```
tcojean, line 430: In this file, the `exec` now needs to be added to the documentations

include/ginkgo/core/base/scoped_device_id.hpp
```diff
@@ -0,0 +1,229 @@
...
+ */
+class cuda_scoped_device_id : public generic_scoped_device_id {
+public:
+    explicit cuda_scoped_device_id(int device_id);
+
+    ~cuda_scoped_device_id() noexcept(false) override;
+
+    cuda_scoped_device_id(cuda_scoped_device_id&& other) noexcept;
+
+    cuda_scoped_device_id& operator=(cuda_scoped_device_id&& other) noexcept;
```
tcojean, line 99: What I'm missing for now is what happens when not compiling `CUDA`. These methods are never defined then, no? Shouldn't they be to at least `GKO_NOT_COMPILED` like for executor functions? Since users could be using these anywhere since they are publicly accessible.

Same for HIP.

cuda/test/base/scoped_device_id.cu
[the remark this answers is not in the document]
tcojean: This should indeed be usable since I yesterday fixed the machines and runners, and reenabled nla-gpu1 as a target.

cuda/test/base/scoped_device_id.cu
```diff
@@ -0,0 +1,86 @@
...
+
+class ScopedDeviceId : public ::testing::Test {
+protected:
+    ScopedDeviceId()
+        : ref(gko::ReferenceExecutor::create()),
+          hip(gko::CudaExecutor::create(0, ref))
+    {}
+
+    std::shared_ptr<gko::ReferenceExecutor> ref;
+    std::shared_ptr<gko::CudaExecutor> hip;
```
tcojean, line 59: Read similar to a variable `int three = 2;`
```suggestion
    std::shared_ptr<gko::CudaExecutor> cuda;
```
(or just `exec`)

tcojean (approved):
LGTM!

I only have some small comments. In general, there's some documentation to be updated, in terms of having `exec` as a parameter or not, and to update to the new ways to `create` a distributed vector/matrix.

Did you check the sonarqube errors? You might have to manually bypass it, I believe it will not like the `noexcept(false)` on the destructor for example.

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -252,7 +252,7 @@ class Matrix
           Matrix<next_precision<ValueType>, LocalIndexType, GlobalIndexType>>,
       public DistributedBase {
     friend class EnableCreateMethod<Matrix>;
-    friend class EnablePolymorphicObject<Matrix, LinOp>;
+    friend class polymorphic_object_traits<Matrix>;
```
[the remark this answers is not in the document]
tcojean: It's maybe also worth showing the communicator creation now to express the executor is given there.

tcojean:
As a small comment or to begin a discussion, I think we might need to explain a bit better the `communicator` now. Because I think our wrapper deviates from MPI in terms of logic, while it makes perfect sense from an implementation point of view. We are now tying two independent things together, the MPI communicator which is a list of processes (and a shared global view), with the `local` device that the user wants the `current` process to execute on top of.

test/mpi/solver/solver.cpp
```diff
@@ -251,7 +252,8 @@ class Solver : public ::testing::Test {
             static_cast<gko::size_type>(part->get_part_size(comm.rank())),
             static_cast<gko::size_type>(nrhs)};
         auto dist_result =
-            DistVecType::create(ref, comm, global_size, local_size, stride);
+            DistVecType::create(gko::mpi::communicator(comm.get(), ref),
```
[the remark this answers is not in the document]
MarcelKoch: No, that doesn't work. I guess the curly brace initializer does not play well with the forwarding from the create method

include/ginkgo/core/distributed/vector.hpp
```diff
@@ -63,7 +63,7 @@ class Partition;
  * vector should be filled using the read_distributed method, e.g.
  * ```
  * auto part = Partition<...>::build_from_mapping(...);
- * auto vector = Vector<...>::create(exec, comm);
+ * auto vector = Vector<...>::create(exec, ref_comm);
```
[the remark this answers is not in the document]
MarcelKoch: I think that was a renaming error. But the documentation still needs to be updated

include/ginkgo/core/base/mpi.hpp
MarcelKoch: Overall, the communicator documentation needs to be overhauled a bit.

include/ginkgo/core/base/scoped_device_id.hpp
MarcelKoch: I added these to the device_hooks. I think that should be enough

MarcelKoch:
> There's also the question if we expect these executors to be unique when a device ID is involved within the communicators (process ranks must be unique within a communicator), and maybe a list of other things which might become confusing when tying these two concepts together.

I don't think we need to impose anything on the device IDs. Most (?) gpus allow you to run multiple mpi processes on the same gpu, but of course with less efficiency. So as long as the device id is valid, the program will run correctly. To run efficiently, the user has to take care of setting the device id correctly. I hope the `map_rank_to_device_id` can help with that.

MarcelKoch:
Now that I think of it, we could also add the executor to each communicator method. Then the communicator class would again be just a wrapper, without combining the executor.

core/base/mpi.cpp
```diff
@@ -0,0 +1,101 @@
...
+        // environment variable describing the node local rank, so we
+        // prioritize it. If no suitable environment variable is found
+        // we determine the node-local rank with MPI calls.
+        int local_rank;
+        int compare_result;
+        int is_initialized;
+        GKO_ASSERT_NO_MPI_ERRORS(MPI_Initialized(&is_initialized));
+        if (is_initialized) {
+            GKO_ASSERT_NO_MPI_ERRORS(
+                MPI_Comm_compare(comm, MPI_COMM_WORLD, &compare_result));
```
upsj, line 73: why does this function need to work at all if MPI is not yet initialized? I would actually prefer if it could fail loudly in that case

upsj (commented):
@MarcelKoch I think I would actually prefer passing the executor to MPI functions, as then we keep the MPI and Ginkgo concepts of communicators and executors separate and unchanged.

MarcelKoch:
@upsj I'm also leaning towards using the executor as a function argument. This will require quite some rework...

core/device_hooks/cuda_hooks.cpp
```diff
@@ -154,6 +160,26 @@ void CudaExecutor::set_gpu_property() {}
...
+
+
+cuda_scoped_device_id::cuda_scoped_device_id(int device_id) GKO_NOT_IMPLEMENTED;
+
+cuda_scoped_device_id::~cuda_scoped_device_id() noexcept(false)
+    GKO_NOT_IMPLEMENTED;
+
+
+cuda_scoped_device_id::cuda_scoped_device_id(
+    gko::detail::cuda_scoped_device_id&& other) noexcept GKO_NOT_IMPLEMENTED;
```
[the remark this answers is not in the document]
MarcelKoch: You're right. I will just implement the move operations fully, because they don't need any cuda/hip stuff anyway.

core/base/mpi.cpp
MarcelKoch: If the id was only deduced from the env variables, it could work without MPI initialized. But I think it's better to also have the comm-splitting approach available. This can be used if ginkgo should not use MPI_COMM_WORLD, but a different communicator, so I would like to keep that approach.

I think if I remove the checks, then it will throw if MPI is not initialized, which should be loud enough.

tcojean:
> I don't think we need to impose anything on the device IDs. Most (?) gpus allow you to run multiple mpi processes on the same gpu, but of course with less efficiency. So as long as the device id is valid, the program will run correctly. To run efficiently, the user has to take care of setting the device id correctly. I hope the `map_rank_to_device_id` can help with that.

I generally agree, I was mostly raising this point as an example of something which isn't obvious when looking at this new combined communicator. 

While that would require some rework, I think that having the executor as an argument to the communicator functions would feel more natural, e.g. when sending or receiving data or synchronizing, it feels natural to me that you might need an object representing the memory location/processor being used.

MarcelKoch:
I've now switched to passing an executor to each MPI call that uses a buffer. For now, I didn't add the executor, and also the device id guard to synchronization MPI calls (barrier, wait, ...). At least for the barrier that shouldn't matter, but I have to test if the wait works correctly.

I've dismissed the reviews because of the change.

include/ginkgo/core/base/mpi.hpp
```diff
@@ -503,9 +511,11 @@ class communicator {
      *                   MPI_Datatype.
      */
     template <typename SendType>
-    void send(const SendType* send_buffer, const int send_count,
-              const int destination_rank, const int send_tag) const
+    void send(std::shared_ptr<const Executor> exec, const SendType* send_buffer,
```
tcojean, line 514: The same story as before, now the function documentations are the ones which need to add the `exec` as `@param`. 

Also, it might make sense to give an `@note` in the `communicator` class to explain that communicator functions tied to a memory location require a `gko::Executor` to represent that, on top of the usual MPI parameters.

tcojean (approved):
LGTM!

Thanks for the changes, I think it's much better organized this way.

core/test/base/mtx_io.cpp
```diff
@@ -870,7 +870,7 @@ class DummyLinOp
       public gko::EnableCreateMethod<DummyLinOp<ValueType, IndexType>>,
       public gko::ReadableFromMatrixData<ValueType, IndexType>,
       public gko::WritableToMatrixData<ValueType, IndexType> {
-    friend class gko::EnablePolymorphicObject<DummyLinOp, gko::LinOp>;
+    friend class gko::polymorphic_object_traits<DummyLinOp>;
```
pratikvn, line 873: Are you sure this works as expected ? This class no longer has the `copy_from`, `clone` etc impls that `EnableAbstractPolymorphicObject` provided through the `EnablePolymorphicObject` interface.

include/ginkgo/core/base/scoped_device_id.hpp
```diff
@@ -0,0 +1,229 @@
...
+     * @param device_id  The device id to use within the scope.
+     */
+    scoped_device_id(const HipExecutor* exec, int device_id)
+        : scope_(std::make_unique<detail::hip_scoped_device_id>(device_id))
+    {}
+
+    /**
+     * Create a scoped device id from an OmpExecutor.
+     *
+     * This will pick the noop_scoped_device_id.
```
pratikvn, line 195:
```suggestion
     * Create a scoped device id from an DpcppExecutor.
     *
     * This will pick the noop_scoped_device_id.
```

include/ginkgo/core/matrix/sellp.hpp
```diff
@@ -258,8 +258,8 @@ class Sellp : public EnableLinOp<Sellp<ValueType, IndexType>>,
     /**
      * @copydoc Sellp::val_at(size_type, size_type, size_type)
      */
-    value_type val_at(size_type row, size_type slice_set, size_type idx) const
-        noexcept
+    value_type val_at(size_type row, size_type slice_set,
+                      size_type idx) const noexcept
```
pratikvn, line 262: I think the formatting is off. Maybe you have a different `clang-format` version?

pratikvn (commented):
A question on `polymorphic_object_traits`, but otherwise looks good. Could you maybe also do a quick evaluation if these device id set and reset affect the performance ? IIRC, previously even CUDA had a few issues with `cudaGetDeviceID` having a large overhead.

core/test/base/mtx_io.cpp
MarcelKoch: The classes still derive from `EnableAbstractPolymorphicObject` somewhere in the hierarchy. This only changes which class has access to the protected methods, mainly the constructor.

pratikvn: As far as I can see, only `LinOp`, `ReorderingBase` and `Criterion` and the `AbstractFactory` derive from `EnableAbstractPolymorphicObject`. So, any classes that derive from these should be okay. But the rest will not be ?

MarcelKoch:
@pratikvn Regarding the performance, I've tested this quite a while ago on crusher, and I didn't experience any notable difference. I've rerun a single benchmark, and it still has the same results, so I would say that this has no negative performance impact.

pratikvn (approved):
LGTM!

MarcelKoch:
format!

MarcelKoch:
rebase!

include/ginkgo/core/base/mpi.hpp
```diff
@@ -78,6 +76,16 @@ static constexpr bool is_gpu_aware()
...
 
 
+/**
+ * Maps each MPI rank to a single device id in a round robin manner.
+ * @param comm  used to determine the node-local rank, if no suitable
+ *              environment variable is available.
+ * @param num_devices  the number of devices per node.
+ * @return  device id that this rank should use.
+ */
+int map_rank_to_device_id(MPI_Comm comm, int num_devices);
```
upsj, line 86: This should hopefully find its way into #704 at some point

include/ginkgo/core/base/scoped_device_id.hpp
```diff
@@ -0,0 +1,228 @@
...
+
+    hip_scoped_device_id& operator=(hip_scoped_device_id&& other) noexcept;
+
+private:
+    int original_device_id_;
+    bool need_reset_;
+};
+
+
+}  // namespace detail
```
upsj, line 124: these could be moved to the .cpp file

include/ginkgo/core/base/scoped_device_id.hpp
```diff
@@ -0,0 +1,228 @@
...
+     * Create a scoped device id from an OmpExecutor.
+     *
+     * This will pick the noop_scoped_device_id.
+     *
+     * @param exec  Not used.
+     * @param device_id  Not used.
+     */
+    scoped_device_id(const DpcppExecutor* exec, int device_id)
+        : scope_(std::make_unique<detail::noop_scoped_device_id>())
+    {}
```
upsj, line 201: implementations don't need to live in the header

upsj (approved):
LGTM! Only minor suggestions

include/ginkgo/core/base/scoped_device_id.hpp
MarcelKoch: the class definition has to be in some header, because the implementation happens both in core and in cuda/hip. So we could keep it in the public headers or in a new private header. I would lean towards the private header

upsj: agree :)

include/ginkgo/core/base/scoped_device_id.hpp
MarcelKoch: I think this doesn't work. I've put the implementation of `EXEC::get_scoped_device_id` into the `ginkgo_*_device` library, so these need the constructor to be available. But since these libraries are not linked against core, this results in linker errors.
I think we have two options:
1. `scoped_device_id` in core, `EXEC::get_scoped_device_id` in header
2. `scoped_device_id` in header, `EXEC::get_scoped_device_id` in devices
The easiest would be 2., because then I would just drop the last commit. Not sure if there is an actual benefit to one of them.

MarcelKoch:
format!

core/base/scoped_device_id.hpp
```diff
@@ -0,0 +1,63 @@
+#ifndef GINKGO_CORE_BASE_SCOPED_DEVICE_ID_HPP
```
upsj: missing license header

include/ginkgo/core/base/scoped_device_id.hpp
```diff
@@ -155,50 +108,42 @@ class scoped_device_id {
...
-     * This will pick the noop_scoped_device_id.
+     * The resulting object will be a noop.
      *
      * @param exec  Not used.
      * @param device_id  Not used.
      */
-    scoped_device_id(const OmpExecutor* exec, int device_id)
-        : scope_(std::make_unique<detail::noop_scoped_device_id>())
-    {}
+    scoped_device_id(const OmpExecutor* exec, int device_id);
```
upsj, line 127: Since this type of operation is sometimes called scope guard, I would suggest trying to get `guard` into the class name

upsj (approved):

include/ginkgo/core/base/scoped_device_id.hpp
MarcelKoch: you're right, I also thought about that, but I seem to have forgotten about it again.

cuda/base/scoped_device_id.cpp
```diff
@@ -43,57 +40,60 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-class device_guard {
-public:
-    device_guard(int device_id) : original_device_id{}, need_reset{}
-    {
-        GKO_ASSERT_NO_CUDA_ERRORS(cudaGetDevice(&original_device_id));
-        if (original_device_id != device_id) {
-            GKO_ASSERT_NO_CUDA_ERRORS(cudaSetDevice(device_id));
-            need_reset = true;
-        }
+namespace detail {
```
pratikvn, line 47:
```suggestion
namespace detail {


```

pratikvn (approved):
LGTM!

PR #1056, opened 2022-06-18
Fix potential memory leak in Logger

upsj:
loggers store a shared_ptr to an executor, but don't provide any access to it (this appears to be an oversight). When adding a logger containing an executor to this executor, we prevent both from being freed, since they create an ownership cycle:

```cpp
{
    auto exec = ReferenceExecutor::create() // refcount 1
    auto logger = SomeLogger::create(exec); // refcount exec 2, refcount logger 1
    exec->add_logger(logger); // refcount exec 2, logger 2
} // refcount exec 1, logger 1, thus none of the two can be destroyed
```

Thus this PR removes the executor member and deprecates all corresponding constructors, prodiving replacements without the executor parameter.

### DIFF
--- a/include/ginkgo/core/log/papi.hpp
+++ b/include/ginkgo/core/log/papi.hpp
@@ -195,13 +195,23 @@ public:
      * Creates a Papi Logger.
      *
      * @param enabled_events  the events enabled for this Logger
-     * @param handle  the papi handle
+     */
+    [[deprecated("use single-parameter create")]] static std::shared_ptr<Papi>
+    create(std::shared_ptr<const gko::Executor>,
+           const Logger::mask_type& enabled_events = Logger::all_events_mask)
+    {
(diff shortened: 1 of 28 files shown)


upsj:
In all logger events, you have access to either the executor or a PolymorphicObject with a corresponding executor, so this should not be a limitation.

include/ginkgo/core/log/papi.hpp
```diff
@@ -213,10 +223,15 @@ class Papi : public Logger {
     const std::string get_handle_name() const { return name; }
 
 protected:
+    [[deprecated("use single-parameter constructor")]] explicit Papi(
+        const Logger::mask_type& enabled_events = Logger::all_events_mask)
+        : Papi(enabled_events)
```
MarcelKoch, line 228:
```suggestion
        : Logger(enabled_events)
```

MarcelKoch (approved):

PR #1058, opened 2022-06-21
numerical LU factorization

upsj:
This adds numerical LU factorization kernels to initialize the factors from the matrix and run the factorization. It uses the Csr lookup structures to implement the required sparse vector additions, and a sync-free waiting approach similar to our custom triangular solvers.


common/cuda_hip/factorization/lu_kernels.hpp.inc
```diff
@@ -0,0 +1,204 @@
...
+    if (row >= num_rows) {
+        return;
+    }
+    const auto warp =
+        group::tiled_partition<config::warp_size>(group::this_thread_block());
+    // first zero out this row of the factor
+    const auto factor_begin = factor_row_ptrs[row];
+    const auto factor_end = factor_row_ptrs[row + 1];
+    const auto lane = static_cast<int>(warp.thread_rank());
+    for (auto nz = factor_begin + lane; nz < factor_end; nz++) {
```
fritzgoebel, line 59:
```suggestion
    for (auto nz = factor_begin + lane; nz < factor_end; nz += config::warp_size) {
```
I think this should be strided access with the warp or did I miss something?

fritzgoebel (commented):
Looks great, but omp and dpcpp are missing completely. Maybe add them as GKO_NOT_IMPLEMENTED?

common/cuda_hip/factorization/lu_kernels.hpp.inc
upsj: great catch!

common/cuda_hip/factorization/lu_kernels.hpp.inc
```diff
@@ -0,0 +1,206 @@
...
+        const auto dep_block = dep / (default_block_size / config::warp_size);
+        const auto dep_local = dep % (default_block_size / config::warp_size);
+        auto val = vals[lower_nz];
+        const auto diag_idx = diag_idxs[dep];
+        const auto dep_end = row_ptrs[dep + 1];
+        assert(dep < row);
+        if (dep_block == block) {
+            // wait for a local dependency
+            while (!load(sh_ready, dep_local)) {
+                __threadfence();
```
MarcelKoch, line 126: would `__threadfence_block();` suffice here?

common/cuda_hip/factorization/lu_kernels.hpp.inc
```diff
@@ -0,0 +1,206 @@
...
+template <typename ValueType, typename IndexType>
+void factorize(std::shared_ptr<const DefaultExecutor> exec,
+               const IndexType* lookup_offsets, const int64* lookup_descs,
+               const int32* lookup_storage, const IndexType* diag_idxs,
+               matrix::Csr<ValueType, IndexType>* factors,
+               array<int>& tmp_storage)
+{
+    const auto num_rows = factors->get_size()[0];
+    if (num_rows > 0) {
+        tmp_storage.resize_and_reset(num_rows + 1);
```
MarcelKoch, line 193: isn't `tmp_storage.resize_and_reset(2)` enough? You are only using the first and last entry, and ignoring everything in between.

MarcelKoch (approved):
Looks really good, especially the cuda/hip implementation. I still left a few questions there.
Besides that, I guess tests with edge cases (0x0 matrix, ...) will follow at a later point.

common/cuda_hip/factorization/lu_kernels.hpp.inc
```diff
@@ -0,0 +1,206 @@
...
+    const int32* __restrict__ storage, const int64* __restrict__ row_descs,
+    const IndexType* __restrict__ diag_idxs, ValueType* __restrict__ vals,
+    int* __restrict__ ready, int* __restrict__ block_counter,
+    size_type num_rows)
+{
+    __shared__ int threadblock_offset;
+    __shared__ int sh_ready[default_block_size / config::warp_size];
+    if (threadIdx.x == 0) {
+        threadblock_offset = atomic_add(block_counter, 1) * default_block_size;
+    }
```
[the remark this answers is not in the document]
upsj: CUDA doesn't guarantee that new blocks are scheduled in a monotonic order (even though some NVIDIA libraries like cub rely on it), so just to be safe, we added this workaround. I'm also pretty annoyed with them for this behavior, but it might change due to more complex scheduling on multi-chiplet GPUs or so, I could imagine? Anyways, without this monotonicity, we could livelock by starving previous rows of compute time.

common/cuda_hip/factorization/lu_kernels.hpp.inc
upsj: probably, yes

common/cuda_hip/factorization/lu_kernels.hpp.inc
upsj: we use tmp_storage for two things: the atomic counter and 0/1 flags for each row

MarcelKoch: Oh, I see, I didn't realize that the `store` does the indexing internally.

common/cuda_hip/components/volatile.hpp.inc
```diff
@@ -0,0 +1,70 @@
...
+{
+    auto real = reinterpret_cast<const ValueType*>(values);
+    auto imag = real + 1;
+    return {load(real, 2 * index), load(imag, 2 * index)};
+}
+
+template <typename ValueType, typename IndexType>
+__device__ __forceinline__ void store(
+    ValueType* values, IndexType index,
+    std::enable_if_t<!is_complex_s<ValueType>::value, ValueType> value)
```
thoasm, line 55: I would prefer if you could have the SFINAE in the return type (`void`) of the `store` function, and not in the argument type itself (especially for readability reasons).

reference/test/factorization/lu_kernels.cpp
```diff
@@ -0,0 +1,212 @@
...
+        : ref(gko::ReferenceExecutor::create()),
+          storage_offsets{ref},
+          storage{ref},
+          row_descs{ref}
+    {}
+
+    void setup(const char* mtx_filename, const char* mtx_lu_filename)
+    {
+        std::ifstream s_mtx{mtx_filename};
+        std::ifstream s_mtx_lu{mtx_lu_filename};
```
thoasm, line 76: Requires header `<fstream>` (which is currently not included).

reference/test/factorization/lu_kernels.cpp
```diff
@@ -0,0 +1,212 @@
...
+            row_descs.get_data(), storage.get_data());
+    }
+
+    std::shared_ptr<const gko::ReferenceExecutor> ref;
+    gko::size_type num_rows;
+    std::unique_ptr<matrix_type> mtx;
+    std::unique_ptr<matrix_type> mtx_lu;
+    gko::array<index_type> storage_offsets;
+    gko::array<gko::int32> storage;
+    gko::array<gko::int64> row_descs;
```
thoasm, line 102: Requires both `<ginkgo/core/base/types.hpp>` and `<ginkgo/core/base/array.hpp>`

test/factorization/lu_kernels.cpp
```diff
@@ -0,0 +1,258 @@
...
+    void TearDown()
+    {
+        if (exec != nullptr) {
+            ASSERT_NO_THROW(exec->synchronize());
+        }
+    }
+
+    void initialize_data(const char* mtx_filename)
+    {
+        std::ifstream s_mtx{mtx_filename};
```
thoasm, line 90: Missing header `<fstream>`

test/factorization/lu_kernels.cpp
```diff
@@ -0,0 +1,258 @@
...
+    std::shared_ptr<matrix_type> mtx;
+    std::shared_ptr<matrix_type> mtx_lu;
+    std::shared_ptr<matrix_type> dmtx;
+    std::shared_ptr<matrix_type> dmtx_lu;
+    gko::array<index_type> storage_offsets;
+    gko::array<index_type> dstorage_offsets;
+    gko::array<gko::int32> storage;
+    gko::array<gko::int32> dstorage;
+    gko::array<gko::int64> row_descs;
+    gko::array<gko::int64> drow_descs;
```
thoasm, line 133: Missing header `ginkgo/core/base/types.hpp` and `ginkgo/core/base/array.hpp`

thoasm (approved):
LGTM!

thoasm:
Nit: maybe add a short description of this PR (before you merge).

upsj:
I'll wait for the pipelines to finish here, then rebase and merge directly, since the previous changes only related to Github Actions.

upsj:
The clang-cuda issues seem unrelated, so I'll still proceed

upsj:
rebase!

Issue #988, opened 2022-03-18
Detect and handle in-place operations


upsj:
We rarely document which operators support in-place operations `A->apply(x, x)`, but we should do that at some point. A possible solution would be comparing the pointers b and x to check whether they are equal, and in that case, and if something like `A->supports_inplace_apply()` is false, we create a temporary clone of x before applying it. Otherwise we could just throw an exception.

PR #1059, opened 2022-06-21
ScaledReordered interface

fritzgoebel:
This PR adds an interface for wrapping an inner operator (like a direct or iterative solver) into a (for now symmetrical, e.g. RCM) reordering via a reordering factory and diagonal scaling via one row and one column scaling diagonal matrix.

The `ScaledReordered` LinOp applies the given scaling and reordering to the system matrix and generates the inner operator on the modified problem. It manages three vectors, a right-hand side and a solution vector for the inner operator, and an intermediate vector used for the permutations.

With a permutation matrix `P`, a row scaling matrix `R`, and a column matrix `C`, the system matrix and vectors are scaled and permuted such that instead of the original system of equations
`A*x = b`
the inner operator solves 
`P*R*A*C*P^T*y = P*R*b`.
Hence, when generating a `ScaledReordered` LinOp, the new system matrix `P*R*A*C*P^T` is computed. When it is applied to vectors `b` and `x`, the right-hand side vector is modified to `P*R*b`, and the solution `y` to the inner system gives the solution `x` to the original system via `x = C*P^T*y`.
When using an inner operator that uses an initial guess, currently the initial guess for the inner system is computed by `y_0 = P*C^(-1)*x_0`, which is the same as using `x_0` for the system `P*R*A*x = P*R*b`. For the inverse diagonal scaling here I added an `inverse_apply` to the `matrix::Diagonal`. I still need to add tests for that.

Any feedback is welcome, especially on ideas regarding how to handle algorithms like RCM that generate a reordering, but also be able to handle MC64 that generates both a reordering and scaling coefficients and simpler equilibration techniques that only generate scaling coefficients.


fritzgoebel:
format!

fritzgoebel:
format!

yhmtsai (commented):
I am thinking about removing the ValueType and IndexType from the ScaledReordered.
set_cache_to you can use create_with_config_of to create cache.
some requirement of ValueType / IndexType can be done by run<type-list>
Will it use ScaledReordered even when only need scale or reorder part?
Not a Scaled and Reordered class to handle them separately and ScaledReordered inherit from Scaled and Reordered.

include/ginkgo/core/base/reordered.hpp
```diff
@@ -0,0 +1,240 @@
...
+        auto exec = this->get_executor();
+
+        system_matrix_ = gko::clone(exec, system_matrix);
+
+        // Scale the system matrix if scaling coefficients are provided
+        if (parameters_.row_scaling) {
+            GKO_ASSERT_EQUAL_DIMENSIONS(parameters_.row_scaling,
+                                        system_matrix_);
+            row_scaling_ = parameters_.row_scaling;
+            row_scaling_->apply(system_matrix_.get(), system_matrix_.get());
```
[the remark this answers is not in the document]
upsj: #988 <---
for some LinOps, this is perfectly safe - in-place scaling is one of them.

include/ginkgo/core/base/reordered.hpp
```diff
@@ -0,0 +1,240 @@
...
+            cache_.inner_b =
+                matrix::Dense<value_type>::create(this->get_executor());
+            cache_.inner_x =
+                matrix::Dense<value_type>::create(this->get_executor());
+            cache_.intermediate =
+                matrix::Dense<value_type>::create(this->get_executor());
+        }
+        cache_.inner_b->copy_from(b);
+        cache_.inner_x->copy_from(x);
+        cache_.intermediate->copy_from(b);
```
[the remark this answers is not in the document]
upsj: `apply_uses_initial_guess()`?

include/ginkgo/core/base/reordered.hpp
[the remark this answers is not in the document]
yhmtsai: Yes, at least in my imagine about what compiler works. TBH, I wouldn't like to challenge the compiler did the same thing as imagine.

include/ginkgo/ginkgo.hpp
```diff
@@ -62,6 +62,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/base/precision_dispatch.hpp>
 #include <ginkgo/core/base/range.hpp>
 #include <ginkgo/core/base/range_accessors.hpp>
+#include <ginkgo/core/base/reordered.hpp>
```
pratikvn, line 65: I would rename this to `scaled_reordering` ? Just `reordered` is a bit vague.

include/ginkgo/core/reorder/reordering_base.hpp
```diff
@@ -60,6 +60,12 @@ namespace reorder {
  * specific reordering to decide what to do with the data that is passed to it.
  */
 class ReorderingBase : public EnableAbstractPolymorphicObject<ReorderingBase> {
+public:
+    virtual std::shared_ptr<const gko::LinOp> get_permutation() const = 0;
+
+    virtual std::shared_ptr<const gko::LinOp> get_inverse_permutation()
+        const = 0;
```
pratikvn, line 67: Is it necessary to have these getters here ?

pratikvn (approved):
Nice work. LGTM!

yhmtsai:
AFAIR, combing scaled and reordered together allows some algorithms uses only one operation.
how do you do that when a new LinOp?

include/ginkgo/core/reorder/reordering_base.hpp
fritzgoebel: As the scaled reordered factory takes a `ReorderingBaseFactory`, I think these are necessary here rather than only in the algorithm implemenations.

include/ginkgo/core/base/reordered.hpp
```diff
@@ -0,0 +1,240 @@
...
+#include <ginkgo/core/base/lin_op.hpp>
+#include <ginkgo/core/base/types.hpp>
+#include <ginkgo/core/matrix/dense.hpp>
+#include <ginkgo/core/matrix/diagonal.hpp>
+#include <ginkgo/core/matrix/identity.hpp>
+#include <ginkgo/core/matrix/permutation.hpp>
+#include <ginkgo/core/reorder/reordering_base.hpp>
+
+
+namespace gko {
```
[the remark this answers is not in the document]
fritzgoebel: Yes, before merging this we should agree on naming and placement.

fritzgoebel:
> AFAIR, combing scaled and reordered together allows some algorithms uses only one operation. how do you do that when a new LinOp?

At the moment, not at all. So far the operations are not merged yet.

yhmtsai:
will the merged operations be in this class or another class inherited from this class?
if there's another class, the clone will only give the clone of ScaledReordered class not itself.

include/ginkgo/core/reorder/reordering_base.hpp
```diff
@@ -60,6 +60,12 @@ namespace reorder {
  * specific reordering to decide what to do with the data that is passed to it.
  */
 class ReorderingBase : public EnableAbstractPolymorphicObject<ReorderingBase> {
+public:
+    virtual std::shared_ptr<const gko::LinOp> get_permutation() const = 0;
+
+    virtual std::shared_ptr<const gko::LinOp> get_inverse_permutation()
```
thoasm, line 66: nit:
```suggestion
    virtual std::shared_ptr<const LinOp> get_permutation() const = 0;

    virtual std::shared_ptr<const LinOp> get_inverse_permutation()
```
Also note that this breaks the interface of RCM (as you had `std::shared_ptr<matrix::Permutation>` as the return type).

include/ginkgo/core/reorder/rcm.hpp
```diff
@@ -109,7 +109,7 @@ class Rcm
      *
      * @return the permutation (permutation matrix)
      */
-    std::shared_ptr<const PermutationMatrix> get_permutation() const
+    std::shared_ptr<const LinOp> get_permutation() const override
```
thoasm, line 112: This change is an actual interface break.
Previously, you could have called
```
rcm->get_permutation()->get_permutation_size();
```
However, now, you need to do it with a cast since you no longer know the actual type of the permutation:
```
as<matrix::Permutation>(rcm->get_permutation())->get_permutation_size();
```
Currently, this affects only RCM (as there is no other reordering).
Is it necessary to make the permutation type a LinOp (meaning are you expecting other permutation types)?

reference/test/base/reordered.cpp
```diff
@@ -0,0 +1,487 @@
...
+}
+
+
+TYPED_TEST(ScaledReordered, BuildsWithRowScaling)
+{
+    using SR = typename TestFixture::SR;
+    using Mtx = typename TestFixture::Mtx;
+    auto scaled_reordered_fact =
+        SR::build().with_row_scaling(this->diag).on(this->exec);
+    auto scaled_reordered = scaled_reordered_fact->generate(this->mtx);
```
thoasm, line 188: nit  (for AAA separation):
```suggestion
        SR::build().with_row_scaling(this->diag).on(this->exec);
        
    auto scaled_reordered = scaled_reordered_fact->generate(this->mtx);
```

reference/test/base/reordered.cpp
```diff
@@ -0,0 +1,487 @@
...
+}
+
+
+TYPED_TEST(ScaledReordered, BuildsWithColScaling)
+{
+    using SR = typename TestFixture::SR;
+    using Mtx = typename TestFixture::Mtx;
+    auto scaled_reordered_fact =
+        SR::build().with_col_scaling(this->diag).on(this->exec);
+    auto scaled_reordered = scaled_reordered_fact->generate(this->mtx);
```
thoasm, line 201: same here:
```suggestion
        SR::build().with_col_scaling(this->diag).on(this->exec);
        
    auto scaled_reordered = scaled_reordered_fact->generate(this->mtx);
```

reference/test/base/reordered.cpp
```diff
@@ -0,0 +1,487 @@
...
+
+TYPED_TEST(ScaledReordered, BuildsWithRowAndColScaling)
+{
+    using SR = typename TestFixture::SR;
+    using Mtx = typename TestFixture::Mtx;
+    auto scaled_reordered_fact = SR::build()
+                                     .with_row_scaling(this->diag)
+                                     .with_col_scaling(this->diag)
+                                     .on(this->exec);
+    auto scaled_reordered = scaled_reordered_fact->generate(this->mtx);
```
thoasm, line 216:
```suggestion
                                     .on(this->exec);
                                     
    auto scaled_reordered = scaled_reordered_fact->generate(this->mtx);
```

reference/test/base/reordered.cpp
```diff
@@ -0,0 +1,487 @@
...
+}
+
+
+TYPED_TEST(ScaledReordered, BuildsWithRcmReordering)
+{
+    using SR = typename TestFixture::SR;
+    using Mtx = typename TestFixture::Mtx;
+    auto scaled_reordered_fact =
+        SR::build().with_reordering(this->rcm_factory).on(this->exec);
+    auto scaled_reordered = scaled_reordered_fact->generate(this->rcm_mtx);
```
thoasm, line 229:
```suggestion
        SR::build().with_reordering(this->rcm_factory).on(this->exec);
        
    auto scaled_reordered = scaled_reordered_fact->generate(this->rcm_mtx);
```

reference/test/base/reordered.cpp
```diff
@@ -0,0 +1,487 @@
...
+    auto scaled_reordered = scaled_reordered_fact->generate(this->mtx);
+
+    auto before_system_matrix = scaled_reordered->get_system_matrix();
+    auto before_inner_operator = scaled_reordered->get_inner_operator();
+
+    auto copied = SR::build().on(this->exec)->generate(this->rcm_mtx);
+    copied->copy_from(scaled_reordered.get());
+
+    ASSERT_EQ(before_system_matrix, copied->get_system_matrix());
+    ASSERT_EQ(before_inner_operator, copied->get_inner_operator());
```
thoasm, line 264: nit: for cleaner AAA separation:
```suggestion
    using SR = typename TestFixture::SR;
    auto scaled_reordered_fact = SR::build()
                                     .with_row_scaling(this->diag)
                                     .with_col_scaling(this->diag)
                                     .with_reordering(this->rcm_factory)
                                     .with_inner_operator(this->solver_factory)
                                     .on(this->exec);
    auto scaled_reordered = scaled_reordered_fact->generate(this->mtx);
    auto before_system_matrix = scaled_reordered->get_system_matrix();
    auto before_inner_operator = scaled_reordered->get_inner_operator();

    auto copied = SR::build().on(this->exec)->generate(this->rcm_mtx);
    copied->copy_from(scaled_reordered.get());

    ASSERT_EQ(before_system_matrix, copied->get_system_matrix());
    ASSERT_EQ(before_inner_operator, copied->get_inner_operator());
```

reference/test/base/reordered.cpp
```diff
@@ -0,0 +1,487 @@
...
+    using SR = typename TestFixture::SR;
+    auto scaled_reordered_fact = SR::build()
+                                     .with_row_scaling(this->diag)
+                                     .with_col_scaling(this->diag)
+                                     .with_reordering(this->rcm_factory)
+                                     .with_inner_operator(this->solver_factory)
+                                     .on(this->exec);
+    auto scaled_reordered = scaled_reordered_fact->generate(this->mtx);
+
+    auto before_system_matrix = scaled_reordered->get_system_matrix();
```
thoasm, line 279: same here
```suggestion
    auto scaled_reordered = scaled_reordered_fact->generate(this->mtx);
    auto before_system_matrix = scaled_reordered->get_system_matrix();
```

reference/test/base/reordered.cpp
```diff
@@ -0,0 +1,487 @@
...
```
thoasm, line 301: and here:
```suggestion
    auto scaled_reordered = scaled_reordered_fact->generate(this->mtx);
    auto before_system_matrix = scaled_reordered->get_system_matrix();
```

include/ginkgo/core/base/reordered.hpp
```diff
@@ -0,0 +1,244 @@
...
+        if (parameters_.reordering) {
+            auto reordering = parameters_.reordering->generate(system_matrix_);
+            auto perm = reordering->get_permutation();
+            permutation_ =
+                clone(exec, as<matrix::Permutation<index_type>>(perm));
+            auto permutation_array =
+                make_array_view(exec, permutation_->get_size()[0],
+                                permutation_->get_permutation());
+            system_matrix_ = as<Permutable<index_type>>(system_matrix_)
+                                 ->permute(&permutation_array);
```
thoasm, line 165: This requires your reordering to return a `matrix::Permutation`. However, this is not documented and not intuitive (your `ReorderingBaseFactory` needs `get_permutation()` to return a `matrix::Permutation`).
Is there a reason why you don't call the `apply` method on the permutation matrix? That would allow for more formats than just `matrix::Permutation` and would work a lot better with your new interface.

include/ginkgo/ginkgo.hpp
thoasm: I second that.

include/ginkgo/core/base/reordered.hpp
```diff
@@ -0,0 +1,244 @@
...
+            cache_.inner_x->copy_from(x);
+        }
+    }
+
+private:
+    std::shared_ptr<LinOp> system_matrix_{};
+    std::shared_ptr<const LinOp> inner_operator_{};
+    std::shared_ptr<const matrix::Diagonal<value_type>> row_scaling_{};
+    std::shared_ptr<const matrix::Diagonal<value_type>> col_scaling_{};
+    std::shared_ptr<matrix::Permutation<index_type>> permutation_{};
```
thoasm, line 216: Same here: if you want it to work for all reordering options, it should be a plain `LinOp` or a new data format (it should match the `get_permutation()` return type of your `ReorderingBase`).

thoasm (changes requested):
There is an interface break at the moment (the only reason why I don't approve this PR).
Unfortunately, it will be quite difficult to find a solution that does not break the existing interface while being intuitive to use (for example, adding `get_permutation_2` would not break the interface, but be quite confusing for the user).
Additionally, the actual reordering inside the constructor of `ScaledReordered` needs a bit more work.

fritzgoebel:
format!

include/ginkgo/core/base/reordered.hpp
fritzgoebel: I went with an array now, that works for all kinds of permutations and does not assume we have a permutation matrix.

include/ginkgo/core/reorder/scaled_reordered.hpp
```diff
@@ -0,0 +1,257 @@
...
+         * and reordered system matrix.
+         */
+        std::shared_ptr<const LinOpFactory> GKO_FACTORY_PARAMETER_SCALAR(
+            inner_operator, nullptr);
+
+        /**
+         * The reordering that is to be applied to the system matrix.
+         */
+        std::shared_ptr<const ReorderingBaseFactory>
+            GKO_FACTORY_PARAMETER_SCALAR(reordering, nullptr);
```
thoasm, line 107: Requires a note that the system matrix must be of type `Permutable<index_type>` if a reordering is specified.

include/ginkgo/core/reorder/scaled_reordered.hpp
```diff
@@ -0,0 +1,257 @@
...
+            col_scaling_->rapply(system_matrix_.get(), system_matrix_.get());
+        }
+
+        // If a reordering factory is provided, generate the reordering and
+        // permute the system matrix accordingly.
+        if (parameters_.reordering) {
+            auto reordering = parameters_.reordering->generate(system_matrix_);
+            permutation_array_ = reordering->get_permutation_array();
+            system_matrix_ = as<Permutable<index_type>>(system_matrix_)
+                                 ->permute(permutation_array_.get());
```
thoasm, line 165: Without a note, this `as<Permutable<index_type>>` might surprise some users.

thoasm (approved):
Thanks for the improvements. I have only minor requests, after addressing these, you can feel free and merge it.

upsj (approved):
LGTM for what we can currently change - I think there is still a lot of work to be done in the Permutation, Diagonal and ScaledPermutation area, but that needs to happen in Ginkgo 2.0

include/ginkgo/core/reorder/scaled_reordered.hpp
```diff
@@ -0,0 +1,257 @@
...
+        cache_struct& operator=(const cache_struct&) { return *this; }
+
+        cache_struct& operator=(cache_struct&&) { return *this; }
+
+        std::unique_ptr<matrix::Dense<value_type>> inner_b{};
+
+        std::unique_ptr<matrix::Dense<value_type>> inner_x{};
+
+        std::unique_ptr<matrix::Dense<value_type>> intermediate{};
+
```
yhmtsai, line 247:
```suggestion
        std::unique_ptr<matrix::Dense<value_type>> inner_b{};
        std::unique_ptr<matrix::Dense<value_type>> inner_x{};
        std::unique_ptr<matrix::Dense<value_type>> intermediate{};
```
nit: these are data

include/ginkgo/core/reorder/scaled_reordered.hpp
```diff
@@ -0,0 +1,257 @@
...
+                                 ->permute(permutation_array_.get());
+        }
+
+        // Generate the inner operator with the scaled and reordered system
+        // matrix. If none is provided, use the Identity.
+        if (parameters_.inner_operator) {
+            inner_operator_ =
+                parameters_.inner_operator->generate(system_matrix_);
+        } else {
+            inner_operator_ = gko::matrix::Identity<value_type>::create(exec);
```
yhmtsai, line 174: need to put the size, apply will check the size with b, x

reference/test/reorder/scaled_reordered.cpp
```diff
@@ -0,0 +1,488 @@
...
+    auto before_inner_operator = scaled_reordered->get_inner_operator();
+
+    auto cloned = scaled_reordered->clone();
+
+    ASSERT_EQ(before_system_matrix, cloned->get_system_matrix());
+    ASSERT_EQ(before_inner_operator, cloned->get_inner_operator());
+}
+
+
+TYPED_TEST(ScaledReordered, SolvesSingleRhsWithOnlyInnerOperator)
```
yhmtsai, line 312: need a test for no inner_factory

yhmtsai (changes requested):
LGTM in general. I have discussed with Fritz about the combining the permutation and scaling. It will change the implementation not a new class, so my above worry will not happen. 

I put it is request changes because it may give error when inner_operator is nullptr (generated Identity)

include/ginkgo/core/reorder/reordering_base.hpp
```diff
@@ -59,7 +59,14 @@ namespace reorder {
...
  * specific reordering to decide what to do with the data that is passed to it.
  */
-class ReorderingBase : public EnableAbstractPolymorphicObject<ReorderingBase> {
+template <typename IndexType = int32>
+class ReorderingBase
+    : public EnableAbstractPolymorphicObject<ReorderingBase<IndexType>> {
+public:
+    using index_type = IndexType;
+
+    virtual const array<index_type>& get_permutation_array() const = 0;
```
yhmtsai, line 68: you can also provide the setter for permutation_array in protected and the permutation data in private.
without implement it in each class

include/ginkgo/core/reorder/reordering_base.hpp
```diff
@@ -81,36 +88,6 @@ struct ReorderingBaseArgs {
 };
 
 
-/**
- * Declares an Abstract Factory specialized for ReorderingBases
- */
-using ReorderingBaseFactory =
```
yhmtsai, line 87: public interface?

yhmtsai (approved):
LGTM in general. two concern: 1. public interface from using alias 2. move the get_permutation_array into ReorderBase.

include/ginkgo/core/reorder/reordering_base.hpp
fritzgoebel: I added a template argument to these two. But I think technically, the second one is still an interface break since in case someone used the `PolymorphicBase` template argument, this would no longer work...

thoasm: I'm fine if you break the interface at that point as I don't think a lot of people (my assumption is 0 outside of us) were using it. If it is compiled with `c++17`, the interface is no longer broken, so I would consider it as acceptable.

fritzgoebel:
Depends on  #1171  now.

fritzgoebel:
Ready to merge after #1171

PR #1060, opened 2022-06-22
relax time limit for test/matrix/matrix_exec

yhmtsai:
This PR relaxes the time limit for test/matrix/matrix_exec because CI sometimes picks slower machine such that matrix_exec timeout.

Related Jobs: https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/2617142630


yhmtsai:
It does not need to be CI level. For cuda test, we should only request the CUDA flag for CI machine not specify which machine we should use.

yhmtsai:
I understand what you meaning. you mean putting the timeout into the ci-job.
I thought putting it into CI machine enironment or specify one kind of machine

PR #1067, opened 2022-07-01
Use needs not dependencies to avoid waiting entire prevstage

yhmtsai:
This PR changes the slurm test job and gh-pages jobs to use needs keyword
The slurm test job only waits for the corresponding build job not build stage
gh-pages jobs only waits for the sync job

Others still follows the stage step such that we do not introduce all jobs of pipeline to the queue.
It makes the build/test jobs are in the queue first.

image


upsj (approved):

pratikvn (approved):
LGTM!

yhmtsai:
rebase!

yhmtsai:
I merge this pr without waiting for the gitlab ci because the dependency is already tested before rebase and gh-page will be tested after merging.

PR #1068, opened 2022-07-01
Move generated documentation to separate repo

upsj:
Our repository is rather big, which is mostly (500 MB) due to the generated documentation on the webpage.
Users shouldn't need to clone this, so I would suggest we move everything to a separate repository.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -917,7 +917,7 @@ gh-pages:
     - cd gh-pages-repo
     - git add -A
     - git diff --quiet HEAD ||
-      (git commit -m "Update documentation from ${CURRENT_SHA}" && git push)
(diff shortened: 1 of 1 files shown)


tcojean (approved):
LGTM! Good idea

 I think we will also need to update this https://github.com/ginkgo-project/ginkgo-project.github.io

yhmtsai:
Could you also change the commit message 
"Update documentation from ${CURRENT_SHA}"
->
"Update documentation from ginkgo-project/ginkgo@${CI_COMMIT_SHORT_SHA}"
and do not need to get the sha from git
the new repo needs to have gh-pages branch. I am thinking whether moving the past branch to there makes sense.
note: If this pr is merged, change the wiki documentation link

yhmtsai (approved):

PR #1069, opened 2022-07-04
Fix minor example issues

gfcas:
Clearify the functionality of a dealii function in the external-lib-example.

### DIFF
--- a/examples/external-lib-interfacing/external-lib-interfacing.cpp
+++ b/examples/external-lib-interfacing/external-lib-interfacing.cpp
@@ -906,7 +906,7 @@ void AdvectionProblem<dim>::solve()
                      preconditioner);
     *******************************************************/
 
-    // Give the solution back to deall.ii
+    // Distribute the possible hanging nodes constraints
     hanging_node_constraints.distribute(solution);
 }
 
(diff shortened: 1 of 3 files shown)


gfcas:
Can you please explain me, what is the app_exec in poisson-solver. I do not understand it, for commenting.

upsj:
The "application" code is CPU only, so it needs to use the host executor accessible via `exec->get_master()` instead of the device executor.

gfcas:
> The "application" code is CPU only, so it needs to use the host executor accessible via `exec->get_master()` instead of the device executor.

Thanks, this means for the external library code I only need to run the solver on exec (GPU) not the other parts:
```cpp
auto             b       = vec::create(exec,
                         gko::dim<2>(num_rows, 1),
                         val_array::view(exec, num_rows, rhs.data()),
                         1);
    auto             x       = vec::create(exec, gko::dim<2>(num_rows, 1));
    auto             A       = mtx::create(exec,
                         gko::dim<2>(num_rows),
                         system_matrix.n_nonzero_elements());
```
Here I would everywhere replace it with app_exec.

This is taken from external-library example, as I want to integrate it in dealii.

MarcelKoch (approved):

upsj (approved):

upsj:
label!

upsj:
check-format!

tcojean (approved):

PR #1070, opened 2022-07-04
Add comment to simple-solver

gfcas:
Add a comment to simple-solver example.


MarcelKoch:
Thanks for your PR @gfcas, would you mind merging this one with the other one #1069? I think they fit well together.

gfcas:
Yes, I can do this.

PR #1071, opened 2022-07-04
Add fix for GCC 12

MarcelKoch:
This adds an include that is missing in gcc-12. It seems that we relied on which headers the standard header include internally for `std::memcpy`.


tcojean (approved):

upsj (approved):

upsj:
Thanks for this, I actually also just compiled ginkgo with gcc 12 to test my new laptop, and ran into the same issue :)

upsj:
I'll go ahead and merge this, since it's a tiny fix, and I could use the fix myself.

PR #1074, opened 2022-07-12
Bwuni fixes

tcojean:
Fix two issues with tests on bwuni

The issues seem to be with the hwloc-related tests, esp. with the case where we only have access to a part of the machine.

### DIFF
--- a/cuda/test/base/cuda_executor_topology.cu
+++ b/cuda/test/base/cuda_executor_topology.cu
@@ -139,12 +139,12 @@ TEST_F(CudaExecutor, CanBindToCores)
 TEST_F(CudaExecutor, ClosestCpusIsPopulated)
 {
     cuda = gko::CudaExecutor::create(0, gko::OmpExecutor::create());
-    auto close_cpus0 = cuda->get_closest_pus();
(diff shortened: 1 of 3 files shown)


upsj (approved):
LGTM! Thanks for fixing this, it also caused failures on my WSL instance.

pratikvn (approved):
LGTM!

cuda/test/base/cuda_executor_topology.cu
```diff
@@ -149,9 +152,12 @@ TEST_F(CudaExecutor, KnowsItsNuma)
 {
     cuda = gko::CudaExecutor::create(0, gko::OmpExecutor::create());
     auto numa0 = cuda->get_closest_numa();
-    auto close_cpu0 = cuda->get_closest_pus()[0];
+    auto close_cpus0 = cuda->get_closest_pus();
```
thoasm, line 155: nit: naming does no longer fit IMO
```suggestion
    auto close_cpus = cuda->get_closest_pus();
```

thoasm (approved):
LGTM!
This also fixes the previously failing test `cuda/test/base/cuda_executor_topology` on my machine.

Issue #1051, opened 2022-06-06
Input tolerance vs output error


bp-intact-solutions:
Hello,
I am running ginkgo/1.4.0 using Conan-center, on Mac.
I would like to have a full control on the tolerance of the solver.

I have setting up a logger doing the following:

```
auto iter_stop = gko::share(
        gko::stop::Iteration::build().with_max_iters(1000).on(exec));
    auto tol_stop = gko::share(gko::stop::ResidualNorm<double>::build()
                                   .with_reduction_factor(1e-20)
                                   .on(exec));

```

Then, I am running a solver that way:

```
using bj = gko::preconditioner::Jacobi<ValueType, IndexType>;
    auto solver_gen =
        cg::build()
            .with_criteria(iter_stop, tol_stop)
            // Add preconditioner, these 2 lines are the only
            // difference from the simple solver example
            .with_preconditioner(bj::build()
                                     .with_max_block_size(16u)
                                     .with_storage_optimization(
                                         gko::precision_reduction::autodetect())
                                     .on(exec))
            .on(exec);
```
At the end, I would like to make sure the output error is lower than the input tolerance.
If I am not mistaken, it seems there are two ways to do it:

1.
```
    auto impl_res = gko::as<real_vec>(logger->get_implicit_sq_resnorm());
    std::cout << "Implicit residual norm (r):\n";
    std::cout << std::sqrt(impl_res->at(1, 0)) << std::endl;
```

2. Something similar to 
```
    auto one = gko::initialize<vec>({1.0}, exec);
    auto neg_one = gko::initialize<vec>({-1.0}, exec);
    auto res = gko::initialize<real_vec>({0.0}, exec);
    A->apply(lend(one), lend(x), lend(neg_one), lend(b));
    b->compute_norm2(lend(res));
```
The second one has a runtime error with 1.4.0 on Mac.
With the first one, the value is much much lower than 1e-20.

What would you recommend?
Thank you.
Best regards. 

Note: I can attach the full code if needed

pratikvn:
If you set the residual tolerance to 1e-20, then the solver will try to get an error of atleast 1e-20, but it might be lower.

For CG, methods 1. and 2. should be equivalent, because the internal residual norm of CG should track the backward error calculated through $|| b - A*x ||_2$. 

For your runtime error for 2., I think if exec is a GPU executor and then try to print that out, then you have memory access issues.

upsj:
[...]

PR #1079, opened 2022-07-14
Fix the edge case in residual stopping criterion

keldu:
Regarding #1051 this would fix the edge case with an rhs vector filled with zeroes.

Used '<=' instead of '<' in residual operations. Test cases will be added

### DIFF
--- a/include/ginkgo/core/stop/residual_norm.hpp
+++ b/include/ginkgo/core/stop/residual_norm.hpp
@@ -52,13 +52,13 @@ namespace stop {
  * The mode for the residual norm criterion.
  *
  * - absolute:        Check for tolerance against residual norm.
- *                    $ || r || < \tau $
+ *                    $ || r || \leq \tau $
  *
  * - initial_resnorm: Check for tolerance relative to the initial residual norm.
- *                    $ \frac{|| r ||}{|| r_0||} < \tau $
+ *                    $ || r || \leq \tau \times || r_0|| $
  *
  * - rhs_norm:        Check for tolerance relative to the rhs norm.
- *                    $ \frac{|| r ||}{|| b ||} < \tau $
+ *                    $ || r || \leq \tau \times || b || $
  *
  * @ingroup stop
  */
@@ -118,10 +118,11 @@ private:
  * The ResidualNorm class is a stopping criterion which
  * stops the iteration process when the actual residual norm is below a
  * certain threshold relative to
- * 1. the norm of the right-hand side, norm(residual) / norm(right_hand_side)
- *                                                                  < threshold
- * 2. the initial residual, norm(residual) / norm(initial_residual) < threshold.
- * 3. one,  norm(residual) < threshold.
+ * 1. the norm of the right-hand side, norm(residual) $\leq$ < threshold *
+ *    norm(right_hand_side).
+ * 2. the initial residual, norm(residual) $\leq$ threshold *
+ *    norm(initial_residual).
+ * 3. one,  norm(residual) $\leq$ threshold.
  *
  * For better performance, the checks are run on the executor
  * where the algorithm is executed.
@@ -176,11 +177,11 @@ protected:
  * The ImplicitResidualNorm class is a stopping criterion which
  * stops the iteration process when the implicit residual norm is below a
  * certain threshold relative to
- * 1. the norm of the right-hand side, implicit_resnorm / norm(right_hand_side)
- *                                                          < threshold
- * 2. the initial residual, implicit_resnorm / norm(initial_residual) <
- *                                                          < threshold.
- * 3. one, implicit_resnorm < threshold.
+ * 1. the norm of the right-hand side, implicit_resnorm $\leq$ < threshold *
+ * norm(right_hand_side)
+ * 2. the initial residual, implicit_resnorm $\leq$ threshold *
+ * norm(initial_residual) .
+ * 3. one,  implicit_resnorm $\leq$ threshold.
  *
  * @note To use this stopping criterion there are some dependencies. The
  * constructor depends on either `b` or the `initial_residual` in order to
--- a/reference/test/stop/residual_norm_kernels.cpp
+++ b/reference/test/stop/residual_norm_kernels.cpp
@@ -108,50 +108,22 @@ TYPED_TEST(ResidualNorm, CheckIfResZeroConverges)
     using Mtx = typename TestFixture::Mtx;
     using NormVector = typename TestFixture::NormVector;
     using T = typename TestFixture::ValueType;
+    using mode = typename gko::stop::mode;
     std::shared_ptr<gko::LinOp> mtx = gko::initialize<Mtx>({1.0}, this->exec_);
     std::shared_ptr<gko::LinOp> rhs = gko::initialize<Mtx>({0.0}, this->exec_);
     std::shared_ptr<gko::LinOp> x = gko::initialize<Mtx>({0.0}, this->exec_);
     std::shared_ptr<gko::LinOp> res_norm =
         gko::initialize<NormVector>({0.0}, this->exec_);
 
-    {
+    for (auto baseline :
+         {mode::rhs_norm, mode::initial_resnorm, mode::absolute}) {
+        gko::remove_complex<T> factor =
+            (baseline == mode::absolute) ? 0.0 : r<T>::value;
         auto criterion = gko::stop::ResidualNorm<T>::build()
-                             .with_reduction_factor(r<T>::value)
+                             .with_reduction_factor(factor)
+                             .with_baseline(baseline)
                              .on(this->exec_)
                              ->generate(mtx, rhs, x.get(), nullptr);
-        constexpr gko::uint8 RelativeStoppingId{1};
-        bool one_changed{};
-        gko::array<gko::stopping_status> stop_status(this->exec_, 1);
-        stop_status.get_data()[0].reset();
-
-        EXPECT_TRUE(criterion->update().residual_norm(res_norm).check(
-            RelativeStoppingId, true, &stop_status, &one_changed));
-        EXPECT_TRUE(stop_status.get_data()[0].has_converged());
-        EXPECT_TRUE(one_changed);
-    }
-    {
-        auto criterion = gko::stop::ResidualNorm<T>::build()
-                             .with_reduction_factor(r<T>::value)
-                             .with_baseline(gko::stop::mode::initial_resnorm)
-                             .on(this->exec_)
-                             ->generate(mtx, rhs, x.get(), nullptr);
-        constexpr gko::uint8 RelativeStoppingId{1};
-        bool one_changed{};
-        gko::array<gko::stopping_status> stop_status(this->exec_, 1);
-        stop_status.get_data()[0].reset();
-
-        EXPECT_TRUE(criterion->update().residual_norm(res_norm).check(
-            RelativeStoppingId, true, &stop_status, &one_changed));
-        EXPECT_TRUE(stop_status.get_data()[0].has_converged());
-        EXPECT_TRUE(one_changed);
-    }
-    {
-        auto criterion =
-            gko::stop::ResidualNorm<T>::build()
-                .with_reduction_factor(gko::zero<gko::remove_complex<T>>())
-                .with_baseline(gko::stop::mode::absolute)
-                .on(this->exec_)
-                ->generate(mtx, rhs, x.get(), nullptr);
         constexpr gko::uint8 RelativeStoppingId{1};
         bool one_changed{};
         gko::array<gko::stopping_status> stop_status(this->exec_, 1);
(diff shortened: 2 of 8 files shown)


yhmtsai:
@keldu what's the status of this pull request? It seems to be good for review.

greole:
Quick question, this currently stands +4000 LOC changes and 174 files changed, which seems to be to much for fixing an edge case. Could it be that it is not based on `dev`?

keldu:
So. I'm not fully sure about the tests. I did a slightly small one, but I am wholly unsure if it even works.
What's missing anyway for now where I am stumped. Since I keep coming back without gaining any ground.

- [ ] Check all comments for the res_norm < \tau
- [ ] Check test and add proper test.

keldu:
And I misclicked. It is not ready for review :facepalm:

keldu:
> Quick question, this currently stands +4000 LOC changes and 174 files changed, which seems to be to much for fixing an edge case. Could it be that it is not based on `dev`?

Yeah. I rebase'd on an incorrect develop  branch. Caused a temporary issue which I didn't realised. I guess rebasing on origin/develop is the fix for that.

test/stop/residual_norm_kernels.cpp
```diff
@@ -110,6 +110,28 @@ TYPED_TEST(ResidualNorm, CanIgorneResidualNorm)
...
+        constexpr gko::uint8 RelativeStoppingId{1};
+        bool one_changed{};
+        gko::array<gko::stopping_status> stop_status(this->ref, 1);
+        stop_status.get_data()[0].reset();
+        stop_status.set_executor(this->exec);
+
+        ASSERT_TRUE(criterion->update().residual_norm(res_norm).check(
+            RelativeStoppingId, true, &stop_status, &one_changed));
+        ASSERT_TRUE(stop_status.get_data()[0].has_converged());
+        ASSERT_TRUE(one_changed);
```
yhmtsai, line 132:
```suggestion
            RelativeStoppingId, true, &stop_status, &one_changed));
        stop_status.set_executor(this->ref);
        ASSERT_TRUE(stop_status.get_data()[0].has_converged());
        ASSERT_TRUE(one_changed);
```
It should solve the segmentation fault

keldu: I'd generally prefer if I could just setup the residual_norm to 0 and test that against the stopping status.
the whole step with the criterion seems to be unnecessary. Though I have no clue how to fix that if that is at all possible.

keldu: Also should I check with rel_factory and abs_factory?

yhmtsai (dismissed):
Miss the implicit residual test.
I think one of abs/rel should be enough? because they all go through the same kernel and the baseline calculation should be also tested in the existing tests.

MarcelKoch (approved):

keldu:
I'm pretty sure my test is wrong btw. The problem mostly lies with the fact that I don't know what some of the functions do.
And the examples don't cover initalizing already solved systems by hand.

keldu:
> Miss the implicit residual test. I think one of abs/rel should be enough? because they all go through the same kernel and the baseline calculation should be also tested in the existing tests.

I've added a rel criterion. Still not sure about my previous comment.

include/ginkgo/core/stop/residual_norm.hpp
```diff
@@ -52,13 +52,13 @@ namespace stop {
  * The mode for the residual norm criterion.
  *
  * - absolute:        Check for tolerance against residual norm.
- *                    $ || r || < \tau $
+ *                    $ || r || <= \tau $
```
upsj, line 55: nit: (same below)
```suggestion
 *                    $ || r || \leq \tau $
```

upsj (changes requested):
We are not testing the reference kernels yet. Otherwise LGTM!

upsj:
@keldu A simple way to check the test does what it's supposed to to is to revert the other changes and only leave the new test in, it should fail.

upsj (approved):
Sorry, my mistake the reference test is included in the test for every backend.

test/stop/residual_norm_kernels.cpp
```diff
@@ -110,6 +110,45 @@ TYPED_TEST(ResidualNorm, CanIgorneResidualNorm)
...
+TYPED_TEST(ResidualNorm, CheckIfResZeroConverges)
+{
+    using Mtx = typename TestFixture::Mtx;
+    using NormVector = typename TestFixture::NormVector;
+    auto initial_res = gko::initialize<Mtx>({10.0}, this->exec);
+    std::shared_ptr<gko::LinOp> rhs = gko::initialize<Mtx>({10.0}, this->exec);
+    auto criterion =
+        this->factory->generate(nullptr, rhs, nullptr, initial_res.get());
+    auto rel_criterion =
+        this->rel_factory->generate(nullptr, rhs, nullptr, initial_res.get());
```
yhmtsai, line 122: you need to set the reduction_factor to zero such that it indeed uses `=`.
For all cases with the original issue corcern,
1. absolute: reduction_factor = 0,
2. initial_resnorm: initial_res = 0 
3. rhs_norm: rhs = 0

create the new absolute, initial_resnorm, rhs_norm here not from the object in TestFixture.
You may need to add `using ValueType = T;` in class

yhmtsai (dismissed):
The test does not reach `=` condition.
If you would also like to add the check more close to what we use in solver.
```
// it checks the rhs is zero and initial guess is the solution
std::shared_ptr<gko::LinOp> rhs = gko::initialize<Mtx>({0.0}, this->exec);
std::shared_ptr<gko::LinOp> mtx = gko::initialize<Mtx>({1.0}, this->exec);
std::shared_ptr<gko::LinOp> x = gko::initialize<Mtx>({0.0}, this->exec);
std::shared_ptr<gko::LinOp> res_norm = gko::initialize<NormVector>({0.0}, this->exec);
auto criterion =
         gko::stop::ResidualNorm<T>::build()
                      .with_reduction_factor(r<T>::value)
                      .on(this->exec)->generate(mtx, rhs, x.get(), nullptr);
// like the other test setting

ASSERT_TRUE(criterion->update().residual_norm(res_norm)check(
    RelativeStoppingId, true, &stop_status, &one_changed));
```
I think @upsj correct. you miss the reference test although they are almost identical between unified and reference in this case.
all tests also need the similar version for implicit Residual

include/ginkgo/core/stop/residual_norm.hpp
```diff
@@ -52,13 +52,13 @@ namespace stop {
  * The mode for the residual norm criterion.
  *
  * - absolute:        Check for tolerance against residual norm.
- *                    $ || r || < \tau $
+ *                    $ || r || \leq \tau $
  *
  * - initial_resnorm: Check for tolerance relative to the initial residual norm.
- *                    $ \frac{|| r ||}{|| r_0||} < \tau $
+ *                    $ \frac{|| r ||}{|| r_0||} <= \tau $
```
yhmtsai, line 58:
```suggestion
 *                    $ || r || \leq \tau \times || r_0||$
```

include/ginkgo/core/stop/residual_norm.hpp
```diff
@@ -52,13 +52,13 @@ namespace stop {
...
  *
  * - rhs_norm:        Check for tolerance relative to the rhs norm.
- *                    $ \frac{|| r ||}{|| b ||} < \tau $
+ *                    $ \frac{|| r ||}{|| b ||} <= \tau $
```
yhmtsai, line 61:
```suggestion
 *                    $ || r || \leq \tau \times || b ||$
```

test/stop/residual_norm_kernels.cpp
```diff
@@ -110,6 +110,45 @@ TYPED_TEST(ResidualNorm, CanIgorneResidualNorm)
                  gko::NotSupported);
 }
 
+TYPED_TEST(ResidualNorm, CheckIfResZeroConverges)
```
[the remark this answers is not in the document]
keldu: Ah. I missed the lower TestFixtures. Thanks for pointing it out

yhmtsai (commented):
LGTM. Some compiling issues and the tests need fix.
Could you also rearrange the commit history by combining commits with the purpose? Also formatting in the end.
I do not change my reviews status yet until the tests are correct.

reference/test/stop/residual_norm_kernels.cpp
```diff
@@ -820,6 +846,32 @@ TYPED_TEST(ImplicitResidualNorm, CanCreateFactory)
...
+    auto criterion =
+        gko::stop::ResidualNorm<T>::build().with_reduction_factor(r<T>::value);
+    .on(this->exec)->generate(mtx, rhs, x.get(), nullptr);
+    constexpr gko::uint8 RelativeStoppingId{1};
+    bool one_changed{};
+    gko::array<gko::stopping_status> stop_status(this->ref, 1);
+    stop_status.get_data()[0].reset();
+    stop_status.set_executor(this->exec);
+
+    ASSERT_TRUE(criterion->update().residual_norm(res_norm).check(
```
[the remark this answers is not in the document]
keldu: Yeah. Your're right. Will change. I just did a quick commit since I had some time.

keldu:
!format

keldu:
format!

reference/test/stop/residual_norm_kernels.cpp
```diff
@@ -102,6 +102,32 @@ TYPED_TEST(ResidualNorm, CanCreateFactory)
     ASSERT_EQ(this->abs_factory_->get_executor(), this->exec_);
 }
 
+TYPED_TEST(ResidualNorm, CheckIfResZeroConverges)
+{
+    using Mtx = typename TestFixture::Mtx;
+    using NormVector = typename TestFixture::NormVector;
+    std::shared_ptr<gko::LinOp> mtx = gko::initialize<Mtx>({1.0}, this->exec);
```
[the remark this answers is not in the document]
keldu: Why though? I mean I see it now. But that is a weird thing to have two different sets of naming schemes in the tests.

include/ginkgo/core/stop/residual_norm.hpp
```diff
@@ -52,13 +52,13 @@ namespace stop {
  * The mode for the residual norm criterion.
  *
  * - absolute:        Check for tolerance against residual norm.
- *                    $ || r || < \tau $
+ *                    $ || r || \leq \tau $
```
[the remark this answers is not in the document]
keldu: Just did it

reference/test/stop/residual_norm_kernels.cpp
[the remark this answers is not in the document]
keldu: I changed it. Still baffled why it's named differently in the reference.

keldu:
> LGTM. Some compiling issues and the tests need fix. Could you also rearrange the commit history by combining commits with the purpose? Also formatting in the end. I do not change my reviews status yet until the tests are correct.

I mean I could generally squash them? It's not a big change, so I would prefer to do it that way.

keldu:
format!

yhmtsai:
format!

yhmtsai:
the first commit adds the tests which are failed in pipeline https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/pipelines/1059541223 because it requires <= not <.
adding <= and updating the doc are in the following commit.

reference/test/CMakeLists.txt
```diff
@@ -12,3 +12,6 @@ add_subdirectory(reorder)
 add_subdirectory(solver)
 add_subdirectory(stop)
 add_subdirectory(utils)
+
+
+
```
tcojean, line 17: unneeded
```suggestion
```

reference/test/stop/residual_norm_kernels.cpp
```diff
@@ -102,6 +103,67 @@ TYPED_TEST(ResidualNorm, CanCreateFactory)
...
+
+    {
+        auto criterion = gko::stop::ResidualNorm<T>::build()
+                             .with_reduction_factor(r<T>::value)
+                             .on(this->exec_)
+                             ->generate(mtx, rhs, x.get(), nullptr);
+        constexpr gko::uint8 RelativeStoppingId{1};
+        bool one_changed{};
+        gko::array<gko::stopping_status> stop_status(this->exec_, 1);
+        stop_status.get_data()[0].reset();
```
tcojean, line 125: not important: since this is always the same except for the baseline, maybe there is some worth in making this a function? same for the checks

tcojean (approved):
LGTM

reference/test/stop/residual_norm_kernels.cpp
yhmtsai: done

yhmtsai:
format!

MarcelKoch (approved):

upsj (approved):

PR #1084, opened 2022-07-20
Push mirroring

tcojean:
This adds push mirroring for the current branch.

The way it works is simple, add the GitLab remote, the ssh key, push force, and prune as needed.

See: https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/pipelines/592229289

Questions: 
* can we assume there is always a branch?
* is there any implication of this change for our current workflow, do labels work as intended, etc?
* should we try pushing all branches simultaneously instead of only the current branch?
* ginkgo-bot will always be the triggerer in this way, should we reuse the author from the last commit instead?

### DIFF
--- a/.github/mirror.sh
+++ b/.github/mirror.sh
@@ -16,7 +16,7 @@ ssh-keyscan -t rsa gitlab.com github.com >>~/.ssh/known_hosts
 
 # Fetch from github
 git fetch fork "$BRANCH_NAME"
-git checkout -b "$BRANCH_NAME"
+git checkout -B "$BRANCH_NAME"
 git reset --hard fork/"$BRANCH_NAME"
 # Push to gitlab
-git push -u --force --prune gitlab HEAD:$BRANCH_NAME
+git push -u --force gitlab HEAD:$BRANCH_NAME
--- a/.github/mirror.sh
+++ b/.github/mirror.sh
@@ -13,8 +13,6 @@ echo "${BOT_KEY}" | tr -d '\r' | ssh-add - >/dev/null
 mkdir -p ~/.ssh
 chmod 700 ~/.ssh
 ssh-keyscan -t rsa gitlab.com github.com >>~/.ssh/known_hosts
-git config user.email "ginkgo.library@gmail.com"
-git config user.name "Ginkgo Bot"
 
 # Fetch from github
 git fetch fork "$BRANCH_NAME"
--- a/.github/mirror.sh
+++ b/.github/mirror.sh
@@ -18,6 +18,7 @@ git config user.name "Ginkgo Bot"
 
 # Fetch from github
 git fetch fork "$BRANCH_NAME"
-git checkout -b fork/$BRANCH_NAME
+git checkout -b "$BRANCH_NAME"
+git reset --hard fork/"$BRANCH_NAME"
 # Push to gitlab
-git push --force --prune gitlab $BRANCH_NAME
+git push -u --force --prune gitlab HEAD:$BRANCH_NAME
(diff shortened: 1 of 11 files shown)


upsj:
Since I've pushed to Gitlab manually before, I'd say this should work w.r.t. labels and statuses.
Not 100% happy with it from the security perspective - looks like we're pulling `mirror.sh` from the current branch. If we used pull_request_target and the sync operation, we'd always rely on the script in `develop`/`master`. That could also be extended to PR comments etc. We'd only lose the ability to automatically sync non-PR branches.

tcojean:
In terms of security, I agree we should pull from develop in the future like it is done for the other scripts, but that would require first merging this PR before adapting it.

Originally, I also wanted to do this through the PR setup, but I think we should still provide CI for non-PR branches as well so I did it this way.

upsj:
We can do something like I did with https://github.com/upsj/ginkgo-bot-playground to make sure things work before merging.

yhmtsai:
depends on who's the trigger and use develop or pull request one?
using wget to get the develop should always work?

tcojean:
Here is a new version of the script which works like the rest, but is still on `push` rather than on PR. It works even for develop (I reset `ginkgo-public-ci` to the correct commit afterward:

https://github.com/tcojean/ginkgo-bot-playground/runs/7433510967
https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/pipelines/592683601

.github/mirror.sh
```diff
@@ -0,0 +1,23 @@
...
+chmod 700 ~/.ssh
+ssh-keyscan -t rsa gitlab.com github.com >>~/.ssh/known_hosts
+git config user.email "ginkgo.library@gmail.com"
+git config user.name "Ginkgo Bot"
+
+# Fetch from github
+git fetch fork "$BRANCH_NAME"
+git checkout -b fork/$BRANCH_NAME
+# Push to gitlab
+git push --force --prune gitlab $BRANCH_NAME
```
yhmtsai, line 23: I need to use `HEAD:$BRANCH_NAME` or `fork/$BRACNH_NAME:$BRANCH_NAME`
otherwise, I get "error: src refspec test does not match any."

yhmtsai: `git checkout -b fork/$BRANCH_NAME` only creates the branch on the current folder, so it does not switch to the branch from fork

yhmtsai (changes requested):
it does not set correctly from the github action output in the playground
```
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.
Please make sure you have the correct access rights
and the repository exists.
```

.github/mirror.sh
yhmtsai: it seems not to delete the unexisting branch?

tcojean:
It now works also on the playground, it was mostly due to an error in the playground setup rather than in the script (ssh key setup).
See: https://github.com/tcojean/ginkgo-bot-playground/runs/7450389602

I already deleted the branch on the `ginkgo-public-ci` side.

.github/mirror.sh
```diff
@@ -0,0 +1,24 @@
...
+git remote -v
+
+# Setup ssh
+eval $(ssh-agent -s)
+echo "${BOT_KEY}" | tr -d '\r' | ssh-add - >/dev/null
+mkdir -p ~/.ssh
+chmod 700 ~/.ssh
+ssh-keyscan -t rsa gitlab.com github.com >>~/.ssh/known_hosts
+git config user.email "ginkgo.library@gmail.com"
+git config user.name "Ginkgo Bot"
```
upsj, line 17: nit: should be unnecessary, since you don't commit anything

upsj (approved):
LGTM! Thanks for tackling this!

yhmtsai (approved):
LGTM

PR #1087, opened 2022-07-22
Enable changing the residual tolerance after a batch solver is created

Slaedr:
This adds an interface to change the residual tolerance for subsequent runs of a solver, so that a new factory does not have to be created just for changing the tolerance. In cases where batch scaling is needed, this removes the need for redoing the setup task of scaling just for changing the tolerance, among potentially other things in the future.

A new, non-templated base class `BatchSolver` is added. This contains all functionality that is agnostic of the concrete type - preconditioner, generated preconditioner, scaling operators, residual tolerance, maximum iteration count. Thus, the user only needs a pointer to `BatchLinOp` for applying the solver, and a dynamic cast to `BatchSolver` to get and set/get solver-agnostic parameters like the residual tolerance.

The factory parameter is now renamed to `default_residual_tol`, and this is used to initialize the `residual_tol_` values of generated solvers. The `set_residual_tolerance(double)` member function directly sets the `residual_tol_` on the solver object.

Drawback:
- Casts are needed to the corresponding `real_type`s of the solvers. This was done so that the residual tolerance can be updated independent of the concrete batched solver type - the application does not always need to know the concrete type.

### DIFF
--- a/benchmark/solver/batch_solver.hpp
+++ b/benchmark/solver/batch_solver.hpp
@@ -295,8 +295,8 @@ std::unique_ptr<gko::BatchLinOpFactory> generate_solver(
     if (description == "richardson") {
         using Solver = gko::solver::BatchRichardson<etype>;
         return Solver::build()
-            .with_max_iterations(static_cast<int>(FLAGS_max_iters))
-            .with_residual_tol(
+            .with_default_max_iterations(static_cast<int>(FLAGS_max_iters))
+            .with_default_residual_tol(
                 static_cast<gko::remove_complex<etype>>(FLAGS_rel_res_goal))
             .with_preconditioner(prec_fact)
             .with_relaxation_factor(static_cast<gko::remove_complex<etype>>(
@@ -308,8 +308,8 @@ std::unique_ptr<gko::BatchLinOpFactory> generate_solver(
     } else if (description == "bicgstab") {
         using Solver = gko::solver::BatchBicgstab<etype>;
         return Solver::build()
-            .with_max_iterations(static_cast<int>(FLAGS_max_iters))
-            .with_residual_tol(
+            .with_default_max_iterations(static_cast<int>(FLAGS_max_iters))
+            .with_default_residual_tol(
                 static_cast<gko::remove_complex<etype>>(FLAGS_rel_res_goal))
             .with_preconditioner(prec_fact)
             .with_tolerance_type(toltype)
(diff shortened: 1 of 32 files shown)


include/ginkgo/core/solver/batch_solver.hpp
```diff
@@ -107,6 +109,21 @@ class EnableBatchSolver
         return right_scaling_;
     }
 
+    /**
+     * Get the residual tolerance used by the solver.
+     *
+     * @return The residual tolerance.
+     */
+    double get_residual_tolerance() const { return residual_tol_; }
```
pratikvn, line 117: I think we should maintain just one copy of the tolerance instead of two separate copies. If the users sets a tolerance, then we just modify the existing one instead of adding one more parameter. the `parameters_` struct already has a getter as well

upsj: Alternatively: Get rid of the parameters member entirely. It's hidden in one of the factory macros, but TBH I dislike how well they hide the additional member they add. I worked entirely without the macros in #1082, and it felt much cleaner :)

upsj: There are two types of parameters: Ones that describe how to generate an object, but can be discarded afterwards, and ones that describe the runtime behavior of the object. IMO, we should only keep the latter inside the generated object.

Slaedr: @upsj That has the advantage that I could add the residual tolerance, getter and setter only to the base class. The problem with the factory macros is that the entire struct needs to be in the derived class, so we'd need to duplicate the getter and setter on every batched iterative solver, which should ideally be unnecessary.

pratikvn: But I think having two things settable in two different ways might bring a lot more confusion, so we should unify those.

Slaedr:
format!

benchmark/solver/batch_solver.hpp
```diff
@@ -295,8 +295,8 @@ std::unique_ptr<gko::BatchLinOpFactory> generate_solver(
     if (description == "richardson") {
         using Solver = gko::solver::BatchRichardson<etype>;
         return Solver::build()
-            .with_max_iterations(static_cast<int>(FLAGS_max_iters))
-            .with_residual_tol(
+            .with_default_max_iterations(static_cast<int>(FLAGS_max_iters))
```
pratikvn, line 298: I think this is a nice idea. But `default` may be a bit confusing. How about `prescribed` or `prescr` for short ?

pratikvn (approved):
I like the renaming idea, though that might mean Cody might have to change a bit in his interface. I guess the XGC interface will also need to be updated. LGTM!

benchmark/solver/batch_solver.hpp
pratikvn: I guess `default` is also okay. So you can leave it as it is.

pratikvn (commented):
CI issue

core/solver/batch_dispatch.hpp
```diff
@@ -195,7 +196,9 @@ class BatchSolverDispatch {
         const gko::batch_dense::UniformBatch<const device_value_type>& b_b,
         const gko::batch_dense::UniformBatch<device_value_type>& x_b)
     {
-        if (!precon_) {
+        if (!precon_ ||
+            dynamic_cast<const matrix::BatchIdentity<device_value_type>*>(
+                precon_)) {
```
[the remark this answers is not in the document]
pratikvn: Yeah, the issue seems to be only on HIP and some CUDA versions (https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/pipelines/655579466)

PR #1088, opened 2022-07-27
Add algorithm choice for triangular solvers

fritzgoebel:
This PR adds a parameter to the triangular solvers for choosing the algorithm to be used rather than making this depend on the matrix spmv strategy. Currently it is only relevant for CUDA, as for the other executors there is only one algorithm available.


include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -590,6 +590,14 @@ class EnablePreconditionedIterativeSolver
 };
 
 
+/**
+ * A helper for algorithm selection in the triangular solvers.
+ * It currently only matters for the Cuda executor as there,
+ * we have a choice between the Ginkgo and Cusparse implementations.
```
upsj, line 596:
```suggestion
 * we have a choice between the Ginkgo and cuSPARSE implementations.
```

include/ginkgo/core/solver/lower_trs.hpp
```diff
@@ -116,6 +116,15 @@ class LowerTrs : public EnableLinOp<LowerTrs<ValueType, IndexType>>,
...
         bool GKO_FACTORY_PARAMETER_SCALAR(unit_diagonal, false);
+
+        /**
+         * Select the implementation which is supposed to be used for
+         * the triangular solver. This only matters for the Cuda
+         * executor where the choice is between the Ginkgo and the
+         * Cusparse implementation. Default is Ginkgo.
+         */
+        trisolve_algorithm GKO_FACTORY_PARAMETER_SCALAR(
+            algorithm, trisolve_algorithm::ginkgo);
```
upsj, line 127: I think the cuSPARSE algorithm is more robust, especially in CUDA >= 11.3, so I would use that one by default.

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -590,6 +590,14 @@ class EnablePreconditionedIterativeSolver
 };
 
 
+/**
```
upsj, line 593: Probably not the correct file for this. We could merge upper_trs.hpp and lower_trs.hpp though, leaving those two files to import a triangular.hpp header, since they are so closely related.

upsj (approved):
LGTM! Thanks for taking care of this!

cuda/solver/lower_trs_kernels.cu
```diff
@@ -73,9 +73,10 @@ template <typename ValueType, typename IndexType>
 void generate(std::shared_ptr<const CudaExecutor> exec,
               const matrix::Csr<ValueType, IndexType>* matrix,
               std::shared_ptr<solver::SolveStruct>& solve_struct,
-              bool unit_diag, const gko::size_type num_rhs)
+              bool unit_diag, const solver::trisolve_algorithm algorithm,
+              const gko::size_type num_rhs)
 {
-    if (matrix->get_strategy()->get_name() == "sparselib") {
+    if (algorithm == solver::trisolve_algorithm::sparselib) {
```
thoasm, line 79: What happens if `trisolve_algorithm::ginkgo` is chosen? I did not see any branch for that option. It it actually not doing anything?

thoasm (commented):
Mostly comments about the previous existing `gko::size_type` (unnecessary `gko::`).
But one question: Where is the action defined in case of `solver::trisolve_algorithm::ginkgo`? I did not find it anywhere.

cuda/solver/lower_trs_kernels.cu
thoasm: Looks like you don't have / need a generation step in that case, so I will mark it as resolved.

thoasm:
Currently, you are dependent on both the strategy of the matrix (for the solve) and the algorithm enum (for the generate).
Is there a reason why you did not make both the solving and the generation dependent on the algorithm enum `solver::trisolve_algorithm`?

fritzgoebel:
> Currently, you are dependent on both the strategy of the matrix (for the solve) and the algorithm enum (for the generate). Is there a reason why you did not make both the solving and the generation dependent on the algorithm enum `solver::trisolve_algorithm`?

Good catch, thanks! I completely overlooked the solve part.. will update this in a minute

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -590,6 +590,14 @@ class EnablePreconditionedIterativeSolver
...
+ * A helper for algorithm selection in the triangular solvers.
+ * It currently only matters for the Cuda executor as there,
+ * we have a choice between the Ginkgo and cuSPARSE implementations.
+ */
+enum class trisolve_algorithm { sparselib, ginkgo };
```
upsj, line 598: I think we can be more explicit about what the algorithm does. May add a `level` algorithm later?
```suggestion
enum class trisolve_algorithm { sparselib, syncfree };
```

core/solver/upper_trs_kernels.hpp
```diff
@@ -58,17 +58,19 @@ namespace upper_trs {
...
 
-#define GKO_DECLARE_UPPER_TRS_GENERATE_KERNEL(_vtype, _itype)              \
-    void generate(std::shared_ptr<const DefaultExecutor> exec,             \
-                  const matrix::Csr<_vtype, _itype>* matrix,               \
-                  std::shared_ptr<gko::solver::SolveStruct>& solve_struct, \
-                  bool unit_diag, const gko::size_type num_rhs)
+#define GKO_DECLARE_UPPER_TRS_GENERATE_KERNEL(_vtype, _itype)                 \
+    void generate(std::shared_ptr<const DefaultExecutor> exec,                \
+                  const matrix::Csr<_vtype, _itype>* matrix,                  \
+                  std::shared_ptr<gko::solver::SolveStruct>& solve_struct,    \
```
upsj, line 64: missing change:
```suggestion
                  std::shared_ptr<solver::SolveStruct>& solve_struct,    \
```

include/ginkgo/core/solver/solver_base.hpp
upsj, line 598: This should go into a separate file, I think

cuda/test/solver/lower_trs_kernels.cpp
```diff
@@ -133,12 +133,14 @@ TEST_F(LowerTrs, CudaLowerTrsFlagCheckIsCorrect)
 }
 
 
-TEST_F(LowerTrs, CudaSingleRhsApplyClassicalIsEquivalentToRef)
+TEST_F(LowerTrs, CudaSingleRhsApplySparselibIsEquivalentToRef)
```
yhmtsai, line 136: I think the path of classical is different from the sparselib

include/ginkgo/core/solver/lower_trs.hpp
```diff
@@ -116,6 +116,15 @@ class LowerTrs : public EnableLinOp<LowerTrs<ValueType, IndexType>>,
...
         bool GKO_FACTORY_PARAMETER_SCALAR(unit_diagonal, false);
+
+        /**
+         * Select the implementation which is supposed to be used for
+         * the triangular solver. This only matters for the Cuda
+         * executor where the choice is between the Ginkgo and the
+         * cuSPARSE implementation. Default is Ginkgo.
+         */
+        trisolve_algorithm GKO_FACTORY_PARAMETER_SCALAR(
+            algorithm, trisolve_algorithm::sparselib);
```
yhmtsai, line 127: the default does not match the description

include/ginkgo/core/solver/solver_base.hpp
yhmtsai: Is the syncfree only on certain architecture like Volta and Ampere?

include/ginkgo/core/solver/upper_trs.hpp
```diff
@@ -116,6 +116,15 @@ class UpperTrs : public EnableLinOp<UpperTrs<ValueType, IndexType>>,
...
         bool GKO_FACTORY_PARAMETER_SCALAR(unit_diagonal, false);
+
+        /**
+         * Select the implementation which is supposed to be used for
+         * the triangular solver. This only matters for the Cuda
+         * executor where the choice is between the Ginkgo and the
+         * Cusparse implementation. Default is Ginkgo.
+         */
+        trisolve_algorithm GKO_FACTORY_PARAMETER_SCALAR(
+            algorithm, trisolve_algorithm::sparselib);
```
yhmtsai, line 127: same. use cuSPARSE like lower_trs

include/ginkgo/core/solver/solver_base.hpp
upsj: no, it also works (with a slightly different kernel) on Pascal etc. This is an algorithmic difference: Fine-grained parallelism or level parallelism

cuda/test/solver/lower_trs_kernels.cpp
fritzgoebel: Yeah, now it should be the default (sparselib) and syncfree

include/ginkgo/core/solver/triangular.hpp
```diff
@@ -0,0 +1,384 @@
...
+
+struct SolveStruct;
+
+
+/**
+ * A helper for algorithm selection in the triangular solvers.
+ * It currently only matters for the Cuda executor as there,
+ * we have a choice between the Ginkgo syncfree and cuSPARSE implementations.
+ */
+enum class trisolve_algorithm { sparselib, syncfree };
```
yhmtsai, line 67: I think it is the only necessary part you need in the triangular.hpp. upper_trs and lower_trs include triangular.hpp should be enough

upsj: The triangular solvers are entirely identical except for a bool parameter. A more maintainable implementation in the long-term would be making the differentiation between lower and upper a factory or template parameter. The former would require a default (for which there is no sensible choice) and the latter would require only type aliases.

yhmtsai: Yes, I agree with using template type and using type aliases. 
Because we do not get any benefit from just putting them together, I will prefer not moving the code.

include/ginkgo/core/solver/upper_trs.hpp
```diff
@@ -34,181 +34,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-    // transposed input vector
-    constexpr static int transposed_b = 0;
-    // transposed output vector
-    constexpr static int transposed_x = 1;
-};
-
-
-}  // namespace solver
-}  // namespace gko
+#include <ginkgo/core/solver/triangular.hpp>
```
yhmtsai, line 37: could you add the deprecation warning? same in lower_trs.

include/ginkgo/ginkgo.hpp
```diff
@@ -125,6 +125,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/solver/multigrid.hpp>
 #include <ginkgo/core/solver/solver_base.hpp>
 #include <ginkgo/core/solver/solver_traits.hpp>
+#include <ginkgo/core/solver/triangular.hpp>
```
yhmtsai, line 128: only include triangular.hpp not upper_trs/lower_trs

yhmtsai (approved):
LGTM in general.
If you add the warning into upper_trs/lower_trs and only use triangular.hpp in other files, the moving code will make sense to me.

include/ginkgo/ginkgo.hpp
upsj: need to adapt the script that generates it

upsj:
rebase!

PR #1089, opened 2022-07-29
WIP: Add a multi-level Schwarz preconditioner

pratikvn:
This PR adds a multi-level Schwarz preconditioner to Ginkgo. The main objective is to be able to use this as preconditioners for Krylov solvers, but combined with IR, it can be used as a solver as well. 

### TODO:

+ [ ] Kernels for CUDA/HIP/DPCPP 
+ [ ] Store coarse solvers as Multigrid levels (RAP)
+ [ ] Incorporate Overlap.

Currently, this is a single GPU version, but the core class and implementation should carry over to the distributed setup.


pratikvn:
format!

pratikvn:
format!

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -817,6 +819,35 @@ class Dense
      */
     void compute_norm1(LinOp* result, array<char>& tmp) const;
 
+    /**
+     * Create a submatrix from the original matrix.
+     * Warning: defining stride for this create_submatrix method might cause
+     * wrong memory access. Better use the create_submatrix(rows, columns)
+     * method instead.
```
yhmtsai, line 826: @copydoc?

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -817,6 +819,35 @@ class Dense
...
+     *
+     * @param rows     row span
+     * @param columns  column span
+     * @param stride   stride of the new submatrix.
+     */
+    std::unique_ptr<const Dense> create_submatrix(const span& rows,
```
yhmtsai, line 832:
```suggestion
    std::unique_ptr<const Dense> create_const_submatrix(const span& rows,
```
maybe use create_const_submatrix to make explict call like others `get_values/get_const_value`

include/ginkgo/core/preconditioner/schwarz.hpp
```diff
@@ -0,0 +1,268 @@
...
+         *        The matrix must be sorted for this preconditioner to work.
+         *
+         * The `system_matrix`, which will be given to this factory, must be
+         * sorted (first by row, then by column) in order for the algorithm
+         * to work. If it is known that the matrix will be sorted, this
+         * parameter can be set to `true` to skip the sorting (therefore,
+         * shortening the runtime).
+         * However, if it is unknown or if the matrix is known to be not sorted,
+         * it must remain `false`, otherwise, this preconditioner might be
+         * incorrect.
```
yhmtsai, line 180: repeated description?

include/ginkgo/core/preconditioner/schwarz.hpp
```diff
@@ -0,0 +1,268 @@
...
+         *
+         * @param coarse_operator The object that stores the MultigridLevel
+         *                        object, which in turn stores the coarse
+         *                        operator, restriction operator and the
+         *                        prolongation operator
+         *
+         * @param coarse_factory The LinOpFactory object that stores the
+         *                       solver to be generated on the coarse operator
+         */
+        coarse_solver_type(
```
yhmtsai, line 136: I think it can be done from outside?

core/preconditioner/schwarz.cpp
```diff
@@ -0,0 +1,208 @@
...
+    const auto exec = this->get_executor();
+    auto csr_mtx =
+        convert_to_with_sorting<csr_type>(exec, system_matrix, skip_sorting);
+    size_type offset = 0;
+    auto mat_size = system_matrix->get_size();
+    std::vector<size_type> subd_sizes;
+    if (parameters_.subdomain_sizes.size() == 0) {
+        subd_sizes = std::vector<size_type>(num_subdomains_,
+                                            mat_size[0] / num_subdomains_);
+        subd_sizes[0] += mat_size[0] - (num_subdomains_ * subd_sizes[1]);
```
yhmtsai, line 171: do you make the first one containing the rest part?
it's common to combine the rest part into one block, or use another block for the rest

core/preconditioner/schwarz.cpp
```diff
@@ -0,0 +1,208 @@
...
+            size_type offset = 0;
+            // TODO Replace with BlockApprox
+            for (size_type i = 0; i < this->num_subdomains_; ++i) {
+                size_type l_num_rows = parameters_.subdomain_sizes[i];
+                auto rspan = gko::span(offset, offset + l_num_rows);
+                const auto b_view = dense_b->create_submatrix(
+                    rspan, gko::span(0, dense_b->get_size()[1]));
+                auto x_view = dense_x->create_submatrix(
+                    rspan, gko::span(0, dense_x->get_size()[1]));
+                this->subdomain_solvers_[i]->apply(dense_alpha, b_view.get(),
```
yhmtsai, line 132: advanced_apply does not do the same thing as apply

yhmtsai (changes requested):

core/preconditioner/schwarz.cpp
tcojean, line 171: to make this slightly clearer, store `mat_size[0] / num_subdomains_` in a variable and use the variable instead?
Also, couldn't this break if `num_subdomains_ ==1` ?

core/preconditioner/schwarz.cpp
```diff
@@ -0,0 +1,208 @@
...
+        size_type l_num_rows = subd_sizes[i];
+        auto rspan = gko::span(offset, offset + l_num_rows);
+        auto cspan = gko::span(offset, offset + l_num_rows);
+        subdomain_matrices_.emplace_back(
+            gko::share(csr_mtx->create_submatrix(rspan, cspan)));
+        offset += l_num_rows;
+    }
+    if (parameters_.generated_inner_solvers.size() > 0) {
+        GKO_ASSERT(parameters_.generated_inner_solvers.size() ==
+                   num_subdomains_);
```
tcojean, line 187: Maybe propagate them to `subdomain_solvers_` to have a coherent state?

core/preconditioner/schwarz.cpp
```diff
@@ -0,0 +1,208 @@
...
+    } else if (parameters_.inner_solver) {
+        for (size_type i = 0; i < num_subdomains_; ++i) {
+            subdomain_solvers_.emplace_back(
+                parameters_.inner_solver->generate(subdomain_matrices_[i]));
+        }
+    } else {
+        GKO_NOT_IMPLEMENTED;
+    }
+    if (parameters_.coarse_solver.uses_coarse_solver_) {
```
tcojean, line 196: Instead of the `uses_coarse_solver_` boolean, can't you rely on whether `coarse_solver_` is nullptr and remove the need for the bool?

core/preconditioner/schwarz.cpp
tcojean, line 132: This might also not work when the user fills `generated_inner_solvers` as that is not propagated to that variable.

core/preconditioner/schwarz.cpp
```diff
@@ -0,0 +1,208 @@
...
+        auto restricted_r = Vector::create(
+            exec,
+            gko::dim<2>(restrict_op->get_size()[0], dense_x->get_size()[1]));
+        restrict_op->apply(r.get(), restricted_r.get());
+        auto restricted_e = Vector::create(
+            exec,
+            gko::dim<2>(restrict_op->get_size()[0], dense_x->get_size()[1]));
+        restricted_e->fill(zero<ValueType>());
+        parameters_.coarse_solver.coarse_solver_->apply(restricted_r.get(),
+                                                        restricted_e.get());
```
tcojean, line 107: You can just use the `coarse_solver_` from `this` here?

core/preconditioner/schwarz.cpp
```diff
@@ -0,0 +1,208 @@
...
+        auto x_view = dense_x->create_submatrix(
+            rspan, gko::span(0, dense_x->get_size()[1]));
+        this->subdomain_solvers_[i]->apply(b_view.get(), x_view.get());
+        offset += l_num_rows;
+    }
+    if (parameters_.coarse_solver.uses_coarse_solver_) {
+        auto r = Vector::create_with_config_of(dense_b);
+        r->copy_from(dense_b);
+        auto restrict_op =
+            parameters_.coarse_solver.coarse_op_->get_restrict_op();
```
tcojean, line 93: Since this is pretty long, you could also store a pointer to the `coarse_op_` in this class?

core/preconditioner/schwarz.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+template <typename ValueType, typename IndexType>
+void Schwarz<ValueType, IndexType>::generate(const LinOp* system_matrix,
+                                             bool skip_sorting)
+{
+    GKO_ASSERT_IS_SQUARE_MATRIX(system_matrix);
+    using csr_type = matrix::Csr<ValueType, IndexType>;
+    const auto exec = this->get_executor();
+    auto csr_mtx =
+        convert_to_with_sorting<csr_type>(exec, system_matrix, skip_sorting);
```
tcojean, line 164: I wonder, in the distributed case do you get the `system_matrix` as something of the type `Matrix` (distributed) or do you get directly the local matrix which you can convert? Is that natural (no user involvement) in terms of interface? I guess that will depend on the distributed solver implementations as well.

reference/test/preconditioner/schwarz.cpp
```diff
@@ -0,0 +1,225 @@
...
+    Schwarz()
+        : exec(gko::ReferenceExecutor::create()),
+          mtx(gko::matrix::Csr<value_type, index_type>::create(
+              exec, gko::dim<2>{5}, 13))
+    {
+        inner_factory =
+            gko::share(InnerSolver::build().with_max_block_size(5u).on(exec));
+        schwarz_factory = gko::share(
+            Prec::build()
+                .with_subdomain_sizes(std::vector<gko::size_type>{2, 2, 1})
```
tcojean, line 81: It might make sense to try out with diverging `subdomain_sizes` and `num_subdomains` values. As well as try some edge cases (e.g., only one subdomain).

tcojean (approved):
I think this is quite good, but I have some comments and questions.

include/ginkgo/core/preconditioner/schwarz.hpp
```diff
@@ -0,0 +1,268 @@
...
+    std::vector<std::shared_ptr<LinOp>> get_subdomain_matrices() const
+    {
+        return subdomain_matrices_;
+    }
+
+    /**
+     * A struct that stores the coarse operator and solver and generates the
+     * coarse solver with the given operator
+     */
+    struct coarse_solver_type {
```
MarcelKoch, line 113: Why is that necessary? Can't you use a `Multigrid` solver with 2 levels?

core/preconditioner/schwarz.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+template <typename ValueType, typename IndexType>
+void Schwarz<ValueType, IndexType>::apply_dense_impl(
+    const matrix::Dense<ValueType>* dense_b,
+    matrix::Dense<ValueType>* dense_x) const
+{
+    using Vector = matrix::Dense<ValueType>;
+    auto exec = this->get_executor();
+    auto one_op = initialize<Vector>({one<ValueType>()}, exec);
+    auto neg_one_op = initialize<Vector>({-one<ValueType>()}, exec);
```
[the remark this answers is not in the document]
pratikvn: It is not a solver, so it does not have a workspace it.

include/ginkgo/core/preconditioner/schwarz.hpp
```diff
@@ -0,0 +1,268 @@
...
+        /**
+         * Array of subdomain sizes.
+         */
+        std::vector<size_type> GKO_FACTORY_PARAMETER_VECTOR(
+            subdomain_sizes, std::vector<size_type>{});
+
+        /**
+         * Number of subdomains.
+         */
+        size_type GKO_FACTORY_PARAMETER_SCALAR(num_subdomains, 1);
```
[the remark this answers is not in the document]
pratikvn: Only one of the two is used. I will throw if they are not consistent

core/preconditioner/schwarz.cpp
pratikvn: I think in general it is easier to add it to the first block, especially in this case as in general you could expect the number of rows in the rest should be fairly small. In any case, someone who wants to use an optimized should not be using this anyway and should be providing specific subdomain sizes.

core/preconditioner/schwarz.cpp
pratikvn: Do you mean for the coarse solver ?

examples/schwarz-preconditioned-solver/doc/builds-on
```diff
@@ -0,0 +1 @@
+simple-solver
```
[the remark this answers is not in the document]
pratikvn: I wouldnt say this particularly builds on that example, but maybe is an alternative to that.

include/ginkgo/core/preconditioner/schwarz.hpp
pratikvn: No, a multigrid is different because it restricts the error and only does a solve on the coarse grid. What I want to possibly do is solve on each level. So, what I need is the functionality of MultigridLevel (which stores the R, A and P) and a LinOpFactory for each level, so I think this struct combines both and generates the coarse solvers from the `MultigridLevel`s.

core/preconditioner/schwarz.cpp
yhmtsai: Yes, the apply contains optional coarse solver

include/ginkgo/core/preconditioner/schwarz.hpp
MarcelKoch: What do you mean by solve on each level? Why is that different from setting the coarsest level to the one you are doing the solve on, and just leave out the rest?

core/preconditioner/schwarz.cpp
```diff
@@ -0,0 +1,207 @@
...
+                Vector::create(exec, gko::dim<2>(restrict_op->get_size()[0],
+                                                 dense_x->get_size()[1]));
+            restrict_op->apply(r.get(), restricted_r.get());
+            auto restricted_e =
+                Vector::create(exec, gko::dim<2>(restrict_op->get_size()[0],
+                                                 dense_x->get_size()[1]));
+            restricted_e->fill(zero<ValueType>());
+            coarse_solvers_[i]->apply(restricted_r.get(), restricted_e.get());
+            prolong_op->apply(one_op.get(), restricted_e.get(), one_op.get(),
+                              dense_x);
```
yhmtsai, line 108: isn't it be the 2-level multigrid->apply(r, x)?

yhmtsai: if it is, I think it can be another LinOpFactory more general

include/ginkgo/core/preconditioner/schwarz.hpp
```diff
@@ -0,0 +1,234 @@
...
+         * Generated Inner solvers.
+         */
+        std::vector<std::shared_ptr<const LinOp>> GKO_FACTORY_PARAMETER_VECTOR(
+            generated_inner_solvers, nullptr);
+
+        /**
+         * Coarse Operators as MultigridLevel
+         */
+        std::vector<std::shared_ptr<const multigrid::MultigridLevel>>
+            GKO_FACTORY_PARAMETER_VECTOR(coarse_operators, nullptr);
```
upsj, line 148: Storing an object by an interface base class is not really a good approach, since you can't take care of using the correct executor, create copies etc. This is also an interface issue with Multigrid, but we need to handle that separately.

include/ginkgo/core/preconditioner/schwarz.hpp
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+
+    GKO_CREATE_FACTORY_PARAMETERS(parameters, Factory)
+    {
+        /**
+         * Array of subdomain sizes.
+         */
+        std::vector<size_type> GKO_FACTORY_PARAMETER_VECTOR(
+            subdomain_sizes, std::vector<size_type>{});
```
upsj, line 115: structural blocks may work well work block Jacobi, but I don't think this maps well to larger domains, and even less so to overlaps (requiring explicit reordering to make it work generically). I think we probably need to integrate the Partition and corresponding kernels somehow.

include/ginkgo/core/preconditioner/schwarz.hpp
```diff
@@ -0,0 +1,234 @@
...
+ * @tparam ValueType  precision of matrix elements
+ * @tparam IndexType  integral type of the preconditioner
+ *
+ * @ingroup schwarz
+ * @ingroup precond
+ * @ingroup LinOp
+ */
+template <typename ValueType = default_precision, typename IndexType = int32>
+class Schwarz : public EnableLinOp<Schwarz<ValueType, IndexType>>,
+                public WritableToMatrixData<ValueType, IndexType>,
```
upsj, line 72: Schwarz is a preconditioner that doesn't explicitly build its inverses. I don't think we should claim it can write out a matrix representation. If it's used for debugging, maybe add a different function that writes out the subdomain matrices?

upsj (changes requested):
I am hestitant to use the same design for both distributed and single-executor Schwarz preconditioners. The stored data (global matrix vs. local matrix), ownership of the subdomains (all on the same rank vs. one per rank) and partitioning mode (diagonal blocks vs. general Partition) make it hard to imagine the two being implemented behind the same interface. Can you elaborate how this would be extended to the distributed case?

In general, what is the use case for this implementation vs. a distributed one?

include/ginkgo/core/preconditioner/schwarz.hpp
pratikvn: Yes, I agree that for larger subdomains structural blocks might not be ideal, but I think for a single executor implementation, it is a good start. For single executor implementation, you can use an index_set or partition as you mention to construct non-contiguous blocks, but the problem with those is that you have to deal with a lot of complications: map rhs and solution vector to correct indices/maintain copies and copy back to the vector. So in general, it might be better to reorder the system matrix before hand and use contiguous blocks.

include/ginkgo/core/preconditioner/schwarz.hpp
pratikvn: I agree, but I am not sure there is another option right now other than storing the three operators separately. Maybe we fix it for both at once by adding some concrete class that actually stores the three operators that the MultigridLevel provides getters for.

pratikvn:
@upsj What do you think of two separate classes, one for distributed and one for single GPU , differing in namespaces `gko::preconditioner::Schwarz` and `gko::distributed::preconditioner:Schwarz`.

upsj:
@pratikvn sorry for leaving this hanging. Yes, that's what I had in mind. It might be that there is some overlap (haha 😁) in the kernels that can be used to set up overlap information, but the interfaces are probably too different to be useful. Still, I think the interfaces should both be designed with overlap in mind, do you have a prototype with overlap capabilities somewhere that we can use as guidance?

pratikvn:
Closing as the same interface for distributed and non-distributed is not reasonable.

pratikvn:
@upsj, I had a preliminary overlap interface somewhere, I will have to dig that up. But it was still only contiguous overlaps (ranges) and not disjoint. To be more generic, It would be better to have an interface that supports disjoint overlaps, which I still need to think about a bit.

PR #1091, opened 2022-08-15
add hip unsafe atomic option

yhmtsai:
AMD introduced the `unsafe-fp-atomics` options to make atomicAdd faster.
unsafe atomicAdd is correct when the memory is on coarse grain.
hipMalloc always allocates memory on coarse grain, so we can use it by default.
although `unsafe-fp-atomics` should be added in clang-12, it is not always true when hip uses clang on their own repo.
llvm: https://github.com/llvm/llvm-project/commit/6054a456daf323294d6f9101b9a5a2345d6b5400
amd llvm: https://github.com/RadeonOpenCompute/llvm-project/commit/189310a140fa1c33f8f4838560f567bab9e99245
but amd shows the rocm >= 5 should use the llvm with unsafe atomic support.

Another note: only 5.2 requires compiler to use unsafe, but the rocm < 5.2 only suggests compiler such that it may not really use unsafe.


common/cuda_hip/components/atomic.hpp.inc
```diff
@@ -147,6 +147,17 @@ GKO_BIND_ATOMIC_ADD(unsigned int);
...
+    GINKGO_HIP_PLATFORM_HCC
+
+
+// the double atomicAdd is added after 4.3
+GKO_BIND_ATOMIC_ADD(double);
+
+
+#endif  // defined(__HIPCC__) && defined(__HIP_DEVICE_COMPILE__) &&
+        // GINKGO_HIP_PLATFORM_HCC
+
```
upsj, line 160: Can we merge this with the CUDA condition, so we have only a single place where we control access to double atomics?

cmake/hip.cmake
```diff
@@ -212,9 +212,13 @@ if(GINKGO_HIP_AMDGPU)
     endforeach()
 endif()
 
+
 set(GINKGO_HIPCC_OPTIONS ${GINKGO_HIP_COMPILER_FLAGS} "-std=c++14 -DGKO_COMPILING_HIP")
 set(GINKGO_HIP_NVCC_OPTIONS ${GINKGO_HIP_NVCC_COMPILER_FLAGS} ${GINKGO_HIP_NVCC_ARCH} ${GINKGO_HIP_NVCC_ADDITIONAL_FLAGS})
 set(GINKGO_HIP_CLANG_OPTIONS ${GINKGO_HIP_CLANG_COMPILER_FLAGS} ${GINKGO_AMD_ARCH_FLAGS})
+if(GINKGO_HIP_AMD_UNSAFE_ATOMIC AND HIP_VERSION VERSION_GREATER_EQUAL 5)
+set(GINKGO_HIP_CLANG_OPTIONS ${GINKGO_HIP_CLANG_OPTIONS} "-munsafe-fp-atomics")
```
upsj, line 220: nit: indentation + CMake stuff
```suggestion
    list(APPEND GINKGO_HIP_CLANG_OPTIONS -munsafe-fp-atomics)
```

CMakeLists.txt
```diff
@@ -67,6 +67,7 @@ set(GINKGO_CUDA_COMPILER_FLAGS "" CACHE STRING
 set(GINKGO_CUDA_ARCHITECTURES "Auto" CACHE STRING
     "A list of target NVIDIA GPU achitectures. See README.md for more detail.")
 option(GINKGO_CUDA_DEFAULT_HOST_COMPILER "Tell Ginkgo to not automatically set the CUDA host compiler" OFF)
+option(GINKGO_HIP_AMD_UNSAFE_ATOMIC "Compiler uses unsafe floating point atomic (only for AMD GPU and ROCM >= 5). Default is ON because we use hipMalloc, which is always on coarse grain. Must turn off when allocating memory on fine grain" ON)
```
upsj, line 70: We are lacking an explanation what coarse grain and fine grain allocations are somewhere in the documentation. Are those the terms AMD uses somewhere? Maybe link to that

upsj (approved):

tcojean (approved):
LGTM. I'm curious about documentation on these unsafe atomic and "fine-grained" vs "coarse-grained" memory allocations, because I could not easily find something clear on that topic.

yhmtsai:
I only found the reference here.
https://docs.olcf.ornl.gov/systems/crusher_quick_start_guide.html#floating-point-fp-atomic-operations-and-coarse-fine-grained-memory-allocations

CMakeLists.txt
upsj: So basically hipMalloc is fine, only hipMallocHost causes issues? We don't use it, so that's fine :) I think the documentation should mainly refer to the consequences of this choice.

common/cuda_hip/components/atomic.hpp.inc
yhmtsai: yes in practice with considering __HIP_DEVICE_COMPILE__ is always true in atomic_add.hpp.inc when compiled by HIPCC.
otherwise, it will not reduce the condition.

CMakeLists.txt
yhmtsai: yes, I add the ref in the comment

PR #1100, opened 2022-08-21
fix warning: comparison of integer expressions of different signedness

lahwaacz:
This fixes a compiler warning in the `mpi.hpp` file:
```
/home/lahwaacz/.local/include/ginkgo/core/base/mpi.hpp: In function ‘std::vector<gko::mpi::status> gko::mpi::wait_all(std::vector<gko::mpi::request>&)’:
/home/lahwaacz/.local/include/ginkgo/core/base/mpi.hpp:277:20: warning: comparison of integer expressions of different signedness: ‘int’ and ‘std::vector<gko::mpi::request>::size_type’ {aka ‘long unsigned int’} [-Wsign-compare]
  277 |     for (auto i = 0; i < req.size(); ++i) {
      |                  ~~^~~~~~~~~~~~
```
Here `auto` cannot be used, because it takes the type of `0`, which is `int`. Later it is compared with `req.size()` which leads to the warning.

### DIFF
--- a/include/ginkgo/core/base/mpi.hpp
+++ b/include/ginkgo/core/base/mpi.hpp
@@ -274,7 +274,7 @@ private:
 inline std::vector<status> wait_all(std::vector<request>& req)
 {
     std::vector<status> stat;
-    for (auto i = 0; i < req.size(); ++i) {
+    for (std::size_t i = 0; i < req.size(); ++i) {
         stat.emplace_back(req[i].wait());
     }
     return stat;


upsj:
An alternative fix would be to make the return type of size() signed and use `int` instead, in the spirit of #430. The MPI wrappers are not yet part of a release, so changing their interface is fine

lahwaacz:
It's `std::vector<gko::mpi::request>::size_type`, you can't easily change that.

upsj:
Ah sorry, I misread the diff ;)

MarcelKoch (approved):
LGTM, thanks for fixing this!

upsj (approved):

upsj:
label!

upsj:
check-format!

pratikvn:
@lahwaacz, thank you for the contribution. Please add yourself to the [contributors list](https://github.com/ginkgo-project/ginkgo/blob/develop/contributors.txt) before we merge this PR.

pratikvn (approved):

lahwaacz:
@pratikvn Is it really necessary? I don't need to be mentioned and I wouldn't like to make my email address public. If it's about the license statement, can I make it here in the discussion? Though I think it's overly formal for a one-line fix, obviously I wouldn't make the PR if I had problems with the license so I think something like a "take it or leave it" approach would work best for me.

upsj:
Your email address is already visible in the git history, but I agree that for such a trivial change, it is not necessary to list you as a contributor.

tcojean (approved):

PR #1101, opened 2022-08-22
Deprecates specialized residual norm classes

MarcelKoch:
This PR adds the `[[deprecate]]` attribute to the specialized residual norm classes, in order to make it clearer that they should not be used.


include/ginkgo/core/stop/residual_norm.hpp
```diff
@@ -326,13 +326,14 @@ class ImplicitResidualNorm : public ResidualNormBase<ValueType> {
  * `residual_norm` or the `residual` being set. When any of those is not
  * correctly provided, an exception ::gko::NotSupported() is thrown.
  *
- * @deprecated Please use the class ResidualNorm with the factory parameter
- *             baseline = mode::initial_resnorm
- *
```
yhmtsai, line 331: we can still keep this for doxygen

yhmtsai (approved):

test/test_install/test_install.cpp
```diff
@@ -504,18 +504,6 @@ int main()
...
-                           .on(exec);
-
-        auto rel_res =
-            gko::stop::RelativeResidualNorm<>::build().with_tolerance(1e-10).on(
-                exec);
-
-        auto abs_res =
-            gko::stop::AbsoluteResidualNorm<>::build().with_tolerance(1e-10).on(
-                exec);
-
```
pratikvn, line 518: I think we should still test that these are available as they are still available but only deprecated.

pratikvn (approved):
LGTM!

test/test_install/test_install.cpp
upsj: Testing for them leads to deprecation warnings, which breaks our -Werror CI job - difficult balance 🙁

pratikvn: Maybe we can disable -Werror for this one file  ?

upsj: removing flags is much harder than adding them - I'd try to avoid this

tcojean (approved):
LGTM

include/ginkgo/core/stop/residual_norm.hpp
```diff
@@ -312,6 +312,18 @@ class ImplicitResidualNorm : public ResidualNormBase<ValueType> {
...
+// The following classes are deprecated, but they internally reference
+// themselves. To reduce unnecessary warnings, we disable deprecation warnings
+// for the definition of these classes.
+#if defined(__GNUC__) || defined(__clang__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#elif defined(_MSC_BUILD) || defined(__INTEL_LLVM_COMPILER)
+#pragma warning(push)
+#pragma warning(disable : 4996)
+#endif
```
yhmtsai, line 324: it only avoids the warning when we use it internally, but it still gives warning when users use it, right?

yhmtsai (approved):

include/ginkgo/core/stop/residual_norm.hpp
MarcelKoch: Yes, this only disables the warning for the class definitions. After that, the previous warning state is restored.

MarcelKoch:
rebase!

PR #1105, opened 2022-08-27
Move implementation from headers to source files

upsj:
This PR moves almost all function implementations from headers to source files. This should make the headers easier to read and maybe also faster to compile.

To finish this up, I would like to suggest additional changes:

* I held off on MPI, since the interface is still in flux, but lots of the non-templated functions could be moved to source files as well
* Dense::at should be deprecated in favor of a function that returns a row-major accessor. This gives us a single consistent, but efficient way of accessing entries.
* Ell::val_at and Ell::col_at should in the same way be replaced by a column-major accessor
* For Sellp, Fbcsr and Csr, it might make sense to also provide accessors with variable-size semantics, but I haven't thought this part through yet. At least Sellp has similar access functions to Ell, so it would probably be a good start.

TODO:
- [ ] Deal with ExecutorBase - that probably requires creating a new library `ginkgo_executor` that has all executor-specific functionality, and implement it for different backends. That would also slim down our link dependencies a bit. graphviz

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1202932419787352

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -496,6 +496,7 @@ build/amd/mvapich2/gcc/rocm45/release/shared:
     - .quick_test_condition
     - .use_gko-rocm45-mvapich2-gnu8-llvm8
   variables:
+    BUILD_MPI: "ON"
     BUILD_OMP: "ON"
     BUILD_HIP: "ON"
     RUN_EXAMPLES: "ON"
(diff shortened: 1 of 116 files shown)


upsj:
format-rebase!

upsj:
format-rebase!

upsj:
This is quite outdated and likely needs to be redone from scratch, I'll close it for now

PR #1109, opened 2022-08-31
Fix PackageConfig and Create the test

### LINKED ISSUE #1104: Linking to static Ginkgo libs using pkg-config does not work
I'm building a custom CUDA solver with Ginkgo static libs using the flags specified in [ginkgo.pc](https://github.com/ginkgo-project/ginkgo/blob/develop/cmake/ginkgo.pc.in#L11) and getting this error:
```
g++ -pthread -L /opt/cuda/lib64/ -o MyApp MyApp.cuo -lstdc++fs -lcudart -ldl -lrt -lz -ltinyxml2 -L /tmp/project/build/ginkgo_Release/lib -lginkgo -lginkgo_device /usr/lib/libhwloc.so -lginkgo_omp /usr/lib/libgomp.so /usr/lib/libpthread.a -lginkgo_cuda /opt/cuda/targets/x86_64-linux/lib/libcudart.so /opt/cuda/targets/x86_64-linux/lib/stubs/libcublas.so /opt/cuda/targets/x86_64-linux/lib/stubs/libcusparse.so /opt/cuda/targets/x86_64-linux/lib/stubs/libcurand.so /opt/cuda/targets/x86_64-linux/lib/stubs/libcufft.so -lginkgo_reference -lginkgo_hip -lginkgo_dpcpp  -lgomp
/usr/bin/ld: /tmp/project/build/ginkgo_Release/lib/libginkgo_cuda.a(executor.cpp.o): in function `std::_Sp_counted_deleter<gko::CudaExecutor*, gko::CudaExecutor::create(int, std::shared_ptr<gko::Executor>, bool, gko::allocation_mode)::{lambda(gko::CudaExecutor*)#1}, std::allocator<void>, (__gnu_cxx::_Lock_policy)2>::_M_dispose()':
executor.cpp:(.text+0x41f): undefined reference to `gko::nvidia_device::get_mutex(int)'
/usr/bin/ld: executor.cpp:(.text+0x444): undefined reference to `gko::nvidia_device::get_num_execs(int)'
/usr/bin/ld: /tmp/project/build/ginkgo_Release/lib/libginkgo_cuda.a(executor.cpp.o): in function `gko::CudaExecutor::create(int, std::shared_ptr<gko::Executor>, bool, gko::allocation_mode) [clone .cold]':
executor.cpp:(.text.unlikely+0x3dea): undefined reference to `gko::nvidia_device::get_mutex(int)'
/usr/bin/ld: executor.cpp:(.text.unlikely+0x3e10): undefined reference to `gko::nvidia_device::get_num_execs(int)'
/usr/bin/ld: /tmp/project/build/ginkgo_Release/lib/libginkgo_cuda.a(executor.cpp.o): in function `gko::CudaExecutor::increase_num_execs(unsigned int)':
executor.cpp:(.text+0xa9): undefined reference to `gko::nvidia_device::get_mutex(int)'
/usr/bin/ld: executor.cpp:(.text+0xbf): undefined reference to `gko::nvidia_device::get_num_execs(int)'
/usr/bin/ld: /tmp/project/build/ginkgo_Release/lib/libginkgo_cuda.a(executor.cpp.o): in function `gko::CudaExecutor::decrease_num_execs(unsigned int)':
executor.cpp:(.text+0xf9): undefined reference to `gko::nvidia_device::get_mutex(int)'
/usr/bin/ld: executor.cpp:(.text+0x10f): undefined reference to `gko::nvidia_device::get_num_execs(int)'
/usr/bin/ld: /tmp/project/build/ginkgo_Release/lib/libginkgo_cuda.a(executor.cpp.o): in function `gko::CudaExecutor::get_num_execs(unsigned int)':
executor.cpp:(.text+0x149): undefined reference to `gko::nvidia_device::get_mutex(int)'
/usr/bin/ld: executor.cpp:(.text+0x15f): undefined reference to `gko::nvidia_device::get_num_execs(int)'
/usr/bin/ld: /tmp/project/build/ginkgo_Release/lib/libginkgo_hip.a(executor.cpp.o): in function `gko::HipExecutor::get_num_execs(int)':
executor.cpp:(.text+0xc9): undefined reference to `gko::amd_device::get_mutex(int)'
/usr/bin/ld: executor.cpp:(.text+0xdf): undefined reference to `gko::amd_device::get_num_execs(int)'
collect2: error: ld returned 1 exit status
make: *** [Makefile:216: MyApp] Error 1
make: *** Waiting for unfinished jobs....
```
The problem is that unlike shared libs, the order of static libs on the command line matters. I had to specify another `-lginkgo_device` at the end to fix this.

yhmtsai:
This PR create the test and fix for PackageConfig
The error is mentioned in #1104.
close #1104 

I add the existed lib into the list, and remove the duplication in the reversed to keep the last not first library in the list.
To keep the ordering of the libraries, this pr can not support circular dependence.
Also, pkgconfig of ginkgo isn't supported in MSVC.

From CMake PkgConfig TARGET, they only put the `-framework`, `-L`, `-l` to link_libraries but others in link_options.
Thus, the linking library with absolute path will be in link_option and cmake will put it before the compiled target such that it destroy the lib order. `g++ <link_options> *.o <link_libraries>` Ref: https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/FindPkgConfig.cmake#L251


test/test_exportbuild_pkgconfig/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+target_link_libraries(test_exportbuild_pkgconfig PRIVATE ${GINKGO_LDFLAGS})
+target_include_directories(test_exportbuild_pkgconfig PRIVATE ${GINKGO_INCLUDE_DIRS})
+target_compile_features(test_exportbuild_pkgconfig PUBLIC cxx_std_14)
+target_link_libraries(test_exportbuild_pkgconfig PRIVATE ${GINKGO_CFLAGS})
+
+# pkg-config only takes care of ginkgo
+if(GINKGO_HAVE_PAPI_SDE)
+    message("Link to PAPI")
+    target_link_libraries(test_exportbuild_pkgconfig PRIVATE PAPI::PAPI)
+endif()
```
upsj, line 27: nit:
```suggestion
endif()

```

CMakeLists.txt
```diff
@@ -336,15 +336,19 @@ set(GINKGO_TEST_INSTALL_SRC_DIR "${Ginkgo_SOURCE_DIR}/test/test_install/")
 set(GINKGO_TEST_INSTALL_BIN_DIR "${Ginkgo_BINARY_DIR}/test/test_install/")
 set(GINKGO_TEST_EXPORTBUILD_SRC_DIR "${Ginkgo_SOURCE_DIR}/test/test_exportbuild/")
 set(GINKGO_TEST_EXPORTBUILD_BIN_DIR "${Ginkgo_BINARY_DIR}/test/test_exportbuild/")
+set(GINKGO_TEST_EXPORTBUILD_PKGCONFIG_SRC_DIR "${Ginkgo_SOURCE_DIR}/test/test_exportbuild_pkgconfig/")
+set(GINKGO_TEST_EXPORTBUILD_PKGCONFIG_BIN_DIR "${Ginkgo_BINARY_DIR}/test/test_exportbuild_pkgconfig/")
 if(MSVC)
```
upsj, line 341: nit: This should probably use [GENERATOR_IS_MULTI_CONFIG](https://cmake.org/cmake/help/v3.24/prop_gbl/GENERATOR_IS_MULTI_CONFIG.html) instead of only supporting MSVC

test/test_exportbuild_pkgconfig/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
+cmake_minimum_required(VERSION 3.9)
+project(GinkgoExportBuildWithPkgConfigTest LANGUAGES CXX)
+
+find_package(PkgConfig REQUIRED)
+# find_package(Ginkgo REQUIRED)
+pkg_check_modules(GINKGO REQUIRED ginkgo)
```
upsj, line 6:
```suggestion
pkg_check_modules(GINKGO REQUIRED IMPORTED_TARGET ginkgo)
```

test/test_exportbuild_pkgconfig/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+message("GINKGO_LIBRARIES ${GINKGO_LIBRARIES}")
+message("GINKGO_INCLUDE_DIRS ${GINKGO_INCLUDE_DIRS}")
+message("GINKGO_CFLAGS ${GINKGO_CFLAGS}")
+message("GINKGO_LDFLAGS ${GINKGO_LDFLAGS}")
+add_executable(test_exportbuild_pkgconfig ../test_install/test_install.cpp)
+target_link_libraries(test_exportbuild_pkgconfig PRIVATE ${GINKGO_LIBRARIES})
+target_link_libraries(test_exportbuild_pkgconfig PRIVATE ${GINKGO_LDFLAGS})
+target_include_directories(test_exportbuild_pkgconfig PRIVATE ${GINKGO_INCLUDE_DIRS})
+target_compile_features(test_exportbuild_pkgconfig PUBLIC cxx_std_14)
+target_link_libraries(test_exportbuild_pkgconfig PRIVATE ${GINKGO_CFLAGS})
```
upsj, line 21: the pkgconfig support already provides CMake target support. We should probably also add the `-std=c++14` flag to `cflags`
```suggestion
target_link_libraries(test_exportbuild_pkgconfig PRIVATE PkgConfig::GINKGO)
```

test/test_exportbuild_pkgconfig/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+pkg_check_modules(GINKGO REQUIRED ginkgo)
+
+# Here, we use test install without any data. We instantiate the
+# interface only.
+message("GINKGOCMAKE ${CMAKE_APPBUNDLE_PATH} ${CMAKE_PREFIX_PATH} ${CMAKE_FRAMEWORK_PATH}")
+message("GINKGO ${GINKGO_PREFIX}")
+message("GINKGO_LIBRARIES ${GINKGO_LIBRARIES}")
+message("GINKGO_INCLUDE_DIRS ${GINKGO_INCLUDE_DIRS}")
+message("GINKGO_CFLAGS ${GINKGO_CFLAGS}")
+message("GINKGO_LDFLAGS ${GINKGO_LDFLAGS}")
```
upsj, line 15: Is this debugging output or necessary?

cmake/information_helpers.cmake
```diff
@@ -10,6 +10,8 @@ function(filter_generator_expressions INPUT OUTPUT)
     string(REGEX REPLACE "[$<A-Z_:]*SHELL:(.+)>*" "\\1" TMP "${TMP}")
     string(REGEX REPLACE ".+INTERFACE:.+>" "" TMP "${TMP}")
     string(REGEX REPLACE "\$<COMMA>" "," TMP "${TMP}")
+    # Remove the left : or > 
+    string(REGEX REPLACE ":|>" "" TMP "${TMP}")
```
upsj, line 14: I don't like the amount of complexity we're collecting here. I think we should consider using CMake's features to evaluate the generator expressions if possible.

upsj (commented):
Partial review, didn't look at the algorithm yet

cmake/information_helpers.cmake
upsj: Could you check if `file(GENERATE ...)` already evaluates the expressions? It should according to [the documentation](https://cmake.org/cmake/help/latest/command/file.html#generate), but some expressions have scopes which may not be available in this step.

test/test_exportbuild_pkgconfig/CMakeLists.txt
yhmtsai: I think using package config is more close to the original way not the CMake way

test/test_exportbuild_pkgconfig/CMakeLists.txt
yhmtsai: yes debugging

cmake/information_helpers.cmake
yhmtsai: I already checked in the beginning. It requires CMake 3.15 to support TARGET. `file(GENERATE ... TARGET ginkgo)` uses the property of ginkgo to generate the file. Moreover, it generates the file what we want but it throw the error about multiple modification are different in the file. I do not dig into the issue yet

upsj: Seems to be even 3.19, so this would be a pretty big jump. I guess we can deal with the current approach for a while longer, maybe add a TODO to remove this once we bump the version requirements

test/test_exportbuild_pkgconfig/CMakeLists.txt
upsj: I would prefer the CMake approach, since that way we can't accidentally miss any part of the pkg-config settings.

yhmtsai: Although using the target approach make make adds `-pthread` to compile flag, it changes the linking order to lead failure.
```
/usr/bin/c++ /usr/lib/libhwloc.so /usr/lib/libpapi.so -Wl,-rpath -Wl,/usr/lib -Wl,--enable-new-dtags -pthread \
/usr/lib/libmpi.so CMakeFiles/test_exportbuild_pkgconfig.dir/root/ginkgo/test/test_install/test_install.cpp.o -o \
test_exportbuild_pkgconfig  -Wl,-rpath,/usr/local/lib /usr/local/lib/libginkgo.so /usr/local/lib/libginkgo_omp.so \
/usr/local/lib/libginkgo_cuda.so /usr/local/lib/libginkgo_reference.so /usr/local/lib/libginkgo_hip.so \
/usr/local/lib/libginkgo_dpcpp.so /usr/local/lib/libginkgo_device.so 
```
leads `/usr/bin/ld: CMakeFiles/test_exportbuild_pkgconfig.dir/root/ginkgo/test/test_install/test_install.cpp.o: undefined reference to symbol 'papi_sde_init'`

ginkgo.pc
```
-L/usr/local/lib -lginkgo /usr/lib/libhwloc.so -lginkgo_omp -lginkgo_cuda -lginkgo_reference -lginkgo_hip \
-lginkgo_dpcpp -lginkgo_device /usr/lib/libpapi.so -Wl,-rpath -Wl,/usr/lib -Wl,--enable-new-dtags -pthread \
/usr/lib/libmpi.so
```
it seems to put the lib with absolute path like hwloc, papi, mpi before the object file.
However, they should be after object file.
Do you know any documentation or keyword for this behavior?

thoasm (approved):
LGTM!

CMakeLists.txt
```diff
@@ -336,21 +343,19 @@ set(GINKGO_TEST_INSTALL_SRC_DIR "${Ginkgo_SOURCE_DIR}/test/test_install/")
...
-    set(GINKGO_TEST_INSTALL_CMD ${GINKGO_TEST_INSTALL_BIN_DIR}/test_install)
-    set(GINKGO_TEST_EXPORTBUILD_CMD ${GINKGO_TEST_EXPORTBUILD_BIN_DIR}/test_exportbuild)
-    if(GINKGO_BUILD_CUDA)
-        set(GINKGO_TEST_INSTALL_CUDA_CMD ${GINKGO_TEST_INSTALL_BIN_DIR}/test_install_cuda)
-    endif()
+set(GINKGO_TEST_EXPORTBUILD_PKGCONFIG_SRC_DIR "${Ginkgo_SOURCE_DIR}/test/test_exportbuild_pkgconfig/")
+set(GINKGO_TEST_EXPORTBUILD_PKGCONFIG_BIN_DIR "${Ginkgo_BINARY_DIR}/test/test_exportbuild_pkgconfig/")
+get_property(GINKGO_USE_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
+# GINKGO_CONFIG_PREFIX contains / in the end.
+set(GINKGO_CONFIG_PREFIX "$<$<BOOL:${GINKGO_USE_MULTI_CONFIG}>:$<CONFIG>/>")
```
upsj, line 350: It's not entirely clean, but this is a really nice way to handle both types of generators.

CMakeLists.txt
```diff
@@ -320,6 +320,13 @@ if(GINKGO_BUILD_DOC)
     add_subdirectory(doc)
 endif()
 
+
+# add escape character '\' and convert the list to string
+string(REPLACE " " "\ " PKG_CMAKE_INSTALL_PREFIX "${CMAKE_INSTALL_PREFIX}")
+list(TRANSFORM GINKGO_INTERFACE_LINK_FLAGS REPLACE " " "\\\\ ")
+list(TRANSFORM GINKGO_INTERFACE_CXX_FLAGS REPLACE " " "\\\\ ")
+string(REPLACE ";" " " GINKGO_INTERFACE_CXX_FLAGS "${GINKGO_INTERFACE_CXX_FLAGS}")
+string(REPLACE ";" " " GINKGO_INTERFACE_LINK_FLAGS "${GINKGO_INTERFACE_LINK_FLAGS}")
```
upsj, line 329: Can you please elaborate a bit more on what this transformation is intended to do? Also it's not entirely obvious why it's necessary.

cmake/information_helpers.cmake
```diff
@@ -96,19 +96,20 @@ macro(ginkgo_interface_information)
     ginkgo_interface_libraries_recursively("${GINKGO_INTERFACE_LINK_LIBRARIES}")
 
     # Format and store the interface libraries found
+    # remove duplicates on the reversed list to keep the dependecy in the end of list.
```
upsj, line 99: nice 😄

cmake/information_helpers.cmake
```diff
@@ -10,67 +12,65 @@ function(filter_generator_expressions INPUT OUTPUT)
...
-                # Get the link flags and treat them
-                get_target_property(GINKGO_INTERFACE_LIBS_LINK_FLAGS "${_libs}"
-                    INTERFACE_LINK_OPTIONS)
-                if (GINKGO_INTERFACE_LIBS_LINK_FLAGS)
-                    filter_generator_expressions("${GINKGO_INTERFACE_LIBS_LINK_FLAGS}"
-                        GINKGO_INTERFACE_LIB_NAME)
-                endif()
-                if (NOT "${GINKGO_INTERFACE_LIB_NAME}" IN_LIST GINKGO_INTERFACE_LIBS_FOUND)
-                    list(APPEND GINKGO_INTERFACE_LIBS_FOUND "${GINKGO_INTERFACE_LIB_NAME}")
+                list(APPEND GINKGO_INTERFACE_LIBS_FOUND "${GINKGO_INTERFACE_LIB_NAME}")
```
upsj, line 32: can we do this with a generator expression? `$<$<CONFIG:Debug>:${CMAKE_DEBUG_POSTFIX}>`
This would also make things work in multi-config generators

upsj (commented):
I'm trying to figure out how to deal correctly with pkg-config, but it might be best not to use the IMPORTED_TARGET approach, otherwise things might get reordered arbitrarily, as you said.

yhmtsai:
Using `IMPORTED_TARGET` can handle the pthread in compile and linking.
if using the original way, I can not put CFLAGS into compile_options directly.
that's why I still change the property of target not use the original plain link/compile

upsj:
Makes sense, then let's leave it how it is

CMakeLists.txt
yhmtsai: the first three commands change the space ` ` to `\ `. string replace is plain text replacement such that it can use `\ `
However, list transform replace is regex replacement such that I need to put double escape. (`\\\\ `(escape \ for regex) -> `\\ ` (escape in the replacement) -> `\ ` in the final stage)
the last two are moved from inside to here.
I will also add the description into the file later

upsj: So you escape spaces instead of putting them in quotes?

yhmtsai: yes

upsj: and why not leave them in quotes instead? 😄

test/test_exportbuild_pkgconfig/CMakeLists.txt
upsj: I think the correct way to do things would be to output `-L/usr/lib -lhwloc -lpapi -lmpi`. Since pkg-config is Linux-only, we can rely on libraries being prefixed `lib` with a `.so` or `.a` suffix. I can't find any pkg-config file on my system that uses absolute paths

upsj:
I attempted to get this right, since there are at least two bugs in the current implementation:
1. the link line should be `-Lpath -lname`
2. compile definitions do not need to have a leading `-D`, so we may need to add it.

Finally, more recent CMake versions add a `$<HOST_LINK:...>` generator expression to the MPI link flags, which honestly just makes me want to give up on this entirely.

The cleanest thing I could come up with is https://github.com/ginkgo-project/ginkgo/commit/a59cc01ecdade47239ef2b0347d8a1684518267c, which includes a bit of cleanup

test/test_exportbuild_pkgconfig/CMakeLists.txt
yhmtsai: I do not think it is the requirement from the png-config. cmake also change -L -l to the absolute path in the end. pkg-config may prefer we have hwloc.pc, papi.pc, mpi.pc, and any pkg-config file from the outside or third-party package. pkg-config is also supported on Windows, but I can not make it work in this pr

yhmtsai:
we can not give it up because at least one user is using it.
If the absolute link path is not correct, we should fix it in our own target. If the target is made from the package, we should keep it.
When is the compile definition without `-D`? Is it from cmake?

test/test_exportbuild_pkgconfig/CMakeLists.txt
upsj: pkg-config has few explicit limitations, but on my entire system, I couldn't find a pkg-config file that uses absolute library paths directly, so we are definitely doing something unusual here. CMake's pkg-config support relies on that implicit assumption, treating every -l and -L parameter as library information and all other parameters as link options. This causes the order to be lost.
On the Windows side: We can use `CMAKE_SHARED_LIBRARY_SUFFIX` and friends to make it portable.

CMakeLists.txt
```diff
@@ -361,21 +370,19 @@ set(GINKGO_TEST_INSTALL_SRC_DIR "${Ginkgo_SOURCE_DIR}/test/test_install/")
...
-        set(GINKGO_TEST_INSTALL_CUDA_CMD ${GINKGO_TEST_INSTALL_BIN_DIR}/$<CONFIG>/test_install_cuda)
-    endif()
-else()
-    set(GINKGO_TEST_INSTALL_CMD ${GINKGO_TEST_INSTALL_BIN_DIR}/test_install)
-    set(GINKGO_TEST_EXPORTBUILD_CMD ${GINKGO_TEST_EXPORTBUILD_BIN_DIR}/test_exportbuild)
-    if(GINKGO_BUILD_CUDA)
-        set(GINKGO_TEST_INSTALL_CUDA_CMD ${GINKGO_TEST_INSTALL_BIN_DIR}/test_install_cuda)
-    endif()
+set(GINKGO_TEST_EXPORTBUILD_PKGCONFIG_SRC_DIR "${Ginkgo_SOURCE_DIR}/test/test_exportbuild_pkgconfig/")
+set(GINKGO_TEST_EXPORTBUILD_PKGCONFIG_BIN_DIR "${Ginkgo_BINARY_DIR}/test/test_exportbuild_pkgconfig/")
```
tcojean, line 374: This is not exportbuild anymore which is a specific CMake thing. Maybe simply name all this `TEST_PKGCONFIG`?

tcojean (approved):
LGTM. The new logic here seems good to me. I think getting a generic solution is out of scope for this so it's OK that it cannot deal with circular deps or MSVC. We only want to provide something reasonable that the user can work from if they really have to.

This at least seems to be working with OpenCARP which relies on pkg-config.

PR #1110, opened 2022-09-01
Add dpcpp device timing

yhmtsai:
This PR adds the ability to enable the profiling of the queue and allow use the device timer in benchmark.

We only use forward declaration in executor.hpp, so I can not use the property as default input directly, which leads the incomplete type error. I use the enum and convert it into the property type when create the queue. Any idea on this is welcome.

Another is to add the device timing to benchmark. I add two dummy kernels as the marker. the timing is the interval between end time of start event and start time of end event.


dpcpp/base/executor.dp.cpp
```diff
@@ -234,10 +235,24 @@ void delete_queue(sycl::queue* queue)
 }
 
 
+::cl::sycl::property_list get_property_list(dpcpp_queue_property property)
+{
+    if (property == dpcpp_queue_property::in_order) {
+        return {sycl::property::queue::in_order{}};
+    } else if (property == dpcpp_queue_property::enable_profiling) {
+        GKO_NOT_SUPPORTED(property);
+        return {sycl::property::queue::enable_profiling{}};
```
upsj, line 244: why return something if you fail in the line before?

yhmtsai: this is to avoid the warning about the condition return nothing matching the return type

upsj: the compiler should already figure out that this code is unreachable: https://godbolt.org/z/TMnbKPbY4

yhmtsai: I see. thanks for the example.

benchmark/utils/general.hpp
```diff
@@ -290,30 +290,35 @@ void backup_results(rapidjson::Document& results)
 
 
 // executor mapping
-const std::map<std::string, std::function<std::shared_ptr<gko::Executor>()>>
+const std::map<std::string, std::function<std::shared_ptr<gko::Executor>(bool)>>
```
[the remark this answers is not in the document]
yhmtsai: I think it will require the `timer.hpp` include the `general.hpp` or include `general.hpp` earlier than `timer.hpp`.
it only uses one input so I prefer not introducing another headers hierarchy.

thoasm (commented):
I have one concern about the use of `enum class`, other than that, it looks good.

include/ginkgo/core/base/executor.hpp
```diff
@@ -104,12 +104,37 @@ constexpr allocation_mode default_hip_alloc_mode =
...
+     */
+    enable_profiling = 2
+};
+
+GKO_ATTRIBUTES GKO_INLINE dpcpp_queue_property operator|(dpcpp_queue_property a,
+                                                         dpcpp_queue_property b)
+{
+    return static_cast<dpcpp_queue_property>(static_cast<int>(a) |
+                                             static_cast<int>(b));
+}
```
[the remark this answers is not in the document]
yhmtsai: I add one enum item for 1 | 2. I do not like this way because it is hard to extend for more options by providing all possible combination. Do you have any idea for it?
I only think store it in integer and provide the operation/assignment between enum and integer such that user can use enum to select the property.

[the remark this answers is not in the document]
upsj: IMO using enums for flags is well-established, so I'd be fine with keeping just the two entries.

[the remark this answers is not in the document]
thoasm: I had to do a lot of reading to understand what the standard says, as [cppreference](https://en.cppreference.com/w/cpp/language/enum) was IMO not clear about this.
According to C++14 ([here is the open draft](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3797.pdf)) [dcl.enum] p8:
> For an enumeration whose underlying type is fixed, the values of the enumeration are the values of the
underlying type.

It is legal for an enum type to have a value that is not part of the enumeration (which I didn't know).
That means, your original implementation was fine and I was wrong, sorry for that.

However, as a small nit, I would like to have a comment for the enum definition itself that specifies that the enumerators are allowed to be used in combination by performing a bitwise or (`|`).

include/ginkgo/core/base/executor.hpp
```diff
@@ -104,12 +104,42 @@ constexpr allocation_mode default_hip_alloc_mode =
...
 class queue;
 
+
 }  // namespace sycl
 }  // namespace cl
 
 
+/**
+ * The enum class for the dpcpp queue property.
+ */
```
thoasm, line 117: I would like to have a comment that it is possible (and legal) to combine multiple properties with a binary or (`|`).

include/ginkgo/core/base/executor.hpp
```diff
@@ -104,12 +104,42 @@ constexpr allocation_mode default_hip_alloc_mode =
...
+
+    /**
+     * queue enables the profiling
+     */
+    enable_profiling = 2,
+
+    /**
+     * queue enables the profiling and executes in order
+     */
+    in_order_and_enabling_profiling = 3
```
thoasm, line 132: This can be removed (see my previous reply: https://github.com/ginkgo-project/ginkgo/pull/1110#discussion_r989077382) since it is legal for an `enum class` to have values other than specified in the enumerator-list (as long as it fits in the underlying type).

thoasm (approved):
LGTM, sorry for the confusion I caused with the `enum class` (I learned something new during this review).

include/ginkgo/core/base/executor.hpp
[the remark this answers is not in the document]
yhmtsai: thanks!

PR #1116, opened 2022-09-10
Make IDR random init deterministic

### LINKED ISSUE #1024: Make IDR deterministic random initialization equal between devices
In the tests for #973, I noticed that the IDR implementation uses different approaches to initialize the random matrix across executors. To improve testability, we should initialize the data in core.

upsj:
Fixes #1024


tcojean (approved):
LGTM

Issue #286, opened 2019-04-02
Current NOT_IMPLEMENTED kernels


thoasm:
We should discuss what kernels we should implement before the release. Since I lost the overview, I looked in the code and searched for kernels which are marked as `GKO_NOT_IMPLEMENTED`.

Interestingly, we do not support `Coo::transpose` and `Coo::conj_transpose` anywhere, which should be fairly trivial to implement (swapping `col_idxs_` and `row_idxs_` with `std::move`, followed by complex conjugation for `conj_transpose`).

The big question is: what should we implement before the first release?

The following are lists of `GKO_NOT_IMPLEMENTED` kernels (which are actually not implemented and not just have a case which is not implemented):
### CUDA
- Coo
  - [ ] `transpose`
  - [ ] `conj_transpose`
- Csr
  - [ ] `move_to_dense` _(can be replaced with `convert_to_dense`)_
- Dense
  - [ ] `move_to_csr` _(can be replaced with `convert_to_csr`)_
  - [ ] `move_to_ell` _(can be replaced with `convert_to_ell`)_
  - [ ] `convert_to_hybrid`
  - [ ] `move_to_hybrid` _(can be replaced with `convert_to_hybrid`)_
  - [ ] `move_to_sellp`  _(can be replaced with `convert_to_sellp`)_
- Hybrid
  - [ ] `convert_to_dense`

__NOTE:__ All `move_to_XXX` kernels in `Dense` can easily be removed, and the functionality replaced by `convert_to_XXX` (which we should definitively do).
### Reference
- Coo
  - [ ] `transpose`
  - [ ] `conj_transpose`

### OMP
- Coo
  - [ ] `transpose`
  - [ ] `conj_transpose`
- Csr
  - [ ] `convert_to_ell` __after__ #250
  - [ ] `calculate_max_nnz_per_row` __after__ #250
  - [ ] `convert_to_sellp` __after__ #279
  - [ ] `calculate_total_cols` __after__ #279
- Ell
  - [ ] `convert_to_csr` __after__ #250
  - [ ] `count_nonzeros` __after__ #250
- Sellp
  - [ ] `convert_to_csr` __after__ #279
  - [ ] `count_nonzeros` __after__ #279
- Gmres
  - [ ] All Kernels (@fritzgoebel is working on this)

## Current conversions supported (maybe not fully implemented)

- __Coo__
  -> Csr
  -> Dense
- __Csr__
  -> Coo
  -> Dense
[...]

PR #1117, opened 2022-09-12
Unified tests

### LINKED ISSUE #483: Unifying kernel tests

upsj:
This is a big one 😀 Most of our tests are 90% - 100% identical between different executors. In terms of maintenance, I don't think there is value in keeping them separate. If there is specific behavior we need to test, we can+
* do that in an isolated executor-specific test
* #ifdef the test away for other executors or
* test it anyways, since more tests is more good (:

TODO:
- [x] Csr kernel tests

Depends on #1123
Closes #483


upsj:
format!

upsj:
format!

test/stop/criterion_kernels.cpp
```diff
@@ -30,49 +30,48 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ******************************<GINKGO LICENSE>*******************************/
 
-#include <ginkgo/core/stop/criterion.hpp>
```
yhmtsai, line 33: "core/stop/criterion_kernels.hpp"

test/stop/residual_norm_kernels.cpp
```diff
@@ -0,0 +1,783 @@
...
+    using NormVector = typename TestFixture::NormVector;
+    auto initial_res = gko::initialize<Mtx>({100.0}, this->exec);
+    std::shared_ptr<gko::LinOp> rhs = gko::initialize<Mtx>({10.0}, this->exec);
+    auto criterion =
+        this->factory->generate(nullptr, rhs, nullptr, initial_res.get());
+    auto rel_criterion =
+        this->rel_factory->generate(nullptr, rhs, nullptr, initial_res.get());
+    auto abs_criterion =
+        this->abs_factory->generate(nullptr, rhs, nullptr, initial_res.get());
+    {
```
yhmtsai, line 104: the following three scope does not share too much common factor, maybe separate them to three test?

test/utils/executor.hpp
```diff
@@ -33,56 +33,93 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-void init_executor(std::shared_ptr<gko::ReferenceExecutor> ref,
-                   std::shared_ptr<gko::CudaExecutor>& exec)
+inline std::shared_ptr<gko::CudaExecutor> init_executor(
+    std::shared_ptr<gko::ReferenceExecutor> ref, gko::CudaExecutor*)
 {
-    ASSERT_GT(gko::CudaExecutor::get_num_devices(), 0);
-    exec = gko::CudaExecutor::create(0, ref);
+    {
+        if (gko::CudaExecutor::get_num_devices() == 0) {
+            throw std::runtime_error{"No suitable HIP devices"};
```
yhmtsai, line 76:
```suggestion
            throw std::runtime_error{"No suitable CUDA devices"};
```

test/utils/executor.hpp
```diff
@@ -33,56 +33,93 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
 {
-    exec = gko::OmpExecutor::create();
+    return gko::OmpExecutor::create();
 }
 
 
-void init_executor(std::shared_ptr<gko::ReferenceExecutor> ref,
-                   std::shared_ptr<gko::CudaExecutor>& exec)
+inline std::shared_ptr<gko::CudaExecutor> init_executor(
+    std::shared_ptr<gko::ReferenceExecutor> ref, gko::CudaExecutor*)
```
yhmtsai, line 72: maybe use template type not the dummy pointer?

yhmtsai (commented):
thanks for this impressive work

core/test/utils.hpp
```diff
@@ -189,7 +189,8 @@ struct reduction_factor {
     using nc_output = remove_complex<OutputType>;
     using nc_precision = remove_complex<Precision>;
     static constexpr nc_output value{
-        std::numeric_limits<nc_precision>::epsilon() * nc_output{10}};
+        std::numeric_limits<nc_precision>::epsilon() * nc_output{10} *
+        (gko::is_complex<Precision>() ? nc_output{1.4142} : one<nc_output>())};
```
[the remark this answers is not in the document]
upsj: yes, though it was surprisingly complicated to get that value in there, so I hand-wrote it

omp/test/stop/residual_norm_kernels.cpp
```diff
@@ -53,24 +53,24 @@ class ResidualNorm : public ::testing::Test {
 
     ResidualNorm()
     {
-        exec_ = gko::OmpExecutor::create();
```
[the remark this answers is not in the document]
upsj: true, forgot to delete it, I had it there for reference

reference/test/stop/residual_norm_kernels.cpp
```diff
@@ -368,7 +368,7 @@ TYPED_TEST(ResidualNorm, SelfCalulatesAndWaitsTillResidualGoal)
         ASSERT_EQ(stop_status.get_data()[0].has_converged(), false);
         ASSERT_EQ(one_changed, false);
 
-        solution->at(0) = rhs_val - r<T>::value * T{0.9} * rhs_norm->at(0);
+        solution->at(0) = rhs_val - r<T>::value * T{0.5} * rhs_norm->at(0);
```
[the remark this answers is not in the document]
upsj: It's related to the sqrt(2) changes in `r::value`

test/components/prefix_sum_kernels.cpp
```diff
@@ -89,17 +82,17 @@ class PrefixSum : public ::testing::Test {
...
 
 
-TYPED_TEST(PrefixSum, TrivialCasesEqualReference)
+TYPED_TEST(PrefixSum, EqualsReference)
 {
-    this->test(0);
-    this->test(1);
+    using gko::size_type;
+    for (auto size :
+         {size_type{0}, size_type{1}, size_type{100}, this->total_size}) {
```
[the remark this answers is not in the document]
upsj: good idea!

test/factorization/par_ilu_kernels.cpp
```diff
@@ -0,0 +1,312 @@
...
+    std::default_random_engine rand_engine;
+    std::shared_ptr<const Csr> mtx;
+    std::shared_ptr<const Csr> dmtx;
+
+    ParIlu() : rand_engine(18)
+    {
+        std::string file_name(gko::matrices::location_ani4_mtx);
+        auto input_file = std::ifstream(file_name, std::ios::in);
+        auto mtx_temp = gko::read<Csr>(input_file, ref);
+        auto dmtx_temp = gko::clone(exec, mtx_temp);
```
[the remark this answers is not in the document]
upsj: done

test/matrix/csr_kernels2.cpp
```diff
@@ -79,41 +82,74 @@ class Csr : public ::testing::Test {
...
+    {
+#ifdef GKO_COMPILING_OMP
+        throw std::runtime_error{"We shouldn't be testing this"};
+#else
+        strategy = std::make_shared<typename Mtx::automatical>(exec);
+#endif
+    }
+
+    template <typename Mtx>
+    void set_up_strategy(std::shared_ptr<typename Mtx::sparselib>& strategy)
```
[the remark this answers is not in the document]
upsj: some strategies take a statically typed executor of restricted type, some don't take an executor at all, we need a generic way to set them up

test/matrix/csr_kernels2.cpp
```diff
@@ -79,41 +82,74 @@ class Csr : public ::testing::Test {
...
-    {
-        ASSERT_GT(gko::DpcppExecutor::get_num_devices("all"), 0);
-        ref = gko::ReferenceExecutor::create();
-        dpcpp = gko::DpcppExecutor::create(0, ref);
-    }
-
-    void TearDown()
+    template <typename MtxType>
+    std::unique_ptr<MtxType> gen_mtx(int num_rows, int num_cols,
+                                     int min_nnz_row, int max_nnz_row)
```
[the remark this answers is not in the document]
upsj: that's essentially what we're doing here, only you can't default a parameter to another value

test/matrix/csr_kernels2.cpp
```diff
@@ -964,4 +1199,55 @@ TEST_F(Csr, CreateSubMatrixIsEquivalentToRef)
 }
 
 
-}  // namespace
+#ifndef GKO_COMPILING_DPCPP
+
+
+TEST_F(Csr, CanDetectMissingDiagonalEntry)
```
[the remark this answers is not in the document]
upsj: I added them to https://github.com/ginkgo-project/ginkgo/issues/286

test/matrix/ell_kernels.cpp
```diff
@@ -450,7 +433,7 @@ TEST_F(Ell, AdvancedByAtomicApplyIsEquivalentToRef)
     mtx->apply(alpha.get(), y.get(), beta.get(), expected.get());
     dmtx->apply(dalpha.get(), dy.get(), dbeta.get(), dresult.get());
 
-    GKO_ASSERT_MTX_NEAR(dresult, expected, r<vtype>::value);
+    GKO_ASSERT_MTX_NEAR(dresult, expected, r<vtype>::value * 10);
```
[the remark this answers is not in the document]
upsj: I think it was CUDA, but don't quote me on that. With the atomics, the results are a bit flaky, so I wanted to add some slack

test/matrix/ell_kernels.cpp
```diff
@@ -432,14 +412,17 @@ TEST_F(Ell, MixedAdvancedApplyWithStrideToDenseMatrixIsEquivalentToRef3)
 }
 
 
+#ifndef GKO_COMPILING_OMP
```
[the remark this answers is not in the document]
upsj: good point, we can leave them in

test/matrix/fbcsr_kernels.cpp
```diff
@@ -181,19 +165,19 @@ TYPED_TEST(Fbcsr, SpmvIsEquivalentToRefSorted)
...
-    rand_cuda->copy_from(gko::lend(this->rsorted_ref));
-    auto x_ref = Dense::create(
-        this->ref, gko::dim<2>(this->rsorted_ref->get_size()[1], 1));
+    auto drand = Mtx::create(this->exec);
+    drand->copy_from(gko::lend(this->rsorted));
+    auto x_ref =
+        Dense::create(this->ref, gko::dim<2>(this->rsorted->get_size()[1], 1));
     this->generate_sin(x_ref.get());
-    auto x_cuda = Dense::create(this->cuda);
+    auto x_cuda = Dense::create(this->exec);
```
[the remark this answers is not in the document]
upsj: done

test/matrix/dense_kernels.cpp
```diff
@@ -135,17 +121,23 @@ class Dense : public ::testing::Test {
         dc_y = gko::clone(exec, c_y);
         dalpha = gko::clone(exec, alpha);
         dc_alpha = gko::clone(exec, c_alpha);
+        result = Mtx::create(ref, gko::dim<2>{1, num_vecs});
+        dresult = Mtx::create(exec, gko::dim<2>{1, num_vecs});
     }
 
     void set_up_apply_data()
     {
         x = gen_mtx<Mtx>(40, 25);
```
fritzgoebel, line 130: That's the size from the OMP kernels. I'd suggest taking the bigger size from the other executors to include at least two warps everywhere

test/preconditioner/jacobi_kernels.cpp
```diff
@@ -323,20 +307,23 @@ TEST_F(Jacobi, HipPreconditionerEquivalentToRefWithBlockSize32Unsorted)
 }
 
 
-#if GINKGO_HIP_PLATFORM_HCC
-TEST_F(Jacobi, HipPreconditionerEquivalentToRefWithBlockSize64)
+#ifdef GKO_COMPILING_HIP
```
pratikvn, line 310: Do we not test hip on CUDA GPUs anymore ?

test/matrix/csr_kernels2.cpp
```diff
@@ -240,9 +275,35 @@ TEST_F(Csr, SimpleApplyIsEquivalentToRefWithOnemkl)
 }
 
 
-TEST_F(Csr, SimpleApplyIsEquivalentToRefWithOnemklUnsorted)
```
pratikvn, line 243: Isnt this this specfic to DPCPP ?

test/matrix/csr_kernels2.cpp
```diff
@@ -438,16 +519,77 @@ TEST_F(Csr, SimpleApplyToCsrMatrixIsEquivalentToRef)
...
+    mtx1->apply(mtx2.get(), square_mtx.get());
+    dmtx1->apply(dmtx2.get(), square_dmtx.get());
+
+    GKO_ASSERT_MTX_EQ_SPARSITY(square_dmtx, square_mtx);
+    GKO_ASSERT_MTX_NEAR(square_dmtx, square_mtx, 1e-14);
+    ASSERT_TRUE(square_dmtx->is_sorted_by_column_index());
+}
+
+
+// TODO: broken in ROCm <= 4.5
```
pratikvn, line 559: That is due to a hipsparse bug ? And in case it works in 5.x, then maybe only disable it for 4.5 and not for all HIP versions ?

pratikvn (commented):
Really nice work! A few comments and questions.

test/preconditioner/jacobi_kernels.cpp
upsj: Blocksize 64 is only enabled for ROCm, we check the warp size in the test body

test/matrix/csr_kernels2.cpp
upsj: We test the behavior in SimpleApplyIsEquivalentToRefWithSparselibUnsorted

test/matrix/csr_kernels2.cpp
upsj: Yes, I'll add a workaround later, but we didn't have this test for HIP before, so it's not a regression

test/solver/cb_gmres_kernels.cpp
```diff
@@ -259,15 +236,15 @@ class CbGmres : public ::testing::Test {
 };
 
 
-TEST_F(CbGmres, DpcppCbGmresInitialize1IsEquivalentToRef)
+TEST_F(CbGmres, CbGmresInitialize1IsEquivalentToRef)
```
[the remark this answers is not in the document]
upsj: The executor is always visible from context, e.g. ctest output or which binary you launched. Most tests don't have this prefix

fritzgoebel (commented):
Thanks for tackling this!
It is a lot to go through so I'd suggest a few of us do it.

reference/test/stop/residual_norm_kernels.cpp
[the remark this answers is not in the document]
upsj: no, because I relaxed the complex error bounds to include a sqrt(2) factor

test/solver/idr_kernels.cpp
```diff
@@ -42,46 +43,44 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/base/array.hpp>
 #include <ginkgo/core/base/exception.hpp>
 #include <ginkgo/core/base/executor.hpp>
+#include <ginkgo/core/base/mtx_io.hpp>
+#include <ginkgo/core/matrix/csr.hpp>
 #include <ginkgo/core/matrix/dense.hpp>
+#include <ginkgo/core/solver/bicgstab.hpp>
+#include <ginkgo/core/solver/gmres.hpp>
```
MarcelKoch, line 50: unused
```suggestion
```

test/solver/multigrid_kernels.cpp
```diff
@@ -42,36 +42,21 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/base/exception.hpp>
 #include <ginkgo/core/base/executor.hpp>
 #include <ginkgo/core/matrix/dense.hpp>
+#include <ginkgo/core/solver/multigrid.hpp>
```
MarcelKoch, line 45: surprisingly(?) unused
```suggestion
```

test/stop/criterion_kernels.cpp
```diff
@@ -30,49 +30,48 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
...
-
 #include <gtest/gtest.h>
 
 
+#include <ginkgo/core/stop/criterion.hpp>
 #include <ginkgo/core/stop/iteration.hpp>
 
 
-namespace {
+#include "core/test/utils.hpp"
```
MarcelKoch, line 40: unused
```suggestion
```

MarcelKoch (approved):
looks very good. I've left some smaller remarks. I guess we need to see if the coverage changes due to this.
I would also suggest getting more than 2 approvals, because a single person won't find all issues that might be hiding somewhere.

test/factorization/par_ilut_kernels.cpp
```diff
@@ -55,31 +55,29 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
 protected:
     using value_type =
         typename std::tuple_element<0, decltype(ValueIndexType())>::type;
     using index_type =
         typename std::tuple_element<1, decltype(ValueIndexType())>::type;
     using Dense = gko::matrix::Dense<value_type>;
+    using ComplexDense = gko::matrix::Dense<std::complex<value_type>>;
     using Coo = gko::matrix::Coo<value_type, index_type>;
     using Csr = gko::matrix::Csr<value_type, index_type>;
+    using ComplexCsr = gko::matrix::Csr<std::complex<value_type>, index_type>;
```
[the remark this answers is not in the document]
upsj: true, this is no longer necessary, with the tests being fully templated now

core/test/utils.hpp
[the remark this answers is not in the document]
upsj: nice suggestion, thanks!

test/components/prefix_sum_kernels.cpp
```diff
@@ -44,23 +44,17 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/base/array.hpp>
 
 
-#include "core/test/utils.hpp"
-
-
-namespace {
+#include "cuda/test/utils.hpp"
```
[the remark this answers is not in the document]
upsj: no, that was leftover from the original :)

test/matrix/csr_kernels2.cpp
```diff
@@ -928,32 +1078,124 @@ TEST_F(Csr, ComputeSubmatrixIsEquivalentToRef)
...
+    gko::index_set<IndexType> drset(this->exec, rset);
+    gko::index_set<IndexType> dcset(this->exec, cset);
+    auto size = this->mtx2->get_size();
+    auto row_nnz = gko::array<int>(this->ref, rset.get_num_elems() + 1);
+    row_nnz.fill(gko::zero<int>());
+    auto drow_nnz = gko::array<int>(this->exec, row_nnz);
+
+    gko::kernels::reference::csr::calculate_nonzeros_per_row_in_index_set(
+        this->ref, this->mtx2.get(), rset, cset, row_nnz.get_data());
+    gko::kernels::omp::csr::calculate_nonzeros_per_row_in_index_set(
```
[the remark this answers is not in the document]
upsj: great catch!

test/stop/criterion_kernels.cpp
upsj: curious how you caught that one. Did you use a tool? It's definitely not obvious!

test/matrix/csr_kernels2.cpp
```diff
@@ -408,9 +459,39 @@ TEST_F(Csr, AdvancedApplyToDenseMatrixIsEquivalentToRefWithMergePath)
 }
 
 
+TEST_F(Csr, OneAutomaticalWorksWithDifferentMatrices)
+{
+    auto automatical = std::make_shared<Mtx::automatical>(exec);
+#ifdef GKO_COMPILING_CUDA
```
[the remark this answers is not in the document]
upsj: We're not very consistent about this surprisingly, but most cases use `#ifdef`, not `#if defined`:

```bash
grep -rh "#if" test | sort | uniq -c
      2 #if !(GINKGO_COMMON_SINGLE_MODE)
     25 #if GINKGO_COMMON_SINGLE_MODE
      1 #if GKO_HAVE_PAPI_SDE
      3 #if HAS_REFERENCE
      2 #if defined(GKO_COMPILING_CUDA) || defined(GKO_COMPILING_HIP) || \
      1 #if defined(GKO_COMPILING_OMP) || defined(GKO_COMPILING_CUDA) || \
      1 #if defined(HAS_CUDA)
      1 #if defined(HAS_CUDA) || defined(HAS_HIP)
      2 #if defined(HAS_HIP) || defined(HAS_CUDA)
      6 #ifdef GINKGO_FAST_TESTS
      6 #ifdef GKO_COMPILING_CUDA
      1 #ifdef GKO_COMPILING_DPCPP
      6 #ifdef GKO_COMPILING_HIP
      4 #ifdef GKO_COMPILING_OMP
      2 #ifndef GKO_COMPILING_DPCPP
      2 #ifndef GKO_COMPILING_HIP
      1 #ifndef GKO_COMPILING_OMP
      1 #ifndef GKO_TEST_UTILS_EXECUTOR_HPP_
```

test/components/absolute_array_kernels.cpp
```diff
@@ -48,43 +48,27 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-
-
-class AbsoluteArray : public ::testing::Test {
+class AbsoluteArray : public CommonTestFixture {
 protected:
 #if GINKGO_COMMON_SINGLE_MODE
     using value_type = float;
 #else
     using value_type = double;
 #endif
```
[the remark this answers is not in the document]
upsj: good point!

test/stop/criterion_kernels.cpp
MarcelKoch: I just trust clion on these kinds of things

core/test/utils.hpp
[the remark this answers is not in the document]
upsj: unfortunately this doesn't work, because it needs to be constexpr

test/preconditioner/isai_kernels.cpp
```diff
@@ -476,180 +464,177 @@ TEST_F(Isai, HipIsaiGeneratePartialExcessIsEquivalentToRef)
...
+TEST_F(Isai, IsaiScatterExcessSolutionSpdIsEquivalentToRef)
 {
     initialize_data(matrix_type::spd, 100, 64);
     const auto num_rows = mtx->get_size()[0];
     gko::array<index_type> a1(ref, num_rows + 1);
     auto a2 = a1;
     gko::kernels::reference::isai::generate_general_inverse(
-        ref, mtx.get(), inverse.get(), a1.get_data(), a2.get_data(), true);
-    gko::array<index_type> da1(hip, a1);
+        ref, mtx.get(), inverse.get(), a1.get_data(), a2.get_data(), false);
```
yhmtsai, line 598: the next excess solution also uses true. Should it be true or false?

test/preconditioner/isai_kernels.cpp
```diff
@@ -275,43 +263,43 @@ TEST_F(Isai, HipIsaiGenerateUinverseLongIsEquivalentToRef)
...
         ref, mtx.get(), inverse.get(), a1.get_data(), a2.get_data(), false);
-    gko::kernels::hip::isai::generate_general_inverse(
-        hip, d_mtx.get(), d_inverse.get(), da1.get_data(), da2.get_data(),
+    gko::kernels::EXEC_NAMESPACE::isai::generate_general_inverse(
+        exec, d_mtx.get(), d_inverse.get(), da1.get_data(), da2.get_data(),
         false);
 
     GKO_ASSERT_MTX_EQ_SPARSITY(inverse, d_inverse);
-    GKO_ASSERT_MTX_NEAR(inverse, d_inverse, 10 * r<value_type>::value);
+    GKO_ASSERT_MTX_NEAR(inverse, d_inverse, 100 * r<value_type>::value);
```
yhmtsai, line 282: maybe need to be careful?

test/solver/gmres_kernels.cpp
```diff
@@ -43,46 +43,46 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-            ASSERT_NO_THROW(dpcpp->synchronize());
-        }
+        mtx = gen_mtx(123, 123);
+        d_mtx = gko::clone(exec, mtx);
+        exec_gmres_factory =
+            Solver::build()
+                .with_criteria(
+                    gko::stop::Iteration::build().with_max_iters(246u).on(exec),
+                    gko::stop::ResidualNorm<value_type>::build()
+                        .with_reduction_factor(value_type{1e-15})
```
yhmtsai, line 74: still surprised at float germs can use 1e-15 as reduction_factor?

yhmtsai (approved):
the excess system in Isai uses true/false in different tests.
I am not in favor of that the init_executor uses overload by passing nullptr not template to choose which implementation.

test/matrix/csr_kernels2.cpp
```diff
@@ -188,309 +218,416 @@ class Csr : public ::testing::Test {
...
     auto trans = mtx->transpose();
     auto d_trans = dmtx->transpose();
 
     mtx->apply(trans.get(), square_mtx.get());
     dmtx->apply(d_trans.get(), square_dmtx.get());
 
-    GKO_ASSERT_MTX_NEAR(square_dmtx, square_mtx, 1e-14);
+    GKO_ASSERT_MTX_NEAR(square_dmtx, square_mtx, r<value_type>::value);
+    GKO_ASSERT_MTX_EQ_SPARSITY(square_dmtx, square_mtx);
+    ASSERT_TRUE(square_dmtx->is_sorted_by_column_index());
```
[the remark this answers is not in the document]
upsj: all the assertions take the matrix by const pointer, so this shouldn't be an issue

test/preconditioner/isai_kernels.cpp
upsj: I like to increase the error bounds with big margins to make sure we don't have any flaky tests if some compiler optimizations change the evaluation order. Do you think this is an issue here?

test/preconditioner/isai_kernels.cpp
upsj: it should stay `true`, good catch!

fritzgoebel (approved):
LGTM!

upsj:
format!

test/solver/gmres_kernels.cpp
yhmtsai: yes, unit roundoff here. it's hard to reach that when in float.
I see, I miss the iteration criterion.

test/preconditioner/isai_kernels.cpp
yhmtsai: I do not know actually.
Without knowing the testing problem, I can somehow accept 10 r as rounding issue. 100 r will be a little higher such that I start wondering whether there's something wrong or the wrong codes may also pass the test

PR #1128, opened 2022-10-03
Preserve the executor in generation

yhmtsai:
This PR will preserve the executor of given LinOp in generation.
clone still moves everything to the same executor.

- [ ]: should the clone use is_accessible not the pointer?

### DIFF
--- a/include/ginkgo/core/solver/solver_base.hpp
+++ b/include/ginkgo/core/solver/solver_base.hpp
@@ -337,7 +337,11 @@ public:
     EnableSolverBase(std::shared_ptr<const MatrixType> system_matrix)
         : SolverBase<MatrixType>{self()->get_executor()}
     {
-        set_system_matrix(std::move(system_matrix));
+        // Use the given matrix to generate the solver.
+        if (system_matrix) {
+            GKO_ASSERT_IS_SQUARE_MATRIX(system_matrix);
+        }
+        this->set_system_matrix_base(system_matrix);
     }
 
     /**


upsj:
My suggestion at the time of the value semantics PR was making the cross-executor copy explicit by putting it into a PreserveExecutor LinOp(Factory) that also avoids the temporary allocation this approach would require at each cross-executor apply call.

yhmtsai:
I mean it introduces the copy in creation not during the copy. `.on` is already quite specific request for me.

upsj:
To recap my arguments from last time:
* Relying on the implicit temporary copy for a repeated `apply` comes with a performance penalty due to repeated allocations and deallocations
* `on` can't be a strong statement of intent if it is mandatory. I have some plans to provide functionality to avoid specifying the executor, because in some cases it even causes segfaults and the like (stopping criteria), but even then, avoiding the aforementioned penalties would require special-case code in each class made from a factory, instead of cleanly separating the responsibility of cross-executor copies into a single-purpose class.
* The same logic also goes for precision conversion boundaries: While having the temporary conversion makes sense for some cases, if you need good performance, I would like to provide a separate precision conversion class that stores the temporaries explicitly.

The suggested approach is a natural extension of the Reordered, ScaledReordered, ... design into other Ginkgo functionality (precisions and executors)

yhmtsai:
Some algorithm only work on the cpu then it will convert twice in apply.
Creating two executors and using different executor on different object is more hard than using one executor for everything.
When user create two executors, they know it may gives some penalty on copy or something else.
I feel it is a conflict design against the executors - we provide several kinds of executor then we only allow using one of them.

MarcelKoch:
closed due to conceptual changes. Now we actually enforce that objects are copied to the right executor.

PR #1136, opened 2022-10-05
Adds simple functions to create Dense view

MarcelKoch:
This PR adds functionality to create a view on an existing Dense vector in a simple manner. With this, a view can be created by
```
auto view = gko::make_dense_view(gko::lend(other));
auto view = Dense::create_view_of(gko::lend(other));
```
If `other` is derived from `Dense`, then `view` will have the same runtime type.

This will make it easier to create distributed vectors (#1133) from existing local vectors, while keeping the local vectors valid.

### DIFF
--- a/core/test/matrix/dense.cpp
+++ b/core/test/matrix/dense.cpp
@@ -412,6 +412,8 @@ TYPED_TEST(Dense, CanMakeMutableView)
     auto view = gko::make_dense_view(this->mtx.get());
 
     ASSERT_EQ(view->get_values(), this->mtx->get_values());
+    ASSERT_EQ(view->get_executor(), this->mtx->get_executor());
+    GKO_ASSERT_MTX_NEAR(view, this->mtx, 0.0);
 }
 
 
@@ -420,6 +422,8 @@ TYPED_TEST(Dense, CanMakeConstView)
     auto view = gko::make_const_dense_view(this->mtx.get());
 
     ASSERT_EQ(view->get_const_values(), this->mtx->get_const_values());
+    ASSERT_EQ(view->get_executor(), this->mtx->get_executor());
+    GKO_ASSERT_MTX_NEAR(view, this->mtx, 0.0);
 }
 
 
@@ -448,8 +452,7 @@ private:
     std::unique_ptr<gko::matrix::Dense<>> create_view_of_impl() override
     {
         auto view = create(this->get_executor(), {}, this->get_data());
-        (*static_cast<gko::matrix::Dense<>*>(view.get())) =
-            std::move(*gko::matrix::Dense<>::create_view_of_impl());
+        gko::matrix::Dense<>::create_view_of_impl()->move_to(this);
         return view;
     }
 
--- a/core/test/matrix/dense.cpp
+++ b/core/test/matrix/dense.cpp
@@ -452,7 +452,7 @@ private:
     std::unique_ptr<gko::matrix::Dense<>> create_view_of_impl() override
     {
         auto view = create(this->get_executor(), {}, this->get_data());
-        gko::matrix::Dense<>::create_view_of_impl()->move_to(this);
+        gko::matrix::Dense<>::create_view_of_impl()->move_to(view.get());
         return view;
     }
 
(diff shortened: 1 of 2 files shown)


core/test/matrix/dense.cpp
```diff
@@ -407,4 +407,65 @@ TYPED_TEST(Dense, CanCreateRealView)
...
+
+TEST(DenseView, ConstViewKeepsRuntimeType)
+{
+    auto vector = CustomDense::create(gko::ReferenceExecutor::create(),
+                                      gko::dim<2>{3, 4});
+
+    auto view = gko::make_const_dense_view(vector.get());
+
+    ASSERT_EQ(view->get_const_values(), vector->get_const_values());
+    ASSERT_TRUE(dynamic_cast<const CustomDense*>(view.get()));
```
yhmtsai, line 467: If CustomDense contains some additional data, does it keeps those data?
For example, CustomDense contains a int and set it to 5 (2 by default).
will the view use 5 or 2?

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -1176,6 +1235,34 @@ struct temporary_clone_helper<matrix::Dense<ValueType>> {
 }  // namespace detail
 
 
+/**
+ * Creates a view of a given Dense vector.
+ *
+ * @tparam ValueType  the underlying value type of the vector
+ * @param vector  the vector on which to create the view
```
yhmtsai, line 1242:
```suggestion
 * @tparam ValueType  the underlying value type of the vector
 * 
 * @param vector  the vector on which to create the view
```
nit

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -990,6 +1014,41 @@ class Dense
...
+     * @return  A immutable Dense matrix that is a view of this.
+     */
+    virtual std::unique_ptr<const Dense> create_const_view_of_impl() const
+    {
+        auto exec = this->get_executor();
+        auto view = this->create_default(exec);
+        view->set_size(this->get_size());
+        view->values_ = gko::make_array_view(
+            exec, this->get_num_stored_elements(),
+            const_cast<value_type*>(this->get_const_values()));
```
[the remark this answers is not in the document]
thoasm: I agree, this should use `make_const_array_view` to avoid the `const_cast`.

core/test/matrix/dense.cpp
```diff
@@ -407,4 +407,65 @@ TYPED_TEST(Dense, CanCreateRealView)
 }
 
 
+TYPED_TEST(Dense, CanMakeMutableView)
+{
+    auto view = gko::make_dense_view(this->mtx.get());
+
+    ASSERT_EQ(view->get_values(), this->mtx->get_values());
```
thoasm, line 414: You should test the executor, stride, and dimensions as well.

core/test/matrix/dense.cpp
```diff
@@ -407,4 +407,65 @@ TYPED_TEST(Dense, CanCreateRealView)
...
+}
+
+
+TYPED_TEST(Dense, CanMakeConstView)
+{
+    auto view = gko::make_const_dense_view(this->mtx.get());
+
+    ASSERT_EQ(view->get_const_values(), this->mtx->get_const_values());
```
thoasm, line 422: Same here (test for executor, stride and dimensions).

thoasm (approved):
LGTM!

core/test/matrix/dense.cpp
MarcelKoch: The view will not keep the data. I don't think there is a way to let it keep the data, without overriding `create_view_of_impl` by the user. I will remove that behavior and always just return a `Dense`.

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -207,6 +207,30 @@ class Dense
         return (*other).create_with_type_of_impl(exec, size, stride);
     }
 
+    /**
+     * Creates a Dense matrix, where the underlying array is a view of another
+     * Dense matrix' array.
+     *
+     * @param other  The other matrix on which to create the view
+     * @return  A Dense matrix that is a view of other
```
yhmtsai, line 215:
```suggestion
     * @param other  The other matrix on which to create the view
     * 
     * @return  A Dense matrix that is a view of other
```
nit

yhmtsai (approved):

MarcelKoch:
format-rebase!

PR #1139, opened 2022-10-12
Allow `rebase!` and `format-rebase!` to preserve merge commits

pratikvn:
A small fix to allow `rebase!` and `format-rebase!` commands to preserve the merge commits. This is useful for long-standing feature branches so that the merge commit history is still preserved when rebasing.


yhmtsai (approved):
from the documentation
> Any resolved merge conflicts or manual amendments in these merge commits will have to be resolved/re-applied manually.

but I think it does not apply to our merge commit because the merge always simply merge the branches without any changes

pratikvn:
Yes, I think our merge commits are simple commits, so should not have any conflicts, so this should be fine for that case. If it fails in case there are merge conflicts, then it fails anyway, so it should be okay.

tcojean (approved):

PR #1147, opened 2022-10-19
Fixes convergence/papi logger for distributed vectors

MarcelKoch:
This PR fixes the use of the convergence and papi logger with distributed vector. The record logger is not affected by the addition of the distributed types. The stream logger requires significantly more effort to also enable it for distributed types and should probably better be handled in a new logger. 

This can wait until #1133 is merged.

### DIFF
--- a/core/distributed/helpers.hpp
+++ b/core/distributed/helpers.hpp
@@ -144,7 +144,7 @@ bool is_distributed(Arg* linop, Rest*... rest)
  * @param args  The additional arguments of f.
  */
 template <typename ValueType, typename T, typename F, typename... Args>
-void vector_dispatch(T* linop, F&& f, Args... args)
+void vector_dispatch(T* linop, F&& f, Args&&... args)
 {
 #if GINKGO_BUILD_MPI
     if (is_distributed(linop)) {
--- a/core/distributed/helpers.hpp
+++ b/core/distributed/helpers.hpp
@@ -159,7 +159,11 @@ void vector_dispatch(T* linop, F&& f, Args... args)
         using type = std::conditional_t<std::is_const<T>::value,
                                         const matrix::Dense<ValueType>,
                                         matrix::Dense<ValueType>>;
-        f(dynamic_cast<type*>(linop), std::forward<Args>(args)...);
+        if (auto concrete_linop = dynamic_cast<type*>(linop)) {
+            f(concrete_linop, std::forward<Args>(args)...);
+        } else {
+            GKO_NOT_SUPPORTED(linop);
+        }
     }
 }
 
@@ -168,4 +172,4 @@ void vector_dispatch(T* linop, F&& f, Args... args)
 }  // namespace gko
 
 
-#endif
+#endif  // GKO_CORE_DISTRIBUTED_HELPERS_HPP_
--- a/core/distributed/helpers.hpp
+++ b/core/distributed/helpers.hpp
@@ -30,6 +30,10 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ******************************<GINKGO LICENSE>*******************************/
 
+#ifndef GKO_CORE_DISTRIBUTED_HELPERS_HPP_
+#define GKO_CORE_DISTRIBUTED_HELPERS_HPP_
+
+
 #include <memory>
 
 
@@ -162,3 +166,6 @@ void vector_dispatch(T* linop, F&& f, Args... args)
 
 }  // namespace detail
 }  // namespace gko
+
+
+#endif
--- a/core/distributed/helpers.hpp
+++ b/core/distributed/helpers.hpp
@@ -125,7 +125,7 @@ bool is_distributed(Arg* linop, Rest*... rest)
 
 
 template <typename ValueType, typename T, typename F, typename... Args>
-void run_vector(T* linop, F&& f, Args... args)
+void vector_dispatch(T* linop, F&& f, Args... args)
 {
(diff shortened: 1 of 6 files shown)


pratikvn:
@MarcelKoch, I think this can also be included in the release ?

MarcelKoch:
@pratikvn Yes, but I didn't want to create more pressure with the release than there currently is.

pratikvn (approved):
LGTM! Thanks for the fix.

pratikvn:
Yes, but I think this is simple enough to review and probably an important one to have to allow users to log convergence data, which is probably one of the most important thing that users care about when using the solvers.

core/distributed/helpers.hpp
```diff
@@ -124,5 +128,44 @@ bool is_distributed(Arg* linop, Rest*... rest)
...
+template <typename ValueType, typename T, typename F, typename... Args>
+void vector_dispatch(T* linop, F&& f, Args... args)
+{
+#if GINKGO_BUILD_MPI
+    if (is_distributed(linop)) {
+        using type = std::conditional_t<
+            std::is_const<T>::value,
+            const experimental::distributed::Vector<ValueType>,
+            experimental::distributed::Vector<ValueType>>;
+        f(dynamic_cast<type*>(linop), std::forward<Args>(args)...);
```
yhmtsai, line 155: it should check whether dynamic_cast<type*>(linop) is a nullptr

core/distributed/helpers.hpp
```diff
@@ -124,5 +128,44 @@ bool is_distributed(Arg* linop, Rest*... rest)
...
+    }
+}
+
+
 }  // namespace detail
 }  // namespace gko
+
+
+#endif
```
yhmtsai, line 171:
```suggestion
#endif // GKO_CORE_DISTRIBUTED_HELPERS_HPP_
```

core/log/papi.cpp
```diff
@@ -243,12 +246,13 @@ void Papi<ValueType>::on_criterion_check_completed(
...
-        auto tmp_res_norm = Vector::create(residual->get_executor(),
-                                           dim<2>{1, residual->get_size()[1]});
-        auto dense_r = as<Vector>(residual);
-        dense_r->compute_norm2(tmp_res_norm.get());
-        residual_norm_d =
-            static_cast<double>(std::real(tmp_res_norm->at(0, 0)));
+        detail::vector_dispatch<ValueType>(residual, [&](const auto* dense_r) {
+            auto tmp_res_norm = Vector::create(
+                residual->get_executor(), dim<2>{1, residual->get_size()[1]});
+            dense_r->compute_norm2(tmp_res_norm.get());
```
yhmtsai, line 252: another way is to add feature class for the compute_norm2, and then we do not need to know whether it is distributed or not. It's not for this pr

core/test/mpi/distributed/helpers.cpp
```diff
@@ -0,0 +1,113 @@
...
+#include "core/distributed/helpers.hpp"
+
+
+#include "core/test/utils.hpp"
+
+
+int run_function(gko::experimental::distributed::Vector<>*) { return 1; }
+
+
+int run_function(const gko::experimental::distributed::Vector<>*) { return 2; }
```
yhmtsai, line 46:
```suggestion

int run_function(const gko::experimental::distributed::Vector<>*) { return 2; }
```
overloaded function can use 1 empty line

core/test/mpi/distributed/helpers.cpp
```diff
@@ -0,0 +1,113 @@
...
+public:
+    std::shared_ptr<gko::ReferenceExecutor> exec =
+        gko::ReferenceExecutor::create();
+};
+
+
+TEST_F(RunVector, PicksDistributedVectorCorrectly)
+{
+    auto dist_vector =
+        gko::experimental::distributed::Vector<>::create(exec, MPI_COMM_WORLD);
```
yhmtsai, line 65: need to use `std::unique_ptr<LinOp>` if this want to simulate the situation when input is LinOp.
Also for others.

yhmtsai (changes requested):

core/test/mpi/distributed/helpers.cpp
MarcelKoch: you're right, thanks for the suggestion

core/distributed/helpers.hpp
MarcelKoch: Could you expand on that? I don't think any of the two branches makes sense in that case, so I could only throw. Is that your suggestion?

core/log/papi.cpp
yhmtsai: Yes, it's not for this pr

core/distributed/helpers.hpp
yhmtsai: yes, it is. when the linop can not be the type, it will pass the nullptr and lead segmentation fault, I think

yhmtsai (approved):
LGTM. only need to throw the error when the type is incorrect

MarcelKoch:
format!

core/test/mpi/distributed/helpers.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+    gko::detail::vector_dispatch<double>(
+        dense_vector.get(), [&](auto* dense) { result = run_function(dense); });
+
+    ASSERT_EQ(result,
+              run_function(gko::as<gko::matrix::Dense<>>(dense_vector.get())));
+}
+
+
+TEST_F(RunVector, PicksConstDenseVectorCorrectly)
```
tcojean, line 102: Add a test with a random type (Csr?) and ensure it throws?

MarcelKoch:
format!

tcojean (approved):
LGTM. From my side, I think this can be merged.

PR #1148, opened 2022-10-20
Fixes distributed Matrix constructor with LinOp* input

MarcelKoch:
This PR fixes the distributed matrix creation with local and non-local templates passed in as LinOp*. Before, the LinOp* overload would only be used if the input argument has exactly the type `const LinOp*`. In all other cases, eg an input as `LinOp*` or `Csr<...>*`, the templated constructor would be chosen, because, after template deduction, that constructor matches better than the LinOp* one. To prevent this from happening, I've enabled the templated constructor only if the template parameter has a `create<ValueType, IndexType>(exec)` function.

### DIFF
--- a/core/test/mpi/distributed/CMakeLists.txt
+++ b/core/test/mpi/distributed/CMakeLists.txt
@@ -1,1 +1,1 @@
-ginkgo_create_test(matrix MPI_SIZE 3)
+ginkgo_create_test(matrix MPI_SIZE 1)
--- a/core/test/mpi/distributed/matrix.cpp
+++ b/core/test/mpi/distributed/matrix.cpp
@@ -79,15 +79,12 @@ protected:
 template <typename ValueLocalGlobalIndexType>
 class MatrixBuilder : public ::testing::Test {
 protected:
-    using value_type =
-        typename std::tuple_element<0, decltype(
-                                           ValueLocalGlobalIndexType())>::type;
-    using local_index_type =
-        typename std::tuple_element<1, decltype(
-                                           ValueLocalGlobalIndexType())>::type;
-    using global_index_type =
-        typename std::tuple_element<2, decltype(
-                                           ValueLocalGlobalIndexType())>::type;
+    using value_type = typename std::tuple_element<
+        0, decltype(ValueLocalGlobalIndexType())>::type;
+    using local_index_type = typename std::tuple_element<
+        1, decltype(ValueLocalGlobalIndexType())>::type;
+    using global_index_type = typename std::tuple_element<
+        2, decltype(ValueLocalGlobalIndexType())>::type;
     using dist_mtx_type =
         gko::experimental::distributed::Matrix<value_type, local_index_type,
                                                global_index_type>;
@@ -118,9 +115,10 @@ protected:
         {
             SCOPED_TRACE("With Csr with strategy");
             using ConcreteCsr = Csr<value_type, local_index_type>;
-            f(gko::with_matrix_type<Csr>(
-                  std::make_shared<typename ConcreteCsr::classical>()),
-              ConcreteCsr::create(this->ref), [](const gko::LinOp* local_mat) {
+            auto strategy = std::make_shared<typename ConcreteCsr::classical>();
+            f(gko::with_matrix_type<Csr>(strategy),
+              ConcreteCsr::create(this->ref, strategy),
+              [](const gko::LinOp* local_mat) {
                   auto local_csr = gko::as<ConcreteCsr>(local_mat);
 
                   ASSERT_NO_THROW(gko::as<typename ConcreteCsr::classical>(
@@ -141,7 +139,7 @@ protected:
         {
             SCOPED_TRACE("With Fbcsr with block_size");
             f(gko::with_matrix_type<Fbcsr>(5),
-              Fbcsr<value_type, local_index_type>::create(this->ref),
+              Fbcsr<value_type, local_index_type>::create(this->ref, 5),
               [](const gko::LinOp* local_mat) {
                   auto local_fbcsr =
                       gko::as<Fbcsr<value_type, local_index_type>>(local_mat);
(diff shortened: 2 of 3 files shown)


core/test/mpi/distributed/CMakeLists.txt
```diff
@@ -1 +1 @@
-ginkgo_create_test(matrix MPI_SIZE 3)
+ginkgo_create_test(matrix MPI_SIZE 1)
```
tcojean, line 1: Is this change normal? It doesn't seem related to the PR here.

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -467,20 +486,23 @@ class Matrix
...
-                     local_matrix_template
-                         .template create<ValueType, LocalIndexType>(exec)
-                         .get()),
-                 static_cast<const LinOp*>(
-                     non_local_matrix_template
-                         .template create<ValueType, LocalIndexType>(exec)
-                         .get()))
+                 local_matrix_template
+                     .template create<ValueType, LocalIndexType>(exec)
+                     .get(),
```
tcojean, line 502: Minor: I guess this means `Dense` cannot be a local/non-local type since it has no index type. It's probably not important considering we probably don't have a use case for that.

tcojean (approved):
LGTM

include/ginkgo/core/distributed/matrix.hpp
MarcelKoch: It's still possible to create distributed matrix with local/non-local dense matrix type by using the `LinOp*` overloads. But in general I think a dense matrix wouldn't be distributed row-wise, so our approach doesn't make much sense there anyway.

core/test/mpi/distributed/CMakeLists.txt
MarcelKoch: That test actually doesn't use any communication, so it can run on only one process.

MarcelKoch:
@pratikvn:
> * Users providing their own matrix formats. For example, something similar to the `custom-matrix-format` example. Not necessarily matrix free, but a application specific format.

That is already possible. The user can always pass in a LinOp*. If the format is templated by `<ValueType, IndexType>`, then they can also use the `with_matrix_type` helper. (This is already in the core/test/mpi/distributed/matrix.cpp test)

> * Future matrix formats with multiple `IndexType` options: `MatrixType<ValueType, IndexType1,IndexType2>` ? I guess we can add another constructor for that. Maybe we can generalize this ?

Again, that will immediately work with the LinOp* overloads. I don't think the templated overloads make sense there, because they  will just instantiate the local/non-local matrix type with the same value and index type as the distributed matrix. If the local/non-local types have a different number of template parameters, that approach doesn't make sense.
We might still extend this approach for additional matrix template parameters, but until we have a concrete use case it might be too much of a hassle to think about.

yhmtsai (approved):
In general LGTM.
~~For the non-const LinOp* input, does not it use `std::enable_if_t<!std::is_same<const LinOp* , const MatrixType>::value>` ?~~
It does not work, it needs to support the `with_matrix_type`, too.

core/test/mpi/distributed/matrix.cpp
```diff
@@ -279,4 +278,60 @@ TYPED_TEST(MatrixBuilder, BuildWithCustomLinOp)
...
+    using index_type = typename TestFixture::local_index_type;
+    using dist_mat_type = typename TestFixture::dist_mtx_type;
+    this->template forall_matrix_types([this](auto with_matrix_type,
+                                              auto expected_type_ptr,
+                                              auto additional_test) {
+        using expected_type = typename std::remove_pointer<
+            decltype(expected_type_ptr.get())>::type;
+
+        auto mat = dist_mat_type ::create(this->ref, this->comm,
+                                          expected_type_ptr.get());
```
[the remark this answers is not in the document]
MarcelKoch: Probably a copy-paste error

MarcelKoch:
@yhmtsai could you elaborate what doesn't work? Do you mean something like `create(..., const Csr*)` doesn't work or is that related to something else?

MarcelKoch:
format!

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -123,7 +141,7 @@ template <template <typename, typename> class MatrixType, typename... Args>
 auto with_matrix_type(Args&&... create_args)
 {
     return detail::MatrixTypeBuilderFromValueAndIndex<MatrixType, Args...>{
-        std::make_tuple(create_args...)};
+        std::tuple<Args...>{std::forward<Args>(create_args)...}};
```
upsj, line 144: Args pack members are most likely a reference type here - is that intended?

upsj (changes requested):

include/ginkgo/core/distributed/matrix.hpp
MarcelKoch: The result of `make_tuple` won't be `tuple<Args...>` but something else. I'm now changing it a bit to be able to use make_tuple. That would be a bit nicer in c++17...

upsj: are you looking for `https://en.cppreference.com/w/cpp/utility/tuple/forward_as_tuple`?

upsj: What I mean here: The reference collapsing rules lead to Args binding to the most suitable reference types, so you are passing reference types to the arguments to std::tuple, is that what you intended? Do you want to pass the parameters by lvalue reference (std::tie), by value (std::make_tuple) or by the most suitable kind of reference (std::forward_as_tuple)?

MarcelKoch: I think passing references here is fine. This is only called in the Matrix constructor and `create` is called immediately on the result, so the references will stay valid. Then `std::forward_as_tuple` seems to be the best choice.

MarcelKoch:
format!

MarcelKoch:
format!

PR #1171, opened 2022-11-02
Reverts addition of polymorphic_object_traits

MarcelKoch:
The addition of `polymorphic_object_traits` was an interface breaking change. The change would break user classes which derive from `EnablePolymorphicObject`, because having `EnablePolymorphicObject` as a friend class was no longer enough. Instead, the user code would need to change so that the class has `polymorphic_object_traits` as a friend class.

To fix the interface break and still support classes that can only be created from an executor + a communicator, this PR adds a separate class hierarchy for distributed types. This includes the new types `EnableDistributedLinOp` and `EnableDistributedPolymorphicObject`. As a precaution, both types are moved into the experimental namespace.

This PR also adds a helper struct for creating a target object for a temporary precision conversion. This is again needed, because the distributed vector can't be created solely from an executor.


MarcelKoch:
format!

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -912,6 +912,87 @@ class EnableLinOp
...
 
 
+#if GINKGO_BUILD_MPI
+
+
+namespace experimental {
+
+
+/**
+ * The EnableLinOp mixin can be used to provide sensible default implementations
```
pratikvn, line 922:
```suggestion
 * The EnableDistributedLinOp mixin can be used to provide sensible default implementations
```

pratikvn (approved):
LGTM! Thanks for the fix.

cmake/build_helpers.cmake
```diff
@@ -11,10 +11,13 @@ function(ginkgo_default_includes name)
         )
     if(GINKGO_HAVE_HWLOC)
       target_include_directories("${name}"
-        PUBLIC 
+        PUBLIC
         $<BUILD_INTERFACE:${HWLOC_INCLUDE_DIRS}>
         )
     endif()
+    if(GINKGO_BUILD_MPI)
+        target_include_directories("${name}" PUBLIC ${MPI_CXX_INCLUDE_DIRS})
```
upsj, line 19: Does the target include `mpi.h` itself? Then maybe
```suggestion
        target_link_libraries("${name}" PUBLIC MPI::MPI_CXX)
```
Otherwise this should be propagated from the `ginkgo` target

include/ginkgo/core/base/math.hpp
```diff
@@ -495,6 +495,16 @@ template <typename T>
...
 
 
+/**
+ * Obtains the previous type in the singly-linked precision list.
+ *
+ * @note Currently our lists contains only two elements, so this is the same as
+ *       next_precision.
+ */
+template <typename T>
+using previous_precision = next_precision<T>;
```
upsj, line 505: Where is this used? I must have missed it

upsj (commented):
First look, only need to go over polymorphic_object.hpp and the test again

include/ginkgo/core/distributed/vector.hpp
```diff
@@ -510,34 +510,28 @@ class Vector
...
 
-    static PolymorphicObject* clear_impl(Vector* self)
+    static std::unique_ptr<target_type> create(
+        const source_type* source, std::shared_ptr<const Executor> exec)
     {
-        *self = Vector{self->get_executor(), self->get_communicator()};
-        return self;
+        return target_type::create(exec, source->get_communicator());
     }
 };
```
thoasm, line 531: Why is this restricted to only the `source_type`? I don't see any reason why `source_type` couldn't just be a template for `create` (a name I would like to be changed, see previous comment) as it is usually.
If you don't need it, `previous_precision` can also be removed (as this is the only use case for it at the moment).
If you want / need to keep it here, please remove the `gko::` namespace prefix from it here.

include/ginkgo/core/base/polymorphic_object.hpp
```diff
@@ -708,14 +772,22 @@ class EnablePolymorphicObject
...
 
 private:
     GKO_ENABLE_SELF(ConcreteObject);
 };
 
 
+}  // namespace experimental
+
+
+#endif
```
thoasm, line 788: nit: For this `endif`, it is useful to know which `#if` it corresponds to.
Please also add it to all other `#if GINKGO_BUILD_MPI` `#endif`s.
```suggestion
#endif  // GINKGO_BUILD_MPI
```

include/ginkgo/core/base/math.hpp
thoasm: From what I can tell, it is only used here: https://github.com/ginkgo-project/ginkgo/pull/1171/files#diff-e56543c2eafac5af3339aac83474d0ee4836b912a212c020817bc4742312ed70R524
(in `conversion_target_helper` in `include/ginkgo/core/distributed/vector.hpp`).
I am not sure if it is required though, but that is a different comment of mine.

core/test/mpi/base/polymorphic_object.cpp
```diff
@@ -0,0 +1,360 @@
...
+
+    int x;
+};
+
+
+class EnableDistributedPolymorphicObject : public testing::Test {
+protected:
+    std::shared_ptr<gko::ReferenceExecutor> ref{
+        gko::ReferenceExecutor::create()};
+    std::shared_ptr<gko::OmpExecutor> omp{gko::OmpExecutor::create()};
```
thoasm, line 155: Currently, I don't think you can rely on `OmpExecutor` being usable (as `GINKGO_BUILD_OMP` is not dependent on `GINKGO_BUILD_MPI`). Since this is a `core` test, you can only rely on `ReferenceExecutor` being present (and the presence of MPI as the test is otherwise not built).
Weirdly enough, we do the same thing in the [`core/test/base/polymorphic_object.cpp`](https://github.com/ginkgo-project/ginkgo/blob/develop/core/test/base/polymorphic_object.cpp) file, We also need the `OmpExecutor` in the `core/test/base/{array,utils,allocator,executor}.cpp` tests, so this is not unique to `polymorphic_object`.
As we have the same issue in our code already, I am fine with this being ignored, but we should create an issue about that at least to fix it later on.

thoasm (approved):
My main concern is with the `conversion_target_helper`, as I think the name does not quite fit (especially the `create`).
The reset looks good to me.

core/test/mpi/base/polymorphic_object.cpp
yhmtsai: also, dpcpp module can not find omp by default

include/ginkgo/core/base/math.hpp
MarcelKoch: It's used in the conversion helper struct. The template parameter for that struct is the vector in higher precision and the input parameter is the one in lower precision, so I found it odd to use `next_precision` to get the type for the lower precision vector.

core/test/mpi/base/polymorphic_object.cpp
MarcelKoch: Since we already do this, as Thomas mentioned, I would also keep this here, but add a todo to fix that later.

include/ginkgo/core/distributed/vector.hpp
MarcelKoch: Because the vector only supports converting to the next precision. So every other type wouldn't make sense to use here, and by being explicit with the type, a mismatch can be caught at compile time,

include/ginkgo/core/base/temporary_conversion.hpp
```diff
@@ -107,6 +107,17 @@ class convert_back_deleter<const CopyType, const OrigType> {
...
 
+template <typename TargetType>
+struct conversion_target_helper {
+    template <typename SourceType>
+    static std::unique_ptr<TargetType> create(
+        const SourceType* source, std::shared_ptr<const Executor> exec)
+    {
+        return TargetType::create(exec);
+    }
+};
```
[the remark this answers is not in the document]
MarcelKoch: 1. This is a helper struct used for creating an object that is the target of a conversion, so at least for me the naming makes sense. I will rename the create function as you suggested.
2. see 4.
3. Agreed
4. I will go with the single parameter version, thanks for bringing that up.

cmake/build_helpers.cmake
MarcelKoch: We don't call any MPI functions within the modules, so I didn't want to use `target_link_libraries`. But we now have to include `mpi.h` basically everywhere, because it is included from `polymorphic_object.hpp`. So we need to explicitly either add the include paths or link against the mpi target for all of our modules.
I would still prefer to only include the header, but I'm fine with linking against mpi.

include/ginkgo/core/base/polymorphic_object.hpp
MarcelKoch: I always think that clang-format will add this, but it seems it doesn't, so I forget.

cmake/build_helpers.cmake
upsj: I still don't see why this is necessary: Ginkgo links against `MPI::MPI_CXX` publically, so the dependency should be propagated to any target that includes `polymorphic_object.hpp`. Is something broken in our CMake setup?

include/ginkgo/core/distributed/vector.hpp
thoasm: Don't you have the same issue with your default implementation of `convresion_target_helper`? It is a bit strange to me that the default implementation can handle pretty much all cases, while the `distributed::Vector` implementation only handles specific `ValueType` combinations.
Additionally, since we only have 2 different `ValueType`s here, it makes no difference (and would actually remove the need for `previous_precision`).

cmake/build_helpers.cmake
MarcelKoch: Our core module is linked against MPI::MPI_CXX, but nothing else (except tests). So our modules are not linked against MPI, and honestly I would prefer if we could keep it that way. But due to the changes in `polymorphic_object.hpp`, now every module needs to have the MPI include path available.

MarcelKoch: I could move the new classes into separate files, and that way we might be able to circumvent all of this.

upsj: Anything that links against `ginkgo` should also inherit the dependency on MPI_CXX, does that not happen?

include/ginkgo/core/distributed/vector.hpp
MarcelKoch: In the general implementation, we have no idea, which SourceTypes can be converted into the TargetType, but for the special case, we know exactly for which types this is possible. I think we should use that additional information to make it more type safe.
I get that the `previous_precision` is controversial, but for me having the source type as the next precision, while our conversions are usually from lower to higher precision, seems confusing. Also, each vector type is only convertible to `next_precision<value_type>`, so getting the type that a vector can be a target of a conversion would mean to me that we need the inverse of the `next_precision` operation. I tried to add that with the `previous_precision`.

cmake/build_helpers.cmake
MarcelKoch: The modules are only linked against `ginkgo_device`. For the same reason we can't call core functions within the modules.

upsj: In that case, this looks fine. Ideally, every header file should belong only to a single target, and linking against that target should pull in all dependencies. ATM we can't guarantee that yet.

MarcelKoch: I've moved the classes into separate files, so now we don't need to include MPI in the modules anymore.

upsj: Nice, thanks!

include/ginkgo/core/distributed/vector.hpp
thoasm: I don't really have an issue with `previous_precision` as this is indeed useful to have.
It is just strange to me that we have one implementation where you can use it with pretty much any `SourceType`, and one that heavily restricts it. In theory, you could check `SourceType::value_type` and see if it fulfills your requirement, but I understand that this is also not ideal.
I am fine if you want to keep it that way, but please add a short comment why the restriction exists in that case.

core/test/mpi/base/polymorphic_object.cpp
```diff
@@ -0,0 +1,361 @@
...
+        if (obj) {
+            obj->remove_logger(logger.get());
+        }
+    }
+};
+
+
+TEST_F(EnableDistributedPolymorphicObject, CreatesConcreteClass)
+{
+    // this test passes as soon as an instance of `DummyObject` can be created
```
tcojean, line 182: nit
```suggestion
    // this test passes as soon as an instance of `DummyDistributedObject` can be created
```

tcojean (approved):
LGTM!

include/ginkgo/core/distributed/vector.hpp
MarcelKoch: It's actually quite simple to restrict the SourceType by using SFINAE with `std::is_base_of<ConvertibleTo<TargetType>, SourceType>`. I will add that in the general case, and leave the vector specialization as it is. Thanks for bringing the whole issue up!

MarcelKoch:
rebase!

PR #1172, opened 2022-11-02
Adding R scripts analyzing benchmark results and small test data

stanisic:
I would like to emphasize that this contribution does not impact any other part of the Ginkgo code. Only in the case of the large changes of the `benchmark` outputs file structure, these scripts might need certain adaptations.

These 4 R scripts are doing several types of analysis and plotting of the results for the spmv and solver benchmarks. Structure and the reading part of these scripts is following the same approach as the existing `tuning_heatmap.R`. I have also added small, manually modified, test data that one can use to test scripts and inspect their capabilities.

The author-email-affiliation-license header is just a proposal, please let me know if you would prefer a different approach.


tcojean:
Thanks Luka for the contribution! Currently, we're busy with a release, we will take a look at this afterward. Other ways to analyze our benchmark data are always interesting :smile:.

pratikvn:
Thank you for the contribution Luka. As we have no people who have expertise in R, we decided to not merge this into Ginkgo develop. But I think this is still very useful for us, and therefore, we have decided to archive it as a branch here: https://github.com/ginkgo-project/ginkgo/compare/r_benchmark_analysis

Therefore, we will close this PR now.

Issue #1150, opened 2022-10-22
error:  select_classical_spmv: unable to find an eligible kernel


ztdepztdep:
I have compile the ginkgo lib sucessfully with cuda11.8 support. but when i run an example file. 
it gives me the following error. 

./main 
MapSMtoCores for SM 8.6is undefined. The default value of 64 Cores/SM is used.
terminate called after throwing an instance of 'gko::KernelNotFound'
  what():  /home/ztdep/Downloads/ginkgo/cuda/matrix/csr_kernels.cu:259: select_classical_spmv: unable to find an eligible kernel
Aborted (core dumped)

yhmtsai:
@ztdepztdep thanks for reporting this issue!
The message about SM 8.6 does not lead the following `KernelNotFound` issue.
What's the commit SHA of your current ginkgo?
Could you also add `std::cerr << max_length_per_row << std::endl` before `host_kernel::select_classical_spmv(` at line 467 and line 526 of `ginkgo/cuda/matrix/csr_kernels.cu` and run the example again?

ztdepztdep:
I download the ginkgo from the github. which version should i use?

upsj:
It seems like you are using an outdated version of Ginkgo - the error should appear on line 270, not 259. Can you try the latest state from `develop` and let us know if the issue is still present?

ztdepztdep:
It is working now.. thank you very much for your help.

PR #1175, opened 2022-11-03
Update cuda archCoresPerSM

yhmtsai:
This PR updates the Arch CorePerSM map from 11.8 sample, which add SM 8.x and SM 9.0 for Hopper
They also move the sample from the toolkit to GitHub cuds-samples
It should eliminate the warning message mentioned in #1150

### DIFF
--- a/common/cuda_hip/base/executor.hpp.inc
+++ b/common/cuda_hip/base/executor.hpp.inc
@@ -34,7 +34,7 @@ namespace {
 
 
 // The function is copied from _ConvertSMVer2Cores of
-// cuda-11.0/samples/common/inc/helper_cuda.h
+// cuda-samples:Common/helper_cuda.h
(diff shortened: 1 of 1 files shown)


upsj (approved):

tcojean (approved):

pratikvn:
We probably dont need this function anymore, right ? We can just infer it directly ? `this->get_exec_info().num_computing_units()` gives you the number of cores and you can use that to get `num_pu_per_cu`, which seems to be the only place this is currently used.

thoasm (approved):
LGTM!

yhmtsai:
@pratikvn I do not get it. do you mean there's another attribute we can combine with `num_cu` to get `num_pu_per_cu`?

pratikvn (approved):
I think I misunderstood what the function was for. Ignore my comment.

PR #1178, opened 2022-11-03
Pipeline updates to improve MPI CI jobs.

### LINKED ISSUE #1174: Very Slow MPI CI for GPU backends in some settings
There seems to be an issue on some machines and systems where the CI with MPI-enabled solvers become increasingly slow. Interestingly, that doesn't seem to be the case everywhere, so it may simply be a setup issue. What follows is a non exhaustive list of what I could see:

Slow jobs:
* AMDCI system with ROCm 4.0 only for the `HIP` executor and a docker backend : https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/3260809640
* HoreKa with CUDA only for the `CUDA` executor and enroot+slurm backend (this job was later moved to the nla-gpu machine): https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/3243280182

Not slow:
* nla-gpu machine with ROCm and docker backend: 
https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/3259622866

Questions:
* ~Since this is slow only for GPU backends, is there a problem with the non-GPU-direct communication logic which make all of this terribly slow?~
* Another possibility, are we somehow oversubscribing the GPUs in some settings (multiple processes using the same GPU)? Oversubscription of the GPUs could go very bad and make it very slow.

pratikvn:
+ Move some MPI jobs from AMDCI to nla-gpu.
+ SLURM_GRES=4 for horeka MPI jobs.

Closes #1174


.gitlab-ci.yml
```diff
@@ -501,15 +533,13 @@ build/amd/openmpi/clang/rocm502/release/shared:
   extends:
     - .build_and_test_template
     - .default_variables
-    - .full_test_condition
-    - .use_gko-rocm502-openmpi-gnu11-llvm11-multi-gpu
```
tcojean, line 505: this one would then be unneeded as well I think

pratikvn: You mean remove this job entirely ?

.gitlab/image.yml
```diff
@@ -67,7 +67,6 @@
   image: ginkgohub/rocm:40-openmpi-gnu8-llvm50
   tags:
     - private_ci
```
pratikvn, line 69:
```suggestion
    - private_ci
    - nla-gpu
```

.gitlab-ci.yml
pratikvn: Yeah, I just kept it in case we want to use that later.

.gitlab-ci.yml
```diff
@@ -341,6 +341,38 @@ build/cuda102/nompi/intel/cuda/debug/static:
...
+  extends:
+    - .horeka_test_template
+    - .default_variables
+    - .full_test_condition
+    - .use_gko-cuda110-mvapich2-gnu9-llvm9-intel2020
+  variables:
+    USE_NAME: "cuda110-mvapich2-gcc-${CI_PIPELINE_ID}"
+    SLURM_PARTITION: "accelerated"
+    SLURM_GRES: "gpu:4"
+    SLURM_TIME: "00:45:00"
```
yhmtsai, line 371: I think in another pr, it exceeds the time limit.
Could you extend it to 1:00:00? also for other slurm jobs

.gitlab-ci.yml
```diff
@@ -501,15 +533,13 @@ build/amd/openmpi/clang/rocm502/release/shared:
...
-    - .full_test_condition
-    - .use_gko-rocm502-openmpi-gnu11-llvm11-multi-gpu
+    - .quick_test_condition
+    - .use_gko-rocm502-openmpi-gnu11-llvm11
   variables:
     C_COMPILER: "clang"
     CXX_COMPILER: "clang++"
     BUILD_OMP: "ON"
     BUILD_HIP: "ON"
-    BUILD_MPI: "ON"
```
yhmtsai, line 511: mistake?

yhmtsai (commented):
I think it does not close the issue because we do not figure why rocm 4.0 leads slow at least.

tcojean:
If the pipeline works on nla-gpu, then it closes the issue, because that means the problem was indeed GPU oversubscription (the 3 MPI processes were sharing the one Radeon VII of amdci and that made things *very* slow).

Another comment, currently I believe we are not using this parameter: https://github.com/ginkgo-project/ginkgo/blob/develop/.gitlab/scripts.yml#L32 to use it, it should be `BUILD_MPI` instead of `GINKGO_BUILD_MPI` I think. So is this needed at all?

.gitlab-ci.yml
pratikvn: No, that was a job on the amdci machine, where we do not want MPI jobs anymore. I have updated the job name accordingly now.

yhmtsai (approved):
LGTM.

tcojean (approved):
LGTM!

The jobs indeed seem much faster, so oversubscription was the problem in #1174

pratikvn:
Yes, oversubscription was indeed the issue. I removed all remaining MPI jobs from the AMDCI machine and moved them to the nla-gpu machine.

.gitlab-ci.yml
```diff
@@ -451,30 +483,30 @@ build/amd/openmpi/clang/rocm40/release/static:
     CXX_COMPILER: "clang++"
     BUILD_OMP: "ON"
     BUILD_HIP: "ON"
-    MPI_AS_ROOT: "ON"
     BUILD_MPI: "ON"
+    MPI_AS_ROOT: "ON"
```
tcojean, line 487: Maybe make this a quick test, since the HoreKa one is already a full test?

.gitlab/scripts.yml
```diff
@@ -29,8 +29,8 @@
       export OMPI_ALLOW_RUN_AS_ROOT=1;
       export OMPI_ALLOW_RUN_AS_ROOT_CONFIRM=1;
       fi
-    - if [[ "${GINKGO_BUILD_MPI}" == "ON" ]]; then
-      MPI_STR=-DGINKGO_MPI_EXEC_SUFFIX=${MPI_SUFFIX} -DMPI_RUN_AS_ROOT=${MPI_AS_ROOT};
+    - if [[ "${BUILD_MPI}" == "ON" ]]; then
+      MPI_STR=-DGINKGO_MPI_EXEC_SUFFIX=${MPI_SUFFIX};
```
yhmtsai, line 33:
```suggestion
      MPI_STR=-DGINKGO_MPI_EXEC_SUFFIX=${MPI_SUFFIX} -DMPI_RUN_AS_ROOT=${MPI_AS_ROOT};
```
create_test.cmake needs `MPI_RUN_AS_ROOT` to add `--allow-run-as-root`

pratikvn: No, I think either the flag is needed or the environment variable needs to be set, which we do. See https://www.open-mpi.org/doc/v4.0/man1/mpirun.1.php#sect25

tcojean: In that case, this block as well as the logic here can be dropped entirely, no?
https://github.com/ginkgo-project/ginkgo/blob/develop/cmake/create_test.cmake#L59-L63

PR #1184, opened 2022-11-04
Check number of rhs for CUDA triangular solvers

### LINKED ISSUE #1138: Throw exception if matrix dimensions at analysis and solve don't match for cuSPARSE generic triangular solve
If we don't check this, we run into a data corruption bug (after solving with incorrect number of rhs > 32 when analyzed with number of rhs = 1, all subsequent solves will produce incorrect results).

Required steps:
* [ ] store nrhs in SolveStruct and check for equality at solve
* [ ] adapt test/solver/solver.cpp to regenerate the solver for each input vector separately (probably needs a new trait)

upsj:
This makes specifying the `with_num_rhs` parameter for the triangular solvers a requirement on the CUDA executor to avoid the (apparent) internal data corruption in cuSPARSE that may occur otherwise when analyzing with <= 32 rhs and solving with > 32 rhs.

Additionally, I needed to modify the solver tests, since instead of the typical `forall_solvers(..., forall_vectors(...))`, here we need to use `forall_vectors(..., forall_solvers_with_num_rhs(...))`.

Closes #1138

### DIFF
--- a/cuda/solver/common_trs_kernels.cuh
+++ b/cuda/solver/common_trs_kernels.cuh
@@ -43,6 +43,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <cusparse.h>
 
 
+#include <ginkgo/core/base/exception.hpp>
 #include <ginkgo/core/base/exception_helpers.hpp>
 #include <ginkgo/core/base/math.hpp>
 
@@ -84,6 +85,7 @@ struct CudaSolveStruct : gko::solver::SolveStruct {
     cusparseHandle_t handle;
     cusparseSpSMDescr_t spsm_descr;
     cusparseSpMatDescr_t descr_a;
+    size_type num_rhs;
 
     // Implicit parameter in spsm_solve, therefore stored here.
     array<char> work;
@@ -94,8 +96,12 @@ struct CudaSolveStruct : gko::solver::SolveStruct {
         : handle{exec->get_cusparse_handle()},
           spsm_descr{},
           descr_a{},
+          num_rhs{num_rhs},
           work{exec}
     {
+        if (num_rhs == 0) {
+            return;
+        }
         cusparse::pointer_mode_guard pm_guard(handle);
         spsm_descr = cusparse::create_spsm_descr();
         descr_a = cusparse::create_csr(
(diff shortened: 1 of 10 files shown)


cuda/solver/common_trs_kernels.cuh
```diff
@@ -143,6 +149,17 @@ struct CudaSolveStruct : gko::solver::SolveStruct {
                matrix::Dense<ValueType>* output, matrix::Dense<ValueType>*,
                matrix::Dense<ValueType>*) const
     {
+        if (input->get_size()[1] != num_rhs) {
+            throw gko::ValueMismatch{
```
tcojean, line 153: In effect, this now means there is a `num_rhs` factory parameter requirement to be specified with triangular solvers (or CUDA will throw). Should I then list this change in the "Deprecations and important changes" of the release?

tcojean (approved):
LGTM

cuda/solver/common_trs_kernels.cuh
```diff
@@ -94,8 +96,12 @@ struct CudaSolveStruct : gko::solver::SolveStruct {
         : handle{exec->get_cusparse_handle()},
           spsm_descr{},
           descr_a{},
+          num_rhs{num_rhs},
```
yhmtsai, line 99: does hip also need that?

cuda/solver/common_trs_kernels.cuh
upsj: We should probably mention it, yes

cuda/solver/common_trs_kernels.cuh
upsj: from the interface, it does, but I haven't seen it matter in practice yet. Still, we should probably keep it in here

upsj: Scratch that - the HIP implementation only uses the single-rhs version, so it doesn't matter here

yhmtsai (approved):
LGTM

yhmtsai:
could you also add the short description in the beginning comment of pull request?

cuda/solver/common_trs_kernels.cuh
tcojean: I tried to mention this change in https://github.com/ginkgo-project/ginkgo/discussions/1177

tcojean (approved):

cuda/solver/common_trs_kernels.cuh
upsj: I extended the explanation a bit, thanks

upsj:
This makes specifying the with_num_rhs parameter for the triangular solvers a requirement on the CUDA executor
to avoid the (apparent) internal data corruption in cuSPARSE that may occur otherwise when analyzing
with <= 32 rhs and solving with > 32 rhs.

Additionally, I needed to modify the solver tests, since instead of the typical `forall_solvers(..., forall_vectors(...))`,
here we need to use `forall_vectors(..., forall_solvers_with_num_rhs(...))`.

Related PR: https://github.com/ginkgo-project/ginkgo/pull/1184

PR #1194, opened 2022-11-10
Find CUDA libraries with NVHPC package

tcojean:
As the title says. There might be other fixes piggy-backed in this PR.

### DIFF
--- a/cmake/cuda.cmake
+++ b/cmake/cuda.cmake
@@ -21,13 +21,44 @@ cas_variable_cuda_architectures(GINKGO_CUDA_ARCH_FLAGS
     ARCHITECTURES ${GINKGO_CUDA_ARCHITECTURES}
     UNSUPPORTED "20" "21")
 
+if (${CMAKE_CXX_COMPILER_ID} MATCHES "PGI|NVHPC")
+    find_package(NVHPC REQUIRED
+        HINTS
+        $ENV{NVIDIA_PATH}
+        ${CMAKE_CUDA_COMPILER}/../../..
+        )
+
+    set(CUDA_RUNTIME_LIBS_DYNAMIC ${NVHPC_CUDART_LIBRARY})
+    set(CUDA_RUNTIME_LIBS_STATIC ${NVHPC_CUDART_LIBRARY_STATIC})
+    set(CUBLAS ${NVHPC_CUBLAS_LIBRARY})
+    set(CUSPARSE ${NVHPC_CUSPARSE_LIBRARY})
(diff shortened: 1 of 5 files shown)


cmake/cuda.cmake
```diff
@@ -21,13 +21,44 @@ cas_variable_cuda_architectures(GINKGO_CUDA_ARCH_FLAGS
     ARCHITECTURES ${GINKGO_CUDA_ARCHITECTURES}
     UNSUPPORTED "20" "21")
 
+if (${CMAKE_CXX_COMPILER_ID} MATCHES "PGI|NVHPC")
+    find_package(NVHPC REQUIRED
```
tcojean, line 25: Does this need to be required in this case, or we silently disable CUDA if not found?

include/ginkgo/core/stop/combined.hpp
```diff
@@ -125,7 +125,6 @@ std::shared_ptr<const CriterionFactory> combine(FactoryContainer&& factories)
     switch (factories.size()) {
     case 0:
         GKO_NOT_SUPPORTED(nullptr);
-        return nullptr;
```
yhmtsai, line 128: the first condition in default can also delete the `return nullptr`

omp/distributed/partition_kernels.cpp
```diff
@@ -60,6 +60,8 @@ void build_starting_indices(std::shared_ptr<const DefaultExecutor> exec,
     auto size_per_thread =
         static_cast<size_type>(ceildiv(num_ranges, num_threads));
     vector<LocalIndexType> local_sizes(num_parts * num_threads, 0, {exec});
+    // TODO: PGI/NVHPC(22.7) doesn't like reduction with references
+    int tmp_reduction = 0;
```
yhmtsai, line 64: should we do it in this pr? We do not support omp with PGI/NVHPC, so we can just mention somewhere without changing it

yhmtsai (approved):

cmake/cuda.cmake
yhmtsai: not sure about PGI.
For NVHPC, I will assume it is always shipped with the cuda toolkit

tcojean: the problem is the `nvc++` compiler is identified as `PGI` at least on Perlmutter...

omp/distributed/partition_kernels.cpp
tcojean: fair enough I can remove this

pratikvn (approved):
LGTM!

PR #1196, opened 2022-11-12
Release 1.5.0 to master

tcojean:
Includes all the commits from develop since 1.4.0 release to master. 

There is one extra commit only due to leftover 1.4.0 references.


tcojean:
Moved to #1197 due to CI issues (gitlab push mirroring doesn't like `/` in the names).

PR #1199, opened 2022-11-13
Errors with the pipelines for 1.5.0

tcojean:
Some fixes required if we want passing pipelines for master and v1.5.0 tag

### DIFF
--- a/.github/workflows/joss.yml
+++ b/.github/workflows/joss.yml
@@ -28,7 +28,7 @@ jobs:
         cd doc
         cp -r ../../doc/joss ./
         cd joss
-        pandoc paper.md -o paper.pdf --bibliography ./paper.bib --latex-engine=xelatex
+        pandoc paper.md -o paper.pdf --bibliography ./paper.bib --pdf-engine=xelatex
         popd
     - uses: actions/upload-artifact@v1.0.0
       with:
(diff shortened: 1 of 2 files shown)


.github/workflows/joss.yml
```diff
@@ -28,7 +28,7 @@ jobs:
         cd doc
         cp -r ../../doc/joss ./
         cd joss
-        pandoc paper.md -o paper.pdf --bibliography ./paper.bib --latex-engine=xelatex
+        pandoc paper.md -o paper.pdf --bibliography ./paper.bib --pdf-engine=xelatex
```
yhmtsai, line 31: is it due to the pandoc update?

tcojean: it seems like it, indeed with recent versions there is only mention of `--pdf-engine` in the manual

tcojean:
This is now using two different gh-pages jobs. The `sync` is not called for tags, as it makes no sense. But to accommodate, a special `gh-pages-tags` is added which does not have the `need` clause for tags, the other jobs deploy the documentation like before, by relying on the needs clause. See: https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/pipelines/693321840

yhmtsai (approved):

MarcelKoch (approved):

pratikvn (approved):

PR #1200, opened 2022-11-16
Fix path to openmpi_test.cpp in uses_openmpi try_run

Nordegraf:
This fixes an issue with the uses_openmpi test run to determine the OpenMPI version introduced in commit https://github.com/ginkgo-project/ginkgo/commit/2c05a7949974dfeb0562590d8ae51861b8a5eb8b#.

Currently the path to the openmpi_test.cpp file is given using CMAKE_SOURCE_DIR which resolves to the source direction of the highest cmake script. In M++ we include Ginkgo as subproject which results in an error when compiling, because the test file can not be found.
I switched this to Ginkgo_SOURCE_DIR which resolves to ginkgos project directory. 

This compiles within M++ and as standalone project without errors now.

### DIFF
--- a/contributors.txt
+++ b/contributors.txt
@@ -23,3 +23,4 @@ Olenik Gregor <go@hpsim.de> HPSim
 Ribizel Tobias <mail@upsj.de> Karlsruhe Institute of Technology
 Riemer Lukas <lksriemer@gmail.com> Karlsruhe Institute of Technology
 Tsai Yuhsiang <yhmtsai@gmail.com> National Taiwan University
+Niklas Conen <niklas.conen@student.kit.edu> Karlsruhe Institute of Technology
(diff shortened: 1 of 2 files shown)


contributors.txt
```diff
@@ -23,3 +23,4 @@ Olenik Gregor <go@hpsim.de> HPSim
 Ribizel Tobias <mail@upsj.de> Karlsruhe Institute of Technology
 Riemer Lukas <lksriemer@gmail.com> Karlsruhe Institute of Technology
 Tsai Yuhsiang <yhmtsai@gmail.com> National Taiwan University
+Niklas Conen <niklas.conen@student.kit.edu> Karlsruhe Institute of Technology
```
upsj, line 26: We try to keep this list in alphabetical order, last name + first name, so you can move yourself up a bit :) Also if you have a more permanent mail address that you are willing to use, I'd suggest using that instead.

upsj (approved):
LGTM! Thanks for the fix!

MarcelKoch (approved):

tcojean (approved):
LGTM

MarcelKoch:
label!

PR #1201, opened 2022-11-16
Enable distributed Gmres

MarcelKoch:
This PR enables Gmres for our distributed matrices and vectors.

Thanks to the changes in #861 this was quite simple to adapt. Only adding `create_with_type_of` to `Vector` was really necessary. As a consequence, I also needed to add `create_with_type_of(..., gko::dim<2>, gko::dim<2>, stride)` to `Dense` so that both `Vector` and `Dense` have the same interface. It might be possible to sidestep this, but that would require annoying template meta programming.

I decided against adding `create_submatrix` to `Vector`, because I would need to think a bit more about the interface for that.

### DIFF
--- a/include/ginkgo/core/distributed/vector.hpp
+++ b/include/ginkgo/core/distributed/vector.hpp
@@ -121,6 +121,44 @@ public:
         return (*other).create_with_same_config();
     }
 
+
+    /**
+     * Creates an empty Vector with the same type as another Vector, but on a
+     * different executor.
+     *
+     * @param other  The other multi-vector whose type we target.
+     * @param exec  The executor of the new multi-vector.
+     *
+     * @note  The new multi-vector uses the same communicator as other.
+     *
+     * @returns an empty Vector with the type of other.
+     */
+    static std::unique_ptr<Vector> create_with_type_of(
+        const Vector* other, std::shared_ptr<const Executor> exec)
+    {
+        return (*other).create_with_type_of_impl(exec, {}, {}, 0);
+    }
+
+    /**
+     * Creates an Vector with the same type as another Vector, but on a
+     * different executor and with a different size.
+     *
+     * @param other  The other multi-vector whose type we target.
+     * @param exec  The executor of the new multi-vector.
+     * @param global_size  The global size of the multi-vector.
+     * @param local_size  The local size of the multi-vector.
+     * @param stride  The stride of the new multi-vector.
+     *
+     * @returns a Vector of specified size with the type of other.
+     */
+    static std::unique_ptr<Vector> create_with_type_of(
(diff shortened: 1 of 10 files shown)


include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -449,6 +449,22 @@ class SolverBase {
...
+    template <typename LinOpType>
+    LinOpType* create_workspace_op_with_type_of(int vector_id,
+                                                const LinOpType* vec,
+                                                dim<2> global_size,
+                                                dim<2> local_size) const
+    {
+        return workspace_.template create_or_get_op<LinOpType>(
+            vector_id,
+            [&] {
+                return LinOpType::create_with_type_of(
```
upsj, line 461: Is it feasible to test this behavior?

include/ginkgo/core/distributed/vector.hpp
```diff
@@ -121,6 +121,44 @@ class Vector
...
+     * @param other  The other multi-vector whose type we target.
+     * @param exec  The executor of the new multi-vector.
+     *
+     * @note  The new multi-vector uses the same communicator as other.
+     *
+     * @returns an empty Vector with the type of other.
+     */
+    static std::unique_ptr<Vector> create_with_type_of(
+        const Vector* other, std::shared_ptr<const Executor> exec)
+    {
```
upsj, line 138: nit: all of these functions don't need to live in the header

upsj (approved):
LGTM! In the long run, I would probably try to find a suitable public interface for create_submatrix, but for now this looks fine.

include/ginkgo/core/solver/solver_base.hpp
MarcelKoch: In what sense? I added tests for the `create_with_type_of` for `Vector`. I will also add a test for the two size version of `Dense`.

upsj: specifically the workspace functionality, not the Vector functionality

MarcelKoch: First of all, these are kind of implicitly tested through all the concrete solver tests. Also, if we would test this, I think it makes sense to test the other workspace functions. But, I think that should be tackled in a separate PR.

pratikvn (approved):
LGTM!

MarcelKoch:
format!

fritzgoebel:
On CUDA I get a `CudaInvaliValue` in the `krylov_dim - 1` th iteration. If I set the size of `hessenberg` to `{krylov_dim + 2, krylov_dim * num_rhs}` it works, I'm not quite sure why this is though

MarcelKoch:
rebase!

MarcelKoch:
rebase!

PR #1204, opened 2022-11-18
Add distributed benchmarks for SpMV, Solver, Multi-Vector

MarcelKoch:
This PR adds distributed benchmarks for SpMV, Solver, and Multi-Vector operations.

To reduce code duplication, the non-distributed versions of these benchmarks have been refactored. In each case, a common core has been extracted. Additionally, creating vectors/matrices has been delegated to helper types for the sequential and distributed case. The common code uses templating with duck-typing to be generalizable across sequential and distributed benchmarks.

Also, this enables generated stencil matrices for all benchmarks.

The distributed multi-vector benchmark does not support BLAS level 2 or level 3 operations. The non-distributed still supports the same operations as before.

The distributed solver benchmark supports only the currently available distributed solvers.

Both in the distributed solver and spmv benchmark, the local and non-local format can be specified via the CLI.


benchmark/blas/blas.cpp
```diff
@@ -33,482 +33,99 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+    n: number of rows for vectors and gemm output (required)
+    r: number of columns for vectors (optional, default 1)
+    m: number of columns for gemm output (optional, default n)
+    k: inner dimension of the gemm (optional, default n)
+    stride: storage stride for both vectors (optional, default r)
+    stride_x: stride for input vector x (optional, default r)
+    stride_y: stride for in/out vector y (optional, default r)
+    stride_A: stride for A matrix in gemm (optional, default k)
+    stride_B: stride for B matrix in gemm (optional, default m)
+    stride_C: stride for C matrix in gemm (optional, default m
```
upsj, line 121:
```suggestion
    stride_C: stride for C matrix in gemm (optional, default m)
```

benchmark/utils/loggers.hpp
```diff
@@ -186,6 +189,25 @@ struct StorageLogger : gko::log::Logger {
...
+        gko::size_type total{};
+        for (const auto& e : storage) {
+            total += e.second;
+        }
+        comm.reduce(gko::ReferenceExecutor::create(),
+                    comm.rank() == 0
+                        ? static_cast<gko::size_type*>(MPI_IN_PLACE)
+                        : &total,
+                    &total, 1, MPI_SUM, 0);
+        add_or_set_member(output, "storage", total, allocator);
```
upsj, line 207: TODO for the future: Figure out if we want to do node-local benchmark results as well. Might be useful for identifying load-imbalance.

benchmark/utils/distributed_helpers.hpp
```diff
@@ -0,0 +1,117 @@
...
+
+std::string broadcast_json_input(std::istream& is,
+                                 gko::experimental::mpi::communicator comm)
+{
+    auto exec = gko::ReferenceExecutor::create();
+
+    std::string json_input;
+    if (comm.rank() == 0) {
+        std::string line;
+        while (std::cin >> line) {
```
upsj, line 59:
```suggestion
        while (is >> line) {
```

benchmark/utils/distributed_helpers.hpp
```diff
@@ -0,0 +1,117 @@
...
+    comm.broadcast(exec->get_master(), &input_size, 1, 0);
+    json_input.resize(input_size);
+    comm.broadcast(exec->get_master(), &json_input[0],
+                   static_cast<int>(input_size), 0);
+
+    return json_input;
+}
+
+
+std::unique_ptr<dist_mtx<etype, itype, gko::int64>> create_distributed_matrix(
```
upsj, line 74: maybe provide a type alias for the global index type?

benchmark/utils/distributed_helpers.hpp
```diff
@@ -0,0 +1,117 @@
...
+
+    auto dist_mat = dist_mtx<etype, itype, gko::int64>::create(
+        exec, comm, local_mat.get(), non_local_mat.get());
+    dist_mat->read_distributed(data, part);
+
+    exec->remove_logger(gko::lend(storage_logger));
+    storage_logger->write_data(comm, spmv_case, allocator);
+
+    return dist_mat;
+}
```
upsj, line 97: suggestion to reduce duplication
```suggestion
    rapidjson::Value* spmv_case = nullptr, rapidjson::MemoryPoolAllocator<>* allocator = nullptr)
{
    auto local_mat = formats::matrix_type_factory.at(format_local)(exec);
    auto non_local_mat =
        formats::matrix_type_factory.at(format_non_local)(exec);

    std::shared_ptr<StorageLogger> storage_logger;
    if (spmv_case) {
        storage_logger = std::make_shared<StorageLogger>();
        exec->add_logger(storage_logger);
    }

    auto dist_mat = dist_mtx<etype, itype, gko::int64>::create(
        exec, comm, local_mat.get(), non_local_mat.get());
    dist_mat->read_distributed(data, part);

    if (spmv_case) {
        exec->remove_logger(gko::lend(storage_logger));
        storage_logger->write_data(comm, *spmv_case, *allocator);
    }

    return dist_mat;
}
```

benchmark/utils/generator.hpp
```diff
@@ -0,0 +1,278 @@
...
+    {
+        return gko::share(
+            ::formats::matrix_factory(format_name, std::move(exec), data));
+    }
+
+    static std::shared_ptr<gko::LinOp> generate_matrix_with_default_format(
+        std::shared_ptr<const gko::Executor> exec,
+        const gko::matrix_data<ValueType, itype>& data)
+    {
+        return generate_matrix_with_format(std::move(exec), "coo", data);
```
upsj, line 91: Should coo stay the default? I think csr might be another suitable candidate.

benchmark/utils/generator.hpp
```diff
@@ -0,0 +1,278 @@
...
+        return res;
+    }
+
+    static std::unique_ptr<Vec> create_multi_vector(
+        std::shared_ptr<const gko::Executor> exec, gko::dim<2> size,
+        gko::size_type stride)
+    {
+        auto res = Vec::create(exec, size, stride);
+        return res;
+    }
```
upsj, line 109: These two overloads are a bit too close for comfort. If you forget a `.0` somewhere, you might inadvertently create an uninitialized vector. I would suggest a specific name
```suggestion
    static std::unique_ptr<Vec> create_multi_vector(
        std::shared_ptr<const gko::Executor> exec, gko::dim<2> size,
        ValueType value)
    {
        auto res = Vec::create(exec);
        res->read(gko::matrix_data<ValueType, itype>(size, value));
        return res;
    }

    static std::unique_ptr<Vec> create_multi_vector_strided(
        std::shared_ptr<const gko::Executor> exec, gko::dim<2> size,
        gko::size_type stride)
    {
        auto res = Vec::create(exec, size, stride);
        return res;
    }
```

benchmark/utils/generator.hpp
```diff
@@ -0,0 +1,278 @@
...
+                                                                           1.0),
+            get_engine()));
+        return res;
+    }
+
+    // creates a zero vector
+    static std::unique_ptr<Vec> create_vector(
+        std::shared_ptr<const gko::Executor> exec, gko::size_type size)
+    {
+        return create_multi_vector(std::move(exec), gko::dim<2>{size, 1}, 1);
```
upsj, line 128: the vector is most likely uninitialized, not zero!

upsj (commented):
First pass - a few small issues, but otherwise really nice work!

Are there any notable changes in the operations I should be aware of (except for supporting the generators)? Reviewing moved code is always a bit of a challenge.

benchmark/utils/loggers.hpp
MarcelKoch: That's a good point. For the storage, that should be easy enough. But for timings, that would be quite difficult, since we have the synchronization from the global reductions. That is also why I'm just reporting the timings from rank 0 atm, hoping that there is no significant derivation.

benchmark/utils/generator.hpp
MarcelKoch: I didn't want to change any existing behavior, so I just kept the currently used default format. I don't think I would want to mix that discussion in here.

MarcelKoch:
@upsj I don't think I've changed anything significant in the moved code, except for using the generator to create matrices and vectors.

benchmark/utils/generator.hpp
upsj: it's used in create_vector, see the thread below

upsj: No wait, I'm not entirely sure which overload fits better, ValueType or size_type? How does this even compile? :laughing:

benchmark/utils/loggers.hpp
upsj: As long as we collect a runtime breakdown into individual operations, we might still be able to get something from it?

MarcelKoch: Right, in that case it would be helpful. But I would prefer doing the statistics on the different local runtimes in a different PR.

upsj: definitely!

benchmark/utils/generator.hpp
MarcelKoch: Yes, that one below is definitely wrong :sweat_smile:. But I've found the actual use for that in the blas benchmark, so I will fix it.

benchmark/utils/generator.hpp
MarcelKoch: this is not even used...

benchmark/utils/stencil_matrix.hpp
```diff
@@ -0,0 +1,501 @@
...
+        gko::dim<2>{static_cast<gko::size_type>(global_size),
+                    static_cast<gko::size_type>(global_size)});
+
+    auto global_offset = [&](const int px, const int py) {
+        return static_cast<int>(local_size) * px +
+               static_cast<int>(local_size) * dims[0] * py;
+    };
+
+    auto target_coords = [&](const IndexType i, const int coord) {
+        return 0 <= i && i <= dp - 1 ? coord : i < 0 ? coord - 1 : coord + 1;
```
upsj, line 88: I would suggest adding parentheses for clarity
```suggestion
        return 0 <= i && (i <= dp - 1 ? coord : (i < 0 ? coord - 1 : coord + 1));
```

benchmark/utils/stencil_matrix.hpp
```diff
@@ -0,0 +1,501 @@
...
+        return num_neighbors;
+    };
+    const auto diag_value = static_cast<ValueType>(nnz_in_row() - 1);
+
+    for (IndexType i = 0; i < dp; ++i) {
+        for (IndexType j = 0; j < dp; ++j) {
+            auto row = flat_idx(j, i);
+            for (IndexType d_i : {-1, 0, 1}) {
+                for (IndexType d_j : {-1, 0, 1}) {
+                    if (is_valid_neighbor(d_i, d_j)) {
```
upsj, line 127: this doesn't seem to do any bounds checking at the boundary, is that intended?

benchmark/utils/stencil_matrix.hpp
```diff
@@ -0,0 +1,501 @@
...
+    const auto diag_value = static_cast<ValueType>(nnz_in_row() - 1);
+
+    for (IndexType i = 0; i < dp; ++i) {
+        for (IndexType j = 0; j < dp; ++j) {
+            for (IndexType k = 0; k < dp; ++k) {
+                auto row = flat_idx(k, j, i);
+                for (IndexType d_i : {-1, 0, 1}) {
+                    for (IndexType d_j : {-1, 0, 1}) {
+                        for (IndexType d_k : {-1, 0, 1}) {
+                            if (is_valid_neighbor(d_i, d_j, d_k)) {
```
upsj, line 228: same here

benchmark/utils/stencil_matrix.hpp
```diff
@@ -0,0 +1,501 @@
...
+        dims, coords, target_local_size, restricted);
+}
+
+
+/**
+ * Generates matrix data for a 2D stencil matrix. If restricted is set to true,
+ * creates a 5-pt stencil, if it is false creates a 9-pt stencil. If
+ * strong_scaling is set to true, creates the same problemsize independent of
+ * the number of ranks, if it false the problem size grows with the number of
+ * ranks.
```
upsj, line 324: Could you add a short explanation what makes this communication-optimal?

benchmark/utils/stencil_matrix.hpp
```diff
@@ -0,0 +1,501 @@
...
+    return A_data;
+}
+
+
+/**
+ * Generates matrix data for a 3D stencil matrix. If restricted is set to true,
+ * creates a 7-pt stencil, if it is false creates a 27-pt stencil. If
+ * strong_scaling is set to true, creates the same problemsize independent of
+ * the number of ranks, if it false the problem size grows with the number of
+ * ranks.
```
upsj, line 388: same here

upsj (changes requested):
I think there might be an issue with the boundary nodes in the stencil matrix setup, otherwise no major concerns.

benchmark/blas/blas_common.hpp
```diff
@@ -0,0 +1,511 @@
...
+    gko::size_type r;
+    gko::size_type stride_x;
+    gko::size_type stride_y;
+    gko::size_type stride_A;
+    gko::size_type stride_B;
+    gko::size_type stride_C;
+};
+
+
+gko::size_type get_optional(rapidjson::Value& obj, const char* name,
```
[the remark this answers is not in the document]
MarcelKoch: I think moving it into a lambda function removes all the generality.

benchmark/solver/distributed/solver.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+    std::stringstream ss_rel_res_goal;
+    ss_rel_res_goal << std::scientific << FLAGS_rel_res_goal;
+
+    std::string json_input = FLAGS_overhead
+                                 ? R"(
+[{"filename": "overhead.mtx",
+  "optimal": {"spmv": "csr-csr"}]
+)"
+                                 : broadcast_json_input(std::cin, comm);
```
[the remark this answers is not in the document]
MarcelKoch: I've looked a bit into it and the difference lies in which process have access to stdin. That is something that can be configured when running mpirun, but the default depends on the MPI implementation. I think common to all of them is that rank 0 gets stdin.

benchmark/utils/stencil_matrix.hpp
MarcelKoch: This checks only if the point is valid for the given stencil. The actual in-bounds check is in `is_valid_idx`, which is wrong currently, but I have a fix for that. There still seems to be another issue, because in the distributed case it only converges, if the domain is decomposed in 1D. I've yet to find the reason for that.

benchmark/utils/stencil_matrix.hpp
```diff
@@ -0,0 +1,577 @@
...
+        } else {
+            return static_cast<IndexType>(-1);
+        }
+    };
+
+    auto is_valid_neighbor = [&](const IndexType d_i, const IndexType d_j,
+                                 const IndexType d_k) {
+        return !restricted ||
+               ((d_i == 0 && d_j == 0) || (d_i == 0 && d_k == 0) ||
+                (d_j == 0 && d_k == 0));
```
upsj, line 251: we can compute the Manhattan distance directly
```suggestion
               ((d_i == 0) + (d_j == 0) + (d_k == 0) == 1);
```

benchmark/utils/stencil_matrix.hpp
```diff
@@ -0,0 +1,577 @@
...
+    };
+    const auto diag_value = static_cast<ValueType>(nnz_in_row() - 1);
+
+    for (IndexType row = start; row < end; row++) {
+        auto i = row / dp;
+        auto j = row % dp;
+        for (IndexType d_i = -1; d_i <= 1; d_i++) {
+            for (IndexType d_j = -1; d_j <= 1; d_j++) {
+                if (is_valid_neighbor(d_i, d_j)) {
+                    auto col = j + d_j + (i + d_i) * dp;
```
upsj, line 432: also need a bounds check here

benchmark/utils/stencil_matrix.hpp
```diff
@@ -0,0 +1,577 @@
...
+
+    for (IndexType row = start; row < end; row++) {
+        auto i = row / (dp * dp);
+        auto j = (row % (dp * dp)) / dp;
+        auto k = row % dp;
+        for (IndexType d_i = -1; d_i <= 1; d_i++) {
+            for (IndexType d_j = -1; d_j <= 1; d_j++) {
+                for (IndexType d_k = -1; d_k <= 1; d_k++) {
+                    if (is_valid_neighbor(d_i, d_j, d_k)) {
+                        auto col =
```
upsj, line 505: same here

upsj (approved):

benchmark/utils/stencil_matrix.hpp
MarcelKoch: Do you mean besides the one below that?

benchmark/utils/stencil_matrix.hpp
upsj: oops yes, I was thinking of != 0. If you wanted to code-golf this, you could also do 
```suggestion
               (!d_i+ !d_j + !d_k == 2);
```
but that's hard to understand at a glance :smile:

benchmark/utils/stencil_matrix.hpp
upsj: yes, I mean `j + d_j and i + d_i in [0,dim]`

MarcelKoch: you're right, thanks!

benchmark/utils/generator.hpp
```diff
@@ -0,0 +1,299 @@
...
+
+    std::shared_ptr<gko::LinOp> generate_matrix_with_format(
+        std::shared_ptr<gko::Executor> exec, const std::string& format_name,
+        const gko::matrix_data<value_type, index_type>& data,
+        rapidjson::Value* spmv_case = nullptr,
+        rapidjson::MemoryPoolAllocator<>* allocator = nullptr) const
+    {
+        auto part = gko::experimental::distributed::
+            Partition<itype, global_itype>::build_from_global_size_uniform(
+                exec, comm.size(), static_cast<global_itype>(data.size[0]));
```
pratikvn, line 194: Maybe this Partition should be a parameter as well, allowing users to benchmark different kinds of partitions?

benchmark/utils/generator.hpp
```diff
@@ -0,0 +1,299 @@
...
+        return generate_matrix_with_format(std::move(exec), "coo-coo", data);
+    }
+
+    std::unique_ptr<Vec> create_multi_vector(
+        std::shared_ptr<const gko::Executor> exec, gko::dim<2> size,
+        value_type value) const
+    {
+        auto part = gko::experimental::distributed::
+            Partition<itype, global_itype>::build_from_global_size_uniform(
+                exec, comm.size(), static_cast<global_itype>(size[0]));
```
pratikvn, line 230: Same here.

pratikvn (approved):
I really like the generalization of the matrix and vector generate functions. LGTM!

benchmark/utils/stencil_matrix.hpp
```diff
@@ -0,0 +1,431 @@
...
+                }
+            }
+        }
+        return num_neighbors;
+    };
+    const auto diag_value = static_cast<ValueType>(nnz_in_row() - 1);
+
+    for (IndexType i = 0; i < discretization_points; ++i) {
+        for (IndexType j = 0; j < discretization_points; ++j) {
+            auto row = flat_idx(j, i);
```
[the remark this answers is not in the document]
MarcelKoch: `j` is the fastest changing loop index and in `flat_idx` the fastest changing index is the first argument, so this matches up in my weird convention. But I agree, the other way around would make more sense.

benchmark/utils/generator.hpp
MarcelKoch: I think instead of a function parameter, a better way to customize the partition would be to derive from the class, and override this function. I'm not sure how else the partition can be passed along.
So I will leave it in its current state.

MarcelKoch:
format!

MarcelKoch:
format!

PR #1216, opened 2022-11-29
Adds Kokkos example

### LINKED ISSUE #1213: Add example that interfaces with external kernel frameworks
Our `custom-matrix-format` shows how users can use their own kernels together with our library. But it is restricted to an openmp and cuda kernel, which are implemented separately. It would be good to also have an example (or replace the current one) which shows how to integrate ginkgo with kernel frameworks such as Kokkos, Raja, ... This might also be helpful in cases like the discussion #1190.

MarcelKoch:
This PR adds an example that uses Kokkos in combination with `gko::device_matrix_data` to assemble a 3-pt stencil matrix.

Closes #1213

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -53,6 +53,13 @@ if(GINKGO_BUILD_MPI)
     list(APPEND EXAMPLES_LIST distributed-solver)
 endif()
 
+find_package(Kokkos QUIET)
+if(Kokkos_FOUND)
+    list(APPEND EXAMPLES_LIST kokkos_assembly)
+else()
+    message(STATUS "No Kokkos found, disabling examples with Kokkos assembly.")
+endif()
+
 foreach(example ${EXAMPLES_LIST})
     add_subdirectory(${example})
 endforeach()
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -55,6 +55,12 @@ endif()
 
 find_package(Kokkos QUIET)
 if(Kokkos_FOUND)
+    if(GINKGO_WITH_CCACHE)
+        message(WARNING "The CMAKE_CXX_COMPILER_LAUNCHER is set due to "
+            "GINKGO_WITH_CCACHE=ON which is known to casue issues with CUDA enabled "
+            "Kokkos (https://github.com/kokkos/kokkos/issues/4821) including compilation "
+            "failures. This can be prevented by setting GINKGO_WITH_CCACHE=OFF.")
+    endif()
     list(APPEND EXAMPLES_LIST kokkos_assembly)
 else()
     message(STATUS "No Kokkos found, disabling examples with Kokkos assembly.")
--- a/examples/kokkos_assembly/doc/intro.dox
+++ b/examples/kokkos_assembly/doc/intro.dox
@@ -1,0 +1,23 @@
+<a name="Intro"></a>
+<h1>Introduction</h1>
+This example solves a 1D Poisson equation:
+
+@f[
+    u : [0, 1] \rightarrow R\\
+    u'' = f\\
+    u(0) = u0\\
+    u(1) = u1
+@f]
+
+using a finite difference method on an equidistant grid with `K` discretization
+points (`K` can be controlled with a command line parameter).
+
+The resulting CSR matrix is assembled using Kokkos kernels.
+This example show how to use Ginkgo data with Kokkos kernels.
+
+<h3> Notes </h3>
+
+If this example is build as part of Ginkgo, it is advised to configure Ginkgo with
+`-DGINKGO_WITH_CCACHE=OFF` to prevent incompabilities with Kokkos' compiler wrapper
+for `nvcc`.
+
--- a/examples/kokkos_assembly/doc/intro.dox
+++ b/examples/kokkos_assembly/doc/intro.dox
@@ -17,7 +17,7 @@ This example show how to use Ginkgo data with Kokkos kernels.
 
 <h3> Notes </h3>
 
-If this example is build as part of Ginkgo, it is advised to configure Ginkgo with
+If this example is built as part of Ginkgo, it is advised to configure Ginkgo with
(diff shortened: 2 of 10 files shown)


examples/kokkos_assembly/doc/intro.dox
```diff
@@ -0,0 +1,23 @@
...
+
+using a finite difference method on an equidistant grid with `K` discretization
+points (`K` can be controlled with a command line parameter).
+
+The resulting CSR matrix is assembled using Kokkos kernels.
+This example show how to use Ginkgo data with Kokkos kernels.
+
+<h3> Notes </h3>
+
+If this example is build as part of Ginkgo, it is advised to configure Ginkgo with
```
upsj, line 20:
```suggestion
If this example is built as part of Ginkgo, it is advised to configure Ginkgo with
```

examples/kokkos_assembly/kokkos_assembly.cpp
```diff
@@ -0,0 +1,241 @@
...
+{
+    Kokkos::View<const ValueType*> v_u(u->get_const_values(),
+                                       discretization_points);
+    auto error = 0.0;
+    Kokkos::parallel_reduce(
+        "calculate_error", discretization_points,
+        KOKKOS_LAMBDA(int i, double& lsum) {
+            const auto h = 1.0 / (discretization_points + 1);
+            const auto xi = (i + 1) * h;
+            lsum += Kokkos::Experimental::abs(
```
upsj, line 141: What's experimental about computing the magnitude of a value? :thinking:

MarcelKoch:
You can set the used device-id for kokkos via CLI, but I couldn't figure out how to extract that cleanly. So I just set the device id to 0 for ginkgo, which might lead to inconsistencies.

upsj (approved):
LGTM, nice how straightforward this seems (but I'm sure there were some fights with the build system)

examples/kokkos_assembly/kokkos_assembly.cpp
upsj: That sounds really inconvenient. Thanks for the details!

examples/kokkos_assembly/doc/tooltip
```diff
@@ -0,0 +1 @@
+Assembles a CSR matrix with Kokkos kernels.
```
pratikvn, line 1: Kokkos kernels is a separate library, so I think API is better suited here.
```suggestion
Assembles a CSR matrix with Kokkos API.
```

examples/kokkos_assembly/kokkos_assembly.cpp
```diff
@@ -0,0 +1,241 @@
...
+            auto row = static_cast<IndexType>(i / 3);
+            auto col = row + ofs;
+
+            // To prevent branching, a mask is used to set the entry to
+            // zero, if the column is out-of-bounds
+            auto mask =
+                static_cast<IndexType>(0 <= col && col < discretization_points);
+
+            v_row_idxs[i] = mask * static_cast<IndexType>(row);
+            v_col_idxs[i] = mask * static_cast<IndexType>(col);
```
pratikvn, line 77: static_cast's are not really needed here.
```suggestion
            v_row_idxs[i] = mask * row;
            v_col_idxs[i] = mask * col;
```

examples/kokkos_assembly/kokkos_assembly.cpp
```diff
@@ -0,0 +1,241 @@
...
+void print_solution(ValueType u0, ValueType u1,
+                    const gko::matrix::Dense<ValueType>* u)
+{
+    auto host_u = gko::make_temporary_clone(u->get_executor()->get_master(), u);
+    std::cout << u0 << '\n';
+    for (int i = 0; i < u->get_size()[0]; ++i) {
+        std::cout << host_u->get_const_values()[i] << '\n';
+    }
+    std::cout << u1 << std::endl;
+}
```
pratikvn, line 123: Unused, so can probably be removed.

examples/kokkos_assembly/kokkos_assembly.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+    using vec = gko::matrix::Dense<ValueType>;
+    using mtx = gko::matrix::Csr<ValueType, IndexType>;
+    using cg = gko::solver::Cg<ValueType>;
+    using bj = gko::preconditioner::Jacobi<ValueType>;
+
+    // Figure out where to run the code
+    if (argc == 2 && (std::string(argv[1]) == "--help")) {
+        std::cerr << "Usage: " << argv[0]
+                  << " [discretization_points] [kokkos_assembly-options]"
```
pratikvn, line 167: What are the kokkos assembly options ?

pratikvn (approved):
Nice work! Some minor comments. Otherwise LGTM!

examples/kokkos_assembly/kokkos_assembly.cpp
MarcelKoch: The kokkos CLI options. I will add a link to that.

pratikvn: Does that work even if in this case you have `discretization_points` as the first argument ? I assume the kokkos scoped guard uses argc and argv to parse that. If they dont care about the order, then I guess it should be fine.

examples/CMakeLists.txt
```diff
@@ -53,6 +53,13 @@ if(GINKGO_BUILD_MPI)
     list(APPEND EXAMPLES_LIST distributed-solver)
 endif()
 
+find_package(Kokkos QUIET)
+if(Kokkos_FOUND)
+    list(APPEND EXAMPLES_LIST kokkos_assembly)
```
tcojean, line 58: Could you add a check for the ginkgo CCache setting, and print a warning (or disable it?) in that case?

tcojean (approved):
LGTM

examples/CMakeLists.txt
MarcelKoch: I would print a warning and then leave it to the user to disable that.

PR #1217, opened 2022-11-30
Adds MPI timer for benchmark

MarcelKoch:
This PR adds an MPI timer that wraps another timer. The timer will report the longest duration across all processes. This is mainly helpful to enable the `repetitions=auto` setting.


benchmark/utils/timer_impl.hpp
```diff
@@ -195,4 +200,34 @@ class CpuTimer : public Timer {
...
+
+protected:
+    void tic_impl() override { concrete_timer_->tic_impl(); }
+
+    double toc_impl() override
+    {
+        auto local_duration = concrete_timer_->toc_impl();
+        double duration = local_duration;
+        comm_.template all_reduce(exec_, &local_duration, &duration, 1,
+                                  MPI_MAX);
```
upsj, line 221: Would it make sense to make the mode controllable (report maximum or local time)? Some operations we want to benchmark are bulk-synchronous anyways, so any local timer is a decent representation of global time. Will this timer also be used for the runtime breakdown in detailed mode? There this reduction will probably add a huge overhead. How about computing aggregations locally and only reducing afterwards? That would mean extending the timer interface significantly, though.

MarcelKoch: This is to make mpi runs work with the adaptive repetitions setting. We have to use the maximum to make sure that all process actually perform the same number of repetitions. 

Also, the reduction only happens when we actually take the timing, which is not that often, or only once if we set `repetitions` to a fixed value.

I think being able to use `repetitions=auto` is very valuable, because this allows you to specify multiple cases in the input json. Otherwise, you would need to run the benchmark for each case separately. This makes processing the results a bit more annoying. (Also you need to find appropriate repetitions values for each case.)

upsj: Ah right, I misremembered the implementation. Thanks for the clarification!

benchmark/utils/timer_impl.hpp
```diff
@@ -195,4 +200,34 @@ class CpuTimer : public Timer {
...
+    {}
+
+protected:
+    void tic_impl() override { concrete_timer_->tic_impl(); }
+
+    double toc_impl() override
+    {
+        auto local_duration = concrete_timer_->toc_impl();
+        double duration = local_duration;
+        comm_.template all_reduce(exec_, &local_duration, &duration, 1,
```
upsj, line 220: Is this even necessary when you provide no template parameters? It seems like the disambiguation for the parser is no longer necessary.
```suggestion
        comm_.all_reduce(exec_, &local_duration, &duration, 1,
```

benchmark/utils/timer.hpp
```diff
@@ -106,4 +106,25 @@ std::shared_ptr<Timer> get_timer(std::shared_ptr<const gko::Executor> exec,
     return std::make_shared<CpuTimer>(exec);
 }
 
+
+#if GINKGO_BUILD_MPI
+
+
+/**
+ * Get the timer. If the executor does not support gpu timer, still return the
+ * cpu timer.
```
upsj, line 115: We should probably mention that the timer needs to be synchronous across all MPI ranks and returns the maximum.

upsj (approved):
LGTM!

benchmark/utils/timer_impl.hpp
MarcelKoch: No, but sometimes CLion decides to add this...

tcojean (approved):
LGTM.

benchmark/utils/timer_impl.hpp
```diff
@@ -195,4 +200,37 @@ class CpuTimer : public Timer {
 };
 
 
+#if GINKGO_BUILD_MPI
```
yhmtsai, line 203: can it be like what we did for the other timer in hip/cuda/dpcpp?

yhmtsai (approved):
LGTM in general. could we move the implementation to to additional file like other executor?

benchmark/CMakeLists.txt
```diff
@@ -10,13 +10,26 @@ function(ginkgo_benchmark_add_tuning_maybe name)
     endif()
 endfunction()
 
+
+function(ginkgo_benchmark_add_mpi_timer_maybe name)
+    if(GINKGO_BUILD_MPI)
+        if(NOT TARGET mpi_timer)
+            add_library(mpi_timer ${Ginkgo_SOURCE_DIR}/benchmark/utils/mpi_timer.cpp)
+            target_link_libraries(mpi_timer ginkgo)
+        endif()
```
upsj, line 19: I think this should happen unconditionally outside of the function

MarcelKoch:
format!

MarcelKoch:
rebase!

PR #1218, opened 2022-11-30
add simple check on wiki Changelog

yhmtsai:
this PR adds a small check on Changelog from wiki.
If it is not in the wiki, it should give a failure in pipeline.
The entry must be `[#PR_NUMBER](https://https://github.com/ginkgo-project/ginkgo/issues/${PR_NUMBER})` (for the previous release) or `[#PR_NUMBER](https://https://github.com/ginkgo-project/ginkgo/pull/${PR_NUMBER})` 
If the branch name starts with release or the pr contains the label 1:ST:skip-changelog-wiki-check, it will skip the tests
I use the script based on the rest api to get the latest label in the end.
I do not include the comment usage because it is on develop not link the status to the pr.
We can still rerun the action or push new commit after change the label.
I can link it like gitlab did, but I think it's not quite necessary.

I still keep the commits using action file directly to record the following issue faced and solved.
- yaml considers ! as a special character, so in if condition it can not use ! as the first character.
using ${{ expression }} the evaluation to quote the expression to avoid the issue
- On the pull_request event trigger, needs to use `github.event.pull_request` not `github.event.issue` to get the corresponding data. We can get the same things from rest api via pulls/number and issues/number. One more note: we use pull/number for the link but need to use pulls/number in rest api.
- they suggest github.event.pull_request.head.ref should be accessed by environment not directly in if
- In if condition, we can not access the environment variable like $variable but use env.variable to access it. The if condition is on action side not the runner side, so it does not in the bash yet.
- In bash, I use `echo ${env_variable}`, `echo "$env_variable"` and `echo "${env_variable}"`. The first two are replaced by variable context but the last one print plain text `${env_variable}`
- `github.event.pull_request.labels.*.name` does not update latest status if we only change the labels and rerun the action.

TODO:
- [x] check whether it runs as expected (https://github.com/yhmtsai/ginkgo/pull/7 for the label/wiki behavior and https://github.com/yhmtsai/ginkgo/pull/8 for the release branch)
For https://github.com/yhmtsai/ginkgo/pull/7, please go into the check-wiki action, there are four runs on right top.
first one is pr without label and wiki, second is pr with label but without wiki, the latest on is with wiki

### DIFF
--- a/.github/check-wiki-changelog.sh
+++ b/.github/check-wiki-changelog.sh
@@ -2,20 +2,21 @@
 
 source .github/bot-pr-base.sh
 
-echo "Check whether PR contains 1:ST:skip-changelog-wiki-check label"
+echo "Checking whether the PR contains the 1:ST:skip-changelog-wiki-check label"
 SKIP_CHECK=$(api_get "$PR_URL" | jq -r 'any( [.labels | .[] | .name ] | .[] ; . == "1:ST:skip-changelog-wiki-check")')
 
 if [[ "$SKIP_CHECK" == "true" ]]; then
-  echo "The PR contains 1:ST:skip-changelog-wiki-check label, so skip the wiki check"
+  echo "The PR contains the 1:ST:skip-changelog-wiki-check label. Skipping the wiki check."
 elif [[ "$HEAD_BRANCH" =~ ^release.* ]]; then
-  echo "The PR branch name starts with release, so skip the wiki check"
+  echo "The PR branch name starts with release. Skipping the wiki check."
 else
   curl https://raw.githubusercontent.com/wiki/ginkgo-project/ginkgo/Changelog.md > Changelog.md
   PR="\[\#${PR_NUMBER}\]\(https://github.com/ginkgo-project/ginkgo/(issues|pull)/${PR_NUMBER}\)"
   HAS_WIKI="$(cat Changelog.md | grep -E ${PR} || echo false)"
   if [[ "${HAS_WIKI}" == "false" ]]; then
     echo "This PR does not create the corresponding entry in wiki/Changelog"
-    echo "Please adds [#${PR_NUMBER}](https://github.com/ginkgo-project/ginkgo/pull/${PR_NUMBER}) in wiki/Changelog"
+    echo "Please add [#${PR_NUMBER}](https://github.com/ginkgo-project/ginkgo/pull/${PR_NUMBER}) in the wiki/Changelog page"
+    echo "Alternatively, use the label 1:ST:skip-changelog-wiki-check to skip this check."
     exit 1
   else
     echo "wiki/Changelog already has this PR entry."
--- a/.github/check-wiki-changelog.sh
+++ b/.github/check-wiki-changelog.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+
+source .github/bot-pr-base.sh
+
+echo "Check whether PR contains 1:ST:skip-changelog-wiki-check label"
+SKIP_CHECK=$(api_get "$PR_URL" | jq -r 'any( [.labels | .[] | .name ] | .[] ; . == "1:ST:skip-changelog-wiki-check")')
+
+if [[ "$SKIP_CHECK" == "true" ]]; then
+  echo "The PR contains 1:ST:skip-changelog-wiki-check label, so skip the wiki check"
+elif [[ "$HEAD_BRANCH" =~ ^release.* ]]; then
+  echo "The PR branch name starts with release, so skip the wiki check"
+else
+  curl https://raw.githubusercontent.com/wiki/ginkgo-project/ginkgo/Changelog.md > Changelog.md
+  PR="\[\#${PR_NUMBER}\]\(https://github.com/ginkgo-project/ginkgo/(issues|pull)/${PR_NUMBER}\)"
+  HAS_WIKI="$(cat Changelog.md | grep -E ${PR} || echo false)"
+  if [[ "${HAS_WIKI}" == "false" ]]; then
+    echo "This PR does not create the corresponding entry in wiki/Changelog"
+    echo "Please adds [#${PR_NUMBER}](https://github.com/ginkgo-project/ginkgo/pull/${PR_NUMBER}) in wiki/Changelog"
+    exit 1
+  else
+    echo "wiki/Changelog already has this PR entry."
+  fi
+fi
--- a/.github/check-wiki-changelog.sh
+++ b/.github/check-wiki-changelog.sh
@@ -7,7 +7,7 @@ SKIP_CHECK=$(api_get "$PR_URL" | jq -r 'any( [.labels | .[] | .name ] | .[] ; . 
 
 if [[ "$SKIP_CHECK" == "true" ]]; then
   echo "The PR contains the 1:ST:skip-changelog-wiki-check label. Skipping the wiki check."
-elif [[ "$HEAD_BRANCH" =~ ^release.* ]]; then
+elif [[ "$HEAD_BRANCH" =~ ^release-.* ]]; then
   echo "The PR branch name starts with release. Skipping the wiki check."
 else
   curl https://raw.githubusercontent.com/wiki/ginkgo-project/ginkgo/Changelog.md > Changelog.md
--- a/.github/check-wiki-changelog.sh
+++ b/.github/check-wiki-changelog.sh
@@ -2,11 +2,11 @@
 
 source .github/bot-pr-base.sh
 
-echo "Checking whether the PR contains the 1:ST:skip-changelog-wiki-check label"
-SKIP_CHECK=$(api_get "$PR_URL" | jq -r 'any( [.labels | .[] | .name ] | .[] ; . == "1:ST:skip-changelog-wiki-check")')
+echo "Checking whether the PR contains the 1:ST:no-changelog-entry label"
+SKIP_CHECK=$(api_get "$PR_URL" | jq -r 'any( [.labels | .[] | .name ] | .[] ; . == "1:ST:no-changelog-entry")')
 
 if [[ "$SKIP_CHECK" == "true" ]]; then
-  echo "The PR contains the 1:ST:skip-changelog-wiki-check label. Skipping the wiki check."
+  echo "The PR contains the 1:ST:no-changelog-entry label. Skipping the wiki check."
 elif [[ "$HEAD_BRANCH" =~ ^release-.* ]]; then
   echo "The PR branch name starts with release. Skipping the wiki check."
 else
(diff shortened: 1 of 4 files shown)


.github/workflows/bot-pr-updated.yml
```diff
@@ -60,3 +60,12 @@ jobs:
...
             path: abi.diff
+  wiki-check:
+    runs-on: ubuntu-latest
+    steps:
+      - check the number is existed in wiki
+        run: |
+          PR_NUMBER=${{ github.event.pull_request.number }}
+          curl https://raw.githubusercontent.com/wiki/ginkgo-project/ginkgo/Changelog.md > Changelog.md
+          PR=\[\#${PR_NUMBER}\]\(https://github.com/ginkgo-project/ginkgo/(issues|pull)/${PR_NUMBER}\)
+          cat Changelog.md | grep -E ${PR}
```
tcojean, line 71: just make sure to disable it for tags, master and develop

yhmtsai: it's only open on pull request.
I only add exception for the branch `release*` (for the release branch merged into master) and label `1:ST:skip-wiki-check`

.github/workflows/bot-pr-updated.yml
```diff
@@ -60,3 +60,14 @@ jobs:
         with:
             name: abi
             path: abi.diff
+  wiki-check:
```
pratikvn, line 63: nit: A bit more descriptive
```suggestion
  changelog-wiki-check:
```

.github/workflows/bot-pr-updated.yml
```diff
@@ -60,3 +60,14 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: check the number is existed in wiki
+        # Disable it when label contains 1:ST:skip-wiki-check or the branch name starts with release
```
pratikvn, line 67:
```suggestion
        # Disable it when label contains 1:ST:skip-changelog-wiki-check or the branch name starts with release
```

.github/workflows/bot-pr-updated.yml
```diff
@@ -60,3 +60,14 @@ jobs:
...
+        if: ! contains(github.event.issue.labels.*.name, '1:ST:skip-wiki-check') || ! startsWith( ${{ github.event.pull_request.head.name }}, 'release')
```
pratikvn, line 68:
```suggestion
        if: ! contains(github.event.issue.labels.*.name, '1:ST:skip-changelog-wiki-check') || ! startsWith( ${{ github.event.pull_request.head.name }}, 'release')
```

.github/workflows/bot-pr-updated.yml
```diff
@@ -60,3 +60,14 @@ jobs:
         with:
             name: abi
             path: abi.diff
+  wiki-check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: check the number is existed in wiki
```
pratikvn, line 66:
```suggestion
      - name: check if the PR ID exists in wiki
```

.github/workflows/bot-pr-updated.yml
yhmtsai: I use check-wiki-changelog because others also use verb first

.github/check-wiki-changelog.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+source .github/bot-pr-base.sh
+
+echo "Check whether PR contains 1:ST:skip-changelog-wiki-check label"
```
tcojean, line 5:
```suggestion
echo "Checking whether the PR contains the 1:ST:skip-changelog-wiki-check label."
```

.github/check-wiki-changelog.sh
```diff
@@ -0,0 +1,23 @@
...
+  echo "The PR contains 1:ST:skip-changelog-wiki-check label, so skip the wiki check"
+elif [[ "$HEAD_BRANCH" =~ ^release.* ]]; then
+  echo "The PR branch name starts with release, so skip the wiki check"
+else
+  curl https://raw.githubusercontent.com/wiki/ginkgo-project/ginkgo/Changelog.md > Changelog.md
+  PR="\[\#${PR_NUMBER}\]\(https://github.com/ginkgo-project/ginkgo/(issues|pull)/${PR_NUMBER}\)"
+  HAS_WIKI="$(cat Changelog.md | grep -E ${PR} || echo false)"
+  if [[ "${HAS_WIKI}" == "false" ]]; then
+    echo "This PR does not create the corresponding entry in wiki/Changelog"
+    echo "Please adds [#${PR_NUMBER}](https://github.com/ginkgo-project/ginkgo/pull/${PR_NUMBER}) in wiki/Changelog"
```
tcojean, line 18:
```suggestion
    echo "This PR does not create the corresponding entry in wiki/Changelog"
    echo "Please add [#${PR_NUMBER}](https://github.com/ginkgo-project/ginkgo/pull/${PR_NUMBER}) in the wiki/Changelog page"
    echo "Alternatively, use the label 1:ST:skip-changelog-wiki-check to skip this check."
```

.github/check-wiki-changelog.sh
```diff
@@ -0,0 +1,23 @@
...
+
+source .github/bot-pr-base.sh
+
+echo "Check whether PR contains 1:ST:skip-changelog-wiki-check label"
+SKIP_CHECK=$(api_get "$PR_URL" | jq -r 'any( [.labels | .[] | .name ] | .[] ; . == "1:ST:skip-changelog-wiki-check")')
+
+if [[ "$SKIP_CHECK" == "true" ]]; then
+  echo "The PR contains 1:ST:skip-changelog-wiki-check label, so skip the wiki check"
+elif [[ "$HEAD_BRANCH" =~ ^release.* ]]; then
+  echo "The PR branch name starts with release, so skip the wiki check"
```
tcojean, line 11:
```suggestion
  echo "The PR branch name starts with release. Skipping the wiki check."
```

.github/check-wiki-changelog.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+source .github/bot-pr-base.sh
+
+echo "Check whether PR contains 1:ST:skip-changelog-wiki-check label"
+SKIP_CHECK=$(api_get "$PR_URL" | jq -r 'any( [.labels | .[] | .name ] | .[] ; . == "1:ST:skip-changelog-wiki-check")')
+
+if [[ "$SKIP_CHECK" == "true" ]]; then
+  echo "The PR contains 1:ST:skip-changelog-wiki-check label, so skip the wiki check"
```
tcojean, line 9:
```suggestion
  echo "The PR contains the 1:ST:skip-changelog-wiki-check label. Skipping the wiki check."
```

tcojean (approved):
LGTM in general. Some simple comments.

yhmtsai:
@tcojean I also add the unreleased section to wiki page and use the subsection in the previous release

.github/check-wiki-changelog.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+source .github/bot-pr-base.sh
+
+echo "Checking whether the PR contains the 1:ST:skip-changelog-wiki-check label"
+SKIP_CHECK=$(api_get "$PR_URL" | jq -r 'any( [.labels | .[] | .name ] | .[] ; . == "1:ST:skip-changelog-wiki-check")')
+
+if [[ "$SKIP_CHECK" == "true" ]]; then
+  echo "The PR contains the 1:ST:skip-changelog-wiki-check label. Skipping the wiki check."
+elif [[ "$HEAD_BRANCH" =~ ^release.* ]]; then
```
upsj, line 10: nit: I think the naming scheme was always release-1.2.3, so we can be a bit more precise
```suggestion
elif [[ "$HEAD_BRANCH" =~ ^release-.* ]]; then
```

.github/workflows/bot-pr-updated.yml
```diff
@@ -60,3 +60,15 @@ jobs:
...
+    if: github.event.pull_request.author_association == 'COLLABORATOR' || github.event.pull_request.author_association == 'MEMBER' || github.event.pull_request.author_association == 'OWNER'
+    steps:
+    - name: Checkout the latest code (shallow clone)
+      uses: actions/checkout@v2
+      with:
+        ref: develop
+    - name: Check if PR number exists in wiki/Changelog
+      env:
+        GITHUB_TOKEN: ${{ secrets.BOT_TOKEN }}
+      run: cp --preserve .github/check-wiki-changelog.sh /tmp && /tmp/check-wiki-changelog.sh
```
upsj, line 74: Since the check only uses the API and `develop` and doesn't check out the PR's branch, I think we don't need this level of protection.

.github/workflows/bot-pr-updated.yml
```diff
@@ -60,3 +60,15 @@ jobs:
...
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.author_association == 'COLLABORATOR' || github.event.pull_request.author_association == 'MEMBER' || github.event.pull_request.author_association == 'OWNER'
+    steps:
+    - name: Checkout the latest code (shallow clone)
+      uses: actions/checkout@v2
+      with:
+        ref: develop
+    - name: Check if PR number exists in wiki/Changelog
+      env:
+        GITHUB_TOKEN: ${{ secrets.BOT_TOKEN }}
```
upsj, line 73: Do we even use the token?

upsj (approved):
LGTM!

.github/check-wiki-changelog.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+source .github/bot-pr-base.sh
+
+echo "Checking whether the PR contains the 1:ST:skip-changelog-wiki-check label"
```
upsj, line 5: I think it might be more descriptive if the label states its intention, not its effect: 
```suggestion
echo "Checking whether the PR contains the 1:ST:no-changelog-entry label"
```

pratikvn: The label is intended for the action job, so I think it should say what the action job should do, which is to skip the check in the wiki whether the changelog has been added or not.

.github/workflows/bot-pr-updated.yml
yhmtsai: yes, the script gets the information from the pull request using that to get higher usage rate permission.

pratikvn (approved):
LGTM!

.github/check-wiki-changelog.sh
upsj: To me the question is who is the primary intended audience for the label. The script doesn't really care what it is named, but we as developers can profit from a shorter label describing the PR's effect, not what we expect the Action to do.

yhmtsai:
rebase!

.github/check-wiki-changelog.sh
yhmtsai: from the voting in our slack, we choose `1:ST:no-changelog-entry` label

Issue #1222, opened 2022-12-02
Easy to mismatch residual norm criteria and actual norm used by solver


MarcelKoch:
Some of our iterative solvers use the exact residual norm for their stopping criteria, while others use the implicit residual norm. For a user, it is not possible to know which one is actually used. This results in an extra reduction, or an exception. If an exception is thrown, it might be easy to fix, but the extra reduction will just happen silently. 

We should make it easier for users to select the correct criteria. Some solution ideas:
- add residual type (explicit, implicit, none) to solver traits
- introduce new criteria which allows the solver to decide if `ResidualNorm` or `ImplicitResidualNorm` is used. (this should still allow users to explicitly specify either of them)

Here is a list of solvers and their used residual norm. Both Ir and Multigrid use neither.
Check exact residual norm:
- [Cb_]Gmres
- Idr

Check implicit residual norm:
- Bicg
- Bicgstab
- Cg
- Cgs
- Fcg

yhmtsai:
Any solver can use the ResidualNorm as the stop criterion.
If the solver does not have it during solving, ResidualNorm will introduce an additional spmv and norm as requests.
Thus, solver does not always have this additional operation when they do not need residual norm during operation.

My concern about the automatic ResidualNorm/ImplicitResidualNorm is that it will introduce another confusion out of expectation.
When user use it on Gmres/CbGmres, they may think they get the actual residual reduction. However, when they use it on Cg, they do not get the actual residual norm from their expectation.

In the original settings, if they want to use ResidualNormReduction as stop criterion, they can use it for all solver.
When they use ImplicitResidualNormReduction, they should know it checks the implicit residual norm different from the residual norm.

PR #1224, opened 2022-12-02
Automatic residual criterion

MarcelKoch:
This is a starting point for discussing solutions to #1222.

This PR adds a new stopping criterion `AutomaticResidualNorm` (name subject to change, perhaps Default...?), wrapping either `ResidualNorm` or `ImplicitResidualNorm`. Which of those two gets used depends on what is available by the solver. The aim is to by default reuse the already computed norms or dot products.


MarcelKoch:
I won't work on that right now, so I will retire it for now.

PR #1226, opened 2022-12-05
Updates git-cmake-format

### LINKED ISSUE #1223: pre-commit hook propagates into higher level git project
When using Ginkgo as a submodule the pre-commit hook is propagated into the main git repository. 
This especially applies to the [dummy hook](/ginkgo-project/ginkgo/blob/develop/third_party/dummy-hook/CMakeLists.txt), which blocks commiting to the repository, when setting the development flag for ginkgo. 

Currently I fix this in the M++ CMake file by removing the added pre-commit hook again after the script adds it using the same process for finding the git hooks location.

The script for finding the git hooks location should point to ginkgos git hooks to not block the higher level repository.

MarcelKoch:
This PR updates git-cmake-format to include the latest fix for setting the git hooks directory. This also fixes the directory for the dummy hook.

Fixes #1223.

Todo:
- [x] wait for https://github.com/ginkgo-project/git-cmake-format/pull/11

### DIFF
--- a/third_party/git-cmake-format/CMakeLists.txt
+++ b/third_party/git-cmake-format/CMakeLists.txt
@@ -1,16 +1,15 @@
 message(STATUS "Fetching git-cmake-format")
 include(FetchContent)
 FetchContent_Declare(
-    git_cmake_format
-    GIT_REPOSITORY https://github.com/ginkgo-project/git-cmake-format.git
-    GIT_TAG        e9a82f20d36f1a92d4c52910a8d0a694e7673c54
+        git_cmake_format
+        GIT_REPOSITORY https://github.com/ginkgo-project/git-cmake-format.git
+        GIT_TAG customize-toplevel-dir
 )
 FetchContent_GetProperties(git_cmake_format)
 if(NOT git_cmake_format_POPULATED)
     FetchContent_Populate(git_cmake_format)
 
-    if(NOT DEFINED GCF_FORCE_OVERWRITE)
-        set(GCF_FORCE_OVERWRITE ON CACHE INTERNAL "If true, always overwrite pre-commit hook and script")
-    endif()
+    set(GCF_FORCE_OVERWRITE ON CACHE INTERNAL "If true, always overwrite pre-commit hook and script")
+    set(GCF_GIT_TOP_LEVEL_DIR ${Ginkgo_SOURCE_DIR})
     add_subdirectory(${git_cmake_format_SOURCE_DIR} ${git_cmake_format_BINARY_DIR} EXCLUDE_FROM_ALL)
 endif()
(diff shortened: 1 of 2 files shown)


third_party/git-cmake-format/CMakeLists.txt
```diff
@@ -1,16 +1,15 @@
 message(STATUS "Fetching git-cmake-format")
 include(FetchContent)
 FetchContent_Declare(
-    git_cmake_format
-    GIT_REPOSITORY https://github.com/ginkgo-project/git-cmake-format.git
-    GIT_TAG        e9a82f20d36f1a92d4c52910a8d0a694e7673c54
+        git_cmake_format
+        GIT_REPOSITORY https://github.com/ginkgo-project/git-cmake-format.git
+        GIT_TAG customize-toplevel-dir
```
thoasm, line 6: I assume this is the reason why it is marked as `WIP` (meaning you will revert this back to either the new checksum or `master`)?

thoasm (approved):
LGTM!

third_party/git-cmake-format/CMakeLists.txt
MarcelKoch: Exactly, but I wanted to allow testing it in #1223

third_party/git-cmake-format/CMakeLists.txt
```diff
@@ -1,16 +1,15 @@
...
+        GIT_TAG 966837ee83f2b3b7df9acf59e9d801fcd189dd9b
 )
 FetchContent_GetProperties(git_cmake_format)
 if(NOT git_cmake_format_POPULATED)
     FetchContent_Populate(git_cmake_format)
 
-    if(NOT DEFINED GCF_FORCE_OVERWRITE)
-        set(GCF_FORCE_OVERWRITE ON CACHE INTERNAL "If true, always overwrite pre-commit hook and script")
-    endif()
+    set(GCF_FORCE_OVERWRITE ON CACHE INTERNAL "If true, always overwrite pre-commit hook and script")
```
yhmtsai, line 12: does it need to be always on?

yhmtsai: If there's no GCF_FORCE_OVERWRITE before, it always sets it on.
It's off when user specifies it, right?

MarcelKoch: You're, right. I will use `option` for this. It should be more concise.

yhmtsai: not sure whether you need to use an option for it. keeping it as the original develop looks good to me

upsj (approved):

MarcelKoch:
rebase!

third_party/git-cmake-format/CMakeLists.txt
MarcelKoch: Option does exactly the same, but in 1 line instead of 3.

yhmtsai: I see

yhmtsai (approved):
LGTM

MarcelKoch:
rebase!

PR #1227, opened 2022-12-05
Enables building Partition from local ranges

MarcelKoch:
This PR adds a freestanding function to create a `Partition` from local ranges. This has to be a freestanding function, because adding it to `Partition` would require including MPI, which leads to our device libraries requiring MPI (or at least the headers).


include/ginkgo/core/distributed/partition_helpers.hpp
```diff
@@ -0,0 +1,80 @@
...
+ *          that for a process `i` with `range[i] = [s_i, e_i)` then for process
+ *          `j = i+1` `range[j] = [s_j = e_i, e_j)`.
+ *
+ * @return a Partition where each range has the individual local_start
+ *         and local_ends.
+ */
+template <typename LocalIndexType, typename GlobalIndexType>
+std::unique_ptr<Partition<LocalIndexType, GlobalIndexType>>
+build_partition_from_local_range(std::shared_ptr<const Executor> exec,
+                                 span local_range, mpi::communicator comm);
```
upsj, line 71: local range or local size? The latter would be harder to misuse

core/distributed/partition.cpp
```diff
@@ -32,7 +32,6 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include <ginkgo/core/distributed/partition.hpp>
 
-
```
upsj, line 35: this should probably remain

test/mpi/distributed/partition_helpers.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+
+template<typename IndexType>
+class PartitionHelpers : public CommonMpiTestFixture{
+protected:
+    using index_type = IndexType;
+
+};
+
+TYPED_TEST_SUITE(PartitionHelpers, gko::test::IndexTypes);
```
upsj, line 16: You can probably add a typename type printer here

include/ginkgo/core/distributed/partition_helpers.hpp
```diff
@@ -0,0 +1,80 @@
...
+HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+#ifndef GKO_PUBLIC_CORE_DISTRIBUTED_PARTITION_HELPERS_HPP_
```
upsj, line 33: Could we also put this into partition.hpp directly?

core/distributed/partition_helpers.cpp
```diff
@@ -0,0 +1,92 @@
...
+}
+}  // namespace partition_helpers
+
+
+template <typename LocalIndexType, typename GlobalIndexType>
+std::unique_ptr<Partition<LocalIndexType, GlobalIndexType>>
+build_partition_from_local_range(std::shared_ptr<const Executor> exec,
+                                 span local_range, mpi::communicator comm)
+{
+    GlobalIndexType range[2] = {static_cast<GlobalIndexType>(local_range.begin),
```
upsj, line 60: nit: std::array

upsj (changes requested):
The main thing that comes to mind is the issue of correctness: The current model doesn't allow feedback on whether the ranges are actually consecutive, right? If they aren't, we just pick a handful of values from the range array, which may or may not be in ascending order. What you could do is check via run_kernel_reduce over a boolean whether the input was consecutive with little overhead, or do the check on the host side after a host-side allgather.

include/ginkgo/core/distributed/partition_helpers.hpp
MarcelKoch: No because it would require to link all device libraries against MPI (or at least include the MPI header)

MarcelKoch: yes, I think there are a few where that happens, mostly the read_distributed ones.

include/ginkgo/core/distributed/partition_helpers.hpp
MarcelKoch: I think we can add both. The local size would also require a part-id. I would still keep the local-range one, but make it more robust with your suggestion.

include/ginkgo/core/distributed/partition_helpers.hpp
```diff
@@ -0,0 +1,99 @@
...
+template <typename LocalIndexType, typename GlobalIndexType>
+class Partition;
+
+
+/**
+ * Builds a partition from a local range.
+ *
+ * @param exec  the Executor on which the partition should be built.
+ * @param local_range the start and end indices of the local range.
+ * @param comm  the communicator used to determine the global partition.
```
greole, line 60:
```suggestion
 * @param comm  the communicator used to determine the global partition.
 * @param local_range the start and end indices of the local range.
```
To reflect the order of the arguments.

include/ginkgo/core/distributed/partition_helpers.hpp
```diff
@@ -0,0 +1,99 @@
...
+build_partition_from_local_range(std::shared_ptr<const Executor> exec,
+                                 mpi::communicator comm, span local_range);
+
+
+/**
+ * Builds a partition from a local size.
+ *
+ * @param exec  the Executor on which the partition should be built.
+ * @param local_range the number of the locally owned indices
+ * @param comm  the communicator used to determine the global partition.
```
greole, line 80:
```suggestion
 * @param comm  the communicator used to determine the global partition.
 * @param local_range the number of the locally owned indices
```

core/distributed/partition_helpers.cpp
```diff
@@ -0,0 +1,156 @@
...
+}  // namespace
+}  // namespace partition_helpers
+
+
+template <typename LocalIndexType, typename GlobalIndexType>
+std::unique_ptr<Partition<LocalIndexType, GlobalIndexType>>
+build_partition_from_local_range(std::shared_ptr<const Executor> exec,
+                                 mpi::communicator comm, span local_range)
+{
+    std::array<GlobalIndexType, 2> range{
```
greole, line 76: Maybe a naive question, but couldn't this just forward to `build_partition_from_local_size` since we state that indices need to be in ascending order?

include/ginkgo/core/distributed/partition.hpp
```diff
@@ -83,9 +83,9 @@ namespace distributed {
  * ```
  * starting_index[0] = 0,
  * starting_index[1] = 0,
- * starting_index[2] = 3,  // second range of part 1
+ * starting_index[2] = 3,  // second range of part 0
  * starting_index[3] = 0,
- * starting_index[4] = 5,  // third range of part 1
+ * starting_index[4] = 5,  // third range of part 0
  * ```
  * which you can use to iterate only over the the second range of part 1 (the
```
pratikvn, line 90: Also here :)

core/distributed/partition_helpers.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+    // make all range_start_ends available on each rank
+    auto mpi_exec = exec->get_master();
+    array<GlobalIndexType> ranges_start_end(mpi_exec, comm.size() * 2);
+    ranges_start_end.fill(invalid_index<GlobalIndexType>());
+    std::vector<mpi::request> reqs;
+    reqs.push_back(comm.i_all_gather(mpi_exec, &range[0], 1,
+                                     ranges_start_end.get_data(), 1));
+    reqs.push_back(comm.i_all_gather(
+        mpi_exec, &range[1], 1, ranges_start_end.get_data() + comm.size(), 1));
```
upsj, line 88: why this layout and not communicate the two entries together by interleaving?

core/distributed/partition_helpers.cpp
```diff
@@ -0,0 +1,156 @@
...
+                                              part_ids.get_num_elems()));
+    exec->run(partition_helpers::make_sort_by_range_start(ranges_start_end,
+                                                          part_ids));
+
+    // check for consistency
+    bool consecutive_ranges = false;
+    exec->run(partition_helpers::make_check_consecutive_ranges(
+        ranges_start_end, &consecutive_ranges));
+    if (!consecutive_ranges) {
+        throw Error(__FILE__, __LINE__, "The partition contains gaps.");
```
upsj, line 104: We need a more specific error for invalid user data, I think

core/distributed/partition_helpers.cpp
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    // remove duplicates
```
upsj, line 107: To me, this description of what happens below could be improved, e.g. 
```suggestion
    // join (now consecutive) starts and ends into combined array
```

upsj (commented):
First look, still need to go over the tests

core/distributed/partition_helpers.cpp
MarcelKoch: I would like to have it interleaved, but then I would need iterators with stride=2, which I don't think thrust support. Or is there some way to do that with thrust?

upsj: Something like a counting_iterator + transform_iterator? That's my go-to solution for non-trivial access patterns

core/distributed/partition_helpers.cpp
MarcelKoch: I agree, but I would rather put that into another PR.

reference/distributed/partition_helpers_kernels.cpp
```diff
@@ -0,0 +1,115 @@
...
+{
+    struct range {
+        GlobalIndexType idxs[2];
+    };
+
+    auto part_ids_d = part_ids.get_data();
+    auto num_parts = part_ids.get_num_elems();
+    auto range_it = reinterpret_cast<range*>(range_start_ends.get_data());
+    auto sort_it = detail::make_zip_iterator(range_it, part_ids_d);
+    std::sort(sort_it, sort_it + num_parts, [](const auto& a, const auto& b) {
```
greole, line 60: Currently, partitions with zero size parts are legal but this might fail to produce a valid sorting, maybe change to
```suggestion
    std::stable_sort(sort_it, sort_it + num_parts, [](const auto& a, const auto& b) {
```
A robust solution would also be needed for the repartitioned case where zero size parts might arise.

MarcelKoch: good point. I will check if that is also necessary for the other kernels

pratikvn:
needs to be rebased ?

common/unified/distributed/partition_helpers_kernels.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+template <typename GlobalIndexType>
+void check_consecutive_ranges(std::shared_ptr<const DefaultExecutor> exec,
+                              const array<GlobalIndexType>& range_start_ends,
+                              bool* result)
+{
+    array<uint32> result_uint32{exec, 1};
+    auto num_ranges = range_start_ends.get_num_elems() / 2;
+    // need additional guard because DPCPP doesn't return the initial value for
+    // empty inputs
```
upsj, line 55: we should fix this

reference/distributed/partition_helpers_kernels.cpp
```diff
@@ -0,0 +1,116 @@
...
+    };
+
+    auto part_ids_d = part_ids.get_data();
+    auto num_parts = part_ids.get_num_elems();
+    auto range_it = reinterpret_cast<range*>(range_start_ends.get_data());
+    auto sort_it = detail::make_zip_iterator(range_it, part_ids_d);
+    std::stable_sort(sort_it, sort_it + num_parts,
+                     [](const auto& a, const auto& b) {
+                         return std::get<0>(a).idxs[0] < std::get<0>(b).idxs[0];
+                     });
```
upsj, line 63: This seems really unsafe, I'd prefer sorting a permutation array with a comparator that looks into range_start_ends.

reference/distributed/partition_helpers_kernels.cpp
```diff
@@ -0,0 +1,116 @@
...
+    auto range_it = reinterpret_cast<const end_start*>(
+        range_start_ends.get_const_data() + 1);
+
+    if (num_parts) {
+        *result =
+            std::all_of(range_it, range_it + num_parts - 1,
+                        [](const end_start& r) { return r.end == r.start; });
+    } else {
+        *result = true;
+    }
```
upsj, line 90: I think this would be safer to do with a raw loop, the reinterpret-casting assumes the struct is not padded, which is implementation-defined.

omp/distributed/partition_helpers_kernels.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+template <typename GlobalIndexType>
+void sort_by_range_start(
+    std::shared_ptr<const DefaultExecutor> exec,
+    array<GlobalIndexType>& range_start_ends,
+    array<experimental::distributed::comm_index_type>& part_ids)
+{
+    struct range {
+        GlobalIndexType idxs[2];
+    };
```
upsj, line 54: same here: unsafe

upsj (changes requested):
LGTM in general, but there are some unsafe operations in the CPU implementations.

common/unified/distributed/partition_helpers_kernels.cpp
MarcelKoch: @yhmtsai I think you worked on the dpcpp reduction. Does it now return the initial value for empty inputs?

core/base/iterator_factory.hpp
```diff
@@ -366,6 +366,130 @@ void swap(zip_iterator_reference<Iterators...> a,
...
+
+    explicit permute_iterator(IteratorType it, PermuteFn perm)
+        : it_{std::move(it)}, idx_{}, perm_{std::move(perm)}
+    {}
+
+    permute_iterator& operator=(permute_iterator other)
+    {
+        it_ = other.it_;
+        idx_ = other.idx_;
+        // no perm_ = other.perm_ because lambdas are not copy-assignable
```
upsj, line 388: I think just to be safe, I would prefer adding this line and using functor structs in the tests.

test/distributed/partition_helper_kernels.cpp
```diff
@@ -0,0 +1,279 @@
...
+
+    ASSERT_FALSE(result);
+}
+
+
+ TYPED_TEST(PartitionHelpers, CanCheckConsecutiveRangesWithSingleRange)
+{
+    using index_type = typename TestFixture::index_type;
+    auto start_ends =
+        make_array(this->ref, create_ranges<index_type>(1));
```
thoasm, line 206: This should not work properly (since `begin+1 < end - 1`).

thoasm (commented):
Mostly LGTM with the exception of the two `reinterpret_cast` usages.

core/base/iterator_factory.hpp
MarcelKoch: Just for completion, lambdas have the copy assignment defined in c++20 (if they don't capture). Also, the copy constructor is still defined. So currently I'm hacking this by wrapping the lambda in a std::vector.

dpcpp/distributed/partition_helpers_kernels.dp.cpp
```diff
@@ -0,0 +1,91 @@
...
+HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+// force-top: on
```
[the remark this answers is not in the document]
MarcelKoch: to prevent clang-format from moving these includes around. There seems to be issues otherwise.

test/distributed/partition_helper_kernels.cpp
MarcelKoch: What part should not work? This creates a temporary vector with two elements, so `begin + 1 == end - 1`.

thoasm: My bad, I thought it would create a 1-element array (number of arguments) instead of 1+1.

core/distributed/partition_helpers.cpp
```diff
@@ -0,0 +1,153 @@
...
+    GKO_DECLARE_BUILD_PARTITION_FROM_LOCAL_RANGE);
+
+
+template <typename LocalIndexType, typename GlobalIndexType>
+std::unique_ptr<Partition<LocalIndexType, GlobalIndexType>>
+build_partition_from_local_size(std::shared_ptr<const Executor> exec,
+                                mpi::communicator comm, size_type local_size)
+{
+    auto local_size_gi = static_cast<GlobalIndexType>(local_size);
+    std::vector<GlobalIndexType> sizes(comm.size());
```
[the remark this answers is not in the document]
thoasm: I agree that it is more comfortable, but we avoid our loggers with `std::vector`. And since you don't need resizing or other more advanced features of `std::vector`, using `gko::array` seems more logical to me.

thoasm (approved):
LGTM.
Thanks for removing the `reinterpret_cast`!

core/base/iterator_factory.hpp
MarcelKoch: We need copy assign, not copy construct. Lambdas have a default defined copy constructor, but their copy assign is deleted (unless in c++20), and we need here the assign. And by using `vector::emplace` I can use the copy constructor of the lambda.
I think using the vector results in better assembly (with function: https://godbolt.org/z/znG889nPj, with vector: https://godbolt.org/z/dxG7q7r4P) and since this might also be used by our omp backend, I would use the vector. I could also wrap the vector construction in a type (e.g. `copy_assignable`) to make it clearer what is happening.

thoasm: > Lambdas have a default defined copy constructor, but their copy assign is deleted (unless in c++20), and we need here the assign.

Thanks for the explanation, I didn't know that.
With the name `copy_assignable` and a comment mentioning that copy assign is deleted on lambdas, I am fine with your vector implementation.

However, note that your `std::function` implementation is very likely much faster than the `std::vector` implementation as you don't call `new` and `delete`.

MarcelKoch: The new and delete is only once per iterator creation, but in the function case, you have function calls within the loop, which are not present in the vector case. But I will benchmark it to see if there are differences.

MarcelKoch: To complete this, here is a quick benchmark: https://quick-bench.com/q/i8UiekPceUwmRSbAMOeBQEbzqvE (choose the bar visualization center right). The vector approach is indeed faster, but still takes ~2x than directly increasing the iterator by two.

core/base/iterator_factory.hpp
pratikvn, line 1: I think given that the class is not called `iterator_factory` anymore, we should also consider renaming this file.

test/distributed/partition_helper_kernels.cpp
```diff
@@ -0,0 +1,277 @@
...
+        result.first.begin(), [](const auto i) { return 2 * i; });
+    auto range_end_it = gko::detail::make_permute_iterator(
+        result.first.begin() + 1, [](const auto i) { return 2 * i; });
+    auto zip_it = gko::detail::make_zip_iterator(range_start_it, range_end_it,
+                                                 result.second.begin());
+    std::shuffle(zip_it, zip_it + num_ranges, engine);
+
+    return result;
+}
+
```
pratikvn, line 162:
```suggestion


```

pratikvn (approved):
Nice work! LGTM! Some minor nits.

core/base/iterator_factory.hpp
MarcelKoch: I guess `iterator_factories.hpp` would be better. But I think that is too small of a difference.

MarcelKoch:
format-rebase!

MarcelKoch:
format!

greole (approved):
LGTM

core/base/copy_assignable.hpp
```diff
@@ -0,0 +1,94 @@
...
+    {
+        return obj_[0](std::forward<Args>(args)...);
+    }
+
+    T const& get() const { return obj_[0]; }
+    T& get() { return obj_[0]; }
+
+private:
+    //!< Store wrapped object in a container that has an emplace function
+    std::vector<T> obj_;
```
upsj, line 87: This needs an allocation on copy - I wouldn't expect good performance if we use it for iterators. Should be more like std::optional. I think for now, we should instead use hand-written `struct` functors instead of lambdas. placement-new could be used to handle the copy-construction

test/mpi/distributed/partition_helpers.cpp
```diff
@@ -0,0 +1,135 @@
...
+    gko::span local_range[] = {{4u, 6u}, {9u, 11u}, {0u, 4u}};
+    // Hack because of multiple template arguments in macro
+    auto build_from_local_ranges = [](auto... args) {
+        return gko::experimental::distributed::build_partition_from_local_range<
+            gko::int32, itype>(args...);
+    };
+
+    ASSERT_THROW(build_from_local_ranges(this->exec, this->comm,
+                                         local_range[this->comm.rank()]),
+                 gko::Error);
```
upsj, line 112: InvalidState?

upsj (changes requested):
One thing I would like to see changed, otherwise seems fine

core/base/copy_assignable.hpp
MarcelKoch: In this thread there is already a performance discussion: https://github.com/ginkgo-project/ginkgo/pull/1227#discussion_r1170924286.  I would gladly use `std::optional` when we have c++17.

upsj: The benchmark doesn't use copy-assignment, if you run std::sort, the difference becomes visible
https://quick-bench.com/q/Fo981WOwE6ktI8rsA4kAn-BCTwM

MarcelKoch: Placement-new is significantly faster than the vector, but still 2x slower than optional: https://quick-bench.com/q/sdaz3zrJ9ITINhIx15Belvn2Osw. Also, the assembly doesn't need any allocation: https://godbolt.org/z/zxW8TYb9o. So, I would go with this, until we can replace it with std::optional.

upsj: sounds good, thanks :)

common/unified/distributed/partition_helpers_kernels.cpp
yhmtsai: I wrap it outside

MarcelKoch:
format!

dpcpp/distributed/partition_helpers_kernels.dp.cpp
```diff
@@ -74,11 +74,31 @@ void sort_by_range_start(
         range_start_ends.get_data(), stride{});
     auto end_it = oneapi::dpl::make_permutation_iterator(
         range_start_ends.get_data() + 1, stride{});
+
+    // older versions of oneDPL have a bug when sorting permutation iterators
+#if ONEDPL_VERSION_MAJOR >= 2022 && ONEDPL_VERSION_MINOR >= 1
```
upsj, line 79: I think we should probably move towards a stable version of oneDPL soon, then we could remove this workaround.

core/base/copy_assignable.hpp
```diff
@@ -0,0 +1,130 @@
...
+        return obj_[0](std::forward<Args>(args)...);
+    }
+
+    T const& get() const { return obj_[0]; }
+
+    T& get() { return obj_[0]; }
+
+private:
+    //!< Store wrapped object on the stack, should use std::optional in c++17
+    T* obj_;
```
upsj, line 122: nit: the pointer has not yet been initialized, assigning to it is UB (even though it would probably never fail)
```suggestion
    T* obj_{};
```

core/base/copy_assignable.hpp
```diff
@@ -0,0 +1,130 @@
...
+
+    template <typename... Args>
+    decltype(auto) operator()(Args&&... args) const
+    {
+        return obj_[0](std::forward<Args>(args)...);
+    }
+
+    T const& get() const { return obj_[0]; }
+
+    T& get() { return obj_[0]; }
```
upsj, line 118: nit: `obj_` is not an array, indexing into it seems weird.
```suggestion
    decltype(auto) operator()(Args&&... args) const
    {
        return (*obj_)(std::forward<Args>(args)...);
    }

    T const& get() const { return *obj_; }

    T& get() { return *obj_; }
```

test/distributed/partition_helper_kernels.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+TYPED_TEST(PartitionHelpers, CanCheckConsecutiveRangesWithSingleElement)
+{
+    using index_type = typename TestFixture::index_type;
+    auto start_ends = gko::array<index_type>(this->exec, {1});
+    bool result = false;
+
+    gko::kernels::EXEC_NAMESPACE::partition_helpers::check_consecutive_ranges(
+        this->exec, start_ends, &result);
```
upsj, line 220: nit: why are we taking a pointer and not a reference for this function?

upsj (approved):
LGTM!

dpcpp/distributed/partition_helpers_kernels.dp.cpp
MarcelKoch: We are still testing with and older version, so let's keep it.

MarcelKoch:
rebase!

MarcelKoch:
rebase!

MarcelKoch:
rebase!

PR #1228, opened 2022-12-05
Allows checking if host MPI buffers are required at runtime

MarcelKoch:
This PR adds the option to use host MPI buffers even if `GINKGO_FORCE_GPU_AWARE_MPI` is true. The option can be set in the `communicator` constructor. Additionally, a helper function is added that checks if the combination of executor and communicator needs to use host buffers.

### DIFF
--- a/include/ginkgo/core/base/mpi.hpp
+++ b/include/ginkgo/core/base/mpi.hpp
@@ -1505,6 +1505,14 @@ private:
 
 
 /**
+ * Checks if the combination of Executor and communicator requires passing
+ * MPI buffers from the host memory.
+ */
+bool requires_host_buffer(const std::shared_ptr<const Executor>& exec,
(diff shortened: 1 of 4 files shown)


greole (approved):
LGTM

thoasm (approved):
LGTM!

MarcelKoch:
format-rebase!

include/ginkgo/core/base/mpi.hpp
```diff
@@ -443,7 +443,8 @@ class communicator {
      *
      * @param comm The input MPI_Comm object.
```
upsj, line 444:
```suggestion
     * @param comm The input MPI_Comm object.
     * @param force_host_buffer If set to true, always communicates through host memory
```

MarcelKoch:
format!

PR #1229, opened 2022-12-05
Temporarily disables nla-gpu tests or moves them to other systems

MarcelKoch:
I propose these changes to temporarily deal with nla-gpu not being available. I think we will only miss an MPI + ROCm test, but that seems unavoidable.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -472,21 +472,22 @@ build/amd/nompi/gcc/rocm40/debug/shared:
     BUILD_TYPE: "Debug"
     FAST_TESTS: "ON"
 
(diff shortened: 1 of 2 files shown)


yhmtsai (approved):
https://gitlab.com/gitlab-org/gitlab/-/issues/229851 is related to the job is stuck forever when self-host runner is disconnected

MarcelKoch:
@yhmtsai I don't think that is related to our case. When we start the job, the runner is already not available, while they seem to remove the runner while the job is running.

yhmtsai:
you're right, I thought it's similar originally because our job is also pending forever.

thoasm (approved):
LGTM!

Issue #1036, opened 2022-05-03
Add support for profiler hooks in logger


upsj:
Tools like [TAU](https://www.cs.uoregon.edu/research/tau/docs/newguide/bk05rn01.html) or [NVIDIA Toolkit Extensions](https://docs.nvidia.com/cuda/profiler-users-guide/index.html#nvtx) allow applications to log custom events/phases in a nested fashion. We could profit from their tooling by providing loggers wrapping this functionality.

pratikvn:
This is definitely a good idea. I think we already had some effort from @tcojean using PAPI with software defined events.

Additionally, there are a few libraries that can abstract this for different backends as well. This would allow us to for example to have a common instrumentation functions in our code and have that work for different backends that the library uses. Just to mention some libraries which might be useful for this:

1. Tau,
2. Score-P
3. Caliper.

Of course, we always have the option of using the profiler API that Nvidia and ROCM (and DPC++ ? ) provide, but that might be more work and maintenance.

PR #1232, opened 2022-12-08
Add support for host operations to executor

### LINKED ISSUE #1113: Host operations get lost in benchmarks
ATM, we only attach loggers to the device executor in our benchmarks. That means that host operations like they are used in RCM and symbolic Cholesky are missing entirely from the `components` breakdown. This should be fixed to provide more comprehensive logging.

Possible approaches:
1. modify executor to add all loggers also to the host executor
2. modify benchmarks to add all loggers also to the host executor

### LINKED ISSUE #1211: Add `HostOperation` to enable profiling sequential core functionality
I think it would make sense to be able to wrap a host-side function in `core` that does significant work, but is not necessarily suitable for execution on the device (e.g. elimination tree computation, symbolic LU, ...) in an `Operation` for it to show up in profiles. For this to be useful in our benchmarks, we should probably address #1113 as well.

upsj:
This adds the `GKO_REGISTER_HOST_OPERATION` macro which can be used to annotate a costly `core`-side computation for benchmarking and (later on, once #1036 is merged) profiling. Additionally, it adds the `OperationLogger` to the host executor in GPU executions, which will make host-side operations visible in detailed benchmark results.

TODO:
- [x] Add tests

Closes #1211
Closes #1113

### DIFF
--- a/core/factorization/symbolic.cpp
+++ b/core/factorization/symbolic.cpp
@@ -58,6 +58,7 @@ GKO_REGISTER_OPERATION(cholesky_symbolic,
 GKO_REGISTER_OPERATION(prefix_sum, components::prefix_sum);
 GKO_REGISTER_OPERATION(initialize, lu_factorization::initialize);
 GKO_REGISTER_OPERATION(factorize, lu_factorization::factorize);
+GKO_REGISTER_HOST_OPERATION(compute_elim_forest, compute_elim_forest);
 
 
 }  // namespace
@@ -65,38 +66,40 @@ GKO_REGISTER_OPERATION(factorize, lu_factorization::factorize);
 
 /** Computes the symbolic Cholesky factorization of the given matrix. */
 template <typename ValueType, typename IndexType>
-std::unique_ptr<matrix::Csr<ValueType, IndexType>> symbolic_cholesky(
-    const matrix::Csr<ValueType, IndexType>* mtx)
+void symbolic_cholesky(
+    const matrix::Csr<ValueType, IndexType>* mtx,
+    std::unique_ptr<matrix::Csr<ValueType, IndexType>>& factors)
 {
     using matrix_type = matrix::Csr<ValueType, IndexType>;
     const auto exec = mtx->get_executor();
     const auto host_exec = exec->get_master();
-    const auto forest = compute_elim_forest(mtx);
+    std::unique_ptr<elimination_forest<IndexType>> forest;
+    exec->run(make_compute_elim_forest(mtx, forest));
     const auto num_rows = mtx->get_size()[0];
     array<IndexType> row_ptrs{exec, num_rows + 1};
     array<IndexType> tmp{exec};
     exec->run(
-        make_cholesky_symbolic_count(mtx, forest, row_ptrs.get_data(), tmp));
+        make_cholesky_symbolic_count(mtx, *forest, row_ptrs.get_data(), tmp));
     exec->run(make_prefix_sum(row_ptrs.get_data(), num_rows + 1));
     const auto factor_nnz = static_cast<size_type>(
         exec->copy_val_to_host(row_ptrs.get_const_data() + num_rows));
-    auto factor = matrix_type::create(
+    factors = matrix_type::create(
         exec, mtx->get_size(), array<ValueType>{exec, factor_nnz},
         array<IndexType>{exec, factor_nnz}, std::move(row_ptrs));
-    exec->run(make_cholesky_symbolic(mtx, forest, factor.get(), tmp));
-    factor->sort_by_column_index();
-    auto lt_factor = as<matrix_type>(factor->transpose());
+    exec->run(make_cholesky_symbolic(mtx, *forest, factors.get(), tmp));
+    factors->sort_by_column_index();
+    auto lt_factor = as<matrix_type>(factors->transpose());
     const auto scalar =
         initialize<matrix::Dense<ValueType>>({one<ValueType>()}, exec);
     const auto id = matrix::Identity<ValueType>::create(exec, num_rows);
-    lt_factor->apply(scalar.get(), id.get(), scalar.get(), factor.get());
-    return factor;
+    lt_factor->apply(scalar.get(), id.get(), scalar.get(), factors.get());
 }
 
 
-#define GKO_DECLARE_SYMBOLIC_CHOLESKY(ValueType, IndexType)               \
-    std::unique_ptr<matrix::Csr<ValueType, IndexType>> symbolic_cholesky( \
-        const matrix::Csr<ValueType, IndexType>* mtx)
+#define GKO_DECLARE_SYMBOLIC_CHOLESKY(ValueType, IndexType) \
+    void symbolic_cholesky(                                 \
+        const matrix::Csr<ValueType, IndexType>* mtx,       \
+        std::unique_ptr<matrix::Csr<ValueType, IndexType>>& factors)
 
 GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(GKO_DECLARE_SYMBOLIC_CHOLESKY);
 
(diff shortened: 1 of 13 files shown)


core/factorization/symbolic.cpp
```diff
@@ -58,45 +58,48 @@ GKO_REGISTER_OPERATION(cholesky_symbolic,
...
+void symbolic_cholesky(
+    const matrix::Csr<ValueType, IndexType>* mtx,
+    std::unique_ptr<matrix::Csr<ValueType, IndexType>>& factors)
 {
     using matrix_type = matrix::Csr<ValueType, IndexType>;
     const auto exec = mtx->get_executor();
     const auto host_exec = exec->get_master();
-    const auto forest = compute_elim_forest(mtx);
+    std::unique_ptr<elimination_forest<IndexType>> forest;
+    exec->run(make_compute_elim_forest(mtx, forest));
```
pratikvn, line 77: Is there some drawback in doing `exec->get_master()->run(...)` instead of having a separate call ? Then you dont need the `GKO_REGISTER_HOST_OPERATION` at all, right ?

pratikvn (commented):
I like the approach of having the function as a kernel instead of in the core, but does there need to be a separate register operation  ? Why not use `exec->get_master()->run(...)`  ?

core/factorization/symbolic.cpp
upsj: It's not being logged with the executor the object lives on, which is also the reason that we need the workaround in the benchmarks. Also with the suggested approach, we'd need to pass an additional Executor parameter and provide overloads for both OmpExecutor and ReferenceExecutor and put them inside the `kernels` namespace, all of which doesn't really fit well for core functionality.

pratikvn: In my opinion, anything computationally expensive shouldn't be happening on the host. It should be on the kernel side. That also is in line with the Ginkgo philosophy of kernels doing the hard work and the host orchestrating the calling. It is perfectly fine if these kernels then are serial (with reference or with single threaded OpenMP), but I dont think we should be doing anything intensive on the host.

MarcelKoch (commented):
It seems to be a bit annoying that this approach doesn't allow for return values of the host kernels. Perhaps that could also be managed by adding a `run_host` to Executor (or just an `run` overload) that takes a `HostOperation` instead. I'm thinking of something like: https://godbolt.org/z/TTd71zKnq. With the appropriate macros that should be as simple as what we have right now. In c++17 this could be simpler, see https://stackoverflow.com/a/53673648

upsj:
Yes, I agree this is a bit unergonomic. Do you see a way how to make that work considering that template functions can't be virtual?

core/factorization/symbolic.cpp
pratikvn: I agree that the operations are sequential and probably not suited for GPUs. My point is that they perform computations which are potentially expensive (with scaled up problem size) and hence they should not be on the core side. Regarding the duplication, we can just have a common kernel file (like we do for HIP and CUDA) and have the serial implementation on OpenMP (forcing it on one thread) and reference alike.

I think performing computation on the host side can have implications for future design. For example, if we want to add some stream/asynchronous tasking for these operations. Performing these operations on the kernel side would allow us to keep the asynchronous tasking/streaming kernel interface and keep the core side clean. 

IMO avoiding this also keeps our logging and profiling interface clean, without fragmenting it into host operations and kernel operations.

upsj: I think we need to unpack the meaning of core and kernel a bit more here. core is everything in libginkgo.so, kernel is everything in libginkgo_<backend>.so. You are talking about a conceptional separation that IMO doesn't apply here - we already have some expensive operations in core, mainly factorization-related. The only thing this PR does is make them visible to profilers regardless of the executor the objects live on. The logging interface is not impacted by this change, since to the loggers, host and kernel operations look the same.

If we were to make them kernel operations, we would have to do a lot more work: Kernels can't create Ginkgo objects, so we need to operate on arrays and combine them into Ginkgo objects on the core side again. Also as a minor point, it also doubles compilation time and binary size of the resulting functions.

MarcelKoch:
Which virtual function do you refer to, `Executor::run`, `HostOperation::run`? I think both of them don't need to be virtual.

upsj:
Having an overloaded function that is partially virtual and partially templated seems a bit dangerous, so I think we should choose only one of the two. Note that there is no such thing as a HostOperation ATM, as any potential later addition to what constitutes an Operation would need to be mirrored with HostOperation (e.g. work estimates, relocation and asynchronous execution for runtime systems)

MarcelKoch:
We could use `Executor::run_host`, which might also be helpful to add some verbosity. I think we could derive `HostOperation` from `Operation`, then changes that apply to `Operations` could be easily integrated.
But of course that would not help with return values for device operations. Perhaps with some CRTP magic it could be possible, but that is a bigger head scratcher.

core/factorization/symbolic.cpp
pratikvn: Yes, I would advocate for making them kernel operations, so this might be something we discuss before merging this PR. 

Allocations should not happen inside a operation, and I think that is all the more reason to have allocations on the host side and do the operations on the kernel side. We log and profile the allocations separately anyway, so having the allocations inside the host operation makes the profiling and logging muddled.

upsj: We allocate a lot inside kernels, e.g. SpGEMM/SpGEAM, sparselib bindings, all kernels that use temporary memory, ..., removing these allocations/moving them into core would complicate the control flow significantly.

include/ginkgo/core/base/executor.hpp
```diff
@@ -510,6 +510,56 @@ RegisteredOperation<Closure> make_register_operation(const char* name,
...
+ *     // create the operation
+ *     auto op = make_my_op(5); // x = 5
+ *
+ *     ref->run(op);  // run host kernel
+ * }
+ * ```
+ *
+ * @ingroup Executor
+ */
+#define GKO_REGISTER_HOST_OPERATION(_name, _kernel)                          \
```
yhmtsai, line 550: with this, will you still need to add the logger to host_executor?

upsj: No, but if the host operation calls kernels on the host side, they would not be visible in the benchmark runtime breakdown.

pratikvn:
I will not hold up this PR anymore, because it is fairly small change and even though it adds the interface for host operations, I think removal of these requires update of code in develop. Additionally, as @upsj needs it for benchmarking purposes, from my side we can probably merge this once approved.

core/factorization/symbolic.cpp
MarcelKoch, line 77: I don't know if we have thought of this yet, but this is a nested operation call. Can we handle that correctly in our loggers (or wherever that might be necessary)?

MarcelKoch (approved):
LGTM. The out-parameters are annoying, but that is what we have to use for now. I think we could try to tackle kernels with return values later on. Although, now that I think of it, I don't think that returning non-trivial types from kernels is possible. In most cases, a kernel can't create a ginkgo object, because the constructors are implemented in core. But for host operations, that should still work.

core/factorization/symbolic.cpp
upsj: The benchmark OperationLogger handles it correctly, with `-nested_names` resolving their nested structure. Similar things happen in kernels that call copy or create arrays. Profiling loggers will handle it correctly by design, since the annotation ranges are made for nesting.

yhmtsai:
@upsj Before I approve it, `add test` is still left in TODO. Did you already finish it?

upsj:
@yhmtsai now I have, thanks for the reminder!

core/factorization/symbolic.hpp
```diff
@@ -37,15 +37,24 @@ namespace gko {
 namespace factorization {
 
 
-/** Computes the symbolic Cholesky factorization of the given matrix. */
+/**
+ * Computes the symbolic Cholesky factorization of the given matrix.
+ * @param mtx  the input matrix
+ * @param factors  the output factors stored in a combined pattern
```
yhmtsai, line 43:
```suggestion
 * Computes the symbolic Cholesky factorization of the given matrix.
 * 
 * @param mtx  the input matrix
 * @param factors  the output factors stored in a combined pattern
```
nit

core/factorization/symbolic.hpp
```diff
@@ -37,15 +37,24 @@ namespace gko {
...
-/** Computes the symbolic LU factorization of the given matrix. */
+void symbolic_cholesky(
+    const matrix::Csr<ValueType, IndexType>* mtx,
+    std::unique_ptr<matrix::Csr<ValueType, IndexType>>& factors);
+
+/**
+ * Computes the symbolic LU factorization of the given matrix.
+ * @param mtx  the input matrix
+ * @param factors  the output factors stored in a combined pattern
+ */
```
yhmtsai, line 54:
```suggestion
/**
 * Computes the symbolic LU factorization of the given matrix.
 * 
 * @param mtx  the input matrix
 * @param factors  the output factors stored in a combined pattern
 */
```

yhmtsai (approved):
LGTM. only minor format change.

PR #1234, opened 2022-12-09
removes unnecessary run overrides in concrete executors

MarcelKoch:
The overrides of `run` in our concrete executors are unnecessary. With the addition of the generic device-id scope guard, the implementation in `ExecutorBase` is sufficient.

### DIFF
--- a/include/ginkgo/core/base/executor.hpp
+++ b/include/ginkgo/core/base/executor.hpp
@@ -1330,6 +1330,14 @@ public:
         return {this, 0};
     }
 
+    void run(const Operation& op) const override
+    {
+        this->template log<log::Logger::operation_launched>(this, &op);
+        op.run(std::static_pointer_cast<const ReferenceExecutor>(
+            this->shared_from_this()));
+        this->template log<log::Logger::operation_completed>(this, &op);
+    }
+
 protected:
     ReferenceExecutor()
     {
--- a/include/ginkgo/core/base/executor.hpp
+++ b/include/ginkgo/core/base/executor.hpp
@@ -1155,6 +1155,8 @@ class ExecutorBase : public Executor {
     friend class ReferenceExecutor;
 
 public:
+    using Executor::run;
+
     void run(const Operation& op) const override
     {
         this->template log<log::Logger::operation_launched>(this, &op);
--- a/include/ginkgo/core/base/executor.hpp
+++ b/include/ginkgo/core/base/executor.hpp
@@ -1153,6 +1153,7 @@ public:
     void run(const Operation& op) const override
     {
         this->template log<log::Logger::operation_launched>(this, &op);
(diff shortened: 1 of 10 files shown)


include/ginkgo/core/base/executor.hpp
```diff
@@ -1153,6 +1153,7 @@ class ExecutorBase : public Executor {
     void run(const Operation& op) const override
     {
         this->template log<log::Logger::operation_launched>(this, &op);
+        auto scope_guard = get_scoped_device_id_guard();
```
upsj, line 1156: very minor point: this adds another small allocation to run(...). It's unlikely this will affect performance in any way, but if it were, we could make the scoped guards non-polymorphic with an executor type enum.

upsj (approved):

pratikvn (approved):
LGTM!

include/ginkgo/core/base/executor.hpp
MarcelKoch: You are right, but I would do that in a separate PR.

MarcelKoch:
The override in `ReferenceExecutor` is necessary, because it's derived from `OmpExecutor` and not from `ExecutorBase<ReferenceExecutor>`. IMO, we should change the inheritance here. It seems the only reason for it is code sharing.

upsj:
@MarcelKoch strong agree, though it conflicts with the LambdaOperation, which doesn't have a ReferenceExecutor overload

MarcelKoch:
This might actually be an interface breaking change. Currently, it is possible to call `exec->run(op)` with a user defined operation class on any executor even if the corresponding module is not enabled. But with this change, calling `run` on an executor for which the module is not build throws an exception due to missing device-id guard implementation.

upsj:
We don't have a clear cut definition of interface breaks, but runtime changes are usually fine, especially in such an IMO uncommon use case.

yhmtsai:
MaxOS throws the error in core/test/executor
`C++ exception with description "/Users/runner/work/ginkgo/ginkgo/core/device_hooks/dpcpp_hooks.cpp:118: feature get_scoped_device_id_guard is part of the dpcpp module, which is not compiled on this system" thrown in the test body.`
maybe miss the dpcpp part?

upsj:
Or alternatively we shouldn't avoid attempting to create executors that are not compiled?

MarcelKoch:
format-rebase!

upsj (approved):
new changes LGTM as well. Not sure if this is in the scope of the PR, but can now we make `Executor::create(...)` also fail if it is not compiled? This currently leads to quite delayed error messages only at attempting to do something with the executor. Of course, we need to figure out how to handle disabled host executors when creating device executors. Maybe disallow such inconsistent configurations altogether?

MarcelKoch:
@upsj I think another PR would be better suited for that.

PR #1236, opened 2022-12-11
Add stream support to CUDA and HIP

### LINKED ISSUE #1206: Support custom streams in CUDA and HIP
We could make the integration with other applications more complete if we allowed users to provide custom streams that will be used for all kernels

TODO:
- [ ] Use stream-ordered allocation (`cudaMallocAsync`, `cudaFreeAsync`, though they use a memory pool, so it might still make sense to use a regular synchronizing `cudaMalloc` and `cudaFree` in some cases)
- [x] Pass stream parameter to all kernel launches (`<<<grid, block, 0, stream>>>`)
- [x] Set stream on all sparselib handles (cuBLAS, cuSPARSE, cuFFT, cuRAND?)
- [x] Set stream on all thrust kernel launches (`thrust::cuda::par.on(stream)`)

upsj:
TODO:
- [x] test this behavior, see [cuDF tests](https://github.com/rapidsai/cudf/tree/branch-23.02/cpp/tests/utilities/identify_stream_usage) for how to do it
- [x] add stream to sparselib/blas handles
- [x] add RAII stream handle types for CUDA and HIP

Closes #1206


cuda/base/curand_bindings.hpp
```diff
@@ -35,6 +35,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
 #include <curand.h>
+#include <driver_types.h>
```
upsj, line 38: not sure how that ended up here, can be removed

MarcelKoch (commented):
What happens with a user provided stream, if `device_reset = true` in a cuda/hip executor? For the cuda docs, it seems like using the stream again is undefined behavior.

upsj:
yes, to be honest, I think the device_reset settings exceed the scope of a single executor and should be removed in the long term. I like to think of devices like the MPI environment: They should only be initialized and finalized globally. Streams and devices should usually have a wider scope than executors, unless they are explicitly created internally.

MarcelKoch:
That sounds like a reasonable approach. Perhaps we could deprecate the constructor with the device_reset in another PR.
Right now, maybe we could issue a warning if a user sets `device_reset` and also a custom stream, or don't offer a constructor with both device_reset and stream parameter.
Sidenote: Our cuda/hip executors count the number of active executors on a specific device, and only does the reset if there are no other active executors. So that case should be handled 'correctly', but it might ignore device_reset even if it was set by a user.

upsj:
Ah, thanks for the reminder, that's actually slight worse IMO: This way the order of destruction matters for whether `DeviceReset` gets called.

cuda/base/thrust.cuh
```diff
@@ -0,0 +1,57 @@
...
+
+
+namespace gko {
+namespace kernels {
+namespace cuda {
+
+
+inline auto thrust_policy(std::shared_ptr<const CudaExecutor> exec)
+{
+    return thrust::cuda::par.on(exec->get_stream());
```
MarcelKoch, line 48: Is that part of the public thrust API? I couldn't find it in their online docs. If it's not part of the public API, do you know which versions support this?

upsj: yes, the documentation is a bit bad in that regard. I am not sure whether the generic `thrust::par` would be equivalent. It should be supported by all versions of Thrust.

MarcelKoch: Do you have any link to their docs for either of that? I still can't find anything on that.

upsj: Sorry, I mixed up standardese with Thrust, it should be [thrust::device](https://thrust.github.io/doc/group__execution__policies_ga330d16be82b0518b6a37d026c2811e7e.html#ga330d16be82b0518b6a37d026c2811e7e), though more recent versions have thrust::par as well. The stream parameter is not documented anywhere though.

MarcelKoch: I find that a bit troublesome, but I guess it works? At least I can find it in the headers directly.

upsj: Thrust doesn't have a big team working on it, so they can't focus too much on documentation, sadly :(

MarcelKoch: It is actually mentioned in the release notes of thrust 1.8.0 https://github.com/NVIDIA/thrust/releases/tag/1.8.0. Until it breaks, that is enough for me.

common/cuda_hip/base/device_matrix_data_kernels.hpp.inc
```diff
@@ -39,12 +39,12 @@ void remove_zeros(std::shared_ptr<const DefaultExecutor> exec,
     // due to overly generic assignment operator and constructor leading to
     // ambiguities. So we need to use our own fake_complex type
     using device_value_type = device_member_type<ValueType>;
-    auto value_ptr = thrust::device_pointer_cast(
-        reinterpret_cast<const device_value_type*>(values.get_const_data()));
+    auto value_ptr =
+        reinterpret_cast<const device_value_type*>(values.get_const_data());
```
MarcelKoch, line 43: Do we not need the thrust pointer cast any more because of the new execution policy, or was it always unnecessary? Either way, it is nice that we can do without it.

upsj: it was always unnecessary, I only had some segfaults in the past that made me think it was necessary, but the issue must have been elsewhere

MarcelKoch (approved):
Reviewing this is a bit hard because there are so many smaller changes. I've done some regex searches, and it seems that there are no kernel launches without the stream parameter anymore. (Unless we are not using the <<<...>>> syntax somewhere, but I don't think so.) And the proposed testing might catch the cases we missed.
Other than that, the interface changes on the executor side look good. Perhaps we could add a constructor, so that we have one constructor with the device_reset parameter, but without the stream and vice versa.

upsj:
format!

hip/base/hiprand_bindings.hip.hpp
```diff
@@ -72,11 +72,13 @@ struct is_supported<std::complex<double>> : std::true_type {};
...
 inline hiprandGenerator_t rand_generator(int64 seed,
-                                         hiprandRngType generator_type)
+                                         hiprandRngType generator_type,
+                                         hipStream_t stream)
 {
     hiprandGenerator_t gen;
-    hiprandCreateGenerator(&gen, generator_type);
-    hiprandSetPseudoRandomGeneratorSeed(gen, seed);
+    GKO_ASSERT_NO_HIPRAND_ERRORS(hiprandCreateGenerator(&gen, generator_type));
+    GKO_ASSERT_NO_HIPRAND_ERRORS(
```
yhmtsai, line 80: hiprandSetStream is available from 1.5.0

third_party/identify_stream_usage/identify_stream_usage.cpp
```diff
@@ -0,0 +1,299 @@
+/*
+ * Copyright (c) 2022, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
yhmtsai, line 4: We can directly use the code by keeping license, right? We only use it from third party and in testing purpose.

yhmtsai (approved):
LGTM. 
Some tests needs to fix and
stencil example may also need to add the get_stream or comment

cuda/factorization/cholesky_kernels.cu
```diff
@@ -114,49 +115,16 @@ void cholesky_symbolic_count(
...
+                postorder_parent, row_nnz);
     }
 }
 
 GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
     GKO_DECLARE_CHOLESKY_SYMBOLIC_COUNT);
 
 
-template <typename ValueType, typename IndexType>
-void cholesky_symbolic_factorize(
```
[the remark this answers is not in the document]
upsj: unfortunately, they need Csr sorting, which relies on hipSPARSE/cuSPARSE, so they need to stay separated

cuda/preconditioner/jacobi_kernels.cu
```diff
@@ -69,187 +69,7 @@ constexpr int default_grid_size = 32 * 32 * 128;
...
-    ValueType* result_values, size_type result_stride) GKO_NOT_IMPLEMENTED;
-
-GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
-    GKO_DECLARE_JACOBI_CONVERT_TO_DENSE_KERNEL);
-
-
 }  // namespace jacobi
 }  // namespace cuda
 }  // namespace kernels
-}  // namespace gko
```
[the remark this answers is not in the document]
upsj: yeah definitely, no idea how that got lost

third_party/identify_stream_usage/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
+cmake_minimum_required(VERSION 3.17)
```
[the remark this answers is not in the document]
upsj: this would fail if used with an older version, but since it is guarded by a flag, I think this should be fine?

third_party/identify_stream_usage/identify_stream_usage.cpp
upsj: yes, maybe we should mention it in LICENSE.md, but the licenses are compatible

hip/base/hiprand_bindings.hip.hpp
upsj: which packaged HIP version does that coincide with?

yhmtsai:
I have checked all `>>>`. If they are kernel call, they use stream.

test/utils/executor.hpp
```diff
@@ -54,60 +54,53 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-std::shared_ptr<ExecType> init_executor(
-    std::shared_ptr<gko::ReferenceExecutor>);
-
-
-template <>
-inline std::shared_ptr<gko::ReferenceExecutor>
-    init_executor<gko::ReferenceExecutor>(
-        std::shared_ptr<gko::ReferenceExecutor>)
+inline void init_executor(std::shared_ptr<gko::ReferenceExecutor>,
+                          std::shared_ptr<gko::ReferenceExecutor>& exec)
```
MarcelKoch, line 58: nit: I know it doesn't make a lot of difference, but the single parameter version would work as well in all of these (and the mpi) cases

cuda/test/base/cuda_executor.cu
```diff
@@ -95,11 +95,18 @@ protected:
     void SetUp()
     {
         ASSERT_GT(gko::CudaExecutor::get_num_devices(), 0);
+#ifdef GKO_TEST_NONDEFAULT_STREAM
+        cuda = gko::CudaExecutor::create(
+            0, omp, false, gko::default_cuda_alloc_mode, stream.get());
```
pratikvn, line 100: Maybe I am missing something here. This is still `nullptr` right ? I dont see a `StreamCreate` for association with a non-default stream.

hip/base/hiprand_bindings.hip.hpp
yhmtsai: I think it should be available in all our considered version?

.gitlab/variables.yml
```diff
@@ -15,6 +15,7 @@
     BUILD_MPI: "OFF"
     MPI_AS_ROOT: "OFF"
     FAST_TESTS: "OFF"
+    NONDEFAULT_STREAM: "OFF"
```
MarcelKoch, line 18: Do we have at least one test with the stream check on?

MarcelKoch (approved):
looked at it again, only very minor nits.

test/utils/executor.hpp
upsj: The template specialization version didn't work for me, because it can't differ in number of parameters afaik?

.gitlab/variables.yml
upsj: yes, `build/cuda110/mvapich2/gcc/cuda/debug/shared` has a subsequent `test` job

include/ginkgo/core/base/executor.hpp
```diff
@@ -1532,7 +1541,8 @@ class CudaExecutor : public detail::ExecutorBase<CudaExecutor>,
     static std::shared_ptr<CudaExecutor> create(
         int device_id, std::shared_ptr<Executor> master,
         bool device_reset = false,
-        allocation_mode alloc_mode = default_cuda_alloc_mode);
+        allocation_mode alloc_mode = default_cuda_alloc_mode,
+        CUstream_st* stream = nullptr);
```
MarcelKoch, line 1545: Perhaps just as a reminder, I think it would be good, to have only constructors with either device_reset or stream parameter.

cuda/test/base/cuda_executor.cu
upsj: no, it is constructed with `cudaStreamCreate` in the `gko::cuda_stream` constructor

test/utils/executor.hpp
MarcelKoch: Oh, I missed that. In that case, never mind.

include/ginkgo/core/base/executor.hpp
upsj: maybe it would make sense to handle this in another PR adding custom allocator support (for AMReX's memory pool)

MarcelKoch: The issue is not with the allocation mode, but with the device reset. So I think it fits in here. I think at the top of the PR there is our old discussion, and this was just meant as a reminder for that.

upsj: Yes, I only want to keep the number of interface changes small, since this will likely be another longer discussion :smile:

upsj: It still anticipates a design decision we haven't made yet. We should probably discuss that tomorrow :)

cuda/test/base/cuda_executor.cu
```diff
@@ -95,11 +95,18 @@ protected:
...
+#else
         cuda = gko::CudaExecutor::create(0, omp);
+#endif
         cuda2 = gko::CudaExecutor::create(
-            gko::CudaExecutor::get_num_devices() - 1, omp);
+            gko::CudaExecutor::get_num_devices() - 1, omp, false,
+            gko::default_cuda_alloc_mode, cuda->get_stream());
```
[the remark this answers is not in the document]
upsj: good point, I'd have to look up how cross-GPU copies and streams interact

pratikvn:
Some comments:

1. I think we should support other ways of stream creation ? See `(cuda/hip)StreamCreateWithFlags`
2. I am not sure if we should just accept a device specific stream object type, or have something more generic, that is abstracted over all the executors ? The user specifies a general gko::stream object, which internally stores the device specific stream type ? 
3. Do we have any applications that currently use this ? If not, maybe we should consider sketching out a more generic approach to asynchronous execution ?

upsj:
@pratikvn thanks for pointing to `StreamCreateWithFlags`, I thought this was only available in a more recent CUDA version, but it's already there in CUDA 9.2, so we can support it generically. I think the rest we can discuss in the meeting :)

include/ginkgo/core/base/executor.hpp
upsj: I think I'd prefer sticking with this for now, and changing the new constructor in a subsequent PR

include/ginkgo/core/base/executor.hpp
```diff
@@ -159,10 +159,19 @@ struct cublasContext;
...
 struct cusparseContext;
 
+struct CUstream_st;
+
 struct hipblasContext;
 
 struct hipsparseContext;
 
+#if GINKGO_HIP_PLATFORM_HCC
+struct ihipStream_t;
```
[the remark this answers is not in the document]
upsj: no, afaik the HIP backend doesn't define such an alias

upsj:
format!

common/cuda_hip/factorization/cholesky_kernels.hpp.inc
```diff
@@ -156,3 +156,36 @@ __launch_bounds__(default_block_size) void cholesky_symbolic_factorize_kernel(
         out_cols[out_base] = row;
     }
 }
+
+
+template <typename ValueType, typename IndexType>
```
pratikvn, line 161: I think this is from the cholesky PR ?

common/cuda_hip/factorization/factorization_kernels.hpp.inc
```diff
@@ -361,3 +361,191 @@ __global__ __launch_bounds__(default_block_size) void initialize_l(
 
 
 }  // namespace kernel
+
+
+template <typename ValueType, typename IndexType>
+void add_diagonal_elements(std::shared_ptr<const DefaultExecutor> exec,
```
pratikvn, line 367: Same for the kernels in this file.

pratikvn (dismissed):
I think there are some changes from other PRs here.

common/cuda_hip/factorization/cholesky_kernels.hpp.inc
upsj: no, this code was only moved from cuda/factorization/cholesky_kernels.hpp so we have a single place where we set the stream for CUDA and HIP kernels. The functions were identical except for the kernel call syntax anyways.

common/cuda_hip/factorization/factorization_kernels.hpp.inc
upsj: same here

common/cuda_hip/factorization/cholesky_kernels.hpp.inc
upsj: but thanks for pointing this out, it's maybe not entirely clear with the files being so apart in the listing 😅

PR #1241, opened 2022-12-14
Proof of concept: local operation for mpi

yhmtsai:
The idea is simple: the operation only considers the local matrix (the diagonal part).
`op->apply(b, x) on mtx` -> `op->apply(b.local, x.local)` on mtx.local no communication is needed.
Of course, the generated matrix is only based on the diagonal part, so it might lose some global information.
ILU/IC might not prefer this way. I guess it still works well when the offdiagonal is not heavy/strong
Jacobi should be quite suitable for this case unless we would like to have a cross rank block.
With this, we can also have a distributed IR with local preconditioner to refine solution by global communication.
all LinOpFactory (it can easily extend to LinOp) including Multigrid

### DIFF
--- a/include/ginkgo/core/distributed/local_factory.hpp
+++ b/include/ginkgo/core/distributed/local_factory.hpp
@@ -1,0 +1,129 @@
+/*******************************<GINKGO LICENSE>******************************
+Copyright (c) 2017-2022, the Ginkgo authors
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+
+1. Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+
+2. Redistributions in binary form must reproduce the above copyright
+notice, this list of conditions and the following disclaimer in the
+documentation and/or other materials provided with the distribution.
+
+3. Neither the name of the copyright holder nor the names of its
+contributors may be used to endorse or promote products derived from
+this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS
+IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED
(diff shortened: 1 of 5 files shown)


upsj:
Thanks for adding this! We already discussed this as a special-case RAS without overlap a long time ago, but with the interface for that not being ready, I think it's a good first step. With only a single distributed matrix type, I don't think we need the whole GetLocal indirection though, and the name should probably reflect its intended purpose (local preconditioning) more closely.

include/ginkgo/core/distributed/local_factory.hpp
```diff
@@ -0,0 +1,129 @@
...
+namespace distributed {
+
+
+/**
+ * The LocalFactory class defines a local factory for (MPI-)distributed system
+ *
+ * It will generate individual factory on the local matrix, so there's no
+ * communication in the local factory or corresponding operation.
+ */
+class LocalFactory : public EnableLinOp<LocalFactory> {
```
upsj, line 61: This is a LinOp, not a factory
```suggestion
class LocalPreconditioner : public EnableLinOp<LocalFactory> {
```

tcojean:
As stated in slack, I believe @pratikvn has a RAS somewhere and wanted to wrap it up and generate plots for the ECP milestone.

include/ginkgo/core/distributed/local_factory.hpp
yhmtsai: Thanks for suggestion. I am not good at naming a class.
I would like to avoid Preconditioner because it can be used for any kind of Factory (or generated LinOp)
something like LocalLinOp but I wonder it is confusing.

upsj:
@yhmtsai I think `Preconditioner` is the closest/most important use case. There may be more unusual use cases, but using it as a solver only makes sense when the local parts are independent (so you wouldn't even need MPI). Another possibility would be `LocalOperator`, which would make it harder to discover for users who are looking for a simple preconditioner. Alternatively, we could also call it `distributed::BlockJacobi`, but that may be very far from the normal use of the term (which involves exact solves of the diagonal blocks)

yhmtsai:
it's handled by #1248

PR #602, opened 2020-07-28
Create executor in exec_map only if requested.

greole:
Fixes #594 Create executor only if requested using lambda functions inside the `exec_map`.

PR #1243, opened 2022-12-15
throws in `Executor::create` if module is not compiled

MarcelKoch:
This PR makes all the `ConcreteExecutor::create` throw if the module is not compiled. This is also true for `OmpExecutor` and `ReferenceExecutor`. Also, the `populate_exec_info` function now also throws.

### DIFF
--- a/core/device_hooks/cuda_hooks.cpp
+++ b/core/device_hooks/cuda_hooks.cpp
@@ -52,17 +52,11 @@ version version_info::get_cuda_version() noexcept
 
 std::shared_ptr<CudaExecutor> CudaExecutor::create(
     int device_id, std::shared_ptr<Executor> master, bool device_reset,
-    allocation_mode alloc_mode)
-{
-    return std::shared_ptr<CudaExecutor>(new CudaExecutor(
-        device_id, std::move(master), device_reset, alloc_mode));
-}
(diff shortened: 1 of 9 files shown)


MarcelKoch:
format!

tcojean:
I believe we discussed this a while ago and decided for not doing this?

upsj:
@tcojean Do you think it would be possible to find the discussion again? I'd be curious to understand the rationale, maybe we can even document it in the hooks?

yhmtsai:
I also remembered we discuss it before, but does not find any related discussion yet. I think one of reason was example or benchmark always create all executors such that we need to do that. but after https://github.com/ginkgo-project/ginkgo/pull/602, we should not face this issue.

tcojean:
Good job Mike, I was also looking. I'm quite sure it appeared indeed here: https://github.com/ginkgo-project/ginkgo/pull/554, but the comments were on slack it seems and not in the PR.

Some concern to keep in mind is that for me this is almost an interface break in some form due to changing exception behavior. In all, some user codes would not work anymore depending on how they did their executor selection because of throwing.

MarcelKoch:
In an earlier internal discussion, the majority favored throwing only at kernel launches. Since this is a relatively small change, and it could lead to potential user issues, as pointed out by @tcojean, I will not pursue this further.

PR #1244, opened 2022-12-15
FGMRES

yhmtsai:
This PR adds the FGMRES. I am not sure whether to create FGMRES like FCG 
reference https://www-users.cse.umn.edu/~saad/Calais/PREC.pdf and the paper A Flexible Inner-Outer Preconditioned GMRES Algorithm
TODO:
- [x] create FGMRES? (using `with_flexible(true)`)
- [ ] avoid the unnecessary allocation (partially, it only avoids the allocation of flexible in normal. It may be possible to reduce some allocation in flexible, too)

### DIFF
--- a/core/solver/gmres.cpp
+++ b/core/solver/gmres.cpp
@@ -76,6 +76,7 @@ std::unique_ptr<LinOp> Gmres<ValueType>::transpose() const
             share(as<Transposable>(this->get_preconditioner())->transpose()))
         .with_criteria(this->get_stop_criterion_factory())
         .with_krylov_dim(this->get_krylov_dim())
+        .with_flexible(this->get_parameters().flexible)
         .on(this->get_executor())
         ->generate(
             share(as<Transposable>(this->get_system_matrix())->transpose()));
@@ -90,6 +91,7 @@ std::unique_ptr<LinOp> Gmres<ValueType>::conj_transpose() const
             as<Transposable>(this->get_preconditioner())->conj_transpose()))
         .with_criteria(this->get_stop_criterion_factory())
         .with_krylov_dim(this->get_krylov_dim())
+        .with_flexible(this->get_parameters().flexible)
         .on(this->get_executor())
         ->generate(share(
             as<Transposable>(this->get_system_matrix())->conj_transpose()));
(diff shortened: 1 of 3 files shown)


upsj (approved):
Really nice work, I didn't consider we can just add this to GMRES instead of adding an entirely different solver. Related, should we at some point merge FCG and CG the same way?

Is there an easy way to construct an example where variable-preconditioned GMRES breaks down?

core/solver/gmres.cpp
```diff
@@ -341,9 +347,16 @@ void Gmres<ValueType>::apply_dense_impl(const VectorType* dense_b,
                                     span{local_num_rows * (restart_iter + 1),
                                          local_num_rows * (restart_iter + 2)},
                                     span{0, num_rhs});
-        // preconditioned_vector = get_preconditioner() * this_krylov
+        auto preconditioned_krylov = create_submatrix_helper(
```
upsj, line 350: we could also store a std::unique_ptr outside and allocate it only if necessary?

core/solver/gmres.cpp
```diff
@@ -414,6 +427,9 @@ void Gmres<ValueType>::apply_dense_impl(const VectorType* dense_b,
     auto krylov_bases_small = create_submatrix_helper(
         krylov_bases, dim<2>{num_rows, num_rhs},
         span{0, local_num_rows * (restart_iter + 1)}, span{0, num_rhs});
+    auto preconditioned_krylov_bases_small = create_submatrix_helper(
```
upsj, line 430: same here

MarcelKoch (approved):

core/solver/gmres.cpp
yhmtsai: I move the corresponding into the scope because they are only needed once and read-only

core/solver/gmres.cpp
yhmtsai: done

yhmtsai:
@upsj it avoids some additional allocation from FGMRES in GMRES now.

upsj:
Awesome, thanks for the update!

yhmtsai:
rebase!

MarcelKoch (approved):

PR #1248, opened 2022-12-20
Add support for distributed block-diagonal preconditioners. 

pratikvn:
This PR is the first set of PRs enabling distributed preconditioning for our distributed solvers.

+ As discussed, everything is in the `experimental` namespace and in the `distributed/` folder.
+ ~~An example has been added to show the usage of the preconditioner~~.
+ ~~Some minor fixes have also been done to mpi, the run and dispatch function and also moving some things into the experimental namespace (Some contributed by @MarcelKoch)~~.
+ ~~A distributed stream logger was also added, but this probably needs more discussion (By @MarcelKoch).~~ Removed for now, as it is a separate thing, for which a new PR should be created.

### TODO
+ [x] Add tests.

### DIFF
--- a/core/distributed/preconditioner/schwarz.cpp
+++ b/core/distributed/preconditioner/schwarz.cpp
@@ -1,5 +1,5 @@
 /*******************************<GINKGO LICENSE>******************************
-Copyright (c) 2017-2022, the Ginkgo authors
+Copyright (c) 2017-2023, the Ginkgo authors
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
--- a/core/distributed/preconditioner/schwarz.cpp
+++ b/core/distributed/preconditioner/schwarz.cpp
@@ -76,8 +76,10 @@ void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::apply_dense_impl(
 {
     using Vector = matrix::Dense<ValueType>;
     auto exec = this->get_executor();
-    this->local_solver_->apply(gko::detail::get_local(dense_b),
-                               gko::detail::get_local(dense_x));
+    if (this->local_solver_ != nullptr) {
+        this->local_solver_->apply(gko::detail::get_local(dense_b),
+                                   gko::detail::get_local(dense_x));
+    }
 }
 
 
@@ -97,11 +99,15 @@ void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::apply_impl(
 
 
 template <typename ValueType, typename LocalIndexType, typename GlobalIndexType>
-void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::generate()
+void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::generate(
+    std::shared_ptr<const LinOp> system_matrix)
 {
     if (parameters_.local_solver_factory) {
-        this->local_solver_ =
-            parameters_.local_solver_factory->generate(local_system_matrix_);
+        this->local_solver_ = parameters_.local_solver_factory->generate(
+            as<experimental::distributed::Matrix<ValueType, LocalIndexType,
+                                                 GlobalIndexType>>(
+                system_matrix)
+                ->get_local_matrix());
     } else {
         GKO_NOT_IMPLEMENTED;
     }
--- a/core/distributed/preconditioner/schwarz.cpp
+++ b/core/distributed/preconditioner/schwarz.cpp
@@ -1,0 +1,126 @@
+/*******************************<GINKGO LICENSE>******************************
+Copyright (c) 2017-2022, the Ginkgo authors
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+
+1. Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+
+2. Redistributions in binary form must reproduce the above copyright
+notice, this list of conditions and the following disclaimer in the
+documentation and/or other materials provided with the distribution.
+
+3. Neither the name of the copyright holder nor the names of its
+contributors may be used to endorse or promote products derived from
+this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS
+IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED
+TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A
+PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+#include <ginkgo/core/distributed/preconditioner/schwarz.hpp>
+
+
+#include <memory>
+
+
+#include <ginkgo/core/base/exception_helpers.hpp>
+#include <ginkgo/core/base/executor.hpp>
+#include <ginkgo/core/base/math.hpp>
+#include <ginkgo/core/base/precision_dispatch.hpp>
+#include <ginkgo/core/base/temporary_conversion.hpp>
+#include <ginkgo/core/base/utils.hpp>
+#include <ginkgo/core/distributed/matrix.hpp>
+#include <ginkgo/core/matrix/csr.hpp>
+#include <ginkgo/core/matrix/dense.hpp>
+
+
+#include "core/base/utils.hpp"
+#include "core/distributed/helpers.hpp"
+
+
+namespace gko {
+namespace distributed {
+namespace preconditioner {
+
+
+template <typename ValueType, typename IndexType>
+void Schwarz<ValueType, IndexType>::apply_impl(const LinOp* b, LinOp* x) const
+{
+    precision_dispatch_real_complex_distributed<ValueType>(
+        [this](auto dense_b, auto dense_x) {
+            this->apply_dense_impl(dense_b, dense_x);
+        },
+        b, x);
+}
+
+
+template <typename ValueType, typename IndexType>
+template <typename VectorType>
+void Schwarz<ValueType, IndexType>::apply_dense_impl(const VectorType* dense_b,
+                                                     VectorType* dense_x) const
+{
+    using Vector = matrix::Dense<ValueType>;
+    auto exec = this->get_executor();
+    auto one_op = initialize<Vector>({one<ValueType>()}, exec);
+    auto neg_one_op = initialize<Vector>({-one<ValueType>()}, exec);
+    this->local_solver_->apply(gko::detail::get_local(dense_b),
+                               gko::detail::get_local(dense_x));
+
+    if (coarse_solvers_[0]) {
+        for (auto& coarse : coarse_solvers_) {
+            coarse->apply(dense_b, dense_x);
+        }
+    }
+}
+
+
+template <typename ValueType, typename IndexType>
+void Schwarz<ValueType, IndexType>::apply_impl(const LinOp* alpha,
+                                               const LinOp* b,
+                                               const LinOp* beta,
+                                               LinOp* x) const
+{
+    precision_dispatch_real_complex_distributed<ValueType>(
+        [this](auto dense_alpha, auto dense_b, auto dense_beta, auto dense_x) {
+            auto x_clone = dense_x->clone();
+            this->apply_dense_impl(dense_b, x_clone.get());
+            dense_x->scale(dense_beta);
+            dense_x->add_scaled(dense_alpha, x_clone.get());
+        },
+        alpha, b, beta, x);
+}
+
+
+template <typename ValueType, typename IndexType>
+void Schwarz<ValueType, IndexType>::generate()
+{
+    if (parameters_.generated_local_solver) {
+        this->local_solver_ = parameters_.generated_local_solver;
+    } else {
+        this->local_solver_ =
+            parameters_.local_solver->generate(local_system_matrix_);
+    }
+}
+
+
+#define GKO_DECLARE_SCHWARZ(ValueType, IndexType) \
+    class Schwarz<ValueType, IndexType>
+GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(GKO_DECLARE_SCHWARZ);
+
+
+}  // namespace preconditioner
+}  // namespace distributed
+}  // namespace gko
--- a/core/distributed/preconditioner/schwarz.cpp
+++ b/core/distributed/preconditioner/schwarz.cpp
@@ -57,8 +57,9 @@ namespace distributed {
 namespace preconditioner {
 
 
-template <typename ValueType, typename IndexType>
-void Schwarz<ValueType, IndexType>::apply_impl(const LinOp* b, LinOp* x) const
+template <typename ValueType, typename LocalIndexType, typename GlobalIndexType>
+void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::apply_impl(
+    const LinOp* b, LinOp* x) const
 {
     precision_dispatch_real_complex_distributed<ValueType>(
         [this](auto dense_b, auto dense_x) {
@@ -68,10 +69,10 @@ void Schwarz<ValueType, IndexType>::apply_impl(const LinOp* b, LinOp* x) const
 }
 
 
-template <typename ValueType, typename IndexType>
+template <typename ValueType, typename LocalIndexType, typename GlobalIndexType>
 template <typename VectorType>
-void Schwarz<ValueType, IndexType>::apply_dense_impl(const VectorType* dense_b,
-                                                     VectorType* dense_x) const
+void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::apply_dense_impl(
+    const VectorType* dense_b, VectorType* dense_x) const
 {
     using Vector = matrix::Dense<ValueType>;
     auto exec = this->get_executor();
@@ -80,11 +81,9 @@ void Schwarz<ValueType, IndexType>::apply_dense_impl(const VectorType* dense_b,
 }
 
 
-template <typename ValueType, typename IndexType>
-void Schwarz<ValueType, IndexType>::apply_impl(const LinOp* alpha,
-                                               const LinOp* b,
-                                               const LinOp* beta,
-                                               LinOp* x) const
+template <typename ValueType, typename LocalIndexType, typename GlobalIndexType>
+void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::apply_impl(
+    const LinOp* alpha, const LinOp* b, const LinOp* beta, LinOp* x) const
 {
     precision_dispatch_real_complex_distributed<ValueType>(
         [this](auto dense_alpha, auto dense_b, auto dense_beta, auto dense_x) {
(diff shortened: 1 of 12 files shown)


pratikvn:
#1241 is a clone (with different naming) of what was there in [a previous distributed preconditioner branch](https://github.com/ginkgo-project/ginkgo/tree/distributed-schwarz). 

This PR takes a few initial commits from that distributed-schwarz, which adds the local block preconditioning into the Schwarz preconditioner class, which is what these general block preconditioners are called. This is also a extension of the Schwarz preconditioners from #1089, which was closed as we decided to do this directly for distributed.

core/distributed/helpers.hpp
```diff
@@ -168,6 +172,33 @@ void vector_dispatch(T* linop, F&& f, Args&&... args)
...
+    if (is_distributed(op)) {
+        run<gko::experimental::distributed::Vector, float, double,
+            std::complex<float>, std::complex<double>>(
+            op, [&](auto vector_op) { local = get_local(vector_op); });
+    } else
+#endif
+    {
+        run<gko::matrix::Dense, float, double, std::complex<float>,
+            std::complex<double>>(
+            op, [&](auto vector_op) { local = get_local(vector_op); });
```
greole, line 196: I think the nomenclature is somewhat confusing here. What is the meaning of `get_local`  in the non-distributed case? I always thought get_local to be rank local vector part.

core/distributed/preconditioner/schwarz.cpp
```diff
@@ -0,0 +1,120 @@
+/*******************************<GINKGO LICENSE>******************************
+Copyright (c) 2017-2022, the Ginkgo authors
```
greole, line 2: I guess the year needs to updated to 2023
```suggestion
Copyright (c) 2017-2023, the Ginkgo authors
```

core/distributed/preconditioner/schwarz.cpp
```diff
@@ -0,0 +1,120 @@
...
+#include <ginkgo/core/base/math.hpp>
+#include <ginkgo/core/base/precision_dispatch.hpp>
+#include <ginkgo/core/base/temporary_conversion.hpp>
+#include <ginkgo/core/base/utils.hpp>
+#include <ginkgo/core/distributed/matrix.hpp>
+#include <ginkgo/core/matrix/csr.hpp>
+#include <ginkgo/core/matrix/dense.hpp>
+
+
+#include "core/base/utils.hpp"
```
greole, line 50:
```suggestion
```
that seems to be redundant to the `<ginkgo/core/base/utils.hpp>` if I am not mistaken.

core/distributed/preconditioner/schwarz.cpp
greole: I guess you can just run a sed command to fix that in the other files ...

examples/distributed-preconditioned-solver/distributed-preconditioned-solver.cpp
```diff
@@ -0,0 +1,282 @@
...
+    using bj = gko::preconditioner::Jacobi<ValueType, LocalIndexType>;
+
+    const auto comm = gko::experimental::mpi::communicator(MPI_COMM_WORLD);
+    const auto rank = comm.rank();
+
+    // @sect3{Initialization and User Input Handling}
+    // Since this is an MPI program, we need to initialize and finalize
+    // MPI at the begin and end respectively of our program. This can be easily
+    // done with the following helper construct that uses RAII to automize the
+    // initialization and finalization.
```
greole, line 88: This comment seems to be at the wrong place.

examples/distributed-preconditioned-solver/distributed-preconditioned-solver.cpp
```diff
@@ -0,0 +1,282 @@
...
+                      << std::endl;
+        }
+        std::exit(-1);
+    }
+
+    ValueType t_init = gko::experimental::mpi::get_walltime();
+
+    // User input settings:
+    // - The executor, defaults to reference.
+    // - The number of grid points, defaults to 100.
```
greole, line 102:
```suggestion
    // - The number of grid points, defaults to 100.
    // - The number of iterations, defaults to 1000.
```
To be consistent, also mention the number of iterations

examples/distributed-preconditioned-solver/doc/intro.dox
```diff
@@ -0,0 +1,8 @@
+<a name="Intro"></a>
+<h1>Introduction</h1>
+This distributed solver example should help you understand the basics of using Ginkgo in a distributed setting.
+The example will solve a simple 1D Laplace equation where the system can be distributed row-wise to multiple processes.
+To run the solver with multiple processes, use `mpirun -n NUM_PROCS ./distributed-solver [executor] [num_grid_points] [num_iterations]`.
+
+If you are using GPU devices, please make sure that you run this example with at most as many processes as you have GPU
+devices available.
```
greole, line 8: Maybe this should be rephrased. Since the example will work with more mpi ranks then GPUs.
```suggestion
If you are using GPU devices, please make sure that you run this example with at most as many processes as you have GPU
devices available, since oversubscription might lead to performance degradation
```

examples/distributed-preconditioned-solver/distributed-preconditioned-solver.cpp
greole: I think you removed the code section from the distributed-solver.cpp example, but left the comment in.

greole (commented):
Some minor comments from my side

include/ginkgo/core/log/stream.hpp
```diff
@@ -178,6 +178,7 @@ class Stream : public Logger {
      * @param verbose  whether we want detailed information or not. This
      *                 includes always printing residuals and other information
      *                 which can give a large output.
+     * @param prefix  string to prefix the log message with.
```
MarcelKoch, line 181: I think all the changes in this file should be part of the distributed stream logger PR.

core/distributed/helpers.hpp
```diff
@@ -168,6 +172,33 @@ void vector_dispatch(T* linop, F&& f, Args&&... args)
...
 
+/**
+ * Returns the local vector as a LinOp
+ * @tparam LinOpType either LinOp or const LinOp
+ * @param op  the object from which the local vector is extracted
+ * @return  the local vector
+ */
+template <typename LinOpType, typename = std::enable_if_t<std::is_same<
+                                  LinOp, std::decay_t<LinOpType>>::value>>
+LinOpType* get_local(LinOpType* op)
```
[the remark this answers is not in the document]
MarcelKoch: And there are also a lot of other changes which might not be necessary.

examples/CMakeLists.txt
```diff
@@ -50,7 +50,7 @@ if(GINKGO_HAVE_PAPI_SDE)
 endif()
 
 if(GINKGO_BUILD_MPI)
-    list(APPEND EXAMPLES_LIST distributed-solver)
+    list(APPEND EXAMPLES_LIST distributed-solver distributed-preconditioned-solver)
```
MarcelKoch, line 53: I would be fine with merging the two examples. I don't see much value in having them separated.

examples/distributed-solver/distributed-solver.cpp
```diff
@@ -72,22 +73,19 @@ int main(int argc, char* argv[])
     // non-distributed program. Please note that not all solvers support
     // distributed systems at the moment.
     using solver = gko::solver::Cg<ValueType>;
+    using schwarz =
+        gko::experimental::distributed::preconditioner::Schwarz<ValueType,
+                                                                LocalIndexType>;
+    using bj = gko::preconditioner::Jacobi<ValueType, LocalIndexType>;
```
MarcelKoch, line 79: Not uses. I guess that is some leftover from when this example had a preconditioner?

include/ginkgo/core/distributed/preconditioner/schwarz.hpp
```diff
@@ -0,0 +1,170 @@
...
+class Schwarz : public EnableLinOp<Schwarz<ValueType, IndexType>> {
+    friend class EnableLinOp<Schwarz>;
+    friend class EnablePolymorphicObject<Schwarz, LinOp>;
+
+public:
+    using EnableLinOp<Schwarz>::convert_to;
+    using EnableLinOp<Schwarz>::move_to;
+    using value_type = ValueType;
+    using index_type = IndexType;
+    using mat_data = matrix_data<ValueType, IndexType>;
```
MarcelKoch, line 87:
```suggestion
```
I don't think we need that atm?

pratikvn:
format!

core/distributed/preconditioner/schwarz.cpp
pratikvn: No they are actually separate files. One is a global header and one in `" "` is a local Ginkgo only header.

core/distributed/helpers.hpp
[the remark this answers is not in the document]
pratikvn: All removed now.

examples/CMakeLists.txt
pratikvn: I have merged them.

pratikvn:
This has been rebased and should be ready to review again.

MarcelKoch (changes requested):
I think it's missing an apply test.

examples/distributed-solver/distributed-solver.cpp
```diff
@@ -214,19 +207,38 @@ int main(int argc, char* argv[])
     comm.synchronize();
     ValueType t_read_setup_end = gko::experimental::mpi::get_walltime();
 
+
     // @sect3{Solve the Distributed System}
     // Generate the solver, this is the same as in the non-distributed case.
+    // with a local block diagonal preconditioner.
+    //
```
MarcelKoch, line 214: nit:
```suggestion
```

MarcelKoch (approved):
Only smaller nits.

core/test/mpi/distributed/preconditioner/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+ginkgo_create_test(schwarz MPI_SIZE 1)
```
yhmtsai, line 1: should it be 1?

examples/distributed-solver/distributed-solver.cpp
```diff
@@ -72,22 +73,18 @@ int main(int argc, char* argv[])
     // non-distributed program. Please note that not all solvers support
     // distributed systems at the moment.
     using solver = gko::solver::Cg<ValueType>;
+    using schwarz = gko::experimental::distributed::preconditioner::Schwarz<
+        ValueType, LocalIndexType, GlobalIndexType>;
+    using bj = gko::preconditioner::Jacobi<ValueType, LocalIndexType>;
+
+    const auto comm = gko::experimental::mpi::communicator(MPI_COMM_WORLD);
+    const auto rank = comm.rank();
```
yhmtsai, line 81:
```suggestion
    // Create an MPI communicator wrapper and get the rank.
    const auto comm = gko::experimental::mpi::communicator(MPI_COMM_WORLD);
    const auto rank = comm.rank();
```

examples/distributed-solver/distributed-solver.cpp
```diff
@@ -45,6 +45,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 int main(int argc, char* argv[])
 {
+    const gko::experimental::mpi::environment env(argc, argv);
```
yhmtsai, line 48: maybe add the documentation

examples/distributed-solver/distributed-solver.cpp
```diff
@@ -108,56 +105,52 @@ int main(int argc, char* argv[])
...
+                 int device_id = 0;
+                 if (gko::DpcppExecutor::get_num_devices("gpu")) {
+                     device_id = gko::experimental::mpi::map_rank_to_device_id(
+                         comm, gko::DpcppExecutor::get_num_devices("gpu"));
+                 } else if (gko::DpcppExecutor::get_num_devices("cpu")) {
+                     device_id = gko::experimental::mpi::map_rank_to_device_id(
+                         comm, gko::DpcppExecutor::get_num_devices("cpu"));
                  } else {
-                     throw std::runtime_error("No suitable DPC++ devices");
+                     GKO_NOT_IMPLEMENTED;
```
yhmtsai, line 138: I think the original one is more informatic?

include/ginkgo/core/distributed/preconditioner/schwarz.hpp
```diff
@@ -0,0 +1,171 @@
...
+#define GKO_PUBLIC_CORE_DISTRIBUTED_PRECONDITIONER_SCHWARZ_HPP_
+
+
+#include <ginkgo/config.hpp>
+
+
+#if GINKGO_BUILD_MPI
+
+
+#include <ginkgo/core/base/array.hpp>
```
greole, line 43:
```suggestion
```
Seems to be unused?

include/ginkgo/core/distributed/preconditioner/schwarz.hpp
```diff
@@ -0,0 +1,171 @@
...
+    explicit Schwarz(const Factory* factory,
+                     std::shared_ptr<const LinOp> system_matrix)
+        : EnableLinOp<Schwarz>(factory->get_executor(),
+                               gko::transpose(system_matrix->get_size())),
+          parameters_{factory->get_parameters()},
+          local_system_matrix_{std::move(
+              as<experimental::distributed::Matrix<ValueType, LocalIndexType,
+                                                   GlobalIndexType>>(
+                  system_matrix.get())
+                  ->get_local_matrix())}
```
yhmtsai, line 140: I think it should keep the global one and pass the get_local_matrix when needed

test/mpi/distributed/preconditioner/schwarz.cpp
```diff
@@ -0,0 +1,224 @@
...
+        dist_x->fill(gko::zero<value_type>());
+
+        local_solver_factory =
+            local_prec_type::build().with_max_block_size(1u).on(exec);
+    }
+
+    void SetUp() override { ASSERT_EQ(comm.size(), 3); }
+
+
+    template <typename ValueType, typename IndexType>
```
yhmtsai, line 144: nit
```suggestion

    template <typename ValueType, typename IndexType>
```

include/ginkgo/core/distributed/preconditioner/schwarz.hpp
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+
+    GKO_CREATE_FACTORY_PARAMETERS(parameters, Factory)
+    {
+        /**
+         * Local solver factory.
+         */
+        std::shared_ptr<const LinOpFactory> GKO_FACTORY_PARAMETER_SCALAR(
+            local_solver_factory, nullptr);
```
greole, line 109: Could you add the option to add pre-generated inner solver again? like
```suggestion
            local_solver_factory, nullptr);
            
            std::vector<std::shared_ptr<const LinOp>> GKO_FACTORY_PARAMETER_VECTOR(
            generated_inner_solvers, nullptr);
```

or would that be a separate PR?

greole (commented):
just a quick comment

test/mpi/distributed/preconditioner/schwarz.cpp
```diff
@@ -0,0 +1,224 @@
...
+                 TupleTypenameNameGenerator);
+
+
+TYPED_TEST(SchwarzPreconditioner, CanApplyPreconditionedSolver)
+{
+    using value_type = typename TestFixture::value_type;
+    using csr = typename TestFixture::local_matrix_type;
+    using cg = typename TestFixture::solver_type;
+    using prec = typename TestFixture::dist_prec_type;
+    constexpr double tolerance = 1e-3;
```
yhmtsai, line 202: the tolerance is quite low here? it's not converged at all?

include/ginkgo/core/distributed/preconditioner/schwarz.hpp
```diff
@@ -0,0 +1,171 @@
...
+
+/**
+ * A Schwarz preconditioner is a simple domain decomposition preconditioner that
+ * generalizes the Block Jacobi preconditioner, incorporating options for
+ * different local subdomain solvers and overlaps between the subdomains.
+ *
+ * See Iterative Methods for Sparse Linear Systems (Y. Saad) for a general
+ * treatment and variations of the method.
+ *
+ * @note Currently overlaps are not supported (TODO).
```
yhmtsai, line 68: what's the overlap meaning here?
1. the overlap of domain
2. the overall refinement after local solve

core/distributed/preconditioner/schwarz.cpp
```diff
@@ -0,0 +1,119 @@
...
+    using Vector = matrix::Dense<ValueType>;
+    auto exec = this->get_executor();
+    this->local_solver_->apply(gko::detail::get_local(dense_b),
+                               gko::detail::get_local(dense_x));
+}
+
+
+template <typename ValueType, typename LocalIndexType, typename GlobalIndexType>
+void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::apply_impl(
+    const LinOp* alpha, const LinOp* b, const LinOp* beta, LinOp* x) const
```
yhmtsai, line 86: miss test

yhmtsai (changes requested):
LGTM in general. but the tolerance is quite low and have some question in the comments

core/test/mpi/distributed/preconditioner/CMakeLists.txt
pratikvn: I think for core we are not using any MPI functions, so it should be fine. We do the same for matrix and other distributed tests as well.

test/mpi/distributed/preconditioner/schwarz.cpp
pratikvn: I think we do the same for our solver tests as well. So the tolerances are similar as it is there.

include/ginkgo/core/distributed/preconditioner/schwarz.hpp
pratikvn: The overlap of the domain. The refinement after local solve would be a coarse grid correction, which needs to be added as well. But Schwarz methods are generally with overlaps, so I wanted to make it clear that has not been implemented yet.

test/mpi/distributed/preconditioner/schwarz.cpp
yhmtsai: why do we only set 1e-3 for mpi solver?

include/ginkgo/core/distributed/preconditioner/schwarz.hpp
pratikvn: I think I would do this in a separate PR as there are a few more considerations there.

test/mpi/distributed/preconditioner/schwarz.cpp
pratikvn: I think we take general matrices and make them hpd. So it might be that we they are not well behaved for all solvers. I think those common tests need to be checked and we need to check the tolerances there.

include/ginkgo/core/distributed/preconditioner/schwarz.hpp
pratikvn: I think as this is just a local preconditioner, we should not need the global one atleast now. We can change that later in case we need it.

greole (approved):
LGTM

include/ginkgo/core/distributed/preconditioner/schwarz.hpp
upsj: ~~I'm kind of torn here - by storing the local matrix only, the lifetime of the global system matrix and its components become disconnected, which makes reasoning about them harder. OTOH, once we start overlaps, the local matrix used for preconditioning becomes different, so we need to store it explicitly anyways.~~
Why are we storing the local system matrix at all? We could always get that from the local solver. To me, this seems to be mixing two levels of nesting in one.

include/ginkgo/core/distributed/preconditioner/schwarz.hpp
```diff
@@ -0,0 +1,170 @@
...
+ * @tparam ValueType  precision of matrix elements
+ * @tparam IndexType  integral type of the preconditioner
+ *
+ * @ingroup schwarz
+ * @ingroup precond
+ * @ingroup LinOp
+ */
+template <typename ValueType = default_precision,
+          typename LocalIndexType = int32, typename GlobalIndexType = int64>
+class Schwarz
```
upsj, line 78: This class doesn't have a valid moved-from state. I think a nullptr test in `apply` for `local_solver` should be sufficient though.

core/distributed/preconditioner/schwarz.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+
+template <typename ValueType, typename LocalIndexType, typename GlobalIndexType>
+void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::generate()
+{
+    if (parameters_.local_solver_factory) {
+        this->local_solver_ =
+            parameters_.local_solver_factory->generate(local_system_matrix_);
+    } else {
+        GKO_NOT_IMPLEMENTED;
```
upsj, line 106: GKO_NOT_SUPPORTED would probably be more appropriate, but we don't have a dedicated exception handler for this setting

upsj (approved):
LGTM, I only have a few small things I would like to note

include/ginkgo/core/distributed/preconditioner/schwarz.hpp
pratikvn: If you dont have a generated local solver, then you need to have the local matrix to generate the local solver from the passed in local solver factory. But your point regarding the disconnected components makes sense. I think for now, as @yhmtsai suggested, I will store the global matrix, and will have to do tha dynamic casts when I need the local matrix.

upsj: Do you need it past the `generate` step? Because otherwise you only need to store the generated solver, just like a local preconditioner doesn't need to store the system matrix after generation ;)

pratikvn: Yes, currently I dont. So, I will remove it altogether for now and we can add it later if needed.

PR #1257, opened 2023-01-12
Half precision support

### LINKED ISSUE #73: Verifying Ginkgo with custom value types
Currently we are only compiling Ginkgo for default (`float`, `double`, `std::complex<>` of those) value types when checking the correctness of our implementation. (We're also testing only for `double`, see #32 for discussions on that).

This means that the interface that `ValueType` needs to have may change after a PR without anyone noticing (e.g. someone used a part of the interface of default types, which user-defined types may not have to implement).

I propose we create a dummy `ValueType` implementation (we would create a concept instead if TS concepts were a part of C++11), that could just be a `double` wrapped in a class, and have the interface that Ginkgo's `ValueType` has to satisfy defined. This gives us two benefits:

*   we can try to compile with this type whenever we test Ginkgo and get a compiler error whenever we inadvertently assume something new about `ValueType`'s interface
*   users can look into this type, and see what their custom value type needs to have defined in order to work with Ginkgo

yhmtsai:
It adds the half (fp16) into ginkgo.
Some related collection:
- GCC 12+ supports _Float16
- Clang 15+ supports _Float16
- Cuda 10+ supports 16 bit atomicAdd on Compute Capability >= 7.0 (atomicCAS has same condition although it only mentioned the version in the atomicAdd)
- Hip does not support 16 bit atomic
- Sycl (I am not sure)
- c++-23 supports float16 and bfloat16

~~Cuda 9.2 `__half` does not contain `+=`. (I think it can be added by more operator overload outside)~~
ROCm 4.0 can not convert `__half` to `double`. (I do not think it can be added outside?)
I will disable `half` support for these two version.

There are two additional commit: one is fixing oneAPI6 (#1251) and the other is multigrid experiments.
They will be cleaned afterwards.

The following PR is related to half:
- #1706 
- #1708 
- #1710 
- #1711 
- #1712 
- #1713 
- #1716
- #1736 

Discussion:
Do we still need to use float2half from vendor in gko::half?

Some fixes are extracted from this pr to #1253 
TODO:
- [x] compile everything and run-able
- [x] ~~change the IDR without half~~ when running curand/hiprand, throws the error
- [x] ~~only test the double <-> float~~or to wait the thrust<__half>
  - [x] current testing on double, float. but a few double test with half are failed (I have create the testing/benchmark next_precision chain and also add some TODO at the failed tests)
- [x] thrust<__half> operator. 
  - [x] the stacking issue needs to be resolved
- [ ] MPI Custom type/op for __half (it can be communicated but the reduction will be wrong)
- [x] Csr strategy logic when arch does not support 16bit atomic
- [x] Disable half when it does not need? (with GINKGO_ENABLE_HALF)
- [x] Using Sycl half all not gko::half when using dpcpp
- [x] clean code (might not need some changes under current half design)
- [x] Move gko::half to public interface

Closes #73


MarcelKoch:
I would suggest putting all the todos into separate PRs. Having it compile and run in the first PR is IMO already enough. You could create a github project to track the rest of the todo.

core/multigrid/pgm.cpp
```diff
@@ -176,7 +177,7 @@ void Pgm<ValueType, IndexType>::generate()
     auto abs_mtx = pgm_op->compute_absolute();
     // abs_mtx is already real valuetype, so transpose is enough
     auto weight_mtx = gko::as<weight_csr_type>(abs_mtx->transpose());
-    auto half_scalar = initialize<matrix::Dense<real_type>>({0.5}, exec);
+    auto half_scalar = initialize<matrix::Dense<real_type>>({half(0.5)}, exec);
```
MarcelKoch, line 180: Why use `half` here explicitly?

core/solver/multigrid.cpp
```diff
@@ -314,7 +315,14 @@ void MultigridState::generate(const LinOp* system_matrix_in,
         auto next_nrows = mg_level_list.at(i)->get_coarse_op()->get_size()[0];
         auto mg_level = mg_level_list.at(i);
 
-        run<gko::multigrid::EnableMultigridLevel, float, double,
+        run<gko::multigrid::EnableMultigridLevel,
+#if GINKGO_ENABLE_HALF
```
MarcelKoch, line 319: again, maybe merge the `#if`

MarcelKoch (commented):
First step of my review. So far I've looked at the public headers and core, with some unified kernels. But I haven't looked into the backends yet.

One major thing, the `half.hpp` is included in many files where it doesn't seem necessary. Removing that could probably reduce the number of changed files by half.

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -268,10 +268,14 @@ class Matrix
           Matrix<ValueType, LocalIndexType, GlobalIndexType>>,
       public ConvertibleTo<
           Matrix<next_precision<ValueType>, LocalIndexType, GlobalIndexType>>,
+#if GINKGO_ENABLE_HALF
+      public ConvertibleTo<Matrix<next_precision<next_precision<ValueType>>,
```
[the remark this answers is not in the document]
yhmtsai: I think they need to use two from the logic of friend class.
`Matrix<value_type>: public ConvertibleTo<Matrix<next_precision>>`
with `friend Matrix<next_precision>` does not work. The original without half works because `next = prev`.
`Matrix<value_type>::convert_to(next_precision)` requires `Matrix<next_precision>` declare the `Matrix<value_type>` as a friend.
That is, matrix should use `friend Matrix<prev_precision>` to make `convert_to(next_precision)` work.
Thus, `convert_to next<next<>>` and `friend prev<prev<>>` are paired. `convert_to next<>` and  `friend prev<>` are paired.

include/ginkgo/core/base/half.hpp
```diff
@@ -0,0 +1,731 @@
...
+// #if defined(__CUDA_ARCH__) || defined(__HIP_DEVICE_COMPILE__)
+template <>
+struct basic_float_traits<__half> {
+    using type = __half;
+    static constexpr int sign_bits = 1;
+    static constexpr int significand_bits = 10;
+    static constexpr int exponent_bits = 5;
+    static constexpr bool rounds_to_nearest = true;
+};
+// #endif
```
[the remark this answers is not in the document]
yhmtsai: I need to check this again. I think I have tried it before this in hip. but hip is confusing when it is only shown in device side

include/ginkgo/core/base/half.hpp
```diff
@@ -0,0 +1,731 @@
...
+    {
+        this->float2half(static_cast<float>(val));
+        return *this;
+    }
+
+    GKO_ATTRIBUTES operator float() const noexcept
+    {
+#if defined(__CUDA_ARCH__) || defined(__HIP_DEVICE_COMPILE__)
+        return __half2float(reinterpret_cast<const __half&>(data_));
+#else   // defined(__CUDA_ARCH__) || defined(__HIP_DEVICE_COMPILE__)
```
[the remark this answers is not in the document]
yhmtsai: It's used by the adaptive jacobi internal storage.
I would like to remove it if possible.
It is the reason we can not use half in constexpr

include/ginkgo/core/base/types.hpp
```diff
@@ -419,6 +430,17 @@ GKO_ATTRIBUTES constexpr bool operator!=(precision_reduction x,
     _enable_macro(CudaExecutor, cuda)
 
 
+// cuda half operation is supported from arch 5.3
+#if GINKGO_ENABLE_HALF && (!defined(__CUDA_ARCH__) || __CUDA_ARCH__ >= 530)
```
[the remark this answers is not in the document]
yhmtsai: I do not think so because we can compile several arch.

benchmark/utils/types.hpp
```diff
@@ -67,4 +67,33 @@ using etype = double;
...
+template <typename T>
+struct next_precision_impl<std::complex<T>> {
+    using type = std::complex<typename next_precision_impl<T>::type>;
+};
+
+
+}  // namespace detail
+
+template <typename T>
+using next_precision = typename detail::next_precision_impl<T>::type;
```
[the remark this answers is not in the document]
yhmtsai: yes, I think it is used from mixed-ell or some advanced preconditioner.

include/ginkgo/core/base/half.hpp
```diff
@@ -0,0 +1,731 @@
...
+
+
+}  // namespace gko
+
+
+namespace std {
+
+
+template <>
+class complex<gko::half> {
```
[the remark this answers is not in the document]
yhmtsai: what's the unspecified meaning actually? it is not like the undefined behavior.
another we can do is to put complex<half> in our own xstd and alias for complex<double>.
Does it avoid the issue?

include/ginkgo/core/distributed/matrix.hpp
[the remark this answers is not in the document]
MarcelKoch: But it is defined as `previous_precision = next_precision<next_precision<T>`, so it should be equivalent?

include/ginkgo/core/base/half.hpp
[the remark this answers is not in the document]
MarcelKoch: It worked before, so I don't think it's an actual issue, just something to keep in mind.

thoasm (changes requested):
This review only concerns public headers and the accessor.
Please remove the many includes of `half.hpp` and remove the specializations for `std`.

include/ginkgo/core/base/math.hpp
```diff
@@ -970,7 +1054,7 @@ template <typename T>
 GKO_ATTRIBUTES GKO_INLINE constexpr std::enable_if_t<!is_complex_s<T>::value, T>
 imag_impl(const T&)
 {
-    return T{};
+    return T(0.0);
```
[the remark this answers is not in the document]
yhmtsai: I believe it was related to NVHPC. NVHPC default constructor will not use zero like the other compiler.
It should not be needed anymore after I initialize underlying data to zero specficly in half constructor

accessor/cuda_helper.hpp
```diff
@@ -17,7 +17,15 @@
 #include "utils.hpp"
 
 
+struct __half;
+
+
```
upsj, line 22: There are some ABI/name mangling differences between struct and class in MSVC - are you sure this will be defined as a struct always?

Also for cleaner headers/exports, maybe we should make this conditional on CUDA compilation?

accessor/reference_helper.hpp
```diff
@@ -14,8 +14,10 @@
 
 // CUDA TOOLKIT < 11 does not support constexpr in combination with
 // thrust::complex, which is why constexpr is only present in later versions
-#if defined(__CUDA_ARCH__) && defined(__CUDACC_VER_MAJOR__) && \
-    (__CUDACC_VER_MAJOR__ < 11)
+// TODO: NVC++ constexpr
```
upsj, line 17: Is this TODO addressed with the `constexpr` PR?

benchmark/CMakeLists.txt
```diff
@@ -77,17 +77,25 @@ function(ginkgo_add_single_benchmark_executable name use_lib_linops macro_def ty
...
+                target_link_libraries("${name}" cusparse_linops_${type})
+            endif()
+            if (GINKGO_BUILD_HIP)
+                target_compile_definitions("${name}" PRIVATE HAS_HIP=1)
+                target_link_libraries("${name}" hipsparse_linops_${type})
+            endif()
+            if (GINKGO_BUILD_SYCL)
+                target_compile_definitions("${name}" PRIVATE HAS_DPCPP=1)
+                target_link_libraries("${name}" onemkl_linops_${type})
+            endif()
```
upsj, line 98: Why not make CUDA unconditional

```cmake
if(GINKGO_BUILD_CUDA)
...
endif()
if (NOT ("${type}" STREQUAL "h"))
...
endif()
```

benchmark/utils/generator.hpp
```diff
@@ -132,10 +132,7 @@ struct DefaultSystemGenerator {
     {
         auto res = Vec::create(exec);
         res->read(gko::matrix_data<ValueType, itype>(
-            size,
-            std::uniform_real_distribution<gko::remove_complex<ValueType>>(-1.0,
-                                                                           1.0),
-            get_engine()));
+            size, std::uniform_real_distribution<>(-1.0, 1.0), get_engine()));
```
upsj, line 135: This will potentially give us conversion warnings in the future, but I'm generally fine with it

benchmark/utils/types.hpp
```diff
@@ -31,11 +32,44 @@ using etype = std::complex<double>;
...
 
 using rc_etype = gko::remove_complex<etype>;
 
 
+namespace detail {
+
+
+// singly linked list of all our supported precisions
+template <typename T>
+struct next_precision_impl {};
```
upsj, line 51: This would probably also be a good time to implement `prev_precision`?

common/cuda_hip/base/device_matrix_data_kernels.cpp
```diff
@@ -31,9 +31,13 @@ void remove_zeros(std::shared_ptr<const DefaultExecutor> exec,
     auto value_ptr = as_device_type(values.get_const_data());
     auto size = values.get_size();
     // count nonzeros
-    auto nnz = thrust::count_if(
-        thrust_policy(exec), value_ptr, value_ptr + size,
-        [] __device__(device_value_type value) { return is_nonzero(value); });
+    // __half != is only device, can not call __device__ from a __host__
+    // __device__ (is_nonzero)
```
upsj, line 35: is this resolved with `constexpr`?

common/cuda_hip/base/device_matrix_data_kernels.cpp
```diff
@@ -49,7 +53,8 @@ void remove_zeros(std::shared_ptr<const DefaultExecutor> exec,
                                as_device_type(new_values.get_data())));
         thrust::copy_if(thrust_policy(exec), it, it + size, out_it,
                         [] __device__(tuple_type entry) {
-                            return is_nonzero(thrust::get<2>(entry));
+                            return thrust::get<2>(entry) !=
+                                   zero(thrust::get<2>(entry));
```
upsj, line 57: same here

common/cuda_hip/base/math.hpp
```diff
@@ -11,16 +11,57 @@
...
+    static constexpr auto inf() { return std::numeric_limits<T>::infinity(); }
+    static constexpr auto max() { return std::numeric_limits<T>::max(); }
+    static constexpr auto min() { return std::numeric_limits<T>::min(); }
+};
+
+template <>
+struct device_numeric_limits<__half> {
+    // from __half documentation, it accepts unsigned short
+    // __half does not have constexpr
+    static GKO_ATTRIBUTES GKO_INLINE auto inf()
```
upsj, line 45: can still be marked `constexpr`, right?

dev_tools/scripts/generate_cuda_memory_ptx.py
```diff
@@ -186,14 +181,109 @@ class type_desc:
...
+    for o in memory_orderings:
+        membar_expression = "" if o.is_relaxed else f"membar_acq_rel{s.fn_suffix}();"
+        const_ptr_expr = s.ptr_expr.format(
+            ptr=f"const_cast<{t.name}*>(ptr)")
+        mut_ptr_expr = s.ptr_expr.format(ptr="ptr")
+        print(f"""
+__device__ __forceinline__ {t.name} load{o.fn_load_suffix}{s.fn_suffix}(const {t.name}* ptr)
+{{
+    {t.parent_name} result;
+    asm volatile("{{\\n\\t"
```
upsj, line 210: This complexity makes me ask: Do we need 16 bit precision triangular solves? IIRC the results from Fritz were not showing notable speedups.

include/ginkgo/core/base/matrix_data.hpp
```diff
@@ -39,7 +40,7 @@ template <typename ValueType, typename Distribution, typename Generator>
 typename std::enable_if<!is_complex_s<ValueType>::value, ValueType>::type
 get_rand_value(Distribution&& dist, Generator&& gen)
 {
-    return dist(gen);
+    return ValueType(dist(gen));
```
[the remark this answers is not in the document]
upsj: This is probably related to my comment on the benchmarks? I would also prefer something more explicit here, or no change at all, and we could fix the issue at the source (by providing a converting wrapper for the distribution)

include/ginkgo/core/base/mpi.hpp
```diff
@@ -117,6 +118,12 @@ GKO_REGISTER_MPI_TYPE(unsigned long long, MPI_UNSIGNED_LONG_LONG);
 GKO_REGISTER_MPI_TYPE(float, MPI_FLOAT);
 GKO_REGISTER_MPI_TYPE(double, MPI_DOUBLE);
 GKO_REGISTER_MPI_TYPE(long double, MPI_LONG_DOUBLE);
+#if GINKGO_ENABLE_HALF
+// OpenMPI 5.0 have support from MPIX_C_FLOAT16 and MPICHv3.4a1 MPIX_C_FLOAT16
+// TODO: it only works on the transferring
+GKO_REGISTER_MPI_TYPE(half, MPI_UNSIGNED_SHORT);
+GKO_REGISTER_MPI_TYPE(std::complex<half>, MPI_FLOAT);
```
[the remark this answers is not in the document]
upsj: Fully agree here, distributed half precision seems a bit out of scope right now.

include/ginkgo/core/base/precision_dispatch.hpp
```diff
@@ -49,12 +49,14 @@ make_temporary_conversion(Ptr&& matrix)
     using Pointee = detail::pointee<Ptr>;
     using Dense = matrix::Dense<ValueType>;
     using NextDense = matrix::Dense<next_precision<ValueType>>;
+    using NextNextDense =
+        matrix::Dense<next_precision<next_precision<ValueType>>>;
     using MaybeConstDense =
         std::conditional_t<std::is_const<Pointee>::value, const Dense, Dense>;
     auto result = detail::temporary_conversion<
-        MaybeConstDense>::template create<NextDense>(matrix);
+        MaybeConstDense>::template create<NextDense, NextNextDense>(matrix);
```
upsj, line 57: `PrevDense`?

upsj (commented):
I took a first pass over the PR. I believe we should see if we can limit the scope of half support a bit or split this into smaller PRs, starting with simple things (SpMV, Dense) and then moving on to solvers. I'm still not convinced we need to be able to build preconditioners or factorizations in half precision. We can also build them in a higher precision and then convert down. I expect a much higher complexity for half integration in those more complex algorithms.

dev_tools/scripts/generate_cuda_memory_ptx.py
yhmtsai: Yes, we got improvement in Petr's example

benchmark/utils/generator.hpp
yhmtsai: I think it is handled in the get_rand_value,

common/cuda_hip/base/math.hpp
upsj: but this is a function?

dev_tools/scripts/generate_cuda_memory_ptx.py
upsj: but what we need for half precision is only 2-4 functions (no need for acqrel, and in shared memory it's likely more efficient to use float), can we just write them out explicitly?

common/cuda_hip/base/device_matrix_data_kernels.cpp
yhmtsai: no. constexpr does not just need marked from #1695 actually.
It still lead to __host__ tries to call __device__ function, which leads an error

common/cuda_hip/base/math.hpp
yhmtsai: I can not make it constexpr variable, so I turn it to be function.
I thought it is no way to be constexpr, so I do not try to mark it to constexpr.

yhmtsai:
do not fully agree on the half generation might not useful in practice.
we need to provide them in practice, or no one can check whether it is useful or not.
Also, some of generation and apply are not in different classes.

include/ginkgo/core/distributed/matrix.hpp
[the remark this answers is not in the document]
yhmtsai: no, previous_precision is mapped to next_precision here.
if you delete friend next<next>, you need to delete function prev<prev> not next<next>
It is the confusing in the original code because the direction between friend and function is not consistent.

yhmtsai:
close this PR, as the corresponding PRs are merged in develop

PR #1263, opened 2023-01-26
Fix handling of missing diagonal in symbolic factorizations

upsj:
This artificially adds the diagonal entry for all symbolic factorizations by using a sentinel value (Cholesky) or adding the diagonal explicitly if missing (LU).

TODO:

- [x] fix cholesky sparsity pattern computation

### DIFF
--- a/common/cuda_hip/factorization/cholesky_kernels.hpp.inc
+++ b/common/cuda_hip/factorization/cholesky_kernels.hpp.inc
@@ -46,7 +46,7 @@ __global__ __launch_bounds__(default_block_size) void build_postorder_cols(
     auto lower_end = row_begin;
     for (auto nz = row_begin; nz < row_end; nz++) {
         const auto col = cols[nz];
-        if (col <= row) {
+        if (col < row) {
             postorder_cols[lower_end] = inv_postorder[cols[nz]];
             lower_end++;
         }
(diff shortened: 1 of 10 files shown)


upsj:
format-rebase!

common/cuda_hip/factorization/cholesky_kernels.hpp.inc
```diff
@@ -110,25 +113,28 @@ template <int subwarp_size, typename IndexType>
...
+        const IndexType* postorder_cols, const IndexType* postorder,
+        const IndexType* postorder_parent, const IndexType* out_row_ptrs,
+        IndexType* out_cols)
 {
     const auto row = thread::get_subwarp_id_flat<subwarp_size, IndexType>();
     if (row >= num_rows) {
         return;
     }
     const auto row_begin = row_ptrs[row];
+    const auto diag_postorder = inv_postorder[row];
```
yhmtsai, line 126: Could you remind me what's inv_postoder and postorder?

reference/test/factorization/cholesky_kernels.cpp
```diff
@@ -162,7 +162,7 @@ TYPED_TEST(Cholesky, KernelSymbolicCountSeparable)
          {0, 0, 0, 1, 1, 0, 0, 0, 0, 0},
          {0, 0, 0, 1, 1, 1, 0, 0, 0, 1},
          {0, 0, 0, 0, 1, 1, 0, 0, 0, 0},
-         {0, 0, 0, 0, 0, 0, 1, 0, 0, 1},
+         {0, 0, 0, 0, 0, 0, 1, 1, 0, 1},
```
yhmtsai, line 165: symmetric for cholesky? It's still correct in previous version because we only use the lower part?

yhmtsai (commented):
LGTM in other parts. the diag_postorder part is not reviewed yet

reference/test/factorization/cholesky_kernels.cpp
upsj: yes, precisely

common/cuda_hip/factorization/cholesky_kernels.hpp.inc
upsj: postorder is the sequence of row numbers such that each child comes before its parents (and thus a permutation), inv_postorder is the inverse permutation

fritzgoebel (approved):
LGTM! 
A comment about the purpose of `diag_postorder` would be helpful.

common/cuda_hip/factorization/cholesky_kernels.hpp.inc
yhmtsai: postorder[i]: i-th node is postorder[i]-th in original graph
row will be considered as inv_postorder[row] in new graph
is postorder_cols like inv_postorder[]?
col[i] will be considered as postorder_col[i] in the new graph

yhmtsai: from build_postorder_cols, it seems to be.

yhmtsai (approved):

upsj:
rebase!

PR #1266, opened 2023-02-04
Fix hwloc version detection

upsj:
The version detection chokes on versions that actually contain alphabet characters instead of the common 0-9. This never came up because we commonly only use versions up to 2.9, but it fails with 1.11.


yhmtsai (approved):
LGTM.
Just a side note: outputformat is supported from cmake 3.13

pratikvn (approved):
Thanks for fixing this. LGTM! I would also maybe go ahead and update the License in the top of the file.

tcojean (approved):
LGTM.

upsj:
rebase!

PR #1267, opened 2023-02-05
Add development container

upsj:
This adds a dockerfile that can be used to build all of Ginkgo, including a few development tools. My long-term goal is to keep an updated version of this container with up-to-date prebuilt Ginkgo around and update it with every `develop` merge to allow people to quickly start working on a new branch without having to rebuild everything every time.

See https://hub.docker.com/repositories/ginkgodev with `unified` and `base` containers containing the full image or just the APT packages

- [x] merge https://github.com/spack/spack/pull/35336


dev_tools/container/Dockerfile
```diff
@@ -0,0 +1,98 @@
...
+    libopencv-dev \
+    pkg-config \
+    ninja-build \
+    # Dependencies for our spack packages
+    autoconf automake bzip2 diffutils libtool m4 tar xz-utils perl pigz re2c openssh-client \    
+    && rm -rf /var/lib/apt/lists/*
+
+RUN (echo "deb http://apt.llvm.org/jammy/ llvm-toolchain-jammy-15 main" > /etc/apt/sources.list.d/llvm15.list) && \
+    (curl https://apt.llvm.org/llvm-snapshot.gpg.key > /etc/apt/trusted.gpg.d/apt.llvm.org.asc) && apt-get update && \
+    apt-get install -y --no-install-recommends clang-15 clang-tools-15 clang-15-doc clangd-15 clang-tidy-15 clang-format-15 \
```
yhmtsai, line 31: should it use format 14 because we use clang-format-14 as our standard in repo

dev_tools/container/Dockerfile
```diff
@@ -0,0 +1,98 @@
...
+
+COPY entrypoint.sh /root/entrypoint.sh
+
+ENV CMAKE_EXPORT_COMPILE_COMMANDS=ON
+
+# allow mpi to run with root user
+ENV OMPI_ALLOW_RUN_AS_ROOT=1 OMPI_ALLOW_RUN_AS_ROOT_CONFIRM=1
+
+# fixup broken rocm package
+RUN cd /opt/rocm-5.4.0/hiprand/lib/cmake/ && ln -s ../../../lib/cmake/hiprand/hiprand-fortran-config.cmake && \
```
yhmtsai, line 95: does rocm-5.4 from apt still use the old rocm-cmake?
It should be fixed after https://github.com/RadeonOpenCompute/rocm-cmake/pull/107 merged

dev_tools/container/packages.yaml
```diff
@@ -0,0 +1,63 @@
+packages:
```
yhmtsai, line 1: the packages listed here are installed from the apt, right?

dev_tools/container/Dockerfile
upsj: yes, but 5.5 should fix that https://github.com/ROCmSoftwarePlatform/rocRAND/issues/315

dev_tools/container/packages.yaml
upsj: yes, they are not detectable by spack (yet), so I need to specify their location manually

dev_tools/container/Dockerfile
upsj: I think we could install just clang-format-14 and use 15 for the rest, I would like to profit from the latest clang features

examples/kokkos_assembly/kokkos_assembly.cpp
```diff
@@ -124,9 +124,8 @@ double calculate_error(int discretization_points,
         KOKKOS_LAMBDA(int i, double& lsum) {
             const auto h = 1.0 / (discretization_points + 1);
             const auto xi = (i + 1) * h;
-            lsum += Kokkos::Experimental::abs(
-                (v_u(i) - correct_u(xi)) /
-                Kokkos::Experimental::abs(correct_u(xi)));
+            lsum += Kokkos::abs((v_u(i) - correct_u(xi)) /
+                                Kokkos::abs(correct_u(xi)));
```
MarcelKoch, line 128: I think we should then require Kokkos >= 3.7

MarcelKoch:
can you change the target branch to `unified_build_fixes` so we can filter out the changes from #1333?

PR #1268, opened 2023-02-05
Fix FindPAPI with non-implicit include path

upsj:
When `papi.h` is installed in a path that is not available by default (i.e. without specifying `-I /path/to/include`), the feature detection fails. To fix this, we need to forward the include path to the `try_run` invocation.


pratikvn (approved):
LGTM!

yanjen (approved):
LGTM

thoasm (approved):
LGTM!

tcojean (approved):

PR #1269, opened 2023-02-06
Add Distributed Multigrid.

pratikvn:
This PR updates the multigrid class to handle distributed matrices and hence allows preconditioning and solution with distributed multigrid. 

## Major changes

1. Store row and column partition objects in the Matrix class to use within Multigrid.
2. Template the memory allocation and multigrid core functions on VectorType and allow dynamic switching between the two.
3. Store matrix_data object in the distributed matrix class to be able to generate coarse matrices.

Of course, as there is no coarse generation method, we cannot still use distributed multigrid automatically, but that will be added in a future PR.

## Points of discussion

1. We probably need to store the partition objects in the distributed matrix class, but I am open to other alternatives.
2. For ease, we also probably need to store the matrix_data object (or devie_matrix_data), but I am not very happy about this.

## Issues

1. The mixed precision version of distributed multigrid does not yet work and needs to be looked into.

### DIFF
--- a/core/solver/multigrid.cpp
+++ b/core/solver/multigrid.cpp
@@ -15,6 +15,8 @@
 #include <ginkgo/core/base/math.hpp>
 #include <ginkgo/core/base/utils.hpp>
 #include <ginkgo/core/base/utils_helper.hpp>
+#include <ginkgo/core/distributed/matrix.hpp>
+#include <ginkgo/core/distributed/vector.hpp>
 #include <ginkgo/core/factorization/lu.hpp>
 #include <ginkgo/core/matrix/dense.hpp>
 #include <ginkgo/core/preconditioner/jacobi.hpp>
@@ -27,6 +29,7 @@
 
 #include "core/base/dispatch_helper.hpp"
 #include "core/components/fill_array_kernels.hpp"
+#include "core/distributed/helpers.hpp"
 #include "core/solver/ir_kernels.hpp"
 #include "core/solver/multigrid_kernels.hpp"
 #include "core/solver/solver_base.hpp"
@@ -78,26 +81,6 @@ std::enable_if_t<!is_complex_s<ValueType>::value && is_complex_s<T>::value,
 casting(const T& x)
 {
     return static_cast<ValueType>(real(x));
-}
-
-/**
- * as_vec gives a shortcut for casting pointer to dense.
- */
-template <typename ValueType>
-auto as_vec(std::shared_ptr<LinOp> x)
-{
-    return std::static_pointer_cast<matrix::Dense<ValueType>>(x);
-}
-
-
-/**
- * as_real_vec gives a shortcut for casting pointer to dense with real type.
- */
-template <typename ValueType>
-auto as_real_vec(std::shared_ptr<LinOp> x)
-{
-    return std::static_pointer_cast<matrix::Dense<remove_complex<ValueType>>>(
-        x);
 }
 
 
@@ -212,16 +195,42 @@ struct MultigridState {
     /**
      * allocate_memory is a helper function to allocate the memory of one level
      *
-     * @tparam ValueType  the value type of memory
+     * @tparam VectorType  the vector type
      *
      * @param level  the current level index
      * @param cycle  the multigrid cycle
      * @param current_nrows  the number of rows of current fine matrix
      * @param next_nrows  the number of rows of next coarse matrix
      */
-    template <typename ValueType>
+    template <typename VectorType>
     void allocate_memory(int level, multigrid::cycle cycle,
                          size_type current_nrows, size_type next_nrows);
+
+#if GINKGO_BUILD_MPI
+    /**
+     * allocate_memory is a helper function to allocate the memory of one level
+     *
+     * @tparam VectorType  the vector type
+     *
+     * @param level  the current level index
+     * @param cycle  the multigrid cycle
+     * @param current_comm  the communicator of the current fine matrix
+     * @param next_comm  the communicator of the next coarse matrix
+     * @param current_nrows  the number of rows of the current fine matrix
+     * @param next_nrows  the number of rows of the next coarse matrix
+     * @param current_local_nrows  the number of rows of the local operator of
+     *                             the current fine matrix
+     * @param next_local_nrows  the number of rows of the local operator of the
+     *                          next coarse matrix
+     */
+    template <typename VectorType>
+    void allocate_memory(int level, multigrid::cycle cycle,
+                         experimental::mpi::communicator& current_comm,
+                         experimental::mpi::communicator& next_comm,
+                         size_type current_nrows, size_type next_nrows,
+                         size_type current_local_nrows,
+                         size_type next_local_nrows);
+#endif
 
     /**
      * run the cycle of the level
@@ -240,11 +249,11 @@ struct MultigridState {
     /**
      * @copydoc run_cycle
      *
-     * @tparam ValueType  the value type
+     * @tparam VectorType  the vector type
      *
      * @note it is the version with known ValueType
      */
-    template <typename ValueType>
+    template <typename VectorType>
     void run_cycle(multigrid::cycle cycle, size_type level,
                    const std::shared_ptr<const LinOp>& matrix, const LinOp* b,
                    LinOp* x, cycle_mode mode);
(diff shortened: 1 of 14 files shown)


greole:
Would that allow to use Multigrid as a preconditioner without Schwarz?

MarcelKoch:
Right now, you use the partition only to get the local size of the matrix, which you can also get from the local matrix. The stored matrix data is not used at all. I would suggest removing these until they are actually necessary.

pratikvn:
@greole , Yes, but a coarse generation algorithm ( that is distributed capable) is necessary. Meaning we need to have the equivalent of AMGX which generates the triplet (R, A_c, P).

pratikvn:
@MarcelKoch, yes, I dont intend to merge this yet. I just wanted to show what changes could be necessary. At present we do not need the partition or the matrix_data.

yhmtsai:
@pratikvn could you rebase it? I think some changes are related to the schwarz?

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -508,6 +526,27 @@ class Matrix
                     ptr_param<const LinOp> local_matrix_template,
                     ptr_param<const LinOp> non_local_matrix_template);
 
+    explicit Matrix(std::shared_ptr<const Executor> exec,
```
greole, line 529: This could use a docstring. Also, what is the use case for this constructor, since it will leave non_local_matrix_ initialized? Is it for testing purposes only?

yhmtsai: It's for testing currently.

core/test/mpi/distributed/solver/multigrid.cpp
```diff
@@ -0,0 +1,180 @@
...
+            .with_max_levels(2u)
+            .with_mg_level(mg_level::build())
+            .with_criteria(gko::stop::Iteration::build().with_max_iters(1u))
+            .with_min_coarse_rows(1u);
+
+    gko::size_type n = 5;
+    gko::size_type global_n = n * this->comm.size();
+    auto mtx =
+        dist_mtx_type::create(this->ref, this->comm, gko::dim<2>{global_n},
+                              DummyLinOp::create(this->ref, gko::dim<2>{n}));
```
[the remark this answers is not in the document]
yhmtsai: I think the original distributed matrix `create` interface does not build the matrix fully yet. It requires the following read to ensure the corresponding data correct. I use the new one to simply set up the testing linop.

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -339,6 +355,8 @@ class Matrix
         ptr_param<const Partition<local_index_type, global_index_type>>
             col_partition);
 
+    dim<2> get_local_size() const override { return local_mtx_->get_size(); }
+
```
greole, line 359: If we add this we should also have 
```suggestion
    dim<2> get_local_size() const override { return local_mtx_->get_size(); }
     
    dim<2> get_non_local_size() const override { return non_local_mtx_->get_size(); }

```

yhmtsai: I was thinking adding it to the same feature class. However, distributed vector does not have this information.
I currently only have get_local_size in vector and matrix. Maybe we can add non local version when we need it afterwards

include/ginkgo/core/distributed/matrix.hpp
yhmtsai: I also use it to generate the local matrix-only distributed op in distributed pgm

core/test/mpi/distributed/solver/multigrid.cpp
```diff
@@ -0,0 +1,212 @@
...
+    void apply_impl(const gko::LinOp* b, gko::LinOp* x) const override {}
+
+    void apply_impl(const gko::LinOp* alpha, const gko::LinOp* b,
+                    const gko::LinOp* beta, gko::LinOp* x) const override
+    {}
+};
+
+
+template <typename ValueType>
+class DummyMultigridLevelWithFactory
```
greole, line 50: I think this needs some documentation why it is actually needed. Also, can't the MultigridLevel implementation adapted to provide that functionality? This seems a bit much for testing purposes.

core/test/mpi/distributed/solver/multigrid.cpp
```diff
@@ -0,0 +1,212 @@
...
+using value_type = float;
+using local_index_type = int;
+using global_index_type = int;
+using dist_mtx_type =
+    gko::experimental::distributed::Matrix<value_type, local_index_type,
+                                           global_index_type>;
+using dist_vec_type = gko::experimental::distributed::Vector<value_type>;
+
+
+class DummyLinOp : public gko::EnableLinOp<DummyLinOp>,
```
greole, line 32: I guess that is needed to construct a DistributedMatrix without fully constructing `local_mtx` and `non_local_mtx` but getting access to the size. Isn't that the same as CustomLinOp? Also if it is implemented at several places maybe we should have a common implementation and reuse it.

core/solver/multigrid.cpp
```diff
@@ -212,17 +195,43 @@ struct MultigridState {
...
+     * @param next_nrows  the number of rows of the next coarse matrix
+     * @param current_local_nrows  the number of rows of the local operator of
+     *                             the current fine matrix
+     * @param next_local_nrows  the number of rows of the local operator of the
+     *                          next coarse matrix
+     */
+    template <typename VectorType>
+    void allocate_memory(int level, multigrid::cycle cycle,
+                         experimental::mpi::communicator& current_comm,
+                         experimental::mpi::communicator& next_comm,
```
greole, line 229: Can't/Shouldn't they be const?
```suggestion
                         const experimental::mpi::communicator& current_comm,
                         const experimental::mpi::communicator& next_comm,
```

greole (commented):
Some more comments from my side. Looks good to me in general. Maybe the tests can be simplified to avoid the implementation of the Dummy classes.

core/test/mpi/distributed/solver/multigrid.cpp
yhmtsai: Yes, it is entirely for the testing process. It's good to have when we do not really have the distributed multigridlevel generation at the moment. Also, the simple dummy component can simplify the test dependency, which means we do not need to rely on another maybe-failed components.

include/ginkgo/core/distributed/base.hpp
```diff
@@ -67,6 +67,21 @@ class DistributedBase {
 };
 
 
+/**
+ * DistributedLocalSize is a feature class providing `get_local_size` which
+ * return the size of local operator.
+ */
+class DistributedLocalSize {
```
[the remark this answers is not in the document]
yhmtsai: I was also thinking about putting it into the `DistributedBase`, but I wonder it is out of the scope because it only handles the communicator now.
`non_local_size` can not go into this base because the vector does not have the idea about non_local_size.
Or, we just put non_local_size of vector is dim<2>(0, 0).
I do not have an idea about the usage of non_local_size out of the matrix

core/test/mpi/distributed/solver/multigrid.cpp
```diff
@@ -0,0 +1,212 @@
...
+        dist_mtx_type::create(this->ref, this->comm, gko::dim<2>{global_n},
+                              DummyLinOp::create(this->ref, gko::dim<2>{n})));
+    auto b = dist_vec_type::create(this->ref, this->comm,
+                                   gko::dim<2>{global_n, 1}, gko::dim<2>{n, 1});
+    auto x = dist_vec_type::create(this->ref, this->comm,
+                                   gko::dim<2>{global_n, 1}, gko::dim<2>{n, 1});
+
+    auto mg = mg_factory->generate(mtx);
+
+    ASSERT_NO_THROW(mg->apply(b, x));
```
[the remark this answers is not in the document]
yhmtsai: I think distributed coarsening method usually different from the non-distributed one?

yhmtsai:
@greole yes, the coarsening may affect the non_local_matrix, but coarsening method will take care of it not the multigrid itself.
In distributed multigrid, we have prepared the distributed matrix each level already, so we only need to deal with the vector according to each distributed matrix (the usage of local size here for creating the vector).

include/ginkgo/core/distributed/base.hpp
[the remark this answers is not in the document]
yhmtsai: No, I only mean the functionality but don't mean put it into the DistributedLocalSize.
I agree with user may expect something similar to `get_local_size()` for `non_local_matrix()`.
For me, if we need to get_non_local_size(), we need to put `get_local_size()` into `DistributedLocalSize` and `get_non_local_size()` into `DistributedNonLocalSize` because vector does not have non_local information such that we can not put them together into `DistrbutedBase`. One in `DistrbutedBase` and another in `DistributedNonLocalSize` are also inconsistent in my mind. (unless we put non_local_size = 0 in vector cases)

core/solver/multigrid.cpp
yhmtsai: good catch!

core/test/mpi/distributed/solver/multigrid.cpp
yhmtsai: partially. they are not the same. the dummy class usually only do the minimal thing for the test. It's kind of trade-off in my mind. but I would say collecting them into a common type is a bit out of this pr scope

core/test/mpi/distributed/solver/multigrid.cpp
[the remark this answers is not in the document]
pratikvn: As you are testing the solver here, I think checking if the two solutions are within some tolerance makes sense.

include/ginkgo/core/distributed/base.hpp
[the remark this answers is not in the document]
pratikvn: I think it should be okay to have only local_size in DistributedBase. A separate interface class is too clunky IMO. But given that this probably will be overwritten, I dont have a strong opinion, and we can discuss this later.

pratikvn (commented):
I think from my side, this looks good to be merged.

core/test/mpi/distributed/solver/multigrid.cpp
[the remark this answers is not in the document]
yhmtsai: ah, for this situation, I can not do it because we do not have actual distributed coarsening method.
I think the current test should be enough because the distributed multigrid actually only handle the vector allocation. These tests ensure the allocation is correct and the corresponding apply works correctly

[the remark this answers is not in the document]
pratikvn: Ah, yes. Maybe make a note/TODO here for now, and it should be okay to add it later

include/ginkgo/core/distributed/base.hpp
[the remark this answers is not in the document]
yhmtsai: I move it to DistributedBase now.

yhmtsai (approved):
due to github pr limitation, approve on behalf of @pratikvn

greole (approved):
LGTM! Just a final question, in order to use it we also need to merge a distributed coarsening method, like PGM?

pratikvn:
@greole, yes, #1403 would need to be merged as well. @yhmtsai , please feel to merge this when you are ready. If you need me to merge it, let me know.

PR #1271, opened 2023-02-07
Use CMAKE_FULL_INSTALL_*_DIR for pkg-config file

keldu:
The ${prefix}/GINKGO_INSTALL_LIBRARY_DIR has the problem that in case of absolute paths a simple string concat produces an invalid path. Added the FULL variable to correctly handle absolute paths.


cmake/install_helpers.cmake
```diff
@@ -5,6 +5,9 @@ include(GNUInstallDirs)
 set(GINKGO_INSTALL_INCLUDE_DIR "${CMAKE_INSTALL_INCLUDEDIR}")
 set(GINKGO_INSTALL_LIBRARY_DIR "${CMAKE_INSTALL_LIBDIR}")
 set(GINKGO_INSTALL_RUNTIME_DIR "${CMAKE_INSTALL_BINDIR}")
+set(GINKGO_INSTALL_FULL_INCLUDE_DIR "${CMAKE_INSTALL_FULL_INCLUDEDIR}")
+set(GINKGO_INSTALL_FULL_RUNTIME_DIR "${CMAKE_INSTALL_FULL_BINDIR}")
+set(GINKGO_INSTALL_FULL_LIB_DIR "${CMAKE_INSTALL_FULL_LIBDIR}")
```
yhmtsai, line 10:
```suggestion
set(GINKGO_INSTALL_FULL_LIBRARY_DIR "${CMAKE_INSTALL_FULL_LIBDIR}")
```
or change the name in another file?

keldu:
Do we need to add a changelog in the wiki?

cmake/install_helpers.cmake
keldu: You are right, I don't know how I missed this.
I tested it on my side with a patch file which added the cmake variable instead of a ginkgo helper.

yhmtsai (approved):
LGTM. I think so. you can put a rough description like pkg-config related fix. we can collect the further pkg-config fix into the same list.

pratikvn (approved):

upsj (approved):

yhmtsai:
rebase!

PR #1272, opened 2023-02-09
Fixes read for skew-symmetric matrices

MarcelKoch:
This PR adds a check preventing the duplication of diagonal entries during the skew-symmetric read. Of course, the diagonal element can only be zero, but even duplicating that leads to issues somewhere in our code. This is helpful if the matrix should explicitly contain the diagonal entries.

### DIFF
--- a/core/base/mtx_io.cpp
+++ b/core/base/mtx_io.cpp
@@ -405,7 +405,9 @@ private:
             matrix_data<ValueType, IndexType>& data) const override
         {
             data.nonzeros.emplace_back(row, col, entry);
-            data.nonzeros.emplace_back(col, row, -entry);
+            if (row != col) {
+                data.nonzeros.emplace_back(col, row, -entry);
+            }
         }
 
         /**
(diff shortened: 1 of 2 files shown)


upsj (approved):
LGTM! Not sure if the MatrixMarket even allows diagonal entries for skew-symmetric matrices, but this is a useful fix in any case.

yhmtsai (approved):
LGTM. Could you describe what the error you face in the code without this fix?

MarcelKoch:
@yhmtsai I had a .mtx file with a skew-sym matrix and I had to explicitly store the zero diagonal entries, because I wanted to call `add_scaled_identity` on that matrix later on. That only works, if the diagonal entries are present. But with the old code, the matrix_data would contain duplicate entries, which leads to incorrect results somewhere down the line. I think I've mentioned the issue with duplicated matrix_data entries somewhere else before, but can't find it right now.

MarcelKoch:
rebase!

MarcelKoch:
rebase!

PR #1275, opened 2023-02-12
Add vector mean computation

greole:
This PR adds the computation of column wise arithmetic mean to dense and the distributed vector classes.


greole:
format!

include/ginkgo/core/distributed/vector.hpp
```diff
@@ -404,6 +404,32 @@ class Vector
...
+    /**
+     * Computes the column-wise mean of this (multi-)vector using a global
+     * reduction.
+     *
+     * @param result  a Dense row matrix, used to store the mean
+     *                (the number of columns in result must match the number
+     *                of columns of this)
+     * @param tmp  the temporary storage to use for partial sums during the
+     *             reduction computation. It may be resized and/or reset to the
+     *             correct executor.
```
yhmtsai, line 416: it does not have the parameter here

greole: Also, I guess, `ptr_param<LinOp> result` is the way to go here?

greole:
format!

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -1215,6 +1236,14 @@ class Dense
      */
     virtual void compute_squared_norm2_impl(LinOp* result) const;
 
+    /**
+     * @copydoc compute_mean(LinOp*) const
+     *
+     * @deprecated  This function will be removed in the future,
```
MarcelKoch, line 1242: Is that comment correct?

include/ginkgo/core/distributed/vector.hpp
```diff
@@ -404,6 +404,42 @@ class Vector
...
+     * global reduction.
+     *
+     * @param result  a Dense row matrix, used to store the mean
+     *                (the number of columns in result must match the number
+     *                of columns of this)
+     * @param tmp  the temporary storage to use for partial sums during the
+     *             reduction computation. It may be resized and/or reset to the
+     *             correct executor.
+     */
+    void compute_mean(LinOp* result, array<char>& tmp) const;
```
MarcelKoch, line 441: I don't think this one is necessary, since there is the version with `ptr_param`

MarcelKoch: Also, it is not implemented

core/distributed/vector.cpp
```diff
@@ -573,19 +573,62 @@ void Vector<ValueType>::compute_squared_norm2(ptr_param<LinOp> result,
...
+
+template <typename ValueType>
+void Vector<ValueType>::compute_mean(ptr_param<LinOp> result,
+                                     array<char>& tmp) const
+{
+    using MeanVector = local_vector_type;
+    const auto global_size = this->get_size()[0];
+    const auto local_size = this->get_local_vector()->get_size()[0];
+    const auto num_vecs = static_cast<int>(this->get_size()[1]);
+    GKO_ASSERT_EQUAL_DIMENSIONS(result, dim<2>(1, num_vecs));
```
MarcelKoch, line 592: There is `GKO_ASSERT_EQUAL_COLS`, which might be better suited

core/distributed/vector.cpp
```diff
@@ -573,19 +573,62 @@ void Vector<ValueType>::compute_squared_norm2(ptr_param<LinOp> result,
...
+    auto exec = this->get_executor();
+    const auto comm = this->get_communicator();
+    auto dense_res = make_temporary_clone(exec, as<MeanVector>(result));
+    this->get_local_vector()->compute_mean(dense_res.get());
+
+    // scale by its weight ie ratio of local to global size
+    auto weight = initialize<matrix::Dense<remove_complex<ValueType>>>(
+        1, {static_cast<remove_complex<ValueType>>(local_size) / global_size},
```
MarcelKoch, line 600: the `1` should be unnecessary

core/matrix/dense.cpp
```diff
@@ -496,6 +497,29 @@ void Dense<ValueType>::compute_squared_norm2(ptr_param<LinOp> result,
...
+    auto exec = this->get_executor();
+    this->compute_mean_impl(make_temporary_output_clone(exec, result).get());
+}
+
+
+template <typename ValueType>
+void Dense<ValueType>::compute_mean(ptr_param<LinOp> result,
+                                    array<char>& tmp) const
+{
+    GKO_ASSERT_EQUAL_DIMENSIONS(result, dim<2>(1, this->get_size()[1]));
```
MarcelKoch, line 512: again, the col version can be used.

core/matrix/dense.cpp
```diff
@@ -506,6 +530,17 @@ void Dense<ValueType>::compute_squared_norm2_impl(LinOp* result) const
 }
 
 
+template <typename ValueType>
+void Dense<ValueType>::compute_mean_impl(LinOp* result) const
```
MarcelKoch, line 534: either this should call `compute_mean(ptr_param<LinOp>, array` directly, or it should also get the `array` parameter and get called by `compute_mean`.

common/unified/matrix/dense_kernels.template.cpp
```diff
@@ -278,6 +278,22 @@ void compute_norm1(std::shared_ptr<const DefaultExecutor> exec,
...
+template <typename ValueType>
+void compute_mean(std::shared_ptr<const DefaultExecutor> exec,
+                  const matrix::Dense<ValueType>* x,
+                  matrix::Dense<ValueType>* result, array<char>& tmp)
+{
+    using ValueType_nc = gko::remove_complex<ValueType>;
+    run_kernel_col_reduction_cached(
+        exec,
+        [] GKO_KERNEL(auto i, auto j, auto x, auto total_size) {
+            return x(i, j) / static_cast<ValueType_nc>(total_size);
```
MarcelKoch, line 290: it's a very minor thing, but I would suggest precomputing the inverse outside the kernel.

reference/matrix/dense_kernels.cpp
```diff
@@ -397,6 +397,28 @@ void compute_norm1(std::shared_ptr<const ReferenceExecutor> exec,
...
+                  const matrix::Dense<ValueType>* x,
+                  matrix::Dense<ValueType>* result, array<char>&)
+{
+    using ValueType_nc = gko::remove_complex<ValueType>;
+    for (size_type j = 0; j < x->get_size()[1]; ++j) {
+        result->at(0, j) = zero<ValueType>();
+    }
+
+    for (size_type i = 0; i < x->get_size()[0]; ++i) {
+        const ValueType_nc alpha = static_cast<ValueType_nc>(i) / (i + 1);
```
MarcelKoch, line 411: maybe also do this in two steps (compute sum, then scaling), instead of using these weights.

reference/test/matrix/dense_kernels.cpp
```diff
@@ -700,6 +700,30 @@ TYPED_TEST(Dense, ComputesNorm1Mixed)
...
 
+TYPED_TEST(Dense, ComputesMean)
+{
+    using Mtx = typename TestFixture::Mtx;
+    using T = typename TestFixture::value_type;
+    auto result = Mtx::create(this->exec, gko::dim<2>{1, 3});
+
+    this->mtx4->compute_mean(result.get());
+
+    GKO_EXPECT_NEAR(result->at(0, 0), T{0.5}, 1e-6);
```
MarcelKoch, line 711: instead of `1e-6` use `r<T>::value * C`, with `C=10` or `100` to make it pass.

MarcelKoch (changes requested):
LGTM, some minor comments.

core/matrix/dense.cpp
```diff
@@ -496,6 +497,29 @@ void Dense<ValueType>::compute_squared_norm2(ptr_param<LinOp> result,
...
+                                    array<char>& tmp) const
+{
+    GKO_ASSERT_EQUAL_DIMENSIONS(result, dim<2>(1, this->get_size()[1]));
+    auto exec = this->get_executor();
+    if (tmp.get_executor() != exec) {
+        tmp.clear();
+        tmp.set_executor(exec);
+    }
+    auto dense_res = make_temporary_conversion<ValueType>(result);
+    exec->run(dense::make_compute_mean(this, dense_res.get(), tmp));
```
yhmtsai, line 519: call `this->compute_mean_impl(make_temporary_output_clone(exec, result).get(), tmp);` directly

include/ginkgo/core/distributed/vector.hpp
yhmtsai, line 441: because the ptr_param already handles this case, you do not need that here

reference/matrix/dense_kernels.cpp
```diff
@@ -397,6 +397,28 @@ void compute_norm1(std::shared_ptr<const ReferenceExecutor> exec,
...
+        const ValueType_nc beta = static_cast<ValueType_nc>(1) / (i + 1);
+        for (size_type j = 0; j < x->get_size()[1]; ++j) {
+            result->at(0, j) = alpha * result->at(0, j) + beta * x->at(i, j);
+        }
```
yhmtsai, line 415: Is it due to the numerical stability?
Otherwise, you can sum all first and then average.

reference/test/matrix/dense_kernels.cpp
```diff
@@ -700,6 +700,30 @@ TYPED_TEST(Dense, ComputesNorm1Mixed)
...
+    GKO_EXPECT_NEAR(result->at(0, 1), T{4.0}, 1e-6);
+    GKO_EXPECT_NEAR(result->at(0, 2), T{1.0}, 1e-6);
```
yhmtsai, line 713: use ` r<T>::value` I think?

test/mpi/vector.cpp
```diff
@@ -675,6 +675,27 @@ TYPED_TEST(VectorReductions, ComputeSquaredNorm2WithTmpIsSameAsDense)
                         r<value_type>::value);
 }
 
+TYPED_TEST(VectorReductions, ComputesMeanIsSameAsDense)
```
yhmtsai, line 678:
```suggestion


TYPED_TEST(VectorReductions, ComputesMeanIsSameAsDense)
```

test/mpi/vector.cpp
```diff
@@ -675,6 +675,27 @@ TYPED_TEST(VectorReductions, ComputeSquaredNorm2WithTmpIsSameAsDense)
...
+    using value_type = typename TestFixture::value_type;
+    this->init_result();
+
+    this->x->compute_mean(this->res);
+    this->dense_x->compute_mean(this->dense_res);
+
+    GKO_ASSERT_MTX_NEAR(this->res, this->dense_res, r<value_type>::value);
+}
+
+TYPED_TEST(VectorReductions, ComputesMeanWithTmpIsSameAsDense)
```
yhmtsai, line 689:
```suggestion
}


TYPED_TEST(VectorReductions, ComputesMeanWithTmpIsSameAsDense)
```

test/mpi/vector.cpp
```diff
@@ -675,6 +675,27 @@ TYPED_TEST(VectorReductions, ComputeSquaredNorm2WithTmpIsSameAsDense)
...
+    using value_type = typename TestFixture::value_type;
+    this->init_result();
+
+    this->x->compute_mean(this->res, this->tmp);
+    this->dense_x->compute_mean(this->dense_res, this->dense_tmp);
+
+    GKO_ASSERT_MTX_NEAR(this->res, this->dense_res, r<value_type>::value);
+}
 
 TYPED_TEST(VectorReductions, ComputeDotCopiesToHostOnlyIfNecessary)
```
yhmtsai, line 700:
```suggestion
}


TYPED_TEST(VectorReductions, ComputeDotCopiesToHostOnlyIfNecessary)
```

include/ginkgo/core/distributed/vector.hpp
yhmtsai: actually, there's no implementation of it in vector.cpp

tcojean (approved):
LGTM. No other comment.

core/distributed/vector.cpp
```diff
@@ -573,19 +573,62 @@ void Vector<ValueType>::compute_squared_norm2(ptr_param<LinOp> result,
...
+        this->get_executor());
```
yhmtsai, line 601: Not sure whether it is performance critical. 
Another way to do that is to pass the total size into compute_mean, and then we can use reduce directly without additional kernel.

yhmtsai (approved):
only some nit

reference/matrix/dense_kernels.cpp
greole: yes numerical stability is the main reason here. Imagine having a large number of entries which might lead to overflow.

upsj: `reference` is usually a simple implementation, it doesn't need to have the best numerical stability compared to the other kernels. I also think a simpler approach would be appropriate.

include/ginkgo/core/matrix/dense.hpp
greole: Ah, I copy/pasted it and didn't change it. However, what are we trying to convey with the deprecated note here?

MarcelKoch: I think there was a switch between cublas/etc and ginkgo implementation for dot products or similar.

reference/matrix/dense_kernels.cpp
greole: Problem here is numerical stability, I would like to have a non-overflowing version for large vectors. Initially, I thought about having a flag to dispatch between a safe and unsafe version.

MarcelKoch: Ok, that sounds fine for the reference case.

core/distributed/vector.cpp
```diff
@@ -573,19 +573,62 @@ void Vector<ValueType>::compute_squared_norm2(ptr_param<LinOp> result,
...
+
+template <typename ValueType>
+void Vector<ValueType>::compute_mean(ptr_param<LinOp> result,
+                                     array<char>& tmp) const
+{
+    using MeanVector = local_vector_type;
+    const auto global_size = this->get_size()[0];
+    const auto local_size = this->get_local_vector()->get_size()[0];
+    const auto num_vecs = static_cast<int>(this->get_size()[1]);
+    GKO_ASSERT_EQUAL_COLS(result, dim<2>(1, num_vecs));
```
MarcelKoch, line 592:
```suggestion
    GKO_ASSERT_EQUAL_COLS(result, this);
```
a few characters shorter

core/matrix/dense.cpp
```diff
@@ -496,6 +497,29 @@ void Dense<ValueType>::compute_squared_norm2(ptr_param<LinOp> result,
...
+    auto exec = this->get_executor();
+    this->compute_mean_impl(make_temporary_output_clone(exec, result).get());
+}
+
+
+template <typename ValueType>
+void Dense<ValueType>::compute_mean(ptr_param<LinOp> result,
+                                    array<char>& tmp) const
+{
+    GKO_ASSERT_EQUAL_COLS(result, dim<2>(1, this->get_size()[1]));
```
MarcelKoch, line 512: same

MarcelKoch (approved):

common/unified/matrix/dense_kernels.template.cpp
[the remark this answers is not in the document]
greole: Same as my other comment, I would leave it as it is for now, and conduct some more experiments if it matters at a later point.

core/distributed/vector.cpp
```diff
@@ -573,19 +573,62 @@ void Vector<ValueType>::compute_squared_norm2(ptr_param<LinOp> result,
...
+
 template <typename ValueType>
 ValueType Vector<ValueType>::at_local(size_type row,
                                       size_type col) const noexcept
 {
     return local_.at(row, col);
 }
 
+
 template <typename ValueType>
```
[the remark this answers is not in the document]
yhmtsai: I think it's not fixed?

yhmtsai:
Before merging it, could you rephrase the commit history?

PR #1277, opened 2023-02-14
Improve kernel operation naming

upsj:
following a discussion with @MarcelKoch in #1270, I realized that we are not using the full name of the kernel in GKO_REGISTER_OPERATION, which would improve its uniqueness. Since we don't have any name collisions between kernels, I think it makes sense to use the full name below `gko::kernels::EXEC::` instead of just the kernel name with the number of arguments added.

Before:
```
            "jacobi-32-0,0": {
                "generate": {
                    "components": {
                        "allocate": 0.0000084584,
                        "free": 0.0000048576,
                        "find_blocks": 0.0000039604,
                        "generate": 0.0003623217
                    },
                    "time": 0.0003646641,
                    "repetitions": 10
                },
                "apply": {
                    "components": {
                        "simple_apply": 0.0003813105
                    },
                    "time": 0.0003896164,
                    "repetitions": 10
                },
                "completed": true
            }
```

After:
```
            "jacobi-32-0,0": {
                "generate": {
                    "components": {
                        "allocate": 0.0000084584,
                        "free": 0.0000048576,
                        "jacobi::find_blocks": 0.0000039604,
                        "jacobi::generate": 0.0003623217
                    },
                    "time": 0.0003646641,
                    "repetitions": 10
                },
                "apply": {
                    "components": {
                        "jacobi::simple_apply": 0.0003813105
                    },
                    "time": 0.0003896164,
                    "repetitions": 10
                },
                "completed": true
            }
```

### DIFF
--- a/include/ginkgo/core/base/executor.hpp
+++ b/include/ginkgo/core/base/executor.hpp
@@ -332,7 +332,7 @@ namespace detail {
  * It is used to implement the @ref GKO_REGISTER_OPERATION macro.
  *
  * @tparam Closure  the type of the functor of taking parameters
- *                  (std::shared_ptr<const Executor>, int)
+ *                  (std::shared_ptr<const Executor>)
  */
 template <typename Closure>
 class RegisteredOperation : public Operation {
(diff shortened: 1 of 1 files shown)


yhmtsai:
Should we still keep the #arg for distinguishing some overloading case?

upsj:
@yhmtsai I can't find any case where we actually overloaded an operation name. Do you see an example?

yhmtsai (approved):
LGTM. no, I do not check whether there's overload or not. I only think it is a case will merge two kernel together

upsj:
@yhmtsai I went through all kernels (though not very thoroughly) and couldn't find an example, so I think this should not be a concern.

thoasm (approved):
LGTM!
Since `RegisteredOperation` is in namespace `detail`, this is not an interface break (thought so at first).

upsj:
rebase!

tcojean (approved):
LGTM

PR #1279, opened 2023-02-15
Allow passing smart pointers to functions directly (interface changes only)

upsj:
To simplify reviewing #1261, this contains only the public interface changes


MarcelKoch (approved):
Since there are no new additions compared to the other PR, as far as I can see, I just copy my review.

thoasm (approved):
LGTM!
for some nits, see my comments in PR #1261:
https://github.com/ginkgo-project/ginkgo/pull/1261#pullrequestreview-1287845165

include/ginkgo/core/base/utils_helper.hpp
```diff
@@ -328,6 +372,43 @@ inline const typename std::decay<T>::type* as(const U* obj)
...
+ * @tparam T  requested result type
+ * @tparam U  static type of the passed object
+ *
+ * @param obj  the object which should be converted
+ *
+ * @return If successful, returns a pointer to the subtype, otherwise throws
+ *         NotSupported.
+ */
+template <typename T, typename U>
+inline typename std::decay<T>::type* as(pointer_param<U> obj)
```
yhmtsai, line 387: we only accept the unique_ptr<U>&& in the above line for the clear ownership.
Won't it destory the unique_ptr case?
pass unique_ptr and implicitly convert to pointer_param and get the raw pointer
It also gives the different type not like the other case return the same type (shared->shared, raw->raw..)

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -154,6 +155,36 @@ class Dense
...
+    using ConvertibleTo<Hybrid<ValueType, int64>>::convert_to;
+    using ConvertibleTo<Hybrid<ValueType, int64>>::move_to;
+    using ConvertibleTo<Sellp<ValueType, int32>>::convert_to;
+    using ConvertibleTo<Sellp<ValueType, int32>>::move_to;
+    using ConvertibleTo<Sellp<ValueType, int64>>::convert_to;
+    using ConvertibleTo<Sellp<ValueType, int64>>::move_to;
+    using ConvertibleTo<SparsityCsr<ValueType, int32>>::convert_to;
+    using ConvertibleTo<SparsityCsr<ValueType, int32>>::move_to;
+    using ConvertibleTo<SparsityCsr<ValueType, int64>>::convert_to;
+    using ConvertibleTo<SparsityCsr<ValueType, int64>>::move_to;
```
yhmtsai, line 187: could we mark `convert_to_impl` virtual and `convert_to` using the convert_to_impl in ConvertiableTo? (`convert_to` still need to be virtual to keep public interface)
then all class override `convert_to_impl` not `convert_to`

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -456,26 +491,6 @@ class Dense
...
-     * @param gather_indices  pointer to an array containing row indices
-     *                        from this matrix. It may contain duplicates.
-     * @param row_collection  pointer to a Dense matrix that will store the
-     *                      gathered rows:
-     *                      `row_collection(i,j) = input(gather_indices(i), j)`
-     *                      It must have the same number of columns as this
-     *                      matrix and `gather_indices->get_num_elems()` rows.
-     */
-    void row_gather(const array<int32>* gather_indices,
-                    Dense* row_collection) const;
```
yhmtsai, line 471: where be this moved to?

yhmtsai (changes requested):
how about to use `pointer` or `raw_pointer`?
the main concern is `as` on `pointer_param` from `unique_ptr`

core/matrix/dense.cpp
```diff
@@ -1431,17 +1432,19 @@ std::unique_ptr<LinOp> Dense<ValueType>::inverse_row_permute(
...
 {
-    this->inverse_row_permute_impl(permutation_indices, output);
+    this->inverse_row_permute_impl(permutation_indices, output.get());
 }
 
 
 template <typename ValueType>
 void Dense<ValueType>::inverse_row_permute(
-    const array<int64>* permutation_indices, Dense<ValueType>* output) const
+    const array<int64>* permutation_indices,
```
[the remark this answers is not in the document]
upsj: this should be a reference, not a pointer IMO, there is no reason to put arrays into a smart pointer, so using pointer_param here seems wrong.

include/ginkgo/core/base/utils_helper.hpp
```diff
@@ -56,29 +56,71 @@ namespace gko {
...
+ * This class is used for function parameters in the place of raw pointers.
+ * It can be converted to from raw pointers, shared pointers and unique pointers
+ * of the specified type or any derived type. This allows functions to be called
+ * without having to use gko::lend or calling .get() for every pointer argument.
+ * It probably has no use outside of function parameters, as it is immutable.
+ *
+ * @tparam T  the pointed-to type
+ */
+template <typename T>
+class pointer_param {
```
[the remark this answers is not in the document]
upsj: this should only be used as a parameter, thus the `_param` suffix. otherwise people might accidentally use it for variables in their, see the discussion in #179

include/ginkgo/core/base/utils_helper.hpp
upsj: This will never happen, because unique_ptr isn't an exact match for pointer_param, thus template argument deduction will fail.

include/ginkgo/core/base/executor.hpp
```diff
@@ -879,6 +879,8 @@ class Executor : public log::EnableLogging<Executor> {
         this->EnableLogging<Executor>::remove_logger(logger);
     }
 
+    using EnableLogging<Executor>::remove_logger;
+
```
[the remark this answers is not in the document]
upsj: the overload with pointer_param still needs to be pulled in, similar to all the matrix types. For a detailed discussion, see https://github.com/ginkgo-project/ginkgo/pull/1261#discussion_r1094302484

include/ginkgo/core/base/polymorphic_object.hpp
```diff
@@ -712,6 +753,8 @@ template <typename ConcreteType, typename ResultType = ConcreteType>
 class EnablePolymorphicAssignment : public ConvertibleTo<ResultType> {
 public:
     using result_type = ResultType;
+    using ConvertibleTo<result_type>::convert_to;
+    using ConvertibleTo<result_type>::move_to;
```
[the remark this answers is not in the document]
upsj: https://github.com/ginkgo-project/ginkgo/pull/1261#discussion_r1094302484

include/ginkgo/core/matrix/coo.hpp
```diff
@@ -96,6 +96,12 @@ class Coo : public EnableLinOp<Coo<ValueType, IndexType>>,
 public:
     using EnableLinOp<Coo>::convert_to;
     using EnableLinOp<Coo>::move_to;
+    using ConvertibleTo<Coo<next_precision<ValueType>, IndexType>>::convert_to;
+    using ConvertibleTo<Coo<next_precision<ValueType>, IndexType>>::move_to;
+    using ConvertibleTo<Csr<ValueType, IndexType>>::convert_to;
+    using ConvertibleTo<Csr<ValueType, IndexType>>::move_to;
+    using ConvertibleTo<Dense<ValueType>>::convert_to;
+    using ConvertibleTo<Dense<ValueType>>::move_to;
```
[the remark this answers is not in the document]
upsj: exactly, https://github.com/ginkgo-project/ginkgo/pull/1261#discussion_r1094302484 for reference

include/ginkgo/core/matrix/dense.hpp
upsj: pointer_param<LinOp> accepts Dense*, so this overload is no longer necessary

include/ginkgo/core/matrix/dense.hpp
upsj: This seems more complicated to me than the current solution. This is only temporary anyways, until we can change the virtual functions to take pointer_param directly with the next major release.

include/ginkgo/core/base/polymorphic_object.hpp
```diff
@@ -504,10 +544,11 @@ std::unique_ptr<R, std::function<void(R*)>> copy_and_convert_to_impl(
 {
     auto obj_as_r = dynamic_cast<R*>(obj);
     if (obj_as_r != nullptr && obj->get_executor() == exec) {
+        // FIXME: this breaks lifetimes
         return {obj_as_r, [](R*) {}};
```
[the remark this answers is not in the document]
upsj: It returns a unique_ptr, which may or may not delete the object. Thus the user has no way of knowing if this will respect the lifetime of the object being passed in.

core/distributed/vector.cpp
```diff
@@ -175,16 +177,65 @@ void Vector<ValueType>::read_distributed(
 
 
 template <typename ValueType>
-template <typename LocalIndexType, typename GlobalIndexType>
 void Vector<ValueType>::read_distributed(
-    const matrix_data<ValueType, GlobalIndexType>& data,
-    const Partition<LocalIndexType, GlobalIndexType>* partition)
+    const device_matrix_data<ValueType, int64>& data,
+    pointer_param<const Partition<int64, int64>> partition)
```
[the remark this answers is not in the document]
upsj: like with gko::as(unique_ptr), templated functions can't be used here, because their parameters need to be an exact match, no implicit conversions allowed.

CMakeLists.txt
```diff
@@ -234,7 +234,7 @@ endif()
 set(GINKGO_HAVE_GPU_AWARE_MPI OFF)
 set(GINKGO_FORCE_SPMV_BLOCKING_COMM OFF)
 if(GINKGO_BUILD_MPI)
-    find_package(MPI REQUIRED)
+    find_package(MPI 3.1 COMPONENTS CXX REQUIRED)
```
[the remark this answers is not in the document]
upsj: done

include/ginkgo/core/base/polymorphic_object.hpp
```diff
@@ -175,8 +175,20 @@ class PolymorphicObject : public log::EnableLogging<PolymorphicObject> {
      * @param other  the object to move from
      *
      * @return this
+     *
+     * @tparam Derived  the actual pointee type of the parameter, it needs to be
+     *                  derived from PolymorphicObject.
+     * @tparam Deleter  the deleter of the unique_ptr parameter
```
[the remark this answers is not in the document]
upsj: Don't we usually put @param before @tparam? I'm not sure we are consistent there :smile:

include/ginkgo/core/base/utils_helper.hpp
```diff
@@ -56,29 +56,71 @@ namespace gko {
 class Executor;
 
 
-namespace detail {
+/**
+ * This class is used for function parameters in the place of raw pointers.
```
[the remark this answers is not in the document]
upsj: done

MarcelKoch:
Could `raw_ptr` as a type name make sense? Or `raw_ptr_param`? That would clear up the ownership issue. But then it's not clear that you can pass in a `unique_ptr` or a `shared_ptr`.

core/distributed/matrix.cpp
```diff
@@ -66,15 +66,16 @@ Matrix<ValueType, LocalIndexType, GlobalIndexType>::Matrix(
 template <typename ValueType, typename LocalIndexType, typename GlobalIndexType>
 Matrix<ValueType, LocalIndexType, GlobalIndexType>::Matrix(
     std::shared_ptr<const Executor> exec, mpi::communicator comm,
-    const LinOp* local_matrix_type)
+    pointer_param<const LinOp> local_matrix_type)
```
MarcelKoch, line 69: why did I call that one `_type`???
I think using the template suffix is a bit cleaner, like for the other constructor.
```suggestion
    pointer_param<const LinOp> local_matrix_template)
```

include/ginkgo/core/base/utils_helper.hpp
yhmtsai: you are right. pointer_param always requires the <U> which is not from the constructor for unique_ptr.

include/ginkgo/core/base/polymorphic_object.hpp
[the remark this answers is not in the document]
yhmtsai: we should probably move it to the core header not public header.
from raw_pointer, I do not think we can safely maintain the lifetime.

include/ginkgo/core/base/polymorphic_object.hpp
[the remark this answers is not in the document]
yhmtsai: we are not consistent for the doxygen.
both cases are in the current code

include/ginkgo/core/matrix/dense.hpp
yhmtsai: I think the original one also accepts Dense*, but it can avoid casting to LinOp* and recasting back the Dense?

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -1267,30 +1291,33 @@ struct temporary_clone_helper<matrix::Dense<ValueType>> {
...
  *
  * @param vector  the vector on which to create the view
  */
-template <typename ValueType>
-std::unique_ptr<const matrix::Dense<ValueType>> make_const_dense_view(
-    const matrix::Dense<ValueType>* vector)
+template <typename VecPtr>
+std::unique_ptr<
+    const matrix::Dense<typename detail::pointee<VecPtr>::value_type>>
+make_const_dense_view(VecPtr&& vector)
```
yhmtsai, line 1317: side note: Because create_const_view_of works only for Dense, it should be fine.
Otherwise, it can accept any type with value_type

yhmtsai (approved):
LGTM in general. For the existed user LinOp, they are not required to put the `using ConvertiableTo<UserLinOp>::convert_to`, right? The UserLinOp works as origin but need the lend or .get.

upsj:
@yhmtsai exactly, only the pointer_param overloads are missing if you omit the `using` statements

include/ginkgo/core/matrix/dense.hpp
upsj: Basically, we can't have two overloads that take `pointer_param<Dense>` and `pointer_param<LinOp>`, because the call would then be ambiguous for any `Dense` pointer we pass in. The reason for this is that with raw pointers, more generic types (LinOp) are a worse fit than more specific types (Dense), but this is not true if we have converting constructors like the ones for `pointer_param(T*)`

yhmtsai: With `pointer_param`, we will only allow the BaseType as the public interface, and then we need to dynamic_cast to the accepted types.

upsj: That is true, though dynamic_cast hasn't really been a performance issue so far. I think in the long term (Ginkgo 2.0), we should aim for making the parameter a `Vector` base class, instead of the current overloads being either too specific `Dense<double>` or too generic `LinOp`

yhmtsai: after checking https://godbolt.org/z/8bcbYK8G4 again, the shared_ptr does not work, too. It only work when have derived and base function signature

upsj: yes, we always need exact matches, otherwise the ambiguity will come into play

include/ginkgo/core/base/temporary_clone.hpp
```diff
@@ -218,14 +219,16 @@ detail::temporary_clone<T> make_temporary_clone(
  * @param exec  the executor where the uninitialized clone will be created
  * @param ptr  a pointer to the object of which the clone will be created
  */
-template <typename T>
-detail::temporary_clone<T> make_temporary_output_clone(
-    std::shared_ptr<const Executor> exec, T* ptr)
+template <typename Ptr>
+detail::temporary_clone<std::remove_reference_t<decltype(*std::declval<Ptr>())>>
+make_temporary_output_clone(std::shared_ptr<const Executor> exec, Ptr&& ptr)
```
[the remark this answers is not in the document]
upsj: I think this should probably be using `pointee`, is that sufficient?

upsj:
rebase!

PR #1281, opened 2023-02-16
Remove .cpp include in CB-GMRES test

thoasm:
Remove the `#include "core/solver/cb_gmres.cpp"` that was included in the core CB-GMRES test.
This does not (yet) lead to compiler errors, however, when adding non-templated functions in `core/solver/cb_gmres.cpp`, it fails (with confusing error messages).


pratikvn (approved):

upsj (approved):

yhmtsai (approved):

upsj:
rebase!

PR #1283, opened 2023-02-17
Remove cygwin pipeline

upsj:
Hopefully nobody uses Cygwin for HPC, and it takes forever to build


tcojean (approved):
lgtm

pratikvn (approved):

yhmtsai (approved):
LGTM. The slow issue is partial from the machine limit. 
We can only use 2 thread to avoid the memory issue on these machine.
Could you also delete the cygwin in the README?

PR #1284, opened 2023-02-17
Accept only unique_ptr for distributed vector constructors

MarcelKoch:
Inspired by #1279 I decided to clear up the ownership transfer in the distributed vector constructor. Currently, distributed vectors can be created from existing local vectors by passing them as raw pointers to the constructor. But the constructor will move from the passed in pointer, so this is potentially unsafe. 
One argument in favor of raw pointers was that working with views of local vectors would be simpler (I think?). Since now there is a helper function to create a view of a dense vector, this can now be done as easily with a `unique_ptr` parameter.

To make this work, I had to add `create_const` functions, but I guess that might be useful anyway.

### DIFF
--- a/benchmark/solver/distributed/solver.cpp
+++ b/benchmark/solver/distributed/solver.cpp
@@ -56,11 +56,9 @@ struct Generator : public DistributedDefaultSystemGenerator<SolverGenerator> {
     {
         return Vec::create(
             exec, comm, gko::dim<2>{system_matrix->get_size()[0], FLAGS_nrhs},
-            gko::as<typename LocalGenerator::Vec>(
-                local_generator.generate_rhs(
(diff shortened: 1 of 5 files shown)


upsj (approved):
LGTM! We should probably merge this after the pointer_param PR is merged?

MarcelKoch:
@upsj yes, since this is based on that PR, that makes the most sense.

MarcelKoch:
format-rebase!

fritzgoebel (approved):
LGTM

thoasm (changes requested):
Tests for new functions `create_const` are missing.

thoasm (approved):
LGTM!

upsj (approved):

MarcelKoch:
rebase!

MarcelKoch:
rebase!

PR #1286, opened 2023-02-21
bump the clang-format to 14 for format base

yhmtsai:
This PR bumps the clang-format to 14 and ubuntu-22.04 from github-hosted runner
ubuntu-18.04 is deprecated
https://github.blog/changelog/2022-08-09-github-actions-the-ubuntu-18-04-actions-runner-image-is-being-deprecated-and-will-be-removed-by-12-1-22/

20.04 can use 10.0.0, 11.0.0, 12.0.0 (from apt, there are 6.0, 7, 8, 9, 10, 12)
22.04 can use 12.0.1, 13.0.1, 14.0.0 (from apt, there are 11, 12, 13, 14)
We choose 14 as our base in repo


upsj:
Should we combine this with a run of clang-format over the entire code? I remember we had issues with jacobi.hpp, a few of the common cuda_hip files and so on

MarcelKoch:
I get that choosing the version is quite arbitrary, but I would prefer using the newest one that is available. We are using ubuntu 22 for this, so it doesn't matter what ubuntu 20 supports.

test/utils/executor.hpp
```diff
@@ -61,8 +61,7 @@ std::shared_ptr<ExecType> init_executor(
 
 template <>
 inline std::shared_ptr<gko::ReferenceExecutor>
-    init_executor<gko::ReferenceExecutor>(
-        std::shared_ptr<gko::ReferenceExecutor>)
+init_executor<gko::ReferenceExecutor>(std::shared_ptr<gko::ReferenceExecutor>)
```
yhmtsai, line 64: should we adjust the setting to indent this situation?

upsj (approved):

test/utils/executor.hpp
upsj: I think this is fine (and also will be changed in #1236)

dpcpp/components/reduction.dp.hpp
```diff
@@ -225,13 +225,13 @@ void reduce_add_array(dim3 grid, dim3 block, size_type dynamic_shared_memory,
...
         cgh.parallel_for(
-            sycl_nd_range(grid, block), [=
-        ](sycl::nd_item<3> item_ct1) [[sycl::reqd_sub_group_size(
-                                            DeviceConfig::subgroup_size)]] {
-                reduce_add_array<DeviceConfig>(
-                    size, source, result, item_ct1,
-                    *block_sum_acc_ct1.get_pointer());
-            });
+            sycl_nd_range(grid, block),
+            [=](sycl::nd_item<3> item_ct1)
```
yhmtsai, line 229: BTW, it keeps the old style when using 12 but `[=]` when using 14

CMakeLists.txt
```diff
@@ -338,19 +338,19 @@ if(NOT "${BASH}" STREQUAL "BASH-NOTFOUND" AND GINKGO_DEVEL_TOOLS)
     find_program(GIT git)
     if(NOT "${GIT}" STREQUAL "GIT-NOTFOUND")
         add_custom_target(format_header
-            COMMAND echo "format header on the modified code files except build/examples/third_party/ginkgo.hpp"
+            COMMAND echo "format header on the modified code files except build/examples/third_party/accessor/dev_tools/ginkgo.hpp"
             COMMAND bash -c "git diff --name-only origin/develop...HEAD | \
-                grep -Ev 'build|examples|third_party|ginkgo.hpp' | \
+                grep -Ev 'build|examples|third_party|accessor/|dev_tools/|ginkgo.hpp' | \
```
MarcelKoch, line 343: nit: are the `/` necessary?
```suggestion
                grep -Ev 'build|examples|third_party|accessor|dev_tools|ginkgo.hpp' | \
```

CMakeLists.txt
```diff
@@ -338,19 +338,19 @@ if(NOT "${BASH}" STREQUAL "BASH-NOTFOUND" AND GINKGO_DEVEL_TOOLS)
...
             WORKING_DIRECTORY ${Ginkgo_SOURCE_DIR}
             VERBATIM)
     endif()
     unset(GIT CACHE)
     add_custom_target(format_header_all
-        COMMAND echo "format header on all code files except build/examples/third_party/ginkgo.hpp"
+        COMMAND echo "format header on all code files except build/examples/third_party/accessor/dev_tools/ginkgo.hpp"
         COMMAND bash -c "find * -type f | \
-                grep -Ev 'build|examples|third_party|ginkgo.hpp' | \
+                grep -Ev 'build|examples|third_party|accessor/|dev_tools/|ginkgo.hpp' | \
```
MarcelKoch, line 353: same as above
```suggestion
                grep -Ev 'build|examples|third_party|accessor|dev_tools|ginkgo.hpp' | \
```

MarcelKoch (approved):

CMakeLists.txt
upsj: I think we have headers whose name contains `accessor`, so the former is definitely useful

yhmtsai: Yes, some name contains accessor, so I only remove the accessor folder
dev_tools can be without `/`

PR #1288, opened 2023-02-22
Extend iteration complete event

MarcelKoch:
This PR extends the `iteration_complete` to also capture the stopping status at the end of the iteration. More specifically, it add the two parameters `const gko::array<stopping_status>* status, bool all_stopped` to the function. The other two overloads are marked as deprecated and just call the new overload. All of our loggers support the new overload.

The big difference to our current state is that this PR allows attaching the `Convergence` logger directly to the solver. IMO, this is much more user-friendly. I changed the examples to reflect this.

Evaluate:
- [x] rename event to `iteration_completed` to fit in the naming scheme
- [x] capture the right-hand-side in the event. This would allow reconstructing the residual if necessary (e.g. multigrid)
- [x] perhaps only the `gko::array<stopping_status>* status` is enough. I'm not sure if the `all_stopped` gives additional information

### DIFF
--- a/benchmark/utils/loggers.hpp
+++ b/benchmark/utils/loggers.hpp
@@ -172,21 +172,13 @@ template <typename ValueType>
 struct ResidualLogger : gko::log::Logger {
     using rc_vtype = gko::remove_complex<ValueType>;
 
-    // TODO2.0: Remove when deprecating simple overload
-    void on_iteration_complete(const gko::LinOp* solver,
-                               const gko::size_type& it,
+    void on_iteration_complete(const gko::LinOp*, const gko::size_type&,
                                const gko::LinOp* residual,
                                const gko::LinOp* solution,
-                               const gko::LinOp* residual_norm) const override
-    {
-        on_iteration_complete(solver, it, residual, solution, residual_norm,
-                              nullptr);
-    }
-
-    void on_iteration_complete(
-        const gko::LinOp*, const gko::size_type&, const gko::LinOp* residual,
-        const gko::LinOp* solution, const gko::LinOp* residual_norm,
-        const gko::LinOp* implicit_sq_residual_norm) const override
+                               const gko::LinOp* residual_norm,
+                               const gko::LinOp* implicit_sq_residual_norm,
+                               const gko::array<gko::stopping_status>* status,
+                               bool all_stopped) const override
     {
         timestamps.PushBack(std::chrono::duration<double>(
                                 std::chrono::steady_clock::now() - start)
@@ -264,7 +256,9 @@ struct IterationLogger : gko::log::Logger {
     void on_iteration_complete(const gko::LinOp*,
                                const gko::size_type& num_iterations,
                                const gko::LinOp*, const gko::LinOp*,
-                               const gko::LinOp*) const override
+                               const gko::LinOp*, const gko::LinOp*,
+                               const gko::array<gko::stopping_status>*,
+                               bool) const override
     {
         this->num_iters = num_iterations;
     }
--- a/benchmark/utils/loggers.hpp
+++ b/benchmark/utils/loggers.hpp
@@ -172,9 +172,11 @@ template <typename ValueType>
 struct ResidualLogger : gko::log::Logger {
     using rc_vtype = gko::remove_complex<ValueType>;
 
-    void on_iteration_complete(const gko::LinOp*, const gko::size_type&,
-                               const gko::LinOp* residual,
+    void on_iteration_complete(const gko::LinOp*,
+                               const gko::LinOp* right_hand_side,
                                const gko::LinOp* solution,
+                               const gko::size_type&,
+                               const gko::LinOp* residual,
                                const gko::LinOp* residual_norm,
                                const gko::LinOp* implicit_sq_residual_norm,
                                const gko::array<gko::stopping_status>* status,
@@ -253,10 +255,11 @@ private:
 
 // Logs the number of iteration executed
 struct IterationLogger : gko::log::Logger {
-    void on_iteration_complete(const gko::LinOp*,
+    void on_iteration_complete(const gko::LinOp*, const gko::LinOp*,
+                               const gko::LinOp*,
                                const gko::size_type& num_iterations,
                                const gko::LinOp*, const gko::LinOp*,
-                               const gko::LinOp*, const gko::LinOp*,
+                               const gko::LinOp*,
                                const gko::array<gko::stopping_status>*,
                                bool) const override
     {
--- a/core/log/convergence.cpp
+++ b/core/log/convergence.cpp
@@ -114,10 +114,12 @@ void Convergence<ValueType>::on_iteration_complete(
                                            dim<2>{1, residual->get_size()[1]});
                     dense_r->compute_norm2(this->residual_norm_);
                 });
-        } else if (dynamic_cast<const solver::SolverBase*>(solver) &&
+        } else if (dynamic_cast<const solver::detail::SolverBaseLinOp*>(
+                       solver) &&
                    b != nullptr && x != nullptr) {
-            auto system_mtx = dynamic_cast<const solver::SolverBase*>(solver)
-                                  ->get_system_matrix();
+            auto system_mtx =
+                dynamic_cast<const solver::detail::SolverBaseLinOp*>(solver)
+                    ->get_system_matrix();
             using Vector = matrix::Dense<ValueType>;
             using NormVector = matrix::Dense<remove_complex<ValueType>>;
             detail::vector_dispatch<ValueType>(b, [&](const auto* dense_b) {
--- a/core/log/convergence.cpp
+++ b/core/log/convergence.cpp
@@ -36,6 +36,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/base/array.hpp>
 #include <ginkgo/core/base/math.hpp>
 #include <ginkgo/core/distributed/vector.hpp>
+#include <ginkgo/core/solver/solver_base.hpp>
 #include <ginkgo/core/stop/criterion.hpp>
 #include <ginkgo/core/stop/stopping_status.hpp>
 
@@ -85,6 +86,7 @@ void Convergence<ValueType>::on_iteration_complete(
     const bool stopped) const
 {
     if (stopped) {
+        // TODO need to throw if not solver or x
         array<stopping_status> tmp(status->get_executor()->get_master(),
                                    *status);
         this->convergence_status_ = true;
--- a/core/log/convergence.cpp
+++ b/core/log/convergence.cpp
@@ -58,9 +58,9 @@ void Convergence<ValueType>::on_criterion_check_completed(
     const array<stopping_status>* status, const bool& one_changed,
     const bool& stopped) const
 {
-    this->on_iteration_complete(nullptr, num_iterations, residual, nullptr,
-                                residual_norm, implicit_sq_resnorm, status,
-                                stopped);
+    this->on_iteration_complete(nullptr, nullptr, solution, num_iterations,
+                                residual, residual_norm, implicit_sq_resnorm,
+                                status, stopped);
 }
 
 
@@ -80,10 +80,10 @@ void Convergence<ValueType>::on_criterion_check_completed(
 
 template <typename ValueType>
 void Convergence<ValueType>::on_iteration_complete(
-    const LinOp* solver, const size_type& num_iterations, const LinOp* residual,
-    const LinOp* x, const LinOp* residual_norm,
-    const LinOp* implicit_resnorm_sq, const array<stopping_status>* status,
-    const bool stopped) const
+    const LinOp* solver, const LinOp* b, const LinOp* x,
+    const size_type& num_iterations, const LinOp* residual,
+    const LinOp* residual_norm, const LinOp* implicit_resnorm_sq,
+    const array<stopping_status>* status, const bool stopped) const
 {
     if (stopped) {
         array<stopping_status> tmp(status->get_executor()->get_master(),
@@ -114,6 +114,24 @@ void Convergence<ValueType>::on_iteration_complete(
                                            dim<2>{1, residual->get_size()[1]});
                     dense_r->compute_norm2(this->residual_norm_);
                 });
+        } else if (dynamic_cast<const solver::SolverBase*>(solver) &&
+                   b != nullptr && x != nullptr) {
+            auto system_mtx = dynamic_cast<const solver::SolverBase*>(solver)
+                                  ->get_system_matrix();
+            using Vector = matrix::Dense<ValueType>;
+            using NormVector = matrix::Dense<remove_complex<ValueType>>;
+            detail::vector_dispatch<ValueType>(b, [&](const auto* dense_b) {
+                detail::vector_dispatch<ValueType>(x, [&](const auto* dense_x) {
+                    auto exec = system_mtx->get_executor();
+                    auto residual = dense_b->clone();
+                    this->residual_norm_ = NormVector::create(
+                        exec, dim<2>{1, residual->get_size()[1]});
+                    system_mtx->apply(initialize<Vector>({-1.0}, exec), dense_x,
+                                      initialize<Vector>({1.0}, exec),
+                                      residual);
+                    residual->compute_norm2(this->residual_norm_);
+                });
+            });
         }
     }
 }
--- a/core/log/convergence.cpp
+++ b/core/log/convergence.cpp
@@ -86,7 +86,6 @@ void Convergence<ValueType>::on_iteration_complete(
     const bool stopped) const
 {
     if (stopped) {
-        // TODO need to throw if not solver or x
         array<stopping_status> tmp(status->get_executor()->get_master(),
                                    *status);
         this->convergence_status_ = true;
--- a/core/log/convergence.cpp
+++ b/core/log/convergence.cpp
@@ -57,37 +57,9 @@ void Convergence<ValueType>::on_criterion_check_completed(
     const array<stopping_status>* status, const bool& one_changed,
     const bool& stopped) const
 {
-    if (stopped) {
-        array<stopping_status> tmp(status->get_executor()->get_master(),
-                                   *status);
-        this->convergence_status_ = true;
-        for (int i = 0; i < status->get_num_elems(); i++) {
-            if (!tmp.get_data()[i].has_converged()) {
-                this->convergence_status_ = false;
-                break;
-            }
-        }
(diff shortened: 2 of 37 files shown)


upsj:
If we're already touching the iteration events, would it be much more effort to add a true iteration_started/completed pair like we discussed in the context of profiler hooks? Also, the residual can theoretically be queried based on the workspace

MarcelKoch:
You're right, that fits the scope of this PR. So I will work on that.

upsj:
@MarcelKoch yes, that is a valid point - we could also consider adding "markers", which are at least supported by NVTX and maybe VTune, to annotate the iterations without an associated range. In that case, I would rename it `iteration` instead of `iteration_complete/...` in the long term

MarcelKoch:
I mean in principle you could also get a profiler range just from the iteration_complete, but of course without the first iteration.

upsj:
yes, that would require additional logic for dealing with iterative solvers specifically, so I'm not 100% happy about it. We could maybe detect whether the solver has a stopping criterion, and based on that begin and end the first/last iteration range with `apply`

MarcelKoch:
I don't think this event would/should be called from anything else than an iterative solver, or maybe I misunderstood you there. But of course you're right, the additional logic might not be worth it in the end.

upsj:
It's simpler than I thought, see #1290 :)

yhmtsai (commented):
IR will not compute the last residual in the last max iteration.

core/log/convergence.cpp
```diff
@@ -56,8 +57,36 @@ void Convergence<ValueType>::on_criterion_check_completed(
...
+
+template <typename ValueType>
+void Convergence<ValueType>::on_iteration_complete(
+    const LinOp* solver, const size_type& num_iterations, const LinOp* residual,
+    const LinOp* x, const LinOp* residual_norm,
+    const LinOp* implicit_resnorm_sq, const array<stopping_status>* status,
+    const bool stopped) const
 {
     if (stopped) {
+        // TODO need to throw if not solver or x
```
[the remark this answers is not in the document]
MarcelKoch: The solution vector. But I need to remove the comment anyway, because both are null if called from the criterion event.

core/log/stream.cpp
```diff
@@ -442,30 +442,24 @@ void Stream<ValueType>::on_criterion_check_completed(
...
+    bool stopped) const
 {
     *os_ << prefix_ << "iteration " << num_iterations
          << " completed with solver " << demangle_name(solver)
          << " with residual " << demangle_name(residual) << ", solution "
          << demangle_name(solution) << ", residual_norm "
          << demangle_name(residual_norm) << " and implicit_sq_residual_norm "
-         << demangle_name(implicit_sq_residual_norm) << std::endl;
+         << demangle_name(implicit_resnorm_sq);
+    if (status) {
```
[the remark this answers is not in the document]
MarcelKoch: No, `stopped` is only valid if `status` is not null.

examples/mixed-multigrid-solver/mixed-multigrid-solver.cpp
```diff
@@ -216,7 +215,8 @@ int main(int argc, char* argv[])
     auto toc = std::chrono::steady_clock::now();
     time += std::chrono::duration_cast<std::chrono::nanoseconds>(toc - tic);
 
-    // Calculate residual
+    // Calculate residual explicitly, because the residual is not
+    // available inside of the multigrid solver
```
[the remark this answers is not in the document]
MarcelKoch: But that is not available for the iteration complete event (normally). I may try if I can get it from the stopping criterion.

examples/distributed-solver/distributed-solver.cpp
```diff
@@ -250,35 +247,25 @@ int main(int argc, char* argv[])
...
-    A_host->apply(minus_one, x_host, one, b_host);
-    auto res_norm = gko::initialize<vec>({0.0}, exec->get_master());
-    b_host->compute_norm2(res_norm);
-
     // Take timings.
     comm.synchronize();
     ValueType t_end = gko::experimental::mpi::get_walltime();
 
+    // Get the residual.
+    auto res_norm = gko::as<vec>(logger->get_residual_norm());
```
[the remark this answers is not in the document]
MarcelKoch: Yes, but we use it only with `write`, so it doesn't matter. But thinking of it, maybe the convergence logger should store the result on the host (or copy it there on access). Do you have an opinion on that?

examples/iterative-refinement/iterative-refinement.cpp
```diff
@@ -146,10 +143,8 @@ int main(int argc, char* argv[])
     auto toc = std::chrono::steady_clock::now();
     time += std::chrono::duration_cast<std::chrono::nanoseconds>(toc - tic);
 
-    // Calculate residual
-    auto res = gko::initialize<real_vec>({0.0}, exec);
-    A->apply(one, x, neg_one, b);
-    b->compute_norm2(res);
+    // Get residual
+    auto res = gko::as<real_vec>(logger->get_residual_norm());
```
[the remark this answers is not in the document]
MarcelKoch: Yes, I think that should also have been an issue previously. Maybe we can live with that. If we stop due to the iteration limit, then we probably only wanted to do a fixed number of IR iterations anyway, or we didn't converge. In both cases, storing the last or second to last residual might not make a big difference.

benchmark/utils/loggers.hpp
```diff
@@ -172,21 +172,13 @@ template <typename ValueType>
 struct ResidualLogger : gko::log::Logger {
     using rc_vtype = gko::remove_complex<ValueType>;
 
-    // TODO2.0: Remove when deprecating simple overload
-    void on_iteration_complete(const gko::LinOp* solver,
-                               const gko::size_type& it,
+    void on_iteration_complete(const gko::LinOp*, const gko::size_type&,
```
pratikvn, line 175: Isnt this technically interface breaking ?

MarcelKoch: I don't think we care about interface changes in our benchmarks.

pratikvn: :+1: I missed that this was in the benchmarks.

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -378,7 +378,6 @@ class EnablePreconditionable : public Preconditionable {
  * @ingroup solver
  * @ingroup LinOp
  */
-template <typename MatrixType = LinOp>
```
yhmtsai, line 381: it's a public interface break  and we use it for LowerTrs at least

include/ginkgo/core/log/stream.hpp
```diff
@@ -156,6 +156,14 @@ class Stream : public Logger {
         const bool& one_changed, const bool& all_converged) const override;
 
     /* Internal solver events */
+    void on_iteration_complete(const LinOp* solver, const LinOp* b,
+                               const LinOp* x, const size_type& num_iterations,
+                               const LinOp* residual,
+                               const LinOp* residual_norm,
+                               const LinOp* implicit_resnorm_sq,
+                               const array<stopping_status>* status,
+                               bool stopped) const override;
```
yhmtsai, line 165: Is the default implementation not enough for the following old interface?

yhmtsai (commented):
should the convergence check be applied to the solver?
AFAIR, it's mentioned several times before but we somehow decline it because it does not fit the logger property well.
maybe @tcojean

MarcelKoch:
> should the convergence check be applied to the solver? 

Could you expand on what you mean here? Do you mean the convergence logger?

include/ginkgo/core/solver/solver_base.hpp
MarcelKoch: It breaks interface, yes. But IMO that interface was pretty useless. And I've fixed the use case of LowerTrs. So, for us the change is fine, since the `EnableSolverBase` has the concrete LinOp type.

yhmtsai:
@MarcelKoch I sent it to you in private chat. 
After reading the discussion, it seems to decline 
1) record the num_iteration as solver member directly -> the issue from calling solver by multithreads and mutable variable
2) propagate the logger to the stopping criterion.
this pr does not meet two opinions.
Whether should the logger record the data which is not in the object?
For example, the solver might only need residual, not the residual norm. 
Whether should the convergence logger still record that? The residual norm is required by the stopping criterion, not the solver.

core/log/convergence.cpp
```diff
@@ -86,25 +114,29 @@ void Convergence<ValueType>::on_criterion_check_completed(
                                            dim<2>{1, residual->get_size()[1]});
                     dense_r->compute_norm2(this->residual_norm_);
                 });
+        } else if (dynamic_cast<const solver::SolverBase*>(solver) &&
+                   b != nullptr && x != nullptr) {
```
yhmtsai, line 118: I think we will also need the IterationLogger for getting iteration only.
Convergence will compute the residual norm (with additional residual computation in some cases)

yhmtsai (approved):
LGTM. if it does not face the concerned issue.

core/log/convergence.cpp
MarcelKoch: I don't think this is relevant to this PR. The convergence logger should compute the residual norm, that is the point of it. I added this case, because sometimes the residual is not available (e.g. MG or first check in IR). Now the logger should be able to always compute the residual.

yhmtsai: IterationLogger should be in another pr, I mean.
If the user just needs the iteration count, the convergence requires more operation than what the user needs.

include/ginkgo/core/solver/solver_base.hpp
yhmtsai: Does it introduce any bug? If not, we should keep the interface

MarcelKoch: With the current interface it is not possible to get the stored system matrix from a solver passed in as a `LinOp`. You would need to try casting the LinOp to `SolverBase<T>` for different types. And for a custom solver with a custom user type, it would be completely impossible to get the operator.

examples/distributed-solver/distributed-solver.cpp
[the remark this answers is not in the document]
yhmtsai: it should keep the residual norm in the corresponding executor if it still uses the gko::Dense to keep. Otherwise, it is hard to notice those dense vectors are in different executor

include/ginkgo/core/solver/solver_base.hpp
yhmtsai: without breaking interface, you can make `SolverBase<MatrixType>` inherit from `SolverLinOpBase`
`SolverLinOpBase` provides the following interface (maybe you need `_linop` suffix)

examples/distributed-solver/distributed-solver.cpp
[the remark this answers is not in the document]
MarcelKoch: Which different executor?

include/ginkgo/core/solver/solver_base.hpp
MarcelKoch: I can do that, but that is also not ideal. For that, I would deprecate `SolverBase<...>`, because it serves no purpose. The base logic (e.g. the workspace) is handled in `SolverBaseLinOp`, while the concrete matrix type is known in `EnableSolverBase`. But we still need to derive from `SolverBase<...>` in `EnableSolverBase` to keep the interface stability you propose, which will result in a lot of deprecation warnings.
IMO, breaking the interface here is the cleaner approach. I doubt that part of our interface has been used anyway, because of the issue I mentioned, and users should not derive directly from `SolverBase`.

examples/distributed-solver/distributed-solver.cpp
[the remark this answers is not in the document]
MarcelKoch: No, they should use the same executor as their solver counterparts. That is also how it worked before. I didn't change anything related to that.

include/ginkgo/core/solver/solver_base.hpp
yhmtsai: you do not need to deprecate `SolverBase<>` because it's still in use.
It's a public interface without any detail/internal namespace, so I can not say it is not used.
(unless we change the definition)
If it is the bug, we must fix it with breaking interface sadly. Other than the bug, we should keep the public interface as much as possible although it introduces much pain.
From the current changelog, we only change the behavior, not the interface (multigrid, executor) and the experimental feature (distributed)

examples/distributed-solver/distributed-solver.cpp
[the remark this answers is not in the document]
yhmtsai: The above comments are my opinion on the question whether to store residual in the host or not.
They are not the comments to the code

include/ginkgo/core/solver/solver_base.hpp
MarcelKoch: IMO `SolverBase<...>` needs to be deprecated, because it should be removed. We could also remove it without deprecation, but I find that can cause even more issues on the user side.
I know that it might be used, but I argued that I find that extremely unlikely.

MarcelKoch: I've now kept the old interface and deprecated it. To reduce clutter, I disabled the corresponding warning for the header.

include/ginkgo/core/log/logger.hpp
```diff
@@ -422,6 +423,28 @@ public:                                                              \
...
+    /**
+     * Register the `iteration_complete` event which logs every completed
+     * iterations.
+     *
+     * @param solver  the solver executing the iteration
+     * @param b  the right-hand-side vector
+     * @param x  the solution vector
+     * @param it  the current iteration count
+     * @param r  the residual (optional)
+     * @param tau  the residual norm (optional)
```
tcojean, line 435: nit: maybe like above (criterion)
```suggestion
     * @param tau  the implicit residual norm squared (optional)
```

examples/mixed-multigrid-solver/mixed-multigrid-solver.cpp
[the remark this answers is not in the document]
tcojean: Shouldn't this work now? Multigrid has `SolverBase` and can at least fall back to the bode with `b` and `x`?

core/test/log/record.cpp
```diff
@@ -573,25 +573,39 @@ TEST(Record, CatchesIterations)
...
     GKO_ASSERT_MTX_NEAR(gko::as<Dense>(data->solution.get()), solution, 0);
     GKO_ASSERT_MTX_NEAR(gko::as<Dense>(data->residual_norm.get()),
                         residual_norm, 0);
     GKO_ASSERT_MTX_NEAR(gko::as<Dense>(data->implicit_sq_residual_norm.get()),
                         implicit_sq_residual_norm, 0);
+    ASSERT_EQ(data->status.get_const_data()->has_stopped(), true);
+    ASSERT_EQ(data->status.get_const_data()->get_id(),
+              stop_status.get_const_data()->get_id());
+    ASSERT_EQ(data->status.get_const_data()->is_finalized(), true);
+    ASSERT_TRUE(data->all_stopped);
```
tcojean, line 608: I don't think we should check the stop status here, it's not touched by the iteration complete logger, which is the point of this test?

core/test/log/convergence.cpp
```diff
@@ -113,12 +123,13 @@ TYPED_TEST(Convergence, CanLogData)
 TYPED_TEST(Convergence, DoesNotLogIfNotStopped)
 {
     auto logger = gko::log::Convergence<TypeParam>::create(
-        gko::log::Logger::criterion_events_mask);
+        gko::log::Logger::criterion_events_mask |
+        gko::log::Logger::iteration_complete_mask);
```
tcojean, line 127: We should maybe make these the default of `gko::log::Convergence::create()` actually.

tcojean (approved):
LGTM

core/test/log/record.cpp
MarcelKoch: As I understood it, the record logger just stores all input parameters of a logged event, so it should also store the `all_stopped` parameter.

core/test/log/convergence.cpp
MarcelKoch: I changed the constructor of `log::Convergence` as you suggested

core/test/log/record.cpp
tcojean: My bad, I had forgotten that change during the review.

benchmark/utils/loggers.hpp
```diff
@@ -172,21 +172,15 @@ template <typename ValueType>
 struct ResidualLogger : gko::log::Logger {
     using rc_vtype = gko::remove_complex<ValueType>;
 
-    // TODO2.0: Remove when deprecating simple overload
-    void on_iteration_complete(const gko::LinOp* solver,
-                               const gko::size_type& it,
-                               const gko::LinOp* residual,
+    void on_iteration_complete(const gko::LinOp*,
```
upsj, line 175: Does this compile without warnings? GCC warns about partially overridden functions, which is why I added the other override and the TODO

core/log/profiler_hook.cpp
```diff
@@ -275,24 +275,14 @@ void ProfilerHook::on_criterion_check_completed(
...
-    const LinOp* solution, const LinOp* residual_norm,
-    const LinOp* implicit_sq_residual_norm) const
+    const LinOp* solver, const LinOp* right_hand_side, const LinOp* solution,
+    const size_type& num_iterations, const LinOp* residual,
+    const LinOp* residual_norm, const LinOp* implicit_sq_residual_norm,
+    const array<stopping_status>* status, bool stopped) const
 {
     if (num_iterations > 0 &&
-        dynamic_cast<const solver::IterativeBase*>(solver)) {
+        dynamic_cast<const solver::IterativeBase*>(solver) && !stopped) {
```
upsj, line 285: this has to do with the order reversal between the logger events in the solvers?

upsj (commented):
still need to go over some of the implementation changes, but a first look

benchmark/utils/loggers.hpp
MarcelKoch: No issues at least with gcc11 and our default settings for debug. Do we have an option to increase the warning level?

core/log/profiler_hook.cpp
MarcelKoch: I think so. Otherwise we would get the last iteration twice.

benchmark/utils/loggers.hpp
MarcelKoch: Still nothing in that file, but so many other warnings. Perhaps we should add more warnings to our default debug build.

MarcelKoch:
format-rebase!

MarcelKoch:
format!

Issue #135, opened 2018-09-20
Eigensolver support in Ginkgo


gflegar:
Introduction
----------------

This is an issue to discuss how eigensolvers could be added to Ginkgo.

In general, eigensolvers compute an [eigendecomposition](https://en.wikipedia.org/wiki/Eigendecomposition_of_a_matrix) **A &Lambda; = Q &Lambda;** (or equivalently **A = Q &Lambda; Q<sup>-1</sup>**) of  a matrix **A**, where **Q** is a square matrix whose columns are the eigenvectors, and **&Lambda;** is a diagonal matrix containing the eigenvalues of the matrix.

### The basics

Here is a reminder of some basic linear algebra concerning the eigendecomposition, which might have to be considered in the discussion (_please add anything that is missing and should be considered_):
*   not all matrices have an eigendecomposition - the closes thing you can get is [Jordan normal form](https://en.wikipedia.org/wiki/Jordan_normal_form), though AFAIK this is never used in numerics
*   even if a real matrix has an eigendecomposition, the corresponding **&Lambda;** (and **Q**) do not have to be real
*   if **A** is normal (**A<sup>\*</sup>A = AA<sup>\*</sup>**), the decomposition exists and **Q** is unitary.
*   if **A** is Hermitian (**A<sup>\*</sup> = A**), **&Lambda;** is real (and since Hermitian => normal, **Q** is unitary)
*   if **A** is real symmetric (**A<sup>T</sup> = A**), **&Lambda;** is real, and **Q** is orthogonal.

Numerical methods
--------------------------

The first step for deciding on the interface is to see what the methods actually do; what inputs they need, what outputs they provide, and how they compute it.
If anyone knows a good online reference of various methods, please link it here, I don't know of one in English. I'll try to quickly go over the basic methods (using my _very_ limited understanding of eigenvalue algorithms) to give a starting point.

While writing this section I found that there is [Templates for the Solution of Algebraic Eigenvalue Problems](http://www.netlib.org/utk/people/JackDongarra/etemplates/book.html). Not sure how good it is, probably the same style as the templates for linear system. It's in HTML format, so not very pleasant to read, and I did not find a (non-shady) PDF available.

### Rayleigh quotient

The first observation that can be made is that it's all about the eigenvectors.
[...]

PR #1289, opened 2023-02-22
Add chebyshev Iteration

yhmtsai:
It adds the Chebshev iteration in https://en.wikipedia.org/wiki/Chebyshev_iteration
The second-order richardson uses a similar formula but the scalars are constant in all iteration. Chebyshev Iteration update the scalar from the previous one (the scalar are the same if upper/lower eigval does not change)


MarcelKoch:
Do you intend adding an eigenvalue estimation? I think that would be very helpful, because most times users don't have that available. I think PETSc is also doing that.
I briefly searched for that and here are some references that might be interesting for that:
- https://doi.org/10.1137/0907057
- https://doi.org/10.1137/0613050
- https://doi.org/10.1007/BF01397475

yhmtsai:
@MarcelKoch thanks for providing these reference.

I do not think I will put them in this pull request.

From https://doi.org/10.1137/0907057, they introduce the algorithm
**adaptive chebyshev iteration**
Perform m-step GMRES and use the information from the Hessenberg matrix to get the estimation of eigenvalue (I may be wrong),
and then perform https://link.springer.com/article/10.1007/BF01389971 to get the parameter for chebchev
Repeat these two steps until converge
This will be more standalone solver not smoother for multigrid because it will require several steps on GMRES, which is too expensive IMO.

From https://petsc.org/release/docs/manualpages/KSP/KSPCHEBYSHEV/#kspchebyshev,
More precisely, https://petsc.org/release/docs/manualpages/KSP/KSPChebyshevEstEigSet/
They use Lanczo or GMRES to get the maximum estimation to decide the parameter by following
min-eig-boundary = a * min-eigest + b * max-eigest  = 0.1 max-eigest (default), and
max-eig-boundary = c * min-eigest + d * max-eigest = 1.1 max-eigest (default)
(because min_eigest is usually inaccurate)
It only did once in generation, so this is more usable for Multigrid

For GMRES, we need to get the Hessenberg out and compute eigenvalue on it.
GMRES and Lanczo are related to #135

reference/test/solver/chebyshev_kernels.cpp
```diff
@@ -0,0 +1,419 @@
...
+TYPED_TEST(Chebyshev, ApplyWithGivenInitialGuessModeIsEquivalentToRef)
+{
+    using Mtx = typename TestFixture::Mtx;
+    using value_type = typename TestFixture::value_type;
+    using initial_guess_mode = gko::solver::initial_guess_mode;
+    auto ref_solver =
+        gko::solver::Chebyshev<value_type>::build()
+            .with_criteria(
+                gko::stop::Iteration::build().with_max_iters(1u).on(this->exec))
+            .with_foci(value_type{0.9}, value_type{1.1})
```
MarcelKoch, line 387: Perhaps extract this into a partial factory, so that you can just add the `.wiht_default_initial_guess(guess)` below?

core/solver/chebyshev.cpp
```diff
@@ -0,0 +1,390 @@
...
+    }
+    experimental::precision_dispatch_real_complex_distributed<ValueType>(
+        [this, guess](auto dense_b, auto dense_x) {
+            prepare_initial_guess(dense_b, dense_x, guess);
+            this->apply_dense_impl(dense_b, dense_x, guess);
+        },
+        b, x);
+}
+
+
```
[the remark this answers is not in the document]
MarcelKoch: I guess the reference for this is 'Templates for the Solution of Linear Systems: Building Blocks for Iterative Methods, 2nd Edition'

include/ginkgo/core/solver/chebyshev.hpp
```diff
@@ -0,0 +1,295 @@
...
+ * residual. It has another term for the difference of solution. Moreover, this
+ * method requires knowledge about the spectrum of the matrix.
+ *
+ * ```
+ * solution = initial_guess
+ * while not converged:
+ *     residual = b - A solution
+ *     error = solver(A, residual)
+ *     solution = solution + alpha_i * error + beta_i * (solution_i -
+ * solution_{i-1})
```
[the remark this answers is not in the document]
MarcelKoch: After looking a bit more into it, my proposal doesn't make too much sense. Perhaps instead it would be possible to extract the common residual computation.

core/test/solver/chebyshev.cpp
```diff
@@ -0,0 +1,375 @@
...
+
+
+TYPED_TEST(Chebyshev, CanSetEigenRegion)
+{
+    using Solver = typename TestFixture::Solver;
+    using value_type = typename TestFixture::value_type;
+    std::shared_ptr<Solver> chebyshev_solver =
+        Solver::build()
+            .with_criteria(
+                gko::stop::Iteration::build().with_max_iters(3u).on(this->exec))
```
MarcelKoch, line 181:
```suggestion
```
not necessary for this test

core/test/solver/chebyshev.cpp
```diff
@@ -0,0 +1,375 @@
...
+{
+    using Solver = typename TestFixture::Solver;
+    using value_type = typename TestFixture::value_type;
+    auto chebyshev_factory =
+        Solver::build()
+            .with_criteria(
+                gko::stop::Iteration::build().with_max_iters(3u).on(this->exec),
+                gko::stop::ResidualNorm<value_type>::build()
+                    .with_reduction_factor(r<value_type>::value)
+                    .on(this->exec))
```
MarcelKoch, line 201:
```suggestion
            .with_criteria(
                gko::stop::Iteration::build().with_max_iters(3u).on(this->exec)
```
one criteria is enough here

core/test/solver/chebyshev.cpp
```diff
@@ -0,0 +1,375 @@
...
+            .on(this->exec);
+    auto solver = chebyshev_factory->generate(this->mtx);
+    auto inner_solver = solver->get_solver();
+
+    ASSERT_NE(inner_solver.get(), nullptr);
+    ASSERT_EQ(inner_solver.get(), chebyshev_solver.get());
+}
+
+
+TYPED_TEST(Chebyshev, CanSetCriteriaAgain)
```
MarcelKoch, line 243:
```suggestion
TYPED_TEST(Chebyshev, CanSetCriteria)
```

MarcelKoch (changes requested):
Besides the comments left below, I want to mention the `num_keep` and `num_generated` mechanic. That needs some explaination, because right now I don't understand what that is for. It seems like some sort of restart mechanic, but I might be wrong there. Also, exposing this to the user seems confusing to me.

include/ginkgo/core/solver/chebyshev.hpp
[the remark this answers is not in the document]
yhmtsai: IR can add the moment parameter to have fixed beta * (sol_i - sol_{i-1}).
chebyshev alpha and beta are changed during iteration, so it is not quite suitable to put it in IR.
adding foci to IR with some flag may be able to switch the version.

include/ginkgo/core/solver/chebyshev.hpp
```diff
@@ -0,0 +1,295 @@
...
+         * Default initial guess mode. The available options are under
+         * initial_guess_mode.
+         */
+        initial_guess_mode GKO_FACTORY_PARAMETER_SCALAR(
+            default_initial_guess, initial_guess_mode::provided);
+
+        /**
+         * The number of scalar to keep
+         */
+        int GKO_FACTORY_PARAMETER_SCALAR(num_keep, 2);
```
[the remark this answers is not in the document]
yhmtsai: no, `num_keep` is to keep the generated scalar in the storage.
alpha and beta are changed by iteration, but they are fixed by the given bound.
It's used for fixed iteration runs because we do not need to refill these scalars for different apply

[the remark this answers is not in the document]
MarcelKoch: But shouldn't we then just keep all scalars?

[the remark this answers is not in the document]
yhmtsai: when using it as normal solver, we may not have the maximum iteration information.
allocating one big dense matrix and then moving them to another one when full may not be a good approach.
we can also add the ability of workspace such that it can handle the std::vector<dense> then it should be more flexible for uncertain size.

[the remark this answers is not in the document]
MarcelKoch: I think we can and should assume that this solver (which is mostly going to be used as a preconditioner anyway) has an iteration criterion. Then we can always get the max number of iteration, or throw if the iteration criterion is not there.
TBH I'm opposed to this factory parameter, because it's too much of an implementation detailed revealed to the user.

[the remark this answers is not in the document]
MarcelKoch: That would also remove a bit of the complexity in the implementation.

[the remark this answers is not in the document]
yhmtsai: I can not assume that there's an iteration criterion. It can contain the residual norm or time criterion.
If we assume that, we should provide the standalone iteration parameter.
Users do not need to know the implementation details. 
The statement is to keep how many scalars for chebyshev to avoid the refill overhead.
I can introduce this usage later when we have the use case.
I think it should help the situation when kernel launch overhead is noticeable.

include/ginkgo/core/solver/chebyshev.hpp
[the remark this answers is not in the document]
MarcelKoch: what I mean is that both have the following structure:
```
while not converged:
  residual = b - A * solution
  error = M^-1 * residual
  solution = update(residual, solution, ...)
```
And the only difference is `update(...)`. So perhaps both could use a unified implementation in that way.

include/ginkgo/core/solver/chebyshev.hpp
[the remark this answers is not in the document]
MarcelKoch: You are right, this is very useful to negate the overhead. Still, I think we can assume that there is an iteration criterion somewhere in the combined one and use that. If not, we would just throw.

vasilisge0 (commented):
I am fine with merging this PR when corrections in parts of the documentation that are pointed out (comments) take place. I could understand what num_keep variable was used for but perhaps a more descriptive name would be better.

include/ginkgo/core/solver/chebyshev.hpp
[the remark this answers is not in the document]
MarcelKoch: You could set the default value to `unspecified`, which you define before. When the solver is generated, you can check if the parameter is `unspecified` and then try to extract an iteration criterion from the criteria. If that doesn't work, you throw with a message to either pass an iteration criterion or set this parameter to something else.
Or alternatively, just replace the criteria parameter with a `iterations` parameter and move the class into the preconditioner namespace. I think that is also a fine option, since that would be the main use case anyway.

include/ginkgo/core/solver/chebyshev.hpp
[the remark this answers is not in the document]
yhmtsai: Yes, it's true and applies almost all iterative methods, I think.

include/ginkgo/core/solver/chebyshev.hpp
```diff
@@ -0,0 +1,295 @@
...
+                                       initial_guess_mode guess) const override;
+
+    void set_relaxation_factor(
+        std::shared_ptr<const matrix::Dense<ValueType>> new_factor);
+
+    explicit Chebyshev(std::shared_ptr<const Executor> exec)
+        : EnableLinOp<Chebyshev>(std::move(exec))
+    {}
+
+    explicit Chebyshev(const Factory* factory,
```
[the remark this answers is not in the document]
yhmtsai: I moved it. We need to set a guideline for what should be in cpp.
The constructor is quite simple, so put definition with declaration such that I do not need to look for the definition.

include/ginkgo/core/solver/chebyshev.hpp
```diff
@@ -0,0 +1,295 @@
...
+        /**
+         * Already generated solver. If one is provided, the factory `solver`
+         * will be ignored.
+         */
+        std::shared_ptr<const LinOp> GKO_FACTORY_PARAMETER_SCALAR(
+            generated_solver, nullptr);
+
+        /**
+         * The pair of foci of ellipse. It is usually be {lower bound of eigval,
+         * upper bound of eigval} for real matrices.
```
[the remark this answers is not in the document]
yhmtsai: I was not sure whether the inner solver has the same meaning/effect as preconditioner in the beginning.
From Templates for the Solution of Linear Systems: Building Blocks for Iterative Methods, 2nd Edition, they also mention is is a preconditioner, so calling it preconditioner should be safe.
Yes, it is bounded for preconditioned system from the textbook.
But I do not quite understanding using CG and GMRES to estimate the eigenvalue now.
Do they use Cg and GMRES with only few iterations such that the M^{-1}A is not I?

core/solver/chebyshev.cpp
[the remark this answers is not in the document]
yhmtsai: thanks for providing the reference.

include/ginkgo/core/solver/chebyshev.hpp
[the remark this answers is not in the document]
MarcelKoch: I've also not looked into the eigenvalue estimation, but I think that can be handled in another PR.

include/ginkgo/core/solver/chebyshev.hpp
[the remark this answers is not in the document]
MarcelKoch: For me everything should be in the .cpp except for getters, or other functions which should be inlined.

include/ginkgo/core/solver/chebyshev.hpp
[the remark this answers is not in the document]
MarcelKoch: You're right. But what makes the Chebychev more similar to IR than the other solvers is the `if(iter == 0)` block. This is exactly the same, so perhaps this part could be extracted into a residual update function.

include/ginkgo/core/solver/chebyshev.hpp
```diff
@@ -0,0 +1,295 @@
...
+
+
+namespace gko {
+namespace solver {
+
+
+/**
+ * Chebyshev iteration is an iterative method that uses another coarse
+ * method to approximate the error of the current solution via the current
+ * residual. It has another term for the difference of solution. Moreover, this
```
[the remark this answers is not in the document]
yhmtsai: the solution x_i is also based on the $x_{i-1} - x_{i-2}$

include/ginkgo/core/solver/chebyshev.hpp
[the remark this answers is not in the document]
yhmtsai: Yes, I need to think how to extract them. It will contain many input parameters, is it sutiable?

include/ginkgo/core/solver/chebyshev.hpp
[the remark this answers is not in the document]
MarcelKoch: TBH I don't see the relevance of that. Has that any effect for the user?

include/ginkgo/core/solver/chebyshev.hpp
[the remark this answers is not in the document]
MarcelKoch: Can some of these parameters be passed via the workspace?

include/ginkgo/core/solver/chebyshev.hpp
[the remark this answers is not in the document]
MarcelKoch: Thanks, I think it should be fine the way it is now.

include/ginkgo/core/solver/chebyshev.hpp
```diff
@@ -0,0 +1,295 @@
...
+ *
+ * ```
+ * solution = initial_guess
+ * while not converged:
+ *     residual = b - A solution
+ *     error = solver(A, residual)
+ *     solution = solution + alpha_i * error + beta_i * (solution_i -
+ * solution_{i-1})
+ * ```
+ *
```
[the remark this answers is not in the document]
yhmtsai: I only verify alpha in eq. 9 in https://encyclopediaofmath.org/index.php?title=Chebyshev_iteration_method
note the beta in algorithm is different from the beta in the webpage.
The following might not use quite precise word.
alpha beta are chosen to form the chebyshev polynomial to lead minimal error.
alpha beta is not directly the chebyshev form ( $T_{n+1}(x) = 2 x T_n(x) - T_{n-1}(x)$ )

include/ginkgo/core/solver/chebyshev.hpp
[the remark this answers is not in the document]
yhmtsai: no, I only try to explain the algorithm's difference from IR. IR using $x += \alpha M^{-1}r$, but chebyshev uses $x += \alpha_i M^{-1} r + \beta_i (x_{i-1} - x_{i-2})$ $(x_{i-1} - x_{i-2})$ is the additional term against IR

include/ginkgo/core/solver/chebyshev.hpp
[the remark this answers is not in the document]
MarcelKoch: `residual` is also part of the workspace, so my ideal would be using it like:
```
residual_update(solver, iter, dense_x, dense_b, criterion, log)
```
And everything else we can get from the workspace.

TBH, do we actually still need the `residual_ptr`? I think it's only relevant for the `iter == 0` branch, and otherwise its `residual_ptr = residual`. So we could just pass-in the initial guess mode.

core/test/solver/chebyshev.cpp
yhmtsai: it is not neccessary, but the solver does not allow empty criterion which leads infinite loops.

reference/test/solver/chebyshev_kernels.cpp
yhmtsai: No, I can only extract the first two settings which might not be too helpful because I would like to ensure the setting does not affect each other.

core/test/solver/chebyshev.cpp
yhmtsai: I prefer staying the `CanSetCriteriaAgain` because it set criteria first from factory and then reset it by member function

include/ginkgo/core/solver/chebyshev.hpp
```diff
@@ -0,0 +1,295 @@
...
+ * @tparam ValueType  precision of matrix elements
+ *
+ * @ingroup solvers
+ * @ingroup LinOp
+ */
+template <typename ValueType = default_precision>
+class Chebyshev : public EnableLinOp<Chebyshev<ValueType>>,
+                  public EnableSolverBase<Chebyshev<ValueType>>,
+                  public EnableIterativeBase<Chebyshev<ValueType>>,
```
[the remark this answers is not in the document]
MarcelKoch: TBH, I don't think it's necessary for IR to be handled differently. I think we can just use `with_preconditioner` uniformly, there is no functional difference between the inner solver and a preconditioner.

include/ginkgo/core/solver/chebyshev.hpp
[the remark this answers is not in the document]
yhmtsai: passing initial guess mode also takes one argument as passing `residual_ptr`, right?
I have thought about getting something from the workspace.
`create_...` is protected, so we need to use get_workspace_op with casting (need to cast const for residual) or lambda
It requires the vectors' names to be the same in IR/Chebyshev. It's true now, but we do not have any restrictions for that.
Thus, I think passing everything is simple

include/ginkgo/core/solver/chebyshev.hpp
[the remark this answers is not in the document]
yhmtsai: I agree there's no functional difference. I think many call it as inner solver of IR not preconditioner of IR. Do you see any reference calls preconditioned-IR?

core/test/solver/chebyshev.cpp
```diff
@@ -0,0 +1,373 @@
...
+    this->assert_same_matrices(static_cast<const Mtx*>(copy_mtx.get()),
+                               this->mtx.get());
+}
+
+
+TYPED_TEST(Chebyshev, CanBeCloned)
+{
+    using Mtx = typename TestFixture::Mtx;
+    using Solver = typename TestFixture::Solver;
+    auto clone = this->solver->clone();
```
MarcelKoch, line 147:
```suggestion

    auto clone = this->solver->clone();
```

core/test/solver/chebyshev.cpp
```diff
@@ -0,0 +1,373 @@
...
+    auto clone_mtx = static_cast<Solver*>(clone.get())->get_system_matrix();
+    this->assert_same_matrices(static_cast<const Mtx*>(clone_mtx.get()),
+                               this->mtx.get());
+}
+
+
+TYPED_TEST(Chebyshev, CanBeCleared)
+{
+    using Solver = typename TestFixture::Solver;
+    this->solver->clear();
```
MarcelKoch, line 159: same

core/test/solver/chebyshev.cpp
```diff
@@ -0,0 +1,373 @@
...
+{
+    ASSERT_TRUE(this->solver->apply_uses_initial_guess());
+}
+
+
+TYPED_TEST(Chebyshev, CanSetEigenRegion)
+{
+    using Solver = typename TestFixture::Solver;
+    using value_type = typename TestFixture::value_type;
+    std::shared_ptr<Solver> chebyshev_solver =
```
MarcelKoch, line 178: same

core/test/solver/chebyshev.cpp
```diff
@@ -0,0 +1,373 @@
...
+            .with_foci(value_type{0.2}, value_type{1.2})
+            .on(this->exec)
+            ->generate(this->mtx);
+
+    ASSERT_EQ(chebyshev_solver->get_parameters().foci,
+              std::make_pair(value_type{0.2}, value_type{1.2}));
+}
+
+
+TYPED_TEST(Chebyshev, CanSetInnerSolverInFactory)
```
MarcelKoch, line 191: also not AAA

I think a lot of the tests here don't have correct AAA spacing

core/test/solver/chebyshev.cpp
```diff
@@ -0,0 +1,373 @@
...
+
+
+TYPED_TEST(Chebyshev, CanSetEigenRegion)
+{
+    using Solver = typename TestFixture::Solver;
+    using value_type = typename TestFixture::value_type;
+    std::shared_ptr<Solver> chebyshev_solver =
+        Solver::build()
+            .with_criteria(
+                gko::stop::Iteration::build().with_max_iters(3u).on(this->exec))
```
MarcelKoch, line 181: all the `.on` can be removed after rebase

core/test/solver/chebyshev.cpp
```diff
@@ -0,0 +1,373 @@
...
+    auto chebyshev_factory =
+        Solver::build().with_criteria(init_crit).on(this->exec);
+
+    ASSERT_EQ((chebyshev_factory->get_parameters().criteria).back(), init_crit);
+
+    auto solver = chebyshev_factory->generate(this->mtx);
+    std::shared_ptr<gko::stop::CriterionFactory> new_crit =
+        gko::stop::Iteration::build().with_max_iters(5u).on(this->exec);
+
+    solver->set_stop_criterion_factory(new_crit);
```
MarcelKoch, line 254: I think we already had a discussion somewhere that setting the criterion doesn't make sense. Perhaps we can start here to remove that functionality. So I would suggest to override `set_stop_criterion_factory` in the chebyshev solver and make that throw/deprecated.

reference/test/solver/chebyshev_kernels.cpp
```diff
@@ -0,0 +1,558 @@
...
+    auto x = gko::initialize<Mtx>({0.0, 0.0, 0.0}, this->exec);
+
+    solver->apply(b.get(), x.get());
+
+    auto alpha = gko::as<gko::matrix::Dense<value_type>>(
+        solver->get_workspace_op(gko::solver::workspace_traits<Solver>::alpha));
+    auto beta = gko::as<gko::matrix::Dense<value_type>>(
+        solver->get_workspace_op(gko::solver::workspace_traits<Solver>::beta));
+    // the iteration is more than default
+    ASSERT_EQ(alpha->get_size(), (gko::dim<2>{1, 7}));
```
MarcelKoch, line 173: I guess the values `alpha(i)` `i>2` are not checked because it will get too complicated?

reference/test/solver/chebyshev_kernels.cpp
```diff
@@ -0,0 +1,558 @@
...
+    auto beta = gko::as<gko::matrix::Dense<value_type>>(
+        solver->get_workspace_op(gko::solver::workspace_traits<Solver>::beta));
+    // if the iteration limit less than the default value, it will use the
+    // default value.
+    ASSERT_EQ(alpha->get_size(), (gko::dim<2>{1, 4}));
+    ASSERT_EQ(beta->get_size(), (gko::dim<2>{1, 4}));
+}
+
+
+TYPED_TEST(Chebyshev, CheckStoredAlphaBeta)
```
MarcelKoch, line 149: Maybe a check to see if alpha, beta stay the same after running the solver twice can be added?

MarcelKoch (changes requested):
LGTM, but I would like to resolve some of my earlier issues. Other than that only minor nits.

core/test/solver/chebyshev.cpp
yhmtsai: No, I think it's AAA because alias is not any AAA.
We only put them next to the real code in ginkgo testings.
That why I only put them together.

core/solver/chebyshev.cpp
```diff
@@ -0,0 +1,399 @@
...
+                num_max_generation_ = std::max(
+                    num_max_generation_, iter_stop->get_parameters().max_iters);
+            }
+        }
+    } else if (auto iter_stop = std::dynamic_pointer_cast<
+                   const gko::stop::Iteration::Factory>(
+                   this->get_stop_criterion_factory())) {
+        num_max_generation_ = std::max(num_max_generation_,
+                                       iter_stop->get_parameters().max_iters);
+    }
```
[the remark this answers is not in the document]
yhmtsai: Thanks for good abstraction! It can solve all case even with more nested stopping criterion

reference/test/solver/chebyshev_kernels.cpp
yhmtsai: I put it into AlphaBetaFromChangingCriterion, which also check the value with less iteration limit

include/ginkgo/core/solver/chebyshev.hpp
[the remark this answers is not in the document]
yhmtsai: I try to rephrase it again. Could you take a look?

reference/test/solver/chebyshev_kernels.cpp
yhmtsai: Yes. Checking the following values requires me to write the same recursive function as core, so I think it's not a real test?
another way is to generate these value first and compare the value directly

include/ginkgo/core/solver/chebyshev.hpp
```diff
@@ -0,0 +1,245 @@
...
+
+    struct parameters_type
+        : enable_preconditioned_iterative_solver_factory_parameters<
+              parameters_type, Factory> {
+        /**
+         * The pair of foci of ellipse, which covers the eigenvalues of
+         * preconditioned system. It is usually be {lower bound of eigval, upper
+         * bound of eigval} of preconditioned real matrices.
+         */
+        std::pair<value_type, value_type> GKO_FACTORY_PARAMETER_VECTOR(
```
MarcelKoch, line 151: Is there a reason to use the old-style parameters, instead of the deferred ones?

include/ginkgo/core/solver/chebyshev.hpp
```diff
@@ -0,0 +1,245 @@
...
+ * ```
+ *
+ *
+ * @tparam ValueType  precision of matrix elements
+ *
+ * @ingroup solvers
+ * @ingroup LinOp
+ */
+template <typename ValueType = default_precision>
+class Chebyshev
```
MarcelKoch, line 81: nit: I would like to push making our classes final. I think I put down my reasons for this elsewhere.
```suggestion
class Chebyshev final
```

core/test/solver/chebyshev.cpp
MarcelKoch: You're right. I guess we first have to put in some effort to deprecate it and show users better alternatives.

include/ginkgo/core/solver/chebyshev.hpp
yhmtsai: I think deferred factory, which handles the Factory with/without on, only for those Factory which requires `.on(exec)`. For the other value or pure LinOp parameter, we still use the original style.

reference/test/solver/chebyshev_kernels.cpp
MarcelKoch: I think it is fine as it is. Generating the values might be a bit tricky, since those would have to be regenerated if the tests or implementation change.

MarcelKoch (approved):

include/ginkgo/core/solver/chebyshev.hpp
```diff
@@ -0,0 +1,236 @@
...
+
+/**
+ * Chebyshev iteration is an iterative method that can solving nonsymeetric
+ * problems. Chebyshev Iterations avoids the inner products for computation
+ * which may be the bottleneck for distributed system. Chebyshev Iteration is
+ * developed based on the Chebyshev polynomials of the first kind. Moreover,
+ * this method requires knowledge about the spectrum of the preconditioned
+ * matrix. This implementation follows the algorithm in "Templates for the
+ * Solution of Linear Systems: Building Blocks for Iterative Methods, 2nd
+ * Edition".
```
upsj, line 36: I would rearrange this slightly (the fact that we need to know the spectrum is important, so we should mention it early), and try to be consistent with capitalization of iteration.
```suggestion
 * Chebyshev iteration is an iterative method for solving nonsymmeetric
 * problems based on some knowledge of the spectrum of the (preconditioned) system matrix. It avoids the computation of inner products,
 * which may be a performance bottleneck for distributed system. Chebyshev iteration is
 * developed based on Chebyshev polynomials of the first kind. 
 * This implementation follows the algorithm in "Templates for the
 * Solution of Linear Systems: Building Blocks for Iterative Methods, 2nd
 * Edition".
```

include/ginkgo/core/solver/chebyshev.hpp
```diff
@@ -0,0 +1,236 @@
...
+                       std::shared_ptr<const LinOp> system_matrix);
+
+private:
+    std::shared_ptr<const LinOp> solver_{};
+    // num_generated_scalar_ tracks the number of generated scalar alpha
+    // and beta.
+    mutable size_type num_generated_scalar_ = 0;
+    // num_max_generation_ is the number of generated scalar kept in the
+    // workspace.
+    mutable size_type num_max_generation_ = 3;
```
upsj, line 190: I'm a bit wary of mutable members beyond well-controlled cases like the workspace. They might make future work on thread-safety more complicated. Not that I have a good alternative for it, but I want to bring it up.

upsj (changes requested):
There are a few questions around data recycling and mutability I would like to discuss before merging this

include/ginkgo/core/solver/chebyshev.hpp
yhmtsai: Isn't wokrspace handled in similar way?

core/solver/chebyshev.cpp
```diff
@@ -0,0 +1,390 @@
...
+    this->setup_workspace();
+
+    GKO_SOLVER_VECTOR(residual, dense_b);
+    GKO_SOLVER_VECTOR(inner_solution, dense_b);
+    GKO_SOLVER_VECTOR(update_solution, dense_b);
+
+    auto old_num_max_generation = num_max_generation_;
+    // Use the scalar first
+    // get the iteration information from stopping criterion.
+    visit_criteria(
```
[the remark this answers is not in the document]
yhmtsai: It was. but @MarcelKoch suggested that getting it from stopping criterion to reduce the parameter, I think.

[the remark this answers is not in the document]
MarcelKoch: If we put it as a separate parameter, then we should not allow for a stopping criteria parameter. This would be fine with me, but I think it causes other issues, as this can't be considered a solver anymore.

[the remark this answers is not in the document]
upsj: The smoother doesn't have much puprpose with norm-based stopping criteria, because that would negate the benefit of avoiding reductions, right? And can it run correctly when it uses more iterations than the number of stored scalar values? If yes, then we could have two separate parameters iterations and subspace_dimension (or similar), if no, we could have just a single parameter iterations, and internally create an Iteration stopping criterion

yhmtsai:
@upsj To avoid the mutable and reusing value from workspace, I think moving them into generatio such that prefill all alpha and beta before apply. I think it will also give better performance because we do not need to do some tiny fill before the limitation during apply.
Also, refill it when the foci or criterion change (or specfic factor for that).

upsj:
If those are just static coefficients independent of the matrix, that sounds very reasonable. I tend to agree with what I believe @MarcelKoch said, which is that we shouldn't consider changeable stopping criteria here. We also currently don't seem to have a way to propagate information from other parts of the solver hierarchy to compute appropriate focus values.

include/ginkgo/core/solver/chebyshev.hpp
upsj: We don't have any mutability in our LinOps right now beyond vector caches and workspaces, most of the `mutable` uses are in loggers etc. So the mutability is constrained into a single class. That's not the case here.

In general, if I write code, I expect a `const` object to not change internally, and the only reasons we don't follow this requirement fully right now seems to be the need for cache data structures to avoid reallocation, and the fact that loggers are `const` and can't store data themselves.

yhmtsai:
It should cover the eigenvalye range from the matrix itself. Without estimation from our side, the value is given by user. They should be changed together with matrices. Something like L1 on SPD matrix can give the value based on the property without knowing the detail of matrices. I will come back for this after the release because it is not the urgent case.

reference/test/solver/chebyshev_kernels.cpp
```diff
@@ -0,0 +1,514 @@
...
+    auto b = gko::initialize<Mtx>({3.9, 9.0, 2.2}, this->exec);
+    auto x = gko::initialize<Mtx>({0.5, 1.0, 2.0}, this->exec);
+
+    solver->apply(alpha.get(), b.get(), beta.get(), x.get());
+
+    GKO_ASSERT_MTX_NEAR(x, l({1.5, 5.0, 2.0}), r<value_type>::value * 1e1);
+}
+
+
+TYPED_TEST(Chebyshev, SolvesTriangularSystemUsingAdvancedApplyMixed)
```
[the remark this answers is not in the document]
yhmtsai: good catch! I also fixed the ir corresponding part

core/solver/chebyshev.cpp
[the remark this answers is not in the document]
yhmtsai: yes it can run correctly even more iterations than the preallocated storage

core/solver/ir.cpp
```diff
@@ -225,52 +224,19 @@ void Ir<ValueType>::apply_dense_impl(const VectorType* dense_b,
...
-                                   .solution(dense_x)
-                                   .check(relative_stopping_id, true,
-                                          &stop_status, &one_changed);
+        auto log_func = [this](auto solver, auto dense_b, auto dense_x,
+                               auto iter, auto residual_ptr,
+                               array<stopping_status>& stop_status,
+                               bool all_stopped) {
             this->template log<log::Logger::iteration_complete>(
-                this, dense_b, dense_x, iter, residual_ptr, nullptr, nullptr,
+                solver, dense_b, dense_x, iter, residual_ptr, nullptr, nullptr,
```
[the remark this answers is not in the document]
yhmtsai: I think I still need `this` for `this->template log...`

yhmtsai:
@upsj I have changed it to accept the host scalar now, so we do not need to store them in advanced.
I think we can take a look how much difference between passing by device pointer or host value in the future.
do you think it will change a lot in this case?

MarcelKoch (approved):

include/ginkgo/core/solver/chebyshev.hpp
```diff
@@ -0,0 +1,236 @@
...
+#include <ginkgo/core/matrix/dense.hpp>
+#include <ginkgo/core/matrix/identity.hpp>
+#include <ginkgo/core/solver/solver_base.hpp>
+#include <ginkgo/core/stop/combined.hpp>
+#include <ginkgo/core/stop/criterion.hpp>
+#include <ginkgo/core/stop/iteration.hpp>
+
+
+namespace gko {
+namespace solver {
```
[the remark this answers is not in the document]
upsj: this still needs to be discussed: Is this the right place to put it?

core/solver/ir.cpp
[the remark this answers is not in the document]
upsj: right, sorry, missed that.

common/unified/CMakeLists.txt
```diff
@@ -25,6 +25,8 @@ set(UNIFIED_SOURCES
     solver/bicgstab_kernels.cpp
     solver/cg_kernels.cpp
     solver/cgs_kernels.cpp
+    solver/cgs_kernels.cpp
```
upsj, line 28: duplicate!

common/unified/solver/chebyshev_kernels.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include "common/unified/base/kernel_launch.hpp"
+
+
+namespace gko {
+namespace kernels {
+namespace GKO_DEVICE_NAMESPACE {
+namespace chebyshev {
+
+
+template <typename ValueType, typename ScalarType>
```
upsj, line 18: Does this actually need a second precision, or can we always use the highest precision available? This is memory-bound after all.

upsj (dismissed):
I mainly want to discuss the need for mixed precision instantiation here

common/unified/CMakeLists.txt
yhmtsai: good catch!

common/unified/solver/chebyshev_kernels.cpp
yhmtsai: I have the same thought and thinking maybe we just use double or complex<double> directly?

upsj: If there is a numerical benefit to using a higher-precision multiplication every time, I would go for it

yhmtsai: after rethink about it, I do not have a good way to use double or complex double without any information from the vector or matrix. Maybe still keep the same way as how we handle IR/richardson now. also I need the ScalarType, which is the ValueType from Chebyshev class

yhmtsai: currently keep the value type of class but change fosi, alpha, beta type to double/complex<double>
when dpcpp only support float not double, it will cast to float/complex<float> internally to keep the same interface among all backends

include/ginkgo/core/solver/chebyshev.hpp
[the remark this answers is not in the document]
pratikvn: In literature, I think it is still called a solver (https://netlib.org/linalg/html_templates/node44.html), and the users would also expect it to be within the solver namespace. IMO adding another namespace would be more confusing.

yhmtsai:
@MarcelKoch @upsj it is updated to passing the alpha/beta from the host to kernel directly.

common/unified/solver/chebyshev_kernels.cpp
```diff
@@ -0,0 +1,114 @@
...
+    run_kernel(
+        exec,
+        [] GKO_KERNEL(auto row, auto col, auto alpha, auto inner_sol,
+                      auto update_sol, auto output) {
+            const auto inner_val = static_cast<type>(inner_sol(row, col));
+            update_sol(row, col) =
+                static_cast<device_type<ValueType>>(inner_val);
+            output(row, col) = static_cast<device_type<ValueType>>(
+                static_cast<type>(output(row, col)) +
+                static_cast<type>(alpha) * inner_val);
```
MarcelKoch, line 72: I think this has a couple unnecessary casts. The `run_kernel` should apply `as_device_type` to all arguments, so casting `static_cast<type>(alpha)` should be unnecessary.
```suggestion
            update_sol(row, col) = inner_sol(row, col);
            output(row, col) = static_cast<device_type<ValueType>>(
                static_cast<type>(output(row, col)) +
                alpha * static_cast<type>(inner_sol(row, col)));
```

common/unified/solver/chebyshev_kernels.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+    auto alpha_val = static_cast<actual_coeff_type>(alpha);
+    auto beta_val = static_cast<actual_coeff_type>(beta);
+
+    run_kernel(
+        exec,
+        [] GKO_KERNEL(auto row, auto col, auto alpha, auto beta, auto inner_sol,
+                      auto update_sol, auto output) {
+            const auto val = static_cast<type>(inner_sol(row, col)) +
+                             static_cast<type>(beta) *
```
MarcelKoch, line 97: same here.

core/solver/chebyshev.cpp
```diff
@@ -0,0 +1,341 @@
...
+#include "core/solver/update_residual.hpp"
+
+
+namespace gko {
+namespace solver {
+namespace chebyshev {
+namespace {
+
+
+GKO_REGISTER_OPERATION(initialize, ir::initialize);
```
MarcelKoch, line 26: nit: maybe move this into an `ir` namespace.

core/solver/chebyshev.cpp
```diff
@@ -0,0 +1,341 @@
...
+Chebyshev<ValueType>::Chebyshev(const Factory* factory,
+                                std::shared_ptr<const LinOp> system_matrix)
+    : EnableLinOp<Chebyshev>(factory->get_executor(),
+                             gko::transpose(system_matrix->get_size())),
+      EnablePreconditionedIterativeSolver<ValueType, Chebyshev<ValueType>>{
+          std::move(system_matrix), factory->get_parameters()},
+      parameters_{factory->get_parameters()}
+{
+    this->set_default_initial_guess(parameters_.default_initial_guess);
+    center_ = (std::get<0>(parameters_.foci) + std::get<1>(parameters_.foci)) /
```
MarcelKoch, line 67: maybe we should have some checks on the `foci` parameter, i.e. `foci[0,1] != 0, foci[0] <= foci[1]`.

core/test/solver/chebyshev.cpp
```diff
@@ -0,0 +1,329 @@
...
+                  .on(exec)),
+          solver(chebyshev_factory->generate(mtx))
+    {}
+
+    std::shared_ptr<const gko::Executor> exec;
+    std::shared_ptr<Mtx> mtx;
+    std::shared_ptr<typename Solver::Factory> chebyshev_factory;
+    std::unique_ptr<gko::LinOp> solver;
+
+    static void assert_same_matrices(const Mtx* m1, const Mtx* m2)
```
MarcelKoch, line 48: Can't this be replaced with our test macro?

reference/solver/chebyshev_kernels.cpp
```diff
@@ -0,0 +1,69 @@
...
+                 matrix::Dense<ValueType>* output)
+{
+    using type = coeff_type<ValueType>;
+    for (size_t row = 0; row < output->get_size()[0]; row++) {
+        for (size_t col = 0; col < output->get_size()[1]; col++) {
+            const auto inner_val = static_cast<type>(inner_sol->at(row, col));
+            update_sol->at(row, col) = static_cast<ValueType>(inner_val);
+            output->at(row, col) =
+                static_cast<ValueType>(static_cast<type>(output->at(row, col)) +
+                                       static_cast<type>(alpha) * inner_val);
```
MarcelKoch, line 33:
```suggestion
                                       alpha * inner_val);
```

reference/solver/chebyshev_kernels.cpp
```diff
@@ -0,0 +1,69 @@
...
+            const coeff_type<ValueType> alpha, const coeff_type<ValueType> beta,
+            matrix::Dense<ValueType>* inner_sol,
+            matrix::Dense<ValueType>* update_sol,
+            matrix::Dense<ValueType>* output)
+{
+    using type = coeff_type<ValueType>;
+    for (size_t row = 0; row < output->get_size()[0]; row++) {
+        for (size_t col = 0; col < output->get_size()[1]; col++) {
+            const auto val = static_cast<type>(inner_sol->at(row, col)) +
+                             static_cast<type>(beta) *
```
MarcelKoch, line 52:
```suggestion
                             beta *
```

reference/solver/chebyshev_kernels.cpp
```diff
@@ -0,0 +1,69 @@
...
+                                 static_cast<type>(update_sol->at(row, col));
+            inner_sol->at(row, col) = static_cast<ValueType>(val);
+            update_sol->at(row, col) = static_cast<ValueType>(val);
+            output->at(row, col) =
+                static_cast<ValueType>(static_cast<type>(output->at(row, col)) +
+                                       static_cast<type>(alpha) * val);
```
MarcelKoch, line 58:
```suggestion
                                       alpha * val);
```

reference/test/solver/chebyshev_kernels.cpp
```diff
@@ -0,0 +1,472 @@
...
+
+TYPED_TEST_SUITE(Chebyshev, gko::test::ValueTypes, TypenameNameGenerator);
+
+
+TYPED_TEST(Chebyshev, KernelInitUpdate)
+{
+    using value_type = typename TestFixture::value_type;
+    using Mtx = typename TestFixture::Mtx;
+    value_type alpha(0.5);
+    auto inner_sol = gko::initialize<Mtx>(
```
MarcelKoch, line 54: those vectors seem to be repeated quite often, maybe pull them into the test class.

test/solver/solver.cpp
```diff
@@ -176,6 +177,21 @@ struct Ir : SimpleSolverTest<gko::solver::Ir<solver_value_type>> {
...
+    static double tolerance() { return 1e7 * r<value_type>::value; }
+
+    static typename solver_type::parameters_type build_preconditioned(
+        std::shared_ptr<const gko::Executor> exec,
+        gko::size_type iteration_count, bool check_residual = true)
+    {
+        return SimpleSolverTest<gko::solver::Chebyshev<solver_value_type>>::
+            build(exec, iteration_count, check_residual)
+                .with_preconditioner(
+                    precond_type::build().with_max_block_size(1u).on(exec));
```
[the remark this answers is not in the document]
MarcelKoch: I think this has not been resolved.

test/solver/solver.cpp
```diff
@@ -177,6 +178,21 @@ struct Ir : SimpleSolverTest<gko::solver::Ir<solver_value_type>> {
...
 
+struct Chebyshev : SimpleSolverTest<gko::solver::Chebyshev<solver_value_type>> {
+    static double tolerance() { return 1e7 * r<value_type>::value; }
+
+    static typename solver_type::parameters_type build_preconditioned(
+        std::shared_ptr<const gko::Executor> exec,
+        gko::size_type iteration_count, bool check_residual = true)
+    {
+        return SimpleSolverTest<gko::solver::Chebyshev<solver_value_type>>::
+            build(exec, iteration_count, check_residual)
```
MarcelKoch, line 189: what about the foci values here?

core/solver/chebyshev.cpp
```diff
@@ -0,0 +1,341 @@
...
+
+template <typename ValueType>
+std::unique_ptr<LinOp> Chebyshev<ValueType>::conj_transpose() const
+{
+    return build()
+        .with_generated_preconditioner(share(
+            as<Transposable>(this->get_preconditioner())->conj_transpose()))
+        .with_criteria(this->get_stop_criterion_factory())
+        .with_foci(conj(std::get<0>(parameters_.foci)),
+                   conj(std::get<1>(parameters_.foci)))
```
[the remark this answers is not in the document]
yhmtsai: yes, it is from the conjugate not from the transpose.
the eigenvalue of conjugated matrix is the conjugate eigenvalue of original matrix like determinant.

core/solver/chebyshev.cpp
```diff
@@ -0,0 +1,341 @@
...
+    auto exec = this->get_executor();
+    this->setup_workspace();
+
+    GKO_SOLVER_VECTOR(residual, dense_b);
+    GKO_SOLVER_VECTOR(inner_solution, dense_b);
+    GKO_SOLVER_VECTOR(update_solution, dense_b);
+
+    GKO_SOLVER_ONE_MINUS_ONE();
+
+    auto alpha_ref = coeff_type{1} / center_;
```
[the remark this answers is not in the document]
MarcelKoch: IMO `alpha_host` would be better suited in that case.

common/unified/solver/chebyshev_kernels.cpp
MarcelKoch: I don't understand your point, there is no mixed precision here, there is only one value type.

test/solver/solver.cpp
yhmtsai: it use the default one as 0, 1. It might not converge. We only check whether it gives the same result as the reference like the others from random matrices.

core/solver/chebyshev.cpp
MarcelKoch: > In L1 scalar Jacobi on SPD matrix, we use 0, 1 here, which provided by the math property

Can you explain this a bit more?

test/solver/solver.cpp
MarcelKoch: That's fine, I just missed the default values for the foci.

core/solver/chebyshev_kernels.hpp
```diff
@@ -0,0 +1,64 @@
...
+#include "core/base/kernel_declaration.hpp"
+
+
+namespace gko {
+namespace kernels {
+namespace chebyshev {
+
+
+template <typename T>
+using coeff_type =
```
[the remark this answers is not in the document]
yhmtsai: it was to reduce the header dependence, but I guess I do not need to worry about it now.

common/unified/solver/chebyshev_kernels.cpp
yhmtsai: you are right. I thought I do not use the truth of double/complex<double> is also the arithmetic type.
type is more on the arithmetic type so the other needs to casting to the highest precision.
delete the casting for the type but I will change the type to arithmetic_type not device_coeff_type for the purpose

common/unified/solver/chebyshev_kernels.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+template <typename ValueType>
+void init_update(std::shared_ptr<const DefaultExecutor> exec,
+                 const coeff_type<ValueType> alpha,
+                 const matrix::Dense<ValueType>* inner_sol,
+                 matrix::Dense<ValueType>* update_sol,
+                 matrix::Dense<ValueType>* output)
+{
+    using actual_coeff_type = if_single_only_type<coeff_type<ValueType>>;
+    using type = device_type<actual_coeff_type>;
```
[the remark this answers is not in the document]
yhmtsai: call it arithmetic_type currently

core/solver/chebyshev.cpp
MarcelKoch: Ok, I see, you meant the range of the eigenvalues under preconditioning. But I don't understand what this has to do with checking the input values for the foci parameter.

include/ginkgo/core/solver/chebyshev.hpp
```diff
@@ -0,0 +1,234 @@
...
+
+    class Factory;
+
+    struct parameters_type
+        : enable_preconditioned_iterative_solver_factory_parameters<
+              parameters_type, Factory> {
+        /**
+         * The pair of foci of ellipse, which covers the eigenvalues of
+         * preconditioned system. It is usually be {lower bound of eigval, upper
+         * bound of eigval} of preconditioned real matrices.
```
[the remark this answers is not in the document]
yhmtsai: I wanted to emphasis the real eigenvalue such that lower/upper bound makes sense, but non-complex matrix can have complex eigenvalue.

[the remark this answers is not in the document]
yhmtsai: I have updated it a bit.

core/solver/chebyshev.cpp
yhmtsai: It only checks the real(foci[1]) >= real(foci[0]) now

MarcelKoch: I would think that's enough.

MarcelKoch (changes requested):
I think only the `!=0` check is required.

MarcelKoch (approved):

Issue #1264, opened 2023-01-26
AMG defaults


MarcelKoch:
I think we should discuss how reasonable defaults for our AMG look like. IMO, it should be possible to write
```
auto solver = Cg<>::build().with_preconditioner(Multigrid::build().on(exec))...
```
and expect that a reasonable AMG is constructed.
Since our AMG is not called AMG, but Multigrid, which implies more generality, I would also think that the following would be fine to specify how the levels are build:
```
auto solver = Cg<>::build().with_preconditioner(Multigrid::build()
                                                  .with_mg_level(Pgm<>::build().on(exec)).on(exec))...
```
But neither work right now. They result in an exception, due to missing stopping criterion, or, if that is fulfilled, in non-convergence.

Before going into specifics, we need to clarify if we want to support the syntax above or not. Any thoughts on this are welcome.

MarcelKoch:
I would like to bring up that the configuration in the example `multigrid-preconditioned-solver` also doesn't work. Of course, in the example the solver converges, but if that configuration is applied to a simple finite difference discretization of the Laplace operator, the needed solver iteration scale linearly with the mesh width. IMO, that means that the AMG is not correctly configured, and thus sets a bad example.
BTW, the issue seems to be that the coarse grid solve is not exact. If an exact solver (iterative solver with tol ~ eps) is used than is the iteration number doesn't scale linearly anymore.

yhmtsai:
for the laplace operator, should it use the geometric coasrening method?
The criterion issue is applied to all solvers. They always need to set the criterion to use.

MarcelKoch:
You are right, I'm missing the criterion. But the argument still holds. Other preconditioner/solver work with just the criterion, but the AMG doesn't.
A different coarsening method might be better suited for the laplace operator, but as I mentioned, that does not seem to be the issue. Solving exactly on the coarse grid seems to fix the convergence rate.

yhmtsai:
I do not understand what's the meaning of incorrect settings.
Do you mean we should have a better coarse solver as the default?

yhmtsai:
I think current setting without any passing presmoother or coarse solver will give you the Identity like IR's inner solver or preconditioner from other solvers

MarcelKoch:
[...]

Issue #1162, opened 2022-11-01
Improving factory parameters


upsj:
I think the current factory parameter setup could use with a few improvements:

1. The factory parameters generated by our macros are `mutable` by default! That's a big code smell IMO, though we can't change it without potentially breaking user code
2. The factory parameters get injected into the generated LinOp by default. That means that we need to carry them around and cannot discard them after generation, which may be useful in some cases, like solvers, where we no longer need the preconditioner factory after generation. Again, this can't be changed without an interface break.
3. The fluent interface `with_parameter` is templated and forwards its argument(s) into a brace-initializer. That means that a parameter of type `unsigned` can't be initialized with an argument of type `int`.
4. The current design assumes that there is a 1:1 relationship between LinOpFactories and LinOps. That is a strong limitation, as there are many cases where it makes sense to use a variety of different algorithms (= factories) to generate the same kind of object, e.g. factorizations.
5. Redundancy: We specify the same `stopping_criterion`, `preconditioner`, [`initial_guess`](https://github.com/ginkgo-project/ginkgo/pull/981), [`logger`](https://github.com/ginkgo-project/ginkgo/pull/1140) for all iterative solvers. What if we built a factory parameter type hierarchy that mirrors the underlying solver hierarchy? That way, users have a single place where to look up how to set up a solver, and we only need to maintain a single piece of documentation. Most solvers will not need any additional parameters, and the ones that do will only need a handful of parameters, like the restart in GMRES.

As a first step, I built #1082 with suggested fixes for most of these issues in mind.

PR #1291, opened 2023-02-23
Change default AMG settings

MarcelKoch:
Based on the discussion in #1264 this PR changes some of the default settings of the AMG. The overall goal it to make it possible that our default AMG (no extra settings except the coarsening and stopping criterion) helps to reduce the necessary number of iterations adequately.

I've made each change in the defaults its own commit, with a brief description of my reasoning.

TODO:
- [x] update the documentation of the example
- [x] evaluate if we should also default the coarsening and/or stopping criterion
  - no to both of them.


MarcelKoch:
There seem to be some incompatibilities between the `initial_guess_mode` and other settings of the multigrid. I'm not sure exactly which settings that touches on. But I've experienced that the mode `provided` can lead to no convergence at all. I think this needs to be investigated and then we need to warn the user if some incompatible settings are used.

MarcelKoch:
While working on this, I come to realize that the default parameters are much better realized with the factory improvements mentioned in #1162. For example using
```
using create_factory_t = std::function<std::shared_ptr<LinOpFactory>(std::shared_ptr<const Executor>)>;
std::vector<create_factory_t> pre_smoother;
```
would allow defining the defaults directly in the parameter declaration:
```
std::vector<create_factory_t> pre_smoother{[](std::shared_ptr<const Executor> exec) {
  return Ir::build().with_solver(Jacobi::build().on(exec)).on(exec);
};
```
Perhaps with some helper functionality/change in factory it could be really concise.

fritzgoebel (changes requested):

include/ginkgo/core/solver/multigrid.hpp
```diff
@@ -323,6 +324,9 @@ class Multigrid : public EnableLinOp<Multigrid>,
 
         /**
          * Coarsest factory list.
+         *
+         * If not set, then GMRES preconditioned with a scalar Jacobi will be
+         * used as solver on the coarsest level.
```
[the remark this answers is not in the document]
MarcelKoch: No, I forgot that this is actually working now.

core/solver/multigrid.cpp
```diff
@@ -554,17 +568,26 @@ void Multigrid::generate()
             using value_type =
                 typename std::decay_t<decltype(*mg_level)>::value_type;
             auto exec = this->get_executor();
+            // default coarse grid solver, direct LU
+            // TODO: maybe remove fixed index type
+            auto gen_default_solver = [&] {
+                return experimental::solver::Direct<value_type, int32>::build()
```
yhmtsai, line 574: Could we use the experimental class in another public class?

examples/multigrid-preconditioned-solver-customized/multigrid-preconditioned-solver-customized.cpp
```diff
@@ -0,0 +1,210 @@
...
+    auto ic_gen = gko::share(
+        ic::build()
+            .with_factorization_factory(
+                gko::factorization::Ic<ValueType, int>::build().on(exec))
+            .on(exec));
+    auto smoother_gen = gko::share(
+        gko::solver::build_smoother(ic_gen, 2u, static_cast<ValueType>(0.9)));
+    // Use Pgm as the MultigridLevel factory (this is also the default).
+    auto mg_level_gen =
+        gko::share(pgm::build().with_deterministic(true).on(exec));
```
yhmtsai, line 141: Is it default mg_level? I do not see when it is set

core/solver/multigrid.cpp
MarcelKoch: I don't see why not. The default initialization is internal anyway.

examples/multigrid-preconditioned-solver-customized/multigrid-preconditioned-solver-customized.cpp
MarcelKoch: You're right, I will remove that comment

examples/ir-ilu-preconditioned-solver/doc/builds-on
```diff
@@ -1 +1 @@
-ilu-preconditioned-solver iterative-refinement
+ic-preconditioned-solver iterative-refinement
```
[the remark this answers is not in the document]
MarcelKoch: don't know how that got in there...

core/solver/multigrid.cpp
MarcelKoch: The experimental features can't be turned off. Only the experimental MPI stuff is disabled if MPI is not available.

yhmtsai (approved):

MarcelKoch:
rebase!

PR #1294, opened 2023-03-09
add more method for timer output

yhmtsai:
This PR adds more statistical methods like average, median, best, and worst to compute the time result.

When the timing use x iteration which is more than 1 iteration, timer will record the average time x times.
Handle repetition_growth_factor <=1 case such that next_timing always be next iteration.
User can set it to get the same overhead operation in each repetitions

### DIFF
--- a/benchmark/blas/blas_common.hpp
+++ b/benchmark/blas/blas_common.hpp
@@ -425,7 +425,7 @@ void apply_blas(const char* operation_name, std::shared_ptr<gko::Executor> exec,
         for (auto _ : ic.run()) {
             op->run();
         }
-        const auto runtime = ic.compute_average_time();
+        const auto runtime = ic.compute_time(FLAGS_timer_output);
         const auto flops = static_cast<double>(op->get_flops());
         const auto mem = static_cast<double>(op->get_memory());
         const auto repetitions = ic.get_num_repetitions();
--- a/benchmark/blas/blas_common.hpp
+++ b/benchmark/blas/blas_common.hpp
@@ -425,7 +425,7 @@ void apply_blas(const char* operation_name, std::shared_ptr<gko::Executor> exec,
         for (auto _ : ic.run()) {
             op->run();
         }
-        const auto runtime = ic.compute_time(FLAGS_timer_output);
+        const auto runtime = ic.compute_time(FLAGS_timer_method);
         const auto flops = static_cast<double>(op->get_flops());
         const auto mem = static_cast<double>(op->get_memory());
         const auto repetitions = ic.get_num_repetitions();
--- a/benchmark/utils/general.hpp
+++ b/benchmark/utils/general.hpp
@@ -115,9 +115,10 @@ DEFINE_uint32(max_repetitions, std::numeric_limits<unsigned int>::max(),
               "If 'repetitions = auto' is used, the maximal number of"
               " repetitions for a single benchmark.");
 
-DEFINE_double(repetition_growth_factor, 1.5,
-              "If 'repetitions = auto' is used, the factor with which the"
-              " repetitions between two timings increase.");
+DEFINE_double(
+    repetition_growth_factor, 1.5,
+    "The factor with which the repetitions between two timings increase. If it "
+    "is lower than or equal to 1, the timing region is always 1 repetition.");
 
 
 /**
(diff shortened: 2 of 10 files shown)


benchmark/utils/general.hpp
```diff
@@ -581,10 +581,27 @@ class IterationControl {
...
+     * @param method  the statistical method
+     * @param same_rep  whether timer managed outside has the same repetitions
+     *                  as IterationControl. If the timer is managed outside and
+     *                  the timer does not have the same repetitions as
+     *                  InterationControl, the method is only allowed "average".
+     *
+     * @return the statistical time
+     */
+    double compute_time(const std::string& method = "average",
+                        bool same_rep = true) const
```
pratikvn, line 596: Do you mean to set this to `false`  ?

benchmark/utils/timer_impl.hpp
```diff
@@ -63,13 +63,18 @@ class Timer {
 
     /**
      * Finish the timer
+     *
+     * @param num  the number of operation for the timing range
```
yanjen, line 67: "the number of samples for the statistical timing" might be better?

benchmark/utils/general.hpp
```diff
@@ -581,10 +581,25 @@ class IterationControl {
...
+            this->get_num_repetitions()) {
+            return status_run_.managed_timer.compute_time(method);
+        } else {
+            assert(method == "average");
+            return status_run_.managed_timer.get_total_time() /
+                   this->get_num_repetitions();
+        }
     }
 
     IndexType get_num_repetitions() const { return status_run_.cur_it; }
```
yanjen, line 605: It was not from this pr but I think the word "samples" would describe better than the word "repetitions". Does anyone feel the same way?

benchmark/blas/blas_common.hpp
```diff
@@ -425,7 +425,7 @@ void apply_blas(const char* operation_name, std::shared_ptr<gko::Executor> exec,
         for (auto _ : ic.run()) {
             op->run();
         }
-        const auto runtime = ic.compute_average_time();
+        const auto runtime = ic.compute_time(FLAGS_timer_output);
```
yanjen, line 428: FLAGS_timer_type or might be better? Same for the other places.

yanjen (approved):
LGTM, but I have some suggestions on the naming.

MarcelKoch (commented):
I'm not sure if this makes sense. Currently, we are not timing each iteration individually, in order to reduce the timing overhead. Perhaps it makes sense to revisit that. But for now, without timings of the individual iterations, it doesn't make sense to compute the median, or min/max.

MarcelKoch (commented):
Currently, we take timestamps after a varying number of iterations, for example at repetition `1, 2, 4, 8, 16, 32, ...` (not exactly but similar). So with the proposed changes we would take averages of a growing number of repetitions. That might have useful information, especially considering that the kernel launch and synchronization overhead is larger for fewer repetitions. But we wouldn't provide actual medians, or min/max.
Still, I won't hold this up.

benchmark/utils/general.hpp
yhmtsai: I think the repetitions is better. Sample to me is like subset of data.

benchmark/blas/blas_common.hpp
yhmtsai: I change it to timer_method

yhmtsai:
rebase!

benchmark/utils/general.hpp
```diff
@@ -115,9 +115,10 @@ DEFINE_uint32(max_repetitions, std::numeric_limits<unsigned int>::max(),
               "If 'repetitions = auto' is used, the maximal number of"
               " repetitions for a single benchmark.");
 
-DEFINE_double(repetition_growth_factor, 1.5,
-              "If 'repetitions = auto' is used, the factor with which the"
-              " repetitions between two timings increase.");
+DEFINE_double(
+    repetition_growth_factor, 1.5,
+    "The factor with which the repetitions between two timings increase. If it "
+    "is lower than or equal to 1, the timing region always be 1 repetition.");
```
upsj, line 121:
```suggestion
    "is lower than or equal to 1, the timing region is always 1 repetition.");
```

benchmark/utils/timer_impl.hpp
```diff
@@ -87,13 +92,31 @@ class Timer {
...
-        return this->get_total_time() / this->get_num_repetitions();
+        if (method == "average") {
+            return this->get_total_time() / this->get_num_repetitions();
+        }
+        auto copy = duration_sec_;
+        std::sort(copy.begin(), copy.end());
+        if (method == "best") {
+            return copy.front();
+        } else if (method == "worst") {
+            return copy.back();
```
upsj, line 111: how about using min and max? For times, this should be unambiguous and is closer to what is used in statistical analysis

PR #1296, opened 2023-03-13
Add METIS integration

upsj:
This adds a METIS CMake find module and a simple integration. It also serves as an example for how to use LinOpFactory to represent a reordering algorithm by creating Permutation matrix from a Csr matrix, which might be a potential future interface for reorderings. ~~I guess we also need to add METIS to a handful of containers to test this.~~

Since METIS gets hiccups from diagonal entries, I added `sparsity_csr::remove_diagonal_elements`  kernels for all backends.


upsj:
format-rebase!

cmake/Modules/FindMETIS.cmake
```diff
@@ -0,0 +1,67 @@
...
+# Result variables
+# ^^^^^^^^^^^^^^^^
+#
+# This module will set the following variables in your project:
+#
+# ``METIS_INCLUDE_DIRS``
+#   where to find metis.h or metis64.h
+#
+# ``METIS_HEADER``
+#   either metis64.h or metis.h, dependent in which was found
```
upsj, line 24:
```suggestion
#   either metis64.h or metis.h, dependent on which was found
```

test/reorder/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+if (GINKGO_HAVE_METIS)
+    ginkgo_create_common_test(nested_dissection)
+endif()
```
upsj, line 3:
```suggestion
endif()

```

test/reorder/nested_dissection.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+TYPED_TEST_SUITE(NestedDissection, gko::test::IndexTypes,
+                 TypenameNameGenerator);
+
+
+TYPED_TEST(NestedDissection, ResultIsEquivalentToRef)
+{
+    auto perm = this->nd_factory->generate(this->mtx);
+    auto dperm = this->dnd_factory->generate(this->dmtx);
+}
```
upsj, line 80: assertion

include/ginkgo/core/reorder/nested_dissection.hpp
```diff
@@ -0,0 +1,146 @@
...
+#include <ginkgo/core/matrix/permutation.hpp>
+
+
+namespace gko {
+/**
+ * @brief The Reorder namespace.
+ *
+ * @ingroup reorder
+ */
+namespace reorder {
```
upsj, line 65: `experimental` unless we decide this new interface is stable enough

upsj:
format-rebase!

test/reorder/nested_dissection.cpp
```diff
@@ -0,0 +1,87 @@
...
+TYPED_TEST(NestedDissection, ResultIsEquivalentToRef)
+{
+    auto perm = this->nd_factory->generate(this->mtx);
+    auto dperm = this->dnd_factory->generate(this->dmtx);
+
+    auto perm_array = gko::make_array_view(this->ref, this->mtx->get_size()[0],
+                                           perm->get_permutation());
+    auto dperm_array = gko::make_array_view(
+        this->exec, this->mtx->get_size()[0], dperm->get_permutation());
+    GKO_ASSERT_ARRAY_EQ(perm_array, dperm_array);
```
pratikvn, line 86:
```suggestion
    auto dperm_array = gko::make_array_view(
        this->exec, this->mtx->get_size()[0], dperm->get_permutation());
        
    GKO_ASSERT_ARRAY_EQ(perm_array, dperm_array);
```

core/reorder/nested_dissection.cpp
```diff
@@ -0,0 +1,200 @@
...
+              IndexType* iperm)
+{
+    auto nvtxs = static_cast<idx_t>(num_rows);
+    vector<idx_t> tmp_row_ptrs(row_ptrs, row_ptrs + num_rows + 1, {host_exec});
+    vector<idx_t> tmp_col_idxs(col_idxs, col_idxs + row_ptrs[num_rows],
+                               {host_exec});
+    vector<idx_t> tmp_perm(num_rows, {host_exec});
+    vector<idx_t> tmp_iperm(num_rows, {host_exec});
+    auto result = METIS_NodeND(&nvtxs, tmp_row_ptrs.data(), tmp_col_idxs.data(),
+                               nullptr, const_cast<idx_t*>(options.data()),
```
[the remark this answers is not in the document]
pratikvn: Just a random idea, not sure if it makes a lot of sense. :)

MarcelKoch:
+1 for ParMETIS, but not in this PR.

test/reorder/nested_dissection.cpp
upsj: would you consider this part of Act or Assert? I was also unsure, but decided to place it with the assertion

pratikvn: I would say it is separate ? But I see that it also makes sense like you have it.

core/reorder/nested_dissection.cpp
```diff
@@ -0,0 +1,200 @@
...
+                               tmp_perm.data(), tmp_iperm.data());
+    if (result != METIS_OK) {
+        throw MetisError(__FILE__, __LINE__, "METIS_NodeND",
+                         metis_error_message(result));
+    }
```
yhmtsai, line 113: replaced by metis_nd<idx_t>?

include/ginkgo/core/reorder/nested_dissection.hpp
```diff
@@ -0,0 +1,148 @@
...
+    using permutation_type = matrix::Permutation<index_type>;
+
+    struct parameters_type
+        : public enable_parameters_type<
+              parameters_type, NestedDissection<ValueType, IndexType>> {
+        /**
+         * The options to be passed on to METIS, stored as key-value pairs.
+         * Any options that are not set here use their default value.
+         */
+        std::unordered_map<int, int> options;
```
yhmtsai, line 102: I think the unodered_map<int, int> is for METIS only. If the NestedDissection is only available for METIS, I will prefer having METIS in class

test/reorder/nested_dissection.cpp
```diff
@@ -0,0 +1,87 @@
...
+    std::unique_ptr<reorder_type> dnd_factory;
+    std::shared_ptr<matrix_type> mtx;
+    std::shared_ptr<matrix_type> dmtx;
+};
+
+TYPED_TEST_SUITE(NestedDissection, gko::test::IndexTypes,
+                 TypenameNameGenerator);
+
+
+TYPED_TEST(NestedDissection, ResultIsEquivalentToRef)
```
yhmtsai, line 77: It tests with reference results, but the reference test is missing

yhmtsai (changes requested):
the reference test and core test is missing.
Also, whether is NestedDissection only from METIS?

common/unified/matrix/sparsity_csr_kernels.cpp
```diff
@@ -0,0 +1,146 @@
...
+        exec,
+        [] GKO_KERNEL(auto row, auto row_ptrs, auto col_idxs, auto prefix_sum) {
+            const auto begin = row_ptrs[row];
+            const auto end = row_ptrs[row + 1];
+            IndexType count = 0;
+            for (auto nz = begin; nz < end; nz++) {
+                if (col_idxs[nz] == row) {
+                    count++;
+                }
+            }
```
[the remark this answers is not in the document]
upsj: yes, I just wanted to make the algorithm robust to broken inputs. METIS crashes with diagonal elements, so we should avoid that at all costs :laughing:

core/reorder/nested_dissection.cpp
```diff
@@ -0,0 +1,200 @@
...
+    case METIS_ERROR_INPUT:
+        return "METIS_ERROR_INPUT";
+    case METIS_ERROR_MEMORY:
+        return "METIS_ERROR_MEMORY";
+    case METIS_ERROR:
+        return "METIS_ERROR";
+    case METIS_OK:
+        return "METIS_OK";
+    default:
+        return "<unknown>";
```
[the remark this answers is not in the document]
upsj: great suggestion, thanks!

core/reorder/nested_dissection.cpp
```diff
@@ -0,0 +1,200 @@
...
+    METIS_SetDefaultOptions(result.data());
+    for (auto pair : options) {
+        if (pair.first < 0 || pair.first >= METIS_NOPTIONS) {
+            throw MetisError(__FILE__, __LINE__, "build_metis_options",
+                             "Invalid option ID " + std::to_string(pair.first));
+        }
+        result[pair.first] = pair.second;
+    }
+    // make sure users don't accidentally switch on 1-based indexing
+    result[METIS_OPTION_NUMBERING] = 0;
```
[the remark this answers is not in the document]
upsj: good point, I'll check

core/reorder/nested_dissection.cpp
upsj: The __LINE__ info should be sufficient to differentiate the two cases, and METIS_NodeND is likely more familiar to users. What do you think?

include/ginkgo/core/reorder/nested_dissection.hpp
upsj: At the moment it's only METIS, but in the future I would add interfaces to [KaHIP](https://kahip.github.io/) and [Scotch/PT-Scotch](https://www.labri.fr/perso/pelegrin/scotch/)

test/reorder/nested_dissection.cpp
upsj: Building a ground-truth for METIS is pretty hard, since there are many orderings that may produce equivalent fill-in. What we could do is test that the permutation reduces Cholesky fill-in in a pathological case?

include/ginkgo/core/reorder/nested_dissection.hpp
yhmtsai: do they also use the unordered_map for their parameters?

upsj: Not really, KaHIP has only a single `mode` parameter and Scotch has an opaque Strategy pointer. But we can add those when we add support for them. Do you think it should be called `metis_options` instead?

test/reorder/nested_dissection.cpp
yhmtsai: do they provide some example matrix with the result?
if they have, we can use it to ensure we indeed run through METIS

upsj: no, and graph partitioning has many degrees of freedom that produce equivalent results, which is why I used the simple star graph as an unambiguous example.

core/reorder/nested_dissection.cpp
yhmtsai: ah, my statement leads confusion,
I mean line 112-118 can call the following `metis_nd<idx_t>`.  but it is just a nit

include/ginkgo/core/reorder/nested_dissection.hpp
yhmtsai: maybe keeping it as options to leave possiblity to reuse it

yhmtsai (approved):

core/reorder/nested_dissection.cpp
upsj: ah right thanks, I'll see if this works easily.

upsj:
rebase!

PR #1302, opened 2023-03-17
Add symbolic LU and Cholesky benchmark

upsj:
Together with a generic validation tool for factorizations. 
Also removes the `strategy` parameter (we don't use strategies for `sparse_blas` algorithms), renames `validate_results` to `validate` and allows operations to output their own custom stats (in this case for the amount of fill-in).
Finally, I added prefix_sum benchmarks to BLAS, which highlighted a horrible performance on CUDA. ~~replacing them by Thrust gives us at least 100x speedup even with the overflow check.~~ See #1303

### DIFF
--- a/benchmark/sparse_blas/operations.cpp
+++ b/benchmark/sparse_blas/operations.cpp
@@ -38,6 +38,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
 #include "benchmark/sparse_blas/operations.hpp"
+#include "benchmark/utils/json.hpp"
 #include "core/factorization/elimination_forest.hpp"
 #include "core/factorization/symbolic.hpp"
 #include "core/matrix/csr_kernels.hpp"
@@ -631,6 +632,13 @@ public:
 
     void run() override { gko::factorization::symbolic_lu(mtx_, result_); }
 
+    void write_stats(rapidjson::Value& object,
+                     rapidjson::MemoryPoolAllocator<>& allocator) override
+    {
+        add_or_set_member(object, "factor_nonzeros",
+                          result_->get_num_stored_elements(), allocator);
+    }
+
 private:
     const Mtx* mtx_;
     std::unique_ptr<Mtx> result_;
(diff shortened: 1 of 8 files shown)


benchmark/sparse_blas/operations.cpp
```diff
@@ -566,6 +569,98 @@ class LookupOperation : public BenchmarkOperation {
 };
 
 
+bool validate_factorization(const Mtx* input, const Mtx* factors)
+{
+    const auto host_exec = input->get_executor()->get_master();
+    const auto host_input = gko::clone(host_exec, input);
+    const auto host_factors = gko::clone(host_exec, factors);
```
MarcelKoch, line 576: `make_temporary_clone`? But I guess the performance is not important here.

benchmark/sparse_blas/operations.cpp
```diff
@@ -566,6 +569,98 @@ class LookupOperation : public BenchmarkOperation {
 };
 
 
+bool validate_factorization(const Mtx* input, const Mtx* factors)
```
MarcelKoch, line 572: perhaps 
```suggestion
bool validate_symbolic_factorization(const Mtx* input, const Mtx* factors)
```
because I was confused for a while wondering why there is no comparison between values.

MarcelKoch (dismissed):
LGTM, only two smaller comments.

benchmark/sparse_blas/operations.cpp
upsj: good point, originally I wanted to sort the input, but the ordering is always guaranteed by the algorithms

upsj:
format-rebase!

benchmark/sparse_blas/operations.cpp
```diff
@@ -566,6 +570,114 @@ class LookupOperation : public BenchmarkOperation {
...
+
+    std::pair<bool, double> validate() const override
+    {
+        return std::make_pair(
+            validate_symbolic_factorization(mtx_, result_.get()), 0.0);
+    }
+
+    gko::size_type get_flops() const override { return 0; }
+
+    gko::size_type get_memory() const override { return 0; }
```
yhmtsai, line 660: could we get the memory?

yhmtsai (dismissed):
suggest to move the prefix sum related changes to another pr

benchmark/sparse_blas/operations.cpp
upsj: doesn't make much sense to compute a bandwidth here, since we use both CPU and GPU for different parts of the algorithm

MarcelKoch (approved):
LGTM

benchmark/blas/blas_common.hpp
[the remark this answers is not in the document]
upsj: done

upsj:
rebase!

tcojean (approved):
LGTM

PR #1304, opened 2023-03-20
Draft: MPI Logging Events

MarcelKoch:
This PR adds Logger events for MPI calls.

The new implementation differs from the other Logger events in Ginkgo. Normally, we have a nearly one-to-one relation between events and the functions registering the events. But for MPI events, I don't think that is the right approach, because there are so many MPI calls we would log. Creating events for each call (e.g. 1 for `send`, 1 for `recv`, 1 for `all_to_all`, 1 for `all_to_all_v`, etc.) seems prohibitive. Besides for many logging purposes, the exact function might not be relevant. For example, logging the total amount of data sent needs to know only the size argument for each MPI call.

To prevent creating too many events, I tried to categorize MPI calls along multiple axes. Currently, I'm using the following axes:
1. point-to-point vs collective
2. blocking vs non-blocking (can be extended to also contain persistent)
3. within collective: all-to-all vs all-to-one vs one-to-all vs scan vs barrier (this is related to the distinction in the MPI Docs §6.2.2)
4. within collective: fixed message sizes vs variable message sizes

Using these four axes, it is possible to model all of our current `mpi::communicator` calls. Also, this leads to (IMO) concise logging calls. For example logging an blocking recv is done with:
```
log<...>(exec, blocking{}, "recv", comm, pt2pt{ /*no send buffer*/ {}, {recv_buffer, count, &type}, rank, tag, &status});
```
A variable, blocking all gather is done with:
```
log<...>(exec, blocking{}, "all_gather", comm, all_to_all<fixed, variable>{
    {send_buffer, send_count, &type}, 
    {recv_buffer, {recv_sizes, recv_offsets}, &type
}});
```
Logging non-blocking calls is more difficult. IMO, the `*completed` event can only be logged, after the communication is complete, so after waiting on the MPI request. So, I pass-in a lambda that captures the necessary MPI call information to the request, which then calls this function after the wait has completed. For example, a non-blocking recv looks like this:
```
request req([=, comm = *this](request* req) {
    comm.template log<...>(
        exec, non_blocking{req->get()}, "recv", comm.comm_.get(),
        pt2pt{{}, {recv_buffer, count. &type}, rank, tag});
});
```

Only in 1. the distinction leads to different events. The other axes are realized using `std::variant`. In particular, each case is realized by a struct that captures only the necessary information. 

MPI Type handling:
All MPI types (e.g. `MPI_Comm`, `MPI_Op`, ...) are passed in as `void*`. My reasoning for this is to prevent including the MPI header in `logger.hpp` and conditionally enable the events. Including the MPI header would make it necessary to also link all of our device libraries against MPI (or at least the header), because `logger.hpp` gets included through `executor.hpp`.

I've also enabled MPI events for the profile hook logger and the record logger. Updating the profile hook logger was pretty easy, but the record logger requires a bit more effort. Still, I think this shows how integration into existing loggers can be done.

TODO:
- [ ] exam assembly code if logging is too expensive
- [ ] wrap all communicator calls
- [ ] figure out how to handle one-sided calls

### DIFF
--- a/core/log/logger.cpp
+++ b/core/log/logger.cpp
@@ -86,8 +86,6 @@ constexpr Logger::mask_type
     Logger::mpi_point_to_point_communication_completed_mask;
 constexpr Logger::mask_type Logger::mpi_collective_communication_started_mask;
 constexpr Logger::mask_type Logger::mpi_collective_communication_completed_mask;
-constexpr Logger::mask_type Logger::mpi_reduction_started_mask;
-constexpr Logger::mask_type Logger::mpi_reduction_completed_mask;
 
 constexpr Logger::mask_type Logger::mpi_point_to_point_events_mask;
 constexpr Logger::mask_type Logger::mpi_collective_events_mask;
--- a/core/log/logger.cpp
+++ b/core/log/logger.cpp
@@ -76,8 +76,7 @@ constexpr Logger::mask_type Logger::criterion_check_completed_mask;
 
 constexpr Logger::mask_type Logger::iteration_complete_mask;
 
-constexpr Logger::mask_type Logger::mpi_blocking_communication_mask;
-constexpr Logger::mask_type Logger::mpi_non_blocking_communication_mask;
+constexpr Logger::mpi_mode_mask_type Logger::all_mpi_modes_mask;
 
 constexpr Logger::mask_type
     Logger::mpi_point_to_point_communication_started_mask;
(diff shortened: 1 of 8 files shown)


MarcelKoch:
I've decided to close this PR. Logging/tracing of MPI calls is very hard to do and there are other tools that can give an overview of the communication (e.g. tau or hpctoolkit). I don't think the overhead of managing these ginkgo logging functions is worth the benefit we could gain.
Alternatively, we could add logging to specific communication functions outside the communicator or window class, such as the communication function within the distributed matrix class. But that should be handled in a new PR.

upsj:
If we added a LinOp representing a distributed sparse gather/scatter and used it inside `Matrix::apply`, the corresponding `SparseGather::apply` call would also represent the communication well

MarcelKoch:
@upsj yes, that's what I would imagine.

PR #1305, opened 2023-03-20
Fixes stencil matrix generation

MarcelKoch:
This PR fixes the stencil matrix generation for distributed benchmarks. The issue was in a messed up index reordering.


yhmtsai:
in line 149, `auto row = flat_idx(i, j);` does not change the order as input.
could you also mention the order of point in the comment? or I miss the documentation somewhere.

benchmark/utils/stencil_matrix.hpp
```diff
@@ -116,19 +116,18 @@ gko::matrix_data<ValueType, IndexType> generate_2d_stencil_box(
...
 
-    auto flat_idx = [&](const IndexType ix, const IndexType iy) {
+    auto flat_idx = [&](const IndexType iy, const IndexType ix) {
         auto tpx = target_position(ix, positions[0]);
         auto tpy = target_position(iy, positions[1]);
         if (is_in_box(tpx, dims[0]) && is_in_box(tpy, dims[1])) {
-            return global_offset(tpx, tpy) +
-                   target_local_idx(ix) * discretization_points +
-                   target_local_idx(iy);
+            return global_offset(tpy, tpx) + target_local_idx(ix) +
```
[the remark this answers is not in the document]
MarcelKoch: I reversed the order of the function parameters so that the fastest changing indices are on the right. The position don't matter too much, because it just defines the processor grid. So there is little difference in the end if the grid is [3 x 5] or [5 x 3].

MarcelKoch:
@yhmtsai I've updated the loop variable names to make it clear at in which position they should go.

benchmark/utils/stencil_matrix.hpp
```diff
@@ -209,11 +210,11 @@ gko::matrix_data<ValueType, IndexType> generate_3d_stencil_box(
...
 
-    auto global_offset = [&](const int position_x, const int position_y,
-                             const int position_z) {
-        return position_x * static_cast<int>(local_size) * dims[0] * dims[1] +
+    auto global_offset = [&](const int position_z, const int position_y,
+                             const int position_x) {
+        return position_x * static_cast<int>(local_size) +
                position_y * static_cast<int>(local_size) * dims[0] +
-               position_z * static_cast<int>(local_size);
+               position_z * static_cast<int>(local_size) * dims[0] * dims[1];
```
yhmtsai, line 217: global_offset may overflow?

benchmark/utils/stencil_matrix.hpp
```diff
@@ -266,15 +266,17 @@ gko::matrix_data<ValueType, IndexType> generate_3d_stencil_box(
...
+        for (IndexType iy = 0; iy < discretization_points; ++iy) {
+            for (IndexType ix = 0; ix < discretization_points; ++ix) {
+                auto row = flat_idx(iz, iy, ix);
+                for (IndexType dz : {-1, 0, 1}) {
+                    for (IndexType dy : {-1, 0, 1}) {
+                        for (IndexType dx : {-1, 0, 1}) {
+                            if (is_valid_neighbor(dz, dy, dx)) {
+                                auto col = flat_idx(iz + dz, iy + dy, ix + dx);
                                 if (is_in_box(col, static_cast<IndexType>(
                                                        global_size))) {
```
yhmtsai, line 281: A_data contains the global index so the A_data `IndexType` uses the `GlobalIndexType`, right?

yhmtsai (approved):
LGTM for this pr purpose. 
I have some questions about the overflow.

benchmark/utils/stencil_matrix.hpp
```diff
@@ -116,20 +116,19 @@ gko::matrix_data<ValueType, IndexType> generate_2d_stencil_box(
                             : discretization_points - i);
     };
 
-    auto flat_idx = [&](const IndexType ix, const IndexType iy) {
+    auto flat_idx = [&](const IndexType iy, const IndexType ix) {
         auto tpx = target_position(ix, positions[0]);
         auto tpy = target_position(iy, positions[1]);
```
yhmtsai, line 121: In the develop branch, the others don't use the same order as here to lead the wrong generation?

benchmark/utils/stencil_matrix.hpp
MarcelKoch: Thank you, good point. I will cast everything to `IndexType`, that should fix it.

benchmark/utils/stencil_matrix.hpp
MarcelKoch: Yes. Since the stencil generation can also be used for plain CSR matrices, `IndexType` may also be just `int32`. But in the distributed case your statement is true.

benchmark/utils/stencil_matrix.hpp
MarcelKoch: Exactly, that is why the fix is needed.

pratikvn (approved):
LGTM!

upsj (commented):
It looks like this is a sufficient fix, but for future maintainability, I would suggest adding a few more comments describing the purpose of the individual lambdas and defining the terms we are using ("position", "i", "ix", "iy", "target", ...)
suggestions: box_position, delta, local_x, local_y. Can't come up with a good term for target on the spot.

PR #1306, opened 2023-03-21
Fix tolerance and mpi binding

yhmtsai:
This PR fixes the issue found in frontier.

- Relax some tolerance
@yanjen also face the same issue with different compiler, could you provide your compiler and tolerance?
- Change MPI_CXX_COMPLEX to MPI_C_COMPLEX
MPI also deprecated the CXX binding
https://github.com/mpi-forum/mpi-forum-historic/issues/150


MarcelKoch:
Just a small note, the `MPI_CXX_*_COMPLEX` datatypes still appear in the MPI standard, see https://www.mpi-forum.org/docs/mpi-4.0/mpi40-report.pdf table §3.4. But it is framed in a weird way, so changing it is fine with me.

yhmtsai:
I can also use the macro to choose the type if you think we should stick the CXX TYPE as much as possible
`#if defined(MPI_CXX_FLOAT_COMPLEX) && MPI_CXX_FLOAT_COMPLEX != MPI_DATATYPE_NULL` using CXX type

tcojean (approved):
LGTM.

I think we should not stick with the `MPI_CXX` at all and directly move to `MPI_C`. It has been deprecated since MPI 2.2 I believe.
https://www.mpi-forum.org/docs/mpi-2.2/mpi22-report/node328.htm

upsj (approved):
LGTM! I agree with Terry though that we should just use the C types, they are binary compatible to the C++ types

include/ginkgo/core/base/mpi.hpp
```diff
@@ -116,8 +116,9 @@ GKO_REGISTER_MPI_TYPE(long long, MPI_LONG_LONG_INT);
 GKO_REGISTER_MPI_TYPE(float, MPI_FLOAT);
 GKO_REGISTER_MPI_TYPE(double, MPI_DOUBLE);
 GKO_REGISTER_MPI_TYPE(long double, MPI_LONG_DOUBLE);
-GKO_REGISTER_MPI_TYPE(std::complex<float>, MPI_CXX_FLOAT_COMPLEX);
-GKO_REGISTER_MPI_TYPE(std::complex<double>, MPI_CXX_DOUBLE_COMPLEX);
+// CXX BINDING is deprecated
```
MarcelKoch, line 119: The `MPI_CXX_*_COMPLEX` are not part of the deprecated MPI CXX binding. So perhaps just remove this comment.

PR #1307, opened 2023-03-22
fix the additional residual computation when passing residual criterion

yhmtsai:
In #981, splitting the criterion for iteration and the residual check when users only pass the iteration criterion.
However, it will compute the residual norm from the solution in the iteration update when users pass the residual check.

It fixes the issue by only computing the residual from the solution in the finalized ste

### DIFF
--- a/core/stop/residual_norm.cpp
+++ b/core/stop/residual_norm.cpp
@@ -193,14 +193,17 @@ bool ResidualNormBase<ValueType>::check_impl(
     const NormVector* dense_tau;
     if (updater.residual_norm_ != nullptr) {
         dense_tau = as<NormVector>(updater.residual_norm_);
+    } else if (updater.ignore_residual_check_) {
+        // If solver already provide the residual norm, we will still store it.
+        // Otherwise, we skip the residual check.
+        return false;
     } else if (updater.residual_ != nullptr) {
         norm_dispatch<ValueType>(
             [&](auto dense_r) { dense_r->compute_norm2(u_dense_tau_); },
             updater.residual_);
         dense_tau = u_dense_tau_.get();
-    } else if (set_finalized && updater.solution_ != nullptr &&
-               system_matrix_ != nullptr && b_ != nullptr) {
-        // Only compute the residual from solution in the finalized step
+    } else if (updater.solution_ != nullptr && system_matrix_ != nullptr &&
+               b_ != nullptr) {
         auto exec = this->get_executor();
         norm_dispatch<ValueType>(
             [&](auto dense_b, auto dense_x) {
@@ -210,10 +213,6 @@ bool ResidualNormBase<ValueType>::check_impl(
             },
             b_.get(), updater.solution_);
         dense_tau = u_dense_tau_.get();
-    } else if (!set_finalized) {
-        // If it is not the finalized step and does not contain residual, we
-        // skip the residual norm check
-        return false;
     } else {
         GKO_NOT_SUPPORTED(nullptr);
     }
--- a/core/stop/residual_norm.cpp
+++ b/core/stop/residual_norm.cpp
@@ -198,8 +198,9 @@ bool ResidualNormBase<ValueType>::check_impl(
             [&](auto dense_r) { dense_r->compute_norm2(u_dense_tau_); },
             updater.residual_);
         dense_tau = u_dense_tau_.get();
-    } else if (updater.solution_ != nullptr && system_matrix_ != nullptr &&
-               b_ != nullptr) {
+    } else if (set_finalized && updater.solution_ != nullptr &&
+               system_matrix_ != nullptr && b_ != nullptr) {
+        // Only compute the residual from solution in the finalized step
         auto exec = this->get_executor();
         norm_dispatch<ValueType>(
             [&](auto dense_b, auto dense_x) {
@@ -209,6 +210,10 @@ bool ResidualNormBase<ValueType>::check_impl(
             },
             b_.get(), updater.solution_);
         dense_tau = u_dense_tau_.get();
+    } else if (!set_finalized) {
+        // If it is not the finalized step and does not contain residual, we
+        // skip the residual norm check
+        return false;
     } else {
         GKO_NOT_SUPPORTED(nullptr);
     }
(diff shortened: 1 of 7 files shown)


core/stop/residual_norm.cpp
```diff
@@ -198,8 +198,9 @@ bool ResidualNormBase<ValueType>::check_impl(
             [&](auto dense_r) { dense_r->compute_norm2(u_dense_tau_); },
             updater.residual_);
         dense_tau = u_dense_tau_.get();
-    } else if (updater.solution_ != nullptr && system_matrix_ != nullptr &&
-               b_ != nullptr) {
+    } else if (set_finalized && updater.solution_ != nullptr &&
+               system_matrix_ != nullptr && b_ != nullptr) {
+        // Only compute the residual from solution in the finalized step
```
MarcelKoch, line 203: nit:
```suggestion
        // Only compute the residual from the solution in the finalized step
```

core/stop/residual_norm.cpp
```diff
@@ -209,6 +210,10 @@ bool ResidualNormBase<ValueType>::check_impl(
             },
             b_.get(), updater.solution_);
         dense_tau = u_dense_tau_.get();
+    } else if (!set_finalized) {
+        // If it is not the finalized step and does not contain residual, we
```
MarcelKoch, line 214:
```suggestion
        // If this is not the finalized step and the updater does not contain residual, we
```

MarcelKoch (approved):
LGTM. I guess this could lead to infinite loops in user code, if they created their own solver, but used `set_finalized` incorrectly. But I think that is highly unlikely.
Alternatively, you could throw (some special error) in the `if(!set_finalized)` case and use a try-catch block in IR. But I think that might be too much overhead.

include/ginkgo/core/stop/criterion.hpp
```diff
@@ -123,6 +125,7 @@ class Criterion : public EnableAbstractPolymorphicObject<Criterion> {
     mutable _type* _name##_ {}
 
         GKO_UPDATER_REGISTER_PARAMETER(size_type, num_iterations);
+        GKO_UPDATER_REGISTER_PARAMETER(bool, ignore_residual_check);
```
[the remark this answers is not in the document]
yhmtsai: Yes, the macro initialized {} which should be false for bool.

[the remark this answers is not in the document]
pratikvn: Yes. or maybe even just add a comment here that is initialized to false by default.

upsj (approved):
LGTM! Is it possible to add a regression test based on a logger counting the number of norm2 operations the criterion performs? (and check that it fails on `develop`)

core/test/solver/ir.cpp
```diff
@@ -52,6 +53,22 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+    while (pos < str.length()) {
+        // no overlapped cases
+        pos = str.find(substr, pos);
+        if (pos != std::string::npos) {
+            pos += substr.length();
+            occurrance++;
+        }
+    }
+    return occurrance;
+}
```
pratikvn, line 69:
```suggestion
int count_occurrence(const std::string& str, const std::string& substr)
{
    int occurrence = 0;
    std::string::size_type pos = 0;
    while (pos < str.length()) {
        // no overlapped cases
        pos = str.find(substr, pos);
        if (pos != std::string::npos) {
            pos += substr.length();
            occurrence++;
        }
    }
    return occurrence;
}
```

core/test/solver/ir.cpp
```diff
@@ -458,4 +475,25 @@ TYPED_TEST(Ir, SmootherBuildWithFactory)
...
+    auto logger = gko::share(gko::log::Stream<TypeParam>::create(
+        gko::log::Logger::operation_launched_mask, out));
+    this->exec->add_logger(logger);
+
+    // solver reaches the iteration limit
+    this->solver->apply(b, x);
+
+    // Contains make_residual_norm 3 times: check in initialization and two
+    // iterations. The last iteration exits due to iteration limit.
+    ASSERT_EQ(count_occurrance(out.str(), "make_residual_norm"), 3);
```
pratikvn, line 495:
```suggestion
    ASSERT_EQ(count_occurrence(out.str(), "make_residual_norm"), 3);
```

pratikvn (approved):
Minor typo

yhmtsai:
@upsj good idea.
failed test in https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/pipelines/831036678
```
Expected equality of these values:
  count_occurrance(out.str(), "make_residual_norm")
    Which is: 5
  3
```
three iteration:
two additional residual norm (0th and 1st) in the develop.
2nd iteration does not have it because it exits the iteration check in criterion combine (not due to splitting in IR)

include/ginkgo/core/stop/criterion.hpp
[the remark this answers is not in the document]
yhmtsai: I also add the test to check the default parameter.

yhmtsai:
I need to change stream logger to profilerhook summary because MSVC does not contain enough function infomation.

MSVC: `class gko::detail::RegisteredOperation<class <lambda_0ccdc7866f99fbdf308616cae4f60c00> >`
others with cxxabi demangle: `gko::detail::RegisteredOperation<gko::stop::residual_norm::(anonymous namespace)::make_residual_norm<gko::matrix::Dense<double> const*&, gko::matrix::Dense<double>*, double&, unsigned char&, bool&, gko::array<gko::stopping_status>*&, gko::array<bool>*, bool*, bool*&>(gko::matrix::Dense<double> const*&, gko::matrix::Dense<double>*&&, double&, unsigned char&, bool&, gko::array<gko::stopping_status>*&, gko::array<bool>*&&, bool*&&, bool*&)::{lambda(auto:1)#1}>`
other without cxxabie demangle: `N3gko6detail19RegisteredOperationIZNS_4stop13residual_norm12_GLOBAL__N_118make_residual_normIJRPKNS_6matrix5DenseIdEEPS8_RdRhRbRPNS_5arrayINS_15stopping_statusEEEPNSG_IbEEPbRSN_EEEDaDpOT_EUlT_E_EE,0x7ffe591e4c60`
MSVC does not have information for the operation, so I can not

PR #1310, opened 2023-03-23
Add the L1 jacobi

yhmtsai:
This add the L1 to Jacobi. The detail is shown in https://epubs.siam.org/doi/abs/10.1137/100798806
original preconditioner is on $M$ (which is diagonal of A for scalar Jacobi or diagonal block of A)
L1 smoother is on $M' = M + D^{\mathcal{l}_1}$ whose $D^{\mathcal{l}_1}$ is a diagonal matrix with

$$D_{ii}^{\mathcal{l}_1} = \sum_{j \in \text{the off-diagonal block of corresponding block of i}}{|A_{ij}|}$$

I am not sure whether the theorems in the paper still work for negative diagonal entries.


MarcelKoch:
I'm a bit confused. The paper talks about the L1 Jacobi only in a distributed setting, ie the off-diagonal entries you mentioned are the entries of our non-local matrix in the distributed matrix. But your implementation is for non-distributed matrices. Can you perhaps provide another reference for the non-distributed case?

MarcelKoch:
I think sec. 6.2 means using the preconditioner from 6. `M = M_H + D^l` with `M_H = D`. The `D^l` is still build from the non-local columns. I'm not saying it still works for non-distributed matrices, just that the paper description applies to distributed matrices. (The L1 smoother for non-distributed matrices would just be the standard Jacobi.)

yhmtsai:
If the partition is only designed with the mpi (let me use the mpi here to distinguish), L1 is indeed normal Jacobi
i.e using Schwarz with L1 will give this behavior.
But the paper from second paragraph of section 5 makes me think it works for any distributed (any block partition). design the partition by the mpi is one example

MarcelKoch:
The L1 smoother from sec. 6 is an improvement on the hybrid smoother from sec 5. The hybrid smoother applies a local smoother only on the local matrix block of a distributed matrix and ignores the non-local part. The L1 smoother than takes the non-local part into account through the `D^l` matrix.
Again, this might also work for non-distributed matrices, but maybe it would not be the L1 smoother people would expect. But from the other paper you linked, people seem to be not very careful about this definition.

MarcelKoch:
Perhaps naming it 'lumped-Jacobi' would be better suited to differentiate between those two variants. And I guess the L1 from the first paper could be implemented as another distributed preconditioner.
Also, the paper you mentioned does a full row sum, so it also includes the diagonal value.

yhmtsai:
The first paper considers the diagonal entries are positive, so they are equivalent.
Distributed on mpi or distributed on the block to me are the same thing, which are ways to cut the matrices.
Applying L1 to LocalPreconditioner and to Schwarz are enough to distinguish.
With lumped, we still need to indicate it is L1 in the comments

MarcelKoch:
My argument is that what is implemented in this PR is not the L1 from the paper, so we should not call it that. But honestly, I don't have a great overview of how the term L1 smoother is usually interpreted.

pratikvn:
Maybe I am misunderstanding something, but in the non-distributed case, this is basically diagonal relaxation with weights equal to row sums, right ? So, I think it definitely makes sense to have as a smoother as an alternative to the scalar Jacobi smoother.

I think the name L1 is well suited because you are taking the L1 norm on the row space vectors ?  $|v| = \sum_{n}|v_i|$. 

Not sure if there is some mathematical basis for this yet, we could also try seeing if any other p-norms also make sense, Particularly with $p=\infty$, $p=2$.

MarcelKoch:
If we go with @pratikvn suggestion, and base this version on the L1 row norm, then we need to take the absolute value of the diagonal.

yhmtsai:
I am not sure whether to take the absolute value of the diagonal value.
It might make sense for the scalar version but not for the block version. 
What's the meeting of the absolute of diagonal block to the diagonal value?

yhmtsai:
Yes, from the block-version, it only updates the diagonal value.
I think l1 is mostly from taking the summation of the absolute value in non-local block's row, not the entire row.
(l1 norm of non-local-block rows) L1 does not apply to the local block.
From the formula, it does not change the other values in the block.
In the smoother paper, they only consider the positive value in diagonal (see 5.5), although they have additional properties more than positive diagonal values.
In another paper, they consider the SPD cases, so the diagonal values are also positive.
When the diagonal value is positive, it does not change the result no matter whether we take the absolute value of the diagonal value or not.
Another treatment is to $A_{ii} + sign(A_{ii}) D^{l_1}$ if -A is SPD for scalar Jacobi,

From the paper or formula perspective, I prefer only adding values on the diagonal value without take absolute on the diagonal value first.
For the negative diagonal value, we do not know whether it works or not from the papers currently.

include/ginkgo/core/preconditioner/jacobi.hpp
```diff
@@ -371,6 +371,13 @@ class Jacobi : public EnableLinOp<Jacobi<ValueType, IndexType>>,
         gko::array<index_type> GKO_FACTORY_PARAMETER_VECTOR(block_pointers,
                                                             nullptr);
 
+        /**
+         * Use L1 Jacboi.
+         * If it is true, it generates the preconditioner on A + Diag(sum_{k in
+         * off diagonal block of i} A_ik) not A.
+         */
+        bool GKO_FACTORY_PARAMETER_SCALAR(l1, false);
```
MarcelKoch, line 379: I'm still not convinced that `l1` is a good name. Perhaps `aggregate`, or `aggregate_l1`.
One issue with just `l1` is that we will also have the distributed L1 preconditioner, as mentioned in the original paper. But that is quite different from the one here, so we need better naming to prevent confusion.

yhmtsai: I think they are the same thing. They only use different way to define block  (by node, diagonal block, or value), but they all add the L1 norm from the row in off-block-diag to the diag.

MarcelKoch: They are not the same. Using the l1 defined here will not give you a distributed preconditioner. Also the distributed version can be used both with local Jacobi and Gauss-Seidel preconditioner.

yhmtsai: I do not get it.
If you have the block list, the diagonal block { $B_i$ } and the off diagonal block { $O_i$ }
the L1 does the diagonal value update
the diagonal value $d_k$ in $B_i$ is updated by $d_k$ += sum abs( value in the same row as $d_k$ in $O_i$ )

For sure, L1 in this pr is specific for Jacobi because we can only know the block of block Jacobi in generation.
scalar L1 Jacobi takes each diagonal value as the block
block L1 Jacobi takes the blocks generated or given as the block
distributed L1 takes the local matrix (assume squared) as the block.

The diagonal treatment are the same

MarcelKoch: The point is this is a variation of the sequential jacobi. In the distributed case, we would have a separate L1 preconditioner (or with a different name) that is not specific to jacobi.
Having both with the same name would be confusing.

yhmtsai: I see

reference/preconditioner/jacobi_kernels.cpp
```diff
@@ -724,6 +724,59 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_AND_INDEX_TYPE(
...
+    for (IndexType row = 0; row < csr->get_size()[0]; row++) {
+        auto off_diag = zero<ValueType>();
+        for (auto i = csr->get_const_row_ptrs()[row];
+             i < csr->get_const_row_ptrs()[row + 1]; i++) {
+            if (csr->get_const_col_idxs()[i] == row) {
+                continue;
+            }
+            off_diag += abs(csr->get_const_values()[i]);
+        }
+        diag->get_values()[row] += off_diag;
```
MarcelKoch, line 741:
```suggestion
        diag->get_values()[row] = abs(diag->get_values()[row]) + off_diag;
```

reference/test/preconditioner/jacobi_kernels.cpp
```diff
@@ -1108,4 +1110,156 @@ TYPED_TEST(Jacobi, ConvertsEmptyToDense)
 }
 
 
+TYPED_TEST(Jacobi, ScalarL1Diag)
+{
+    /* example matrix:
+        1   1
+        2   2
+       -5       1
+        1      -1
```
MarcelKoch, line 1119:
```suggestion
        1   1   0   0
        2   2   0   0
       -5   0   1   0
        1   0   -1   0
```

reference/test/preconditioner/jacobi_kernels.cpp
```diff
@@ -1108,4 +1110,156 @@ TYPED_TEST(Jacobi, ConvertsEmptyToDense)
...
+     */
+    using Mtx = typename TestFixture::Mtx;
+    using index_type = typename TestFixture::index_type;
+    using value_type = typename TestFixture::value_type;
+    auto mtx = Mtx::create(this->exec, gko::dim<2>{4}, 8);
```
MarcelKoch, line 1124:
```suggestion
    auto mtx = Mtx::create(this->exec, gko::dim<2>{4, 4}, 8);
```
IMO the single parameter `dim` constructor is a bit confusing.

reference/test/preconditioner/jacobi_kernels.cpp
```diff
@@ -1108,4 +1110,156 @@ TYPED_TEST(Jacobi, ConvertsEmptyToDense)
...
+}
+
+
+TYPED_TEST(Jacobi, BlockL1)
+{
+    /* example matrix:
+        1   1
+        2   2
+       -5       1
+        1      -1   1
```
MarcelKoch, line 1149:
```suggestion
        1   1   0   0
        2   2   0   0
       -5   0   1   0
        1   0  -1   1
```

reference/test/preconditioner/jacobi_kernels.cpp
```diff
@@ -1108,4 +1110,156 @@ TYPED_TEST(Jacobi, ConvertsEmptyToDense)
...
+     */
+    using Mtx = typename TestFixture::Mtx;
+    using index_type = typename TestFixture::index_type;
+    using value_type = typename TestFixture::value_type;
+    auto mtx = Mtx::create(this->exec, gko::dim<2>{4}, 9);
```
MarcelKoch, line 1154:
```suggestion
    auto mtx = Mtx::create(this->exec, gko::dim<2>{4, 4}, 9);
```

reference/test/preconditioner/jacobi_kernels.cpp
```diff
@@ -1108,4 +1110,156 @@ TYPED_TEST(Jacobi, ConvertsEmptyToDense)
...
+    gko::array<index_type> block_ptr(this->exec, {0, 1, 2, 4});
+
+    gko::kernels::reference::jacobi::block_l1(this->exec, 3, block_ptr,
+                                              mtx.get());
+
+    GKO_ASSERT_MTX_NEAR(mtx,
+                        l({{2.0, 1.0, 0.0, 0.0},
+                           {2.0, 4.0, 0.0, 0.0},
+                           {-5.0, 0.0, 6.0, 0.0},
+                           {1.0, 0.0, -1.0, 2.0}}),
```
MarcelKoch, line 1169: This doesn't seem right. Are the off-diagonal values of the diagonal block also added to the diagonal of the diagonal block? IMO that shouldn't be the case, only values from the off-diagonal blocks should be used (at least according to the original reference).

test/preconditioner/jacobi_kernels.cpp
```diff
@@ -252,6 +257,66 @@ TEST_F(Jacobi,
 }
 
 
+TEST_F(Jacobi, ScalarInLargeMatrixEquivalentToRef)
```
MarcelKoch, line 260:
```suggestion
TEST_F(Jacobi, ScalarL1LargeMatrixEquivalentToRef)
```

MarcelKoch (changes requested):
Thanks for changing the name, for me the name is fine now.
I do have some questions about the block version though.

reference/test/preconditioner/jacobi_kernels.cpp
yhmtsai: I think it is the case, right?
the blocks here are (0, 0), (1, 1), and (2:4, 2:4)

reference/preconditioner/jacobi_kernels.cpp
yhmtsai: I think we discussed this a little bit before, but did not get a reasonable choice when the diagonal values are negative.
L1 is originally designed for postive diagonal value

reference/test/preconditioner/jacobi_kernels.cpp
MarcelKoch: You're right, I didn't understand the blocking of the matrix. Maybe the blocks can be denoted in the comment above?

include/ginkgo/core/preconditioner/jacobi.hpp
```diff
@@ -371,6 +371,14 @@ class Jacobi : public EnableLinOp<Jacobi<ValueType, IndexType>>,
         gko::array<index_type> GKO_FACTORY_PARAMETER_VECTOR(block_pointers,
                                                             nullptr);
 
+        /**
+         * Use L1 Jacboi, which is introduced in the paper A. H. Baker et al.
+         * "Multigrid smoothers for ultraparallel computing."
+         * If it is true, it generates the preconditioner on A + Diag(sum_{k in
+         * off diagonal block of i} A_ik) not A.
```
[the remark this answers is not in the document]
yhmtsai: I think it more means adding L1 norm of off-block to the diagonal value not make diagonal value as L1 norm (it is also not L1 norm of row in general)
I will add the comments for the description about the matrix property.
The paper discusses the matrix with the following property. 
$$
a_{ii} \geq \theta \sum_{j \in \text{off diagonal block}} |a_{ij}|, \theta \geq 0
$$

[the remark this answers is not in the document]
yhmtsai: from the paper description, the new matrix is $A + D^\ell$, so we only touch off-block part

include/ginkgo/core/preconditioner/jacobi.hpp
```diff
@@ -371,6 +371,14 @@ class Jacobi : public EnableLinOp<Jacobi<ValueType, IndexType>>,
         gko::array<index_type> GKO_FACTORY_PARAMETER_VECTOR(block_pointers,
                                                             nullptr);
 
+        /**
+         * Use L1 Jacboi, which is introduced in the paper A. H. Baker et al.
+         * "Multigrid smoothers for ultraparallel computing."
```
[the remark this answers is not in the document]
yhmtsai: I do not think the description limited by using parallel resource to define the block.

reference/test/preconditioner/jacobi_kernels.cpp
yhmtsai: I still keep the original format like the example matrix in the other tests

include/ginkgo/core/preconditioner/jacobi.hpp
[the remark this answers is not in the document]
yhmtsai: That's another reason I prefer not to adding the absolute on the diagonal value. That makes the formula different and hard to write the description on the matrices which is not discussed by the reference paper.

common/unified/preconditioner/jacobi_kernels.cpp
```diff
@@ -124,6 +124,72 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE(
...
+        [] GKO_KERNEL(auto row, auto row_ptrs, auto col_idxs, auto vals,
+                      auto diag) {
+            auto off_diag = zero(vals[0]);
+            for (auto i = row_ptrs[row]; i < row_ptrs[row + 1]; i++) {
+                if (col_idxs[i] == row) {
+                    continue;
+                }
+                off_diag += abs(vals[i]);
+            }
+            diag[row] += off_diag;
```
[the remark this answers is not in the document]
yhmtsai: I prefer not to adding this. It has the same effect when the positive diagonal, which is discussed by the reference paper. I add some note to indicate that it is unclear when applying to the negative diagonal value

reference/test/preconditioner/jacobi_kernels.cpp
yhmtsai: done

include/ginkgo/core/preconditioner/jacobi.hpp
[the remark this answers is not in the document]
MarcelKoch: I think it would be fine to keep this reference, but explicitly mention that this is not the distributed version of the L1 Jacobi.

include/ginkgo/core/preconditioner/jacobi.hpp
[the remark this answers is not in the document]
MarcelKoch: Ok, I see. then it should be noted, that this only works with matrices that have a positive diagonal.

reference/test/preconditioner/jacobi_kernels.cpp
MarcelKoch: IMO this should still be updated, since the original comment is missing the last column.

include/ginkgo/core/preconditioner/jacobi.hpp
```diff
@@ -371,6 +371,19 @@ class Jacobi : public EnableLinOp<Jacobi<ValueType, IndexType>>,
...
                                                             nullptr);
 
+        /**
+         * Use L1 Jacboi, which is introduced in the paper A. H. Baker et al.
+         * "Multigrid smoothers for ultraparallel computing." This paper
+         * discusses this type of smoother with the following matrix property.
+         * $ A_{ii} \geq \theta \sum_{j \in \text{off diagonal block}} |A_{ij}|,
+         * \theta \geq 0 $
+         * If it is true, it generates the preconditioner on A +
+         * Diag(sum_{k in off-diagonal block of i} |A_ik|) not A. We aggregate
```
MarcelKoch, line 381: nit:
```suggestion
         * Diag(sum_{k in off-diagonal block of i} |A_ik|) instead of A. We aggregate
```

include/ginkgo/core/preconditioner/jacobi.hpp
```diff
@@ -371,6 +371,19 @@ class Jacobi : public EnableLinOp<Jacobi<ValueType, IndexType>>,
...
+         * the L1 norm of the value with the same row in the off-diagonal block
```
MarcelKoch, line 382: nit:
```suggestion
         * the absolute value of the entries with the same row in the off-diagonal block
```

test/preconditioner/jacobi_kernels.cpp
```diff
@@ -252,6 +257,66 @@ TEST_F(Jacobi,
...
+                    .on(exec)
+                    ->generate(mtx);
+
+    GKO_ASSERT_MTX_NEAR(gko::as<Bj>(d_bj.get()), gko::as<Bj>(bj.get()), 1e-13);
+}
+
+
+TEST_F(Jacobi, BlockL1InLargeMatrixEquivalentToRef)
+{
+    /* example matrix duplicated 50 times:
```
MarcelKoch, line 289: same here, it would be good to have also show the last column

MarcelKoch (approved):
only smaller nits left on the documentation.

reference/test/preconditioner/jacobi_kernels.cpp
yhmtsai: I see the point. I will update it

PR #1311, opened 2023-03-23
Batched Tridiagonal Matrix Format and Solver

aggarwal2000:
This PR adds:

1. Batched Tridiagonal Matrix Format
2. Batched Tridiagonal Solver (based on Wang and Mou - Parallel Gaussian Elimination algorithm)


include/ginkgo/core/matrix/batch_tridiagonal.hpp
```diff
@@ -0,0 +1,557 @@
...
+
+
+    /**
+     * Creates a BatchTridiagonal matrix by duplicating BatchTridiagonal matrix
+     *
+     * @param exec  Executor associated to the matrix
+     * @param num_duplications  The number of times to duplicate
+     * @param input  The matrix to be duplicated.
+     */
+    // NOTE: Currently, this only works for a uniform batch
```
pratikvn, line 487: I would then add an `GKO_ASSERT(this->get_size().stores_equal_sizes() == true)`

include/ginkgo/core/solver/batch_tridiagonal_solver.hpp
```diff
@@ -0,0 +1,278 @@
...
+
+/**
+ * The approach to be used to solve the batched trdiagonal system on CUDA/HIP
+ * executor.
+ *
+ * Both "WM_pGE_app1" and "WM_pGE_app2" are based on the Wang and Mou parallel
+ * Gaussian Elimination algorithm. The only difference between the two lies is
+ * the parallel scheme; in "WM_pGE_app1" each thread of the subwarp operates on
+ * a single row of the matrix tile, while in "WM_pGE_2", each thread of the
+ * subwarp handles two adjacent rows of the matrix tile.
```
pratikvn, line 58: Can add a sentence on which strategy the paper implements as well ?

include/ginkgo/core/solver/batch_tridiagonal_solver.hpp
```diff
@@ -0,0 +1,278 @@
...
+ *
+ * The approach- "vendor_provided" makes use of the "gtsv2StridedBatched"
+ * function provided by cuSparse/hipSparse.
+ *
+ */
+enum class batch_tridiag_solve_approach {
+    WM_pGE_app1,
```
pratikvn, line 65: I think we dont need to so specific here to have varied capitalization ? Can we just use `wm_pge_app1` and `wm_pge_app2`  ? Or maybe something a bit more descriptive would be better `wm_pge_1row` and `wm_pge_2row`  ?

include/ginkgo/core/solver/batch_tridiagonal_solver.hpp
```diff
@@ -0,0 +1,278 @@
...
+     * taken up by the "vendor-provided" approach to get the actual computation
+     * time.
+     *
+     * Note: The time gets added up in each solver apply step, therefore,
+     * accessing this time after performing n solver applies means that the time
+     * returned is the total time taken up by buffersize calculation +
+     * allocation in n steps.
+     *
+     */
+    double get_time_in_millisec_to_be_subtracted() const
```
pratikvn, line 132: I think it is better to rename this to preprocess time.

include/ginkgo/core/solver/batch_tridiagonal_solver.hpp
```diff
@@ -0,0 +1,278 @@
...
+    {
+        return millisec_to_be_subtracted_;
+    }
+
+
+    void set_time_in_millisec_to_be_subtracted(const double time)
```
pratikvn, line 138: I get why we need to have a getter, but why do we need a setter function ?

common/cuda_hip/solver/batch_tridiagonal_solver_kernels.hpp.inc
```diff
@@ -0,0 +1,878 @@
...
+        // Approach: a thread in the subwarp handles one row of the matrix or to
+        // be precise, a row in the matrix tile
+        const auto tile_size = subwarp_size;
+        const auto num_tiles = ceildiv(nrows, tile_size);
+        const bool is_last_tile_similar = ((nrows % tile_size) == 0);
+        const int final_group_size = pow(2, num_WM_steps);
+
+        if (final_group_size > tile_size) {
+            printf(
+                "\n Number of WM steps ans subwarp size should satisfy the "
```
pratikvn, line 255:
```suggestion
                "\n Number of WM steps and subwarp size should satisfy the "
```

common/cuda_hip/solver/batch_tridiagonal_solver_kernels.hpp.inc
```diff
@@ -0,0 +1,878 @@
...
+                "inequality: pow(2, WM_steps) <=  tile_size, where tile_size = "
+                "subwarp_size");
```
pratikvn, line 257: I think condition requirement should be mentioned and maybe even verified when launching the kernel. So, we shouldn't be checking this condition here.

common/cuda_hip/solver/batch_tridiagonal_solver_kernels.hpp.inc
```diff
@@ -0,0 +1,878 @@
...
+        const auto tile_size = 2 * subwarp_size;
+        const auto num_tiles = ceildiv(nrows, tile_size);
+        const bool is_last_tile_similar = ((nrows % tile_size) == 0);
+        const int final_group_size = pow(2, num_WM_steps);
+        if (final_group_size > tile_size) {
+            printf(
+                "\n Number of WM steps ans subwarp size should satisfy the "
+                "inequality: pow(2, WM_steps) <= tile_size, where tile_size = "
+                "2 "
+                "* subwarp_size");
```
pratikvn, line 761: Same here as well. Can be moved to the kernel call location, or maybe even to the core function.

pratikvn (commented):
Really excellent work! One question from my side:

tile_size, wm_steps and subwarp_size are correlated right ? Atleast one condition that I see needs to be satisfied is: $2^{wm_steps} \leq 2 *subwarp_size$. 

On this front I think it is more intuitive for the user to set the `tile_size` and we can derive the `subwarp_size` from that  ?

The `wm_steps` can also be seen as how many times you want to recursively do the divide and conquer approach, right ? So how about renaming that to `num_recursive_steps`  ?

pratikvn:
format!

pratikvn:
format!

PR #1312, opened 2023-03-23
Avoid transpose in symbolic Cholesky benchmark

upsj:
The transposition is pretty slow on OpenMP, so to have a fair performance baseline, I want to skip all the operations needed to symmetrize the matrix.

### DIFF
--- a/benchmark/sparse_blas/operations.cpp
+++ b/benchmark/sparse_blas/operations.cpp
@@ -647,20 +647,27 @@ private:
 
 class SymbolicCholeskyOperation : public BenchmarkOperation {
 public:
-    explicit SymbolicCholeskyOperation(const Mtx* mtx) : mtx_{mtx}, result_{} {}
+    explicit SymbolicCholeskyOperation(const Mtx* mtx, bool symmetric)
+        : mtx_{mtx}, symmetric_{symmetric}, result_{}
+    {}
 
     std::pair<bool, double> validate() const override
     {
-        const auto exec = mtx_->get_executor();
-        const auto symm_result = result_->clone();
-        const auto lt_factor = gko::as<Mtx>(symm_result->transpose());
-        const auto scalar = gko::initialize<gko::matrix::Dense<etype>>(
-            {gko::one<etype>()}, exec);
-        const auto id =
-            gko::matrix::Identity<etype>::create(exec, mtx_->get_size()[0]);
-        lt_factor->apply(scalar, id, scalar, symm_result);
-        return std::make_pair(
-            validate_symbolic_factorization(mtx_, symm_result.get()), 0.0);
+        if (symmetric_) {
+            return std::make_pair(
+                validate_symbolic_factorization(mtx_, result_.get()), 0.0);
+        } else {
+            const auto exec = mtx_->get_executor();
+            const auto symm_result = result_->clone();
+            const auto lt_factor = gko::as<Mtx>(symm_result->transpose());
+            const auto scalar = gko::initialize<gko::matrix::Dense<etype>>(
+                {gko::one<etype>()}, exec);
+            const auto id =
+                gko::matrix::Identity<etype>::create(exec, mtx_->get_size()[0]);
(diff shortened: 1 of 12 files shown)


core/factorization/symbolic.hpp
```diff
@@ -44,14 +44,22 @@ namespace factorization {
...
  * @param mtx  the input matrix
+ * @param symmetrize  should we output the pattern of L + L^T or just L?
  * @param factors  the output factors stored in a combined pattern
+ * @param forest  the elimination forest of the input matrix
  */
 template <typename ValueType, typename IndexType>
-void symbolic_cholesky(const matrix::Csr<ValueType, IndexType>*,
-                       std::unique_ptr<matrix::Csr<ValueType, IndexType>>&,
-                       std::unique_ptr<elimination_forest<IndexType>>&);
+void symbolic_cholesky(
```
MarcelKoch, line 52: perhaps it makes sense to always just return the pattern of `L`? Then you would have to add the symmetrization to the places where you call this, but IMO it would be clearer what the function does.

yhmtsai (commented):
Maybe still need to keep one version including transpose time?

core/factorization/symbolic.hpp
MarcelKoch: TBH, I don't think that is the purpose of `develop`. Why not use a branch that has everything you need in it?

upsj:
@yhmtsai agreed, I added it

core/factorization/symbolic.hpp
```diff
@@ -44,14 +44,22 @@ namespace factorization {
  * Computes the symbolic Cholesky factorization of the given matrix.
  *
  * @param mtx  the input matrix
+ * @param symmetrize  should we output the pattern of L + L^T or just L?
```
thoasm, line 47: nit: please specify which of these two options correspond to `true` and `false`.

benchmark/sparse_blas/operations.cpp
```diff
@@ -705,8 +722,12 @@ const std::map<std::string,
...
              return std::make_unique<SymbolicLuOperation>(mtx);
          }},
-        {"symbolic_cholesky", [](const Mtx* mtx) {
-             return std::make_unique<SymbolicCholeskyOperation>(mtx);
+        {"symbolic_cholesky",
+         [](const Mtx* mtx) {
+             return std::make_unique<SymbolicCholeskyOperation>(mtx, false);
+         }},
+        {"symbolic_cholesky_symmetric", [](const Mtx* mtx) {
+             return std::make_unique<SymbolicCholeskyOperation>(mtx, true);
```
thoasm, line 730: It is not great that the old name `symbolic_cholesky` does something different now compared to previously.
Would it not be possible to keep the old functionality with the old name and have a new name for the new functionality? Maybe `symbolic_cholesky_nonsymmetric`?
Since you were talking about reproducibility, this would make things significantly harder to reproduce an old benchmark that was using `symbolic_cholesky`.

thoasm (approved):
I am not a fan of renaming an old functionality instead of renaming the new one, as this makes it difficult to re-generate old benchmark results.

benchmark/sparse_blas/operations.cpp
upsj: The benchmark has only been there for two days #1302, so I see no issue to renaming it

thoasm: Good point, I didn't know it was that young. In that case, I agree that there should not be any issue renaming it.

upsj:
@yhmtsai To clarify: I want the code I am benchmarking on to be as close as possible to `develop`, it doesn't necessarily need to be merged, but it should be reviewed (I won't merge this now, because there seem to have been some compiler changes in MSVC which break the overflow check)

core/factorization/symbolic.hpp
upsj: good point!

upsj:
Found the MSVC issue, we were reading uninitialized data in the prefix_sum kernel, I fixed it and added tests

PR #1318, opened 2023-04-03
Draft: Interface for 'local-only' matrices

MarcelKoch:
This PR shows the bare interface for 'local-only' matrices, without any implementation details. 

In contrast to our current distributed `Vector` and `Matrix`, this format does not need any global information. Instead, the key element is describing the communication in terms of shared indices. These are indices that appear in more than one local index set. It is not necessary that these indices have unique global indices, all that we require is that the relation between local indices and which rank also shares them and the indices local to their numbering.

This supports both unique and shared ownership of the shared DOFs. More concretely, a DOF that appears in the local index set of multiple ranks may be exclusively owned by one rank or equally by all ranks. The difference is currently realized as an enum, but it might be useful in the future to allow more arbitrary weighting between the ranks. At the moment, the ownership for all shared DOFs has to be the same, but that might be relaxed.

The distributed operator is very simple. It just applies the local operator and then issues a communication update to the result vector.

The distributed vector is a bit more complicated. It has a routine to update the shared DOFs with the values from other ranks. This is called making the vector consistent. Assuming a consistent vector, this step only has to happen if non-pointwise operations are executed. In all other cases, e.g. vector additions, scaling, the vector stays consistent. Additionally, the vector reductions have to be handled with special care. In the case of exclusive ownership, non-owned DOFs have to be set to zero, while in the shared ownership mode, the DOFs have to be scaled by `sqrt(1/#Owners)` to compensate for adding the same values multiple times.

Usage example for an overlapping domain decomposition:
```
auto shared_idxs = /* compute shared idxs */;
auto info = std::make_shared<sharing_info<value_type>>(comm, shared_idxs, sharing_mode::set);

auto x = LocalVector<>create(exec, comm, info);
auto y = LocalVector<>create(exec, comm, info);
/* fill x and y like Dense */

auto A = Csr<>create(exec);
/* fill A as usual */

auto local_cg = Cg<>::build().on(exec)->generate(A);

auto cg = Cg<>::build().with_generated_preconditioner(
    ConsistentOp::create(exec, local_cg))
  .on(exec)->generate(ConsistentOp::create(exec, A));
cg->apply(x, y);
```

I'm open to any criticism, or suggestions on this topic.

TODO:

- [ ] find better names
- [ ] evaluate reordering and non-blocking option
- [ ] evaluate user functions for sharing mode

### DIFF
--- a/include/ginkgo/core/distributed/lin_op.hpp
+++ b/include/ginkgo/core/distributed/lin_op.hpp
@@ -109,6 +109,30 @@ protected:
 };
 
 
+class ConsistentOp : public EnableLinOp<ConsistentOp> {
+protected:
+    void apply_impl(const LinOp* b, LinOp* x) const override
+    {
+        // need to run both statements through precision dispatch
+        local_op->apply_impl(as<LocalVector>(b)->get_dense(),
+                             as<LocalVector>(x)->get_dense());
+        as<LocalVector>(x)->make_consistent();
+    }
+
+    void apply_impl(const LinOp* alpha, const LinOp* b, const LinOp* beta,
+                    LinOp* x) const override
+    {
+        // need to run both statements through precision dispatch
+        local_op->apply(alpha, as<LocalVector>(b)->get_dense(), beta,
+                        as<LocalVector>(x)->get_dense());
+        as<LocalVector>(x)->make_consistent();
+    }
+
+private:
+    std::unique_ptr<LinOp> local_op;
+};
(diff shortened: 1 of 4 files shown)


MarcelKoch:
@fritzgoebel Could you give this a look to see if that would fit your needs for the BDDC?

MarcelKoch:
The communication class is under re-design and will be put into it's own PR.

PR #1320, opened 2023-04-05
remove the implicit IR in multigrid construction

yhmtsai:
This PR removes the implicit smoother wrapper in the multigrid.

The original design thinks putting a solver without considering input (usually from non-iterative methods) as smoother does not make sense, so add IR on it.
With initial_guess_mode = zero, the IR does not need to use the initial guess from input.
In the original code, it adds another IR, which is not expected.

It will only put the solver from user input directly.


MarcelKoch (approved):

vasilisge0 (approved):

yhmtsai:
rebase!

PR #1324, opened 2023-04-18
generate the test pipeline if the build job require it with specific runner

yhmtsai:
This PR is to add the flexibility if horeka is down.

I have tried some possible way, but they do not always work as my initial thought
1. (Work) check the server status, run the two build/test jobs via slurm or the build+test in our own server
adds some fallback job if horeka is not reachable.
checks the reachability of horeka before launching the horeka related jobs.
When using `on_failure` with `needs`, the job will be skipped if one of job is skipped
Thus, the fallback can only rely on the check.
https://gitlab.com/gitlab-org/gitlab/-/issues/388866
https://gitlab.com/gitlab-org/gitlab/-/issues/36610
2. (Should Work but limit resource) unified build between docker and slurm, which select the build script from `CI_RUNNER_TAGS`
Due to current slurm runner design, it will pass the script from gitlab to runner.
Thus, it will combine build and test on gpu partition, which is not good for the resource.
Not sure it is easy or not, if scripts write gpu_scripts to cache, slurm runner will run the generated script with gpu job additionally. 
3. (Not Work)runner dyanmic tag
from https://docs.gitlab.com/ee/ci/yaml/#tags, it can support CI/CD variable.
However, it is decided when job create not run. After creating the manual job, resetting the runner_tag works but it does not affect runner selection
`private_ci` and `$RUNNER_TAG` are used in `tags`
-https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/4141132266
image
in the right side, the tags are still `private_ci` `status-jobs`, but the `RUNNER_TAG` is shown as `nla-gpu`
-https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/4141192635
image
if we do not define the default `RUNNER_TAG`, it will use plain text ${RUNNER_TAG} directly and still can not be overwritten in the runner selection.
-The pipeline variable should work because it is overwrited before job creation. However, we do not know the server status before the job creation. (or, one additional pipeline to trigger the current pipeline)
-It is similar to the `rules`, after generation, the job is decided to create or not already
-rules can not depends on artifact, so we can not use artifact like variable to select job. https://gitlab.com/gitlab-org/gitlab/-/issues/215100
4. (Not work for the github status) run manual job by job according to `CI_RUNNER_TAGS`
have the corresponding test job in manual mode, if build job requires it based on `CI_RUNNER_TAGS`, run the test job via API.
However, the manual job can be allow_failure or not. if it allows failure, we will not know the status from the github status. if it doesn't allow, it will show blocked (failed/pending) if they are not executed.
5. (Work) create the child test pipeline
If build job requests the specific runner to run the test, it will write a small job into yml. Thus, each job can use different runner. The later job collects these to generate child pipeline. note. if no job is generated, it will add a dummy job to avoid empty pipeline issue. Note. to upload artifact, the host server also needs to install gitlab-runner.

TODO:
- [x] clean it up and apply it to all related jobs

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -2,10 +2,8 @@ stages:
   - init-status
   - sync
   - trigger_pipeline
-  - check
   - build
   - test
-  - fallback
   - code_quality
   - deploy
   - QoS_tools
@@ -85,13 +83,6 @@ trigger_pipeline:
         echo "Can not find the corresponding Pull Request"
       fi
 
-check/horeka:
-  stage: check
-  extends:
-    - .use_gko-nocuda-nompi-gnu9-llvm8
-  script: 
-    - apt update && apt install -y netcat
-    - nc -w 20 -z horeka.scc.kit.edu 22
 
 # Build jobs
 # Job with example runs.
@@ -286,7 +277,7 @@ build/cuda102/nompi/intel/cuda/debug/static:
 # cuda 11.0 and friends on HoreKa with tests
 build/cuda110/mvapich2/gcc/cuda/debug/shared:
   extends:
-    - .build_template
+    - .unified_build_template
     - .default_variables
     - .full_test_condition
     - .use_gko-cuda110-mvapich2-gnu9-llvm9-intel2020
@@ -301,25 +292,12 @@ build/cuda110/mvapich2/gcc/cuda/debug/shared:
     USE_NAME: "cuda110-mvapich2-gcc-${CI_PIPELINE_ID}"
     KEEP_CONTAINER: "ON"
     USE_SLURM: 0
-
-test/cuda110/mvapich2/gcc/cuda/debug/shared:
-  extends:
-    - .horeka_test_template
-    - .default_variables
-    - .full_test_condition
-    - .use_gko-cuda110-mvapich2-gnu9-llvm9-intel2020
-  variables:
-    USE_NAME: "cuda110-mvapich2-gcc-${CI_PIPELINE_ID}"
-    SLURM_PARTITION: "accelerated"
-    SLURM_GRES: "gpu:4"
-    SLURM_TIME: "02:00:00"
-  dependencies: null
-  needs: [ "build/cuda110/mvapich2/gcc/cuda/debug/shared" ]
+    TEST_SLURM_GRES: "gpu:4"
 
 
 build/cuda110/nompi/clang/cuda/release/static:
   extends:
-    - .build_template
+    - .unified_build_template
     - .default_variables
     - .full_test_condition
     - .use_gko-cuda110-mvapich2-gnu9-llvm9-intel2020
@@ -334,20 +312,7 @@ build/cuda110/nompi/clang/cuda/release/static:
     USE_NAME: "cuda110-nompi-clang-${CI_PIPELINE_ID}"
     KEEP_CONTAINER: "ON"
     USE_SLURM: 0
-
-test/cuda110/nompi/clang/cuda/release/static:
-  extends:
-    - .horeka_test_template
-    - .default_variables
-    - .full_test_condition
-    - .use_gko-cuda110-mvapich2-gnu9-llvm9-intel2020
-  variables:
-    USE_NAME: "cuda110-nompi-clang-${CI_PIPELINE_ID}"
-    SLURM_PARTITION: "accelerated"
-    SLURM_GRES: "gpu:1"
-    SLURM_TIME: "01:30:00"
-  dependencies: null
-  needs: [ "build/cuda110/nompi/clang/cuda/release/static" ]
+    TEST_SLURM_GRES: "gpu:1"
 
 
 build/cuda110/nompi/intel/cuda/debug/static:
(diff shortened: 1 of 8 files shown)


upsj (approved):
approving as a quick workaround, but we should revert this after we have figured out a more stable setup

.gitlab-ci.yml
```diff
@@ -359,6 +368,7 @@ build/cuda110/nompi/intel/cuda/debug/static:
     USE_NAME: "cuda110-nompi-intel-${CI_PIPELINE_ID}"
     KEEP_CONTAINER: "ON"
     USE_SLURM: 0
+  needs: [ "check/horeka" ]
```
pratikvn, line 371: I think you need to do this for all the horeka jobs below as well ?

yhmtsai:
Actually, a unified build is not too hard because gitlab provides CI_RUNNER_TAGS.
Thus, we can select the slurm script when CI_RUNNER_TAGS contains horeka or slurm or normal scrip otherwise.
However, it will need to merge the build and test on horeka.
@tcojean 
```
KEEP_CONTAINER: "ON"
USE_SLURM: 0
```
are only used for build
```
SLURM_PARTITION: "accelerated"
SLURM_GRES: "gpu:1"
SLURM_TIME: "02:00:00"
```
are only used for tests.
They will not infect each other, right?

.gitlab-ci.yml
```diff
@@ -277,7 +277,7 @@ build/cuda102/nompi/intel/cuda/debug/static:
 # cuda 11.0 and friends on HoreKa with tests
```
pratikvn, line 277: Maybe note here that the build and tests run in the same step.

pratikvn (commented):
Currently you seem to run a step that generates the optional yml file, fills it with the correct variables and then in another step depending on the tag add that file to the job.yml file in unified_build_template.

Would it not be better to have the variables from the job itself ? I dont see a reason why the additional complexity of first filling a .yml file and then substituting the variables with a script is necessary ?

yhmtsai:
> Would it not be better to have the variables from the job itself ? I dont see a reason why the additional complexity of first filling a .yml file and then substituting the variables with a script is necessary ?

Could you elaborate on it?

pratikvn:
If I understand what you are doing, in the `test/generate_test_pipeline`, you generare the yml files based on whether horeka is available or not and filling in the variables. I am wondering if it possible to have the yml files generated beforehand (one for horeka and one for the amdci) and just directly run the test jobs without having the `generate_test_pipeline`. This would also avoid the varsub stuff in the unified_template ?

pratikvn:
My point is mainly that we know beforehand the variables we need. Instead of the job which generates the yml file filling in the variables, we can just select the one for the appropriate machine that we already have prepared in the repo. 

But I guess this is more of my preference, than any concern, so I am happy with your current approach.

pratikvn:
I see. Then let's just leave it as it is. Thanks for the clarifications.

pratikvn (approved):
LGTM!

tcojean:
> Actually, a unified build is not too hard because gitlab provides CI_RUNNER_TAGS. Thus, we can select the slurm script when CI_RUNNER_TAGS contains horeka or slurm or normal scrip otherwise. However, it will need to merge the build and test on horeka. @tcojean
> 
> ```
> KEEP_CONTAINER: "ON"
> USE_SLURM: 0
> ```
> 
> are only used for build
> 
> ```
> SLURM_PARTITION: "accelerated"
> SLURM_GRES: "gpu:1"
> SLURM_TIME: "02:00:00"
> ```
> 
> are only used for tests. They will not infect each other, right?

To answer this, if you merge build and test, that means you want to run both on a compute node? In that case, you will need only the second part (the one for test). The `USE_SLURM: 0` is obviously to run the build on the login node. The `KEEP_CONTAINER: ON` is to keep the container in storage so that we reuse the built image in the test.

upsj:
I fully agree with merging build and test jobs, and if we use a full compute node, the builds also become blazingly fast :sparkles: That also saves us the complexity of interdependent jobs

yhmtsai:
@tcojean Yes, it is easy for merging them.
However, the bill or limitation of computing resources is the issue to avoid the merging way when I ask for this usage.
That's why I go the generated pipeline in the end.

MarcelKoch:
Do we want to keep working on this, or can this PR be closed?

MarcelKoch:
Asking again, will this be updated, or should it be closed? If there is no comment in 1 week, I will close this PR.

upsj:
IMO we should be aiming to make our pipelines more uniform so they work across different gitlab runner executors and hardware platforms. My current plans for the CI setup goes in the same direction, where this change would become superfluous.

yhmtsai:
If we do not have limit on CI resources from the slurm system, unified build is easy and good as the second point I mentioned.
Unfortunately, we have limited resources in the project for our CI pipeline, and I do not find a better way to handle two/one stage for build uniformly.

upsj:
We have so far used only 24% of our CPU time and 80% of our GPU time - CPU time is not our limiting factor here, so it doesn't hurt to simplify our CI pipelines.

yhmtsai:
It is considered from the build time and test time.
From the current develop pipeline https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/pipelines/986527225
with mpi, build: 27min, test: 58 min -> require 46% more
clang without mpi, build: 33min, test: 8 min (why is it so fast?) -> require 400% more
intel without mpi, build: 43min, test: 70min -> require 61% more
It may use less compile time when we build it in node but it may also use less time for testing with ctest_resource.
From the number you gave, it requires 30% more (CPU time for build and GPU time for testing I assume? although the summation is not 100%)

upsj:
I don't think these numbers provide any useful information - they are for shared builds on a login node with something like 8 threads, while a full node has > 100 cores. Together with ctest resources, I expect this to cut down even our slowest build/test executions to < 30 min, release builds probably even < 10 min. On top of Horeka, we also have FTP resources, which don't cut into our hours at all.

yhmtsai:
I know it's not the real case for combining build/test on computing nodes, but it is the only real number we have for guessing the situation. As I said, it may reduce build time on compute node but also reduce testing time with ctest_resource.  I do not have access to the project on horeka I do not know what it is the real number on that.

PR #1328, opened 2023-04-26
Wrap SuiteSparse AMD

upsj:
This adds some files from [SuiteSparse](https://github.com/DrTimothyAldenDavis/SuiteSparse) implementing the symmetric AMD reordering algorithm as a stop-gap measure until we have our own AMD implementation.
It also implements a fallback Csr sorting algorithm for CUDA/HIP 64 bit indices using Thrust to make a handful of tests pass.


upsj:
format-rebase!

core/test/reorder/amd.cpp
```diff
@@ -0,0 +1,276 @@
...
+    gko::factorization::symbolic_cholesky(this->mtx.get(), true, factorized_mtx,
+                                          forest);
+    gko::factorization::symbolic_cholesky(permuted_mtx.get(), true,
+                                          factorized_permuted_mtx, forest);
+    int fillin_mtx = factorized_mtx->get_num_stored_elements() -
+                     this->mtx->get_num_stored_elements();
+    int fillin_permuted = factorized_permuted_mtx->get_num_stored_elements() -
+                          permuted_mtx->get_num_stored_elements();
+    ASSERT_LE(fillin_permuted, fillin_mtx * 2 / 5);
+}
```
fritzgoebel, line 276:
```suggestion
}


}  // namespace
```

core/test/reorder/amd.cpp
```diff
@@ -0,0 +1,276 @@
...
+#include <ginkgo/core/matrix/csr.hpp>
+
+
+#include "core/factorization/symbolic.hpp"
+#include "core/test/utils.hpp"
+#include "core/test/utils/assertions.hpp"
+#include "core/test/utils/unsort_matrix.hpp"
+#include "matrices/config.hpp"
+
+
```
fritzgoebel, line 53:
```suggestion

namespace {


```

core/test/reorder/amd.cpp
```diff
@@ -0,0 +1,276 @@
...
+    gko::size_type num_rows;
+    gko::array<index_type> permutation_ref;
+    std::shared_ptr<matrix_type> mtx;
+    std::shared_ptr<gko::experimental::reorder::Amd<index_type>> amd;
+};
+
+TYPED_TEST_SUITE(Amd, gko::test::ValueIndexTypes, PairTypenameNameGenerator);
+
+
+TYPED_TEST(Amd, WorksAndReducesFillIn)
```
yhmtsai, line 216: Aren't these tests suitable for reference test?

core/test/reorder/amd.cpp
```diff
@@ -0,0 +1,276 @@
...
+            gko::factorization::symbolic_cholesky(this->mtx.get(), true,
+                                                  factorized_mtx, forest);
+            gko::factorization::symbolic_cholesky(
+                permuted_mtx.get(), true, factorized_permuted_mtx, forest);
+            int fillin_mtx = factorized_mtx->get_num_stored_elements() -
+                             this->mtx->get_num_stored_elements();
+            int fillin_permuted =
+                factorized_permuted_mtx->get_num_stored_elements() -
+                permuted_mtx->get_num_stored_elements();
+            ASSERT_LE(fillin_permuted, fillin_mtx - this->fillin_reduction);
```
yhmtsai, line 243: Is `fillin_reduction` also an answer here?

core/test/reorder/amd.cpp
```diff
@@ -0,0 +1,276 @@
...
+    std::unique_ptr<matrix_type> factorized_permuted_mtx;
+    gko::factorization::symbolic_cholesky(this->mtx.get(), true, factorized_mtx,
+                                          forest);
+    gko::factorization::symbolic_cholesky(permuted_mtx.get(), true,
+                                          factorized_permuted_mtx, forest);
+    int fillin_mtx = factorized_mtx->get_num_stored_elements() -
+                     this->mtx->get_num_stored_elements();
+    int fillin_permuted = factorized_permuted_mtx->get_num_stored_elements() -
+                          permuted_mtx->get_num_stored_elements();
+    ASSERT_LE(fillin_permuted, fillin_mtx * 2 / 5);
```
yhmtsai, line 275: where's 2/5 from?

core/reorder/amd.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+    array<IndexType> permutation{host_exec, num_rows};
+    array<IndexType> row_lengths{host_exec, num_rows};
+    for (size_type row = 0; row < num_rows; row++) {
+        row_lengths.get_data()[row] =
+            row_ptrs.get_data()[row + 1] - row_ptrs.get_data()[row];
+    }
+    // we use this much space for the column index workspace, the rest for
+    // row workspace
+    const auto col_idxs_plus_workspace_size = nnz + nnz / 5 + 2 * num_rows;
```
yhmtsai, line 194: maybe move this before allocating `col_idxs_plus_workspace`, with the size of (col_idxs_plus_workspace_size + 6 * num_rows)

yhmtsai: It only fits the suggestion of `iwlen : 1.2 * pfree +n` , right?

yhmtsai (commented):
do we need to extract the amd reordering file into our repo?

yhmtsai (commented):
do we need to extract the amd reordering file into our repo?

core/test/reorder/amd.cpp
upsj: the reordering leads to a 60% fill-in reduction, I wanted to avoid hard-coding the fill-in

upsj:
IMO Ginkgo already has a lot of different build configurations to keep track of, and with AMD being rather small, and to still enable building Ginkgo without internet access, I would prefer building the code every time. In the long term, we will probably replace it by our own AMD implementation

yhmtsai:
I think building ginkgo without internet access is a different thing and should use a different way from putting all codes in this repo. For example, when we release or merge pr into develop, there's an action or compressed files for the installation without internet access except for those GPL packages. 
Putting code directly in our repo is hard to track the package version.

From the license in the third party, some packages contain the GPL license. Will the license file be also under GPL?

upsj:
@yhmtsai all our dependencies are licensed under BSD 3-clause. The AMD implementation in Suitesparse has proper version numbers, and has only changed superficially since 2013, so I believe we don't need to worry about having to track upstream changes. Vendoring CAS was also done to make builds of libginkgo (excluding tests etc) possible from airgapped systems. I don't think we should give up that property for effectively 3 source code files from a large repository.

Technically a license file itself is under copyright owned by the author, but since all licenses listed in there require reproduction of the license text in derivative works, we are good. We could instead only include the AMD/Doc/LICENSE.txt though, which would slim down the PR a bit.

yhmtsai (approved):
as the decision in the meeting. LGTM

core/test/reorder/amd.cpp
upsj: I'm not using an unnamed namespace around the tests here

core/reorder/amd.cpp
```diff
@@ -0,0 +1,223 @@
...
+            auto scalar = initialize<real_scalar>({one<double>()}, exec);
+            auto id = real_identity::create(exec, conv_csr->get_size()[0]);
+            // compute A^T + A
+            conv_csr->transpose()->apply(scalar, id, scalar, conv_csr);
+        }
+        d_nnz = conv_csr->get_num_stored_elements();
+        d_row_ptrs = conv_csr->get_row_ptrs();
+        d_col_idxs = conv_csr->get_col_idxs();
+        converted = std::move(conv_csr);
+    }
```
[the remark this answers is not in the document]
upsj: true, that reduces the memory requirements slightly

fritzgoebel (approved):
LGTM

core/test/reorder/amd.cpp
upsj: just the specific test case

PR #1329, opened 2023-04-27
Relax METIS requirements for shared builds and update version

upsj:
From the issues @nbeams reported, it would be easier if we searched for METIS only where absolutely required (static builds), this reduces the likelihood we accidentally call another FindMETIS.cmake module. Additionally, this bumps the CMake library version to make it possible to identify post-release Ginkgo.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,7 @@ if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.18)
     cmake_policy(SET CMP0104 OLD)
 endif()
 
-project(Ginkgo LANGUAGES C CXX VERSION 1.5.0 DESCRIPTION "A numerical linear algebra library targeting many-core architectures")
+project(Ginkgo LANGUAGES C CXX VERSION 1.6.0 DESCRIPTION "A numerical linear algebra library targeting many-core architectures")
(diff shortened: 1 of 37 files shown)


tcojean (approved):
LGTM

nbeams:
When I configure with CMake on this branch, if METIS is found (which it is at the moment, because I have the module loaded), I see this in GinkgoTargest.cmake:
```
INTERFACE_LINK_LIBRARIES "Ginkgo::ginkgo_device;Ginkgo::ginkgo_omp;Ginkgo::ginkgo_cuda;Ginkgo::ginkgo_reference;Ginkgo::ginkgo_hip;Ginkgo::ginkgo_dpcpp;METIS::METIS"
```

If I'm building serial MFEM (so, no METIS), won't this still cause the same problem, because it will try to link to `lMETIS::METIS` since it doesn't look for METIS itself to import the target?

nbeams:
It's shared.

upsj:
some of the dependencies are PUBLIC when they don't necessarily need to be, I'll check if we can change that

pratikvn (approved):

nbeams (approved):
I tested the latest version of the branch, and it fixes the issue I was seeing when building serial MFEM with Ginkgo support. 🎉

Issue #1191, opened 2022-11-09
problematic `reinterpret_cast` in truncated float types


upsj:
```cpp
GKO_ATTRIBUTES operator float_type() const noexcept
{
    const auto bits = static_cast<full_bits_type>(data_)
                      << component_position;
    return reinterpret_cast<const float_type&>(bits);
}
```

This is technically UB, but it seems like no compilers actually rely on the type information here, so we're in the clear ATM. A safer alternative might be using `union` for type punning, since this is allowed by gcc (and other compilers?) explicitly. `using std::memcpy; memcpy(...)` might be the safest, as long as it doesn't need to be `constexpr`, since this is explicitly supported in CUDA.

PR #1331, opened 2023-05-04
Add NVHPC CI and fix issues

tcojean:
This is no longer a WIP, but issues remain:

- [x] ~~Investigate `matrix_generator_test` likely miscompilation~~ I just disabled the offending tests, if somebody wants to investigate, check why Release builds before `disable miscompiling matrix_generator tests` fail.
- [x] Investigate `jacobi_kernels` adaptive/reduced precision issues

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -84,6 +84,22 @@ trigger_pipeline:
       fi
 
 
+# Build jobs
+# Job with example runs.
+# cuda 9.2 and friends
+build/cuda92/nompi/gcc/all/release/shared:
+  extends:
+    - .build_and_test_template
+    - .default_variables
+    - .quick_test_condition
+    - .use_gko-cuda92-mvapich2-gnu7-llvm50-intel2017
+  variables:
+    BUILD_OMP: "ON"
+    BUILD_CUDA: "ON"
+    BUILD_HIP: "ON"
+    BUILD_TYPE: "Release"
+    CUDA_ARCH: 61
+
 # cuda 10.1 and friends
 # Build CUDA NVIDIA without omp
 # Make sure that our jobs run when HWLOC is
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -84,89 +84,40 @@ trigger_pipeline:
       fi
 
 
-# Build jobs
-# Job with example runs.
-# cuda 9.2 and friends
-build/cuda92/nompi/gcc/all/release/shared:
-  extends:
-    - .build_and_test_template
-    - .default_variables
-    - .quick_test_condition
-    - .use_gko-cuda92-mvapich2-gnu7-llvm50-intel2017
-  variables:
-    BUILD_OMP: "ON"
-    BUILD_CUDA: "ON"
-    BUILD_HIP: "ON"
-    BUILD_TYPE: "Release"
-    RUN_EXAMPLES: "ON"
-    CUDA_ARCH: 61
-
-# cuda 10.0 and friends
-# Make sure that our jobs run when using self-installed
-# third-party HWLOC.
-build/cuda100/mvapich2/gcc/all/debug/shared:
-  extends:
-    - .build_template
-    - .default_variables
-    - .quick_test_condition
-    - .use_gko-cuda100-mvapich2-gnu7-llvm60-intel2018
-  variables:
-    BUILD_MPI: "ON"
-    BUILD_OMP: "ON"
-    BUILD_CUDA: "ON"
-    BUILD_HIP: "ON"
-    BUILD_TYPE: "Debug"
-    FAST_TESTS: "ON"
-    CUDA_ARCH: 35
-
+# cuda 10.1 and friends
+# Build CUDA NVIDIA without omp
 # Make sure that our jobs run when HWLOC is
 # forcibly switched off
-build/cuda100/nompi/clang/all/release/static:
-  extends:
-    - .build_template
-    - .default_variables
-    - .full_test_condition
-    - .use_gko-cuda100-mvapich2-gnu7-llvm60-intel2018
-  variables:
-    C_COMPILER: "clang"
-    CXX_COMPILER: "clang++"
-    BUILD_OMP: "ON"
-    BUILD_CUDA: "ON"
-    BUILD_HIP: "ON"
-    BUILD_HWLOC: "OFF"
-    BUILD_TYPE: "Release"
-    BUILD_SHARED_LIBS: "OFF"
-    CUDA_ARCH: 35
-
-# Build CUDA NVIDIA without omp
-build/cuda100/nompi/clang/cuda_wo_omp/release/shared:
-  extends:
-    - .build_template
-    - .default_variables
-    - .full_test_condition
-    - .use_gko-cuda100-mvapich2-gnu7-llvm60-intel2018
-  variables:
-    C_COMPILER: "clang"
-    CXX_COMPILER: "clang++"
-    BUILD_CUDA: "ON"
-    BUILD_HIP: "ON"
-    BUILD_HWLOC: "OFF"
-    BUILD_TYPE: "Release"
-    CUDA_ARCH: 35
-
-# cuda 10.1 and friends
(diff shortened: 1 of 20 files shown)


cuda/test/solver/upper_trs_kernels.cpp
```diff
@@ -174,21 +174,31 @@ TEST_F(UpperTrs, CudaMultipleRhsApplySyncfreeIsEquivalentToRef)
...
+#else
+    // The cuSPARSE 12 solver is even worse: It only works if the stride is
+    // equal to the number of columns.
+    auto in_stride = 3;
+    auto out_stride = 3;
+#endif
+#else
+    auto in_stride = 4;
+    auto out_stride = 5;
+#endif
```
pratikvn, line 192: This is a problem where the cusparse spsm is called, so maybe we should be handling it there instead?

include/ginkgo/core/base/std_extensions.hpp
```diff
@@ -69,6 +70,16 @@ template <typename... Ts>
...
 
 
+// Disable deprecation warnings when using standard > C++14
+inline bool uncaught_exception() noexcept {
+#if __cplusplus > 201402L
+    return std::uncaught_exceptions() > 0;
+#else
+    return std::uncaught_exception();
+#endif
+}
```
pratikvn, line 80: Is this strictly necessary  ? I dont think we should have this in the public header either way. Might have unexpected consequences for integrations.

cuda/test/solver/upper_trs_kernels.cpp
upsj: you mean throw an exception?

pratikvn: Yes. I think a thrown exception with some nice message is better than some weird unreadable error, which cusparse exits with

upsj: We don't throw an error right now, the results are just incorrect.

cuda/test/solver/lower_trs_kernels.cpp
```diff
@@ -174,21 +174,31 @@ TEST_F(LowerTrs, CudaMultipleRhsApplySyncfreeIsEquivalentToRef)
...
+#if CUDA_VERSION < 12000
+    // The cuSPARSE Generic SpSM implementation uses the wrong stride here
+    // so the input and output stride need to match
+    auto in_stride = 4;
+    auto out_stride = 4;
+#else
+    // The cuSPARSE 12 solver is even worse: It only works if the stride is
+    // equal to the number of columns.
+    auto in_stride = 3;
+    auto out_stride = 3;
```
tcojean, line 187: Is this with 12 only? I thought I also had to pick stride = nrhs for 11.7, but maybe I misremember. It's puzzling this decrease in stride support over time.

upsj: Might be that they introduced the new algorithm at 11.7 already, I only remember hearing there were some changes around 12.

include/ginkgo/core/base/std_extensions.hpp
tcojean: What would be the consequences? I added this because the version we use of `std::uncaught_exceptions` got deprecated in C++17, therefore rising many errors when compiling with an higher standard (which nvhpc does). Here, I select the new int version with high standards and keep the same one with lower standards. 
See: https://en.cppreference.com/w/cpp/error/uncaught_exception

pratikvn: I see. Thanks for the clarification. I misunderstood this usage.

cuda/test/solver/upper_trs_kernels.cpp
pratikvn: Maybe we should throw an error ?

omp/solver/idr_kernels.cpp
```diff
@@ -191,10 +195,13 @@ void initialize(std::shared_ptr<const OmpExecutor> exec, const size_type nrhs,
...
-        for (size_type j = 0; j < num_cols; j++) {
-            norm += squared_norm(subspace_vectors->at(row, j));
-        }
+        run_kernel_reduction(
+            exec,
+            [](auto col, auto row, auto subspace_vectors) {
+                return squared_norm(subspace_vectors(row, col));
+            },
+            GKO_KERNEL_REDUCE_SUM(remove_complex<ValueType>), &norm, num_cols,
+            static_cast<int64>(row), subspace_vectors);
```
yhmtsai, line 204: Is it for fix or optimization?

.gitlab-ci.yml
```diff
@@ -84,67 +84,16 @@ trigger_pipeline:
       fi
 
 
-# Build jobs
-# Job with example runs.
-# cuda 9.2 and friends
-build/cuda92/nompi/gcc/all/release/shared:
```
yhmtsai, line 90: I will suggest to bump the version after release.

.gitlab/image.yml
```diff
@@ -69,6 +57,18 @@
     - private_ci
     - nvidia-gpu
 
+.use_gko_nvhpc233-cuda120-openmpi-gnu12-llvm16:
+  image: ginkgohub/nvhpc:233-cuda120-openmpi-gnu12-llvm16
+  tags:
+    - private_ci
+    - nvidia-gpu
```
yhmtsai, line 64: also add the oneapi:2022.1

omp/solver/idr_kernels.cpp
upsj: It's a first step towards unifying it, but also a workaround for NVHPC's limited `pragma omp reduction` feature set

yhmtsai: Do you have any specific feature name for the unsupported one and the corresponding error message?

yhmtsai (commented):
others LGTM. Some of CI tests are not passed, so I do not put my approval yet

omp/solver/idr_kernels.cpp
upsj: basically `pragma omp (parallel) for reduction` is supported, but `omp declare reduction` is not

tcojean:
format!

.gitlab-ci.yml
tcojean: We still have compatibility, but to make space for the new CUDA versions I remove the old ones.

tcojean:
This should now be ""working"" including for Jacobi. I am also trying the new containers at the same time with the fixed environment.

.gitlab-ci.yml
yhmtsai: After removing them, it's hard to track whether the new functionalities are available in the earliest version or not.
Keeping one job for 9.2 to ensure we still testing our minimal requirement.
same for dpcpp 2022.1

omp/solver/idr_kernels.cpp
yhmtsai: I see. because omp `run_kernel_reduction` does not use omp reduction (declare for complex), it works via `run_kernel_reduction`

yhmtsai (changes requested):
LGTM. only have comments on testing version - should keep at least one job for the minimal supported version like cuda 9.2 and dpcpp 2022.1

.gitlab-ci.yml
upsj: I agree with CUDA, but Intel's ecosystem is still so much of a moving target that I would rather stay at head

omp/CMakeLists.txt
```diff
@@ -44,6 +44,15 @@ target_sources(ginkgo_omp
 ginkgo_compile_features(ginkgo_omp)
 target_compile_definitions(ginkgo_omp PRIVATE GKO_COMPILING_OMP)
 
+# TODO FIXME: Currently nvhpc 22.7+ optimizations break the omp jacobi's custom
+# precision implementation (mantissa segmentation)
```
upsj, line 48: maybe related to #1191?

upsj (approved):
LGTM (though somebody else should review my changes)

PR #1333, opened 2023-05-05
Enable an ensemble build

upsj:
With these fixes, it's possible to build Ginkgo for all backends simultaneously on the `ginkgodev/unified` container image. Includes a few test fixes.

Corresponding CMake invocation: 

`cmake -GNinja -DCMAKE_CXX_COMPILER=dpcpp -DOpenMP_CXX_FLAGS=-qopenmp -DOpenMP_CXX_LIB_NAMES=libiomp5 -DOpenMP_libiomp5_LIBRARY=/opt/intel/oneapi/compiler/latest/linux/compiler/lib/intel64_lin/libiomp5.so ..`

TODO:
- [x] ~~maybe fix deprecation warnings~~
- [x] ~~add a CI job for this~~ that will happen in #1267


upsj:
rebase!

test/mpi/distributed/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
-ginkgo_create_common_and_reference_test(matrix MPI_SIZE 3)
-ginkgo_create_common_and_reference_test(vector MPI_SIZE 3)
+ginkgo_create_common_and_reference_test(matrix MPI_SIZE 3 DISABLE_EXECUTORS dpcpp)
+ginkgo_create_common_and_reference_test(vector MPI_SIZE 3 DISABLE_EXECUTORS dpcpp)
```
MarcelKoch, line 2: I already disabled tests which don't have their kernels implemented directly in the files, is that not enough anymore?

MarcelKoch (approved):
LGTM, except for the test folder structure.

test/mpi/distributed/solver/solver.cpp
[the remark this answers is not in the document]
upsj: agreed, I think we should be using `distributed` for kernel tests and `mpi` for runs using MPI

test/mpi/distributed/CMakeLists.txt
upsj: *something* broke, I don't remember what it was. I'll check tomorrow

thoasm (approved):
LGTM

test/matrix/fft_kernels.cpp
```diff
@@ -160,6 +163,9 @@ TYPED_TEST(Fft, Apply1DInverseIsEqualToReference)
 
 TYPED_TEST(Fft, ApplyStrided1DInverseIsEqualToReference)
 {
+#if defined(GKO_COMPILING_HIP) && GINKGO_HIP_PLATFORM_HCC
+    GTEST_SKIP() << "rocFFT has a bug related to strided 1D FFT";
+#endif
```
yhmtsai, line 168: is something related to this https://github.com/ROCmSoftwarePlatform/rocFFT/issues/353 ?
but it should pass all in the current CI setup

test/solver/idr_kernels.cpp
```diff
@@ -40,6 +40,11 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <gtest/gtest.h>
 
 
+#ifdef GKO_COMPILING_DPCPP
+#include <CL/sycl.hpp>
+#endif
```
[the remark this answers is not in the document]
upsj:
```cpp
if (exec->get_queue()->get_device().is_gpu()) {
    GTEST_SKIP() << "skip the test because oneMKL GEMM on gpu may give NaN "
                    "(under investigation)";
}
```

test/matrix/fft_kernels.cpp
upsj: No, this is a different bug (causing incorrect results, not an error) I didn't bother creating a reproducer for yet.

test/mpi/distributed/CMakeLists.txt
upsj: `matrix_dpcpp` fails because `build_starting_indices` is not implemented

MarcelKoch: If we want to keep at least part of the test, we could build the partition on `ref` instead of `exec`.

upsj: Maybe we can do that in a subsequent PR, since it requires more significant changes?

upsj:
rebase!

PR #1335, opened 2023-05-07
Improve FindVTune

upsj:
When `vtune` is available in ´PATH`, we can use the location to find the headers and library automatically.

### DIFF
--- a/cmake/Modules/FindVTune.cmake
+++ b/cmake/Modules/FindVTune.cmake
@@ -17,6 +17,9 @@
 #
 # This module will set the following variables in your project:
 #
+# ``VTune_EXECUTABLE``
+#   path to the vtune executable
+#
 # ``VTune_INCLUDE_DIRS``
 #   where to find ittnotify.h
 #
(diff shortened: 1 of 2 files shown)


pratikvn (approved):
LGTM!

cmake/Modules/FindVTune.cmake
```diff
@@ -26,19 +29,22 @@
...
+        find_library(VTune_LIBRARY NAMES ittnotify HINTS ${VTune_PATH}/lib64)
     else()
-        find_library(VTune_LIBRARY NAMES ittnotify HINTS ${VTUNE_PATH}/lib32)
+        find_library(VTune_LIBRARY NAMES ittnotify HINTS ${VTune_PATH}/lib32)
     endif()
 endif()
 
 include(FindPackageHandleStandardArgs)
-find_package_handle_standard_args(VTune REQUIRED_VARS VTune_LIBRARY VTune_INCLUDE_DIR)
+find_package_handle_standard_args(VTune REQUIRED_VARS VTune_EXECUTABLE VTune_LIBRARY VTune_INCLUDE_DIR)
```
yhmtsai, line 47: does `VTune_EXECUTABLE` need to be required vars?

yhmtsai (approved):
LGTM

cmake/Modules/FindVTune.cmake
upsj: It seems like Intel always bundles the executable with the library and header, so this seems cleanest to me. Without the executable being present, `VTune_PATH` would be incorrect, so I would need more complex logic for finding the library and header.

PR #1336, opened 2023-05-10
Factory improvements

### LINKED ISSUE #766: Simplify setup of factory parameters of factory type
Currently, setting up stopping criteria requires a lot of boilerplate code:

```cpp
gko::ResidualNormReduction<ValueType>::build().with_reduction_factor(...).on(exec)
```

In this example, there are two points that I believe could or should be eliminated completely, and one piece of syntactic sugar that could help:

1. Stopping criterion factories need to be given an executor, even though they can never exist on another executor than the solver itself. So I would like to have some way to omit the `on(exec)` part. This could be approached either by storing an incomplete object (factory parameter struct), deferring the actual factory generation to when the executor for the outer solver factory is set, or by removing the concept of a stopping criterion factory from the solver factory itself completely, only using it inside the solver. I guess the former would be preferable here, since it doesn't interfere with custom stopping criteria. A possible implementation would be to wrap a stopping criterion factory's parameter struct inside a
```cpp
std::function<std::shared_ptr<StopCriterionFactory>(std::shared_ptr<const Executor>)>
```
that gets called in the `on` function of the solver factory.
2. For a stopping criterion, it doesn't matter whether it is dealing with vectors of type double, float or complex double. It is only given a single norm value per rhs (double or float, though we could limit that to double, since these kernels don't really matter in terms of runtime) and basically stores the result of the comparison with a target tolerance in another variable. So by using the dynamic type of the input vectors, it might be possible to remove the value type parameter from all residual-based stopping criteria.
3. With the above simpifications, it should be possible to provide syntactic sugar like `stop::rel_residual(...)`, `stop::residual_reduction(...)` or `stop::implicit_rel_residual(...)` to turn the whole stopping criterion setup into a single function call each.

What do you think?

upsj:
* Make all iterative solver parameter classes based on the same structure
* Allow passing factory parameter instances instead of full factories for criteria, preconditioners etc.
* Remove now unnecessary `.on(...)` calls.
* More strict const-correctness requirements for `with_*` functions

TODO:
- [x] Test the old and new behavior
- [x] Test `with_logger` propagating to factory and product
- [x] Adapt `Multigrid`

Closes #766
Partially addresses #1162


yhmtsai:
TBH, I prefer the old macro way to do this job.
the proposed way contains many parts which need to be checked manually.
for example,
`std::shared_ptr<LinOpFactory> GKO_FACTORY_PARAMETER_SCALAR(factorization_factory, nullptr);`
we get the type, scalar/vector, parameter name (including with_function), the default value in a place
```
std::shared_ptr<const LinOpFactory> factorization_factory{};
parameters_type& with_factorization(
            std::shared_ptr<const LinOpFactory> factory)
        {
            this->factorization_generator =
                [factory](std::shared_ptr<const Executor>)
                -> std::shared_ptr<const LinOpFactory> { return factory; };
            return *this;
        }
// another part for accepting without `.on(exec)`
```
the overload for supporting with/without factory can be handled in macro, too.
I will also suggest the deprecated factory parameter in another pr.

MarcelKoch:
I think this needs to be rebased to remove the changes from #1337

include/ginkgo/core/solver/direct.hpp
```diff
@@ -86,8 +87,52 @@ class Direct : public EnableLinOp<Direct<ValueType, IndexType>>,
...
+        parameters_type& with_factorization(
+            FactorizationParameters factorization_parameters)
+        {
+            this->factorization_generator =
+                [factorization_parameters](std::shared_ptr<const Executor> exec)
+                -> std::shared_ptr<const LinOpFactory> {
+                return factorization_parameters.on(exec);
+            };
+            return *this;
+        }
```
yhmtsai, line 119: I think it is already handled by the deferred_factory_parameter constructor

yhmtsai (commented):
I think there are two macro that can represent these changes.
```
type GKO_DEFERRED_FACTORY_PARAMETER(name, default) 
type name{default};
const parameters_type& with_##name(deferred_factory_parameter<LinOpFactory> _value) const {
  this->_name##_generator_ = std::move(_value);
}
private:
deferred_factory_parameter<const LinOpFactory> _name##_generator_;
public:
```
and corresponding vector version.

yhmtsai:
I think the required and optional parameters should be checked by the LinOp self like current state.

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -1119,8 +1119,7 @@ public:                                                                      \
     mutable _name{__VA_ARGS__};                                              \
                                                                              \
     template <typename... Args>                                              \
-    auto with_##_name(Args&&... _value)                                      \
-        const->const std::decay_t<decltype(*this)>&                          \
+    auto with_##_name(Args&&... _value)->std::decay_t<decltype(*this)>&      \
```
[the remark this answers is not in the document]
upsj: I would argue that modifying a `const` object is a bug and we should treat it as such

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -859,6 +861,116 @@ class EnablePreconditionedIterativeSolver
...
+    /**
+     * Provides stopping criteria via stop::CriterionFactory instances to be
+     * used by the iterative solver in a fluent interface.
+     */
+    template <typename... Args>
+    Parameters& with_criteria(Args&&... value)
+    {
+        this->criterion_generators = {
+            deferred_factory_parameter<stop::CriterionFactory>{
+                std::forward<Args>(value)}...};
```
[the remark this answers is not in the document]
upsj: deferred_factory_parameter already adds `const`

upsj:
What I'm talking about can best be seen in the `Multigrid::factory_parameters::on` function: Different parameters need to be handled differently on whether they support `nullptr`

MarcelKoch (changes requested):
Most of my suggestions are optional. The one I would like to see addressed is the deferred factory constructor with nullptr.

include/ginkgo/core/base/abstract_factory.hpp
```diff
@@ -274,6 +274,268 @@ class enable_parameters_type {
...
+ * @tparam FactoryType  the type of factory that can be instantiated from this
+ * object.
+ */
+template <typename FactoryType>
+class deferred_factory_parameter {
+public:
+    deferred_factory_parameter() = default;
+
+    /** Creates an empty deferred factory parameter. */
+    deferred_factory_parameter(std::nullptr_t) {}
```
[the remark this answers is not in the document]
upsj: good idea!

include/ginkgo/core/preconditioner/ilu.hpp
```diff
@@ -131,25 +131,97 @@ class Ilu : public EnableLinOp<
...
+        std::shared_ptr<const LinOpFactory> factorization_factory{};
+
+        [[deprecated("use with_l_solver instead")]] parameters_type&
+        with_l_solver_factory(
+            deferred_factory_parameter<typename l_solver_type::Factory> solver)
+        {
+            return with_l_solver(std::move(solver));
+        }
+
+        parameters_type& with_l_solver(
```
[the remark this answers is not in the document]
upsj: I think this won't work entirely, since I am also renaming the factory parameter at the same time. The variable is still named _factory for interface-stability.

include/ginkgo/core/preconditioner/ilu.hpp
```diff
@@ -131,25 +131,97 @@ class Ilu : public EnableLinOp<
...
+
+        parameters_type& with_factorization(
+            deferred_factory_parameter<LinOpFactory> factorization)
+        {
+            this->factorization_generator = std::move(factorization);
+            return *this;
+        }
+
+        /**
+         *
```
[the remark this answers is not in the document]
upsj: .on can be removed here

include/ginkgo/core/base/abstract_factory.hpp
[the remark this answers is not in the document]
upsj: They can't be, because I am using this as a parameter type in all of the with_* functions.

include/ginkgo/core/solver/solver_base.hpp
```diff
@@ -859,6 +861,116 @@ class EnablePreconditionedIterativeSolver
 };
 
 
+struct iterative_solver_factory_parameters {
```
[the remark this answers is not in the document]
upsj: With the current design, all parameter types share a common type (iterative_solver_factory_parameters) that can be passed to non-templated functions doing something with it. In the pure mixin setting, every solver has its own specific parameter type with no relation between them. Not sure if this is worth it, but it seems like a useful property.

CMakeLists.txt
```diff
@@ -511,3 +511,8 @@ else()
     FILE(READ ${PROJECT_BINARY_DIR}/minimal.log GINKGO_LOG_SUMMARY)
 endif()
 MESSAGE(STATUS "${GINKGO_LOG_SUMMARY}")
+
+# make sure no build files get committed accidentally
+if(NOT EXISTS ${CMAKE_CURRENT_BINARY_DIR}/.gitignore)
+    file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/.gitignore "*")
```
MarcelKoch, line 517: won't that create a .gitignore in the binary dir that the user has to add to the top-level .gitignore?

include/ginkgo/core/base/abstract_factory.hpp
```diff
@@ -271,9 +278,336 @@ class enable_parameters_type {
      * Loggers to be attached to the factory and generated object.
      */
     std::vector<std::shared_ptr<const log::Logger>> loggers{};
+
+    std::unordered_map<std::string,
```
MarcelKoch, line 282: maybe a small doc string?

include/ginkgo/core/base/abstract_factory.hpp
```diff
@@ -271,9 +278,336 @@ class enable_parameters_type {
...
+ *
+ * @tparam FactoryType  the type of factory that can be instantiated from this
+ * object.
+ */
+template <typename FactoryType>
+class deferred_factory_parameter {
+public:
+    deferred_factory_parameter() = default;
+
+    /** Creates an empty deferred factory parameter. */
```
MarcelKoch, line 323: I think the doc is not up-to-date

MarcelKoch (changes requested):
only some comments on documentation left

CMakeLists.txt
upsj: I think gitignore is slightly more relaxed about dotfiles than bash globs, [the documentation](https://git-scm.com/docs/gitignore) states
> An asterisk "*" matches anything except a slash [...]

MarcelKoch (approved):

include/ginkgo/core/base/abstract_factory.hpp
```diff
@@ -271,9 +278,346 @@ class enable_parameters_type {
...
+                                              auto& params) {                \
+            if (!params._name##_generator_.empty()) {                        \
+                params._name.clear();                                        \
+                for (auto& generator : params._name##_generator_) {          \
+                    params._name.push_back(generator.on(exec));              \
+                }                                                            \
+            }                                                                \
+        };                                                                   \
+        return *this;                                                        \
+    }                                                                        \
```
pratikvn, line 592: nit:
```suggestion
    }                                                                        \
                                                                             \
```

core/test/solver/bicgstab.cpp
```diff
@@ -64,10 +64,9 @@ class Bicgstab : public ::testing::Test {
           bicgstab_factory(
               Solver::build()
                   .with_criteria(
-                      gko::stop::Iteration::build().with_max_iters(3u).on(exec),
+                      gko::stop::Iteration::build().with_max_iters(3u),
                       gko::stop::ResidualNorm<value_type>::build()
-                          .with_reduction_factor(gko::remove_complex<T>{1e-6})
-                          .on(exec))
+                          .with_reduction_factor(gko::remove_complex<T>{1e-6}))
```
pratikvn, line 69: Could you add a test to ensure that the previous behaviour works as well ? Otherwise, it breaks all existing code.

pratikvn (approved):
Mostly looks good to me. I am a bit concerned that the large number of changes might break interface, and we currently have no way of checking that as we update all the tests as well.

Maybe we should have something like interface stability tests, that ensure that the updated code does not break existing interface. Something similar to `test_install` would probably be sufficient and we dont need to make it elaborate.

core/test/solver/bicgstab.cpp
upsj: done

upsj:
format!

PR #1338, opened 2023-05-16
Reusable permutation and transpose

### LINKED ISSUE #1157: Reusable transpose
We need a way to update the transpose of a matrix with new values, but fixed sparsity pattern. Under the hood, this is nothing but a permutation on the values, but computing the permutation indices requires either `reinterpret_cast`ing an identity permutation of integers into `double` and calling the cuSPARSE routine (hoping it doesn't flip any bits) or using our own sorting-based fallback transpose kernel (which shouldn't be that bad, since cub RadixSort is blazingly fast)

upsj:
As a tool for implementing reusable factories, this adds reusable functionality for all Csr permutation and transpose functions.

It also takes a first step towards making `Permutation` the default representation of a reordering.

Closes #1157

### DIFF
--- a/core/matrix/csr.cpp
+++ b/core/matrix/csr.cpp
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: 2017 - 2024 The Ginkgo authors
+// SPDX-FileCopyrightText: 2017 - 2025 The Ginkgo authors
 //
 // SPDX-License-Identifier: BSD-3-Clause
 
@@ -26,6 +26,7 @@
 #include "core/components/absolute_array_kernels.hpp"
 #include "core/components/fill_array_kernels.hpp"
 #include "core/components/format_conversion_kernels.hpp"
+#include "core/components/precision_conversion_kernels.hpp"
 #include "core/components/prefix_sum_kernels.hpp"
 #include "core/matrix/csr_kernels.hpp"
 #include "core/matrix/ell_kernels.hpp"
@@ -48,6 +49,7 @@ GKO_REGISTER_OPERATION(spgeam, csr::spgeam);
 GKO_REGISTER_OPERATION(convert_idxs_to_ptrs, components::convert_idxs_to_ptrs);
 GKO_REGISTER_OPERATION(convert_ptrs_to_idxs, components::convert_ptrs_to_idxs);
 GKO_REGISTER_OPERATION(fill_in_dense, csr::fill_in_dense);
+GKO_REGISTER_OPERATION(fill_seq_array, components::fill_seq_array);
 GKO_REGISTER_OPERATION(compute_slice_sets, sellp::compute_slice_sets);
 GKO_REGISTER_OPERATION(convert_to_sellp, csr::convert_to_sellp);
 GKO_REGISTER_OPERATION(compute_max_row_nnz, ell::compute_max_row_nnz);
@@ -83,6 +85,7 @@ GKO_REGISTER_OPERATION(is_sorted_by_column_index,
                        csr::is_sorted_by_column_index);
 GKO_REGISTER_OPERATION(extract_diagonal, csr::extract_diagonal);
 GKO_REGISTER_OPERATION(fill_array, components::fill_array);
+GKO_REGISTER_OPERATION(convert_precision, components::convert_precision);
 GKO_REGISTER_OPERATION(prefix_sum_nonnegative,
                        components::prefix_sum_nonnegative);
 GKO_REGISTER_OPERATION(inplace_absolute_array,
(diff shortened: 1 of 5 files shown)


MarcelKoch:
Just a comment, but Csr now has 18 (!) function for permutation. IMO that is too much, at least the `array` ones should be deprecated, since, as far as I understood it, we would move to the `Permutation` interface.

include/ginkgo/core/matrix/csr.hpp
```diff
@@ -878,6 +966,18 @@ class Csr : public EnableLinOp<Csr<ValueType, IndexType>>,
         return values_.get_const_data();
     }
 
+    /**
+     * Creates a Dense view of the value array of this matrix as a column
+     * vector of dimensions nnz x 1.
+     */
+    std::unique_ptr<Dense<ValueType>> create_value_view();
```
MarcelKoch, line 973: maybe keep those as free functions in our core?

core/matrix/csr.cpp
```diff
@@ -618,6 +621,83 @@ void Csr<ValueType, IndexType>::write(mat_data& data) const
...
+    static_assert(alignof(FloatIndexType) == alignof(IndexType));
+    auto exec = input->get_executor();
+    auto in_size = input->get_size();
+    auto transformed = Csr<ValueType, IndexType>::create(exec, out_size, nnz);
+    // transform matrix with integer values from 0 to nnz - 1 reinterpret_cast
+    // as float
+    array<IndexType> iota_values{exec, nnz};
+    exec->run(csr::make_fill_seq_array(iota_values.get_data(), nnz));
+    auto iota_float_view = make_array_view(
+        exec, nnz, reinterpret_cast<FloatIndexType*>(iota_values.get_data()));
```
MarcelKoch, line 642: TBH I'm a bit skeptical about this. Isn't this UB?

MarcelKoch (changes requested):
In general LGTM, I would like to discuss the difference between returning both the reuse info with the transposed/permuted matrix and just the reuse info.
If the use case for the reuse info is to be used only once or twice, the current approach makes sense.
If, however, it should be applied in a loop (or similar) then I think returning only the reuse info should lead to more uniform control flow.

include/ginkgo/core/matrix/csr.hpp
```diff
@@ -754,6 +757,42 @@ class Csr : public EnableLinOp<Csr<ValueType, IndexType>>,
...
+     * original matrix.
+     * ```
+     * auto [transposed, reuse] = matrix->transpose_reuse();
+     * change_values(matrix);
+     * reuse->update_values(matrix, transposed);
+     * ```
+     * @return the reuse info struct that can be used to update values in the
+     *         transposed matrix.
+     */
+    std::pair<std::unique_ptr<Csr>, permuting_reuse_info> transpose_reuse()
```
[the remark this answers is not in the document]
upsj: doing the operation twice seems like an unnecessary pessimization - maybe alternatively we could somehow make the reuse object a reference parameter, but that makes initialization more complicated, and makes it easier to misuse (what if the existing reuse info has the correct size but incorrect data? We can't really recognize that). It might make sense to give the reuse info a default constructor or at least size-independent constructor though, to allow a pattern like
```
std::unique_ptr<Csr> transposed;
permuting_reuse_info reuse;
...
if(!transposed) {
std::tie(transposed, permuting_reuse_info) = std::move(mtx->transpose_reuse());
}
```

core/matrix/csr.cpp
upsj: Yes, but we are provoking the same kind of nasal demons elsewhere - we don't really care about object provenance for index and value types, because we allocate the data using a pointer allocated via an untyped `::operator new`. The point here is that the compiler can't do anything with this information, because we always pass it through the ABI boundary between the ginkgo and ginkgo_... libraries, only the kernels actually access these values through regular typed pointers instead of a memcpy.

MarcelKoch: Thanks for the clarification. If the bit patterns match, then I guess we can live with it.

include/ginkgo/core/matrix/csr.hpp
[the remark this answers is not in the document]
upsj: I made the type default-constructible now.

include/ginkgo/core/matrix/csr.hpp
```diff
@@ -766,7 +805,9 @@ class Csr : public EnableLinOp<Csr<ValueType, IndexType>>,
      *              If permute_mode::rows is set, the rows will be permuted.
      *              If permute_mode::columns is set, the columns will be
      *              permuted.
-     * @return  The permuted matrix.
+     * @return an std::pair consisting of the transposed matrix and the reuse
+     *         info struct that can be used to update values in the transposed
+     *         matrix.
```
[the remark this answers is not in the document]
upsj: fixed

include/ginkgo/core/matrix/csr.hpp
upsj: Where would be the right place to put them? I could create a new internal csr header, but maybe that's overkill?

include/ginkgo/core/matrix/csr.hpp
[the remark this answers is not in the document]
upsj: Also maybe to be more precise, my goal is for this to be used internally to implement reuse information for complex solvers, usage by applications should be restricted to advanced use cases

include/ginkgo/core/matrix/csr.hpp
MarcelKoch: I thought they were only used in `csr.cpp`, but now I see that they are also used in tests, so I'm fine with them as it is.

include/ginkgo/core/matrix/csr.hpp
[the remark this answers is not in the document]
MarcelKoch: That makes sense. Anyway, fine with me as it is.

core/matrix/csr.cpp
```diff
@@ -618,6 +621,92 @@ void Csr<ValueType, IndexType>::write(mat_data& data) const
...
+    return std::make_pair(
+        std::move(transformed),
+        typename Csr<ValueType, IndexType>::permuting_reuse_info{
+            std::move(transform_permutation)});
+}
+
+
+template <typename ValueType, typename IndexType>
+Csr<ValueType, IndexType>::permuting_reuse_info::permuting_reuse_info()
+    : permuting_reuse_info{nullptr}
```
MarcelKoch, line 675: Isn't this just the default constructor?

MarcelKoch (approved):

core/matrix/csr.cpp
upsj: this is calling the other permuting_reuse_info constructor, not the value_permutation constructor, if that's what you mean?

MarcelKoch: Oh, I see. But doesn't that still have the same effect as the default constructor?

upsj: yes, I just wanted to base everything on a single implementation of the constructor, no deep other reason behind it.

PR #1342, opened 2023-05-25
Add support for profiler and file input to benchmarks

upsj:
To simplify #1341, this pulls out the flag changes

### DIFF
--- a/benchmark/solver/solver_common.hpp
+++ b/benchmark/solver/solver_common.hpp
@@ -610,6 +610,17 @@ void run_solver_benchmarks(std::shared_ptr<gko::Executor> exec,
     }
 
     auto& allocator = test_cases.GetAllocator();
+    auto profiler_hook = create_profiler_hook(exec);
+    if (profiler_hook) {
+        exec->add_logger(profiler_hook);
+    }
+    auto annotate =
+        [profiler_hook](const char* name) -> gko::log::profiling_scope_guard {
+        if (profiler_hook) {
+            return profiler_hook->user_range(name);
+        }
+        return {};
+    };
 
     for (auto& test_case : test_cases.GetArray()) {
         try {
@@ -628,6 +639,12 @@ void run_solver_benchmarks(std::shared_ptr<gko::Executor> exec,
                        })) {
                 continue;
             }
+            // annotate the test case
+            // This string needs to outlive `test_case_range` to make sure we
+            // don't use its const char* c_str() after it was freed.
+            auto test_case_str = system_generator.describe_config(test_case);
+            auto test_case_range = annotate(test_case_str.c_str());
+
             if (do_print) {
                 std::clog << "Running test case: " << test_case << std::endl;
             }
@@ -660,16 +677,20 @@ void run_solver_benchmarks(std::shared_ptr<gko::Executor> exec,
                               allocator);
             auto precond_solver_name = begin(precond_solvers);
             for (const auto& solver_name : solvers) {
+                auto solver_range = annotate(solver_name.c_str());
                 for (const auto& precond_name : preconds) {
                     if (do_print) {
                         std::clog
                             << "\tRunning solver: " << *precond_solver_name
                             << std::endl;
                     }
-                    solve_system(solver_name, precond_name,
-                                 precond_solver_name->c_str(), exec, timer,
-                                 system_matrix, b.get(), x.get(), test_case,
-                                 allocator);
+                    {
+                        auto precond_range = annotate(precond_name.c_str());
+                        solve_system(solver_name, precond_name,
+                                     precond_solver_name->c_str(), exec, timer,
+                                     system_matrix, b.get(), x.get(), test_case,
+                                     allocator);
+                    }
                     if (do_print) {
                         backup_results(test_cases);
                     }
@@ -685,6 +706,9 @@ void run_solver_benchmarks(std::shared_ptr<gko::Executor> exec,
                 add_or_set_member(test_case, "error", msg_value, allocator);
             }
         }
+    }
+    if (profiler_hook) {
+        exec->remove_logger(profiler_hook);
     }
 }
 
--- a/benchmark/solver/solver_common.hpp
+++ b/benchmark/solver/solver_common.hpp
@@ -614,13 +614,7 @@ void run_solver_benchmarks(std::shared_ptr<gko::Executor> exec,
     if (profiler_hook) {
         exec->add_logger(profiler_hook);
     }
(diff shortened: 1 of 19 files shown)


benchmark/utils/general.hpp
```diff
@@ -93,6 +98,15 @@ DEFINE_bool(keep_errors, true,
...
 DEFINE_bool(nested_names, false, "If set, separately logs nested operations");
 
+DEFINE_bool(profile, false,
+            "If set, enables profiler mode: 1 repetition, 0 warmup "
+            "repetitions, profiler_hook=auto (if it is not otherwise set)");
+
+DEFINE_string(
+    profiler_hook, "none",
+    "Which profiler annotation mode to use, if any. Options are nvtx, roctx, "
+    "vtune, tau, debug, auto (choose based on executor).");
```
yhmtsai, line 108:
```suggestion
    "Which profiler annotation mode to use, if any. Options are none, nvtx, roctx, "
    "vtune, tau, debug, auto (choose based on executor).");
```

benchmark/utils/general.hpp
```diff
@@ -202,6 +248,26 @@ std::vector<std::string> split(const std::string& s, char delimiter = ',')
...
+            return nullptr;
+        }
+        if (input_str.back() == ']') {
+            return std::make_unique<std::stringstream>(input_str);
+        }
+        return std::make_unique<std::ifstream>(input_str);
+    }();
+    if (stream) {
+        return *stream;
+    }
```
yhmtsai, line 266: will the data exist when you only take the content of unique_ptr after the unique_ptr is destroyed?

benchmark/solver/solver_common.hpp
```diff
@@ -628,6 +639,12 @@ void run_solver_benchmarks(std::shared_ptr<gko::Executor> exec,
                        })) {
                 continue;
             }
+            // annotate the test case
+            // This string needs to outlive `test_case_range` to make sure we
+            // don't use its const char* c_str() after it was freed.
+            auto test_case_str = system_generator.describe_config(test_case);
```
yhmtsai, line 645: should we just use `const std::string` to keep the string data in the profiler?

benchmark/utils/general.hpp
upsj: the lifetime of the static object ends past the end of the application, so I think we can ignore that possibility.

benchmark/solver/solver_common.hpp
upsj: It's a bit of a balance between allocation overhead and convenience - using `std::string` would mean that the caller needs to allocate storage, while `const char*` works efficiently with string literals `"foo"`, but may require allocation otherwise, e.g. dynamically constructed strings. We could probably improve that overhead, e.g. in `ProfilerHook::on_apply_started`, by providing a small buffer on the stack that handles all but the most extreme cases.

benchmark/utils/general.hpp
upsj: good point!

yhmtsai (commented):
Should we put the annotation to include detail, warmup, and repetition?
IMO, we should only put the annotation (or another annotation to distinguish) for the repetition
Detail and warmup can be eliminated by parameters, but warmup might still show some benefit for profiling especially when we have cache for repeated apply.
Thus, we still profile it with some warmup.

upsj:
Sounds sensible, I'll add the annotations, but only add the respective annotations if warmup > 0 or repetitions > 0

upsj:
On second thought, I think this should best be handled once the refactoring in #1323 is done, because it requires passing around the same kind of annotation function in many more places, the duplication of which I would like to avoid to a degree.

benchmark/utils/general.hpp
```diff
@@ -93,6 +98,15 @@ DEFINE_bool(keep_errors, true,
 
 DEFINE_bool(nested_names, false, "If set, separately logs nested operations");
 
+DEFINE_bool(profile, false,
+            "If set, enables profiler mode: 1 repetition, 0 warmup "
```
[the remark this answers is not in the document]
MarcelKoch: Should we also set `detailed=false`?

MarcelKoch (approved):
LGTM. Only some minor suggestion. I would also favor using `std::string` instead of just `char*` for the annotation.

benchmark/sparse_blas/sparse_blas.cpp
```diff
@@ -180,8 +193,12 @@ int main(int argc, char* argv[])
             }
             auto& sp_blas_case = test_case[benchmark_name];
             std::clog << "Running test case: " << test_case << std::endl;
-            auto data =
-                DefaultSystemGenerator<>::generate_matrix_data(test_case);
+            // annotate the test case
+            // This string needs to outlive `test_case_range` to make sure we
+            // don't use its const char* c_str() after it was freed.
+            auto test_case_str = generator.describe_config(test_case);
+            auto test_case_range = annotate(test_case_str.c_str());
```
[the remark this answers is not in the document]
MarcelKoch: I'm fine either way, btw.

benchmark/solver/solver_common.hpp
upsj: as a compromise, how about we either add a `string_profiling_scope_guard` that also owns the corresponding string or return a struct containing both the string and the scope guard from `annotate`?

include/ginkgo/core/log/profiler_hook.hpp
```diff
@@ -450,7 +453,7 @@ class profiling_scope_guard {
 
     profiling_scope_guard& operator=(const profiling_scope_guard&) = delete;
 
-    profiling_scope_guard& operator=(profiling_scope_guard&&) = delete;
```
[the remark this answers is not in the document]
upsj: actually I don't think we need it at all, this was from a previous implementation where I had a plain `profiling_scope_guard` lying around instead of returning from two branches of `annotate`

benchmark/utils/general.hpp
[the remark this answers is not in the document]
upsj: good point, I changed it

benchmark/sparse_blas/sparse_blas.cpp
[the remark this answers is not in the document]
upsj: I moved it down, good point

benchmark/solver/solver_common.hpp
upsj: I went for the latter solution, since this seems specific to the benchmark use case for now.

PR #1344, opened 2023-06-06
Improve memory ordering of sync-free kernels

upsj:
This adds load_relaxed(_shared), load_acquire(_shared) and store_relaxed(_shared) and store_release(_shared) functions to provide limited atomic load/store support in NVIDIA GPUs.

TODO
- [x] Figure out deadlock on post-Volta


common/cuda_hip/components/syncfree.hpp.inc
```diff
@@ -93,48 +93,31 @@ public:
...
-            __threadfence();
-            return finished;
-        } else {
-            // peek at a global dependency
-            auto finished = load(global.status, dependency);
-            __threadfence();
-            return finished;
-        }
-    }
+    __device__ __forceinline__ bool peek(IndexType dependency) { return false; }
```
upsj, line 110: this needs to be reverted

thoasm (commented):
I would like to have some small documentation and explanation inside the `memory.cuh` file to clarify things.
Also, I feel like the names of the functions aren't accurate as they do different things depending on the architecture.
But I am a fan of these new functions!

cuda/components/memory.cuh
```diff
@@ -0,0 +1,789 @@
...
+// for reasoning behind this implementation
+#if (!defined(__clang__) && __CUDACC_VER_MAJOR__ >= 11)
+    return static_cast<uint32>(__cvta_generic_to_shared(ptr));
+#elif (!defined(__clang__) && CUDACC_VER_MAJOR__ == 10 && \
+       __CUDACC_VER_MINOR__ >= 2)
+    return __nvvm_get_smem_pointer(ptr);
+#else
+    uint32 smem_ptr;
+    asm("{{ .reg .u64 smem_ptr; cvta.to.shared.u64 smem_ptr, %1; cvt.u32.u64 "
+        "%0, smem_ptr; }}"
```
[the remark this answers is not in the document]
upsj: Python format strings require the use of double braces to not be interpreted as variable placeholders

cuda/components/memory.cuh
```diff
@@ -0,0 +1,789 @@
...
+    asm volatile("ld.volatile.shared.b32 %0, [%1];"
+                 : "=r"(result)
+                 : "r"(convert_generic_ptr_to_smem_ptr(ptr))
+                 : "memory");
+#else
+    asm volatile("ld.relaxed.cta.shared.b32 %0, [%1];"
+                 : "=r"(result)
+                 : "r"(convert_generic_ptr_to_smem_ptr(ptr))
+                 : "memory");
+#endif
```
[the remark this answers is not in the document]
upsj: To quote the PTX documentation: 
> [`ld.volatile` has the same memory synchronization semantics as `ld.relaxed.sys`.](https://docs.nvidia.com/cuda/archive/9.0/parallel-thread-execution/index.html#data-movement-and-conversion-instructions-ld)

This is equivalent to how libcu++ [implements its atomics](https://github.com/NVIDIA/libcudacxx/blob/a57dbed580e49b14ac1ad7f98496176407208aa4/include/cuda/std/detail/libcxx/include/support/atomic/atomic_cuda_generated.h#L12)

cuda/components/memory.cuh
```diff
@@ -0,0 +1,789 @@
...
+}
+
+
+__device__ __forceinline__ uint32 membar_acq_rel_shared()
+{
+#if __CUDA_ARCH__ < 700
+    asm volatile("membar.cta;" ::: "memory");
+#else
+    asm volatile("fence.acq_rel.cta;" ::: "memory");
+#endif
```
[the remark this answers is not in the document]
upsj: pre-Volta doesn't have relaxed/acq/rel memory semantics, see [this example](https://godbolt.org/z/dh8qEY1Tx)

dev_tools/scripts/generate_cuda_memory_ptx.py
```diff
@@ -0,0 +1,192 @@
+#!/usr/bin/env python3
+import os
+memory_spaces = [(".shared", ".cta", "_shared", "convert_generic_ptr_to_smem_ptr(ptr)", "r"), ("", ".gpu", "", "ptr", "l")]
+memory_orderings = [
+    (".relaxed", "_relaxed", ".relaxed", "_relaxed", True),
+    (".acquire", "_acquire", ".release", "_release", False)
+    ]
+sizes=[(".b32", "r", "int32", 4), (".b64", "l", "int64", 8), (".f32", "f", "float", 4), (".f64", "d", "double", 8)]
```
MarcelKoch, line 8: Maybe using dataclasses could make it a bit clearer what the tuple elements are for:
```suggestion
import dataclasses


@dataclass
class size:
    suffix: str
    constraint: str,
    typename: str
    length: int


sizes=[size(".b32", "r", "int32", 4), size(".b64", "l", "int64", 8), size(".f32", "f", "float", 4), size(".f64", "d", "double", 8)]
# or
sizes=[size(suffix=".b32", constraint="r", typename="int32", length=4), ...]
```

cuda/components/memory.cuh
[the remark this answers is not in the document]
thoasm: My fear was actually that we might run into old CUDA versions that don't support `fence`, but since we officially only support CUDA 10.1+, this is fine.

cuda/components/memory.cuh
```diff
@@ -0,0 +1,789 @@
...
+
+#include "cuda/base/types.hpp"
+
+
+namespace gko {
+namespace kernels {
+namespace cuda {
+
+
+__device__ __forceinline__ uint32 convert_generic_ptr_to_smem_ptr(void* ptr)
```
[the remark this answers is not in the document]
thoasm: This is the only thing left from my review. I will approve it after comments are added.

cuda/components/memory.cuh
```diff
@@ -0,0 +1,796 @@
...
+#include "common/cuda_hip/components/memory.hpp.inc"
+
+
+__device__ __forceinline__ int32 load_relaxed_shared(const int32* ptr)
+{
+    int32 result;
+#if __CUDA_ARCH__ < 700
+    asm volatile("ld.volatile.shared.b32 %0, [%1];"
+                 : "=r"(result)
+                 : "r"(convert_generic_ptr_to_smem_ptr((void*)ptr))
```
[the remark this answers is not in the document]
upsj: technically I only need the const_cast, because previously the load functions were not const-correct, but this is shorter. Would `const_cast<void*>` suffice for you?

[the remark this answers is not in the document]
upsj: In the other non-shmem cases, I needed to actually use the typename in the `const_cast`, so this was just the easier solution. In assembly, there is no distinction between different pointer types anyways

thoasm (approved):
LGTM!

cuda/components/memory.cuh
```diff
@@ -0,0 +1,804 @@
...
+//     include/cutlass/arch/memory_sm75.h#L90
+// for reasoning behind this implementation
+#if (!defined(__clang__) && __CUDACC_VER_MAJOR__ >= 11)
+    return static_cast<uint32>(__cvta_generic_to_shared(ptr));
+#elif (!defined(__clang__) && CUDACC_VER_MAJOR__ == 10 && \
+       __CUDACC_VER_MINOR__ >= 2)
+    return __nvvm_get_smem_pointer(ptr);
+#else
+    uint32 smem_ptr;
+    asm("{{ .reg .u64 smem_ptr; cvta.to.shared.u64 smem_ptr, %1; cvt.u32.u64 "
```
yhmtsai, line 72: Maybe it is a stupid question: from https://docs.nvidia.com/cuda/inline-ptx-assembly/index.html#incorrect-optimization, `volatile` ensures it is not deleted or moved. I think the location of this ptx does not affect anything, but if it is deleted? or does the delete possibility only happen in combination(optimization) or no output?

cuda/components/memory.cuh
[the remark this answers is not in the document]
yhmtsai: the requirement (fense require sm>=70) is also listed in the end of https://docs.nvidia.com/cuda/archive/9.0/parallel-thread-execution/index.html#parallel-synchronization-and-communication-instructions-membar

cuda/components/memory.cuh
```diff
@@ -0,0 +1,804 @@
...
+#endif
+
+    return result;
+}
+
+
+__device__ __forceinline__ void store_relaxed_shared(int64* ptr, int64 result)
+{
+#if __CUDA_ARCH__ < 700
+    asm volatile("st.volatile.shared.b64 [%0], %1;" ::"r"(
```
yhmtsai, line 161:
```suggestion
    asm volatile("st.volatile.shared.s64 [%0], %1;" ::"r"(
```

cuda/components/memory.cuh
```diff
@@ -0,0 +1,804 @@
...
+#else
+    asm volatile("fence.acq_rel.cta;" ::: "memory");
+#endif
+}
+
+
+#include "common/cuda_hip/components/memory.hpp.inc"
+
+
+__device__ __forceinline__ int32 load_relaxed_shared(const int32* ptr)
```
yhmtsai, line 104: the following codes are repeated up to the type/shared and the corresponding PTX (except for complex)
maybe macros like LOAD_ACQUIRE(TYPE, PTX_TYPE) -> give load_* and load_*_shared, which may require moving CUDA_ARCH macro out of this kind of macro.

cuda/components/memory.cuh
```diff
@@ -0,0 +1,804 @@
...
+#endif
+
+    return result;
+}
+
+
+__device__ __forceinline__ void store_relaxed_shared(int32* ptr, int32 result)
+{
+#if __CUDA_ARCH__ < 700
+    asm volatile("st.volatile.shared.b32 [%0], %1;" ::"r"(
```
yhmtsai, line 126:
```suggestion
    asm volatile("st.volatile.shared.s32 [%0], %1;" ::"r"(
```

cuda/components/memory.cuh
yhmtsai: ah, you did that in python. Isn't macro enough for that?

dev_tools/scripts/generate_cuda_memory_ptx.py
```diff
@@ -0,0 +1,241 @@
...
+          ptr_expr="convert_generic_ptr_to_smem_ptr({ptr})", ptr_constraint="r"),
+    space(ptx_space_suffix="", ptx_scope_suffix=".gpu", fn_suffix="", ptr_expr="{ptr}", ptr_constraint="l")]
+memory_orderings = [
+    ordering(ptx_load_suffix=".relaxed", fn_load_suffix="_relaxed",
+             ptx_store_suffix=".relaxed", fn_store_suffix="_relaxed", is_relaxed=True),
+    ordering(ptx_load_suffix=".acquire", fn_load_suffix="_acquire",
+             ptx_store_suffix=".release", fn_store_suffix="_release", is_relaxed=False)
+]
+types = [type_desc(ptx_type_suffix=".b32", val_constraint="r", name="int32"),
+         type_desc(ptx_type_suffix=".b64", val_constraint="l", name="int64"),
```
yhmtsai, line 41:
```suggestion
types = [type_desc(ptx_type_suffix=".s32", val_constraint="r", name="int32"),
         type_desc(ptx_type_suffix=".s64", val_constraint="l", name="int64"),
```

common/cuda_hip/components/syncfree.hpp.inc
```diff
@@ -93,18 +93,18 @@ public:
...
+                while (!load_acquire_shared(local.status + dep_local)) {
+                }
+            } else {
+                // wait for a global dependency
+                while (!load_acquire(global.status + dependency)) {
+                }
             }
         }
-        __threadfence();
+        group::tiled_partition<subwarp_size>(group::this_thread_block()).sync();
```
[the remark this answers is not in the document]
upsj: yes, since only a single lane is waiting for the data here, we need to make sure the other threads wait here as well. It might be necessary to keep a threadfence here as well, though IIRC syncwarp tends to do that implicitly.

cuda/components/memory.cuh
upsj: it can only be deleted if the optimizer manages to remove the dependency on the value the assembly computes. Since the following load/store is `volatile`, it cannot be optimized away.

cuda/components/memory.cuh
upsj: good catch, though it probably doesn't make a difference. this is what `libcudacxx` uses in their atomic loads/stores, they only use signed/unsigned types for arithmetic operations in atomics (+,-,max,min).

cuda/components/memory.cuh
upsj: I think a macro would be worse in terms of readability, since we are building strings out of many different components. Python allows us to at least give everything names in the template and parameter set (i.e. both where we define the macro and where we call it).

yhmtsai: macro should still give name for the parameter although you need to specify all instantiation manually not from for loop.
The python was an issue for me about the generated code and source.
I first review this file and try to figure out whether there's a missing combination.  Then figure out there's another python file for it. There's no strong connection between python and generated code especially when it is the final code not the intermediate state.
Could you at least add some comment about the code is generated by the python file?

common/cuda_hip/components/syncfree.hpp.inc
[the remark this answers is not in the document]
yhmtsai: the other threads also has the same dependency here, right?

yhmtsai (approved):

common/cuda_hip/components/syncfree.hpp.inc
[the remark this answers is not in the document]
upsj: They inherit the dependency from the lane 0 because they wait for it

upsj:
I'll go ahead and merge this already, since only DPC++ and OpenMP pipelines are outstanding, and those files were unmodified. Then we can move forward the other PRs soon.

PR #1345, opened 2023-06-08
Use shared storage for ELL and COO in Hybrid

upsj:
Currently the only thing preventing reusable reads working via int-to-float-to-int conversions like in #1338 is the fact that Hybrid uses two separate allocations to store the values for the ELL and COO parts. This PR uses a shared storage and views into it instead.

To fix the tests, I also had to make sure that they specify the correct input dimensions, because ELL and COO can no longer be reallocated separately.

### DIFF
--- a/core/matrix/hybrid.cpp
+++ b/core/matrix/hybrid.cpp
@@ -91,6 +91,11 @@ Hybrid<ValueType, IndexType>& Hybrid<ValueType, IndexType>::operator=(
     if (&other != this) {
         EnableLinOp<Hybrid>::operator=(other);
         auto exec = this->get_executor();
+        this->resize(other.get_size(),
+                     other.get_ell_num_stored_elements_per_row(),
+                     other.get_coo_num_stored_elements());
+        // need to make sure the COO and ELL storage is aliased to
+        // this Hybrid's storage.
         *coo_ = *other.coo_;
         *ell_ = *other.ell_;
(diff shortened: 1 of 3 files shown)


yhmtsai:
could you elaborate more on the issue from the separate storage issue in int-to-float-to-int conversions?

upsj:
I want to implement reusable read as a gather operation that gathers values from the input `device_matrix_data` (or just value array) into the matrix. For this to be possible, the `values` array needs to be contiguous, so I can implement this as
```cpp
auto [gather_idxs, ...] = matrix->read_reuse(matrix_data);
// ...
auto mtx_value_vector = matrix->create_value_view(); // returns Dense
value_vector->row_gather(gather_idxs, mtx_value_vector);
```
Depending on the actual interface we end up with, this may or may not be necessary. If we want to expose the value array as a single storage, we will need it though.

upsj:
On second thought, if we hide the read reuse behind a more generic interface, we might be able to get by without this change, because it can also be implemented with two separate `gather` calls in the current `Hybrid` implementation, and for distributed matrices, we can't use the same aliasing trick generically, so we would need to defer to the generic reuse functionality and at least two separate `gather` calls anyways.

upsj:
Shelving this, it seems like there are cleaner ways to enable reusable reads for Hybrid that also work for distributed matrices

Issue #570, opened 2020-06-15
Rename CMake options from `GINKGO_*` to `GKO_*`


upsj:
`GKO_` has a lot less potential for typos than `GINKGO_`, which is especially important when you have to set many options at configure-time. I would thus propose to replace the `GINKGO_` prefix by `GKO_` throughout all of our files, especially the CMake-related files.

@tcojean correctly pointed out that we currently keep these prefixes separate to avoid issues with CMake-configured files like config.hpp.in, but I believe we should be able to work around them.

PR #1350, opened 2023-06-14
add icpx support

yhmtsai:
This PR adds the icpx support.
add `-fsycl` when compiling any device function or queue operation and add the corresponding include path in all ginkgo.


dpcpp/CMakeLists.txt
```diff
@@ -82,8 +82,10 @@ configure_file(preconditioner/jacobi_common.hpp.in preconditioner/jacobi_common.
 ginkgo_compile_features(ginkgo_dpcpp)
 target_compile_definitions(ginkgo_dpcpp PRIVATE GKO_COMPILING_DPCPP _ONEDPL_COMPILE_KERNEL=0)
 
+set(GINKGO_DPCPP_FLAGS "-fsycl")
```
tcojean, line 85: Does this flag also work with the `dpcpp` compiler?

CMakeLists.txt
```diff
@@ -56,7 +56,7 @@ set(GINKGO_VERBOSE_LEVEL "1" CACHE STRING
 if(MSVC)
     set(GINKGO_COMPILER_FLAGS "" CACHE STRING
         "Set the required CXX compiler flags, mainly used for warnings. Current default is ``")
-elseif(GINKGO_BUILD_DPCPP OR CMAKE_CXX_COMPILER MATCHES "dpcpp")
```
tcojean, line 59: Maybe to prepare for more compilers in the future, we should add an option `GINKGO_BUILD_SYCL` and give `GINKGO_BUILD_DPCPP` the value of `GINKGO_BUILD_SYCL` by default.

dpcpp/CMakeLists.txt
tcojean: Also, we could try rewriting this this form:
https://github.com/oneapi-src/oneAPI-samples/pull/1721/files#diff-cbe919c8af7559493969dbc2c78bf849e49a594bd82fc018982f948266dbee24R1

Note: it requires a higher CMake version, also probably a high oneAPi version. I see IntelSYCLConfig provided in 2023.1. Before it was IntelDPCPP. See:
2023.0 instructions: https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2023-0/use-cmake-with-the-compiler.html
2023.1 instructions: https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2023-1/use-cmake-with-the-compiler.html

tcojean (commented):
Generally LGTM. Some suggestions inline.

dpcpp/CMakeLists.txt
yhmtsai: yes, dpcpp has the flag, too.
I went through IntelDpcpp.config. there's no `add_sycl_to_target` in this module.

CMakeLists.txt
yhmtsai: Maybe add them when we support other compilers not Intel's one?
Do we need to have `GINKGO_BUILD_DPCPP` and `GINKGO_BUILD_SYCL` at the same time?

upsj: I think we could use `GINKGO_BUILD_DPCPP` to initialize `GINKGO_BUILD_SYCL` if it is available, otherwise only use `BUILD_SYCL`

tcojean: Yes, that was exactly my idea. This way we don't break build scripts at the moment, while allowing a switch to `BUILD_SYCL`.

upsj: If I may suggest something, we could use a function like
```
gko_renamed_variable(GINKGO_BUILD_DPCPP GINKGO_BUILD_SYCL)
```
that expands to
```
if (EXISTS GINKGO_BUILD_DPCPP)
    if (EXISTS GINKGO_BUILD_SYCL)
        message(FATAL_ERROR "Both GINKGO_BUILD_SYCL and GINKGO_BUILD_DPCPP were specified, please remove the latter variable")
    else()
        message(WARNING "GINKGO_BUILD_DPCPP was deprecated, please use GINKGO_BUILD_SYCL instead")
        set(GINKGO_BUILD_SYCL ${GINKGO_BUILD_DPCPP})
    endif()
endif()
```
That would also allow us to implement things like #570 in the future

yhmtsai: I can try, but it is not really deprecated meaning.
after the first setting, setting DPCPP does not have the same functionality as previous one.

yhmtsai: done, I only provide the cache version because I think we do not need deprecation for internal variable.

cmake/rename.cmake
```diff
@@ -0,0 +1,25 @@
...
+        if(DEFINED ${actual})
+            message("actual ${actual} and deprecated ${deprecated}")
+            if("${${actual}}" STREQUAL "${${deprecated}}")
+                # They are the same, so only throw warning
+                message(WARNING "${deprecated} was deprecated, please only use ${actual} instead.")
+            else()
+                # They are different
+                set(${deprecated}_copy ${${deprecated}})
+                unset(${deprecated} CACHE)
+                message(FATAL_ERROR "Both ${deprecated} and ${actual} were specified, please use ${actual} instead.  "
```
upsj, line 13: This error message would vanish after rerunning CMake, is that intended? Maybe people might miss it under the large amount of output, and just rerun CMake, silently losing their configuration option.

cmake/rename.cmake
```diff
@@ -0,0 +1,25 @@
...
+                    "We remove ${deprecated}:${${deprecated}_copy} and keep ${actual}:${${actual}}")
+            endif()
+        else()
+            # Only set `deprecated`, move it to `actual`.
+            message(WARNING "${deprecated} was deprecated, please use ${actual} instead.  "
+                "We copy ${${deprecated}} to ${actual} and unset ${deprecated}.")
+            set(${actual} ${${deprecated}} CACHE ${type} "")
```
upsj, line 20: This way we have no description for the variable, maybe we can find a way to keep it?

cmake/sycl.cmake
```diff
@@ -0,0 +1,33 @@
+# IntelSYCL for dpcpp and icpx if the config is existed and cmake reaches the requirement
+if(CMAKE_CXX_COMPILER MATCHES "dpcpp|icpx")
+    if(CMAKE_HOST_WIN32 AND CMAKE_VERSION VERSION_GREATER_EQUAL 3.25)
+        find_package(IntelSYCL QUIET)
+    elseif(CMAKE_VERSION VERSION_GREATER_EQUAL 3.20.5)
```
upsj, line 5: I think we should require 3.20 for SYCL if that is necessary to make IntelLLVM work correctly.

cmake/sycl.cmake
```diff
@@ -0,0 +1,33 @@
+# IntelSYCL for dpcpp and icpx if the config is existed and cmake reaches the requirement
+if(CMAKE_CXX_COMPILER MATCHES "dpcpp|icpx")
+    if(CMAKE_HOST_WIN32 AND CMAKE_VERSION VERSION_GREATER_EQUAL 3.25)
+        find_package(IntelSYCL QUIET)
```
upsj, line 4: should this be optional?
```suggestion
        find_package(IntelSYCL QUIET REQUIRED)
```

upsj (approved):
LGTM!

tcojean (approved):
LGTM

cmake/rename.cmake
yhmtsai: I currently do not unset the deprecated variable.
It will keep showing the warning or error when the deprecated is there

cmake/rename.cmake
yhmtsai: can pass it thought the doc_string now

cmake/sycl.cmake
yhmtsai: no, it is optional. if there's no config and cmake, simply adding `-fsycl` in the following code is enough currently.

cmake/sycl.cmake
yhmtsai: yes it is optional. If there's no IntelSYCL, we have -fsycl for that.

PR #1354, opened 2023-06-23
Add reordering algorithms to sparse_blas benchmark

upsj:
We need a way to benchmark RCM, ND and AMD. I believe the straightforward solution is to use the `sparse_blas` benchmark, The operations happen exclusively on the CPU, so the executor choice has only minor impact on the actual algorithm, but it highlights the cost of the data movement involved, so it might still be useful to have executors involved here.

### DIFF
--- a/benchmark/sparse_blas/operations.cpp
+++ b/benchmark/sparse_blas/operations.cpp
@@ -695,11 +695,111 @@ private:
 };
 
 
+class ReorderRcmOperation : public BenchmarkOperation {
+    using reorder_type = gko::reorder::Rcm<etype, itype>;
+
+public:
+    explicit ReorderRcmOperation(const Mtx* mtx)
+        : mtx_{mtx->clone()},
+          factory_{reorder_type::build().on(mtx->get_executor())}
+    {}
+
+    std::pair<bool, double> validate() const override
+    {
+        // validating RCM correctness is hard, let's leave it out for now
+        return {true, 0.0};
+    }
+
+    gko::size_type get_flops() const override { return 0; }
+
+    gko::size_type get_memory() const override { return 0; }
+
+    void prepare() override {}
+
+    void run() override { reorder_ = factory_->generate(mtx_); }
+
+private:
+    std::shared_ptr<Mtx> mtx_;
+    std::unique_ptr<reorder_type::Factory> factory_;
+    std::unique_ptr<reorder_type> reorder_;
+};
+
+
+#if GKO_HAVE_METIS
+
+
+class ReorderNestedDissectionOperation : public BenchmarkOperation {
+    using factory_type =
+        gko::experimental::reorder::NestedDissection<etype, itype>;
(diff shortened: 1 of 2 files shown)


pratikvn (approved):
LGTM!

We should probably update our benchmarking documentation at some point. I think it is fairly out of date right now.

tcojean (approved):
LGTM.

Agree with Pratik on the documentation. People were already pointing this out to me several times. It's read more than I expected.

benchmark/sparse_blas/operations.cpp
```diff
@@ -695,11 +695,111 @@ class SymbolicCholeskyOperation : public BenchmarkOperation {
...
                std::function<std::unique_ptr<BenchmarkOperation>(const Mtx*)>>
-    operation_map{
-        {"spgemm",
-         [](const Mtx* mtx) { return std::make_unique<SpgemmOperation>(mtx); }},
+    operation_map
+{
+    {"spgemm",
+     [](const Mtx* mtx) { return std::make_unique<SpgemmOperation>(mtx); }},
         {"spgeam",
          [](const Mtx* mtx) { return std::make_unique<SpgeamOperation>(mtx); }},
```
yhmtsai, line 804: weird indentation

benchmark/sparse_blas/operations.cpp
```diff
@@ -726,9 +826,25 @@ const std::map<std::string,
...
+         }},
+        {"reorder_amd", [](const Mtx* mtx) {
+             return std::make_unique<ReorderApproxMinDegOperation>(mtx);
+         }},
+#if GKO_HAVE_METIS
+    {
+        "reorder_nd", [](const Mtx* mtx) {
+            return std::make_unique<ReorderNestedDissectionOperation>(mtx);
+        }
+    }
```
yhmtsai, line 845: same here

yhmtsai (approved):
nit on format

benchmark/sparse_blas/operations.cpp
```diff
@@ -726,9 +826,25 @@ const std::map<std::string,
...
              return std::make_unique<SymbolicCholeskyOperation>(mtx, true);
-         }}};
+         }},
+        {"reorder_rcm",
+         [](const Mtx* mtx) {
+             return std::make_unique<ReorderRcmOperation>(mtx);
+         }},
+        {"reorder_amd", [](const Mtx* mtx) {
+             return std::make_unique<ReorderApproxMinDegOperation>(mtx);
+         }},
```
yhmtsai, line 839: I am surprised at c++ allows `,` in the end without anything

upsj: That's allowed in many languages to make diffs in lists nicer. But I changed it, so this is no longer relevant.

benchmark/sparse_blas/operations.cpp
upsj: Pulled the ifdef inside to make it easier for the formatter.

PR #1356, opened 2023-06-30
Adds row scatter functionality to dense

MarcelKoch:
This PR add row scatter functionality to dense. The target rows can be specified either by a plain `array<index_type>` or by `index_set<index_type>`. Note that the index_set implies that the target rows are ordered. Signature:
```c++
void row_scatter(const array<index_type>*, LinOp*);
void row_scatter(const index_set<index_type>*, LinOp*);
```


yhmtsai (changes requested):
the function description and one test variable

core/matrix/dense.cpp
```diff
@@ -1158,6 +1159,41 @@ void Dense<ValueType>::row_gather_impl(const Dense<ValueType>* alpha,
...
+template <typename ValueType>
+template <typename OutputType, typename IndexType>
+void Dense<ValueType>::row_scatter_impl(const array<IndexType>* row_idxs,
+                                        Dense<OutputType>* target) const
+{
+    auto exec = this->get_executor();
+    dim<2> expected_dim{row_idxs->get_num_elems(), this->get_size()[1]};
+    GKO_ASSERT_EQUAL_DIMENSIONS(expected_dim, this);
+    GKO_ASSERT_EQUAL_COLS(this, target);
+    // @todo check that indices are inbounds for target
```
yhmtsai, line 1171: not sure whether to check the indices here because it will affect the performance. I guess it will have the same runtime as `make_row_scatter`

core/matrix/dense.cpp
```diff
@@ -1158,6 +1159,41 @@ void Dense<ValueType>::row_gather_impl(const Dense<ValueType>* alpha,
...
 
+template <typename ValueType>
+template <typename OutputType, typename IndexType>
+void Dense<ValueType>::row_scatter_impl(const array<IndexType>* row_idxs,
+                                        Dense<OutputType>* target) const
+{
+    auto exec = this->get_executor();
+    dim<2> expected_dim{row_idxs->get_num_elems(), this->get_size()[1]};
+    GKO_ASSERT_EQUAL_DIMENSIONS(expected_dim, this);
+    GKO_ASSERT_EQUAL_COLS(this, target);
```
yhmtsai, line 1170: it is also checked by the `EQUAL_DIMENSIONS`

yhmtsai (dismissed):

core/matrix/dense.cpp
MarcelKoch: the first one checks `expected_dim` and `this`, while the second one checks `this` and `target`. So, I don't think it's covered.

core/matrix/dense.cpp
MarcelKoch: I added an in-kernel check to make sure we don't write out-of-bounds. I'm open to discussions on how to handle that, because I don't think there are many places where we already use something like this. In principle, we could also do this for the gather kernels, although I don't think the issue is as big there, since we are only reading out-of-bounds. Also, we could wrap the check in some `#ifdef` to disable it perhaps in release builds.

core/matrix/dense.cpp
yhmtsai: you are right.

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -539,6 +543,42 @@ class Dense
                     ptr_param<const LinOp> beta,
                     ptr_param<LinOp> row_collection) const;
 
+    /**
+     * Copies this matrix into the given rows of the target matrix.
+     *
+     * @tparam IndexType  the index type, either int32 or int 64
```
greole, line 549:
```suggestion
     * @tparam IndexType  the index type, either int32 or int64
```

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -539,6 +543,42 @@ class Dense
                     ptr_param<const LinOp> beta,
                     ptr_param<LinOp> row_collection) const;
 
+    /**
+     * Copies this matrix into the given rows of the target matrix.
```
greole, line 547: I think this needs to be a bit more expressive in what it means to "copy this matrix into the given rows". Could you add an example here.

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -539,6 +543,42 @@ class Dense
...
+     *          identical.
+     */
+    template <typename IndexType>
+    void row_scatter(const array<IndexType>* scatter_indices,
+                     ptr_param<LinOp> target) const;
+
+    /**
+     * Copies this matrix into the given rows of the target matrix.
+     *
+     * @tparam IndexType  the index type, either int32 or int 64
```
greole, line 569:
```suggestion
     * @tparam IndexType  the index type, either int32 or int64
```

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -539,6 +543,42 @@ class Dense
...
+     *          for indices `i, j` with `scatter_indices[i] ==
+     *          scatter_indices[j]` the rows `i, j` of this matrix are
+     *          identical.
+     */
+    template <typename IndexType>
+    void row_scatter(const array<IndexType>* scatter_indices,
+                     ptr_param<LinOp> target) const;
+
+    /**
+     * Copies this matrix into the given rows of the target matrix.
```
greole, line 567: Same comment as above.

reference/test/matrix/dense_kernels.cpp
```diff
@@ -2361,6 +2361,137 @@ TYPED_TEST(Dense, SquareMatrixGatherRowsIntoDenseFailsForWrongDimensions64)
...
+                                    exec);
+    auto row_collection = gko::initialize<Mtx>(
+        {{3.0, 2.7, 6.5}, {0.7, 1.1, 4.0}, {3.4, 3.8, 7.8}}, exec);
+    gko::index_set<gko::int32> permute_idxs{exec, {1, 0, 4}};
+
+    row_collection->row_scatter(&permute_idxs, mtx);
+
+    GKO_ASSERT_MTX_NEAR(mtx,
+                        l<T>({{3.0, 2.7, 6.5},
+                              {0.7, 1.1, 4.0},
```
greole, line 2450: I am confused, looks to me like row 0 and 1 should be switched? Doesn't `permute_idxs{exec, {1, 0, 4}}` say `{3.0, 2.7, 6.5}` goes into row 1 ?

greole (commented):
Not a full review yet, just some quick comments/questions?

reference/test/matrix/dense_kernels.cpp
MarcelKoch: An index set is always ordered. So `permute_idxs{exec, {1, 0, 4}}` is equivalent to `permute_idxs{exec, {0, 1, 4}}`. I will add a comment to `row_scatter`

include/ginkgo/core/matrix/dense.hpp
MarcelKoch: TBH, since this is the brief description, I think it's fine. But the detailed description is mangled up in the description of the parameters, so perhaps moving that into a separate description block is better.

common/cuda_hip/matrix/dense_kernels.hpp.inc
```diff
@@ -392,6 +392,39 @@ __global__ __launch_bounds__(default_block_size) void fill_in_sellp(
 }
 
 
+template <typename ValueType, typename OutputType, typename IndexType>
+__global__ __launch_bounds__(default_block_size) void row_scatter(
```
upsj, line 396: Do we need this kernel for `index_set`? I don't see it used anywhere else in practice

common/unified/matrix/dense_kernels.template.cpp
```diff
@@ -456,6 +456,33 @@ void advanced_row_gather(std::shared_ptr<const DefaultExecutor> exec,
...
+    array<bool> invalid_access_arr{exec, {false}};
+    run_kernel(
+        exec,
+        [num_rows = target->get_size()[0]] GKO_KERNEL(
+            auto row, auto col, auto orig, auto rows, auto scattered,
+            auto* invalid_access_ptr) {
+            if (rows[row] >= num_rows) {
+                *invalid_access_ptr = true;
+                return;
+            }
```
upsj, line 474: maybe leave a TODO hear for when we have established a uniform error handling framework.

core/matrix/dense.cpp
```diff
@@ -1287,6 +1288,42 @@ void Dense<ValueType>::row_gather_impl(const Dense<ValueType>* alpha,
...
+
+    bool invalid_access = false;
+
+    exec->run(dense::make_row_scatter(
+        make_temporary_clone(exec, row_idxs).get(), orig,
+        make_temporary_clone(exec, target).get(), invalid_access));
+
+    if (invalid_access) {
+        GKO_INVALID_STATE(
+            "Out-of-bounds access detected during kernel execution.");
```
upsj, line 1322: I think we can be more specific here
```suggestion
            "Out-of-bounds scatter index.");
```

upsj (commented):
LGTM in general, though I would consider making this more strongly typed than just an `array`, then we also have a well-defined place where we can check for invalid data #770. I think we should already keep an eye out for a potential future `scatter_add` interface, which may profit from some preprocessing to avoid atomics, which would also encourage a dedicated type.

upsj:
@MarcelKoch my goal is to deprecate the `array` interfaces in favor of `Permutation` and related classes in GInkgo 2.0, for consistency it would seem sensible to resolve this topic before adding new interfaces which might need to be deprecated shortly after. `index_set` IMO is an overly specific type for this purpose, as its storage layout favors a specific kind of subset, which doesn't give any performance benefit in the use cases for row scattering I can think of (namely halo exchange) on general meshes.

MarcelKoch:
Ok, that makes more sense. I will look into providing a `Scatter` class instead.

MarcelKoch:
After looking into it a bit, I don't think adding an extra class for scattering is the right choice at the moment. Without the `scatter_add` this is a fairly simple operation, so an extra class would be overkill. E.g. if the scatter is not done repeatedly, users would have to write:
```
RowScatterer<int32>::create(exec, size, idxs)->apply(a, b);
```
which to me seems to be quite cumbersome, compared to `a->row_scatter(idxs, b)`.

upsj:
To me this is an abstraction level mismatch - `array` is a low-level type with no interpretation of the data associated with it, `Dense` is a high-level type that gives an interpretation to the underlying data. The `array` you pass into `row_scatter` could be anything. Otherwise the same kind of argument could be made for `Permutation`. IMO we should aim for interfaces that don't mix different abstraction levels.

As for the discussion between `scatter_add` and `scatter`, I believe we should consider making `scatter_add` the default, since otherwise the operation can produce race conditions and non-deterministic behavior trivially.

MarcelKoch:
Making `scatter_add` the default will require quite a bit of work. Since I don't think this feature is urgently needed, I will remove this PR from the upcoming release.

PR #1361, opened 2023-07-06
Remove CUDA_ARCH from GPU-enabled CI jobs

upsj:
These are all jobs running on amdci right now. We can specify the CUDA architecture from environment variables of the gitlab-runner (I already added it to the config), which makes the test runs a bit more flexible as well, since I have a fallback node with only compute capability 6.0, where the tests currently fail.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -98,7 +98,6 @@ build/cuda92/nompi/gcc/all/release/shared:
     BUILD_CUDA: "ON"
     BUILD_HIP: "ON"
     BUILD_TYPE: "Release"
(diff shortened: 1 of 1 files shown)


MarcelKoch (approved):

MarcelKoch:
I guess the docker image for the unified runner could be updated: https://gitlab.com/nvidia/container-images/cuda/blob/master/doc/support-policy.md

upsj:
Yes, we need to update that in our docker image generation scripts, this was just pulled from dockerhub.

pratikvn (approved):

PR #1362, opened 2023-07-07
Adds test for zero size common-kernel reduction

MarcelKoch:
This PR adds test for inputs of size zero to our common-kernel reductions. This ensures that for empty inputs, the output will be the provided `identity` value. 

Additionally, it fixes the dpcpp kernels to ensure the correct behavior.


upsj (approved):
The bounds listed in the comments don't match the parameters, otherwise LGTM!

pratikvn (approved):

yhmtsai (approved):
thanks for picking this up with the intensive tests.

yhmtsai:
could you also add `fix` keyword into the pr title such that we do not ignore it from the changelog?

MarcelKoch:
rebase!

PR #1365, opened 2023-07-10
Makes Gitlab CI Interruptible

MarcelKoch:
This PR makes our Gitlab CI interruptible, similar to the setup that we have for github. Only CI that does not run on master/develop branch can be interrupted.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -25,7 +25,7 @@ include:
   # [1] https://gitlab.com/gitlab-org/gitlab/-/issues/194023#note_1225906002
   - local: '.gitlab/add-interrupt.yml'
     rules:
-      - if: $CI_COMMIT_BRANCH != "master" && $CI_COMMIT_BRANCH != "develop"
+      - if: $CI_COMMIT_BRANCH != "master" && $CI_COMMIT_BRANCH != "develop" && CI_COMMIT_TAG !~ /^v\d+\.\d+\.\d+/
 
 sync:
   stage: sync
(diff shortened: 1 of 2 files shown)


upsj (approved):
LGTM! Awesome that his now works :)

.gitlab-ci.yml
```diff
@@ -19,6 +19,13 @@ include:
   - local: '.gitlab/rules.yml'
   - local: '.gitlab/scripts.yml'
   - local: '.gitlab/variables.yml'
+  # This is a workaround to conditionally make the branch pipelines
+  # interruptible, because the flag does not directly support rules [1].
+  #
+  # [1] https://gitlab.com/gitlab-org/gitlab/-/issues/194023#note_1225906002
+  - local: '.gitlab/add-interrupt.yml'
+    rules:
+      - if: $CI_COMMIT_BRANCH != "master" && $CI_COMMIT_BRANCH != "develop"
```
yhmtsai, line 28: also the tagged branch `$CI_COMMIT_TAG` can not be interruptible.
We also use the tag for the release although the tagged version should only be pushed once.

yhmtsai:
also @tcojean needs to adjust the option https://docs.gitlab.com/ee/ci/pipelines/settings.html#auto-cancel-redundant-pipelines to make it work

.gitlab-ci.yml
```diff
@@ -859,10 +867,10 @@ threadsanitizer:
...
     - LD_PRELOAD=/usr/local/lib/libomp.so
       CC=clang CXX=clang++
-        ctest -V -S cmake/CTestScript.cmake -DCTEST_BUILD_CONFIGURATION=TSAN
-        -DCTEST_MEMORYCHECK_TYPE=ThreadSanitizer
-        -DCTEST_MEMORYCHECK_SANITIZER_OPTIONS=ignore_noninstrumented_modules=1
-        --timeout 6000
+      ctest -V -S cmake/CTestScript.cmake -DCTEST_BUILD_CONFIGURATION=TSAN
+      -DCTEST_MEMORYCHECK_TYPE=ThreadSanitizer
+      -DCTEST_MEMORYCHECK_SANITIZER_OPTIONS=ignore_noninstrumented_modules=1
+      --timeout 6000
```
tcojean, line 873: please keep the alignment for all the scripts

MarcelKoch:
I think the interrupt doesn't work if we force push... That makes it not as helpful with our workflow

yhmtsai:
It's sad they do not handle the forced push. if it does not lead another error, it's still helpful partially.
I guess they go through all previous sha not the branch name.
Another annoying way is to use the API to get all pipeline with the same branch name and then cancel them by API.

MarcelKoch:
rebase!

PR #1369, opened 2023-07-11
Fix OMP column reduction kernel

thoasm:
The kernel requested more memory than necessary in most scenarios because of a faulty temporary storage estimation.
`rows` was used in the calculation instead of `cols`.

### DIFF
--- a/omp/base/kernel_launch_reduction.hpp
+++ b/omp/base/kernel_launch_reduction.hpp
@@ -65,7 +65,8 @@ void run_kernel_reduction_impl(std::shared_ptr<const OmpExecutor> exec,
     const auto ssize = static_cast<int64>(size);
     // Limit the number of threads to the number of columns
     const auto num_threads = std::min<int64>(omp_get_max_threads(), ssize);
-    const auto work_per_thread = ceildiv(ssize, num_threads);
+    const auto work_per_thread =
+        ceildiv(ssize, std::max<int64>(num_threads, 1));
     const auto required_storage = sizeof(ValueType) * num_threads;
     if (tmp.get_num_elems() < required_storage) {
         tmp.resize_and_reset(required_storage);
@@ -74,17 +75,20 @@ void run_kernel_reduction_impl(std::shared_ptr<const OmpExecutor> exec,
 #pragma omp parallel num_threads(num_threads)
     {
         const auto thread_id = omp_get_thread_num();
-        const auto begin = thread_id * work_per_thread;
-        const auto end = std::min(ssize, begin + work_per_thread);
+        if (thread_id < num_threads) {
+            const auto begin = thread_id * work_per_thread;
+            const auto end = std::min(ssize, begin + work_per_thread);
 
-        auto local_partial = identity;
-        for (auto i = begin; i < end; i++) {
-            local_partial = op(local_partial, fn(i, map_to_device(args)...));
+            auto local_partial = identity;
+            for (auto i = begin; i < end; i++) {
+                local_partial =
+                    op(local_partial, fn(i, map_to_device(args)...));
+            }
+            partial[thread_id] = local_partial;
         }
-        partial[thread_id] = local_partial;
     }
-    *result = finalize(std::accumulate(
-        partial, partial + required_storage / sizeof(ValueType), identity, op));
+    *result =
+        finalize(std::accumulate(partial, partial + num_threads, identity, op));
 }
 
 
@@ -102,7 +106,7 @@ void run_kernel_reduction_sized_impl(syn::value_list<int, remainder_cols>,
     const auto cols = static_cast<int64>(size[1]);
     // Limit the number of threads to the number of columns
     const auto num_threads = std::min<int64>(omp_get_max_threads(), rows);
-    const auto work_per_thread = ceildiv(rows, num_threads);
+    const auto work_per_thread = ceildiv(rows, std::max<int64>(num_threads, 1));
     const auto required_storage = sizeof(ValueType) * num_threads;
     if (tmp.get_num_elems() < required_storage) {
         tmp.resize_and_reset(required_storage);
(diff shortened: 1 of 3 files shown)


upsj (approved):
LGTM, thanks for the fix. Small naming nit: this is a column reduction, because we are reducing a column to a single value. I'd love to see a test for this behavior, but it's kind of hard since the size depends on `omp_max_threads`. We can query the size of the temporary allocation from the array we pass in optionally.

MarcelKoch (approved):

yhmtsai (approved):
LGTM. that will be great if you have test for it.
How about to create a larger or equal size tmp array with filling NaN?
check the value in the required size are not NaN and the value out of the required size are NaN.

thoasm:
I just pushed multiple tests for the various reductions that monitor the cache size.

test/base/kernel_launch_generic.cpp
```diff
@@ -468,6 +543,53 @@ void run2d_row_reduction(std::shared_ptr<gko::EXEC_TYPE> exec)
...
+                static_assert(is_same<decltype(j), int64>::value, "index");
+                return i + j + 2;
+            },
+            [] GKO_KERNEL(auto i, auto j) { return std::max(i, j); },
+            [] GKO_KERNEL(auto j) { return j; }, int64{}, output.get_data(),
+            result_stride, dim, temp);
+
+        GKO_ASSERT_ARRAY_EQ(host_ref, output);
+        ASSERT_LT(temp.get_num_elems() / sizeof(int64),
+                  max_tmp_elems * max_tmp_elems);
```
yhmtsai, line 574: I do not get this criterion.
if `rows >= 4 * num_thread` or `cols < row`, tmp is not required.
or, it will allocate `valuetype * rows * num_threads.`
`max_tmp_elems^2` = 1048576
1048576/500000 = 2.~, how does it pass the test?

test/base/kernel_launch_generic.cpp
```diff
@@ -468,6 +543,53 @@ void run2d_row_reduction(std::shared_ptr<gko::EXEC_TYPE> exec)
 TEST_F(KernelLaunch, ReductionRow2D) { run2d_row_reduction(exec); }
 
 
+void run2d_row_reduction_cached(std::shared_ptr<gko::EXEC_TYPE> exec,
+                                std::vector<gko::dim<2>> dims)
+{
+    // This assumes at most 256 OpenMP Threads
+    constexpr int64_t max_tmp_elems = 4 * 256;
```
upsj, line 550: That's not a future-proof assumption, and conflicts with other executors with more parallelism. Why does the other approach (tmp storage smaller than input) not work here?

test/base/kernel_launch_generic.cpp
```diff
@@ -517,3 +639,49 @@ void run2d_col_reduction(std::shared_ptr<gko::EXEC_TYPE> exec)
...
+                static_assert(is_same<decltype(j), int64>::value, "index");
+                return i + j + 2;
+            },
+            [] GKO_KERNEL(auto i, auto j) { return std::max(i, j); },
+            [] GKO_KERNEL(auto j) { return j; }, int64{}, output.get_data(),
+            dim, temp);
+
+        GKO_ASSERT_ARRAY_EQ(host_ref, output);
+        // This assumes at most 256 OpenMP Threads
+        const size_type temp_elem_limit =
```
upsj, line 669: same here

test/base/kernel_launch_generic.cpp
```diff
@@ -468,6 +543,53 @@ void run2d_row_reduction(std::shared_ptr<gko::EXEC_TYPE> exec)
...
+        for (int64 i = 0; i < host_ref.get_num_elems(); ++i) {
+            host_ref.get_data()[i] = dim[1] + i + 1;
+        }
+
+        gko::kernels::EXEC_NAMESPACE::run_kernel_row_reduction_cached(
+            exec,
+            [] GKO_KERNEL(auto i, auto j) {
+                static_assert(is_same<decltype(i), int64>::value, "index");
+                static_assert(is_same<decltype(j), int64>::value, "index");
+                return i + j + 2;
```
upsj, line 566: we already check the reductions are correct elsewhere. Here we could simplify things
```suggestion
                return 1;
```

test/base/kernel_launch_generic.cpp
thoasm: Because for small input vectors, the temporary storage can exceed the actual vector size with the current implementations.
I can set it as a minimum and require that `tmp_storage` is either smaller than `max_tmp_elems` or smaller than input size.

upsj: That sounds like a mistake to me. I don't remember the implementation details clearly, would it be possible to fix this easily?

thoasm: I just pushed an update that should address this for all reductions (in this test file).
I tried to change as little of the implementation as possible while limiting the temporary storage to the input size.

test/base/kernel_launch_generic.cpp
thoasm: This was the worst-case scenario, as there was a chance for it to allocate `rows * cols`.
I changed the OMP reduction behavior slightly so it does not blindly use all available threads.

test/base/kernel_launch_generic.cpp
```diff
@@ -364,6 +365,33 @@ void run1d_reduction(std::shared_ptr<gko::EXEC_TYPE> exec)
...
+            exec, [] GKO_KERNEL(auto i) { return i + 1; },
+            [] GKO_KERNEL(auto i, auto j) { return std::max(i, j); },
+            [] GKO_KERNEL(auto j) { return j; }, int64{}, output.get_data(),
+            size, temp);
+
+        ASSERT_EQ(exec->copy_val_to_host(output.get_const_data()),
+                  static_cast<int64>(size));
+        // The temporary storage (used for partial sums) must be smaller than
+        // the input array
+        ASSERT_LE(temp.get_num_elems() / sizeof(int64), size);
```
upsj, line 385: why not the other way round? seems more straightforward to me
```suggestion
        ASSERT_LE(temp.get_num_elems(), size * sizeof(int64));
```

test/base/kernel_launch_generic.cpp
```diff
@@ -432,6 +460,44 @@ void run2d_reduction(std::shared_ptr<gko::EXEC_TYPE> exec)
...
+            exec, [] GKO_KERNEL(auto i, auto j) { return i + j + 2; },
+            [] GKO_KERNEL(auto i, auto j) { return std::max(i, j); },
+            [] GKO_KERNEL(auto j) { return j; }, int64{}, output.get_data(),
+            dim, temp);
+
+        ASSERT_EQ(exec->copy_val_to_host(output.get_const_data()),
+                  static_cast<int64>(dim[0] + dim[1]));
+        // The temporary storage (used for partial sums) must be smaller than
+        // the input array
+        ASSERT_LE(temp.get_num_elems() / sizeof(int64), dim[0] * dim[1]);
```
upsj, line 480:
```suggestion
        ASSERT_LE(temp.get_num_elems(), dim[0] * dim[1] * sizeof(int64));
```

test/base/kernel_launch_generic.cpp
```diff
@@ -481,6 +547,49 @@ void run2d_row_reduction(std::shared_ptr<gko::EXEC_TYPE> exec)
...
+        gko::kernels::EXEC_NAMESPACE::run_kernel_row_reduction_cached(
+            exec, [] GKO_KERNEL(auto i, auto j) { return i + j + 2; },
+            [] GKO_KERNEL(auto i, auto j) { return std::max(i, j); },
+            [] GKO_KERNEL(auto j) { return j; }, int64{}, output.get_data(),
+            result_stride, dim, temp);
+
+        GKO_ASSERT_ARRAY_EQ(host_ref, output);
+        // The temporary storage (used for partial sums) must be smaller than
+        // the input array
+        ASSERT_LE(temp.get_num_elems() / sizeof(int64), dim[0] * dim[1]);
```
upsj, line 572:
```suggestion
        ASSERT_LE(temp.get_num_elems(), dim[0] * dim[1] * sizeof(int64));
```

test/base/kernel_launch_generic.cpp
```diff
@@ -530,3 +639,43 @@ void run2d_col_reduction(std::shared_ptr<gko::EXEC_TYPE> exec)
...
+        }
+
+        gko::kernels::EXEC_NAMESPACE::run_kernel_col_reduction_cached(
+            exec, [] GKO_KERNEL(auto i, auto j) { return i + j + 2; },
+            [] GKO_KERNEL(auto i, auto j) { return std::max(i, j); },
+            [] GKO_KERNEL(auto j) { return j; }, int64{}, output.get_data(),
+            dim, temp);
+
+        GKO_ASSERT_ARRAY_EQ(host_ref, output);
+        ASSERT_LE(temp.get_num_elems() / sizeof(int64), dim[0] * dim[1]);
```
upsj, line 663:
```suggestion
        ASSERT_LE(temp.get_num_elems(), dim[0] * dim[1] * sizeof(int64));
```

omp/base/kernel_launch_reduction.hpp
```diff
@@ -82,8 +83,8 @@ void run_kernel_reduction_impl(std::shared_ptr<const OmpExecutor> exec,
         }
         partial[thread_id] = local_partial;
     }
-    *result =
-        finalize(std::accumulate(partial, partial + num_threads, identity, op));
+    *result = finalize(std::accumulate(
+        partial, partial + required_storage / sizeof(ValueType), identity, op));
```
upsj, line 87: is this change necessary? The result seems to be identical to before

omp/base/kernel_launch_reduction.hpp
```diff
@@ -147,8 +149,8 @@ void run_kernel_reduction_sized_impl(syn::value_list<int, remainder_cols>,
...
```
upsj, line 153: same here, even slightly less readable

upsj (approved):

PR #954, opened 2022-01-24
Change random number generator and update gtest

yhmtsai:
This PR change ranlux generator to default generator to avoid MSVC SFIANE issue with gtest.
~~Moreover, it picks small fix of hybrid -> csr from @upsj #905~~ already merged
Update Gtest to release v1.11.0, which also fix the gcc 11.2 issue as #914 mentioned

PR #1371, opened 2023-07-20
Add BatchMultiVector class, core, kernels and tests

pratikvn:
This PR adds the `BatchMultiVector` class, which implements the batched multi-vector object. Following our discussion, this is now separate from the `LinOp`/`BatchLinOp` hierarchy. In addition, the following things are also added in this PR:

1. `batch_dim<>` functionality which is a simple wrapper over `dim<>` for storing uniform batched objects.
2. batch_struct, to ease the access and implementation of the backend kernels.
3. `batch_initialize` functions used to initialize the all matrix formats in a fashion similar to `Dense<>::initialize(...)`.
4. Backend kernels and tests for OpenMP, CUDA, HIP and DPCPP and their tests.

There are no apply kernels as this is only a Multi-vector object and hence does not lend itself to the `apply` functionality.

To be general enough, a MultiVector is supported even though the solvers in `batch_develop` currently only support single vector.

This PR is first in the series of the functionality that brings batched functionality to Ginkgo develop and part of the functionality has been in `batch-develop`.


pratikvn:
format!

include/ginkgo/core/base/batch_multi_vector.hpp
```diff
@@ -0,0 +1,800 @@
...
+     * Computes the column-wise dot product of each matrix in this batch and its
+     * corresponding entry in `b`. If the vector has complex value_type, then
+     * the conjugate of this is taken.
+     *
+     * @param b  a BatchMultiVector matrix of same dimension as this
+     * @param result  a BatchMultiVector row vector, used to store the dot
+     * product (the number of column in the vector must match the number of
+     * columns of this)
+     */
+    void compute_dot(ptr_param<const BatchMultiVector<ValueType>> b,
```
MarcelKoch, line 345: this is the conjugated dot product, so perhaps it should be renamed

include/ginkgo/core/base/batch_multi_vector.hpp
```diff
@@ -0,0 +1,800 @@
...
+        gko::detail::const_array_view<ValueType>&& values)
+    {
+        // cast const-ness away, but return a const object afterwards,
+        // so we can ensure that no modifications take place.
+        return std::unique_ptr<const BatchMultiVector>(new BatchMultiVector{
+            exec, sizes, gko::detail::array_const_cast(std::move(values))});
+    }
+
+private:
+    inline batch_dim<2> compute_batch_size(
```
MarcelKoch, line 389: I don't think this, or the function below, need to be part of the class. If you move the constructors to the .cc you can just put these functions also in the .cc

include/ginkgo/core/base/batch_multi_vector.hpp
```diff
@@ -0,0 +1,800 @@
...
+    }
+
+    /**
+     * Creates a BatchMultiVector matrix with the same configuration as the
+     * callers matrix.
+     *
+     * @returns a BatchMultiVector matrix with the same configuration as the
+     * caller.
+     */
+    virtual std::unique_ptr<BatchMultiVector> create_with_same_config() const
```
MarcelKoch, line 525: I don't think we should have any virtual functions here. Letting users derive from a type like this makes little sense, and it is also more complicated than expected.

include/ginkgo/core/base/batch_multi_vector.hpp
```diff
@@ -0,0 +1,800 @@
...
+{
+    using batch_multi_vector = BatchMultiVector<typename Matrix::value_type>;
+    size_type num_batch_entries = vals.size();
+    auto vals_begin = begin(vals);
+    size_type common_num_rows = vals_begin->size();
+    auto common_size = dim<2>(common_num_rows, 1);
+    for (size_type b = 0; b < num_batch_entries; ++b) {
+        GKO_ASSERT_EQ(common_num_rows, vals_begin->size());
+        vals_begin++;
+    }
```
MarcelKoch, line 620: range based for loop?
```suggestion
    for (const auto& val: vals) {
        GKO_ASSERT_EQ(common_num_rows, val.size());
    }
```
Or perhaps just merge it directly with the loop below.

include/ginkgo/core/base/batch_dim.hpp
```diff
@@ -0,0 +1,164 @@
...
+ * @tparam Dimensionality  number of dimensions of the dim objects
+ * @tparam DimensionType  datatype used to represent each dimension
+ *
+ * @param x  first object
+ * @param y  second object
+ *
+ * @return `!(x == y)`
+ */
+template <size_type Dimensionality, typename DimensionType>
+inline bool operator!=(const batch_dim<Dimensionality, DimensionType>& x,
```
MarcelKoch, line 136: either this is also part of the class, or `==` is also not part of the class.

core/base/batch_struct.hpp
```diff
@@ -0,0 +1,144 @@
...
+GKO_ATTRIBUTES GKO_INLINE batch_multi_vector::BatchEntry<ValueType> batch_entry(
+    ValueType* const batch_values, const size_type stride, const int num_rows,
+    const int num_rhs, const size_type batch_idx)
+{
+    return {batch_values + batch_idx * stride * num_rows, stride, num_rows,
+            num_rhs};
+}
+
+template <typename ValueType>
+GKO_ATTRIBUTES GKO_INLINE ValueType* batch_entry_ptr(
```
MarcelKoch, line 132: This one is not used yet, so perhaps put it into a later PR?

core/base/batch_multi_vector.cpp
```diff
@@ -0,0 +1,256 @@
...
+    GKO_ASSERT_EQ(alpha->get_num_batch_entries(),
+                  this->get_num_batch_entries());
+    GKO_ASSERT_EQUAL_ROWS(alpha->get_common_size(), dim<2>(1, 1));
+    for (size_type b = 0; b < alpha->get_num_batch_entries(); ++b) {
+        if (alpha->get_common_size()[1] != 1) {
+            // different alpha for each column
+            GKO_ASSERT_EQUAL_COLS(this->get_common_size(),
+                                  alpha->get_common_size());
+        }
+    }
```
MarcelKoch, line 79: I don't think this loop does anything.
```suggestion
      if (alpha->get_common_size()[1] != 1) {
          GKO_ASSERT_EQUAL_COLS(this->get_common_size(),
                                alpha->get_common_size());
      }
```

core/base/batch_multi_vector.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+template <typename ValueType>
+void BatchMultiVector<ValueType>::add_scaled_impl(
+    const BatchMultiVector<ValueType>* alpha,
+    const BatchMultiVector<ValueType>* b)
+{
+    GKO_ASSERT_EQ(alpha->get_num_batch_entries(),
+                  this->get_num_batch_entries());
+    GKO_ASSERT_EQUAL_ROWS(alpha->get_common_size(), dim<2>(1, 1));
+    for (size_type b = 0; b < alpha->get_num_batch_entries(); ++b) {
```
MarcelKoch, line 92: same as above

core/base/batch_multi_vector.cpp
```diff
@@ -0,0 +1,256 @@
...
+        if (alpha->get_common_size()[1] != 1) {
+            // different alpha for each column
+            GKO_ASSERT_EQUAL_COLS(this->get_common_size(),
+                                  alpha->get_common_size());
+        }
+    }
+    GKO_ASSERT_EQ(b->get_num_batch_entries(), this->get_num_batch_entries());
+    GKO_ASSERT_EQUAL_DIMENSIONS(this->get_common_size(), b->get_common_size());
```
MarcelKoch, line 100: Perhaps an assertion `GKO_ASSERT_EQUAL_BATCHES` that combines the two would help.

core/base/batch_multi_vector.cpp
```diff
@@ -0,0 +1,256 @@
...
+    this->get_executor()->run(
+        batch_multi_vector::make_compute_dot(this, b, result));
+}
+
+
+template <typename ValueType>
+void BatchMultiVector<ValueType>::compute_norm2_impl(
+    BatchMultiVector<remove_complex<ValueType>>* result) const
+{
+    GKO_ASSERT_EQ(this->get_num_batch_entries(),
```
MarcelKoch, line 135: Perhaps an assertion `GKO_ASSERT_EQUAL_BATCH_COLS` would help?

core/base/batch_multi_vector.cpp
```diff
@@ -0,0 +1,256 @@
...
+    this->convert_to(result);
+}
+
+
+template <typename MatrixType, typename MatrixData>
+inline void read_impl(MatrixType* mtx, const std::vector<MatrixData>& data)
+{
+    auto common_size = data[0].size;
+    auto batch_size = batch_dim<2>(data.size(), common_size);
+    size_type ind = 0;
```
MarcelKoch, line 167:
```suggestion
```
it's not used

core/base/batch_multi_vector.cpp
```diff
@@ -0,0 +1,256 @@
...
+    for (const auto& b : data) {
+        auto b_size = b.size;
+        GKO_ASSERT_EQUAL_DIMENSIONS(common_size, b_size);
+    }
+    auto tmp =
+        MatrixType::create(mtx->get_executor()->get_master(), batch_size);
+    for (size_type b = 0; b < data.size(); ++b) {
+        size_type ind = 0;
+        for (size_type row = 0; row < data[b].size[0]; ++row) {
```
MarcelKoch, line 176: If you fill the array with zeros, you wouldn't need this nested for-loop + if branch, which would make this clearer

core/base/batch_multi_vector.cpp
```diff
@@ -0,0 +1,256 @@
...
+template <typename ValueType>
+void BatchMultiVector<ValueType>::move_to(
+    BatchMultiVector<next_precision<ValueType>>* result)
+{
+    this->convert_to(result);
+}
+
+
+template <typename MatrixType, typename MatrixData>
+inline void read_impl(MatrixType* mtx, const std::vector<MatrixData>& data)
```
MarcelKoch, line 163: I don't think the inline does anything here.

core/base/batch_multi_vector.cpp
```diff
@@ -0,0 +1,256 @@
...
+}
+
+
+template <typename MatrixType, typename MatrixData>
+inline void write_impl(const MatrixType* mtx, std::vector<MatrixData>& data)
+{
+    std::unique_ptr<const BatchMultiVector<typename MatrixData::value_type>>
+        op{};
+    const MatrixType* tmp{};
+    if (mtx->get_executor()->get_master() != mtx->get_executor()) {
```
MarcelKoch, line 214: how about `make_temporary_clone`?

MarcelKoch (commented):
I think I will break up my review in a few pieces. This first one looks mostly at the interface and core stuff.

MarcelKoch:
I fell that some of my comments are maybe too detailed and might not be critical to this PR. If you think that's the case, then those comments can be handled in another PR.

core/base/batch_multi_vector.cpp
```diff
@@ -0,0 +1,256 @@
...
+    BatchMultiVector<next_precision<ValueType>>* result)
+{
+    this->convert_to(result);
+}
+
+
+template <typename MatrixType, typename MatrixData>
+inline void read_impl(MatrixType* mtx, const std::vector<MatrixData>& data)
+{
+    auto common_size = data[0].size;
```
upsj, line 165: This segfaults if data is empty!

core/base/batch_struct.hpp
```diff
@@ -0,0 +1,144 @@
...
+    int num_rhs;
+};
+
+/**
+ * A 'simple' structure to store a global uniform batch of dense matrices.
+ *
+ * It is uniform in the sense that all matrices in the batch have common sizes.
+ */
+template <typename ValueType>
+struct UniformBatch {
```
upsj, line 64: this type is not polymorphic, so it should be named in snake_case
```suggestion
struct uniform_batch {
```

core/base/batch_struct.hpp
```diff
@@ -0,0 +1,144 @@
...
+ * Extract one object (matrix, vector etc.) from a batch of objects
+ *
+ * This overload is for batch multi-vectors.
+ * These overloads are intended to be called from within a kernel.
+ *
+ * @param batch  The batch of objects to extract from
+ * @param batch_idx  The position of the desired object in the batch
+ */
+template <typename ValueType>
+GKO_ATTRIBUTES GKO_INLINE batch_multi_vector::BatchEntry<ValueType> batch_entry(
```
upsj, line 114: `entry` seems ambiguous to me, how about `tile`?

include/ginkgo/core/base/batch_dim.hpp
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    /**
+     * Get the cumulative storage size offset
+     *
+     * @param b the batch id
+     *
+     * @return the cumulative offset
+     */
+    size_type get_cumulative_offset(size_type b) const
```
upsj, line 84: I like the name :) For autocompletion, the parameter name is visible first before the documentation, maybe call it `batch_id`?

include/ginkgo/core/base/batch_lin_op_helpers.hpp
```diff
@@ -0,0 +1,126 @@
...
+
+    virtual ~BatchReadableFromMatrixData() = default;
+
+    /**
+     * Reads a batch matrix from a std::vector of matrix_data objects.
+     *
+     * @param data  the std::vector of matrix_data objects
+     */
+    virtual void read(
+        const std::vector<matrix_data<ValueType, IndexType>>& data) = 0;
```
upsj, line 77: This interface makes it quite easy to specify mismatching sizes, and doesn't differentiate between uniform and non-uniform. This is probably the part of the interface I am looking for better solutions the hardest.

include/ginkgo/core/base/batch_multi_vector.hpp
```diff
@@ -0,0 +1,800 @@
...
+    void write(std::vector<mat_data>& data) const override;
+
+    void write(std::vector<mat_data32>& data) const override;
+
+    /**
+     * Unbatches the batched dense and creates a std::vector of Dense matrices
+     *
+     * @return  a std::vector containing the Dense matrices.
+     */
+    std::vector<std::unique_ptr<unbatch_type>> unbatch() const
```
upsj, line 136: In most cases, we have a huge number of small problems, instantiating a polymorphic type for each of them is quite inefficient. This seems mostly useful for testing, does it have other (user-facing) applications? I would suggest to have functions we use for testing kept inside test-specific headers first, and consider other more efficient ways of exposing this if necessary.

include/ginkgo/core/base/batch_multi_vector.hpp
```diff
@@ -0,0 +1,800 @@
...
+std::unique_ptr<Matrix> batch_initialize(
+    std::initializer_list<std::initializer_list<
+        std::initializer_list<typename Matrix::value_type>>>
+        vals,
+    std::shared_ptr<const Executor> exec, TArgs&&... create_args)
+{
+    using batch_multi_vector = BatchMultiVector<typename Matrix::value_type>;
+    size_type num_batch_entries = vals.size();
+    auto vals_begin = begin(vals);
+    size_type common_num_rows = vals_begin->size();
```
upsj, line 668: same here

include/ginkgo/core/base/batch_multi_vector.hpp
```diff
@@ -0,0 +1,800 @@
...
+ */
+template <typename Matrix, typename... TArgs>
+std::unique_ptr<Matrix> batch_initialize(
+    const size_type num_matrices,
+    std::initializer_list<std::initializer_list<typename Matrix::value_type>>
+        vals,
+    std::shared_ptr<const Executor> exec, TArgs&&... create_args)
+{
+    using batch_multi_vector = BatchMultiVector<typename Matrix::value_type>;
+    auto common_size = dim<2>(vals.size(), begin(vals)->size());
```
upsj, line 777: same here

omp/base/batch_multi_vector_kernels.cpp
```diff
@@ -0,0 +1,167 @@
...
+namespace omp {
+/**
+ * @brief The BatchMultiVector matrix format namespace.
+ * @ref BatchMultiVector
+ * @ingroup batch_multi_vector
+ */
+namespace batch_multi_vector {
+
+
+#include "reference/base/batch_multi_vector_kernels.hpp.inc"
```
upsj, line 59: this approach limits us in our optimizations, as reference is supposed to be unoptimized

upsj (commented):
First coarse pass with a handful of specific suggestions:
* we will be doing a lot of scheduling that is the same everywhere - looping over all batches in thread blocks, looping over all entries inside a batch. I would suggest to provide abstractions for those to a) give a single place where we can work on scheduling b) reduce the likelihood of indexing mistakes during development and c) give explicit names to the different levels of parallelism we are using here
* we don't need raw reduction implementations, we have abstractions for those already.
* with certain functions (read, write, unbatch), it seems like they are written mainly for testability purposes, I think we should find clear demarcations which functions are supposed to be used in applications, and which are mostly there for testing purposes (and decide whether we even want to expose them to users easily)

pratikvn:
format!

include/ginkgo/core/base/batch_multi_vector.hpp
```diff
@@ -0,0 +1,800 @@
...
+        ptr_param<const BatchMultiVector> other)
+    {
+        // De-referencing `other` before calling the functions (instead of
+        // using operator `->`) is currently required to be compatible with
+        // CUDA 10.1.
+        // Otherwise, it results in a compile error.
+        return (*other).create_with_same_config();
+    }
+
+    friend class BatchMultiVector<next_precision<ValueType>>;
```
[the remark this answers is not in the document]
pratikvn: This is currently the case in `matrix::Dense` as well. But I think we dont need it to be public.

include/ginkgo/core/base/batch_multi_vector.hpp
```diff
@@ -0,0 +1,800 @@
...
+     * @param other  The other matrix whose configuration needs to copied.
+     */
+    static std::unique_ptr<BatchMultiVector> create_with_config_of(
+        ptr_param<const BatchMultiVector> other)
+    {
+        // De-referencing `other` before calling the functions (instead of
+        // using operator `->`) is currently required to be compatible with
+        // CUDA 10.1.
+        // Otherwise, it results in a compile error.
+        return (*other).create_with_same_config();
```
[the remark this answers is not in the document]
pratikvn: I think for simple functions like getters, it should be okay to have them in the header file. Otherwise, I feel the source file gets too crowded.

include/ginkgo/core/base/batch_multi_vector.hpp
pratikvn: I think we probably need to discuss this in a larger group. I agree that having everything in the header file might increase our compilation time and probably also our library size. But having the declarations in the header and the implementation in the source, duplicates the code a bit and reduces the readability of the source file. 

But maybe given our large compile times, we should eventually move everything to the source files :)

core/base/batch_multi_vector.cpp
pratikvn: I do have that in the batch_develop branch, but there we also have more complex checks as we support non-uniform batches. I think for these simple ones, I feel an additional macro is probably unnecessary.

pratikvn:
format!

core/base/batch_multi_vector.cpp
MarcelKoch: I would still prefer if that macro would exist. Remembering to use one macro is simpler than remembering to use two different ones. But I think that can be handled in a different PR.

core/base/batch_multi_vector.cpp
MarcelKoch: The GKO_ASSERT does nothing in release mode. So perhaps use `GKO_INVALID_STATE`, which will throws.

core/base/batch_multi_vector.cpp
MarcelKoch: Since you marked this as resolved, will you change it, or did I miss something?

include/ginkgo/core/base/batch_multi_vector.hpp
```diff
@@ -0,0 +1,813 @@
...
+     */
+    std::unique_ptr<BatchMultiVector> create_with_same_config() const
+    {
+        return BatchMultiVector::create(this->get_executor(), this->get_size());
+    }
+
+    /**
+     * @copydoc scale(const BatchMultiVector *)
+     */
+    void scale_impl(const BatchMultiVector<ValueType>* alpha);
```
MarcelKoch, line 561: Without the virtual stuff, we don't need the two staged implementation `scale->scale_impl` anymore. (Same for the other ones)

include/ginkgo/core/base/batch_multi_vector.hpp
```diff
@@ -0,0 +1,813 @@
...
+    std::initializer_list<std::initializer_list<
+        std::initializer_list<typename Matrix::value_type>>>
+        vals,
+    std::shared_ptr<const Executor> exec, TArgs&&... create_args)
+{
+    using batch_multi_vector = BatchMultiVector<typename Matrix::value_type>;
+    size_type num_batch_entries = vals.size();
+    GKO_ASSERT(num_batch_entries > 0);
+    auto vals_begin = begin(vals);
+    size_type common_num_rows = vals_begin->size();
```
MarcelKoch, line 690: Alternative without assertions:
```suggestion
    size_type common_num_rows = vals_begin ? vals_begin->size() : 0;
```
This would allow to create 0x0 batches.

core/base/batch_struct.hpp
```diff
@@ -0,0 +1,137 @@
...
+#include <ginkgo/core/base/lin_op.hpp>
+#include <ginkgo/core/base/types.hpp>
+
+
+namespace gko {
+namespace batch_multi_vector {
+
+
+/**
+ * Encapsulates one matrix from a batch of dense matrices (vectors).
```
MarcelKoch, line 47:
```suggestion
 * Encapsulates one multi-vector from a batch of mult-vectors.
```
Same goes for the other comments in this file.

common/cuda_hip/base/batch_multi_vector_kernels.hpp.inc
```diff
@@ -0,0 +1,293 @@
...
+                                                               const ValueType>
+                                                               y,
+                                                   const gko::
+                                                       batch_multi_vector::
+                                                           UniformBatch<
+                                                               ValueType>
+                                                               result)
+{
+    for (size_type ibatch = blockIdx.x; ibatch < x.num_batch_entries;
+         ibatch += gridDim.x) {
```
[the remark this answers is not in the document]
pratikvn: Do you mean have a common kernel launcher or declare a lambda in each kernel ?

core/base/batch_struct.hpp
pratikvn: I feel `tile` is more confusing as we are handling one batch entry here. You can again have tiles inside each batch entry in the context of parallelism.

include/ginkgo/core/base/batch_dim.hpp
```diff
@@ -0,0 +1,164 @@
...
+ *
+ * @return a batch_dim object with dimensions swapped
+ */
+template <typename DimensionType>
+inline batch_dim<2, DimensionType> transpose(
+    const batch_dim<2, DimensionType>& input)
+{
+    return batch_dim<2, DimensionType>(input.get_num_batch_entries(),
+                                       gko::transpose(input.get_common_size()));
+}
```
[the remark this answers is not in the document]
pratikvn: Yes, it is needed for cases where you need to perform a transpose for each batch entry. I also think it is well-defined in this case.

include/ginkgo/core/base/batch_multi_vector.hpp
pratikvn: Yes, unfortunately, applications have asked for this. Though it is definitely more of a utility function rather than a performance related function.

upsj: Did they ask specifically for this interface, or for a way to extract single systems from a batch system? Because due to the data layout, I think it would be easier if we allowed users to access a view to a single vector of the batch vector, instead of copying them all out. Something like `std::unique_ptr<Dense> batch_vector->create_single_view(batch_id)`. The same is possible for batch matrices, and incurs essentially no performance overhead.

include/ginkgo/core/base/batch_multi_vector.hpp
```diff
@@ -0,0 +1,800 @@
...
+ *
+ * The values in each of the batches are stored in row-major format (values
+ * belonging to the same row appear consecutive in the memory).
+ *
+ * @tparam ValueType  precision of matrix elements
+ *
+ * @ingroup batch_multi_vector
+ */
+template <typename ValueType = default_precision>
+class BatchMultiVector
```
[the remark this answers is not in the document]
upsj: I'd also prefer BatchVector

include/ginkgo/core/base/batch_dim.hpp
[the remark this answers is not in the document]
upsj: There is currently no functionality inside this PR that uses it though, right? The question whether it is well-defined depends on whether we view it as a tensor or a block-diagonal matrix (or block-stacked matrix in the vector case?).

common/cuda_hip/base/batch_multi_vector_kernels.hpp.inc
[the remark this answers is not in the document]
upsj: have a function that implements this kind of loop, separately for each backend (most importantly GPUs, because there we need this strided indexing).

core/base/batch_multi_vector.cpp
pratikvn: Are you checking the latest change ? This file has changed a lot, so line numbers dont match anymore

include/ginkgo/core/base/batch_dim.hpp
```diff
@@ -0,0 +1,164 @@
...
+     * @return num_batch_entries
+     */
+    size_type get_num_batch_entries() const { return num_batch_entries_; }
+
+    /**
+     * Get the common size of the batches
+     *
+     * @return common_size
+     */
+    dim<dimensionality, dimension_type> get_common_size() const
```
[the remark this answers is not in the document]
pratikvn: I think we just have the uniform case for now. This only supports the uniform case.

include/ginkgo/core/base/batch_multi_vector.hpp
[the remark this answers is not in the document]
pratikvn: I prefer MultiVector for a few reasons:

1. It is semantically more accurate.
2. Will not be confused with std::vector and gko::vector etc.
3. When Dense will be split into a matrix format and a vector format later, I think that would be a MultiVector as well, and that would make this more natural.

include/ginkgo/core/base/batch_multi_vector.hpp
```diff
@@ -0,0 +1,800 @@
...
+    {
+        return batch_size_.get_num_batch_entries();
+    }
+
+    /**
+     * Returns the common size of the batch entries.
+     *
+     * @return the common size stored
+     */
+    dim<2> get_common_size() const { return batch_size_.get_common_size(); }
```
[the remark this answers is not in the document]
pratikvn: I think those are a bit more verbose. So it is easier to have these, making code easier to read

include/ginkgo/core/base/batch_multi_vector.hpp
```diff
@@ -0,0 +1,800 @@
...
+    /**
+     * Returns a pointer to the array of values of the vector.
+     *
+     * @return the pointer to the array of values
+     */
+    value_type* get_values(size_type batch) noexcept
+    {
+        GKO_ASSERT(batch < this->get_num_batch_entries());
+        return values_.get_data() +
+               this->get_size().get_cumulative_offset(batch);
```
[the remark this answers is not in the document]
pratikvn: It will be useful in Dense as well, because we have only uniform size per batch.

cuda/base/batch_struct.hpp
```diff
@@ -0,0 +1,112 @@
...
+#include "cuda/base/config.hpp"
+#include "cuda/base/types.hpp"
+
+
+namespace gko {
+namespace kernels {
+namespace cuda {
+
+
+/** @file batch_struct.hpp
```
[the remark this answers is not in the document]
pratikvn: yes

common/cuda_hip/base/batch_multi_vector_kernels.hpp.inc
```diff
@@ -0,0 +1,272 @@
...
+{
+    ValueType val = zero<ValueType>();
+
+    for (int r = subgroup.thread_rank(); r < x.num_rows; r += subgroup.size()) {
+        val += conj_map(x.values[r * x.stride + rhs_index]) *
+               y.values[r * y.stride + rhs_index];
+    }
+
+    // subgroup level reduction
+    val = reduce(subgroup, val, thrust::plus<ValueType>{});
```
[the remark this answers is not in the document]
pratikvn: I think it makes sense to unify this ? But that is another discussion. But I would like to keep it like this because technically this is cuda/hip, so subgroup makes more sense than subwarp/subwavefront ?

common/cuda_hip/base/batch_multi_vector_kernels.hpp.inc
```diff
@@ -0,0 +1,272 @@
...
+                                                           batch_multi_vector::
+                                                               uniform_batch<
+                                                                   const ValueType>
+                                                                   y,
+                                                       const gko::
+                                                           batch_multi_vector::
+                                                               uniform_batch<
+                                                                   ValueType>
+                                                                   result,
+                                                       Mapping map)
```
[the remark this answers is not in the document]
pratikvn: Yes, I agree. But I just let clang-format handle it.

include/ginkgo/core/base/batch_dim.hpp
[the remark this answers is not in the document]
upsj: I think we have enough discussions in this PR, this is a really minor one without wide-reaching impact, so I'll close it.

include/ginkgo/core/base/batch_multi_vector.hpp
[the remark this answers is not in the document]
upsj: On 1., I think we need to balance semantic accuracy with practicability. In the most common case, people will use just a single vector. If I read the typename, the first half of it (Multi) is irrelevant to that use case, for such important types a concise name is much more useful IMO. Also in terms of implementation effort, do we (or applications) really need multivectors, or are vectors sufficient? It seems like a lot of effort went into the former, but we only did it because the rest of Ginkgo does it that way. Is that sufficient justification?
Related to naming precedents, I think we already made that decision via `distributed::Vector`. gko::vector is an internal type alias that we can freely rename, I suggested `host_vector`.

hip/base/batch_multi_vector_kernels.hip.cpp
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * @brief The BatchMultiVector matrix format namespace.
+ *
+ * @ingroup batch_multi_vector
+ */
+namespace batch_multi_vector {
+
+
+constexpr auto default_block_size = 256;
+constexpr int sm_multiplier = 4;
```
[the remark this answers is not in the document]
pratikvn: Yes, you are right. They are two different things. I will have to look into that later when tuning the performance.

include/ginkgo/core/base/batch_multi_vector.hpp
```diff
@@ -0,0 +1,800 @@
...
+     * @return the pointer to the array of values
+     */
+    value_type* get_values() noexcept { return values_.get_data(); }
+
+    /**
+     * Returns a pointer to the array of values of the vector.
+     *
+     * @return the pointer to the array of values
+     */
+    value_type* get_values(size_type batch) noexcept
```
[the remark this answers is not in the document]
upsj: While that is a nice trick, semantically the two functions do different things, I think they might even deserve different names. `get_values_for_tile(...)` comes to mind (still haven't found a better name than tile, but batch refers to the whole as well, so it's not a good name for the part)

core/test/base/batch_multi_vector.cpp
```diff
@@ -0,0 +1,447 @@
...
+                                          this->exec);
+    auto mat2 = gko::initialize<DenseMtx>({{1.0, 2.5, 3.0}, {1.0, 2.0, 3.0}},
+                                          this->exec);
+
+    auto m = gko::BatchMultiVector<TypeParam>::create(
+        this->exec, std::vector<DenseMtx*>{mat1.get(), mat2.get()});
+    auto m_ref = gko::BatchMultiVector<TypeParam>::create(
+        this->exec, std::vector<DenseMtx*>{mat1.get(), mat2.get(), mat1.get(),
+                                           mat2.get(), mat1.get(), mat2.get()});
+    auto m2 = gko::BatchMultiVector<TypeParam>::create(this->exec, 3, m.get());
```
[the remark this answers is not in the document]
pratikvn: Yes, if constructing from Dense, we want to copy the data over to new object. The main use-case is as a utility for bencharmking and testing.

core/test/base/batch_multi_vector.cpp
```diff
@@ -0,0 +1,447 @@
...
+}
+
+
+TYPED_TEST(BatchMultiVector, CanBeDoubleListConstructed)
+{
+    using value_type = typename TestFixture::value_type;
+    using T = value_type;
+    auto m = gko::batch_initialize<gko::BatchMultiVector<TypeParam>>(
+        {{I<T>{1.0, 1.0, 0.0}, I<T>{2.0, 4.0, 3.0}, I<T>{3.0, 6.0, 1.0}},
+         {I<T>{1.0, 2.0, -1.0}, I<T>{3.0, 4.0, -2.0}, I<T>{5.0, 6.0, -3.0}}},
```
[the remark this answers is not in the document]
pratikvn: Yes, it is batch_id and then linear id

include/ginkgo/core/base/batch_multi_vector.hpp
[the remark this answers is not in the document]
pratikvn: In this case the implementation is exactly the same and is guaranteed not change when pointing to the beginning of the vector. IMO `get_values_for_x` is unnecesarrily verbose. I really feel `tile` is not suitable here as the batch entry can have tiles themselves from a parallelization perspective, and we should not make things confusing.

common/cuda_hip/base/batch_multi_vector_kernels.hpp.inc
```diff
@@ -0,0 +1,272 @@
...
+    }
+}
+
+
+template <typename ValueType, typename Mapping>
+__device__ __forceinline__ void compute_gen_dot_product(
+    const gko::batch_multi_vector::batch_entry<const ValueType>& x,
+    const gko::batch_multi_vector::batch_entry<const ValueType>& y,
+    const gko::batch_multi_vector::batch_entry<ValueType>& result,
+    Mapping conj_map)
```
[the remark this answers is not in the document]
pratikvn: No, but I think as it refers to whether it is mapping to a conjugate function or not, it should be okay ?

include/ginkgo/core/base/batch_multi_vector.hpp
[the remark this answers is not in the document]
pratikvn: While I agree that the common use-case is indeed a single vector, some applications definitely use multiple vectors. I think `MultiVector` captures the meaning when a single vector is used, but Vector does not for the case of MultiVectors, so if we support multiple vectors, then IMO, `Vector` is just incorrect. 

Now that I think about it and look at the `distributed::Vector` class, I think it should be renamed to `MultiVector`. Honestly, I really dislike that the C++ standard uses vector as a class name for its dynamic arrays.

On `BatchMultiVector` v/s `batch::MultiVector`, I am a bit conflicted. I like that it uniformizes the naming, but with the possibility of `using xxxxx` shortcuts, I feel like it has a lot of potential for misuse and confusion.

core/base/batch_multi_vector.cpp
yhmtsai: the matrix data does not contain m * n entries.
It currently goes over all m * n and fill the value into the matrices. The ind will be m*n in the end.
Because you already initial the vectors to zero, you only need to go over the mat_data.size to fill the value into the matrix.

core/test/base/batch_multi_vector.cpp
```diff
@@ -0,0 +1,428 @@
...
+    BatchMultiVector()
+        : exec(gko::ReferenceExecutor::create()),
+          mtx(gko::batch_initialize<gko::BatchMultiVector<value_type>>(
+              {{{-1.0, 2.0, 3.0}, {-1.5, 2.5, 3.5}},
+               {{1.0, 2.5, 3.0}, {1.0, 2.0, 3.0}}},
+              exec))
+    {}
+
+
+    static void assert_equal_to_original_mtx(
```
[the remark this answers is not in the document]
pratikvn: I think for the simple case, it is useful to have this in a explicit fashion.

include/ginkgo/core/base/batch_multi_vector.hpp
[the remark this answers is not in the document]
upsj: I think we should be putting everything new we added here under the `batch` namespace, otherwise we are inconsistent related to `distributed`. On Vector vs. MultiVector, my opinion is not that strong, but batch::Vector or batch::MultiVector definitely reads more smoothly than BatchMultiVector, especially with the prefix being optional inside the namespace.

`using` shortcuts are fine, as long as their scope is limited. There have been enough blog posts written about why `using namespace std` is a bad habit for me to be unconcerned about this eventuality 😄

reference/test/base/batch_multi_vector_kernels.cpp
```diff
@@ -0,0 +1,425 @@
...
+    std::unique_ptr<DenseMtx> mtx_20;
+    std::unique_ptr<DenseMtx> mtx_21;
+    std::unique_ptr<Mtx> mtx_3;
+    std::unique_ptr<DenseMtx> mtx_30;
+    std::unique_ptr<DenseMtx> mtx_31;
+    std::unique_ptr<Mtx> mtx_4;
+    std::unique_ptr<Mtx> mtx_5;
+    std::unique_ptr<Mtx> mtx_6;
+
+    std::ranlux48 rand_engine;
```
yhmtsai, line 127: change the ranlux48 to the default one to avoid SFIANE issue in Windows https://github.com/ginkgo-project/ginkgo/pull/954

test/base/batch_multi_vector_kernels.cpp
```diff
@@ -0,0 +1,295 @@
...
+        dy = gko::clone(exec, y);
+        dalpha = gko::clone(exec, alpha);
+        dbeta = gko::clone(exec, beta);
+        expected = Mtx::create(
+            ref, gko::batch_dim<2>(batch_size, gko::dim<2>{1, num_vecs}));
+        dresult = Mtx::create(
+            exec, gko::batch_dim<2>(batch_size, gko::dim<2>{1, num_vecs}));
+    }
+
+    std::ranlux48 rand_engine;
```
yhmtsai, line 95: same here

dpcpp/base/batch_struct.hpp
```diff
@@ -0,0 +1,93 @@
...
+
+namespace gko {
+namespace kernels {
+namespace dpcpp {
+
+
+/** @file batch_struct.hpp
+ *
+ * Helper functions to generate a batch struct from a batch LinOp,
+ * while also shallow-casting to the requried DPCPP scalar type.
```
yhmtsai, line 53:
```suggestion
 * while also shallow-casting to the required DPCPP scalar type.
```

dpcpp/base/batch_multi_vector_kernels.dp.cpp
```diff
@@ -0,0 +1,318 @@
...
+           const BatchMultiVector<ValueType>* const alpha,
+           BatchMultiVector<ValueType>* const x)
+{
+    const auto alpha_ub = get_batch_struct(alpha);
+    const auto x_ub = get_batch_struct(x);
+
+    const auto num_batches = x_ub.num_batch_entries;
+    auto device = exec->get_queue()->get_device();
+    auto group_size =
+        device.get_info<sycl::info::device::max_work_group_size>();
```
yhmtsai, line 82: I still think the max_work_group_size will waste many resources from CPU especially if the group only handle a small batch

hip/base/batch_struct.hip.hpp
```diff
@@ -0,0 +1,93 @@
...
+
+namespace gko {
+namespace kernels {
+namespace hip {
+
+
+/** @file batch_struct.hpp
+ *
+ * Helper functions to generate a batch struct from a batch LinOp,
+ * while also shallow-casting to the requried Hip scalar type.
```
yhmtsai, line 54:
```suggestion
 * while also shallow-casting to the required Hip scalar type.
```

include/ginkgo/core/base/batch_dim.hpp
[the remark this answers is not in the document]
yhmtsai: With this, you can not really handle the non-uniform case in the future unless making this function return wrong result in non-uniform.

cuda/base/batch_struct.hpp
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * Helper functions to generate a batch struct from a batch LinOp,
+ * while also shallow-casting to the requried CUDA scalar type.
```
yhmtsai, line 54:
```suggestion
 * while also shallow-casting to the required CUDA scalar type.
```

core/base/batch_struct.hpp
```diff
@@ -0,0 +1,136 @@
...
+
+
+/**
+ * Encapsulates one matrix from a batch of multi-vectors.
+ */
+template <typename ValueType>
+struct batch_entry {
+    using value_type = ValueType;
+    ValueType* values;
+    int stride;
```
yhmtsai, line 53: because BatchMultiVector does not support stride, do you still need that?
I think the gmres need to create a view on dense such that it required though.

yhmtsai (approved):
LGTM in general. There's some potential performance improvement we can try.
Should we also have the stride for the BatchMultiVector?

thoasm (commented):
I'm personally in favor of creating a `gko::batch` namespace and removing the `Batch` and `batch` prefixes for all classes (that is what namespaces are for).

Some typos Github actions found:
siginificant -> significant
requried -> required

core/base/batch_struct.hpp
```diff
@@ -0,0 +1,136 @@
...
+    int num_rows;
+    int num_rhs;
```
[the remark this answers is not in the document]
yhmtsai: I think batch only aims for the small matrices. 2^31-1 quite exceed the usage of batch.

reference/test/base/batch_multi_vector_kernels.cpp
```diff
@@ -0,0 +1,425 @@
...
+
+template <typename T>
+class BatchMultiVector : public ::testing::Test {
+protected:
+    using value_type = T;
+    using size_type = gko::size_type;
+    using Mtx = gko::BatchMultiVector<value_type>;
+    using DenseMtx = gko::matrix::Dense<value_type>;
+    using ComplexMtx = gko::to_complex<Mtx>;
+    using RealMtx = gko::remove_complex<Mtx>;
```
MarcelKoch, line 62: unused
```suggestion
```

reference/test/base/batch_multi_vector_kernels.cpp
```diff
@@ -0,0 +1,425 @@
...
+    this->mtx_10->scale(ualpha[0].get());
+    this->mtx_11->scale(ualpha[1].get());
+
+    auto res = this->mtx_1->unbatch();
+    GKO_ASSERT_MTX_NEAR(res[0].get(), this->mtx_10.get(), 0.);
+    GKO_ASSERT_MTX_NEAR(res[1].get(), this->mtx_11.get(), 0.);
+}
+
+
+TYPED_TEST(BatchMultiVector, ScalesDataWithStride)
```
MarcelKoch, line 168: What stride does this use?

reference/test/base/batch_multi_vector_kernels.cpp
```diff
@@ -0,0 +1,425 @@
...
+    using Mtx = typename TestFixture::Mtx;
+    auto result =
+        Mtx::create(this->exec, gko::batch_dim<2>(2, gko::dim<2>{1, 2}));
+    auto result2 =
+        Mtx::create(this->exec, gko::batch_dim<2>(2, gko::dim<2>{1, 2}));
+
+    ASSERT_THROW(this->mtx_0->compute_dot(this->mtx_1.get(), result.get()),
+                 gko::DimensionMismatch);
+    ASSERT_THROW(this->mtx_0->compute_dot(this->mtx_1.get(), result2.get()),
+                 gko::DimensionMismatch);
```
MarcelKoch, line 272: Aren't `result` and `result2` the same?
```suggestion
    auto result =
        Mtx::create(this->exec, gko::batch_dim<2>(2, gko::dim<2>{1, 2}));

    ASSERT_THROW(this->mtx_0->compute_dot(this->mtx_1.get(), result.get()),
                 gko::DimensionMismatch);
```
same for conj_dot

core/base/batch_struct.hpp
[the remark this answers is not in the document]
pratikvn: Yes, `int` should be more than enough for all cases. Using 32 bit instead of 64 also enables lesser register usage.

include/ginkgo/core/base/batch_multi_vector.hpp
```diff
@@ -0,0 +1,769 @@
...
+     * @param matrices  The matrix::Dense objects that need to be batched.
+     *
+     * @note This is a utility function that can serve as a first step to port
+     * to batched data-structures and solvers. Even if the matrices are in
+     * device memory, this method can have siginificant overhead, as new
+     * allocations and deep copies are necessary and hence this constructor must
+     * not be used in performance sensitive applications
+     */
+    BatchMultiVector(std::shared_ptr<const Executor> exec,
+                     const std::vector<matrix::Dense<ValueType>*>& matrices)
```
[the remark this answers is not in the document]
pratikvn: Yes, I didnt use `stride` to simplify things. But maybe I should be using `stride`.

core/test/base/batch_multi_vector.cpp
```diff
@@ -0,0 +1,451 @@
...
+
+TYPED_TEST(BatchMultiVector, CanGetValuesForEntry)
+{
+    using value_type = typename TestFixture::value_type;
+
+    ASSERT_EQ(this->mtx->get_values_for_entry(1)[0], value_type{1.0});
+}
+
+
+TYPED_TEST(BatchMultiVector, CanBeCopied)
```
MarcelKoch, line 119: I think this and `CanBeMoved` `CanBeCloned` should rather be in the `test/...` test, because they depend on the executor.

test/base/batch_multi_vector_kernels.cpp
```diff
@@ -0,0 +1,295 @@
...
+
+
+#include <memory>
+#include <random>
+
+
+#include <gtest/gtest.h>
+
+
+#include <ginkgo/core/base/array.hpp>
```
MarcelKoch, line 43: unused
```suggestion
```

test/base/batch_multi_vector_kernels.cpp
```diff
@@ -0,0 +1,295 @@
...
+class BatchMultiVector : public CommonTestFixture {
+protected:
+    using Mtx = gko::BatchMultiVector<value_type>;
+    using NormVector = gko::BatchMultiVector<gko::remove_complex<value_type>>;
+    using ComplexMtx = gko::BatchMultiVector<std::complex<value_type>>;
+
+    BatchMultiVector() : rand_engine(15) {}
+
+    template <typename MtxType>
+    std::unique_ptr<MtxType> gen_mtx(const size_t batch_size, int num_rows,
```
MarcelKoch, line 63: `batch_size` shadows data member, so perhaps rename one of them

test/base/batch_multi_vector_kernels.cpp
```diff
@@ -0,0 +1,295 @@
...
+#include "core/test/utils/assertions.hpp"
+#include "core/test/utils/batch_helpers.hpp"
+#include "test/utils/executor.hpp"
+
+
+class BatchMultiVector : public CommonTestFixture {
+protected:
+    using Mtx = gko::BatchMultiVector<value_type>;
+    using NormVector = gko::BatchMultiVector<gko::remove_complex<value_type>>;
+    using ComplexMtx = gko::BatchMultiVector<std::complex<value_type>>;
```
MarcelKoch, line 58: This is not actually used (only to declare the data members). I think the tests which would use that (probably dot/conj_dot) are missing.

MarcelKoch (commented):
My last batch of comments.

include/ginkgo/core/base/batch_multi_vector.hpp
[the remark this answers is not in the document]
thoasm: Do you mean the `friend class` statement?
It should make no difference whether that is `private` or `public`:
> [Access specifiers](https://en.cppreference.com/w/cpp/language/access) have no effect on the meaning of friend declarations (they can appear in private: or in public: sections, with no difference).

-- https://en.cppreference.com/w/cpp/language/friend

core/base/batch_multi_vector.cpp
```diff
@@ -0,0 +1,333 @@
...
+        this, make_temporary_output_clone(exec, result).get()));
+}
+
+
+template <typename ValueType>
+void BatchMultiVector<ValueType>::convert_to(
+    BatchMultiVector<next_precision<ValueType>>* result) const
+{
+    result->values_ = this->values_;
+    result->set_size(this->get_size());
```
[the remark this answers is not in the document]
pratikvn: stride will be added in a later PR.

core/test/base/batch_multi_vector.cpp
pratikvn: We want to test the core functionality here, so tests using reference executor should be sufficient I think.

common/cuda_hip/base/batch_multi_vector_kernels.hpp.inc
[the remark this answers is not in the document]
pratikvn: I think this makes sense. I will pick this up later. I have logged this in #1376

core/base/batch_struct.hpp
pratikvn: yes, I think it might be good to leave this for when the stride will be added, so I dont have to update the kernels again.

include/ginkgo/core/base/range.hpp
```diff
@@ -864,7 +864,7 @@ GKO_BIND_UNARY_RANGE_OPERATION_TO_OPERATOR(transpose_operation, transpose);
 
 
 #define GKO_DEPRECATED_SIMPLE_BINARY_OPERATION(_deprecated_name, _name) \
-    struct [[deprecated("Please use " #_name)]] _deprecated_name : _name{};
+    struct [[deprecated("Please use " #_name)]] _deprecated_name : _name {}
```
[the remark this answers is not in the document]
pratikvn: No, I added it, to remove those warnings

dpcpp/base/batch_multi_vector_kernels.dp.cpp
pratikvn: performance related. will be tackled later. #1376

dpcpp/base/batch_multi_vector_kernels.dp.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+    const dim3 block(group_size);
+    const dim3 grid(num_batches);
```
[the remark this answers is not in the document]
pratikvn: will be tackled later. Noted in #1376

common/cuda_hip/base/batch_multi_vector_kernels.hpp.inc
```diff
@@ -0,0 +1,272 @@
...
+        const int col = li % x.num_rhs;
+
+        x.values[row * x.stride + col] =
+            alpha.values[map(col)] * x.values[row * x.stride + col];
+    }
+}
+
+template <typename ValueType, typename Mapping>
+__global__
+__launch_bounds__(default_block_size, sm_multiplier) void scale_kernel(
```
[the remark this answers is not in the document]
pratikvn: #1376

common/cuda_hip/base/batch_multi_vector_kernels.hpp.inc
```diff
@@ -0,0 +1,272 @@
...
+    constexpr auto tile_size = config::warp_size;
+    auto thread_block = group::this_thread_block();
+    auto subgroup = group::tiled_partition<tile_size>(thread_block);
+    const auto subgroup_id = static_cast<int>(threadIdx.x / tile_size);
+    const int num_subgroups_per_block = ceildiv(blockDim.x, tile_size);
+
+    for (int rhs_index = subgroup_id; rhs_index < x.num_rhs;
+         rhs_index += num_subgroups_per_block) {
+        one_norm2(x, rhs_index, result, subgroup);
+    }
```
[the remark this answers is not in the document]
pratikvn: #1376

pratikvn:
format!

pratikvn:
format!

include/ginkgo/core/base/batch_dim.hpp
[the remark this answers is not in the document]
pratikvn: I think if we extend this class to non-uniform objects, then we can just have this return a zero size or maybe even throw.

cuda/base/batch_multi_vector_kernels.cu
```diff
@@ -0,0 +1,82 @@
...
+constexpr auto default_block_size = 256;
+constexpr int sm_multiplier = 4;
+
+// NOTE: DO NOT CHANGE THE ORDERING OF THE INCLUDES
+// force-top: on
+#include "common/cuda_hip/base/batch_multi_vector_kernels.hpp.inc"
+// force-top: off
+
+
+#include "common/cuda_hip/base/batch_multi_vector_kernel_launcher.hpp.inc"
```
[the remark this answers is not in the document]
pratikvn: I prefer to include it here and not have any includes in `.inc` files.

pratikvn:
I see. I think I now get what you mean. That is a good idea. Thanks.

pratikvn:
format!

pratikvn:
@upsj , I have now added a `create_view_for_item()` function and used it to implement the unbatch, read and write. I did not use it for the duplicating constructors as I think using `exec->copy_from(...)` is simple and easier to read than having to go through `matrix::Dense` functionality for these cases.

pratikvn:
@upsj, I am not sure there is a way to remove constructors from the public interface. The `create` method picks them up anyway. I guess we could have a separate named function and remove the constructor altogether. 

As I said, we have applications that use `unbatch`. `read` and `write` are not only used in testing, but useful in other cases. If I understand it correctly, you are not happy with the `std::vector<>` being used there. But given what we want to represent is reading and writing of distinct items, I think `std::vector` captures that quite well. For applications, who also generally store them in distinct problems, it serves as a good starting point to test out ginkgo's code. So, I would say it is quite important to have that in the public interface.

MarcelKoch (approved):
My comments have been address, thanks for that.

upsj:
By remove a constructor from the public interface, I mean delete the function and implement it using other functionality.

If we restrict ourselves by what the application use in `batch-develop`, we are not able to have an honest review of this interface. My point is that `read` and `write` are at the same time overly specific in that they only provide very narrow functionality with poor performance, and overly generic, as they statically allow inputs that cannot be represented in the object. `unbatch` and the duplicating constructor strike me as functions that were added as a quick fix for necessary functionality, but we could do much better with a cleaner interface. Call it a hunch, I believe these will be the functions that would be annoying us the most in the future, so I suggest an interface that lets us do all that (and more) and does not carry the aforementioned issues.

If I look at code that uses these functions, I see everything happening in a loop anyways, so there is no difference in complexity between
```cpp
vec = BatchVector::create(...);
for (int i = 0; i < n; i++) {
    data.push_back(read_batch_item(i, ...));
}
vec->read(data);
```
or
```cpp
auto vectors = vec->unbatch();
for (int i = 0; i < vectors.size(); i++) {
    do_something(vectors[i]);
}
```
and the counterpart with the simpler interface
```cpp
vec = BatchVector::create(...);
for (int i = 0; i < n; i++) {
    vec->create_item_view(i)->read(read_batch_item(i, ...));
}
```
or
```cpp
for (int i = 0; i < vectors.size(); i++) {
    do_something(vector->create_item_view(i));
}
```

pratikvn:
@upsj, if I understand you correctly, you are suggesting that we remove the entire `BatchReadable/BatchWritable` interface (along with their `read` and `write` functions and in places where `batch::MultiVector::read(...)` or `batch::MultiVector::write(...)`, just replace them with the for loop and call the `Dense::read` with `create_item_view` ? 

Therefore, the `MultiVector` class will not have any `read` `write` functions at all. 

I guess I could remove the duplicating constructor and add that to some named function within a Ginkgo internal header along with the `unbatch()` so that it can be used in the tests.

upsj:
@pratikvn exactly, I'm worried about baking in these functions at such an early stage, maybe later on we come up with a better approach, but right now I think we should be conservative in what we put into the interface.

pratikvn:
format!

upsj (approved):
Public interface LGTM! If we can do without `at` (to be improved later) for now, I would be really happy, as I had planned to replace that function by a better solution with Ginkgo 2.0
We can also already prototype that solution in a subsequent PR.

include/ginkgo/core/base/batch_multi_vector.hpp
```diff
@@ -0,0 +1,658 @@
...
+    }
+
+    /**
+     * @copydoc MultiVector::at(size_type, size_type, size_type)
+     */
+    value_type at(size_type batch_id, size_type row, size_type col) const
+    {
+        GKO_ASSERT(batch_id < this->get_num_batch_items());
+        return values_.get_const_data()[linearize_index(batch_id, row, col)];
+    }
```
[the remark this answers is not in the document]
pratikvn: Yes, this is again some sort of a test or a utility function, which maybe we can move to Ginkgo internal. I think it is quite helpful in tests for better readability, but probably should not be used inside performance critical kernels.  I think your suggestions above make sense, but I think we can handle that in a later PR.

.gitlab/scripts.yml
```diff
@@ -38,6 +38,7 @@
         -DCMAKE_CUDA_COMPILER=${CUDA_COMPILER} -DCMAKE_BUILD_TYPE=${BUILD_TYPE}
         -DCMAKE_CXX_FLAGS="${CXX_FLAGS}" -DBUILD_SHARED_LIBS=${BUILD_SHARED_LIBS}
         ${EXTRA_CMAKE_FLAGS} ${CUDA_ARCH_STR} ${CUDA_HOST_STR}
+        -DGINKGO_COMPILER_FLAGS=${GKO_COMPILER_FLAGS}
```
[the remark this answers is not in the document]
pratikvn: [I think it does not ?](https://cdn.artifacts.gitlab-static.net/b3/71/b371625aef4555ed2d68bdfb8a378c365272a348af12d3a1ddb806372e332781/2023_08_03/4794272738/5253030938/job.log?response-content-type=text%2Fplain%3B%20charset%3Dutf-8&response-content-disposition=inline&Expires=1691057648&KeyName=gprd-artifacts-cdn&Signature=C3hGM8GLq2rOv2rh6cZNb2YSVTo=) See the initial output of `GINKGO_COMPILER_FLAGS` that is `<empty>`

.gitlab-ci.yml
```diff
@@ -613,6 +613,7 @@ build/dpcpp/2022-1/cpu/release/static:
     C_COMPILER: "gcc"
     CXX_COMPILER: "dpcpp"
     BUILD_DPCPP: "ON"
+    GKO_COMPILER_FLAGS: "-ffp-model=precise"
```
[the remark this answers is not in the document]
pratikvn: Yes, the idea was to remove `-Wpedantic` to not have any warnings in the CI builds.

.gitlab/scripts.yml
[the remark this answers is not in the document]
upsj: Ah, I thought you did this for DPC++. Which warnings are you trying to hide from us? 😆

.gitlab-ci.yml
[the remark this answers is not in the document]
upsj: I think with DPC++, the flags actually get overwritten, see CMakeLists.txt:62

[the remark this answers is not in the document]
yhmtsai: why do you need to remove `-Wpedantic`?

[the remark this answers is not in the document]
pratikvn: I just wanted to remove the `-Wpedantic` from the CI builds as that unnecessariy floods the CI logs and for example, the gitlab logs exceeds its size and does not log anything anymore, which makes it difficult to see the errors, if any.

.gitlab/scripts.yml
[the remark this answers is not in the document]
pratikvn: :) No, just wanted to remove the warning flood in CI logs.

.gitlab-ci.yml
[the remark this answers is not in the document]
upsj: If we have warnings, we should fix them ;)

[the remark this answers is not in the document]
pratikvn: [It does not seem to get overwritten for DPCPP as well](https://cdn.artifacts.gitlab-static.net/b3/71/b371625aef4555ed2d68bdfb8a378c365272a348af12d3a1ddb806372e332781/2023_08_03/4794272743/5253237023/job.log?response-content-type=text%2Fplain%3B%20charset%3Dutf-8&response-content-disposition=inline&Expires=1691059405&KeyName=gprd-artifacts-cdn&Signature=0fBw1KQasjHQJttLRLktD466C-I=).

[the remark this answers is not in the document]
pratikvn: Its mostly deprecated warnings (from Ginkgo), and warnings from CUDA/HIP related quirks on some compilers. So, we cannot really fix them

PR #1373, opened 2023-07-24
Enable CTest Resources

MarcelKoch:
This PR enables the use ctest resources to better handle the available hardware for inter-test parallelism. Enabling it required some changes to make our tests run on devices other than the one with device id 0.

What is not included: actually enabling the test parallelism. (The resources.json file is only temporary)

To get the full benefits, it is necessary to define a resource file that specifies the available resources. I see two options on how to handle this file:
1. Generate the file from CMake (this is used in rapids.ai)
2. Create a file for each of our gitlab/github runners by hand
Getting 1. to work robustly across all kind of different hardware might be difficult, while option 2. requires a lot of manual set-up and maintanance. 

I've noticed some issues with MPI implementations and OMP (mostly mvapich). It is important to figure out the right environment variables, otherwise the MPI tests won't schedule their threads to different cores. On Leconte I used `MV_ENABLE_AFFINITY=0` and `OMP_NUM_THREADS=dummy_value`.

A timing comparison using ICL's Leconte (40 cores):
- with resources: ~80s (it is more random than without)
- without resources: 400s

### Default Values
Number of OMP thread: 4
Oversubscription of hardware threads: 20% (i.e. specify more slots than HW threads)
Concurrent GPU tests on a single device: 4 (excluding MPI tests)


MarcelKoch:
Should this allow for tests with multiple backends enabled at once? Currently it assumes only one type of GPUs.

MarcelKoch:
Should CPU and GPU test run completely independent? Or should a GPU test also occupy a single HW thread? Currently, it is assumed that they can run independently. Doing it otherwise will complicate the set-up again.

upsj:
Different pipelines sometimes want to use different devices (e.g. SYCL CPU vs. GPU via `SYCL_DEVICE_FILTER` or `ONEAPI_DEVICE_SELECTOR`, or just generally `CUDA_VISIBLE_DEVICES`/`ROCM_VISIBLE_DEVICES`, so dynamically creating the resource file might be preferable. On the other hand, this is mostly relevant to SYCL, where we don't really have multiple devices available on a single system anyways, and we can't control how many CPUs the SYCL device uses (except for maybe the taskset OpenCL environment variable, but that is pretty advanced and doesn't work for level zero).

Since it is heavily influenced by the environment (variables and hardware), and might otherwise harm building the tests on non-GPU enabled login nodes, we should make the entire resource configuration optional.

Collecting the information with the native language (cudaGetNumDevices, omp_get_max_threads etc.) should be straightforward. Alternatively, we could provide a separate tool that generates these resource files?

Finally, if there is nothing fundamentally that would make it hard to implement, I would prefer if we had one resource type per device executor. I think using a single CPU thread as the minimum resource for each test seems sensible.

MarcelKoch:
> Since it is heavily influenced by the environment (variables and hardware), and might otherwise harm building the tests on non-GPU enabled login nodes, we should make the entire resource configuration optional.

I'm not sure if I follow here. This changes only how the tests are run, not how they are build. But mentioning non-GPU login nodes, these might be an issue for dynamically creating the resource file. I think it might not be a problem, if cmake is run again on the compute nodes.

> Finally, if there is nothing fundamentally that would make it hard to implement, I would prefer if we had one resource type per device executor. I think using a single CPU thread as the minimum resource for each test seems sensible.

I'm not sure what this relates to. Do you mean that a device executor occupies the full device? So removing the `RESOURCE_PRECENTAGE` stuff?

upsj:
> But mentioning non-GPU login nodes, these might be an issue for dynamically creating the resource file.

Exactly, that's what I meant - the test properties need to be captured at configure time, if we want to do things dynamically. But mounting a resource file in the container and specifying it to CMake should also be easily doable, we just need a set of binaries that can produce the necessary information (number of devices of each type), and a script to combine them into a JSON file.

> I think it might not be a problem, if cmake is run again on the compute nodes.

I am a bit unhappy with that requirement, because IMO the configuration should only capture the external environment once and be stable afterwards (as much as possible).

MarcelKoch:
> This relates to the question whether we should have a resource type for each GPU vendor. I would say yes, if it's not too complicated.

If I understand you correctly, that would be using `cuda-gpu`, `amd-gpu`, etc., in the resource file. It would complicate things a bit, but should still be doable.

> Exactly, that's what I meant - the test properties need to be captured at configure time, if we want to do things dynamically. But mounting a resource file in the container and specifying it to CMake should also be easily doable, we just need a set of binaries that can produce the necessary information (number of devices of each type), and a script to combine them into a JSON file.

One very simple approach would be to predefine the resource file for all of our supported systems and just put them into our repository.

upsj:
> One very simple approach would be to predefine the resource file for all of our supported systems and just put them into our repository.

That would be hard to maintain, as it requires rebuilding the containers every time we add a new CI system. Doing that on the administrative side is much cleaner and easier to scale.

MarcelKoch:
If we continue with this, I think we have to revisit our gitlab-runners. We need to decide how many jobs we want to allow to run in parallel, and how to limit the resources accordingly.

upsj:
I can take care of setting up the test environments for the runners. I would suggest using an environment variable `CTEST_EXTRA_VARIABLES` that we control from the runner, so we can keep things from breaking on systems without resource files.

cmake/create_test.cmake
```diff
@@ -1,22 +1,32 @@
...
 function(ginkgo_build_test_name test_name target_name)
     file(RELATIVE_PATH REL_BINARY_DIR
-        ${PROJECT_BINARY_DIR} ${CMAKE_CURRENT_BINARY_DIR})
+            ${PROJECT_BINARY_DIR} ${CMAKE_CURRENT_BINARY_DIR})
     string(REPLACE "/" "_" TEST_TARGET_NAME "${REL_BINARY_DIR}/${test_name}")
     set(${target_name} ${TEST_TARGET_NAME} PARENT_SCOPE)
-endfunction(ginkgo_build_test_name)
+endfunction()
+
+function(ginkgo_create_gtest_main)
```
upsj, line 14: we don't seem to be using the targets

core/test/gtest/environments.hpp
```diff
@@ -0,0 +1,188 @@
...
+                std::cerr << omp_threads << " CPU threads" << std::endl;
+            }
+        }
+        // CUDA GPUs
+        if (auto rs_cuda_env = get_ctest_group("cudagpu", rank)) {
+            auto resource = parse_ctest_resources(rs_cuda_env);
+            cuda_device_id = resource.id;
+            if (rank == 0) {
+                std::cerr << "CUDA device " << cuda_device_id << std::endl;
+            }
```
upsj, line 107: Maybe we should be outputting the GPU IDs for all ranks here?

cuda/test/base/CMakeLists.txt
```diff
@@ -1,13 +1,13 @@
 ginkgo_create_cuda_test(array)
 ginkgo_create_cuda_test(cuda_executor)
-ginkgo_create_test(index_set)
-ginkgo_create_test(cuda_executor_reset ADDITIONAL_LIBRARIES Threads::Threads)
-if(GINKGO_HAVE_HWLOC)
+ginkgo_create_cuda_test(index_set)
+ginkgo_create_cuda_test(cuda_executor_reset ADDITIONAL_LIBRARIES Threads::Threads RESOURCE_PERCENTAGE 100)
```
upsj, line 4: These were deleted in `develop`

hip/test/base/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 ginkgo_create_hip_test(hip_executor)
 ginkgo_create_test(index_set)
-ginkgo_create_test(hip_executor_reset ADDITIONAL_LIBRARIES Threads::Threads)
+ginkgo_create_test(hip_executor_reset ADDITIONAL_LIBRARIES Threads::Threads RESOURCE_PERCENTAGE 100)
```
upsj, line 3: same here

hip/test/utils.hip.hpp
```diff
@@ -40,31 +40,22 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-          exec(gko::HipExecutor::create(
-              0, ref, false, gko::default_hip_alloc_mode, stream.get()))
-#else
-          exec(gko::HipExecutor::create(0, ref))
-#endif
+          stream(ResourceEnvironment::hip_device_id),
+          exec(gko::HipExecutor::create(ResourceEnvironment::hip_device_id, ref,
+                                        false, gko::default_hip_alloc_mode,
+                                        stream.get())),
+          guard(exec->get_scoped_device_id_guard())
```
upsj, line 58: This should still differentiate between default and non-default stream

resources.json
```diff
@@ -0,0 +1,51 @@
+{
```
upsj, line 1: no longer needed

test/mpi/preconditioner/schwarz.cpp
```diff
@@ -162,13 +162,14 @@ class SchwarzPreconditioner : public CommonMpiTestFixture {
...
     {
+        auto host_row_part = gko::make_temporary_clone(ref, row_part);
         auto l_dist_vec = dist_vec->get_local_vector();
         auto vec_view = local_vec_type::create_const(
             exec, l_dist_vec->get_size(),
             gko::array<value_type>::const_view(
                 exec, l_dist_vec->get_size()[0],
                 local_vec->get_const_values() +
-                    row_part->get_range_bounds()[comm.rank()]),
+                    host_row_part->get_range_bounds()[comm.rank()]),
```
upsj, line 172: should already be fixed in `develop`

test/tools/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+add_executable(resource_file_generator resource_file_generator.cpp)
+target_link_libraries(resource_file_generator Ginkgo::ginkgo )
```
upsj, line 2: nit: 
```suggestion
target_link_libraries(resource_file_generator Ginkgo::ginkgo)
```

test/utils/executor.hpp
```diff
@@ -119,16 +93,17 @@ inline void init_executor(std::shared_ptr<gko::ReferenceExecutor> ref,
     if (gko::HipExecutor::get_num_devices() == 0) {
         throw std::runtime_error{"No suitable HIP devices"};
     }
-    exec = gko::HipExecutor::create(0, ref, false, gko::default_hip_alloc_mode,
-                                    stream);
+    exec = gko::HipExecutor::create(ResourceEnvironment::hip_device_id, ref,
+                                    false, gko::default_hip_alloc_mode, stream);
```
upsj, line 97: needs to be rebased to use allocators

test/utils/executor.hpp
```diff
@@ -119,16 +93,17 @@ inline void init_executor(std::shared_ptr<gko::ReferenceExecutor> ref,
...
 }
 
 
 inline void init_executor(std::shared_ptr<gko::ReferenceExecutor> ref,
                           std::shared_ptr<gko::DpcppExecutor>& exec)
 {
     if (gko::DpcppExecutor::get_num_devices("gpu") > 0) {
-        exec = gko::DpcppExecutor::create(0, ref, "gpu");
+        exec = gko::DpcppExecutor::create(ResourceEnvironment::sycl_device_id,
+                                          ref, "gpu");
```
upsj, line 106: I guess we just don't run SYCL CPU tests in parallel, right? Controlling their parallelism seems hard

upsj (approved):
Even though part of the code was written by me, I'll add some review comments anyways

LGTM, if we add a `CTEST_EXTRA_ARGS` environment variable to be used in `ctest` invocations, we can even do the resource configuration transparently to the `.gitlab-ci.yml`

cmake/create_test.cmake
MarcelKoch: Yes, due to the changes of the gtest-main we no longer compile the main separately. Each test now compiles its own main, since the main uses different compile definitions depending on the test type. Still, I hope that by using ccache that should not add too much of an overhead.

upsj: you could take the `TARGET_FILES` object library approach we use in our executors to avoid duplication

hip/test/utils.hip.hpp
MarcelKoch: Is there any documentation on that? Because I had errors when running the tests on different devices (although I don't remember the details anymore)

MarcelKoch: I have to check again, might be that the issue was somewhere else.

test/utils/executor.hpp
```diff
@@ -146,14 +121,21 @@ class CommonTestFixture : public ::testing::Test {
...
+          ref{gko::ReferenceExecutor::create()}
     {
-#if defined(GKO_TEST_NONDEFAULT_STREAM) && \
-    (defined(GKO_COMPILING_CUDA) || defined(GKO_COMPILING_HIP))
+#if defined(GKO_COMPILING_CUDA) || defined(GKO_COMPILING_HIP)
         init_executor(ref, exec, stream.get());
 #else
         init_executor(ref, exec);
 #endif
+        guard = exec->get_scoped_device_id_guard();
```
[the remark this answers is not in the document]
MarcelKoch: we call some of our kernels directly and not through `exec->run`. Thereby, we don't set the device id for the kernel call. I added the guard here, instead of adding it to all instances where we call the kernels directly.

cmake/create_test.cmake
MarcelKoch: would that help? I would need to create 10 (5 per executor, 2 per mpi) object libraries, and then I could just use normal libraries for that.

hip/test/utils.hip.hpp
MarcelKoch: This commit has more details on the issue I had: c11df1a9

cmake/create_test.cmake
upsj: good point, then no 😆

MarcelKoch:
format!

core/test/gtest/environments.hpp
```diff
@@ -0,0 +1,220 @@
...
+
+
+#ifdef GKO_COMPILING_CUDA
+
+class CudaEnvironment : public ::testing::Environment {
+public:
+    void TearDown() override
+    {
+        gko::kernels::cuda::reset_device(ResourceEnvironment::cuda_device_id);
+    }
```
upsj, line 193: Same with the device ID and maybe the device name (cudaGetDeviceProperties, might need a wrapper inside ginkgo_cuda)

core/test/gtest/environments.hpp
```diff
@@ -0,0 +1,220 @@
...
+
+
+#ifdef GKO_COMPILING_HIP
+
+class HipEnvironment : public ::testing::Environment {
+public:
+    void TearDown() override
+    {
+        gko::kernels::hip::reset_device(ResourceEnvironment::hip_device_id);
+    }
```
upsj, line 210: same here

core/test/gtest/environments.hpp
```diff
@@ -0,0 +1,220 @@
...
+};
+
+#else
+
+class HipEnvironment : public ::testing::Environment {};
+
+#endif
+
```
upsj, line 218: SYCL environment that outputs the device name (GPU, or CPU if none available)

cmake/create_test.cmake
```diff
@@ -87,9 +151,9 @@ endfunction()
 function(ginkgo_create_test test_name)
     ginkgo_build_test_name(${test_name} test_target_name)
     add_executable(${test_target_name} ${test_name}.cpp)
-    target_link_libraries(${test_target_name} PRIVATE ${create_test_ADDITIONAL_LIBRARIES})
+    target_link_libraries(${test_target_name})
     ginkgo_set_test_target_properties(${test_target_name} ${ARGN})
-    ginkgo_add_test(${test_name} ${test_target_name} ${ARGN})
+    ginkgo_add_test(${test_name} ${test_target_name} ${ARGN} RESOURCE_TYPE ref)
 endfunction(ginkgo_create_test)
```
yhmtsai, line 157: if this allows RESOURCE_TYPE input, the re-naming and changing test macro in CMakeFile is not neccessary.
like `ginkgo_create_test(memory RESOURCE_TYPE CUDA)` not `ginkgo_create_cuda_test(memory)` with linking additional library

cmake/create_test.cmake
```diff
@@ -100,7 +164,7 @@ function(ginkgo_create_dpcpp_test test_name)
     target_compile_options(${test_target_name} PRIVATE ${GINKGO_DPCPP_FLAGS})
     target_link_options(${test_target_name} PRIVATE -fsycl-device-code-split=per_kernel)
     ginkgo_set_test_target_properties(${test_target_name} ${ARGN})
-    ginkgo_add_test(${test_name} ${test_target_name} ${ARGN})
+    ginkgo_add_test(${test_name} ${test_target_name} ${ARGN} RESOURCE_TYPE syclgpu)
```
yhmtsai, line 167: we have syclcpu testing

core/test/gtest/environments.hpp
```diff
@@ -0,0 +1,220 @@
...
+
+#ifdef GKO_COMPILING_HIP
+
+class HipEnvironment : public ::testing::Environment {
+public:
+    void TearDown() override
+    {
+        gko::kernels::hip::reset_device(ResourceEnvironment::hip_device_id);
```
yhmtsai, line 209: The TearDown are executed from several ranks, right?
If a gpu is assigned to several rank, the reset_device will be called several times.
ResetDevice does not contain the counter now, so it leads the issue.

yhmtsai (changes requested):
It will make syclcpu test wrong and the renaming suffix is not necessary imo.

core/test/gtest/environments.hpp
MarcelKoch: I don't think this can happen. Each rank owns a single gpu exclusively.

cuda/test/utils/CMakeLists.txt
```diff
@@ -1 +1 @@
-ginkgo_create_test(assertions_test)
+ginkgo_create_cuda_test(assertions_test)
```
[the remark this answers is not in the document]
MarcelKoch: I forgot that I can set manually the resource type, so you are right. But I would still like to keep it that way, because now it is consistent.

test/tools/resource_file_generator.cpp
```diff
@@ -0,0 +1,105 @@
...
+{
+    std::string json;
+    json.append(R"({
+  "version": {
+    "major": 1,
+    "minor": 0
+  },
+  "local": [
+    {
+)");
```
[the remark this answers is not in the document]
MarcelKoch: too bad. I will just turn it of for this block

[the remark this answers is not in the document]
MarcelKoch: No, I think it looks correct. The indent is due to the '{' at the top.

test/tools/resource_file_generator.cpp
```diff
@@ -0,0 +1,105 @@
...
+    json.append(R"(    }
+  ]
+})");
+    return json;
+}
+
+
+int main()
+{
+    auto num_cpu_threads = std::max(std::thread::hardware_concurrency(), 1u);
```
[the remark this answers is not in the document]
upsj: This tool is meant to be used manually to generate the resource config file and put it into the runner config. This can always be changed afterwards.

test/tools/resource_file_generator.cpp
```diff
@@ -0,0 +1,105 @@
...
+    auto num_cuda_gpus = gko::CudaExecutor::get_num_devices();
+    auto num_hip_gpus = gko::HipExecutor::get_num_devices();
+    auto num_sycl_gpus = gko::DpcppExecutor::get_num_devices("gpu");
```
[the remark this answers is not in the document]
upsj: For SYCL CPUs, we just disable resource configuration, because we cannot easily restrict how many cores it will use.

core/test/gtest/environments.hpp
upsj: Yes, the MPI processes are completely independent, so device resets don't influence each other

cmake/create_test.cmake
upsj: For that, we just ignore the resource config

omp/test/base/CMakeLists.txt
```diff
@@ -1,4 +1,2 @@
-ginkgo_create_test(kernel_launch)
-target_compile_definitions(omp_test_base_kernel_launch PRIVATE GKO_COMPILING_OMP)
-target_link_libraries(omp_test_base_kernel_launch PRIVATE OpenMP::OpenMP_CXX)
-ginkgo_create_test(index_set)
+ginkgo_create_omp_test(kernel_launch)
+ginkgo_create_omp_test(index_set)
```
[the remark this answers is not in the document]
MarcelKoch: Is this the same issue as with the cuda tests? If so, I would also like to keep it consistent here.

core/test/gtest/environments.hpp
```diff
@@ -0,0 +1,220 @@
...
+                      << std::endl;
+        }
+        // HIP GPUs
+        if (auto rs_hip_env = get_ctest_group("hipgpu", rank)) {
+            auto resource = parse_ctest_resources(rs_hip_env);
+            hip_device_id = resource.id;
+            std::cerr << "Rank " << rank << ": HIP device " << cuda_device_id
+                      << std::endl;
+        }
+        // SYCL GPUs (no other devices!)
```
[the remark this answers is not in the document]
MarcelKoch: sycl in general seems hard to handle. AFAIK the sycl backend is chosen at runtime, so we would need to specify resources for both gpu and cpu. Also, is there a way to limit the parallelism of sycl cpu?

test/tools/resource_file_generator.cpp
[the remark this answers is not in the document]
MarcelKoch: I've added a query for the default number of omp threads to the OmpExecutor (can also move that somewhere else).

core/test/gtest/environments.hpp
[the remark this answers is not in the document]
upsj: For the OpenCL backend, we can control the taskset with an environment variable, but that's probably much harder from inside the process. IMO it's not worth it, we can handle a single test that doesn't run in parallel.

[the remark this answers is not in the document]
MarcelKoch: If I understand the API correctly, I think `sycl::device::create_sub_devices` could be used. But I don't know how nicely that would play with our executor.

core/test/gtest/environments.hpp
```diff
@@ -0,0 +1,220 @@
...
+            omp_threads = resource.slots;
+            if (rank == 0) {
+                std::cerr << omp_threads << " CPU threads" << std::endl;
+            }
+        }
+        // CUDA GPUs
+        if (auto rs_cuda_env = get_ctest_group("cudagpu", rank)) {
+            auto resource = parse_ctest_resources(rs_cuda_env);
+            cuda_device_id = resource.id;
+            std::cerr << "Rank " << rank << ": CUDA device " << cuda_device_id
```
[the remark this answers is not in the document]
MarcelKoch: In MPI runs different ranks may interleave, but that is not really controllable.

cmake/create_test.cmake
MarcelKoch: The renaming seems cleaner to me, since it handles all tests uniformly. Otherwise, why are some tests just `ginkgo_create_test` while others are `ginkgo_create_cuda_test` when both are testing cuda functionality.

core/test/gtest/environments.hpp
yhmtsai: three ranks on the same GPU if test on amdci
but from https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1gef69dd5c6d0206c2b8d099abac61f217, it only reset the associated stuff from the current process. It should be okay, right?

cmake/create_test.cmake
yhmtsai: ginkgo_create_cuda_test means the file contains some cuda functionalities and needs nvcc compiler to compile.
ginkgo_create_test only uses host compiler.
If it only tests the functionality without requiring nvcc, it should use host compiler

core/test/gtest/environments.hpp
MarcelKoch: that should not happen. amdci has only one cuda/hip gpu, so the resource can't be allocated. I think it will then skip the test, but I have to check that.

test/tools/resource_file_generator.cpp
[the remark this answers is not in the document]
MarcelKoch: I've added a query for the default number of omp threads to the OmpExecutor (can also move that somewhere else).

MarcelKoch:
I think I have to stall this until we finally update CMake #1368. The issue is that I can't add compile definitions to the hip tests (under /hip/test). I tried using `target_compile_definitions` but using it or not doesn't change the build at all. I need the compile definitions to set up the hip environment properly. Since this is only related to the device reset and printing the used device, we could just ignore it. But I would prefer if we could do it correct now, instead of fixing it later.

upsj:
@yhmtsai No, you need to set them via the compiler flags beforehands. This inflexibility is probably the main reason behind #1334

MarcelKoch:
I also tried modifying the flags passed to `hip_add_executable`, but I couldn't get them propagated to the .cpp file properly.

MarcelKoch:
Thinking about this, perhaps we can just ignore the issues with the /hip/test. It's only for printing some configuration and resetting the device. It should still pick the correct device id.

cmake/create_test.cmake
upsj: Since we should be able to run CPU tests anywhere, I think it would be best if we allowed setting the resource type to CPU for SYCL, ideally setting it to the number of cores that SYCL would be using (I'm sure we'll figure out how to determine that)

MarcelKoch: I think the first question is if we can determine during cmake, if it will run sycl on cpu or gpus.
Regarding setting the number of cores, maybe `ONEAPI_DEVICE_SELECTOR` (https://github.com/intel/llvm/blob/sycl/sycl/doc/EnvironmentVariables.md#oneapi-device-selector) is the one we want. But I don't know how well that is supported.

MarcelKoch: I guess, we need to wrap our dpcpp test in a script to set this variable from the ctest variables.

upsj: I would use either a variable `GINKGO_DPCPP_TEST_GPU_RESOURCES` or indeed parse the SYCL device list to check whether there is a GPU available.

test/utils/executor.hpp
upsj: I just played around with SYCL's create_sub_devices a bit, and it seems like this might actually not be that hard.
```cpp
sycl::device d = sycl::cpu_selector().select_device();
std::cout << d.get_info<sycl::info::device::max_compute_units>() << std::endl;
auto devices = d.create_sub_devices<sycl::info::partition_property::partition_equally>(4);
for (auto dev : devices) {
    std::cout << dev.get_info<sycl::info::device::max_compute_units>() << std::endl;
}
```
The only difference is that OpenCL actually binds to specific cores, while OpenMP doesn't give such fine control

test/utils/executor.hpp
```diff
@@ -122,18 +98,31 @@ inline void init_executor(std::shared_ptr<gko::ReferenceExecutor> ref,
...
+            exec = gko::DpcppExecutor::create(0, ref, "cpu");
+        } else {
+            auto devices = create_cpu_sub_devices();
+            if (ResourceEnvironment::sycl_cpu_id >= devices.size()) {
+                throw gko::OutOfBoundsError(__FILE__, __LINE__,
+                                            ResourceEnvironment::sycl_cpu_id,
+                                            devices.size());
+            }
+            exec = gko::DpcppExecutor::create(
+                *devices[ResourceEnvironment::sycl_cpu_id], ref);
```
upsj, line 124: Alternatively: 
```suggestion
            exec = gko::DpcppExecutor::create(
                *devices[ResourceEnvironment::sycl_cpu_id % devices.size()], ref);
```
which always works, but potentially oversubscribes the CPU

CMakeLists.txt
```diff
@@ -89,6 +89,10 @@ option(GINKGO_INSTALL_RPATH_ORIGIN "Add $ORIGIN (Linux) or @loader_path (MacOS)
 option(GINKGO_INSTALL_RPATH_DEPENDENCIES "Add dependencies to the installation RPATH." OFF)
 option(GINKGO_FORCE_GPU_AWARE_MPI "Assert that the MPI library is GPU aware. This forces Ginkgo to assume that GPU aware functionality is available (OFF (default) or ON), but may fail
      catastrophically in case the MPI implementation is not GPU Aware, and GPU aware functionality has been forced" OFF)
+set(GINKGO_TEST_OMP_PARALLELISM "4" CACHE STRING
```
MarcelKoch, line 92: Perhaps `GINKGO_CI_TEST_OMP_PARALLELISM`, because it doesn't matter if the tests are run individually.

test/utils/executor.hpp
MarcelKoch: I think we should over-subscribe the resources a bit anyway. It might help us with load balancing issues.
But maybe that could be done at the runner level.

test/utils/executor.hpp
```diff
@@ -109,7 +111,18 @@ inline void init_executor(std::shared_ptr<gko::ReferenceExecutor> ref,
         exec = gko::DpcppExecutor::create(ResourceEnvironment::sycl_device_id,
                                           ref, "gpu");
     } else if (gko::DpcppExecutor::get_num_devices("cpu") > 0) {
-        exec = gko::DpcppExecutor::create(0, ref, "cpu");
+        if (ResourceEnvironment::sycl_cpu_id < 0) {
+            exec = gko::DpcppExecutor::create(0, ref, "cpu");
+        } else {
+            auto devices = create_cpu_sub_devices();
+            if (ResourceEnvironment::sycl_cpu_id >= devices.size()) {
```
MarcelKoch, line 118: You also need to check for `sycl_cpu_id == -1`, since that might also be set.

test/utils/executor.hpp
upsj: In the latest version, we just have a single SYCL device that may be CPU or GPU, allowing limited oversubscription. That should be sufficient.

cmake/create_test.cmake
upsj: Now the resource is just called `sycl` and represents either a SYCL GPU or CPU, that should resolve this question.

upsj:
format!

upsj:
One thing which I would like to change, but that is up for discussion: The current setup seems to assume that OpenMP is available, otherwise it will disable any possible parallelism in `resource_test_generator`. I think we should still use the physical number of cores and not the information queried from OpenMP to generate the resource file, especially in SYCL builds, where CMake can't correctly detect OpenMP at the moment.

core/test/gtest/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+function(add_gtest_main suffix definitions)
+    add_library(ginkgo_gtest_main${suffix} STATIC ginkgo_main.cpp resources.cpp)
```
yhmtsai, line 2: any reason for STATIC?

core/test/gtest/environments.hpp
yhmtsai: Although the reset device concern is resolved, we should not skip the test if there's only one GPU in the machine. Multiple ranks should be able to use the same GPU device.

core/test/gtest/environments.hpp
```diff
@@ -0,0 +1,137 @@
...
+    }
+#elif defined(GKO_COMPILING_DPCPP)
+    void print_environment() const
+    {
+        auto device_id = ResourceEnvironment::sycl_device_id;
+        std::stringstream ss;
+        ss << "Rank " << rank_ << ": SYCL device "
+           << gko::kernels::dpcpp::get_device_name(device_id) << " ID "
+           << device_id << std::endl;
+        std::cerr << ss.str();
```
yhmtsai, line 126:
```suggestion
        std::cerr << "Rank " << rank_ << ": SYCL device "
           << gko::kernels::dpcpp::get_device_name(device_id) << " ID "
           << device_id << std::endl;
```
directly to stderr?

cuda/test/base/CMakeLists.txt
```diff
@@ -1,13 +1,13 @@
 ginkgo_create_cuda_test(array)
 ginkgo_create_cuda_test(cuda_executor)
-ginkgo_create_test(index_set)
+ginkgo_create_cuda_test(index_set)
 if(GINKGO_HAVE_HWLOC)
     find_package(NUMA REQUIRED)
     ginkgo_create_cuda_test(cuda_executor_topology ADDITIONAL_LIBRARIES NUMA::NUMA)
-endif()
+endif ()
```
yhmtsai, line 7:
```suggestion
endif()
```
because other modifications in this pr are from `endif ()` to `endif()`

cuda/test/base/CMakeLists.txt
```diff
@@ -1,13 +1,13 @@
...
 ginkgo_create_cuda_test(exception_helpers)
 ginkgo_create_cuda_test(kernel_launch)
 ginkgo_create_cuda_test(lin_op)
 ginkgo_create_cuda_test(math)
-ginkgo_create_test(memory)
-ginkgo_create_cuda_test(scoped_device_id)
+ginkgo_create_cuda_test(memory)
```
yhmtsai, line 12: I do not like using unnecessary nvcc compilers to compile the no-cuda code. We usually use the host compiler to call ginkgo not the nvcc compiler

cmake/create_test.cmake
```diff
@@ -1,26 +1,19 @@
...
-    target_sources(gtest_mpi_main
-      PRIVATE
-      ${PROJECT_SOURCE_DIR}/core/test/mpi/gtest/mpi_listener.cpp)
-    find_package(MPI 3.1 COMPONENTS CXX REQUIRED)
-    target_link_libraries(gtest_mpi_main PRIVATE GTest::GTest MPI::MPI_CXX)
-endfunction(ginkgo_create_gtest_mpi_main)
+  file(RELATIVE_PATH REL_BINARY_DIR
+       ${PROJECT_BINARY_DIR} ${CMAKE_CURRENT_BINARY_DIR})
+  string(REPLACE "/" "_" TEST_TARGET_NAME "${REL_BINARY_DIR}/${test_name}")
+  set(${target_name} ${TEST_TARGET_NAME} PARENT_SCOPE)
```
yhmtsai, line 11:
```suggestion
    file(RELATIVE_PATH REL_BINARY_DIR
         ${PROJECT_BINARY_DIR} ${CMAKE_CURRENT_BINARY_DIR})
    string(REPLACE "/" "_" TEST_TARGET_NAME "${REL_BINARY_DIR}/${test_name}")
    set(${target_name} ${TEST_TARGET_NAME} PARENT_SCOPE)
```
others use 4 space for indentation

cmake/create_test.cmake
```diff
@@ -31,26 +24,75 @@ function(ginkgo_set_test_target_properties test_target_name)
...
+    elseif(add_rr_RESOURCE_TYPE STREQUAL "cpu")
+        if(NOT add_rr_RESOURCE_LOCAL_CORES)
+            set(add_rr_RESOURCE_LOCAL_CORES ${GINKGO_TEST_OMP_PARALLELISM})
+        endif()
+        if(NOT add_rr_RESOURCE_LOCAL_CORES MATCHES "^[0-9]+")
+            message(FATAL_ERROR "Resource specification is invalid: RESOURCE_LOCAL_CORES=${add_rr_RESOURCE_LOCAL_CORES}")
+        endif()
+
+        set(single_resource "cpu:${add_rr_RESOURCE_LOCAL_CORES}")
+    elseif(add_rr_RESOURCE_TYPE MATCHES "^(cudagpu|hipgpu|sycl)$")
```
yhmtsai, line 62:
```suggestion
    elseif(add_rr_RESOURCE_TYPE MATCHES "^(cudagpu|hipgpu|syclgpu)$")
```
syclgpu?

cmake/create_test.cmake
```diff
@@ -31,26 +24,75 @@ function(ginkgo_set_test_target_properties test_target_name)
...
+        if(NOT add_rr_RESOURCE_PERCENTAGE)
+            set(add_rr_RESOURCE_PERCENTAGE 25)
+        endif()
+        if(add_rr_MPI_SIZE GREATER 1)
+            set(add_rr_RESOURCE_PERCENTAGE 100)
+        endif()
```
yhmtsai, line 68: it's unnecessary.

cmake/create_test.cmake
yhmtsai: nevermind, it is different from RESOURCE_GROUPS

cuda/test/utils.hpp
```diff
@@ -41,32 +41,22 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-#ifdef GKO_TEST_NONDEFAULT_STREAM
-          stream(0),
+          stream(ResourceEnvironment::cuda_device_id),
           exec(gko::CudaExecutor::create(
-              0, ref, std::make_shared<gko::CudaAllocator>(), stream.get()))
-#else
-          exec(gko::CudaExecutor::create(0, ref))
-#endif
+              ResourceEnvironment::cuda_device_id, ref,
+              std::make_shared<gko::CudaAllocator>(), stream.get())),
```
yhmtsai, line 58: it means we always test the nondefault stream, we still need to keep the test for default stream, I think.

test/tools/resource_file_generator.cpp
```diff
@@ -0,0 +1,107 @@
...
+            gpus += R"(  {"id": ")" + std::to_string(i) + R"(", "slots": 100})";
+        }
+        if (num_devices) {
+            gpus.append("\n]");
+        }
+    };
+    add_devices(num_cuda_gpus, "cudagpu");
+    add_devices(num_hip_gpus, "hipgpu");
+    // SYCL GPUs, fall back to CPU
+    add_devices(std::max(1, num_sycl_gpus), "sycl");
```
yhmtsai, line 104: maybe I miss something, here uses sycl but the other uses syclgpu.

CMakeLists.txt
```diff
@@ -307,6 +309,10 @@ configure_file(${Ginkgo_SOURCE_DIR}/include/ginkgo/config.hpp.in
 # propagated to the other parts of Ginkgo in case of building as static libraries
 add_subdirectory(devices)        # Basic device functionalities. Always compiled.
 add_subdirectory(common)         # Import list of unified kernel source files
+if(GINKGO_BUILD_TESTS)
+    # use custom target `tests` to build only test binaries
+    add_custom_target(tests)
```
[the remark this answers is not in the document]
upsj: `test` runs the tests, but `tests` makes it easier to build only the tests (not examples and benchmarks)

core/test/gtest/CMakeLists.txt
upsj: Windows seems to have issues with global variables being declared in a shared library, but referenced from another library. This seems the easiest fix

core/test/gtest/environments.hpp
upsj: That makes it more likely for race conditions to jumble the output. This is safer in MPI tests

test/tools/resource_file_generator.cpp
upsj: Should be `sycl` everywhere, where am I still using the old one?

cmake/create_test.cmake
upsj: I think we had some issues with MPI tests being extremely slow if oversubscribing a single GPU. Since we have multi-GPU machines available for each architecture where we test MPI, we can still keep this restriction.

cuda/test/utils.hpp
upsj: Agreed, we discussed this already here https://github.com/ginkgo-project/ginkgo/pull/1373#discussion_r1280830747
I need to try and reproduce the issues Marcel had.

cmake/create_test.cmake
upsj: let's keep this discussion in a single place: https://github.com/ginkgo-project/ginkgo/pull/1373#discussion_r1315754500

cuda/test/base/CMakeLists.txt
upsj: This can be changed easily, I tend to agree

hip/test/utils.hip.hpp
upsj: I could not reproduce the issue, I'm checking to see now if this pops up in CI.

test/utils/executor.hpp
```diff
@@ -151,18 +127,22 @@ class CommonTestFixture : public ::testing::Test {
 
     CommonTestFixture()
         :
-#if defined(GKO_TEST_NONDEFAULT_STREAM) && \
```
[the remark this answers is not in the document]
upsj: see https://github.com/ginkgo-project/ginkgo/pull/1373#discussion_r1324223618

core/test/gtest/environments.hpp
upsj: I checked quickly, there seems to be little runtime difference in whether we oversubscribe GPUs or not. Since we run MPI tests only on multi-GPU systems, and this only applies to CI jobs, I think exclusive GPU access should be more reliable.

For now, I replaced the RESOURCE_PERCENTAGE approach by a single slot per GPU. If we explicitly want to oversubscribe GPUs, we can configure that on the CI runner side with the resources.json file.

cuda/solver/common_trs_kernels.cuh
```diff
@@ -120,12 +120,13 @@ struct CudaSolveStruct : gko::solver::SolveStruct {
         const auto rows = matrix->get_size()[0];
         // workaround suggested by NVIDIA engineers: for some reason
         // cusparse needs non-nullptr input vectors even for analysis
+        // also make sure they are aligned by 16 bytes
```
yhmtsai, line 123: 16bytes or 8 bytes?

test/utils/executor.hpp
```diff
@@ -151,18 +127,22 @@ class CommonTestFixture : public ::testing::Test {
...
-    (defined(GKO_COMPILING_CUDA) || defined(GKO_COMPILING_HIP))
-          stream{0},
+#if defined(GKO_TEST_NONDEFAULT_STREAM) && defined(GKO_COMPILING_CUDA)
+          stream(ResourceEnvironment::cuda_device_id),
+#endif
```
yhmtsai, line 132: when test default stream, it will use the default constructor of stream which is the default stream, right?

test/tools/resource_file_generator.cpp
```diff
@@ -0,0 +1,107 @@
...
+    auto add_devices = [&](int num_devices, const std::string& name) {
+        if (num_devices) {
+            gpus.append(",\n");
+            gpus += '"' + name + "\": [\n";
+        }
+        for (int i = 0; i < num_devices; i++) {
+            if (i > 0) {
+                gpus.append(",\n");
+            }
+            gpus += R"(  {"id": ")" + std::to_string(i) + R"(", "slots": 1})";
```
yhmtsai, line 95: when we need to oversubscripting it, we change the slot number here, right?
in case developers use ctest_resource to test mpi on single gpu node.

cuda/test/utils.hpp
```diff
@@ -41,32 +41,22 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-    testing::AddGlobalTestEnvironment(new CudaEnvironment);
-
-
 class CudaTestFixture : public ::testing::Test {
 protected:
     CudaTestFixture()
         : ref(gko::ReferenceExecutor::create()),
-#ifdef GKO_TEST_NONDEFAULT_STREAM
-          stream(0),
+          stream(ResourceEnvironment::cuda_device_id),
```
yhmtsai, line 55: default_stream like test/utils/executor.hpp

cuda/test/base/CMakeLists.txt
yhmtsai: do you mean it will be changed in another pr?

cmake/create_test.cmake
```diff
@@ -31,26 +24,62 @@ function(ginkgo_set_test_target_properties test_target_name)
...
+    else()
+        message(FATAL_ERROR "Unrecognized resource type ${add_rr_RESOURCE_TYPE}, allowed are: ref, cpu, cudagpu, hipgpu, sycl.")
+    endif()
+
+    if(NOT add_rr_MPI_SIZE)
+        set(add_rr_MPI_SIZE 1)
+    endif()
+    set_property(TEST ${test_name}
+                 PROPERTY
+                 RESOURCE_GROUPS "${add_rr_MPI_SIZE},${single_resource}")
```
yhmtsai, line 73: when using gpu executor, we do not have limitation on cpu side.
The gpu testing which uses omp/ref for reference answer will use full cpu

hip/test/utils.hip.hpp
```diff
@@ -41,32 +41,22 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
     HipTestFixture()
         : ref(gko::ReferenceExecutor::create()),
-#ifdef GKO_TEST_NONDEFAULT_STREAM
-          stream(0),
-          exec(gko::HipExecutor::create(
-              0, ref, std::make_shared<gko::HipAllocator>(), stream.get()))
-#else
-          exec(gko::HipExecutor::create(0, ref))
-#endif
+          stream(ResourceEnvironment::hip_device_id),
```
yhmtsai, line 55: default_stream like test/utils/executor.hpp?

test/utils/executor.hpp
upsj: exactly

cuda/solver/common_trs_kernels.cuh
upsj: double complex needs to be aligned by 16 bytes, since `thrust::complex<double>` has higher alignment requirements to enable vectorized loads/stores

test/tools/resource_file_generator.cpp
upsj: yes, I manually changed the output for the CI runs to match with our number of parallel jobs.

CMakeLists.txt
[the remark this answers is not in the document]
upsj: you can build `ginkgo` exclusively, but maybe this change is not necessary after all?

hip/test/utils.hip.hpp
upsj: good catch, thanks!

CMakeLists.txt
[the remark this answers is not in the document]
upsj: I removed it

cuda/test/base/CMakeLists.txt
upsj: It requires a few more changes, but by removing the `ref` resource type and letting the `-j` parameter take care of overall test parallelism, I can do it ;)

upsj: done

hip/test/utils.hip.hpp
upsj: Seems to be fine without the changes

cmake/create_test.cmake
```diff
@@ -31,26 +24,56 @@ function(ginkgo_set_test_target_properties test_target_name)
...
+        endif()
+        if(NOT add_rr_RESOURCE_LOCAL_CORES MATCHES "^[0-9]+")
+            message(FATAL_ERROR "Resource specification is invalid: RESOURCE_LOCAL_CORES=${add_rr_RESOURCE_LOCAL_CORES}")
+        endif()
+
+        set(single_resource "cpu:${add_rr_RESOURCE_LOCAL_CORES}")
+    elseif(add_rr_RESOURCE_TYPE MATCHES "^(cudagpu|hipgpu|sycl)$")
+        set(single_resource "${add_rr_RESOURCE_TYPE}:1")
+    else()
+        message(FATAL_ERROR "Unrecognized resource type ${add_rr_RESOURCE_TYPE}, allowed are: ref, cpu, cudagpu, hipgpu, sycl.")
```
yhmtsai, line 59:
```suggestion
        message(FATAL_ERROR "Unrecognized resource type ${add_rr_RESOURCE_TYPE}, allowed are: cpu, cudagpu, hipgpu, sycl.")
```

yhmtsai (approved):
LGTM

cuda/solver/common_trs_kernels.cuh
yhmtsai: but `0xDEAD0` only use more than two bytes, does it only has four byte?

upsj: this is a pointer, the 0 at the least significant digit makes it divisible by 16

cmake/create_test.cmake
upsj: ah thanks, I missed that

cuda/test/reorder/CMakeLists.txt
```diff
@@ -1 +1 @@
-ginkgo_create_test(rcm_kernels)
\ No newline at end of file
+ginkgo_create_cuda_test(rcm_kernels)
```
upsj, line 1:
```suggestion
ginkgo_create_test(rcm_kernels RESOURCE_TYPE cudagpu)
```

cuda/test/utils/CMakeLists.txt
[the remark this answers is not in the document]
upsj:
```suggestion
ginkgo_create_test(assertions_test RESOURCE_TYPE cudagpu)
```

dpcpp/test/matrix/CMakeLists.txt
```diff
@@ -1 +1 @@
-ginkgo_create_test(fbcsr_kernels)
+ginkgo_create_dpcpp_test(fbcsr_kernels)
```
upsj, line 1: ginkgo_create_test(fbcsr_kernels RESOURCE_TYPE sycl)

dpcpp/test/preconditioner/CMakeLists.txt
```diff
@@ -1 +1 @@
-ginkgo_create_test(jacobi_kernels)
+ginkgo_create_dpcpp_test(jacobi_kernels)
```
upsj, line 1:
```suggestion
ginkgo_create_test(jacobi_kernels RESOURCE_TYPE sycl)
```

.gitlab/image.yml
```diff
@@ -72,15 +72,13 @@
   image: ginkgohub/rocm:45-mvapich2-gnu8-llvm8
   tags:
     - private_ci
-    - amdci
-    - gpu
+    - amd-gpu
```
MarcelKoch, line 75: Are these changes from the PR #1394 ?

omp/test/matrix/CMakeLists.txt
```diff
@@ -1 +1 @@
-ginkgo_create_test(fbcsr_kernels)
+ginkgo_create_omp_test(fbcsr_kernels)
```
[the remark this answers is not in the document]
upsj:
```suggestion
ginkgo_create_test(fbcsr_kernels RESOURCE_TYPE cpu)
```

omp/test/base/CMakeLists.txt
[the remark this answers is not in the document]
upsj:
```suggestion
ginkgo_create_test(index_set RESOURCE_TYPE cpu)
```

omp/test/reorder/CMakeLists.txt
```diff
@@ -1 +1 @@
-ginkgo_create_test(rcm_kernels)
+ginkgo_create_omp_test(rcm_kernels)
```
upsj, line 1:
```suggestion
ginkgo_create_test(rcm_kernels RESOURCE_TYPE cpu)
```

cmake/create_test.cmake
```diff
@@ -1,26 +1,19 @@
...
+set(gko_test_resource_args "RESOURCE_LOCAL_CORES;RESOURCE_TYPE")
+set(gko_test_single_args "MPI_SIZE;${gko_test_resource_args}")
 set(gko_test_multi_args "DISABLE_EXECUTORS;ADDITIONAL_LIBRARIES;ADDITIONAL_INCLUDES")
+set(gko_test_option_args "NO_RESOURCES")
 
 ## Replaces / by _ to create valid target names from relative paths
 function(ginkgo_build_test_name test_name target_name)
     file(RELATIVE_PATH REL_BINARY_DIR
-        ${PROJECT_BINARY_DIR} ${CMAKE_CURRENT_BINARY_DIR})
+         ${PROJECT_BINARY_DIR} ${CMAKE_CURRENT_BINARY_DIR})
```
MarcelKoch, line 9: This file contains some formatting changes. We should really find a way to get a consistent cmake formatting.

cmake/create_test.cmake
```diff
@@ -87,8 +113,8 @@ endfunction()
 function(ginkgo_create_test test_name)
     ginkgo_build_test_name(${test_name} test_target_name)
     add_executable(${test_target_name} ${test_name}.cpp)
-    target_link_libraries(${test_target_name} PRIVATE ${create_test_ADDITIONAL_LIBRARIES})
-    ginkgo_set_test_target_properties(${test_target_name} ${ARGN})
+    target_link_libraries(${test_target_name})
+    ginkgo_set_test_target_properties(${test_target_name} "" ${ARGN})
```
MarcelKoch, line 117: Maybe use this 
```suggestion
    ginkgo_set_test_target_properties(${test_target_name} "_cpu" ${ARGN})
```
so that it might be clearer what the `""` parameter is for.

MarcelKoch: Same for the omp test

cmake/create_test.cmake
```diff
@@ -214,20 +256,38 @@ function(ginkgo_create_common_test_internal test_name exec_type exec)
     if(exec IN_LIST common_test_DISABLE_EXECUTORS)
         return()
     endif()
+    if (exec STREQUAL reference)
+        set(test_resource_type "")
```
MarcelKoch, line 260: Doesn't this prevent the reference test to participate in the resource allocation?

cuda/test/base/array.cpp
MarcelKoch, line 1: I think the renames are not necessary anymore

MarcelKoch (commented):
only some minor remarks

.gitlab/image.yml
upsj: no, I just noticed we are barely using nla1, and this enables more jobs to be run there.

cmake/create_test.cmake
upsj: The reference tests don't have resources assigned, so they run with the maximal amount of parallelism that `-j` specifies.

MarcelKoch: Doesn't that lead to oversubscription, at least in the omp case? Or is it guarenteed that all the reference+core test will be done before the omp/device tests?

upsj: Right, I missed that. So in theory yes, in practice the reference and core tests are really fast (except for maybe FFT and ISAI) and only use a single core, so at most we have a bit of oversubscription on the OpenMP execution, but since those also tend to spend more time in reference kernels than OpenMP kernels (since they are hopefully more efficient), I am not sure this matters much to us?

MarcelKoch: I guess we can just leave it and see if that actually makes any issues.

dpcpp/test/matrix/CMakeLists.txt
upsj: actually no, this messes up the SINGLE_MODE definition

dpcpp/test/preconditioner/CMakeLists.txt
upsj: actually no, this messes up the SINGLE_MODE definition

cuda/test/base/array.cpp
upsj: Can you elaborate? I changed the CMakeLists.txt to use host tests, so this is looking for the .cpp file

MarcelKoch: I was just thinking less change is better, but it doesn't really matter.

cmake/create_test.cmake
```diff
@@ -31,26 +24,56 @@ function(ginkgo_set_test_target_properties test_target_name)
...
     target_include_directories(${test_target_name} PRIVATE ${Ginkgo_BINARY_DIR} ${set_properties_ADDITIONAL_INCLUDES})
-    target_link_libraries(${test_target_name} PRIVATE ginkgo ${gtest_main} GTest::GTest ${set_properties_ADDITIONAL_LIBRARIES})
+    target_link_libraries(${test_target_name} PRIVATE ginkgo GTest::GTest ${set_properties_ADDITIONAL_LIBRARIES})
+endfunction()
+
+function(ginkgo_add_resource_requirement test_name)
+    cmake_parse_arguments(PARSE_ARGV 1 add_rr "${gko_test_option_args}" "${gko_test_single_args}" "")
+    if(add_rr_NO_RESOURCES OR (NOT add_rr_RESOURCE_TYPE))
+        return()
+    endif ()
```
tcojean, line 45: I would support Marcel's request for a CMake format though there's few maintained ones. There's many inconsistencies through this file and some pure style changes.

cmake/create_test.cmake
```diff
@@ -1,26 +1,19 @@
-set(gko_test_single_args "MPI_SIZE")
+set(gko_test_resource_args "RESOURCE_LOCAL_CORES;RESOURCE_TYPE")
```
tcojean, line 1: This file is becoming fairly large and complex. Do we have a good way of splitting it up at least, maybe in terms of functionality? Maybe unified vs non unified? Or support vs non support?

core/test/gtest/ginkgo_mpi_main.cpp
```diff
@@ -373,17 +378,33 @@ class MPIWrapperPrinter : public ::testing::TestEventListener {
...
-    ::testing::AddGlobalTestEnvironment(new GTestMPIListener::MPIEnvironment);
+    MPI_Comm comm(MPI_COMM_WORLD);
+    int rank;
+    int size;
+    MPI_Comm_rank(comm, &rank);
+    MPI_Comm_size(comm, &size);
+
+    testing::AddGlobalTestEnvironment(new GTestMPIListener::MPIEnvironment);
+    ::testing::AddGlobalTestEnvironment(new ResourceEnvironment(rank, size));
+    ::testing::AddGlobalTestEnvironment(new DeviceEnvironment(rank));
```
tcojean, line 400: Why the different style?

tcojean (approved):

cmake/create_test.cmake
upsj: I'll try to address this in a subsequent PR

cmake/create_test.cmake
upsj: subsequent PR

PR #1375, opened 2023-07-30
Split compilation of dense kernels

upsj:
The compilation time, especially for DPC++, is dominated by `dense_kernels.cpp`, more precisely the reduction kernels with their large number of combinations. By splitting them up, we can improve the parallel efficiency of the builds significantly.


common/unified/matrix/dense_kernels_norm1.cpp
```diff
@@ -0,0 +1,70 @@
...
+namespace kernels {
+namespace GKO_DEVICE_NAMESPACE {
+/**
+ * @brief The Dense matrix format namespace.
+ *
+ * @ingroup dense
+ */
+namespace dense {
+
+template <typename ValueType>
```
yhmtsai, line 52:
```suggestion
namespace dense {


template <typename ValueType>
```

yhmtsai (approved):

pratikvn:
Can we somehow make this automatic and not manually split this into files ? I think this unnecessarily complicates our file structure. How about something like a .tpp file, which marked with macros and using CMake to automatically generate the separate compilation units at build time to enable parallelism ?

upsj:
@pratikvn I don't believe that's possible in the way we do it, e.g. for Jacobi. We would most likely need to do this on the individual sub-configuration instantiation like we do with jacobi_kernels.cu, which is not really possible with the current file structure of the reduction kernels. At least it would mean a separation of the kernel and its configurations into two different files.

upsj:
@pratikvn or maybe I am misunderstanding your suggestion, can you give an example?

yhmtsai:
I think some script to split all functions with the GKO_INSTANTIATION function into different files.
for each GKO_INSTANTIATION function,
```
#include "*.cpp" // to get all necessary file
namespace list{
GKO_INSTANTIATION(function name);
}
```
each instantiation creates a file and compiles those files.
note. it may not work if there's a function without template or inline or anonymous namespace

pratikvn:
Also, how much of an effect does this splitting have ? If possible, I would like to avoid this fragmentation. I think it makes finding functions quite hard and also not really scalable as each reduction operation now has its own file.

upsj:
Ah thanks, I like the idea. We only have a single file with enough compile time where this is necessary though, maybe we can pick this idea up again if we have more such files at some point?

upsj:
@pratikvn Compiling the full file takes 5m53.388s, compiling the individual files in parallel takes 1m1.452s. In my tests, the total build was stuck at least 2-3 minutes at this single file.

pratikvn (approved):
In general I would like to avoid this type of fragmentation. But given that the benefit seems to be large, as a temporary measure, it looks acceptable to be. But I think we should try to do this automatically if possible, so that it is scalable, and the build system can handle the splitting into different instantiation units. See suggestion from @yhmtsai above.

upsj:
@pratikvn this is not meant as a temporary measure. We do the same thing already for Jacobi and ParILUT kernels. If we want to split things into definition and instantiation, we don't really need to use automation (and I don't see any advantage of that), but it still leads to the same number of files, with the small disadvantage that IDEs will in my experience have a harder time displaying information about the function templates without instantiations available.

yhmtsai:
from the splitting definition and instantiation, it generates more files than the origin.
One Cpp has several def and $n$ instantiation (only count macro)
->
One Cpp + $n$ file for different function instantiation.
For me, the 5 min does not sound like a problem.

upsj:
superseded by #1378

PR #1378, opened 2023-07-31
Split compilation of large files automatically

upsj:
As an alternative to https://github.com/ginkgo-project/ginkgo/pull/1375, here I automatically generate files that instantiate a subset of the templates in a source file each. Due to the format of the instantiation source file, it can be included both in its original form and split into different files.


upsj:
To put some number on these changes: `time ninja ginkgo` gives the following output (256 threads)
```

rocm45 before
real    2m44.885s
user    34m59.771s
sys     10m28.850s

rocm45 after
real    1m48.505s
user    18m8.664s
sys     2m23.337s

rocm502 before
real    3m55.766s
user    19m59.029s
sys     2m21.003s

rocm502 after
real    2m37.208s
user    49m14.886s
sys     11m57.626s

NVHPC23 before
real    4m24.576s
user    119m53.211s
sys     24m25.492s

NVHPC23 after
real    3m22.483s
user    8m36.878s
sys     2m18.913s
```

CMakeLists.txt
```diff
@@ -303,6 +303,7 @@ configure_file(${Ginkgo_SOURCE_DIR}/include/ginkgo/config.hpp.in
 # propagated to the other parts of Ginkgo in case of building as static libraries
 add_subdirectory(devices)        # Basic device functionalities. Always compiled.
 add_subdirectory(common)         # Import list of unified kernel source files
+message("${GKO_UNIFIED_COMMON_INSTANTIATE_SOURCES}")
```
upsj, line 306:
```suggestion
```

cmake/template_instantiation.cmake
```diff
@@ -0,0 +1,68 @@
+function(add_instantiation_files source_dir source_file output_files_var)
+    # read full file into variable
+    set(source_path "${source_dir}/${source_file}")
+    file(READ "${source_path}" file_contents)
```
yhmtsai, line 4: if the \r\n is not critical, you can also use `file(STRINGS ...)` which ignores \r

cmake/template_instantiation.cmake
```diff
@@ -0,0 +1,68 @@
...
+        # concatenate header, content and footer and turn semicolons into newlines
+        string(REPLACE ";" "\n" content "${header};${content};${footer}")
+        # and escaped semicolons into regular semicolons again
+        string(REPLACE "<semicolon>" ";" content "${content}")
+        # create a .tmp file, but only copy it over if source file changed
+        # this way, we don't rebuild unnecessarily
+        file(WRITE "${target_path}.tmp" "${content}")
+        add_custom_command(
+            OUTPUT "${target_path}"
+            COMMAND ${CMAKE_COMMAND} -E copy "${target_path}.tmp" "${target_path}"
```
yhmtsai, line 62:
```suggestion
            COMMAND ${CMAKE_COMMAND} -E copy_if_different "${target_path}.tmp" "${target_path}"
```
reduce a little more.

common/CMakeLists.txt
```diff
@@ -1,31 +1,2 @@
...
-    solver/common_gmres_kernels.cpp
-    solver/fcg_kernels.cpp
-    solver/gcr_kernels.cpp
-    solver/gmres_kernels.cpp
-    solver/ir_kernels.cpp
-    )
-list(TRANSFORM UNIFIED_SOURCES PREPEND ${CMAKE_CURRENT_SOURCE_DIR}/unified/)
-set(GKO_UNIFIED_COMMON_SOURCES ${UNIFIED_SOURCES} PARENT_SCOPE)
+add_subdirectory(unified)
+set(GKO_UNIFIED_COMMON_SOURCES ${GKO_UNIFIED_COMMON_SOURCES} PARENT_SCOPE)
```
yhmtsai, line 2: do you need to set that to `PARENT_SCOPE` again? it is set in the common/unified/CMakeLists.txt

common/unified/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+    solver/cg_kernels.cpp
+    solver/cgs_kernels.cpp
+    solver/common_gmres_kernels.cpp
+    solver/fcg_kernels.cpp
+    solver/gcr_kernels.cpp
+    solver/gmres_kernels.cpp
+    solver/ir_kernels.cpp
+    )
+list(TRANSFORM UNIFIED_SOURCES PREPEND ${CMAKE_CURRENT_SOURCE_DIR}/)
+set(GKO_UNIFIED_COMMON_SOURCES ${UNIFIED_SOURCES} PARENT_SCOPE)
```
yhmtsai, line 30:
```suggestion
set(GKO_UNIFIED_COMMON_SOURCES ${UNIFIED_SOURCES} PARENT_SCOPE)

```
nit

cuda/CMakeLists.txt
```diff
@@ -1,4 +1,9 @@
 add_library(ginkgo_cuda $<TARGET_OBJECTS:ginkgo_cuda_device> "")
+include(${PROJECT_SOURCE_DIR}/cmake/template_instantiation.cmake)
+add_instantiation_files(. matrix/csr_kernels.instantiate.cu CSR_INSTANTIATE)
+add_instantiation_files(. matrix/fbcsr_kernels.instantiate.cu FBCSR_INSTANTIATE)
+# we don't split up the dense kernels into distinct compilations
+list(APPEND GKO_UNIFIED_COMMON_SOURCES ${PROJECT_SOURCE_DIR}/common/unified/matrix/dense_kernels.instantiate.cpp)
```
yhmtsai, line 6: the flexibility of tuning on/off is nice. I wonder whether it affects the dpcpp per_source situation.

dpcpp/CMakeLists.txt
```diff
@@ -55,6 +57,7 @@ target_sources(ginkgo_dpcpp
     stop/criterion_kernels.dp.cpp
     stop/residual_norm_kernels.dp.cpp
     ${GKO_UNIFIED_COMMON_SOURCES}
+    ${DENSE_INSTANTIATE}
```
yhmtsai, line 60: the matrix/dense.dp.cpp is not deleted

dev_tools/scripts/config
```diff
@@ -32,6 +32,12 @@
     - FixInclude: "common/unified/base/kernel_launch_solver.hpp"
 - "(cuda|hip|dpcpp|omp)/base/kernel_launch_solver\."
     - FixInclude: "common/unified/base/kernel_launch_solver.hpp"
+- "dense_kernels.template.cpp"
+    - FixInclude: "core/matrix/dense_kernels.hpp"
+- "/csr_kernels.template.*"
+    - FixInclude: "core/matrix/csr_kernels.hpp"
+- "/fbcsr_kernels.template.*"
+    - FixInclude: "core/matrix/fbcsr_kernels.hpp"
```
yhmtsai, line 40:
```suggestion
- "*\.template\.*"
    - PathIgnore: "1"
    - PathPrefix: "core"
```
it should also work

yhmtsai (approved):
LGTM. some nit and wrong variable

cmake/template_instantiation.cmake
```diff
@@ -0,0 +1,68 @@
...
+        message(FATAL_ERROR "Nothing to split")
+    endif()
+    if (begin_location GREATER_EQUAL end_location)
+        message(FATAL_ERROR "Incorrect begin/end order")
+    endif()
+    # determine which lines belong to the header and footer
+    set(range_begins ${begin_location} ${split_locations})
+    set(range_ends ${split_locations} ${end_location})
+    list(LENGTH begin_locations range_count)
+    list(LENGTH split_locations range_count_minus_one)
```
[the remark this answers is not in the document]
yhmtsai: range_count is not used actually

cmake/template_instantiation.cmake
upsj: It seems like `STRINGS` does not properly escape semicolons, thus we lose them in the output. This is the reason I do this weird escaping step first

common/CMakeLists.txt
upsj: yes, we need to propagate it up two levels. BTW, this change is not strictly necessary, I removed the code that required it.

dev_tools/scripts/config
upsj: ah, does everything past the first `.` get ignored?

cuda/CMakeLists.txt
upsj: you mean the size of the binaries?

cmake/template_instantiation.cmake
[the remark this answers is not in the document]
upsj: good catch, it was used in a previous version of the implementation. I'll remove it

dev_tools/scripts/config
yhmtsai: Actually, it should be `.*\.template\..*`, and it only apply to the csr and fbcsr not common/unified .
the PathIgnore value is the number of folder levels ignored.

You can add the `- CoreSuffix: "\.template"` into the following two sections.

for unified
```
- "common/unified/.*.cpp"
    - PathIgnore: "2"
    - PathPrefix: "core"
```
and for normal kernels
```
- ".*"
    - PathPrefix: "core"
    - PathIgnore: "1"
```
script will ignore the `.template` such that it is like the orignal.

dpcpp/CMakeLists.txt
yhmtsai: ah, I see. I thought it is the similar file with csr and fbcsr.

cmake/template_instantiation.cmake
upsj: So my understanding of the dependencies is as follows:
* changing the `.instantiate.cpp` file causes both `cmake` to be called (which leads to the `.tmp` files being regenerated) as well as the `.instantiate.*.cpp` files to be regenerated by copying them from the `.tmp` files.
* changing the `.template.cpp` files causes all `.instantiate.*.cpp` files to be recompiled
* calling `cmake` does *not* cause recompilation, because changes to the `.tmp` files do not lead to the `.instantiate.*.cpp` files being recompiled.

Does this match your understanding of the code?

cmake/template_instantiation.cmake
```diff
@@ -0,0 +1,68 @@
+function(add_instantiation_files source_dir source_file output_files_var)
```
MarcelKoch, line 1: maybe instead of an output variable, we could pass in the target the sources should be added to. Then we would not need to add them manually.

MarcelKoch (approved):
Thanks for working on this. Anything that brings down compile time is appreciated. Since the comment came up, maybe it makes sense to add a CMake option to disable the splitting completely.

cmake/template_instantiation.cmake
upsj: That wouldn't be sufficient for HIP and CUDA, where we need to set some properties on the files on top of adding them to a target.

cmake/template_instantiation.cmake
```diff
@@ -0,0 +1,68 @@
...
+    string(REGEX REPLACE "[\r\n]" ";" file_contents "${file_contents}")
+    # find location of // begin|split|end comments
+    set(begin_location)
+    set(end_location)
+    set(split_locations)
+    list(LENGTH file_contents total_length)
+    set(counter 0)
+    foreach(line IN LISTS file_contents)
+        if(line MATCHES "// begin")
+            set(begin_location ${counter})
```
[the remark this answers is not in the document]
upsj: good point, I'll add it

dev_tools/scripts/config
upsj: thanks for the fix :)

upsj:
I added a switch to enable and disable the split: `GINKGO_SPLIT_TEMPLATE_INSTANTIATIONS`

cmake/template_instantiation.cmake
MarcelKoch: What do you mean by that? At least I can't find another place where the output variable is used except for `target_sources`.

upsj: https://github.com/ginkgo-project/ginkgo/blob/dbd43cf0624fba5fb8fe7d8850955180f0e1e7c6/cuda/CMakeLists.txt#L60

MarcelKoch: I guess we could get the files from the target's `SOURCES` property. But maybe that is not worth the trouble.

upsj: Yes, with native HIP support we could consider it

PR #1379, opened 2023-08-04
Add a batch::BatchLinOp hierarchy and core tests

pratikvn:
This PR adds the BatchLinOp hierarchy to be used for the batched matrix formats, solvers and preconditioners. As discussed this hierarchy is separate from the LinOp hierarchy.

The multi-vector apply, `apply(MultiVector* b, MultiVector* x)` will be handled by the concrete classes for now as this simplifies code as `MultiVector` is not a `BatchLinOp`, therefore needs special treatment in terms of either using variants, type-erased classes or a base class, due to having a `ValueType`.

If we figure out a cleaner approach in the future, we can create adopt a variant/type-erased approach to handle this behaviour throughout.


include/ginkgo/core/base/batch_lin_op.hpp
```diff
@@ -0,0 +1,439 @@
...
+     *
+     * Performs the operation x = op(b), where op is this batch linear operator.
+     *
+     * @param b  the input batch vector(s) on which the batch operator is
+     *           applied
+     * @param x  the output batch vector(s) where the result is stored
+     *
+     * @return this
+     */
+    BatchLinOp* apply(ptr_param<const BatchLinOp> b, ptr_param<BatchLinOp> x)
```
MarcelKoch, line 106: I guess this has not been updated with the batch::MultiVector in mind. The question is do we need to have both apply functions, one applying to vectors and one applying to general linops?

MarcelKoch, line 106: I don't get why we have a non-const apply. First, I think it makes more sense to only have a const version. And second, we don't have a non-const apply_impl, so even if a linop implementation would want to mutate some internal data during the apply, it couldn't.

MarcelKoch, line 106: Perhaps I'm opening another can of worms, but do we actually want to return something from the apply? I fail to imagine that this returns anything else than the object that apply is called on, so there seems very little value in the return type. If we wouldn't use it, we can skip the `EnableLinOp` stuff.

include/ginkgo/core/base/exception_helpers.hpp
```diff
@@ -147,6 +148,22 @@ inline dim<2> get_size(const T& op)
 inline dim<2> get_size(const dim<2>& size) { return size; }
 
 
+template <typename T>
+inline batch_dim<2> get_batch_size(const T& op)
```
MarcelKoch, line 152: maybe just split this into two, one for the multivector, and one for linops. That would be a bit safer.

core/test/base/batch_lin_op.cpp
```diff
@@ -0,0 +1,405 @@
...
+                        public gko::EnableCreateMethod<DummyBatchLinOp> {
+public:
+    DummyBatchLinOp(std::shared_ptr<const gko::Executor> exec,
+                    gko::batch_dim<2> size = gko::batch_dim<2>{})
+        : gko::batch::EnableBatchLinOp<DummyBatchLinOp>(exec, size)
+    {}
+
+    void access() const { last_access = this->get_executor(); }
+
+    mutable std::shared_ptr<const gko::Executor> last_access;
```
MarcelKoch, line 62: I'm not sure what this is going to be used for. Since this is a core test, this can only either be null, or a ReferenceExecutor. So I guess a boolean would be better suited to check if it was accessed.

core/test/base/batch_lin_op.cpp
```diff
@@ -0,0 +1,405 @@
...
+TEST_F(EnableBatchLinOp, ExtendedApplyFailsOnWrongBetaDimension)
+{
+    auto wrong =
+        DummyBatchLinOp::create(ref, gko::batch_dim<2>(1, gko::dim<2>{2, 5}));
+
+    ASSERT_THROW(op->apply(alpha, b, wrong, x), gko::DimensionMismatch);
+}
+
+
+TEST_F(EnableBatchLinOp, ApplyDoesNotCopyBetweenSameMemory)
```
MarcelKoch, line 286: I think this and the following test are better done by using a logger to check if a copy event was issued.

include/ginkgo/core/base/batch_lin_op.hpp
MarcelKoch: I think this is actually a more difficult problem. The multi-vector is templated, so we can't just use it as a parameter for the virtual functions. So we would probably need to introduce a non-templated base class for the multi-vector. The only other alternative I can think of right now would be using variants. But that might lead to more issues than it's worth.

pratikvn: I think that is another discussion we should have at some point, but the aim was to allow chaining. `linop->apply(b1,x)->apply(b2,x)`. I think it is a nice feature to have. I would recommend keeping it for now and having a discussion later.

include/ginkgo/core/base/exception_helpers.hpp
pratikvn: Can you elaborate on what you mean by safer ? It would give a compile time error if `get_size` is not available, so I dont think there are any issues there ?

include/ginkgo/core/base/batch_lin_op.hpp
pratikvn: What do you think about circumventing this whole thing and having only the concrete LinOp/BatchLinOp handling the apply with `MultiVector` parameters ? I think that significantly simplifies things without too much code in the concrete classes.

MarcelKoch: I'm not sure I follow. So the `BatchLinOp` would not have an apply at all?

pratikvn: `BatchLinOp` has apply only to `BatchLinOp`, but the `apply(MultiVector<> b ... )` is only in the concrete classes, `batch::Dense`, `batch::Csr` etc.

MarcelKoch: For me that seems very odd, since now there is a linear operator that can't be applied to vectors, which is the whole definition of a linear operator. Also, you could not use `BatchLinOp` in solvers or similar (although that might not be as important as for the `LinOp`).

core/test/base/batch_lin_op.cpp
pratikvn: I think these tests are not relevant as they mainly test behaviour of `make_temporary_clone`, so I have removed them.

include/ginkgo/core/base/batch_lin_op.hpp
MarcelKoch: Will the `apply(BatchLinOp*, BatchLinOp*)` be needed in the future? I think an argument can also be made to drop that entirely, because it only has very limited use, eg. SpGEAM but only for CSR. So maybe just remove it and the BatchLinOp only has it's size.

include/ginkgo/core/base/exception_helpers.hpp
MarcelKoch: IMO, the compiler error will be easier to understand if there are only two overloads. Otherwise you might even get linker instead of compiler errors.

include/ginkgo/core/base/batch_lin_op.hpp
pratikvn: It would be used for GEMM, which exists for all formats, and maybe also for user created LinOps ?

upsj: I think we wanted to make a clear distinction between Dense used as a vector and as a matrix, see #796
In that setting, GEMM should be a separate function, not an `apply`.

upsj: No, since you can only pass a vector to apply, it cannot be used to call GEMM.

MarcelKoch: Perhaps to get back to my point, I don't think we will have any meaning full implementation of `appy(BatchLinOp*, BatchLinOp*)`, so we can just remove it here.

pratikvn: So your suggestion would be to have a separate `gemm`, `geam` in all matrix formats instead ?

MarcelKoch: TBH, for me there is no difference between GEMM and `apply(MultiVector*, MultiVector*)`, or I'm misunderstanding something.

upsj: > So your suggestion would be to have a separate gemm, geam in all matrix formats instead ?

No, my suggestion would be to only add specific functions if we need them and they are feasible to implement. Most sparse storage formats are not well-suited for implementing gemm or geam.

> TBH, for me there is no difference between GEMM and apply(MultiVector*, MultiVector*), or I'm misunderstanding something.

If I say GEMM I mean dense * dense, not dense * multivector. The latter can of course use GEMM internally.

pratikvn: The underlying data might be the same (this is only the case for dense though), but the semantics are different and you usually have dense objects that you call gemm on, and if you want to use the multivector apply, then you will have to create new multivector objects for those. 

For SpGEMM and SpGEAM, that is not the case, and is not representable with the multivector apply.

pratikvn:
format!

pratikvn:
format!

include/ginkgo/core/base/lin_op.hpp
```diff
@@ -43,6 +43,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/base/device_matrix_data.hpp>
 #include <ginkgo/core/base/dim.hpp>
 #include <ginkgo/core/base/exception_helpers.hpp>
+#include <ginkgo/core/base/lin_op_helpers.hpp>
```
MarcelKoch, line 46: I think all these changes will clash with other PRs, so perhaps it's best to leave them out.

MarcelKoch (approved):
LGTM, I would keep the changes to lin_op.hpp out.

include/ginkgo/core/base/lin_op.hpp
pratikvn: It reduces the duplication quite a bit, particularly for the macros for Factory parameters, so I would say it is necessary. 

But, I think the Factory parameters PR might be merged before this PR and I will rebase it after that anyway and see if these changes still make sense then.

upsj: IIRC there I move most of the factory macros to `abstract_factory.hpp`, so that should help

pratikvn: No changes in lin_op now.

core/test/base/batch_lin_op.cpp
```diff
@@ -0,0 +1,164 @@
...
+};
+
+
+class EnableBatchLinOp : public ::testing::Test {
+protected:
+    EnableBatchLinOp()
+        : ref{gko::ReferenceExecutor::create()},
+          ref2{gko::ReferenceExecutor::create()},
+          op{DummyBatchLinOp::create(ref2,
+                                     gko::batch_dim<2>(1, gko::dim<2>{3, 5}))},
```
yhmtsai, line 68:
```suggestion
          op{DummyBatchLinOp::create(ref,
                                     gko::batch_dim<2>(1, gko::dim<2>{3, 5}))},
```

yhmtsai: Are the second op2 and ref2 necessary for test? there's no interaction between two op.

core/test/base/batch_lin_op.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+    DummyBatchLinOpWithFactory(const Factory* factory,
+                               std::shared_ptr<const gko::batch::BatchLinOp> op)
+        : gko::batch::EnableBatchLinOp<DummyBatchLinOpWithFactory>(
+              factory->get_executor()),
+          parameters_{factory->get_parameters()},
+          op_{op}
+    {}
+
+    std::shared_ptr<const gko::batch::BatchLinOp> op_;
```
yhmtsai, line 120:
```suggestion
    std::shared_ptr<const gko::batch::BatchLinOp> op;
```
nit: because it is a public member

include/ginkgo/core/base/batch_lin_op.hpp
```diff
@@ -0,0 +1,386 @@
...
+    /**
+     * Creates a batch operator storing items of uniform sizes.
+     *
+     * @param exec        the executor where all the operations are performed
+     * @param batch_size  the size the batched operator, as a batch_dim object
+     */
+    explicit BatchLinOp(std::shared_ptr<const Executor> exec,
+                        const batch_dim<2>& batch_size)
+        : EnableAbstractPolymorphicObject<BatchLinOp>(exec), size_{batch_size}
+    {}
```
upsj, line 141: These two constructors essentially do the same thing. If we want to have both of them, then one of them should be implemented in terms of the other, so we have less potential for inconsistency.
```suggestion
    explicit BatchLinOp(std::shared_ptr<const Executor> exec,
                        const size_type num_batch_items = 0,
                        const dim<2>& common_size = dim<2>{})
        : BatchLinOp{std::move(exec), num_batch_items > 0 ? batch_dim<2>(num_batch_items, common_size)
                                    : batch_dim<2>{}}
    {}

    /**
     * Creates a batch operator storing items of uniform sizes.
     *
     * @param exec        the executor where all the operations are performed
     * @param batch_size  the size the batched operator, as a batch_dim object
     */
    explicit BatchLinOp(std::shared_ptr<const Executor> exec,
                        const batch_dim<2>& batch_size)
        : EnableAbstractPolymorphicObject<BatchLinOp>(exec), size_{batch_size}
    {}
```

include/ginkgo/core/base/batch_lin_op.hpp
```diff
@@ -0,0 +1,386 @@
...
+class BatchLinOp : public EnableAbstractPolymorphicObject<BatchLinOp> {
+public:
+    /**
+     * Returns the number of items in the batch operator.
+     *
+     * @return number of items in the batch operator
+     */
+    size_type get_num_batch_items() const noexcept
+    {
+        return size_.get_num_batch_items();
```
upsj, line 91:
```suggestion
        return get_size().get_num_batch_items();
```

include/ginkgo/core/base/batch_lin_op.hpp
```diff
@@ -0,0 +1,386 @@
...
+    }
+
+    /**
+     * Returns the common size of the batch items.
+     *
+     * @return the common size stored
+     */
+    dim<2> get_common_size() const { return size_.get_common_size(); }
```
upsj, line 99:
```suggestion
    dim<2> get_common_size() const { return get_size().get_common_size(); }
```

include/ginkgo/core/base/batch_lin_op.hpp
```diff
@@ -0,0 +1,386 @@
...
+template <typename ConcreteBatchLinOp, typename PolymorphicBase = BatchLinOp>
+class EnableBatchLinOp
+    : public EnablePolymorphicObject<ConcreteBatchLinOp, PolymorphicBase>,
+      public EnablePolymorphicAssignment<ConcreteBatchLinOp> {
+public:
+    using EnablePolymorphicObject<ConcreteBatchLinOp,
+                                  PolymorphicBase>::EnablePolymorphicObject;
+
+protected:
+    GKO_ENABLE_SELF(ConcreteBatchLinOp);
```
upsj, line 239: This doesn't seem necessary any longer, after `apply` has been removed

include/ginkgo/core/base/batch_lin_op.hpp
```diff
@@ -0,0 +1,386 @@
...
+ * BatchLinOpFactory. A function of this class is to provide a generate method,
+ * which internally cals the generate_impl(), which the concrete BatchLinOps
+ * have to implement.
+ *
+ * Example: using BatchCG in Ginkgo
+ * ---------------------------
+ *
+ * ```c++
+ * // Suppose A is a batch matrix, batch_b a batch rhs vector, and batch_x an
+ * // initial guess
```
yhmtsai, line 162:
```suggestion
 * // Suppose A is a batch matrix, batch_b is a batch rhs vector, and batch_x is an
 * // initial guess
```

include/ginkgo/core/log/logger.hpp
```diff
@@ -58,6 +58,13 @@ class PolymorphicObject;
 class Operation;
 class stopping_status;
 
+
+namespace batch {
+class BatchLinOp;
+class BatchLinOpFactory;
+}  // namespace batch
```
yhmtsai, line 65:
```suggestion
namespace batch {


class BatchLinOp;
class BatchLinOpFactory;


}  // namespace batch
```

include/ginkgo/core/base/exception_helpers.hpp
```diff
@@ -147,6 +148,22 @@ inline dim<2> get_size(const T& op)
...
+{
+    return op->get_size();
```
yhmtsai, line 154:
```suggestion
    return op->get_size();
```
should it still use `get_batch_size()`?

upsj (approved):
LGTM, only minor consistency suggestions. BatchLinOp should get an `apply(MultiVectorBase, MultiVectorBase)` function soon though, otherwise the type has little utility on its own. OTOH, I think that much of what makes polymorphism useful in Ginkgo does not really apply to the batch case, since we always need to instantiate the fused kernels for the exact solver/matrix/preconditioner/type combinations we need.

include/ginkgo/core/base/batch_lin_op.hpp
```diff
@@ -0,0 +1,386 @@
...
+ * have no communication/information exchange between them. Therefore, any
+ * collective operations between the batches is not possible and not
+ * implemented. This allows for each batch to be computed and operated on in an
+ * embarrassingly parallel fashion.
+ *
+ * A key difference between the LinOp and the BatchLinOp class is that the apply
+ * between BatchLinOps is no longer supported. The user can apply a BatchLinOp
+ * to a batch::MultiVector but not to any general BatchLinOp. Therefore, the
+ * BatchLinOp serves only as a base class providing necessary core functionality
+ * from Polymorphic object and store the dimensions of the batched object.
```
yhmtsai, line 78: also mention the apply currently handled by the concrete type.

include/ginkgo/core/base/batch_lin_op.hpp
```diff
@@ -0,0 +1,386 @@
...
+     * @param size to be set
+     */
+    void set_size(const batch_dim<2>& size) { size_ = size; }
+
+    /**
+     * Creates a batch operator storing items of uniform sizes.
+     *
+     * @param exec        the executor where all the operations are performed
+     * @param num_batch_items the number of batch items to be stored in the
+     * operator
```
yhmtsai, line 121:
```suggestion
     * @param num_batch_items  the number of batch items to be stored in the
     * operator
```

core/test/base/batch_lin_op.cpp
```diff
@@ -0,0 +1,164 @@
...
+        DummyBatchLinOp::create(ref, gko::batch_dim<2>(1, gko::dim<2>{3, 5})));
+    auto factory = DummyBatchLinOpWithFactory<>::build().with_value(6).on(ref);
+
+    auto op = factory->generate(dummy);
+
+    ASSERT_EQ(op->get_executor(), ref);
+    ASSERT_EQ(op->get_parameters().value, 6);
+    ASSERT_EQ(op->op_.get(), dummy.get());
+}
+
```
yhmtsai, line 162: miss FactoryGenerateIsLogged, CopiesLinOpToOtherExecutor testing because BatchLinOpFactory has the generate function and logger.

yhmtsai (changes requested):
TBH, I think BatchLinOp should have the signature for `apply(batch::MultiVectorBase...)`
the BatchLinOp only serves the base class with some `get_size` function.
There's no much from EnableBatchLinOp.
Because it will not be the interface break in the future, I do not request the changes from the apply signature now but for the missing tests.

pratikvn:
format!

core/test/base/batch_lin_op.cpp
pratikvn: I think it is okay in the tests here as it makes it easier to differentiate that it is a member of the `op` object.

include/ginkgo/core/base/exception_helpers.hpp
pratikvn: No, the BatchLinOp getter is also `get_size()` and it returns a `batch_dim` object.

include/ginkgo/core/base/batch_lin_op.hpp
```diff
@@ -0,0 +1,388 @@
...
+ * std::cout << my_op->get_my_parameters().my_value;  // prints 0
+ * ```
+ *
+ * @note It is possible to combine both the #GKO_CREATE_FACTORY_PARAMETER_*()
+ * macros with this one in a unique macro for class __templates__ (not with
+ * regular classes). Splitting this into two distinct macros allows to use them
+ * in all contexts. See <https://stackoverflow.com/q/50202718/9385966> for more
+ * details.
+ *
+ * @param _lin_op  concrete operator for which the factory is to be created
```
tcojean, line 333:
```suggestion
 * @param _batch_lin_op  concrete operator for which the factory is to be created
```

include/ginkgo/core/base/batch_lin_op.hpp
```diff
@@ -0,0 +1,388 @@
...
+};
+
+
+/**
+ * This is an alias for the EnableDefaultFactory mixin, which correctly sets the
+ * template parameters to enable a subclass of BatchLinOpFactory.
+ *
+ * @tparam ConcreteFactory  the concrete factory which is being implemented
+ *                          [CRTP parameter]
+ * @tparam ConcreteLinOp  the concrete BatchLinOp type which this factory
```
tcojean, line 251:
```suggestion
 * @tparam ConcreteBatchLinOp  the concrete BatchLinOp type which this factory
```

include/ginkgo/core/base/batch_lin_op.hpp
```diff
@@ -0,0 +1,388 @@
...
+ * produces, needs to have a constructor which takes a const ConcreteFactory *,
+ * and an std::shared_ptr<const BatchLinOp> as parameters.
+ * @tparam ParametersType  a subclass of enable_parameters_type template which
+ *                         defines all of the parameters of the factory
+ * @tparam PolymorphicBase  parent of ConcreteFactory in the polymorphic
+ *                          hierarchy, has to be a subclass of LinOpFactory
```
tcojean, line 257:
```suggestion
 *                          hierarchy, has to be a subclass of BatchLinOpFactory
```

tcojean (approved):
LGTM. Minor documentation issues.

include/ginkgo/core/base/batch_lin_op.hpp
```diff
@@ -0,0 +1,388 @@
...
+ * have no communication/information exchange between them. Therefore, any
+ * collective operations between the batches is not possible and not
+ * implemented. This allows for each batch to be computed and operated on in an
+ * embarrassingly parallel fashion.
+ *
+ * A key difference between the LinOp and the BatchLinOp class is that the apply
+ * between BatchLinOps is no longer supported. The user can apply a BatchLinOp
+ * to a batch::MultiVector but not to any general BatchLinOp. This apply to a
+ * batch::MultiVector is handled by the concrete LinOp and may be moved to the
+ * base BatchLinOp class in the future.
```
tcojean, line 78: While having this in the text is good, having it concisely as a note would be better. It's a really important point that needs to be highlighted and may confuse people. Maybe:
`@note Currently the apply method is not in the BatchLinOp base class and only applies to batch::MultiVector.`

PR #1382, opened 2023-08-07
Remove CUDA 9.2 support

upsj:
Pulled out of #1368, this removes CUDA 9.2 support and the corresponding workarounds

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ For Ginkgo core library:
 
 The Ginkgo CUDA module has the following __additional__ requirements:
 
-*   _CUDA 9.2+_ or _NVHPC Package 22.7+_
+*   _CUDA 10.1+_ or _NVHPC Package 22.7+_
 *   Any host compiler restrictions your version of CUDA may impose also apply
     here. For the newest CUDA version, this information can be found in the
     [CUDA installation guide for Linux](https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html)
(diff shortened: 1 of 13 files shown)


yhmtsai (approved):

pratikvn (commented):
Can we enforce from CMake that the CMAKE_CUDA_COMPILER_VERSION is >10,1. For example checking with `CMAKE_CUDA_COMPILER_VERSION` and exiting if not correct ? 

Can we remove the setting of the following explicit flags we needed to add: `--expt-relaxed-constexpr`, `--expt-extended-lambda`  ?

upsj:
@pratikvn We already have version requirements listed in README, do we need to be more explicit? Anyways, that should be the subject of another PR.

upsj:
AFAIK the lambda/constexpr flags are still required even for CUDA 12.1

pratikvn:
I think it is better to realize during CMake build time that a later CUDA version is necessary rather than having a weird warning thrown (with no indication that the issue is due to the older CUDA version) during compile time. People might not necessarily read the README.

MarcelKoch:
@pratikvn I think the cmake update #1368 would be necessary to easily check the cuda version. So it could be part of that PR.

README.md
```diff
@@ -58,7 +58,7 @@ The Ginkgo HIP module has the following __additional__ requirements:
 * _ROCm 4.5+_
 *    the HIP, hipBLAS, hipSPARSE, hip/rocRAND and rocThrust packages compiled with either:
     * _AMD_ backend (using the `clang` compiler)
-    * _9.2 <= CUDA < 11_ backend
```
tcojean, line 61: As a sidenote, we would need to check but on the hipSPARSE side they keep adding ifdefs for larger CUDA version. Maybe they caught up to 12 in the last versions.

https://github.com/ROCmSoftwarePlatform/hipSPARSE/blob/develop/library/src/nvidia_detail/hipsparse.cpp#L974

tcojean (approved):
LGTM.

pratikvn (approved):

PR #1385, opened 2023-08-10
Allow specifying allocator for benchmarks

upsj:
This adds an `-allocator` flag to specify how to allocate data in the benchmarks.

### DIFF
--- a/benchmark/sparse_blas/sparse_blas.cpp
+++ b/benchmark/sparse_blas/sparse_blas.cpp
@@ -127,9 +127,12 @@ void apply_sparse_blas(const char* operation_name,
                               allocator);
             auto gen_logger = create_operations_logger(
                 FLAGS_gpu_timer, FLAGS_nested_names, exec,
-                test_case[operation_name]["components"], allocator, 1);
+                test_case[operation_name]["components"], allocator,
+                repetitions);
             exec->add_logger(gen_logger);
-            op->run();
+            for (unsigned i = 0; i < repetitions; i++) {
+                op->run();
+            }
             exec->remove_logger(gen_logger);
         }
         op->write_stats(test_case[operation_name], allocator);
--- a/benchmark/utils/general.hpp
+++ b/benchmark/utils/general.hpp
@@ -45,6 +45,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ostream>
 #include <random>
 #include <sstream>
+#include <stdexcept>
 #include <string>
 #include <type_traits>
 #include <utility>
@@ -58,6 +59,9 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <rapidjson/prettywriter.h>
 
 
+#include <ginkgo/core/base/memory.hpp>
+
+
 #include "benchmark/utils/json.hpp"
 #include "benchmark/utils/timer.hpp"
 #include "benchmark/utils/types.hpp"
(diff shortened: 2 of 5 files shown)


cuda/base/memory.cpp
```diff
@@ -33,6 +33,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/base/memory.hpp>
 
 
+#include <cuda.h>
```
upsj, line 36: CUDA_VERSION was undefined previously

upsj:
format-rebase!

benchmark/utils/general.hpp
```diff
@@ -329,6 +337,40 @@ void backup_results(rapidjson::Document& results)
 }
 
 
+inline std::shared_ptr<gko::CudaAllocatorBase> create_cuda_allocator()
+{
+    std::string flag{FLAGS_allocator};
+    if (flag == "default") {
+        return std::make_shared<gko::CudaAllocator>();
+    } else if (flag == "async") {
```
pratikvn, line 345: Is there any point in having an async allocator if you are not using a stream ?

upsj: The default stream is also a stream, the stream-ordered semantics of the allocator still apply

pratikvn: Does stream==nullptr mean that it is the default stream ? I cannot seem to find it in the docs anywhere.

upsj: It's listed as 0 in there

pratikvn (approved):

benchmark/sparse_blas/sparse_blas.cpp
```diff
@@ -127,9 +127,12 @@ void apply_sparse_blas(const char* operation_name,
...
             auto gen_logger = create_operations_logger(
                 FLAGS_gpu_timer, FLAGS_nested_names, exec,
-                test_case[operation_name]["components"], allocator, 1);
+                test_case[operation_name]["components"], allocator,
+                repetitions);
             exec->add_logger(gen_logger);
-            op->run();
+            for (unsigned i = 0; i < repetitions; i++) {
+                op->run();
+            }
```
yhmtsai, line 135: should the repetitions be in detail?

benchmark/utils/general.hpp
yhmtsai: but the default stream does not allow other streams created by `cudaStreamNonBlocking` execute at the same time, right?

core/device_hooks/cuda_hooks.cpp
```diff
@@ -75,6 +75,10 @@ bool CudaAsyncAllocator::check_environment(int device_id,
     GKO_NOT_COMPILED(cuda);
 
 
+CudaUnifiedAllocator::CudaUnifiedAllocator(int device_id)
```
yhmtsai, line 78: Why did it not throw undefined reference before?

benchmark/sparse_blas/sparse_blas.cpp
upsj: We do the same for preconditioners, and it's useful to have reliable timings for the different components in a runtime breakdown

core/device_hooks/cuda_hooks.cpp
upsj: Because we did not call that constructor anywhere, we used the one with `flags` IIRC

benchmark/sparse_blas/sparse_blas.cpp
yhmtsai: We do not do the same for solver. It will be slower. but I am fine with both.

yhmtsai (approved):

benchmark/utils/general.hpp
upsj: yes, but the benchmarks only use a single stream, so we are fine for now. As soon as we move to multi-stream setups, we need to use a custom stream or `CU_STREAM_PER_THREAD`

PR #1387, opened 2023-08-14
Allow passing matrix filename as benchmark input

upsj:
When running benchmarks by hand, I always have to type `-input '[{"filename":"..."}'` by hand. The MatrixMarket and Ginkgo binary formats can be differentiated from JSON files by the first character: `%` is MM, `G` is binary, everything else is potentially JSON. So I had the idea to pass the matrix filename directly as input for the benchmark, either with the same flag (like in this PR) or with a different flag.

Not sure whether this overloads the `-input` flag too much, but I wanted to suggest it anyways.

### DIFF
--- a/benchmark/test/reference/spmv.matrix.stderr
+++ b/benchmark/test/reference/spmv.matrix.stderr
@@ -1,0 +1,31 @@
+This is Ginkgo 1.7.0 (develop)
+    running with core module 1.7.0 (develop)
+Running on reference(0)
+Running with 2 warm iterations and 10 running iterations
+The random seed for right hand sides is 42
+The formats are coo
+The number of right hand sides is 1
+Running test case
+{
+    "filename": "../../matrices/test/ani1.mtx",
+    "spmv": {}
+}
+Matrix is of size (36, 36)
+Current state:
+[
+    {
+        "filename": "../../matrices/test/ani1.mtx",
+        "spmv": {
+            "coo": {
+                "storage": 3328,
+                "max_relative_norm2": 1.0,
+                "time": 1.0,
+                "repetitions": 10,
+                "completed": true
+            }
+        },
+        "size": 36,
+        "nnz": 208,
+        "optimal": {}
+    }
+]
--- a/benchmark/test/reference/spmv.matrix.stderr
+++ b/benchmark/test/reference/spmv.matrix.stderr
@@ -7,14 +7,14 @@ The formats are coo
 The number of right hand sides is 1
 Running test case
 {
-    "filename": "../../matrices/test/ani1.mtx",
+    "filename": "",
     "spmv": {}
 }
 Matrix is of size (36, 36)
 Current state:
 [
     {
-        "filename": "../../matrices/test/ani1.mtx",
+        "filename": "",
         "spmv": {
             "coo": {
                 "storage": 3328,
--- a/benchmark/utils/general_matrix.hpp
+++ b/benchmark/utils/general_matrix.hpp
@@ -67,9 +67,17 @@ void initialize_argument_parsing_matrix(
                 << "-input and -input_matrix cannot be used simultaneously\n";
             std::exit(1);
         }
-        auto input_json = R"([{"filename":")" + input_matrix_str + "\"" +
-                          additional_matrix_file_json + "}]";
-        input_stream = std::make_unique<std::stringstream>(input_json);
+        // create JSON for the filename via RapidJSON to ensure the string is
+        // correctly escaped
+        rapidjson::Document d;
+        auto json_template =
+            R"([{"filename":"")" + additional_matrix_file_json + "}]";
+        d.Parse(json_template.c_str());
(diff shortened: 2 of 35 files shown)


MarcelKoch:
I don't know if `-input` was part of the last release, but perhaps just rename that to `-input_matrix` and expect it to be a filename. There may not be much benefit of `-input [{'filename': '...'}]` over `echo [{filename': '...'}] | benchmark`.

MarcelKoch:
From some stackoverflow (https://stackoverflow.com/a/8422306) it seems like you can do this:
```
> gdb ./exec
> run < input.json
```
And also, perhaps it's enough for the debugging to have one input matrix.

upsj:
That doesn't allow changing the parameters in a debugging session, since you cannot do `echo ... | run`.

MarcelKoch:
you can do 
```
echo ... > input.json
gdb ./exec
run < input.json
```
Or do you want to change stdin during the `run`?

MarcelKoch:
In newer python version (>=3.7) you can just pass a string to `subprocess.run(args, input=some_input)`. But I might not understand your workflow anyway.
Still, I think it would be better to have this parameter for the very simple case of using exactly one input file.

upsj:
I simply don't see a reason to remove the functionality when it can make scripting things much easier. Running benchmarks for multiple BLAS parameters on multiple GPUs using `parallel` is another example case where
`parallel command -input ... :: parameters` is simpler than `parallel bash -c 'echo ... | command' :: parameters`

MarcelKoch:
TBH, I'm still not sure if it makes scripting easier. But anyway, perhaps just add this under a new parameter name, to not overload the old one.

upsj:
The reason I reused the name is purely practical: The parameter would either need to be put into `general.hpp` and thus be inapplicable for blas/multivector/matrix_generator or be duplicated into individual functions or put into an entirely separate header. I'll use `-input_matrix`

benchmark/utils/general_matrix.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+#include <gflags/gflags.h>
+
+
+#include "benchmark/utils/general.hpp"
+
+
+DEFINE_string(input_matrix, "",
+              "Filename of a matrix to be used as the single input. Overwrites "
+              "the value of the -input flag");
```
yhmtsai, line 48: I am considering `input` ignores `input_matrix`

benchmark/utils/general_matrix.hpp
```diff
@@ -0,0 +1,72 @@
...
+ *                                     will be used as input for the benchmark
+ *                                     if the `-input_matrix` flag is used.
+ */
+void initialize_argument_parsing_matrix(
+    int* argc, char** argv[], std::string& header, std::string& format,
+    std::string additional_matrix_file_json = "")
+{
+    initialize_argument_parsing(argc, argv, header, format);
+    std::string input_matrix_str{FLAGS_input_matrix};
+    if (!input_matrix_str.empty()) {
```
yhmtsai, line 64: another way is to change the `FLAGS_input` here such that the get_input_stream is not changed, but I do not like changing the user input state.

yhmtsai (approved):

benchmark/test/reference/spmv.matrix.stderr
```diff
@@ -0,0 +1,31 @@
...
+The random seed for right hand sides is 42
+The formats are coo
+The number of right hand sides is 1
+Running test case
+{
+    "filename": "../../matrices/test/ani1.mtx",
+    "spmv": {}
+}
+Matrix is of size (36, 36)
+Current state:
```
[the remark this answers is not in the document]
upsj: This is how it will be after #1323, at the moment, there is still JSON in both stderr and stdout

benchmark/utils/general_matrix.hpp
upsj: We cannot use both flags at the same time. Either one overrides the other, or we throw an error if they are used simultaneously.

benchmark/utils/general_matrix.hpp
```diff
@@ -0,0 +1,72 @@
...
+        auto input_json = "[{\"filename\":\"" + input_matrix_str + "\"" +
```
MarcelKoch, line 65: again, R-string?

benchmark/utils/general_matrix.hpp
```diff
@@ -0,0 +1,72 @@
...
+                          additional_matrix_file_json + "}]";
+        input_stream = std::make_unique<std::stringstream>(input_json);
+    }
+}
+
+
+#endif  // GKO_BENCHMARK_UTILS_GENERAL_MATRIX_HPP_
```
MarcelKoch, line 72:
```suggestion
#endif  // GKO_BENCHMARK_UTILS_GENERAL_MATRIX_HPP_

```

MarcelKoch (approved):
LGTM.

benchmark/utils/general_matrix.hpp
upsj: I added an error message

PR #1388, opened 2023-08-14
Fix circular dep check by using the same preference Threads 

yhmtsai:
It fixes the compilation error in the circular dependence check.

The issue happens when Threads uses the `-lpthread` not `-pthread`.
FindHIP also uses FindCUDA from cmake but faces the compilation issue.
Also, the hip link is `target_link_libraries(ginkgo_hip PUBLIC /usr/local/cuda/lib64/libcudart_static.a;Threads::Threads;dl;/usr/lib/x86_64-linux-gnu/librt.so )` from FindCUDA in HIP but I think cmake does not guarantee the linking order is the same as providing. 
Either `-pthread /usr/.../libcudart_static.a` or `/usr/.../libcudart_static.a -lpthread` works in hip test linking.
It's not the issue in cuda because cuda has another `-lpthread` in the end.

Compilation twice solves the issue because we have set `set(THREADS_PREFER_PTHREAD_FLAG ON)`, then it will change the Threads in the second round.
Note. It's hard to reproduce because it usually uses `-pthread` in the end for the first compile, unlike the CI job. Maybe we do not find the actual reason yet.

The first generations and second generations give different Threads target.
It can be found by the several texts after `get_target_property(GINKGO_LIBS_INTERFACE_DEFS Threads::Threads` by using `--debug-find --trace-expand` with `cmake` command

TODO:
- [x] revert the job dependency

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,8 @@ cmake_minimum_required(VERSION 3.16)
 project(Ginkgo LANGUAGES C CXX VERSION 1.7.0 DESCRIPTION "A numerical linear algebra library targeting many-core architectures")
 set(Ginkgo_VERSION_TAG "develop")
 set(PROJECT_VERSION_TAG ${Ginkgo_VERSION_TAG})
+# Cuda and Hip also look for Threads. Set it before any find_package to ensure the Threads setting is not changed.
+set(THREADS_PREFER_PTHREAD_FLAG ON)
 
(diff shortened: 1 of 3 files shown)


upsj (approved):
LGTM! Maybe squash the commits to remove the temporary changes from the history

tcojean (approved):
LGTM

yhmtsai:
rebase!

PR #1390, opened 2023-08-15
WIP: Reuse license

MarcelKoch:
This PR changes how we handle our license declaration. It moves from the current block of license (~30 lines) to 3/4 lines as defined by https://reuse.software/. The new license header will now read as:
```
// SPDX-FileCopyrightText: 2017-2023 The Ginkgo authors
//
```

Additionally, a license is added to all files in our repository. This includes adding a CC-BY-4.0 license for all of our images. Since the images are binary files, an extra file with a `.license` suffix is created in all cases. 

The MatrixMarket files are properly attributed, but since these files should not be modified (see the license section in https://sparse.tamu.edu/about) this is handled in the `.reuse/dep5` file. We could also do the same as with the images for those.

Some files have a CC0-1.0 license which basically means no copyright. The files are the `*.stderr` and `*.stdout` in the benchmark tests, and the files `examples/*/doc/`.

Some files have dual license. These are `FindHWLOC.cmake` (MORSE + our) and `external-lib-interfacing.cpp` (deal.II + our). I'm currently using an `AND` clause to combine the licenses, which means that a user must comply with all licenses simultaneously. The alternative is an `OR` clause, which would mean that the user can choose a license.

For now I've disabled our pre-commit license check, and instead added a github action for that.

### Todo
- [ ] enable pre-commit check for license again
- [ ] figure out the license of the matrices in `matrices/test`
- [ ] switch to CC0 license for files with little originality, eg gitignore, clang-format, etc

### Benefits of this PR
- Removes ~30k LOC which is just our license
- Fewer 'ignored' lines in upcoming PRs
- Clearly states the license for all files (no ambiguity)


LICENSE.md
```diff
@@ -0,0 +1,82 @@
...
+**The source code of Ginkgo is released under the terms of the BSD 3-Clause "New" or "Revised" License**.
+However, the licensing terms for assets, input files, and included third-party software may differ.
+All licenses used by this project are stored in the `LICENSES` directory.
+
+If you want to use Ginkgo, but need a different license, please
+[contact us](https://github.com/ginkgo-project/ginkgo/wiki#contact-us).
+
+## Third-Party Licenses
+
+Ginkgo source distribution does not bundle any third-party software. However,
```
upsj, line 16: This is no longer true since AMD was added

upsj:
Maybe we can do this in two separate PRs? First replace the license by the SPDX header, then handle automation?

MarcelKoch:
@upsj This PR handles only the license replacement atm. It only disables our current handling, because that doesn't mix well. The automation (if wanted) should best be handled by [pre-commit](https://pre-commit.com/).

But as you noted, we need to adjust the license for files with very little originality.

upsj:
I mean shortening only the existing license headers, everything else is a bigger effort

MarcelKoch:
I will split this PR up into two parts:
1. changing only existing license (as @upsj suggested)
2. adding license to other files using the reuse automation.

I will close this PR in favor of the other two.

PR #1396, opened 2023-08-22
SYCL deprecation fixes

upsj:
Fixes the deprecation warnings for accessors to local memory

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -572,24 +572,6 @@ build/nocuda-nomixed/nompi/clang/omp/debug/static:
     BUILD_SHARED_LIBS: "OFF"
     MIXED_PRECISION: "OFF"
 
-build/dpcpp/2022-1/cpu/release/shared:
-  extends:
-    - .build_and_test_template
-    - .default_variables
-    - .quick_test_condition
-    - .use_gko-oneapi-cpu
-  variables:
-    CXX_COMPILER: "dpcpp"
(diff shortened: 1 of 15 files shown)


upsj:
I've decided the complexity involved in sub-devices just to get #1373 working on SYCL CPUs isn't worth it yet, so I'll be removing those changes, only fixing the deprecation warnings

pratikvn:
We should either remove bump the SYCL version support or create a special case for it, as this update [currently breaks for SYCL 2022-1](https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/6271510855).

upsj:
There are some additional issues with `cooperative_groups::ballot` using `reduce_over_group` slightly wrong, but that can be fixed easily

yhmtsai (commented):
Is the PR description old? I do not see the changes for device.
LGTM in code itself.
However, I would keep intel compiler 2022 as the 2023 does not work properly.
you can make local_accessor function to the old accessor with target::local when it is not supported

rscohn2:
These changes are needed to build ginkgo with icpx 2024.1 & later.

yhmtsai:
Thanks for updating.
ReadME version support is not updated.
you can also change the image `ginkgohub/oneapi:2024.0.2.20231213` to `ginkgohub/oneapi:latest`

pratikvn (approved):

yhmtsai (approved):

PR #1400, opened 2023-08-29
Rename `array` members to match standard library

upsj:
Instead of our custom `get_num_elems` and `get_data`, having names like `size` and `data` makes things much more familiar to C++ developers that are used to the standard library.


include/ginkgo/core/base/array.hpp
```diff
@@ -121,14 +121,27 @@ class const_array_view {
      *
      * @return the number of elements in the array view
      */
-    size_type get_num_elems() const noexcept { return num_elems_; }
+    size_type size() const noexcept { return num_elems_; }
+
+    /** @copydoc size() */
+    [[deprecated("use size() instead")]] size_type get_size() const noexcept
```
MarcelKoch, line 127: We need to deprecate the old function:
```suggestion
    [[deprecated("use size() instead")]] size_type get_num_elems() const noexcept
```
Same for the device_matrix_data

upsj: oops, right, thanks!

MarcelKoch (dismissed):
I like the changes, especially `get_num_elems` was a bit lengthy. The deprecation warning in `array` still has to be fixed.

include/ginkgo/core/base/matrix_data.hpp
```diff
@@ -484,14 +484,19 @@ struct matrix_data {
...
-    void ensure_row_major_order()
+    void sort_row_major()
     {
         std::sort(
             begin(nonzeros), end(nonzeros), [](nonzero_type x, nonzero_type y) {
                 return std::tie(x.row, x.column) < std::tie(y.row, y.column);
             });
     }
 
+    [[deprecated("Use sort_row_major() instead")]] void ensure_row_major_order()
```
MarcelKoch, line 495: there is still usage of this:
- this file line 263
- this file line 517
- test/factorization/cholesky_kernels.cpp line 314

MarcelKoch (changes requested):
I only checked if the old versions are called anywhere. Since it compiles and passes tests, I'm assuming that all instance have been correctly replaced.

include/ginkgo/core/base/device_matrix_data.hpp
```diff
@@ -175,14 +175,14 @@ class device_matrix_data {
      *
      * @return the number of stored elements of the matrix.
      */
-    size_type get_num_elems() const { return values_.get_num_elems(); }
```
MarcelKoch, line 178: No deprecation of old function.
But I think this is a more tricky case, since this class already has a `get_size`, which returns something different then `size`.
IMO, we can't rename it here, and should just leave the old name.

include/ginkgo/core/base/index_set.hpp
```diff
@@ -280,7 +280,7 @@ class index_set {
      *
      * @return  number of indices stored in the index set
      */
-    index_type get_num_elems() const { return this->num_stored_indices_; };
```
MarcelKoch, line 283: Same issue as with `device_matrix_data`.
But, TBH, I think the naming of this and `get_size` is confusing anyway.
For an index set that covers the interval `[0, n)` `get_size` returns `n`, but that doesn't need to correlate to the size of any stored arrays. So, to me, renaming it to `get_end`, or maybe just `end`, seems clearer.
As for `get_num_elems`, it's honestly really hard to tell what this is for. I think it is the sum over the length of all stored subsets. So I would assume it's the size of the array returned by `to_global_indices`, but judging from its implementation it's not??? I'm very confused...

MarcelKoch (changes requested):
I think the renaming in `device_matrix_data` and `index_set` are not appropriate.

include/ginkgo/core/base/device_matrix_data.hpp
upsj: I wasn't aware we have another .get_num_elems member function. I'm looking for a better name, maybe match it with `Csr::get_num_stored_elements()`?

nbeams:
I would like to point out that these changes will affect the MFEM integration (use of `get_const_data`); not sure about any other integrations. For MFEM, it would be simple to update, but could this PR be merged close to a tag or minor release if possible, to make it easier to check the Ginkgo version to decide which version to use?

include/ginkgo/core/base/index_set.hpp
MarcelKoch: Perhaps `get_num_global_indices` would describe this a bit better? @pratikvn any suggestions?

include/ginkgo/core/base/device_matrix_data.hpp
MarcelKoch: I think that is a good approach.

upsj:
I made the changes less intrusive (see PR description) and fixed the mistaken replacements.

include/ginkgo/core/base/device_matrix_data.hpp
MarcelKoch: I think renaming it to `get_num_stored_elements` would still be appropriate.

include/ginkgo/core/base/index_set.hpp
MarcelKoch: I still think this needs a better name.

MarcelKoch (changes requested):
LGTM. I would still like to see the STL versions at some point.

Missed replacement:
- omp/test/base/index_set.cpp (a lot of places in there)
- reference/test/base/index_set.cpp:332:348

include/ginkgo/core/base/index_set.hpp
upsj: get_superset_size?

upsj: I'll do this in a new PR

include/ginkgo/core/base/matrix_data.hpp
```diff
@@ -481,7 +489,7 @@ struct matrix_data {
      */
     void sum_duplicates()
     {
-        ensure_row_major_order();
+        sort_row_major();
```
yhmtsai, line 492:
```suggestion
        this->sort_row_major();
```

include/ginkgo/core/base/array.hpp
```diff
@@ -93,6 +93,14 @@ class const_array_view {
      *
      * @return the number of elements in the array view
      */
+    size_type get_size() const noexcept { return num_elems_; }
```
pratikvn, line 96: Maybe for uniformity and clarity, this internal member should also be renamed.
```suggestion
    size_type get_size() const noexcept { return size_; }
```

pratikvn (approved):
LGTM! I think with the `get_` prefix and only for array, this makes a lot more sense. Should we also replace the internal member of array `num_elems_` to `size_` as well ?

yhmtsai (commented):
code LGTM. but I would like to ensure the others are okay with this naming change

test/base/device_matrix_data_kernels.cpp
```diff
@@ -132,7 +132,7 @@ TYPED_TEST(DeviceMatrixData, CopyConstructsOnOtherExecutorCorrectly)
         this->exec->get_master(), device_data};
 
     ASSERT_EQ(device_data.get_size(), host_data.get_size());
-    ASSERT_EQ(device_data.get_num_elems(), host_data.get_num_elems());
+    ASSERT_EQ(device_data.get_size(), host_data.get_size());
```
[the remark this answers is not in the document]
upsj: great catch, thanks! I found the same kind of replacement errors in other places, but missed this one

examples/kokkos_assembly/kokkos_assembly.cpp
```diff
@@ -3,13 +3,15 @@
...
+#include <Kokkos_Core.hpp>
 #include <map>
 #include <string>
 
 
 #include <omp.h>
+
+
 #include <ginkgo/ginkgo.hpp>
-#include <Kokkos_Core.hpp>
```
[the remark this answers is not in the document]
upsj: I think we can handle that with #1358?

thoasm (approved):
LGTM!

include/ginkgo/core/base/array.hpp
```diff
@@ -569,12 +572,12 @@ class array {
      *
      * If the array is not assigned an executor, an exception will be thrown.
      *
-     * @param num_elems  the amount of memory (expressed as the number of
+     * @param size  the amount of memory (expressed as the number of
      *                   `value_type` elements) allocated on the Executor
```
yhmtsai, line 576: offset

yhmtsai (approved):

PR #1402, opened 2023-08-30
Fix memory_order invocations to be inline with C++20 changes

### LINKED ISSUE #1401: Use of C++17 Atomic Exchange Causes Compilation Error with C++20
I'm trying to compile ginkgo with C++20, and I get the following error due to [this line](https://github.com/ginkgo-project/ginkgo/blob/3d396152217bbb94de69926ae075e5e2a98b120c/omp/reorder/rcm_kernels.cpp#L235). 

```
error: ‘memory_order_acq_rel’ is not a member of ‘std::memory_order
```
This happens due to two things:
1. C++ made the memory order enum strongly typed in C++20 and hence changed the naming to prevent accidental usage.
2. The code in the link above [depends on](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html#index-_005f_005fatomic_005fcompare_005fexchange_005fn) implicit conversion of weakly-typed `enum` to `int`. 

A fix would be to replace all occurrences of memory order with the newer version for newer language modes. There are a couple of ways I can think of to approach this:
1. Define a memory order macro, something like
```
#ifdef CPP20
#define MEMORY_ORDER_RELAXED std::memory_order::relaxed
#else 
#define MEMORY_ORDER_RELAXED std::memory_order::memory_order_relaxed
#endif
```
2. Replace call sites of all functions with an `#ifdef` everywhere.

jbadwaik:
Fixes #1401

### DIFF
--- a/omp/reorder/rcm_kernels.cpp
+++ b/omp/reorder/rcm_kernels.cpp
@@ -235,8 +235,8 @@ struct UbfsLinearQueue {
 #define GKO_CMPXCHG_IMPL(ptr, ptr_expected, replace_with) \
     return __atomic_compare_exchange_n(                   \
         ptr, ptr_expected, replace_with, true,            \
-        std::memory_order::memory_order_acq_rel,          \
-        std::memory_order::memory_order_acquire);
+        static_cast<int>(std::memory_order_acq_rel),      \
+        static_cast<int>(std::memory_order_acquire));
 #endif
 
 /**


omp/reorder/rcm_kernels.cpp
```diff
@@ -235,8 +235,8 @@ struct UbfsLinearQueue {
 #define GKO_CMPXCHG_IMPL(ptr, ptr_expected, replace_with) \
     return __atomic_compare_exchange_n(                   \
```
yhmtsai, line 236: clang and gcc indeed use int for memory_order although clang doc uses `__atomic_compare_exchange_N` not `__atomic_compare_exchange_n`

yhmtsai (approved):

omp/reorder/rcm_kernels.cpp
jbadwaik: Should I fix that line as well? Or is it something for other authors/maintainers?

tcojean (approved):

tcojean:
I think we should be able to merge this as is. While using OpenMP or a standard way of doing atomics would be an improvement, this can be exported to an issue and adressed in another MR.

upsj:
after #1436 is merged, we should rebase this and run `check_pr` again

omp/reorder/rcm_kernels.cpp
yhmtsai: It should be okay now

yhmtsai:
rebase!

yhmtsai:
format!

PR #1403, opened 2023-09-06
Distributed pgm

yhmtsai:
This pr adds distributed pgm support.
The distributed pgm only creates the aggregation map locally. That is, there's no aggregation across ranks.
1. build the aggregation map locally
2. communicate the aggregation map
3. aggregate the off-diagonal part
4. create coarse matrix

There are two new constructors for distributed matrix.
1. create the matrix by the existing mapping information

TODO:
- [x] it uses dispatch now ~~move the new constructor of distributed matrix to another pr. check the local-only matrix more carefully.~~


include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -536,6 +586,20 @@ class Matrix
                     ptr_param<const LinOp> local_matrix_template,
                     ptr_param<const LinOp> non_local_matrix_template);
 
+    // build local only matrix from existed data
+    explicit Matrix(std::shared_ptr<const Executor> exec,
+                    mpi::communicator comm, dim<2> size,
+                    std::shared_ptr<LinOp> local_linop);
```
MarcelKoch, line 592: What is this used for? If you are using only a local linop without communication (i.e. no send/recv sizes, offsets, etc.) then you can just use the linop directly.

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -155,6 +155,22 @@ template <typename ValueType>
 class Vector;
 
 
+// TODO: move the data into this class
+template <typename IndexType>
+class MatrixBase {
```
MarcelKoch, line 160: I think it is more appropriate to move the getters for the communication data (and subsequently the data itself) into its own class instead of the base class approach. I was working on something similar for the local matrix stuff.

MarcelKoch (commented):
My initial impression is that we really need to update our distributed matrix constructors. Allowing creating it from some pre-existing communication data + linops seems crucial and should be the first step (probably in another PR).

core/multigrid/pgm.cpp
```diff
@@ -227,10 +237,238 @@ void Pgm<ValueType, IndexType>::generate()
...
 
-    this->set_multigrid_level(prolong_row_gather, coarse_matrix,
-                              restrict_sparsity);
+
+template <typename ValueType, typename IndexType>
+void Pgm<ValueType, IndexType>::generate()
+{
+    using csr_type = matrix::Csr<ValueType, IndexType>;
+#if GINKGO_BUILD_MPI
+    if (auto matrix = std::dynamic_pointer_cast<
```
greole, line 314: This implementation is quite lengthy, maybe you can split that up into several sub methods to make the if/else clause easier to read.

core/multigrid/pgm.cpp
```diff
@@ -227,10 +237,238 @@ void Pgm<ValueType, IndexType>::generate()
...
+            const experimental::distributed::MatrixBase<IndexType>>(
+            system_matrix_)) {
+        // only work for the square local matrix
```
greole, line 317: This comment seems to be out of place. I cannot see testing for the "squareness" of the matrix here.

include/ginkgo/core/multigrid/pgm.hpp
```diff
@@ -189,6 +190,11 @@ class Pgm : public EnableLinOp<Pgm<ValueType, IndexType>>,
 
     void generate();
 
+    std::tuple<std::shared_ptr<LinOp>, std::shared_ptr<LinOp>,
```
greole, line 193: Eventhough it is not a public method, some short documentation on what generate_local is expected to do, would be good.

core/multigrid/pgm.cpp
```diff
@@ -227,10 +237,238 @@ void Pgm<ValueType, IndexType>::generate()
 
     // Construct the coarse matrix
     // TODO: improve it
```
greole, line 239:
```suggestion
```
Either this should be removed, or more specific on how to improve it.

include/ginkgo/core/multigrid/pgm.hpp
```diff
@@ -189,6 +190,11 @@ class Pgm : public EnableLinOp<Pgm<ValueType, IndexType>>,
...
+               std::shared_ptr<LinOp>>
+    generate_local(
```
greole, line 195: I think it would make sense to add some documentation on what `generate_local `does and what the return type represent, to simplify understanding.

greole (commented):
Some quick comments

core/multigrid/pgm.cpp
```diff
@@ -227,10 +237,238 @@ void Pgm<ValueType, IndexType>::generate()
...
+        auto result = this->generate_local(pgm_local_op);
+        auto non_local_matrix = matrix->get_non_local_matrix();
+        auto non_local_size = non_local_matrix->get_size()[1];
+        array<IndexType> non_local_agg(exec, non_local_size);
+        // prolong, restrict, only needs the local infomation
+        // get agg information (prolong_row_gather row idx)
+        communicate(matrix, agg_, non_local_agg);
+        // generate non_local_col_map
+        non_local_agg.set_executor(exec->get_master());
+        array<IndexType> non_local_col_map(exec->get_master(), non_local_size);
```
MarcelKoch, line 339: I think computing the `non_local_col_map` here can be merged with a similar part in the `Matrix::read_distributed`. In both cases, what we are doing is, for an index sequence that is segmented by keys (the target rank), we map each index into the interval `[0, U)`, where `U` is the number of unique indices in that segment. I will extract the relevant parts. That should simplify the following code a lot.

MarcelKoch: The branch `compress-indices` contains the relevant part of `read_distributed` extracted. Using the part_id as keys and the non_local_agg as indices should give you a mapping from the fine non-local columns to the coarse non-local columns. Since the output will also contain the part_ids for the coarse unique columns, it should also simplify computing the gather_idxs, and recv_sizes. I would guess that both can be computed nearly identically to the approach in `read_distibuted`

core/multigrid/pgm.cpp
```diff
@@ -227,10 +237,238 @@ void Pgm<ValueType, IndexType>::generate()
...
+        }
+        renumber.get_data()[non_local_size] =
+            renumber.get_data()[non_local_size - 1] + 1;
+        // create col map
+        // for each thread i, col_map[tuple[i].index] = map[i]
+        for (int i = 0; i < non_local_size; i++) {
+            non_local_col_map.get_data()[index.get_data()[i]] =
+                renumber.get_data()[i];
+        }
+        // get new recv_size and recv_offsets
```
MarcelKoch, line 376: I think you can also easily compute the send_sizes/offsets here, instead of communicating them.

common/unified/multigrid/pgm_kernels.cpp
```diff
@@ -311,6 +311,22 @@ GKO_INSTANTIATE_FOR_EACH_NON_COMPLEX_VALUE_AND_INDEX_TYPE(
     GKO_DECLARE_PGM_ASSIGN_TO_EXIST_AGG);
 
 
+template <typename IndexType>
+void gather_index(std::shared_ptr<const DefaultExecutor> exec,
```
pratikvn, line 315: nit:
```suggestion
void gather_indices(std::shared_ptr<const DefaultExecutor> exec,
```

core/distributed/matrix.cpp
```diff
@@ -89,6 +88,62 @@ Matrix<ValueType, LocalIndexType, GlobalIndexType>::Matrix(
...
+                      recv_offsets_.data(), gather_idxs_.get_data(),
+                      send_sizes_.data(), send_offsets_.data());
+    if (use_host_buffer) {
+        gather_idxs_.set_executor(exec);
+        recv_gather_idxs_.set_executor(exec);
+    }
+
+    one_scalar_.init(exec, dim<2>{1, 1});
+    one_scalar_->fill(one<value_type>());
+}
```
pratikvn, line 145: Do we merge this PR first, or the distributed stack first ?

test/mpi/matrix.cpp
```diff
@@ -156,6 +156,117 @@ TYPED_TEST(MatrixCreation, ReadsDistributedWithColPartition)
...
+    x->read_distributed(vec_md, part);
+    y->read_distributed(vec_md, part);
+
+    mat->apply(x, y);
+
+    GKO_ASSERT_MTX_NEAR(y->get_local_vector(), result[rank], 0);
+}
+
+
+TYPED_TEST(MatrixCreation, BuildFromExistedData)
```
pratikvn, line 206:
```suggestion
TYPED_TEST(MatrixCreation, BuildFromExistingData)
```

core/multigrid/pgm.cpp
```diff
@@ -112,40 +120,43 @@ std::shared_ptr<matrix::Csr<ValueType, IndexType>> generate_coarse(
...
 
 
 template <typename ValueType, typename IndexType>
-void Pgm<ValueType, IndexType>::generate()
+std::tuple<std::shared_ptr<LinOp>, std::shared_ptr<LinOp>,
+           std::shared_ptr<LinOp>>
+Pgm<ValueType, IndexType>::generate_local(
+    std::shared_ptr<const matrix::Csr<ValueType, IndexType>> local_matrix)
 {
     using csr_type = matrix::Csr<ValueType, IndexType>;
```
pratikvn, line 142: unused

core/multigrid/pgm.cpp
```diff
@@ -199,10 +209,240 @@ void Pgm<ValueType, IndexType>::generate()
...
+                                    : non_local_agg.get_data();
+    exec->synchronize();
+    comm.all_to_all_v(use_host_buffer ? exec->get_master() : exec, send_ptr,
+                      send_sizes.data(), send_offsets.data(), type, recv_ptr,
+                      recv_sizes.data(), recv_offsets.data(), type);
+    if (use_host_buffer) {
+        exec->copy_from(exec->get_master(), total_recv_size, recv_ptr,
+                        non_local_agg.get_data());
+    }
+    return;
```
pratikvn, line 267: not needed ?

pratikvn (commented):
In general, LGTM. But I think some of Marcel's comments regarding the simplifications and unification with the distributed code has not yet been completed ? 

I am also not sure of the merge order here. Unlike distributed multigrid, this does heavily depend on the distributed stack, so maybe that should be merged first ?

test/mpi/multigrid/pgm.cpp
```diff
@@ -0,0 +1,125 @@
...
+    using value_type = typename TestFixture::value_type;
+    using dist_mtx_type = typename TestFixture::dist_mtx_type;
+    using local_matrix_type = typename TestFixture::local_matrix_type;
+    auto pgm_factory = pgm::build().on(this->exec);
+    auto rank = this->comm.rank();
+    I<I<value_type>> res_local[] = {{{8}}, {{5, 0}, {0, 5}}, {{6, 0}, {0, 6}}};
+    // the rank 2 part of non local matrix of rank 1 are reordered.
+    // [0 -1 -2 0], 1st and 3rd are aggregated to the first group but the rest
+    // are aggregated to the second group. Thus, the aggregated result should be
+    // [-2 -1] not [-1, -2]
```
greole, line 112: I find this comment quite hard to understand.

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -533,6 +585,31 @@ class Matrix
...
+     * the corresponding mapping.
+     *
+     * @note It use the input to build up the distributed matrix
+     *
+     * @param exec  Executor associated with this matrix.
+     * @param comm  Communicator associated with this matrix.
+     * @param size  the global size
+     * @param local_linop  the local linop
+     * @param non_local_linop  the non-local linop
+     * @param recv_sizes  the size of non-local recevier
```
greole, line 599: fix typo
```suggestion
     * @param recv_sizes  the size of non-local receiver
```

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -533,6 +585,31 @@ class Matrix
                                           mpi::communicator comm, dim<2> size,
                                           std::shared_ptr<LinOp> local_linop);
 
+    /**
+     * Creates distributed matrix with existent local and non-local LinOp and
+     * the corresponding mapping.
```
greole, line 590: What exactly is the mapping here? I don't find it in the list of params either?

core/distributed/matrix.cpp
```diff
@@ -89,6 +88,62 @@ Matrix<ValueType, LocalIndexType, GlobalIndexType>::Matrix(
     local_mtx_ = local_linop;
 }
 
+template <typename ValueType, typename LocalIndexType, typename GlobalIndexType>
+Matrix<ValueType, LocalIndexType, GlobalIndexType>::Matrix(
```
greole, line 92: This introduces a new way to create and fill distributed matrices compared to the read_distributed approach. I find that a bit confusing because we don't really communicate why in one situation on use  `Matrix::create(..)` followed by `Matrix::read_distributed()` vs `Matrix::create(..)` and passing the local and non-local matrix directly.

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -388,6 +408,37 @@ class Matrix
...
+    };
+    std::vector<comm_index_type> get_recv_offsets() const override
+    {
+        return recv_offsets_;
+    };
+    std::vector<comm_index_type> get_send_offsets() const override
+    {
+        return send_offsets_;
+    }
+    array<local_index_type> get_gather_idxs() const override
```
greole, line 428: Just for my understanding, why don't we return  `const array<local_index_type>&` ?

greole (commented):
Some comments from my side. Also, do we have an example or test demonstrating the use of distributed PGM in combination with the distributed multigrid solver?

common/unified/multigrid/pgm_kernels.cpp
yhmtsai: we also use `gather_index` in dense, so I keep the same name

core/multigrid/pgm.cpp
yhmtsai: it's used in weight_csr_type

include/ginkgo/core/distributed/matrix.hpp
yhmtsai: After discussing with Marcel, we decide to use `friend`.

test/mpi/multigrid/pgm.cpp
yhmtsai: I try to update it a little bit. I hope it is better now.

include/ginkgo/core/multigrid/pgm.hpp
```diff
@@ -161,6 +163,25 @@ class Pgm : public EnableLinOp<Pgm<ValueType, IndexType>>,
 
     void generate();
 
+    /**
+     * This function only handles the local matrix coarsening method.
```
greole, line 167: Maybe better?
```suggestion
     * This function generates the local matrix coarsening operators.
```

core/base/dispatch_helper.hpp
```diff
@@ -132,6 +151,39 @@ auto run(T obj, Func&& f, Args&&... args)
...
+ * @tparam T  the type of input object
+ * @tparam Func  the function will run if the object can be converted to K
+ * @tparam ...Args  the additional arguments for the Func
+ *
+ * @param obj  the input object waiting converted
+ * @param f  the function will run if obj can be converted successfully
+ * @param args  the additional arguments for the function
+ *
+ * @note  This assumes that each invocation of f with types (std::shared_ptr<K>,
+ * std::shared_ptr<Types>...) returns the same type
```
greole, line 169:
```suggestion
 * @note  This assumes that the return type of f is independent of the input types (std::shared_ptr<K>,
 * std::shared_ptr<Types>...)
```

core/base/dispatch_helper.hpp
```diff
@@ -132,6 +151,39 @@ auto run(T obj, Func&& f, Args&&... args)
...
 
 
+/**
+ * run uses template to go through the list and select the valid
+ * template and run it.
+ *
+ * @tparam K  the current shared_ptr type tried in the conversion
+ * @tparam ...Types  the other types will be tried in the conversion if K fails
+ * @tparam T  the type of input object
+ * @tparam Func  the function will run if the object can be converted to K
```
greole, line 161:
```suggestion
 * @tparam Func  function type that is invoked  if the object can be converted to K
```

core/multigrid/pgm.cpp
```diff
@@ -199,10 +210,290 @@ void Pgm<ValueType, IndexType>::generate()
...
+#endif
+
+
+template <typename IndexType>
+void generate_non_local_map(
+    const std::vector<experimental::distributed::comm_index_type>& recv_offsets,
+    array<IndexType>& non_local_agg, array<IndexType>& non_local_col_map,
+    array<IndexType>& renumber)
+{
+    auto exec = renumber.get_executor();
```
pratikvn, line 278: Can `renumber` be guaranteed to be on host ? Maybe it should be 
```suggestion
    auto exec = renumber.get_executor()->get_master();
```
if you want to make sure it is on the host ? 

and also probably `GKO_ASSERT(exec == exec->get_master())` in the functions where any `array` is needs to be accessed from host would be good.

core/multigrid/pgm.cpp
```diff
@@ -199,10 +210,290 @@ void Pgm<ValueType, IndexType>::generate()
...
+    renumber.get_data()[non_local_size] =
+        renumber.get_data()[non_local_size - 1] + 1;
+    // create col map
+    // for each thread i, col_map[tuple[i].index] = map[i]
+    for (int i = 0; i < non_local_size; i++) {
+        non_local_col_map.get_data()[index.get_data()[i]] =
+            renumber.get_data()[i];
+    }
+}
+
```
pratikvn, line 316:
```suggestion


```

core/multigrid/pgm.cpp
```diff
@@ -199,10 +210,290 @@ void Pgm<ValueType, IndexType>::generate()
...
+        new_recv_size.at(i) =
+            renumber.get_const_data()[recv_offsets.at(i + 1)] -
+            renumber.get_const_data()[recv_offsets.at(i)];
+        new_recv_offsets.at(i + 1) =
+            new_recv_offsets.at(i) + new_recv_size.at(i);
+    }
+    IndexType non_local_num_agg = new_recv_offsets.back();
+    new_recv_gather_idxs.resize_and_reset(non_local_num_agg);
+    for (int i = 0; i < non_local_agg.get_size(); i++) {
+        new_recv_gather_idxs.get_data()[renumber.get_const_data()[i]] =
```
pratikvn, line 336: I am not sure the implicit assumption that new_recv_gather_idxs is on the host is safe.

core/multigrid/pgm.cpp
```diff
@@ -199,10 +210,290 @@ void Pgm<ValueType, IndexType>::generate()
...
+            // build csr from row and col map
+            // unlike non-distributed version, generate_coarse uses different
+            // row and col maps.
+            auto result_non_local_csr = generate_coarse(
+                exec, non_local_csr.get(),
+                static_cast<IndexType>(std::get<1>(result)->get_size()[0]),
+                agg_, non_local_num_agg, non_local_col_map);
+            // use local and non-local to build coarse matrix
+            // also restriction and prolongation (Local-only-global matrix)
+            int64 coarse_size = std::get<1>(result)->get_size()[0];
```
pratikvn, line 447: static_cast to remove warning ?

pratikvn (approved):
LGTM! A minor concern about using implictly assuming that an array is on the host. I think an assertion `GKO_ASSERT(exec == exec->get_master())` would probably be sufficient in those functions.

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -242,6 +252,7 @@ class Matrix
     friend class EnableDistributedPolymorphicObject<Matrix, LinOp>;
     friend class Matrix<next_precision<ValueType>, LocalIndexType,
                         GlobalIndexType>;
+    friend class multigrid::Pgm<ValueType, LocalIndexType>;
```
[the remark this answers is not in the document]
yhmtsai: Yes, because previous approach require additional features class with many functions. Because we only use these in PGM currently and there are some future changes, we use friend to put these still internal. When we have more than one class for that, we definitely should think about how to extract them to feature class or alwasy use dispatch.

core/multigrid/pgm.cpp
yhmtsai: Yes, it will be on the host. I add the assert to check

core/base/dispatch_helper.hpp
```diff
@@ -132,6 +153,42 @@ auto run(T obj, Func&& f, Args&&... args)
...
+ *               converted to std::shared_ptr<K>
+ * @tparam ...Args  the additional arguments for the Func
+ *
+ * @param obj  the input object waiting converted
+ * @param f  the function will get invoked if obj can be converted successfully
+ * @param args  the additional arguments for the function
+ *
+ * @note  This assumes that the return type of f is independent of the input
+ *        types (std::shared_ptr<K>, std::shared_ptr<Types>...) returns the same
+ *        type
```
greole, line 173:
```suggestion
 *        types (std::shared_ptr<K>, std::shared_ptr<Types>...)
```

core/base/dispatch_helper.hpp
```diff
@@ -132,6 +153,42 @@ auto run(T obj, Func&& f, Args&&... args)
...
+ * template and run it.
+ *
+ * @tparam K  the current shared_ptr type tried in the conversion
+ * @tparam ...Types  the other types will be tried in the conversion if K fails
+ * @tparam T  the type of input object
+ * @tparam Func  the function type that is invoked if the object can be
+ *               converted to std::shared_ptr<K>
+ * @tparam ...Args  the additional arguments for the Func
+ *
+ * @param obj  the input object waiting converted
```
greole, line 167:
```suggestion
 * @param obj  the  input object that should be converted
```

PR #1404, opened 2023-09-11
Reduce license header

MarcelKoch:
This PR changes how we handle our license declaration. It moves from the current block of license (~30 lines) to 3/4 lines as defined by https://reuse.software/. The new license header will now read as:

// SPDX-FileCopyrightText: 2017-2023 The Ginkgo authors
//

It also updates the `add_license.sh` script accordingly ~and changes the top-level license file. The latter might lead to issues with githubs automatic license detection.~

Todo:
- [x] check github's license detection - will just use the current license file


dev_tools/scripts/add_license.sh
```diff
@@ -4,12 +4,11 @@ THIS_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" &>/dev/null && pwd )
 # ${THIS_DIR} is expected to be: ${GINKGO_ROOT_DIR}/dev_tools/scripts
 
 # Use local paths, so there is less chance of a newline being in a path of a found file
-cd "${THIS_DIR}/../.."
+cd "${THIS_DIR}/../.." || exit
 GINKGO_ROOT_DIR="."
 
-
-LICENSE_FILE="${GINKGO_ROOT_DIR}/LICENSE"
-GINKGO_LICENSE_BEACON="******************************<GINKGO LICENSE>******************************"
```
yhmtsai, line 12: does the structure allow the beacon?
I am wondering about some scripts that rely on the BEACON detection.

MarcelKoch: Sorry, I didn't realize that other scripts also use the beacon. I will adjust them.

greole:
REUSE provides a python [tool](https://pypi.org/project/reuse/) to verify reuse compliance. We could also add this to our CI pipeline. However, this will additionally require to add some license information to our cmake files.

MarcelKoch:
@greole I'm very much in favor of that. Here is the old pr for that: #1390. This PR is a first step towards that.

greole (approved):
LGTM

MarcelKoch:
**Note:** Since the format CI is based on the jobs defined in `develop`, that job will always fail. I will mirror this PR in my fork to check if the CI behaves as expected.

MarcelKoch:
See https://github.com/MarcelKoch/ginkgo/pull/10 for CI check. The new license header is used properly in the CI. It was necessary to change the `develop` branch in my fork to already enable the changes to `check-format` as they are in this PR.

yhmtsai:
I saw you change some script to adapt the changes of beacon
Does SPDX allow the following? With the following, we do not need to change many scripts
```
/*******************************<GINKGO LICENSE>******************************
* SPDX-FileCopyrightText: 2017-2023 The Ginkgo authors
*
* SPDX-License-Identifier: BSD-3-Clause
```
With `//` approach, it require all script recognize the SPDX fully. Previously, it is only handled through add_license.sh and others only detect or add beacon.

MarcelKoch:
I would like to keep the new license as short as possible, that's why I'm using `//`. Also, I think don't think the beacon approach is  supported by the `reuse` tool, to which I will move to in the next step.

upsj (approved):
LGTM! As the license is always the first part of the file, and much more structured, I also tend to agree we no longer need the beacon

yhmtsai:
I do not fully agree with it's more structured. beacon way does not hurt the structure but gives easy way to detect the block. something like https://github.com/fsfe/reuse-example/blob/main/src/main.c

MarcelKoch:
I meant that the reuse tool doesn't support it for adding new license header. Also, I don't see why we need to detect the block. Adding the license will be handled by reuse, which can detect it correctly, and the formatting by clang.

dev_tools/oneapi/convert_source.sh
```diff
@@ -167,8 +167,10 @@ fi
 rm "${OUTPUT_FILE}"
 echo "#define GET_QUEUE 0" >> "${OUTPUT_FILE}"
 # add empty ginkgo license such that format_header recognize some header before header def macro
-echo "/*${GINKGO_LICENSE_BEACON}" >> "${OUTPUT_FILE}"
-echo "${GINKGO_LICENSE_BEACON}*/" >> "${OUTPUT_FILE}"
+CURRENT_YEAR=$(date +%Y)
+echo "${GINKGO_LICENSE_BEGIN} 2017-${CURRENT_YEAR} The Ginkgo authors" > "${OUTPUT_FILE}"
+echo "//" >> "${OUTPUT_FILE}"
+echo "${GINKGO_LICENSE_END} BSD-3-Clause" >> "${OUTPUT_FILE}"
```
yhmtsai, line 173:
```suggestion
echo "${GINKGO_LICENSE_BEGIN} 2017-${CURRENT_YEAR} The Ginkgo authors" >> "${OUTPUT_FILE}"
echo "//" >> "${OUTPUT_FILE}"
echo "${GINKGO_LICENSE_END} BSD-3-Clause" >> "${OUTPUT_FILE}"
```
It should be controlled by add_licences.
It's fine if it can be fixed automatically when they are different from the add_licences.sh
need to use append >>

dev_tools/scripts/add_license.sh
```diff
@@ -41,9 +40,15 @@ if ! command -v cut &> /dev/null; then
...
 fi
+if ! command -v date &> /dev/null; then
+    echo 'The command `date` is required for this script to work, but not supported by your system.' 1>&2
+    exit 1
+fi
 
-
-echo -e "/*${GINKGO_LICENSE_BEACON}\n$(cat ${LICENSE_FILE})\n${GINKGO_LICENSE_BEACON}*/\n" > "${COMMENTED_LICENSE_FILE}"
+CURRENT_YEAR=$(date +%Y)
+echo "${GINKGO_LICENSE_BEGIN} 2017-${CURRENT_YEAR} The Ginkgo authors" > "${COMMENTED_LICENSE_FILE}"
```
yhmtsai, line 49: some files use `the` Ginkgo authors, could you also change one of them? the or The

yhmtsai (changes requested):
I consider it maybe from another side.
We need to pay additional attension on changing beacon to `//`.  handling license in two or three places may not be good for reuse/consistense.
I originally think it only need to change or create another file (`SPDX...`)for add_licences/format_header.  Other scripts work smoothly.
beacon way is widely used by tranditional license in my mind, so I am also suprised reuse tool may not support it.
request changes mainly for the wrong script and the inconsistence between "the" and "The"

dev_tools/oneapi/convert_source.sh
MarcelKoch: Do you mean to remove this part, and just let add_license.sh handle this?

dev_tools/scripts/add_license.sh
MarcelKoch: I'm now using always `The Ginkgo authors`, since `the` was only used in two files.

dev_tools/oneapi/convert_source.sh
yhmtsai: No, I think add_license.sh needs `// SPDX` to add the corresponding license?
I would like to ensure the add_license.sh is the final license control to ensure it will be correct even if it is wrong here.
The license is hard-coded in more than one place, so it may be easy to miss one of them in the future

cmake/Modules/CudaArchitectureSelector.cmake
```diff
@@ -1,5 +1,5 @@
 ################################ License #######################################
-# Copyright (c) 2017-2023, the Ginkgo authors
+# Copyright (c) 2017-2023 The Ginkgo authors
```
yhmtsai, line 2: should this also use `SPDX`?

yhmtsai (approved):
I do not hold the concern for beacon now.

dev_tools/oneapi/convert_source.sh
MarcelKoch: The add_license.sh script is getting removed in the next PR anyway, so perhaps we shouldn't dwell on this too long. I will create another PR showing how it will be removed.

cmake/Modules/CudaArchitectureSelector.cmake
MarcelKoch: Yes, thanks for catching that.

dev_tools/oneapi/convert_source.sh
MarcelKoch: BTW, here is the PR that removes the script: #1460

pratikvn (approved):

PR #1405, opened 2023-09-13
Use fixed clang format

MarcelKoch:
This PR enables us to force users to use the same version of clang-format as our CI. This is done by requiring contributors to install [pre-commit](https://pre-commit.com/) (which can be done easily through pip/pipx). pre-commit allows us to fix the used version of clang-format, so that contributors will use exactly the same version as our CI.

If pre-commit can't be detected during cmake an error is thrown. Otherwise, the pre-commit hooks are installed.

Since this replaces our git-cmake-format, I have removed it.

To see it in action: here is an PR in a fork that shows the CI use cases https://github.com/MarcelKoch/ginkgo/pull/9

# Warning

For those using git-worktrees, if one working tree is updated with this PR it will overwrite the git hooks for all working trees (I don't think you can have different hooks for different working trees). The easy fix is of course to just update all used working trees accordingly.

### DIFF
--- a/.github/format-rebase.sh
+++ b/.github/format-rebase.sh
@@ -12,7 +12,7 @@ git rebase --rebase-merges --empty=drop --no-keep-empty \
     --exec "cp /tmp/add_license.sh /tmp/format_header.sh /tmp/update_ginkgo_header.sh dev_tools/scripts/ && \
             dev_tools/scripts/add_license.sh && dev_tools/scripts/update_ginkgo_header.sh && \
             for f in \$($DIFF_COMMAND | grep -E '$FORMAT_HEADER_REGEX'); do dev_tools/scripts/format_header.sh \$f; done && \
-            pipx run pre-commit run --files $TO_FORMAT && \
+            pipx run pre-commit run && \
             git checkout dev_tools/scripts && (git diff >> /tmp/difflog; true) && (git diff --quiet || git commit -a --amend --no-edit --allow-empty)" \
     base/$BASE_BRANCH 2>&1 || bot_error "Rebase failed, see the related [Action]($JOB_URL) for details"
 
--- a/.github/format-rebase.sh
+++ b/.github/format-rebase.sh
@@ -1,25 +1,7 @@
 #!/usr/bin/env bash
 
-source .github/bot-pr-base.sh
-
-git remote add base "$BASE_URL"
-git remote add fork "$HEAD_URL"
-
-git remote -v
-
-git fetch base $BASE_BRANCH
-git fetch fork $HEAD_BRANCH
-
-git config user.email "$USER_EMAIL"
-git config user.name "$USER_NAME"
-
-LOCAL_BRANCH=rebase-tmp-$HEAD_BRANCH
-git checkout -b $LOCAL_BRANCH fork/$HEAD_BRANCH
-
-# save scripts from develop
-pushd dev_tools/scripts
-cp add_license.sh format_header.sh update_ginkgo_header.sh /tmp
-popd
+cp .github/bot-pr-format-base.sh /tmp
+source /tmp/bot-pr-format-base.sh
 
 bot_delete_comments_matching "Error: Rebase failed"
 
@@ -30,7 +12,7 @@ git rebase --rebase-merges --empty=drop --no-keep-empty \
     --exec "cp /tmp/add_license.sh /tmp/format_header.sh /tmp/update_ginkgo_header.sh dev_tools/scripts/ && \
             dev_tools/scripts/add_license.sh && dev_tools/scripts/update_ginkgo_header.sh && \
             for f in \$($DIFF_COMMAND | grep -E '$FORMAT_HEADER_REGEX'); do dev_tools/scripts/format_header.sh \$f; done && \
-            for f in \$($DIFF_COMMAND | grep -E '$FORMAT_REGEX'); do $CLANG_FORMAT -i \$f; done && \
+            pipx run pre-commit run --files $TO_FORMAT && \
             git checkout dev_tools/scripts && (git diff >> /tmp/difflog; true) && (git diff --quiet || git commit -a --amend --no-edit --allow-empty)" \
     base/$BASE_BRANCH 2>&1 || bot_error "Rebase failed, see the related [Action]($JOB_URL) for details"
 
--- a/.github/workflows/bot-pr-comment.yml
+++ b/.github/workflows/bot-pr-comment.yml
@@ -1,7 +1,9 @@
+name: OnCommentPR
+
 on:
   issue_comment:
     types: [created]
-name: OnCommentPR
+
 jobs:
   label:
     runs-on: ubuntu-latest
@@ -15,25 +17,12 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ secrets.BOT_TOKEN }}
       run: cp --preserve .github/label.sh /tmp && /tmp/label.sh
+
   check_format:
     name: check-format
-    runs-on: ubuntu-22.04
     if: github.event.issue.pull_request != '' && github.event.comment.body == 'check-format!' && (github.event.comment.author_association == 'COLLABORATOR' || github.event.comment.author_association == 'MEMBER' || github.event.comment.author_association == 'OWNER')
-    steps:
-    - name: Checkout the latest code (shallow clone)
-      uses: actions/checkout@v3
-      with:
-        ref: develop
-    - name: Check for formatting changes
-      env:
-        GITHUB_TOKEN: ${{ secrets.BOT_TOKEN }}
-      run: cp --preserve .github/check-format.sh /tmp && /tmp/check-format.sh
-    - name: Upload code formatting patch
-      if: failure()
-      uses: actions/upload-artifact@v3
-      with:
-          name: patch
-          path: format.patch
+    uses: ./.github/workflows/check-formatting.yml
+
   format:
     name: format
     runs-on: ubuntu-22.04
@@ -48,6 +37,7 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ secrets.BOT_TOKEN }}
       run: cp --preserve .github/format.sh /tmp && /tmp/format.sh
+
   rebase:
     name: rebase
     if: github.event.issue.pull_request != '' && github.event.comment.body == 'rebase!' && (github.event.comment.author_association == 'COLLABORATOR' || github.event.comment.author_association == 'MEMBER' || github.event.comment.author_association == 'OWNER')
@@ -63,6 +53,7 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ secrets.BOT_TOKEN }}
       run: cp --preserve .github/rebase.sh /tmp && /tmp/rebase.sh
+
   format-rebase:
     name: format-rebase
     if: github.event.issue.pull_request != '' && github.event.comment.body == 'format-rebase!' && (github.event.comment.author_association == 'COLLABORATOR' || github.event.comment.author_association == 'MEMBER' || github.event.comment.author_association == 'OWNER')
--- a/.github/workflows/bot-pr-updated.yml
+++ b/.github/workflows/bot-pr-updated.yml
@@ -1,29 +1,18 @@
+name: OnSyncPR
+
 on:
   pull_request_target:
     types: [opened,synchronize]
-name: OnSyncPR
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
   cancel-in-progress: true
+
 jobs:
   check-format:
-    runs-on: ubuntu-22.04
     if: github.event.pull_request.author_association == 'COLLABORATOR' || github.event.pull_request.author_association == 'MEMBER' || github.event.pull_request.author_association == 'OWNER'
-    steps:
-      - name: Checkout the latest code (shallow clone)
-        uses: actions/checkout@v3
-        with:
-          ref: develop
-      - name: Check for formatting changes
-        env:
-          GITHUB_TOKEN: ${{ secrets.BOT_TOKEN }}
-        run: cp .github/check-format.sh /tmp && /tmp/check-format.sh
-      - name: Upload code formatting patch
-        if: failure()
-        uses: actions/upload-artifact@v3
-        with:
-          name: patch
-          path: format.patch
+    uses: ./.github/workflows/check-formatting.yml
+
   abidiff:
     runs-on: ubuntu-latest
     if: github.event.pull_request.author_association == 'COLLABORATOR' || github.event.pull_request.author_association == 'MEMBER' || github.event.pull_request.author_association == 'OWNER'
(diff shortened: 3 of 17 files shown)


MarcelKoch:
Note: the changes to the CI will not be reflected in the CI runs here.

yhmtsai:
If we make git-cmake-format be able to handle max (or fixed version), does it also fit the same purpose?

.github/format-rebase.sh
```diff
@@ -30,7 +30,7 @@ git rebase --rebase-merges --empty=drop --no-keep-empty \
     --exec "cp /tmp/add_license.sh /tmp/format_header.sh /tmp/update_ginkgo_header.sh dev_tools/scripts/ && \
             dev_tools/scripts/add_license.sh && dev_tools/scripts/update_ginkgo_header.sh && \
             for f in \$($DIFF_COMMAND | grep -E '$FORMAT_HEADER_REGEX'); do dev_tools/scripts/format_header.sh \$f; done && \
-            for f in \$($DIFF_COMMAND | grep -E '$FORMAT_REGEX'); do $CLANG_FORMAT -i \$f; done && \
+            pipx run pre-commit run --files $TO_FORMAT && \
```
yhmtsai, line 33: $TO_FORMAT is defined in .github/format.sh but it is not sourced

.pre-commit-config.yaml
```diff
@@ -0,0 +1,7 @@
+repos:
+-   repo: https://github.com/pre-commit/mirrors-clang-format
+    rev: 'v14.0.0'  # The default in Ubuntu 22.04, which is used in our CI
```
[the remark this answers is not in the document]
MarcelKoch: I don't really know, how the user environment is changes. But what I can tell you is that on my system, where I use pre-commit, I don't have clang-14 in my path, even though it is used by pre-commit.

MarcelKoch:
Here is a PR in my fork showing the updated CI: https://github.com/MarcelKoch/ginkgo/pull/11

.pre-commit-config.yaml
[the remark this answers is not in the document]
yhmtsai: After some search, it is handled by python virtualenv.
The default folder is ~/.cache/pre-commit `https://[pre-commit](https://pre-commit.com/#managing-ci-caches).com/#managing-ci-caches`
there's a folder (py_env-python3.9) under a random folder, `source py_env-python3.9/bin/activate` can run clang-format from the environment.
I am still looking for how to get the path correctly and use it in editor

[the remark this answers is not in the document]
MarcelKoch: For what do you need it in your editor? You can just run `pre-commit run` and it will do the formatting.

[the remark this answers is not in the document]
yhmtsai: I usually run several times clang-format during coding with the default shortcut.
pre-commit requires us to run via command and pass files (or all-files)

.github/workflows/bot-pr-updated.yml
```diff
@@ -1,65 +1,18 @@
...
-        env:
-          GITHUB_TOKEN: ${{ secrets.BOT_TOKEN }}
-        run: cp .github/check-format.sh /tmp && /tmp/check-format.sh
-      - name: Upload code formatting patch
-        if: failure()
-        uses: actions/upload-artifact@v3
-        with:
-          name: patch
-          path: format.patch
-  abidiff:
```
yhmtsai, line 27: abidiff is deleted accidentally

.github/workflows/intel.yml
```diff
@@ -21,8 +21,7 @@ jobs:
       fail-fast: false
       matrix:
         config:
```
yhmtsai, line 23: the following change should be rebased error

.github/workflows/windows-msvc-ref.yml
```diff
@@ -27,10 +27,8 @@ jobs:
       fail-fast: false
       matrix:
         config:
-        # Debug shared exceeds symbol limit
```
yhmtsai, line 30: same here

.gitlab-ci.yml
```diff
@@ -297,11 +295,9 @@ test/cuda110/nompi/clang/cuda/release/static:
   variables:
     USE_NAME: "cuda110-nompi-clang-${CI_PIPELINE_ID}"
     SLURM_PARTITION: "accelerated"
-    SLURM_GRES: "gpu:4"
```
yhmtsai, line 300: same here?

.github/workflows/bot-pr-comment.yml
```diff
@@ -46,8 +35,9 @@ jobs:
         persist-credentials: false
     - name: Commit formatting changes
       env:
-        GITHUB_TOKEN: ${{ secrets.BOT_TOKEN }}
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
upsj, line 38: What's the reason behind this change?

CMakeLists.txt
```diff
@@ -354,13 +354,58 @@ if(GINKGO_BUILD_BENCHMARKS)
...
+        execute_process(COMMAND ${GIT_EXECUTABLE} "rev-parse" "--git-dir"
+                        WORKING_DIRECTORY ${Ginkgo_SOURCE_DIR}
+                        RESULT_VARIABLE GINKGO_IS_GIT_DIR_ERROR)
+        if(${GINKGO_IS_GIT_DIR_ERROR})
+            message_git_warning()
+            return()
+        endif()
+
+        find_program(PRE_COMMIT pre-commit)
+        if(${PRE_COMMIT} STREQUAL "PRE_COMMIT-NOTFOUND")
```
upsj, line 379: this should also work:
```suggestion
        if(NOT PRE_COMMIT)
```

CMakeLists.txt
```diff
@@ -354,13 +354,58 @@ if(GINKGO_BUILD_BENCHMARKS)
...
+                          VERBATIM)
+
+        add_custom_target(add_license
+                          COMMAND ${Ginkgo_SOURCE_DIR}/dev_tools/scripts/add_license.sh
+                          WORKING_DIRECTORY ${Ginkgo_SOURCE_DIR})
         add_dependencies(format add_license)
-    endif()
+    endfunction()
+
+    enable_format_check()
```
upsj, line 408: Do you use a separate function for variable scoping?

CMakeLists.txt
```diff
@@ -354,13 +354,58 @@ if(GINKGO_BUILD_BENCHMARKS)
...
-    add_custom_target(add_license
-        COMMAND ${Ginkgo_SOURCE_DIR}/dev_tools/scripts/add_license.sh
-        WORKING_DIRECTORY ${Ginkgo_SOURCE_DIR})
-    # if git-cmake-format can not build format target, do not add the dependencies
-    if(TARGET format)
+    function(enable_format_check)
+        macro(message_git_warning)
+            message(WARNING "No viable Git setup found, but GINKGO_DEVEL_TOOLS=ON is set. "
+                    "It is recommended to enable this option only with the intention of "
+                    "adding changes to Ginkgo via Git.")
```
upsj, line 361: I think we should fail here

upsj (commented):
The general setup looks good to me. I definitely see it clashing with my `clangd` installation which handles the formatting, but it will be caught at commit time, so it should be acceptable. There are a few unrelated changes in the .gitlab-ci.yml we should revert. Also the github token usage leads to the comments coming from the wrong user.

.github/workflows/bot-pr-comment.yml
MarcelKoch: To get the CI running in my fork. I still have to remove some commits before I would merge this,

MarcelKoch: To get the CI running on my fork. I will remove this and other commits before merging.

CMakeLists.txt
MarcelKoch: I'm open to that, and that would make the code quite a bit easier (I wouldn't need the function anymore). But I think we don't fail currently in this case, so I kept that behavior.

CMakeLists.txt
MarcelKoch: No, because I wanted to use fast returns with the corresponding warning. Otherwise, I would get quite deep if-else nesting.

.pre-commit-config.yaml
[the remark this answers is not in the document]
yhmtsai: It may work in almost case but failed with the comment and string because clang-format does not try to concat them back if break early.
local clang-format breaks the string/comment eariler due to different indentation or format. the pre-commit does not fix it if they are still under 80 column limit (but not correct)

CMakeLists.txt
```diff
@@ -354,13 +354,58 @@ if(GINKGO_BUILD_BENCHMARKS)
...
+                        OUTPUT_VARIABLE pre-commit-output
+                        ERROR_VARIABLE pre-commit-error)
+        if(${pre-commit-result})
+            message(FATAL_ERROR
+                    "Failed to install the git hooks via pre-commit. Please check the error message:\n"
+                    "${pre-commit-output}\n${pre-commit-error}")
+        endif()
+
+        add_custom_target(format
+                          COMMAND bash -c "pre-commit run"
```
upsj, line 398:
```suggestion
                          COMMAND bash -c "${PRE_COMMIT} run"
```

.github/workflows/check-formatting.yml
```diff
@@ -0,0 +1,21 @@
...
+    steps:
+      - name: Checkout the latest code (shallow clone)
+        uses: actions/checkout@v4
+      - name: Run Pre-Commit checks
+        run: cp .github/check-format.sh /tmp && /tmp/check-format.sh
+        id: pre-commit
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Upload code formatting patch
+        if: ${{ failure() && steps.pre-commit.outcome == 'failure' }}
```
yhmtsai, line 17: why does it need two failure?

MarcelKoch:
AFAIK, pre-commit is mostly language agnostic. Which language is supported or not depends on the hook. Also, it is always possible to restrict the hooks to certain file sets, e.g. based on filename suffix.

yhmtsai (commented):
from https://github.com/pre-commit/pre-commit/issues/1468, pre-comment does not intend to export the repo usage. It's a bit sad because the inconsistence between local and pre-commit may introduce some annoying difference without notice.

.github/workflows/bot-pr-updated.yml
MarcelKoch: yes, I will revert that, when I prepare this branch for merging

.github/workflows/check-formatting.yml
MarcelKoch: because I don't want to trigger this if any other step failed, e.g. the git checkout step. But I guess it's not necessary, since the uploading would just also fail. 
I will use only one failure.

.pre-commit-config.yaml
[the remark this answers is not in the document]
MarcelKoch: I will check that out.

.github/workflows/check-formatting.yml
yhmtsai: I think keeping the second one should be enough? or the second one requires `failure()` to properly catch

yhmtsai:
**It's not `pre-commit` intended usage. It might be failed if they change the cache way.**
The following script such that we can use the auto installation in editor easily, too.
```
#!/usr/bin/env bash

PRE_COMMIT_CACHE="${HOME}/.cache/pre-commit"
# get the python environment folder from pre-commit
VENV="$(sqlite3 -header -csv ${PRE_COMMIT_CACHE}/db.db "select * from repos;" | grep clang-format,v14.0.0 | sed -E 's/.*,.*,(.*)/\1/g')"
ACTIVATE="$(find ${VENV} -name activate -print -quit)"
# activate env
source "${ACTIVATE}"
# forward the arguments
clang-format "$@"
```

yhmtsai (approved):

.github/workflows/check-formatting.yml
MarcelKoch: The first one is necessary, see: https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions. If `failure()` is not used, the `if` will implicitly use `success()`.

yhmtsai: I see. Thanks for explanation. Keeping both looks good to me

.pre-commit-config.yaml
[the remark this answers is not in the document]
MarcelKoch: So, I don't think there is a clang-format option for that. There is `ReflowComments`, but that only splits up long lines, and does not concat short lines, see: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#reflowcomments. All other clang-format options related to comments seem not to do the concatenation. Which clang-format version are you using?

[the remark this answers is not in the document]
yhmtsai: No, I mean no-concatenation will introduce issue from inconsistent clang-format.
local clang-format may break some comments which are considered long with this format. but pre-commit one does not break in this way.
For example, assuming `pre-commit` clang-format only put the normal indentation for arguments, but local clang-format aligns the arguments
1 is the original code
```
void func(...,
          tttt // some text ...
```
2 is from local clang-format (break when it is too long)
```
void func(...,
          tttt // some text
               // more text
```
3 is format 1 from pre-commit clang-format, less indention so no need for breaking
```
void func(...,
    tttt // some text ...
```
4 is format 2 from pre-commit clang-format, no-concatenation if it is already split
```
void func(...,
    tttt // some text
         // more text
```
3 and 4 are different. If we apply format directly, we should get 3. 
With inconsistent clang-format, we may get 4 in the end.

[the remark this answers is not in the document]
MarcelKoch: can you still provide your clang-format version? I will still look into the issue you mentioned.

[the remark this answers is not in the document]
MarcelKoch: TBH, I think this might be one of those issues which just change depending on the clang format versions.
Just as an example, I used this:
```
    LinOp* apply(ptr_param<const LinOp> alpha,
                 ptr_param<const LinOp> b, // tis a long comment describing the parametr b for what ever reasons
                 ptr_param<const LinOp> beta,
                 ptr_param<LinOp> x)
```
and the pre-commit gave me:
```
    LinOp* apply(ptr_param<const LinOp> alpha,
                 ptr_param<const LinOp> b,  // tis a long comment describing the
                                            // parametr b for what ever reasons
                 ptr_param<const LinOp> beta, ptr_param<LinOp> x)
```
which is your 2. But maybe it depends more on the actual example.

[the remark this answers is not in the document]
yhmtsai: The issue happens when the different indention format clang-format, that's why the example is under the assumption.
In more realistic example, you need to look for the attribute usage.
Clang-format does not give any consistence between version if no option to control behavior such that we may have different indention. It's also why we need to fix a version for repo code not just rely on the config version requirement. There are always some undefined behavior across versions.
Otherwise, the config only requires 9+ and format_header only requires 12+. 
clang-format 9, 12, 14, 17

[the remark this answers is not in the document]
MarcelKoch: I'm not sure if I understand your point. The whole point of this PR is to fix a clang-format version that is used both in the CI and each local repo.

PR #1407, opened 2023-09-22
OpenMP parallel triangular solves and factorizations

upsj:
Still need to check performance on this, but we can use the atomic operations based on OpenMP primitives.

### DIFF
--- a/omp/factorization/cholesky_kernels.cpp
+++ b/omp/factorization/cholesky_kernels.cpp
@@ -40,12 +40,14 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/matrix/csr.hpp>
 
 
+#include "core/base/allocator.hpp"
 #include "core/base/iterator_factory.hpp"
 #include "core/components/fill_array_kernels.hpp"
 #include "core/components/format_conversion_kernels.hpp"
 #include "core/factorization/elimination_forest.hpp"
 #include "core/factorization/lu_kernels.hpp"
 #include "core/matrix/csr_lookup.hpp"
+#include "omp/components/atomic.hpp"
 
 
 namespace gko {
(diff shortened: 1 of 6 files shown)


omp/factorization/cholesky_kernels.cpp
```diff
@@ -248,34 +250,42 @@ void factorize(std::shared_ptr<const DefaultExecutor> exec,
...
                 }
             }
         }
-        ValueType diag = vals[row_diag];
-        for (auto lower_nz = row_begin; lower_nz < row_diag; lower_nz++) {
-            diag -= squared_norm(vals[lower_nz]);
+        const auto diag_val = sqrt(vals[row_diag]);
+        for (auto upper_nz = row_diag + 1; upper_nz < row_end; upper_nz++) {
+            vals[upper_nz] /= diag_val;
             // copy the lower triangular entries to the transpose
```
fritzgoebel, line 284: I think you're copying the upper entries now.

omp/factorization/cholesky_kernels.cpp
```diff
@@ -248,34 +250,42 @@ void factorize(std::shared_ptr<const DefaultExecutor> exec,
     const auto row_ptrs = factors->get_const_row_ptrs();
     const auto cols = factors->get_const_col_idxs();
     const auto vals = factors->get_values();
+    vector<int> ready(num_rows, {exec});
+#pragma omp parallel for schedule(monotonic : auto)
     for (size_type row = 0; row < num_rows; row++) {
```
fritzgoebel, line 255: Just a question: could you start with the rows that have no dependencies here? I am not sure, are those the roots in the elimination forest? I see that for GPUs that doesn't really make a difference because you start many many threads, but maybe here it would?

omp/solver/upper_trs_kernels.cpp
```diff
@@ -97,26 +101,35 @@ void solve(std::shared_ptr<const OmpExecutor> exec,
...
                 if (col == row) {
                     diag = vals[k];
                 }
             }
             if (!unit_diag) {
-                x->at(row, j) /= diag;
+                val /= diag;
+            }
+            if (is_nan(val)) {
+                val = zero<ValueType>();
```
fritzgoebel, line 130: For CUDA, we warn that a `NaN` was produced in the solve, for reference we just keep the `NaN`s in the solution if any are produced. I think we should warn here to, such as not to suggest we produced a correct solution if we didn't.

fritzgoebel (changes requested):
LGTM in general, but I suggest warning if a `NaN` is found, like for CUDA.

PR #1408, opened 2023-09-22
Support reordering in all benchmarks

### LINKED ISSUE #1374: Evaluating solver performance in BENCHMARK with different reordering algorithms
I would like to benchmark the impact of different reordering algorithms (most importantly RCM) already implemented in Ginkgo on the total time to solve some matrices. Ideally, this would be a BENCHMARK option similar to `PRECONDS`, but I understand that implementing and documenting this properly might take some time. Until then, I would highly appreciate any quick, dirty and temporary patch/suggestion :)

BTW I already tried the code implemented in PR #1354, and it indeed helps evaluating performance of reordering algorithms themselves. However, I am not confident that I know what the correct way would be to integrate something similar into `solver` benchmark.

upsj:
This adds another `reorder` parameter to the input JSON, allowing users to specify a reordering algorithm to be applied on the matrix before proceeding. The reordering time itself is not being measured in the benchmark (Should it be?). I am also a bit unsure where to document this.

For consistency, I also added a new interface for RCM that creates a `Permutation`

Fixes #1374

### DIFF
--- a/benchmark/conversion/conversion.cpp
+++ b/benchmark/conversion/conversion.cpp
@@ -118,6 +118,8 @@ struct ConversionBenchmark : Benchmark<gko::device_matrix_data<etype, itype>> {
     {
         gko::matrix_data<etype, itype> data;
         data = Generator::generate_matrix_data(test_case);
+        // no reordering here, as it doesn't impact conversions beyond
+        // dense-sparse conversions
         std::clog << "Matrix is of size (" << data.size[0] << ", "
                   << data.size[1] << "), " << data.nonzeros.size() << std::endl;
         test_case["rows"] = data.size[0];
--- a/benchmark/preconditioner/preconditioner.cpp
+++ b/benchmark/preconditioner/preconditioner.cpp
@@ -183,6 +183,7 @@ struct PreconditionerBenchmark : Benchmark<preconditioner_benchmark_state> {
     {
         preconditioner_benchmark_state state;
         auto data = Generator::generate_matrix_data(test_case);
+        reorder(data, test_case);
 
         state.system_matrix =
             formats::matrix_factory(FLAGS_formats, exec, data);
--- a/benchmark/utils/general_matrix.hpp
+++ b/benchmark/utils/general_matrix.hpp
@@ -51,22 +51,27 @@ std::string reordering_algorithm_desc =
 #if GKO_HAVE_METIS
     "    nd - Nested Dissection reordering algorithm\n"
 #endif
-    "    rcm - Reverse Cuthill-McKee reordering algorithm";
+    "    rcm - Reverse Cuthill-McKee reordering algorithm\n"
+    "This is a preprocessing step whose runtime will not be included\n"
+    "in the measurements.";
 
 
 DEFINE_string(input_matrix, "",
               "Filename of a matrix to be used as the single input. Overwrites "
               "the value of the -input flag");
 
+
+#ifndef GKO_BENCHMARK_DISTRIBUTED
 DEFINE_string(reorder, "none", reordering_algorithm_desc.c_str());
+#endif
 
 
 template <typename ValueType, typename IndexType>
 std::unique_ptr<gko::matrix::Permutation<IndexType>> reorder(
-    gko::matrix_data<ValueType, IndexType>& data, json& test_case,
-    bool is_distributed = false)
+    gko::matrix_data<ValueType, IndexType>& data, json& test_case)
 {
-    if (FLAGS_reorder == "none" || is_distributed) {
+#ifndef GKO_BENCHMARK_DISTRIBUTED
+    if (FLAGS_reorder == "none") {
         return nullptr;
     }
     using Csr = gko::matrix::Csr<ValueType, IndexType>;
@@ -100,6 +105,10 @@ std::unique_ptr<gko::matrix::Permutation<IndexType>> reorder(
     gko::as<Csr>(mtx->permute(&perm_arr))->write(data);
     test_case["reordered"] = FLAGS_reorder;
     return perm;
+#else
+    // no reordering for distributed benchmarks
+    return nullptr;
+#endif
 }
 
 
--- a/benchmark/utils/general_matrix.hpp
+++ b/benchmark/utils/general_matrix.hpp
@@ -41,11 +41,79 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
 #include "benchmark/utils/general.hpp"
+#include "benchmark/utils/generator.hpp"
+
+
+std::string reordering_algorithm_desc =
+    "Reordering algorithm to apply to the input matrices:\n"
+    "    none - no reordering\n"
+    "    amd - Approximate Minimum Degree reordering algorithm\n"
+#if GKO_HAVE_METIS
+    "    nd - Nested Dissection reordering algorithm\n"
+#endif
+    "    rcm - Reverse Cuthill-McKee reordering algorithm";
 
 
 DEFINE_string(input_matrix, "",
               "Filename of a matrix to be used as the single input. Overwrites "
               "the value of the -input flag");
+
+DEFINE_string(reorder, "none", reordering_algorithm_desc.c_str());
+
+
+template <typename ValueType, typename IndexType>
+std::unique_ptr<gko::matrix::Permutation<IndexType>> reorder(
+    gko::matrix_data<ValueType, IndexType>& data, json& test_case,
+    bool is_distributed = false)
+{
+    if (FLAGS_reorder == "none" || is_distributed) {
+        return nullptr;
+    }
+    using Csr = gko::matrix::Csr<ValueType, IndexType>;
+    auto ref = gko::ReferenceExecutor::create();
+    auto mtx = gko::share(Csr::create(ref));
+    mtx->read(data);
+    std::unique_ptr<gko::matrix::Permutation<IndexType>> perm;
+    if (FLAGS_reorder == "amd") {
+        perm = gko::experimental::reorder::Amd<IndexType>::build()
+                   .on(ref)
+                   ->generate(mtx);
+#if GKO_HAVE_METIS
+    } else if (FLAGS_reorder == "nd") {
+        perm = gko::experimental::reorder::NestedDissection<ValueType,
+                                                            IndexType>::build()
+                   .on(ref)
+                   ->generate(mtx);
+#endif
+    } else if (FLAGS_reorder == "rcm") {
+        perm = gko::reorder::Rcm<ValueType, IndexType>::build()
+                   .on(ref)
+                   ->generate(mtx)
+                   ->get_permutation()
+                   ->clone();
+    } else {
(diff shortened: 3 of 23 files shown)


MarcelKoch:
I think this is a good time to clarify when a parameter for the benchmark is part of the json config or part of the program arguments. IIRC, currently we only have the matrix file or stencil config, and the optimal spmv setting, as part of the json, while other parameters are arguments. I think the most similar parameter is the matrix format, which is an argument.

I'm not saying the reordered parameter should also be an argument, I'm just curious about argument for or against it.

upsj:
I agree that this is an important issue, I guess at the core of it is that we take a "cross-product" approach to solver-preconditioner combinations, adding another factor to this product and integrating it into the benchmark seems not the right path at this time. Also, reorderings matter for most if not all benchmarks - cache effects will have impact on all algorithms that operate on the matrix (spmv, sparse_blas, conversions), numerical and fill-in consideration impact all factorization-based preconditioners and solvers.

upsj:
Thinking about it again, I don't think we need to support multiple reorderings at the same time. If we want to benchmark multiple, we can do that in separate benchmark invocations. This also fits with what seems like the normal approach to benchmarking we use. Duplicating the same filename with multiple reorderings seems hard to handle in post-processing.
So I see two possible approaches:
* make `reorder` an input JSON object parameter
* make `reorder` a command-line parameter

An unrelated issue here is that the reordering doesn't affect the RHS, so we may need additional processing for the solver benchmark.

benchmark/utils/generator.hpp
```diff
@@ -55,23 +55,56 @@ struct DefaultSystemGenerator {
...
+                                                       perm.get_data());
+                gko::as<Csr>(mtx->permute(&perm_array))->write(data);
+#endif
+            } else if (reorder_type == "rcm") {
+                auto perm = gko::reorder::Rcm<ValueType, IndexType>::build()
+                                .on(ref)
+                                ->generate(mtx);
+                gko::as<Csr>(mtx->permute(&perm->get_permutation_array()))
+                    ->write(data);
+            }
```
MarcelKoch, line 99: A final `else` is missing if the reorder type didn't match any of the previous options.

benchmark/utils/generator.hpp
```diff
@@ -86,16 +119,22 @@ struct DefaultSystemGenerator {
...
             ss << "stencil(" << config["size"].get<gko::int64>() << ", "
                << config["stencil"].get<std::string>() << ")";
-            return ss.str();
+            raw_desc = ss.str();
         } else {
             throw std::runtime_error("No known way to describe config.");
         }
+        if (config.contains("reorder")) {
+            auto reorder_name = config["reorder"].get<std::string>();
+            raw_desc = reorder_name + "(" + raw_desc + ")";
```
MarcelKoch, line 135: That might just be my perception, but I think this puts too much emphasis on the reorder. Maybe `matrix_desc(reorder)` would be better.

MarcelKoch (commented):
For handling the RHS in the solver, it might be necessary to store the permutation in the generator. Otherwise it would need to be regenerated.

Also, the changelog comment should not mention all benchmarks, since this only works for the non-distributed ones.

benchmark/utils/generator.hpp
```diff
@@ -55,23 +55,56 @@ struct DefaultSystemGenerator {
...
+        if (config.contains("reorder")) {
+            using Csr = gko::matrix::Csr<ValueType, IndexType>;
+            auto ref = gko::ReferenceExecutor::create();
+            auto mtx = gko::share(Csr::create(ref));
+            mtx->read(data);
+            auto reorder_type = config["reorder"].get<std::string>();
+            if (reorder_type == "amd") {
+                auto perm = gko::experimental::reorder::Amd<IndexType>::build()
+                                .on(ref)
+                                ->generate(mtx);
```
pratikvn, line 79: I think it makes sense to measure the time it takes to reorder as well. It does not have to be here, but could be added as a separate part within the detailed benchmark.

pratikvn (commented):
I agree that we don't need to have the "cross-product" for reordering. It does not really lend itself to all options: For example, it requires the CSR format.

I think it would be quite useful to have the runtime of the reordering itself to evaluate the overall effectiveness when comparing different reorderings.

benchmark/utils/generator.hpp
upsj: This is already possible through `sparse_blas`, I would like to avoid supporting this in every benchmark executable, since it mixes the setup phase with the benchmarking phase. Does requiring the users to execute two benchmarks sound acceptable?

include/ginkgo/core/reorder/rcm.hpp
```diff
@@ -224,6 +160,71 @@ class Rcm : public EnablePolymorphicObject<Rcm<ValueType, IndexType>,
...
+
+
+using rcm_starting_strategy = gko::reorder::starting_strategy;
+
+
+/**
+ * @copydoc gko::reorder::Rcm
+ */
+template <typename IndexType = int32>
+class Rcm : public EnablePolymorphicObject<Rcm<IndexType>, LinOpFactory>,
```
MarcelKoch, line 176: I guess this class should replace the current one, right? Then the current one should be deprecated.

benchmark/utils/generator.hpp
pratikvn: I think two benchmarks should be okay.

include/ginkgo/core/reorder/rcm.hpp
MarcelKoch: If you don't want to replace the old one, then perhaps this would fit better in the benchmark directory.

pratikvn (changes requested):
Two things, but otherwise looks good.

1. I think you should separate the updated Rcm interface class to a new PR and only have the benchmark changes in this PR.
2. Not sure if necessary, but maybe it is good to ensure row major ordering after reordering the data.

include/ginkgo/core/reorder/rcm.hpp
```diff
@@ -224,6 +160,71 @@ class Rcm : public EnablePolymorphicObject<Rcm<ValueType, IndexType>,
...
+            public EnablePolymorphicAssignment<Rcm<IndexType>> {
```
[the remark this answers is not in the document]
upsj: One is the old interface based on `ReorderBase`, the other one is the new one base on `Permutation`. We cannot remove the old one, and it doesn't play nice with the new interface.

benchmark/utils/general_matrix.hpp
```diff
@@ -41,12 +41,82 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+        perm = gko::experimental::reorder::Rcm<IndexType>::build()
+                   .on(ref)
+                   ->generate(mtx);
+    } else {
+        throw std::runtime_error{"Unknown reordering algorithm " +
+                                 FLAGS_reorder};
+    }
+    auto perm_arr =
+        gko::array<IndexType>::view(ref, data.size[0], perm->get_permutation());
+    gko::as<Csr>(mtx->permute(&perm_arr))->write(data);
```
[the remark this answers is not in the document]
upsj: column and symmetric permutations ensure that the matrix is sorted afterwards, row permutations don't impact the sortedness, so both cases should be fine.

include/ginkgo/core/reorder/rcm.hpp
upsj: Yes, it is intended for deprecation, but I wanted to avoid having too many tests that require deprecation warnings turned off.

MarcelKoch: Ok, just wanted to know the future direction of this. But in that case we will not get around deprecating the old one. I would also favor @pratikvn suggestion of moving all these changes into a separate PR.

include/ginkgo/core/reorder/rcm.hpp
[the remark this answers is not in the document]
pratikvn: I would agree with @MarcelKoch that it should be deprecated, otherwise, I think it will be confusing for users.

pratikvn (approved):
LGTM!

MarcelKoch:
Do you want to base this PR on the new RCM PR?

upsj:
@MarcelKoch no, it works independently, and only requires ~4 lines changed, so I will merge this first

yhmtsai (commented):
I am not sure how the reordering is used in practice.
Unless the reordered solver is used several times for different right hand side, should the solving time also include reorder time?
Because the flag only accepts one reorder algorithm, there's still a need for several files to store the result when comparing the reordering effect. For me, the cross product is easier to test several reordering, but the multiple runs should also work with additional script

benchmark/conversion/conversion.cpp
```diff
@@ -118,6 +118,8 @@ struct ConversionBenchmark : Benchmark<gko::device_matrix_data<etype, itype>> {
     {
         gko::matrix_data<etype, itype> data;
         data = Generator::generate_matrix_data(test_case);
+        // no reordering here, as it doesn't impact conversions beyond
+        // dense-sparse conversions
```
tcojean, line 122: Now that we have neat classes, we could document this, and generally what the benchmark is doing, by documenting the main struct?
This is rather a comment for the future.

benchmark/utils/general_matrix.hpp
```diff
@@ -41,12 +41,84 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
 #include "benchmark/utils/general.hpp"
+#include "benchmark/utils/generator.hpp"
+
+
+std::string reordering_algorithm_desc =
+    "Reordering algorithm to apply to the input matrices:\n"
+    "    none - no reordering\n"
```
tcojean, line 49: maybe 
```suggestion
    "    none - no reordering (default)\n"
```

benchmark/preconditioner/preconditioner.cpp
```diff
@@ -183,6 +183,7 @@ struct PreconditionerBenchmark : Benchmark<preconditioner_benchmark_state> {
     {
         preconditioner_benchmark_state state;
         auto data = Generator::generate_matrix_data(test_case);
+        reorder(data, test_case);
```
tcojean, line 186: Do these do anything in practice, or is it only to note that it is potentially reordered? Since the permutation isn't stored.

tcojean (approved):
LGTM

benchmark/preconditioner/preconditioner.cpp
upsj: the functions modify `data` and add an entry to `test_case`

benchmark/utils/general_matrix.hpp
upsj: The default value is already displayed in the `--help` output

yhmtsai (approved):

benchmark/conversion/conversion.cpp
upsj: Do you mean the individual benchmarks, or the benchmark runner?

PR #1409, opened 2023-09-23
Fix HIP deprecation warnings

upsj:
This fixes most warnings present in current ROCm builds:
* use `hipsparse/hipsparse.h` instead of `hipsparse.h`, similar for other includes
* hide warnings related to `-funsafe-fp-atomics` in the depfile generation

### DIFF
--- a/benchmark/utils/hip_linops.hip.cpp
+++ b/benchmark/utils/hip_linops.hip.cpp
@@ -36,9 +36,6 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <memory>
 
 
-#include <hipsparse.h>
-
-
 #include "benchmark/utils/sparselib_linops.hpp"
 #include "benchmark/utils/types.hpp"
 #include "hip/base/hipsparse_bindings.hip.hpp"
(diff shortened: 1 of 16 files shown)


upsj:
This needs more work, as the header changes happened some time after 5.0

CMakeLists.txt
```diff
@@ -63,8 +63,6 @@ set(GINKGO_CUDA_COMPILER_FLAGS "" CACHE STRING
     "Set the required NVCC compiler flags, mainly used for warnings. Current default is an empty string")
 set(GINKGO_CUDA_ARCHITECTURES "Auto" CACHE STRING
     "A list of target NVIDIA GPU architectures. See README.md for more detail.")
-# the details of fine/coarse grain memory and unsafe atomic are available https://docs.olcf.ornl.gov/systems/crusher_quick_start_guide.html#floating-point-fp-atomic-operations-and-coarse-fine-grained-memory-allocations
-option(GINKGO_HIP_AMD_UNSAFE_ATOMIC "Compiler uses unsafe floating point atomic (only for AMD GPU and ROCM >= 5). Default is ON because we use hipMalloc, which is always on coarse grain. Must turn off when allocating memory on fine grain" ON)
```
MarcelKoch, line 67: maybe `mark_as_advanced`? I don't think there are many cases, where a user would want to change that.

MarcelKoch (approved):
LGTM, only one small comment.

CMakeLists.txt
upsj: I dug some deeper, the reason we get a warning here is the depfile generation, not the actual compilation. I'll see if there is a nice workaround, otherwise I'll revert it entirely.

yhmtsai:
is not `-munsafe-fp-atomics` supported anymore?
If you only mean the warning output, the flag is somehow parsed in two places (device and host, I guess) but it's only available in device part.
as we showed the performance improvement in slack, I do not think it is not supported.

upsj:
See the latest changes, the warning came from another compiler pass, I added a workaround for it and removed the other changes.

upsj:
format-rebase!

yhmtsai (approved):

PR #1412, opened 2023-09-28
Fix SLURM GPU allocation

upsj:
the Horeka runner now uses CTest resources, but that fails if we don't allocate enough GPUs.


MarcelKoch (approved):

tcojean (approved):
LGTM

Issue #1376, opened 2023-07-31
Batched classes todo list


pratikvn:
This issue tracks the different aspects of the batched functionality that reviewers have suggested, but still need to be implemented:

### `batch::MultiVector`
1. Performance related:
+ [ ] Handling multiple RHS, or special case for single rhs 
+ [ ] Workgroup sizes for DPCPP

2. Utility related:
+ [ ] Add stride capability.

3. Code simplifications:
+ [ ] Using a common launcher syntax, something similar to `parfor_batch_blockwise(x, [](ibatch) { ... })`

PR #1413, opened 2023-10-03
Add a Dense batched class and kernels

pratikvn:
This PR adds support for the Batch Dense matrix format and currently adds very minimal functionality that is necessary.

## TODO

+ [x]  Add CUDA kernels and tests  
+ [x]  Add HIP kernels and tests 
+ [x]  Add DPCPP kernels and tests

### DIFF
--- a/common/cuda_hip/matrix/batch_dense_kernels.hpp.inc
+++ b/common/cuda_hip/matrix/batch_dense_kernels.hpp.inc
@@ -33,7 +33,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 template <typename ValueType>
 __device__ __forceinline__ void simple_apply(
-    const gko::batch::matrix::batch_dense::batch_item<const ValueType>& mat,
+    const gko::batch::matrix::dense::batch_item<const ValueType>& mat,
     const ValueType* const __restrict__ b, ValueType* const __restrict__ x)
 {
     constexpr auto tile_size = config::warp_size;
@@ -65,10 +65,9 @@ template <typename ValueType>
 __global__ __launch_bounds__(
     default_block_size,
     sm_oversubscription) void simple_apply_kernel(const gko::batch::matrix::
-                                                      batch_dense::
-                                                          uniform_batch<
-                                                              const ValueType>
-                                                              mat,
+                                                      dense::uniform_batch<
+                                                          const ValueType>
+                                                          mat,
                                                   const gko::batch::
                                                       multi_vector::
                                                           uniform_batch<
@@ -94,7 +93,7 @@ __global__ __launch_bounds__(
 template <typename ValueType>
 __device__ __forceinline__ void advanced_apply(
     const ValueType alpha,
-    const gko::batch::matrix::batch_dense::batch_item<const ValueType>& mat,
+    const gko::batch::matrix::dense::batch_item<const ValueType>& mat,
     const ValueType* const __restrict__ b, const ValueType beta,
     ValueType* const __restrict__ x)
 {
@@ -132,10 +131,9 @@ __global__ __launch_bounds__(
                                                                 const ValueType>
                                                                 alpha,
                                                     const gko::batch::matrix::
-                                                        batch_dense::
-                                                            uniform_batch<
-                                                                const ValueType>
-                                                                mat,
+                                                        dense::uniform_batch<
+                                                            const ValueType>
+                                                            mat,
                                                     const gko::batch::
                                                         multi_vector::
                                                             uniform_batch<
--- a/common/cuda_hip/matrix/batch_dense_kernels.hpp.inc
+++ b/common/cuda_hip/matrix/batch_dense_kernels.hpp.inc
@@ -52,10 +52,8 @@ __device__ __forceinline__ void simple_apply(
             temp += val * b[j];
         }
 
-#pragma unroll
-        for (int i = static_cast<int>(tile_size) / 2; i > 0; i /= 2) {
-            temp += subwarp_grp.shfl_down(temp, i);
-        }
+        // subgroup level reduction
+        temp = reduce(subgroup, temp, thrust::plus<ValueType>{});
 
         if (subwarp_grp.thread_rank() == 0) {
             x[row] = temp;
@@ -116,10 +114,8 @@ __device__ __forceinline__ void advanced_apply(
             temp += alpha * val * b[j];
         }
 
-#pragma unroll
-        for (int i = static_cast<int>(tile_size) / 2; i > 0; i /= 2) {
-            temp += subwarp_grp.shfl_down(temp, i);
-        }
+        // subgroup level reduction
+        temp = reduce(subgroup, temp, thrust::plus<ValueType>{});
 
         if (subwarp_grp.thread_rank() == 0) {
             x[row] = temp + beta * x[row];
--- a/common/cuda_hip/matrix/batch_dense_kernels.hpp.inc
+++ b/common/cuda_hip/matrix/batch_dense_kernels.hpp.inc
@@ -1,0 +1,170 @@
+/*******************************<GINKGO LICENSE>******************************
+Copyright (c) 2017-2023, the Ginkgo authors
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+
+1. Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+
+2. Redistributions in binary form must reproduce the above copyright
+notice, this list of conditions and the following disclaimer in the
+documentation and/or other materials provided with the distribution.
+
+3. Neither the name of the copyright holder nor the names of its
+contributors may be used to endorse or promote products derived from
+this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS
+IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED
+TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A
+PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+
+template <typename ValueType>
+__device__ __forceinline__ void simple_apply(
+    const gko::batch::matrix::batch_dense::batch_item<const ValueType>& mat,
+    const ValueType* const __restrict__ b, ValueType* const __restrict__ x)
+{
+    constexpr auto tile_size = config::warp_size;
+
+    auto thread_block = group::this_thread_block();
+    auto subwarp_grp = group::tiled_partition<tile_size>(thread_block);
+    const auto subwarp_grp_id = static_cast<int>(threadIdx.x / tile_size);
+    const int num_subwarp_grps_per_block = ceildiv(blockDim.x, tile_size);
+
+    for (int row = subwarp_grp_id; row < mat.num_rows;
+         row += num_subwarp_grps_per_block) {
+        ValueType temp = zero<ValueType>();
+        for (int j = subwarp_grp.thread_rank(); j < mat.num_cols;
+             j += subwarp_grp.size()) {
+            const ValueType val = mat.values[row * mat.stride + j];
+            temp += val * b[j];
+        }
+
+#pragma unroll
+        for (int i = static_cast<int>(tile_size) / 2; i > 0; i /= 2) {
+            temp += subwarp_grp.shfl_down(temp, i);
+        }
+
+        if (subwarp_grp.thread_rank() == 0) {
+            x[row] = temp;
+        }
+    }
+}
+
+template <typename ValueType>
+__global__ __launch_bounds__(
+    default_block_size,
+    sm_oversubscription) void simple_apply_kernel(const gko::batch::matrix::
+                                                      batch_dense::
+                                                          uniform_batch<
+                                                              const ValueType>
+                                                              mat,
+                                                  const gko::batch::
+                                                      multi_vector::
+                                                          uniform_batch<
+                                                              const ValueType>
+                                                              b,
+                                                  const gko::batch::
+                                                      multi_vector::
+                                                          uniform_batch<
+                                                              ValueType>
+                                                              x)
+{
+    for (size_type batch_id = blockIdx.x; batch_id < mat.num_batch_items;
+         batch_id += gridDim.x) {
+        const auto mat_b =
+            gko::batch::matrix::extract_batch_item(mat, batch_id);
+        const auto b_b = gko::batch::extract_batch_item(b, batch_id);
+        const auto x_b = gko::batch::extract_batch_item(x, batch_id);
+        simple_apply(mat_b, b_b.values, x_b.values);
+    }
+}
+
+
+template <typename ValueType>
+__device__ __forceinline__ void advanced_apply(
+    const ValueType alpha,
+    const gko::batch::matrix::batch_dense::batch_item<const ValueType>& mat,
+    const ValueType* const __restrict__ b, const ValueType beta,
+    ValueType* const __restrict__ x)
+{
+    constexpr auto tile_size = config::warp_size;
+
+    auto thread_block = group::this_thread_block();
+    auto subwarp_grp = group::tiled_partition<tile_size>(thread_block);
+    const auto subwarp_grp_id = static_cast<int>(threadIdx.x / tile_size);
+    const int num_subwarp_grps_per_block = ceildiv(blockDim.x, tile_size);
+
+    for (int row = subwarp_grp_id; row < mat.num_rows;
+         row += num_subwarp_grps_per_block) {
+        ValueType temp = zero<ValueType>();
+        for (int j = subwarp_grp.thread_rank(); j < mat.num_cols;
+             j += subwarp_grp.size()) {
+            const ValueType val = mat.values[row * mat.stride + j];
+            temp += alpha * val * b[j];
+        }
+
+#pragma unroll
+        for (int i = static_cast<int>(tile_size) / 2; i > 0; i /= 2) {
+            temp += subwarp_grp.shfl_down(temp, i);
+        }
+
+        if (subwarp_grp.thread_rank() == 0) {
+            x[row] = temp + beta * x[row];
+        }
+    }
+}
+
+template <typename ValueType>
+__global__ __launch_bounds__(
+    default_block_size,
+    sm_oversubscription) void advanced_apply_kernel(const gko::batch::
+                                                        multi_vector::
+                                                            uniform_batch<
+                                                                const ValueType>
+                                                                alpha,
+                                                    const gko::batch::matrix::
+                                                        batch_dense::
+                                                            uniform_batch<
+                                                                const ValueType>
+                                                                mat,
+                                                    const gko::batch::
+                                                        multi_vector::
+                                                            uniform_batch<
+                                                                const ValueType>
+                                                                b,
+                                                    const gko::batch::
+                                                        multi_vector::
+                                                            uniform_batch<
+                                                                const ValueType>
+                                                                beta,
+                                                    const gko::batch::
+                                                        multi_vector::
+                                                            uniform_batch<
+                                                                ValueType>
+                                                                x)
+{
+    for (size_type batch_id = blockIdx.x; batch_id < mat.num_batch_items;
+         batch_id += gridDim.x) {
+        const auto mat_b =
+            gko::batch::matrix::extract_batch_item(mat, batch_id);
+        const auto b_b = gko::batch::extract_batch_item(b, batch_id);
+        const auto x_b = gko::batch::extract_batch_item(x, batch_id);
+        const auto alpha_b = gko::batch::extract_batch_item(alpha, batch_id);
+        const auto beta_b = gko::batch::extract_batch_item(beta, batch_id);
+        advanced_apply(alpha_b.values[0], mat_b, b_b.values, beta_b.values[0],
+                       x_b.values);
+    }
+}
--- a/common/cuda_hip/matrix/batch_dense_kernels.hpp.inc
+++ b/common/cuda_hip/matrix/batch_dense_kernels.hpp.inc
@@ -39,15 +39,15 @@ __device__ __forceinline__ void simple_apply(
     constexpr auto tile_size = config::warp_size;
 
     auto thread_block = group::this_thread_block();
-    auto subwarp_grp = group::tiled_partition<tile_size>(thread_block);
-    const auto subwarp_grp_id = static_cast<int>(threadIdx.x / tile_size);
-    const int num_subwarp_grps_per_block = ceildiv(blockDim.x, tile_size);
+    auto subgroup = group::tiled_partition<tile_size>(thread_block);
+    const auto subgroup_id = static_cast<int>(threadIdx.x / tile_size);
+    const int num_subgroups_per_block = ceildiv(blockDim.x, tile_size);
 
-    for (int row = subwarp_grp_id; row < mat.num_rows;
-         row += num_subwarp_grps_per_block) {
+    for (int row = subgroup_id; row < mat.num_rows;
+         row += num_subgroups_per_block) {
         ValueType temp = zero<ValueType>();
-        for (int j = subwarp_grp.thread_rank(); j < mat.num_cols;
-             j += subwarp_grp.size()) {
+        for (int j = subgroup.thread_rank(); j < mat.num_cols;
+             j += subgroup.size()) {
             const ValueType val = mat.values[row * mat.stride + j];
             temp += val * b[j];
         }
@@ -55,7 +55,7 @@ __device__ __forceinline__ void simple_apply(
         // subgroup level reduction
         temp = reduce(subgroup, temp, thrust::plus<ValueType>{});
 
-        if (subwarp_grp.thread_rank() == 0) {
+        if (subgroup.thread_rank() == 0) {
             x[row] = temp;
         }
     }
@@ -101,15 +101,15 @@ __device__ __forceinline__ void advanced_apply(
     constexpr auto tile_size = config::warp_size;
 
     auto thread_block = group::this_thread_block();
-    auto subwarp_grp = group::tiled_partition<tile_size>(thread_block);
-    const auto subwarp_grp_id = static_cast<int>(threadIdx.x / tile_size);
-    const int num_subwarp_grps_per_block = ceildiv(blockDim.x, tile_size);
+    auto subgroup = group::tiled_partition<tile_size>(thread_block);
+    const auto subgroup_id = static_cast<int>(threadIdx.x / tile_size);
+    const int num_subgroups_per_block = ceildiv(blockDim.x, tile_size);
 
-    for (int row = subwarp_grp_id; row < mat.num_rows;
-         row += num_subwarp_grps_per_block) {
+    for (int row = subgroup_id; row < mat.num_rows;
+         row += num_subgroups_per_block) {
         ValueType temp = zero<ValueType>();
-        for (int j = subwarp_grp.thread_rank(); j < mat.num_cols;
-             j += subwarp_grp.size()) {
+        for (int j = subgroup.thread_rank(); j < mat.num_cols;
+             j += subgroup.size()) {
             const ValueType val = mat.values[row * mat.stride + j];
             temp += alpha * val * b[j];
         }
@@ -117,7 +117,7 @@ __device__ __forceinline__ void advanced_apply(
         // subgroup level reduction
         temp = reduce(subgroup, temp, thrust::plus<ValueType>{});
 
-        if (subwarp_grp.thread_rank() == 0) {
+        if (subgroup.thread_rank() == 0) {
             x[row] = temp + beta * x[row];
         }
     }
--- a/core/base/batch_multi_vector.cpp
+++ b/core/base/batch_multi_vector.cpp
@@ -294,11 +294,12 @@ void MultiVector<ValueType>::move_to(
 template <typename ValueType>
 void MultiVector<ValueType>::convert_to(matrix::Dense<ValueType>* result) const
 {
-    auto exec = result->get_executor() != nullptr ? result->get_executor()
-                                                  : this->get_executor();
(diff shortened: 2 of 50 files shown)


pratikvn:
format!

pratikvn:
format!

cuda/matrix/batch_dense_kernels.cu
```diff
@@ -0,0 +1,86 @@
...
+
+// clang-format off
+
+// NOTE: DO NOT CHANGE THE ORDERING OF THE INCLUDES
+
+#include "common/cuda_hip/matrix/batch_dense_kernels.hpp.inc"
+
+
+#include "common/cuda_hip/matrix/batch_dense_kernel_launcher.hpp.inc"
+
```
MarcelKoch, line 79: why are these two files split?

core/matrix/batch_struct.hpp
```diff
@@ -0,0 +1,125 @@
...
+GKO_ATTRIBUTES GKO_INLINE batch_dense::batch_item<ValueType> extract_batch_item(
+    const batch_dense::uniform_batch<ValueType>& batch,
+    const size_type batch_idx)
+{
+    return {batch.values + batch_idx * batch.stride * batch.num_rows,
+            batch.stride, batch.num_rows, batch.num_cols};
+}
+
+template <typename ValueType>
+GKO_ATTRIBUTES GKO_INLINE batch_dense::batch_item<ValueType> extract_batch_item(
```
MarcelKoch, line 111: not used. Will it be used later?

core/matrix/batch_struct.hpp
```diff
@@ -0,0 +1,125 @@
...
+        return num_rows * stride * sizeof(value_type);
+    }
+};
+
+
+}  // namespace batch_dense
+
+
+template <typename ValueType>
+GKO_ATTRIBUTES GKO_INLINE batch_dense::batch_item<const ValueType> to_const(
```
MarcelKoch, line 86: I don't think this and the other `to_const` are used. Will they be used later?

core/matrix/batch_struct.hpp
```diff
@@ -0,0 +1,125 @@
...
+    using value_type = ValueType;
+    using entry_type = batch_item<ValueType>;
+
+    ValueType* values;
+    size_type num_batch_items;
+    int stride;
+    int num_rows;
+    int num_cols;
+
+    size_type get_entry_storage() const
```
MarcelKoch, line 75: also not used

core/matrix/batch_dense.cpp
```diff
@@ -0,0 +1,251 @@
...
+    gko::detail::const_array_view<ValueType>&& values)
+{
+    // cast const-ness away, but return a const object afterwards,
+    // so we can ensure that no modifications take place.
+    return std::unique_ptr<const Dense>(new Dense{
+        exec, sizes, gko::detail::array_const_cast(std::move(values))});
+}
+
+
+inline const batch_dim<2> get_col_sizes(const batch_dim<2>& sizes)
```
MarcelKoch, line 181: unused

core/matrix/batch_dense.cpp
```diff
@@ -0,0 +1,251 @@
...
+    this->get_executor()->run(
+        dense::make_advanced_apply(alpha, this, b, beta, x));
+}
+
+
+template <typename ValueType>
+void Dense<ValueType>::convert_to(
+    Dense<next_precision<ValueType>>* result) const
+{
+    result->values_ = this->values_;
```
MarcelKoch, line 233:
```suggestion
    result->values_ = array<next_precision<ValueType>(result->get_executor, this->values_);
```

common/cuda_hip/matrix/batch_dense_kernels.hpp.inc
```diff
@@ -0,0 +1,166 @@
...
+                                                              const ValueType>
+                                                              b,
+                                                  const gko::batch::
+                                                      multi_vector::
+                                                          uniform_batch<
+                                                              ValueType>
+                                                              x)
+{
+    for (size_type batch_id = blockIdx.x; batch_id < mat.num_batch_items;
+         batch_id += gridDim.x) {
```
MarcelKoch, line 84: This it not necessary, since the number of thread blocks is set to the number of batches (unless we want to change that). Same for the advanced apply

dpcpp/matrix/batch_dense_kernels.dp.cpp
```diff
@@ -0,0 +1,176 @@
...
+namespace dpcpp {
+/**
+ * @brief The Dense matrix format namespace.
+ *
+ * @ingroup batch_dense
+ */
+namespace batch_dense {
+
+
+#include "dpcpp/matrix/batch_dense_kernels.hpp.inc"
```
MarcelKoch, line 74: is there a benefit from putting the kernels into a separate file?

reference/matrix/batch_dense_kernels.hpp.inc
```diff
@@ -0,0 +1,88 @@
...
+    if (beta != gko::zero<ValueType>()) {
+        for (int row = 0; row < c.num_rows; ++row) {
+            for (int col = 0; col < c.num_rhs; ++col) {
+                c.values[row * c.stride + col] *= beta;
+            }
+        }
+    } else {
+        for (int row = 0; row < c.num_rows; ++row) {
+            for (int col = 0; col < c.num_rhs; ++col) {
+                c.values[row * c.stride + col] *= gko::zero<ValueType>();
```
MarcelKoch, line 74:
```suggestion
                c.values[row * c.stride + col] = gko::zero<ValueType>();
```

reference/test/matrix/batch_dense_kernels.cpp
```diff
@@ -0,0 +1,246 @@
...
+    this->mtx_01->apply(this->b_01.get(), this->x_01.get());
+
+    auto res = gko::batch::unbatch<gko::batch::MultiVector<T>>(this->x_0.get());
+
+    GKO_ASSERT_MTX_NEAR(res[0].get(), this->x_00.get(), 0.);
+    GKO_ASSERT_MTX_NEAR(res[1].get(), this->x_01.get(), 0.);
+}
+
+
+TYPED_TEST(Dense, AppliesLinearCombinationWithSameAlphaToBatchMultiVector)
```
MarcelKoch, line 132: I don't think this test adds much value in combination with the next test.

test/matrix/batch_dense_kernels.cpp
```diff
@@ -0,0 +1,128 @@
...
+        dalpha = gko::clone(exec, alpha);
+        dbeta = gko::clone(exec, beta);
+        expected = MVec::create(
+            ref,
+            gko::batch_dim<2>(batch_size, gko::dim<2>{num_rows, num_vecs}));
+        expected->fill(gko::one<value_type>());
+        dresult = gko::clone(exec, expected);
+    }
+
+    std::ranlux48 rand_engine;
```
MarcelKoch, line 93:
```suggestion
    std::default_random_engine rand_engine;
```
didn't this (or something like this) gave us some trouble a while back?

MarcelKoch (dismissed):
Looks good, but some mostly minor comments left.

include/ginkgo/core/matrix/batch_dense.hpp
```diff
@@ -0,0 +1,384 @@
...
+ * @tparam ValueType  precision of matrix elements
+ *
+ * @ingroup batch_dense
+ * @ingroup mat_formats
+ * @ingroup BatchLinOp
+ */
+template <typename ValueType = default_precision>
+class Dense final : public EnableBatchLinOp<Dense<ValueType>>,
+                    public EnableCreateMethod<Dense<ValueType>>,
+                    public ConvertibleTo<Dense<next_precision<ValueType>>> {
```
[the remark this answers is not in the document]
pratikvn: I think I would like to avoid cases where one has to convert from Dense to MultiVector, and instead prefer people to just directly create it from views, especially as the ConvertibleTo interface in this case adds functions, that I dont think will be useful and is just extra code.

include/ginkgo/core/base/batch_multi_vector.hpp
```diff
@@ -81,21 +90,25 @@ class MultiVector
...
       public EnablePolymorphicAssignment<MultiVector<ValueType>>,
       public EnableCreateMethod<MultiVector<ValueType>>,
-      public ConvertibleTo<MultiVector<next_precision<ValueType>>> {
+      public ConvertibleTo<MultiVector<next_precision<ValueType>>>,
+      public ConvertibleTo<matrix::Dense<ValueType>> {
     friend class EnableCreateMethod<MultiVector>;
     friend class EnablePolymorphicObject<MultiVector>;
     friend class MultiVector<to_complex<ValueType>>;
     friend class MultiVector<next_precision<ValueType>>;
+    friend class matrix::Dense<ValueType>;
```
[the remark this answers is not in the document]
pratikvn: Same argument as above.

include/ginkgo/core/matrix/batch_dense.hpp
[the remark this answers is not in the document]
MarcelKoch: I would agree there, but then the conversion MultiVector->Dense should also be replaced by views.

core/base/batch_multi_vector.cpp
```diff
@@ -290,6 +291,26 @@ void MultiVector<ValueType>::move_to(
 }
 
 
+template <typename ValueType>
+void MultiVector<ValueType>::convert_to(matrix::Dense<ValueType>* result) const
+{
+    auto exec = result->get_executor() != nullptr ? result->get_executor()
+                                                  : this->get_executor();
+    auto tmp = gko::batch::matrix::Dense<ValueType>::create_const(
```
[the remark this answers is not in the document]
pratikvn: But it should be copied to the executor of result, but you are right that the array_view is on the incorrect executor.

cuda/matrix/batch_dense_kernels.cu
pratikvn: The kernels themselves will be reused in solvers, while the kernel launcher will not be and is used only to reduce duplication.

core/base/batch_multi_vector.cpp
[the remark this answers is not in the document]
MarcelKoch: IMO, result should handle that in it's `copy_from`. Alternatively, you could use two executors, one for the tmp dense and one for the array view.

core/matrix/batch_struct.hpp
pratikvn: Yes, in solvers. If it turns out that I dont need them, I will remove them later.

core/matrix/batch_struct.hpp
pratikvn: Again will be used in solvers, to uniformly get the storage for different matrix formats and other batch objects.

core/matrix/batch_dense.cpp
pratikvn: array has a overloaded copy operator with the necessary checks, so it is cleaner to use that.

dpcpp/matrix/batch_dense_kernels.dp.cpp
pratikvn: Same as for cuda/hip, will be useful in solvers.

core/matrix/batch_dense.cpp
MarcelKoch: right, I somehow forgot about that.

upsj: side-note: This can mess with submatrices/strided storage, since it also overwrites padding if the target is already allocated. I'm not sure if we do that correctly in `matrix::Dense`, but it should be mentioned.

test/matrix/batch_dense_kernels.cpp
upsj: some compilation issues with MSVC, fixed in https://github.com/ginkgo-project/ginkgo/pull/954

include/ginkgo/core/matrix/batch_dense.hpp
```diff
@@ -0,0 +1,384 @@
...
+    template <typename ValuesArray>
+    Dense(std::shared_ptr<const Executor> exec, const batch_dim<2>& size,
+          ValuesArray&& values)
+        : EnableBatchLinOp<Dense>(exec, size),
+          values_{exec, std::forward<ValuesArray>(values)}
+    {
+        // Ensure that the values array has the correct size
+        auto num_elems = compute_num_elems(size);
+        GKO_ENSURE_IN_BOUNDS(num_elems, values_.get_num_elems() + 1);
+    }
```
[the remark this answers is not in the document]
pratikvn: This is a templated constructor, so it needs to be in the header.

include/ginkgo/core/base/batch_multi_vector.hpp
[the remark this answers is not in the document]
pratikvn: No, I mean no explicit `create_dense_view`, but use views on the underlying data and use `MultiVector::create(..., array_view)` instead.

common/cuda_hip/matrix/batch_dense_kernels.hpp.inc
pratikvn: I think it does not hurt to be explicit, as this is used in other places such as solvers as well.

thoasm (commented):
Pretty much only nits and two questions:
1. Why is it `batch_dim` and not `batch::dim`?
2. Wouldn't a type-alias be better (which fixes the bit-size of the integer) in `core/matrix/batch_struct.hpp`?

include/ginkgo/core/base/batch_multi_vector.hpp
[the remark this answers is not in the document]
MarcelKoch: But there is `create_multi_vector_view` already. I really don't think we should have only one of those. Either we have both, or none. Anything else is inconsistent.

include/ginkgo/core/matrix/batch_dense.hpp
```diff
@@ -0,0 +1,384 @@
...
+    friend class EnablePolymorphicObject<Dense, BatchLinOp>;
+    friend class Dense<to_complex<ValueType>>;
+    friend class Dense<next_precision<ValueType>>;
+
+public:
+    using EnableBatchLinOp<Dense>::convert_to;
+    using EnableBatchLinOp<Dense>::move_to;
+
+    using value_type = ValueType;
+    using index_type = int32;
```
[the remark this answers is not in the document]
pratikvn: Yes, for batched cases, you dont need index values more than 2e9, but indeed this `using` is probably is not necessary for batched dense.

dpcpp/matrix/batch_dense_kernels.hpp.inc
```diff
@@ -0,0 +1,98 @@
...
+
+    for (int row = subgroup_id; row < mat.num_rows; row += num_subgroups) {
+        ValueType temp = zero<ValueType>();
+        for (int j = subgroup.get_local_id(); j < mat.num_cols;
+             j += subgroup_size) {
+            const ValueType val = mat.values[row * mat.stride + j];
+            temp += val * b.values[j];
+        }
+
+        temp = ::gko::kernels::dpcpp::reduce(
```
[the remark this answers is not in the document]
pratikvn: This is in general expected to be used as a header file and included in a variety of namespaces, so I think it is better to fully specify the namespaces, to avoid any confusion.

include/ginkgo/core/base/batch_multi_vector.hpp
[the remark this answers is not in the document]
pratikvn: I will remove the `create_multi_vector_view`.

pratikvn:
format!

include/ginkgo/core/matrix/batch_dense.hpp
```diff
@@ -0,0 +1,384 @@
...
+     * @param size  the dimensions of the matrix
+     * @param values  the value array of the matrix
+     *
+     * @return A smart pointer to the constant matrix wrapping the input
+     * array (if it resides on the same executor as the matrix) or a copy of the
+     * array on the correct executor.
+     */
+    static std::unique_ptr<const Dense<value_type>> create_const(
+        std::shared_ptr<const Executor> exec, const batch_dim<2>& sizes,
+        gko::detail::const_array_view<ValueType>&& values);
```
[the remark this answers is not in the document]
pratikvn: MSVC for some reason has an issue with lookup when both both `gko::batch::detail` and `gko::detail` exist. And it only seems to look in `gko::batch::detail` without the full qualification.

include/ginkgo/core/matrix/batch_dense.hpp
```diff
@@ -0,0 +1,384 @@
...
+    void apply(const MultiVector<value_type>* alpha,
+               const MultiVector<value_type>* b,
+               const MultiVector<value_type>* beta,
+               MultiVector<value_type>* x) const
+    {
+        this->apply_impl(alpha, b, beta, x);
+    }
+
+private:
+    inline size_type compute_num_elems(const batch_dim<2>& size)
```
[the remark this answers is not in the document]
pratikvn: Its definitely an option, but I feel it might create some significant confusion whether the dim object is a batch::dim or a gko::dim and everything will need to be clarified, which is more annoying to handle.

reference/test/matrix/batch_dense_kernels.cpp
yhmtsai: agree because the alpha is still duplicated to 2 batch_item

reference/test/matrix/batch_dense_kernels.cpp
```diff
@@ -0,0 +1,244 @@
...
+        BMVec::create(this->exec, gko::batch_dim<2>{2, gko::dim<2>{3, 3}});
+    auto alpha =
+        BMVec::create(this->exec, gko::batch_dim<2>{2, gko::dim<2>{2, 1}});
+    auto beta =
+        BMVec::create(this->exec, gko::batch_dim<2>{2, gko::dim<2>{1, 1}});
+
+    ASSERT_THROW(
+        this->mtx_0->apply(alpha.get(), res.get(), beta.get(), this->x_0.get()),
+        gko::DimensionMismatch);
+}
```
yhmtsai, line 244: I was thinking of the test for multiple rhs throw not_implemented, but maybe it's unnecessary

include/ginkgo/core/matrix/batch_dense.hpp
```diff
@@ -0,0 +1,368 @@
...
+     *
+     * @param batch_id  the id of the batch item.
+     *
+     * @return the pointer to the array of values
+     */
+    value_type* get_values_for_item(size_type batch_id) noexcept
+    {
+        GKO_ASSERT(batch_id < this->get_num_batch_items());
+        return values_.get_data() +
+               this->get_size().get_cumulative_offset(batch_id);
```
yhmtsai, line 211: I still think the batch_dim is not a good place for get_cumulative_offset especially when we add the stride in the future.

include/ginkgo/core/matrix/batch_dense.hpp
```diff
@@ -0,0 +1,368 @@
...
+     */
+    void apply(const MultiVector<value_type>* b,
+               MultiVector<value_type>* x) const
+    {
+        this->apply_impl(b, x);
+    }
+
+    /**
+     * Apply the matrix to a multi-vector with a linear combination of the given
+     * input vector. Represents the matrix vector multiplication, x = alpha* A *
```
yhmtsai, line 272:
```suggestion
     * input vector. Represents the matrix vector multiplication, x = alpha * A *
```

core/test/matrix/batch_dense.cpp
```diff
@@ -0,0 +1,476 @@
...
+    auto m_ref =
+        gko::batch::create_from_item<gko::batch::matrix::Dense<value_type>>(
+            this->exec,
+            std::vector<DenseMtx*>{mat1.get(), mat2.get(), mat1.get(),
+                                   mat2.get(), mat1.get(), mat2.get()});
+
+    auto m2 = gko::batch::duplicate<gko::batch::matrix::Dense<value_type>>(
+        this->exec, 3, m.get());
+
+    GKO_ASSERT_BATCH_MTX_NEAR(m2.get(), m_ref.get(), 1e-14);
```
yhmtsai, line 319:
```suggestion
    GKO_ASSERT_BATCH_MTX_NEAR(m2.get(), m_ref.get(), 0);
```
It only copy the data, so it should not have the error

core/test/matrix/batch_dense.cpp
```diff
@@ -0,0 +1,476 @@
...
+    ASSERT_EQ(m->get_common_size(), gko::dim<2>(3, 3));
+    EXPECT_EQ(m->at(0, 0), value_type{1.0});
+    EXPECT_EQ(m->at(0, 1), value_type{1.0});
+    EXPECT_EQ(m->at(0, 2), value_type{0.0});
+    ASSERT_EQ(m->at(0, 3), value_type{2.0});
+    EXPECT_EQ(m->at(0, 4), value_type{4.0});
+    EXPECT_EQ(m->at(1, 0), value_type{1.0});
+    EXPECT_EQ(m->at(1, 1), value_type{2.0});
+    EXPECT_EQ(m->at(1, 2), value_type{-1.0});
+    ASSERT_EQ(m->at(1, 3), value_type{3.0});
```
yhmtsai, line 392:
```suggestion
    EXPECT_EQ(m->at(0, 3), value_type{2.0});
    EXPECT_EQ(m->at(0, 4), value_type{4.0});
    EXPECT_EQ(m->at(1, 0), value_type{1.0});
    EXPECT_EQ(m->at(1, 1), value_type{2.0});
    EXPECT_EQ(m->at(1, 2), value_type{-1.0});
    EXPECT_EQ(m->at(1, 3), value_type{3.0});
```
miss some entries check

yhmtsai (approved):

core/matrix/batch_dense.cpp
```diff
@@ -0,0 +1,194 @@
...
+    // Otherwise, it results in a compile error.
+    return (*other).create_with_same_config();
+}
+
+
+template <typename ValueType>
+std::unique_ptr<Dense<ValueType>> Dense<ValueType>::create_with_same_config()
+    const
+{
+    return Dense<ValueType>::create(this->get_executor(), this->get_size());
```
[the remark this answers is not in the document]
pratikvn: no stride for batch::Dense yet. Might be added later.

MarcelKoch (changes requested):
Not much left. I would like to get the connection between `Dense` and `Multivector` clarified, and I think adding the validate function to BatchedLinOp would be nice.

include/ginkgo/core/base/batch_multi_vector.hpp
```diff
@@ -81,21 +90,25 @@ class MultiVector
     : public EnablePolymorphicObject<MultiVector<ValueType>>,
       public EnablePolymorphicAssignment<MultiVector<ValueType>>,
       public EnableCreateMethod<MultiVector<ValueType>>,
-      public ConvertibleTo<MultiVector<next_precision<ValueType>>> {
+      public ConvertibleTo<MultiVector<next_precision<ValueType>>>,
+      public ConvertibleTo<matrix::Dense<ValueType>> {
```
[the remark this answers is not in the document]
pratikvn: The ConvertibleTo interface is necessary (only in this PR) to have batch::initialize<> working. I have updated it in the next PR and will remove this ConvertibleTo<Dense> in the batch::Ell PR and everything will then go through matrix_data and hence no unnecessary ConvertibleTo interfaces will exist.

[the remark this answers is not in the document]
MarcelKoch: Thanks for clarifying!

dpcpp/matrix/batch_dense_kernels.hpp.inc
```diff
@@ -0,0 +1,98 @@
...
+    const gko::batch::multi_vector::batch_item<const ValueType>& b,
+    const gko::batch::multi_vector::batch_item<ValueType>& x,
+    sycl::nd_item<3>& item_ct1)
+{
+    constexpr auto tile_size = config::warp_size;
+    auto subg =
+        group::tiled_partition<tile_size>(group::this_thread_block(item_ct1));
+    const auto subgroup = static_cast<sycl::sub_group>(subg);
+    const int subgroup_id = subgroup.get_group_id();
+    const int subgroup_size = subgroup.get_local_range().size();
```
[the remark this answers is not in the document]
pratikvn: Noted in #1376 and will tackle it later in a general fashion.

yhmtsai:
forgot to mention. missing the construction in test_install

include/ginkgo/core/matrix/batch_dense.hpp
pratikvn: Removed from batch_dim now and added to the classes themselves.

MarcelKoch (approved):

tcojean (commented):
First part of my review.

core/matrix/batch_dense_kernels.hpp
```diff
@@ -0,0 +1,84 @@
...
+#include "core/base/kernel_declaration.hpp"
+
+
+namespace gko {
+namespace kernels {
+
+
+#define GKO_DECLARE_BATCH_DENSE_SIMPLE_APPLY_KERNEL(_type)         \
+    void simple_apply(std::shared_ptr<const DefaultExecutor> exec, \
+                      const batch::matrix::Dense<_type>* a,        \
```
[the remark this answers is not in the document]
pratikvn: Do you mean `gko/matrix/dense.hpp` ? I dont use it here. I use only the batched classes and the Multivector.

reference/matrix/batch_struct.hpp
```diff
@@ -0,0 +1,100 @@
...
+#include "core/base/batch_struct.hpp"
+
+
+namespace gko {
+namespace kernels {
+/**
+ * @brief A namespace for shared functionality between omp and reference
+ *  executors.
+ */
+namespace host {
```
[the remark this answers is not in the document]
pratikvn: Currently, yes, but I am not sure I want to unify reference code and dpcpp code.

test/matrix/batch_dense_kernels.cpp
```diff
@@ -0,0 +1,128 @@
...
+    std::unique_ptr<BMVec> dalpha;
+    std::unique_ptr<BMVec> dbeta;
+};
+
+
+TEST_F(Dense, SingleVectorApplyIsEquivalentToRef)
+{
+    set_up_apply_data(1);
+
+    x->apply(y.get(), expected.get());
```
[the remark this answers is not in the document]
pratikvn: I think when reading `GKO_ASSERT_BATCH_MTX_NEAR` `expected` makes more sense ? I
 think it makes it clear which object is being compared to what. 

I will change `x` to `mat`.

core/matrix/batch_struct.hpp
```diff
@@ -0,0 +1,124 @@
...
+    int32 num_rows;
+    int32 num_cols;
+};
+
+
+/**
+ * A 'simple' structure to store a global uniform batch of dense matrices.
+ */
+template <typename ValueType>
+struct uniform_batch {
```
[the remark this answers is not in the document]
pratikvn: I will check. But maybe, at a later point once the solver and preconditioner has been implemented. I have noted it down in the technical debt issue (https://github.com/ginkgo-project/ginkgo/issues/1376)

common/cuda_hip/matrix/batch_dense_kernels.hpp.inc
```diff
@@ -0,0 +1,164 @@
...
+        if (subgroup.thread_rank() == 0) {
+            x[row] = temp;
+        }
+    }
+}
+
+template <typename ValueType>
+__global__ __launch_bounds__(
+    default_block_size,
+    sm_oversubscription) void simple_apply_kernel(const gko::batch::matrix::
```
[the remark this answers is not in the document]
pratikvn: Thanks. This has been noted in #1376 and will be tackled in a performance related PR in the future.

include/ginkgo/core/matrix/batch_dense.hpp
```diff
@@ -0,0 +1,372 @@
...
+     * @return  a gko::matrix::Dense object with the data from the batch item
+     * at the given index.
+     */
+    std::unique_ptr<unbatch_type> create_view_for_item(size_type item_id);
+
+    /**
+     * @copydoc create_view_for_item(size_type)
+     */
+    std::unique_ptr<const unbatch_type> create_const_view_for_item(
+        size_type item_id) const;
```
tcojean, line 125: Somewhat unrelated, but everywhere in Ginkgo, do we actually need to have different names for the const version? Usually the compiler does a good job at selecting the correct version with const overloading.
https://isocpp.org/wiki/faq/const-correctness#const-overloading

The stdlib also usually has the same method for both const & non const access, see e.g.:
https://en.cppreference.com/w/cpp/ranges/view_interface/data
https://en.cppreference.com/w/cpp/container/vector/data

I see that we never mentioned this in either #402 or #1400

include/ginkgo/core/matrix/batch_dense.hpp
```diff
@@ -0,0 +1,372 @@
...
+    {
+        return values_.get_data()[linearize_index(batch_id, idx)];
+    }
+
+    /**
+     * @copydoc Dense::at(size_type, size_type, size_type)
+     */
+    ValueType at(size_type batch_id, size_type idx) const noexcept
+    {
+        return values_.get_const_data()[linearize_index(batch_id, idx)];
```
tcojean, line 209: Don't you also need `GKO_ASSERT(batch_id < this->get_num_batch_items());` here?

include/ginkgo/core/matrix/batch_dense.hpp
```diff
@@ -0,0 +1,372 @@
...
+    /**
+     * Get the cumulative storage size offset
+     *
+     * @param batch_id the batch id
+     *
+     * @return the cumulative offset
+     */
+    size_type get_cumulative_offset(size_type batch_id) const
+    {
+        return batch_id * this->get_common_size()[0] *
```
tcojean, line 136: I guess as assert could go here as well.

core/matrix/batch_dense.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+}  // namespace
+}  // namespace dense
+
+
+template <typename ValueType>
+std::unique_ptr<gko::matrix::Dense<ValueType>>
+Dense<ValueType>::create_view_for_item(size_type item_id)
+{
+    auto exec = this->get_executor();
```
tcojean, line 71: Maybe also an early assert here? Although the `get_values_for_item` should trigger an error.

core/test/matrix/batch_dense.cpp
```diff
@@ -0,0 +1,484 @@
...
+            std::vector<DenseMtx*>{mat1.get(), mat1.get(), mat1.get()});
+    auto m =
+        gko::batch::create_from_item<gko::batch::matrix::Dense<value_type>>(
+            this->exec, 3, mat1.get());
+
+    GKO_ASSERT_BATCH_MTX_NEAR(bat_m.get(), m.get(), 0);
+}
+
+
+TYPED_TEST(Dense, CanBeConstructedByDuplicatingDenseMatrices)
```
tcojean, line 296: nit: this test and the previous one don't respect AAA.

core/test/matrix/batch_dense.cpp
```diff
@@ -0,0 +1,484 @@
...
+    using DenseMtx = typename TestFixture::DenseMtx;
+    using size_type = gko::size_type;
+
+    auto mat1 = gko::initialize<DenseMtx>({{-1.0, 2.0, 3.0}, {-1.5, 2.5, 3.5}},
+                                          this->exec);
+    auto mat2 = gko::initialize<DenseMtx>({{1.0, 2.5, 3.0}, {1.0, 2.0, 3.0}},
+                                          this->exec);
+
+    auto m =
+        gko::batch::create_from_item<gko::batch::matrix::Dense<value_type>>(
```
tcojean, line 308: We could use `ASSERT_DEATH` or `ASSERT_ANY_THROW` to test the out of bound calls.

tcojean (approved):
LGTM. Some small points.

thoasm (approved):
LGTM!

include/ginkgo/core/matrix/batch_dense.hpp
pratikvn: I think we discussed this sometime back and decided that for uniformity, we have named functions for const qualified related operations. I think this was around the time we were creating `const_view` in `gko::array`

PR #1414, opened 2023-10-04
use arithmetic_type more in mixed-precision ELL

yhmtsai:
It uses the arithmeic_type until writing the result to the global memory in the mixed precision ELL


upsj (approved):
LGTM! Thanks for noticing the issues. Side-note: With some carefully-designed inputs (e.g. matrix filled with float values cast to double compared against double matrix), it should be possible to check the achieved precision, highlighting where we use the wrong type for arithmetic.

thoasm (approved):
LGTM!

PR #1417, opened 2023-10-10
Fix missing destruction of rand_generator from cuda/hip

### LINKED ISSUE #1406: CURAND generators not destroyed
Not sure if destruction is necessary, the API reference does not state that. But it has a function for destruction of generators and on my device I'm getting an a curand allocation exception after a couple of iterations of Idr solver with Cuda executor.

yhmtsai:
This is raised by https://github.com/ginkgo-project/ginkgo/issues/1406

Because we do not face the allocation exception from IDR test in ci, I am not entirely sure whether it fixes the issue.
I close the issue after this pr, but I will add a comment about reopening it when it is still an issue.
At least, we should destroy the rand generator.

close #1406


upsj (approved):

MarcelKoch (approved):

tcojean (approved):

upsj:
rebase!

yhmtsai:
rebase!

yhmtsai:
rebase!

PR #1419, opened 2023-10-11
Fix PAPI segmentation fault

yhmtsai:
This PR destroys the logger before shutdown papi_sde_shutdown. It sometimes lead to segmentation faults before.
papi_sde_shutdown seems to unregister the counter during shutdown such that the `~papi_queue` reach the invalid memory.
I don't know whether papi_sde_shutdown always unregister counter in different versions/for different types, so I only reorder the destruction.

### DIFF
--- a/include/ginkgo/core/log/papi.hpp
+++ b/include/ginkgo/core/log/papi.hpp
@@ -209,8 +209,9 @@ public:
            const Logger::mask_type& enabled_events = Logger::all_events_mask)
     {
         return std::shared_ptr<Papi>(new Papi(enabled_events), [](auto logger) {
-            papi_sde_shutdown(logger->get_handle());
+            auto handle = logger->get_handle();
             delete logger;
+            papi_sde_shutdown(handle);
         });
     }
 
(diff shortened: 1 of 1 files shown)


include/ginkgo/core/log/papi.hpp
```diff
@@ -209,8 +209,9 @@ class Papi : public Logger {
            const Logger::mask_type& enabled_events = Logger::all_events_mask)
     {
         return std::shared_ptr<Papi>(new Papi(enabled_events), [](auto logger) {
-            papi_sde_shutdown(logger->get_handle());
+            auto handle = logger->get_handle();
             delete logger;
+            papi_sde_shutdown(handle);
```
upsj, line 214: This should be calling the other `create` function

upsj (approved):

tcojean (approved):

include/ginkgo/core/log/papi.hpp
yhmtsai: Do you mean the same fix for the deprecated `create`? I have done it.
Or, you mean different things

upsj: I mean there should only be a single place this needs to be fixed, the deprecated version should call the normal version instead of reimplementing it. (also it's probably more appropriate to put this into the source file

yhmtsai:
rebase!

PR #1420, opened 2023-10-12
auto add tweak version

yhmtsai:
This pr adds github action with a new tweak version to allow user to follow the develop changes.

the pr is not completed with the components.
I collected some possible ways

tweak version:
CMake find_package can use tweak version
Although CMake only allows `major.minor.patch` in `write_basic_package_version_file`, it should only affect COMPATIBILITY ExactVersion mode, which is not the case for us.
Or, we only use it for ginkgo header such that user can switch by macro.

use patch version?
use patch version for release and develop. develop 1.7.10 -> develop 1.7.11 -> release 1.7.12,
but I think it destroys the patch only for bug fix

action possibility: (github action can not push to protected branch, so it needs to use PAT)
1. use comment with !merge to merge: (it need to check some status beforehand in script)
  1.1. bump the version only in local branch then merge (require PAT bypassing protection rule)
    - commit-status: (last_author_commit (develop(#pr)) -> bump_version (develop) -> merge (develop(#pr)))
    - pro: does not change the pull request content. 
    - con: PAT, mixed the develop changes and pr changes, complex check to ensure the branch meet the requirement
  1.2 bump the version and push back to the pr branch and then merge it immediately (We can also do that manually before merge or only provide the action for adding version)
    - commit-status: (last_author_commit (develop(#pr)) -> bump_version (develop(#pr)) -> merge (develop(#pr)))
    - pro: not need PAT (I think?), bump version is in the pull request 
    - con: the pr status on the pr page does not contain ci job running, complex check to ensure the branch meet the requirement
2. bump the version when branch is merged (require PAT)
  - commit-status: (last_author_commit (develop(#pr)) -> merge(develop(#pr)) -> bump_version(develop))
  - pro: also work for the external PR without the permission for upstream
  - con: PAT, always show the bumping version commit as the latest commit, the pipeline is from merge not from bumping
3. embed the bumping version into merge commit (I like the type, but someone said it is evil)

Currently, I prefer setting up a PAT for action (option 2) because we do not need to worry about the status check in merge! comment.


MarcelKoch:
Could you elaborate what this is for? If I understand correctly, this will just give us the number of the merged PRs.

yhmtsai:
Yes, it gives the number of merged pr.
It's mainly for users who want to use the latest function or would like to avoid deprecation warnings.
```
#if GINKGO_VERSION > version
new_function or new function signature
#else
old function or the deprecated after certain version
#endif
```

MarcelKoch:
But that is already possible? The version in develop is already the one for the next release. So it would only make sense if a user wants to fix to a certain commit of `develop` before a release.

yhmtsai:
Actually, if we only aim for deprecation warnings, maybe we can force us to add a predefined macro like `#define GKO_DEPRECATED_*` with the corresponding header. It may indeed be a good approach with a more meaningful name.
tweak version may still be needed in the following case
- we would like to have a stable or comparable version (not sha)
- It's also available when downing the compressed file
- new function (do it also `#GINKGO_NEW_*` ?)

yhmtsai:
Yes, you are right.
I guess it is also related to that we usually have a large number of updates between releases.

MarcelKoch:
What would the macro be used for? Is it something to wrap the `[[deprecated]]`?

> Yes, you are right. I guess it is also related to that we usually have a large number of updates between releases.

IMO, then the answer should be more releases.

upsj:
I don't think avoiding deprecation warnings is a good idea. They are there explicitly to tell users that they are using functionality that will go away in the future, so they need to update their code. If they want to use an older interface, they can use an older version of Ginkgo.

yhmtsai:
they update the code but still support the old ginkgo, so they can not just remove the deprecated one

MarcelKoch:
Your suggestion of using that compile time switch is still possible without the changes in this PR.

upsj:
That is not our responsibility. If people want to support a wide range of versions, they need to take care of switching between different interfaces themselves, like we do with cuSPARSE.

yhmtsai:
Yes, I am not saying we need to do switch for users but we need to provide a way such that they can switch the function.
cuSPARSE only provides the release to user, so they have version for each downloadable package.
ginkgo develop does not provide it with the downloadable package in develop.
More releases can somehow reduce the problem although I think it is sometimes hard especially we would like to have something released with the components.
I do not think we have a way to compile-time switch the version in different develop commit?

upsj:
We never delete interfaces between minor releases, so they can use the version number to decide which code to use. `develop` itself cannot be stable, otherwise we would be regularly locking ourselves into untested interfaces.

MarcelKoch:
> I do not think we have a way to compile-time switch the version in different develop commit?

That is the point I have trouble understanding. Why should we have different versions for develop?

yhmtsai:
Like a third party using ginkgo with latest feature (currently only in develop), they do not know the user ginkgo develop version.
the version check can not avoid using the old develop from user side.
I think all interface should be tested before merge?
but I agree with @upsj in terms of showing some unstable interface/changes to user although it can still be solved with develop version
More release somehow turns this tweak version into minor version and then we need to take care of the interface

upsj:
`develop` is a moving target, so people need to always build ginkgo from the latest state if they want to rely on it. We always update the version number after a release, so it's possible to recognize the pre-release version from that. If we add a new version number to it, we are implying that there is actually such an intermediate version, and I think that's a bad idea.

yhmtsai:
It's indeed the intermediate version without any guarantee.  
maybe not consider it as a version number if it's confusing with the semantic version.
It's just a comparable increasing tag.

PR #1421, opened 2023-10-12
Update tau.cpp

khuck:
Fixing order of includes. The ifdef check has to happen after config.h has been included, or else the perfstubs header file won't get included. Also, adding an argument name for the `end_tau()` function.

### DIFF
--- a/core/log/tau.cpp
+++ b/core/log/tau.cpp
@@ -30,15 +30,13 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ******************************<GINKGO LICENSE>*******************************/
 
+#include <ginkgo/core/base/exception_helpers.hpp>
+#include <ginkgo/core/log/profiler_hook.hpp>
+
 #if GKO_HAVE_TAU
 #define PERFSTUBS_USE_TIMERS
 #include <perfstubs_api/timer.h>
 #endif
-
-
-#include <ginkgo/core/base/exception_helpers.hpp>
-#include <ginkgo/core/log/profiler_hook.hpp>
-
 
 namespace gko {
 namespace log {
@@ -56,7 +54,7 @@ void begin_tau(const char* name, profile_event_category)
 }
 
 
-void end_tau(const char*, profile_event_category)
+void end_tau(const char* name, profile_event_category)
 {
     PERFSTUBS_STOP_STRING(name);
 }


core/log/tau.cpp
```diff
@@ -30,16 +30,14 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ******************************<GINKGO LICENSE>*******************************/
 
+#include <ginkgo/core/base/exception_helpers.hpp>
+#include <ginkgo/core/log/profiler_hook.hpp>
```
yhmtsai, line 34:
```suggestion
#include <ginkgo/core/base/exception_helpers.hpp>
#include <ginkgo/core/log/profiler_hook.hpp>
#include <ginkgo/config.hpp>
```
nit: because it uses GKO_HAVE_TAU

yhmtsai (approved):
Good Catch!
After this change, do you run you code with tau logger successfully?

tcojean (approved):

pratikvn (approved):
Thanks for the fix! LGTM.

pratikvn:
@khuck, if you can confirm that the you are able to successfully run with the TAU logger with these changes, then I will go ahead and merge this PR.

core/log/tau.cpp
yhmtsai: ah. this order fits the ginkgo include order more.
profiler_hook.hpp should be the main header which includes the config.
```
#include <ginkgo/core/log/profiler_hook.hpp>


#if GKO_HAVE_TAU
#define PERFSTUBS_USE_TIMERS
#include <perfstubs_api/timer.h>
#endif


```
@tcojean do you think we still need to include the config after exception_herlpers?

khuck:
THanks - I haven't tried running yet, I'll do that now...

upsj (changes requested):
We should base this on `develop` and not `batch-develop`, otherwise LGTM

pratikvn:
@upsj, we cannot do that anymore as `batch-develop` and `develop` have diverged and `batch-develop` cannot be rebased on `develop` anymore. But we should create a PR to have these changes in `develop` as well.

upsj:
Sure we can, the patch applies just fine on `develop`.

@khuck could you please reset your branch onto `develop`, cherry-pick 2b16c0f162a5d16cc5834adcbd384e95cff61e64 and push again? Then we can change the base of the PR to our development branch.

yhmtsai:
I thought we just need two pr: one for batch-develop and another for develop.
after this can be merged, create another pr for develop

upsj:
`batch-develop` was last updated 2.5 months ago, I think we should focus on our active development branch.

pratikvn:
@khuck , thank you again for the PR. Please create a PR for our main development branch, `develop` as well, as these changes are necessary there as well.

PR #1423, opened 2023-10-13
Refactor code duplications in dense_kernels

upsj:
Pulled out of #1415, this refactors the dense reference tests to get instantiated for all index types automatically.


reference/test/matrix/dense_kernels.cpp
```diff
@@ -780,3510 +779,2192 @@ TYPED_TEST(Dense, MovesToPrecision)
 }
 
 
-TYPED_TEST(Dense, ConvertsToCoo32)
+TYPED_TEST(Dense, ConvertsEmptyToPrecision)
```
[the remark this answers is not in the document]
upsj: Done, I first changed the tests and then grouped them together.

reference/test/matrix/dense_kernels.cpp
```diff
@@ -780,3061 +779,2103 @@ TYPED_TEST(Dense, MovesToPrecision)
...
+class DenseWithIndexType
+    : public Dense<
+          typename std::tuple_element<0, decltype(ValueIndexType())>::type> {
+public:
+    using value_type =
+        typename std::tuple_element<0, decltype(ValueIndexType())>::type;
+    using index_type =
+        typename std::tuple_element<1, decltype(ValueIndexType())>::type;
+
+    index_type invalid_index = gko::invalid_index<index_type>();
```
MarcelKoch, line 1322: is that one even used? In the `assert_***_eq_mtx` functions the invalid index is defined again.

reference/test/matrix/dense_kernels.cpp
```diff
@@ -780,3061 +779,2103 @@ TYPED_TEST(Dense, MovesToPrecision)
...
-TYPED_TEST(Dense,
-           SquareMatrixInversePermuteIntoDenseFailsForWrongPermutationSize)
+TYPED_TEST(Dense, ScaleAddIdentityRectangular)
 {
-    using Mtx = typename TestFixture::Mtx;
-    auto exec = this->mtx5->get_executor();
-    gko::array<gko::int32> permute_idxs{exec, {0, 1}};
+    using T = typename TestFixture::value_type;
+    using Vec = typename TestFixture::Mtx;
+    using MixedVec = typename TestFixture::MixedMtx;
```
MarcelKoch, line 1300: unused
```suggestion
```

MarcelKoch (approved):
~I think this is still too complex to review thoroughly. But I think it might be generally impossible to get this into an easy-to-review state due to the reordering. The only way I can think of is to split this into many more smaller PRs (probably 10 each with 200 LOC changes). But that is unreasonable in our workflow.~

~So I would be fine merging this, as long as the tests pass.~
LGTM, reviewing the commits individually in VSCode makes it bearable.

reference/test/matrix/dense_kernels.cpp
```diff
@@ -780,3061 +779,2103 @@ TYPED_TEST(Dense, MovesToPrecision)
...
-    using T = typename TestFixture::value_type;
-    using ComplexMtx = typename TestFixture::ComplexMtx;
+    using Mtx = typename TestFixture::Mtx;
+    using value_type = typename TestFixture::value_type;
+    using index_type = typename TestFixture::index_type;
     auto exec = this->mtx5->get_executor();
+    gko::array<index_type> inverse_permute_idxs{exec, {1, 2, 0}};
 
-    auto complex_mtx = ComplexMtx::create(exec);
+    auto inv_row_permute =
```
MarcelKoch, line 2694: nit: unnecessary rename, especially since that is not done consistently.

pratikvn (approved):
LGTM. Only concern is that the tests that used stride=num_cols were updated to use non-standard stride instead, for example for ELL and Hybrid. But in general, it is probably okay as that is the more general case.

yhmtsai (approved):
~~The tests with DenseWithIndexType are in the middle of the test with Dense.
one more commit to reorder them again?~~
```
class Dense{};

TYPED_TEST(Dense, ...)

class DenseWithIndexType{};

TYPED_TEST(DenseWithIndexType)

class DenseComplex{};

TYPED_TEST(DenseComplex)
```
~~Also make the fixture close to the testing~~
Ignore this comment. I review the code in the reversed order

PR #1422, opened 2023-10-13
Backport TAU logger fix

upsj:
This is a cherry-pick of #1421

PR #1424, opened 2023-10-13
Update tau.cpp

khuck:
Fixing order of includes. The ifdef check has to happen after config.h has been included, or else the perfstubs header file won't get included. Also, adding an argument name for the `end_tau()` function.

This is a cherry-pick from the batch-develop branch.


upsj:
@khuck Thank you again! We didn't want to impose any more work on you, so we already created a PR ourselves, which was merged recently #1422.

PR #1427, opened 2023-10-13
WIP read_distributed with local and non_local_data

greole:
This PR allows to read distributed from pregenerated local and non_local_data


upsj:
Suggestion to make writing such functionality and passing communication data easier: We should add a `distributed::RowGatherer` (or RowGather? Gatherer is a handful) LinOp that does what `communicate` does inside `distributed::Matrix`. It gets constructed from an array of global indices and a corresponding Partition. Additionally, we should have the dual `RowScatter` that scatter-adds the values. Combining these two, we get row-distributed, column-distributed and overlapping distributed matrices.

That gives us a single place where we construct communication patterns and deal with MPI, everything else just passes information through.

cc @MarcelKoch

PR #1431, opened 2023-10-15
Add a two-level Schwarz preconditioner

pratikvn:
This PR uses the distributed coarse level generation from PGM to use as a coarse level for the additive Schwarz preconditioner. The only requirement is that the galerkin product generator ($RAP$) generate a `multigrid::MultigridLevel` which is a triplet of `restrict`, `prolong` and `coarse` operators. The user additionally also needs to set the solver for the coarse level.

### TODO

~~+ [ ] Add options to switch between multiplicative and additive~~
~~+ [ ] Options for arbitrary number of levels.~~

### Possible issues
1. The coarse level solve significantly reduces the number of iterations, but can be extremely expensive. Maybe makes sense to apply coarse solve, only a few times instead of every preconditioner apply (need to be careful of consistency of preconditioner apply) ? 
2. Weighting between the local solver solution and the coarse solution is unclear. 
3. Though the coarse solve and the local solve are independent, we have no way of performing them in parallel. This will probably require a rewrite of the apply interface to enable asynchronicity, in addition to support for multiple streams/queues.

### DIFF
--- a/core/distributed/preconditioner/schwarz.cpp
+++ b/core/distributed/preconditioner/schwarz.cpp
@@ -183,6 +183,15 @@ void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::generate(
         this->set_solver(
             gko::share(parameters_.local_solver->generate(local_matrix)));
     }
+
+    auto dist_mat =
+        as<experimental::distributed::Matrix<ValueType, LocalIndexType,
+                                             GlobalIndexType>>(system_matrix);
+
+    if (parameters_.coarse_solver_factory) {
+        this->coarse_solver_ = as<multigrid::MultigridLevel>(
+            share(parameters_.coarse_solver_factory->generate(dist_mat)));
+    }
 }
 
 
--- a/core/distributed/preconditioner/schwarz.cpp
+++ b/core/distributed/preconditioner/schwarz.cpp
@@ -87,10 +87,37 @@ template <typename VectorType>
 void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::apply_dense_impl(
     const VectorType* dense_b, VectorType* dense_x) const
 {
+    using Vector = matrix::Dense<ValueType>;
+    using dist_vec = experimental::distributed::Vector<ValueType>;
     auto exec = this->get_executor();
+
     if (this->local_solver_ != nullptr) {
         this->local_solver_->apply(gko::detail::get_local(dense_b),
                                    gko::detail::get_local(dense_x));
+    }
+
+    if (this->coarse_solver_ != nullptr && this->galerkin_ops_ != nullptr) {
+        auto restrict = this->galerkin_ops_->get_restrict_op();
+        auto prolong = this->galerkin_ops_->get_prolong_op();
+        auto coarse =
+            as<experimental::distributed::Matrix<ValueType, LocalIndexType,
+                                                 GlobalIndexType>>(
+                this->galerkin_ops_->get_coarse_op());
+        auto comm = coarse->get_communicator();
+
+        auto cs_ncols = dense_x->get_size()[1];
+        auto cs_local_nrows = coarse->get_local_matrix()->get_size()[0];
+        auto cs_global_nrows = coarse->get_size()[0];
+        auto cs_local_size = dim<2>(cs_local_nrows, cs_ncols);
+        auto cs_global_size = dim<2>(cs_global_nrows, cs_ncols);
+        auto csol = dist_vec::create(exec, comm, cs_global_size, cs_local_size,
+                                     dense_x->get_stride());
+        restrict->apply(dense_b, csol);
+        auto tmp = csol->clone();
+        this->coarse_solver_->apply(csol, tmp);
+        auto one = gko::initialize<Vector>({0.5}, exec);
+        auto zero = gko::initialize<Vector>({0.5}, exec);
+        prolong->apply(one, tmp, zero, dense_x);
     }
 }
 
@@ -188,9 +215,19 @@ void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::generate(
         as<experimental::distributed::Matrix<ValueType, LocalIndexType,
                                              GlobalIndexType>>(system_matrix);
 
-    if (parameters_.coarse_solver_factory) {
-        this->coarse_solver_ = as<multigrid::MultigridLevel>(
-            share(parameters_.coarse_solver_factory->generate(dist_mat)));
+    if (parameters_.local_solver) {
+        this->set_solver(gko::share(
+            parameters_.local_solver->generate(dist_mat->get_local_matrix())));
+    } else {
+        this->set_solver(parameters_.generated_local_solver);
+    }
+
+
+    if (parameters_.galerkin_ops_factory && parameters_.coarse_solver_factory) {
+        this->galerkin_ops_ = as<multigrid::MultigridLevel>(
+            share(parameters_.galerkin_ops_factory->generate(dist_mat)));
+        this->coarse_solver_ = parameters_.coarse_solver_factory->generate(
+            this->galerkin_ops_->get_coarse_op());
     }
 }
 
--- a/core/distributed/preconditioner/schwarz.cpp
+++ b/core/distributed/preconditioner/schwarz.cpp
@@ -107,31 +107,41 @@ void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::apply_dense_impl(
                                    gko::detail::get_local(dense_x));
     }
 
-    if (this->coarse_solver_ != nullptr && this->coarse_level_ != nullptr) {
-        auto restrict = as<gko::multigrid::MultigridLevel>(this->coarse_level_)
-                            ->get_restrict_op();
-        auto prolong = as<gko::multigrid::MultigridLevel>(this->coarse_level_)
-                           ->get_prolong_op();
-        auto coarse =
-            as<experimental::distributed::Matrix<ValueType, LocalIndexType,
-                                                 GlobalIndexType>>(
+    if (parameters_.coarse_correction.mode ==
+        coarse_correction_mode::additive) {
+        if (this->coarse_solver_ != nullptr && this->coarse_level_ != nullptr) {
+            auto restrict =
                 as<gko::multigrid::MultigridLevel>(this->coarse_level_)
-                    ->get_coarse_op());
+                    ->get_restrict_op();
+            auto prolong =
+                as<gko::multigrid::MultigridLevel>(this->coarse_level_)
+                    ->get_prolong_op();
+            auto coarse =
+                as<experimental::distributed::Matrix<ValueType, LocalIndexType,
+                                                     GlobalIndexType>>(
+                    as<gko::multigrid::MultigridLevel>(this->coarse_level_)
+                        ->get_coarse_op());
 
-        // Coarse solve vector cache init
-        // Should allocare only in the first apply call.
-        auto cs_ncols = dense_x->get_size()[1];
-        auto cs_local_nrows = coarse->get_local_matrix()->get_size()[0];
-        auto cs_global_nrows = coarse->get_size()[0];
-        auto cs_local_size = dim<2>(cs_local_nrows, cs_ncols);
-        auto cs_global_size = dim<2>(cs_global_nrows, cs_ncols);
-        auto comm = coarse->get_communicator();
-        csol_cache_.init(exec, comm, cs_global_size, cs_local_size);
+            // Coarse solve vector cache init
+            // Should allocare only in the first apply call.
+            auto cs_ncols = dense_x->get_size()[1];
+            auto cs_local_nrows = coarse->get_local_matrix()->get_size()[0];
+            auto cs_global_nrows = coarse->get_size()[0];
+            auto cs_local_size = dim<2>(cs_local_nrows, cs_ncols);
+            auto cs_global_size = dim<2>(cs_global_nrows, cs_ncols);
+            auto comm = coarse->get_communicator();
+            csol_cache_.init(exec, comm, cs_global_size, cs_local_size);
 
-        restrict->apply(dense_b, csol_cache_.get());
-        this->coarse_solver_->apply(csol_cache_.get(), csol_cache_.get());
-        prolong->apply(this->coarse_weight_, csol_cache_.get(),
-                       this->local_weight_, dense_x);
+            restrict->apply(dense_b, csol_cache_.get());
+            this->coarse_solver_->apply(csol_cache_.get(), csol_cache_.get());
+            prolong->apply(this->coarse_weight_, csol_cache_.get(),
+                           this->local_weight_, dense_x);
+        }
+    } else if (parameters_.coarse_correction.mode ==
+               coarse_correction_mode::multiplicative) {
+        GKO_NOT_IMPLEMENTED;
+    } else {
+        GKO_NOT_IMPLEMENTED;
     }
 }
 
@@ -227,13 +237,22 @@ void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::generate(
             gko::share(parameters_.local_solver->generate(local_matrix)));
     }
 
+    if ((parameters_.coarse_level && !parameters_.coarse_solver) ||
+        (!parameters_.coarse_level && parameters_.coarse_solver)) {
+        GKO_INVALID_STATE(
+            "Requires both coarse solver and coarse level to be set.");
+    }
     auto dist_mat =
         as<experimental::distributed::Matrix<ValueType, LocalIndexType,
                                              GlobalIndexType>>(system_matrix);
+    // this->local_weight_ = gko::initialize<matrix::Dense<ValueType>>(
+    //     {ValueType{1} - parameters_.coarse_weight}, this->get_executor());
+    // this->coarse_weight_ = gko::initialize<matrix::Dense<ValueType>>(
+    //     {parameters_.coarse_weight}, this->get_executor());
     this->local_weight_ = gko::initialize<matrix::Dense<ValueType>>(
-        {ValueType{1} - parameters_.coarse_weight}, this->get_executor());
+        {ValueType{1}}, this->get_executor());
     this->coarse_weight_ = gko::initialize<matrix::Dense<ValueType>>(
-        {parameters_.coarse_weight}, this->get_executor());
+        {ValueType{1}}, this->get_executor());
 
     if (parameters_.local_solver) {
         this->set_solver(gko::share(
@@ -246,17 +265,17 @@ void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::generate(
     if (parameters_.coarse_level && parameters_.coarse_solver) {
         this->coarse_level_ =
             share(parameters_.coarse_level->generate(system_matrix));
-        if (as<gko::multigrid::MultigridLevel>(this->coarse_level_)
-                ->get_coarse_op()) {
-            auto coarse =
-                as<experimental::distributed::Matrix<ValueType, LocalIndexType,
-                                                     GlobalIndexType>>(
-                    as<gko::multigrid::MultigridLevel>(this->coarse_level_)
-                        ->get_coarse_op());
-            auto exec = coarse->get_executor();
-            auto comm = coarse->get_communicator();
-            this->coarse_solver_ =
-                share(parameters_.coarse_solver->generate(coarse));
+        if (as<multigrid::MultigridLevel>(this->coarse_level_)) {
+            auto coarse = as<multigrid::MultigridLevel>(this->coarse_level_)
+                              ->get_coarse_op();
+            // Cannot guarantee that coarse op is not nullptr
+            if (coarse != nullptr) {
+                this->coarse_solver_ =
+                    share(parameters_.coarse_solver->generate(
+                        as<experimental::distributed::Matrix<
+                            ValueType, LocalIndexType, GlobalIndexType>>(
+                            coarse)));
+            }
         } else {
             GKO_NOT_SUPPORTED(this->coarse_level_);
         }
--- a/core/distributed/preconditioner/schwarz.cpp
+++ b/core/distributed/preconditioner/schwarz.cpp
@@ -59,8 +59,8 @@ Schwarz<ValueType, LocalIndexType, GlobalIndexType>::parse(
     if (auto& obj = config.get("l1_smoother")) {
         params.with_l1_smoother(obj.get_boolean());
     }
-    if (auto& obj = config.get("galerkin_ops")) {
-        params.with_galerkin_ops(
+    if (auto& obj = config.get("coarse_level")) {
+        params.with_coarse_level(
             gko::config::parse_or_get_factory<const LinOpFactory>(
                 obj, context, td_for_child));
     }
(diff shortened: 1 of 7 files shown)


pratikvn:
As a first step, I just added a two-level preconditioner with equal weighting for the local solution and the coarse solution. I think it maybe makes sense to try this out in some applications and then think about arbitrary number of levels, additive/multiplicative etc.

pratikvn:
format!

include/ginkgo/core/distributed/preconditioner/schwarz.hpp
```diff
@@ -89,6 +91,19 @@ class Schwarz
          */
         std::shared_ptr<const LinOp> GKO_FACTORY_PARAMETER_SCALAR(
             generated_local_solver, nullptr);
+
+        /**
+         * Operator factory to generate the triplet (prolong_op, coarse_op,
+         * restrict_op).
+         */
+        std::shared_ptr<const LinOpFactory> GKO_DEFERRED_FACTORY_PARAMETER(
+            galerkin_ops);
```
MarcelKoch, line 100: I'm not too convinced of the name and the doc. At least to me, `galerkin_ops` is too unspecific, maybe something like `coarse_level_factory` would be better suited. Also the doc should contain that this will be used to create a coarse level system, maybe 
```
Operator factory to generate the coarse system `A_c = R * A * P`. 
```

core/distributed/preconditioner/schwarz.cpp
```diff
@@ -77,11 +87,40 @@ void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::apply_dense_impl(
...
+        auto cs_local_nrows = coarse->get_local_matrix()->get_size()[0];
+        auto cs_global_nrows = coarse->get_size()[0];
+        auto cs_local_size = dim<2>(cs_local_nrows, cs_ncols);
+        auto cs_global_size = dim<2>(cs_global_nrows, cs_ncols);
+        auto comm = coarse->get_communicator();
+        csol_cache_.init(exec, comm, cs_global_size, cs_local_size);
+
+        restrict->apply(dense_b, csol_cache_.get());
+        this->coarse_solver_->apply(csol_cache_.get(), csol_cache_.get());
+        prolong->apply(this->half_, csol_cache_.get(), this->half_, dense_x);
```
MarcelKoch, line 122: I don't think it's typical to weight these solutions. For example the book Domain Decomposition by Smith, Bjorstad, and Gropp, just sums all contributions up (p. 47) regardless of if they are from the subdomains, or the coarse grid.

core/distributed/preconditioner/schwarz.cpp
```diff
@@ -129,16 +170,35 @@ void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::generate(
...
-                ->get_local_matrix())));
-
+        this->set_solver(gko::share(
+            parameters_.local_solver->generate(dist_mat->get_local_matrix())));
     } else {
         this->set_solver(parameters_.generated_local_solver);
     }
+
+
+    if (parameters_.galerkin_ops && parameters_.coarse_solver) {
```
MarcelKoch, line 185: maybe there should be an exception, if only one of those is set. I think that would be a configuration error, and the user should be notified.

core/distributed/preconditioner/schwarz.cpp
```diff
@@ -129,16 +170,35 @@ void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::generate(
...
+        this->galerkin_ops_ =
+            share(parameters_.galerkin_ops->generate(system_matrix));
+        if (as<gko::multigrid::MultigridLevel>(this->galerkin_ops_)
+                ->get_coarse_op()) {
```
MarcelKoch, line 189: Can this actually be null? If so, maybe that should be considered an error.

core/distributed/preconditioner/schwarz.cpp
MarcelKoch, line 1: I think adjustments to the copy/move are necessary. However, they seem to be already incomplete as of now, so that might also be done in a separate PR.

core/distributed/preconditioner/schwarz.cpp
```diff
@@ -129,16 +170,35 @@ void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::generate(
...
+            auto coarse =
+                as<experimental::distributed::Matrix<ValueType, LocalIndexType,
+                                                     GlobalIndexType>>(
+                    as<gko::multigrid::MultigridLevel>(this->galerkin_ops_)
+                        ->get_coarse_op());
+            auto exec = coarse->get_executor();
+            auto comm = coarse->get_communicator();
```
MarcelKoch, line 196: unused
```suggestion
```

examples/distributed-solver/distributed-solver.cpp
```diff
@@ -79,6 +80,7 @@ int main(int argc, char* argv[])
         static_cast<gko::size_type>(argc >= 3 ? std::atoi(argv[2]) : 100);
     const auto num_iters =
         static_cast<gko::size_type>(argc >= 4 ? std::atoi(argv[3]) : 1000);
+    std::string schw_type = argc >= 5 ? argv[4] : "multi-level";
```
MarcelKoch, line 83: Just as a note, I was running this example with
```
mpirun -n 4 ./distributed-solver reference 10000 1000
```
and the `one-level` version was faster and used less iteration. So maybe it should be the default.

MarcelKoch (changes requested):
Generally looks good, I left some minor nits.
Did you have a case, where the coarse grid solve was beneficial? Right now the provided example doesn't seem to benefit from the coarse grid solve.
Also, since the coarse and local solves are not computed in parallel, as you mentioned, which will probably be the case for a while, maybe implement it as a multiplicative version directly.

core/distributed/preconditioner/schwarz.cpp
pratikvn: For multiplicative versions that is valid, but I think in this case it generally results in an over-correction and leads to solver divergence.

core/distributed/preconditioner/schwarz.cpp
pratikvn: Tracking it here: https://github.com/ginkgo-project/ginkgo/issues/1779

core/distributed/preconditioner/schwarz.cpp
pratikvn: I think we need in general a better way to validate our configuration. If multiple factory parameters, then I think there should be a way to communicate that directly to the user rather than having the logic in the core file. Maybe we can think about adding some dependency support checks in the factory parameter setup agnostic to the concrete class.

MarcelKoch: We could do that at a later point, but for now we should still throw if only one is set. I don't think it would be much of a burden here.

examples/distributed-solver/distributed-solver.cpp
pratikvn: Yes, it seems to depend on the problem size. I see that 100000 multi-level reduces the number of iterations, but it is slower. But I will change the default to one-level.

core/distributed/preconditioner/schwarz.cpp
pratikvn: Yes, sure.

core/distributed/preconditioner/schwarz.cpp
```diff
@@ -130,15 +178,53 @@ void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::generate(
...
-                ->get_local_matrix())));
+    if ((parameters_.coarse_level && !parameters_.coarse_solver) ||
+        (!parameters_.coarse_level && parameters_.coarse_solver)) {
+        GKO_INVALID_STATE(
+            "Requires both coarse solver and coarse level to be set.");
+    }
+    auto dist_mat =
+        as<experimental::distributed::Matrix<ValueType, LocalIndexType,
+                                             GlobalIndexType>>(system_matrix);
+    if (parameters_.coarse_weight > 0) {
```
MarcelKoch, line 189: I think a check for `coarse_weight > 1` wouldn't hurt.

core/distributed/preconditioner/schwarz.cpp
```diff
@@ -130,15 +178,53 @@ void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::generate(
...
     } else {
         this->set_solver(parameters_.generated_local_solver);
     }
+
+
+    if (parameters_.coarse_level && parameters_.coarse_solver) {
+        this->coarse_level_ =
+            share(parameters_.coarse_level->generate(system_matrix));
+        if (as<multigrid::MultigridLevel>(this->coarse_level_)) {
+            auto coarse = as<multigrid::MultigridLevel>(this->coarse_level_)
```
MarcelKoch, line 214: you could use `if(auto coarse_level = std::dynamic_pointer_cast<const multigrid::MultigridLevel>(this->coarse_level_))` to remove redundancy.

MarcelKoch (changes requested):

core/distributed/preconditioner/schwarz.cpp
pratikvn: IMO, the current one is easier to understand.

core/distributed/preconditioner/schwarz.cpp
```diff
@@ -47,6 +48,19 @@ Schwarz<ValueType, LocalIndexType, GlobalIndexType>::parse(
             gko::config::parse_or_get_factory<const LinOpFactory>(
                 obj, context, td_for_child));
     }
+    if (auto& obj = config.get("coarse_level")) {
```
MarcelKoch, line 51: all config parameters are missing tests.

MarcelKoch (changes requested):

core/distributed/preconditioner/schwarz.cpp
```diff
@@ -130,15 +182,52 @@ void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::generate(
...
+        (!parameters_.coarse_level && parameters_.coarse_solver)) {
+        GKO_INVALID_STATE(
+            "Requires both coarse solver and coarse level to be set.");
+    }
+    auto dist_mat =
+        as<experimental::distributed::Matrix<ValueType, LocalIndexType,
+                                             GlobalIndexType>>(system_matrix);
+    if (parameters_.coarse_weight > 0 && parameters_.coarse_weight <= 1) {
+        this->local_weight_ = gko::initialize<matrix::Dense<ValueType>>(
+            {ValueType{1} - ValueType{parameters_.coarse_weight}},
```
yhmtsai, line 195: because we only use it in ValueType, the type of coarse_weight should directly use ValueType not double

core/distributed/preconditioner/schwarz.cpp
```diff
@@ -130,15 +182,52 @@ void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::generate(
...
         this->set_solver(parameters_.generated_local_solver);
     }
+
+    if (parameters_.coarse_level && parameters_.coarse_solver) {
+        this->coarse_level_ =
+            share(parameters_.coarse_level->generate(system_matrix));
+        if (as<multigrid::MultigridLevel>(this->coarse_level_)) {
+            auto coarse = as<multigrid::MultigridLevel>(this->coarse_level_)
+                              ->get_coarse_op();
+            // Cannot guarantee that coarse op is not nullptr
```
yhmtsai, line 219: should it throw the error rather than make the operation silent?

examples/distributed-solver/distributed-solver.cpp
```diff
@@ -63,10 +64,12 @@ int main(int argc, char* argv[])
     // - The executor, defaults to reference.
     // - The number of grid points, defaults to 100.
     // - The number of iterations, defaults to 1000.
+    // - One-level or two-level preconditioner, defaults to one-level.
```
yhmtsai, line 67: it also has another option for no-preconditioned in the code

core/distributed/preconditioner/schwarz.cpp
```diff
@@ -77,10 +91,46 @@ void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::apply_dense_impl(
...
 {
     using Vector = matrix::Dense<ValueType>;
+    using dist_vec = experimental::distributed::Vector<ValueType>;
     auto exec = this->get_executor();
-    if (this->local_solver_ != nullptr) {
+
+    // Two-level
+    if (this->coarse_solver_ != nullptr && this->coarse_level_ != nullptr) {
         this->local_solver_->apply(gko::detail::get_local(dense_b),
                                    gko::detail::get_local(dense_x));
```
yhmtsai, line 100: I think you can move the local_solver part to the beginning without another condition.
```
if (local_solver_ != nullptr) {...}
if (this->coarse_solver_ ...) {...}
```

yhmtsai (changes requested):
#1806 seems to be merged first. We will need to discuss the two-level should take the matrix after or before $\ell_1$ modification.

include/ginkgo/core/distributed/preconditioner/schwarz.hpp
```diff
@@ -89,6 +97,32 @@ class Schwarz
          */
         std::shared_ptr<const LinOp> GKO_FACTORY_PARAMETER_SCALAR(
             generated_local_solver, nullptr);
+
+        /**
+         * Coarse weighting.
+         *
+         * By default the coarse and the local solutions are added together
+         * (when the coarse weight is < 0). A weighting can instead be provided
```
yhmtsai, line 105:
```suggestion
         * (when the coarse weight is < 0 or > 1). A weighting can instead be provided
```

core/distributed/preconditioner/schwarz.cpp
```diff
@@ -130,15 +182,52 @@ void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::generate(
...
+        this->local_weight_ = gko::initialize<matrix::Dense<ValueType>>(
+            {ValueType{1} - ValueType{parameters_.coarse_weight}},
+            this->get_executor());
+        this->coarse_weight_ = gko::initialize<matrix::Dense<ValueType>>(
+            {ValueType{parameters_.coarse_weight}}, this->get_executor());
+    } else {
+        this->local_weight_ = gko::initialize<matrix::Dense<ValueType>>(
+            {ValueType{1.0}}, this->get_executor());
+        this->coarse_weight_ = gko::initialize<matrix::Dense<ValueType>>(
+            {ValueType{1.0}}, this->get_executor());
```
yhmtsai, line 203: putting coarse_weight = 1 or 0 to the above formula will not get the same weight here, right?

pratikvn: Yes, coarse_weight 1 should mean that only the coarse level solve is used, and coarse_weight 0 means that only the local solves are used. If no coarse_weight is used (implying to weighting between two solutions), then the solutions are just added up.

core/distributed/preconditioner/schwarz.cpp
pratikvn: I prefer to keep them in separate code paths for better code readability.

yhmtsai:
`pgm_factory` and `coarse_solver_factory` are not initialized

core/distributed/preconditioner/schwarz.cpp
pratikvn: I guess that depends on whether we consider generation of a coarse level valid with empty triplet (R, A, P = nullptr)

pratikvn:
Regarding the L1 smoother, I think it should not affect the coarse solve. The levels (fine-local and coarse) should still be generated from the input matrix rather than an implicitly modified matrix. 

IMO, the L1 smoother should have been separate from Schwarz class and not included in the Schwarz class at all. It should be equivalent to the user calling `diag_mat->apply(one, I, one, dist_mat)`, right (After summing the non-local matrix rows) ? I am not entirely sure if it should be a part of the Schwarz preconditioner class directly. I think that makes sense for the Jacobi preconditioner, but I am not sure it makes sense for the other preconditioners ?

core/distributed/preconditioner/schwarz.cpp
```diff
@@ -87,10 +102,48 @@ template <typename VectorType>
...
+        // Additive apply of coarse correction
+        restrict_op->apply(dense_b, crhs_cache_.get());
+        // TODO: Does it make sense to restrict dense_x (to csol_cache) to
+        // provide a good initial guess for the coarse solver ?
+        this->coarse_solver_->apply(crhs_cache_.get(), csol_cache_.get());
+        prolong_op->apply(this->coarse_weight_, csol_cache_.get(),
+                          this->local_weight_, dense_x);
+    } else {
+        // One-level
+        if (this->local_solver_ != nullptr) {
```
MarcelKoch, line 143: nit: else if (...)

yhmtsai:
for L1, yes it is equivalent. However, user will not be able to handle it in adding the additional computation by themselves because it should only be used in the preconditioner. the solver will not adapt the process.

core/distributed/preconditioner/schwarz.cpp
```diff
@@ -183,6 +241,39 @@ void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::generate(
...
+        this->local_weight_ = gko::initialize<matrix::Dense<ValueType>>(
+            {ValueType{1.0}}, this->get_executor());
+        this->coarse_weight_ = gko::initialize<matrix::Dense<ValueType>>(
+            {ValueType{1.0}}, this->get_executor());
+    }
+
+    if (parameters_.coarse_level && parameters_.coarse_solver) {
+        this->coarse_level_ =
+            share(parameters_.coarse_level->generate(system_matrix));
+        if (as<multigrid::MultigridLevel>(this->coarse_level_)) {
```
MarcelKoch, line 263: This can only return false, if `coarse_level_` has dynamic type `MultigridLevel`, but is the nullptr. Can this actually happen? If it doesn't have the correct dynamic type, then `as` will already throw an exception.

MarcelKoch (approved):

core/distributed/preconditioner/schwarz.cpp
yhmtsai: I assume you mean only return true?

core/distributed/preconditioner/schwarz.cpp
yhmtsai: I think we should throw an error now. The previous issue here is caused by that the test does not initialize the coarse solver and coarse level properly. The slight operation on this will give user wrong performance thought. If someone comes to use for this behavior later, we can add another parameter for that but not the default behavior.

include/ginkgo/core/distributed/preconditioner/schwarz.hpp
```diff
@@ -99,6 +107,32 @@ class Schwarz
          * local solver.
          */
         bool GKO_FACTORY_PARAMETER_SCALAR(l1_smoother, false);
+
+        /**
+         * Coarse weighting.
+         *
+         * By default the coarse and the local solutions are added together
+         * (when the coarse weight is < 0 or > 1). A weighting can instead be
```
yhmtsai, line 115:
```suggestion
         * (when the coarse weight is <= 0 or > 1). A weighting can instead be
```

core/distributed/preconditioner/schwarz.cpp
```diff
@@ -183,6 +241,39 @@ void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::generate(
...
     }
+
+    gko::remove_complex<ValueType> cweight =
+        gko::detail::real_impl(parameters_.coarse_weight);
+    if (cweight > 0.0 && cweight <= 1.0) {
+        this->local_weight_ = gko::initialize<matrix::Dense<ValueType>>(
+            {ValueType{1} - ValueType{parameters_.coarse_weight}},
+            this->get_executor());
+        this->coarse_weight_ = gko::initialize<matrix::Dense<ValueType>>(
+            {ValueType{parameters_.coarse_weight}}, this->get_executor());
```
yhmtsai, line 252:
```suggestion
            {static_cast<ValueType>(parameters_.coarse_weight)}, this->get_executor());
```
there is narrow-down warning here

core/distributed/preconditioner/schwarz.cpp
```diff
@@ -183,6 +241,39 @@ void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::generate(
...
+    } else {
+        this->local_weight_ = gko::initialize<matrix::Dense<ValueType>>(
+            {ValueType{1.0}}, this->get_executor());
```
yhmtsai, line 255:
```suggestion
            {one<ValueType>()}, this->get_executor());
```
same for next one to avoid the narrow-down warning.

include/ginkgo/core/distributed/preconditioner/schwarz.hpp
```diff
@@ -99,6 +107,32 @@ class Schwarz
          * local solver.
          */
         bool GKO_FACTORY_PARAMETER_SCALAR(l1_smoother, false);
```
yhmtsai, line 109: add a comment about it is unsupported with two level

include/ginkgo/core/distributed/preconditioner/schwarz.hpp
yhmtsai: changing the condition in the core might be better for symmetric

test/mpi/preconditioner/schwarz.cpp
```diff
@@ -264,6 +276,22 @@ TYPED_TEST(SchwarzPreconditioner, CanApplyPreconditioner)
 }
 
 
+TYPED_TEST(SchwarzPreconditioner, CanApplyMultilevelPreconditioner)
```
yhmtsai, line 279: The test is not enough because it does not verify the result based on something we tested before.
You can do the workflow again like you did in core. also need two coarse weight testcase: one is in the [0, 1] and the other is default.

I was thinking whether extracting the same thing from core to here is considered valid or not.
It seems to be duplication but the routines here can be considered verified by the other tests.

core/distributed/preconditioner/schwarz.cpp
MarcelKoch: No, but the first comma is confusing, so I remove it.

core/distributed/preconditioner/schwarz.cpp
pratikvn: No, the issue is that when pgm (or MultigridLevel) is generated with an empty matrix, then the coarse op is null. We should probably ensure that the triplet generated by MultigridLevel cannot be null. If we throw here then it means that the Schwarz preconditioner will not be in a valid state.

include/ginkgo/core/distributed/preconditioner/schwarz.hpp
pratikvn: It is supported with two-level. L1 being a local smoother is separate from the coarse solve.

test/mpi/preconditioner/schwarz.cpp
pratikvn: Which core tests do you mean ? I dont think there are any tests in core which do a proper apply.

pratikvn: Should be updated now. I just use Multigrid as a preconditioner to emulate the coarse apply and a local precond separately and then add the solutions together.

core/distributed/preconditioner/schwarz.cpp
```diff
@@ -183,6 +241,34 @@ void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::generate(
...
+
+    if (parameters_.coarse_level && parameters_.coarse_solver) {
+        this->coarse_level_ =
+            share(parameters_.coarse_level->generate(system_matrix));
+        if (auto coarse = as<multigrid::MultigridLevel>(this->coarse_level_)
+                              ->get_coarse_op()) {
+            this->coarse_solver_ = share(parameters_.coarse_solver->generate(
+                as<Matrix<ValueType, LocalIndexType, GlobalIndexType>>(
+                    coarse)));
+        }
```
yhmtsai, line 270: throw an error or assert if coarse_level_ or coarse_solver are nullptr.
Do you need the fallback option? 
For example, the problem is small enough or different reason such that multigrid level does not generate proper one.
User seems to easily have a wrong expectation. we do not perform the two-level schwarz.

core/distributed/preconditioner/schwarz.cpp
```diff
@@ -87,8 +102,46 @@ template <typename VectorType>
...
 {
+    using Vector = matrix::Dense<ValueType>;
+    using dist_vec = experimental::distributed::Vector<ValueType>;
     auto exec = this->get_executor();
-    if (this->local_solver_ != nullptr) {
+
+    // Two-level
+    if (this->coarse_solver_ != nullptr && this->coarse_level_ != nullptr) {
+        this->local_solver_->apply(gko::detail::get_local(dense_b),
+                                   gko::detail::get_local(dense_x));
```
yhmtsai, line 112: you need to wrap if (this->local_solver) to ensure it is not nullptr

yhmtsai (changes requested):

core/distributed/preconditioner/schwarz.cpp
pratikvn: I will add an assert in the generate.

core/distributed/preconditioner/schwarz.cpp
pratikvn: I would say that is out of scope of what Schwarz should do. I think the user has to ensure that the coarse level has valid inputs. The fallback option will be that there is no coarse level solve.

PR #1432, opened 2023-10-15
Enable unpacking combined factorizations

upsj:
This implements the `Factorization::unpack` function for LU and Cholesky factorizations. There are no new kernels, so I only test the core functionality.

### DIFF
--- a/core/factorization/factorization.cpp
+++ b/core/factorization/factorization.cpp
@@ -33,18 +33,88 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <ginkgo/core/factorization/factorization.hpp>
 
 
+#include <ginkgo/core/base/exception_helpers.hpp>
 #include <ginkgo/core/base/types.hpp>
 #include <ginkgo/core/matrix/csr.hpp>
+
+
+#include "core/factorization/factorization_kernels.hpp"
 
 
 namespace gko {
 namespace experimental {
 namespace factorization {
+namespace {
+
+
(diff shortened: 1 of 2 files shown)


yhmtsai (approved):
also need to check the composition and symm composition unpack

core/factorization/factorization.cpp
```diff
@@ -33,18 +33,88 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+Factorization<ValueType, IndexType>::unpack() const
+{
+    const auto exec = this->get_executor();
+    const auto size = this->get_size();
+    switch (this->get_storage_type()) {
+    case storage_type::empty:
+        GKO_NOT_SUPPORTED(nullptr);
+    case storage_type::composition:
+    case storage_type::symm_composition:
+        return this->clone();
```
pratikvn, line 72: Why return a clone for the symmetric version here ?

pratikvn (approved):

core/factorization/factorization.cpp
upsj: the function returns a unique_ptr, we cannot fulfill that requirement from `this`

upsj: should probably add a note to the documentation on this.

pratikvn: And that is not the case for a non-symmetric composition ?

upsj: this is a fallthrough, both `composition` and `symm_composition` lead to the statement being executed.

PR #1434, opened 2023-10-16
Use proper atomics for ParILU(T) and ParIC(T)

upsj:
I was always a bit surprised by the large discrepancy between OpenMP and GPU ParILU(T) convergence. Recently I found out why that is: GPUs don't have coherent L1 caches, so the current implementation of ParILU (without any memory fences that might cause L1 cache flushes) may read stale values instead of the latest values. This is not an issue on x86 CPUs, because they have a fully coherent cache hierarchy. By replacing the regular (weak) loads/stores by relaxed loads/stores, we get more up-to-date values.

In some preliminary experiments, this reduced the number of iterations to reach a threshold by at least a factor 2.

Before merging, I'll attach some benchmarks to evaluate the runtime impact of these changes.

### DIFF
--- a/common/cuda_hip/factorization/par_ic_kernels.hpp.inc
+++ b/common/cuda_hip/factorization/par_ic_kernels.hpp.inc
@@ -78,16 +78,18 @@ __global__ __launch_bounds__(default_block_size) void ic_sweep(
         auto l_col = l_col_idxs[l_row_begin];
         auto lh_row = l_col_idxs[lh_col_begin];
         if (l_col == lh_row && l_col < last_entry) {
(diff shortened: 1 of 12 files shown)


thoasm (approved):
LGTM!

upsj:
I'm seeing a slight decrease in time-to-solution on the ani* matrices with BiCGStab and ParILUT (and an even slighter reduction with ParILU) due to the higher preconditioner quality, despite the runtime of the sweeps increasing by 1.05x (ParILUT) and 1.45x (ParILU).
On the largest problem (ani7), the time-to-solution is reduced from 11.6 s to 10.2s, mostly due to the reduced number of iterations (73 vs. 84).

So from my perspective, this PR is good to go :)

PR #1436, opened 2023-10-18
Add dpcpp csr check_diagonal_entries and add_scaled_identity

yhmtsai:
This PR adds the missing components related to csr diag in dpcpp
- `check_diagonal_entries`
- `add_scaled_identity`

### DIFF
--- a/common/cuda_hip/matrix/csr_kernels.hpp.inc
+++ b/common/cuda_hip/matrix/csr_kernels.hpp.inc
@@ -826,15 +826,19 @@ __global__ __launch_bounds__(default_block_size) void add_scaled_identity(
     auto tile_grp =
         group::tiled_partition<warp_size>(group::this_thread_block());
     const auto warpid = thread::get_subwarp_id_flat<warp_size, IndexType>();
-    const auto num_warps = thread::get_subwarp_num_flat<warp_size, IndexType>();
     if (warpid < num_rows) {
         const auto tid_in_warp = tile_grp.thread_rank();
         const IndexType row_start = row_ptrs[warpid];
         const IndexType num_nz = row_ptrs[warpid + 1] - row_start;
+        const auto beta_val = beta[0];
+        const auto alpha_val = alpha[0];
         for (IndexType iz = tid_in_warp; iz < num_nz; iz += warp_size) {
-            values[iz + row_start] *= beta[0];
-            if (col_idxs[iz + row_start] == warpid) {
-                values[iz + row_start] += alpha[0];
+            if (beta_val != one<ValueType>()) {
+                values[iz + row_start] *= beta_val;
+            }
+            if (col_idxs[iz + row_start] == warpid &&
+                alpha_val != zero<ValueType>()) {
+                values[iz + row_start] += alpha_val;
             }
         }
     }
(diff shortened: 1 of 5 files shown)


pratikvn (approved):
LGTM

dpcpp/matrix/csr_kernels.dp.cpp
```diff
@@ -871,6 +871,72 @@ void extract_diagonal(size_type diag_size, size_type nnz,
...
+    if (row < num_rows) {
+        const auto tid_in_warp = tile_grp.thread_rank();
+        const auto row_start = row_ptrs[row];
+        const auto num_nz = row_ptrs[row + 1] - row_start;
+        for (IndexType iz = tid_in_warp; iz < num_nz; iz += subgroup_size) {
+            values[iz + row_start] *= beta[0];
+            if (col_idxs[iz + row_start] == row) {
+                values[iz + row_start] += alpha[0];
+            }
+        }
```
thoasm, line 933: It might make sense to check the scaling value itself first before performing these operations (in case of beta == 1, you don't have to read and write most of the values).
```suggestion
        const auto num_nz = row_ptrs[row + 1] - row_start;
        const auto beta_v = beta[0];
        const auto alpha_v = alpha[0];
        for (IndexType iz = tid_in_warp; iz < num_nz; iz += subgroup_size) {
            if (beta_v != one<ValueType>()) {
                values[iz + row_start] *= beta_v;
            }
            if (col_idxs[iz + row_start] == row) {
                if (alpha_v != zero<ValueType>()) {
                    values[iz + row_start] += alpha_v;
                }
            }
        }
```

thoasm (approved):
LGTM!

common/cuda_hip/matrix/csr_kernels.hpp.inc
```diff
@@ -826,15 +826,19 @@ __global__ __launch_bounds__(default_block_size) void add_scaled_identity(
...
         const auto tid_in_warp = tile_grp.thread_rank();
         const IndexType row_start = row_ptrs[warpid];
         const IndexType num_nz = row_ptrs[warpid + 1] - row_start;
+        const auto beta_val = beta[0];
+        const auto alpha_val = alpha[0];
         for (IndexType iz = tid_in_warp; iz < num_nz; iz += warp_size) {
-            values[iz + row_start] *= beta[0];
-            if (col_idxs[iz + row_start] == warpid) {
-                values[iz + row_start] += alpha[0];
+            if (beta[0] != one<ValueType>()) {
```
thoasm, line 836: nit: forgot to use the `beta_val` variable for this check
```suggestion
            if (beta_val != one<ValueType>()) {
```

dpcpp/matrix/csr_kernels.dp.cpp
thoasm: I didn't check it, but the condition should be the same for all threads, so I was hoping it would be optimized properly.
In any case, if you only want to scale the diagonal, the check should significantly improve performance (maybe the `alpha` check is not that necessary, but the `beta` check does make sense IMO).

tcojean (approved):

dpcpp/matrix/csr_kernels.dp.cpp
yhmtsai: agree. I think alpha check does not hurt because we already have the check for diagonal

PR #1437, opened 2023-10-19
Fix batch ell const apply infinite loop

yhmtsai:
This PR fixes the infinite loop of `const apply` call, which is raised by sonarcloud
I also reverse the calling path between `const apply` and `apply`.
`apply` will call `const apply` now not `const apply` called `apply`


upsj:
Throwing my vote in again to make `apply` void and remove the chaining functionality in 2.0

upsj (approved):

pratikvn (approved):
Thanks for the fix! I was a bit lazy to create a PR for this and had the fix directly in my batch bicgstab branch. But I think we can merge this and I can rebase.

yhmtsai:
rebase!

PR #1441, opened 2023-10-25
Use deferred factory parameters in Combined

MarcelKoch:
@yhmtsai If you want, you can also put that into your PR.

I've not seen any other instance, where the old style parameters could be replaced by the deferred ones.


upsj (approved):

MarcelKoch:
Already in #1439

PR #1399, opened 2023-08-23
Update contributors.txt

phu0ngng:

PR #1443, opened 2023-10-26
Add CUDA, HIP and DPCPP batch bicgstab kernels

pratikvn:
This PR adds the batch bicgstab solver kernels for CUDA, HIP and DPCPP backends. Some additional single rhs vector kernels are also added into the batch multivector kernels.

### TODO

+ [x] Add DPCPP kernels


dpcpp/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -0,0 +1,449 @@
...
+                                     sycl::nd_item<3> item_ct1)
+{
+    for (int r = item_ct1.get_local_linear_id(); r < num_rows;
+         r += item_ct1.get_local_range().size()) {
+        x_shared_entry[r] = x_shared_entry[r] + alpha * p_hat_shared_entry[r];
+    }
+}
+
+
+template <typename StopType, const int n_shared_total, const bool sg_kernel_all,
```
MarcelKoch, line 212: Is `sg_kernel_all` the same as `prec_shared_bool` in the cuda/hip kernels?
Also, the const is unnecessary here.

dpcpp/solver/batch_bicgstab_kernels.dp.cpp
```diff
@@ -33,39 +33,247 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+#include "dpcpp/solver/batch_bicgstab_kernels.hpp.inc"
+
+
 template <typename T>
 using settings = gko::kernels::batch_bicgstab::settings<T>;
 
 
+__dpct_inline__ int get_group_size(int value, int simd_len = 32)
+{
+    int num_sg = (value + simd_len - 1) / simd_len;
```
MarcelKoch, line 82: ceildiv

dpcpp/solver/batch_bicgstab_kernels.dp.cpp
```diff
@@ -33,39 +33,247 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+            exec_,
+            sconf.gmem_stride_bytes * num_batch_items / sizeof(ValueType));
+        assert(sconf.gmem_stride_bytes % sizeof(ValueType) == 0);
+
+        ValueType* const workspace_data = workspace.get_data();
+        int n_shared_total = sconf.n_shared + int(sconf.prec_shared);
+
+        // template
+        // launch_apply_kernel<StopType, SIMDLEN, n_shared_total, sg_kernel_all>
+        if (num_rows <= 32 && n_shared_total == 10)
```
MarcelKoch, line 188: cuda/hip uses 9 vectors in shmem. Why does this check for 10? Also the kernel only checks until `n_shared_total == 9`

dpcpp/solver/batch_bicgstab_kernels.dp.cpp
```diff
@@ -33,39 +33,247 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+            device.get_info<sycl::info::device::max_work_group_size>();
+        if (group_size > num_rows) group_size = get_group_size(num_rows);
+
+        size_type shmem_per_blk =
+            device.get_info<sycl::info::device::local_mem_size>() -
+            (group_size + 5) * sizeof(ValueType) -
+            2 * sizeof(
+                    real_type);  // reserve 5 for intermediate rho-s, norms,
+                                 // alpha, omega, temp and for reduce_over_group
+        if (shmem_per_blk < 0) shmem_per_blk = 0;
```
MarcelKoch, line 167: fuse with above and use std::max

core/solver/batch_bicgstab_kernels.hpp
```diff
@@ -90,6 +90,103 @@ inline int local_memory_requirement(const int num_rows, const int num_rhs)
...
 
+struct storage_config {
+    // preconditioner storage
+    bool prec_shared;
+    // total number of shared vectors
+    int n_shared;
+    // number of vectors in global memory
+    int n_global;
+    // global stride from one batch entry to the next
+    int gmem_stride_bytes;
```
MarcelKoch, line 101: is it necessary to store this in bytes? In the instances I've seen it used, it was always divided again by the value type size

dpcpp/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -0,0 +1,449 @@
...
+        item_ct1.barrier(sycl::access::fence_space::local_space);
+
+        // v = A * p_hat
+        simple_apply_kernel(mat_global_entry, p_hat_sh, v_sh, item_ct1);
+        item_ct1.barrier(sycl::access::fence_space::local_space);
+
+        // alpha = rho_new / < r_hat , v>
+        compute_alpha<sg_kernel_all>(num_rows, rho_new_sh[0], r_hat_sh, v_sh,
+                                     alpha_sh[0], item_ct1);
+        // item_ct1.barrier(sycl::access::fence_space::local_space);
```
MarcelKoch, line 391: why is that sync not necessary? Same for the commented out one below.

dpcpp/solver/batch_bicgstab_kernels.hpp.inc
MarcelKoch: Also, that name is very non-descriptive.

MarcelKoch (commented):
I think the kernels look good so far. I have mostly comments outside of those.

Here are some things to be tackled later:
- use dispatch instead of manual switch
- make reductions work with more than 1 warp

common/cuda_hip/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -0,0 +1,378 @@
...
+                   static_cast<ValueType>(1.0), r_shared_entry);
+    __syncthreads();
+
+    if (threadIdx.x / config::warp_size == 0) {
+        single_rhs_compute_norm2(subgroup, num_rows, r_shared_entry, res_norm);
+    } else if (threadIdx.x / config::warp_size == 1) {
+        // Compute norms of rhs
+        single_rhs_compute_norm2(subgroup, num_rows, b_global_entry, rhs_norm);
+    }
+    __syncthreads();
```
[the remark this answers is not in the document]
pratikvn: Diverging paths between subwarps. To ensure consistency, I think it is good to synchronize them.

core/solver/batch_bicgstab_kernels.hpp
```diff
@@ -90,6 +90,103 @@ inline int local_memory_requirement(const int num_rows, const int num_rhs)
...
+                                      const int num_rows, const int num_nz,
+                                      const int num_rhs)
+{
+    using real_type = remove_complex<ValueType>;
+    const int vec_size = num_rows * num_rhs * sizeof(ValueType);
+    const int num_main_vecs = 9;
+    const int prec_storage =
+        Prectype::dynamic_work_size(num_rows, num_nz) * sizeof(ValueType);
+    int rem_shared = shared_mem_per_blk;
+    storage_config sconf{false, 0, num_main_vecs, 0, num_rows};
```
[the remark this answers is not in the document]
pratikvn: I think it is better to explicitly set the default values for these types of structs. It should be fairly straightforward to lookup which value stands for what member.

core/solver/batch_bicgstab_kernels.hpp
```diff
@@ -90,6 +90,103 @@ inline int local_memory_requirement(const int num_rows, const int num_rhs)
...
+    sconf.n_shared += num_vecs_shared;
+    sconf.n_global -= num_vecs_shared;
+    if (sconf.n_global > 0) {
+        set_gmem_stride_bytes<align_bytes>(sconf, vec_size, prec_storage);
+        return sconf;
+    }
+    rem_shared -= num_vecs_shared * vec_size;
+    if (rem_shared >= prec_storage && prec_storage > 0) {
+        sconf.prec_shared = true;
+        rem_shared -= prec_storage;
```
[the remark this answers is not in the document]
pratikvn: For correctness, I think this is necessry. For example, if someone wants to print out the `rem_shared` at the end of this function.

core/solver/batch_bicgstab_kernels.hpp
```diff
@@ -90,6 +90,103 @@ inline int local_memory_requirement(const int num_rows, const int num_rhs)
...
+ *   should be prioritized, the cache configuration must be updated separately
+ *   and the needed space should be subtracted before passing to this
+ *   function.
+ * @param num_rows  Size of the matrix.
+ * @param num_nz  Number of nonzeros in the matrix
+ * @param num_rhs  Number of right-hand-sides in the vectors.
+ * @return  A struct containing allocation information specific to Bicgstab.
+ */
+template <typename Prectype, typename ValueType, int align_bytes = 32>
+storage_config compute_shared_storage(const int shared_mem_per_blk,
```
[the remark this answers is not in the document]
pratikvn: I think splitting this into more functions makes it more harder to understand what each function is doing, especially as each function reads and writes into sconf. 

I will add inline comments to make it clear what each line is doing.

cuda/solver/batch_bicgstab_kernels.cu
```diff
@@ -56,19 +75,227 @@ namespace cuda {
...
+            exec_,
+            sconf.gmem_stride_bytes * num_batch_items / sizeof(value_type));
+        assert(sconf.gmem_stride_bytes % sizeof(value_type) == 0);
+
+        value_type* const workspace_data = workspace.get_data();
+
+        // Template parameters launch_apply_kernel<StopType, n_shared,
+        // prec_shared)
+        if (sconf.prec_shared)
+            launch_apply_kernel<StopType, 9, 1>(
```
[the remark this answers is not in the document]
pratikvn: `prec_shared` is a `bool`

cuda/solver/batch_bicgstab_kernels.cu
```diff
@@ -56,19 +75,227 @@ namespace cuda {
...
+    template <typename BatchMatrixType, typename PrecType, typename StopType,
+              typename LogType>
+    void call_kernel(
+        LogType logger, const BatchMatrixType& mat, PrecType prec,
+        const gko::batch::multi_vector::uniform_batch<const value_type>& b,
+        const gko::batch::multi_vector::uniform_batch<value_type>& x) const
+    {
+        using real_type = gko::remove_complex<value_type>;
+        const size_type num_batch_items = mat.num_batch_items;
+        constexpr int align_multiple = 8;
```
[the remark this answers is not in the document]
pratikvn: configure_shared_memroy_banks does it for the device, while this is for when we allocate and set the offsets for the different vectors

core/solver/batch_bicgstab_kernels.hpp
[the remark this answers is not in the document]
MarcelKoch: The default values are set with `sconf{}`. If you remove the fast returns (which I think is possible), then you don't read any member before it is written. IMO, until we have c++20 designated initializers, it is not straight forward to look up the members.

cuda/solver/batch_bicgstab_kernels.cu
```diff
@@ -56,19 +75,227 @@ namespace cuda {
...
+            get_max_dynamic_shared_memory<StopType, PrecType, LogType,
+                                          BatchMatrixType, value_type>(exec_,
+                                                                       0);
+        const int block_size =
+            get_num_threads_per_block<StopType, PrecType, LogType,
+                                      BatchMatrixType, value_type>(
+                exec_, mat.num_rows);
+        assert(block_size >= 2 * config::warp_size);
+
+        const size_t prec_size =
```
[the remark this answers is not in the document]
pratikvn: storage_config is meant to be a struct to be passed into the kernel. Ideally, we would like to minimize its size, while trying to keep the necessary info. Its not expensive to compute this, so I think it should be okay to do it when necessary.

cuda/solver/batch_bicgstab_kernels.cu
```diff
@@ -33,21 +33,40 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "core/solver/batch_bicgstab_kernels.hpp"
 
 
+#include <thrust/functional.h>
+#include <thrust/transform.h>
```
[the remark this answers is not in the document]
pratikvn: Yes, mainly due to included files from batch_multi_vector_kernels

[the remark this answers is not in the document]
MarcelKoch: I think it can be hard to figure out which kernel file needs which includes. Perhaps it makes sense if there is the kernel implementation in `some_kernels.cpp.inc` then to have a file `some_kernels_includes.hpp.inc` (or just use `.hpp.inc`). But that might be done later.

dpcpp/solver/batch_bicgstab_kernels.hpp.inc
pratikvn: No, it is actually different. With `sg_kernel_all` we create local memory objects that are [shared between all work-items of a group](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_oneapi_local_memory.asciidoc#allocating-local-memory). I think we observed better performance for smaller group sizes with this strategy.

dpcpp/solver/batch_bicgstab_kernels.dp.cpp
pratikvn: the strategy is slightly different. Here the count includes the prec_shared vector.  The number of shared vectors is always 9, so you can only check until 9. If it is greater than 9, then you know that the prec is also in shared memory.

dpcpp/solver/batch_bicgstab_kernels.hpp.inc
pratikvn: re-enabled now. I think it is a bit difficult to clearly say where a sync is necessary. But for correctness, it is best to have they local_space syncs everywhere.

dpcpp/solver/batch_bicgstab_kernels.hpp.inc
MarcelKoch: Then, I don't think the current name is appropriate. At least, I could not guess that this is the function of that parameter.

dpcpp/solver/batch_bicgstab_kernels.dp.cpp
MarcelKoch: but isn't that what `storage_config::prec_shared` is there for?

core/solver/batch_bicgstab_kernels.hpp
pratikvn: But I think as we want to align to a certain byte size, it is cleaner to store it in bytes and divide by value_type size as necessary.

pratikvn:
format!

dpcpp/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -0,0 +1,455 @@
...
+
+    if constexpr (sg_kernel_all) {
+        using tile_value_t = ValueType[5];
+        tile_value_t& values =
+            *sycl::ext::oneapi::group_local_memory_for_overwrite<tile_value_t>(
+                group);
+        using tile_real_t = real_type[2];
+        tile_real_t& reals =
+            *sycl::ext::oneapi::group_local_memory_for_overwrite<tile_real_t>(
+                group);
```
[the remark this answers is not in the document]
yhmtsai: they can also be passed like the dynamic memory, which is supported without extension I think.
dpct and the other ginkgo kernels currently use this way to allocate static shared memory

core/test/utils/batch_helpers.hpp
```diff
@@ -250,7 +250,7 @@ LinearSystem<MatrixType> generate_batch_linear_system(
     // A * x^{*} = b
     sys.matrix->apply(sys.exact_sol, sys.rhs);
     const gko::batch_dim<2> norm_dim(num_batch_items, gko::dim<2>(1, num_rhs));
-    sys.rhs_norm = real_vec::create(exec, norm_dim);
+    sys.rhs_norm = real_vec::create(exec->get_master(), norm_dim);
```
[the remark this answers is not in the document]
pratikvn: That is by design, because I only use it for relative residual check and for that I need to explicitly copy to host otherwise.

cuda/base/kernel_config.cuh
```diff
@@ -0,0 +1,65 @@
...
+ * \tparam ValueType  The scalar type used for computations.
+ */
+template <typename ValueType>
+inline void configure_shared_memory_banks()
+{
+    if (sizeof(ValueType) == 4) {
+        cudaDeviceSetSharedMemConfig(cudaSharedMemBankSizeFourByte);
+    } else if (sizeof(ValueType) % 8 == 0) {
+        cudaDeviceSetSharedMemConfig(cudaSharedMemBankSizeEightByte);
+    }
```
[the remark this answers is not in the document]
pratikvn: No, I dont think that is necessary. Only a value of 8 is recommended for `double` to avoid bank conflicts. You can just set it to 4 for `half` I think .

dpcpp/matrix/batch_ell_kernels.hpp.inc
```diff
@@ -45,22 +43,19 @@ __dpct_inline__ void simple_apply_kernel(
             if (col_idx == invalid_index<IndexType>()) {
                 break;
             } else {
-                temp += mat.values[tidx + idx * mat.stride] *
-                        b.values[col_idx * b.stride];
+                temp += mat.values[tidx + idx * mat.stride] * b[col_idx];
             }
         }
-        x.values[tidx * x.stride] = temp;
+        x[tidx] = temp;
```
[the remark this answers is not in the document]
pratikvn: I just use the plain pointers as arguments here. I guess technically we should have another stride parameter to the function, but I think that is unnecessary for now and we can add that when we support stride later.

core/solver/batch_bicgstab_kernels.hpp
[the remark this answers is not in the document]
MarcelKoch: That has nothing to do with correctness, only with a particular debugging style. As mentioned somewhere else, this modification would make more sense to me if it was stored as part of `sconf`.

core/solver/batch_bicgstab_kernels.hpp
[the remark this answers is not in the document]
MarcelKoch: That should help. But I would still suggest to remove the early returns. This function is not time-critical, so we can prioritize readability.

cuda/solver/batch_bicgstab_kernels.cu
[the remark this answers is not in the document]
MarcelKoch: The cuda/hip/dpcpp kernels only use `gmem_strided_bytes` and `padded_vec_len`, so you could reduce it even further. But I would suggest to rather use two structs, one for the core code, and one for the kernels, as the batch stuff is doing in a lot of places.

core/solver/batch_bicgstab_kernels.hpp
```diff
@@ -90,6 +90,112 @@ inline int local_memory_requirement(const int num_rows, const int num_rhs)
...
+{
+    using real_type = remove_complex<ValueType>;
+    const int vec_size = num_rows * num_rhs * sizeof(ValueType);
+    const int num_main_vecs = 9;
+    const int prec_storage =
+        Prectype::dynamic_work_size(num_rows, num_nz) * sizeof(ValueType);
+    int rem_shared = shared_mem_per_blk;
+    // Set default values. All vecs are in global.
+    storage_config sconf{false, 0, num_main_vecs, 0, num_rows};
+    // If available shared mem, is zero, set all vecs to global.
```
MarcelKoch, line 169: That comment doesn't describe what the code below does.

dpcpp/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -0,0 +1,455 @@
...
+__dpct_inline__ void initialize(
+    const int num_rows, const BatchMatrixType_entry& mat_global_entry,
+    const ValueType* const b_global_entry,
+    const ValueType* const x_global_entry, ValueType& rho_old, ValueType& omega,
+    ValueType& alpha, ValueType* const x_shared_entry,
+    ValueType* const r_shared_entry, ValueType* const r_hat_shared_entry,
+    ValueType* const p_shared_entry, ValueType* const v_shared_entry,
+    typename gko::remove_complex<ValueType>& rhs_norm,
+    typename gko::remove_complex<ValueType>& res_norm,
+    sycl::nd_item<3> item_ct1)
```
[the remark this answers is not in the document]
pratikvn: I think the compiler does try to use it, but like in CUDA, they seem to recommend using `__ldg()` instead.

cuda/solver/batch_bicgstab_kernels.cu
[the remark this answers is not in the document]
MarcelKoch: So, that alignment is only relevant if the vectors are stored in global memory, right?

dpcpp/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -0,0 +1,455 @@
...
+    } else {
+        ValueType values[5];
+        real_type reals[2];
+        rho_old_sh = &values[0];
+        rho_new_sh = &values[1];
+        alpha_sh = &values[2];
+        omega_sh = &values[3];
+        temp_sh = &values[4];
+        norms_rhs_sh = &reals[0];
+        norms_res_sh = &reals[1];
```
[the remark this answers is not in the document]
pratikvn: From the documentation, it seems to be deallocated only when the kernel has finished or rather when all work-items in the group have finished executing the kernel

> Constructs an object of type T in an address space accessible by all work items in the workgroup g, using the default initialization. The object is initialized upon or before the first call to the group_local_memory_for_overwrite function. The storage for the object is allocated upon or before the first call to the group_local_memory_for_overwrite function, and deallocated when all work items in the workgroup have completed executing the kernel. 

[See reference](https://www.intel.com/content/www/us/en/docs/oneapi-fpga-add-on/optimization-guide/2023-2/fpga-local-memory-function.html)

cuda/solver/batch_bicgstab_kernels.cu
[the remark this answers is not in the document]
pratikvn: No, also for vector lengths stored in shared memory.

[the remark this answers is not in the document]
MarcelKoch: Should the alignment then be the same as the bank size?

core/solver/batch_bicgstab_kernels.hpp
pratikvn: If the remaining shared is zero, then in the function, we have `sconf.n_global` set to all vectors, and we allocate everything on global memory. Maybe 
```
// If available shared memory is zero, then set all vectors to be in global memory, including the preconditioner workspace 
```
is more clear ?

MarcelKoch: This is again an issue with the early return. None of the code in the if-branch sets `sconf.n_global`. It just uses the default setting from above. Again, it would be much easier to describe what is happening in this function without the early return.

cuda/solver/batch_bicgstab_kernels.cu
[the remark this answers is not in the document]
pratikvn: alignment here is to ensure that the vectors are not across block boundaries. But I guess there is a relation between configuring the shared memory banks and padding the rows. I have not experimented with that yet. 8 should always work I think. But maybe when using `float` 4 is also enough.

cuda/solver/batch_bicgstab_kernels.cu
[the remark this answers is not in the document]
pratikvn: I would prefer to keep it like this for now as all the other batched solvers (not implemented yet in develop) follow a similar structure. I have [noted these suggestions down](https://github.com/ginkgo-project/ginkgo/issues/1376) and update them for all solvers later.

dpcpp/solver/batch_bicgstab_kernels.dp.cpp
pratikvn: I think it is a bit easier with looking at `n_shared` as 10 vectors. Otherwise, `prec_shared` will need to be a template parameter as well. But I understand your point that it makes the cuda/dpcpp kernels more confusing to compare.

test/solver/batch_bicgstab_kernels.cpp
```diff
@@ -224,13 +248,15 @@ TEST_F(BatchBicgstab, CanSolveLargeHpdSystem)
                                                  &logger->get_num_iterations());
     auto res_norm = gko::make_temporary_clone(exec->get_master(),
                                               &logger->get_residual_norm());
-    GKO_ASSERT_BATCH_MTX_NEAR(res.x, linear_system.exact_sol, comp_tol);
+    GKO_ASSERT_BATCH_MTX_NEAR(res.x, linear_system.exact_sol, tol * 50);
     for (size_t i = 0; i < num_batch_items; i++) {
-        auto comp_res_norm =
-            exec->copy_val_to_host(res.res_norm->get_const_values() + i);
+        auto comp_res_norm = res.res_norm->get_const_values()[i] /
+                             linear_system.rhs_norm->get_const_values()[i];
```
yhmtsai, line 254: I will prefer the original version because all vectors in res use the same executor

cuda/solver/batch_bicgstab_kernels.cu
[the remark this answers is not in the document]
MarcelKoch: Ok. I guess we can leave it as it is. But perhaps `vector_alignment` might be a better name?

core/solver/batch_bicgstab_kernels.hpp
```diff
@@ -90,6 +90,112 @@ inline int local_memory_requirement(const int num_rows, const int num_rhs)
...
+                           const int multi_vector_size_bytes,
+                           const int prec_storage_bytes)
+{
+    int gmem_stride = sconf.n_global * multi_vector_size_bytes;
+    if (!sconf.prec_shared) {
+        gmem_stride += prec_storage_bytes;
+    }
+    // align global memory chunks
+    sconf.gmem_stride_bytes =
+        gmem_stride > 0 ? ((gmem_stride - 1) / align_bytes + 1) * align_bytes
```
yhmtsai, line 118:
```suggestion
        gmem_stride > 0 ? ceildiv(gmem_stride, aligh_bytes) * align_bytes
```

dpcpp/solver/batch_bicgstab_kernels.hpp.inc
MarcelKoch: maybe `use_sg_mem_scalars`?

dpcpp/solver/batch_bicgstab_kernels.dp.cpp
MarcelKoch: I would prefer the additional template parameter then. But that might also be done later.

core/solver/batch_bicgstab_kernels.hpp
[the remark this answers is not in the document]
pratikvn: I think ensuring the variables have correct values is useful for self-documenting purposes, even if they are not used later in the scope. I dont think it makes sense to have rem_shared in `sconf`. Maybe when we split it into two parts, one for core and one for kernels as you suggested, it might. But for now, we can leave it as it is.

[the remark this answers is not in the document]
MarcelKoch: My issue with dead code is that if someone without deep knowledge of this functions reads this, they might think that something is missing. `rem_shared` is modified, but not used, so why is that there? The reader might then come to the conclusion that there is an error in the code.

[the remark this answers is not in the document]
pratikvn: In combination with the documentation inline, I think there should be no ambiguity. I think within the scope the variable should always have the correct value, so I would not necessarily say that it is dead code.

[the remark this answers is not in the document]
MarcelKoch: The above is just my interpretation of the code, and my reasoning for removing it. But this is nothing I would hold up the PR for.

core/solver/batch_bicgstab_kernels.hpp
MarcelKoch: The function could be interpreted as a sequence of operations like:
```
set_number_of_vectors(...);
set_preconditioner(...);
set_global_memory_stride(...);
```
But the early return makes that hard to follow (internally these functions could use early return).
Maybe this can help to illustrate my difficulty understanding the control flow. If you put in the corresponding else-branch for each if-branch, then the final return is *three* levels deep. IMO, something like this is hard to follow.

common/cuda_hip/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -0,0 +1,378 @@
...
+
+template <typename ValueType>
+__device__ __forceinline__ void update_p(
+    const int num_rows, const ValueType& rho_new, const ValueType& rho_old,
+    const ValueType& alpha, const ValueType& omega,
+    const ValueType* const r_shared_entry,
+    const ValueType* const v_shared_entry, ValueType* const p_shared_entry)
+{
+    for (int r = threadIdx.x; r < num_rows; r += blockDim.x) {
+        const ValueType beta = (rho_new / rho_old) * (alpha / omega);
```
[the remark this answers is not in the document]
pratikvn: I think technically it is not necessary as it is not read from until it has been first updated. But I will add it in any case.

test/solver/batch_bicgstab_kernels.cpp
pratikvn: Given that this is a struct is mainly for tests, I think it becomes easier to read and generally less error-prone if we dont have to write copy_val_to_host everywhere.

dpcpp/base/batch_multi_vector_kernels.dp.cpp
```diff
@@ -296,9 +366,14 @@ void copy(std::shared_ptr<const DefaultExecutor> exec,
...
     const auto num_batches = x_ub.num_batch_items;
+    const int num_rows = x->get_common_size()[0];
     auto device = exec->get_queue()->get_device();
-    auto group_size =
+    constexpr int max_subgroup_size = config::warp_size;
+    long max_group_size =
         device.get_info<sycl::info::device::max_work_group_size>();
+    int group_size =
+        std::max(ceildiv(num_rows, max_subgroup_size) * max_subgroup_size,
+                 max_group_size);
```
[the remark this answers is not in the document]
pratikvn: Good catch. Thanks

core/solver/batch_bicgstab_kernels.hpp
pratikvn: IMO this type of early return makes the code easier to read, but I understand your concern. I guess it should be possible to refactor this into separate functions like you suggested such that the `set_num_vec...`, `set_prec...` take `rem_shared` and `sconf` and set the values and return `sconf`, but maybe we postpone this refactoring for now. 

But I think eventually I would like to unify this across all batched solver classes, because essentially the computation is similar. I have noted it down in #1376 and we can come back to this later.

common/cuda_hip/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -0,0 +1,378 @@
...
+        single_rhs_compute_norm2(subgroup, num_rows, r_shared_entry, res_norm);
+    } else if (threadIdx.x / config::warp_size == 1) {
+        // Compute norms of rhs
+        single_rhs_compute_norm2(subgroup, num_rows, b_global_entry, rhs_norm);
+    }
+    __syncthreads();
+
+    for (int iz = threadIdx.x; iz < num_rows; iz += blockDim.x) {
+        r_hat_shared_entry[iz] = r_shared_entry[iz];
+        p_shared_entry[iz] = zero<ValueType>();
```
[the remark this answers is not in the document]
yhmtsai: Yes, I also realize that afterward, so I add the comment in another pr and no change needed here

cuda/solver/batch_bicgstab_kernels.cu
```diff
@@ -56,19 +75,217 @@ namespace cuda {
...
+#include "common/cuda_hip/solver/batch_bicgstab_kernels.hpp.inc"
+
+
+template <typename StopType, typename PrecType, typename LogType,
+          typename BatchMatrixType, typename ValueType>
+int get_num_threads_per_block(std::shared_ptr<const DefaultExecutor> exec,
+                              const int num_rows)
+{
+    int num_warps = std::max(num_rows / 4, 2);
+    constexpr int warp_sz = static_cast<int>(config::warp_size);
```
yhmtsai, line 91: from this, it seems to use a warp to handle a row.
but the following check is using a thread to handle a row.

cuda/solver/batch_bicgstab_kernels.cu
```diff
@@ -56,19 +75,217 @@ namespace cuda {
...
+                                       BatchMatrixType, ValueType>);
+    const int num_regs_used = funcattr.numRegs;
+    int max_regs_blk = 0;
+    cudaDeviceGetAttribute(&max_regs_blk, cudaDevAttrMaxRegistersPerBlock,
+                           exec->get_device_id());
+    const int max_threads_regs =
+        ((max_regs_blk / static_cast<int>(num_regs_used)) / warp_sz) * warp_sz;
+    int max_threads = std::min(max_threads_regs, device_max_threads);
+    max_threads = max_threads <= 1024 ? max_threads : 1024;
+    return std::min(num_warps * warp_sz, max_threads);
```
yhmtsai, line 107: I assume `num_regs_used` always larger than 0 and less than max_regs_blk. If it is not, the max_threads_regs can be zero or infinite.
from the first block of this function, I guess you would like to ensure 2 * warp_size, so you may need `return std::max(std::min(num_warps * warp_sz, max_threads), 2 * warp_size);`.

dpcpp/base/batch_multi_vector_kernels.hpp.inc
```diff
@@ -67,6 +67,56 @@ __dpct_inline__ void add_scaled_kernel(
...
 
+template <int tile_size = config::warp_size, typename ValueType>
+__dpct_inline__ void single_rhs_compute_conj_dot(
+    const int num_rows, const ValueType* const __restrict__ x,
+    const ValueType* const __restrict__ y, ValueType& result,
+    sycl::nd_item<3> item_ct1)
+{
+    auto grp =
+        group::tiled_partition<tile_size>(group::this_thread_block(item_ct1));
+    // auto grp = group::this_thread_block(item_ct1);
```
yhmtsai, line 78: If it does not introduce issue, you can delete the unused code.

dpcpp/solver/batch_bicgstab_kernels.dp.cpp
```diff
@@ -53,19 +67,217 @@ namespace dpcpp {
...
+        // launch_apply_kernel<StopType, SIMDLEN, n_shared_total, sg_kernel_all>
+        if (num_rows <= 32 && n_shared_total == 10)
+            launch_apply_kernel<StopType, 16, 10, true>(
+                sconf, logger, prec, mat, b.values, x.values, workspace_data,
+                group_size, shared_size);
+        else if (num_rows <= 256 && n_shared_total == 10)
+            launch_apply_kernel<StopType, 32, 10, true>(
+                sconf, logger, prec, mat, b.values, x.values, workspace_data,
+                group_size, shared_size);
+        else {
```
yhmtsai, line 198:
```suggestion
        else if (num_rows <= 256 && n_shared_total == 10) {
            launch_apply_kernel<StopType, 32, 10, true>(
                sconf, logger, prec, mat, b.values, x.values, workspace_data,
                group_size, shared_size);
        } else {
```
also for the others

dpcpp/solver/batch_bicgstab_kernels.dp.cpp
```diff
@@ -53,19 +67,217 @@ namespace dpcpp {
...
+            group_size = get_group_size(num_rows);
+        };
+
+        // reserve 5 for intermediate rho-s, norms,
+        // alpha, omega, temp and for reduce_over_group
+        // If the value available is negative, then set it to 0
+        size_type shmem_per_blk = std::max(
+            device.get_info<sycl::info::device::local_mem_size>() -
+                (group_size + 5) * sizeof(ValueType) - 2 * sizeof(real_type),
+            static_cast<size_type>(0));
```
yhmtsai, line 169: because local_mem_size returns uint64, it may reach another side if the usage is larger than return.

dpcpp/solver/batch_bicgstab_kernels.dp.cpp
```diff
@@ -53,19 +67,217 @@ namespace dpcpp {
...
+        auto workspace = gko::array<ValueType>(
+            exec_,
+            sconf.gmem_stride_bytes * num_batch_items / sizeof(ValueType));
+        GKO_ASSERT(sconf.gmem_stride_bytes % sizeof(ValueType) == 0);
+
+        ValueType* const workspace_data = workspace.get_data();
+        int n_shared_total = sconf.n_shared + int(sconf.prec_shared);
+
+        // template
+        // launch_apply_kernel<StopType, SIMDLEN, n_shared_total, sg_kernel_all>
```
yhmtsai, line 189:
```suggestion
        // launch_apply_kernel<StopType, subgroup_size, n_shared_total, sg_kernel_all>
```

hip/solver/batch_bicgstab_kernels.hip.cpp
```diff
@@ -57,19 +74,189 @@ namespace hip {
...
+{
+    int nwarps = num_rows / 4;
+    if (nwarps < 2) {
+        nwarps = 2;
+    }
+    const int min_block_size = 2 * config::warp_size;
+    const int device_max_threads =
+        ((std::max(num_rows, min_block_size)) / config::warp_size) *
+        config::warp_size;
+    const int num_regs_used_per_thread = 64;
```
yhmtsai, line 96: why is it 64 here?

hip/solver/batch_bicgstab_kernels.hip.cpp
```diff
@@ -57,19 +74,189 @@ namespace hip {
...
+    void call_kernel(
+        LogType logger, const BatchMatrixType& mat, PrecType prec,
+        const gko::batch::multi_vector::uniform_batch<const value_type>& b,
+        const gko::batch::multi_vector::uniform_batch<value_type>& x) const
+    {
+        using real_type = gko::remove_complex<value_type>;
+        const size_type num_batch_items = mat.num_batch_items;
+        constexpr int align_multiple = 8;
+        const int padded_num_rows =
+            ((mat.num_rows + align_multiple - 1) / align_multiple) *
```
yhmtsai, line 150: ceildiv

hip/solver/batch_bicgstab_kernels.hip.cpp
```diff
@@ -57,19 +74,189 @@ namespace hip {
...
+
+        value_type* const workspace_data = workspace.get_data();
+
+        // Template parameters launch_apply_kernel<StopType, n_shared,
+        // prec_shared)
+        if (sconf.prec_shared)
+            launch_apply_kernel<StopType, 9, 1>(
+                sconf, logger, prec, mat, b.values, x.values, workspace_data,
+                block_size, shared_size);
+        else {
```
yhmtsai, line 182:
```suggestion
        if (sconf.prec_shared) {
            launch_apply_kernel<StopType, 9, 1>(
                sconf, logger, prec, mat, b.values, x.values, workspace_data,
                block_size, shared_size);
        } else {
```

hip/solver/batch_bicgstab_kernels.hip.cpp
```diff
@@ -57,19 +74,189 @@ namespace hip {
...
+            case 8:
+                launch_apply_kernel<StopType, 8, 0>(
+                    sconf, logger, prec, mat, b.values, x.values,
+                    workspace_data, block_size, shared_size);
+                break;
+            case 9:
+                launch_apply_kernel<StopType, 9, 0>(
+                    sconf, logger, prec, mat, b.values, x.values,
+                    workspace_data, block_size, shared_size);
+                break;
```
yhmtsai, line 233:
```suggestion
            case 9:
                launch_apply_kernel<StopType, 9, 0>(
                    sconf, logger, prec, mat, b.values, x.values,
                    workspace_data, block_size, shared_size);
                break;
            default:
                GKO_NOT_IMPLEMENTED;
```

hip/solver/batch_bicgstab_kernels.hip.cpp
```diff
@@ -57,19 +74,189 @@ namespace hip {
...
+            }
+        }
+
+        GKO_HIP_LAST_IF_ERROR_THROW;
```
yhmtsai, line 237: does it need that? It will not capture the error without synchronization, but we already capture error with synchronization.

hip/matrix/batch_struct.hip.hpp
```diff
@@ -92,7 +92,8 @@ get_batch_struct(batch::matrix::Dense<ValueType>* const op)
  * Generates an immutable uniform batch struct from a batch of ell matrices.
  */
 template <typename ValueType, typename IndexType>
-inline batch::matrix::ell::uniform_batch<const hip_type<ValueType>, IndexType>
+inline batch::matrix::ell::uniform_batch<const hip_type<ValueType>,
+                                         const IndexType>
```
yhmtsai, line 96: I think the to_const usually face this issue.
Could you check the other const version also correct?
If all related to this issue are not in public interface, it are not urgent before release

hip/solver/batch_bicgstab_kernels.hip.cpp
yhmtsai: also miss the same thing in cuda side

test/solver/batch_bicgstab_kernels.cpp
yhmtsai: then, maybe make the variable name clear like `host_res_norm`

cuda/solver/batch_bicgstab_kernels.cu
pratikvn: I think the strategy (warp to row or thread to row) can depend on the matrix format. Hence, this just ensures that a minimum number of warps is always used.

cuda/solver/batch_bicgstab_kernels.cu
pratikvn: Yes, I would assume that `0 < num_regs_used < max_regs_blk`. But I think what you suggest is good to have in case of some issues with those functions.

hip/solver/batch_bicgstab_kernels.hip.cpp
pratikvn: That is the [value for maximum occupancy](https://gpuopen.com/learn/amd-lab-notes/amd-lab-notes-register-pressure-readme/). HIP does not have a function to query the number of registers for a given function like CUDA.

hip/solver/batch_bicgstab_kernels.hip.cpp
pratikvn: I think this gives a more specific error, with a string `cudaGetErrorString` being called, but I guess we could technically have that in `GKO_CUDA_ERROR()` as well.

yhmtsai: what do you mean specific error? I thought it will give the same message when you call `exec->synchronize()` somewhere else

yhmtsai: cudaGetErrorString is to get the error from previous error from synchronize or asynchronize (if it is finished).
synchronize action will pop up with the asynchronization error.
I think we always have the error check for synchronization call/kernel, so I think we do not need that.
It may catch the invalid configuration like <<<1, 1025>>>

upsj: the _correct_ way of doing this would be using `cudaLaunchKernel` directly and checking the return value. This catches all launch configuration errors.

common/cuda_hip/solver/batch_bicgstab_kernels.hpp.inc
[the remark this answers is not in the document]
upsj: Not requesting any changes, but I wanted to elaborate on this a bit. I agree here, I think we could take a page from CUB's book, where they ensure synchronization always happens inside functions that require it (i.e. SpMVs and reductions) and are entirely absent from the code otherwise.
To make this work, you need a "default" work assignment (like the default `for (int iz = threadIdx.x; iz < num_rows; iz += blockDim.x)` loop) and every time you read from values outside your own assigned set, you have a threadsync before, and if you write to values outside your set (also computing reductions), you have a threadsync after. This may even allow you to keep all values in registers most of the time, as long as you don't have huge blocks. But that is an optional detail.

Outside of this, there is also some potential for "kernel fusion" (i.e. removing the `__syncthreads` and computing directly on values in registers) by computing the dot product on the result of the SpMV, but I don't have a clear idea how large the runtime impact of that would be.

cuda/base/kernel_config.cuh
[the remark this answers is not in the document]
upsj: This is kind of problematic - it configures the entire device, but we only run on a single stream. At the very least, we need to revert it after the kernel finished, otherwise we interfere with other applications' performance

hip/base/exception.hip.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+
+namespace gko {
+
+
+#define GKO_HIP_LAST_IF_ERROR_THROW                                          \
+    {                                                                        \
+        hipError_t err = hipGetLastError();                                  \
+        if (err != hipSuccess) {                                             \
+            printf(" Hip kernel error: %s\n", hipGetErrorString(err));       \
```
upsj, line 47: Why `printf`? stderr is more appropriate for it, and `std::cerr` is more C++ style. Also this needs to respect `GKO_VERBOSE_LEVEL`, which is based on an xSDK policy.

common/cuda_hip/stop/batch_criteria.hpp.inc
```diff
@@ -0,0 +1,80 @@
...
+
+    __device__ __forceinline__ SimpleRelResidual(
+        const real_type rel_res_tol, const real_type* const rhs_b_norms)
+        : rel_tol_{rel_res_tol}, rhs_norms_{rhs_b_norms}
+    {}
+
+    __device__ __forceinline__ bool check_converged(
+        const real_type* const residual_norms) const
+    {
+        return (residual_norms[0] / rhs_norms_[0] <= rel_tol_);
```
yhmtsai, line 50:
```suggestion
    {
        return (residual_norms[0] <= rel_tol_ * rhs_norm_[0]);
```

core/base/batch_utilities.hpp
```diff
@@ -201,6 +201,8 @@ std::unique_ptr<OutputType> read(
                            std::forward<TArgs>(create_args)...);
 
     for (size_type b = 0; b < num_batch_items; ++b) {
+        if (data.at(b).size != data.at(0).size)
```
yhmtsai, line 204: Miss {}

core/device_hooks/common_kernels.inc.cpp
```diff
@@ -401,6 +402,15 @@ GKO_STUB_VALUE_AND_INDEX_TYPE(GKO_DECLARE_DIAGONAL_FILL_IN_MATRIX_DATA_KERNEL);
 }  // namespace diagonal
 
 
+namespace batch_bicgstab {
+
+
+GKO_STUB_VALUE_TYPE(GKO_DECLARE_BATCH_BICGSTAB_APPLY_KERNEL);
```
yhmtsai, line 408: Are these not already in another pr?

pratikvn: I think this is just a rebase issue.

hip/solver/batch_bicgstab_kernels.hip.cpp
pratikvn: I just use `synchronize()` now.

pratikvn:
format!

cuda/solver/batch_bicgstab_kernels.cu
```diff
@@ -56,19 +75,219 @@ namespace cuda {
...
+    return funcattr.maxDynamicSharedSizeBytes;
+}
+
+
 template <typename T>
 using settings = gko::kernels::batch_bicgstab::settings<T>;
 
 
+template <typename CuValueType>
+class kernel_caller {
```
MarcelKoch, line 135: this (and hip) uses snake_case, while dpcpp/omp/reference use CamelCase. Perhaps just fix one.

include/ginkgo/core/base/executor.hpp
```diff
@@ -1607,6 +1612,14 @@ class CudaExecutor : public detail::ExecutorBase<CudaExecutor>,
         return this->get_exec_info().max_subgroup_size;
     }
 
+    /**
+     * Get maximum shared memory per block.
+     */
+    int get_max_shared_memory_per_block() const noexcept
```
MarcelKoch, line 1618: does this function need to be part of the executor? It could just as well be implemented just in the cuda/hip backend

dpcpp/base/batch_multi_vector_kernels.dp.cpp
```diff
@@ -183,29 +194,59 @@ void compute_dot(std::shared_ptr<const DefaultExecutor> exec,
...
-                auto group = item_ct1.get_group();
-                auto group_id = group.get_group_linear_id();
-                const auto x_b = batch::extract_batch_item(x_ub, group_id);
-                const auto y_b = batch::extract_batch_item(y_ub, group_id);
-                const auto res_b = batch::extract_batch_item(res_ub, group_id);
-                compute_gen_dot_product_kernel(x_b, y_b, res_b, item_ct1,
-                                               [](auto val) { return val; });
-            });
-    });
+    if (x->get_common_size()[1] == 1) {
```
MarcelKoch, line 210: why is this switch made for dpcpp, but not for cuda/hip?

core/solver/batch_bicgstab_kernels.hpp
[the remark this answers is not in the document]
MarcelKoch: I guess that is something that we might tackle later.

cuda/solver/batch_bicgstab_kernels.cu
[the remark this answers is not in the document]
MarcelKoch: maybe add that to the to-do list

MarcelKoch (changes requested):
Most of my comments have been addressed. Only the change to the executor should be discussed. All other comments are relatively minior, and can be changed afterwards.

cuda/solver/batch_bicgstab_kernels.cu
pratikvn: I think this was again a rebase issue. Everything should be using snake_case

include/ginkgo/core/base/executor.hpp
pratikvn: I think this could be useful in other kernels as well.

MarcelKoch: It's not useful for core right now. If we come to the situation where it is needed for core, we can revisit this. But since we are quite rushing all of these PRs, I think it's good to remove any interface change that is not absolutely necessary.

pratikvn: now moved to internal

MarcelKoch: thanks!

MarcelKoch (approved):
All my comments have been addressed, except some minor nits which can also be addressed later.

dpcpp/solver/batch_bicgstab_kernels.dp.cpp
```diff
@@ -53,19 +67,220 @@ namespace dpcpp {
...
+#include "dpcpp/matrix/batch_dense_kernels.hpp.inc"
+#include "dpcpp/matrix/batch_ell_kernels.hpp.inc"
+#include "dpcpp/solver/batch_bicgstab_kernels.hpp.inc"
+
+
 template <typename T>
 using settings = gko::kernels::batch_bicgstab::settings<T>;
 
 
+__dpct_inline__ int get_group_size(int value, int subgroup_size = 32)
```
yhmtsai, line 80:
```suggestion
__dpct_inline__ int get_group_size(int value, int subgroup_size =config::warp_size)
```

dpcpp/solver/batch_bicgstab_kernels.dp.cpp
```diff
@@ -53,19 +67,220 @@ namespace dpcpp {
...
+            device.get_info<sycl::info::device::max_work_group_size>();
+        if (group_size > num_rows) {
+            group_size = get_group_size(num_rows);
+        };
+
+        // reserve 5 for intermediate rho-s, norms,
+        // alpha, omega, temp and for reduce_over_group
+        // If the value available is negative, then set it to 0
+        const int static_var_mem =
+            (group_size + 5) * sizeof(ValueType) - 2 * sizeof(real_type);
```
yhmtsai, line 166: what is - 2 sizeof(real_type)?
or it is indeed ( group_size + 3 ) * sizeof(ValueType) + 2 sizeof(real_type)?
if it this formula, the original one will be wrong when ValueType is real because it is eliminated by the latter.

yhmtsai (commented):
Could you also take the #1399 with the correct ordering in this pr?

include/ginkgo/core/base/executor.hpp
```diff
@@ -1318,9 +1318,9 @@ class EnableDeviceReset {
      *
      * @return the current status of the device reset boolean for this executor.
      */
-    [[deprecated(
-        "device_reset is no longer supported, call "
-        "cudaDeviceReset/hipDeviceReset manually")]] bool
+    [
```
yhmtsai, line 1321: format issue. I think you do not need to touch this file, so it should be no-changes

hip/solver/batch_bicgstab_kernels.hip.cpp
yhmtsai: Could you add a comment for that? it will be helpful for the others who wants to changes the parameters in the future.

hip/solver/batch_bicgstab_kernels.hip.cpp
```diff
@@ -57,19 +73,192 @@ namespace hip {
...
+    template <typename BatchMatrixType, typename PrecType, typename StopType,
+              typename LogType>
+    void call_kernel(
+        LogType logger, const BatchMatrixType& mat, PrecType prec,
+        const gko::batch::multi_vector::uniform_batch<const value_type>& b,
+        const gko::batch::multi_vector::uniform_batch<value_type>& x) const
+    {
+        using real_type = gko::remove_complex<value_type>;
+        const size_type num_batch_items = mat.num_batch_items;
+        constexpr int align_multiple = 8;
```
yhmtsai, line 146: does not it need to use the size of the ValueType?

yhmtsai (changes requested):

dpcpp/solver/batch_bicgstab_kernels.dp.cpp
pratikvn: Yes, you are right. It was a copy paste issue. It is 5 + 2

pratikvn:
format!

dpcpp/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -0,0 +1,453 @@
...
+    item_ct1.barrier(sycl::access::fence_space::local_space);
+
+    if constexpr (sg_kernel_all) {
+        if (sg_id == 0) {
+            single_rhs_compute_norm2_sg(num_rows, r_shared_entry, res_norm,
+                                        item_ct1);
+        } else if (sg_id == 1) {
+            single_rhs_compute_norm2_sg(num_rows, b_global_entry, rhs_norm,
+                                        item_ct1);
+        }
```
yhmtsai, line 78: not pass the subgroup_size information, so it will give the wrong answer when subgroup_size is 16 (previously).
it will be the norm of first half of r and second half of b

yhmtsai (commented):
the comment is not submitted successfully previously

pratikvn:
@yhmtsai , can you please have another look ? I think the test matrices were badly conditioned and BiCGSTAB was unable to solve them and hence we observed these failures, showing up mainly for float and with parallel reductions. The [Intel2020 bug reported](https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/5458636202) by @upsj was also fixed.

pratikvn:
format!

cuda/base/kernel_config.hpp
```diff
@@ -0,0 +1,98 @@
...
+        } else {
+            GKO_ASSERT_NO_CUDA_ERRORS(
+                cudaDeviceSetSharedMemConfig(cudaSharedMemBankSizeDefault));
+        }
+    }
+
+
+    ~shared_memory_config_guard()
+    {
+        auto error_code = cudaDeviceSetSharedMemConfig(original_config_);
```
yhmtsai, line 73: I think`cudaDeviceSetSharedMemConfig` is enough?

dpcpp/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -0,0 +1,455 @@
...
+        x_shared_entry[iz] = x_global_entry[iz];
+        r_shared_entry[iz] = b_global_entry[iz];
+    }
+    item_ct1.barrier(sycl::access::fence_space::local_space);
+
+    // r = b - A*x
+    advanced_apply_kernel(static_cast<ValueType>(-1.0), mat_global_entry,
+                          x_shared_entry, static_cast<ValueType>(1.0),
+                          r_shared_entry, item_ct1);
+    item_ct1.barrier(sycl::access::fence_space::local_space);
```
yhmtsai, line 69: you need to ensure the x, r are indeed in the local shared memory, or it may get the old result from global memory.

yhmtsai (changes requested):
The main core (not device) apply does not need the synchronize like the other operations.
dpcpp require careful handle the barrier setting because some of them may not be in the shared memory.

include/ginkgo/core/base/executor.hpp
```diff
@@ -1336,8 +1336,8 @@ class EnableDeviceReset {
 
     [[deprecated(
         "device_reset is no longer supported, call "
-        "cudaDeviceReset/hipDeviceReset manually")]] EnableDeviceReset(bool
-                                                                           device_reset)
+        "cudaDeviceReset/hipDeviceReset "
+        "manually")]] EnableDeviceReset(bool device_reset)
```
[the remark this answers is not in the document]
pratikvn: This was formatted by the github actions. So, I believe this should be correct.

hip/solver/batch_bicgstab_kernels.hip.cpp
```diff
@@ -57,19 +73,195 @@ namespace hip {
...
+                launch_apply_kernel<StopType, 9, false>(
+                    sconf, logger, prec, mat, b.values, x.values,
+                    workspace_data, block_size, shared_size);
+                break;
+            default:
+                GKO_NOT_IMPLEMENTED;
+            }
+        }
+
+        exec_->synchronize();
```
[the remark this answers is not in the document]
pratikvn: I think calling synchronize here does not hurt though. Technically, this is only a stream synchronize now.

cuda/base/kernel_config.hpp
pratikvn: Not sure what you mean here.

include/ginkgo/core/base/executor.hpp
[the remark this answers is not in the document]
yhmtsai: No, I think it should not be touched in the beginning so the GitHub action should not check it.
clang-format can not concat two string, so if it is splited first then they can not recover it.
To ensure it is correct, you need to combine these two string and then run clang-format.

cuda/base/kernel_config.hpp
yhmtsai: ah, copy wrong.
I mean `GKO_ASSERT_NO_CUDA_ERRORS` like above

hip/solver/batch_bicgstab_kernels.hip.cpp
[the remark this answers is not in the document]
yhmtsai: it is not necessary synchronize. The host still need to wait for it which may postpone other kernel submission. We do not want to block the operation when it is not required

dpcpp/solver/batch_bicgstab_kernels.hpp.inc
pratikvn: I dont think that is necessary. If it is in global memory, all the subgroup reads and writes should be consistent. A variable does not switch between local shared and global. If it is global memory, it always remains in global and if it is in local shared it always remains in local shared until the end when it is copied back. Only when it is in local shared memory, consistency between different work-items needs to be enforced with a `fence_space::local_space`

upsj: you are probably looking for a `group_barrier(item_ct1)`? As far as I understand the documentation on the [Codeplay website](https://developer.codeplay.com/products/computecpp/ce/2.11.0/api-reference/namespacecl_1_1sycl_1_1access), this applies only to local memory.

cuda/base/kernel_config.hpp
pratikvn: No, I think the reason is that you dont want to throw in a destructor, so you output an error and `std::exit`

upsj: Maybe pull out the macro `GKO_EXIT_ON_CUDA_ERROR` from `memory.cpp` into a header and use it here

dpcpp/solver/batch_bicgstab_kernels.hpp.inc
pratikvn: Is there even a `group_barrier` function ? I think `local_space` is the local shared memory, so enforcing a fence on the local memory should already ensure that all the work-items in the workgroup are synchronized ?

hip/solver/batch_bicgstab_kernels.hip.cpp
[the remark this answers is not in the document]
pratikvn: Sure. I can remove it.

include/ginkgo/core/base/executor.hpp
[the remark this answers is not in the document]
pratikvn: Unfortunately, that is a bit hard now, as the changes to the file were introduced sometime back and it is hard to roll back only for this file without completely destroying the history, particularly because my `clang-format` seems to somehow format differently. I guess there are two options:

1. We have this version for now and fix it later when some other changes are introduced to this file.
2. You or someone else who has the same `clang-format` version can explicitly remove the new line and call `clang-format` on this file to restore it back to the original formatting.

cuda/base/kernel_config.hpp
pratikvn: I think that is fine with me. I will just remove the exit, as it is probably not necessary here, though if for some reason the reset to the `original_config_` fails, there may be some inexplicable performance degradations.

dpcpp/solver/batch_bicgstab_kernels.hpp.inc
upsj: According to the SYCL 2020 standard there is. There is a difference beetween visibility of memory operations between different threads and their execution synchronization. The SYCL standard doesn't say anything about global memory coherence when synchronizing local memory, so we can't rely on it. The group_barrier function has a stronger contract. It may work on specific hardware, but may still give unexpected results on other hardware.

upsj: In practical terms, subgroups (and likely also workgroups) will use the same L1 cache, so they will likely have the same view of global memory, but the compiler may still be allowed to reorder operations that it knows happen in global memory and not local memory.

pratikvn: Indeed, it seems that with SYCL 2020, `group_barrier()` is the equivalent of `__syncthreads()`, or the equivalent `fence_space` seems to be `fence_space::global_and_local`. But it [does seem that both `group_barrier()` and `fence_space::global_and_local` are significantly slower than `fence_space::local_space`](https://github.com/intel/llvm/issues/7069). IIRC, Phuong also observed this in her experiments.

pratikvn: But I think for correctness, I will change this to `fence_space::global_and_local`, and we can investigate the performance at a later point.

yhmtsai: synchronize with global and local should definitely be slower than only with local without doubt.
If we do not adapt it with the shared/global switch, we can only use global_and_local for correctness unfortunately.
using local will work if all threads always access the same global address in whole kernel.
I think it is usually true except for norm/dot.

cuda/solver/batch_bicgstab_kernels.cu
```diff
@@ -56,19 +74,219 @@ namespace cuda {
...
+    cudaFuncAttributes funcattr;
+    cudaFuncGetAttributes(&funcattr,
+                          apply_kernel<StopType, 9, true, PrecType, LogType,
+                                       BatchMatrixType, ValueType>);
+    const int num_regs_used = funcattr.numRegs;
+    int max_regs_blk = 0;
+    cudaDeviceGetAttribute(&max_regs_blk, cudaDevAttrMaxRegistersPerBlock,
+                           exec->get_device_id());
+    const int max_threads_regs =
+        ((max_regs_blk / static_cast<int>(num_regs_used)) / warp_sz) * warp_sz;
```
yhmtsai, line 103: I do not get this function.
max_regs_blk is the #register of a block and num_regs_used is #register for function.
I would think something related to SM occupancy assign more blocks when it is required less register

dpcpp/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -0,0 +1,453 @@
...
+    if constexpr (sg_kernel_all) {
+        if (sg_id == 0) {
+            single_rhs_compute_conj_dot_sg(num_rows, r_hat_shared_entry,
+                                           v_shared_entry, alpha, item_ct1);
+        }
+        item_ct1.barrier(sycl::access::fence_space::global_and_local);
+        if (tid == 0) {
+            alpha = rho_new / alpha;
+        }
+        item_ct1.barrier(sycl::access::fence_space::global_and_local);
```
yhmtsai, line 132: need to move to end because the other path also update alpha

dpcpp/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -0,0 +1,453 @@
...
+            if (sg_id == 0)
+                single_rhs_compute_norm2_sg(num_rows, r_sh, norms_res_sh[0],
+                                            item_ct1);
+            if (tid == group_size - 1) {
+                rho_old_sh[0] = rho_new_sh[0];
+            }
+            item_ct1.barrier(sycl::access::fence_space::global_and_local);
+        } else {
+            single_rhs_compute_norm2(num_rows, r_sh, norms_res_sh[0], item_ct1);
+            rho_old_sh[0] = rho_new_sh[0];
```
yhmtsai, line 444: tid == group_size - 1
and sync

dpcpp/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -0,0 +1,453 @@
...
+                                           s_shared_entry, omega, item_ct1);
+        } else if (sg_id == 1) {
+            single_rhs_compute_conj_dot_sg(num_rows, t_shared_entry,
+                                           t_shared_entry, temp, item_ct1);
+        }
+        item_ct1.barrier(sycl::access::fence_space::global_and_local);
+        if (tid == 0) {
+            omega /= temp;
+        }
+        item_ct1.barrier(sycl::access::fence_space::global_and_local);
```
yhmtsai, line 180: move to the end

yhmtsai: it's fine now because you still call sync after the function.
If you always call sync after function, maybe removing it here also makes sense

include/ginkgo/core/solver/batch_solver_base.hpp
```diff
@@ -182,25 +182,13 @@ class BatchSolver {
...
-};
+struct preconditioned_iterative_solver_factory_parameters {};
 
 
 template <typename Parameters, typename Factory>
 struct enable_preconditioned_iterative_solver_factory_parameters
-    : enable_parameters_type<Parameters, Factory>,
-      preconditioned_iterative_solver_factory_parameters {
+    : enable_parameters_type<Parameters, Factory> {
+    using parameters_type = Parameters;
```
yhmtsai, line 191: this is unnecessary now.

include/ginkgo/core/solver/batch_solver_base.hpp
```diff
@@ -182,25 +182,13 @@ class BatchSolver {
...
-     */
-    std::shared_ptr<const BatchLinOpFactory> preconditioner{nullptr};
-
-    /**
-     * Already generated preconditioner. If one is provided, the factory
-     * `preconditioner` will be ignored.
-     */
-    std::shared_ptr<const BatchLinOp> generated_preconditioner{nullptr};
-};
+struct preconditioned_iterative_solver_factory_parameters {};
```
yhmtsai, line 185: can be deleted now

pratikvn:
format!

yhmtsai (approved):

pratikvn:
Turns out the `no-circular-deps` job is terribly slow. I verified (with the same config and flags as the job, inside the same image with a docker container) that it builds successfully with `GINKGO_CHECK_CIRCULAR_DEPS=on`, so I will go ahead and merge this.

PR #1444, opened 2023-10-29
Uniform handling of disabled deprecation warnings

upsj:
We keep copy-pasting the same (or slightly different) pragmas for disabling deprecation warnings throughout many files, instead I propose that we keep around a single pair of macros for dealing with them.

### DIFF
--- a/include/ginkgo/core/base/types.hpp
+++ b/include/ginkgo/core/base/types.hpp
@@ -89,11 +89,26 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
 // Handle deprecated notices correctly on different systems
-#if defined(_WIN32) || defined(__CYGWIN__)
-#define GKO_DEPRECATED(msg) __declspec(deprecated(msg))
+// clang-format off
+#define GKO_DEPRECATED(_msg) [[deprecated(_msg)]]
+#ifdef __NVCOMPILER
+#define GKO_BEGIN_DISABLE_DEPRECATION_WARNINGS _Pragma("diag_suppress 1445")
+#define GKO_END_DISABLE_DEPRECATION_WARNINGS _Pragma("diag_warning 1445")
+#elif defined(__GNUC__) || defined(__clang__)
+#define GKO_BEGIN_DISABLE_DEPRECATION_WARNINGS                      \
+    _Pragma("GCC diagnostic push")                                  \
+    _Pragma("GCC diagnostic ignored \"-Wdeprecated-declarations\"")
+#define GKO_END_DISABLE_DEPRECATION_WARNINGS _Pragma("GCC diagnostic pop")
+#elif defined(_MSC_VER)
+#define GKO_BEGIN_DISABLE_DEPRECATION_WARNINGS        \
+    _Pragma("warning(push)")                          \
+    _Pragma("warning(disable : 5211 4973 4974 4996)")
+#define GKO_END_DISABLE_DEPRECATION_WARNINGS _Pragma("warning(pop)")
 #else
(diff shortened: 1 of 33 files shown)


include/ginkgo/core/base/types.hpp
```diff
@@ -89,11 +89,26 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
 // Handle deprecated notices correctly on different systems
-#if defined(_WIN32) || defined(__CYGWIN__)
-#define GKO_DEPRECATED(msg) __declspec(deprecated(msg))
+// clang-format off
+#define GKO_DEPRECATED(_msg) [[deprecated(_msg)]]
```
MarcelKoch, line 93: Since deprecations are a language feature, I would prefer if we use them directly, instead of this macro. (I didn't even know that we have this macro before).

MarcelKoch (approved):
For me replacing `[[deprecated(...)]]` is not necessary. But if others want it, than it's fine with me.

include/ginkgo/core/base/types.hpp
yhmtsai: same commet if [[deprecated()]] is also allowed in WIN32 and CYGWIN

upsj:
By using a macro, we make it easier to make the warnings configurable (e.g. a config.hpp switch that disables deprecation warnings?), but I'm not particularly attached to that change. Only we probably want to keep the macro around for interface stability.

include/ginkgo/core/base/types.hpp
upsj: By using a macro, we make it easier to make the warnings configurable (e.g. a config.hpp switch that disables deprecation warnings?), but I'm not particularly attached to that change. Only we probably want to keep the macro around for interface stability.

yhmtsai (approved):
I prefer using the GKO_DEPRECATED macro when someone request it or we have corresponding option to control. but both are fine to me

thoasm (approved):
The changes look good to me! I'm currently investigating if we can get rid of more unnecessary deprecation warnings.

thoasm:
Unfortunately, I haven't found a good way to remove the `deprecated` Rcm warnings. I tried the following things:
1. Renaming Rcm to `RcmDeprecated` and deprecate an alias named `Rcm`. This removes all current warnings, but also any future warning unless you create an alias to `Rcm` itself.
2. Move the deprecation to the constructor. This reduced the warning count a bit, but a lot of them are still there (in `EnableDefaultFactory` because it calls the constructor). Shielding `EnableDefaultFactory` from warnings removes all possible warnings.
3. Marking `EnablePolymorphicObject` to ignore deprecation warnings might have removed some of the warnings, but a huge amount still shows up.

I'm currently in favor of simply proceeding with 1. or removing the deprecation warning for `Rcm` completely. It is significantly less annoying to compile Ginkgo that way and makes it less likely to miss an important warning. We are free to improve this deprecation warning in the future, but IMO, the current state is way too annoying.

upsj:
I'd be fine with both 1. and removing the deprecation warning entirely.

thoasm:
I decided to remove the deprecation warnings entirely from Rcm (and the disabling of the warnings in all test files for Rcm).

MarcelKoch:
format!

include/ginkgo/core/base/types.hpp
thoasm: For now, I decided to leave the macro in place without an option to disable it. That means we just use the C++14 method and still have the option to disable it in the future if we want to (but, as far as I can see, we don't emit any false-positives anymore),
We do, however, still emit CUDA and HIP deprecation warnings for some compilers, but no functions that are marked as deprecated inside Ginkgo.

PR #1445, opened 2023-10-30
Symbolic LU for near-symmetric matrices

upsj:
This uses a symbolic Cholesky factorization of $A + A^T$ and the numerical factorization to compute a symbolic LU factorization

TODO:
- [x] CUDA and HIP kernels
- [x] benchmark integration
- [x] factory integration


MarcelKoch:
@upsj should this be part of the release?

upsj:
@MarcelKoch would be nice to have, but not a necessity

MarcelKoch (approved):
Do you know how much gain this gives over the general factorization? If I understand it correctly, you are computing the cholesky factorization, then you filter it, and then you do the numeric factorization. Algorithm-wise, the filtering seems to be quite similar to the numeric factorization, so I guess it would be similarily expensive.

Interface-wise it looks good. I didn't find any issues on the algorithmic side, but I'm also not too familiar with that.

Sidenote: I think I mentioned that before, but we should find a way to remove the verbosity from creating the lookup structure.

upsj:
@MarcelKoch on ani7, we get a speedup of 3.5x (AMD-reordered) to 18x (input ordering) on `ani7_crop` on an A100 vs. an Intel Xeon Platinum 8368 (Horeka). So this is definitely worth it.

thoasm (approved):
Do you really only use the `float` values of the CSR to store `1` and `0`?
Isn't 32bit a bit overkill for it (pretty sure I had the same question for the Symbolic LU before)?
Additionally, some of the AMD GPU pipelines fail, so they need to be taken care of before we merge it: https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/5447633534

MarcelKoch:
@upsj that's good to hear! Thanks for checking that out.

upsj:
@thoasm 32 bit makes it easier to use `Csr` to pass around data, but at the same time, I believe 32 bit atomics are easier on the memory controller? After the release, I might give it a try to use bytes or even bits, but the latter might make it hard to use more of the available parallelism, since we are merging the dependency structure of a lot of unrelated rows. I'll have to experiment with that.

PR #1446, opened 2023-10-31
Fix OpenMPI check in cross-compiling mode

MarcelKoch:
This PR replaces try-run with try-compile in OpenMPI check.

try-run can't be run in cross-compiling mode, which happens for example on Macs. So, the sufficient OpenMPI version check has been re-written to run at compile time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -248,7 +248,6 @@ if(GINKGO_BUILD_MPI)
         ${Ginkgo_SOURCE_DIR}/cmake/openmpi_test.cpp
                 COMPILE_DEFINITIONS -DCHECK_HAS_OPEN_MPI=1
         LINK_LIBRARIES MPI::MPI_CXX
-                OUTPUT_VARIABLE output
         )
     if(uses_openmpi)
         try_compile(valid_openmpi_version
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -242,14 +242,22 @@ if(GINKGO_BUILD_MPI)
         set(GINKGO_HAVE_GPU_AWARE_MPI OFF)
     endif()
 
-    try_run(uses_openmpi gko_result_unused
+    # use try_compile instead of try_run to prevent cross-compiling issues
+    try_compile(uses_openmpi
         ${Ginkgo_BINARY_DIR}
         ${Ginkgo_SOURCE_DIR}/cmake/openmpi_test.cpp
+                COMPILE_DEFINITIONS -DCHECK_HAS_OPEN_MPI=1
         LINK_LIBRARIES MPI::MPI_CXX
-        RUN_OUTPUT_VARIABLE openmpi_version
+                OUTPUT_VARIABLE output
         )
     if(uses_openmpi)
-        if(openmpi_version VERSION_LESS "4.1")
+        try_compile(valid_openmpi_version
+                    ${Ginkgo_BINARY_DIR}
+                    ${Ginkgo_SOURCE_DIR}/cmake/openmpi_test.cpp
+                    COMPILE_DEFINITIONS -DCHECK_OPEN_MPI_VERSION=1
+                    LINK_LIBRARIES MPI::MPI_CXX
+        )
+        if(NOT valid_openmpi_version)
             message(WARNING
                 "OpenMPI v4.0.x has a bug that forces us to use blocking communication in our distributed "
                 "matrix class. To enable faster, non-blocking communication, consider updating your OpenMPI version or "
--- a/cmake/openmpi_test.cpp
+++ b/cmake/openmpi_test.cpp
@@ -39,12 +39,12 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 int main()
 {
 #if CHECK_HAS_OPEN_MPI && defined(OPEN_MPI) && OPEN_MPI
-    static_assert(true);
-#elif CHECK_OPEN_MPI_VERSION
+    static_assert(true, "Check availability of OpenMPI");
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -242,14 +242,22 @@ if(GINKGO_BUILD_MPI)
...
 
-    try_run(uses_openmpi gko_result_unused
+    # use try_compile instead of try_run to prevent cross-compiling issues
+    try_compile(uses_openmpi
         ${Ginkgo_BINARY_DIR}
         ${Ginkgo_SOURCE_DIR}/cmake/openmpi_test.cpp
+                COMPILE_DEFINITIONS -DCHECK_HAS_OPEN_MPI=1
         LINK_LIBRARIES MPI::MPI_CXX
-        RUN_OUTPUT_VARIABLE openmpi_version
+                OUTPUT_VARIABLE output
```
yhmtsai, line 251: unused

greole (approved):
LGTM!

cmake/openmpi_test.cpp
```diff
@@ -38,11 +38,13 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+#if CHECK_HAS_OPEN_MPI && defined(OPEN_MPI) && OPEN_MPI
+    static_assert(true);
+#elif CHECK_OPEN_MPI_VERSION
+    static_assert(OMPI_MAJOR_VERSION > 4 ||
+                  (OMPI_MAJOR_VERSION == 4 && OMPI_MINOR_VERSION >= 1));
 #else
-    return 0;
+    static_assert(false);
 #endif
+    return 1;
```
yhmtsai, line 49:
```suggestion
    return 0;
```
Although it does not matter now, `return 0` is success from code

yhmtsai (approved):

CMakeLists.txt
```diff
@@ -242,14 +242,22 @@ if(GINKGO_BUILD_MPI)
...
         )
     if(uses_openmpi)
-        if(openmpi_version VERSION_LESS "4.1")
+        try_compile(valid_openmpi_version
+                    ${Ginkgo_BINARY_DIR}
+                    ${Ginkgo_SOURCE_DIR}/cmake/openmpi_test.cpp
+                    COMPILE_DEFINITIONS -DCHECK_OPEN_MPI_VERSION=1
+                    LINK_LIBRARIES MPI::MPI_CXX
+        )
+        if(NOT valid_openmpi_version)
```
yhmtsai, line 260: By the way, how's the bug in OpenMPI v3.*?

CMakeLists.txt
```diff
@@ -242,14 +242,22 @@ if(GINKGO_BUILD_MPI)
...
             message(WARNING
                 "OpenMPI v4.0.x has a bug that forces us to use blocking communication in our distributed "
```
greole, line 262: Somewhat unrelated, but can we refer to a specific issue here?

cmake/openmpi_test.cpp
```diff
@@ -38,11 +38,13 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 int main()
 {
-#if defined(OPEN_MPI) && OPEN_MPI
-    std::printf("%d.%d.%d", OMPI_MAJOR_VERSION, OMPI_MINOR_VERSION,
-                OMPI_RELEASE_VERSION);
-    return 1;
+#if CHECK_HAS_OPEN_MPI && defined(OPEN_MPI) && OPEN_MPI
+    static_assert(true);
+#elif CHECK_OPEN_MPI_VERSION
```
pratikvn, line 43: for correctness

```suggestion
#elif CHECK_OPEN_MPI_VERSION && defined(OPEN_MPI) && OPEN_MPI
```

pratikvn (approved):
LGTM!

CMakeLists.txt
MarcelKoch: Not checked and I don't care. We only support openMPI >= 4. All older releases are officially retired.

CMakeLists.txt
MarcelKoch: I think I didn't fully investigate that. So I have no idea what the actual bug was.

cmake/openmpi_test.cpp
upsj: you can also remove it completely, `main` returns 0 implicitly

CMakeLists.txt
MarcelKoch: I think the more interesting question is about openMPI 5.0.0, which was released a few days ago. But I guess I will look into that after the release.

yhmtsai: I see.

yhmtsai: does OpenMPI support MPI 3.1 after OpenMPI v4?
Because our documentation claim we support MPI3.1+

thoasm (changes requested):
The idea is good, but you need to change all `static_assert` statements to have a message (without a message works for C++17 onward).

cmake/openmpi_test.cpp
```diff
@@ -38,11 +38,13 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+    static_assert(OMPI_MAJOR_VERSION > 4 ||
+                  (OMPI_MAJOR_VERSION == 4 && OMPI_MINOR_VERSION >= 1));
 #else
-    return 0;
+    static_assert(false);
```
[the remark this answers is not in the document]
MarcelKoch: I guess I just can't leave my hands of c++17 :sweat_smile:

thoasm (approved):
LGTM!

PR #1447, opened 2023-10-31
pass slurm job failure (not a fix)

yhmtsai:
slurm always reports error even if all tests are passed. It disallows the following tests in pipeline.

Because we are busy with release deadline and the following tests are sometimes useful, we just always pass it now.
but before merge, we should check the result in the pipeline.
It should be fixed or replaced by alternative way in the future


upsj (approved):

thoasm (approved):
LGTM!

PR #1450, opened 2023-11-05
Add a CSR batched matrix format, CUDA, HIP and DPCPP kernels

pratikvn:
This PR adds a batched Csr matrix format, that stores the same sparsity pattern for all entries, but different values for each batch. Only simple and advanced apply to `batch::MultiVector` are supported for now.

### TODO

+ [x] Add CUDA/HIP kernels
+ [x] Add DPCPP kernels
+ [x] Add to batch::Bicgstab dispatch
+ [x] Update docs

### DIFF
--- a/common/cuda_hip/matrix/batch_csr_kernels.hpp.inc
+++ b/common/cuda_hip/matrix/batch_csr_kernels.hpp.inc
@@ -1,35 +1,7 @@
-/*******************************<GINKGO LICENSE>******************************
-Copyright (c) 2017-2023, the Ginkgo authors
-All rights reserved.
-
-Redistribution and use in source and binary forms, with or without
-modification, are permitted provided that the following conditions
-are met:
-
-1. Redistributions of source code must retain the above copyright
-notice, this list of conditions and the following disclaimer.
-
-2. Redistributions in binary form must reproduce the above copyright
-notice, this list of conditions and the following disclaimer in the
-documentation and/or other materials provided with the distribution.
-
-3. Neither the name of the copyright holder nor the names of its
-contributors may be used to endorse or promote products derived from
-this software without specific prior written permission.
-
-THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS
-IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED
-TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A
-PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
-HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
-SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
-LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
-DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
-THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
-(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
-OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-******************************<GINKGO LICENSE>*******************************/
-
+// Copyright (c) 2017-2023, the Ginkgo authors
+// SPDX-FileCopyrightText: 2017-2023 The Ginkgo authors
+//
+// SPDX-License-Identifier: BSD-3-Clause
 
 template <typename ValueType, typename IndexType>
 __device__ __forceinline__ void simple_apply(
--- a/common/cuda_hip/matrix/batch_csr_kernels.hpp.inc
+++ b/common/cuda_hip/matrix/batch_csr_kernels.hpp.inc
@@ -39,14 +39,13 @@ __device__ __forceinline__ void simple_apply(
     const auto num_rows = mat.num_rows;
     const auto val = mat.values;
     const auto col = mat.col_idxs;
-    for (int tidx = threadIdx.x; tidx < num_rows; tidx += blockDim.x) {
+    for (int row = threadIdx.x; row < num_rows; row += blockDim.x) {
         auto temp = zero<ValueType>();
-        for (auto nnz = mat.row_ptrs[tidx]; nnz < mat.row_ptrs[tidx + 1];
-             nnz++) {
+        for (auto nnz = mat.row_ptrs[row]; nnz < mat.row_ptrs[row + 1]; nnz++) {
             const auto col_idx = col[nnz];
             temp += val[nnz] * b[col_idx];
         }
-        x[tidx] = temp;
+        x[row] = temp;
     }
 }
 
@@ -90,14 +89,13 @@ __device__ __forceinline__ void advanced_apply(
     const auto num_rows = mat.num_rows;
     const auto val = mat.values;
     const auto col = mat.col_idxs;
-    for (int tidx = threadIdx.x; tidx < num_rows; tidx += blockDim.x) {
+    for (int row = threadIdx.x; row < num_rows; row += blockDim.x) {
         auto temp = zero<ValueType>();
-        for (auto nnz = mat.row_ptrs[tidx]; nnz < mat.row_ptrs[tidx + 1];
-             nnz++) {
+        for (auto nnz = mat.row_ptrs[row]; nnz < mat.row_ptrs[row + 1]; nnz++) {
             const auto col_idx = col[nnz];
             temp += alpha * val[nnz] * b[col_idx];
         }
-        x[tidx] = temp + beta * x[tidx];
+        x[row] = temp + beta * x[row];
     }
 }
 
--- a/common/cuda_hip/matrix/batch_csr_kernels.hpp.inc
+++ b/common/cuda_hip/matrix/batch_csr_kernels.hpp.inc
@@ -1,0 +1,144 @@
+/*******************************<GINKGO LICENSE>******************************
+Copyright (c) 2017-2023, the Ginkgo authors
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+
+1. Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+
+2. Redistributions in binary form must reproduce the above copyright
+notice, this list of conditions and the following disclaimer in the
+documentation and/or other materials provided with the distribution.
+
+3. Neither the name of the copyright holder nor the names of its
+contributors may be used to endorse or promote products derived from
+this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS
+IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED
+TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A
+PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
+
+
+template <typename ValueType, typename IndexType>
+__device__ __forceinline__ void simple_apply(
+    const gko::batch::matrix::csr::batch_item<const ValueType, IndexType>& mat,
+    const ValueType* const __restrict__ b, ValueType* const __restrict__ x)
+{
+    const auto num_rows = mat.num_rows;
+    const auto val = mat.values;
+    const auto col = mat.col_idxs;
+    for (int tidx = threadIdx.x; tidx < num_rows; tidx += blockDim.x) {
+        auto temp = zero<ValueType>();
+        for (auto nnz = mat.row_ptrs[tidx]; nnz < mat.row_ptrs[tidx + 1];
+             nnz++) {
+            const auto col_idx = col[nnz];
+            temp += val[nnz] * b[col_idx];
+        }
+        x[tidx] = temp;
+    }
+}
+
+template <typename ValueType, typename IndexType>
+__global__ __launch_bounds__(
+    default_block_size,
+    sm_oversubscription) void simple_apply_kernel(const gko::batch::matrix::
+                                                      csr::uniform_batch<
+                                                          const ValueType,
+                                                          IndexType>
+                                                          mat,
+                                                  const gko::batch::
+                                                      multi_vector::
+                                                          uniform_batch<
+                                                              const ValueType>
+                                                              b,
+                                                  const gko::batch::
+                                                      multi_vector::
+                                                          uniform_batch<
+                                                              ValueType>
+                                                              x)
+{
+    for (size_type batch_id = blockIdx.x; batch_id < mat.num_batch_items;
+         batch_id += gridDim.x) {
+        const auto mat_b =
+            gko::batch::matrix::extract_batch_item(mat, batch_id);
+        const auto b_b = gko::batch::extract_batch_item(b, batch_id);
+        const auto x_b = gko::batch::extract_batch_item(x, batch_id);
+        simple_apply(mat_b, b_b.values, x_b.values);
+    }
+}
+
+
+template <typename ValueType, typename IndexType>
+__device__ __forceinline__ void advanced_apply(
+    const ValueType alpha,
+    const gko::batch::matrix::csr::batch_item<const ValueType, IndexType>& mat,
+    const ValueType* const __restrict__ b, const ValueType beta,
+    ValueType* const __restrict__ x)
+{
+    const auto num_rows = mat.num_rows;
+    const auto val = mat.values;
+    const auto col = mat.col_idxs;
+    for (int tidx = threadIdx.x; tidx < num_rows; tidx += blockDim.x) {
+        auto temp = zero<ValueType>();
+        for (auto nnz = mat.row_ptrs[tidx]; nnz < mat.row_ptrs[tidx + 1];
+             nnz++) {
+            const auto col_idx = col[nnz];
+            temp += alpha * val[nnz] * b[col_idx];
+        }
+        x[tidx] = temp + beta * x[tidx];
+    }
+}
+
+template <typename ValueType, typename IndexType>
+__global__ __launch_bounds__(
+    default_block_size,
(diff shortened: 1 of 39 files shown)


common/cuda_hip/matrix/batch_csr_kernels.hpp.inc
```diff
@@ -0,0 +1,144 @@
...
+__device__ __forceinline__ void advanced_apply(
+    const ValueType alpha,
+    const gko::batch::matrix::csr::batch_item<const ValueType, IndexType>& mat,
+    const ValueType* const __restrict__ b, const ValueType beta,
+    ValueType* const __restrict__ x)
+{
+    const auto num_rows = mat.num_rows;
+    const auto val = mat.values;
+    const auto col = mat.col_idxs;
+    for (int tidx = threadIdx.x; tidx < num_rows; tidx += blockDim.x) {
```
upsj, line 93: nit: naming
```suggestion
    for (int row = threadIdx.x; row < num_rows; row += blockDim.x) {
```

common/cuda_hip/matrix/batch_csr_kernels.hpp.inc
```diff
@@ -0,0 +1,144 @@
...
+
+template <typename ValueType, typename IndexType>
+__device__ __forceinline__ void simple_apply(
+    const gko::batch::matrix::csr::batch_item<const ValueType, IndexType>& mat,
+    const ValueType* const __restrict__ b, ValueType* const __restrict__ x)
+{
+    const auto num_rows = mat.num_rows;
+    const auto val = mat.values;
+    const auto col = mat.col_idxs;
+    for (int tidx = threadIdx.x; tidx < num_rows; tidx += blockDim.x) {
```
upsj, line 42:
```suggestion
    for (int row = threadIdx.x; row < num_rows; row += blockDim.x) {
```

core/matrix/csr.cpp
```diff
@@ -471,8 +471,9 @@ void Csr<ValueType, IndexType>::read(device_mat_data&& data)
     auto arrays = data.empty_out();
     this->row_ptrs_.resize_and_reset(size[0] + 1);
     this->set_size(size);
-    this->values_ = std::move(arrays.values);
-    this->col_idxs_ = std::move(arrays.col_idxs);
+    // Need to copy instead of move, as move into view data is not supported.
+    this->values_ = arrays.values;
+    this->col_idxs_ = arrays.col_idxs;
```
upsj, line 476: This is a pessimization for many use cases, and defeats the purposes of this being an rvalue overload. It sounds like calling `copy_from` might be more appropriate here.

core/test/matrix/batch_csr.cpp
```diff
@@ -0,0 +1,549 @@
...
+}
+
+
+TYPED_TEST(Csr, SparseMtxKnowsItsSizeAndValues)
+{
+    this->assert_equal_to_original_sparse_mtx(this->sp_mtx.get());
+}
+
+
+// TYPED_TEST(Csr, CanBeEmpty)
```
upsj, line 164: I don't think we should commit commented out code. you can disable the tests by prefixing them with `DISABLED_` IIRC

dpcpp/matrix/batch_csr_kernels.hpp.inc
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        temp = ::gko::kernels::dpcpp::reduce(
+            sg, temp, [](ValueType a, ValueType b) { return a + b; });
+
+
+        if (sg.get_local_id() == 0) {
+            c[row] = alpha * temp + beta * c[row];
+        }
+    }
```
upsj, line 83: This is a different level of optimization than the other kernels. Is that intended?

omp/matrix/batch_ell_kernels.cpp
```diff
@@ -37,7 +37,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
 #include <ginkgo/core/base/batch_multi_vector.hpp>
-#include <ginkgo/core/matrix/batch_dense.hpp>
+#include <ginkgo/core/matrix/batch_ell.hpp>
```
upsj, line 40: leftover?

upsj (dismissed):
This PR is pretty late, so I only went over the implementation, not the tests. Most of it LGTM, there is just a behavioral change in here that I believe overshoots in fixing a specific edge case.

pratikvn:
@upsj , this was not meant to be in the release. I had a bit of time, so decided to work on it. Thanks for reviewing, but this is still a WIP, that is why I havent requested any reviews or marked it ready-for-review  yet.

upsj:
@pratikvn phew, thank you, that would have been a night shift otherwise 😄

core/matrix/csr.cpp
pratikvn: Yes, I agree. I was also planning on checking if the arrays are `is_owning()` and move if so, and copy otherwise

pratikvn:
format!

pratikvn:
format!

core/test/matrix/batch_csr.cpp
```diff
@@ -0,0 +1,542 @@
...
+    auto mat1 =
+        gko::initialize<CsrMtx>({{1.0, 0.0, 0.0}, {0.0, 2.0, 0.0}}, this->exec);
+    auto bat_m = gko::batch::create_from_item<BatchCsrMtx>(
+        this->exec, std::vector<CsrMtx*>{mat1.get(), mat1.get(), mat1.get()},
+        2);
+
+    auto m =
+        gko::batch::create_from_item<BatchCsrMtx>(this->exec, 3, mat1.get(), 2);
+
+    GKO_ASSERT_BATCH_MTX_NEAR(bat_m.get(), m.get(), 1e-14);
```
MarcelKoch, line 332: Why does this need `eps > 0`? I think all operations are just copies.

core/test/matrix/batch_csr.cpp
```diff
@@ -0,0 +1,542 @@
...
+        this->exec, std::vector<CsrMtx*>{mat1.get(), mat2.get()}, 2);
+    auto m_ref = gko::batch::create_from_item<BatchCsrMtx>(
+        this->exec,
+        std::vector<CsrMtx*>{mat1.get(), mat2.get(), mat1.get(), mat2.get(),
+                             mat1.get(), mat2.get()},
+        2);
+
+    auto m2 = gko::batch::duplicate<BatchCsrMtx>(this->exec, 3, m.get(), 2);
+
+    GKO_ASSERT_BATCH_MTX_NEAR(m2.get(), m_ref.get(), 1e-14);
```
MarcelKoch, line 355: same

core/test/matrix/batch_csr.cpp
```diff
@@ -0,0 +1,542 @@
...
+    GKO_ASSERT_MTX_NEAR(csr_mats[1].get(), mat2.get(), 0.);
+}
+
+
+TYPED_TEST(Csr, CanBeListConstructed)
+{
+    using value_type = typename TestFixture::value_type;
+    using index_type = typename TestFixture::index_type;
+    using BatchCsrMtx = typename TestFixture::BatchCsrMtx;
+    using CsrMtx = typename TestFixture::CsrMtx;
```
MarcelKoch, line 380:
```suggestion
```

core/test/matrix/batch_csr.cpp
```diff
@@ -0,0 +1,542 @@
...
+
+    this->assert_equal_to_original_sparse_mtx(m.get());
+}
+
+
+TYPED_TEST(Csr, ThrowsForDataWithDifferentNnz)
+{
+    using value_type = typename TestFixture::value_type;
+    using index_type = typename TestFixture::index_type;
+    using BatchCsrMtx = typename TestFixture::BatchCsrMtx;
```
MarcelKoch, line 483:
```suggestion
```

core/test/matrix/batch_csr.cpp
```diff
@@ -0,0 +1,542 @@
...
+        gko::batch::detail::assert_same_sparsity_in_batched_data(vec_data),
+        gko::NotImplemented);
+}
+
+
+TYPED_TEST(Csr, ThrowsForDataWithDifferentSparsity)
+{
+    using value_type = typename TestFixture::value_type;
+    using index_type = typename TestFixture::index_type;
+    using BatchCsrMtx = typename TestFixture::BatchCsrMtx;
```
MarcelKoch, line 505:
```suggestion
```

reference/test/matrix/batch_csr_kernels.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+TYPED_TEST(Csr, AppliesToBatchMultiVector)
+{
+    using T = typename TestFixture::value_type;
+
+    this->mtx_0->apply(this->b_0.get(), this->x_0.get());
+
+    this->mtx_00->apply(this->b_00.get(), this->x_00.get());
+    this->mtx_01->apply(this->b_01.get(), this->x_01.get());
+    auto res = gko::batch::unbatch<gko::batch::MultiVector<T>>(this->x_0.get());
```
MarcelKoch, line 125:
```suggestion

    auto res = gko::batch::unbatch<gko::batch::MultiVector<T>>(this->x_0.get());
```
for AAA

MarcelKoch: More generally, AAA seems to not be followed correctly in this file.

MarcelKoch (changes requested):
looks good, only some smaller comments

pratikvn:
format!

core/matrix/csr.cpp
```diff
@@ -443,8 +443,26 @@ void Csr<ValueType, IndexType>::read(device_mat_data&& data)
     auto arrays = data.empty_out();
     this->row_ptrs_.resize_and_reset(size[0] + 1);
     this->set_size(size);
-    this->values_ = std::move(arrays.values);
-    this->col_idxs_ = std::move(arrays.col_idxs);
+    // Need to copy instead of move, as move into view data is not supported.
```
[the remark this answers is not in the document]
pratikvn: No, move into view is not supported. It does not make sense to move data from an owning array to a non-owning one. I think the behaviour in array is correct. To have correct behaviour, we can either have all cases just copy, or copy only when the arrays of `this` are non-owning, and I think it is better to avoid deep copies wherever possible, so I think the implementation here is a good solution.

reference/test/matrix/batch_csr_kernels.cpp
pratikvn: I believe it should be. Act is only the batch matrix apply. The normal csr matrix apply comes in the assert block.

MarcelKoch: Ok, I got a bit confused and thought the `mtx_0->apply` was the arrange part. But still I think we usually group applies of the to-test object together with the reference object. We do this in the common tests (reference and device apply) and in the MPI tests. So I would still suggest grouping all applies into the act category and then the unbatch into the assert caategory.

core/matrix/csr.cpp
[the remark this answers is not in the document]
MarcelKoch: I will create that PR.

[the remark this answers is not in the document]
pratikvn: I think the behaviour of `const device_mat_data&` overload does copy. This function is just a helper that reduces code duplication by only taking in a temporary object. 

I agree that it is an edge-case, but it is very useful in implementing read for batched objects. If and when we implement non-uniform batch objects, I think we should follow a similar approach.

reference/test/matrix/batch_csr_kernels.cpp
pratikvn: I think that defeats the purpose of the AAA structure. IMO, the Act block should only contain function calls/declarations that are being tested, otherwise it is unclear what is being tested.

core/matrix/csr.cpp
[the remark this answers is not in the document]
MarcelKoch: Sadly the `const device_mat_data&` overload only copies the input, but it does not copy into the matrix's arrays, it moves again. So IMO, making the `const device_mat_data&` overload also copy the arrays would fix the issue here, while keeping the current move semantics.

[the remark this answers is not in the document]
MarcelKoch: I would suggest that the input itself is not copied, but instead only the arrays are copied. This should have no overhead.

Moving data into views doesn't fail, it does exactly what is should do. The move overload takes ownership of the input data, and thus the matrix takes ownership of the stored arrays. So afterwards the matrix can't have array views anymore. I agree that we should clarify this, but I think this is the correct behavior.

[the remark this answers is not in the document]
upsj: I tend to agree, reading into a view was never something we really considered before. This only matters for matrix types that reuse the arrays anyways (Csr and Coo), so it's not a notable maintenance overhead to duplicate a tiny bit of the implementation. We should have tests for both cases in place though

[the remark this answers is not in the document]
pratikvn: I am fine with copying the arrays instead of copying the input. As we probably do that in another PR, maybe we push this PR forward with these changes ?

pratikvn:
format!

dpcpp/matrix/batch_csr_kernels.hpp.inc
```diff
@@ -0,0 +1,44 @@
...
+    const auto val = mat.values;
+    const auto col = mat.col_idxs;
+    for (int row = item_ct1.get_local_linear_id(); row < num_rows;
+         row += item_ct1.get_local_range().size()) {
+        auto temp = zero<ValueType>();
+        for (auto nnz = mat.row_ptrs[row]; nnz < mat.row_ptrs[row + 1]; nnz++) {
+            const auto col_idx = col[nnz];
+            temp += alpha * val[nnz] * b[col_idx];
+        }
+        x[row] = temp + beta * x[row];
```
yhmtsai, line 42:
```suggestion
            temp += val[nnz] * b[col_idx];
        }
        x[row] = alpha * temp + beta * x[row];
```

reference/test/solver/batch_bicgstab_kernels.cpp
```diff
@@ -247,6 +249,42 @@ TYPED_TEST(BatchBicgstab, CanSolveEllSystem)
 }
 
 
+TYPED_TEST(BatchBicgstab, CanSolveCsrSystem)
```
yhmtsai, line 252: Maybe this can be change to the uniform test for different matrix format.
(not need in the pr)

include/ginkgo/core/matrix/batch_csr.hpp
```diff
@@ -0,0 +1,341 @@
...
+    {
+        return values_.get_size();
+    }
+
+    /**
+     * Returns the number of stored elements in each batch item.
+     *
+     * @return the number of stored elements per batch item.
+     */
+    size_type get_num_elements_per_item() const noexcept
```
yhmtsai, line 167: we use `get_num_stored_elements()` in other place.
should it be `get_num_stored_elements_per_item()` for consistence? it's quite long though

include/ginkgo/core/matrix/batch_csr.hpp
```diff
@@ -0,0 +1,341 @@
...
+namespace gko {
+namespace batch {
+namespace matrix {
+
+
+/**
+ * Csr is a general sparse matrix format that stores the column indices for each
+ * nonzero entry and a cumulative sum of the number of nonzeros in each row. It
+ * is one of the most popular sparse matrix formats due to its versatility and
+ * ability to store a wide range of sparsity patterns in an efficient fashion.
```
yhmtsai, line 33: not update documentation with batch like batchEll?

include/ginkgo/core/matrix/batch_csr.hpp
```diff
@@ -0,0 +1,341 @@
...
+        : EnableBatchLinOp<Csr>(exec, size),
+          values_{exec, std::forward<ValuesArray>(values)},
+          col_idxs_{exec, std::forward<ColIdxsArray>(col_idxs)},
+          row_ptrs_{exec, std::forward<RowPtrsArray>(row_ptrs)}
+    {
+        // Ensure that the value and col_idxs arrays have the correct size
+        auto max_num_elems = this->get_common_size()[0] *
+                             this->get_common_size()[1] *
+                             this->get_num_batch_items();
+        GKO_ASSERT(values_.get_size() <= max_num_elems);
```
yhmtsai, line 317: do you need to check this? it is just unused from the batch csr

yhmtsai: ah, the get_num_stored_elements is based on values.size(), so it is required to only store the batch value

reference/test/matrix/batch_csr_kernels.cpp
```diff
@@ -0,0 +1,230 @@
...
+    Csr()
+        : exec(gko::ReferenceExecutor::create()),
+          mtx_0(gko::batch::initialize<BMtx>(
+              {{I<T>({1.0, -1.0, 1.5}), I<T>({-2.0, 2.0, 3.0})},
+               {{1.0, -2.0, -0.5}, {1.0, -2.5, 4.0}}},
+              exec, 6)),
+          mtx_00(gko::initialize<CsrMtx>(
+              {I<T>({1.0, -1.0, 1.5}), I<T>({-2.0, 2.0, 3.0})}, exec)),
+          mtx_01(gko::initialize<CsrMtx>(
+              {I<T>({1.0, -2.0, -0.5}), I<T>({1.0, -2.5, 4.0})}, exec)),
```
yhmtsai, line 46: maybe use sparse data not the full dense one?

reference/test/matrix/batch_csr_kernels.cpp
```diff
@@ -0,0 +1,230 @@
...
+    std::unique_ptr<DenseMtx> b_01;
+    std::unique_ptr<BMVec> x_0;
+    std::unique_ptr<DenseMtx> x_00;
+    std::unique_ptr<DenseMtx> x_01;
+
+    std::ranlux48 rand_engine;
+};
+
+
+TYPED_TEST_SUITE(Csr, gko::test::ValueTypes, TypenameNameGenerator);
```
yhmtsai, line 86:
```suggestion
TYPED_TEST_SUITE(Csr, gko::test::ValueTypes, TypenameNameGenerator);
```

reference/test/solver/batch_bicgstab_kernels.cpp
pratikvn: Yes, maybe I will handle it in another PR

include/ginkgo/core/matrix/batch_csr.hpp
yhmtsai: I think we try to avoid nnz because the stored element can be zeros

include/ginkgo/core/matrix/batch_csr.hpp
yhmtsai: It's like CSR not batchCSR

pratikvn: The note below clarifies that this is a batched matrix fomat, and how the batches are stored. I think being within the namespace also makes that clear. It is similar to `batch::Ell`. But I think can add more clarification to make it clear.

core/test/matrix/batch_csr.cpp
```diff
@@ -0,0 +1,511 @@
...
+        {2, 3}, {{0, 0, 1.0}, {1, 1, 2.0}, {1, 2, 3.0}}));
+
+    auto m = gko::batch::read<value_type, index_type, BatchCsrMtx>(this->exec,
+                                                                   vec_data, 3);
+
+    this->assert_equal_to_original_sparse_mtx(m.get());
+}
+
+
+TYPED_TEST(Csr, ThrowsForDataWithDifferentNnz)
```
MarcelKoch, line 450: This and the next test are already done in batch_ell. And generally, these tests are not matrix type specific, so they should be in their own file.

reference/test/matrix/batch_csr_kernels.cpp
```diff
@@ -0,0 +1,229 @@
...
+    this->mtx_01->apply(this->b_01.get(), this->x_01.get());
+    auto res = gko::batch::unbatch<gko::batch::MultiVector<T>>(this->x_0.get());
+    GKO_ASSERT_MTX_NEAR(res[0].get(), this->x_00.get(), r<T>::value);
+    GKO_ASSERT_MTX_NEAR(res[1].get(), this->x_01.get(), r<T>::value);
+}
+
+
+TYPED_TEST(Csr, AppliesLinearCombinationToBatchMultiVector)
+{
+    using BMtx = typename TestFixture::BMtx;
```
MarcelKoch, line 119: unused
```suggestion
```

MarcelKoch (approved):
LGTM. I think the testing strategy should be revisited at some point. A lot of tests in core test functionality that is composed of smaller parts, but the implementation is the same for all different matrix types. For example, `batch::write` only requires that the type has `create_const_view_for_item`. Because of that many tests are repetetive and could probably be unified.

pratikvn:
As the previous pipeline had fully passed, and I only removed a couple of tests, I will just go ahead and merge this PR, not waiting for the whole pipeline to complete.

PR #1460, opened 2023-11-14
Use Reuse tool for license handling

MarcelKoch:
This PR enables using the reuse tool [[1]](https://reuse.software/dev/) for adding our license to our files. This replaces the current `add_license.sh` script. The tool is automatically invoked as part of our pre-commit hooks.

Sidenote: There is an official pre-commit hook for the reuse tool, but only for the linting. This has several drawbacks, 1. it requires that EVERY file has a license (I think we should still do that, but in another PR), 2. our current workflow is to add the license automatically, not just warn about it. There is some discussion for adding an official hook to do the annotation, so perhaps we can remove our manual hook in the future.


.pre-commit-config.yaml
```diff
@@ -0,0 +1,22 @@
...
+      types_or: [c, c++, cuda, inc]
+      exclude: third_party/SuiteSparse/AMD/.*
+-   repo: local
+    hooks:
+    # The official reuse hook only supports calling lint, so we need our own hook
+    - id: reuse-annotate
+      name: reuse-annotate
+      entry: reuse annotate --license BSD-3-Clause --copyright 'The Ginkgo authors' --year 2017-2023 --style c
+      language: python
+      additional_dependencies: [reuse]
```
yhmtsai, line 15: Don't you need another entry to describe reuse?

yhmtsai: no, I thought there should be one entry to declare the reuse tool like mirror-clang-format.
I do not find it with the repo url. Is reuse already a keyword in pre-commit

MarcelKoch: No this hook works different. It uses `repo: local`, which means it will just execute the `entry`. How it will be executed is specified by the `language`. To make sure that the `reuse` tool is installed, I added it to the `additional_dependencies`. Again, the language specifies how these dependencies are handled, and in the case of python, it just means that the dependencies will be installed through pip.

yhmtsai: Thanks for explanation.
Maybe I miss some part of documentation.
python is not specified about support for `additional_dependencies` from https://pre-commit.com/#python, but conda is.
I try it on my side. It installed reuse and used the package when running pre-commit.

yhmtsai:
something is not rebased yet

greole:
> Sidenote: There is an official pre-commit hook for the reuse tool, but only for the linting. This has several drawbacks, 1. it requires that EVERY file has a license (I think we should still do that, but in another PR)
Cant this be avoided by a `.reuse/dep5` file, which would some entries like

```
Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/    
Upstream-Name: NeoFOAM                                                       
Upstream-Contact: NeoFOAM authors    
Source: https://git.example.com/jane/my-project
                                                                                                               
Files: .github/*                                                                
Copyright: 2023 NeoFOAM authors                                                 
License: MPL-2.0
```

MarcelKoch:
@greole yes, but the major drawback is that the official hook doesn't automatically add the license.

.github/check-format.sh
```diff
@@ -9,7 +9,7 @@ pipx run pre-commit run --show-diff-on-failure --color=always --from-ref "origin
 
 echo -n "Collecting information on changed files"
 
-git checkout -- dev_tools/scripts/*.sh dev_tools/scripts/update-ginkgo-header.py
+git checkout -- dev_tools/scripts/update-ginkgo-header.py
```
yhmtsai, line 12: pr-format should also pick the other sh file?

CMakeLists.txt
```diff
@@ -377,11 +377,6 @@ if(GINKGO_DEVEL_TOOLS)
                       COMMAND bash -c "${PRE_COMMIT} run"
                       WORKING_DIRECTORY ${Ginkgo_SOURCE_DIR}
                       VERBATIM)
-
-    add_custom_target(add_license
-                      COMMAND ${Ginkgo_SOURCE_DIR}/dev_tools/scripts/add_license.sh
-                      WORKING_DIRECTORY ${Ginkgo_SOURCE_DIR})
-    add_dependencies(format add_license)
```
yhmtsai, line 384: the format will handle the license now, right?

.github/check-format.sh
yhmtsai: also I do not see this file in this pr. Is the base of this pr wrong one?

CMakeLists.txt
MarcelKoch: the pre-commit hook will handle it.

include/ginkgo/core/log/papi.hpp
```diff
@@ -12,11 +12,11 @@
 #if GKO_HAVE_PAPI_SDE
 
 
+#include <sde_lib.h>
 #include <cstddef>
 #include <iostream>
 #include <map>
 #include <mutex>
-#include <sde_lib.h>
```
yhmtsai, line 19: should it be changed?

include/ginkgo/core/log/papi.hpp
```diff
@@ -12,11 +12,11 @@
 #if GKO_HAVE_PAPI_SDE
 
 
+#include <sde_lib.h>
```
pratikvn, line 15: Was this automatic ? Should this probably be in a separate section ?

.github/check-format.sh
```diff
@@ -9,6 +9,8 @@ pipx run pre-commit run --show-diff-on-failure --color=always --from-ref "origin
 
 echo -n "Collecting information on changed files"
 
+git checkout -- dev_tools/scripts/*.sh
```
pratikvn, line 12: Why is this needed ?

pratikvn (approved):
LGTM!

.github/check-format.sh
MarcelKoch: It was there before. I deleted it by mistake. But it's needed to filter out formatting changes on the dev_tools scripts.

yhmtsai (approved):

include/ginkgo/core/log/papi.hpp
MarcelKoch: I think this is missing `format_header`. But I think we should just leave it, since it will most likely change again in #1466.

pratikvn (approved):

yhmtsai (approved):

PR #1461, opened 2023-11-14
Use fixed clang format

MarcelKoch:
**Old PR: #1404**

This PR enables us to force users to use the same version of clang-format as our CI. This is done by requiring contributors to install [pre-commit](https://pre-commit.com/) (which can be done easily through pip/pipx). pre-commit allows us to fix the used version of clang-format, so that contributors will use exactly the same version as our CI.

If pre-commit can't be detected during cmake an error is thrown. Otherwise, the pre-commit hooks are installed.

Since this replaces our git-cmake-format, I have removed it.

To see it in action: here is an PR in a fork that shows the CI use cases https://github.com/MarcelKoch/ginkgo/pull/9

Additional changes:

- change not supported values in `.clang-format`
- run pre-commit checks over all files

# Warning

For those using git-worktrees, if one working tree is updated with this PR it will overwrite the git hooks for all working trees (I don't think you can have different hooks for different working trees). The easy fix is of course to just update all used working trees accordingly.

**Note:** I think the previous PR #1405 got closed because it was based not on develop and it was from my fork, but who knows?


MarcelKoch:
Putting the link to the old PR more prominently: #1405

MarcelKoch:
When this is merged, we should merge ginkgo-project/git-cmake-format#13

upsj (approved):

MarcelKoch:
I think the current `check-format` error is again due to the CI using the configuration from develop. So we have to ignore it again.

examples/kokkos_assembly/kokkos_assembly.cpp
```diff
@@ -8,8 +8,8 @@
 
 
 #include <omp.h>
-#include <Kokkos_Core.hpp>
 #include <ginkgo/ginkgo.hpp>
+#include <Kokkos_Core.hpp>
```
yhmtsai, line 12: I think it should be separated

.github/workflows/check-formatting.yml
```diff
@@ -0,0 +1,21 @@
+name: Check formatting
+on: workflow_call
+
+jobs:
+  pre-commit:
+    name: Run pre-commit hooks
+    runs-on: ubuntu-latest
```
yhmtsai, line 7:
```suggestion
    runs-on: ubuntu-22.04
```
because format_header currently still use the system clang-format-14, using 22.04 might be more stable

yhmtsai (changes requested):
should old pr be #1405 ?
Could you add the pre-commit requirement into the README.md and pre-commit and mirrors-clang-format license into ABOUT-LICENSING.md.
I just noticed ABOUT-LICENSING render is broken. Could you also fix them?

.github/workflows/check-formatting.yml
MarcelKoch: I will change this, but what is the purpose of `format_header.sh` anyway? It seems to reorder the includes, but that should be covered by clang-format already. Special handling like the `force-on-top` could be handled by disabling clang-format.

MarcelKoch:
@yhmtsai I'm not sure if we need to add either pre-commit, or the clang-format hook to our licensing description. Our ginkgo package (the source code + whatever is in the third_party stuff) does not contain any code or binaries from either pre-commit or clang-format. The pre-commit config could be licensed, but we didn't get into that yet.

yhmtsai:
just copy paste the comment from #1405 if want to use clang-format from pre-commit in the editor. (easier find in the future)
**It's not pre-commit intended usage. It might be failed if they change the cache way.**
The following script such that we can use the auto installation in editor easily, too.
```
#!/usr/bin/env bash

PRE_COMMIT_CACHE="${HOME}/.cache/pre-commit"
# get the python environment folder from pre-commit
VENV="$(sqlite3 -header -csv ${PRE_COMMIT_CACHE}/db.db "select * from repos;" | grep clang-format,v14.0.0 | sed -E 's/.*,.*,(.*)/\1/g')"
ACTIVATE="$(find ${VENV} -name activate -print -quit)"
# activate env
source "${ACTIVATE}"
# forward the arguments
clang-format "$@"
```

.github/workflows/check-formatting.yml
yhmtsai: no, clang-format reordering does not fit our main header (it's somehow complex, so I would also like to change) and the ordering can not be handled across sections (or across-section reorder will only contain one space line for sep?)

yhmtsai: the format_header did two things:
1. put the correct main header on the top
2. reorder the header with clang-format 12+ and then replace one line sep -> two line sep.

MarcelKoch: There is a way to achieve 1. with clang-format. But that requires changing
```
#include <ginkgo/core/what/ever.hpp>
```
into
```
#include "ginkgo/core/what/ever.hpp"
```
So clang-format recognizes the main header of a source file only if its in `""`, but not if its in `<>`.
I will also look into 2.

yhmtsai: No, the main header issue is we have several rules for main header. Please take a look at https://github.com/ginkgo-project/ginkgo/blob/develop/dev_tools/scripts/config
Some of them could be merged, but it's still complex if we do not have certain forced rules.
We try writing the general rules in https://github.com/ginkgo-project/ginkgo/wiki/Contributing-guidelines#main-header
From https://clang.llvm.org/docs/ClangFormatStyleOptions.html#includeismainregex, you can only have simple regex for all of them.

Thanks for looking at that, look forward to avoiding complex format_header.
I suggest it should be another pr if you find something easy to maintain.

yhmtsai (commented):
Something I can not put comments directly but needs updates:
- git-cmake-format can be deleted in ABOUT-LICENSING.md
- INSTALL.md replaces git-cmake-format by pre-commit
- delete third-party git-cmake-format
- dummy_hook/CMakeLists.txt needs to update

yhmtsai (approved):

include/ginkgo/core/log/papi.hpp
```diff
@@ -16,6 +16,8 @@
 #include <iostream>
 #include <map>
 #include <mutex>
+
+
 #include <sde_lib.h>
```
yhmtsai, line 21: actually, I do not get this changes.
do you add new breaks and only run format without format_header?

MarcelKoch: Yes, I forgot the format_header. I will also run this.

yhmtsai: although I do not know why the format_header change it, it shuold keep everything in `#if` block

yhmtsai: just go through again the format_header. the thing in #if only raises a warning only. the header is still considered in clang-format with regroup (in format_header)

MarcelKoch:
format!

PR #1463, opened 2023-11-15
Fix libc++ incompatibility

upsj:
This is an issue detected in https://github.com/conan-io/conan-center-index/pull/21058, which uses a different integer type for `std::chrono::nanosecond`

### DIFF
--- a/core/base/timer.cpp
+++ b/core/base/timer.cpp
@@ -96,7 +96,7 @@ void CpuTimer::wait(time_point& time) {}
 std::chrono::nanoseconds CpuTimer::difference_async(const time_point& start,
                                                     const time_point& stop)
 {
-    return std::chrono::duration_cast<std::chrono::nanoseconds, int64>(
+    return std::chrono::duration_cast<std::chrono::nanoseconds>(
         stop.data_.chrono - start.data_.chrono);
 }
 


MarcelKoch (approved):

core/base/timer.cpp
```diff
@@ -96,7 +96,7 @@ void CpuTimer::wait(time_point& time) {}
 std::chrono::nanoseconds CpuTimer::difference_async(const time_point& start,
                                                     const time_point& stop)
 {
-    return std::chrono::duration_cast<std::chrono::nanoseconds, int64>(
+    return std::chrono::duration_cast<std::chrono::nanoseconds>(
         stop.data_.chrono - start.data_.chrono);
```
yhmtsai, line 100: Just to ensure I got it correctly.
duration_cast signature:
```
template< class ToDuration, class Rep, class Period >
constexpr ToDuration duration_cast( const std::chrono::duration<Rep,Period>& d );
```
The issue is from type definition of `std::chrono::nanoseconds`
`std::chrono::duration</* int64 */, std::nano>` standard only require it at least 64 bit signed integer.
Thus, it might be not int64.
The type also affect `.count()` return type. 
Could you check that the `.count()` has static_cast to desired type if it is the case?

yhmtsai (approved):
LGTM. but maybe need to check `.count()` type with `static_cast`

core/base/timer.cpp
upsj: The issue is basically that I used to return int64 before and now return nanoseconds, and I didn't change the function call. It only worked accidentally.

upsj: I checked, we only use `.count()` in places relevant to this issue where it will be cast to `double`, so no problems expected

PR #1466, opened 2023-11-16
Remove format_header.sh

MarcelKoch:
This PR removes the format_header.sh in favor of purely relying on clang-format. The .clang-format is updated to allow regrouping of includes. In total, out of the affected 1000 files, only ~200 have non-whitespace changes. Many of those are regarding the ginkgo main header, or previously missing 

## Include groups (ordered by their prority)
1. `<oneapi/...` since we used `force-top` with that a lot.
2. `<hip/hip_runtime.h>` same
3. 'main' include
4. STL
5. backend library, i.e. cuda/hip/omp/mpi/...
6. third party library
7. Ginkgo main header
8. Ginkgo headers
9. Other `<...>` includes
10. All `"..."` includes

This is mostly based on the already used regrouping in `format_header.sh`.

## Notable Changes
- The 'main' header of an implementation file now uses `"..."`, instead of `<...>`. This is required, since clang-format can't deduce the main header otherwise.
- `force-top` is removed and instead handling by using negative priorities in .clang-format. In very few cases this is not sufficient, and there `clang-format on/off` achieves the same results.
- Only **1** empty line between separate groups of includes. It is not possible to adjust that with a clang-format setting.
- The `<ginkgo/ginkgo.hpp>` header is now ordered after standard library includes
- Some 'main' headers are not recognized. This concerns especially backend headers, which add stuff to code headers, for example `cuda/matrix/batch_struct.hpp` will not have `core/matrix/batch_struct.hpp` as its main header anymore.
- Split up implementations, like for the Jacobi preconditioner, don't recognize the 'main' header correctly. One solution could be to use the same splitting approach as for CSR/FBCSR kernels.

Since the `format_header.sh` script is very complex, I think these changes are a small price to removing it.


upsj:
I like it, if it performs well, this should make IDE integration of our header ordering much smoother

MarcelKoch:
The non-whitespace diff can be obtained with:
```
git diff -b -B -w -z --ignore-space-at-eol --ignore-blank-lines 195e5bf3..195e5bf3~1
```
The diff is also attached here.
[diff.txt](https://github.com/ginkgo-project/ginkgo/files/13379882/diff.txt)

upsj (approved):
LGTM!

core/base/batch_multi_vector_kernels.hpp
```diff
@@ -6,13 +6,11 @@
 #define GKO_CORE_BASE_BATCH_MULTI_VECTOR_KERNELS_HPP_
```
pratikvn, line 8: Change this to also a single newline ?

core/base/array.cpp
```diff
@@ -2,12 +2,10 @@
 //
 // SPDX-License-Identifier: BSD-3-Clause
 
-#include <ginkgo/core/base/array.hpp>
-
+#include "ginkgo/core/base/array.hpp"
```
pratikvn, line 5: Is it not possible to have these main headers also in `<...>` fashion ? IMO it looks a bit weird with initially having `" ..."` and then `<...>` .

MarcelKoch: No that is not supported by clang-format

core/base/batch_multi_vector_kernels.hpp
MarcelKoch: This is not handled by clang-format. clang-format only changes the number of empty lines within the include declaration block. For everything before or after, the normal rule of max. 2 empty lines holds.

yhmtsai:
`<oneapi/..` -> `<oneapi/dpl...` only the dpl need to be on the top. the other oneapi before dpl may also introduce issues.
clang format does not take care of the "" for local and <> from global, so it may still destroy the regroup
hip_runtime should be in hip group. only a few hip tests request it on top which can be handled by clang-format off
9 can be merged into 6? 
It changes a lot of the format rules in the documentation
I would say this pr changes the format style not just remove format_header

yhmtsai (commented):
the regroup does not take care the header. for example cb_gmres example

MarcelKoch:
The main goal of the PR is to remove `format_header.sh`. The formatting changes are a by product of that. 

> 9 can be merged into 6?

I don't follow, can you elaborate?

> the regroup does not take care the header. for example cb_gmres example

Do you mean the wrong comment on the main ginkgo header file?

MarcelKoch:
> clang format does not take care of the "" for local and <> from global, so it may still destroy the regroup

I think it does. Any `<>` is priority 1-4, the `<ginkgo/` headers are priority 5-6, and everything else, which includes `""` is priority 7.

yhmtsai:
For me, removing format_header means that we have the equivalent way to handle it. 
Removing something that drives the guideline is a bit dangerous to me although it may be the same thing in the end.

> 9 can be merged into 6?

Do we consider all the other library from outside (`<...>` not `"..."`) is third party? I think it should be the case for ginkgo.

> the regroup does not take care the header. for example cb_gmres

It was shown in my first review, but it is not there anymore.
```
// ginkgo main header
#include <ginkgo/ginkgo.hpp>

#include <....>
```
->
```
// ginkgo main header
#include <...>

#include <ginkgo/ginkgo.hpp>
```
regroup does not move the header with the comments.
It should be checked for this pr to ensure that the comments are also moved to correct place.



> clang format does not take care of the "" for local and <> from global, so it may still destroy the regroup
I think it does. Any <> is priority 1-4, the <ginkgo/ headers are priority 5-6, and everything else, which includes "" is priority 7.

I mean if we put "ginkgo/..." it will be priority 7 because the script does not correct `<>/""`  usage now.

MarcelKoch:
Another argument to remove the script is that it's slow. The clang and reuse pre-commit check each take ~1s when applying to all files, but the format-header check takes more than a minute. The reason is most likely that the script can only be run sequentially, while the other checks can run in parallel.

pratikvn (approved):
LGTM!

upsj (approved):
LGTM!

reference/test/solver/direct.cpp
```diff
@@ -2,26 +2,21 @@
...
-
 #include <gtest/gtest.h>
 
-
 #include <ginkgo/core/base/exception.hpp>
 #include <ginkgo/core/base/executor.hpp>
 #include <ginkgo/core/base/math.hpp>
 #include <ginkgo/core/factorization/lu.hpp>
 #include <ginkgo/core/matrix/dense.hpp>
+#include <ginkgo/core/solver/direct.hpp>
```
yhmtsai, line 14: main header

reference/test/solver/idr_kernels.cpp
```diff
@@ -2,21 +2,17 @@
...
-#include <ginkgo/core/solver/idr.hpp>
-
-
 #include <gtest/gtest.h>
 
-
 #include <ginkgo/core/base/exception.hpp>
 #include <ginkgo/core/base/executor.hpp>
 #include <ginkgo/core/matrix/dense.hpp>
+#include <ginkgo/core/solver/idr.hpp>
```
yhmtsai, line 10: main header

reference/test/solver/multigrid_kernels.cpp
```diff
@@ -16,11 +12,11 @@
 #include <ginkgo/core/preconditioner/jacobi.hpp>
 #include <ginkgo/core/solver/cg.hpp>
 #include <ginkgo/core/solver/ir.hpp>
+#include <ginkgo/core/solver/multigrid.hpp>
```
yhmtsai, line 15: main header

reference/test/stop/combined.cpp
```diff
@@ -2,19 +2,17 @@
...
 // SPDX-License-Identifier: BSD-3-Clause
 
-#include <ginkgo/core/stop/combined.hpp>
-
-
 #include <chrono>
 #include <thread>
+
+#include <ginkgo/core/stop/combined.hpp>
 #if defined(_WIN32) || defined(__CYGWIN__)
```
yhmtsai, line 9:
```suggestion
#include <ginkgo/core/stop/combined.hpp>

#if defined(_WIN32) || defined(__CYGWIN__)
```

reference/test/stop/criterion_kernels.cpp
```diff
@@ -2,12 +2,9 @@
 //
 // SPDX-License-Identifier: BSD-3-Clause
 
-#include <ginkgo/core/stop/criterion.hpp>
-
-
 #include <gtest/gtest.h>
 
-
+#include <ginkgo/core/stop/criterion.hpp>
```
yhmtsai, line 7: main header

reference/test/stop/iteration.cpp
```diff
@@ -2,11 +2,10 @@
 //
 // SPDX-License-Identifier: BSD-3-Clause
 
-#include <ginkgo/core/stop/iteration.hpp>
-
-
 #include <gtest/gtest.h>
 
+#include <ginkgo/core/stop/iteration.hpp>
```
yhmtsai, line 7: main header

reference/test/stop/residual_norm_kernels.cpp
```diff
@@ -2,18 +2,13 @@
...
 #include <type_traits>
 
-
 #include <gtest/gtest-typed-test.h>
 #include <gtest/gtest.h>
 
-
 #include <ginkgo/core/base/math.hpp>
-
+#include <ginkgo/core/stop/residual_norm.hpp>
```
yhmtsai, line 11: main header

reference/test/stop/time.cpp
```diff
@@ -2,11 +2,10 @@
 //
 // SPDX-License-Identifier: BSD-3-Clause
 
-#include <ginkgo/core/stop/time.hpp>
-
-
 #include <chrono>
 #include <thread>
+
+#include <ginkgo/core/stop/time.hpp>
```
yhmtsai, line 8: main header

test/base/timer.cpp
```diff
@@ -2,15 +2,12 @@
...
-#include <ginkgo/core/base/timer.hpp>
-
-
 #include <map>
 #include <thread>
 
-
 #include <gtest/gtest.h>
 
+#include <ginkgo/core/base/timer.hpp>
```
yhmtsai, line 10: main header

test/matrix/csr_kernels2.cpp
```diff
@@ -2,19 +2,15 @@
...
 #include <stdexcept>
 
-
 #include <gtest/gtest.h>
 
-
 #include <ginkgo/core/base/exception.hpp>
 #include <ginkgo/core/base/executor.hpp>
 #include <ginkgo/core/matrix/coo.hpp>
+#include <ginkgo/core/matrix/csr.hpp>
```
yhmtsai, line 13: main header

yhmtsai (changes requested):
summary of my comments:
1. oneapi except for dpl should be in the same group as cu/hip...
2. sde_lib is also from papi. and maybe we can put the priority 4 and 5 together?
3. some main headers are missed or wrong due to no `""`.
4. because dpl in priority -2, we do not need the clang-format on/off to force it. It may apply to runtime but I am not sure the priority 2 is enough for solving the compiling issue.

cuda/test/base/array.cpp
```diff
@@ -2,15 +2,11 @@
 //
 // SPDX-License-Identifier: BSD-3-Clause
 
-#include <ginkgo/core/base/array.hpp>
-
-
 #include <gtest/gtest.h>
 
-
+#include <ginkgo/core/base/array.hpp>
```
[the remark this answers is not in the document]
MarcelKoch: This is not the main header. The conversation about main headers only makes sense for files which implement functions defined in a header file. This is not the case for the test files, so there is no main header here.
I think we had a different interpretation of main header previously, which was based on *\~vibes\~*. Now we just follow clangs definition.
So all the comments on the test files are unnecessary. Sorry for not making that clearer before.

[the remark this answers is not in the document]
yhmtsai: with this definition, the opposite part needs to change.
`cuda/test/matrix/coo_kernels.cpp` uses `core/matrix/coo_kernels.hpp` as the main header because they are in the mainHeaderSource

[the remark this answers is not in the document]
MarcelKoch: do you mean `cuda/matrix/coo_kernels.cpp`? Because there is no `cuda/test/matrix/coo_kernels.cpp`.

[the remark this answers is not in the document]
yhmtsai: sorry, I just randomly pick something might have issue.
something like this: `cuda/test/solver/lower_trs_kernels.cu` uses `core/solver/lower_trs_kernels.hpp`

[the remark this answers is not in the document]
MarcelKoch: I see. This poses an issue, because there doesn't seem to be a way to prevent clang from detecting `core/solver/lower_trs_kernels.hpp` as main header. I think the simplest choice is to consider these headers as the main header again.

Side note: this particular header is unnecessary. The test doesn't call any kernel directly.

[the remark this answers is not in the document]
MarcelKoch: Or we can just ignore it and live with these (small) inconsistencies. That means there will be some tests which have a main header and those that don't.

yhmtsai (approved):
I ignore main header inconsistentence for now.
the onedpl with disabling the clang-format is not changed and some wrong usage of header ""/<> is not resolved yet.

core/base/batch_multi_vector_kernels.hpp
```diff
@@ -6,13 +6,11 @@
...
 
 
-#include <ginkgo/core/base/batch_multi_vector.hpp>
-
+#include "ginkgo/core/base/batch_multi_vector.hpp"
```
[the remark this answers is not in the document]
MarcelKoch: you're right, I will update this (and the others) accordingly

MarcelKoch:
I've rebased it after the changes from #1616. It went pretty smoothly. I don't think there are additional changes that were not reviewed before, so I will go a head and merge it when the tests pass.

PR #1467, opened 2023-11-19
Add benchmark comparison script

upsj:
This script allows us to easily compare runtimes, storage and iteration counts between different benchmark runs on the same input.

TODO
- [x] allow listing outliers for larger benchmark runs
- [x] different output modes: csv, markdown table, json

### DIFF
--- a/benchmark/tools/compare.py
+++ b/benchmark/tools/compare.py
@@ -160,7 +160,7 @@ def compare_main(args: list):
 
     results = {}
 
-    for key in set(baseline_json.keys()).intersection(comparison_json.keys()):
+    for key in sorted_key_intersection(baseline_json, comparison_json):
         results[key] = compare(baseline_json[key], comparison_json[key])
 
     outliers = {}
--- a/benchmark/tools/compare.py
+++ b/benchmark/tools/compare.py
@@ -1,0 +1,135 @@
+#!/usr/bin/env python3
+import sys
+import json
+import argparse
+
+parser = argparse.ArgumentParser(description="Compare to Ginkgo benchmark outputs")
+parser.add_argument("--outlier-threshold")
+parser.add_argument("--output")
+parser.add_argument("baseline")
+parser.add_argument("comparison")
+args = parser.parse_args()
+keys = {"stencil", "size", "filename", "n", "r", "k", "m"}
+
+
+def key_to_str(key: tuple) -> str:
+    """Restore a JSON output from a key tuple"""
+    result = {}
+    for key_name, key_val in zip(keys, key):
+        if key_val is not None:
+            result[key_name] = key_val
+    return json.dumps(result)
+
+
+def parse_json_matrix(filename: str) -> dict:
+    """Parse a JSON file into a key -> test_case dict"""
+    parsed = json.load(open(filename))
+    result = {}
+    assert isinstance(parsed, list)
+    for case in parsed:
+        assert isinstance(case, dict)
+        assert not keys.isdisjoint(case.keys())
+        dict_key = tuple(case.get(key, None) for key in keys)
+        if dict_key in result.keys():
+            print(
+                "WARNING: Duplicate key {}".format(key_to_str(dict_key)),
+                file=sys.stderr,
+            )
+        result[dict_key] = case
+    return result
+
+
+def warn_on_inconsistent_keys(baseline: dict, comparison: dict, context: str):
+    """Print a warning message for non-matching keys between baseline/comparison using the given context string"""
+    baseline_only = set(baseline.keys()).difference(comparison.keys())
+    comparison_only = set(comparison.keys()).difference(baseline.keys())
+    for dict_key in baseline_only:
+        print(
+            "WARNING: Key {} found in baseline only in context {}".format(
+                key_to_str(dict_key), context
+            ),
+            file=sys.stderr,
+        )
+    for dict_key in comparison_only:
+        print(
+            "WARNING: Key {} found in comparison only in context {}".format(
+                key_to_str(dict_key), context
+            ),
+            file=sys.stderr,
+        )
+
+
+def ratio(baseline: int | float, comparison: int | float) -> float:
+    """Compares the ratio between baseline and comparison. For runtimes, this is the speedup."""
+    return baseline / comparison
+
+
+def compare_benchmark(baseline: dict, comparison: dict, context: str):
+    """Compares a handful of keys and component breakdowns recursively, writing them with a suffix to the output"""
+    comparison_keys = {"time", "storage", "iterations"}
+    suffix = ".ratio"
+    warn_on_inconsistent_keys(baseline, comparison, context)
+    result = {}
+    for key in baseline.keys():
+        sub_context = "{}.{}".format(context, key)
+        if key == "components":
+            assert isinstance(baseline[key], dict)
+            assert isinstance(comparison[key], dict)
+            warn_on_inconsistent_keys(baseline[key], comparison[key], sub_context)
+            result[key + suffix] = {
+                sub_key: ratio(baseline[key][sub_key], comparison[key][sub_key])
+                for sub_key in baseline[key]
+            }
+        elif isinstance(baseline[key], dict):
+            result[key] = compare_benchmark(baseline[key], comparison[key], sub_context)
+        elif key in comparison_keys:
+            result[key + suffix] = ratio(baseline[key], comparison[key])
+    return result
+
+
+def compare(baseline: dict, comparison: dict, context: str) -> dict:
+    """Compares a test case, keeping root-level values and recursing into benchmarks"""
+    warn_on_inconsistent_keys(baseline, comparison, context)
+    result = {}
+    for key in baseline.keys():
+        # we don't have lists on the test case root level
+        assert not isinstance(baseline[key], list)
+        if isinstance(baseline[key], dict):
+            benchmark_result = {}
+            warn_on_inconsistent_keys(
+                baseline[key], comparison[key], "{}.{}".format(context, key)
+            )
+            for benchmark_name in baseline[key].keys():
+                if isinstance(baseline[key][benchmark_name], dict):
+                    comparison_result = compare_benchmark(
+                        baseline[key][benchmark_name],
+                        comparison[key][benchmark_name],
+                        "{}.{}.{}".format(context, key, benchmark_name),
+                    )
+                    if len(comparison_result) > 0:
+                        benchmark_result[benchmark_name] = comparison_result
+            if len(benchmark_result) > 0:
+                result[key] = benchmark_result
+        else:
+            # everything that's not a dict should only depend on the key in the root level
+            if baseline[key] != comparison[key]:
+                print(
+                    "WARNING: Inconsistent value for {}: {} != {} in context {}".format(
+                        key, baseline[key], comparison[key], context
+                    ),
+                    file=sys.stderr,
+                )
+            result[key] = baseline[key]
+    return result
+
+
+baseline_json = parse_json_matrix(args.baseline)
+comparison_json = parse_json_matrix(args.comparison)
+warn_on_inconsistent_keys(baseline_json, comparison_json, "root")
+
+results = []
+
+for key in baseline_json.keys():
+    results.append(compare(baseline_json[key], comparison_json[key], key_to_str(key)))
+
+json.dump(results, sys.stdout, indent=4)
--- a/benchmark/tools/compare.py
+++ b/benchmark/tools/compare.py
@@ -249,4 +249,4 @@ def compare_main(args: list):
 
 
 if __name__ == "__main__":
-    compare(sys.argv)
+    compare_main(sys.argv)
--- a/benchmark/tools/compare.py
+++ b/benchmark/tools/compare.py
@@ -1,90 +1,65 @@
 #!/usr/bin/env python3
+# SPDX-FileCopyrightText: 2017-2023 The Ginkgo authors
+# SPDX-License-Identifier: BSD-3-Clause
 import sys
 import json
 import argparse
 import math
 import pandas as pd
+import tabulate  # for pandas markdown output
+from frozendict import frozendict
 
-parser = argparse.ArgumentParser(description="Compare to Ginkgo benchmark outputs")
-parser.add_argument(
-    "--outliers", action="store_true", help="List outliers from the results"
-)
-parser.add_argument(
-    "--outlier-threshold",
-    type=float,
-    default=10,
-    help="At what percentage of deviation (above or below) should outliers be reported",
-)
-parser.add_argument(
-    "--outlier-count",
-    type=int,
-    default=1000,
-    help="How many outliers should be reported per benchmark",
-)
-parser.add_argument("--output", choices=["json", "csv", "markdown"], default="json")
-parser.add_argument("baseline")
-parser.add_argument("comparison")
-args = parser.parse_args()
+
 keys = {"stencil", "size", "filename", "n", "r", "k", "m"}
 comparison_keys = {"time", "storage", "iterations"}
 suffix = ".ratio"
 
 
-def key_to_json(key: tuple) -> dict:
-    """Restore a JSON output from a key tuple"""
-    result = {}
-    for key_name, key_val in zip(keys, key):
-        if key_val is not None:
-            result[key_name] = key_val
-    return result
-
-
-def key_to_str(key: tuple) -> str:
-    """Restore a JSON output string from a key tuple"""
-    mapped = key_to_json(key)
-    if "filename" in mapped.keys():
-        return mapped["filename"]
-    if "stencil" in mapped.keys():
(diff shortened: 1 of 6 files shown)


benchmark/tools/compare.py
```diff
@@ -0,0 +1,261 @@
+#!/usr/bin/env python3
+import sys
+import json
+import argparse
+import math
+import pandas as pd
+
```
MarcelKoch, line 7: requires `tabulate` for markdown output.

benchmark/tools/compare.py
```diff
@@ -0,0 +1,261 @@
...
+                (case_key, value)
+                for case_key, value in benchmark_results
+                if is_outlier(value)
+            ],
+            key=lambda x: math.fabs(math.log(x[1])),
+            reverse=True,
+        )
+        outliers[benchmark_name] = outlier[: min(len(outlier), args.outlier_count)]
+
+if args.output == "json":
```
MarcelKoch, line 196: why is the json output not handled by pandas?

benchmark/tools/compare.py
```diff
@@ -0,0 +1,261 @@
...
+                for key, values in benchmarks.items()
+            ],
+            [],
+        ),
+        columns=columns,
+    )
+    if args.output == "csv":
+        table.to_csv(sys.stdout, index=False)
+    else:
+        table.to_markdown(sys.stdout, index=False)
```
MarcelKoch, line 236: These two have empty output if `outliers` is not set. That is inconsistent with the json output.

MarcelKoch (commented):
some quick first comments

benchmark/tools/compare.py
```diff
@@ -0,0 +1,261 @@
...
+    return result
+
+
+def key_to_str(key: tuple) -> str:
+    """Restore a JSON output string from a key tuple"""
+    mapped = key_to_json(key)
+    if "filename" in mapped.keys():
+        return mapped["filename"]
+    if "stencil" in mapped.keys():
+        return "stencil({}, {})".format(mapped["stencil"], mapped["size"])
```
greole, line 48: Which python version are we using. If python > 3.6 fstrings are more concise
```suggestion
        return f"stencil({mapped["stencil"]}, {mapped["size"]})"
```

benchmark/tools/compare.py
```diff
@@ -0,0 +1,261 @@
...
+    parsed = json.load(open(filename))
+    result = {}
+    assert isinstance(parsed, list)
+    for case in parsed:
+        assert isinstance(case, dict)
+        assert not keys.isdisjoint(case.keys())
+        dict_key = tuple(case.get(key, None) for key in keys)
+        if dict_key in result.keys():
+            print(
+                "WARNING: Duplicate key {}".format(key_to_str(dict_key)),
```
greole, line 63: For python > 3.6
```suggestion
                f"WARNING: Duplicate key {key_to_str(dict_key)}"
```

benchmark/tools/compare.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+def warn_on_inconsistent_keys(baseline: dict, comparison: dict, context: str):
+    """Print a warning message for non-matching keys between baseline/comparison using the given context string"""
+    baseline_only = set(baseline.keys()).difference(comparison.keys())
+    comparison_only = set(comparison.keys()).difference(baseline.keys())
+    for dict_key in baseline_only:
+        print(
+            "WARNING: Key {} found in baseline only in context {}".format(
+                key_to_str(dict_key), context
```
greole, line 77:
```suggestion
            f"WARNING: Key {key_to_str(dict_key)} found in baseline only in context {context}"
```

benchmark/tools/compare.py
```diff
@@ -0,0 +1,261 @@
...
+    """Compares the ratio between baseline and comparison. For runtimes, this is the speedup."""
+    return baseline / comparison
+
+
+def compare_benchmark(baseline: dict, comparison: dict, context: str):
+    """Compares a handful of keys and component breakdowns recursively, writing them with a suffix to the output"""
+    warn_on_inconsistent_keys(baseline, comparison, context)
+    result = {}
+    for key in baseline.keys():
+        sub_context = "{}.{}".format(context, key)
```
greole, line 100:
```suggestion
        sub_context = f"{context}.{key}"
```

benchmark/tools/compare.py
```diff
@@ -0,0 +1,261 @@
...
+        print(
+            "WARNING: Key {} found in baseline only in context {}".format(
+                key_to_str(dict_key), context
+            ),
+            file=sys.stderr,
+        )
+    for dict_key in comparison_only:
+        print(
+            "WARNING: Key {} found in comparison only in context {}".format(
+                key_to_str(dict_key), context
```
greole, line 84:
```suggestion
            f"WARNING: Key {key_to_str(dict_key)} found in comparison only in context {context}"
```

benchmark/tools/compare.py
```diff
@@ -0,0 +1,261 @@
...
+            benchmark_result = {}
+            warn_on_inconsistent_keys(
+                baseline[key], comparison[key], "{}.{}".format(context, key)
+            )
+            for benchmark_name in baseline[key].keys():
+                if isinstance(baseline[key][benchmark_name], dict):
+                    comparison_result = compare_benchmark(
+                        baseline[key][benchmark_name],
+                        comparison[key][benchmark_name],
+                        "{}.{}.{}".format(context, key, benchmark_name),
```
greole, line 133:
```suggestion
                       f"{context}.{key}.{benchmark_name}",
```

benchmark/tools/compare.py
```diff
@@ -0,0 +1,261 @@
...
+    """Compares a test case, keeping root-level values and recursing into benchmarks"""
+    warn_on_inconsistent_keys(baseline, comparison, context)
+    result = {}
+    for key in baseline.keys():
+        # we don't have lists on the test case root level
+        assert not isinstance(baseline[key], list)
+        if isinstance(baseline[key], dict):
+            benchmark_result = {}
+            warn_on_inconsistent_keys(
+                baseline[key], comparison[key], "{}.{}".format(context, key)
```
greole, line 126:
```suggestion
                baseline[key], comparison[key], f"{context}.{key}".
```

greole (commented):
Some quick fstring suggestions. F-strings are great, helps to get rid of most str.format()

benchmark/tools/compare.py
```diff
@@ -0,0 +1,261 @@
...
+                    "WARNING: Inconsistent value for {}: {} != {} in context {}".format(
+                        key, baseline[key], comparison[key], context
+                    ),
+                    file=sys.stderr,
+                )
+            result[key] = baseline[key]
+    return result
+
+
+def extract_benchmark_results(
```
MarcelKoch, line 152: maybe make that an out-of-place operation? No other (except warn*) is like that.

benchmark/tools/compare.py
```diff
@@ -0,0 +1,261 @@
...
+            file=sys.stderr,
+        )
+
+
+def ratio(baseline: int | float, comparison: int | float) -> float:
+    """Compares the ratio between baseline and comparison. For runtimes, this is the speedup."""
+    return baseline / comparison
+
+
+def compare_benchmark(baseline: dict, comparison: dict, context: str):
```
MarcelKoch, line 95: nit:
```suggestion
def compare_benchmark(baseline: dict, comparison: dict, context: str) -> dict:
```

benchmark/tools/compare.py
```diff
@@ -0,0 +1,261 @@
...
+results = {}
+
+for key in baseline_json.keys():
+    results[key] = compare(baseline_json[key], comparison_json[key], key_to_str(key))
+
+outliers = {}
+if args.outliers:
+    benchmarks = {}
+    for key, value in results.items():
+        extract_benchmark_results(value, benchmarks, key, None)
```
MarcelKoch, line 183: This also needs to happen outside the `if`, otherwise the csv/markdown output is not correct.

MarcelKoch (changes requested):
My biggest issue is storing the keys to the benchmarks as tuples. This throws away too much information that is then still implicitly relied on.
Also, the output for the different formats should be unified, or perhaps it is best to only output json.

benchmark/tools/compare.py
upsj: right, I'm still used to older python styles ;) I'll update those

benchmark/tools/compare.py
```diff
@@ -0,0 +1,261 @@
...
+    "--outlier-count",
+    type=int,
+    default=1000,
+    help="How many outliers should be reported per benchmark",
+)
+parser.add_argument("--output", choices=["json", "csv", "markdown"], default="json")
+parser.add_argument("baseline")
+parser.add_argument("comparison")
+args = parser.parse_args()
+keys = {"stencil", "size", "filename", "n", "r", "k", "m"}
```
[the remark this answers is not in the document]
upsj: here I am relying on an unchanging set providing consistent ordering, but that's maybe stretching the standard requirements too much. I wanted to avoid relying on external packages as much as possible, but I guess with `pandas` and `tabulate`, that ship has already sailed ;)

benchmark/tools/compare.py
```diff
@@ -0,0 +1,261 @@
...
+
+def parse_json_matrix(filename: str) -> dict:
+    """Parse a JSON file into a key -> test_case dict"""
+    parsed = json.load(open(filename))
+    result = {}
+    assert isinstance(parsed, list)
+    for case in parsed:
+        assert isinstance(case, dict)
+        assert not keys.isdisjoint(case.keys())
+        dict_key = tuple(case.get(key, None) for key in keys)
```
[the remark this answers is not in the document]
upsj: makes sense

benchmark/tools/compare.py
```diff
@@ -0,0 +1,261 @@
...
+                        comparison[key][benchmark_name],
+                        "{}.{}.{}".format(context, key, benchmark_name),
+                    )
+                    if len(comparison_result) > 0:
+                        benchmark_result[benchmark_name] = comparison_result
+            if len(benchmark_result) > 0:
+                result[key] = benchmark_result
+        else:
+            # everything that's not a dict should only depend on the key in the root level
+            if baseline[key] != comparison[key]:
```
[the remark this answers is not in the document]
upsj: Not necessarily, these can also be things like `rows`, `columns` and `nonzeros`, which should be the same but don't need to be part of the keys (and I would like to keep them out of there, because inconsistencies in these values point to mismatching inputs)

benchmark/tools/compare.py
```diff
@@ -0,0 +1,261 @@
...
+def key_to_json(key: tuple) -> dict:
+    """Restore a JSON output from a key tuple"""
+    result = {}
+    for key_name, key_val in zip(keys, key):
+        if key_val is not None:
+            result[key_name] = key_val
+    return result
+
+
+def key_to_str(key: tuple) -> str:
```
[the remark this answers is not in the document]
upsj: that's a nice suggestion, I like it.

benchmark/tools/compare.py
upsj: Is there a nice way to implement joining on matching keys? Here we need to use `list + list` on the values

benchmark/tools/compare.py
upsj: Right, I only checked with outliers.

benchmark/tools/compare.py
upsj: I have two output styles in mind: The JSON output can handle arbitrary nesting, the tabular style flattens everything. I want to preserve the nested structure for the JSON output, which is hard to represent in Pandas, since there is no `orient` matching this structure well.

benchmark/tools/compare.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+baseline_json = parse_json_matrix(args.baseline)
+comparison_json = parse_json_matrix(args.comparison)
+warn_on_inconsistent_keys(baseline_json, comparison_json, "root")
+
+results = {}
+
+for key in baseline_json.keys():
+    results[key] = compare(baseline_json[key], comparison_json[key], key_to_str(key))
```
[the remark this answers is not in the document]
upsj: done, right now I was only printing warnings, but still proceeding with the traversal

benchmark/tools/compare.py
```diff
@@ -0,0 +1,261 @@
...
+    return result
+
+
+def warn_on_inconsistent_keys(baseline: dict, comparison: dict, context: str):
+    """Print a warning message for non-matching keys between baseline/comparison using the given context string"""
+    baseline_only = set(baseline.keys()).difference(comparison.keys())
+    comparison_only = set(comparison.keys()).difference(baseline.keys())
+    for dict_key in baseline_only:
+        print(
+            "WARNING: Key {} found in baseline only in context {}".format(
```
greole, line 76: Instead of crafting warning messages manually and printing it out to sys.stderr, you could use pythons logging module https://docs.python.org/3/library/logging.html (or check warnings  https://docs.python.org/3/library/warnings.html together)

greole (commented):
Some more minor things

benchmark/tools/compare.py
```diff
@@ -0,0 +1,261 @@
...
+                    ),
+                    file=sys.stderr,
+                )
+            result[key] = baseline[key]
+    return result
+
+
+def extract_benchmark_results(
+    input: dict, benchmarks: dict, case_key: tuple, context: str | None
+):
```
greole, line 154: nit 
```suggestion
): -> None
```

benchmark/tools/compare.py
MarcelKoch: using `defaultdict(list)` instead of plain `dict` should allow you to just append to any key.

benchmark/tools/compare.py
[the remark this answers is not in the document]
MarcelKoch: another possibility would be to use sorted `tuple((k,v), ...)` for the keys

benchmark/tools/compare.py
```diff
@@ -0,0 +1,261 @@
+#!/usr/bin/env python3
```
greole, line 1:
```suggestion
#!/usr/bin/env python3
# SPDX-FileCopyrightText: 2017-2023 The Ginkgo authors
# SPDX-License-Identifier: BSD-3-Clause
```

benchmark/tools/compare.py
```diff
@@ -0,0 +1,261 @@
...
+    default=10,
+    help="At what percentage of deviation (above or below) should outliers be reported",
+)
+parser.add_argument(
+    "--outlier-count",
+    type=int,
+    default=1000,
+    help="How many outliers should be reported per benchmark",
+)
+parser.add_argument("--output", choices=["json", "csv", "markdown"], default="json")
```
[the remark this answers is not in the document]
upsj: I would like to also support CSV for follow-up analysis

benchmark/tools/compare.py
upsj: I'd like to test everything using capsys and pytest, is it easy to capture warnings?

MarcelKoch (approved):
LGTM. Are the test run automatically or manually?

upsj:
@MarcelKoch currently only manually, I wasn't sure about the directory layout and common practices w.r.t. available packages, so I wanted to collect some feedback first.

yhmtsai:
I do not look into the detail but I feel it is more suitable in ginkgo-data repo

benchmark/tools/compare.py
```diff
@@ -0,0 +1,252 @@
...
+            benchmark_name = benchmark_name[: -len(suffix)]
+            if benchmark_name not in benchmarks.keys():
+                benchmarks[benchmark_name] = []
+            benchmarks[benchmark_name].append((case_key, value))
+        elif isinstance(value, dict):
+            extract_benchmark_results(value, benchmarks, case_key, benchmark_name)
+
+
+def is_outlier(value: float, args) -> bool:
+    """returns true iff the value exceeds 1.0 above the outlier threshold"""
```
greole, line 131: Iff that is on purpose keep it, otherwise change it.

upsj: iff = if and only if

upsj:
@yhmtsai I am building this for development purposes, modeled after what [nvbench_compare.py](https://github.com/NVIDIA/nvbench/blob/main/scripts/nvbench_compare.py) provides for nvbench, I don't think this is specific to ginkgo-data.

yhmtsai:
It somehow post-processes the data generated by benchmark, right? Its purpose is like the script for aggregating the json or plotting spmv performance comparison in ginkgo-data.

upsj:
An argument for putting this into `ginkgo` directly is that it is likely to be used directly in conjunction with the benchmarks, while it is unlikely you will need to check out `ginkgo-data` when optimizing something, because it is more stable to run the baseline manually instead of relying on archived values.

greole (approved):
LGTM!

PR #1469, opened 2023-11-24
fixup checking out branch for CI

MarcelKoch:
This PR fixes a mistake I made merging #1461. The issue was that the `bot_pr_format_base.sh` script tried to create the same branch twice.


pratikvn (approved):

MarcelKoch:
Note that the `check-format` job still fails, because it uses the script from the current develop.

greole (approved):
LGTM

PR #1471, opened 2023-11-24
Remove old variable in CMake `gko_rename_cache`

upsj:
I wanted to propose this change to get rid of warning messages on my long-standing repositories, be creative in thinking of potential negative side-effects this might have!

This removes the old variable from cache instead of leaving it around, which leads to the error message only being shown once instead of continuously. Removing variables from cache requires modifying the `CMakeCache.txt`, which isn't something we need to ask of them.

### DIFF
--- a/cmake/rename.cmake
+++ b/cmake/rename.cmake
@@ -5,7 +5,8 @@ macro(gko_rename_cache deprecated actual type doc_string)
             message("actual ${actual} and deprecated ${deprecated}")
             if("${${actual}}" STREQUAL "${${deprecated}}")
                 # They are the same, so only throw warning
-                message(WARNING "${deprecated} was deprecated, please only use ${actual} instead.")
+                message(WARNING "${deprecated} was deprecated, please only use ${actual} instead. We removed the old variable.")
+                unset(${deprecated} CACHE)
             else()
                 # They are different
                 message(FATAL_ERROR "Both ${deprecated} and ${actual} were specified differently, please only use ${actual} instead.")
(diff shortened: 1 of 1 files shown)


cmake/rename.cmake
```diff
@@ -13,8 +13,9 @@ macro(gko_rename_cache deprecated actual type doc_string)
         else()
             # Only set `deprecated`, move it to `actual`.
             message(WARNING "${deprecated} was deprecated, please use ${actual} instead.  "
-                "We copy ${${deprecated}} to ${actual}")
+                "We copy ${${deprecated}} to ${actual} and remove the old variable")
```
MarcelKoch, line 16: nit:
```suggestion
                "We copied ${${deprecated}} to ${actual} and removed the old variable")
```

cmake/rename.cmake
```diff
@@ -13,8 +13,9 @@ macro(gko_rename_cache deprecated actual type doc_string)
...
             set(${actual} ${${deprecated}} CACHE ${type} "${doc_string}")
+            unset(${${deprecated}} CACHE)
```
MarcelKoch, line 18: should this be outside the `if(DEFINED ${actual})`? After both branches, there is still the deprecated variable and the new one.

MarcelKoch (approved):

cmake/rename.cmake
upsj: That's probably okay to do, I just didn't want to touch cases where the function was already run previously.

upsj: I added `unset` in to places instead, to make the control-flow involving `FATAL_ERROR` more clear

yhmtsai (commented):
As we discussed in #1350 , I do not have strong opinions on either one.
I do not get why it introduces the issue. It only throws the warning and should not block any operation. user can unset the variable from cmake command

yhmtsai (approved):
I can not comment on whether user do not care or not on some hidden changes.

upsj:
As a developer, I definitely care :nerd_face: I think deprecation warnings make sense if somebody is actively doing something deprecated, but if it's just a variable they still have in their CMakeCache.txt, we can fix it for them.

PR #1472, opened 2023-11-25
Replace `volatile` by proper memory ordering in HIP

upsj:
`rocm-clang` supports the GCC `__atomic` intrinsics, which we can use to implement the atomic operations instead of using `volatile` and memory fences.

This is a follow-up to #1344, so I requested reviews from the same reviewers


dev_tools/scripts/generate_cuda_memory_ptx.py
```diff
@@ -28,45 +28,52 @@ class type_desc:
...
+        ptr_constraint="r",
+    ),
+    space(
+        ptx_space_suffix="",
+        ptx_scope_suffix=".gpu",
+        fn_suffix="",
+        ptr_expr="{ptr}",
+        ptr_constraint="l",
+    ),
+]
```
yhmtsai, line 45: I prefer the old format in my opinion

yhmtsai (commented):
do you have the documentation about rocm-clang support gcc atomic on GPU?
Is there only one compiler rocm-clang from hipcc on AMD GPU?

dev_tools/scripts/generate_cuda_memory_ptx.py
upsj: Right, I think I accidentally formatted with `black` instead of `autopep8`. I'll revert the changes

upsj:
Relevant: https://github.com/ROCm-Developer-Tools/llvm-project/blob/6b9c186b2d4c8bd315034a9655a28d32bcf745ab/clang/test/SemaCUDA/atomic-ops.cu#L6

hip/components/memory.hip.hpp
```diff
@@ -20,71 +21,203 @@ namespace kernels {
...
+template <int memorder, int scope, typename ValueType>
+__device__ __forceinline__ ValueType load_generic(const ValueType* ptr)
+{
+    using atomic_type = typename gcc_atomic_intrinsic_type_map<ValueType>::type;
+    static_assert(sizeof(atomic_type) == sizeof(ValueType), "invalid map");
+    static_assert(alignof(atomic_type) == sizeof(ValueType), "invalid map");
+    auto cast_value = HIP_ATOMIC_LOAD(reinterpret_cast<const atomic_type*>(ptr),
+                                      memorder, scope);
+    ValueType result{};
+    memcpy(&result, &cast_value, sizeof(ValueType));
```
yhmtsai, line 86: `reinterpret_cast` the result to the `atomic_type`, too?

hip/components/memory.hip.hpp
```diff
@@ -20,71 +21,203 @@ namespace kernels {
...
+__device__ __forceinline__ void store_release_shared(ValueType* ptr,
+                                                     ValueType value)
+{
+    store_generic<__ATOMIC_RELEASE, HIP_SCOPE_THREADBLOCK>(ptr, value);
 }
 
 
 template <typename ValueType>
-__device__ __forceinline__ ValueType load_relaxed_shared(const ValueType* ptr)
+__device__ __forceinline__ thrust::complex<ValueType> load_relaxed(
```
yhmtsai, line 171: maybe only for thrust::complex<double>?
thrust::complex<float> can be done with int64

hip/components/memory.hip.hpp
```diff
@@ -6,6 +6,7 @@
 #define GKO_HIP_COMPONENTS_MEMORY_HIP_HPP_
 
 
+#include <cstring>
```
yhmtsai, line 9: is it used?

yhmtsai (commented):
For the code self, LGTM.
I am still not sure whether gcc intrinsics work on GPU address. compiling to libstdc++ only mean for CPU side not GPU, right?

hip/components/memory.hip.hpp
```diff
@@ -20,71 +21,203 @@ namespace kernels {
...
+    static_assert(alignof(atomic_type) == sizeof(ValueType), "invalid map");
+    atomic_type cast_value{};
+    memcpy(&cast_value, &value, sizeof(ValueType));
+    HIP_ATOMIC_STORE(reinterpret_cast<atomic_type*>(ptr), cast_value, memorder,
+                     scope);
+}
+
+
+template <typename ValueType,
+          gcc_atomic_intrinsic_type_map<ValueType>* = nullptr>
```
MarcelKoch, line 105: does this need an `enable_if`? Or what is the purpose of the second parameter otherwise?

hip/components/memory.hip.hpp
MarcelKoch, line 86: should it be `std::memcpy`? Or does this use the C version? TBH, I'm not sure how this is then handled e.g. with the issue Mike pointed out.

cuda/components/memory.cuh
```diff
@@ -15,761 +15,12 @@
...
-    asm volatile("st.volatile.v2.f64 [%0], {%1, %2};" ::"l"(ptr),
-                 "d"(real_result), "d"(imag_result)
-                 : "memory");
-#else
-    asm volatile("st.relaxed.gpu.v2.f64 [%0], {%1, %2};" ::"l"(ptr),
-                 "d"(real_result), "d"(imag_result)
-                 : "memory");
-#endif
-}
+#include "common/cuda_hip/components/memory.nvidia.hpp.inc"
```
MarcelKoch, line 23: I would also be fine with putting the header in `cuda/...`. Including that one from hip would be reasonable for me.

MarcelKoch (changes requested):
I think this needs quite a bit of documentation, since this relies on functionality that is not in the official documentation. It needs at least links to where you found the`__hip_atomic_load|store`, and the different scopes. If the references themselves don't have much documentation, then also properly documenting them would be necessary.

hip/components/memory.hip.hpp
upsj: `memcpy` uses `void*`, so it doesn't matter

hip/components/memory.hip.hpp
upsj: `memcpy` is defined in string.h, not sure if this is actually necessary though

hip/components/memory.hip.hpp
upsj: this is necessary to avoid compilation failures (not substitution failures) for thrust::complex inputs

hip/components/memory.hip.hpp
upsj: probably no difference between them, I'll use `std::`

hip/components/memory.hip.hpp
upsj: I don't follow, we only use thrust::complex with float value types

cuda/components/memory.cuh
upsj: it's only an incomplete header, the full header has namespaces around, so I believe this is cleaner, since it uses `common` for incomplete files

upsj:
@yhmtsai The tests run fine, and this was a suggestion by an AMD engineer, so I'm confident we can use them. If the intrinsics weren't supported, it would fail to compile instead.

hip/components/memory.hip.hpp
MarcelKoch: I don't see how this is necessary. I would expect that the complex overloads are going to be picked anyway because they are more specialized (e.g. https://godbolt.org/z/vxab9fxrz). Also, the second template parameter can also be instantiated for std::complex, so I still don't understand what it's doing there.

hip/components/memory.hip.hpp
yhmtsai: good to hear they give the same codegen in the end. Thanks for trying both

cuda/components/memory.cuh
yhmtsai: I would also prefer staying in the common because it is used from both

MarcelKoch (approved):

PR #1473, opened 2023-11-27
Scripts as precommit hook

MarcelKoch:
This PR make the pre-commit hook call the scripts `format_header.sh` and `ginkgo_update_header.sh`. It was necessary to modify `format_header.sh` to accept multiple files.

Additionally, the `.pre-commit-config.yaml` is now correctly staged before `pre-commit` is run.


upsj:
This means #1466 isn't sufficient and should be closed? I'd be really happy if we could make that work without manual scripts

MarcelKoch:
@upsj #1466 will be a successor to this. I just wanted to make the CI work again, with minimal changes. I'm still planning to remove the `format_header.sh` script.

upsj:
nit: you should change the API token back to the @ginkgo-bot token we have stored as a secret

upsj:
I think you could also use this to remove `rebase!` and `format-rebase!`

upsj (approved):
LGTM! Can you confirm the only changes you did to format_header.sh are the outer for-loop and replacing `$1`  by `${current_file}`?

MarcelKoch:
> LGTM! Can you confirm the only changes you did to format_header.sh are the outer for-loop and replacing `$1` by `${current_file}`?

I tested it locally with multiple inputs and it worked as expected. But there is no way to show this, except putting this in my fork.

upsj:
@MarcelKoch I just meant that the diff makes it a bit unclear whether other things changed except for the indentation. But I see that that's not the case

pratikvn (approved):

PR #1474, opened 2023-11-27
Make dim constructor explicit

upsj:
To avoid things like `dim<2>{1, 2} == 3` working accidentally, we need to make the single-parameter constructor `explicit`.


MarcelKoch (approved):
LGTM.
Might this be considered interface breaking? I wouldn't hold it up, just curious.

pratikvn (approved):
LGTM! This is technically interface breaking, but as we have previously broken interface for wrong behaviour, it should be okay, I think.

upsj:
@MarcelKoch Yes, that was in https://github.com/ginkgo-project/ginkgo/pull/650

Is this interface-breaking?

PR #1476, opened 2023-11-28
Copy device_matrix_data into arrays

MarcelKoch:
This PR changes the behavior for the `read(const device_matrix_data&)` overload for `Coo` and `Csr`. 

Currently, the `const &` read overload copies the input device_matrix_data and moves it into the `&&` overload. The `&&` overload will cause the internal arrays of the matrix to use the arrays of the device_matrix_data. This means that if the matrix was created from views, the matrix after the read will have owning arrays. 

For the `&&` overload this is justified, since here the matrix takes ownership of the input data. But for the `const &` it's unexpected, since no ownership is transferred. So, this PR makes the `const &` copy the input device matrix data directly into the matrix's arrays. The ownership property of the internal arrays will thus be preserved.

~**Note:** The behavior of the `&&` overload is a bit inconsistent across our matrix types. For `Coo` and `Csr` the `&&` overload will result in the matrix arrays having ownership. But for all other matrix types, the ownership property is not changed. Maybe we should consolidate that.~ I was under the impression that device_matrix_data always owns its arrays. But that is not the case. The device_matrix_data can also be constructed from array, which might be views.

### DIFF
--- a/core/matrix/coo.cpp
+++ b/core/matrix/coo.cpp
@@ -19,6 +19,7 @@
 #include <ginkgo/core/matrix/dense.hpp>
 
 
+#include "core/base/device_matrix_data_kernels.hpp"
 #include "core/components/absolute_array_kernels.hpp"
 #include "core/components/fill_array_kernels.hpp"
 #include "core/components/format_conversion_kernels.hpp"
@@ -43,6 +44,7 @@ GKO_REGISTER_OPERATION(inplace_absolute_array,
                        components::inplace_absolute_array);
 GKO_REGISTER_OPERATION(outplace_absolute_array,
                        components::outplace_absolute_array);
+GKO_REGISTER_OPERATION(aos_to_soa, components::aos_to_soa);
 
 
 }  // anonymous namespace
@@ -180,15 +182,36 @@ void Coo<ValueType, IndexType>::resize(dim<2> new_size, size_type nnz)
 template <typename ValueType, typename IndexType>
 void Coo<ValueType, IndexType>::read(const mat_data& data)
 {
-    this->read(device_mat_data::create_from_host(this->get_executor(), data));
+    auto size = data.size;
+    auto exec = this->get_executor();
+    this->set_size(size);
+    row_idxs_.resize_and_reset(data.nonzeros.size());
+    col_idxs_.resize_and_reset(data.nonzeros.size());
+    values_.resize_and_reset(data.nonzeros.size());
+    device_mat_data view{exec, size, row_idxs_.as_view(), col_idxs_.as_view(),
+                         values_.as_view()};
+    const auto host_data =
+        make_array_view(exec->get_master(), data.nonzeros.size(),
+                        const_cast<matrix_data_entry<ValueType, IndexType>*>(
+                            data.nonzeros.data()));
+    exec->run(
+        coo::make_aos_to_soa(*make_temporary_clone(exec, &host_data), view));
 }
 
 
 template <typename ValueType, typename IndexType>
 void Coo<ValueType, IndexType>::read(const device_mat_data& data)
 {
-    // make a copy, read the data in
-    this->read(device_mat_data{this->get_executor(), data});
+    this->set_size(data.get_size());
+    this->values_ = make_const_array_view(data.get_executor(),
+                                          data.get_num_stored_elements(),
+                                          data.get_const_values());
+    this->col_idxs_ = make_const_array_view(data.get_executor(),
+                                            data.get_num_stored_elements(),
+                                            data.get_const_col_idxs());
+    this->row_idxs_ = make_const_array_view(data.get_executor(),
+                                            data.get_num_stored_elements(),
+                                            data.get_const_row_idxs());
 }
 
 
--- a/core/matrix/coo.cpp
+++ b/core/matrix/coo.cpp
@@ -185,11 +185,11 @@ void Coo<ValueType, IndexType>::read(const mat_data& data)
     auto size = data.size;
     auto exec = this->get_executor();
     this->set_size(size);
-    row_idxs_.resize_and_reset(data.nonzeros.size());
-    col_idxs_.resize_and_reset(data.nonzeros.size());
-    values_.resize_and_reset(data.nonzeros.size());
-    device_mat_data view{exec, size, row_idxs_.as_view(), col_idxs_.as_view(),
-                         values_.as_view()};
+    this->row_idxs_.resize_and_reset(data.nonzeros.size());
+    this->col_idxs_.resize_and_reset(data.nonzeros.size());
+    this->values_.resize_and_reset(data.nonzeros.size());
+    device_mat_data view{exec, size, this->row_idxs_.as_view(),
+                         this->col_idxs_.as_view(), this->values_.as_view()};
     const auto host_data =
         make_array_view(exec->get_master(), data.nonzeros.size(),
                         const_cast<matrix_data_entry<ValueType, IndexType>*>(
@@ -203,6 +203,9 @@ template <typename ValueType, typename IndexType>
 void Coo<ValueType, IndexType>::read(const device_mat_data& data)
 {
     this->set_size(data.get_size());
+    // copy the arrays from device matrix data into the arrays of
+    // this. Compared to the read(device_mat_data&&) version, the internal
+    // arrays keep their current ownership status
     this->values_ = make_const_array_view(data.get_executor(),
                                           data.get_num_stored_elements(),
                                           data.get_const_values());
--- a/core/matrix/csr.cpp
+++ b/core/matrix/csr.cpp
@@ -426,12 +426,16 @@ void Csr<ValueType, IndexType>::read(const mat_data& data)
 {
     auto size = data.size;
     auto exec = this->get_executor();
-    row_ptrs_.resize_and_reset(size[0] + 1);
-    col_idxs_.resize_and_reset(data.nonzeros.size());
-    values_.resize_and_reset(data.nonzeros.size());
+    this->set_size(size);
+    this->row_ptrs_.resize_and_reset(size[0] + 1);
+    this->col_idxs_.resize_and_reset(data.nonzeros.size());
+    this->values_.resize_and_reset(data.nonzeros.size());
+    // the device matrix data contains views on the column indices
+    // and values array of this matrix, and an owning array for the
+    // row indices (which doesn't exist in this matrix)
     device_mat_data view{exec, size,
                          array<IndexType>{exec, data.nonzeros.size()},
-                         col_idxs_.as_view(), values_.as_view()};
+                         this->col_idxs_.as_view(), this->values_.as_view()};
     const auto host_data =
         make_array_view(exec->get_master(), data.nonzeros.size(),
                         const_cast<matrix_data_entry<ValueType, IndexType>*>(
@@ -440,7 +444,7 @@ void Csr<ValueType, IndexType>::read(const mat_data& data)
         csr::make_aos_to_soa(*make_temporary_clone(exec, &host_data), view));
     exec->run(csr::make_convert_idxs_to_ptrs(view.get_const_row_idxs(),
                                              view.get_num_stored_elements(),
-                                             size[0], get_row_ptrs()));
+                                             size[0], this->get_row_ptrs()));
     this->make_srow();
 }
 
@@ -452,6 +456,9 @@ void Csr<ValueType, IndexType>::read(const device_mat_data& data)
     auto exec = this->get_executor();
     this->row_ptrs_.resize_and_reset(size[0] + 1);
     this->set_size(size);
+    // copy the column indices and values array from the device matrix data
+    // into this. Compared to the read(device_mat_data&&) version, the internal
+    // arrays keep their current ownership status.
     this->values_ = make_const_array_view(data.get_executor(),
                                           data.get_num_stored_elements(),
                                           data.get_const_values());
@@ -476,7 +483,7 @@ void Csr<ValueType, IndexType>::read(device_mat_data&& data)
     auto size = data.get_size();
     auto exec = this->get_executor();
     auto arrays = data.empty_out();
-    this->row_ptrs_ = array<IndexType>{exec, size[0] + 1};
+    this->row_ptrs_.resize_and_reset(size[0] + 1);
     this->set_size(size);
     this->values_ = std::move(arrays.values);
     this->col_idxs_ = std::move(arrays.col_idxs);
--- a/core/matrix/csr.cpp
+++ b/core/matrix/csr.cpp
@@ -95,6 +95,7 @@ GKO_REGISTER_OPERATION(inv_scale, csr::inv_scale);
 GKO_REGISTER_OPERATION(add_scaled_identity, csr::add_scaled_identity);
 GKO_REGISTER_OPERATION(check_diagonal_entries,
                        csr::check_diagonal_entries_exist);
+GKO_REGISTER_OPERATION(aos_to_soa, components::aos_to_soa);
 
 
 }  // anonymous namespace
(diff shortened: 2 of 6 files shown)


MarcelKoch:
This fixes an issue in #1450

core/matrix/csr.cpp
```diff
@@ -441,7 +456,7 @@ void Csr<ValueType, IndexType>::read(device_mat_data&& data)
     auto size = data.get_size();
     auto exec = this->get_executor();
     auto arrays = data.empty_out();
-    this->row_ptrs_.resize_and_reset(size[0] + 1);
+    this->row_ptrs_ = array<IndexType>{exec, size[0] + 1};
```
pratikvn, line 459: resize and reset is the better option here as it does not resize if the array was of correct size.

core/matrix/coo.cpp
```diff
@@ -187,8 +187,13 @@ void Coo<ValueType, IndexType>::read(const mat_data& data)
...
 {
-    // make a copy, read the data in
-    this->read(device_mat_data{this->get_executor(), data});
+    this->set_size(data.get_size());
+    this->values_ = make_const_array_view(
+        data.get_executor(), data.get_num_elems(), data.get_const_values());
+    this->col_idxs_ = make_const_array_view(
+        data.get_executor(), data.get_num_elems(), data.get_const_col_idxs());
+    this->row_idxs_ = make_const_array_view(
+        data.get_executor(), data.get_num_elems(), data.get_const_row_idxs());
```
pratikvn, line 196: I think just simple copies should also work  ?
```suggestion
    this->values_ = data.arrays.values;
    this->col_idxs_ = data.arrays.col_idxs;
    this->row_idxs_ = data.arrays.row_idxs;
```

upsj: `arrays` isn't a member of data, and the internal arrays are not accessible to `Csr`, so this needs to stay. Though it is a bit confusing, since assigning a non-const view to an array causes it to be reassigned, while this causes a copy

core/matrix/csr.cpp
upsj: If we want `read` to never turn a view into an owning object, we should keep `resize_and_reset`, otherwise this seems fine? It's definitely a slight pessimization performance-wise.

include/ginkgo/core/base/array.hpp
```diff
@@ -547,6 +547,61 @@ class array {
...
+        }
+        if (other.get_executor() == nullptr) {
+            this->clear();
+            return *this;
+        }
+
+        if (this->is_owning()) {
+            this->resize_and_reset(other.get_num_elems());
+        } else {
+            GKO_ENSURE_COMPATIBLE_BOUNDS(other.get_num_elems(),
```
upsj, line 585: This should check ==, not <=

include/ginkgo/core/base/array.hpp
```diff
@@ -547,6 +547,61 @@ class array {
...
+     *
+     * @param other  the const_array_view to copy from
+     * @tparam OtherValueType  the value type of `other`
+     *
+     * @return this
+     */
+    template <typename OtherValueType>
+    std::enable_if_t<std::is_convertible<OtherValueType, ValueType>::value,
+                     array>&
+    operator=(const detail::const_array_view<OtherValueType>& other)
```
upsj, line 571: not sure I like this - I think we should be using `resize_and_reset` and `exec->copy_from` explicitly, because it mixes the behavior between assigning a view and a const view.

upsj (changes requested):

core/matrix/csr.cpp
MarcelKoch: This is changed to make `row_ptrs` owning. Since `col_idxs` and `values` will be owning, I thought this would be more consistent. It has worse performance if the same matrix is read multiple times into. For now, I don't think this is urgent to fix.

include/ginkgo/core/base/array.hpp
MarcelKoch: Should the whole macro change, or just here?

include/ginkgo/core/base/array.hpp
MarcelKoch: I don't think it's mixing behavior. In case of a non-const view, `a = view_b` also does a copy. Only the `a = std::move(view_b)` is different.

upsj: That's what you are doing in your `read(const device_matrix_data)` function - we have rvalue overloads for regular arrays, but not for const views.

include/ginkgo/core/base/array.hpp
upsj: Considering the small number of places we use it, probably the former

core/matrix/csr.cpp
MarcelKoch: Somehow I forgot that `device_matrix_data` can also be views. In that case, I will revert this change.

include/ginkgo/core/base/array.hpp
MarcelKoch: I'm not sure what you are getting at. Should I also add move assignment for const views?

upsj: I mean if we cannot match the semantics we have for array views with const_array_view (which is impossible without const_cast, and thus unsafe), I think we should avoid expanding the use of `const_array_view` beyond it being a wrapper type we can use in conjunction with `create_const`

MarcelKoch: IMO copying a `const_array_view` into an array is a perfectly fine use case. Otherwise copying from a `const T*` into an array is quite complicated and involves a lot of boilerplate. Basically each time you want to do such a copy, you have to copy-paste the implemented copy-assigment to catch all edge cases.

I would also go further and allow for copy-construction.

upsj: I think this will be fine, the distinction between move-assignment and copy-assignment is probably not that important for any but very advanced users, and they should be able to figure out the difference between a const and non-const rhs.

include/ginkgo/core/base/exception_helpers.hpp
```diff
@@ -757,16 +757,16 @@ inline T ensure_allocated_impl(T ptr, const std::string& file, int line,
...
  * Ensures that two dimensions have compatible bounds, in particular before a
- * copy operation. This means the target should have at least as much elements
+ * copy operation. This means the target should have the same number of elements
  * as the source.
  *
  * @param _source  the source of the expected copy operation
  * @param _target  the destination of the expected copy operation
  *
- * @throw OutOfBoundsError  if `_source > _target`
+ * @throw OutOfBoundsError  if `_source == _target`
```
upsj, line 766:
```suggestion
 * @throw OutOfBoundsError  if `_source != _target`
```

upsj (approved):
LGTM!

pratikvn (approved):

MarcelKoch:
format!

MarcelKoch:
Another question: how should `const matrix_data&` be handled. Currently it calls `device_matrix_data&&` internally, so it moves into the stored arrays. Should that stay, or also call `const device_matrix_data&`?

upsj:
This only applies to Csr and Coo only, right? In that case, I would suggest calling `aos_to_soa` directly onto the arrays inside the matrix, i.e. combining the copy with the transformation

core/test/matrix/coo.cpp
```diff
@@ -215,6 +235,112 @@ TYPED_TEST(Coo, CanBeReadFromMatrixAssemblyData)
...
+        gko::device_matrix_data<value_type, index_type>::create_from_host(
+            this->exec, data.get_ordered_data());
+
+    m->read(std::move(device_data));
+
+    this->assert_equal_to_original_mtx(m);
+    ASSERT_NE(row_idxs.get_data(), m->get_row_idxs());
+    ASSERT_NE(col_idxs.get_data(), m->get_col_idxs());
+    ASSERT_NE(values.get_data(), m->get_values());
+}
```
upsj, line 341: maybe check that the pointers are the same that were inside `device_data`?

core/test/matrix/csr.cpp
```diff
@@ -230,6 +260,116 @@ TYPED_TEST(Csr, CanBeReadFromMatrixAssemblyData)
...
+    auto device_data =
+        gko::device_matrix_data<value_type, index_type>::create_from_host(
+            this->exec, data.get_ordered_data());
+
+    m->read(device_data);
+
+    this->assert_equal_to_original_mtx(m);
+    ASSERT_EQ(device_data.get_num_stored_elements(),
+              m->get_num_stored_elements());
+    GKO_ASSERT_EQUAL_DIMENSIONS(&device_data, m);
```
upsj, line 284: same

core/test/matrix/csr.cpp
```diff
@@ -230,6 +260,116 @@ TYPED_TEST(Csr, CanBeReadFromMatrixAssemblyData)
...
+    data.set_value(0, 2, 2.0);
+    data.set_value(1, 1, 5.0);
+    auto device_data =
+        gko::device_matrix_data<value_type, index_type>::create_from_host(
+            this->exec, data.get_ordered_data());
+
+    m->read(std::move(device_data));
+
+    this->assert_equal_to_original_mtx(m);
+    GKO_ASSERT_EQUAL_DIMENSIONS(&device_data, gko::dim<2>{});
```
upsj, line 336: same

core/test/matrix/csr.cpp
```diff
@@ -230,6 +260,116 @@ TYPED_TEST(Csr, CanBeReadFromMatrixAssemblyData)
...
+        gko::device_matrix_data<value_type, index_type>::create_from_host(
+            this->exec, data.get_ordered_data());
+
+    m->read(std::move(device_data));
+
+    this->assert_equal_to_original_mtx(m);
+    ASSERT_EQ(row_ptrs.get_data(), m->get_row_ptrs());
+    ASSERT_NE(col_idxs.get_data(), m->get_col_idxs());
+    ASSERT_NE(values.get_data(), m->get_values());
+}
```
upsj, line 370: same: check equal pointers

include/ginkgo/core/base/array.hpp
```diff
@@ -550,6 +550,60 @@ class array {
...
```
upsj, line 574: Do we need the convertible aspect? That's a part of `array` that I regret adding, because it mixes a very basic data structure with kernel calls, so it makes untangling the library harder.

MarcelKoch: I'm fine with disabling it. I don't need it here.

include/ginkgo/core/base/exception_helpers.hpp
```diff
@@ -757,16 +757,16 @@ inline T ensure_allocated_impl(T ptr, const std::string& file, int line,
...
+ * copy operation. This means the target should have the same number of elements
  * as the source.
  *
  * @param _source  the source of the expected copy operation
  * @param _target  the destination of the expected copy operation
  *
- * @throw OutOfBoundsError  if `_source > _target`
+ * @throw OutOfBoundsError  if `_source != _target`
  */
 #define GKO_ENSURE_COMPATIBLE_BOUNDS(_source, _target)                       \
```
pratikvn, line 768: I dont think the macro should change, but the place where you were using it should use `GKO_ASSERT_EQ` instead. I think this macro is doing the correct thing. 

Actually, looks like this macro is not needed at all. We can consider removing it.

include/ginkgo/core/base/array.hpp
```diff
@@ -550,6 +550,52 @@ class array {
...
+        }
+        if (other.get_executor() == nullptr) {
+            this->clear();
+            return *this;
+        }
+
+        if (this->is_owning()) {
+            this->resize_and_reset(other.get_size());
+        } else {
+            GKO_ENSURE_COMPATIBLE_BOUNDS(other.get_size(), this->get_size());
```
pratikvn, line 585: this should be updated instead
```suggestion
            GKO_ASSERT_EQ(other.get_size(), this->get_size());
```

pratikvn (changes requested):
Macro change and usage of `this->`

pratikvn (approved):
Thanks for updating this! LGTM!

include/ginkgo/core/base/array.hpp
```diff
@@ -550,6 +550,52 @@ class array {
...
+        }
+
+        if (this->is_owning()) {
+            this->resize_and_reset(other.get_size());
+        } else {
+            GKO_ASSERT_EQ(other.get_size(), this->get_size());
+        }
+        array tmp{this->exec_};
+        const ValueType* source = other.get_const_data();
+        // if we are on different executors: copy, then convert
```
upsj, line 589:
```suggestion
        // if we are on different executors: copy
```

upsj (approved):
LGTM! The macro is part of our public headers, but to me, it's only intended for internal usage, so I'm fine with removing it. There is no such thing as a `detail` macro after all.

yhmtsai (changes requested):
I do not think the device_matrix_data view means the matrix also be a view on it.
If user want the Coo view on the device_matrix_data view, they can use the Coo constructor rather than the read

include/ginkgo/core/base/exception_helpers.hpp
```diff
@@ -755,25 +755,6 @@ inline T ensure_allocated_impl(T ptr, const std::string& file, int line,
...
- * Ensures that two dimensions have compatible bounds, in particular before a
- * copy operation. This means the target should have at least as much elements
- * as the source.
- *
- * @param _source  the source of the expected copy operation
- * @param _target  the destination of the expected copy operation
- *
- * @throw OutOfBoundsError  if `_source > _target`
- */
-#define GKO_ENSURE_COMPATIBLE_BOUNDS(_source, _target)                       \
```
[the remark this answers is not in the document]
MarcelKoch: As @upsj technically it is, but we don't care since this should only be used internally.

core/matrix/csr.cpp
```diff
@@ -423,15 +424,56 @@ void Csr<ValueType, IndexType>::move_to(Fbcsr<ValueType, IndexType>* result)
...
-    this->read(device_mat_data::create_from_host(this->get_executor(), data));
+    auto size = data.size;
+    auto exec = this->get_executor();
+    this->set_size(size);
+    this->row_ptrs_.resize_and_reset(size[0] + 1);
+    this->col_idxs_.resize_and_reset(data.nonzeros.size());
+    this->values_.resize_and_reset(data.nonzeros.size());
+    // the device matrix data contains views on the column indices
+    // and values array of this matrix, and an owning array for the
+    // row indices (which doesn't exist in this matrix)
```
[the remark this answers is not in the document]
upsj: we had the same view + owning split previously when moving from a `device_matrix_data`, because row pointers need to be allocated separately in any case.

include/ginkgo/core/base/exception_helpers.hpp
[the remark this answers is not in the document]
MarcelKoch: I will bring it back, because I want to keep the behavior of the current assignment operators.

MarcelKoch:
> I do not think the device_matrix_data view means the matrix also be a view on it.
> If user want the Coo view on the device_matrix_data view, they can use the Coo constructor rather than the read

I'm not understanding this. Do you want to see something changed?

core/test/matrix/csr.cpp
```diff
@@ -230,6 +260,118 @@ TYPED_TEST(Csr, CanBeReadFromMatrixAssemblyData)
 }
 
 
+TYPED_TEST(Csr, CanBeReadFromDeviceMatrixData)
```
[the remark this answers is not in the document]
MarcelKoch: I don't think so. There it's only tested if it's equivalent to reference

core/test/matrix/coo.cpp
```diff
@@ -215,6 +235,115 @@ TYPED_TEST(Coo, CanBeReadFromMatrixAssemblyData)
...
+    auto orig_row_idxs = device_data.get_row_idxs();
+    auto orig_col_idxs = device_data.get_col_idxs();
+    auto orig_values = device_data.get_values();
+
+    m->read(std::move(device_data));
+
+    this->assert_equal_to_original_mtx(m);
+    ASSERT_EQ(orig_row_idxs, m->get_row_idxs());
+    ASSERT_EQ(orig_col_idxs, m->get_col_idxs());
+    ASSERT_EQ(orig_values, m->get_values());
```
yhmtsai, line 343: matrix array view is now to use the device_data view not the original one

yhmtsai (approved):

PR #1479, opened 2023-11-30
Preconditioner and Factorization config

yhmtsai:
This pr contains the config interface for preconditioner and factorization.

TODO:
- [x] how to distinguish the ILU between preconditioner and factorization. I currently use Factorization as prefix. (it is only internal)

### DIFF
--- a/core/config/config_helper.hpp
+++ b/core/config/config_helper.hpp
@@ -43,7 +43,15 @@ enum class LinOpFactoryType : int {
     CbGmres,
     Direct,
     LowerTrs,
-    UpperTrs
+    UpperTrs,
+    Factorization_Ic,
+    Factorization_Ilu,
+    Cholesky,
+    Lu,
+    ParIc,
+    ParIct,
+    ParIlu,
+    ParIlut
 };
 
 
@@ -223,6 +231,24 @@ get_value(const pnode& config)
         return solver::initial_guess_mode::provided;
     }
     GKO_INVALID_STATE("Wrong value for initial_guess_mode");
+}
+
+
+template <typename Csr>
+inline std::shared_ptr<typename Csr::strategy_type> get_strategy(
+    const pnode& config)
+{
+    auto str = config.get_string();
+    std::shared_ptr<typename Csr::strategy_type> strategy_ptr;
+    // automatical and load_balance requires the executor
+    if (str == "sparselib" || str == "cusparse") {
+        strategy_ptr = std::make_shared<typename Csr::sparselib>();
+    } else if (str == "merge_path") {
+        strategy_ptr = std::make_shared<typename Csr::merge_path>();
+    } else if (str == "classical") {
+        strategy_ptr = std::make_shared<typename Csr::classical>();
+    }
+    return std::move(strategy_ptr);
 }
 
 
--- a/core/config/config_helper.hpp
+++ b/core/config/config_helper.hpp
@@ -10,7 +10,6 @@
 #include <type_traits>
 
 
-#include <ginkgo/core/base/array.hpp>
 #include <ginkgo/core/base/exception_helpers.hpp>
 #include <ginkgo/core/base/lin_op.hpp>
 #include <ginkgo/core/base/math.hpp>
@@ -25,6 +24,15 @@
 
 namespace gko {
 namespace config {
+
+
+#define GKO_INVALID_CONFIG_VALUE(_entry, _value)                           \
+    GKO_INVALID_STATE(_value + std::string(" is invalid for the entry ") + \
+                      _entry)
+
+
+#define GKO_MISS_CONFIG_ENTRY(_entry) \
+    GKO_INVALID_STATE(std::string("miss the entry ") + _entry)
 
 
 /**
@@ -257,7 +265,7 @@ get_value(const pnode& config)
     } else if (val == "provided") {
         return solver::initial_guess_mode::provided;
     }
-    GKO_INVALID_STATE("Wrong value for initial_guess_mode");
+    GKO_INVALID_CONFIG_VALUE("default_initial_guess", val);
 }
 
 
@@ -273,31 +281,6 @@ get_value(const pnode& config)
     }
     GKO_INVALID_STATE("should use size 2 array");
 }
-
-
-// template <typename T>
-// struct is_array_t : std::false_type {};
-
-// template <typename V>
-// struct is_array_t<array<V>> : std::true_type {};
-
-// template <typename ArrayType>
-// inline typename std::enable_if<is_array_t<ArrayType>::value, ArrayType>::type
-// get_value(const pnode& config, std::shared_ptr<const Executor> exec)
-// {
-//     using T = typename ArrayType::value_type;
-//     std::vector<T> res;
-//     // for loop in config
-//     if (config.get_tag() == pnode::tag_t::array) {
-//         for (const auto& it : config.get_array()) {
-//             res.push_back(get_value<T>(it));
-//         }
-//     } else {
-//         // only one config can be passed without array
-//         res.push_back(get_value<T>(config));
-//     }
-//     return ArrayType(exec, res.begin(), res.end());
-// }
 
 
 template <typename Csr>
--- a/core/config/config_helper.hpp
+++ b/core/config/config_helper.hpp
@@ -26,13 +26,14 @@ namespace gko {
 namespace config {
 
 
-#define GKO_INVALID_CONFIG_VALUE(_entry, _value)                           \
-    GKO_INVALID_STATE(_value + std::string(" is invalid for the entry ") + \
-                      _entry)
+#define GKO_INVALID_CONFIG_VALUE(_entry, _value)                            \
+    GKO_INVALID_STATE(std::string("The value >" + _value +                  \
+                                  "< is invalid for the entry >" + _entry + \
+                                  "<"))
 
 
 #define GKO_MISS_CONFIG_ENTRY(_entry) \
-    GKO_INVALID_STATE(std::string("miss the entry ") + _entry)
+    GKO_INVALID_STATE(std::string("The entry >") + _entry + "< is missing")
 
 
 /**
@@ -104,26 +105,20 @@ deferred_factory_parameter<T> parse_or_get_factory(const pnode& config,
                                                    const type_descriptor& td);
 
 template <typename T>
-inline deferred_factory_parameter<typename T::Factory> get_specific_factory(
-    const pnode& config, const registry& context, const type_descriptor& td)
+inline deferred_factory_parameter<typename T::Factory>
+parse_or_get_specific_factory(const pnode& config, const registry& context,
+                              const type_descriptor& td)
 {
     using T_non_const = std::remove_const_t<T>;
-    // static_assert(std::is_convertible<T_non_const*, LinOpFactory*>::value,
-    //               "only LinOpFactory");
-    deferred_factory_parameter<typename T::Factory> ptr;
 
     if (config.get_tag() == pnode::tag_t::string) {
-        ptr =
-            detail::registry_accessor::get_data<typename T_non_const::Factory>(
-                context, config.get_string());
+        return detail::registry_accessor::get_data<
+            typename T_non_const::Factory>(context, config.get_string());
     } else if (config.get_tag() == pnode::tag_t::map) {
-        ptr = T_non_const::parse(config, context, td);
+        return T_non_const::parse(config, context, td);
     } else {
         GKO_INVALID_STATE("The data of config is not valid.");
     }
-    GKO_THROW_IF_INVALID(!ptr.is_empty(), "Parse get nullptr in the end");
-
-    return ptr;
 }
 
 
@@ -296,6 +291,8 @@ inline std::shared_ptr<typename Csr::strategy_type> get_strategy(
         strategy_ptr = std::make_shared<typename Csr::merge_path>();
     } else if (str == "classical") {
         strategy_ptr = std::make_shared<typename Csr::classical>();
+    } else {
+        GKO_INVALID_CONFIG_VALUE("strategy", str);
     }
     return std::move(strategy_ptr);
 }
--- a/core/config/config_helper.hpp
+++ b/core/config/config_helper.hpp
@@ -10,6 +10,7 @@
 #include <type_traits>
 
 
+#include <ginkgo/core/base/array.hpp>
 #include <ginkgo/core/base/exception_helpers.hpp>
 #include <ginkgo/core/base/lin_op.hpp>
 #include <ginkgo/core/base/math.hpp>
@@ -51,7 +52,11 @@ enum class LinOpFactoryType : int {
     ParIc,
     ParIct,
     ParIlu,
-    ParIlut
+    ParIlut,
+    Ic,
+    Ilu,
+    Isai,
+    Jacobi
 };
 
 
@@ -90,14 +95,36 @@ deferred_factory_parameter<T> parse_or_get_factory(const pnode& config,
                                                    const registry& context,
                                                    const type_descriptor& td);
 
-/**
- * specialize for const LinOpFactory
- */
+template <typename T>
+inline deferred_factory_parameter<typename T::Factory> get_specific_factory(
+    const pnode& config, const registry& context, const type_descriptor& td)
+{
+    using T_non_const = std::remove_const_t<T>;
+    // static_assert(std::is_convertible<T_non_const*, LinOpFactory*>::value,
+    //               "only LinOpFactory");
+    deferred_factory_parameter<typename T::Factory> ptr;
+
+    if (config.get_tag() == pnode::tag_t::string) {
+        ptr =
+            detail::registry_accessor::get_data<typename T_non_const::Factory>(
+                context, config.get_string());
+    } else if (config.get_tag() == pnode::tag_t::map) {
+        ptr = T_non_const::parse(config, context, td);
+    } else {
+        GKO_INVALID_STATE("The data of config is not valid.");
+    }
+    GKO_THROW_IF_INVALID(!ptr.is_empty(), "Parse get nullptr in the end");
+
+    return ptr;
+}
+
+
 template <>
 deferred_factory_parameter<const LinOpFactory>
 parse_or_get_factory<const LinOpFactory>(const pnode& config,
                                          const registry& context,
                                          const type_descriptor& td);
+
 
 /**
  * specialize for const stop::CriterionFactory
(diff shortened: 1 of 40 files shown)


core/factorization/cholesky.cpp
```diff
@@ -36,6 +39,27 @@ GKO_REGISTER_OPERATION(factorize, cholesky::factorize);
...
+template <typename ValueType, typename IndexType>
+typename Cholesky<ValueType, IndexType>::parameters_type
+Cholesky<ValueType, IndexType>::parse(
+    const config::pnode& config, const config::registry& context,
+    const config::type_descriptor& td_for_child)
+{
+    using sparsity_pattern_type =
+        Cholesky<ValueType, IndexType>::sparsity_pattern_type;
+    auto params = Cholesky<ValueType, IndexType>::build();
+    if (auto& obj = config.get("symbolic_factorization")) {
```
MarcelKoch, line 51: we have some options that require their values to be keys in the registry. Maybe we need to specify this somewhere. It should only concern parameters which are stored as linops in the factory, so maybe we can add somthing about this to the guidelines.

core/config/config_helper.hpp
```diff
@@ -216,6 +251,63 @@ get_value(const pnode& config)
...
+    using T = typename Type::storage_type;
+    if (config.get_tag() == pnode::tag_t::array &&
+        config.get_array().size() == 2) {
+        return Type(get_value<T>(config.get(0)), get_value<T>(config.get(1)));
+    }
+    GKO_INVALID_STATE("should use size 2 array");
+}
+
+
+// template <typename T>
```
MarcelKoch, line 268: all the commented stuff has to go

core/config/config_helper.hpp
```diff
@@ -82,13 +95,35 @@ deferred_factory_parameter<T> get_factory(const pnode& config,
                                           const registry& context,
                                           const type_descriptor& td);
 
-/**
- * specialize for const LinOpFactory
- */
+template <typename T>
+inline deferred_factory_parameter<typename T::Factory> get_specific_factory(
```
MarcelKoch, line 99: is this different from `get_factory` followed by a dynamic cast?
If not, this function can be simplified.

core/config/config_helper.hpp
```diff
@@ -10,6 +10,7 @@
 #include <type_traits>
 
 
+#include <ginkgo/core/base/array.hpp>
```
MarcelKoch, line 13: unnecessary

core/config/preconditioner_config.cpp
```diff
@@ -0,0 +1,295 @@
...
+
+template <>
+deferred_factory_parameter<gko::LinOpFactory> parse<LinOpFactoryType::Ilu>(
+    const pnode& config, const registry& context, const type_descriptor& td)
+{
+    auto updated = update_type(config, td);
+    auto dispatch_solver = [&](auto reverse_apply)
+        -> deferred_factory_parameter<gko::LinOpFactory> {
+        using ReverseApply = decltype(reverse_apply);
+        // always use symmetric solver for USolverType
```
MarcelKoch, line 178: is guess for now that is a sufficient simplification.

core/config/preconditioner_config.cpp
```diff
@@ -0,0 +1,295 @@
...
+
+
+template <>
+deferred_factory_parameter<gko::LinOpFactory> parse<LinOpFactoryType::Isai>(
+    const pnode& config, const registry& context, const type_descriptor& td)
+{
+    auto updated = update_type(config, td);
+    if (auto& obj = config.get("isai_type")) {
+        auto str = obj.get_string();
+        if (str == "lower") {
```
MarcelKoch, line 246: I think for Isai, you could just pass in three type selectors instead of the ifs here.

MarcelKoch: The same might also be possible for Ilu/Ic, but it would require some changes to dispatch, so it's not a high priority.

core/factorization/lu.cpp
```diff
@@ -40,6 +43,41 @@ GKO_REGISTER_HOST_OPERATION(symbolic_lu_near_symm,
...
+    if (auto& obj = config.get("symbolic_algorithm")) {
+        auto str = obj.get_string();
+        if (str == "general") {
+            params.with_symbolic_algorithm(symbolic_type::general);
+        } else if (str == "near_symmetric") {
+            params.with_symbolic_algorithm(symbolic_type::near_symmetric);
+        } else if (str == "symmetric") {
+            params.with_symbolic_algorithm(symbolic_type::symmetric);
+        } else {
+            GKO_INVALID_STATE("Wrong value for symbolic_type");
```
MarcelKoch, line 70: nit: the error message for an invalid string value seems to differ every time. Perhaps just fix one blueprint and use that (or maybe a new assertion macro).

core/factorization/par_ilut.cpp
```diff
@@ -160,6 +163,41 @@ struct ParIlutState {
 }  // namespace
 
 
+template <typename ValueType, typename IndexType>
+typename ParIlut<ValueType, IndexType>::parameters_type
+ParIlut<ValueType, IndexType>::parse(
+    const config::pnode& config, const config::registry& context,
+    const config::type_descriptor& td_for_child)
+{
+    using matrix_type =
```
MarcelKoch, line 172: `preconditioner::ParIlut` already has a `matrix_type` alias, maybe use that.

core/preconditioner/jacobi.cpp
```diff
@@ -50,6 +54,44 @@ GKO_REGISTER_OPERATION(initialize_precisions, jacobi::initialize_precisions);
...
+    if (auto& obj = config.get("max_block_size")) {
+        params.with_max_block_size(gko::config::get_value<uint32>(obj));
+    }
+    if (auto& obj = config.get("max_block_stride")) {
+        params.with_max_block_stride(gko::config::get_value<uint32>(obj));
+    }
+    if (auto& obj = config.get("skip_sorting")) {
+        params.with_skip_sorting(gko::config::get_value<bool>(obj));
+    }
+    // SET_VALUE_ARRAY(factory, gko::array<IndexType>, block_pointers,
```
MarcelKoch, line 74: I would recommend not supporting these kind of parameters, see my comment in #1480

core/preconditioner/jacobi.cpp
```diff
@@ -50,6 +54,44 @@ GKO_REGISTER_OPERATION(initialize_precisions, jacobi::initialize_precisions);
...
+    // config);
+    // storage_optimization_type is not public. It uses precision_reduction
+    // as input. Also, it allows value and array input
+    // Each precision_reduction is created by two values.
+    // [x, y] -> one precision_reduction (value mode)
+    // [[x, y], ...] -> array mode
+    if (auto& obj = config.get("storage_optimization")) {
+        // only support value mode
+        // TODO: more than one precision_reduction -> array mode.
```
MarcelKoch, line 83: same, array mode should not be supported

core/test/config/preconditioner.cpp
```diff
@@ -0,0 +1,415 @@
...
+                    std::shared_ptr<const gko::Executor> exec)
+    {
+        config_map["skip_sorting"] = pnode{true};
+        param.with_skip_sorting(true);
+        config_map["sparsity_power"] = pnode{2};
+        param.with_sparsity_power(2);
+        config_map["excess_limit"] = pnode{32};
+        param.with_excess_limit(32u);
+        config_map["excess_solver_reduction"] = pnode{1e-4};
+        param.with_excess_solver_reduction(1e-4);
```
MarcelKoch, line 215: using float here `1e-4f` will fix the error

core/test/config/preconditioner.cpp
```diff
@@ -0,0 +1,415 @@
...
+    {
+        config_map["value_type"] = pnode{"float32"};
+        config_map["index_type"] = pnode{"int64"};
+    }
+
+    template <bool from_reg, typename ParamType>
+    static void set(pnode::map_type& config_map, ParamType& param, registry reg,
+                    std::shared_ptr<const gko::Executor> exec)
+    {
+        config_map["max_block_size"] = pnode{32};
```
MarcelKoch, line 271: maybe use a different value than the default block size?

core/test/config/preconditioner.cpp
```diff
@@ -0,0 +1,415 @@
...
+        param.with_max_block_stride(32u);
+        config_map["skip_sorting"] = pnode{true};
+        param.with_skip_sorting(true);
+        // config_map["block_pointers"] = pnode{{{0}, {3}, {17}}};
+        // param.with_block_pointers(gko::array<gko::int64>(exec, {0, 3, 17}));
+        config_map["storage_optimization"] =
+            pnode{std::vector<pnode>{pnode{0}, pnode{1}}};
+        param.with_storage_optimization(gko::precision_reduction(0, 1));
+        config_map["accuracy"] = pnode{1e-2};
+        param.with_accuracy(1e-2);
```
MarcelKoch, line 283: same fix using float

core/test/config/preconditioner.cpp
```diff
@@ -0,0 +1,415 @@
...
+
+        ASSERT_EQ(res_so.is_block_wise, ans_so.is_block_wise);
+        ASSERT_EQ(res_so.of_all_blocks, ans_so.of_all_blocks);
+        GKO_ASSERT_ARRAY_EQ(res_so.block_wise, ans_so.block_wise);
+        ASSERT_EQ(res_param.accuracy, ans_param.accuracy);
+    }
+};
+
+
+// struct JacobiByArray : Jacobi {
```
MarcelKoch, line 307: to be removed

core/config/preconditioner_config.cpp
MarcelKoch: Well, it's not as easy as I thought for the isai (because of the enum). So I guess this is something that can be fixed later.

MarcelKoch (changes requested):
lgtm. The same comment on array parameters as in #1480 apply. I would also suggest a uniform error message if an unsupported string value is provided.

core/config/preconditioner_config.cpp
yhmtsai: I have setuped the generalized version but it was complained in the old pr.
In short, you can use integral_constant to map value to the type and then decouple it in the dispatch step.
I prefer that way, but I also agree it is tiny usage.

core/preconditioner/jacobi.cpp
yhmtsai: although it can accept shorter list without knowing the matrix size, I remove it now.

core/config/preconditioner_config.cpp
yhmtsai: I also prefer #1029 in the future without instantiate each type

core/config/config_helper.hpp
yhmtsai: It has the specific type directly, so it will go the the type::parse without the `type` or `value_type`

core/config/config_helper.hpp
```diff
@@ -212,7 +255,39 @@ get_value(const pnode& config)
...
+    auto str = config.get_string();
+    std::shared_ptr<typename Csr::strategy_type> strategy_ptr;
+    // automatical and load_balance requires the executor
+    if (str == "sparselib" || str == "cusparse") {
+        strategy_ptr = std::make_shared<typename Csr::sparselib>();
+    } else if (str == "merge_path") {
+        strategy_ptr = std::make_shared<typename Csr::merge_path>();
+    } else if (str == "classical") {
+        strategy_ptr = std::make_shared<typename Csr::classical>();
+    }
```
MarcelKoch, line 289: does this need an else with `GKO_INVALID_CONFIG_VALUE`?

core/test/config/preconditioner.cpp
```diff
@@ -0,0 +1,400 @@
...
+                    std::shared_ptr<const gko::Executor> exec)
+    {
+        config_map["max_block_size"] = pnode{16};
+        param.with_max_block_size(16u);
+        config_map["max_block_stride"] = pnode{32u};
+        param.with_max_block_stride(32u);
+        config_map["skip_sorting"] = pnode{true};
+        param.with_skip_sorting(true);
+        // config_map["block_pointers"] = pnode{{{0}, {3}, {17}}};
+        // param.with_block_pointers(gko::array<gko::int64>(exec, {0, 3, 17}));
```
MarcelKoch, line 279: can be removed

MarcelKoch (changes requested):
looks good. I would adjust the error message a bit,

core/config/config_helper.hpp
```diff
@@ -82,15 +103,37 @@ deferred_factory_parameter<T> build_or_get_factory(const pnode& config,
...
+    if (config.get_tag() == pnode::tag_t::string) {
+        ptr =
+            detail::registry_accessor::get_data<typename T_non_const::Factory>(
+                context, config.get_string());
+    } else if (config.get_tag() == pnode::tag_t::map) {
+        ptr = T_non_const::parse(config, context, td);
+    } else {
+        GKO_INVALID_STATE("The data of config is not valid.");
+    }
+    GKO_THROW_IF_INVALID(!ptr.is_empty(), "Parse get nullptr in the end");
```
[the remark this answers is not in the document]
yhmtsai: I was thinking of the nullptr in the dataset, but it should be detected in gko::as.
I remove it now

MarcelKoch (approved):

core/preconditioner/ic.cpp
thoasm, line 1: The `core/preconditioner` files for IC and ILU are a special case: They seem to be header-only preconditioners by relying on the templated solver implementation. I see that you currently need the `gko::config::parse_or_get_factory` function from the non-public header `core/config/config_helper.hpp`, so you can't just put this into the header itself.
I looked at it in more detail, and it seems you can't put this into the public header because you rely on `gko::config::registry::get_data`, which is `protected`, so you need the non-public `registry_accessor` to get access.
Do you see any way to just put this into the public header?

If it's not in the public header, you are heavily limiting the `l_solver` and `factorization` parameters to known and compiled values, which is a bit hard to maintain (as it requires changes to this file and `core/config/precontitioner_config.cpp`) and the current list doesn't even include all options. Additionally, you didn't document which `l_solver` or `factorization` you support in the configuration (apart from the error you throw in `core/config/preconditioner_config.cpp`).

core/preconditioner/ilu.cpp
thoasm, line 1: Same in this case.

include/ginkgo/core/factorization/cholesky.hpp
```diff
@@ -88,6 +90,26 @@ class Cholesky
     /** Creates a new parameter_type to set up the factory. */
     static parameters_type build() { return {}; }
 
+    const parameters_type& get_parameters() const { return parameters_; }
```
thoasm, line 93: Please put this below the non-const version and add a doxygen `copydoc` comment.

thoasm (commented):
I would like a short discussion about the IC and ILU preconditioner. The rest looks good to me!

core/preconditioner/ic.cpp
yhmtsai: Yes, it is quite limited. I only put four options which I think should be the usual cases.
If I can put everything in public header, user still need to instantiate the potential options on their own side and then I also need to generate the corresponding selection map based on these possible options in the header. Then it somehow goes back to the first header-only implementation which needs some template/macro trick to make it work. Also, I think it requires user carefully to setup the possible list.

I would rather look forward to the IC/ILU without the specific solver type like https://github.com/ginkgo-project/ginkgo/pull/1029

yhmtsai: I do not have a good idea for this documentation because the limitation is in an internal step.
we only have two public interface for that. 
`parse(...)` for LinOpFactory::parameters is for all cases
`Ic<....>::parse(...)` already knows the type so the limitation does not happen in this function
How about putting the solver limitation in the Ic class? User now can not instantiate the other solver type now.

yhmtsai: I am thinking of removing them now because it destorys the header only class.
I am not sure whether any application or library already uses some specific IC/ILU preconditioner, so this change might break the usage. (kind of interface break?)

MarcelKoch:
@thoasm can you either approve the PR or request changes?

core/preconditioner/ic.cpp
thoasm: I would like to get this resolved before approving.

thoasm (changes requested):

include/ginkgo/core/preconditioner/parse_limit.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+
+namespace gko {
+namespace preconditioner {
+namespace detail {
+
+
+// true_type if ExplicitType is an instantiation of TemplateType.
+template <typename ExplicitType, template <typename...> class TemplateType>
+struct is_instance_of : std::false_type {};
```
MarcelKoch, line 22: IMO `is_instantiation_of` would be a better name (it also matches the docs better)

include/ginkgo/core/preconditioner/parse_limit.hpp
MarcelKoch, line 1: I don't get the name for this file.

MarcelKoch (approved):

include/ginkgo/core/preconditioner/parse_limit.hpp
yhmtsai: I do not come out a good name for it. I put this just because it is the condition for the parse. this function is not general enough, so I do not think should put into some base header?

yhmtsai:
I wonder whether the instantiation of preconditioner Ic/Ilu in config will leads the duplicated symbol if user also use the same instantiation in their own code.
For function, we can use inline. but I was not sure for the class.

from https://en.cppreference.com/w/cpp/language/inline
> A function defined entirely inside a [class/struct/union definition](https://en.cppreference.com/w/cpp/language/classes), whether it's a member function or a non-member friend function, is implicitly an inline function unless it is attached to a [named module](https://en.cppreference.com/w/cpp/language/modules#Module_declarations)(since C++20).

should it be safe?

upsj:
> For function, we can use inline. but I was not sure for the class.

`template` functions automatically emit weak symbols, so we don't need to worry about that

include/ginkgo/core/preconditioner/parse_limit.hpp
yhmtsai: I change it to utils.hpp

include/ginkgo/core/preconditioner/utils.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+
+namespace gko {
+namespace preconditioner {
+namespace detail {
+
+
+// true_type if ExplicitType is an instantiation of TemplateType.
+template <typename ExplicitType, template <typename...> class TemplateType>
+struct is_instantiation_of : std::false_type {};
```
thoasm, line 22: Instantiation doesn't fit in this context as it is used to describe instances of a class.
What about: `is_parametered_template_of` or `is_typed_version_of`?

thoasm (approved):
Well done extending the implementation without considerable changes to our build system or public headers!
I would like to have one more name change. The rest looks good to me!

PR #1485, opened 2023-12-04
Add single-value access functions for `array` access

upsj:
This adds bounds-checked `array::load_value(size_type)` and `array::store_value(size_type, T)` functions and uses them where possible


MarcelKoch (changes requested):
More places to replace:

- common/unified/matrix/dense_kernels.template.cpp:281
- core/baser/index_set.cpp:84
- core/factorization/par_ict.cpp:170|189|191
- core/matrix/sparsity_csr.cpp:136
- dpcpp/matrix/csr_kernels.dp.cpp:2632

include/ginkgo/core/base/array.hpp
```diff
@@ -637,6 +637,46 @@ class array {
...
+     * Returns a single value from the array.
+     *
+     * This involves a bounds check, polymorphic calls and potentially a
+     * device-to-host copy, so it is not suitable for accessing many elements
+     * in performance-critical code.
+     *
+     * @param index  the array element index.
+     * @return the value at index.
+     */
+    value_type get_value(size_type index) const
```
pratikvn, line 650: This will always involve a device to host copy. So, IMO, the function name is misleasing and has potential for misuse. 

I dont see any applications that would need to use this. I am not sure it is necessary to have this as a member of the `gko::array` class. As I see it, the main benefit is to simplify the verbosity of `exec->copy_val_to_host(...)`. Therefore, I would suggest making it a `static` function in a ginkgo internal header and not expose it in the public interface.

include/ginkgo/core/base/array.hpp
```diff
@@ -637,6 +637,46 @@ class array {
...
+     * Sets a single entry in the array to a new value.
+     *
+     * This involves a bounds check, polymorphic calls and potentially a
+     * host-to-device copy, so it is not suitable for accessing many elements
+     * in performance-critical code.
+     *
+     * @param index  the array element index.
+     * @param value  the new value.
+     */
+    void set_value(size_type index, value_type value)
```
pratikvn, line 670: Same here. I think it is better to have this as a `static` function in a ginkgo internal header.

upsj:
@pratikvn Our types are first and foremost used by our code, and an array class that doesn't provide access to its values is at least inconvenient. Thrust has `device_vector`, RAPIDS has `device_uvector`, all of which provide this kind of access and more. There are many ways of getting poor performance out of a high performance library - we have the same issue with automatic cross-executor and cross-precision copies. I don't believe this is sufficient reason to avoid adding this to our fundamental storage type. To me, the amount of simplifications this allows speaks for itself.
If people do encounter poor performance, we have both the ProfilerHook and PerformanceHint loggers to help people find the source of their performance issues.

MarcelKoch (approved):

include/ginkgo/core/base/array.hpp
thoasm, line 650: Maybe change the name to make sure it is clear that it can be copied from a device to the host.
```suggestion
    value_type get_value_to_host(size_type index) const
```

upsj: I replaced them by `load/store_value`, which seemed to be the favored approach

pratikvn (approved):
I still think the names `load_value` and `store_value` are misleading. In particular, because in my view `load` and `store` are verbs that are generally used for low level operations.

But I wont hold back the PR.

upsj:
@pratikvn We have a kind of tendency in Ginkgo to give things elaborate names `get_num_stored_elements`, `get_num_elems`, `copy_val_to_host`, which I think we should cut back on. Arrays are a fundamental type, fundamental operations in fundamental types get short, concise (and **established**) names. This is also why I pushed against @thoasm suggestion of `_to_host`/`_from_host` suffixes.

I agree that `load/store` generally describe a lower abstraction level, but these functions should have concise names, and `get/set_value` wasn't considered clear enough.

upsj:
To expand upon my previous statement, after looking through a few libraries, I feel like `load/store` is used for both more low-level (atomics) and high-level (file system access) functions, but not really for the level in-between.

I looked at some examples in other libraries, I see the following approaches:
```
thrust_device_vector[index] // implicit accessor
sycl_buffer.get_access()[index] // explicit accessor
rapids_device_uvector.get_element(index, stream) // .get access function
```

This makes me question the `load/store` pair again - we are adding a new naming scheme instead of following established patterns. Is this the right way to go?

Expanding on this a bit, I also considered adding a Thrust-like `device_reference` type that could be used to safely implement `Dense::at` and similar functions also on device (e.g. retrieving a norm value) in the long term, that would follow the Thrust/SYCL approach to things. Would that be a better solution, e.g. following the SYCL approach of providing a function to get an accessor/reference?

upsj:
In the latest state I added an accessor/device_reference-based interface, which has a slight readability advantage for writing values:
```cpp
array.set_value(0, 1); // which is the index, which the value?
array.get_access()[0] = 0; // here it's clear
```
Still, these kinds of accessors bring all kinds of issues (e.g. the need to implement a lot of custom operators for cases where the matching operator is implemented with a template function, e.g. in std::complex)

MarcelKoch:
Maybe going back one step and just implement
```
void set_element(array, index, value);
T get_element(array, index);
```
in a private header would be the better choice right now. From the discussion, this would be rather uncontroversial and already fixes the use cases we want to target here.

upsj:
As the least intrusive version, I now moved to an internal access function

core/base/array_access.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+
+/**
+ * Returns a single value from an array.
+ *
+ * This involves a bounds check, polymorphic calls and potentially a
+ * device-to-host copy, so it is not suitable for accessing many elements
+ * in performance-critical code.
+ *
+ * @param index  the array element index.
```
pratikvn, line 23:
```suggestion
 * @param array  the array to get the element from.
 * @param index  the array element index.
```

core/base/array_access.hpp
```diff
@@ -0,0 +1,65 @@
...
+ * @tparam ValueType  the value type of the array.
+ * @return the value at index.
+ */
+template <typename ValueType>
+ValueType get_element(const array<ValueType>& array, size_type index)
+{
+    // TODO2.0 add bounds check for negative indices
+    if (index >= array.get_size()) {
+        throw OutOfBoundsError{__FILE__, __LINE__, index, array.get_size()};
+    }
```
pratikvn, line 33: Can be simplified: 
```suggestion
    GKO_ENSURE_IN_BOUNDS(index, array.get_size());
```

core/base/array_access.hpp
```diff
@@ -0,0 +1,65 @@
...
+ * @param value  the new value.
+ * @tparam ValueType  the value type of the array.
+ */
+template <typename ValueType>
+void set_element(array<ValueType>& array, size_type index, ValueType value)
+{
+    // TODO2.0 add bounds check for negative indices
+    if (index >= array.get_size()) {
+        throw OutOfBoundsError{__FILE__, __LINE__, index, array.get_size()};
+    }
```
pratikvn, line 56:
```suggestion
    GKO_ENSURE_IN_BOUNDS(index, array.get_size());
```

core/base/array_access.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+
+/**
+ * Sets a single entry in the array to a new value.
+ *
+ * This involves a bounds check, polymorphic calls and potentially a
+ * host-to-device copy, so it is not suitable for accessing many elements
+ * in performance-critical code.
+ *
+ * @param index  the array element index.
```
pratikvn, line 46:
```suggestion
 *
 * @param array  the array to set the element.
 * @param index  the array element index.
```

pratikvn (approved):
LGTM!

MarcelKoch (approved):

PR #1497, opened 2023-12-06
OpenMP Merge-SpMV algorithm from Merril & Garland for CSR SpMV

stanisic:
This PR brings Merge-SpMV algorithm from Merril & Garland for CSR SpMV. The implementation is based on the original article https://github.com/dumerrill/merge-spmv/raw/master/merge-based-spmv-sc16-preprint.pdf and source code https://github.com/dumerrill/merge-spmv/blob/master/cpu_spmv.cpp from Merril & Garland work back in 2016. It is using a "2D" approach to do a better load-balancing between the threads. In my extensive investigation on 6 Arm and x86 machines with 8 SuiteSparse matrices using Ginkgo 1.6, this algorithm was performing between -4% to 80% better than the current CSR SpMV implementation.

Implementation in this PR was tested and optimized only for RHS=1, but in principle it should work for RHS>1 as well. Note that I have also tested the second variant of Merge-SpMV algorithm (Algorithm 4 in https://github.com/dumerrill/merge-spmv/raw/master/merge-based-spmv-sc16-preprint.pdf), but it was always performing worse than the original Merge-SpMV algorithm (except for a few matrices on A64FX machine, but I believe this can be ignored).

At the moment the code in this PR still lacks 3 things that I leave to the experts to propose what is the best way to do it:

 1. Integration with the rest of Ginkgo.
 2. If necessary, addition of any tests or comparisons with the default CSR SpMV.
 3. I guess my name and affiliation ("Luka Stanisic <luka.stanisic@huawei.com>, Huawei Technologies Duesseldorf GmbH") as well as Merge-SpMV project (3-Clause BSD license https://github.com/dumerrill/merge-spmv/blob/master/LICENSE.TXT ) should be mentioned somewhere.


upsj (commented):
First pass

core/matrix/csr_kernels.hpp
```diff
@@ -45,6 +45,13 @@ namespace kernels {
                        const matrix::Dense<OutputValueType>* beta,            \
                        matrix::Dense<OutputValueType>* c)
 
+#define GKO_DECLARE_CSR_MERGE_SPMV_KERNEL(MatrixValueType, InputValueType, \
+                                    OutputValueType, IndexType)            \
+    void merge_spmv(std::shared_ptr<const DefaultExecutor> exec,           \
+              const matrix::Csr<MatrixValueType, IndexType>* a,            \
+              const matrix::Dense<InputValueType>* b,                      \
+              matrix::Dense<OutputValueType>* c)
```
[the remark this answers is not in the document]
stanisic: I see now, so you suggest removing this and adding branches in `csr_kernels.cpp` based `a->get_strategy()`, similar to `cuda/matrix/csr_kernels.template.cu`, right?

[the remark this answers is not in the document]
upsj: exactly

omp/matrix/csr_kernels.cpp
```diff
@@ -126,6 +126,112 @@ GKO_INSTANTIATE_FOR_EACH_MIXED_VALUE_AND_INDEX_TYPE(
...
+    const IndexType* row_end_offsets = row_ptrs + 1; // Merge list A: row end offsets
+    const auto num_merge_items = num_rows + nnz; // Merge path total length
+    const auto items_per_thread = (num_merge_items + num_threads - 1) / num_threads; // Merge items per thread
+    array<IndexType> row_carry_out{exec, num_threads};
+    array<arithmetic_type> value_carry_out{exec, num_threads};
+    auto row_carry_out_ptr = row_carry_out.get_data();
+    auto value_carry_out_ptr = value_carry_out.get_data();
+
+    for (size_type j = 0; j < c->get_size()[1]; ++j) {
+#pragma omp parallel for schedule(static)
```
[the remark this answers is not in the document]
upsj: Without modifying the body of the loop, that makes sense, as you need to do more binary searches, and have the scheduling overhead on top of that. I have an idea how to remove the search overhead though, I'll add a prototype soon.

omp/matrix/csr_kernels.cpp
```diff
@@ -126,6 +126,112 @@ GKO_INSTANTIATE_FOR_EACH_MIXED_VALUE_AND_INDEX_TYPE(
...
+    const size_type num_threads = omp_get_max_threads();
+    const IndexType* row_end_offsets = row_ptrs + 1; // Merge list A: row end offsets
+    const auto num_merge_items = num_rows + nnz; // Merge path total length
+    const auto items_per_thread = (num_merge_items + num_threads - 1) / num_threads; // Merge items per thread
+    array<IndexType> row_carry_out{exec, num_threads};
+    array<arithmetic_type> value_carry_out{exec, num_threads};
+    auto row_carry_out_ptr = row_carry_out.get_data();
+    auto value_carry_out_ptr = value_carry_out.get_data();
+
+    for (size_type j = 0; j < c->get_size()[1]; ++j) {
```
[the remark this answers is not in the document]
stanisic: Do you suggest something similar to the approach in `omp/matrix/coo_kernels.cpp`? If yes, then it would require a bit more effort for development and testing. Do you see it as a part of this PR or is this a comment for some future optimizations?

omp/matrix/csr_kernels.cpp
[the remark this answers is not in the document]
stanisic: I have done some performance measurements on Huawei Kunpeng 920 with 8 SuiteSparse matrices, and the results show that using `schedule (static)` is actually giving the best performance. Not specifying `schedule` or using `schedule (auto)` is only slightly worse. This applies for both "classical" CSR SpMV currently implemented in Ginkgo and for the proposed MergeSpMV.

(at least in my measurements on Kunpeng 920) Dynamic scheduling performs terribly for "classical" CSR SpMV. For MergeSpMV, the gap between `dynamic` and other scheduling policy is significantly lower, but `dynamic` is still consistently the worst.

Therefore, I propose for this PR to use either `schedule (static)` or not specify scheduling at all. Any optimized scheduling approach (especially when code is improved for multiple RHS) can be added later.

omp/matrix/csr_kernels.cpp
```diff
@@ -126,6 +123,129 @@ GKO_INSTANTIATE_FOR_EACH_MIXED_VALUE_AND_INDEX_TYPE(
...
+ * Computes the begin offsets into A and B for the specific diagonal
+ */
+template <typename IndexType>
+inline void merge_path_search(
+    const int         diagonal,           ///< [in]The diagonal to search
+    const IndexType*        A,                  ///< [in]List A
+    const int         A_len,              ///< [in]Length of A
+    const int         B_len,              ///< [in]Length of B
+    int&              path_coordinate_x,  ///< [out] (x) coordinate where diagonal intersects the merge path
+    int&              path_coordinate_y)  ///< [out] (y) coordinate where diagonal intersects the merge path
```
yhmtsai, line 136: `int` -> `gko::size_type` or `IndexType`

omp/matrix/csr_kernels.cpp
```diff
@@ -126,6 +123,129 @@ GKO_INSTANTIATE_FOR_EACH_MIXED_VALUE_AND_INDEX_TYPE(
...
 
+/**
+ * Computes the begin offsets into A and B for the specific diagonal
+ */
+template <typename IndexType>
+inline void merge_path_search(
+    const int         diagonal,           ///< [in]The diagonal to search
+    const IndexType*        A,                  ///< [in]List A
+    const int         A_len,              ///< [in]Length of A
+    const int         B_len,              ///< [in]Length of B
```
yhmtsai, line 134:
```suggestion
    const IndexType*        a,                  ///< [in]List A
    const int         a_len,              ///< [in]Length of A
    const int         b_len,              ///< [in]Length of B
```
we use snake case for variable (https://github.com/ginkgo-project/ginkgo/wiki/Contributing-guidelines#variables)

omp/matrix/csr_kernels.cpp
```diff
@@ -126,6 +123,129 @@ GKO_INSTANTIATE_FOR_EACH_MIXED_VALUE_AND_INDEX_TYPE(
...
+    int&              path_coordinate_x,  ///< [out] (x) coordinate where diagonal intersects the merge path
+    int&              path_coordinate_y)  ///< [out] (y) coordinate where diagonal intersects the merge path
+{
+    auto x_min = std::max(diagonal - B_len, 0);
```
yhmtsai, line 138:
```suggestion
    auto x_min = std::max(diagonal - B_len, zero<IndexType>());
```
nit

omp/matrix/csr_kernels.cpp
```diff
@@ -126,6 +123,129 @@ GKO_INSTANTIATE_FOR_EACH_MIXED_VALUE_AND_INDEX_TYPE(
...
+    array<IndexType> row_carry_out{exec, num_threads};
+    array<arithmetic_type> value_carry_out{exec, num_threads};
+    auto row_carry_out_ptr = row_carry_out.get_data();
+    auto value_carry_out_ptr = value_carry_out.get_data();
+
+    for (size_type j = 0; j < c->get_size()[1]; ++j) {
+#pragma omp parallel for schedule(static)
+        for (size_type tid = 0; tid < num_threads; tid++) {
+            const auto start_diagonal = std::min(items_per_thread * tid, num_merge_items);
+            const auto end_diagonal = std::min(start_diagonal + items_per_thread, num_merge_items);
```
yhmtsai, line 189: cast to IndexType to ensure the latter usage type

omp/matrix/csr_kernels.cpp
```diff
@@ -126,6 +123,129 @@ GKO_INSTANTIATE_FOR_EACH_MIXED_VALUE_AND_INDEX_TYPE(
...
+            int thread_coord_x, thread_coord_y, thread_coord_end_x, thread_coord_end_y;
```
yhmtsai, line 190: As https://github.com/ginkgo-project/ginkgo/wiki/Contributing-guidelines#variable-declarations mentioned, we need to declare only one variable in one line.

yhmtsai (changes requested):
Thank you for implementing merge spmv for openmp csr
this is 1/2 part of my review. mainly the format issue and partially implementation comments.
If you have build ginkgo with `-DGINKGO_DEVEL_TOOLS=ON` with pre-commit, you can run `pre-commit run --from-ref "origin/develop" --to-ref HEAD` to format this file. or installing clang-format-14 to format this file should be good enough

omp/matrix/csr_kernels.cpp
```diff
@@ -126,6 +123,129 @@ GKO_INSTANTIATE_FOR_EACH_MIXED_VALUE_AND_INDEX_TYPE(
...
+            x_min = x_pivot + 1;    // Contract range up A (down B)
+	} else {
+            x_max = x_pivot;        // Contract range down A (up B)
+	}
+    }
+
+    path_coordinate_x = std::min(x_min, A_len);
+    path_coordinate_y = diagonal - x_min;
+}
+
```
yhmtsai, line 153:
```suggestion
}


```

omp/matrix/csr_kernels.cpp
```diff
@@ -126,6 +123,129 @@ GKO_INSTANTIATE_FOR_EACH_MIXED_VALUE_AND_INDEX_TYPE(
...
+        acc::helper::build_const_rrm_accessor<arithmetic_type>(b);
+    auto c_vals = acc::helper::build_rrm_accessor<arithmetic_type>(c);
+
+    // Merge-SpMV variables
+    const auto num_rows = a->get_size()[0];
+    const auto nnz = a->get_num_stored_elements();
+    const size_type num_threads = omp_get_max_threads();
+    const IndexType* row_end_offsets = row_ptrs + 1; // Merge list A: row end offsets
+    const auto num_merge_items = num_rows + nnz; // Merge path total length
+    const auto items_per_thread = (num_merge_items + num_threads - 1) / num_threads; // Merge items per thread
```
yhmtsai, line 179: can use `ceildiv`

omp/matrix/csr_kernels.cpp
```diff
@@ -126,6 +123,129 @@ GKO_INSTANTIATE_FOR_EACH_MIXED_VALUE_AND_INDEX_TYPE(
...
+                sum += val * b_vals(col, j);
+	    }
+
+            // Save carry-outs
+            row_carry_out_ptr[tid] = thread_coord_end_x;
+            value_carry_out_ptr[tid] = sum;
+          }
+
+        // Carry-out fix-up (rows spanning multiple threads)
+        for (int tid = 0; tid < num_threads - 1; tid++) {
```
yhmtsai, line 220: maybe add a note something like: the last thread does not carry out any partial result because it must compute the result till the last row end.

yhmtsai (changes requested):
LGTM in other part. also need to add the note to the merge_path option in csr.hpp about the omp merge_path is only supported in apply not advanced_apply

stanisic:
Thank you for the review @yhmtsai, I agree with every single comment that you have made. Unfortunately, I am losing the access to the machine from which I can make public contributions. Therefore, I was hoping that Ginkgo community could take over with merging this PR.

yhmtsai:
@stanisic Thank you for the contribution!
We have implemented and adapted with the review in #1810 based on your contribution.
This PR is closed because #1810 is merged

PR #1498, opened 2023-12-06
Print pre-commit install non-default output

MarcelKoch:
This PR adds a message if `pre-commit install`  exits with non-default output. The default output should be:
```
pre-commit installed at <path>
```
Anything else might indicate an issue, for example if already another pre-commit hook exists. 

I decided against issuing a warning specific for the pre-commit legacy mode, since I hope that this can be more generic.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -372,6 +372,15 @@ if(GINKGO_DEVEL_TOOLS)
                 "Failed to install the git hooks via pre-commit. Please check the error message:\n"
                 "${pre-commit-output}\n${pre-commit-error}")
     endif()
+    if(pre-commit-output MATCHES "^Running in migration mode with existing hooks")
+        message(WARNING
+                "An existing git hook was encountered during `pre-commit install`. The old git hook "
+                "will also be executed. Consider removing it with `pre-commit install -f`")
+    elseif(NOT pre-commit-output MATCHES "^pre-commit installed at")
+        message(WARNING
+                "`pre-commit install` did not exit normally. Please check the output message:\n"
+                "${pre-commit-output}")
+    endif()
 
     add_custom_target(format
                       COMMAND bash -c "${PRE_COMMIT} run"


CMakeLists.txt
```diff
@@ -372,6 +372,11 @@ if(GINKGO_DEVEL_TOOLS)
                 "Failed to install the git hooks via pre-commit. Please check the error message:\n"
                 "${pre-commit-output}\n${pre-commit-error}")
     endif()
+    if(NOT pre-commit-output MATCHES "^pre-commit installed at")
+        message(WARNING
+                "`pre-commit install` did not exit normally. Please check the output message:\n"
+                "${pre-commit-output}")
```
yhmtsai, line 378:
```suggestion
                "`pre-commit install` did not exit normally. If the old pre-commit is not necessary, please consider to delete .git/pre-commit.legacy. Please check the output message:\n"
                "${pre-commit-output}")
```
I do not think the pre-commit output is clear enough such that user know git pre-commit hook will still execute `pre-commit.legacy`. For me, it is only like moving the old code to another part like backup.

Also, pre-commit does not take care of the new hook in `pre-commit`,
The new hook will be deleted if you add something into pre-commit.
There's ID in the file:
- delete ID: pre-commit will install a new one and copy the current one to legacy (overwrite the old legacy)
- keep ID: lose everything other than pre-commit generated.

yhmtsai: or "pre-commit.legacy will still be executed"

MarcelKoch: As I mentioned, I decided against issuing a warning specific to the legacy hook, because the pre-commit output might contain something completely different. In that case, mentioning the legacy would be confusing.

This is mostly relevant for us anyway, and we know how to fix the legacy hook, so I think the warning should be enough of a reminder to actually fix it.

yhmtsai: I am not sure whether I can get an idea how to fix with `pre-commit` message before I dig into it by myself.
It still gives somehow a hint, so I guess it's okay.

yhmtsai (approved):

CMakeLists.txt
MarcelKoch: Ok, I added the more the more verbose and specialized warning.

CMakeLists.txt
```diff
@@ -372,6 +372,15 @@ if(GINKGO_DEVEL_TOOLS)
                 "Failed to install the git hooks via pre-commit. Please check the error message:\n"
                 "${pre-commit-output}\n${pre-commit-error}")
     endif()
+    if(pre-commit-output MATCHES "^Running in migration mode with existing hooks")
+        message(WARNING
+                "An existing git hooks was encountered during `pre-commit install`. The old git hook "
+                "will also be executed. Consider removing it with `pre-commit install -f`")
```
pratikvn, line 378:
```suggestion
                "An existing git hook was encountered during `pre-commit install`. The old git hook "
                "will also be executed. Consider removing it with `pre-commit install -f`")
```

pratikvn (approved):

PR #1500, opened 2023-12-06
Update gdb pretty-printer

upsj:
Since the `array` member names changed, we need to fix the pretty-printer. Also we should avoid running it on pointers to arrays.


upsj:
There are multiple cases where `print`ing a variable can accidentally call this pretty-printer, e.g. if you have a `array<IndexType>::value_type` return type somewhere or an `array<IndexType> *` type. We only want to handle the plain `array<IndexType>` case, which is almost caught by restricting this to structs.

pratikvn (approved):

yhmtsai (approved):

Issue #1144, opened 2022-10-15
Add CMake options to install the documentation, benchmarks and examples


lahwaacz:
CMake options to install the documentation, benchmarks and examples would be useful for [packaging](https://github.com/ginkgo-project/ginkgo/discussions/1095). I guess the benchmarks and examples shouldn't go directly into `bin/` as they don't have a common prefix, so the install paths should be configurable. For example, installing the built examples (and source code) to `/usr/share/doc/ginkgo-hpc-git/examples/` would make sense.

stanisic:
From a personal experience, I can say that depending on the installation workflow, `benchmark` might not work so smoothly out-of-the-box. For example, if spack/easybuild/... is used for installing Ginkgo, then Makefiles generated by CMake can have some hardcoded paths that are no longer valid. In overall, to make `benchmark` installation robust, some changes to the `run_all_benchmarks.sh` might be needed.

I dont know enough about the documentation and examples to comment about that part.

pratikvn:
Thanks @lahwaacz for the report. I think installing documentation might be a good idea. In line with the usual linux packages, it probably should go into `doc/`. Benchmarks are generally meant for developers and I dont  think we should install and package them. Examples are meant for interactive usage and packaging them is probably not what we would like to have.

upsj:
What might be interesting (comparing it to the CUDA samples) is installing the source code of the examples, so people can see how to use CMake etc. correctly (also connected to #1152, we don't seem to test out-of-source builds ATM)

upsj:
If we want to approach this, we probably want to use separate CMake install components `install(... COMPONENT Examples|Benchmark|...)`

PR #1502, opened 2023-12-08
Improve packageability of install and install gdb pretty-printer

upsj:
To make packaging Ginkgo easier, we can separate the installable components into `Developement` and `Runtime` (imagine libginkgo and libginkgo-dev packages). By default all packages get installed (we can control this using `EXCLUDE_FROM_ALL`), but you can choose individual components separately.

For some more context, look at https://www.youtube.com/watch?v=m0DwB4OvDXk around 30:00

Related content: https://www.youtube.com/watch?v=sBP17HQAQjk https://www.youtube.com/watch?v=_5weX5mx8hc

Also I noticed that the compiler version check doesn't check whether a language is enabled, so if CUDA is not enabled, it complains about us using the wrong CUDA compiler.

Finally, every time a shared library gets loaded, gdb looks for `libsomething.so.x.x.x-gdb.py` files in its auto-load safe-path. We can use this to make our pretty-printer available to user debug builds.

### DIFF
--- a/cmake/install_helpers.cmake
+++ b/cmake/install_helpers.cmake
@@ -47,14 +47,15 @@ function(ginkgo_install_library name)
     install(TARGETS "${name}"
         EXPORT Ginkgo
         LIBRARY
-        DESTINATION "${CMAKE_INSTALL_FULL_LIBDIR}"
-        COMPONENT Runtime
+            DESTINATION "${CMAKE_INSTALL_FULL_LIBDIR}"
+            COMPONENT Ginkgo_Runtime
+            NAMELINK_COMPONENT Ginkgo_Development
         RUNTIME
-        DESTINATION "${CMAKE_INSTALL_FULL_BINDIR}"
-        COMPONENT Runtime
+            DESTINATION "${CMAKE_INSTALL_FULL_BINDIR}"
+            COMPONENT Ginkgo_Runtime
         ARCHIVE
-        DESTINATION "${CMAKE_INSTALL_FULL_LIBDIR}"
-        COMPONENT Development
+            DESTINATION "${CMAKE_INSTALL_FULL_LIBDIR}"
+            COMPONENT Ginkgo_Development
         )
 endfunction()
 
@@ -63,17 +64,17 @@ function(ginkgo_install)
     install(FILES ${Ginkgo_BINARY_DIR}/ginkgo_$<CONFIG>.pc
         DESTINATION "${GINKGO_INSTALL_PKGCONFIG_DIR}"
         RENAME ginkgo.pc
-        COMPONENT Development)
+        COMPONENT Ginkgo_Development)
 
     # install the public header files
     install(DIRECTORY "${Ginkgo_SOURCE_DIR}/include/"
         DESTINATION "${CMAKE_INSTALL_FULL_INCLUDEDIR}"
-        COMPONENT Development
+        COMPONENT Ginkgo_Development
         FILES_MATCHING PATTERN "*.hpp"
         )
     install(FILES "${Ginkgo_BINARY_DIR}/include/ginkgo/config.hpp"
         DESTINATION "${CMAKE_INSTALL_FULL_INCLUDEDIR}/ginkgo"
-        COMPONENT Development
+        COMPONENT Ginkgo_Development
         )
 
     if  (GINKGO_HAVE_HWLOC AND NOT HWLOC_FOUND)
@@ -81,23 +82,23 @@ function(ginkgo_install)
         file(GLOB HWLOC_LIBS "${HWLOC_LIB_PATH}/libhwloc*")
         install(FILES ${HWLOC_LIBS}
             DESTINATION "${CMAKE_INSTALL_FULL_LIBDIR}"
-            COMPONENT Runtime
+            COMPONENT Ginkgo_Runtime
             )
         # We only use hwloc and not netloc
         install(DIRECTORY "${HWLOC_INCLUDE_DIRS}/hwloc"
             DESTINATION "${CMAKE_INSTALL_FULL_INCLUDEDIR}"
-            COMPONENT Development
+            COMPONENT Ginkgo_Development
             )
         install(FILES "${HWLOC_INCLUDE_DIRS}/hwloc.h"
             DESTINATION "${CMAKE_INSTALL_FULL_INCLUDEDIR}"
-            COMPONENT Development
+            COMPONENT Ginkgo_Development
             )
     endif()
 
     # Install CMake modules
     install(DIRECTORY "${Ginkgo_SOURCE_DIR}/cmake/Modules/"
         DESTINATION "${GINKGO_INSTALL_MODULE_DIR}"
-        COMPONENT Development
+        COMPONENT Ginkgo_Development
         FILES_MATCHING PATTERN "*.cmake"
         )
 
@@ -117,12 +118,12 @@ function(ginkgo_install)
         "${Ginkgo_BINARY_DIR}/GinkgoConfig.cmake"
         "${Ginkgo_BINARY_DIR}/GinkgoConfigVersion.cmake"
         DESTINATION "${GINKGO_INSTALL_CONFIG_DIR}"
-        COMPONENT Development)
+        COMPONENT Ginkgo_Development)
     install(EXPORT Ginkgo
         NAMESPACE Ginkgo::
         FILE GinkgoTargets.cmake
         DESTINATION "${GINKGO_INSTALL_CONFIG_DIR}"
-        COMPONENT Development)
+        COMPONENT Ginkgo_Development)
 
     # Export package for use from the build tree
     if (GINKGO_EXPORT_BUILD_DIR)
@@ -134,6 +135,6 @@ function(ginkgo_install)
             "${Ginkgo_SOURCE_DIR}/dev_tools/scripts/gdb-ginkgo.py"
             DESTINATION "${CMAKE_INSTALL_FULL_LIBDIR}"
             RENAME "$<TARGET_FILE_NAME:ginkgo>-gdb.py"
-            COMPONENT Development)
+            COMPONENT Ginkgo_Development)
     endif()
 endfunction()


CMakeLists.txt
```diff
@@ -91,6 +91,7 @@ option(GINKGO_DPCPP_SINGLE_MODE "Do not compile double kernels for the DPC++ bac
 option(GINKGO_INSTALL_RPATH "Set the RPATH when installing its libraries." ON)
 option(GINKGO_INSTALL_RPATH_ORIGIN "Add $ORIGIN (Linux) or @loader_path (MacOS) to the installation RPATH." ON)
 option(GINKGO_INSTALL_RPATH_DEPENDENCIES "Add dependencies to the installation RPATH." OFF)
+option(GINKGO_INSTALL_GDB_PRETTY_PRINTER "Install the GDB pretty-printer for auto-loading." OFF)
```
pratikvn, line 94: To reduce the number of CMake options we would need to manage, maybe just bundle the pretty-printer with `GINKGO_DEVEL_TOOLS` ?

upsj: Those are two different things - one of them is used for development purposes (where you usually don't install your code), and the other one is for application debugging purposes (on an installed Ginkgo). But it definitely makes sense to copy the pretty-printer next to the compiled libginkgo.so for debugging purposes.

upsj: Another option to approach this (probably also nice for packagers) is to use `install(... COMPONENT Development)`, this also makes sense for #1144

pratikvn: Yes, I think that is a good idea!

cmake/install_helpers.cmake
```diff
@@ -44,66 +44,63 @@ endfunction()
...
+        EXPORT Ginkgo
+        LIBRARY
+        DESTINATION "${CMAKE_INSTALL_FULL_LIBDIR}"
+        COMPONENT Runtime
+        RUNTIME
+        DESTINATION "${CMAKE_INSTALL_FULL_BINDIR}"
+        COMPONENT Runtime
+        ARCHIVE
+        DESTINATION "${CMAKE_INSTALL_FULL_LIBDIR}"
+        COMPONENT Development
```
pratikvn, line 57: I think it is better to have a Ginkgo_ prefix to avoid nameclashes when installed with FetchContent or as submodules.
```suggestion
        COMPONENT Ginkgo_Development
```

cmake/install_helpers.cmake
```diff
@@ -44,66 +44,63 @@ endfunction()
...
-        # install .so and .a files
-        install(TARGETS "${name}"
-            EXPORT Ginkgo
-            LIBRARY DESTINATION "${CMAKE_INSTALL_FULL_LIBDIR}"
-            ARCHIVE DESTINATION "${CMAKE_INSTALL_FULL_LIBDIR}"
+    install(TARGETS "${name}"
+        EXPORT Ginkgo
+        LIBRARY
+        DESTINATION "${CMAKE_INSTALL_FULL_LIBDIR}"
+        COMPONENT Runtime
```
pratikvn, line 51: namelink can go into development component
```suggestion
        COMPONENT Runtime
        NAMELINK_COMPONENT Development
```

cmake/install_helpers.cmake
```diff
@@ -44,66 +44,63 @@ endfunction()
...
+        RUNTIME
+        DESTINATION "${CMAKE_INSTALL_FULL_BINDIR}"
+        COMPONENT Runtime
```
pratikvn, line 54: I think this formatting would be better (with COMPONENT just below DESTINATION, github seems to display it badly): 
```suggestion
        RUNTIME DESTINATION "${CMAKE_INSTALL_FULL_BINDIR}"
                         COMPONENT Runtime
```

pratikvn (commented):
I thought you also wanted to install the benchmark related files ?

cmake/install_helpers.cmake
upsj: Great point! I missed that

cmake/install_helpers.cmake
upsj: Good point on subdir builds, we need to keep that in mind. I would prefer doing this by making the installation itself configurable, like GTest does it, because Development and Runtime seem to be common names, and other libraries might want to install Ginkgo itself with the same component

upsj: I added a `GINKGO_INSTALL` option which, when `OFF`, disables the corresponding `install` functions and `test_install`, does that make sense?

MarcelKoch: I don't think this fixes the issue. There might still be name conflicts with other libraries, since `GINKGO_INSTALL` is on by default. So users would need to know that they have to set it to `OFF` to prevent these name conflicts. I don't know how we could warn users about that.
But the `GINKGO_INSTALL` option can stay, I would only suggest adding the prefixes.

yhmtsai (commented):
Runtime is only used in runtime linking, right?
Users will only use them with some pre-built packages.

MarcelKoch (commented):
Would this be a good opportunity to also fix the `cmake --install . --prefix some/prefix/path` behavior? Right now, the `--prefix` path is just ignored.

cmake/install_helpers.cmake
upsj: I think the issue is much smaller than it seems. A subdir build using Ginkgo would be broken if Ginkgo wasn't being installed, so `GINKGO_INSTALL` being on by default is a sensible value IMO. GTest differs in the respect that we only use it for testing, not for production. Additionally, component installs are an advanced usage of CMake, by default all components get installed directly. If people use the same naming scheme (Runtime and Development), `cmake --install . --component Runtime` is safe to run and installs the application plus Ginkgo, if that is not intended, we can expect advanced users to set a flag manually.

MarcelKoch: I just mean, you already linked a presentation, which criticizes using generic names. As @pratikvn mentioned, if a user includes ginkgo through `FetchContent` and they directly, or another library they are using, has the same generic names there is a conflict. But we can solve that very easily as mentioned in the video.
And again, I'm not arguing against `GINKGO_INSTALL`, only against the generic names.

upsj: Ah right thanks, I was going by memory on the component setup. I'll use a prefix

upsj: In that case we probably don't need `GINKGO_INSTALL`, unless we consider users using Ginkgo as a purely build-time dependency in a subtree build a realistic use case?

MarcelKoch: Yes, but I can't see how users would know that. I think it's better to use the non-generic names and not hope that the user reads our documentation and figures out that they have to disable `GINKGO_INSTALL`.

upsj: Not sure I follow, can you elaborate? What I mean is that if people do a normal (not component-wise) install of their application, they need to install Ginkgo with it, with separate component names, the packageability is still guaranteed. The GTest use case of it being purely a build and test dependency doesn't seem important.

MarcelKoch: That sounds like we agree to use the non-generic names, right? So maybe my last comment was just badly phrased and you can ignore it.

MarcelKoch:
@upsj Ok, I just wanted to check that. Since I'm the only one complaining about it, I will look into it.

upsj:
@pratikvn Installing benchmarks is much more complicated, since it requires installing things like gflags and our timer/linop wrapper libraries. I'll leave that for a subsequent PR

tcojean:
> @pratikvn Installing benchmarks is much more complicated, since it requires installing things like gflags and our timer/linop wrapper libraries. I'll leave that for a subsequent PR

Couldn't we just depend on a package manager (spack, conan, whichever) to first install gflags as most other package does? And if that was not the case, fail loudly. I don't think we need to change that much, and I don't think it's unreasonable to ask people to install that package if they really want the benchmarks to be installed.

It does require an option to separately install dev tools/support compared to the core.

pratikvn (approved):
LGTM!

lahwaacz:
@upsj You mentioned "auto-load safe-path" in https://github.com/ginkgo-project/ginkgo/pull/1502#issue-2033095195 but the `$prefix/lib` directory, such as `/usr/lib`, where the gdb script is installed, is not a safe-path by default (at least on Arch Linux):

```
(gdb) show auto-load safe-path
List of directories from which it is safe to auto-load files is $debugdir:$datadir/auto-load.
(gdb) show data-directory
GDB's data directory is "/usr/share/gdb".
(gdb) show debug-file-directory
The directory where separate debug symbols are searched for is "/usr/lib/debug".
```

Indeed, most Arch Linux packages install gdb scripts in `/usr/share/gdb/auto-load/usr/lib/`:

```
$ pacman -Fx 'usr/lib/lib.*-gdb.py' | sort
usr/lib/libginkgo.so.1.9.0-gdb.py is owned by extra/ginkgo-hpc 1.9.0-1
usr/lib/libginkgo.so.1.9.0-gdb.py is owned by extra/ginkgo-hpc-cuda 1.9.0-1
usr/lib/libginkgo.so.1.9.0-gdb.py is owned by extra/ginkgo-hpc-hip 1.9.0-1
usr/share/gdb/auto-load/usr/lib/libarrow.so.1900.0.0-gdb.py is owned by extra/arrow 19.0.0-1
usr/share/gdb/auto-load/usr/lib/libeo.so.1.28.0-gdb.py is owned by extra/efl 1.28.0-1
usr/share/gdb/auto-load/usr/lib/libglib-2.0.so.0.8200.4-gdb.py is owned by core/glib2-devel 2.82.4-2
usr/share/gdb/auto-load/usr/lib/libgobject-2.0.so.0.8200.4-gdb.py is owned by core/glib2-devel 2.82.4-2
usr/share/gdb/auto-load/usr/lib/libgstreamer-1.0.so.0.2412.0-gdb.py is owned by extra/gstreamer 1.24.12-1
usr/share/gdb/auto-load/usr/lib/libisl.so.23.4.0-gdb.py is owned by core/libisl 0.27-1
usr/share/gdb/auto-load/usr/lib/libstdc++.so.6.0.33-gdb.py is owned by core/gcc 14.2.1+r753+g1cd744a6828f-1
```

Would you consider changing this or making it configurable in ginkgo?

upsj:
Good point, there are two possible approaches we could take: 1. make the path configurable in CMake or 2. move the pretty-printer to a separate install component like Ginkgo_PrettyPrinter. That would allow you to install it with a different prefix. I would probably prefer the latter, since it is easier to implement, and both solution come with little additional work for the packager. What do you think?

lahwaacz:
Hmm does cmake allow to easily install all except one specific components? AFAIK, it can either install everything or one component at a time. This would require me to loop over all components during install...

upsj:
There are two components only, one to build a runtime package, one a development package. If you don‘t have have the same package separation, are three separate calls manageable? Otherwise we can consider the CMake variable approach.

lahwaacz:
It is manageable atm, but I'd have to watch out for the future components :smile:  I still like the cmake variable more, it allows to keep the script in the current component and location by default without breaking the expectations of everybody else.

PR #1507, opened 2023-12-13
Update CITING.md with new papers

pratikvn:
Following the suggestion from @tcojean, this PR adds new papers and sections to the CITING.md file

### DIFF
--- a/CITING.md
+++ b/CITING.md
@@ -1,8 +1,6 @@
 # Citing Ginkgo                                           {#citing_ginkgo}
 
-The main Ginkgo paper describing Ginkgo's purpose, design and interface is
-available through the following reference:
-
+The main Ginkgo interface design, performance paper is
 ``` bibtex
 @article{ginkgo-toms-2022,
 title = {{Ginkgo: A Modern Linear Operator Algebra Framework for High Performance Computing}},
@@ -23,16 +21,11 @@ pages = {2:1--2:33}
 }
 ```
 
-Multiple topical papers exist on Ginkgo and its algorithms. The following papers
-can be used to cite specific aspects of the Ginkgo project.
-
-### The Ginkgo Software
-
-The Ginkgo software itself was reviewed and has a paper published in the Journal
-of Open Source Software, which can be cited with the following reference:
+and the software itself has been reviewed and published in Journal of Open Source Software (JOSS).
 
 ```bibtex
-@article{GinkgoJoss2020,
+
+@article{ginkgo-joss-2020,
   doi = {10.21105/joss.02260},
   url = {https://doi.org/10.21105/joss.02260},
   year = {2020},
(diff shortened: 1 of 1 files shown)


MarcelKoch (commented):
I think it's missing clear statements on which papers to actually cite. The TOMS paper should probably have something like this at the top:

> Please cite this if you are using Ginkgo in your research.

And for the other features there should also be a single paper highlighted, which best represents the feature. (But I get that this might not make sense everywhere.)

tcojean:
Thanks Pratik for taking this up! I agree with Marcel, we should not clutter this file too much. The main point from my PoV is to concentrate citations on e.g. journal papers rather than have a few citations spread over too many papers. It's up to everyone to pick which papers they prefer is cited for their research, but I do think we should keep this list as short as possible.

pratikvn:
I think that is quite difficult to do. In many cases, papers are mostly intertwined with applications, so specifying one paper for the feature is difficult. I mostly added a few main papers for each, and let the user decide which they want to use. Maybe we can make a rule that we add two papers per section to keep this file limited.

I think maybe one approach would be to have a recommendation only for the main Ginkgo paper, but list maybe two/three papers for each major feature, and let the user decide if they want to cite something additional to the main Ginkgo paper.

CITING.md
```diff
@@ -43,22 +36,57 @@ of Open Source Software, which can be cited with the following reference:
...
 
 ``` bibtex
-@misc{tsai2020amdportability,
-    title={Preparing Ginkgo for AMD GPUs -- A Testimonial on Porting CUDA Code to HIP},
-    author={Yuhsiang M. Tsai and Terry Cojean and Tobias Ribizel and Hartwig Anzt},
-    year={2020},
-    eprint={2006.14290},
-    archivePrefix={arXiv},
-    primaryClass={cs.MS}
+@inproceedings{tsaiPreparingGinkgoAMD2021,
```
tcojean, line 56: I think the main paper which summarizes these efforts would be:

```@article{COJEAN2022102902,
title = {{Ginkgo—A math library designed for platform portability}},
journal = {{Parallel Computing}},
volume = {111},
pages = {102902},
year = {2022},
issn = {0167-8191},
doi = {https://doi.org/10.1016/j.parco.2022.102902},
url = {https://www.sciencedirect.com/science/article/pii/S0167819122000096},
author = {Terry Cojean and Yu-Hsiang Mike Tsai and Hartwig Anzt},
keywords = {Porting to GPU accelerators, Platform Portability, Performance portability, AMD, NVIDIA, Intel}
}
```


I generally think it's better to cite journals first or important conferences (e.g., IPDPS).

CITING.md
```diff
@@ -119,15 +148,217 @@ articleno = {2},
 numpages = {26},
 keywords = {irregular matrices, GPUs, Sparse Matrix Vector Product (SpMV)}
 }
+
+@inproceedings{anztEvaluatingPerformanceNVIDIA2020,
```
tcojean, line 152: This paper is hard to put in one place, it's more than sparse to begin with, and also has solvers performance. It's essentially an A100 benchmarking paper, and comparing to V100 performance, I don't know if it fits here.

CITING.md
```diff
@@ -119,15 +148,217 @@ articleno = {2},
...
+  urldate = {2023-12-06},
+  keywords = {ginkgo}
+}
+```
+
+
+### Preconditioners
+
+```bibtex
+@inproceedings{gobelMixedPrecisionIncomplete2021,
```
tcojean, line 169: We could also consider Hartwig and Tobias' ParILUT paper in this section?
https://scholar.google.com/citations?view_op=view_citation&citation_for_view=ctNCF8YAAAAJ:RGFaLdJalmkC

CITING.md
```diff
@@ -119,15 +148,217 @@ articleno = {2},
...
+  publisher = {{Springer Nature Switzerland}},
+  address = {{Cham}},
+  doi = {10.1007/978-3-031-23606-8_3},
+  copyright = {All rights reserved},
+  isbn = {978-3-031-23606-8},
+  langid = {english},
+  keywords = {Batched preconditioners,Batched solvers,GINKGO,GPU,Sparse linear systems}
+}
+
+@inproceedings{kashiBatchedSparseIterative2022,
```
tcojean, line 335: I would put IPDPS first I think.

tcojean (commented):
some comments

PR #1510, opened 2023-12-13
Remove HWLOC

MarcelKoch:
This PR removes HWLOC. Over time the hwloc component in Ginkgo seem to have lost its functionality, and now, it does absolutely nothing. We only use it to create the `machine_topology` class, but at no point we are actually using any of the data it stores.

So, this deprecates/removes the cmake option for HWLOC, and removes all references to it in our code. This again shows that we did not use HWLOC in any relevant capacity.

One point for discussion: I didn't remove `Executor::populate_exec_info(machine_topology*)`, although the `machine_topology` class is basically empty. But that function is protected, so it's part of our interface. Removing it would be a bit simpler, but we would need to decide on that.


MarcelKoch:
This seems to be irrelevant now, after #1513 was merged.

PR #1513, opened 2023-12-15
Disable HWLOC by Default

MarcelKoch:
This PR disables our HWLOC support by default. This is a softer version of #1510. It also changes the behavior of `GINKGO_BUILD_HWLOC`. Now, if set to `ON`, then it is required that HWLOC can be found.

Also, a warning is issued if `GINKGO_BUILD_HWLOC=ON`, since this has no (positive) effect.

Todo:
- [x] maybe disable HWLOC for tests


pratikvn (approved):
LGTM!

upsj (approved):

Issue #1429, opened 2023-10-15
ROCm gfx10 and gfx11 not supported


Madouura:
Looks like only gfx9 and below is supported here ATM.
gfx1010, gfx1012, gfx1030, gfx1100, gfx1101, and gfx1102 are affected.
Here's the error log. I'll link you to the build environment once I submit a PR to nixpkgs.

```console
ginkgo-hpc-hip> In file included from /build/source/build/hip/preconditioner/jacobi_advanced_apply_instantiate.64.hip.cpp:66:
ginkgo-hpc-hip> /build/source/common/cuda_hip/preconditioner/jacobi_advanced_apply_kernel.hpp.inc:84:9: error: no matching function for call to 'tiled_partition'
ginkgo-hpc-hip>         group::tiled_partition<subwarp_size>(group::this_thread_block());
ginkgo-hpc-hip>         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ginkgo-hpc-hip> /build/source/build/hip/preconditioner/jacobi_advanced_apply_instantiate.64.hip.cpp:98:21: note: in instantiation of function template specialization 'gko::kernels::hip::jacobi::kernel::advanced_adaptive_apply<64, 64, 4, float, int>' requested here
ginkgo-hpc-hip>             kernel::advanced_adaptive_apply<max_block_size, subwarp_size,
ginkgo-hpc-hip>                     ^
ginkgo-hpc-hip> /build/source/hip/components/cooperative_groups.hip.hpp:392:5: note: candidate template ignored: requirement '64UL <= kernels::hip::config::warp_size' was not satisfied [with Size = 64]
ginkgo-hpc-hip>     tiled_partition(const Group&)
ginkgo-hpc-hip>     ^
ginkgo-hpc-hip> In file included from /build/source/build/hip/preconditioner/jacobi_advanced_apply_instantiate.64.hip.cpp:66:
ginkgo-hpc-hip> /build/source/common/cuda_hip/preconditioner/jacobi_advanced_apply_kernel.hpp.inc:50:9: error: no matching function for call to 'tiled_partition'
ginkgo-hpc-hip>         group::tiled_partition<subwarp_size>(group::this_thread_block());
ginkgo-hpc-hip>         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #1516, opened 2023-12-19
Unify more CUDA and HIP files into common

upsj:
A lot of the files we have in `common/cuda_hip` don't work standalone, but instead require to be included with certain other includes being available. Here I'm trying to change this, allowing the files to be parsed from tooling more easily, and making it easier to write CUDA and HIP code simultaneously.

I did many of these modifications automatically using the `unify_cuda_hip.py` python script, or using plain text replacement to fix the headers

Some related questions:

- ~~What should be the canonical place of including `hip_runtime.h`? I don't think we should have to do that with every file, instead something like `config.hip.hpp` or `types.hip.hpp` seems most appropriate~~
- Should `hip::config` be used on host or device or both? If we use it on the device only, we should be able to handle #1429 more easily, the host side should instead be handled with a runtime value, like `exec->get_warp_size()`


MarcelKoch:
TBH, I'm not sure if this can be reviewed by humans. 200+ changed files are not easily processed manually. So I guess we have to rely on our CI.  If you could provide some key files to look at, that would help a lot. 
But in general terms, unifying redundant code sounds good to me. Is the goal to unify as many files as possible and only exclude those, where the actual implementation differs? If so, maybe a list of files that are not suitable would be helpful.

upsj:
@MarcelKoch Most of these changes are 90% automatical, but the remaining 10% is what's interesting. I'll see what I can do in terms of providing a diff what changed beyond the automated stuff. I reviewed a similar PR in the past using `meld`'s three-way comparison.

MarcelKoch:
Perhaps the easiest is if you already know where the 10% interesting stuff is, you could just give us a list of files and then we can manage.

upsj:
I had to do so many small changes, I can't really remember, I'll have to reproduce that in an automated fashion

MarcelKoch:
TBH, I'm not sure if the places where you did manual changes are that important. AFAIK the changes are mostly code movement, with the addition of perhaps some wrappers around the moved code. And I'm somewhat confident that our CI would catch errors on that level. If there are any changes besides that, it would be good to know if you can point us to those.
Otherwise, the question of review is rather on the concept itself and not on the individual changes.

upsj:
Link to review only the non-automated changes: https://github.com/ginkgo-project/ginkgo/pull/1516/files/774b3d6b9b7d6d99fb8357685424f8bf6b027f97..e6d1e5fe469574ac8edd6067c9b0316e2cf0fa15

For the first commit, best look at the Python script in the PR description

cuda/preconditioner/jacobi_advanced_apply_kernels.cu
```diff
@@ -7,7 +7,7 @@
 #include "core/matrix/dense_kernels.hpp"
 #include "core/preconditioner/jacobi_kernels.hpp"
 #include "core/synthesizer/implementation_selection.hpp"
-#include "cuda/preconditioner/jacobi_common.hpp"
+#include "preconditioner/jacobi_common.hpp"
```
yhmtsai, line 10: our header search path are all from the top level

hip/stop/batch_criteria.hip.hpp
```diff
@@ -0,0 +1,26 @@
...
+#ifndef GKO_HIP_STOP_BATCH_CRITERIA_HIP_HPP_
+#define GKO_HIP_STOP_BATCH_CRITERIA_HIP_HPP_
+
+
+#include <ginkgo/core/base/math.hpp>
+
+
+namespace gko {
+namespace kernels {
+namespace hip {
```
yhmtsai: why can't this be in cuda_hip?

hip/base/batch_multi_vector_kernels.hip.cpp
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * @brief The MultiVector matrix format namespace.
+ *
+ * @ingroup batch_multi_vector
+ */
+namespace batch_multi_vector {
+
+
+constexpr auto default_block_size = 256;
+constexpr int sm_oversubscription = 4;
```
yhmtsai: can't this be in cuda_hip with setting the config by macro

common/cuda_hip/log/batch_logger.hpp.inc
```diff
@@ -2,19 +2,6 @@
 //
 // SPDX-License-Identifier: BSD-3-Clause
 
-#ifndef GKO_COMMON_CUDA_HIP_LOG_BATCH_LOGGER_HPP_INC_
```
yhmtsai: I do not understand this. It makes hpp to hpp.inc.
I think the original should already be unified.

yhmtsai: do you give the wrong direction of changes?
I have seen the batch_logger.hpp.inc without the namespace in the develop

yhmtsai (commented):
The design of cuda_hip was to do copy/paste only. They intend not to work standalone.
In the beginning, we only allow the kernel implementation there. cuda/hip are distinguished from the path already.

hip/stop/batch_criteria.hip.hpp
upsj: The whole batch ecosystem has an extremely complex include structure, the untangling of which would make this PR much too complicated. I started with it and realized that it would be too much effort, so I reverted all changes related to batch

cuda/preconditioner/jacobi_advanced_apply_kernels.cu
upsj: They don't need to be - the jacobi_common headers differ slighly between CUDA and HIP, and are the only generated headers in the backend, so I believe we can treat them differently from the other headers. There might be a more stable solution, but I didn't attempt to find it for now. Alternatively we could attempt to do weird preprocessor things like `#include #GKO_DEVICE_NAMESPACE "preconditioner/jacobi_common.hpp"`, but I am worried about how portable that would be

hip/base/batch_multi_vector_kernels.hip.cpp
upsj: see [bath_criteria](https://github.com/ginkgo-project/ginkgo/pull/1516#discussion_r1661547031)

common/cuda_hip/log/batch_logger.hpp.inc
upsj: some changes were reverted because `batch` was too messy to unify directly without a lot of additional changes

common/cuda_hip/components/uninitialized_array.hpp
```diff
@@ -63,3 +75,11 @@ class uninitialized_array {
...
     unsigned char data_[sizeof(ValueType) / sizeof(unsigned char) * size];
 };
+
+
+}  // namespace GKO_DEVICE_NAMESPACE
+}  // namespace kernels
+}  // namespace gko
+
+
+#endif  // GKO_COMMON_CUDA_HIP_COMPONENTS_UNINITIALIZED_ARRAY_HPP_INC_
```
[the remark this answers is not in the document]
upsj: should be addressed in https://github.com/ginkgo-project/ginkgo/pull/1516/commits/1ee816494c6a77aa8319d260845ea78210f7b582

common/cuda_hip/base/math.hpp
```diff
@@ -37,3 +49,9 @@ struct truncate_type_impl<thrust::complex<T>> {
 
 
 }  // namespace detail
+
+
+}  // namespace gko
+
+
+#endif  // GKO_COMMON_CUDA_HIP_BASE_MATH_HPP_INC_
```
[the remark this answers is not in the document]
upsj: should be addressed in https://github.com/ginkgo-project/ginkgo/pull/1516/commits/1ee816494c6a77aa8319d260845ea78210f7b582

upsj:
Looking backwards, I think the major reason why we couldn't go further in unifying files then was that we didn't realize the chevron launch syntax `<<<...>>>` was supported since ROCm 2.8 already, which made having different files for CUDA and HIP necessary. The reason I'm trying to eliminate the `.inc` files is that they split up code across 3 files that should be the same file - earlier on it used to be kernel implementation vs. kernel launch and headers, now it is mainly kernel implementation + launch vs. headers. It is only a small step to pull the headers into the shared codebase as well.
With this change, the support for C++ tools is much improved, as they tend to separate files into headers and source files, neither of which the `.inc` files fit into. 

Looking at how much of the code can be unified with a simple script without and diffs, I think this shows that we are already in-facto there, even if we didn't put in the effort of unifying the files.

MarcelKoch:
Just a general comment, I think it would have been good if the python script itself would also have been part of the commits. This would allow to also comment on the script directly, which is not possible right now.

upsj:
@MarcelKoch I added the script, but will remove it again before merging

dev_tools/scripts/unify_cuda_hip.py
```diff
@@ -0,0 +1,135 @@
...
+)
+cuda_replaced = (
+    cuda_lines[:cuda_location] + common_lines + cuda_lines[cuda_location + 1 :]
+)
+hip_replaced = hip_lines[:hip_location] + common_lines + hip_lines[hip_location + 1 :]
+
+cuda_replaced = (
+    subprocess.run(
+        args=[
+            "/home/tribizel/.cache/pre-commit/repoay30okq9/py_env-python3/lib64/python3.9/site-packages/clang_format/data/bin/clang-format",
```
upsj, line 91: you can ignore this hard-coded path, the script will not stick around

MarcelKoch (changes requested):
I think this complicates the directory structure a bit more. My guess is that if we could fully unify both backends, then it would become clearer again. This would mean something like pulling up `common/cuda_hip` to the top-level.

What is missing for the last files (excluding batched) to be unified?
I think there are some files which could work already now, e.g. `ic_kernels`, but are not included, since there wasn't a matching file in `common`

upsj:
@MarcelKoch The list is rather extensive - most of `base`, all factorization preconditioners, CSR, triangular solvers, Jacobi, batch. Especially with the `base` stuff, I don't think that would be feasible. Do all of these changes need to happen in this PR? The other work that is going on makes this a rather moving target, I'm already happy to have made it this far without any major rebasing required.

MarcelKoch:
@upsj I didn't expect `base` to be part of that, so I was guessing it wouldn't be too many files (I'm already excluding batch, because that needs a rework anyway). But if that is not the case, then it should be done in a separate PR.
I would still like to see that done, because the directory structure is rather complicated right now.

MarcelKoch (approved):

upsj:
@MarcelKoch I agree, that would be my next goal

common/cuda_hip/log/batch_logger.hpp.inc
yhmtsai: it makes review a bit hard. Some are based on the changes which are not reviewed yet.

dev_tools/scripts/unify_cuda_hip.py
yhmtsai: you can use the `dev_tools/script/clang-formats.sh`, which can find the corresponding clang-format we used from pre-commit

upsj: this script will be removed before merging, I only added it so we can add review comments to it.

common/cuda_hip/log/batch_logger.hpp.inc
upsj: You can review it in two parts by excluding the reverting commit https://github.com/ginkgo-project/ginkgo/pull/1516/commits/f3c1a103951d54291d66e96aa1d2049b44b5f01f:

* https://github.com/ginkgo-project/ginkgo/pull/1516/files/1e5cdf70fa140f25d23236a38dbff833cd9c8f81..a52527011f8c6f80bbf581aad6d83ee5d004b7bc
* https://github.com/ginkgo-project/ginkgo/pull/1516/files/f3c1a103951d54291d66e96aa1d2049b44b5f01f..15ea33e393e6c014b772f5ee47c13f609a8146db

yhmtsai (changes requested):
two left in my mind: preconditioner/jacobi_common.hpp and header guard

common/cuda_hip/base/kernel_launch.hpp
```diff
@@ -2,6 +2,52 @@
 //
 // SPDX-License-Identifier: BSD-3-Clause
 
+#ifndef GKO_COMMON_UNIFIED_BASE_KERNEL_LAUNCH_HPP_
+#error \
+    "This file can only be used from inside common/unified/base/kernel_launch.hpp"
+#endif
+
+
+#include <thrust/tuple.h>
```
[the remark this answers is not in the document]
upsj: The file is not self-sufficient without the `unified/kernel_launch.hpp` header, just like before, so it does not need one.

cuda/preconditioner/jacobi_advanced_apply_kernels.cu
upsj: I'll revert the Jacobi-related changes for now, it's probably better to do this together with the other Jacobi kernels anyways.

upsj: Found a better solution - we can mirror the directory structure inside the `cuda` and `hip` build directory

hip/CMakeLists.txt
```diff
@@ -110,14 +84,14 @@ foreach(GKO_JACOBI_BLOCK_SIZE IN LISTS GKO_HIP_JACOBI_BLOCK_SIZES)
         ${CMAKE_CURRENT_BINARY_DIR}/preconditioner/jacobi_advanced_apply_kernels.instantiate.${GKO_JACOBI_BLOCK_SIZE}.hip.cpp)
 endforeach()
 string(REPLACE ";" "," GKO_HIP_JACOBI_BLOCK_SIZES_CODE "${GKO_HIP_JACOBI_BLOCK_SIZES}")
-configure_file(preconditioner/jacobi_common.hip.hpp.in preconditioner/jacobi_common.hip.hpp)
+configure_file(preconditioner/jacobi_common.hip.hpp.in common/cuda_hip/preconditioner/jacobi_common.hpp)
```
yhmtsai, line 87: the file from cuda and hip will overwrite each other when enable both.

upsj: no, this happens inside the `cuda` and `hip` subdirectory of your build directory. I tested the builds with a CUDA + ROCm environment the entire time.

yhmtsai: ah, okay. it is in the <ginkgo>/cuda/common/cuda_hip/preconditioner.hpp and <ginkgo>/hip/common/cuda_hip/preconditioner.hpp.
My concern is not solved by this.
I was thinking both are in <ginkgo>/common/cuda_hip/...

upsj: It is not clear to me what the actual concern is - can you please elaborate?

All solutions I've seen proposed so far are much more complicated - this solution preserves the directory structure and chooses a different file for each backend, addressing both concerns you mentioned previously. `#include` statements can't appear inside a macro, so the `CMAKE_DISPATCH` suggestion won't work.

yhmtsai: I think you try to unify the include path of the header by adding the current folder not the root/top folder as the search path.
I mean something like the `#include CMAKE_GENERATED_HEADER(real source file)`, but it is not necessary anyways.
The original one does not change the directory structure and is not complicated. 
I am confused about the location and it sounds unnecessary to me because those files using that are not unified.
I would like avoid exception if it is not necessary. (I call it exception just from the ginkgo experience)

upsj: This only concerns a single header file (and it doesn't look like we will add any more like this in the near future), so I think we should not start building anything complex around it. `jacobi_kernels.cpp` is already unified, and the other Jacobi files will be unified in a subsequent PR (the whole block size instantiation setup requires some more work that I wanted to keep separate).

yhmtsai: I see. I was missing the unified jacobi_kernels.cpp one.
Then, preconditioner/jacobi_common.hpp is okay to me with the requirement from the unified one.

yhmtsai (approved):

PR #1521, opened 2024-01-08
Update copyright to 2024

MarcelKoch:
Happy new year :fireworks: 

I think the format CI should fail, since it will use the old configuration files without the updated year.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,7 +14,7 @@ repos:
   # The official reuse hook only supports calling lint, so we need our own hook
   - id: reuse-annotate
     name: reuse-annotate
-    entry: reuse annotate --license BSD-3-Clause --copyright 'The Ginkgo authors' --year 2017-2023 --style c
+    entry: reuse annotate --license BSD-3-Clause --copyright 'The Ginkgo authors' --year 2024 --style c --merge-copyright
     language: python
     additional_dependencies: [reuse]
     types_or: [c, c++, cuda, inc]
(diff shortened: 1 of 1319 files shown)


yhmtsai (approved):

upsj:
The formatting patch shows a lot more changes than just the copyright year update, maybe we need to rerun it or tune the clang-format config?

MarcelKoch:
I think the extra formatting changes are because we didn't run the `format_header.sh` script on all files? The changes from clang-format are only 3 files, so most of that is actually from the script.

upsj (approved):

.pre-commit-config.yaml
```diff
@@ -14,7 +14,7 @@ repos:
   # The official reuse hook only supports calling lint, so we need our own hook
   - id: reuse-annotate
     name: reuse-annotate
-    entry: reuse annotate --license BSD-3-Clause --copyright 'The Ginkgo authors' --year 2017-2023 --style c
+    entry: reuse annotate --license BSD-3-Clause --copyright 'The Ginkgo authors' --year 2024 --style c --merge-copyright
```
tcojean, line 17: While we're at it, can we automate the year parameter in a platform portable way? e.g., in python something like
```
from datetime import datetime
datetime.now().strftime('%Y')
```
(maybe with timezone UTC)

MarcelKoch: I think this would be something that needed to be changed in reuse. I will see, if I can get a PR there.

PR #1523, opened 2024-01-08
Automatically use current year for license

MarcelKoch:
From @tcojean suggestion, this now uses the current year for the license. The documentation of reuse is not helpful here, since it doesn't clearly describe that it will use the current year if the `--year` option is *not* given.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,7 +14,7 @@ repos:
   # The official reuse hook only supports calling lint, so we need our own hook
   - id: reuse-annotate
     name: reuse-annotate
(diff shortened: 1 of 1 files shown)


tcojean (approved):
LGTM

upsj (approved):

upsj:
This also means that the updates to the license will be staggered (i.e. the year in a file is the year it changed last) which I'm totally fine with, but we should note it.

MarcelKoch:
@upsj we can still update all files yearly. With this change we just don't need to update the pre-commit config each year.

upsj:
I know, I'm just not sure if this is necessary, since the date is supposed to describe when the code was written, so it's fine to leave it unchanged on old files. No strong opinion on either solution though.

PR #1526, opened 2024-01-13
Add a uniform coarse generation algorithm 

pratikvn:
This PR is an updated version of #979 . The idea is to have a very cheap coarse grid generation algorithm that works well for matrices that are uniform, for example, matrices that arise from stencils on structured grids.

For example, for a 7 point stencil on a 3D grid, it is very competitive to PGM with much faster generation and faster apply for larger problems (on OMP executor with 8 cores):

unif_pgm


upsj:
Especially focusing on distributed, I think if we include this, we should try to support at least 2D and maybe 3D coarsening (`i + n * j`) and (`i + n * j + k * n * m`).

MarcelKoch:
Maybe it makes sense to split the `UniformCoarsening` class up. The generated object essentially stores an array with coarse row indices, so it seems natural to allow users to pass in these indices. But I don't think we can add a parameter to the existing class, since then there would be two mutually exclusive parameters.
I think Tobias' approach for the direct solvers, i.e. having multiple factories that generate the same class makes sense here.

reference/test/multigrid/uniform_coarsening_kernels.cpp
```diff
@@ -0,0 +1,445 @@
...
+};
+
+TYPED_TEST_SUITE(UniformCoarsening, gko::test::ValueIndexTypes,
+                 PairTypenameNameGenerator);
+
+
+TYPED_TEST(UniformCoarsening, CanBeCopied)
+{
+    using Mtx = typename TestFixture::Mtx;
+    using MgLevel = typename TestFixture::MgLevel;
```
MarcelKoch, line 179: unused (and in the other tests)

include/ginkgo/core/multigrid/uniform_coarsening.hpp
```diff
@@ -0,0 +1,141 @@
...
+ *
+ * @tparam ValueType  precision of matrix elements
+ * @tparam IndexType  precision of matrix indexes
+ *
+ * @ingroup MultigridLevel
+ * @ingroup Multigrid
+ * @ingroup LinOp
+ */
+template <typename ValueType = default_precision, typename IndexType = int32>
+class UniformCoarsening
```
MarcelKoch, line 37: Should this define copy/move assignment/constructors? The default ones are not completely inline with our normal understanding of those.

reference/test/multigrid/uniform_coarsening_kernels.cpp
```diff
@@ -0,0 +1,445 @@
...
+    auto c4_rows = gko::array<index_type>(
+        this->exec, {0, -1, -1, -1, 1, -1, -1, -1, 2, -1});
+    auto c5_rows = gko::array<index_type>(
+        this->exec, {0, -1, -1, -1, -1, 1, -1, -1, -1, -1});
+    auto c_rows = gko::array<index_type>(this->exec, 10);
+    c_rows.fill(-gko::one<index_type>());
+
+    gko::kernels::reference::uniform_coarsening::fill_incremental_indices(
+        this->exec, 2, &c_rows);
+    GKO_ASSERT_ARRAY_EQ(c_rows, c2_rows);
```
MarcelKoch, line 284: range-based for loop?

MarcelKoch (changes requested):
As I mentioned, I think the `UniformCoarsening` class should be split up. Other than that I have mostly smaller comments.

examples/multigrid-preconditioned-solver/multigrid-preconditioned-solver.cpp
```diff
@@ -73,24 +112,76 @@ int main(int argc, char* argv[])
...
+                                        .with_min_coarse_rows(10u)
+                                        .with_pre_smoother(smoother_gen)
+                                        .with_post_uses_pre(true)
+                                        .with_mg_level(mg_level_gen)
+                                        .with_coarsest_solver(coarsest_gen)
+                                        .with_criteria(iter_stop, tol_stop)
+                                        .on(exec));
+    if (coarse_type == std::string("uniform")) {
+        std::cout << "Using Uniform coarsening" << std::endl;
+        multigrid_gen = gko::share(mg::build()
```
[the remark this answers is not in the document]
pratikvn: This just builds the factories, no actual generation is done.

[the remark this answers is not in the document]
MarcelKoch: doesn't really matter, it's unnecessary duplication.

pratikvn:
@upsj, After a bit of thought, I think this class should not represent Geometric/Structured coarsening. Uniform coarsening is just a cheap coarsening method that just uniformly discards rows and columns to construct the coarse matrix, and hence is not suitable for Geometric coarsening, which needs to create connections between the coarse nodes.

@MarcelKoch , we already provide the users with a way to specify custom indices for coarsening with the [`FixedCoarsening` class](https://github.com/ginkgo-project/ginkgo/blob/8a31cf8317ca9c5223fe2ce86990a9293a8163a7/include/ginkgo/core/multigrid/fixed_coarsening.hpp#L69)

upsj:
@pratikvn If we discard variables for restriction, we still need to interpolate new values for prolongation, so I don't yet see how this makes multi-dimensional coarsening impossible if 1D coarsening is possible?

pratikvn:
In a similar fashion to PGM, the prolongation matrix will be a transpose of the restriction matrix. I think maybe I have given a wrong impression that this would be a 1D coarsening. The fine level vectors will not have extrapolated values, but just "mapped" values.

upsj:
Doesn't PGM do the same thing? And I believe we are mixing two things here - coarsening and interpolation. One is structural, one concerns the actual numerical values.

pratikvn:
Just adding more performance results on a 7pt stencil and some chosen matrices from Suitesparse, to show that this type of coarsening can help in certain cases.

### Apply speedup for cage15 (v/s PGM) matrix and bcsstm39 matrix (v/s no preconditioning)
[speedup_app_ss_bicgstab_cskip.pdf](https://github.com/user-attachments/files/16267051/speedup_app_ss_cg_cskip.pdf)

### Apply speedup for 7pt stencil (v/s PGM)
[speedup_app_7pt_cg_cskip.pdf](https://github.com/user-attachments/files/16267094/speedup_app_7pt_cg_cskip.pdf)

### Generate speedup for 7pt stencil (v/s PGM)
[speedup_gen_7pt_cg_cskip.pdf](https://github.com/user-attachments/files/16267105/speedup_gen_7pt_cg_cskip.pdf)

upsj:
I don't think these results matter, since even for simple stencils, the matrix loses all connectivity:

Take the poisson-solver example and plug in the following code:

```cpp
    auto gen = gko::multigrid::UniformCoarsening<ValueType, IndexType>::build().on(app_exec)->generate(matrix->clone());
    gko::write(std::cout, matrix);
    gko::write(std::cout, gko::as<mtx>(gen->get_composition()->get_operators()[1]));
```

The fine matrix is tridiagonal, but the coarse matrix is just a diagonal matrix.

upsj:
I don't see this PR being useful in its current state. IMO we should either turn it into a regular coarsening algorithm, or abandon it.

PR #1458, opened 2023-11-13
Add scaling operations for batch matrices

pratikvn:
~~This PR adds a batch::Diagonal matrix format, core and device kernels for the `apply(MultiVector*, MultiVector*)`. This is useful when scaling matrices and vectors that are necessary in some applications.~~

This PR now only adds the kernels and free standing functions for two-sided scaling.  

### TODO
+ [x] Add scaling kernels for batched matrices
+ [x] Check docs
+ [x] Add in_place apply core function 
+ [x] Add row, col scaling and update batch solvers to enable scaling.

PR #1527, opened 2024-01-13
Add scaling for batch objects

pratikvn:
This PR pulls out a part of functionality from #1458 to ease the review process. It mainly adds scaling (row and column) for batched matrices and multi-vectors, which is useful in many applications, particularly SUNDIALS and XGC.


common/cuda_hip/matrix/batch_csr_kernels.hpp.inc
```diff
@@ -111,3 +111,47 @@ __global__ __launch_bounds__(
...
+    const ValueType* const __restrict__ row_scale,
+    const IndexType* const __restrict__ col_idxs,
+    const IndexType* const __restrict__ row_ptrs,
+    ValueType* const __restrict__ values)
+{
+    constexpr auto warp_size = config::warp_size;
+    const auto tile =
+        group::tiled_partition<warp_size>(group::this_thread_block());
+    const int tile_rank = threadIdx.x / warp_size;
+    const int num_tiles = (blockDim.x - 1) / warp_size + 1;
```
MarcelKoch, line 128: I don't understand what this is. From the name, `blockDim.x / warp_size` makes much more sense. For the default block size, this is basically `(blockDim / warp_size) - 1`, but if you were to use `blockDim = 512` then you would have `(blockDIm / warp_size) - 2`, which seems inconsistent to me.

common/cuda_hip/matrix/batch_ell_kernels.hpp.inc
```diff
@@ -125,3 +125,43 @@ __global__ __launch_bounds__(
...
 }
+
+
+template <typename ValueType, typename IndexType>
+__device__ __forceinline__ void scale(
+    const ValueType* const __restrict__ col_scale,
+    const ValueType* const __restrict__ row_scale,
+    const gko::batch::matrix::ell::batch_item<ValueType, IndexType>& mat)
+{
+    for (int tidx = threadIdx.x; tidx < mat.num_rows; tidx += blockDim.x) {
```
MarcelKoch, line 136: Csr uses a warp-per-row parallelization, why does this use thread-per-row again?

core/matrix/batch_dense.cpp
```diff
@@ -189,6 +191,28 @@ void Dense<ValueType>::move_to(Dense<next_precision<ValueType>>* result)
 GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE(GKO_DECLARE_BATCH_DENSE_MATRIX);
 
 
+template <typename ValueType>
+void two_sided_scale(const array<ValueType>& col_scale,
+                     const array<ValueType>& row_scale,
+                     batch::matrix::Dense<ValueType>* in_out)
+{
+    GKO_ASSERT_EQ(col_scale.get_size(), (in_out->get_common_size()[1] *
+                                         in_out->get_num_batch_items()));
```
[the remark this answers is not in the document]
MarcelKoch: Maybe something similar to the internal batch structs would be better suited.

core/test/solver/batch_bicgstab.cpp
```diff
@@ -245,6 +245,38 @@ TYPED_TEST(BatchBicgstab, CanSetTolType)
...
+
+    auto solver_factory = Solver::build()
+                              .with_max_iterations(22)
+                              .with_tolerance(static_cast<real_type>(0.25))
+                              .with_col_scaling(col_scale)
+                              .with_row_scaling(row_scale)
+                              .on(this->exec);
+    auto solver = solver_factory->generate(this->mtx);
+
+    ASSERT_EQ(solver->get_parameters().max_iterations, 22);
```
MarcelKoch, line 267: nit: IMO, this test needs to only care about the scaling vectors and can ignore all other parameters.

reference/test/matrix/batch_csr_kernels.cpp
```diff
@@ -165,6 +165,46 @@ TYPED_TEST(Csr, ConstAppliesLinearCombinationToBatchMultiVector)
...
+
+    auto scaled_mtx_0 =
+        gko::batch::initialize<BMtx>({{{6.0, -6.0, 0.0}, {-12.0, 12.0, 18.0}},
+                                      {{6.0, -12.0, 0.0}, {6.0, -15.0, 24.0}}},
+                                     this->exec, 5);
+    GKO_ASSERT_BATCH_MTX_NEAR(this->mtx_0.get(), scaled_mtx_0.get(), 0.);
+}
+
+
+TYPED_TEST(Csr, CanTwoSidedScaleWithDifferentAlpha)
```
MarcelKoch, line 189: What does `alpha` mean here?

reference/test/matrix/batch_csr_kernels.cpp
```diff
@@ -165,6 +165,46 @@ TYPED_TEST(Csr, ConstAppliesLinearCombinationToBatchMultiVector)
...
+{
+    using value_type = typename TestFixture::value_type;
+    using index_type = gko::int32;
+    using BMtx = typename TestFixture::BMtx;
+    auto col_scale = gko::array<value_type>(this->exec, 3 * 2);
+    auto row_scale = gko::array<value_type>(this->exec, 2 * 2);
+    col_scale.fill(2);
+    row_scale.fill(3);
+
+    gko::batch::matrix::two_sided_scale<value_type, index_type>(
```
MarcelKoch, line 178: The template arguments should not be necessary. Same for all other tests.

test/matrix/batch_csr_kernels.cpp
```diff
@@ -80,16 +89,20 @@ class Csr : public CommonTestFixture {
     std::default_random_engine rand_engine;
 
     const gko::size_type batch_size = 11;
-    std::unique_ptr<BMtx> mat;
+    std::shared_ptr<BMtx> mat;
```
MarcelKoch, line 92: why are this and `dmat` now `shared_ptrs`?

test/matrix/batch_dense_kernels.cpp
```diff
@@ -64,16 +76,22 @@ class Dense : public CommonTestFixture {
     std::default_random_engine rand_engine;
 
     const gko::size_type batch_size = 11;
-    std::unique_ptr<BMtx> mat;
+    std::shared_ptr<BMtx> mat;
+    std::shared_ptr<BMtx> mat2;
```
MarcelKoch, line 80: `mat2` and `dmat2` are not used.

MarcelKoch (dismissed):
The parallelization approach seems inconsistent to me, i.e. when warp-per-row and thread-per-row are used. Of course this might be due to a better performance of one or the other, I just wanted to check on that.

Also the reference tests for the multivector changes are missing.

common/cuda_hip/matrix/batch_csr_kernels.hpp.inc
MarcelKoch: Sorry, from the name `num_tiles = ceildiv(blockDim.x, warp_size)` makes more sense.

include/ginkgo/core/matrix/batch_dense.hpp
```diff
@@ -338,6 +349,21 @@ class Dense final : public EnableBatchLinOp<Dense<ValueType>>,
...
+ * @param col_scale  the column scalars
+ * @param row_scale  the row scalars
+ * @param in_out  the matrix to be scaled
+ *
+ * @note the operation is performed in-place
+ */
+template <typename ValueType>
+void two_sided_scale(const array<ValueType>& col_scale,
+                     const array<ValueType>& row_scale,
+                     batch::matrix::Dense<ValueType>* in_out);
```
yhmtsai, line 364: ptr_param?

reference/test/matrix/batch_dense_kernels.cpp
```diff
@@ -67,7 +67,7 @@ class Dense : public ::testing::Test {
     {}
 
     std::shared_ptr<const gko::ReferenceExecutor> exec;
-    std::unique_ptr<BMtx> mtx_0;
+    std::shared_ptr<BMtx> mtx_0;
```
yhmtsai, line 70: the unique_ptr should still work with the new tests

test/matrix/batch_dense_kernels.cpp
```diff
@@ -43,17 +44,28 @@ class Dense : public CommonTestFixture {
             std::normal_distribution<>(-1.0, 1.0), rand_engine, ref);
     }
 
-    void set_up_apply_data(gko::size_type num_rows, gko::size_type num_vecs = 1)
+    void set_up_apply_data(gko::size_type num_rows,
+                           gko::size_type num_cols = 32,
+                           gko::size_type num_vecs = 1)
```
yhmtsai, line 49: any usage for this changes?

yhmtsai (changes requested):

include/ginkgo/core/matrix/batch_csr.hpp
```diff
@@ -333,6 +331,21 @@ class Csr final
...
+ * @param col_scale  the column scalars
+ * @param row_scale  the row scalars
+ * @param in_out  the matrix to be scaled
+ *
+ * @note the operation is performed in-place
+ */
+template <typename ValueType, typename IndexType>
+void two_sided_scale(const array<ValueType>& col_scale,
+                     const array<ValueType>& row_scale,
+                     batch::matrix::Csr<ValueType, IndexType>* in_out);
```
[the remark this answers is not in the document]
pratikvn: I think for scaling, `array` makes more sense. Semantically, storing them in vectors/dense is misleading IMO. A diagonal matrix makes sense, but that would mean you have to use `apply`, which we wanted to go away from.

test/matrix/batch_dense_kernels.cpp
pratikvn: Its more general, but will be useful in https://github.com/ginkgo-project/ginkgo/pull/1528

include/ginkgo/core/matrix/batch_dense.hpp
```diff
@@ -338,6 +349,21 @@ class Dense final : public EnableBatchLinOp<Dense<ValueType>>,
...
+ *
+ * @param col_scale  the column scalars
+ * @param row_scale  the row scalars
+ * @param in_out  the matrix to be scaled
+ *
+ * @note the operation is performed in-place
+ */
+template <typename ValueType>
+void two_sided_scale(const array<ValueType>& col_scale,
+                     const array<ValueType>& row_scale,
```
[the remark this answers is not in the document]
pratikvn: I think we wanted to avoid new Interface classes and rather have free functions instead ?

common/cuda_hip/matrix/batch_ell_kernels.hpp.inc
pratikvn: Csr matrices tend to be more unbalanced, while Ell matrices are balanced.

core/matrix/batch_dense.cpp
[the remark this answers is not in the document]
pratikvn: I think `batch::array` is not worth it, right now. Maybe for non-uniform batches. I think if there are more use cases, we can consider it for the future.

common/cuda_hip/base/batch_multi_vector_kernels.hpp.inc
```diff
@@ -32,6 +33,45 @@ __launch_bounds__(default_block_size, sm_oversubscription) void scale_kernel(
 }
 
 
+template <typename ValueType>
+__device__ __forceinline__ void elem_wise_scale(
```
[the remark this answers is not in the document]
pratikvn: good idea!

include/ginkgo/core/solver/batch_solver_base.hpp
```diff
@@ -188,6 +191,16 @@ struct enable_preconditioned_iterative_solver_factory_parameters
      */
     std::shared_ptr<const BatchLinOp> GKO_FACTORY_PARAMETER_SCALAR(
         generated_preconditioner, nullptr);
+
+    /**
+     * Column scaling vector
+     */
+    array<ValueType> GKO_FACTORY_PARAMETER_SCALAR(col_scaling, {});
```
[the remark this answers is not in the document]
pratikvn: I think it is better to have tolerance in double. I dont see a necessary reason for it be `ValueType` and depend on the class parameter

include/ginkgo/core/solver/batch_solver_base.hpp
```diff
@@ -278,6 +291,31 @@ class EnableBatchSolver
...
+        using batch_ell = batch::matrix::Ell<value_type>;
+
+        if (params.col_scaling.get_executor() &&
+            params.row_scaling.get_executor()) {
+            GKO_ASSERT_EQ(params.col_scaling.get_size(),
+                          system_matrix->get_common_size()[0] *
+                              system_matrix->get_num_batch_items());
+            GKO_ASSERT_EQ(params.col_scaling.get_size(),
+                          params.row_scaling.get_size());
+            if (auto mat = as<batch_dense>(system_matrix)) {
```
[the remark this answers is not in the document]
pratikvn: You would either have to `dynamic_cast` in the helper file or here. An other option would be to create an interface class, and override that in the matrix classes, but I think we wanted to move away from that. I think having them as free functions and checking the types here is a reasonable approach.

include/ginkgo/core/matrix/batch_dense.hpp
pratikvn: I dont think there is a particular advantage in this case of `ptr_param` ?

MarcelKoch: I second Mike's suggestion. This should use `ptr_param` since it doesn't involve ownership.

core/test/solver/batch_bicgstab.cpp
```diff
@@ -245,6 +245,34 @@ TYPED_TEST(BatchBicgstab, CanSetTolType)
...
+    using real_type = typename TestFixture::real_type;
+    auto scale_size = this->num_batch_items * this->num_rows;
+    auto col_scale = gko::array<value_type>(this->exec, scale_size);
+    col_scale.fill(0.5);
+    auto row_scale = gko::array<value_type>(this->exec, scale_size);
+    row_scale.fill(0.8);
+
+    auto solver_factory = Solver::build()
+                              .with_max_iterations(22)
+                              .with_tolerance(static_cast<real_type>(0.25))
```
MarcelKoch, line 261: nit: I would not set these, since they are not relevant for this test.

include/ginkgo/core/solver/batch_solver_base.hpp
```diff
@@ -278,6 +291,28 @@ class EnableBatchSolver
...
+                matrix::scale_in_place(params.col_scaling, params.row_scaling,
+                                       const_cast<batch_dense*>(mat.get()));
```
MarcelKoch, line 307: I think this is actual UB. I guess the cleanest way is to just make the constructor take a mutable pointer for the system matrix. That would also reflect the usage more honestly.

test/matrix/batch_csr_kernels.cpp
MarcelKoch: Is this going to be reverted, or was there an actual reason for this?

reference/test/matrix/batch_ell_kernels.cpp
```diff
@@ -167,6 +167,44 @@ TYPED_TEST(Ell, ConstAppliesLinearCombinationToBatchMultiVector)
 }
 
 
+TYPED_TEST(Ell, CanTwoSidedScale)
+{
+    using value_type = typename TestFixture::value_type;
+    using index_type = gko::int32;
```
MarcelKoch, line 173: unused
```suggestion
```

reference/test/matrix/batch_ell_kernels.cpp
```diff
@@ -167,6 +167,44 @@ TYPED_TEST(Ell, ConstAppliesLinearCombinationToBatchMultiVector)
...
+                                      {{6.0, -12.0, -3.0}, {6.0, -15.0, 24.0}}},
+                                     this->exec);
+    GKO_ASSERT_BATCH_MTX_NEAR(this->mtx_0.get(), scaled_mtx_0.get(), 0.);
+}
+
+
+TYPED_TEST(Ell, CanTwoSidedScaleWithDifferentValues)
+{
+    using value_type = typename TestFixture::value_type;
+    using index_type = gko::int32;
```
MarcelKoch, line 193: same

reference/test/matrix/batch_csr_kernels.cpp
```diff
@@ -165,6 +165,44 @@ TYPED_TEST(Csr, ConstAppliesLinearCombinationToBatchMultiVector)
...
+                                      {{6.0, -12.0, 0.0}, {6.0, -15.0, 24.0}}},
+                                     this->exec, 5);
+    GKO_ASSERT_BATCH_MTX_NEAR(this->mtx_0.get(), scaled_mtx_0.get(), 0.);
+}
+
+
+TYPED_TEST(Csr, CanTwoSidedScaleWithDifferentValues)
+{
+    using value_type = typename TestFixture::value_type;
+    using index_type = gko::int32;
```
MarcelKoch, line 191: same

reference/test/matrix/batch_csr_kernels.cpp
```diff
@@ -165,6 +165,44 @@ TYPED_TEST(Csr, ConstAppliesLinearCombinationToBatchMultiVector)
 }
 
 
+TYPED_TEST(Csr, CanTwoSidedScale)
+{
+    using value_type = typename TestFixture::value_type;
+    using index_type = gko::int32;
```
MarcelKoch, line 171: same

pratikvn:
@MarcelKoch and @yhmtsai, I decided to make `scale` a member function and remove the functionality from the solver entirely. After re-thinking this, I think there is no need to have it within the solver as it only brings complications and it would be a separate kernel anyways, so no real performance optimizations are possible. I think this approach should be much cleaner. The user now decides if they want to scale the matrix and do it before hand. Maybe adding a inverse_scale for the batched MultiVector makes sense.

common/cuda_hip/base/batch_multi_vector_kernel_launcher.hpp.inc
```diff
@@ -12,17 +12,42 @@ void scale(std::shared_ptr<const DefaultExecutor> exec,
...
+            [] __device__(int row, int col, int stride) { return col; });
     }
 }
 
 GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE(
     GKO_DECLARE_BATCH_MULTI_VECTOR_SCALE_KERNEL);
 
 
+template <typename ValueType>
+void element_wise_scale(std::shared_ptr<const DefaultExecutor> exec,
```
yhmtsai, line 34: Is it unused now?

omp/matrix/batch_dense_kernels.cpp
```diff
@@ -83,6 +83,31 @@ GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE(
...
+    const auto num_cols = static_cast<int>(input->get_common_size()[1]);
+    const auto stride = input->get_common_size()[1];
+#pragma omp parallel for
+    for (size_type batch_id = 0; batch_id < input->get_num_batch_items();
+         ++batch_id) {
+        const auto col_scale_b = col_scale_vals + num_cols * batch_id;
+        const auto row_scale_b = row_scale_vals + num_rows * batch_id;
+        const auto input_mat =
+            input_vals + input->get_num_elements_per_item() * batch_id;
+        scale(num_rows, num_cols, stride, col_scale_b, row_scale_b, input_mat);
```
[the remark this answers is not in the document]
yhmtsai: I was thinking in the diagonal matrix fashion. row scaling is diag(R) * A and col scaling is A * diag(C) and the order is more close to the dim ordering

yhmtsai (approved):
I agree with scaling in solver matrix only is a bit complicated especially when it change the const data. I also left some comments in the threads of previous comments about the thought about the row_scale and col_scale ordering

include/ginkgo/core/solver/batch_solver_base.hpp
```diff
@@ -149,7 +149,7 @@ class BatchSolver {
 };
 
 
-template <typename Parameters, typename Factory>
+template <typename ValueType, typename Parameters, typename Factory>
```
MarcelKoch, line 152: The new parameter is not needed anymore, right?

include/ginkgo/core/matrix/batch_ell.hpp
```diff
@@ -18,6 +18,7 @@
 #include <ginkgo/core/base/range_accessors.hpp>
 #include <ginkgo/core/base/types.hpp>
 #include <ginkgo/core/base/utils.hpp>
+#include <ginkgo/core/base/utils_helper.hpp>
```
MarcelKoch, line 21: not needed here?

reference/test/matrix/batch_csr_kernels.cpp
```diff
@@ -69,7 +69,7 @@ class Csr : public ::testing::Test {
     {}
 
     std::shared_ptr<const gko::ReferenceExecutor> exec;
-    std::unique_ptr<BMtx> mtx_0;
+    std::shared_ptr<BMtx> mtx_0;
```
MarcelKoch, line 72: Is this necessary?

reference/test/matrix/batch_dense_kernels.cpp
MarcelKoch, line 70: same

MarcelKoch (changes requested):
Mostly some smaller stuff left. The only larger question is the argument order, that Mike brought up. Although the argument names are clear, their order might be unexpected. The scale operation does `A = diag(r) * A * diag(c)`, but the arguments are `scale(c, r)`. Maybe a consistent ordering would be better, but it's not required if its already used in this order.

include/ginkgo/core/matrix/batch_ell.hpp
pratikvn: Used for `ptr_param`

pratikvn:
I somehow miscalculated the scaling apply. I now changed the user-facing parameter order to (row_scale, col_scale).

MarcelKoch (approved):

PR #1528, opened 2024-01-14
Add `add_scaled_identity` and `scale_add` for batched matrices

pratikvn:
This PR adds `add_scaled_identity` and `scale_add` functionality for batched matrices. Currently for simplicity, `scale_add` is enabled only for `batch::Dense` matrix format.

## TODO

+ [x] Merge #1527 
+ [x] Add device kernels


reference/test/matrix/batch_csr_kernels.cpp
```diff
@@ -201,6 +201,29 @@ TYPED_TEST(Csr, CanTwoSidedScaleWithDifferentValues)
 }
 
 
+TYPED_TEST(Csr, CanAddScaledIdentity)
+{
+    using value_type = typename TestFixture::value_type;
+    using index_type = gko::int32;
```
MarcelKoch, line 207: unused
```suggestion
```

MarcelKoch (changes requested):
Mostly smaller comment. I think the not implemented functions should not be part of this PR.
Also, I don't think there is any benefit of basing this PR on the other scaling PR. It only introduces leftover changes.

core/matrix/batch_csr.cpp
```diff
@@ -181,6 +183,29 @@ void Csr<ValueType, IndexType>::apply_impl(const MultiVector<ValueType>* alpha,
 }
 
 
+template <typename ValueType, typename IndexType>
+void Csr<ValueType, IndexType>::scale_add(
+    ptr_param<const MultiVector<ValueType>> alpha,
+    ptr_param<const batch::matrix::Csr<ValueType, IndexType>> b)
+{
+    GKO_NOT_IMPLEMENTED;
```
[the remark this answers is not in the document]
MarcelKoch: I don't think that is relevant at all here. We use `GKO_NOT_IMPLEMENTED` because some function has to be defined, maybe due to interface requirements or our kernel structure. But none of that applies here. This function simply doesn't need to exist, so there is zero use to declare and define this function. 
If we want to implement it at a later point we can still do so, nothing would stop us if we remove the function right now.

include/ginkgo/core/matrix/batch_csr.hpp
```diff
@@ -265,6 +265,29 @@ class Csr final
...
+        ptr_param<const MultiVector<value_type>> alpha,
+        ptr_param<const batch::matrix::Csr<value_type, index_type>> b);
+
+    /**
+     * Performs the operation this = alpha*I + beta*this.
+     *
+     * @param alpha the scalar for identity
+     * @param beta  the scalar to multiply this matrix
+     *
+     * @note Performs the operation in-place for this batch matrix
```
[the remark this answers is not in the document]
pratikvn: good catch. I added a diag_check now.

include/ginkgo/core/matrix/batch_dense.hpp
```diff
@@ -272,6 +283,28 @@ class Dense final : public EnableBatchLinOp<Dense<ValueType>>,
...
+    /**
+     * Performs the operation this = alpha*this + b.
+     *
+     * @param alpha the scalar to multiply this matrix
+     * @param b  the matrix to add
+     *
+     * @note Performs the operation in-place for this batch matrix
+     */
+    void scale_add(ptr_param<const MultiVector<value_type>> alpha,
+                   ptr_param<const batch::matrix::Dense<value_type>> b);
```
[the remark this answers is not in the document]
pratikvn: I put the parameter to reflect the name of the function, which scales `this` and then adds `b`. `add_scaled` adds a scaled `b`, so the parameter order (alpha, b) is appropriate there. AFAIR, blas does not have a `scale_add`, but you need to do a scale separately and then perform an addition with axpy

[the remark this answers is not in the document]
MarcelKoch: I was just thinking if we want to have `add_scaled` and `scale_add` to have different signatures. Of course this is not an issue right now, but perhaps it makes sense to think about it.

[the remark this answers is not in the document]
pratikvn: Yeah, that is a valid point. But maybe given that the two functions do different operations, it is probably better to distinguish them with different signatures as well.

reference/test/matrix/batch_csr_kernels.cpp
```diff
@@ -201,6 +201,27 @@ TYPED_TEST(Csr, CanTwoSidedScaleWithDifferentValues)
 }
 
 
+TYPED_TEST(Csr, CanAddScaledIdentity)
```
MarcelKoch, line 204: If you feel like it, you could add a test for a matrix that doesn't have a diagonal (same for ell)

MarcelKoch (approved):

common/cuda_hip/matrix/batch_dense_kernels.hpp.inc
```diff
@@ -170,3 +170,74 @@ __global__ void scale_kernel(
...
+    const gko::batch::matrix::dense::batch_item<const ValueType>& mat,
+    const gko::batch::matrix::dense::batch_item<ValueType>& in_out)
+{
+    for (int iz = threadIdx.x; iz < mat.num_rows * mat.num_cols;
+         iz += blockDim.x) {
+        const int row = iz / mat.stride;
+        const int col = iz % mat.stride;
+        in_out.values[row * in_out.stride + col] =
+            alpha * in_out.values[row * in_out.stride + col] +
+            mat.values[row * mat.stride + col];
```
yhmtsai, line 187:
```suggestion
        const int row = iz / mat.num_cols;
        const int col = iz % mat.num_cols;
        in_out.values[row * in_out.stride + col] =
            alpha * in_out.values[row * in_out.stride + col] +
            mat.values[row * mat.stride + col];
```
if you would like to use stride, need to ensure col is less than mat.num_cols and `iz < mat.num_rows * mat.stride`
I think you do not support stride > num_cols, so maybe add the todo note for tracking at least

common/cuda_hip/matrix/batch_ell_kernels.hpp.inc
```diff
@@ -165,3 +165,43 @@ __global__ void scale_kernel(
...
+            const auto ind = tidx + idx * mat.stride;
+            mat.values[ind] *= beta;
+            const auto col_idx = mat.col_idxs[ind];
+            if (col_idx == invalid_index<IndexType>()) {
+                break;
+            } else {
+                if (tidx == col_idx) {
+                    mat.values[ind] += alpha;
+                }
+            }
```
yhmtsai, line 186: nit: you can combine them to `else if`

core/matrix/batch_csr.cpp
```diff
@@ -197,6 +201,31 @@ void Csr<ValueType, IndexType>::scale(const array<ValueType>& row_scale,
...
+    GKO_ASSERT_BATCH_EQUAL_NUM_ITEMS(alpha, beta);
+    GKO_ASSERT_BATCH_EQUAL_NUM_ITEMS(this, beta);
+    GKO_ASSERT_EQUAL_DIMENSIONS(alpha->get_common_size(), gko::dim<2>(1, 1));
+    GKO_ASSERT_EQUAL_DIMENSIONS(beta->get_common_size(), gko::dim<2>(1, 1));
+    auto exec = this->get_executor();
+
+    auto csr_mat = this->create_const_view_for_item(0);
+
+    bool has_diags{false};
+    exec->run(csr::make_check_diagonal_entries(csr_mat.get(), has_diags));
```
yhmtsai, line 218: nice trick!

core/matrix/batch_csr.cpp
```diff
@@ -197,6 +201,31 @@ void Csr<ValueType, IndexType>::scale(const array<ValueType>& row_scale,
...
+    if (!has_diags) {
+        GKO_UNSUPPORTED_MATRIX_PROPERTY(
+            "The matrix has one or more structurally zero diagonal entries!");
```
yhmtsai, line 221: from the kernel, it only detects the missing diagonal issue

core/matrix/batch_csr.cpp
```diff
@@ -197,6 +201,31 @@ void Csr<ValueType, IndexType>::scale(const array<ValueType>& row_scale,
...
+{
+    GKO_ASSERT_BATCH_EQUAL_NUM_ITEMS(alpha, beta);
+    GKO_ASSERT_BATCH_EQUAL_NUM_ITEMS(this, beta);
+    GKO_ASSERT_EQUAL_DIMENSIONS(alpha->get_common_size(), gko::dim<2>(1, 1));
+    GKO_ASSERT_EQUAL_DIMENSIONS(beta->get_common_size(), gko::dim<2>(1, 1));
+    auto exec = this->get_executor();
+
+    auto csr_mat = this->create_const_view_for_item(0);
+
+    bool has_diags{false};
```
yhmtsai, line 217:
```suggestion
    bool has_all_diags{false};
```
nit

dpcpp/matrix/batch_dense_kernels.hpp.inc
```diff
@@ -88,3 +88,55 @@ __dpct_inline__ void scale_kernel(
...
+    constexpr auto tile_size = config::warp_size;
+    auto subg =
+        group::tiled_partition<tile_size>(group::this_thread_block(item_ct1));
+    const auto subgroup = static_cast<sycl::sub_group>(subg);
+    const int subgroup_id = subgroup.get_group_id();
+    const int subgroup_size = subgroup.get_local_range().size();
+    const int num_subgroup = subgroup.get_group_range().size();
+
+    for (int row = subgroup_id; row < mat.num_rows; row += num_subgroup) {
+        for (int col = subgroup.get_local_id(); col < mat.num_cols;
```
yhmtsai, line 134: same as cuda/hip

dpcpp/matrix/batch_ell_kernels.hpp.inc
```diff
@@ -66,3 +66,26 @@ __dpct_inline__ void scale_kernel(
...
+        for (auto k = 0; k < mat.num_stored_elems_per_row; k++) {
+            auto col_idx = mat.col_idxs[row + mat.stride * k];
+            mat.values[row + k * mat.stride] *= beta;
+            if (col_idx == invalid_index<IndexType>()) {
+                break;
+            } else {
+                if (row == col_idx) {
+                    mat.values[row + k * mat.stride] += alpha;
+                }
+            }
```
yhmtsai, line 88: same nit

include/ginkgo/core/matrix/batch_csr.hpp
```diff
@@ -274,6 +274,19 @@ class Csr final
...
+     * @note This operation fails in case this matrix does not have all its
+     *       diagonal entries.
+     */
+    void add_scaled_identity(ptr_param<const MultiVector<value_type>> alpha,
```
yhmtsai, line 287: My first sight of `add_scaled_identity` is `this +=  scaled * I` but I do not have a good idea on it yet

reference/matrix/batch_dense_kernels.hpp.inc
```diff
@@ -72,3 +72,36 @@ inline void scale(const int num_rows, const int num_cols,
...
+
+
+template <typename ValueType>
+inline void add_scaled_identity_kernel(
+    const ValueType alpha, const ValueType beta,
+    const gko::batch::matrix::dense::batch_item<ValueType>& mat)
+{
+    for (int row = 0; row < mat.num_rows; row++) {
+        for (int col = 0; col < mat.num_cols; col++) {
+            auto nnz = row * mat.stride + col;
```
yhmtsai, line 100: same here

reference/test/matrix/batch_csr_kernels.cpp
```diff
@@ -201,6 +201,43 @@ TYPED_TEST(Csr, CanTwoSidedScaleWithDifferentValues)
...
+{
+    using BMtx = typename TestFixture::BMtx;
+    using BMVec = typename TestFixture::BMVec;
+    auto alpha = gko::batch::initialize<BMVec>({{2.0}, {-1.0}}, this->exec);
+    auto beta = gko::batch::initialize<BMVec>({{3.0}, {-2.0}}, this->exec);
+    auto mat = gko::batch::initialize<BMtx>(
+        {{{1.0, 2.0, 0.0}, {3.0, 1.0, 1.0}, {0.0, 1.0, 1.0}},
+         {{2.0, -2.0, 0.0}, {1.0, -1.0, 2.0}, {0.0, 2.0, 1.0}}},
```
yhmtsai, line 212: also the rectangular case

test/matrix/batch_csr_kernels.cpp
```diff
@@ -137,3 +147,14 @@ TEST_F(Csr, TwoSidedScaleIsEquivalentToRef)
 
     GKO_ASSERT_BATCH_MTX_NEAR(dmat, mat, r<value_type>::value);
 }
+
+
+TEST_F(Csr, AddScaledIdentityIsEquivalentToRef)
+{
+    set_up_apply_data(2, 5, 151, 151, true);
```
yhmtsai, line 154: rectangular case

yhmtsai (dismissed):

common/cuda_hip/matrix/batch_ell_kernels.hpp.inc
pratikvn: IMO this is easier to read. The compiler should combine them so you should not lose any performance.

test/matrix/batch_dense_kernels.cpp
```diff
@@ -133,3 +137,25 @@ TEST_F(Dense, TwoSidedScaleIsEquivalentToRef)
...
+    set_up_apply_data(42, 42, 15);
+
+    mat->scale_add(alpha, mat2);
+    dmat->scale_add(dalpha, dmat2);
+
+    GKO_ASSERT_BATCH_MTX_NEAR(dmat, mat, r<value_type>::value);
+}
+
+
+TEST_F(Dense, AddScaledIdentityIsEquivalentToRef)
```
yhmtsai, line 153: miss rectangular

yhmtsai (approved):
miss some rectangular cases. others LGTM.
Some comments not for this pr: I feel `generate_diag_dominant_batch_matrix` a bit complex due to the nonzeros or num_stored_elements setting. Need to change the nonzeros or num_stored_elements according to `is_hermitian`.

PR #1531, opened 2024-01-17
Rename test binaries that collide with C++ standard library header names

### LINKED ISSUE #1530: Ginkgo build crashes
At this point

```
[ 61%] Building CXX object core/test/reorder/CMakeFiles/core_test_reorder_amd.dir/amd.cpp.o
```

the build crashes with the following

```
In file included from /home/oberhuber/Downloads/ginkgo/include/ginkgo/core/base/types.hpp:9,
     2                  from /home/oberhuber/Downloads/ginkgo/core/test/base/types.cpp:5:
     3 ./array:1:1: error: stray â<80><98>\177â<80><99> in program
     4     1 | <U+007F>ELF<U+0002><U+0001><U+0001><U+0003><U+0000><U+0000><U+0000><U+0000><U+0000><U+0000><U+0000><U+0000><U+0003><U+0000>><U+0000><U+0001><U+0000><U+0000><U+0000><c0><fe><U+0006><U+0000><U+0000><U+0000><U+0000><U+0000>@<U+0000><U+0000><U+0000><U+0000><U+0000><U+0000><U+0000><U+0010>P8<U+0000><U+0000><U+0000><U+0000><U+0000><U+0000><U+0000><U+0000><U+0000>@<U+00       00>8<U+0000>
     5       | ^~~~~~~~
     6 ./array:1:5: error: stray â<80><98>\2â<80><99> in program
     7     1 | <U+007F>ELF<U+0002><U+0001><U+0001><U+0003><U+0000><U+0000><U+0000><U+0000><U+0000><U+0000><U+0000><U+0000><U+0003><U+0000>><U+0000><U+0001><U+0000><U+0000><U+0000><c0><fe><U+0006><U+0000><U+0000><U+0000><U+0000><U+0000>@<U+0000><U+0000><U+0000><U+0000><U+0000><U+0000><U+0000><U+0010>P8<U+0000><U+0000><U+0000><U+0000><U+0000><U+0000><U+0000><U+0000><U+0000>@<U+00       00>8<U+0000>
     8       |            ^~~~~~~~
     9 ./array:1:6: error: stray â<80><98>\1â<80><99> in program
    10     1 | <U+007F>ELF<U+0002><U+0001><U+0001><U+0003><U+0000><U+0000><U+0000><U+0000><U+0000><U+0000><U+0000><U+0000><U+0003><U+0000>><U+0000><U+0001><U+0000><U+0000><U+0000><c0><fe><U+0006><U+0000><U+0000><U+0000><U+0000><U+0000>@<U+0000><U+0000><U+0000><U+0000><U+0000><U+0000><U+0000><U+0010>P8<U+0000><U+0000><U+0000><U+0000><U+0000><U+0000><U+0000><U+0000><U+0000>@<U+00       00>8<U+0000>
    11       |                    ^~~~~~~~
    12 ./array:1:7: error: stray â<80><98>\1â<80><99> in program
....
```

and it continues for several more megabytes :). I use the following compilers:

gcc (GCC) 13.2.1 20230801

and

nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2023 NVIDIA Corporation
Built on Tue_Aug_15_22:02:13_PDT_2023
Cuda compilation tools, release 12.2, V12.2.140
Build cuda_12.2.r12.2/compiler.33191640_0

upsj:
This gets rid of all the `CPATH` issues we've been having from time to time by renaming all test binaries that match a C++ standard library header name.

Fixes #1530

### DIFF
--- a/cmake/create_test.cmake
+++ b/cmake/create_test.cmake
@@ -1,5 +1,5 @@
 set(gko_test_resource_args "RESOURCE_LOCAL_CORES;RESOURCE_TYPE")
-set(gko_test_single_args "MPI_SIZE;${gko_test_resource_args}")
+set(gko_test_single_args "MPI_SIZE;EXECUTABLE_NAME;${gko_test_resource_args}")
 set(gko_test_multi_args "DISABLE_EXECUTORS;ADDITIONAL_LIBRARIES;ADDITIONAL_INCLUDES")
 set(gko_test_option_args "NO_RESOURCES")
 
@@ -80,9 +80,13 @@ endfunction()
 function(ginkgo_add_test test_name test_target_name)
     cmake_parse_arguments(PARSE_ARGV 2 add_test "" "${gko_test_single_args}" "${gko_test_multi_args}")
     file(RELATIVE_PATH REL_BINARY_DIR ${PROJECT_BINARY_DIR} ${CMAKE_CURRENT_BINARY_DIR})
-    set_target_properties(${test_target_name} PROPERTIES OUTPUT_NAME ${test_name})
+    set(test_binary_name ${test_name})
+    if (add_test_EXECUTABLE_NAME)
+        set(test_binary_name ${add_test_EXECUTABLE_NAME})
+    endif()
+    set_target_properties(${test_target_name} PROPERTIES OUTPUT_NAME ${test_binary_name})
     if (add_test_MPI_SIZE)
-        add_test(NAME ${REL_BINARY_DIR}/${test_name}
+        add_test(NAME ${REL_BINARY_DIR}/${test_binary_name}
                  COMMAND
                      ${MPIEXEC_EXECUTABLE}
                      ${MPIEXEC_NUMPROC_FLAG}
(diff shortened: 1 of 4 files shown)


pratikvn:
I think it might be a bit annoying to maintain these lists ? Maybe we can prefix all test executables with a unique prefix/suffix like `gkotest`/`gko` ? So the tests would be of the form `gkotest_array`/`gko_array` etc ?

upsj:
I don't think it's that much effort, the list of C++ standard library headers is pretty short and doesn't change a lot, and we don't have many tests that overlap the list. That's something we can do during code review.

upsj:
To go a bit more into detail why I dislike adding a prefix/suffix to all tests: I like being able to just build a single test binary via `ninja rcm_gpu`, the prefix/suffix makes that more annoying. The cause of this PR is a rather obscure issue that rarely impacts us, so I would like to make the workaround as unintrusive as possible.

pratikvn:
I see. That makes sense. Usually in my case, it is multiple targets that are relevant, so I end up doing a regex match with 
```sh
ninja -t targets | grep ^foo | sed 's/:.*//' > my_targets
```
so a suffix/prefix does not affect this.

and then you can just run `ninja $(cat my_targets)` whenever necessary. This could also be something we could add to our CMake build step accepting a string from the user, but maybe that is over-engineering this.

pratikvn (approved):

upsj:
What we could do there (and what I did before) is add meta-targets that compile certain parts of the library, e.g. a `benchmarks` target that compiles all benchmark binaries, a `tests` target that compiles all tests or stuff like that.

tcojean (approved):

MarcelKoch (approved):

PR #1532, opened 2024-01-17
Fixing External Include of Build Directory

MarcelKoch:
This PR extracts the exporting of the binary dir from our install configuration. To properly allow external projects to import Ginkgo directly from the binary tree, it is necessary to create two separate sets of config files. Otherwise, the binary tree config would refer to directories that are part of the install tree, which might not exist.

---
**Old Description:**

The build directory is missing the `GinkgoTargets.cmake` file. This prevents external projects from using the Ginkgo build directory directly, instead they need to use the install directory.

The issue was introduced in 75777b4c. Reverting a small part of it fixes the issue.

I also changed the `test_exportbuild` test to only search Ginkgo using the `Ginkgo_ROOT` path. TBH, I'm not sure if that was the intention of the test, but it can reproduce both the failing and succeeding behavior.


INSTALL.md
```diff
@@ -122,10 +119,15 @@ Ginkgo adds the following additional switches to control what is being built:
...
+Additionally, the following CMake options have effect on the build process:
+
+*  `-DCMAKE_EXPORT_PACKAGE_REGISTRY={ON,OFF}` if set to `ON` the build directory will
+   be stored in the current user's CMake package registry.
+
 For example, to build everything (in debug mode), use:
 
 ```cmake
-cmake  -G "Unix Makefiles" -H. -BDebug -DCMAKE_BUILD_TYPE=Debug -DGINKGO_DEVEL_TOOLS=ON \
+cmake -S. -BDebug -DCMAKE_BUILD_TYPE=Debug -DGINKGO_DEVEL_TOOLS=ON \
```
yhmtsai, line 130:
```suggestion
cmake -S.. -BDebug -DCMAKE_BUILD_TYPE=Debug -DGINKGO_DEVEL_TOOLS=ON \
```
Maybe keep using the build (or any child folder) not the source directory as the build directory

cmake/install_helpers.cmake
```diff
@@ -126,9 +132,7 @@ function(ginkgo_install)
         COMPONENT Ginkgo_Development)
 
     # Export package for use from the build tree
-    if (GINKGO_EXPORT_BUILD_DIR)
-        export(PACKAGE Ginkgo)
-    endif()
+    export(PACKAGE Ginkgo)
```
yhmtsai, line 135: could you explain more on this?
It does nothing without CMAKE_EXPORT_PACKAGE_REGISTRY=ON. Thus, it does not write anything to the user registry like ~/.cmake in the linux. (it does not affect us to use build directory as the external library)
never noticed cmake contains a user package registry before

.gitlab/scripts.yml
```diff
@@ -95,12 +94,12 @@
         -DGINKGO_CONFIG_LOG_DETAILED=${CONFIG_LOG}
         -DGINKGO_DPCPP_SINGLE_MODE=${DPCPP_SINGLE_MODE}
         -DGINKGO_RUN_EXAMPLES=${RUN_EXAMPLES}
-        -DGINKGO_EXPORT_BUILD_DIR=${EXPORT_BUILD_DIR}
     - ninja -j${NUM_CORES} -l${CI_LOAD_LIMIT} install
     - awk '!/^#/ { print ($2 - $1)/1000 " " $4 }' .ninja_log | sort -nr
     - |
         (( $(ctest -N | tail -1 | sed 's/Total Tests: //') != 0 )) || exit 1
     - ctest --output-on-failure --timeout 6000 ${CTEST_EXTRA_ARGS}
+    - ninja test_exportbuild
```
yhmtsai, line 102: to ensure it still works after installation?

yhmtsai (approved):

cmake/install_helpers.cmake
MarcelKoch: I think that already covers it. Without `CMAKE_EXPORT_PACKAGE_REGISTRY=ON` this is a no-op, so we don't need to guard it. Also setting `GINKGO_EXPORT_BUILD_DIR` had no effect on the other CMake option, so no need for that.
The CMake registry basically just stores the paths to the packages.

.gitlab/scripts.yml
upsj: Side-note: The reason we do `ninja install` directly in the beginning is that the HIP compilation doesn't create (or didn't used to create) the correct depfiles or something like that on the first pass, so calling `ninja` and then `ninja install` caused the HIP objects to be rebuilt. If we tackle #1334, that is no longer an issue and we can separate between `ninja` and `ninja install` again

MarcelKoch: TBH, I think we should push the CMake update for HIP, even if that would mean to drop support for rocm < 5.0.1. I don't think it's even possible to find the documentation for rocm 4.5 from visiting amd.com, so it seems to me that these versions are not really supported anymore.

yhmtsai: for the issue that hip needs built twice, we can just run `ninja && ninja` if we do not want `ninja install` rebuilds the stuff.
I do not think the two build is a big issue.
Dropping the support rocm < 5.0.1 is not the main issue from updating cmake for HIP.  
Updating cmake for hip means we drops the nvidia support of hip. In my mind, the new cmake is not completed yet because they claim they can do that on both architectures. Also, recovering the support after deleting is harder than keeping the current one.

upsj: This is an issue of build time, our HIP builds (especially debug) are extremely slow, we don't want to build the binaries twice. It seems to me like the old `hip_add_*` approach hasn't been mentioned in their documentation in a long time, on top of all its downsides I also do not want to keep relying on it. We have full CUDA support via our CUDA backend, to me there are no technical reasons to keep this around, as HIP is just a thin layer over CUDA anyways.

upsj (approved):
LGTM! Thanks for fixing this

pratikvn (approved):
LGTM!

yhmtsai:
We have two version of cmakeConfig and implementation in the code. Is there any way to reduce the possibility of missing partial changes in the future?

PR #1533, opened 2024-01-17
Change write_raw default to sparse output format

upsj:
This leads to a much better worst-case storage size (roughly 3x overhead instead of linear overhead) at the cost of slightly worse best case storage requirement.

Since this is only a utility function, it shouldn't matter much.

### DIFF
--- a/include/ginkgo/core/base/mtx_io.hpp
+++ b/include/ginkgo/core/base/mtx_io.hpp
@@ -118,7 +118,7 @@ enum class layout_type {
  */
 template <typename ValueType, typename IndexType>
 void write_raw(std::ostream& os, const matrix_data<ValueType, IndexType>& data,
-               layout_type layout = layout_type::array);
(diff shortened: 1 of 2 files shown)


include/ginkgo/core/base/mtx_io.hpp
```diff
@@ -118,7 +118,7 @@ enum class layout_type {
  */
 template <typename ValueType, typename IndexType>
 void write_raw(std::ostream& os, const matrix_data<ValueType, IndexType>& data,
-               layout_type layout = layout_type::array);
+               layout_type layout = layout_type::coordinate);
```
yhmtsai, line 121: I think we use array as default in the beginning to keep the same behavior as the old interface:
`void write_raw(std::ostream& os, const matrix_data<ValueType, IndexType>& data);`
but it should not be public interface break with this change

yhmtsai (approved):

PR #1536, opened 2024-01-19
Adding CMake Presets

MarcelKoch:
This PR can be used to discuss the addition of CMake presets. So far, I've added only very basic presets which are compatible with the current workflows.

But these preset can be used to set common standards, especially regarding warnings. ATM, we only enforce the `-Wpedantic` warnings for our development builds. We could use the presets to define a more rigorous set of warnings.

There might also be other options we might want to set here, this is only a starting point for a discussion.


CMakePresets.json
```diff
@@ -0,0 +1,37 @@
...
+        "CMAKE_BUILD_TYPE": "Release"
+      },
+      "warnings": {
+        "dev": false
+      }
+    },
+    {
+      "name": "develop",
+      "inherits": "base",
+      "cacheVariables": {
```
upsj, line 23: From our previous discussions, I would suggest adding `CMAKE_EXPORT_COMPILE_COMMANDS=ON`, `CMAKE_COLOR_DIAGNOSTICS=ON`, `GINKGO_DEVEL_TOOLS=ON`, `CMAKE_LINK_DEPENDS_NO_SHARED=ON`

upsj:
Another useful feature might be the `environment` object, with which we can configure path-independent `CXX` and `CUDACXX` compiler executable names

CMakePresets.json
```diff
@@ -0,0 +1,37 @@
...
+        "CMAKE_BUILD_TYPE": "Debug",
+        "CMAKE_CXX_FLAGS": "$env{CXXFLAGS} -Wpedantic"
```
MarcelKoch, line 25: In terms of warnings I was thinking of 
```
-Wall -Wextra -Wconversion -Wunused
```

MarcelKoch:
> Another useful feature might be the `environment` object, with which we can configure path-independent `CXX` and `CUDACXX` compiler executable names

Are you meaning something like `CXX=g++`? I think that would make sense if we would extend the presets to be used in our CI, but otherwise I'm not sure of the benefit of that.

upsj:
I could imagine `develop-clang` being useful from time to time, especially with its better diagnostics

greole:
I guess that should also work with cmakes [new](https://cmake.org/cmake/help/latest/manual/cmake.1.html#run-a-workflow-preset) `--workflow` option.

CMakePresets.json
```diff
@@ -0,0 +1,37 @@
+{
+  "version": 1,
+  "configurePresets": [
+    {
+      "name": "base",
+      "hidden": true,
+      "generator": "Ninja",
```
greole, line 7: Do we want to require `Ninja` as dependency? Often needs a separate module to be loaded and is not available on some systems.

cmake/build_helpers.cmake
```diff
@@ -113,26 +113,6 @@ macro(ginkgo_modify_flags name)
     string(REPLACE "\"" "\\\"" ${name}_MODIFY "${${name}}")
 endmacro()
 
-# Extract the clang version from a clang executable path
```
greole, line 116: I guess this is superseded by `CMAKE_CXX_COMPILER_VERSION`?

upsj:
@pratikvn I think the issue there is that they depend on loaded modules, and the executable names (gcc etc) are the same otherwise, so the configs may not help and differ that much between different systems.

We could consider adding a '-march=native` preset that also includes other optimizations (mixed precision, maybe jacobi full optimizations)

MarcelKoch:
I tend to agree that adding supported machines may clutter the file a bit too much. But it is possible to include configurations from other files, so perhaps we could put these configs into a separate file.

CMakeCIPresets.json
```diff
@@ -0,0 +1,608 @@
...
+            "name": "enable-msvc",
+            "hidden": true,
+            "generator": "Visual Studio 17 2022",
+            "cacheVariables": {
+                "CMAKE_CXX_COMPILER": "msvc",
+                "CMAKE_CXX_FLAGS_DEBUG": "/MDd /Zi /Ob1 /O1 /Od /RTC1"
+            }
+        },
+        {
+            "name": "ci-cuda-msvc-nompi-release",
```
MarcelKoch, line 137: I chose the naming scheme `ci-[device (multiple connected with +)]-[toolchain]-[nompi|mpi]-[build_type][optional configs]`.  Should this be kept? 
Also, in our ci we seem to use `/` as a separator instead of `-`. We could do the same here, with the side effect, that we would create nested folders.

pratikvn: I think it was just a convention we used when building our containers. IMO using `-` seems reasonable. Does having nested folders have any advantages ?

upsj: The only impact (not sure if it's an advantage) is that the folders used by the Gitlab runner share the same nested structure.

CMakePresets.json
MarcelKoch: If we use presets version `< 3`, then we need to specify a generator. But even if we don't need to specify it, I would still prefer to set `Ninja`, since it seems to handle dependencies better than make, and it's not as verbose. It's always possible to overwrite the settings from the preset, by using the normal CLI arguments.

cmake/build_helpers.cmake
MarcelKoch: Maybe, but this function was never used.

CMakePresets.json
```diff
@@ -0,0 +1,76 @@
...
+            "cacheVariables": {
+                "CMAKE_CXX_FLAGS": "$env{CXXFLAGS} /permissive- /Oy-"
+            },
+            "condition": {
+                "type": "equals",
+                "lhs": "${hostSystemName}",
+                "rhs": "Windows"
+            }
+        }
+    ]
```
greole, line 75: Would it makes sense to also add `buildPresets` like
```suggestion
    ], 
      "buildPresets": [
    {
      "name": "develop-gcc",
      "configurePreset": "develop-gcc"
    }
    ]
```
that way one can build with `cmake --build --preset develop-gcc` without knowing the exact `binaryDir`

CMakePresets.json
```diff
@@ -0,0 +1,76 @@
+{
+    "version": 4,
+    "include": ["cmake/presets/ci.json"],
+    "configurePresets": [
+        {
+            "name": "base",
+            "hidden": true,
+            "binaryDir": "${sourceDir}/build-${presetName}"
```
greole, line 8: My personal preference would be nesting the build folder and adding buildPresets. So that there won't be multiple `build-` folders in the `sourceDir` 
```suggestion
            "binaryDir": "${sourceDir}/build/${presetName}"
```

MarcelKoch: I would agree. Although typically there wouldn't be too many different build dirs anyway.

CMakePresets.json
MarcelKoch: For the 'normal' configuration I can see this making sense. But for the CI configs it would probably be too much.

CMakePresets.json
greole: Thinking about it, I usually build in a `build/<branchName>` directory to keep different builds separated. But I guess that information is not available for CMakePresets.

MarcelKoch: No, but it's always possible to overwrite this from the CLI

CMakePresets.json
greole: I think `buildPresets` make especially sense in combination with the [workflow]( https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#workflow-preset) option. So that a user can configure and build a preset with a single command. However, workflows are only available for version 6 and above, which needs cmake 3.25.

greole: Just another thought on the buildPresets, we could also add a format preset  with a targets
 key
 ```
"targets": [
        "format",
      ]
```
However, I don't know how to then just pick the "current" `binaryDir` via the preset.

Comment (upsj) [reply CMakePresets.json:75]: I think we'll probably move away from that with pre-commit handling the formatting entirely?

Comment (greole) [reply CMakePresets.json:75]: The thing I like about presets is that it allows you to run `cmake --build --preset format` from the source directory without knowing the corresponding build directory. However, you are right that moving towards precommit makes it unnecessary.

File: CMakePresets.json [hunk 8]
```diff
@@ -0,0 +1,76 @@
...
+        },
+        {
+            "name": "base-dev",
+            "hidden": true,
+            "cacheVariables": {
+                "CMAKE_BUILD_TYPE": "Debug",
```
greole, line 14: Nowadays I use presets almost everywhere, thus having a Release version would be also great.

greole (approved):
LGTM! 
- I think, however, we should also mention it the presets in the README.md and maybe give an example of how to use presets in combination with normal flags to enable compiling specific backends.
- And if possible add a release version of the presets.

.gitlab/scripts.yml
```diff
@@ -104,9 +86,31 @@
...
+            export EX_ARG="cuda"
+            ninja run_all_examples
+            ninja validate_all_examples
+        fi
+        if [ "$(get_cache_val GINKGO_BUILD_HIP)" == "ON" ]; then
+            export EX_ARG="hip"
+            ninja run_all_examples
+            ninja validate_all_examples
+        fi
+      fi
```
MarcelKoch, line 109: Was there a replacement for this, or do we just not test this anymore?

.gitlab-ci.yml
```diff
@@ -364,13 +314,7 @@ build/nocuda/openmpi/clang/omp/glibcxx-debug-release/shared:
...
     - .use_gko-nocuda-nompi-gnu9-llvm8
   variables:
-    CXX_COMPILER: "clang++"
-    BUILD_OMP: "ON"
-    MPI_AS_ROOT: "ON"
-    BUILD_MPI: "ON"
-    CXX_FLAGS: "-Wpedantic -D_GLIBCXX_DEBUG=1"
-    # The tests are prohibitively slow in Debug
-    BUILD_TYPE: "Release"
+    CI_CMAKE_PRESET: "ci-omp-clang-mpi-glibcxx+debug-release"
```
MarcelKoch, line 317: does this config even make sense to run with `_GLIBCXX_DEBUG` in release mode?

cmake/presets/ci.json
```diff
@@ -0,0 +1,575 @@
...
+    "version": 4,
+    "configurePresets": [
+        {
+            "name": "base-ci",
+            "hidden": true,
+            "binaryDir": "${sourceDir}/build-${presetName}",
+            "generator": "Ninja",
+            "environment": {
+                "MPI_AS_ROOT": "OFF"
+            },
```
yhmtsai, line 11:
```suggestion
            },
            "condition": {
                "type": "equals",
                "lhs": "$env{GINKGO_PRESETS}",
                "rhs": "CI"
            },
```
We can use the environment to control the list here.
`GINKGO_PRESETS=CI cmake --list-presets configure ..` will show this additionally such that we can hide this long list to the users in the first side.

PR #1538, opened 2024-01-29
Expand hipSPARSE bindings for complex types

nbeams:
hipSPARSE has improved its complex support, so we can add complex types to these routines.


yhmtsai (approved):
LGTM.
side note: hipsparse is 1.11.2 for rocm 4.5, so we do not need to check 1.4 for complex

pratikvn:
We dont test any of these. Maybe we should think about having capabilities to more comprehensive typed unit tests ? Probably not to run with our regular CI, as that would consume too much time, but something that we could run if necessary, or maybe before a release to ensure that all types work well.

upsj (approved):

MarcelKoch (approved):

PR #1543, opened 2024-02-14
Distributed Index Map

MarcelKoch:
This PR adds an `index_map` to be used in our distributed components. This PR only adds the core and reference kernels. device kernels are added in #1579 

I will add another PR that shows how this will be used.

PR Stack:

- [x] #1545 
- [ ] #1543
- [ ] #1579
- [ ] #1544 
- [ ] #1588
- [ ] #1589


reference/test/distributed/index_map_kernels.cpp
```diff
@@ -0,0 +1,169 @@
...
+#include <gtest/gtest.h>
+
+
+#include <ginkgo/core/base/executor.hpp>
+
+
+#include "core/distributed/index_map_kernels.hpp"
+#include "core/test/utils.hpp"
+
+namespace {
```
pratikvn, line 23: Is this anonymous namespace needed ?

pratikvn (changes requested):
A couple of minor things, but mostly looks okay to me.

include/ginkgo/core/distributed/index_map.hpp
```diff
@@ -0,0 +1,178 @@
...
+namespace distributed {
+
+
+/**
+ * \brief Index space classification for the locally stored indices.
+ */
+enum class index_space {
+    local,      //!< indices that are locally owned
+    non_local,  //!< indices that are owned by other processes
+    combined    //!< both local and non_local indices
```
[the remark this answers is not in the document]
MarcelKoch: It's this: https://www.doxygen.nl/manual/docblocks.html#memberdoc
But maybe we are using a different variant of that more commonly.

core/distributed/index_map.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+template <typename LocalIndexType, typename GlobalIndexType>
+array<LocalIndexType> index_map<LocalIndexType, GlobalIndexType>::get_local(
+    const array<GlobalIndexType>& global_ids, index_space is) const
+{
+    array<LocalIndexType> local_ids(exec_);
+
+    exec_->run(index_map_kernels::make_get_local(
+        partition_.get(), remote_target_ids_, remote_global_idxs_, rank_,
+        global_ids, is, local_ids));
```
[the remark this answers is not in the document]
MarcelKoch: I guess the issue is that the enum is already called `index_space`, so I can't use that. Perhaps `index_space_v`?

reference/distributed/index_map_kernels.cpp
```diff
@@ -0,0 +1,222 @@
...
+        for (size_type i = 0; i < global_ids.get_size(); ++i) {
+            auto gid = global_ids.get_const_data()[i];
+            local_ids.get_data()[i] = map_combined(gid);
+        }
+    }
+}
+
+GKO_INSTANTIATE_FOR_EACH_LOCAL_GLOBAL_INDEX_TYPE(
+    GKO_DECLARE_INDEX_MAP_GET_LOCAL_FROM_GLOBAL_ARRAY);
+
```
pratikvn, line 218:
```suggestion


```

pratikvn (approved):
LGTM!

include/ginkgo/core/distributed/index_map.hpp
```diff
@@ -0,0 +1,180 @@
...
+ * $I_l, l \neq k$. In particular, any global index passed to the `get_local`
+ * map that is not part of the specified index space, will be mapped to an
+ * invalid_index.
+ *
+ * \tparam LocalIndexType  type for local indices
+ * \tparam GlobalIndexType  type for global indices
+ */
+template <typename LocalIndexType, typename GlobalIndexType = int64>
+struct index_map {
+    using part_type = Partition<LocalIndexType, GlobalIndexType>;
```
upsj, line 64: I know partition is annoying to type out, but I believe we should avoid abbreviations if they themselves have another meaning in the same context (e.g. I used to refer to a specific rank's data in a partition as a part in a lot of my code, so this is surprising when skimming over)
```suggestion
    using partition_type = Partition<LocalIndexType, GlobalIndexType>;
```

include/ginkgo/core/distributed/index_map.hpp
```diff
@@ -0,0 +1,180 @@
...
+     * filtered out.
+     *
+     * \param exec  the executor
+     * \param part  the partition of the global index set
+     * \param rank  the id of the global index space subset
+     * \param recv_connections  the global indices that are not owned by this
+     *                          rank, but accessed by it
+     */
+    index_map(std::shared_ptr<const Executor> exec,
+              std::shared_ptr<const part_type> part, comm_index_type rank,
```
upsj, line 109:
```suggestion
     * \param partition  the partition of the global index set
     * \param rank  the id of the global index space subset
     * \param recv_connections  the global indices that are not owned by this
     *                          rank, but accessed by it
     */
    index_map(std::shared_ptr<const Executor> exec,
              std::shared_ptr<const part_type> partition, comm_index_type rank,
```

include/ginkgo/core/distributed/index_map.hpp
```diff
@@ -0,0 +1,180 @@
...
+              const array<GlobalIndexType>& recv_connections);
+
+    /**
+     * \brief Creates an empty index map.
+     */
+    index_map() = default;
```
upsj, line 115: What case do you have in mind for an empty index_map? Its executor and its arrays' executors are uninitialized, that seems problematic

upsj (commented):
First look at the interface

include/ginkgo/core/distributed/index_map.hpp
MarcelKoch: I think a empty index map is a valid object, so I would like to keep it. I guess I could require an executor, that wouldn't change much.

include/ginkgo/core/distributed/index_map.hpp
```diff
@@ -0,0 +1,180 @@
...
+     * \param global_ids  the global indices to map
+     * \param index_space_v  the index space in which the returned local indices
+     *                       are defined
+     *
+     * \return  the mapped local indices. Any global index that is not in the
+     *          specified index space is mapped to invalid_index.
+     */
+    [[nodiscard]] array<LocalIndexType> get_local(
+        const array<GlobalIndexType>& global_ids,
+        index_space index_space_v = index_space::combined) const;
```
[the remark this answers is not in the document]
MarcelKoch: I really dislike out-parameters, so I would like to keep this version, but I could add an overload with the out-parameter. Right now, the primary usecase is the `read_distributed` and there, the result buffer is not reused, so the out-parameter version has no benefit.

include/ginkgo/core/distributed/index_map.hpp
MarcelKoch: This is gone now, it's replaced by a constructor with a single executor.

reference/distributed/partition_helpers.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+
+#include <algorithm>
+
+
+#include <ginkgo/core/base/types.hpp>
+#include <ginkgo/core/distributed/partition.hpp>
+
+
+namespace gko {
```
pratikvn, line 16: Maybe also in `detail` namespace ?

core/distributed/index_map.cpp
```diff
@@ -0,0 +1,188 @@
...
+#include "core/distributed/index_map_kernels.hpp"
+
+
+namespace gko {
+namespace index_map_kernels {
+
+
+GKO_REGISTER_OPERATION(build_mapping, index_map::build_mapping);
+GKO_REGISTER_OPERATION(get_local, index_map::get_local);
+
```
pratikvn, line 17: nit:
```suggestion


```

pratikvn (approved):
LGTM!

reference/distributed/index_map_kernels.cpp
```diff
@@ -0,0 +1,210 @@
...
+    auto find_part = [&](GlobalIndexType idx) {
+        auto range_id = find_range(idx, part, 0);
+        return part_ids[range_id];
+    };
+
+    // sort by (part-id, global-id)
+    std::sort(unique_indices.begin(), unique_indices.end(),
+              [&](const auto a, const auto b) {
+                  auto part_a = find_part(a);
+                  auto part_b = find_part(b);
```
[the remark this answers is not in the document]
MarcelKoch: Since this is reference, I would prefer keeping it this way, because otherwise I need more temporary storage.

include/ginkgo/core/distributed/index_map.hpp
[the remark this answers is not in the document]
upsj: sounds reasonable

include/ginkgo/core/distributed/index_map.hpp
```diff
@@ -0,0 +1,176 @@
...
+ * separated into locally owned (index_space::local) and non-locally owned
+ * (index_space::non_local). The locally owned indices are defined as
+ * $[0, \dots, |I_k|)$, and the non-locally owned as $[0, \dots, |R_k|)$.
+ * With these index sets, the following mappings are defined:
+ * - $c_k : \hat{I}_k \mapsto [0, \dots, |\hat{I}_k|)$ which maps global indices
+ *   into the combined/full local index space,
+ * - $l_k: I_k \mapsto [0, \dots, |I_k|)$ which maps global indices into the
+ *   locally owned index space,
+ * - $r_k: R_k \mapsto [0, \dots, |R_k|)$ which maps global indices into the
+ *   non-locally owned index space.
```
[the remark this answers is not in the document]
MarcelKoch: I would feel that the descriptions of $c_k$, $l_k$, $r_k$ are too closely related to the nomenclature introduced above. If I would move them the `index_space` enum, then I think I would need to move everything up there. But I think that part should stay here.
I will add more cross-references, to hopefully clear up the connections.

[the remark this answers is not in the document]
upsj: sounds like a good solution

include/ginkgo/core/distributed/index_map.hpp
```diff
@@ -0,0 +1,176 @@
...
+     */
+    const segmented_array<LocalIndexType>& get_remote_local_idxs() const;
+
+    /**
+     * \brief get the rank ids which contain indices accessed by this rank.
+     *
+     * The order matches the order of the sets in get_remote_global_idxs and
+     * get_remote_local_idxs.
+     */
+    const array<comm_index_type>& get_remote_target_ids() const;
```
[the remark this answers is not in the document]
MarcelKoch: I guess so, maybe I can actually do the same as for csr, i.e. storing the offsets, and local/global indices explicitly.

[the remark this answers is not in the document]
upsj: `get_remote_target_ids` definitely makes sense, I'm just wondering whether `get_remote_global_idxs` will stay that way. But since this is still `experimental`, maybe we can replace it by a more efficient representation before stabilizing the interface.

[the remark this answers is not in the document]
upsj: > I guess so, maybe I can actually do the same as for csr, i.e. storing the offsets, and local/global indices explicitly.

I am worried less about storage overhead than potential for inconsistencies between the two arrays, I'd prefer a single source of truth for the ranges.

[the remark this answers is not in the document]
MarcelKoch: I think I will offer `get_remote_target_sizes` instead of the indices. This is more directly what I need, and then we could still change the internal representation.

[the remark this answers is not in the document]
upsj: Does that map well to sparse communication? Otherwise we might have issues with scaling to huge numbers of ranks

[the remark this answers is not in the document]
MarcelKoch: I realized that I actually use `get_remote_local_idxs` to communicate the local indices for the distributed row-gather. And they are also used without the row-gatherer in the `read_distributed`. So for this one I would at least need direct pointer access.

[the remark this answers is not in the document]
MarcelKoch: I think I also provided `get_remote_global_idxs`, because this is directly defined through the class documentation, while the local indices need more explaining. So I could remove the global indices (but I feel a bit uncomfortable with that because it breaks some sort of symmetry).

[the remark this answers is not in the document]
MarcelKoch: For now, I would leave it as it is. I specifically thought of segmented arrays as storage for these arrays, so changing that seems to have quite far-reaching consequences. Since this all still experimental, we could change it if we find better solutions later on.

core/distributed/index_map.cpp
```diff
@@ -0,0 +1,187 @@
...
+}
+
+
+template <typename LocalIndexType, typename GlobalIndexType>
+index_map<LocalIndexType, GlobalIndexType>&
+index_map<LocalIndexType, GlobalIndexType>::operator=(index_map&& other)
+{
+    if (this != &other) {
+        partition_ = std::move(other.partition_);
+        rank_ = other.rank_;
```
upsj, line 150: nit: should the moved-from state preserve the original rank?

core/distributed/index_map_kernels.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+
+namespace gko {
+namespace kernels {
+
+
+/**
+ * This kernel creates an index map from a partition and global remote indices.
+ *
+ * The index map is defined by the output parameters remove_local_idxs,
```
upsj, line 28:
```suggestion
 * The index map is defined by the output parameters remote_local_idxs,
```

core/test/distributed/index_map.cpp
```diff
@@ -0,0 +1,165 @@
...
+    auto copy = imap;
+    auto imap_remote_global_it =
+        imap.get_remote_global_idxs().get_const_flat_data();
+    auto imap_remote_local_it =
+        imap.get_remote_local_idxs().get_const_flat_data();
+    auto imap_remote_target_it = imap.get_remote_target_ids().get_const_data();
+
+    auto move = std::move(imap);
+
+    GKO_ASSERT_ARRAY_EQ(move.get_remote_target_ids(),
```
upsj, line 103: you could also check the moved-from state here, but we rarely do that, so it's fine if you don't ;)

include/ginkgo/core/distributed/index_map.hpp
```diff
@@ -0,0 +1,181 @@
...
+     *
+     * The indices are ordered by their owning rank and global index.
+     */
+    const segmented_array<GlobalIndexType>& get_remote_global_idxs() const;
+
+    /**
+     * \brief get the index set $R_k$, but mapped to their respective local
+     *        index space.
+     *
+     * The indices are ordered by their owning rank and global index.
```
upsj, line 133: nit: could be a bit more explicit
```suggestion
     * The indices are grouped by their owning rank and sorted according to their global index inside each group.
```

reference/distributed/index_map_kernels.cpp
```diff
@@ -0,0 +1,207 @@
...
+              });
+
+    // make unique by (part-id, global-id)
+    auto unique_indices_end =
+        std::unique(unique_indices.begin(), unique_indices.end(),
+                    [&](const auto a, const auto b) {
+                        auto part_a = find_part(a);
+                        auto part_b = find_part(b);
+                        return std::tie(part_a, a) == std::tie(part_b, b);
+                    });
```
upsj, line 59: it's annoying that there is no suitable implementation to handle both < and ==, then we could get rid of this almost duplication

reference/distributed/index_map_kernels.cpp
```diff
@@ -0,0 +1,207 @@
...
+                       return map_to_local(idx, part, range);
+                   });
+
+    // get part-ids
+    std::vector<comm_index_type> full_part_ids(unique_size);
+    std::transform(unique_indices.begin(), unique_indices_end,
+                   full_part_ids.begin(),
+                   [&](const auto idx) { return find_part(idx); });
+
+    std::vector<comm_index_type> unique_part_ids(full_part_ids);
```
upsj, line 83: executor allocator throughout all std::vectors

upsj (approved):
LGTM! Only minor point is gko::vector usage to track memory allocations

core/distributed/index_map.cpp
```diff
@@ -0,0 +1,187 @@
...
+    : exec_(std::move(exec)),
+      partition_(std::move(partition)),
+      rank_(rank),
+      remote_target_ids_(exec_),
+      remote_local_idxs_(exec_),
+      remote_global_idxs_(exec_)
+{
+    array<LocalIndexType> flat_remote_local_idxs(exec_);
+    array<GlobalIndexType> flat_remote_global_idxs(exec_);
+    array<int64> remote_sizes(exec_);
```
pratikvn, line 107: can be 
```suggestion
    array<GlobalIndexType> remote_sizes(exec_);
```
?

pratikvn (approved):
LGTM!

core/distributed/index_map.cpp
MarcelKoch: The segmented array always uses int64 offsets, and the sizes have to be the same type.

core/distributed/index_map.cpp
MarcelKoch: I think that is sensible. We don't really have an invalid rank value that I could put here.

reference/distributed/index_map_kernels.cpp
MarcelKoch: starship operator <=> pew pew

PR #1544, opened 2024-02-14
Use index_map in distributed::matrix

MarcelKoch:
Application of #1543. Now the `index_map` is used to create the numbering for the non-local columns. The `index_map` is created during the `distributed::matrix::read_distributed` and also returned from that function. This can allow users to provide already separated local and non-local matrices for the `read_distributed` further down the line.
Also, it makes the mapping `non-local -> global` more easily available.

This PR also introduces:
- making `run` easier to use
- specialization of `run` for distributed matrices -> removes need for changes to distributed base

PR Stack:

- [ ] #1545 
- [ ] #1543
- [ ] #1579
- [ ] #1544  
- [ ] #1588
- [ ] #1589


pratikvn (changes requested):
LGTM! The issue of interface needs to be resolved. It is in the `experimental` namespace, but I think we said we would still not break interface in `experimental` ?

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -281,10 +282,12 @@ class Matrix
      *
      * @param data  The device_matrix_data structure.
      * @param partition  The global row and column partition.
+     *
+     * @return the index_map induced by the partitions and the matrix structure
      */
-    void read_distributed(
+    index_map<local_index_type, global_index_type> read_distributed(
```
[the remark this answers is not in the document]
upsj: we made no promises about interface-stability in `experimental`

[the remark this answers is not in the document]
MarcelKoch: TBH, I don't consider that really an interface break. This has an effect only on very serious template meta programming, which I don't expect users to do. Other than that, current code will compile, it just discards the return value.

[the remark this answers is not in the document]
MarcelKoch: There is actually an interface break here, but not due to the return value. Now this function expects a `shared_ptr<Partition>` instead of a raw pointer. But this is done in #1543 if you want to discuss that.

core/distributed/matrix.cpp
```diff
@@ -243,57 +261,68 @@ void Matrix<ValueType, LocalIndexType, GlobalIndexType>::read_distributed(
     recv_offsets_[0] = 0;
 
     // exchange step 2: exchange gather_idxs from receivers to senders
+    auto recv_gather_idxs =
+        make_const_array_view(
+            imap.get_executor(), imap.get_non_local_size(),
+            imap.get_remote_local_idxs().get_const_flat_data())
+            .copy_to_array();
```
pratikvn, line 268: This copy is only needed for the host path, right ?

reference/test/distributed/matrix_kernels.cpp
```diff
@@ -436,27 +416,27 @@ TYPED_TEST(Matrix, BuildsLocalNonLocalMixedWithColPartition)
             this->ref, col_mapping, num_parts);
 
     this->validate(gko::dim<2>{6, 6}, partition, col_partition,
+        // clang-format off
                    {2, 3, 3, 0, 5, 1, 4, 2, 3, 2, 0, 0, 1, 1, 4, 4},
-                   {0, 0, 1, 5, 4, 2, 2, 3, 2, 4, 1, 2, 4, 5, 0, 5},
+                   { 0,  0,  1,  5,  4,  2,  2, 3, 2, 4, 1, 2, 4, 5, 0, 5},
```
pratikvn, line 421: nit: old formatting was better

reference/test/distributed/matrix_kernels.cpp
```diff
@@ -473,76 +453,22 @@ TYPED_TEST(Matrix, BuildsLocalNonLocalNonSquare)
...
-        gko::kernels::reference::distributed_matrix::build_local_nonlocal(
-            this->ref, this->create_input_full_rank(), partition.get(),
-            partition.get(), local_id, this->local_row_idxs,
-            this->local_col_idxs, this->local_values, this->non_local_row_idxs,
-            this->non_local_col_idxs, this->non_local_values, this->gather_idxs,
-            this->recv_sizes, this->non_local_to_global);
-
-        GKO_ASSERT_ARRAY_EQ(result[local_id], this->non_local_to_global);
-    }
-}
```
pratikvn, line 546: Are these tests not relevant anymore ?

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -278,10 +279,12 @@ class Matrix
...
      * @param data  The device_matrix_data structure.
      * @param partition  The global row and column partition.
+     *
+     * @return the index_map induced by the partitions and the matrix structure
      */
-    void read_distributed(
+    index_map<LocalIndexType, GlobalIndexType> read_distributed(
         const device_matrix_data<value_type, global_index_type>& data,
-        ptr_param<const Partition<local_index_type, global_index_type>>
+        std::shared_ptr<const Partition<local_index_type, global_index_type>>
```
pratikvn, line 287: I think we maybe already discussed this, but I think this breaks existing code.

include/ginkgo/core/distributed/matrix.hpp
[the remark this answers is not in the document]
pratikvn: Somehow the old comments I made dont appear anymore, so it is a bit annoying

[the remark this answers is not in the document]
pratikvn: But this is indeed an unfortunate interface break. I think the version with `shared_ptr`  should be okay, but I think users actually depend on `read_distributed`, so it will break existing code.

core/distributed/matrix.cpp
MarcelKoch: No, it's always a copy, since the `non_local_to_global` is stored (for now) in the matrix.

reference/test/distributed/matrix_kernels.cpp
MarcelKoch: for me it is helpful to have the column idxs and the values aligned. I think all of these mappings are based on a drawing I used, and the alignment helped me to prevent mixing things up.

MarcelKoch: Especially if the row and column partitions are different and both are non-trivial.

include/ginkgo/core/distributed/matrix.hpp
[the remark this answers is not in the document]
MarcelKoch: I think the index map needs to store the partition. I've looked again into the `map_to_local` kernels, and there I actually need *all* arrays a partition stores. So it wouldn't make sense to only copy the needed data out of the partition and into the index-map.

pratikvn (approved):

benchmark/test/reference/spmv_distributed.profile.stdout
```diff
@@ -5,7 +5,7 @@
         "comm_pattern": "stencil",
         "spmv": {
             "csr-csr": {
-                "storage": 6564,
+                "storage": 6420,
```
upsj, line 8: Maybe not directly related to this, but I went back to check, there are some places left over in #1543 that still use std::vector instead of gko::vector or gko::array.

core/distributed/matrix.cpp
```diff
@@ -302,8 +312,16 @@ void Matrix<ValueType, LocalIndexType, GlobalIndexType>::read_distributed(
...
+    auto host_recv_targets =
+        make_temporary_clone(exec->get_master(), &imap.get_remote_target_ids());
+    auto host_offsets = make_temporary_clone(
+        exec->get_master(), &imap.get_remote_global_idxs().get_offsets());
+    std::fill(recv_sizes_.begin(), recv_sizes_.end(), 0);
+    for (size_type i = 0; i < host_recv_targets->get_size(); ++i) {
+        recv_sizes_[host_recv_targets->get_const_data()[i]] =
+            host_offsets->get_const_data()[i + 1] -
+            host_offsets->get_const_data()[i];
+    }
```
upsj, line 324: It feels pedantic, but I think having a raw loop next to a bunch of algorithm calls is a mismatch of abstraction levels - this might be better extracted into a separate function, similar to Sean Parent's "no raw loops" suggestion (https://www.youtube.com/watch?v=W2tWOdzgXHA)

Maybe we can call it `compute_recv_sizes`? Then the high-level control flow is easy to follow, and if you need to know how those recv_sizes get determined, you look at the function call.

reference/test/distributed/matrix_kernels.cpp
```diff
@@ -287,26 +271,26 @@ TYPED_TEST(Matrix, BuildsLocalNonLocalMixed)
...
         {std::make_tuple(gko::dim<2>{2, 2}, I<git>{0, 1}, I<git>{1, 0},
                          I<vt>{15, 16}),
          std::make_tuple(gko::dim<2>{2, 2}, I<git>{0, 0, 1}, I<git>{0, 1, 0},
                          I<vt>{11, 12, 18}),
          std::make_tuple(gko::dim<2>{2, 2}, I<git>{0, 0, 1}, I<git>{0, 1, 1},
                          I<vt>{13, 14, 17})},
-        {std::make_tuple(gko::dim<2>{2, 1}, I<git>{1}, I<git>{0}, I<vt>{6}),
-         std::make_tuple(gko::dim<2>{2, 3}, I<git>{0, 0, 1}, I<git>{2, 1, 0},
+        {std::make_tuple(gko::dim<2>{2, 1}, I<git>{1}, I<git>{1}, I<vt>{6}),
+         std::make_tuple(gko::dim<2>{2, 3}, I<git>{0, 0, 1}, I<git>{1, 3, 2},
```
upsj, line 286: These changes are a bit hard to review - can you add a comment to the function explaining what the tuple members represent?

upsj (approved):
LGTM, minor gripe with the memory allocation tracking that is incomplete in index_map, maybe a fix could be included here?

core/distributed/matrix.cpp
MarcelKoch: IMO the issue here is that arrays can't be used with STL. If we had begin, and end for them, this would just be `std::adjacent_difference`

upsj: You can use pointers here , but `adjacent_difference` is not 100% sufficient, since it leaves the first element unchanged and doesn't scatter the results. I just meant extract these 5 lines into a function

core/distributed/matrix.cpp
```diff
@@ -271,29 +270,40 @@ void Matrix<ValueType, LocalIndexType, GlobalIndexType>::read_distributed(
...
+    exec->run(matrix::make_separate_local_nonlocal(
         data, make_temporary_clone(exec, row_partition).get(),
         make_temporary_clone(exec, col_partition).get(), local_part,
         local_row_idxs, local_col_idxs, local_values, non_local_row_idxs,
-        non_local_col_idxs, non_local_values, recv_gather_idxs_,
-        recv_sizes_array, non_local_to_global_));
+        global_non_local_col_idxs, non_local_values));
+
+    auto imap = index_map<local_index_type, global_index_type>(
+        exec, col_partition, comm.rank(), global_non_local_col_idxs);
```
MarcelKoch, line 287: I'm quite confident that it will become necessary to store this map in the matrix class in the future. I'm not doing that for now, because with the distributed PGM, there are now ways to create a non-empty matrix without this read function. I don't know how to create the mapping in the PGM case, so there would be non-empty matrices, but with empty mappings. 
I didn't think this is suitable, so I don't store the mapping. But making it a member in the future would either require unnecessary copies or an interface break.
Any thoughts on this @upsj @pratikvn?

upsj: I think it's a fundamental choice - do we want to consider the matrix storing the implicitly block-reordered matrix (every rank has a consecutive range of indices), or the original matrix? I'm fine with the latter, but it may complicate things as mappings global <--> local are more costly. But at the same time, I'm not sure that would ever be a performance bottleneck.

upsj: Can you elaborate on the interface break?

MarcelKoch: interface break: 
```
auto imap = mat->read_distributed(...);
```
will result in an error if `read_distributed` returns again void. The return type could be changed to `const index_map&` but this would result in copies.

upsj: That's not entirely true - you can operate on the block-reordered indexing just fine, imagine computing an index_map after remapping each rank's indices to be consecutive. But that is still doing 90% of the work towards having a full global <--> local index map, so it's probably to better to do it fully.

pratikvn: IMO, just from a properties perspective, I think it makes sense for the matrix class to have an index map. Any function that creates a distributed matrix, should then also be required to update its index_map. I understand that we dont do that for PGM, right now, but I think the interface choices should not be dependent on that. 

So, I think it should be okay to store an index_map in the matrix class. 

TBH, I am not sure why `read_distributed` should return an index_map. I think we should just store it inside the class and have a public getter to access it where necessary.

MarcelKoch: I think the reordering question is a bit separate from the index-map question. As you said, you could still have an index map after reordering. In general, I think reordering can become quite complicated, see the fun image in petsc's user guide: https://petsc.org/release/manual/mat/#block-matrices

benchmark/test/reference/spmv_distributed.profile.stdout
MarcelKoch: I've only checked `(reference|omp)/distributed` for `std::vector` and found only 4 instances. Did you find more? (I'm updating them btw)

upsj: no, I checked the patch, that should be all. The rest happens inside tests

core/distributed/matrix.cpp
MarcelKoch: You're right, `adjacent_difference` is weird, same as `partial_sum`. I would still argue that the issue here is just that our arrays are not easy to access. If it would read
```
    for (size_type i = 0; i < ecv_targets.size(); ++i) {
        recv_sizes_[recv_targets[i]] = offsets[i + 1] - offsets[i];
    }
```
I don't think you would have the same issue. Only because we have to clutter it with `host_` prefixes, and use very verbose function names it becomes an issue.

upsj: My comment is mostly not about the complexity of the code (`.get_const_data()[]` vs `[]`) but about mixing low-level code with high-level code. I do agree that it would be nice to have a cleaner access to arrays. If your extracted function takes pointers, you would arrive at that code though.

Maybe we could add a `host_array` type that is derived from an `array`, just with `operator[]` and `begin/end()` functions?

benchmark/test/reference/spmv_distributed.profile.stdout
MarcelKoch: I see. Maybe also providing the maximal storage might be interesting to add in the future.

upsj: Yes, good point!

core/distributed/matrix.cpp
MarcelKoch: I think for now I will just remove the return type. The index map is still helpful, since it simplifies the matrix kernels, while the information necessary for PGM is still available. Thus we could add the index map as a member at a later point if we want to.

core/distributed/matrix.cpp
MarcelKoch: I've changed this, but maybe not exactly how you imagined it. I'm using a lambda for this, since I don't want to put it outside the function, due to it's limited use.

PR #1545, opened 2024-02-14
Add segmented array type

MarcelKoch:
This PR adds a class to store a segmented array. A flat ginkgo array is partitioned into multiple segments by an additional index offset array. The segment `i` starts within the flat buffer at index `offsets[i]`, and ends (exclusively) at index `offsets[i + 1]`. The class only provides access to the flat buffer and the offsets.

Used in: #1543 

PR Stack:

- [ ] #1545 
- [ ] #1543
- [ ] #1579
- [ ] #1544 
- [ ] #1588  
- [ ] #1589 

---
#### Old Description

This PR adds a type `collection::array` which is a `std::vector` like class that stores multiple arrays. These arrays are all subviews of a shared buffer.

I think technically this could be considered as a `batch::array`, but I'm not sure if we want to go that way. As a `batch::array` this would interleave the dependencies between the batched and non-batched stuff, and I'm not sure if that is a good idea.

### DIFF
--- a/core/base/segmented_array.cpp
+++ b/core/base/segmented_array.cpp
@@ -1,0 +1,187 @@
+// SPDX-FileCopyrightText: 2017 - 2024 The Ginkgo authors
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <ginkgo/core/base/segmented_array.hpp>
+
+
+#include "core/base/array_access.hpp"
+#include "core/components/prefix_sum_kernels.hpp"
+
+
+namespace gko {
+namespace {
+
+
+GKO_REGISTER_OPERATION(prefix_sum, components::prefix_sum_nonnegative);
+
+
+}
+
+
+template <typename T>
+size_type segmented_array<T>::get_size() const
+{
+    return buffer_.get_size();
+}
+
+
+template <typename T>
+size_type segmented_array<T>::get_segment_count() const
+{
+    return offsets_.get_size() ? offsets_.get_size() - 1 : 0;
+}
+
+
+template <typename T>
+T* segmented_array<T>::get_flat_data()
+{
+    return buffer_.get_data();
+}
+
+
+template <typename T>
+const T* segmented_array<T>::get_const_flat_data() const
+{
+    return buffer_.get_const_data();
+}
+
+
+template <typename T>
+const gko::array<int64>& segmented_array<T>::get_offsets() const
+{
+    return offsets_;
+}
+
+
+template <typename T>
+std::shared_ptr<const Executor> segmented_array<T>::get_executor() const
+{
+    return buffer_.get_executor();
+}
+
+
+template <typename T>
+segmented_array<T>::segmented_array(std::shared_ptr<const Executor> exec)
+    : buffer_(exec), offsets_(exec, 1)
+{
+    offsets_.fill(0);
+}
+
+
+array<int64> sizes_to_offsets(const gko::array<int64>& sizes)
+{
+    auto exec = sizes.get_executor();
+    array<int64> offsets(exec, sizes.get_size() + 1);
+    exec->copy(sizes.get_size(), sizes.get_const_data(), offsets.get_data());
+    exec->run(make_prefix_sum(offsets.get_data(), offsets.get_size()));
+    return offsets;
+}
+
+
+template <typename T>
+segmented_array<T> segmented_array<T>::create_from_sizes(
+    const gko::array<int64>& sizes)
+{
+    return create_from_offsets(sizes_to_offsets(sizes));
+}
+
+
+template <typename T>
+segmented_array<T> segmented_array<T>::create_from_sizes(
+    gko::array<T> buffer, const gko::array<int64>& sizes)
+{
+    return create_from_offsets(std::move(buffer), sizes_to_offsets(sizes));
+}
+
+
+template <typename T>
+segmented_array<T> segmented_array<T>::create_from_offsets(
+    gko::array<int64> offsets)
+{
+    GKO_THROW_IF_INVALID(offsets.get_size() > 0,
+                         "The offsets for segmented_arrays require at least "
+                         "one element.");
+    auto size =
+        static_cast<size_type>(get_element(offsets, offsets.get_size() - 1));
+    return create_from_offsets(array<T>{offsets.get_executor(), size},
+                               std::move(offsets));
+}
+
+
+template <typename T>
+segmented_array<T> segmented_array<T>::create_from_offsets(
+    gko::array<T> buffer, gko::array<int64> offsets)
+{
+    GKO_ASSERT_EQ(buffer.get_size(),
+                  get_element(offsets, offsets.get_size() - 1));
+    segmented_array<T> result(buffer.get_executor());
+    result.offsets_ = std::move(offsets);
+    result.buffer_ = std::move(buffer);
+    return result;
+}
+
+
+template <typename T>
+segmented_array<T>::segmented_array(std::shared_ptr<const Executor> exec,
+                                    segmented_array&& other)
+    : segmented_array(exec)
+{
+    *this = std::move(other);
+}
+
+
+template <typename T>
+segmented_array<T>::segmented_array(std::shared_ptr<const Executor> exec,
+                                    const segmented_array& other)
+    : segmented_array(exec)
+{
+    *this = other;
+}
+
+
+template <typename T>
+segmented_array<T>::segmented_array(const segmented_array& other)
+    : segmented_array(other.get_executor())
+{
+    *this = other;
+}
+
+
+template <typename T>
+segmented_array<T>::segmented_array(segmented_array&& other)
+    : segmented_array(other.get_executor())
+{
+    *this = std::move(other);
+}
+
+
+template <typename T>
+segmented_array<T>& segmented_array<T>::operator=(const segmented_array& other)
+{
+    if (this != &other) {
+        buffer_ = other.buffer_;
+        offsets_ = other.offsets_;
+    }
+    return *this;
+}
+
+
+template <typename T>
+segmented_array<T>& segmented_array<T>::operator=(segmented_array&& other)
+{
+    if (this != &other) {
+        buffer_ = std::move(other.buffer_);
+        offsets_ = std::exchange(other.offsets_,
+                                 array<int64>{other.get_executor(), {0}});
+    }
+    return *this;
+}
+
+
+#define GKO_DECLARE_SEGMENTED_ARRAY(_type) class segmented_array<_type>
+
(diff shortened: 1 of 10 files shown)


upsj:
Do you actually need these individual sub-ranges to be materialized as `array` views? Because otherwise this is just the typical row_ptr - col_idxs structure, and we might be better served with a fully functional `std::span` replacement and a range-of-ranges interface

MarcelKoch:
At least for CPU kernels the arrays are quite handy. But you're right, something like `std::span` would be better suited.

upsj:
I've had it in my mind to provide a nice abstraction over these segmented arrays (including range-for support) for a while already, if you want I can resurrect my prototype code

MarcelKoch:
Sure, if it's not too much work, we could use that. I can also take that over if you want.

MarcelKoch:
@upsj maybe this could be formulated using ranges and accessors. Accessing the segmented array could be thought of as `[i][j]`, where `i` is the array index, and `j` the index within the array. But I don't know enough of ranges and accessor to judge if that would be appropriate.

include/ginkgo/core/base/collection.hpp
```diff
@@ -0,0 +1,200 @@
...
+
+/**
+ * \brief A minimal std::vector-like interface for a collection of arrays.
+ *
+ * The arrays are stored as sub-views of a single contiguous array.
+ *
+ * \tparam T value type stored in the arrays
+ */
+template <typename T>
+struct array {
```
upsj, line 24: I think I would call this a _segmented array_, since the corresponding operations are referred to a segmented sort/reduction/... in papers and the cub library. That would also avoid the need for a separate namespace

include/ginkgo/core/base/collection.hpp
```diff
@@ -0,0 +1,200 @@
...
+        offsets_.fill(0);
+        std::partial_sum(sizes.begin(), sizes.end(), offsets_.get_data() + 1);
+        buffer_.resize_and_reset(offsets_.get_data()[sizes.size()]);
+        // can't use emplace/push_back because then the view will get copied
+        // into owning arrays
+        elems_.resize(sizes.size());
+        for (size_type i = 0; i < sizes.size(); ++i) {
+            elems_[i] = make_array_view(
+                exec, sizes[i], buffer_.get_data() + offsets_.get_data()[i]);
+        }
```
upsj, line 95: I think this design might encourage poor performance - it invites using a kernel invocation for each MPI rank, which might scale poorly. Instead I would try to do this with a single kernel for all ranks, and a corresponding device view.

include/ginkgo/core/base/collection.hpp
```diff
@@ -0,0 +1,200 @@
...
+    const gko::array<size_type>& get_offsets() const { return offsets_; }
+
+    std::shared_ptr<const Executor> get_executor() const
+    {
+        return buffer_.get_executor();
+    }
+
+private:
+    gko::array<T> buffer_;
+    gko::array<size_type> offsets_;
```
upsj, line 128: Considering #430, I would prefer if we used int64 or even a generic integer type.

upsj (commented):
I think this is a useful construct, I would try to make it more generic in some contexts (index type) and more restricted in others (host-accessibility). I would imagine a device adapter based on what I implemented as `segmented_range` in #1582 might be useful here.

Also we would need some tests and documentation before we can merge this

pratikvn (commented):
I agree with Tobias that calling it `segmented_array` would be better. Otherwise, I dont have any other concerns.

core/base/segmented_array.cpp
```diff
@@ -0,0 +1,89 @@
...
+    : buffer_(exec), offsets_(exec)
+{}
+
+
+template <typename T>
+segmented_array<T>::segmented_array(std::shared_ptr<const Executor> exec,
+                                    const std::vector<int64>& sizes)
+    : segmented_array(
+          gko::array<T>(exec,
+                        std::accumulate(sizes.begin(), sizes.end(), int64{})),
```
upsj, line 25: static_cast to size_type

core/base/segmented_array.cpp
```diff
@@ -0,0 +1,89 @@
...
+}
+
+
+template <typename T>
+segmented_array<T>::segmented_array(array<T> buffer, gko::array<int64> offsets)
+    : buffer_(std::move(buffer)), offsets_(std::move(offsets))
+{
+    offsets_.set_executor(this->get_executor());
+    GKO_ASSERT_EQ(buffer_.get_size(),
+                  offsets_.get_size()
```
upsj, line 50: offsets_.size() should never be 0

core/base/segmented_array.cpp
```diff
@@ -0,0 +1,89 @@
...
+}
+
+
+template <typename T>
+segmented_array<T>::segmented_array(gko::array<int64> offsets)
+    : buffer_(offsets.get_executor()), offsets_(std::move(offsets))
+{
+    offsets_.set_executor(this->get_executor());
+    buffer_.resize_and_reset(
+        offsets_.get_size() ? get_element(offsets_, offsets_.get_size() - 1)
```
upsj, line 62: same here

include/ginkgo/core/base/segmented_array.hpp
```diff
@@ -0,0 +1,236 @@
...
+size_type segmented_array<T>::get_size() const
+{
+    return buffer_.get_size();
+}
+
+
+template <typename T>
+size_type segmented_array<T>::get_segment_size() const
+{
+    return offsets_.get_size() ? offsets_.get_size() - 1 : 0;
```
upsj, line 144: same here, also should that be segment_count?

core/base/segmented_array.cpp
MarcelKoch: I was also thinking that. I decided against that, because then I would need to allocate a new array during a move. Do you think that would be acceptable?

MarcelKoch:
@upsj I don't think this will change with the segmented ranges at all. The ranges are only relevant on the backend side. The `index_map` has this type as constructor parameter, and accessors to it, so I would not put it into the detail namespace.

include/ginkgo/core/base/segmented_array.hpp
MarcelKoch: I'm getting a bit tired of all the rebasing, would it be ok to leave the name as it is?

pratikvn:
Maybe also update the PR name ?

include/ginkgo/core/base/segmented_array.hpp
```diff
@@ -0,0 +1,245 @@
...
+    explicit segmented_array(std::shared_ptr<const Executor> exec);
+
+    /**
+     * Creates an uninitialized segmented array with predefined segment sizes.
+     *
+     * @param exec  executor for storage arrays
+     * @param sizes  the sizes of each segment
+     */
+    segmented_array(std::shared_ptr<const Executor> exec,
+                    const std::vector<int64>& sizes);
```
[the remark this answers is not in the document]
MarcelKoch: you're right. I don't remember why I used `std::vector`. I will also use `create_from_*` functions instead.

include/ginkgo/core/base/segmented_array.hpp
```diff
@@ -0,0 +1,245 @@
+// SPDX-FileCopyrightText: 2017 - 2024 The Ginkgo authors
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+#pragma once
```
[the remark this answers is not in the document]
MarcelKoch: These update-every-file changes are always a bit annoying. I would pick up #1466 after the release, perhaps this could be addressed then to keep the disruption minimal.

include/ginkgo/core/base/segmented_array.hpp
MarcelKoch: changed it, since there were other updates anyway.

pratikvn (commented):
Mostly looks good. Just one question regarding the `_const_x` naming for const getters.

include/ginkgo/core/base/segmented_array.hpp
```diff
@@ -0,0 +1,246 @@
...
+    void operator()(pointer ptr) const
+    {
+        *original_ = *ptr;
+        delete ptr;
+    }
+
+private:
+    pointer original_;
+};
+
```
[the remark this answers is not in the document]
MarcelKoch: do you mean the `copy_back_deleter`? For that one it makes sense, but the temorary clone implementations are too different.

include/ginkgo/core/base/segmented_array.hpp
```diff
@@ -0,0 +1,246 @@
...
+     * @return  the flat buffer
+     */
+    T* get_flat_data();
+
+    /**
+     * Const-access to the flat buffer
+     *
+     * @return  the flat buffer
+     */
+    const T* get_flat_data() const;
```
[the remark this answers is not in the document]
MarcelKoch: How about I keep the two overloads and just add `get_const_*`? For me the overloaded versions are much easier to use in generic programming, so I would prefer to keep them.

include/ginkgo/core/base/segmented_array.hpp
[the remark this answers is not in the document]
pratikvn: Yes, that is true. Just for the `copy_back_deleter`.

include/ginkgo/core/base/segmented_array.hpp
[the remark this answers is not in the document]
pratikvn: Yes, I am okay with that.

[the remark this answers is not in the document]
upsj: Don't want to block this, but do you need both a mutable and non-mutable version in the same generic code? I would assume in that case it should also be clear whether the object is `const` or not?

core/base/segmented_array.cpp
upsj: Our moves are mostly not `noexcept` anyways, because we can have cross-executor copies when move-assigning. Performance-wise, allocating <= 8 bytes shouldn't really matter, unless we do really weird things.

upsj (approved):

pratikvn (approved):

include/ginkgo/core/base/segmented_array.hpp
```diff
@@ -0,0 +1,225 @@
...
+size_type segmented_array<T>::get_segment_count() const
+{
+    return offsets_.get_size() ? offsets_.get_size() - 1 : 0;
+}
+
+
+template <typename T>
+T* segmented_array<T>::get_flat_data()
+{
+    return buffer_.get_data();
```
pratikvn, line 159: You can move this directly within the class, right ?

MarcelKoch: I can actually move them into the .cpp. I did that, but I won't wait for the full CI pipeline to merge it.

PR #1546, opened 2024-02-14
Adds sparse communicator class

MarcelKoch:
Based on #1544 

This PR adds a class that handles the communication of a local vector. It uses the neigborhood non-blocking all-to-all for this. The sparse comm is constructed from an `index_map`.

In another PR this will be used in the distributed matrix.


MarcelKoch:
@upsj I think this PR might be a good opportunity to discuss your `distributed::RowGather` idea. In principle this class is already that, but not in the lin-op hierarchy.

include/ginkgo/core/distributed/sparse_communicator.hpp
```diff
@@ -0,0 +1,89 @@
...
+namespace gko {
+namespace experimental {
+namespace distributed {
+
+
+/**
+ * Simplified MPI communicator that handles only neighborhood all-to-all
+ * communication.
+ */
+class sparse_communicator
```
pratikvn, line 32: If it only provides a neighborhood all to all communication, then I think calling it a sparse communicator might be misleading.

include/ginkgo/core/distributed/sparse_communicator.hpp
```diff
@@ -0,0 +1,89 @@
...
+     * Executes non-blocking neighborhood all-to-all.
+     *
+     * @return mpi::request, the recv_buffer is in a valid state only after
+     *         request.wait() has finished
+     */
+    template <typename ValueType>
+    mpi::request communicate(
+        const matrix::Dense<ValueType>* local_vector,
+        const detail::DenseCache<ValueType>& send_buffer,
+        const detail::DenseCache<ValueType>& recv_buffer) const;
```
pratikvn, line 56: A public method of a public class should not be taking objects within the `detail` namespace. Probably need to generalize it to array or Dense ?

core/distributed/sparse_communicator.cpp
```diff
@@ -0,0 +1,275 @@
...
+{
+    // create temporary inverse sparse communicator
+    MPI_Comm sparse_comm;
+    MPI_Info info;
+    GKO_ASSERT_NO_MPI_ERRORS(MPI_Info_create(&info));
+    GKO_ASSERT_NO_MPI_ERRORS(MPI_Dist_graph_create_adjacent(
+        comm.get(), send_ids.get_size(), send_ids.get_const_data(),
+        MPI_UNWEIGHTED, recv_ids.get_size(), recv_ids.get_const_data(),
+        MPI_UNWEIGHTED, info, false, &sparse_comm));
+    GKO_ASSERT_NO_MPI_ERRORS(MPI_Info_free(&info));
```
pratikvn, line 87: Why not 
```suggestion
    GKO_ASSERT_NO_MPI_ERRORS(MPI_Dist_graph_create_adjacent(
        comm.get(), send_ids.get_size(), send_ids.get_const_data(),
        MPI_UNWEIGHTED, recv_ids.get_size(), recv_ids.get_const_data(),
        MPI_UNWEIGHTED, MPI_INFO_NULL, false, &sparse_comm));
```
since no key-value pairs are being set anyway ?

core/distributed/sparse_communicator.cpp
```diff
@@ -0,0 +1,275 @@
...
+    bool use_host_buffer = mpi::requires_host_buffer(exec, comm);
+    return use_host_buffer ? exec->get_master() : exec;
+}
+
+
+template <typename ValueType>
+mpi::request sparse_communicator::communicate(
+    const matrix::Dense<ValueType>* local_vector,
+    const detail::DenseCache<ValueType>& send_buffer,
+    const detail::DenseCache<ValueType>& recv_buffer) const
```
pratikvn, line 199: Looking at this now, I dont get why the sparse_communicator class handles this specific communicate function.

pratikvn (changes requested):
I think this class structure mixes a few things together: the `communicator` class idea, with the specific dense communicate (involving a all_to_all_v). I think it would make sense to separate the two. 

I think it should be possible to make the sparse_communicator a first class communicator and overload only those functions as necessary, for example, all_to_all_v in this case. The distributed object can then just use sparse communicator as a normal communicator and inherits the better all_to_all_v as it uses the sparse communicator for communication. For all other functions where the sparse communicator does not make sense (MPI_Send etc), you transparently use the underlying default communicator.

core/distributed/sparse_communicator.cpp
MarcelKoch: One MPI implementation, I can't remember on which system, crashed without it, so I just kept it.

include/ginkgo/core/distributed/sparse_communicator.hpp
MarcelKoch: The issue is that if the communicator stores the caches (which I also prefer) then the communicator can handle only one async communication at a time. Any new communication would need to wait until the internal buffers are ready to access.

MarcelKoch:
This is superseded by #1588 and #1589

PR #1549, opened 2024-02-16
Use sparse communicator in distributed::Matrix

MarcelKoch:
Based on: #1546 

This PR enables using the sparse-communicator in the `distributed::Matrix`. It is only available if the C++17 standard is used, otherwise the old implementation will be used. But I will also deprecate the old implementation, so we won't maintain two separate code paths.

PR Stack:

- [ ] #1545 
- [ ] #1543
- [ ] #1579
- [ ] #1544 
- [ ] #1548 
- [ ] #1546 
- [ ] #1549

### DIFF
--- a/include/ginkgo/core/distributed/matrix.hpp
+++ b/include/ginkgo/core/distributed/matrix.hpp
@@ -523,7 +523,10 @@ protected:
      *                 shared values is automatically extracted.
      * @return  MPI request for the non-blocking communication.
      */
-    mpi::request communicate(const local_vector_type* local_b) const;
+    [[deprecated(
+        "The pre-C++17 interface of the distributed::Matrix is "
+        "deprecated, please update your C++ standard to 17. ")]] mpi::request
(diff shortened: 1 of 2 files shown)


pratikvn:
This probably needs to be rebased ?

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -521,19 +523,27 @@ class Matrix
      *                 shared values is automatically extracted.
      * @return  MPI request for the non-blocking communication.
      */
-    mpi::request communicate(const local_vector_type* local_b) const;
+    [[deprecated(
+        "The pre-C++17 interface of the distributed::Matrix is "
+        "deprecated, please update your C++ standard to 17. ")]] mpi::request
+    communicate(const local_vector_type* local_b) const;
+#endif
```
pratikvn, line 530: I think the `communicate` function should be the responsibility of this class rather than the sparse_communicator class.

PR #1551, opened 2024-02-16
Fixes preconditioner Benchmark

MarcelKoch:
This PR fixes the preconditioner benchmark by storeing only encoded preconditioner names. The benchmark can run successfully in the current state, but it will emit some exceptions.


yhmtsai (approved):
Could you show some examples on it?

upsj (approved):
LGTM! Can this be caught by the tests?

MarcelKoch:
@upsj @yhmtsai I've added now a test, that would fail with the previous implementation,

upsj:
Can confirm :)

PR #1557, opened 2024-02-27
Allow any characters after expected MM header

MarcelKoch:
This PR allows any characters (esp. `\r`) after the expected matrix market header. Additionally, if the regex search failed, the description line is also printed.


core/base/mtx_io.cpp
```diff
@@ -693,7 +693,9 @@ class mtx_io {
             "    <LAYOUT-TYPE>     is one of: coordinate, array\n"
             "    <VALUE-TYPE>      is one of: real, integer, complex, pattern\n"
             "    <LAYOUT-MODIFIER> is one of: general, symmetric, "
-            "skew-symmetric, hermitian\n");
+            "skew-symmetric, hermitian\n"
+            "Header instead: " +
```
upsj, line 697: I think this could be made clearer, e.g. something like "Found the following header line instead:"

upsj (approved):

PR #1558, opened 2024-02-27
Run examples as part of the test suite

upsj:
Similar to #1341, this moves the script we used previously for comparing the example output with its reference into our regular test suite. To enable platform portability, I replaced the Linux-specific commands by a Python script


examples/compare-output.py
```diff
@@ -0,0 +1,52 @@
...
+    # PATH is hopefully never empty
+    env["PATH"] += ";" + ginkgo_path
+else:
+    if "LD_LIBRARY_PATH" in env:
+        env["LD_LIBRARY_PATH"] += ":" + ginkgo_path
+    else:
+        env["LD_LIBRARY_PATH"] = ginkgo_path
+
+result = subprocess.run(
+    args=args, cwd=cwd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, env=env
```
MarcelKoch, line 29: you can add `text=True` or `universal_newlines=True`, then you don't need `decode` below.

examples/compare-output.py
```diff
@@ -0,0 +1,52 @@
...
+    sys.exit(1)
+
+cwd = sys.argv[1]
+ginkgo_path = sys.argv[2]
+reference = open(sys.argv[3]).readlines()
+args = sys.argv[4:]
+env = os.environ.copy()
+if os.name == "nt":
+    # PATH is hopefully never empty
+    env["PATH"] += ";" + ginkgo_path
```
MarcelKoch, line 21: I don't think doing the same as for `LD_LIBRARY_PATH` would be too much here.

.gitlab-ci.yml
```diff
@@ -127,7 +127,6 @@ build/cuda101/openmpi/gcc/all/debug/shared:
     BUILD_HIP: "ON"
     BUILD_TYPE: "Debug"
     BUILD_PAPI_SDE: "ON"
-    RUN_EXAMPLES: "ON"
```
MarcelKoch, line 130: Just want to note that there also needs `GINKGO_BUILD_EXAMPLES` to be set. I think currently this is done a bit implicitly, but hard-coding it to `ON` in the build scripts. But I guess it could be handled more explicitly with presets.

MarcelKoch (approved):

examples/compare-output.py
upsj: PATH on Windows always contains at least `system32` and similar paths, so we don't really need to handle that case

fritzgoebel (approved):
LGTM

PR #1559, opened 2024-02-27
need sycl::ctz for Aurora 2023.12.30.002

pwxy:
When building on Aurora, sycl::ext::intel::ctz is OK for oneapi/eng-compiler/2023.10.15.002, but not for oneapi/eng-compiler/2023.12.30.002 (build will fail)
For oneapi/eng-compiler/2023.12.30.002, need sycl::ext::intel::ctz -> sycl::ctz
sycl::ctz is fine for 2023.10.15.002

I am unable to test any on other SYCL platform.

### DIFF
--- a/dpcpp/components/intrinsics.dp.hpp
+++ b/dpcpp/components/intrinsics.dp.hpp
@@ -39,13 +39,13 @@ __dpct_inline__ int popcnt(uint64 mask) { return sycl::popcount(mask); }
  */
 __dpct_inline__ int ffs(uint32 mask)
 {
-    return (mask == 0) ? 0 : (sycl::ext::intel::ctz(mask) + 1);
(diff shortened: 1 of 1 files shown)


pwxy:
Doesn't look like I can request reviewers, otherwise I would have requested Pratik Nayak.

upsj:
Tagging #1396 here, which contains the same fixes plus a few other deprecation fixes. I think we should get rid of those remainders of old oneAPI versions, since they clog up our build outputs unnecessarily

pratikvn:
Thank you @pwxy for creating the PR and the fixes. It seems that #1396 from @upsj already incorporates these changes. IMO, we can drop support for the older SYCL versions. We will try to merge that PR soon.

upsj:
There are some additional issues with the latest oneAPI version, so I'll do that in #1396 and close this PR

PR #1573, opened 2024-03-18
Add zero-checks to axpy-like operations

upsj:
Operations like $\alpha A x + 0 y$ may propagate NaNs from y to the output despite the 0 coefficient. This can be avoided by checking the beta scaling factors for zero explicitly.

TODO:
- [ ] add reference tests for matrices
- [ ] add generic tests for Dense

### DIFF
--- a/common/unified/matrix/dense_kernels.template.cpp
+++ b/common/unified/matrix/dense_kernels.template.cpp
@@ -77,14 +77,22 @@ void scale(std::shared_ptr<const DefaultExecutor> exec,
         run_kernel(
             exec,
             [] GKO_KERNEL(auto row, auto col, auto alpha, auto x) {
-                x(row, col) *= alpha[col];
+                if (is_zero(zero(alpha[col]))) {
+                    x(row, col) = zero(alpha[col]);
+                } else {
+                    x(row, col) *= alpha[col];
+                }
             },
             x->get_size(), alpha->get_const_values(), x);
     } else {
         run_kernel(
             exec,
             [] GKO_KERNEL(auto row, auto col, auto alpha, auto x) {
-                x(row, col) *= alpha[0];
+                if (is_zero(alpha[0])) {
+                    x(row, col) = zero(alpha[0]);
+                } else {
+                    x(row, col) *= alpha[0];
+                }
             },
             x->get_size(), alpha->get_const_values(), x);
     }
@@ -130,7 +138,9 @@ void add_scaled(std::shared_ptr<const DefaultExecutor> exec,
         run_kernel(
             exec,
             [] GKO_KERNEL(auto row, auto col, auto alpha, auto x, auto y) {
-                y(row, col) += alpha[0] * x(row, col);
+                if (is_nonzero(alpha[0])) {
+                    y(row, col) += alpha[0] * x(row, col);
+                }
             },
             x->get_size(), alpha->get_const_values(), x, y);
     }
@@ -153,7 +163,9 @@ void sub_scaled(std::shared_ptr<const DefaultExecutor> exec,
         run_kernel(
             exec,
             [] GKO_KERNEL(auto row, auto col, auto alpha, auto x, auto y) {
-                y(row, col) -= alpha[0] * x(row, col);
+                if (is_nonzero(alpha[0])) {
+                    y(row, col) -= alpha[0] * x(row, col);
+                }
             },
             x->get_size(), alpha->get_const_values(), x, y);
     }
@@ -170,7 +182,9 @@ void add_scaled_diag(std::shared_ptr<const DefaultExecutor> exec,
     run_kernel(
         exec,
         [] GKO_KERNEL(auto i, auto alpha, auto diag, auto y) {
-            y(i, i) += alpha[0] * diag[i];
+            if (is_nonzero(alpha[0])) {
+                y(i, i) += alpha[0] * diag[i];
+            }
         },
         x->get_size()[0], alpha->get_const_values(), x->get_const_values(), y);
 }
@@ -186,7 +200,9 @@ void sub_scaled_diag(std::shared_ptr<const DefaultExecutor> exec,
     run_kernel(
         exec,
         [] GKO_KERNEL(auto i, auto alpha, auto diag, auto y) {
-            y(i, i) -= alpha[0] * diag[i];
+            if (is_nonzero(alpha[0])) {
+                y(i, i) -= alpha[0] * diag[i];
+            }
         },
         x->get_size()[0], alpha->get_const_values(), x->get_const_values(), y);
 }
--- a/reference/matrix/ell_kernels.cpp
+++ b/reference/matrix/ell_kernels.cpp
@@ -109,8 +109,8 @@ void advanced_spmv(std::shared_ptr<const ReferenceExecutor> exec,
 
     for (size_type j = 0; j < c->get_size()[1]; j++) {
         for (size_type row = 0; row < a->get_size()[0]; row++) {
-            arithmetic_type result = c->at(row, j);
-            result *= beta_val;
+            arithmetic_type result =
+                is_zero(beta_val) ? zero(beta_val) : beta_val * c->at(row, j);
             for (size_type i = 0; i < num_stored_elements_per_row; i++) {
                 arithmetic_type val = a_vals(row + i * stride);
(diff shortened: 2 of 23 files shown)


fritzgoebel (approved):
LGTM

reference/matrix/ell_kernels.cpp
```diff
@@ -109,8 +109,8 @@ void advanced_spmv(std::shared_ptr<const ReferenceExecutor> exec,
 
     for (size_type j = 0; j < c->get_size()[1]; j++) {
         for (size_type row = 0; row < a->get_size()[0]; row++) {
-            arithmetic_type result = c->at(row, j);
-            result *= beta_val;
+            arithmetic_type result =
+                is_zero(beta_val) ? zero(beta_val) : beta_val * c->at(row, j);
```
MarcelKoch, line 113:
```suggestion
                is_zero(beta_val) ? zero(beta_val) : beta_val * static_cast<arithmetic_type>(c->at(row, j));
```
to make it compile

MarcelKoch (approved):
lgtm in general. Maybe there are still some `static_cast` missing to make it compile.

yhmtsai:
Is there any reason to avoid propagation of NaN? If there's no performance penalty, I think propagation of NaN is easier to know the algorithm does not work out due to some arthmetic error.

MarcelKoch:
@yhmtsai if you are computing for example `y = 0.5 * A * x + 0.0 * y` then propagating NaNs from `y` is unnecessary, since it's mathematical equivalent to just leaving `y` out. This can easily happen, if `y` is not initialized before that computation.

yhmtsai:
no, 0 * NaN should be NaN not zero, so it is not mathimatical equality by just leaving them out.
Yes, it might happen in unitialized memory, but I would say it should be properly initialized or using the call without touching the unitialization put. (for us, we may proparbably provide A->apply(alpha, x, y) for y = alpha * A * x)

yhmtsai:
I know current vendor library usually treat 0 as do not touch due to BLAS.
I am not sure the other routines hold the same rule

upsj:
> 0 * NaN should be NaN not zero

that makes calculations more fragile, we already do similar things (special cases) for zeros inside our solver kernels

MarcelKoch:
@yhmtsai we treat `0 * NaN = 0` only in the case of axpy-like operation. So there will still be NaN propagation in normal SpMV, dot-products, etc. But for these axpy-like operations, users will not care about the IEEE standard. For them, initializing `x` by setting it to 0 and multiplying it with 0 should be equivalent.

common/cuda_hip/matrix/csr_kernels.template.cpp
MarcelKoch, line 1: Just a mental note for me: the load balanced spmv is also covered, since it used the `Dense::scale` function if `beta != nullptr`.

common/unified/matrix/dense_kernels.template.cpp
```diff
@@ -76,14 +76,22 @@ void scale(std::shared_ptr<const DefaultExecutor> exec,
         run_kernel(
             exec,
             [] GKO_KERNEL(auto row, auto col, auto alpha, auto x) {
-                x(row, col) *= alpha[col];
```
MarcelKoch, line 79: I'm wondering if changes like this have a performance impact. I'm a bit skeptical that the compiler is capable of hoisting the inner if-check out of the hot loop. But I also don't want to introduce another if nesting on top.

MarcelKoch (approved):

common/unified/matrix/dense_kernels.template.cpp
upsj: Since this operates column-wise, I don't see an immediate optimization potential. On the other hand, this can be implemented without an actual branch with a CMOV on x86 or something similar on GPUs to avoid the branch overhead. Should even be possible to vectorize with mask registers, but I'm not sure if the auto-vectorizer can already do that.

upsj: No longer relevant, I removed this change

common/cuda_hip/matrix/csr_kernels.template.cpp
yhmtsai: It is inconsistent in SpMV, which should be the same up to the rounding error IMO

yhmtsai:
In AXPBY SpMV, it takes (nnz + 3n) ValueType + (nnz + n + 1) IndexType 
I think the condition are the same in kernel, so we do not have divengent branch but it might change some kerrnel register usage or optimization.
it only changes from  (nnz + 3n) ValueType + (nnz + n + 1) IndexType to  (nnz + 2n) ValueType + (nnz + n + 1) IndexType
if the right hand side is not able to stay in cache, we might still need nnz for the right hand side vector
we will only change from  (2nnz + 2n) ValueType + (nnz + n + 1) IndexType to  (2nnz + n) ValueType + (nnz + n + 1) IndexType
I do not think the benefit worth enough such that user might not notice there are some nan turns out.
To me, the case we discuss in meeting is that user set the beta to zero to use $alpha A x$ not do $Ax$ then scale, where we should provide apply(A, x, y) for that.
I know BLAS treats zero as non-operation (without documentation though?) and many vendors follow that.
To be clear, there is no doubt that we should still aim for performance not nan propagation issue.
However, if there's no more work, we should try to keep propagation if it can help user know something wrong not luckily get the solution.
In this pr, we do more work to avoid the propagation, which seems to be opposite direction.
That's why I am not in favor of this pr.

upsj:
The behavior seems well-documented, e.g. with [GEMV](https://www.netlib.org/lapack/explore-html-3.6.1/d7/d15/group__double__blas__level2_gadd421a107a488d524859b4a64c1901a9.html)

> When BETA is supplied as zero then Y need not be set on input.

This is also not meant as a performance optimization, rather a safety feature. The argument is mainly that this doesn't have a negative performance impact.

What sources of NaNs can we imagine? Either they come from uninitialized memory, or they come from divergent operations, any others I am missing? Regardless of the source, in an application, if I multiply something with an explicit zero, I as a programmer am saying that I don't care what these values were.

I'm having a hard time imagining inside Ginkgo's solvers where we would end up with a zero scalar that could cause problematic NaNs to disappear. As soon as a NaN appears in a vector, it will infect scalars through reductions and other vectors via updates, I really don't see a way how a scalar could become zero _after_ a NaN was introduced in a vector, which is where this issue would pop up. Can you think of a case? I have the impression that we are discussing something that will not happen in practice.

upsj:
Here is the paper Pratik referred to in today's meeting: [Proposed Consistent Exception Handling for the BLAS and LAPACK](https://arxiv.org/abs/2207.09281)

yhmtsai:
We seem to use opposite view to see this problem like user-specific vs numerical error during application.
In my case, it will be more from the divergent case.
It is also a range issue of precision. It quite often some vector goes to inf and them we have some scalar is divided by norm (=zero). It may gives zero and inf in the first place but later start to introduce NaN (such as 0*inf).
zero and inf is before nan. nan is usually introduced by zero and inf combination.
It is in the whole application. It is good that we can catch just whenever it happens, but adding check to each operation affects the performance. We definitely not go this direction.
Users specify zero to me is a intentional action, so they should be able to call another function to perform desired operation. 
rather call another function than set a scalar to zero. 
Currently it mixed the user-defined (explicit) zero and normal (implicit during application) zero.
Good to know they have described the behavior in the document, which reduces my concern a bit.

MarcelKoch:
> Users specify zero to me is a intentional action, so they should be able to call another function to perform desired operation. rather call another function than set a scalar to zero.

Just a note, in some cases users don't have any other option. For example, if you want to store the result of $y = \alpha M x$, then users can (and should) only use the advanced apply with $\beta = 0$. We don't need an extra function for that, we just need the advanced apply to do what is mathematically expected.
I can also imagine that users from a more mathematical background will assume that `y.scale(0)` and `y.fill(0)` will behave the same.

yhmtsai:
That's the reason why I said we should provide `apply(alpha, b, x)` for this behavior. 
from the function call, it is written in $y = \alpha M x + 0 * y$ not $\alpha M x$
If they consider the number can always be represented in computer and do not know anything about the initialization, then it might be a possible thought.
With some programming knowledge, they propabably know it can be anything if they do not initialize value and operate on it.

PR #1578, opened 2024-03-22
Fix misaligned addresses with batched loggers

### LINKED ISSUE #1576: CUDA error with batched solvers
Using Ginkgo batched solvers I sometimes get the following errors

```bash
terminate called after throwing an instance of 'gko::CudaError'
  what():  /tmp/soft/spack-stage/spack-stage-ginkgo-1.7.0-a6wuqlgaxa2t5vkyzdapmgicl73ujdjd/spack-src/cuda/base/executor.cpp:203: synchronize: cudaErrorMisalignedAddress: misaligned address
Abandon
```
This error appears on Ginkgo 1.7.0 release on a V100 GPU with CUDA 12.2.1.

You can find a minimal reproducer in this repository https://github.com/Geoflow/gko_minreproducer/tree/minimal-reproducer. In my case I got the error with a batch size of 1. Am I doing something wrong ?

cc @Geoflow

pratikvn:
This PR fixes errors with cuda misaligned addresses in the batched loggers when mixing 32 bit and 64 bit types reported in #1576. It adds a setup to allow aliasing workspace pointers with arrays of different types, which is necessary to prevent repeated allocations while preventing misaligned issues.

Fixes #1576

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,7 +22,8 @@ repos:
       (?x)^(
         third_party/SuiteSparse/AMD/.*|
         third_party/identify_stream_usage/.*|
-        examples/external-lib-interfacing/external-lib-interfacing.cpp
+        examples/external-lib-interfacing/external-lib-interfacing.cpp|
+        core/base/workspace_aliases.hpp
       )$
   - id: format-headers
     name: format headers
@@ -34,7 +35,8 @@ repos:
         (?x)^(
           third_party/SuiteSparse/AMD/.*|
           third_party/identify_stream_usage/.*|
-          include/ginkgo/ginkgo.hpp
+          include/ginkgo/ginkgo.hpp|
+          core/base/workspace_aliases.hpp
         )$
   - id: update-ginkgo-header
     name: update ginkgo header
--- a/core/base/workspace_aliases.hpp
+++ b/core/base/workspace_aliases.hpp
@@ -117,6 +117,7 @@ public:
 private:
     GKO_ATTRIBUTES void set_bytes_required(std::size_t new_size)
     {
+        using std::max;
         size_ = max(size_, new_size);
     }
 
--- a/core/base/workspace_aliases.hpp
+++ b/core/base/workspace_aliases.hpp
@@ -8,7 +8,12 @@
 #define GKO_CORE_BASE_WORKSPACE_ALIASES_HPP_
 
 
+#include <algorithm>
+#include <cstdlib>
+
+
 #include <ginkgo/config.hpp>
+#include <ginkgo/core/base/types.hpp>
 
 
 // Macros for handling different device error return types uniformly
--- a/core/base/workspace_aliases.hpp
+++ b/core/base/workspace_aliases.hpp
@@ -2,8 +2,8 @@
 //
 // SPDX-License-Identifier: BSD-3-Clause
 
-#ifndef GKO_PUBLIC_CORE_BASE_WORKSPACE_ALIASES_HPP_
-#define GKO_PUBLIC_CORE_BASE_WORKSPACE_ALIASES_HPP_
+#ifndef GKO_CORE_BASE_WORKSPACE_ALIASES_HPP_
+#define GKO_CORE_BASE_WORKSPACE_ALIASES_HPP_
 
 
 #include <ginkgo/config.hpp>
@@ -252,4 +252,4 @@ private:
 }  // namespace gko
 
 
-#endif  // GKO_PUBLIC_CORE_BASE_WORKSPACE_ALIASES_HPP_
+#endif  // GKO_CORE_BASE_WORKSPACE_ALIASES_HPP_
--- a/core/base/workspace_aliases.hpp
+++ b/core/base/workspace_aliases.hpp
@@ -42,7 +42,7 @@ namespace detail {
 template <int num_allocs, int align_bytes = 8>
 GKO_ATTRIBUTES GKO_INLINE GKO_DEVICE_ERROR_TYPE create_workspace_aliases(
     void* workspace_ptr, size_t& num_bytes, void* (&allocations)[num_allocs],
-    size_t (&allocation_sizes)[num_allocs])
+    const size_t (&allocation_sizes)[num_allocs])
 {
     constexpr int align_mask = ~(align_bytes - 1);
 
--- a/core/base/workspace_aliases.hpp
+++ b/core/base/workspace_aliases.hpp
@@ -9,7 +9,22 @@
 
 
 #include <ginkgo/config.hpp>
-#include <ginkgo/core/base/types.hpp>
+
+
+// Macros for handling different device error return types uniformly
+#if defined(__CUDACC__)
+#define GKO_DEVICE_ERROR_TYPE cudaError_t
+#define GKO_DEVICE_ERROR_INVALID cudaErrorInvalidValue
+#define GKO_DEVICE_NO_ERROR cudaSuccess
+#elif defined(__HIPCC__)
+#define GKO_DEVICE_ERROR_TYPE hipError_t
+#define GKO_DEVICE_ERROR_INVALID hipErrorInvalidValue
+#define GKO_DEVICE_NO_ERROR hipSuccess
+#else
+#define GKO_DEVICE_ERROR_TYPE int
+#define GKO_DEVICE_ERROR_INVALID 1
+#define GKO_DEVICE_NO_ERROR 0
+#endif
 
 
 // This code is a modified version of the code from CCCL
--- a/core/base/workspace_aliases.hpp
+++ b/core/base/workspace_aliases.hpp
@@ -31,6 +31,8 @@
 // (https://github.com/NVIDIA/cccl) (cub/detail/temporary_storage.cuh and
 // cub/temporary_storage.cuh), made available through the Apache-2.0 and BSD-3
 // licenses. See ABOUT-LICENSING.md for more details.
+// The main changes: the namespaces, the object/class
+// names and some simplifications that adhere to Ginkgo's code conventions
 
 
 namespace gko {
--- a/core/base/workspace_aliases.hpp
+++ b/core/base/workspace_aliases.hpp
@@ -1,6 +1,8 @@
-// SPDX-FileCopyrightText: 2017 - 2024 The Ginkgo authors
+// SPDX-FileCopyrightText: 2011 - 2023 NVIDIA CORPORATION
+// SPDX-FileCopyrightText: 2021 Duane Merrill
+// SPDX-FileCopyrightText: 2024 The Ginkgo authors
 //
-// SPDX-License-Identifier: BSD-3-Clause
+// SPDX-License-Identifier: BSD-3-Clause AND Apache-2.0
 
 #ifndef GKO_CORE_BASE_WORKSPACE_ALIASES_HPP_
 #define GKO_CORE_BASE_WORKSPACE_ALIASES_HPP_
--- a/core/log/batch_logger.cpp
+++ b/core/log/batch_logger.cpp
@@ -13,16 +13,63 @@
 namespace gko {
 namespace batch {
 namespace log {
+// namespace detail {
+
+
+// template <typename ValueType>
+// log_data<ValueType>::log_data(std::shared_ptr<const Executor> exec,
+//                               size_type num_batch_items)
+//     : res_norms(exec), iter_counts(exec)
+// {
+//     if (num_batch_items > 0) {
+//         iter_counts.resize_and_reset(num_batch_items);
+//         res_norms.resize_and_reset(num_batch_items);
+//     } else {
+//         GKO_INVALID_STATE("Invalid num batch items passed in");
+//     }
+// }
+
+
+// template <typename ValueType>
+// log_data<ValueType>::log_data(std::shared_ptr<const Executor> exec,
+//                               size_type num_batch_items,
+//                               array<unsigned char>& workspace)
+//     : res_norms(exec), iter_counts(exec)
+// {
+//     const size_type workspace_size =
+//         num_batch_items * (sizeof(real_type) + sizeof(idx_type));
+
+//     if (num_batch_items > 0 && !workspace.is_owning() &&
+//         workspace.get_size() >= workspace_size) {
+//         gko::detail::layout<2> workspace_alias;
+//         auto slot_1 = workspace_alias.get_slot(0);
+//         auto slot_2 = workspace_alias.get_slot(1);
+
+//         // Temporary storage mapping
+//         workspace_alias.map_to_buffer(workspace.get_data(), workspace_size);
+//         iter_counts = array<idx_type>::view(
+//             exec, num_batch_items,
+//             slot_1->create_alias<idx_type>(num_batch_items).get());
+//         res_norms = array<real_type>::view(
+//             exec, num_batch_items,
+//             slot_2->create_alias<real_type>(num_batch_items).get());
+//     } else {
+//         GKO_INVALID_STATE("invalid workspace or num batch items passed in");
+//     }
+// }
+
+
+// }  // namespace detail
 
 
 template <typename ValueType>
 void BatchConvergence<ValueType>::on_batch_solver_completed(
-    const array<int64>& iteration_count,
+    const array<int>& iteration_count,
     const array<remove_complex<ValueType>>& residual_norm) const
 {
     if (this->iteration_count_.get_size() == 0) {
-        this->iteration_count_ = gko::array<int64>(
-            iteration_count.get_executor(), iteration_count.get_size());
+        this->iteration_count_ = gko::array<int>(iteration_count.get_executor(),
+                                                 iteration_count.get_size());
     }
     if (this->residual_norm_.get_size() == 0) {
         this->residual_norm_ = gko::array<remove_complex<ValueType>>(
--- a/core/log/batch_logger.cpp
+++ b/core/log/batch_logger.cpp
@@ -17,12 +17,12 @@ namespace log {
 
 template <typename ValueType>
 void BatchConvergence<ValueType>::on_batch_solver_completed(
-    const array<int>& iteration_count,
+    const array<int64>& iteration_count,
     const array<remove_complex<ValueType>>& residual_norm) const
 {
     if (this->iteration_count_.get_size() == 0) {
-        this->iteration_count_ = gko::array<int>(iteration_count.get_executor(),
-                                                 iteration_count.get_size());
+        this->iteration_count_ = gko::array<int64>(
+            iteration_count.get_executor(), iteration_count.get_size());
     }
     if (this->residual_norm_.get_size() == 0) {
         this->residual_norm_ = gko::array<remove_complex<ValueType>>(
--- a/core/log/batch_logger.cpp
+++ b/core/log/batch_logger.cpp
@@ -10,9 +10,66 @@
 #include <ginkgo/core/base/math.hpp>
 
 
+#include "core/base/workspace_aliases.hpp"
+
+
 namespace gko {
 namespace batch {
 namespace log {
+namespace detail {
+
+
+template <typename ValueType>
+log_data<ValueType>::log_data(std::shared_ptr<const Executor> exec,
+                              size_type num_batch_items)
+    : res_norms(exec), iter_counts(exec)
+{
+    if (num_batch_items > 0) {
+        iter_counts.resize_and_reset(num_batch_items);
+        res_norms.resize_and_reset(num_batch_items);
+    } else {
+        GKO_INVALID_STATE("Invalid num batch items passed in");
+    }
+}
+
+
+template <typename ValueType>
+log_data<ValueType>::log_data(std::shared_ptr<const Executor> exec,
+                              size_type num_batch_items,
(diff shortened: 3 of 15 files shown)


include/ginkgo/core/log/batch_logger.hpp
```diff
@@ -101,7 +103,7 @@ class BatchConvergence final : public gko::log::Logger {
     using mask_type = gko::log::Logger::mask_type;
 
     void on_batch_solver_completed(
-        const array<int>& iteration_count,
+        const array<int64>& iteration_count,
```
yhmtsai, line 106: Is it the bug of logger or misused in the kernel?
we can only change the public interface if it is the first case.

pratikvn: I think the logger needs to store int64 to be efficient. I think we can right now safely assume that no one derives from the BatchConvergence logger.

upsj (commented):
Using 64 bit indices doesn't seem to make any sense for batched problems to me. This looks more like a quick patch than a proper fix. I'll try to prototype an example for how we could improve upon this, I think CUB provides a nice blueprint.

pratikvn:
In this case the maximum number of elements possible is equal to the number of batch items. So, I think it is definitely possible to have more than max(int32) for that. Additionally, this is not a shared memory issue, but an issue of using a workspace and storing pointers to different types within the workspace that was causing the misaligned accesses.

Nevertheless, your suggestion on the CUB approach is a nice way to do it, and I will look into that.

upsj:
This change only deals with final iteration counts, which is where IMO 64 bit integers make no sense.

pratikvn:
Yes, only the final iteration counts are being logged, but the number of entries scales with the number of batch items that are being solved, which can possibly more than `max(int32)`

upsj:
The type you use to index an array is independent of the value type of the array, I am talking about the value type

pratikvn:
Ah, yes. I misunderstood your comment. I am looking into the CUB blueprint. That should be usable for correct aligntment of  both the workspace vectors and for the shared memory objects.

MarcelKoch:
I think the PR description and title need to change now.

pratikvn:
I think both the title and description are still valid. I have updated the description now.

include/ginkgo/core/solver/batch_solver_base.hpp
```diff
@@ -289,8 +289,9 @@ class EnableBatchSolver
             auto id = Identity::create(exec, system_matrix->get_size());
             preconditioner_ = std::move(id);
         }
-        const size_type workspace_size = system_matrix->get_num_batch_items() *
-                                         (sizeof(real_type) + sizeof(int));
+        const size_type workspace_size =
+            system_matrix->get_num_batch_items() * 32;
+        // 2 * (sizeof(real_type) + sizeof(int));
```
MarcelKoch, line 294: So `32 == 2 * (sizeof(real_type) + sizeof(int))`? But that is wrong, `2 * (sizeof(real_type) + sizeof(int)) <= 24`

include/ginkgo/core/log/batch_logger.hpp
```diff
@@ -33,12 +34,11 @@ namespace detail {
 template <typename ValueType>
 struct log_data final {
     using real_type = remove_complex<ValueType>;
+    using idx_type = int;
 
     log_data(std::shared_ptr<const Executor> exec, size_type num_batch_items)
```
MarcelKoch, line 39: all of this implementation can be moved to core. Then we don't need the CUB file in our public header.

include/ginkgo/core/base/types.hpp
```diff
@@ -35,6 +35,22 @@
...
+// Macros for handling different device error return types uniformly
+#if defined(__CUDACC__)
+#define GKO_DEVICE_ERROR_TYPE cudaError_t
+#define GKO_DEVICE_ERROR_INVALID cudaErrorInvalidValue
+#define GKO_DEVICE_NO_ERROR cudaSuccess
+#elif defined(__HIPCC__)
+#define GKO_DEVICE_ERROR_TYPE hipError_t
+#define GKO_DEVICE_ERROR_INVALID hipErrorInvalidValue
+#define GKO_DEVICE_NO_ERROR hipSuccess
+#else
```
MarcelKoch, line 47: Since only succeess/failure is needed, you don't need these wrappers.

include/ginkgo/core/log/batch_logger.hpp
```diff
@@ -51,17 +51,22 @@ struct log_data final {
...
         if (num_batch_items > 0 && !workspace.is_owning() &&
             workspace.get_size() >= workspace_size) {
+            gko::detail::layout<2> workspace_alias;
+            auto slot_1 = workspace_alias.get_slot(0);
+            auto slot_2 = workspace_alias.get_slot(1);
+            auto iter_alias = slot_1->create_alias<idx_type>(num_batch_items);
+            auto res_alias = slot_2->create_alias<real_type>(num_batch_items);
+
+            // Temporary storage mapping
+            workspace_alias.map_to_buffer(workspace.get_data(), workspace_size);
```
MarcelKoch, line 65: return code is ignored.

core/log/batch_logger.cpp
```diff
@@ -13,6 +13,53 @@
 namespace gko {
 namespace batch {
 namespace log {
+// namespace detail {
```
MarcelKoch, line 16: what's going on with this?

MarcelKoch (dismissed):
Can we maybe first summarize the issue? If I understand it correctly, the issue was the `res_norm` pointer in `batch_logger.hpp`. The alignment of that pointer depends on how many batch items there are, and the size in bytes of the `iter_counts`, which are using the workspace memory before the `res_norm`. In some cases (`num_batch_items==1`) this could lead to an alignment for `res_norm` less than 8 bytes.

The workspace manager from CUB fixes the alignment by ensuring that each memory region from the work space is aligned to at least 8 bytes.

TBH, the CUB implementation seems a bit overkill for us. We just need a function that like
```
T* next_aligned(void* ptr)
```
which gives you a pointer to the next aligned (to 8 bytes I guess) memory location. 

If the CUB implementation should stay, it needs a lot more documentation. I'm also not a fan of the chain `layout->slot->alias`, it could be `layout->alias` directly.

core/log/batch_logger.cpp
MarcelKoch: maybe the header was missing some `inline`? Still it should be moved into the core.

include/ginkgo/core/base/workspace_aliases.hpp
```diff
@@ -0,0 +1,252 @@
+// SPDX-FileCopyrightText: 2017 - 2024 The Ginkgo authors
+//
+// SPDX-License-Identifier: BSD-3-Clause
```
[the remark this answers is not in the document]
pratikvn: It seems the formatter does not let me do that.

include/ginkgo/core/log/batch_logger.hpp
```diff
@@ -51,17 +51,22 @@ struct log_data final {
              array<unsigned char>& workspace)
         : res_norms(exec), iter_counts(exec)
     {
-        const size_type workspace_size =
-            num_batch_items * (sizeof(real_type) + sizeof(int));
+        const size_type workspace_size = num_batch_items * 32;
```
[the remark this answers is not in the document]
pratikvn: No, we just want to ensure enough of a large workspace. In this case, we align everything to 8 bytes.

include/ginkgo/core/solver/batch_solver_base.hpp
pratikvn: This comment was wrong. Should be fixed now. But we want to align to 8 bytes, size of `double` to ensure there is no misaligned accesses, which may happen when using the workspace for storing both 4 bytes and 8 bytes (int and double) respectively.

include/ginkgo/core/log/batch_logger.hpp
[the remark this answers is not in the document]
upsj: I will take a look at this PR on Wednesday, but I can already tell you that `thrust::complex<double>` needs an alignment of 16 bytes

include/ginkgo/core/base/workspace_aliases.hpp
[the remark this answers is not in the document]
pratikvn: This is needed in the batch logger header, which is needed for the batch solver base classes which cannot be moved to core.

include/ginkgo/core/log/batch_logger.hpp
pratikvn: Unfortunately, I dont think this can be moved to core.

include/ginkgo/core/base/types.hpp
pratikvn: The plan is to use these for aliasing the shared memory pointers in the device kernels as well. In those cases, we would need the return code which would be useful in narrowing the issue.

include/ginkgo/core/log/batch_logger.hpp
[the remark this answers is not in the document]
pratikvn: In this case, we are aligning to the max real floating type, (should be of `remove_complex<T>`), which is 8. But for the general case, yes, I guess we would need to align to 16.

include/ginkgo/core/base/workspace_aliases.hpp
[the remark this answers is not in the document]
MarcelKoch: This is because of our stupid `format_header.sh` script. Maybe format it correctly with reuse, and then commit it with `-n`.

core/base/workspace_aliases.hpp
```diff
@@ -0,0 +1,255 @@
+// SPDX-FileCopyrightText: 2017 - 2024 The Ginkgo authors
+//
+// SPDX-License-Identifier: BSD-3-Clause
```
MarcelKoch, line 3: I really think this should be different. You can add this file to the exludes in the pre-commit config for the format-headers hook. Then the changes to it won't get overwritten anymore.

core/base/workspace_aliases.hpp
```diff
@@ -0,0 +1,255 @@
...
+class alias;
+
+template <int num_slots, int align_bytes>
+class layout;
+
+/**
+ * Represents a slot of memory which can be aliased. See @alias and @layout on
+ * how this can be used.
+ */
+class slot {
```
MarcelKoch, line 77: still no idea why there is both `slot` and `alias`. `alias` alone should be enough. But I won't hold anything up over the implementations in this file.

include/ginkgo/core/log/batch_logger.hpp
```diff
@@ -33,39 +33,12 @@ namespace detail {
 template <typename ValueType>
 struct log_data final {
     using real_type = remove_complex<ValueType>;
+    using idx_type = int;
```
MarcelKoch, line 36: I wouldn't mind
```suggestion
    using index_type = int;
```
I think that would be more consistent with our naming scheme.

include/ginkgo/core/solver/batch_solver_base.hpp
```diff
@@ -289,8 +289,8 @@ class EnableBatchSolver
             auto id = Identity::create(exec, system_matrix->get_size());
             preconditioner_ = std::move(id);
         }
-        const size_type workspace_size = system_matrix->get_num_batch_items() *
-                                         (sizeof(real_type) + sizeof(int));
+        const size_type workspace_size =
+            system_matrix->get_num_batch_items() * 32;
```
MarcelKoch, line 293: still not entirely sure about where the 32 comes from, but not a blocker.

.pre-commit-config.yaml
```diff
@@ -23,6 +23,7 @@ repos:
         third_party/SuiteSparse/AMD/.*|
         third_party/identify_stream_usage/.*|
         examples/external-lib-interfacing/external-lib-interfacing.cpp
+        core/base/workspace_aliases.hpp
```
MarcelKoch, line 26: why is the header excluded here? I thought it would not change the copyright from previous years.

.pre-commit-config.yaml
```diff
@@ -35,6 +36,7 @@ repos:
           third_party/SuiteSparse/AMD/.*|
           third_party/identify_stream_usage/.*|
           include/ginkgo/ginkgo.hpp
+          core/base/workspace_aliases.hpp
```
MarcelKoch, line 39: I think it has to be
```suggestion
        examples/external-lib-interfacing/external-lib-interfacing.cpp|
        core/base/workspace_aliases.hpp
```

core/log/batch_logger.cpp
```diff
@@ -10,9 +10,66 @@
...
+        gko::detail::layout<2, 8> workspace_alias;
+        auto slot_1 = workspace_alias.get_slot(0);
+        auto slot_2 = workspace_alias.get_slot(1);
+        auto iter_alias = slot_1->create_alias<index_type>(num_batch_items);
+        auto res_alias = slot_2->create_alias<real_type>(num_batch_items);
+
+        // Temporary storage mapping
+        auto err = workspace_alias.map_to_buffer(workspace.get_data(),
+                                                 reqd_workspace_size);
+        GKO_ASSERT(err == 0);
```
MarcelKoch, line 55:
```suggestion
        GKO_ASSERT(err == GKO_DEVICE_NO_ERROR);
```

include/ginkgo/core/base/types.hpp
MarcelKoch: btw, does this have to be in a public header? Wouldn't a core header be sufficient?

.pre-commit-config.yaml
MarcelKoch: that should also fix the CI.

MarcelKoch (changes requested):

pratikvn:
the check-format will fail because it uses develop instead of the updated precommit file

include/ginkgo/core/base/types.hpp
yhmtsai: have the same question as MarcelKoch. the current usage only require these alias in the core header not public.

core/base/workspace_aliases.hpp
```diff
@@ -0,0 +1,272 @@
...
+#define GKO_DEVICE_NO_ERROR hipSuccess
+#else
+#define GKO_DEVICE_ERROR_TYPE int
+#define GKO_DEVICE_ERROR_INVALID 1
+#define GKO_DEVICE_NO_ERROR 0
+#endif
+
+
+// This code is a modified version of the code from CCCL
+// (https://github.com/NVIDIA/cccl) (cub/detail/temporary_storage.cuh and
```
yhmtsai, line 31: maybe also mention what the modified part?
I guess only namespace, error code

include/ginkgo/core/log/batch_logger.hpp
```diff
@@ -33,39 +33,12 @@ namespace detail {
...
-            num_batch_items * (sizeof(real_type) + sizeof(int));
-        if (num_batch_items > 0 && !workspace.is_owning() &&
-            workspace.get_size() >= workspace_size) {
-            iter_counts =
-                array<int>::view(exec, num_batch_items,
-                                 reinterpret_cast<int*>(workspace.get_data()));
-            res_norms = array<real_type>::view(
-                exec, num_batch_items,
-                reinterpret_cast<real_type*>(workspace.get_data() +
-                                             (sizeof(int) * num_batch_items)));
```
yhmtsai, line 64: originally, I thought it should be solved by using the view on the larger type first.

core/log/batch_logger.cpp
```diff
@@ -10,9 +10,66 @@
...
+}
+
+
+template <typename ValueType>
+log_data<ValueType>::log_data(std::shared_ptr<const Executor> exec,
+                              size_type num_batch_items,
+                              array<unsigned char>& workspace)
+    : res_norms(exec), iter_counts(exec)
+{
+    const size_type reqd_workspace_size = num_batch_items * 32;
```
yhmtsai, line 42: If I understand it correctly, the mapping is to ensure the memory address is divisible by the size of data type.
It's also why I think it was enough by viewing the larger data type first.
Assume the workspace data is not aliased from the beginning, which should not be the case if the workspace is the alone allocation, and then the actual memory will need a little more than reqd_workspace_size by cutting the first unaligned part. If the reqd_workspace_size is the same as the workspace allocation, it does not help when the workspace is not aligned from the beginning.

yhmtsai:
By the way, do you have the test such that we face the issue in our pipeline?

reference/test/solver/batch_bicgstab_kernels.cpp
```diff
@@ -229,7 +229,7 @@ TYPED_TEST(BatchBicgstab, CanSolveEllSystem)
             .with_tolerance_type(gko::batch::stop::tolerance_type::relative)
             .on(this->exec);
     const int num_rows = 13;
-    const size_t num_batch_items = 2;
+    const size_t num_batch_items = 1;
```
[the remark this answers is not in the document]
pratikvn: Yes, with num_batch_items=1 and without this fix, current develop gives the error

include/ginkgo/core/log/batch_logger.hpp
pratikvn: I havent tested that, but it is possible, but for cases when we need to have different types, it would not help. This approach is more general.

MarcelKoch:
> the check-format will fail because it uses develop instead of the updated precommit file

This is always so annoying....

MarcelKoch (approved):

reference/test/solver/batch_bicgstab_kernels.cpp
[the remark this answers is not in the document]
yhmtsai: wait, I thought the error is from cuda not from reference. did the error also happen in reference?

include/ginkgo/core/log/batch_logger.hpp
yhmtsai: one `if-else` condition should solve all cases

reference/test/solver/batch_bicgstab_kernels.cpp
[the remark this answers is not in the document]
pratikvn: Yes, you are right. I somehow pushed it in the wrong file. It only fails in cuda and not reference.

include/ginkgo/core/log/batch_logger.hpp
pratikvn: The plan would be to use this in all cases, for shared memory inside the kernels as well, so IMO CUB solution is a better and more general approach.

core/base/workspace_aliases.hpp
```diff
@@ -0,0 +1,274 @@
...
+
+
+namespace gko {
+namespace detail {
+
+
+template <int num_allocs, int align_bytes = 8>
+GKO_ATTRIBUTES GKO_INLINE GKO_DEVICE_ERROR_TYPE create_workspace_aliases(
+    void* workspace_ptr, size_t& num_bytes, void* (&allocations)[num_allocs],
+    size_t (&allocation_sizes)[num_allocs])
```
yhmtsai, line 45:
```suggestion
    void* workspace_ptr, size_t& num_bytes, void* (&allocations)[num_allocs],
    const size_t (&allocation_sizes)[num_allocs])
```
just ensure it will not be changed accendentally.

core/log/batch_logger.cpp
yhmtsai: okay, I may misunderstand the `reqd_workspace_size` meaning. I thought it is required workspace size. That's why I thought it will be failed when it is not aligned. The actual size is larger than required size == workspace size
However, it is just the size of workspace pointer in `create_workspace_aliases`.
It will be required workspace size (by function) when the aliases phase is failed.

yhmtsai, line 42:
```suggestion
    // it should at least `num * (sizeof(real_type) + sizeof(int))` with some additional buffer for alias purpose, but we simply request large enough size here.
    const size_type reqd_workspace_size = num_batch_items * 32;
```

include/ginkgo/core/solver/batch_solver_base.hpp
yhmtsai: maybe replaced it by log_data->get_workspace_size(num_batch)? It may help understanding for the future. 
At least, some comments here to indicate what this workspace for.

yhmtsai (approved):
LGTM. the name reqd_workspace_size is still confusing to me.
because it has different meaning in function input/output, I do not have idea about the better name

core/base/workspace_aliases.hpp
```diff
@@ -0,0 +1,279 @@
...
+    /**
+     * @brief Returns an array of type @p T and length @p num_elems
+     */
+    template <typename T>
+    GKO_ATTRIBUTES alias<T> create_alias(std::size_t num_elems = 0);
+
+private:
+    GKO_ATTRIBUTES void set_bytes_required(std::size_t new_size)
+    {
+        size_ = max(size_, new_size);
```
MarcelKoch, line 120: this causes an error in our no-circular-deps test, because nothing provides the `max` function. I think it should be fine to use `std::max`. It should even be available in device code, since it's `constexpr` since c++14. If that is somehow not the case, then removing the `GKO_ATTRIBUTES` should be fine, since this functionality is not used in device codes anyway.

PR #1581, opened 2024-03-26
Enable compatability with kokkos 3.7 and 4.x

MarcelKoch:
This PR enables building our kokkos example both with kokkos 3.7 and 4.x.

### DIFF
--- a/examples/kokkos_assembly/CMakeLists.txt
+++ b/examples/kokkos_assembly/CMakeLists.txt
@@ -6,7 +6,7 @@ if(NOT GINKGO_BUILD_EXAMPLES)
     find_package(Ginkgo 1.8.0 REQUIRED)
 endif()
 
-find_package(Kokkos REQUIRED)
+find_package(Kokkos 3.7 REQUIRED)
 
 add_executable(kokkos-assembly kokkos_assembly.cpp)
 target_link_libraries(kokkos-assembly Ginkgo::ginkgo Kokkos::kokkos)
--- a/examples/kokkos_assembly/kokkos_assembly.cpp
+++ b/examples/kokkos_assembly/kokkos_assembly.cpp
@@ -14,6 +14,18 @@
 #include <ginkgo/ginkgo.hpp>
 
 
+// enable compatibility with both kokkos 3.7 and 4.x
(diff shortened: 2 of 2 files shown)


examples/kokkos_assembly/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ if(NOT GINKGO_BUILD_EXAMPLES)
     find_package(Ginkgo 1.8.0 REQUIRED)
 endif()
 
-find_package(Kokkos REQUIRED)
+find_package(Kokkos 3.7 REQUIRED)
```
yhmtsai, line 9: does Kokkos cmake allow 4 is compatiable with 3.7?
Otherwise, I think it usually only allows >= 3.7 but < 4.0

examples/kokkos_assembly/kokkos_assembly.cpp
```diff
@@ -14,6 +14,14 @@
 #include <ginkgo/ginkgo.hpp>
 
 
+// enable compatibility with both kokkos 3.7 and 4.x
+#if KOKKOS_VERSION / 10000 < 4
+namespace Kokkos {
+using Experimental::abs;
+}
```
yhmtsai, line 21:
```suggestion
namespace Kokkos {


using Experimental::abs;


}
```

examples/kokkos_assembly/CMakeLists.txt
MarcelKoch: I'm not sure if I understand what you mean here.

yhmtsai: I thought that `find_package(Kokkos 3.7 REQUIRED)` only allows the kokkos version is >= 3.7 and < 4.0
kokkos version 4 is not allowed here, so the changes in another file are not necessary.

MarcelKoch: No any version >= 3.7 will be accepted.

yhmtsai: It's a bit out of my expectation, but thanks for explanation

Issue #1563, opened 2024-03-08
BiCGStab OMP issue


blegouix:
Hello,

This issue is the continuation of #1561 with more details. We use BiCGStab to solve a quasi-tridiagonal pds problem with different executors (Serial, OMP and CUDA). We observe a numerical issue with OMP. In our test bench build upon those solvers (which involve more than just Ginkgo) we evaluate an error to `1e-16` with Serial or CUDA, but `~1e-10` with OMP **even with `OMP_NUM_THREADS=1`**.

The problem disappears just by replacing the OMP executor with a Serial one.

I link the matrix and the rhs which is problematic. Note that the size of the matrix has an impact (no problem with n=10 or n=100 ie.)

[csr.txt](https://github.com/ginkgo-project/ginkgo/files/14537723/csr.txt)
[rhs.txt](https://github.com/ginkgo-project/ginkgo/files/14537724/rhs.txt)

Please let me know if it is enough for you to reproduce the problem.

I tried with gcc and clang, with Ginkgo 1.7.0 and current develop branch.

Valgrind does not reveals anything and both cases (with serial or omp executor) "converge" in 3 iterations.

Issue from our side: https://github.com/CExA-project/ddc/issues/332
Problematic ginkgo apply call (all ginkgo stuff is in this file): https://github.com/CExA-project/ddc/blob/cc2942283213cc700b3bd8c09fb00346621997e3/include/ddc/kernels/splines/matrix_sparse.hpp#L207

(`create_gko_exec<Kokkos::Serial>()` is a `gko::ReferenceExecutor::create();` while `gko_exec` is a `gko::OmpExecutor::create();`.)

Regards

MarcelKoch:
From the code you linked, I could see that you are using a block Jacobi preconditioner, with block size 32 for OMP and reference. Have you tried using a block size of 1 and see if any differences appear? Nevertheless, it could still hint at some issue in our OMP implementation.

blegouix:
I tried that just now and observed it was indeed fixing the issue.

MarcelKoch:
I've run our benchmarks with the provided matrices. If I used the same residual reduction as in your code `1e-19` I also get a difference in the final residual:
- reference/cuda: 1e-16
- omp: 1e-13

[...]

PR #1584, opened 2024-04-03
Allow binary input for benchmark vectors

upsj:
We currently only support MatrixMarket files as rhs vectors, this can be extended to allow bitwise exact inputs.

Discovered in #1563


pratikvn (approved):

greole (approved):
LGTM.

PR #1589, opened 2024-04-04
Adds distributed row gatherer

MarcelKoch:
This PR adds a distributed row gatherer. This operator essentially provides the communication required in our matrix apply.

Besides the normal apply (which is blocking), it also provides two asynchronous calls. One version has an additional `workspace` parameter which is used as send buffer. This version can be called multiple times without restrictions, if different workspaces are used for each call. The other version doesn't have a workspace parameter, and instead uses an internal buffer. As a consequence, this function can only be called a second time, if the request of the previous call has been waited on. Otherwise, this function will throw.

This is the second part of splitting up #1546.

It also introduces some intermediate changes, which could be extracted out beforehand:
- ~~a type-erased `DenseCache`~~
- ~~making`detail::run` easier to use~~ now part of #1544 

PR Stack:

- [x] #1780 
- [x] #1707 
- [x] #1639 
- [ ] #1589


MarcelKoch:
One issue that I have is the constructor. It takes a `collective_communicator` and an `index_map`. The `index_map` already defines the communication pattern, so the `collective_communicator` has to match that.
One option might be to have a virtual function like
```
std::unique_ptr<collective_communicator> create_with_same_type(communicator, index_map);
```
If I can't come up with anything better, I guess I will use that.

pratikvn:
Do we need to have the `std::future` setup for the release ? Can we remove that for now and just use a normal synchronous approach ? I think that is a significant change that maybe needs more thought and probably a separate PR.

MarcelKoch:
@pratikvn without it, we can't have any non-blocking communication. That would be quite a downgrade from what is currently there.

pratikvn:
You should still be able to return a `mpi::request` to accomplish that ?

MarcelKoch:
With `mpi::request`, it's only possible to have exactly one request at any given time. So the following will lead to a deadlock:
```
auto req1 = row_gatherer->apply_async(...);
auot req2 = row_gatherer->apply_async(...); // <- dead locks
```
The reason is that row gatherer stores the recv buffer internally, so a second call needs to wait until the first one is finished.

MarcelKoch:
But I could use `mpi::request` and then just throw is a second apply is called. But tbh I think an `std::future` is more flexible. The `mpi::request` will not suffice for any more complex system, so we might as well already use `std::future`, since it has very few restrictions.

pratikvn:
Atleast currently, I dont think there are use-cases of one row_gatherer object calling `apply` multiple times ?  Alternatively, we could also pass in a workspace from the caller (matrix class here), which would allow for as many `row_gatherer->apply()` calls, which should also not have the restriction of the serialized `async` apply calls ?

MarcelKoch:
You're right. I will go with the workspace version, since that is similar to what we already have e.g. for the scalar products.

MarcelKoch:
@pratikvn I'm using now mpi requests.

MarcelKoch:
Since this PR has a lot of conflicts with #1403, I will postpone this PR until after the release.

core/distributed/row_gatherer.cpp
```diff
@@ -0,0 +1,299 @@
...
+                        "provide a host buffer or enable MPI support for "
+                        "device memory.");
+
+                    auto b_local = b_global->get_local_vector();
+
+                    dim<2> send_size(coll_comm_->get_send_size(),
+                                     b_local->get_size()[1]);
+                    workspace.set_executor(mpi_exec);
+                    workspace.resize_and_reset(sizeof(ValueType) *
+                                               send_size[0] * send_size[1]);
```
pratikvn, line 140: I think you dont need to `resize_and_reset` every time, but only if `workspace.get_size() < sizeof(ValueType)*send_size[0]*send_size[1]`. `resize_and_reset` will allocate even if the size is sufficient.

core/distributed/row_gatherer.cpp
```diff
@@ -0,0 +1,299 @@
...
+                        make_array_view(
+                            mpi_exec, send_size[0] * send_size[1],
+                            reinterpret_cast<ValueType*>(workspace.get_data())),
+                        send_size[1]);
+                    b_local->row_gather(&send_idxs_, send_buffer);
+
+                    auto recv_ptr = x_local->get_values();
+                    auto send_ptr = send_buffer->get_values();
+
+                    mpi_exec->synchronize();
```
pratikvn, line 152: I think you need to call synchronize for `b_local->get_executor()` ? `mpi_exec` can also be host and the preceding operation `row_gather` is executed on `b_local->get_executor()`.

pratikvn (commented):
Some comments regarding synchronizations.

core/distributed/row_gatherer.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+    array<LocalIndexType> send_gather_idxs(remote_local_idxs.get_executor(),
+                                           send_offsets.back());
+
+    GKO_ASSERT_NO_MPI_ERRORS(MPI_Neighbor_alltoallv(
+        remote_local_idxs.get_const_data(), recv_sizes.data(),
+        recv_offsets.data(), mpi::type_impl<LocalIndexType>::get_type(),
+        send_gather_idxs.get_data(), send_sizes.data(), send_offsets.data(),
+        mpi::type_impl<LocalIndexType>::get_type(), sparse_comm));
+    GKO_ASSERT_NO_MPI_ERRORS(MPI_Comm_free(&sparse_comm));
```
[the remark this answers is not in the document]
MarcelKoch: this is actually not used anymore, because I also realized that. Seems like I forgot to remove it.

core/distributed/row_gatherer.cpp
```diff
@@ -0,0 +1,261 @@
...
+void RowGatherer<LocalIndexType>::apply_impl(const LinOp* b, LinOp* x) const
+{
+    apply_async(b, x).wait();
+}
+
+
+template <typename LocalIndexType>
+void RowGatherer<LocalIndexType>::apply_impl(const LinOp* alpha, const LinOp* b,
+                                             const LinOp* beta, LinOp* x) const
+    GKO_NOT_IMPLEMENTED;
```
pratikvn, line 37: I think you can also implement the advanced apply by replacing `b_local->row_gather(idxs, buffer)` by `b_local->row_gather(alpha, idxs, beta, buffer)`  ?

pratikvn (approved):
Really nice work! LGTM!

core/distributed/matrix.cpp
```diff
@@ -470,17 +360,23 @@ void Matrix<ValueType, LocalIndexType, GlobalIndexType>::apply_impl(
...
+                this->row_gatherer_->apply_async(dense_b, recv_buffer_.get());
             local_mtx_->apply(local_alpha, dense_b->get_local_vector(),
                               local_beta, local_x);
             req.wait();
 
-            auto exec = this->get_executor();
-            auto use_host_buffer = mpi::requires_host_buffer(exec, comm);
-            if (use_host_buffer) {
-                recv_buffer_->copy_from(host_recv_buffer_.get());
-            }
```
pratikvn, line 483: I cant really comment on the specific lines, but the `col_scale` function needs to also be updated with the new interface.

yhmtsai (changes requested):

core/distributed/row_gatherer.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+
+template <typename LocalIndexType>
+mpi::request RowGatherer<LocalIndexType>::apply_async(ptr_param<const LinOp> b,
+                                                      ptr_param<LinOp> x) const
+{
+    int is_inactive;
+    MPI_Status status;
+    GKO_ASSERT_NO_MPI_ERRORS(
+        MPI_Request_get_status(req_listener_, &is_inactive, &status));
```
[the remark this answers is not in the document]
MarcelKoch: That doesn't really work here, since this function would be a member function of `request`, but I'm using a bare `MPI_Request` (and can't use `request`, because it will try to free the request in the destructor), so it would not be applicable.

include/ginkgo/core/distributed/row_gatherer.hpp
```diff
@@ -0,0 +1,196 @@
...
+     */
+    RowGatherer(std::shared_ptr<const Executor> exec, mpi::communicator comm);
+
+    std::shared_ptr<const mpi::CollectiveCommunicator> coll_comm_;
+
+    array<LocalIndexType> send_idxs_;
+
+    mutable array<char> send_workspace_;
+
+    mutable MPI_Request req_listener_{MPI_REQUEST_NULL};
```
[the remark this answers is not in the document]
MarcelKoch: No, because the destructor of `mpi::request` will try to free the request. But `req_listenser_` doesn't own any requests, so the program would crash.

include/ginkgo/core/distributed/row_gatherer.hpp
```diff
@@ -0,0 +1,196 @@
...
+ * ```c++
+ * auto coll_comm = std::make_shared<mpi::neighborhood_communicator>(comm,
+ * imap); auto rg = distributed::RowGatherer<int32>::create(exec, coll_comm,
+ * imap);
+ *
+ * auto b = distributed::Vector<double>::create(...);
+ * auto x = matrix::Dense<double>::create(...);
+ *
+ * auto future = rg->apply_async(b, x);
+ * // do some computation that doesn't modify b, or access x
```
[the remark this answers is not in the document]
MarcelKoch: I guess this just meant to say that you can't expect any meaningful data when accessing `x` before the `wait` has completed.

core/distributed/row_gatherer.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+                    dim<2> send_size(coll_comm_->get_send_size(),
+                                     b_local->get_size()[1]);
+                    auto send_size_in_bytes =
+                        sizeof(ValueType) * send_size[0] * send_size[1];
+                    workspace.set_executor(mpi_exec);
+                    if (send_size_in_bytes > workspace.get_size()) {
+                        workspace.resize_and_reset(sizeof(ValueType) *
+                                                   send_size[0] * send_size[1]);
+                    }
```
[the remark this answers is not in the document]
MarcelKoch: Combine how? Do you mean like
```
workspace = array<char>(mpi_exec, sizeof(ValueType) * send_size[0] * send_size[1]);
```

core/distributed/row_gatherer.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+                    auto recv_ptr = x_local->get_values();
+                    auto send_ptr = send_buffer->get_values();
+
+                    b_local->get_executor()->synchronize();
+                    mpi::contiguous_type type(
+                        b_local->get_size()[1],
+                        mpi::type_impl<ValueType>::get_type());
+                    req = coll_comm_->i_all_to_all_v(
+                        mpi_exec, send_ptr, type.get(), recv_ptr, type.get());
```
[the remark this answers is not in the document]
MarcelKoch: I have a check above to ensure that the memory space of the recv buffer is accessible from the mpi executor. So if GPU aware MPI is used, it should work (even if send buffer is on the host and recv buffer in the device or vice versa). Otherwise an exception will be thrown.

include/ginkgo/core/distributed/row_gatherer.hpp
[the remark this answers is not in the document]
yhmtsai: I think I get it wrong.
Is the comment here to describe that user can do something safely after the call or the apply_async behavior?
My comment was based on that it is the behavior of the apply_async because apply_async definitely accesses x.
If it is for user action during async and wait, then it is correct.

core/distributed/row_gatherer.cpp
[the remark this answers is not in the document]
yhmtsai: yes, if the one of the condition is matched

yhmtsai (changes requested):

core/test/gtest/ginkgo_mpi_main.cpp
```diff
@@ -356,7 +356,13 @@ int main(int argc, char** argv)
 {
     ::testing::InitGoogleTest(&argc, argv);
 
-    MPI_Init(&argc, &argv);
+    int provided_thread_support;
+    MPI_Init_thread(&argc, &argv, MPI_THREAD_MULTIPLE,
+                    &provided_thread_support);
```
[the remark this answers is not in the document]
MarcelKoch: I think this should not be necessary anymore. I tried to have some threading support, but I dropped that after the discussion with @pratikvn

include/ginkgo/core/distributed/row_gatherer.hpp
```diff
@@ -0,0 +1,192 @@
...
+ * ```
+ * Using apply instead of apply_async will lead to a blocking communication.
+ *
+ * @note Objects of this class are only available as shared_ptr, since the class
+ *       is derived from std::enable_shared_from_this.
+ *
+ * @tparam LocalIndexType  the index type for the stored indices
+ */
+template <typename LocalIndexType = int32>
+class RowGatherer final : public EnableLinOp<RowGatherer<LocalIndexType>>,
```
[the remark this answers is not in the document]
MarcelKoch: I think it's a natural extension of the `matrix::RowGatherer` to distributed vectors. I think for now we should keep both separate, but maybe in the future we could combine them to have only one for both the distributed and non-distributed case.

include/ginkgo/core/distributed/row_gatherer.hpp
```diff
@@ -0,0 +1,192 @@
...
+ *
+ * auto b = distributed::Vector<double>::create(...);
+ * auto x = matrix::Dense<double>::create(...);
+ *
+ * auto req = rg->apply_async(b, x);
+ * // do some computation that doesn't modify b, or access x
+ * req.wait();
+ * // x now contains the gathered rows of b
+ * ```
+ * Using apply instead of apply_async will lead to a blocking communication.
```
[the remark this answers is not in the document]
MarcelKoch: Yes, the issue here is that the conversion back would need to happen when the mpi request is finished. I think there is some option to add some hooks to the mpi request, but I've not managed to do that properly yet. So I will just add a warning.

core/distributed/matrix.cpp
```diff
@@ -535,17 +426,23 @@ void Matrix<ValueType, LocalIndexType, GlobalIndexType>::apply_impl(
...
-            auto req = this->communicate(dense_b->get_local_vector());
+            auto recv_dim =
+                dim<2>{static_cast<size_type>(
+                           row_gatherer_->get_collective_communicator()
+                               ->get_recv_size()),
+                       dense_b->get_size()[1]};
+            auto recv_exec = mpi::requires_host_buffer(exec, comm)
+                                 ? exec->get_master()
+                                 : exec;
+            recv_buffer_.init(recv_exec, recv_dim);
```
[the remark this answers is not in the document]
MarcelKoch: I've added the `host_recv_buffer_` again. It should prevent the extra allocation now.

include/ginkgo/core/distributed/row_gatherer.hpp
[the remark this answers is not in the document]
MarcelKoch: I realized, that there is a functional difference between the distributed and non-distributed `RowGatherer`. The distributed one maps between two vectors of the same size, while the non-distributed one map a vector to a (usually) smaller vector. So I can that a different name might be more appropriate. I will try to think of one.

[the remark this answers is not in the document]
MarcelKoch: Sorry, scratch that. They have the same functionality. In the distributed case, the result is also a (usually) smaller vector.

core/distributed/row_gatherer.cpp
```diff
@@ -0,0 +1,266 @@
...
+#include <ginkgo/core/distributed/dense_communicator.hpp>
+#include <ginkgo/core/distributed/neighborhood_communicator.hpp>
+#include <ginkgo/core/matrix/dense.hpp>
+
+#include "core/base/dispatch_helper.hpp"
+
+namespace gko {
+namespace experimental {
+namespace distributed {
+#if GINKGO_HAVE_OPENMPI_PRE_4_1_X
```
yhmtsai, line 18:
```suggestion
namespace distributed {


#if GINKGO_HAVE_OPENMPI_PRE_4_1_X
```

yhmtsai (changes requested):

core/distributed/matrix.cpp
```diff
@@ -563,33 +467,38 @@ void Matrix<ValueType, LocalIndexType, GlobalIndexType>::col_scale(
...
+        make_const_array_view(exec, n_local_cols,
+                              scaling_factors_ptr->get_const_local_values()));
+
+    auto recv_dim = dim<2>{
+        static_cast<size_type>(
+            row_gatherer_->get_collective_communicator()->get_recv_size()),
+        scaling_factors->get_size()[1]};
+    auto recv_exec =
+        mpi::requires_host_buffer(exec, comm) ? exec->get_master() : exec;
+    recv_buffer_.init(recv_exec, recv_dim);
```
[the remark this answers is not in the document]
MarcelKoch: I should now have fixed it so that the buffers will be allocated only once. I always allocate both device and host buffers, even if this is not needed from the MPI implementation. I just think it simplifies the control flow a bit and the memory overhead shouldn't really matter.

MarcelKoch:
@pratikvn, @yhmtsai I've removed the row gatherer from the LinOp hierarchy for now. It instead directly derives from PolymorphicObject. You might want to give it a second look because of that.
The reasoning for me was that the `apply` will always create temporary clones for the vectors, using the executor from the operator. However, in the case of non-GPU aware MPI, this will always lead to throwing an exception, because then the output vector will always be on the device.
I think it should be possible to handle this correctly with generalized MPI request, but this might take quite a while for me to realize, so I would rather proceed as it is.

core/distributed/row_gatherer.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+    GKO_THROW_IF_INVALID(
+        !use_host_buffer || mpi_exec->memory_accessible(x->get_executor()),
+        "The receive buffer uses device memory, but MPI support of device "
+        "memory is not available or host buffer were explicitly requested. "
+        "Please provide a host buffer or enable MPI support for device "
+        "memory.");
+
+    // dispatch global vector
+    run<Vector, double, float, std::complex<double>, std::complex<float>>(
```
pratikvn, line 67: I think a common global list of types should be used here. If it does not exist, I think we should create one in `types.hpp`. If `ValueType` is of `half` or `bfloat16` , then this dispatch will fail ?

core/distributed/row_gatherer.cpp
```diff
@@ -0,0 +1,266 @@
...
+            using ValueType =
+                typename std::decay_t<decltype(*b_global)>::value_type;
+            // dispatch local vector with the same precision as the global
+            // vector
+            ::gko::precision_dispatch<ValueType>(
+                [&](auto* x_local) {
+                    auto b_local = b_global->get_local_vector();
+
+                    dim<2> send_size(coll_comm_->get_send_size(),
+                                     b_local->get_size()[1]);
```
pratikvn, line 78: `coll_comm_` and the `send_size` are already known during the time when the row gather object is created. I think it should be possible to move the workspace alloc (in case of non-gpu aware memory) to the creation time, rather than in the apply function ?

pratikvn (changes requested):
I like that you added tests to check the number of allocs as well! I have a couple of suggestions, but otherwise looks good!

core/distributed/row_gatherer.cpp
MarcelKoch: The number of RHS is not known at creation. So I will just leave it as it is.

core/distributed/row_gatherer.cpp
MarcelKoch: I agree, but I think this should happen in a separate PR. Otherwise, too many unrelated changes will appear.

core/distributed/row_gatherer.cpp
pratikvn: Most of our applications just use a single rhs. I think without much effort, it should be possible to allocate in the creation and then only re-allocate if the number of rhs > 1.

pratikvn (approved):
lgtm

core/distributed/row_gatherer.cpp
yhmtsai: unresolved

include/ginkgo/core/distributed/row_gatherer.hpp
```diff
@@ -0,0 +1,201 @@
...
+    RowGatherer(RowGatherer&& o) noexcept;
+
+    RowGatherer& operator=(const RowGatherer& o);
+
+    RowGatherer& operator=(RowGatherer&& o);
+
+private:
+    /**
+     * @copydoc RowGatherer::create(std::shared_ptr<const
+     *          Executor>, std::shared_ptr<const mpi::collective_communicator>,
```
yhmtsai, line 173:
```suggestion
     *          Executor>, std::shared_ptr<const mpi::CollectiveCommunicator>,
```

include/ginkgo/core/distributed/row_gatherer.hpp
```diff
@@ -0,0 +1,201 @@
...
+    dim<2> size_;
+    std::shared_ptr<const mpi::CollectiveCommunicator> coll_comm_;
+    array<LocalIndexType> send_idxs_;
+    mutable array<char> send_workspace_;
+    // This object might not hold an actual MPI request, so we can't use the
+    // always owning mpi::request. It's destructor would otherwise make the
+    // program crash.
+    mutable MPI_Request req_listener_{MPI_REQUEST_NULL};
+};
+}  // namespace distributed
```
yhmtsai, line 196:
```suggestion
};


}  // namespace distributed
```

test/mpi/distributed/row_gatherer.cpp
```diff
@@ -0,0 +1,222 @@
...
+TYPED_TEST(RowGatherer, ThrowsOnNonMatchingExecutor)
+{
+    if (this->mpi_exec == this->exec) {
+        GTEST_SKIP();
+    }
+
+    using RowGatherer = typename TestFixture::row_gatherer_type;
+    using Dense = gko::matrix::Dense<double>;
+    using Vector = gko::experimental::distributed::Vector<double>;
+    auto rg = RowGatherer::create(this->exec, this->comm);
```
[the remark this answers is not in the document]
yhmtsai: b seems to be fine on both due to the temporary_clone

benchmark/test/reference/spmv_distributed.profile.stderr
```diff
@@ -116,6 +116,18 @@ DEBUG: begin components::convert_idxs_to_ptrs
...
+DEBUG: begin components::fill_array
+DEBUG: end   components::fill_array
+DEBUG: begin copy
+DEBUG: end   copy
+DEBUG: begin copy
+DEBUG: end   copy
+DEBUG: begin copy
+DEBUG: end   copy
+DEBUG: begin copy
+DEBUG: end   copy
```
yhmtsai, line 130: I only found the one array copy from row_gather when copy the matrix.
Do I miss anything?

yhmtsai (changes requested):
mainly for the test executor. I will not hold this due to the name but it might be confusing.

test/mpi/distributed/row_gatherer.cpp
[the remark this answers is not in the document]
MarcelKoch: It should stay `exec`. Yes, `b` is fine, but the exception will be caused by `x`, since that is not on the host.

benchmark/test/reference/spmv_distributed.profile.stderr
MarcelKoch: The constructor has 
- 2 `set_executor`
- 1 `make_temporary_clone(segmented_array)`

One `set_executor` might not trigger the copy section, the other operations, however, should trigger the copy section. That gives 3 sections already, since the segmented array consists of two arrays internally. These two might also cover the fill_array section.

test/mpi/distributed/row_gatherer.cpp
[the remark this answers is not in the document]
yhmtsai: I see. mpi_exec is the executor respect to the mpi ability not the main one.

benchmark/test/reference/spmv_distributed.profile.stderr
yhmtsai: The reference test should not need any copy in this three function, right?

include/ginkgo/core/distributed/row_gatherer.hpp
```diff
@@ -0,0 +1,201 @@
...
+ *
+ * @note The output vector for the apply_async functions *must* use an executor
+ *       that is compatible with the MPI implementation. In particular, if the
+ *       MPI implementation is not GPU aware, then the output vector *must* use
+ *       a CPU executor. Otherwise, an exception will be thrown.
+ *
+ * @tparam LocalIndexType  the index type for the stored indices
+ */
+template <typename LocalIndexType = int32>
+class RowGatherer final
```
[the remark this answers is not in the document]
MarcelKoch: IMO `RowDistributer` would be more confusing. This type does the same thing as `RowGatherer`, but in a distributed context. The different name of the apply function doesn't really matter, since the functionality is still the same. (Also I expect that the `apply` function will be added some time in the future.)

benchmark/test/reference/spmv_distributed.profile.stderr
MarcelKoch: I found the issue. The `CollectiveCommunicator::create_with_same_type` takes in a `const std::variant<imap<...>...>&`, but this does *not* mean that the `imap` is also passed as a `const &`. Instead this triggers the copy constructor of the `imap`, which leads to the additional calls.
See this example: https://godbolt.org/z/r1WYKv37q

I will try to see if I can prevent this somehow.

MarcelKoch: I changed the `create_with_same_type` a bit now so that it doesn't create temporary index maps. So the overhead should be gone.

include/ginkgo/core/distributed/row_gatherer.hpp
```diff
@@ -129,24 +128,22 @@ class RowGatherer final
      * @tparam GlobalIndexType  the global index type of the index map
      *
      * @param exec  the executor
-     * @param coll_comm  the collective communicator
+     * @param coll_comm_creator  the collective communicator creator_fn
```
yhmtsai, line 131: could you also add the note here about revisiting this function argument and the reason?
I feel a bit weird when passing the function to create a object, but I do not have an idea to resolve the variant copy issue.

yhmtsai (approved):
LGTM. Adding some note about the revisiting and reason about the creator_fn will be good

yhmtsai:
The rowgatherer create function in the test does not adapt the changes yet

include/ginkgo/core/distributed/row_gatherer.hpp
MarcelKoch: I've reverted this change. There are still no temporary copies of the index map, and the collective communicator object is used again, instead of the creator function.

PR #1594, opened 2024-04-08
Simplify the testing types

thoasm:
Add template-type functions to combine and merge type lists. As a result, the GINKGO_DPCPP_SINGLE_MODE only needs to be present once.

Additionally, change the typen name ValueAndIndexType to ComplexAndPODTypes (because gko::size_type is not an IndexType).

TODO:
- [x] Add documentation to the type helper
- [x] go through all existing tests and use the wrapper in all cases we define types on our own (example: CB-Gmres)

### DIFF
--- a/core/test/utils.hpp
+++ b/core/test/utils.hpp
@@ -173,13 +173,13 @@ using ValueTypes = merge_type_list_t<RealValueTypes, ComplexValueTypes>;
 
 using IndexTypes = ::testing::Types<int32, int64>;
 
+using IntegerTypes = merge_type_list_t<IndexTypes, ::testing::Types<size_type>>;
+
 using LocalGlobalIndexTypes =
     ::testing::Types<std::tuple<int32, int32>, std::tuple<int32, int64>,
                      std::tuple<int64, int64>>;
 
-using PODTypes =
-    merge_type_list_t<merge_type_list_t<RealValueTypes, IndexTypes>,
-                      ::testing::Types<size_type>>;
+using PODTypes = merge_type_list_t<RealValueTypes, IntegerTypes>;
 
 using ComplexAndPODTypes = merge_type_list_t<ComplexValueTypes, PODTypes>;
 
--- a/core/test/utils.hpp
+++ b/core/test/utils.hpp
@@ -29,21 +29,136 @@
 
 namespace gko {
 namespace test {
+namespace detail {
 
 
-using ValueTypes =
-#if GINKGO_DPCPP_SINGLE_MODE
-    ::testing::Types<float, std::complex<float>>;
-#else
-    ::testing::Types<float, double, std::complex<float>, std::complex<double>>;
-#endif
+template <typename LeftList, typename RightList, typename... Result>
+struct cartesian_type_product {};
 
-using ComplexValueTypes =
-#if GINKGO_DPCPP_SINGLE_MODE
-    ::testing::Types<std::complex<float>>;
-#else
-    ::testing::Types<std::complex<float>, std::complex<double>>;
-#endif
+template <template <typename...> class OuterWrapper, typename FirstLeft,
+          typename... RemainingLeftArgs, typename... RightArgs,
+          typename... Result>
+struct cartesian_type_product<OuterWrapper<FirstLeft, RemainingLeftArgs...>,
+                              OuterWrapper<RightArgs...>, Result...>
+    : cartesian_type_product<OuterWrapper<RemainingLeftArgs...>,
+                             OuterWrapper<RightArgs...>, Result...,
+                             std::tuple<FirstLeft, RightArgs>...> {};
+
+template <template <typename...> class OuterWrapper, typename... RightArgs,
+          typename... Result>
+struct cartesian_type_product<OuterWrapper<>, OuterWrapper<RightArgs...>,
+                              Result...> {
+    using type = OuterWrapper<Result...>;
+};
+
+template <typename ExistingCombinationList, typename NewElementList,
+          typename... Result>
+struct add_to_cartesian_type_product {};
+
+template <template <typename...> class OuterWrapper,
+          typename... CurrentCombinationArgs,
+          typename... RemainingOldCombinations, typename... NewElementArgs,
+          typename... Result>
+struct add_to_cartesian_type_product<
+    OuterWrapper<std::tuple<CurrentCombinationArgs...>,
+                 RemainingOldCombinations...>,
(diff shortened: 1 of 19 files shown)


core/test/utils.hpp
```diff
@@ -32,154 +32,388 @@
...
-                     gko::size_type>;
-#else
-    ::testing::Types<float, double, std::complex<float>, std::complex<double>,
-                     gko::int32, gko::int64, gko::size_type>;
-#endif
+}  // namespace detail
+
+
+/**
+ * This type alias creates a cartesian product of the types in the left list
```
MarcelKoch, line 240: there is a lot of duplicated documentation. Either here or in the `detail` namespace the documentation can be replaced by either `@copydoc` or just referencing the other doc

core/test/utils.hpp
```diff
@@ -32,154 +32,388 @@
...
+ *     add_internal_wrapper<std::complex, std::tuple<float, double>>;
+ * // new_type = std::tuple<std::complex<float>, std::complex<double>>;
+ * ```
+ *
+ * @tparam NewInnerWrapper  the new wrapper you want to use to wrap each type
+ *                          in the list
+ * @tparam ListType  The list of types where you want to add a wrapper to each
+ */
+template <template <typename...> class NewInnerWrapper, typename ListType>
+using add_internal_wrapper_t =
```
MarcelKoch, line 374: nit: maybe use `inner` instead of `internal` (same for the detail namespace)

core/test/utils/utils_test.cpp
```diff
@@ -0,0 +1,236 @@
...
+{
+    testing::StaticAssertTypeEq<
+        gko::test::change_outer_wrapper_t<std::tuple, testing_types1>,
+        tuple_types1>();
+    testing::StaticAssertTypeEq<
+        gko::test::change_outer_wrapper_t<std::tuple, testing_types2>,
+        tuple_types2>();
+    testing::StaticAssertTypeEq<
+        gko::test::change_outer_wrapper_t<std::tuple, testing_types3>,
+        tuple_types3>();
```
MarcelKoch, line 44: Does this test give any additional information than the two before? (same for the one below)

core/test/utils/utils_test.cpp
```diff
@@ -0,0 +1,236 @@
...
+    testing::StaticAssertTypeEq<
+        gko::test::change_outer_wrapper_t<testing::Types, tuple_types2>,
+        testing_types2>();
+    testing::StaticAssertTypeEq<
+        gko::test::change_outer_wrapper_t<testing::Types, tuple_types3>,
+        testing_types3>();
+}
+
+
+TEST(TypeListHelper, ChangeOuterWrapper2)
```
MarcelKoch, line 57: Maybe more descriptive test name (also for the one above)?

core/test/utils/utils_test.cpp
```diff
@@ -0,0 +1,236 @@
...
+    testing::StaticAssertTypeEq<
+        gko::test::cartesian_type_product_t<testing_types2, testing_types3>,
+        expected_c1>();
+    testing::StaticAssertTypeEq<
+        gko::test::cartesian_type_product_t<tuple_types2, tuple_types3>,
+        gko::test::change_outer_wrapper_t<std::tuple, expected_c1>>();
+}
+
+
+TEST(TypeListHelper, CartesianTypeProductSmall)
```
MarcelKoch, line 142: test for cartesian product with one (or both) empty list?

MarcelKoch (changes requested):
Generally lgtm. The biggest issue for me is the documentation. It's dupliated a lot and should only be in one place (I think I would prefer the aliases in the `gko::test` namespace).

core/test/utils/utils_test.cpp
MarcelKoch: But what is the difference to the two tests before? The first test uses a list of length 1 and the second uses a list of length two. I think a list with one element and a list with two elements can be considered to be quite different, but I don't think there is any more difference between a list of two elements and one with three elements.

thoasm: That is true. The handling of two or three elements should be the same. I mainly added it for 2 purposes:
1. Make sure `testing_typesX` has the same contents as `tuple_typesX`. This should prevent the test types from being modified in the wrong way.
2. Make sure the parameter packs work as intended and are not just special cases for 1-2 parameters.
The 2nd point could also be checked by looking at the implementation, but I thought an additional test wouldn't hurt.

If you think I should remove the 3-element tests, I can do so.

core/test/utils.hpp
thoasm: I used the `@see` doxygen feature (but since we don't produce the documentation, I'm not 100% sure it works). I will try to compile it for myself to see if it works properly.
I also added details on how the implementation works in the `detail` namespace (mostly in `cartesian_type_product`).

thoasm: It is just shown as plain text (and not as a link). But the rest is displayed correctly.

core/test/utils/utils_test.cpp
MarcelKoch: Well, I don't care that much about it.

MarcelKoch (approved):

yhmtsai (approved):
LGTM. 
Because it is only used for testing, it is fine for generating `std::tuple<>` in cartesian product.
Just want to mention that I have also made some similar stuff for different purpose in first draft of file config https://github.com/ginkgo-project/ginkgo/blob/resource_manager/extension/resource_manager/include/resource_manager/base/type_pack.hpp with restricted and explicit outer_type and inner_type, so I do not need to distinguish after 2D caresian.

I think you can merge before half precision.
It might give some annoying rebase step for me, but it makes the half pr more clear.

core/test/utils.hpp
```diff
@@ -29,154 +29,331 @@
...
+template <typename ExistingCombinationList, typename NewElementList,
+          typename... Result>
+struct add_to_cartesian_type_product {};
+
+template <template <typename...> class OuterWrapper,
+          typename... CurrentCombinationArgs,
+          typename... RemainingOldCombinations, typename... NewElementArgs,
+          typename... Result>
+struct add_to_cartesian_type_product<
+    OuterWrapper<std::tuple<CurrentCombinationArgs...>,
```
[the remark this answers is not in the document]
thoasm: Yes, I always expect the inner type to be a `std::tuple` (otherwise, the code is even more complex).

PR #1598, opened 2024-04-17
Add a batch CG solver and reference/omp kernels

pratikvn:
This PR adds a batch::solver::Cg class, which similar to the batch::solver::Bicgstab enables solution for SPD matrices. The functionality is equivalent of the non-batched Cg solver.

### Some comments

1. The .cu/.hip.cpp/.dp.cpp files can be safely ignored. All paths leading to those should result in GKO_NOT_IMPLEMENTED. These device kernels will be added in another PR.
2. The class and the framework is very similar to the bicgstab class, so hopefully it should be fairly easy to review.


reference/solver/batch_cg_kernels.cpp
```diff
@@ -0,0 +1,108 @@
...
+        if (num_rhs > max_num_rhs) {
+            GKO_NOT_IMPLEMENTED;
+        }
+
+        const size_type local_size_bytes =
+            gko::kernels::batch_cg::local_memory_requirement<ValueType>(
+                num_rows, num_rhs) +
+            PrecType::dynamic_work_size(num_rows,
+                                        mat.get_single_item_num_nnz()) *
+                sizeof(ValueType);
```
MarcelKoch, line 72: why does this return the size in elements, and the other one the size in bytes? IMO both should return the same.

omp/solver/batch_cg_kernels.cpp
```diff
@@ -0,0 +1,110 @@
...
+            PrecondType::dynamic_work_size(num_rows,
+                                           mat.get_single_item_num_nnz()) *
+                sizeof(ValueType);
+
+#pragma omp parallel for
+        for (size_type batch_id = 0; batch_id < num_batch_items; batch_id++) {
+            // TODO: Align to cache line boundary
+            // TODO: Allocate and free once per thread rather than once per
+            // work-item.
+            auto local_space = array<unsigned char>(exec_, local_size_bytes);
```
MarcelKoch, line 77: That is quite trivial to do. Just put it outside the loop and add `firstprivate(local_space)` to the pragma

reference/test/solver/batch_cg_kernels.cpp
```diff
@@ -0,0 +1,315 @@
...
+{
+    using value_type = typename TestFixture::value_type;
+    using real_type = gko::remove_complex<value_type>;
+
+    auto res = gko::test::solve_linear_system(this->exec, this->solve_lambda,
+                                              this->solver_settings,
+                                              this->linear_system);
+
+    const int ref_iters = 2;
+    auto iter_array = res.log_data->iter_counts.get_const_data();
```
MarcelKoch, line 104: these are unused, is a test missing?

reference/test/solver/batch_cg_kernels.cpp
```diff
@@ -0,0 +1,315 @@
...
+    GKO_ASSERT_BATCH_MTX_NEAR(res.x, linear_system.exact_sol, tol);
+    for (size_t i = 0; i < num_batch_items; i++) {
+        ASSERT_LE(res.host_res_norm->get_const_values()[i] /
+                      linear_system.host_rhs_norm->get_const_values()[i],
+                  tol);
+    }
+}
+
+
+TYPED_TEST(BatchCg, ApplyLogsResAndIters)
```
MarcelKoch, line 171: nit: I would switch the order of this test and the one before.

test/solver/batch_cg_kernels.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+
+#include "core/base/batch_utilities.hpp"
+#include "core/matrix/batch_dense_kernels.hpp"
+#include "core/test/utils.hpp"
+#include "core/test/utils/batch_helpers.hpp"
+#include "test/utils/executor.hpp"
+
+
+class BatchCg : public CommonTestFixture {
```
MarcelKoch, line 29: Since these tests don't compare agains the ref exec, I think it would be fine to merge the tests from `reference/test/...` into this file. I guess that would save a couple of lines.

core/test/solver/batch_cg.cpp
```diff
@@ -0,0 +1,277 @@
...
+    auto copy = this->solver_factory->generate(Mtx::create(this->exec));
+
+    copy->copy_from(this->solver.get());
+
+    ASSERT_EQ(copy->get_common_size(),
+              gko::dim<2>(this->num_rows, this->num_rows));
+    ASSERT_EQ(copy->get_num_batch_items(), this->num_batch_items);
+    auto copy_mtx = gko::as<Solver>(copy.get())->get_system_matrix();
+    const auto copy_batch_mtx = gko::as<const Mtx>(copy_mtx.get());
+    GKO_ASSERT_BATCH_MTX_NEAR(this->mtx.get(), copy_batch_mtx, 0.0);
```
MarcelKoch, line 112: I guess one could also test that the pointers are not the same.

core/test/solver/batch_cg.cpp
```diff
@@ -0,0 +1,277 @@
...
+    auto copy = this->solver_factory->generate(Mtx::create(this->exec));
+
+    copy->move_from(this->solver);
+
+    ASSERT_EQ(copy->get_common_size(),
+              gko::dim<2>(this->num_rows, this->num_rows));
+    ASSERT_EQ(copy->get_num_batch_items(), this->num_batch_items);
+    auto copy_mtx = gko::as<Solver>(copy.get())->get_system_matrix();
+    const auto copy_batch_mtx = gko::as<const Mtx>(copy_mtx.get());
+    GKO_ASSERT_BATCH_MTX_NEAR(this->mtx.get(), copy_batch_mtx, 0.0);
```
MarcelKoch, line 129: the moved from state is not checked?

core/test/solver/batch_cg.cpp
```diff
@@ -0,0 +1,277 @@
...
+    using real_type = typename TestFixture::real_type;
+    auto solver_factory =
+        Solver::build()
+            .with_max_iterations(22)
+            .with_tolerance(static_cast<real_type>(0.25))
+            .with_tolerance_type(gko::batch::stop::tolerance_type::relative)
+            .on(this->exec);
+    auto solver = solver_factory->generate(this->mtx);
+
+    solver->reset_tolerance(0.5);
```
MarcelKoch, line 193: If we already allow this (which I don't support), then it also needs to set the tolerance of the parameter. Having two different tolerances is nonsense to me.
(Same for the other reset functions)

core/test/utils/batch_helpers.hpp
```diff
@@ -106,11 +106,11 @@ std::unique_ptr<MatrixType> generate_3pt_stencil_batch_matrix(
         {}};
     for (int row = 0; row < num_rows; ++row) {
         if (row > 0) {
-            data.nonzeros.emplace_back(row - 1, row, value_type{-1.0});
+            data.nonzeros.emplace_back(row - 1, row, value_type{-0.5});
```
MarcelKoch, line 109: did the previous value lead to any issues?

MarcelKoch (changes requested):
only some minior suggestions. The largest is probably the local space for the omp implementation.

omp/solver/batch_cg_kernels.cpp
pratikvn: Yes, that is true, but I believe one of the OMP versions in our container does not support it and ends up giving wrong results.

MarcelKoch: I guess we have a non-conforming OMP implementation? We should probably fix that.
Anyway, a `std::vector` of arrays would also work.

core/test/solver/batch_cg.cpp
pratikvn: The pointers will be the same as the only a shallow copy is done. It is the same case in the non-batched as well.

reference/test/solver/batch_cg_kernels.cpp
pratikvn: iters have another test.

core/test/utils/batch_helpers.hpp
pratikvn: Just wanted to test for a more positive matrix. Cg sometimes seems to have issues.

core/test/solver/batch_cg.cpp
MarcelKoch: I guess that's true. Maybe something to think about in the future.

core/test/solver/batch_cg.cpp
pratikvn: I think we decided to make the `parameters_` struct not public facing anymore, so in that case it should be irrelevant. Currently, we cannot set the values of the `parameters_` struct from the `reset_` functions because the `reset_` functions are part of the base class, where there is no concept of the parameters_ struct. In any case, the parameters_ struct is not used in the solver core anymore, with the updated SolverBase structure and only used to pass the parameters to the base solver members.

MarcelKoch: sounds reasonable.

omp/solver/batch_cg_kernels.cpp
pratikvn: Yes, the nvhpc compiler OMP implementations are annoying to deal with, and I believe follow an older OMP standard ?

MarcelKoch: from https://docs.nvidia.com/hpc-sdk/archive/22.7/compilers/hpc-compilers-user-guide/index.html#openmp-use they say that they support the `firstprivate` key word. Also, the failed test with nvhpc isn't about omp.

reference/test/solver/batch_cg_kernels.cpp
```diff
@@ -0,0 +1,201 @@
...
+                                              this->solver_settings,
+                                              this->linear_system);
+
+    auto res_log_array = res.log_data->res_norms.get_const_data();
+    for (size_t i = 0; i < this->num_batch_items; i++) {
+        ASSERT_LE(
+            res_log_array[i] / this->linear_system.host_rhs_norm->at(i, 0, 0),
+            this->solver_settings.residual_tol);
+        ASSERT_NEAR(res_log_array[i], res.host_res_norm->get_const_values()[i],
+                    10 * this->eps);
```
yhmtsai, line 109: does the residual still use eps not the r<T>?

reference/test/solver/batch_cg_kernels.cpp
```diff
@@ -0,0 +1,201 @@
...
+        gko::share(gko::test::generate_diag_dominant_batch_matrix<const Mtx>(
+            this->exec, num_batch_items, num_rows, true));
+    auto linear_system =
+        gko::test::generate_batch_linear_system(diag_dom_mat, num_rhs);
+    auto solver = gko::share(solver_factory->generate(linear_system.matrix));
+
+    auto res =
+        gko::test::solve_linear_system(this->exec, linear_system, solver);
+
+    GKO_ASSERT_BATCH_MTX_NEAR(res.x, linear_system.exact_sol, tol * 500);
```
yhmtsai, line 200: also check the residual

test/solver/batch_cg_kernels.cpp
```diff
@@ -0,0 +1,240 @@
...
+}
+
+
+TEST_F(BatchCg, CanSolveLargeMatrixSizeHpdSystem)
+{
+    const int num_batch_items = 12;
+    const int num_rows = 1025;
+    const int num_rhs = 1;
+    const real_type tol = 1e-5;
+    const int max_iters = num_rows * 2;
```
yhmtsai, line 213: num_rows should be the enough in theory, right?

core/test/solver/batch_cg.cpp
```diff
@@ -0,0 +1,279 @@
...
+}
+
+
+TYPED_TEST(BatchCg, CanBeMoved)
+{
+    using Mtx = typename TestFixture::Mtx;
+    using Solver = typename TestFixture::Solver;
+    auto copy = this->solver_factory->generate(Mtx::create(this->exec));
+
+    copy->move_from(this->solver);
```
yhmtsai, line 122: nit: not copy

core/solver/batch_cg_kernels.hpp
```diff
@@ -0,0 +1,188 @@
...
+    // Compute the number of vecs that can be stored in shared memory and assign
+    // the rest to global memory.
+    const int initial_vecs_available = rem_shared / vec_size;
+    const int num_vecs_shared = min(initial_vecs_available, num_main_vecs);
+    sconf.n_shared += num_vecs_shared;
+    sconf.n_global -= num_vecs_shared;
+    rem_shared -= num_vecs_shared * vec_size;
+    // Set the storage configuration with preconditioner workspace in global if
+    // there are any vectors in global memory.
+    if (sconf.n_global > 0) {
```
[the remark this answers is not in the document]
yhmtsai: Maybe not fully same? rem_shared might still have more than 0 bytes

yhmtsai (commented):
LGTM in general.
Is there any reason to set only 1e-5 tol not something like 1e-12 for double precision?

reference/test/solver/batch_cg_kernels.cpp
pratikvn: I want to set it to a lower tol so check against that tolerance and not r

test/solver/batch_cg_kernels.cpp
pratikvn: Yes, but not in practice. Especially without a preconditioner and for matrices with eigenvalues very close to each other.

core/solver/batch_cg_kernels.hpp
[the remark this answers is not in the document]
MarcelKoch: I think you are right.

reference/solver/batch_cg_kernels.cpp
```diff
@@ -0,0 +1,108 @@
...
+        const auto num_rows = mat.num_rows;
+        const auto num_rhs = b.num_rhs;
+        if (num_rhs > max_num_rhs) {
+            GKO_NOT_IMPLEMENTED;
+        }
+
+        const size_type local_size_bytes =
+            gko::kernels::batch_cg::local_memory_requirement<ValueType>(
+                num_rows, num_rhs) +
+            // TODO: return bytes for dynamic work size as well
```
MarcelKoch, line 70: isn't this comment done?

MarcelKoch (approved):

reference/test/solver/batch_cg_kernels.cpp
yhmtsai: I am not sure checking lower tolerance on the solution is a good idea because it will depends on the matrix condition number and it is not the criterion can help here.

yhmtsai (approved):
LGTM. I am not sure checking lower tolerance on the solution is a good idea because it will depends on the matrix condition number and the criterion can not really ensure it will be achieved.

reference/test/solver/batch_cg_kernels.cpp
pratikvn: I this test I mainly want to check that the logger logs the residual, so I think it should be fine here.

PR #1599, opened 2024-04-17
Adds workflow to update the status of a PR

MarcelKoch:
This PR adds a GH workflow to automatically update the review status of a PR within a GH project. It is assumed that a PR is only part of either 0 or 1 projects. 

The status will be set from the following criteria:
- `>=1` changes-requested -> `in-progress`
- `>=2` approved -> `ready-to-merge`
- `>=1` reviews pending -> `in-review`

Compared to our current GH actions, this uses the graphQL api instead of the rest API. The new projects are not available with rest API so the graphQL API has to be used.

You can see an example of that in a PR in my repo https://github.com/MarcelKoch/gh-projects-playground/pull/1 (maybe ignore the comments in there...) and the project https://github.com/users/MarcelKoch/projects/5. I think you can just add a review there, or add your self as reviewers, and then the project will be automatically updated. 
Other than that, I'm not sure if this can be tested, due to our CI setup.

Todo:

- more robust discovery of project id


MarcelKoch:
Right now this will not work, because the bot token doesn't have the right permissions: https://github.com/ginkgo-project/ginkgo/actions/runs/8724752050/job/23936060577?pr=1599

tcojean:
> Right now this will not work, because the bot token doesn't have the right permissions: https://github.com/ginkgo-project/ginkgo/actions/runs/8724752050/job/23936060577?pr=1599

Works now.

MarcelKoch:
yes thanks @tcojean. It already does work within this PR. Any is free to try it out, i.e. add reviewers, remove reviewers, approve, request changes, etc.

MarcelKoch:
I will close this for now, but may revisit it again for future releases.

PR #1601, opened 2024-04-29
Simple segmented ranges

upsj:
This removes all stride support from #1582, which removes most of the heavy `iterator` functionality added in the other PR.
It relies on `zip_iterator` to provide both enumerating
`for (auto [nonzero_index, column, value] : matrix.enumerated()[row])` and non-enumerating
`for (auto [column, value] : matrix[row])` iteration through a matrix, and uses pairs of pointers to represent ranges of column indices or values.
The two commits add the baseline first and C++17 structured binding support second.

TODO

- [x] Merge #1603 
- [x] Merge #1604


MarcelKoch:
First up, the zip iterator also need to be made constexpr for this to work on the device.

core/base/segmented_range.hpp
```diff
@@ -0,0 +1,361 @@
...
+
+    /**
+     * Constructs a segmented range from combined begin and end pointers.
+     * The `i`th range is given by `[ptrs[i], ptrs[i + 1])`.
+     *
+     * @param ptrs  a pointer to the array of beginning and end indices
+     * @param num_segments  the number of segments, i.e. the size of the
+     *                      ptrs index arrays.
+     */
+    constexpr explicit segmented_range(const index_type* ptrs,
```
MarcelKoch, line 103: this constructor is not tested.

core/base/segmented_range.hpp
```diff
@@ -0,0 +1,361 @@
...
+ */
+template <typename IndexType, typename ValueIterator>
+class segmented_value_range {
+public:
+    using index_type = IndexType;
+    using index_iterator = index_iterator<index_type>;
+    using value_iterator = ValueIterator;
+    using segment = iterator_range<ValueIterator>;
+    using enumerated_range = segmented_value_range<
+        index_type, detail::zip_iterator<index_iterator, value_iterator>>;
```
MarcelKoch, line 190: this class will not work on the device, because `zip_iterator` doesn't work on the device. Maybe different zip iterator types will be used depending on the backend?

core/base/segmented_range.hpp
```diff
@@ -0,0 +1,361 @@
...
+     * Constructs a segmented range from combined begin and end pointers.
+     * The `i`th range is given by `[ptrs[i], ptrs[i + 1])`.
+     *
+     * @param ptrs  a pointer to the array of beginning and end indices
+     * @param values  an iterator pointing to the values into which the
+     *                beginning/end indices point.
+     * @param num_segments  the number of segments, i.e. the size of the
+     *                      ptrs index arrays.
+     */
+    constexpr explicit segmented_value_range(const index_type* ptrs,
```
MarcelKoch, line 272: also not tested

core/test/base/segmented_range.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+TEST(SegmentedRange, Works)
+{
+    std::vector<int> begins{3, 1, 4, 9};
+    std::vector<int> ends{3, 10, 6, 10};
+    std::vector<std::vector<int>> result_indices(begins.size());
+    gko::segmented_range<int> range{begins.data(), ends.data(),
+                                    static_cast<int>(begins.size())};
+
+    for (auto row : gko::irange<int>(begins.size())) {
```
MarcelKoch, line 24: nit:
```suggestion
    for (auto row : gko::irange<int>(range.num_segments())) {
```
since I think this shows better the usage of the range.
Same for the rest of the file.

core/test/base/segmented_range.cpp
MarcelKoch, line 1: maybe also add assertion tests, similar to the other PR?

core/test/base/segmented_range.cpp
```diff
@@ -0,0 +1,136 @@
...
+        }
+    }
+
+    ASSERT_EQ(result_values,
+              std::vector<std::vector<int>>(
+                  {{}, {2, 3, 4, 5, 6, 7, 8, 9, 10}, {5, 6}, {10}}));
+}
+
+
+TEST(SegmentedValueRange, WorksByRangeFor)
```
MarcelKoch, line 59: I think it's possible to only conditionally enable this test when c++17 is present with the `__cplusplus` macro.
Same for the other test with structured bindings.

MarcelKoch (changes requested):
this looks a lot cleaner than the old PR, thanks for that. I have mostly smaller remarks, the biggest issue is the `zip_iterator` which is not available on the device. Maybe it would be possible to use different types for the zip_iterator depending on the backend.

core/base/segmented_range.hpp
```diff
@@ -0,0 +1,372 @@
...
+ * Represents a range of indices that is segmented into contiguous segments.
+ * Each segment has the shape `[begin, end)`, i.e. it is a half-open interval.
+ *
+ * @tparam IndexType  the type of indices used to represent the segments.
+ */
+template <typename IndexType>
+class segmented_range {
+public:
+    using index_type = IndexType;
+    using index_iterator_type = index_iterator<index_type>;
```
MarcelKoch, line 29: not really related to this pr, but maybe a better name for `index_iterator` would be `interval_iterator`.

core/base/segmented_range.hpp
```diff
@@ -0,0 +1,372 @@
...
+ * Represents a range of indices that is segmented into contiguous segments,
+ * mapped into a value array. Each segment has the shape `[begin, end)`, i.e. it
+ * is a half-open interval and points to corresponding entries of the value
+ * array.
+ *
+ * @tparam IndexType  the type of indices used to represent the segments.
+ * @tparam ValueIterator  the iterator type pointing to the values.
+ */
+template <typename IndexType, typename ValueIterator>
+class segmented_value_range {
```
MarcelKoch, line 188: I'm wondering a bit how this will work out with the `device_segmented_array` and device implementations of our classes in general. 
In principle, `device_segmented_array` could be replaced directly by this. One thing I would not like about that is that the `device_segmented_array` basically consists only of data members, and in general I would prefer our device implementations to not have member functions. (My reason for that is that I want to generalize the device implementations so that they can be directly exposed to kokkos or similar.)
So right now, I would imagine conversion functions and having some redundancy, instead of replacing `device_segmented_array`.

MarcelKoch (approved):

core/base/segmented_range.hpp
pratikvn: I think that is a heavy restriction. Is there a reason why having no member functions is necessary to expose these directly to kokkos ?

pratikvn (approved):
LGTM!

core/base/segmented_range.hpp
MarcelKoch: It's not necessary. But I would prefer to not provide methods, as I'm not sure if they would also be available in Kokkos code. I guess they would be, since this can run on the GPU, but I would still be a bit hesitant.

core/base/segmented_range.hpp
upsj: I think we should have clarity on what the iterator typename represents - is it an iterator _returning_ an index or an interator _pointing into_ an interval. Both make sense, but I chose the first one initially.

core/base/segmented_range.hpp
```diff
@@ -0,0 +1,372 @@
...
+
+
+/**
+ * Represents a range of indices that is segmented into contiguous segments.
+ * Each segment has the shape `[begin, end)`, i.e. it is a half-open interval.
+ *
+ * @tparam IndexType  the type of indices used to represent the segments.
+ */
+template <typename IndexType>
+class segmented_range {
```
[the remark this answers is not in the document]
upsj: How about `segmented_index_range`? That also fits well with the related `segmented_value_range` typename.
I think the majority of use cases will be ones where we either have row pointers or non-overlapping segments. In those case, I think talking about a _segmented_ range still makes sense. Only when the ranges overlap does it no longer work. But the same would be true in a four-array CSR4 storage format, so it's not necessarily a new challenge.

core/base/segmented_range.hpp
MarcelKoch: Thinking about it, I now would also prefer how it is currently done. So I would prefer an iterator being named after what it returns.

core/base/segmented_range.hpp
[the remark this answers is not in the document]
MarcelKoch: I think `segmented_index_range` would be fine. I was just a bit caught up over the `range` part, but now coming back to it, I don't think this was reasonable.

core/base/segmented_range.hpp
```diff
@@ -0,0 +1,372 @@
...
+     */
+    class iterator {
+    public:
+        constexpr explicit iterator(segmented_range range, index_type segment)
+            : range_{range}, segment_{segment}
+        {}
+
+        struct enumerated_segment {
+            index_type index;
+            segment_type segment;
```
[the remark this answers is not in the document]
upsj: The "entries" of a segmented range are segments, and enumerating them you get a segment index and segment value back. Do you have a suggestion how to make this more clear through naming?

PR #1604, opened 2024-05-01
Allow `zip_iterator` to be used on the device

upsj:
This replaces `std::tuple` by a custom `device_tuple` implementation to be used on the device.

TODO

- [x] Merge #1603

### DIFF
--- a/core/base/iterator_factory.hpp
+++ b/core/base/iterator_factory.hpp
@@ -13,6 +13,8 @@
 #include <tuple>
 #include <utility>
 
+#include <ginkgo/core/base/types.hpp>
+
 #include "core/base/copy_assignable.hpp"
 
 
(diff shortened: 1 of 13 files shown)


fritzgoebel (approved):
LGTM.

I am not 100% sure I understand all that's going on with the `template <typename... Ts>` so definitely someone else should also give their opinion.

core/base/iterator_factory.hpp
```diff
@@ -24,6 +24,234 @@ template <typename... Iterators>
...
+                                            detail::device_tuple<Ts...>>::type&
+get(const detail::device_tuple<Ts...>& tuple);
+
+
+namespace detail {
+
+
+/** simplified constexpr std::tuple reimplementation for use in device code. */
+template <typename T, typename... Ts>
+class device_tuple {
```
MarcelKoch, line 90: I think this might have come up before, but why do we need our own implementation of `tuple`? Wouldn't providing alias for either thrust, onedpl, or std tuple work as well?

MarcelKoch (approved):

PR #1606, opened 2024-05-07
Reduce the DistributedBase interface

MarcelKoch:
This PR reduces the interface changes to the `DistributedBase` interface. It removes the virtual function `get_local_size`. The only place where this was used was in the multigrid solver. There it has been replaced by using the `run` dispatch with every possible distributed Matrix type.

It also makes `run` easier to use by automatically deducing the const qualifier.

Note: although this is based on #1544, it can be rebased on develop and merged before.

### DIFF
--- a/core/base/dispatch_helper.hpp
+++ b/core/base/dispatch_helper.hpp
@@ -25,7 +25,7 @@ using with_same_constness_t = std::conditional_t<
 /**
  *
  * @copydoc run<typename ReturnType, typename K, typename... Types, typename T,
- *              typename Func, typename... Args>
+ *              typename Func, typename... Args>(T*, Func&&, Args&&...)
  *
  * @note this is the end case
  */
@@ -36,8 +36,8 @@ ReturnType run_impl(T* obj, Func&&, Args&&...)
 }
 
 /**
- * @copydoc run<typename K, typename... Types, typename T, typename Func,
- *              typename... Args>
+ * @copydoc run<typename ReturnType, typename K, typename... Types, typename T,
+ *              typename Func, typename... Args>(T*, Func&&, Args&&...)
  *
  * @note This has additionally the return type encoded.
  */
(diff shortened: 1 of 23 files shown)


core/base/dispatch_helper.hpp
```diff
@@ -120,18 +126,49 @@ ReturnType run_impl(T obj, Func&& f, Args&&... args)
...
+ *               of const Base<K>
+ * @tparam ...Args  the additional arguments for the Func
+ *
+ * @param obj  the input object waiting converted
+ * @param f  the function will run if obj can be converted successfully
+ * @param args  the additional arguments for the function
+ *
+ * @return  the result of f invoked with obj cast to the first matching type
+ */
+template <template <class> class K, typename... Types, typename T,
```
yhmtsai, line 163:
```suggestion
template <template <class> class Base, typename... Types, typename T,
```
also the documentation needs to be adjusted. I have changed some documentations here, you may need to rebase and keep them similar.
With that, I think the following function can be deleted, right?

yhmtsai (commented):
LGTM. I have added the shared_ptr version previously, so I hold my approval until the shared_ptr still works here. It should work seamlessly with these changes though.

core/base/dispatch_helper.hpp
MarcelKoch: Which following function do you mean? This at l.221 or at l.252?

MarcelKoch:
@yhmtsai I've unified the `shared_ptr` implementation with the previous smart pointer implementation.

core/base/dispatch_helper.hpp
```diff
@@ -157,32 +178,54 @@ auto run(T obj, Func&& f, Args&&... args)
...
+ * template and run it.
+ *
+ * @tparam K  the current type to try in the conversion
  * @tparam ...Types  the other types will be tried in the conversion if K fails
- * @tparam T  the type of input object
+ * @tparam T  the type of input object waiting converted, has to be either
+ *            unique_ptr or shared_ptr
  * @tparam Func  the function type that is invoked if the object can be
- *               converted to std::shared_ptr<K>
+ *               converted to pointer of const Base<K>
```
yhmtsai, line 212:
```suggestion
 *               converted to pointer of Base<K>
```
The const depends on the given type now, right?

core/base/dispatch_helper.hpp
```diff
@@ -193,10 +236,9 @@ auto run(std::shared_ptr<T> obj, Func&& f, Args&&... args)
  * template and run it.
  *
  * @tparam Base  the Base class with one template
- * @tparam K  the current template type of B. pointer of const Base<K> is tried
- *            in the conversion.
- * @tparam ...Types  the other types will be tried in the conversion if K fails
- * @tparam T  the type of input object waiting converted
+ * @tparam ...Types  the types that will be tried with Base, i.e. Base<Types>...
+ * @tparam T  the type of input object waiting converted, has to be either
+ *            unique_ptr or shared_ptr
```
yhmtsai, line 241: It can only be `shared_ptr` because `dynamic_pointer_cast` only allow `shared_ptr` and the ownership issue.

core/base/dispatch_helper.hpp
```diff
@@ -68,52 +55,91 @@ ReturnType run_impl(std::shared_ptr<T> obj, Func&& f, Args&&... args)
...
+    using type = std::result_of_t<Func(
+        std::shared_ptr<detail::with_same_constness_t<K, T>>, Args...)>;
+#else
+    using type = std::invoke_result_t<
+        Func, std::shared_ptr<detail::with_same_constness_t<K, T>>, Args...>;
+#endif
+};
+
+template <typename T, typename K, typename Func, typename... Args>
+struct result_of<std::unique_ptr<T>, K, Func, Args...> {
```
yhmtsai, line 129: the unique_ptr is unnecessary because we can not use unique_ptr in these functions

yhmtsai (approved):
LGTM. but unique_ptr part is unneccessary now.

omp/multigrid/pgm_kernels.cpp
```diff
@@ -14,7 +14,6 @@
 
 #include <ginkgo/core/base/exception_helpers.hpp>
 #include <ginkgo/core/base/math.hpp>
-#include <ginkgo/core/multigrid/pgm.hpp>
```
pratikvn, line 17: These includes should not be removed ?

reference/multigrid/pgm_kernels.cpp
```diff
@@ -16,7 +16,6 @@
 #include <ginkgo/core/matrix/csr.hpp>
 #include <ginkgo/core/matrix/dense.hpp>
 #include <ginkgo/core/matrix/diagonal.hpp>
-#include <ginkgo/core/multigrid/pgm.hpp>
```
pratikvn, line 19: same here

pratikvn (approved):
I think the pgm headers should be left in the pgm related files. Otherwise LGTM

omp/multigrid/pgm_kernels.cpp
MarcelKoch: No, they *need* to be removed. The kernels don't depend on anything from the header, so it's not necessary. Besides, the pgm header now includes the mpi header through the distributed matrix, but those are not available, as MPI is not linked against the kernel libraries.
TBH, I'm a bit confused that it actually compiles in our tests. I suspect that the mpi headers are installed in some standard directory, and thus always available in our tests.

yhmtsai: the mpi header include is always wrapped with GINKGO_BUILD_MPI, so it will not be include if we do not build the mpi.
If we build the mpi, those are public of targets, and the tests are always linked to all gko library.

MarcelKoch: I wasn't talking about compiling the tests specifically. Instead I'm meant our CI in general. So I'm confused that the kernels compile in our CI, since they are not linked against MPI.

PR #1600, opened 2024-04-18
Add Batch jacobi device kernels

pratikvn:
This PR adds the device kernels for batch Jacobi preconditioner.

PR #1609, opened 2024-05-10
Add batch::cg solver device kernels

pratikvn:
This PR adds the CUDA/HIP/DPCPP device kernels for the batch CG solver.

A lot of similarities between existing bicgstab kernels and this one, which will be unified at a later stage.

### DIFF
--- a/common/cuda_hip/preconditioner/batch_scalar_jacobi.hpp.inc
+++ b/common/cuda_hip/preconditioner/batch_scalar_jacobi.hpp.inc
@@ -17,7 +17,7 @@ public:
     __host__ __device__ static constexpr int dynamic_work_size(
         const int num_rows, int)
     {
-        return num_rows;
+        return num_rows * sizeof(value_type);
     }
 
     /**
--- a/common/cuda_hip/solver/batch_cg_kernels.hpp.inc
+++ b/common/cuda_hip/solver/batch_cg_kernels.hpp.inc
@@ -1,0 +1,228 @@
+// SPDX-FileCopyrightText: 2017 - 2024 The Ginkgo authors
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+template <typename Group, typename BatchMatrixType_entry, typename PrecType,
+          typename ValueType>
+__device__ __forceinline__ void initialize(
+    Group subgroup, const int num_rows, const BatchMatrixType_entry& mat_entry,
+    const ValueType* const b_global_entry,
+    const ValueType* const x_global_entry, ValueType* const x_shared_entry,
+    ValueType* const r_shared_entry, const PrecType& prec_shared,
+    ValueType* const z_shared_entry, ValueType& rho_old_shared_entry,
+    ValueType* const p_shared_entry,
+    typename gko::remove_complex<ValueType>& rhs_norms_sh)
+{
+    // copy x from global to shared memory
+    // r = b
+    for (int iz = threadIdx.x; iz < num_rows; iz += blockDim.x) {
+        x_shared_entry[iz] = x_global_entry[iz];
+        r_shared_entry[iz] = b_global_entry[iz];
+    }
+    __syncthreads();
+
+    // r = b - A*x
+    advanced_apply(static_cast<ValueType>(-1.0), mat_entry, x_shared_entry,
+                   static_cast<ValueType>(1.0), r_shared_entry);
+    __syncthreads();
+
+    // z = precond * r
+    prec_shared.apply(num_rows, r_shared_entry, z_shared_entry);
+    __syncthreads();
+
+    if (threadIdx.x / config::warp_size == 0) {
+        // Compute norms of rhs
+        single_rhs_compute_norm2(subgroup, num_rows, b_global_entry,
+                                 rhs_norms_sh);
+    } else if (threadIdx.x / config::warp_size == 1) {
+        // rho_old = r' * z
+        single_rhs_compute_conj_dot(subgroup, num_rows, r_shared_entry,
+                                    z_shared_entry, rho_old_shared_entry);
+    }
+
+    // p = z
+    for (int iz = threadIdx.x; iz < num_rows; iz += blockDim.x) {
+        p_shared_entry[iz] = z_shared_entry[iz];
+    }
+}
+
+
+template <typename ValueType>
+__device__ __forceinline__ void update_p(const int num_rows,
+                                         const ValueType& rho_new_shared_entry,
+                                         const ValueType& rho_old_shared_entry,
+                                         const ValueType* const z_shared_entry,
+                                         ValueType* const p_shared_entry)
+{
+    for (int li = threadIdx.x; li < num_rows; li += blockDim.x) {
+        const ValueType beta = rho_new_shared_entry / rho_old_shared_entry;
+        p_shared_entry[li] = z_shared_entry[li] + beta * p_shared_entry[li];
+    }
+}
+
+
+template <typename Group, typename ValueType>
+__device__ __forceinline__ void update_x_and_r(
+    Group subgroup, const int num_rows, const ValueType& rho_old_shared_entry,
+    const ValueType* const p_shared_entry,
+    const ValueType* const Ap_shared_entry, ValueType& alpha_shared_entry,
+    ValueType* const x_shared_entry, ValueType* const r_shared_entry)
+{
+    if (threadIdx.x / config::warp_size == 0) {
+        single_rhs_compute_conj_dot(subgroup, num_rows, p_shared_entry,
+                                    Ap_shared_entry, alpha_shared_entry);
+    }
+    __syncthreads();
+
+    for (int li = threadIdx.x; li < num_rows; li += blockDim.x) {
+        const ValueType alpha = rho_old_shared_entry / alpha_shared_entry;
+        x_shared_entry[li] += alpha * p_shared_entry[li];
+        r_shared_entry[li] -= alpha * Ap_shared_entry[li];
+    }
+}
+
+
+template <typename StopType, const int n_shared, const bool prec_shared_bool,
+          typename PrecType, typename LogType, typename BatchMatrixType,
+          typename ValueType>
+__global__ void apply_kernel(const gko::kernels::batch_cg::storage_config sconf,
+                             const int max_iter,
+                             const gko::remove_complex<ValueType> tol,
+                             LogType logger, PrecType prec_shared,
+                             const BatchMatrixType mat,
+                             const ValueType* const __restrict__ b,
+                             ValueType* const __restrict__ x,
+                             ValueType* const __restrict__ workspace = nullptr)
+{
+    using real_type = typename gko::remove_complex<ValueType>;
+    const auto num_batch_items = mat.num_batch_items;
+    const auto num_rows = mat.num_rows;
+
+    constexpr auto tile_size = config::warp_size;
+    auto thread_block = group::this_thread_block();
+    auto subgroup = group::tiled_partition<tile_size>(thread_block);
+
+    for (size_type batch_id = blockIdx.x; batch_id < num_batch_items;
+         batch_id += gridDim.x) {
+        const int gmem_offset =
+            batch_id * sconf.gmem_stride_bytes / sizeof(ValueType);
+        extern __shared__ char local_mem_sh[];
+
+        ValueType* r_sh;
+        ValueType* z_sh;
+        ValueType* p_sh;
+        ValueType* Ap_sh;
+        ValueType* x_sh;
+        ValueType* prec_work_sh;
+
+        if (n_shared >= 1) {
+            r_sh = reinterpret_cast<ValueType*>(local_mem_sh);
+        } else {
+            r_sh = workspace + gmem_offset;
+        }
+        if (n_shared == 1) {
+            z_sh = workspace + gmem_offset;
+        } else {
+            z_sh = r_sh + sconf.padded_vec_len;
+        }
+        if (n_shared == 2) {
+            p_sh = workspace + gmem_offset;
+        } else {
+            p_sh = z_sh + sconf.padded_vec_len;
+        }
+        if (n_shared == 3) {
+            Ap_sh = workspace + gmem_offset;
+        } else {
+            Ap_sh = p_sh + sconf.padded_vec_len;
+        }
+        if (n_shared == 4) {
+            x_sh = workspace + gmem_offset;
+        } else {
+            x_sh = Ap_sh + sconf.padded_vec_len;
+        }
+        if (!prec_shared_bool && n_shared == 5) {
+            prec_work_sh = workspace + gmem_offset;
+        } else {
+            prec_work_sh = x_sh + sconf.padded_vec_len;
+        }
+
+        __shared__ uninitialized_array<ValueType, 1> rho_old_sh;
+        __shared__ uninitialized_array<ValueType, 1> rho_new_sh;
+        __shared__ uninitialized_array<ValueType, 1> alpha_sh;
+        __shared__ real_type norms_rhs_sh[1];
+        __shared__ real_type norms_res_sh[1];
+
+        const auto mat_entry =
+            gko::batch::matrix::extract_batch_item(mat, batch_id);
+        const ValueType* const b_global_entry =
+            gko::batch::multi_vector::batch_item_ptr(b, 1, num_rows, batch_id);
+        ValueType* const x_global_entry =
+            gko::batch::multi_vector::batch_item_ptr(x, 1, num_rows, batch_id);
+
+        // generate preconditioner
+        prec_shared.generate(batch_id, mat_entry, prec_work_sh);
+
+        // initialization
+        // compute b norms
+        // r = b - A*x
+        // z = precond*r
+        // rho_old = r' * z (' is for hermitian transpose)
+        // p = z
+        initialize(subgroup, num_rows, mat_entry, b_global_entry,
+                   x_global_entry, x_sh, r_sh, prec_shared, z_sh, rho_old_sh[0],
+                   p_sh, norms_rhs_sh[0]);
+        __syncthreads();
+
+        // stopping criterion object
+        StopType stop(tol, norms_rhs_sh);
+
+        int iter = 0;
+        for (; iter < max_iter; iter++) {
+            norms_res_sh[0] = sqrt(abs(rho_old_sh[0]));
+            __syncthreads();
+            if (stop.check_converged(norms_res_sh)) {
+                logger.log_iteration(batch_id, iter, norms_res_sh[0]);
+                break;
+            }
+
+            // Ap = A * p
+            simple_apply(mat_entry, p_sh, Ap_sh);
+            __syncthreads();
+
+            // alpha = rho_old / (p' * Ap)
+            // x = x + alpha * p
+            // r = r - alpha * Ap
+            update_x_and_r(subgroup, num_rows, rho_old_sh[0], p_sh, Ap_sh,
+                           alpha_sh[0], x_sh, r_sh);
+            __syncthreads();
+
+            // z = precond * r
+            prec_shared.apply(num_rows, r_sh, z_sh);
+            __syncthreads();
+
+            if (threadIdx.x / config::warp_size == 0) {
+                // rho_new =  (r)' * (z)
+                single_rhs_compute_conj_dot(subgroup, num_rows, r_sh, z_sh,
+                                            rho_new_sh[0]);
+            }
+            __syncthreads();
+
+            // beta = rho_new / rho_old
+            // p = z + beta * p
+            update_p(num_rows, rho_new_sh[0], rho_old_sh[0], z_sh, p_sh);
+            __syncthreads();
+
+            // rho_old = rho_new
+            if (threadIdx.x == 0) {
+                rho_old_sh[0] = rho_new_sh[0];
+            }
+            __syncthreads();
+        }
+
+        logger.log_iteration(batch_id, iter, norms_res_sh[0]);
+
+        // copy x back to global memory
+        single_rhs_copy(num_rows, x_sh, x_global_entry);
+        __syncthreads();
+    }
+}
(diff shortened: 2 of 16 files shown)


common/cuda_hip/solver/batch_cg_kernels.hpp.inc
```diff
@@ -0,0 +1,228 @@
...
+            prec_work_sh = workspace + gmem_offset;
+        } else {
+            prec_work_sh = x_sh + sconf.padded_vec_len;
+        }
+
+        __shared__ uninitialized_array<ValueType, 1> rho_old_sh;
+        __shared__ uninitialized_array<ValueType, 1> rho_new_sh;
+        __shared__ uninitialized_array<ValueType, 1> alpha_sh;
+        __shared__ real_type norms_rhs_sh[1];
+        __shared__ real_type norms_res_sh[1];
```
MarcelKoch, line 153: why are these normal arrays, while the ones above are `uninitialized_array`?

dpcpp/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -296,11 +295,15 @@ void apply_kernel(const gko::kernels::batch_bicgstab::storage_config sconf,
...
                item_ct1);
     item_ct1.barrier(sycl::access::fence_space::global_and_local);
+    int iter = 0;
 
     // stopping criterion object
     StopType stop(tol, norms_rhs_sh);
+    if (stop.check_converged(norms_res_sh)) {
+        logger.log_iteration(batch_id, iter, norms_res_sh[0]);
+        return;
+    }
```
MarcelKoch, line 305: this is the first thing that happens in the loop. I don't think checking it before is necessary.

dpcpp/solver/batch_cg_kernels.dp.cpp
```diff
@@ -43,12 +43,187 @@ namespace batch_cg {
...
+        // template
+        // launch_apply_kernel<StopType, subgroup_size, n_shared_total>
+        if (num_rows <= 32 && n_shared_total == 6)
+            launch_apply_kernel<StopType, 16, 6>(
+                sconf, logger, prec, mat, b.values, x.values, workspace_data,
+                group_size, shared_size);
+        else if (num_rows <= 256 && n_shared_total == 6)
+            launch_apply_kernel<StopType, 32, 6>(
+                sconf, logger, prec, mat, b.values, x.values, workspace_data,
+                group_size, shared_size);
```
MarcelKoch, line 177: is this case required? If `n_shared_total == 6` than the call in the switch is the same, regardless of `num_rows`.

dpcpp/solver/batch_cg_kernels.hpp.inc
```diff
@@ -0,0 +1,252 @@
...
+    const auto tid = item_ct1.get_local_linear_id();
+    if (sg.get_group_id() == 0) {
+        single_rhs_compute_conj_dot_sg(num_rows, p_shared_entry,
+                                       Ap_shared_entry, alpha_shared_entry,
+                                       item_ct1);
+    }
+    item_ct1.barrier(sycl::access::fence_space::global_and_local);
+    if (tid == 0) {
+        alpha_shared_entry = rho_old_shared_entry / alpha_shared_entry;
+    }
```
MarcelKoch, line 90: don't you also need a barrier after this? In the cuda/hip this is solved by storing this in a local variable instead of shared memory.

dpcpp/solver/batch_cg_kernels.hpp.inc
```diff
@@ -0,0 +1,252 @@
...
+    const auto group = item_ct1.get_group();
+    const auto batch_id = item_ct1.get_group_linear_id();
+
+    // The whole workgroup have the same values for these variables, but
+    // these variables are stored in reg. mem, not on SLM
+    ValueType* rho_old_sh;
+    ValueType* rho_new_sh;
+    ValueType* alpha_sh;
+    real_type* norms_rhs_sh;
+    real_type* norms_res_sh;
```
MarcelKoch, line 130: nit: could merge declaration with definition below.

core/solver/batch_bicgstab_kernels.hpp
```diff
@@ -88,8 +88,7 @@ void set_gmem_stride_bytes(storage_config& sconf,
         gmem_stride += prec_storage_bytes;
     }
     // align global memory chunks
-    sconf.gmem_stride_bytes =
-        gmem_stride > 0 ? ceildiv(gmem_stride, align_bytes) * align_bytes : 0;
+    sconf.gmem_stride_bytes = ceildiv(gmem_stride, align_bytes) * align_bytes;
```
MarcelKoch, line 91: I don't think this is equivalent to the one before. Is that intentional?

MarcelKoch (dismissed):
Looks good. I've left mostly nits and some open-ended questions.

Maybe also list the unrelated changes. So far I gathered these:
- snake_case for bicgstab kernel_caller
- return bytes from scalar jacobi `dynamic_work_size`

common/cuda_hip/solver/batch_cg_kernels.hpp.inc
pratikvn: The types are different. The ones above can be complex, while these are guaranteed to be real. I believe we had some issues previously with using complex types in shared memory, so we use `uninitialized_array` for when using static shared memory.

dpcpp/solver/batch_cg_kernels.hpp.inc
pratikvn: good catch!

core/solver/batch_bicgstab_kernels.hpp
pratikvn: I think this was left over from previous code and not necessary anymore as ceildiv should be okay with gmem_stride being zero

common/cuda_hip/preconditioner/batch_scalar_jacobi.hpp.inc
```diff
@@ -17,7 +17,7 @@ public:
     __host__ __device__ static constexpr int dynamic_work_size(
         const int num_rows, int)
     {
-        return num_rows;
+        return num_rows * sizeof(value_type);
```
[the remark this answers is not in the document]
pratikvn: Yes, but I moved it to #1600 now. I think that will be merged first, so will rebase this on that afterwards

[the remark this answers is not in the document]
MarcelKoch: Then maybe change the base of the PR? Makes it easier to review.

dpcpp/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -296,11 +295,11 @@ void apply_kernel(const gko::kernels::batch_bicgstab::storage_config sconf,
                p_sh, p_hat_sh, v_sh, norms_rhs_sh[0], norms_res_sh[0],
                item_ct1);
     item_ct1.barrier(sycl::access::fence_space::global_and_local);
+    int iter = 0;
```
MarcelKoch, line 298: this code change seems unnecessary.

dpcpp/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -296,11 +295,11 @@ void apply_kernel(const gko::kernels::batch_bicgstab::storage_config sconf,
...
 
     // stopping criterion object
     StopType stop(tol, norms_rhs_sh);
 
-    int iter = 0;
```
MarcelKoch, line 303: combined with this one

MarcelKoch (approved):

cuda/solver/batch_cg_kernels.cu
```diff
@@ -51,12 +51,178 @@ namespace batch_cg {
...
+        GKO_ASSERT(sconf.gmem_stride_bytes % sizeof(value_type) == 0);
+
+        value_type* const workspace_data = workspace.get_data();
+
+        // Template parameters launch_apply_kernel<StopType, n_shared,
+        // prec_shared>
+        if (sconf.prec_shared)
+            launch_apply_kernel<StopType, 5, true>(
+                sconf, logger, prec, mat, b.values, x.values, workspace_data,
+                block_size, shared_size);
```
yhmtsai, line 181:
```suggestion
        if (sconf.prec_shared) {
            launch_apply_kernel<StopType, 5, true>(
                sconf, logger, prec, mat, b.values, x.values, workspace_data,
                block_size, shared_size);
        }
```

dpcpp/solver/batch_cg_kernels.dp.cpp
```diff
@@ -43,12 +43,183 @@ namespace batch_cg {
...
+public:
+    kernel_caller(std::shared_ptr<const DefaultExecutor> exec,
+                  const settings<remove_complex<ValueType>> settings)
+        : exec_{std::move(exec)}, settings_{settings}
+    {}
+
+    template <typename StopType, const int subgroup_size,
+              const int n_shared_total, typename PrecType, typename LogType,
+              typename BatchMatrixType>
+    __dpct_inline__ void launch_apply_kernel(
```
yhmtsai, line 72:
```suggestion
    void launch_apply_kernel(
```
it is not the device code

dpcpp/solver/batch_cg_kernels.dp.cpp
```diff
@@ -43,12 +43,183 @@ namespace batch_cg {
...
+        const size_type num_batch_items = mat.num_batch_items;
+        const auto num_rows = mat.num_rows;
+        const auto num_rhs = b.num_rhs;
+        GKO_ASSERT(num_rhs == 1);
+
+        auto device = exec_->get_queue()->get_device();
+        auto max_group_size =
+            device.get_info<sycl::info::device::max_work_group_size>();
+        int group_size =
+            device.get_info<sycl::info::device::max_work_group_size>();
```
yhmtsai, line 133:
```suggestion
        int group_size = max_group_size;
```
the following `group_size = std::min...` should be the same as you set group_size directly to the `get_group_size(...)`?

dpcpp/solver/batch_cg_kernels.dp.cpp
```diff
@@ -43,12 +43,183 @@ namespace batch_cg {
...
+        if (group_size > num_rows) {
+            group_size = get_group_size(num_rows);
+        }
+        group_size = std::min(
+            std::max(group_size, static_cast<int>(2 * config::warp_size)),
+            static_cast<int>(max_group_size));
+
+        // reserve 3 for intermediate rho, norms,
+        // alpha, and for reduce_over_group
```
yhmtsai, line 142: it does not fully match the comments?

dpcpp/solver/batch_cg_kernels.hpp.inc
```diff
@@ -0,0 +1,247 @@
...
+    ValueType* const __restrict__ r_shared_entry, sycl::nd_item<3> item_ct1)
+{
+    auto sg = item_ct1.get_sub_group();
+    const auto tid = item_ct1.get_local_linear_id();
+    if (sg.get_group_id() == 0) {
+        single_rhs_compute_conj_dot_sg(num_rows, p_shared_entry,
+                                       Ap_shared_entry, alpha_shared_entry,
+                                       item_ct1);
+    }
+    item_ct1.barrier(sycl::access::fence_space::global_and_local);
```
yhmtsai, line 87: maybe todo in the future. the barrier mode depends on whether it is shared or not.

common/cuda_hip/solver/batch_cg_kernels.hpp.inc
```diff
@@ -0,0 +1,228 @@
...
+        StopType stop(tol, norms_rhs_sh);
+
+        int iter = 0;
+        for (; iter < max_iter; iter++) {
+            norms_res_sh[0] = sqrt(abs(rho_old_sh[0]));
+            __syncthreads();
+            if (stop.check_converged(norms_res_sh)) {
+                logger.log_iteration(batch_id, iter, norms_res_sh[0]);
+                break;
+            }
```
yhmtsai, line 186: the check converged place is different from sycl?

hip/solver/batch_cg_kernels.hip.cpp
```diff
@@ -50,12 +50,154 @@ namespace batch_cg {
...
+    // that maximizes the occupancy on an AMD GPU (MI200). HIP does not have an
+    // API to query the number of registers a function uses.
+    const int num_regs_used_per_thread = 64;
+    int max_regs_blk = 0;
+    GKO_ASSERT_NO_HIP_ERRORS(hipDeviceGetAttribute(
+        &max_regs_blk, hipDeviceAttributeMaxRegistersPerBlock,
+        exec->get_device_id()));
+    const int max_threads_regs = (max_regs_blk / num_regs_used_per_thread);
+    int max_threads = std::min(max_threads_regs, device_max_threads);
+    max_threads = max_threads <= 1024 ? max_threads : 1024;
```
yhmtsai, line 75: max_threads may not be the multiple of warp_size

hip/solver/batch_cg_kernels.hip.cpp
```diff
@@ -50,12 +50,154 @@ namespace batch_cg {
...
+        constexpr int align_multiple = 8;
+        const int padded_num_rows =
+            ceildiv(mat.num_rows, align_multiple) * align_multiple;
+        int shmem_per_blk = 0;
+        GKO_ASSERT_NO_HIP_ERRORS(hipDeviceGetAttribute(
+            &shmem_per_blk, hipDeviceAttributeMaxSharedMemoryPerBlock,
+            exec_->get_device_id()));
+        const int block_size =
+            get_num_threads_per_block<BatchMatrixType>(exec_, mat.num_rows);
+        GKO_ASSERT(block_size >= 2 * config::warp_size);
```
yhmtsai, line 131: maybe also check it is divisible with warp_size

yhmtsai (commented):
some sycl algorithm part (not kernel details) are different from cuda hip.

dpcpp/solver/batch_cg_kernels.dp.cpp
pratikvn: Yes, you are right. I will initially set it to `get_group_size` for all now.

dpcpp/solver/batch_cg_kernels.hpp.inc
pratikvn: Yes, noted down here: https://github.com/ginkgo-project/ginkgo/issues/1376

test/solver/batch_cg_kernels.cpp
```diff
@@ -179,7 +179,8 @@ TEST_F(BatchCg, CanSolveLargeBatchSizeHpdSystem)
     auto mat =
         gko::share(gko::test::generate_diag_dominant_batch_matrix<const Mtx>(
             exec, num_batch_items, num_rows, true));
-    auto linear_system = setup_linsys_and_solver(mat, num_rhs, tol, max_iters);
+    auto linear_system =
+        setup_linsys_and_solver(mat, num_rhs, tol / 100, max_iters);
```
yhmtsai, line 183: Stopping by residual norm but checking the true error is still weird to me. the scale is 50000, which is a little high to me.
you also check the residual norm, so I do not hold this pr by this question now

test/solver/batch_cg_kernels.cpp
```diff
@@ -190,7 +191,7 @@ TEST_F(BatchCg, CanSolveLargeBatchSizeHpdSystem)
                                                  &logger->get_num_iterations());
     auto res_norm = gko::make_temporary_clone(exec->get_master(),
                                               &logger->get_residual_norm());
-    GKO_ASSERT_BATCH_MTX_NEAR(res.x, linear_system.exact_sol, tol * 50);
+    GKO_ASSERT_BATCH_MTX_NEAR(res.x, linear_system.exact_sol, tol * 500);
```
yhmtsai, line 194: in the following checking
```
 EXPECT_LE(res_norm->get_const_data()[i] /
                      linear_system.host_rhs_norm->get_const_values()[i],
                  tol * 20);
```
should it be tol/100 as you set the target is tol/100?

test/solver/batch_cg_kernels.cpp
pratikvn: Do you suggest I dont check with the true solution at all, because I am definitely having issues with DPCPP with the tolerance. I also agree that 500 is too high.

yhmtsai: Yes, but the tol needs to be lower than current setup.
If it is the issue only in dpcpp, I think we need to be more careful on this.
For example, using the same n_shared settings, subgroup_size, group_size, and maybe using the same impl of reduction not from reduce_by_group on sycl and cuda side. If they still give quite different result, I think there are something wrong in the sync.

dpcpp/solver/batch_cg_kernels.dp.cpp
```diff
@@ -43,12 +43,179 @@ namespace batch_cg {
...
+        int group_size = get_group_size(num_rows);
+        group_size = std::min(
+            std::max(group_size, static_cast<int>(2 * config::warp_size)),
+            static_cast<int>(max_group_size));
+
+        // reserve 3 for intermediate rho,
+        // alpha, reduce_over_group, and two norms
+        // If the value available is negative, then set it to 0
+        const int static_var_mem =
+            (group_size + 3) * sizeof(ValueType) + 2 * sizeof(real_type);
```
yhmtsai, line 141: still miss group_size?

dpcpp/solver/batch_cg_kernels.hpp.inc
```diff
@@ -0,0 +1,244 @@
...
+    const auto sg = item_ct1.get_sub_group();
+    const int sg_id = sg.get_group_id();
+    const int sg_size = sg.get_local_range().size();
+    const int num_sg = sg.get_group_range().size();
+
+    const auto group = item_ct1.get_group();
+    const auto batch_id = item_ct1.get_group_linear_id();
+
+    // The whole workgroup have the same values for these variables, but
+    // these variables are stored in reg. mem, not on SLM
```
yhmtsai, line 125: will it be only one shared value of whole workgroup?

test/solver/batch_cg_kernels.cpp
```diff
@@ -190,16 +190,15 @@ TEST_F(BatchCg, CanSolveLargeBatchSizeHpdSystem)
                                                  &logger->get_num_iterations());
     auto res_norm = gko::make_temporary_clone(exec->get_master(),
                                               &logger->get_residual_norm());
-    GKO_ASSERT_BATCH_MTX_NEAR(res.x, linear_system.exact_sol, tol * 50);
     for (size_t i = 0; i < num_batch_items; i++) {
         auto comp_res_norm = res.host_res_norm->get_const_values()[i] /
                              linear_system.host_rhs_norm->get_const_values()[i];
         ASSERT_LE(iter_counts->get_const_data()[i], max_iters);
         EXPECT_LE(res_norm->get_const_data()[i] /
```
yhmtsai, line 197: are the` host_res_norm and res_norm from logger different?

test/solver/batch_cg_kernels.cpp
```diff
@@ -190,16 +190,15 @@ TEST_F(BatchCg, CanSolveLargeBatchSizeHpdSystem)
...
                       linear_system.host_rhs_norm->get_const_values()[i],
-                  tol * 20);
+                  tol * 100);
```
yhmtsai, line 199: is the stopping criterion not based on this condition < tol?
It may contain the numerical rounding error from cg itself, but 100 times is 1e-3?

yhmtsai: and later test does not need to change the tol.

dpcpp/solver/batch_cg_kernels.hpp.inc
pratikvn: for each of them, yes.

dpcpp/solver/batch_cg_kernels.dp.cpp
pratikvn: Sorry, I dont understand what you mean.

test/solver/batch_cg_kernels.cpp
pratikvn: Yes, host_res_norm is the explicit residual norm: `||b-Ax||`

dpcpp/solver/batch_cg_kernels.dp.cpp
yhmtsai: the description only mention 3 for the result, right? but what' the group_size * sizeof(ValueType) here for

pratikvn: That was for local memory for reduce_over_group. But I think that was in a previous code. So, removed now.

yhmtsai: does the cuda/hip part need to change? or they indeed use shared_memory?

cuda/solver/batch_cg_kernels.cu
```diff
@@ -51,12 +51,178 @@ namespace batch_cg {
...
+    {
+        using real_type = gko::remove_complex<value_type>;
+        const size_type num_batch_items = mat.num_batch_items;
+        constexpr int align_multiple = 8;
+        const int padded_num_rows =
+            ceildiv(mat.num_rows, align_multiple) * align_multiple;
+        auto shem_guard =
+            gko::kernels::cuda::detail::shared_memory_config_guard<
+                value_type>();
+        const int shmem_per_blk =
```
yhmtsai, line 150: here does not consider the 3 * ValueType and 2 * real_type.

yhmtsai: same for hip

yhmtsai: Okay, here is a bit different from SYCL. It only considers the DnamicSharedMemory and the getter does not contain static shared memory limitation information.

yhmtsai (approved):
LGTM. That will be great if you can confirm the CUDA/HIP only considers the DynamicSharedMemory Size

pratikvn:
@yhmtsai , yes. For CUDA/HIP we only consider dynamic shared memory and only that needs to be passed into the kernel. I dont think it is necessary to check for the static shared memory with CUDA/HIP

PR #1610, opened 2024-05-13
Fix the msvc failure on json config

yhmtsai:
Sorry for another pr on this. I thought the failure from msvc cuda, which is now always failed.
To avoid setting the lib path in msvc, gko makes all test works in the library path (dll location).
I changed the working directory to access testing file in #1607  However, it can not find the dll path now.
follow the `matrices/config.hpp.in` to pass the file path

### DIFF
--- a/dev_tools/scripts/format_header.sh
+++ b/dev_tools/scripts/format_header.sh
@@ -15,6 +15,8 @@ convert_header () {
             fi
         elif [ "${header_file}" = "matrices/config.hpp" ]; then
             echo "#include \"${header_file}\""
+        elif [ "${header_file}" = "extensions/test/config/file_location.hpp" ]; then
+            echo "#include \"${header_file}\""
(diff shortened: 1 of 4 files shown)


MarcelKoch (approved):

upsj (approved):
LGTM! An alternative (for small files) would be creating a temporary file in the working directory and deleting it afterwards. For the large matrices, it is sensible to store them separately, but for just JSON input, we could do it inline

yhmtsai:
I thought the same thing, too. but I also think we can use the filesystem to replace the path when we move to c++17.

upsj:
@yhmtsai `std::filesystem` is still a bit annoying to use on some slightly older compilers, see https://en.cppreference.com/w/cpp/filesystem

> Notes
> 
> Using this library may require additional compiler/linker options. GNU implementation prior to 9.1 requires linking with -lstdc++fs and LLVM implementation prior to LLVM 9.0 requires linking with -lc++fs

so I think we better avoid it.

PR #1392, opened 2023-08-17
Factory config

yhmtsai:
As we discussed, we only focus on the LinOpFactory generation first.
Also, the dispatch only supports type now.
The public interface only provides LinOpFactory, not concrete type generation if we do not want the function with executor arg in Solver.

PR #1613, opened 2024-05-17
Minimal stopping criteria config

MarcelKoch:
This PR adds a minimal specification for configuring stopping criteria. The test now contains an example for this configuration. Since the main implementation to configure the criteria is already implemented, this adds only a bit of dispatching.

This is intended to continue the discussion in #1392, while not blocking the release.

Todo:
- [x] discuss the specification
- [x] test all allowed key-value pairs

The stopping criteria can now be defined as:
```yaml
criteria:
  value_type: float32 # optional 
  # only one of the following is required 
  relative_residual_norm: 1e-6
  initial_residual_norm: 1e-6
  absolute_residual_norm: 1e-6
  relative_implicit_residual_norm: 1e-6
  initial_implicit_residual_norm: 1e-6
  absolute_implicit_residual_norm: 1e-6
  time: 1000
```

### DIFF
--- a/core/config/config_helper.cpp
+++ b/core/config/config_helper.cpp
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: 2017 - 2024 The Ginkgo authors
+// SPDX-FileCopyrightText: 2017 - 2025 The Ginkgo authors
 //
 // SPDX-License-Identifier: BSD-3-Clause
 
@@ -43,7 +43,9 @@ parse_or_get_factory<const stop::CriterionFactory>(const pnode& config,
     if (config.get_tag() == pnode::tag_t::string) {
         return detail::registry_accessor::get_data<stop::CriterionFactory>(
             context, config.get_string());
-    } else if (config.get_tag() == pnode::tag_t::map) {
+    }
+
+    if (config.get_tag() == pnode::tag_t::map) {
         static std::map<std::string,
                         std::function<deferred_factory_parameter<
                             gko::stop::CriterionFactory>(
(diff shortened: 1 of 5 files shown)


core/config/config_helper.cpp
```diff
@@ -62,8 +63,87 @@ parse_or_get_factory<const stop::CriterionFactory>(const pnode& config,
...
+parse_or_get_criteria(const pnode& config, const registry& context,
+                      const type_descriptor& td)
+{
+    if (config.get_tag() == pnode::tag_t::array) {
+        return parse_or_get_factory_vector<const stop::CriterionFactory>(
+            config, context, td);
+    }
+
+    if (config.get_tag() == pnode::tag_t::map) {
+        return parse_minimal_criteria(config, context, td);
```
yhmtsai, line 136:
```suggestion
        auto updated = config::update_type(td);
        return parse_minimal_criteria(config, context, updated);
```
It is to support no valuetype available outside

core/config/config_helper.cpp
```diff
@@ -62,8 +63,87 @@ parse_or_get_factory<const stop::CriterionFactory>(const pnode& config,
...
+    }
+    return res;
+}
+
+
+std::vector<deferred_factory_parameter<const stop::CriterionFactory>>
+parse_or_get_criteria(const pnode& config, const registry& context,
+                      const type_descriptor& td)
+{
+    if (config.get_tag() == pnode::tag_t::array) {
```
yhmtsai, line 130: The array can not accept 1 object now. you can use type to distinguish the path and reuse the parse_or_get_factory

core/config/config_helper.cpp
MarcelKoch: I'm not sure what you mean here. I can specify the value type of the residual nom criterion, as you can see in the tests.

yhmtsai: It's based on the type_descriptor from the outer loop.
I mean something like
```
"stop": {
  "value_type": "float64",
  "residual_norm": ...
}
```
in case no precision information from outside or want to specify certain precision for stop.

MarcelKoch: I will try a bit to make it work. If it's a reasonable effort I will adapt it.

MarcelKoch: It was pretty simple to add, so it's now available.

core/config/config_helper.cpp
```diff
@@ -55,9 +57,101 @@ parse_or_get_factory<const stop::CriterionFactory>(const pnode& config,
...
+             create_residual_mapping("absolute_implicit_residual_norm",
+                                     "absolute", configure_implicit_residual)}};
+
+    auto value_type_it = config.get_map().find("value_type");
+    type_descriptor updated_td =
+        value_type_it == config.get_map().end()
+            ? td
+            : type_descriptor(value_type_it->second.get_string(),
+                              td.get_local_index_typestr(),
+                              td.get_global_index_typestr());
```
yhmtsai, line 119: you can also use `updated_td = config::update_type(td);`

yhmtsai (approved):
LGTM! only nit left is that the updated_td can reuse the existing function

core/config/config_helper.cpp
MarcelKoch: Thanks for the hint!

upsj (approved):
LGTM!

PR #1616, opened 2024-05-20
Prepare for more CUDA/HIP unification

upsj:
To simplify the review of #1516, this does all the necessary changes to automate the majority of the other changes

- Add necessary switching headers
- Provide device namespace macro via compiler definitions
- Add necessary (namespace) aliases
- adapt math lib includes and namespaces
- uniformize files


cuda/distributed/vector_kernels.cu
```diff
@@ -5,6 +5,9 @@
 #include "core/distributed/vector_kernels.hpp"
 
 
+#include <functional>
```
MarcelKoch, line 8: why is this necessary? The kernel doesn't use `std::function`, and I would have guessed that `thrust/functional.h` would be more appropriate.

cuda/factorization/par_ilu_kernels.cu
```diff
@@ -5,12 +5,14 @@
 #include "core/factorization/par_ilu_kernels.hpp"
 
 
+#include <ginkgo/core/base/std_extensions.hpp>
```
MarcelKoch, line 8: why is this necessary?

cuda/factorization/par_ilut_filter_kernels.cu
```diff
@@ -12,15 +12,16 @@
 #include <ginkgo/core/matrix/dense.hpp>
 
 
+#include "common/cuda_hip/base/config.hpp"
+#include "common/cuda_hip/base/runtime.hpp"
```
MarcelKoch, line 16: when is it necessary to add this header? It's not added to the file above for instance.

hip/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.21)
 include(${PROJECT_SOURCE_DIR}/cmake/template_instantiation.cmake)
-add_instantiation_files(. matrix/csr_kernels.instantiate.hip.cpp CSR_INSTANTIATE)
-add_instantiation_files(. matrix/fbcsr_kernels.instantiate.hip.cpp FBCSR_INSTANTIATE)
+add_instantiation_files(. matrix/csr_kernels.instantiate.hip.cpp CSR_kernels.instantiate)
+add_instantiation_files(. matrix/fbcsr_kernels.instantiate.hip.cpp FBCSR_kernels.instantiate)
```
MarcelKoch, line 4: his didn't change for cuda, is that intended?

hip/base/config.hip.hpp
```diff
@@ -6,15 +6,13 @@
...
-#include <hip/hip_runtime.h>
-
-
 #include <hip/device_functions.h>
 
 
 #include <ginkgo/core/base/types.hpp>
 
 
+#include "common/cuda_hip/base/runtime.hpp"
```
MarcelKoch, line 15: I think you shouldn't change the order of the runtime header. It causes compilation errors in some cases.

include/ginkgo/core/base/executor.hpp
```diff
@@ -1602,6 +1602,11 @@ class CudaExecutor : public detail::ExecutorBase<CudaExecutor>,
      */
     cublasContext* get_cublas_handle() const { return cublas_handle_.get(); }
 
+    /**
+     * @copydoc get_cublas_handle()
+     */
+    cublasContext* get_blas_handle() const { return get_cublas_handle(); }
```
MarcelKoch, line 1608: I guess it would make sense to deprecate the other version in the near future.

MarcelKoch (changes requested):
In general, looks good to me. I've left some remarks, which are mostly quite minor.

The only thing I would like to see changed is to revert the header ordering of the hip runtime header. I think it is already causing compilation issues.

Another question is about the `as_cuda_type`, and `as_hip_type`. Should they also be unified?

common/cuda_hip/base/runtime.hpp
```diff
@@ -0,0 +1,18 @@
+// SPDX-FileCopyrightText: 2017 - 2024 The Ginkgo authors
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+#ifndef GKO_COMMON_CUDA_HIP_BASE_RUNTIME_HPP_
+#define GKO_COMMON_CUDA_HIP_BASE_RUNTIME_HPP_
+
+
+#ifdef GKO_COMPILING_CUDA
+// nothing needed here
```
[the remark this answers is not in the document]
upsj: That gets included by `nvcc` automatically

cuda/distributed/vector_kernels.cu
upsj: Doesn't seem to be necessary, it was only added to make HIP and CUDA consistent

hip/base/config.hip.hpp
upsj: That's a symptom of missing includes in our other files. Include order should not matter

include/ginkgo/core/base/executor.hpp
upsj: Agreed

hip/base/config.hip.hpp
MarcelKoch: Is it really just because of our header? I thought you had to include `hip/hip_runtime.h` before other hip includes. I guess it could be worthwhile to investigate which of our header are causing trouble.

cuda/factorization/par_ilu_kernels.cu
upsj: probably isn't, again consistency between CUDA and HIP

cuda/base/cublas_bindings.hpp
```diff
@@ -249,6 +249,20 @@ inline void destroy(cublasHandle_t handle)
...
+
+namespace blas {
+
+
+using namespace cublas;
+
+
+#define BLAS_OP_N CUBLAS_OP_N
+#define BLAS_OP_T CUBLAS_OP_T
+#define BLAS_OP_C CUBLAS_OP_C
```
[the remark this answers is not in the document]
upsj: I would prefer leaving them as macros, since the original ones are as well

hip/base/config.hip.hpp
upsj: the issue was hip/device_functions.h, which is not a self-sufficient header file and caused the compilation issues.

MarcelKoch: Thanks for finding this. THat is a better solution than forcing the include order.

hip/CMakeLists.txt
MarcelKoch: pinging this again

MarcelKoch (approved):

cuda/factorization/par_ilut_filter_kernels.cu
upsj: That's basically an iwyu question - do we include everything directly, or do we use some transitive includes? HIP's runtime is necessary basically any time you call a kernel or refer to some constants like warpSize afaik, I think that cleanup should be part of another PR.

omp/CMakeLists.txt
```diff
@@ -54,7 +54,7 @@ target_sources(ginkgo_omp
     )
 
 ginkgo_compile_features(ginkgo_omp)
-target_compile_definitions(ginkgo_omp PRIVATE GKO_COMPILING_OMP)
+target_compile_definitions(ginkgo_omp PRIVATE GKO_COMPILING_OMP GKO_GKO_DEVICE_NAMESPACE=omp)
```
upsj: replacement error 
```suggestion
target_compile_definitions(ginkgo_omp PRIVATE GKO_COMPILING_OMP GKO_DEVICE_NAMESPACE=omp)
```

hip/CMakeLists.txt
upsj: probably replacement issue without case-sensitive search, I'll revert it

MarcelKoch (approved):

yhmtsai (changes requested):
1. when changing `<hip/hip_runtime.h>` to `"common/cuda_hip/base/runtime.hpp"`, the header order should be also considered.
2. because the files are not merged together in this pr, some hip check in cuda files or some cuda check in hip files should be moved to another pr
3. `EXEC_NAMESPACE` -> `GKO_DEVICE_NAMESPACE` is not necessary

Also, if the code path are quite different between cuda/hip, I will prefer the file-based not the macro to distinguish for readibility.

test/base/index_range.cpp
```diff
@@ -30,7 +30,7 @@ class IndexRange : public CommonTestFixture {
 void run_range_for(std::shared_ptr<gko::EXEC_TYPE> exec,
                    gko::array<int>& result_array)
 {
-    gko::kernels::EXEC_NAMESPACE::run_kernel(
+    gko::kernels::GKO_DEVICE_NAMESPACE::run_kernel(
```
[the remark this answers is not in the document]
upsj: we define the same thing twice, once for `common/unified`, once for the tests, and we use two different macro names, I think it's more consistent this way

cuda/base/types.hpp
```diff
@@ -394,6 +394,10 @@ GKO_INLINE GKO_ATTRIBUTES constexpr
 }
 
 
+using deviceComplex = cuComplex;
```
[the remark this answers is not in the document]
upsj: They will be used in the PR that build on them #1516

test/base/index_range.cpp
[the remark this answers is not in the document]
upsj: An alternative way I just noticed:

```cpp
namespace gko {
namespace kernels {
namespace device {
using namespace GKO_DEVICE_NAMESPACE;
}
}
}
```
That would remove the macros from all test files and limit them to the test headers

cuda/matrix/ell_kernels.cu
```diff
@@ -212,7 +214,7 @@ void spmv(std::shared_ptr<const CudaExecutor> exec,
     const int num_worker_per_row = std::get<2>(data);
 
     /**
-     * info is the parameter for selecting the cuda kernel.
+     * info is the parameter for selecting the device kernel.
```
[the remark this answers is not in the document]
upsj: The intention of these changes is to make the files as identical as possible, so the other PR can consist (almost) entirely of automated changes. Otherwise it would be impossible to review

test/base/index_range.cpp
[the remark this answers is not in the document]
yhmtsai: I see. I prefer specifying them directly in each file.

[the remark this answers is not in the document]
upsj: Then I'll leave it as-is :)

cuda/solver/common_trs_kernels.cuh
```diff
@@ -75,47 +75,47 @@ struct CudaSolveStruct : gko::solver::SolveStruct {
...
 
         work.resize_and_reset(work_size);
 
-        cusparse::spsm_analysis(handle, CUSPARSE_OPERATION_NON_TRANSPOSE,
-                                CUSPARSE_OPERATION_NON_TRANSPOSE,
-                                one<ValueType>(), descr_a, descr_b, descr_c,
-                                CUSPARSE_SPSM_ALG_DEFAULT, spsm_descr,
-                                work.get_data());
+        sparselib::spsm_analysis(handle, CUSPARSE_OPERATION_NON_TRANSPOSE,
+                                 CUSPARSE_OPERATION_NON_TRANSPOSE,
```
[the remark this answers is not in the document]
upsj: There are only macros for cuBLAS, not for cuSPARSE (the interfaces differ too much there)

hip/base/exception.hip.cpp
```diff
@@ -8,7 +8,7 @@
 #include <string>
 
 
-#include <hip/hip_runtime.h>
+#include "common/cuda_hip/base/runtime.hpp"
```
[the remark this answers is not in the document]
upsj: fixed, thanks!

include/ginkgo/core/base/executor.hpp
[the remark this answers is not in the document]
upsj: not really, I'll deprecate them

yhmtsai (changes requested):
For the code self, only some macro can still use the unified one.
because 
But I have some concerns:
Will it make the header more fragment? need to jump many levels and figure out what it is. Also, the distinguishment of two implementations is based on compile flag, so the editor may not find the correct one or always show both. (my editor show pointer_gaurd both implementation but can not jump for blas::gemm)
another concern is that it may destroy iwyu although we do not really follow this rule in all codes seriously.

cuda/base/cusparse_bindings.hpp
```diff
@@ -1458,11 +1448,26 @@ GKO_BIND_CUSPARSE_IC0(float, cusparseScsric02);
...
+
+
+namespace sparselib {
+
+
+using namespace cusparse;
+
+
+#define SPARSELIB_OPERATION_TRANSPOSE CUSPARSE_OPERATION_TRANSPOSE
+#define SPARSELIB_OPERATION_NON_TRANSPOSE CUSPARSE_OPERATION_NON_TRANSPOSE
```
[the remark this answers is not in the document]
upsj: not sure how I missed those, thanks!

yhmtsai (approved):

Issue #1026, opened 2022-04-20
Investigate solver divergence between Reference and devices with only a few iterations


upsj:
In #973 we use really large tolerances, because already after a few iterations, Reference and device kernels diverge noticeably. We need to figure out whether this is just an inherent instability of the solvers on the input matrices we use, or a bug.

PR #1620, opened 2024-05-31
Add debug logger for solvers

upsj:
This debug logger prints out the value of all scalars inside the solver after each iteration:

Here we see the output of the `simple-solver` example with the logger running
```
gko::solver::Cg<double>::apply(0x124f04b60,0x124f04f00) of dimensions (19, 19) and 1 rhs
       Iteration           alpha            beta        prev_rho             rho
               0               0               0               1              19
               1               0             765       5.3109419              19
               2               0       151.51792       5.3109419       2.1541599
               3               0       56.610294      0.96997901       2.1541599
               4               0        22.91651      0.96997901      0.55031615
               5               0       13.637587      0.26380868      0.55031615
               6               0       5.8524907      0.26380868      0.14758275
               7               0       3.2949066      0.10013392      0.14758275
               8               0       2.3026212      0.10013392     0.051851311
               9               0       1.1447493     0.029006197     0.051851311
              10               0      0.64935163     0.029006197    0.0094813452
              11               0      0.22821574    0.0038048001    0.0094813452
              12               0     0.075631164    0.0038048001    0.0020622776
              13               0     0.047698364     0.001020997    0.0020622776
              14               0     0.023026955     0.001020997   0.00026116441
              15               0    0.0072957469   4.3166892e-05   0.00026116441
              16               0   0.00078229883   4.3166892e-05   6.9889839e-06
              17               0   0.00017981764   7.3755266e-07   6.9889839e-06
              18               0   1.9199351e-05   7.3755266e-07   8.2060081e-08
              19               0   1.9866215e-06   8.9249994e-28   8.2060081e-08
```

Should be useful for debugging #1563


include/ginkgo/core/log/solver_debug.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+namespace gko {
+namespace log {
+
+
+/**
+ * This Logger prints the value of all scalar values stored internally by the
+ * solver after each iteration. If the solver is applied to multiple right-hand
+ * sides, only the first right-hand side gets printed.
+ */
```
upsj, line 24:
```suggestion
 * It needs to be attached to the solver you want to debug.
 */
```

upsj:
Also relevant for investigating #1026 (which is where this idea came up originally)

upsj:
Wondering about the name here - it might make sense to also have a version that stores all scalars and intermediate vectors in files - should it be `SolverDebugPrint` or similar?

include/ginkgo/core/log/solver_debug.hpp
```diff
@@ -0,0 +1,80 @@
+// SPDX-FileCopyrightText: 2017 - 2024 The Ginkgo authors
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+#ifndef GKO_PUBLIC_CORE_LOG_SOLVER_DEBUG_HPP_
+#define GKO_PUBLIC_CORE_LOG_SOLVER_DEBUG_HPP_
+
+
+#include <iosfwd>
```
tcojean, line 9: Why iosfwd instead of iostream (or ostream, as that's what you're using)

tcojean: also needs the memory header for shared_ptr

core/log/solver_debug.cpp
```diff
@@ -0,0 +1,126 @@
...
+#include <ginkgo/core/log/solver_debug.hpp>
+
+
+#include <iomanip>
+
+
+#include <ginkgo/core/base/exception_helpers.hpp>
+#include <ginkgo/core/base/name_demangling.hpp>
+#include <ginkgo/core/log/logger.hpp>
+#include <ginkgo/core/solver/solver_base.hpp>
```
tcojean, line 14: minor: needs dense mtx include

tcojean (commented):
Some comments.

Also, to clarify, this would not give very helpful output with `Gmres` right, due to the nature of its "scalars"  (in relation with `krylov_dim` most of the time)?

I think in general, having a version that stores output to files/json/wherever, in particular for vector cases would be really helpful as a general solution.

core/log/solver_debug.cpp
```diff
@@ -0,0 +1,321 @@
...
+#include "core/base/dispatch_helper.hpp"
+
+
+namespace gko {
+namespace log {
+namespace {
+
+
+template <typename Functor>
+static bool dispatch_type(const LinOp* value, Functor fn)
```
MarcelKoch, line 32: you could decouple the type checking from the dispatch, i.e. have
```
bool is_dense(const LinOp* value){
  return dynamic_cast<const ConvertibleTo<matrix::Dense<double>>*>(value) || 
         dynamic_cast<const ConvertibleTo<matrix::Dense<std::complex<double>>>*>(value);
}
```
And then use the run dispatch like
```
run<matrix::Dense, double, float, std::complex<float>, std::complex<double>>(
  value, [&](const auto* dense){
    auto host = make_temporary_clone(dense->get_executor()->get_master(), dense);
    stream << host->at(0, 0);
  });
```
I would at least prefer having the check by itself, because it would simplifiy the if construct in `print_scalar`.

core/test/log/solver_debug.cpp
```diff
@@ -0,0 +1,177 @@
...
+        {"solver_debug_test_0_z", nullptr},
+        {"solver_debug_test_1_alpha", nullptr},
+        {"solver_debug_test_1_beta", nullptr},
+        {"solver_debug_test_1_implicit_sq_residual_norm", this->zero.get()},
+        {"solver_debug_test_1_minus_one", nullptr},
+        {"solver_debug_test_1_one", nullptr},
+        {"solver_debug_test_1_p", nullptr},
+        {"solver_debug_test_1_prev_rho", nullptr},
+        {"solver_debug_test_1_q", nullptr},
+        {"solver_debug_test_1_r", nullptr},
```
MarcelKoch, line 159: why are these (and many others) nullptrs? They should be filled after one iteration.

MarcelKoch (changes requested):
I think the naming needs to be updated, and the internal dispatch in the scalar printer.

core/test/log/solver_debug.cpp
upsj: I didn't want to check each individual file, only one from each "category" (workspace vector, each argument to `on_iteration`)

MarcelKoch: That's fair. Maybe still check if the file exists?

upsj: Good point, I only delete them right now

fritzgoebel (approved):
This looks very useful!
LGTM

core/log/solver_debug.cpp
```diff
@@ -0,0 +1,321 @@
...
+{
+    const auto host_exec = value->get_executor()->get_master();
+    using conv_to_double = ConvertibleTo<matrix::Dense<double>>;
+    using conv_to_complex = ConvertibleTo<matrix::Dense<std::complex<double>>>;
+    if (dynamic_cast<const conv_to_double*>(value)) {
+        auto host_vec = matrix::Dense<double>::create(host_exec);
+        host_vec->copy_from(value);
```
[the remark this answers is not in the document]
upsj: this is not always a clone, in some cases it can be a conversion. If I didn't use `double`, I would have to handle every float type separately - going through the highest precision type for debugging purposes seemed more appropriate to me than duplicating the amount of code.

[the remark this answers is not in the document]
MarcelKoch: In that case, maybe it would be better to use `convert_to`?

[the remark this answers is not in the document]
upsj: might make things clearer, yes.

core/log/solver_debug.cpp
MarcelKoch: pinging this again. IMO the runtime check should at least be separated.

upsj: going through `matrix::Dense` requires me to dispatch to all dense types, while going through ConvertibleTo only needs to choose between real/complex, that wouldn't really simplify the code. Do you mean using `run<ConvertibleTo, matrix::Dense<double>, matrix::Dense<float>>`?

MarcelKoch: My main point is how the dispatch is used. I'm not a fan of the
```c++
} else {
    if (!dispatch_type(
            value, [&](auto vector) { stream << vector->at(0, 0); })) {
        stream << "<unknown>";
    }
}
```
construct.
By separating the runtime check you would have
```c++
} else if(is_dense(vector)) {
    // dispatch how you want
} else {
    stream << "<unknown>";
}
```
which I find more digestible.

upsj: the code became a bit more complex due to the need for `detail::pointee<decltype(value)>::result_type`, but I think it's still a clearer separation, so I applied the suggestion.

core/log/solver_debug.cpp
```diff
@@ -0,0 +1,315 @@
...
+                            decltype(vector)>::result_type;
+                        auto host_vec = vector_type::create(host_exec);
+                        vector->convert_to(host_vec);
+                        stream << host_vec->at(0, 0);
+                    });
+
+            } else {
+                stream << "<unknown>";
+            }
+        }
```
MarcelKoch, line 155: nit:
```suggestion
        } else if (is_dense(value)) {
            auto host_exec = value->get_executor()->get_master();
            run<ConvertibleTo<matrix::Dense<double>>,
                ConvertibleTo<matrix::Dense<std::complex<double>>>>(
                value, [&](auto vector) {
                    using vector_type = typename detail::pointee<
                        decltype(vector)>::result_type;
                    auto host_vec = vector_type::create(host_exec);
                    vector->convert_to(host_vec);
                    stream << host_vec->at(0, 0);
                });
        } else {
            stream << "<unknown>";
        }
```

include/ginkgo/core/log/solver_debug.hpp
```diff
@@ -0,0 +1,73 @@
...
+namespace gko {
+namespace log {
+
+
+/**
+ * This Logger prints the value of all scalar values stored internally by the
+ * solver after each iteration. If the solver is applied to multiple right-hand
+ * sides, only the first right-hand side gets printed.
+ */
+class SolverDebug : public Logger {
```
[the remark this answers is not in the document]
MarcelKoch: Pinging this again, I'm still unhappy with the name. It's rather non-descriptive, so users will have no idea what this is doing just from the name.

MarcelKoch (changes requested):
LGTM, except the name.

include/ginkgo/core/log/solver_debug.hpp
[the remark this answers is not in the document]
upsj: How would you describe what this logger is used for? My first thought was "It's for debugging solvers", so the name made sense. But from a user perspective, it's maybe slightly unintuitive. OTOH, will users need to use this regularly?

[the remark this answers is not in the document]
MarcelKoch: It's for debugging solvers in a very specific sense, by printing the internal state after every iteration. That can be reflected in the name. Just 'debug' can mean anything and nothing.
Regarding usage: it could also be used to achieve the same thing as in #1517, so perhaps users will use it for that.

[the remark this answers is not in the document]
upsj: How about `SolverProgress`?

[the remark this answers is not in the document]
MarcelKoch: TBH, I kinda warmed up to my first suggestion. `SolverProgress` could easily be understood to mean just the residual norm, so I think referencing the state might be useful.

thoasm:
How about `SolverIntricaciesLogger`?

MarcelKoch (approved):

PR #1625, opened 2024-06-17
Fix Ginkgo version in CITATION.cff

greole:
The version in CITATION.cff is not the most recent, this PR bumps the version to the latest released version 1.8.0.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -26,11 +26,12 @@ authors:
 - family-names: "Riemer"
   given-names: "Lukas"
 - family-names: "Tsai"
-  given-names: "Yuhsiang"
+  given-names: "Yu-Hsiang"
 title: "Ginkgo: A Modern Linear Operator Algebra Framework for High Performance Computing"
-version: 1.5.0
-date-released: 2022-11-12
+version: 1.8.0
+date-released: 2024-06-13
 url: "https://github.com/ginkgo-project/ginkgo"
+license: BSD-3-Clause
 preferred-citation:
   type: article
   authors:


tcojean:
I always wondered if we should bump the version or not in the citation. My feeling would be not, because the paper doesn't reflect the library content and capacity of version 1.8.0, it maybe does up to 1.3.0 or something like this, in fact.

greole:
Ah I see your point. I thought the first part is meant for a general citation of the library. And additionally we have the prefered  paper section which should state the library version indepently.

greole:
Looking at the [validation guide](https://github.com/citation-file-format/citation-file-format/blob/main/schema-guide.md#preferred-citation), especially the title and prefrerred-citation section, I understand it as the current version of the software and the prefrerred-citation: is an additional paper which can specify the version separately.

CITATION.cff
```diff
@@ -28,9 +28,10 @@ authors:
 - family-names: "Tsai"
   given-names: "Yuhsiang"
 title: "Ginkgo: A Modern Linear Operator Algebra Framework for High Performance Computing"
-version: 1.5.0
-date-released: 2022-11-12
+version: 1.8.0
+date-released: 2022-06-13
```
tcojean, line 32:
```suggestion
date-released: 2024-06-13
```

tcojean (approved):
Thanks for updating this and checking how it works.

A last problem for me is the authors list in the citation if this is the software citation. In the best case, it should reflect (or directly come from) the `contributors.txt` file, no? Except if we consider main authors only, whatever that means. 

Does anyone have any idea how other packages do it?

greole:
There is an example of a mixed authors list with persons and anonymous entries.

```
authors:
  - given-names: Stephan
    family-names: Druskat
  - name: "The Research Software project"
```

I guess if we want to distinguish between main and other authors we would use something like
`  - name: "The Gingko contributors" 

or similar.`

CITATION.cff
```diff
@@ -28,9 +28,10 @@ authors:
 - family-names: "Tsai"
   given-names: "Yuhsiang"
```
yhmtsai, line 29:
```suggestion
- family-names: "Tsai"
  given-names: "Yu-Hsiang"
```
Could you also change my first name with -?

yhmtsai (approved):
Could you also help me to change my first name?

PR #1626, opened 2024-06-18
C API Support

youwuyou:
This PR adds the initial support of the C API implementation for interoperating with other programming languages, such as Fortran, Python, and Julia. The langauge linkage with key word `extern "c"` is used to enable the interfacing.

**C API Design:**
- Conveys an uniform naming convention as specified [here](https://ginkgo-for-julia.org/dev/programmer-guide/ginkgo-c-library-api/)
- Usage of macros and `std::dynamic_pointer_cast` to avoid code duplication

**Changes:**
- Added header file `include/ginkgo/c_api.h` (define macros, declares function/structs) and source file `core/c_api.cpp` (concretize definitions of declared functions that internally calls `ginkgo`)
- Necessary changes in CMake files `cmake/install_helpers.cmake` and `core/CMakeLists.txt` to include the C API

**Example Usage:**
- Calling ginkgo from C through the C API can be seen [here](https://ginkgo-for-julia.org/dev/programmer-guide/debugging/)
- The Julia package [Ginkgo.jl](https://github.com/youwuyou/Ginkgo.jl/tree/main) calls ginkgo routines with `ccall` through the from Clang generated Julia side [api.jl](https://github.com/youwuyou/Ginkgo.jl/blob/main/src/api.jl), which internally executes compiled dynamic libraries of ginkgo. It supports both locally compiled dynamic libraries or ones that are hosted under an artifact package [ginkgo_jll.jl](https://github.com/JuliaBinaryWrappers/ginkgo_jll.jl).


include/ginkgo/c_api.h
```diff
@@ -0,0 +1,851 @@
...
+#define C_API_H
+
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
+/* ----------------------------------------------------------------------
+ * Data type constants in GINKGO
+ * ---------------------------------------------------------------------- */
+enum _GKO_DATATYPE_CONST {
```
MarcelKoch, line 16: I don't think this is used? Maybe remove it until it becomes necessary.

core/c_api.cpp
```diff
@@ -0,0 +1,492 @@
...
+{
+    if (auto omp_exec = std::dynamic_pointer_cast<gko::OmpExecutor>(
+            exec_st_ptr->shared_ptr)) {
+        return new gko_executor_st{
+            gko::CudaExecutor::create(device_id, exec_st_ptr->shared_ptr)};
+    } else if (auto reference_exec =
+                   std::dynamic_pointer_cast<gko::ReferenceExecutor>(
+                       exec_st_ptr->shared_ptr)) {
+        return new gko_executor_st{
+            gko::CudaExecutor::create(device_id, exec_st_ptr->shared_ptr)};
```
MarcelKoch, line 151: could be combined
```suggestion
    if ( std::dynamic_pointer_cast<gko::OmpExecutor>(
            exec_st_ptr->shared_ptr) ||
         std::dynamic_pointer_cast<gko::ReferenceExecutor>(
            exec_st_ptr->shared_ptr)) {
        return new gko_executor_st{
            gko::CudaExecutor::create(device_id, exec_st_ptr->shared_ptr)};
```
same for hip and sycl

core/c_api.cpp
```diff
@@ -0,0 +1,492 @@
...
+    } else {
+        return nullptr;
+    }
+}
+
+size_t ginkgo_executor_cuda_get_num_devices()
```
MarcelKoch, line 157: this should return an `int` instead of `size_t`, since that's what the ginkgo function returns.
Or is the `size_t` required from the julia side.
This also applies to many of the executor functions.

include/ginkgo/c_api.h
```diff
@@ -0,0 +1,851 @@
...
+ * executor
+ * @return size_t No. of computing units
+ */
+size_t ginkgo_executor_gpu_item_get_num_computing_units(
+    gko_executor exec_st_ptr);
+
+/* ----------------------------------------------------------------------
+ * Library functions for creating arrays and array operations in GINKGO
+ * ---------------------------------------------------------------------- */
+DECLARE_ARRAY_OVERLOAD(int16_t, int16_t, i16)
```
MarcelKoch, line 729: I don't think the `int16` overload is useful. It can't be used to create matrices, or basically anything relevant. Instead, a `size_t` might be more helpful, although this is also not necessary for the functions currently available in the C-API.

MarcelKoch (changes requested):
Thanks for the work, this looks already quite good! I left some remarks below.

The big points I want to bring up are:
- executor function names. I'm not sure `_cpu`, `_gpu` parts are necessary for executor function names. From the rest of the name, it is already implied if it is for a cpu or gpu executor or not. Also this implies a type hierarchy level that is not really present in ginkgo at the moment.
- naming consistency. Should the prefix be `ginkgo` or `gko`. I might tend to `gko`, since that matches our namespace.
- error handling. A lot of Ginkgo functions can throw, and I'm unsure what will happen on the C side in that case. I guess the standard approach would be to catch the exception and return error codes instead.

include/ginkgo/c_api.h
```diff
@@ -0,0 +1,851 @@
...
+
+/* ----------------------------------------------------------------------
+ * Library functions for deferred factory parameters in GINKGO
+ * ---------------------------------------------------------------------- */
+/**
+ * @brief Struct containing the shared pointer to a ginkgo deferred factory
+ * parameter
+ *
+ */
+struct gko_deferred_factory_parameter_st;
```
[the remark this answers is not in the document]
youwuyou: I was thinking to use `gko_` for struct objects and `ginkgo_` for actual methods, and thus I used `gko_linop`, `gko_deferred_factory_parameter`, `gko_executor` etc.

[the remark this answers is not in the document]
MarcelKoch: I think it's more common for c libraries to have a single prefix that they stick to.

include/ginkgo/c_api.h
```diff
@@ -0,0 +1,851 @@
...
+        gko_matrix_dense_##_name_dense x, gko_matrix_dense_##_name_dense beta, \
+        gko_matrix_dense_##_name_dense y);
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+/* ----------------------------------------------------------------------
+ * C memory management
+ * ---------------------------------------------------------------------- */
+void c_char_ptr_free(char* ptr);
```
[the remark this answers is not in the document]
youwuyou: I was experimenting a bit around the I/O interfacing and tried to have dense matrices printed using `_write_mtx` (defined under macro).

include/ginkgo/c_api.h
youwuyou: Yes this was unused indeed, will remove it

include/ginkgo/c_api.h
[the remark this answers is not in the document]
MarcelKoch: That is what I would prefer.

include/ginkgo/c_api.h
```diff
@@ -0,0 +1,851 @@
+// SPDX-FileCopyrightText: 2017 - 2024 The Ginkgo authors
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+// Copyright (c) 2017-2023, the Ginkgo authors
```
greole, line 5:
```suggestion
```
See line 1.

include/ginkgo/c_api.h
```diff
@@ -0,0 +1,851 @@
...
+#ifndef C_API_H
+#define C_API_H
```
greole, line 7:
```suggestion
#ifndef GKO_C_API_H
#define GKO_C_API_H
```
Maybe better to prefix with GKO to avoid potential collisions.

include/ginkgo/c_api.h
```diff
@@ -0,0 +1,851 @@
...
+ * construction by removing the repetitive typing of array's name.
+ *
+ * @param _ctype  Type name of the element type in C
+ *
+ * @param _cpptype  Type name of the element type in C++
+ *
+ * @param _name  Name of the datatype of the array
+ *
+ */
+#define DEFINE_ARRAY_OVERLOAD(_ctype, _cpptype, _name)                         \
```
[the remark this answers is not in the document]
greole: True, since they are only required in `c_api.cpp`, they could also be defined there.

include/ginkgo/c_api.h
```diff
@@ -0,0 +1,851 @@
...
+    char* ginkgo_matrix_dense_##_name##_write_mtx(                           \
+        gko_matrix_dense_##_name mat_st_ptr);
+
+
+/**
+ * @brief A build instruction for defining gko::matrix::Csr<Tv, Ti>, simplifying
+ * its construction by removing the repetitive typing of array's name.
+ *
+ * @param _ctype_value  Type name of the element type in C
+ *
```
greole, line 270:
```suggestion
```
We rather keep all the @param together instead of separating them by an empty line.

include/ginkgo/c_api.h
```diff
@@ -0,0 +1,851 @@
...
+/* ----------------------------------------------------------------------
+ * MACROS for generating structs for wrapping
+ * ---------------------------------------------------------------------- */
+
+/**
+ * @brief A build instruction for defining gko::array<T>, simplifying its
+ * construction by removing the repetitive typing of array's name.
+ *
+ * @param _ctype  Type name of the element type in C
+ *
```
greole, line 36:
```suggestion
```
We rather keep the @param together instead of separating them by an empty line.

include/ginkgo/c_api.h
youwuyou: Thank you! I will edit the ones that do not match the format

include/ginkgo/c_api.h
```diff
@@ -0,0 +1,850 @@
...
+        char* cstr = new char[str.length() + 1];                            \
+        std::strcpy(cstr, str.c_str());                                     \
+        return cstr;                                                        \
+    }
+
+/**
+ * @brief A build instruction for declaring gko::matrix::Dense<T> in the C API
+ * header file
+ */
+#define DECLARE_DENSE_OVERLOAD(_ctype, _cpptype, _name)                      \
```
thoasm, line 225: `GKO_` Prefix

include/ginkgo/c_api.h
```diff
@@ -0,0 +1,850 @@
...
+ * @param _cpptype_value  Type name of the element type in C++
+ *
+ * @param _cpptype_index  Type name of the index type in C++
+ *
+ * @param _name  Name of the datatype of the sparse CSR matrix
+ *
+ * @param _name_dense Name of the datatype of the dense matrix for apply
+ * function
+ */
+#define DEFINE_CSR_OVERLOAD(_ctype_value, _ctype_index, _cpptype_value,        \
```
thoasm, line 281: `GKO_` and move it to `c_api.cpp`

include/ginkgo/c_api.h
```diff
@@ -0,0 +1,850 @@
...
+    {                                                                          \
+        mat_st_ptr->mat->apply(alpha->mat, x->mat, beta->mat, y->mat);         \
+    }
+
+
+/**
+ * @brief A build instruction for declaring gko::matrix::Csr<Tv,Ti> in the C API
+ * header file
+ */
+#define DECLARE_CSR_OVERLOAD(_ctype_value, _ctype_index, _cpptype_value,       \
```
thoasm, line 392: `GKO_` prefix

include/ginkgo/c_api.h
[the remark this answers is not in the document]
thoasm: If you need this as a public function, it must be prefixed (for functions, you seem to use `ginkgo_`).

thoasm (commented):
Good work in general!
Most of my comments are about names and moving things to the CPP file.
Also, you seem to mix the prefix `gko_` (mostly structs) and `ginkgo_` (functions).
I don't think we need to make a distinction here,
I would recommend to use `gko_` for everything (and `GKO_` for macros), as you have already discussed with @MarcelKoch in the comments.

I need to read up a bit more on C APIs to give my final review, so consider this 1/2.

thoasm:
I also have a general question: Should we add some form of CI to check if the API is still working?

include/ginkgo/c_api.h
```diff
@@ -0,0 +1,850 @@
...
+                (*exec).shared_ptr, gko::dim<2>{size.rows, size.cols})};    \
+    }                                                                       \
+                                                                            \
+    gko_matrix_dense_##_name ginkgo_matrix_dense_##_name##_create_view(     \
+        gko_executor exec, gko_dim2_st size, _ctype* values, size_t stride) \
+    {                                                                       \
+        return new gko_matrix_dense_##_name##_st{                           \
+            gko::matrix::Dense<_ctype>::create(                             \
+                (*exec).shared_ptr, gko::dim<2>{size.rows, size.cols},      \
+                gko::array<_ctype>::view((*exec).shared_ptr, size.rows,     \
```
pratikvn, line 117: Just a note.
```suggestion
                gko::array<_ctype>::view((*exec).shared_ptr, size.rows*stride,     \
```

core/c_api.cpp
```diff
@@ -0,0 +1,491 @@
...
+
+/* ----------------------------------------------------------------------
+ * C memory management
+ * ---------------------------------------------------------------------- */
+void c_char_ptr_free(char* ptr) { delete[] ptr; }
+
+/* ----------------------------------------------------------------------
+ * Library functions for retrieving configuration information in GINKGO
+ * ---------------------------------------------------------------------- */
+void ginkgo_version_get()
```
yhmtsai:
```suggestion
void print_ginkgo_version()
```
because it prints the version actually, I will prefer print not get

include/ginkgo/c_api.h
```diff
@@ -0,0 +1,850 @@
...
+ * @return size_t Second dimension
+ */
+size_t ginkgo_dim2_cols_get(gko_dim2_st dim);
+
+/* ----------------------------------------------------------------------
+ * Library functions for executors (Creation, Getters) in GINKGO
+ * ---------------------------------------------------------------------- */
+/**
+ * @brief Struct containing the shared pointer to a ginkgo executor
+ *
```
yhmtsai:
```suggestion
```
no need additional line when finish the comments

include/ginkgo/c_api.h
```diff
@@ -0,0 +1,850 @@
...
+ * @param exec_st_ptr Raw pointer to the shared pointer of the executor to be
+ * deleted
+ */
+void ginkgo_executor_delete(gko_executor exec_st_ptr);
+
+/**
+ * @brief Returns the master OmpExecutor of this Executor.
+ *
+ * @param exec_st_ptr Raw pointer to the shared pointer of the current executor
+ * @return gko_executor Raw pointer to the shared pointer of the master executor
```
yhmtsai:
```suggestion
 * @param exec_st_ptr Raw pointer to the shared pointer of the current executor
 * 
 * @return gko_executor Raw pointer to the shared pointer of the master executor
```
we usually have the new line between different group

include/ginkgo/c_api.h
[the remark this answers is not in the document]
yhmtsai: I think it is good place. With that, I will say core/c_api.cpp should also be there and create additional library rather in ginkgo core

core/c_api.cpp
```diff
@@ -0,0 +1,491 @@
...
+        return nullptr;
+    }
+}
+
+size_t ginkgo_executor_hip_get_num_devices()
+{
+    return gko::HipExecutor::get_num_devices();
+}
+
+size_t ginkgo_executor_gpu_thread_get_num_multiprocessor(
```
yhmtsai: do you need these functionality?
for example `gpu_thread_get_num_multiprocessor` and `gpu_thread_get_num_warps_per_sm`?
I think executor only need create, delete, sync, maybe accessible

core/c_api.cpp
```diff
@@ -0,0 +1,491 @@
...
+        return dpcpp_exec->get_num_computing_units();
+    } else {
+        return 0;
+    }
+}
+
+/* ----------------------------------------------------------------------
+ * Library functions for creating arrays and array operations in GINKGO
+ * ---------------------------------------------------------------------- */
+DEFINE_ARRAY_OVERLOAD(int16_t, int16_t, i16)
```
yhmtsai: because we already have the data type enum, should we just use generic api?
not specific *_f64_i32 but *(data type, data type)
but it can be discussed later no need to be in this pr.

include/ginkgo/c_api.h
youwuyou: unused enum removed in [f4f5769](https://github.com/ginkgo-project/ginkgo/pull/1626/commits/f4f576967ea5584e02f5bf8880b1bfb337160210)

include/ginkgo/c_api.h
youwuyou: addressed in [09af535](https://github.com/ginkgo-project/ginkgo/pull/1626/commits/09af535b804b1a88e43c5824a4b5859dd20d7496)

include/ginkgo/c_api.h
youwuyou: addressed in [09af535](https://github.com/ginkgo-project/ginkgo/pull/1626/commits/09af535b804b1a88e43c5824a4b5859dd20d7496)

include/ginkgo/c_api.h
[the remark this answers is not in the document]
youwuyou: added prefix with [4c78e0f](https://github.com/ginkgo-project/ginkgo/pull/1626/commits/4c78e0f7bca65191a5af6de3e82d2ae6f45a5887)

core/c_api.cpp
youwuyou: renamed in [d3b53f4](https://github.com/ginkgo-project/ginkgo/pull/1626/commits/d3b53f4bd6f10176f43b62009f42684c0a8e119a)

include/ginkgo/c_api.h
youwuyou: added new line between groups for all docstrings in [7f01d72](https://github.com/ginkgo-project/ginkgo/pull/1626/commits/7f01d7224588e2f6cfc2706e3a7ecee9702dd34c)

include/ginkgo/c_api.h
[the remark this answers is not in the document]
youwuyou: Thanks! I renamed both declare and define macros to have prefix `GKO_` in commit [725b7dc](https://github.com/ginkgo-project/ginkgo/pull/1626/commits/725b7dc0dd8d1e897630e6d5814da05e80b1a63e) and moved the define macros to `c_api.cpp` in [890232d](https://github.com/ginkgo-project/ginkgo/pull/1626/commits/890232dba5b0443ef659e1438c264f499cffe135).

include/ginkgo/c_api.h
youwuyou: Thanks, renamed in commit [725b7dc](https://github.com/ginkgo-project/ginkgo/pull/1626/commits/725b7dc0dd8d1e897630e6d5814da05e80b1a63e) and moved the define macros to `c_api.cpp` in [890232d](https://github.com/ginkgo-project/ginkgo/pull/1626/commits/890232dba5b0443ef659e1438c264f499cffe135), I will resolve other related comments to mark as done.

include/ginkgo/c_api.h
youwuyou: removed all redundant newlines with [704abd5](https://github.com/ginkgo-project/ginkgo/pull/1626/commits/704abd512e7a15a021a482f1fb31f29c0cabbac9)

core/c_api.cpp
youwuyou: Thanks! I incorporated the same simplification with commit [9030a56](https://github.com/ginkgo-project/ginkgo/pull/1626/commits/9030a563099f59d0ff1fd15193687146103cc9e0)

include/ginkgo/c_api.h
```diff
@@ -0,0 +1,851 @@
...
+    }                                                                          \
+                                                                               \
+    void ginkgo_array_##_name##_delete(gko_array_##_name array_st_ptr)         \
+    {                                                                          \
+        delete array_st_ptr;                                                   \
+    }                                                                          \
+                                                                               \
+    size_t ginkgo_array_##_name##_get_size(gko_array_##_name array_st_ptr)     \
+    {                                                                          \
+        return (*array_st_ptr).arr.get_size();                                 \
```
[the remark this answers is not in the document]
youwuyou: I picked the `(*struct).member` syntax because using `->` seems to be a bit hard to understand in situations as below:

https://github.com/ginkgo-project/ginkgo/blob/9030a563099f59d0ff1fd15193687146103cc9e0/core/c_api.cpp#L449-L456

then I stayed with it throughout the API

core/c_api.cpp
youwuyou: Yes, I am using these functionalities as a sanity check in the testing of executors in the Julia package, e.g. [here](https://github.com/youwuyou/Ginkgo.jl/blob/develop/test/base/test-executor.jl)

core/c_api.cpp
youwuyou: Thanks @MarcelKoch! If I recalled correctly, the design choice was made on one hand due to the convention in Julia side, on the other hand I encountered some problems with the use of `int`.

As an example, the following function uses `int` as the return type in commit [cd78d48](https://github.com/ginkgo-project/ginkgo/pull/1626/commits/cd78d481fdaaeabb67530628686d2f507375cea0).

https://github.com/ginkgo-project/ginkgo/blob/657c1cdc989510ecb5a6b6643e2528f16e7e3b27/core/c_api.cpp#L822-L826

while I was testing it, the compilation gives the following compilation warning when using `int` as the return type
```
  823 |     gko_log_convergence_f64 logger_st_ptr)
      |                                           
      |                                            -> int
/path/to//ginkgo/core/c_api.cpp:825:12: warning: narrowing conversion from 'size_type' (aka 'unsigned long') to signed type 'int' is implementation-defined [bugprone-narrowing-conversions,cppcoreguidelines-narrowing-conversions]
  825 |     return logger_st_ptr->shared_ptr->get_num_iterations();
```

the `unsigned long` mentioned here seems to be a "system dependent type" according to the warning, this mentioned in the bottom section of an official doc I looked up [here](https://docs.julialang.org/en/v1/manual/calling-c-and-fortran-code/#man-bits-types).

Switching to `size_t` does not result in this warning.

include/ginkgo/c_api.h
youwuyou: Thanks @pratikvn! I fixed the stride issue along with the solvers you newly added as in the commit [2fc73a6](https://github.com/ginkgo-project/ginkgo/pull/1626/commits/2fc73a63c7ddad4cf4cfc8afc9795c55a78fd179), I saw there are in all three commits from you with additional changes in the `electrostatic-runner` branch. I should have incorporated all of them now, where the other two were cherry-picked from your commits directly.

Also tested and [updated](https://github.com/youwuyou/Ginkgo.jl/commit/a81102cf52d24c79db2060ba1d8453c701c324f0) the Julia package and it's working fine.

include/ginkgo/c_api.h
```diff
@@ -0,0 +1,851 @@
...
+ * @param x_st_ptr Raw pointer to the shared pointer of the output vector where
+ * the result is stored
+ */
+void ginkgo_linop_apply(gko_linop A_st_ptr, gko_linop b_st_ptr,
+                        gko_linop x_st_ptr);
+
+//-------------------- Iterative solvers -----------------------------
+gko_linop ginkgo_linop_cg_preconditioned_f64_create(
+    gko_executor exec_st_ptr, gko_linop A_st_ptr,
+    gko_deferred_factory_parameter dfp_st_ptr, double reduction, int maxiter);
```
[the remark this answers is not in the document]
youwuyou: Changed the naming with commit [7ef82bc](https://github.com/ginkgo-project/ginkgo/pull/1626/commits/7ef82bca4a202af472b8662c0595d7acec3c36bd)

include/ginkgo/c_api.h
```diff
@@ -0,0 +1,851 @@
...
+ * @param b_st_ptr Raw pointer to the shared pointer of the input vector(s) on
+ * which the operator is applied
+ * @param x_st_ptr Raw pointer to the shared pointer of the output vector where
+ * the result is stored
+ */
+void ginkgo_linop_apply(gko_linop A_st_ptr, gko_linop b_st_ptr,
+                        gko_linop x_st_ptr);
+
+//-------------------- Iterative solvers -----------------------------
+gko_linop ginkgo_linop_cg_preconditioned_f64_create(
```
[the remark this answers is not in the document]
youwuyou: Thanks, matched the namespace with `solver` while removed `_preconditioned_` with commit [51b8fb3](https://github.com/ginkgo-project/ginkgo/pull/1626/commits/51b8fb3d4a49bafc7a1dad2619280afb5ad5cfc8), also tested for all solvers and updated in the Julia package [here](https://github.com/youwuyou/Ginkgo.jl/pull/35/commits/b693664f8e4bd1124a12360466981a8a0087202a)

include/ginkgo/c_api.h
```diff
@@ -0,0 +1,851 @@
...
+ * @brief Applies a linear operator to a vector (or a sequence of vectors).
+ *
+ * @param A_st_ptr Raw pointer to the shared pointer of the LinOp object
+ * @param b_st_ptr Raw pointer to the shared pointer of the input vector(s) on
+ * which the operator is applied
+ * @param x_st_ptr Raw pointer to the shared pointer of the output vector where
+ * the result is stored
+ */
+void ginkgo_linop_apply(gko_linop A_st_ptr, gko_linop b_st_ptr,
+                        gko_linop x_st_ptr);
```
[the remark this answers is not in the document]
youwuyou: Thanks! Actually no explicit type conversion is required at the C API level here because the polymorphic behavior of classes inheriting from `gko::LinOp` is preserved. We had the actual C++ objects encapsulated within C-compatible structs and accessed them via shared pointers. When methods in `c_api.cpp` are invoked, they execute the underlying C++ code while maintaining proper object access through these shared pointer handles.


I just added a MWE on the Julia package [here](https://github.com/youwuyou/Ginkgo.jl/commit/b374c8220b2166eb8f1e244e5477952668288c26) that demonstrates applying a sparse diagonal matrix where both `b_st_ptr` and `x_st_ptr` are vectors. The type conversions to native Julia types are handled transparently within my package implementation. In this example, `b_st_ptr` (aka `v_in` here) is initialized with all ones:

```julia
julia> include("linop-apply.jl")
10×10 SparseMatrixCSC{Float64, Int64} with 10 stored entries:
 1.0   ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅ 
  ⋅   2.0   ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅ 
  ⋅    ⋅   3.0   ⋅    ⋅    ⋅    ⋅    ⋅    ⋅     ⋅ 
  ⋅    ⋅    ⋅   4.0   ⋅    ⋅    ⋅    ⋅    ⋅     ⋅ 
  ⋅    ⋅    ⋅    ⋅   5.0   ⋅    ⋅    ⋅    ⋅     ⋅ 
  ⋅    ⋅    ⋅    ⋅    ⋅   6.0   ⋅    ⋅    ⋅     ⋅ 
  ⋅    ⋅    ⋅    ⋅    ⋅    ⋅   7.0   ⋅    ⋅     ⋅ 
  ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅   8.0   ⋅     ⋅ 
  ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅   9.0    ⋅ 
  ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅   10.0

Result of applying A to v_in (stored in v_out):
10-element Vector{Float64}:
  1.0
  2.0
  3.0
  4.0
  5.0
  6.0
  7.0
  8.0
  9.0
 10.0
```

matching the behavior as expected in the ginkgo [doc](https://ginkgo-project.github.io/ginkgo-generated-documentation/doc/develop/classgko_1_1LinOp.html#aff0379dbd0d86bd9ce2db03ac05f789f).

Issue #1627, opened 2024-06-18
MM-Format IO Checking


MarcelKoch:
Reading an MM-file doesn't check the file itself for consistency. For example, if the MM file defines a `6x4` matrix, but less then 24 lines after the size definition contain values, not all values of the created matrix/vector will be filled. 
IMO there should be a simple test that after a definition of a `MxN` dense matrix, exactly `M * N` lines should contain a value.
This checking would not apply to the coordinate format, as missing values will be assumed to be zero anyway.

MarcelKoch:
Actually, this is already checked.
But the issue remains if the MM-file contains multiple values per line. Then each value in a row seems to be interpreted as being on a row by itself. Thus the value is not interpreted as the next column value, but instead as a value for the next row.

Consider this file:
```
%%MatrixMarket matrix array real general
4 2
1 1
2 2
3 3
4 4
```
Then reading and writing it again one would expect (without deeper knowledge of the MM standard) the matrix entries in a single list using column major format:
```
%%MatrixMarket matrix array real general
4 2
1
2
3
4
1
2
3
4
```
But instead one gets
```
%%MatrixMarket matrix array real general
4 2
1
1
2
2
3
3
4
4
```

PR #1628, opened 2024-06-19
Discard trailing entries of a line in a MM-format file

MarcelKoch:
This PR will discard any character in a line of a MM-format file after the values have been read in. This addresses #1627, using multiple columns within the MM-file will now lead to an exception. However, the exception is thrown by the `read_entry` function and thus the exception message is not very helpful.

### DIFF
--- a/core/base/mtx_io.cpp
+++ b/core/base/mtx_io.cpp
@@ -517,8 +517,8 @@ private:
                 GKO_CHECK_STREAM(content, "error when reading matrix entry " +
                                               std::to_string(i));
                 modifier->insert_entry(row - 1, col - 1, entry, data);
-                content.ignore(max_streamsize,
-                               '\n');  // discards rest of the line
+                // discards rest of the line
+                content.ignore(max_streamsize, '\n');
             }
             return data;
         }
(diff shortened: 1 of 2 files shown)


thoasm (approved):
LGTM!
I just have one question: Does the `GKO_CHECK_STREAM` also throw in case we don't read enough elements?

nbeams:
> LGTM! I just have one question: Does the `GKO_CHECK_STREAM` also throw in case we don't read enough elements?

With the current `develop` (not this PR) I tried a case where I left off one element, and it threw a read error.

nbeams:
The "real case" was just me making a mistake in the dense/"array" format for Matrix Market (vs coordinate). I didn't check that the entries were supposed to be in column-major order, and instead provided them row-major (the way you would print out the matrix to look at it visually) and without the correct one-per-line formatting. Because I still had the correct total number of entries, Ginkgo did not throw an error (whitespace between values was treated the same as newlines), so I didn't realize the values were being interpreted incorrectly and therefore causing problems with my test.

So it was a case of user error for sure, but the idea behind this PR was that Ginkgo could help users by making this a failure mode -- if you do not have the correct number of lines, it will be an error, and you will realize your file formatting is incorrect. It has the benefit of more strictly adhering to the MM standard, apart from cases of user error.

MarcelKoch:
> Could you update the documentation with this behavior?

I don't think there is anything to update. This just fixes the behavior for our previous wrong parsing.

yhmtsai (approved):

PR #1629, opened 2024-06-24
Split batched solver compilation

MarcelKoch:
This PR splits up the compilation of the batched solvers in order to reduce the compilation times. It splits up the instantiations of the kernel launches depending on the number of vectors in shared memory. This is based on the same CMake mechanism as for the csr and fbcsr kernels.


upsj:
This should have a huge impact, excerpt from the HIP 5.14 debug build log
```
6534.89 hip/CMakeFiles/ginkgo_hip.dir/solver/batch_bicgstab_kernels.hip.cpp.o
```

core/solver/batch_dispatch.hpp
```diff
@@ -167,6 +167,41 @@ enum class log_type { simple_convergence_completion };
 }  // namespace log
 
 
+#define GKO_BATCH_INSTANTIATE_STOP(macro, ...)                          \
+    macro(__VA_ARGS__,                                                  \
+          ::gko::batch::solver::device::batch_stop::SimpleAbsResidual); \
+    template macro(                                                     \
```
MarcelKoch, line 173: the `template` here (and in the other macros below) could be removed, if the value/index type instantiation macros would accept variable number or arguments.

cuda/solver/batch_bicgstab_kernels.cuh
```diff
@@ -0,0 +1,271 @@
...
+           batch::MultiVector<ValueType>* const x,
+           batch::log::detail::log_data<remove_complex<ValueType>>& logdata)
+{
+    using cu_value_type = cuda_type<ValueType>;
+    auto dispatcher = batch::solver::create_dispatcher<ValueType>(
+        kernel_caller<cu_value_type>(exec, settings), settings, mat, precon);
+    dispatcher.apply(b, x, logdata);
+}
+
+GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE(GKO_DECLARE_BATCH_BICGSTAB_APPLY_KERNEL);
```
pratikvn, line 265: not needed here, already in the .cu file

pratikvn (changes requested):
In general, the idea looks good, but the pipelines are failing.

One thing against this approach is the readability and maintainability is seriously affected. The already complex batched code is even more complex and annoying to read now. We should maybe see if instead we dont do this split approach and instead maybe do what Jacobi does and have fewer cases as default, and only have full instantiations when necessary.

core/solver/batch_dispatch.hpp
MarcelKoch: That doesn't work until C++20. A macro with `(arg, ...)` requires *two* arguments before c++20.

MarcelKoch:
IMO the Jacobi instantiation is more complex than what is here. The kernel and the instantiations are directly together, instead of being generated by CMake, which makes it easier to follow for me.
I also merged the two .cpp files per solver, perhaps that can simplify things a bit again.

But I agree that the batch system needs an overhaul in general.

MarcelKoch:
> An alternative approach: https://github.com/ginkgo-project/ginkgo/tree/batch-optim

This seems to be quite orthogonal to this PR. With full optimizations enabled, there would be the same issue as before, so the fix from this PR is still needed. I don't see a reason why we should burden people that want the full optimizations enabled with those long compile times, for which we already have a fix available.
But we could add this into this PR.

pratikvn:
@MarcelKoch, can you please rebase this when you have some time and we can try to get it merged ?

MarcelKoch:
@pratikvn this should now be properly rebased.

cuda/solver/batch_bicgstab_launch.cuh
```diff
@@ -0,0 +1,112 @@
+// SPDX-FileCopyrightText: 2017 - 2024 The Ginkgo authors
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
```
pratikvn, line 4: Needs a header guard.

cuda/solver/batch_cg_launch.cuh
```diff
@@ -0,0 +1,104 @@
+// SPDX-FileCopyrightText: 2017 - 2024 The Ginkgo authors
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
```
pratikvn, line 4: Needs a header guard

hip/solver/batch_cg_launch.hip.hpp
```diff
@@ -0,0 +1,70 @@
+// SPDX-FileCopyrightText: 2017 - 2024 The Ginkgo authors
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
```
pratikvn, line 4: Needs header guard

hip/solver/batch_bicgstab_launch.hip.hpp
```diff
@@ -0,0 +1,78 @@
+// SPDX-FileCopyrightText: 2017 - 2024 The Ginkgo authors
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
```
pratikvn, line 4: Needs header guard

pratikvn (changes requested):
Some header guards are missing, but otherwise looks good to me.

hip/solver/batch_cg_launch.hip.hpp
```diff
@@ -0,0 +1,77 @@
...
+#ifndef GKO_CUDA_SOLVER_BATCH_CG_LAUNCH_HPP_
+#define GKO_CUDA_SOLVER_BATCH_CG_LAUNCH_HPP_
```
pratikvn, line 6:
```suggestion
#ifndef GKO_HIP_SOLVER_BATCH_CG_LAUNCH_HPP_
#define GKO_HIP_SOLVER_BATCH_CG_LAUNCH_HPP_
```

hip/solver/batch_bicgstab_launch.hip.hpp
```diff
@@ -0,0 +1,85 @@
...
+#ifndef GKO_CUDA_SOLVER_BATCH_BICGSTAB_LAUNCH_HIP_HPP_
+#define GKO_CUDA_SOLVER_BATCH_BICGSTAB_LAUNCH_HIP_HPP_
```
pratikvn, line 6:
```suggestion
#ifndef GKO_HIP_SOLVER_BATCH_BICGSTAB_LAUNCH_HIP_HPP_
#define GKO_HIP_SOLVER_BATCH_BICGSTAB_LAUNCH_HIP_HPP_
```

pratikvn (changes requested):
minor typos

hip/solver/batch_cg_launch.hip.hpp
MarcelKoch: we really should switch to `#pragma once`. This is quite annoying

pratikvn: I agree.

pratikvn (approved):

core/matrix/batch_struct.hpp
```diff
@@ -22,14 +22,14 @@ namespace csr {
 /**
  * Encapsulates one matrix from a batch of csr matrices.
  */
-template <typename ValueType, typename IndexType>
+template <typename ValueType, typename IndexType = const int32>
```
yhmtsai, line 25: does it need to be `const int32`?
My personal preference is `int32` and use `const IndexType` when it is necessary

MarcelKoch: It was necessary to add a default parameter, because the CSR and ELL have two template arguments, while Dense only has one. So, to handle them the same in the template instantiation, I had to use either add another template argument to Dense (which will not be used), or add a default argument to CSR and ELL. Since CSR and ELL already expect to only use `int32` as index type, I chose that.
Additionally, it had to be `const`, because the default argument has to match how the solver functions will be called. Otherwise, there will be a mismatch between the instantiated functions and what is actually called, leading to linker errors.

pratikvn (approved):

core/matrix/batch_struct.hpp
MarcelKoch: Any more comments? If not, I will merge this today, or tomorrow morning.

yhmtsai (approved):
Sorry, I did not tend to hold this pr from my previous comments.
one other comments here. when getting number of reg in cuda, you take the max between worst-shared-memory and best-shared-memory case.
I do not see it in the previous release when we still have these optimation selection.
I assume it introduced some illegal configuration.
Could you elaborate it more?

dpcpp/solver/batch_bicgstab_launch.instantiate.dp.cpp
```diff
@@ -0,0 +1,112 @@
...
+                    mat.get_single_item_num_nnz(),
+                    static_cast<ValueType*>(slm_values.get_pointer()), item_ct1,
+                    workspace);
+            });
+    });
+}
+
+
+// begin
+GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE(GKO_DECLARE_BATCH_BICGSTAB_LAUNCH_0);
```
yhmtsai, line 83: It will not work with `GINKGO_DPCPP_SINGLE_MODE=ON`.
we use the instantiation to provide the specialization with unsupported exception on double precision.
with `GKO_BATCH_INSTANTION`, it will be wrong. Only the last one has the specialization, but the others will be instantiated.
`template macro {GKO_UNSUPPORTED;}` ->
```
template first_...;
template second_...;
...
template last  {GKO_UNSUPPORTED;}
```

yhmtsai (changes requested):
I think it will lead issue with single mode

dpcpp/solver/batch_bicgstab_launch.instantiate.dp.cpp
MarcelKoch: You're right. Thanks for bringing this up. I changed the order of macro application, so now it should be fixed.

cuda/solver/batch_bicgstab_launch.instantiate.cu
```diff
@@ -0,0 +1,90 @@
...
+    };
+    const int num_regs_used = std::max(
+        get_num_regs(
+            batch_single_kernels::apply_kernel<StopType, 9, true, PrecType,
+                                               LogType, BatchMatrixType,
+                                               ValueType>),
+        get_num_regs(
+            batch_single_kernels::apply_kernel<StopType, 0, false, PrecType,
+                                               LogType, BatchMatrixType,
+                                               ValueType>));
```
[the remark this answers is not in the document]
pratikvn: I think first one is everything in shared memory, second one is nothing in shared memory.

cuda/solver/batch_bicgstab_launch.instantiate.cu
```diff
@@ -0,0 +1,88 @@
...
+    int max_regs_blk = 0;
+    cudaDeviceGetAttribute(&max_regs_blk, cudaDevAttrMaxRegistersPerBlock,
+                           exec->get_device_id());
+    const int max_threads_regs =
+        ((max_regs_blk / static_cast<int>(num_regs_used)) / warp_sz) * warp_sz;
+    int max_threads = std::min(max_threads_regs, device_max_threads);
+    max_threads = max_threads <= max_bicgstab_threads ? max_threads
+                                                      : max_bicgstab_threads;
```
pratikvn, line 52: Just a comment and something for me to do in the future: I think this whole logic needs to be simplified. It seems it is now also possible to set the max number of registers similar to the launch_bounds with CUDA: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#maximum-number-of-registers-per-thread 

But of course, that means we maybe cannot unify HIP and CUDA anymore, but something we need to investigate.

pratikvn (approved):
I would wait for the CI to finish to merge this (maybe also the Intel SYCL pipelines), but looks good to me otherwise.

yhmtsai (approved):
LGTM. It is a bit hard to understand now though.

PR #1631, opened 2024-06-26
WIP read_distributed with pre-existing data

greole:
This PR adds a version of read_distributed that works on existing local data and non_local_data and comm_pattern.

### DIFF
--- a/core/distributed/matrix.cpp
+++ b/core/distributed/matrix.cpp
@@ -240,6 +240,40 @@ void Matrix<ValueType, LocalIndexType, GlobalIndexType>::move_to(
 
 template <typename ValueType, typename LocalIndexType, typename GlobalIndexType>
 void Matrix<ValueType, LocalIndexType, GlobalIndexType>::read_distributed(
+    const device_matrix_data<value_type, local_index_type>& local_data,
+    const device_matrix_data<value_type, local_index_type>& non_local_data,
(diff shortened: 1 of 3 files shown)


yhmtsai:
Is it more straight way in openFoam application? you can also do it with `create` function from existing linop and comm pattern.

greole:
I'll add an extra constructor/create function. The reasoning is that we basically already `local` and `non_local data `and the communication pattern already in place from OpenFOAM.

greole:
I just realised that the following create method should be sufficient:
```
template <typename ValueType, typename LocalIndexType, typename GlobalIndexType>
std::unique_ptr<Matrix<ValueType, LocalIndexType, GlobalIndexType>>
Matrix<ValueType, LocalIndexType, GlobalIndexType>::create(
    std::shared_ptr<const Executor> exec, mpi::communicator comm, dim<2> size,
    std::shared_ptr<LinOp> local_linop, std::shared_ptr<LinOp> non_local_linop,
    std::vector<comm_index_type> recv_sizes,
    std::vector<comm_index_type> recv_offsets,
    array<local_index_type> recv_gather_idxs)
```

upsj:
@greole Agreed in general, though we should be using the abstractions from #1588 and #1589 once they are available, instead of raw vectors/arrays

PR #1632, opened 2024-06-27
Fix mean computation for zero sized ranks

greole:
This PR makes sure to not  divide by zero when calculating the mean of a distributed vector. This otherwise might occur if some ranks hold a local vector of length zero after repartitioning a matrix or vector.

### DIFF
--- a/common/unified/matrix/dense_kernels.template.cpp
+++ b/common/unified/matrix/dense_kernels.template.cpp
@@ -261,7 +261,7 @@ void compute_mean(std::shared_ptr<const DefaultExecutor> exec,
             return x(i, j) * inv_total_size;
         },
         GKO_KERNEL_REDUCE_SUM(ValueType), result->get_values(), x->get_size(),
-        tmp, x, ValueType_nc{1.} / x->get_size()[0]);
+        tmp, x, ValueType_nc{1.} / std::max(1ul, x->get_size()[0]));
 }
 
 
--- a/common/unified/matrix/dense_kernels.template.cpp
+++ b/common/unified/matrix/dense_kernels.template.cpp
@@ -261,7 +261,7 @@ void compute_mean(std::shared_ptr<const DefaultExecutor> exec,
             return x(i, j) * inv_total_size;
         },
         GKO_KERNEL_REDUCE_SUM(ValueType), result->get_values(), x->get_size(),
-        tmp, x, ValueType_nc{1.} / std::max(1ul, x->get_size()[0]));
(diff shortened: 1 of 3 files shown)


common/unified/matrix/dense_kernels.template.cpp
```diff
@@ -263,7 +263,7 @@ void compute_mean(std::shared_ptr<const DefaultExecutor> exec,
             return x(i, j) * inv_total_size;
         },
         GKO_KERNEL_REDUCE_SUM(ValueType), result->get_values(), x->get_size(),
-        tmp, x, ValueType_nc{1.} / x->get_size()[0]);
+        tmp, x, ValueType_nc{1.} / std::max(1ul, x->get_size()[0]));
```
upsj, line 266: more future-proof:
```suggestion
        tmp, x, ValueType_nc{1.} / std::max<size_type>(1, x->get_size()[0]));
```

reference/matrix/dense_kernels.cpp
```diff
@@ -374,11 +374,15 @@ void compute_mean(std::shared_ptr<const ReferenceExecutor> exec,
                   const matrix::Dense<ValueType>* x,
                   matrix::Dense<ValueType>* result, array<char>&)
 {
+    GKO_ASSERT_EQ(result->get_size()[0], 1);
```
upsj, line 377: Why this assertion? The result of an empty mean can sensibly be defined as 0

reference/test/matrix/dense_kernels.cpp
```diff
@@ -695,6 +695,14 @@ TYPED_TEST(Dense, ComputesMean)
...
 }
 
+TYPED_TEST(Dense, ComputesMeanFailsOnZeroRowResults)
+{
+    using Mtx = typename TestFixture::Mtx;
+    using T = typename TestFixture::value_type;
+    auto result = Mtx::create(this->exec, gko::dim<2>{0, 1});
+
+    ASSERT_THROW(this->mtx4->compute_mean(result), gko::ValueMismatch);
+}
```
upsj, line 705: That would make this test unnecessary

upsj (changes requested):

reference/matrix/dense_kernels.cpp
MarcelKoch, line 377: I think we (mostly) don't throw in our backend libraries. You could just add this check to the core file.

MarcelKoch (changes requested):

MarcelKoch (approved):

upsj (approved):

PR #1639, opened 2024-07-02
Use Index map in PGM

MarcelKoch:
This PR enables using an index map to create the coarse level using PGM for distributed matrices. IMO this simplifies the PGM implementation, since there is no 'manual' implementation needed anymore.
This also has some consequences for the `distributed::Matrix` class. Namely, when creating a matrix with from local and non-local parts, an index map has to be supplied, instead of the list of vectors. Again, I think this should simplify the interface. Note that currently there is no accessor to the index map, since it's not necessary. But it might be useful for consistency.

I guess this is also an interface breaking change.

Todo:
- [x] backend implementation for `gather_as_global_index`
- [x] merge #1707

### DIFF
--- a/benchmark/test/reference/spmv_distributed.profile.stderr
+++ b/benchmark/test/reference/spmv_distributed.profile.stderr
@@ -52,6 +52,10 @@ DEBUG: begin components::fill_array
 DEBUG: end   components::fill_array
 DEBUG: begin components::fill_array
 DEBUG: end   components::fill_array
+DEBUG: begin components::fill_array
+DEBUG: end   components::fill_array
+DEBUG: begin components::fill_array
+DEBUG: end   components::fill_array
 DEBUG: begin copy(<typename>)
 DEBUG: begin copy
 DEBUG: end   copy
@@ -104,8 +108,6 @@ DEBUG: begin components::fill_array
 DEBUG: end   components::fill_array
 DEBUG: begin index_map::map_to_local
 DEBUG: end   index_map::map_to_local
-DEBUG: begin copy
-DEBUG: end   copy
 DEBUG: begin copy
 DEBUG: end   copy
 DEBUG: begin copy
--- a/core/distributed/matrix.cpp
+++ b/core/distributed/matrix.cpp
@@ -13,21 +13,10 @@
 #include <ginkgo/core/matrix/csr.hpp>
 #include <ginkgo/core/matrix/diagonal.hpp>
 
-#include "core/components/precision_conversion_kernels.hpp"
-#include "core/components/prefix_sum_kernels.hpp"
 #include "core/distributed/matrix_kernels.hpp"
 
 
 namespace gko {
-namespace conversion {
-namespace {
-
-
-GKO_REGISTER_OPERATION(convert, components::convert_precision);
-
-
-}  // anonymous namespace
-}  // namespace conversion
 namespace experimental {
 namespace distributed {
 namespace matrix {
--- a/core/distributed/matrix.cpp
+++ b/core/distributed/matrix.cpp
@@ -31,6 +31,62 @@ GKO_REGISTER_OPERATION(separate_local_nonlocal,
 }  // namespace matrix
 
 
+template <typename LocalIndexType, typename GlobalIndexType>
+void initialize_communication_pattern(
+    std::shared_ptr<const Executor> exec, mpi::communicator comm,
+    const index_map<LocalIndexType, GlobalIndexType>& imap,
+    std::vector<comm_index_type>& recv_sizes,
+    std::vector<comm_index_type>& recv_offsets,
+    std::vector<comm_index_type>& send_sizes,
+    std::vector<comm_index_type>& send_offsets,
+    array<LocalIndexType>& gather_idxs)
+{
+    // exchange step 1: determine recv_sizes, send_sizes, send_offsets
+    auto host_recv_targets =
+        make_temporary_clone(exec->get_master(), &imap.get_remote_target_ids());
+    auto host_offsets = make_temporary_clone(
+        exec->get_master(), &imap.get_remote_global_idxs().get_offsets());
+    auto compute_recv_sizes = [](const auto* recv_targets, size_type size,
+                                 const auto* offsets, auto& recv_sizes) {
+        for (size_type i = 0; i < size; ++i) {
+            recv_sizes[recv_targets[i]] = offsets[i + 1] - offsets[i];
+        }
+    };
+    std::fill(recv_sizes.begin(), recv_sizes.end(), 0);
+    compute_recv_sizes(host_recv_targets->get_const_data(),
+                       host_recv_targets->get_size(),
+                       host_offsets->get_const_data(), recv_sizes);
+    std::partial_sum(recv_sizes.begin(), recv_sizes.end(),
+                     recv_offsets.begin() + 1);
+    comm.all_to_all(exec, recv_sizes.data(), 1, send_sizes.data(), 1);
+    std::partial_sum(send_sizes.begin(), send_sizes.end(),
+                     send_offsets.begin() + 1);
+    send_offsets[0] = 0;
+    recv_offsets[0] = 0;
+
+    // exchange step 2: exchange gather_idxs from receivers to senders
+    auto recv_gather_idxs =
+        make_const_array_view(
+            imap.get_executor(), imap.get_non_local_size(),
+            imap.get_remote_local_idxs().get_const_flat_data())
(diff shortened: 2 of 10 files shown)


core/multigrid/pgm.cpp
```diff
@@ -270,20 +277,29 @@ void Pgm<ValueType, IndexType>::communicate(
         send_agg.get_size(), local_agg.get_const_data(),
         gather_idxs.get_const_data(), send_agg.get_data()));
 
+    // temporary index map that contains no remote connections to map
+    // local indices to global
+    experimental::distributed::index_map<IndexType, GlobalIndexType> imap(
+        exec, coarse_partition, comm.rank(), array<GlobalIndexType>{exec});
+    auto seng_global_agg = imap.map_to_global(
+        send_agg, experimental::distributed::index_space::local);
```
yhmtsai, line 285: If I guess it correctly, you map the local index back to the global index to avoid the collision. 
this is what I want to avoid because it needs twice of memory transferring between nodes.

MarcelKoch: I would guess that the actual volume of the data transfer is negligible compared to the latency cost. Also, this is significantly simpler to what was there before.

fritzgoebel (changes requested):
LGTM generally, minor changes

core/distributed/matrix.cpp
```diff
@@ -91,54 +158,28 @@ Matrix<ValueType, LocalIndexType, GlobalIndexType>::Matrix(
...
       send_offsets_(comm.size() + 1),
       send_sizes_(comm.size()),
       recv_offsets_(comm.size() + 1),
       recv_sizes_(comm.size()),
       gather_idxs_{exec},
-      non_local_to_global_{exec},
       one_scalar_{}
 {
-    this->set_size(size);
+    this->set_size({imap_.get_global_size(), imap_.get_global_size()});
```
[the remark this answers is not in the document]
MarcelKoch: Right now, this constructor only allows for square matrices. The index map defines the partition of columns, so for a rectangular matrix an additional partition for the rows would be necessary. Perhaps the index map would also need to be revised a bit.
But I think there should still be the option to get a rectangular matrix using the `read_distributed` functions.

include/ginkgo/core/multigrid/pgm.hpp
```diff
@@ -194,12 +194,28 @@ class Pgm : public EnableLinOp<Pgm<ValueType, IndexType>>,
...
+     *
+     * @tparam GlobalIndexType  Global index type
+     *
+     * @param matrix  a distributed matrix
+     * @param coarse_partition  the coarse partition to compute the new global
+     *                          indices
+     * @param local_agg  the local aggregate indices
+     *
+     * @return  the aggregates for non-local columns. The aggregated indices are
+     *          in the new global indexing
```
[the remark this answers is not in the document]
MarcelKoch: Yes, it's using the coarse index space. I will clarify.

fritzgoebel (approved):

test/mpi/matrix.cpp
```diff
@@ -312,6 +315,77 @@ TYPED_TEST(MatrixCreation, BuildFromExistingData)
...
+    mat->apply(x, y);
+
+    GKO_ASSERT_MTX_NEAR(gko::as<csr>(mat->get_local_matrix()), res_local[rank],
+                        0);
+    GKO_ASSERT_MTX_NEAR(gko::as<csr>(mat->get_non_local_matrix()),
+                        res_non_local[rank], 0);
+    GKO_ASSERT_MTX_NEAR(y->get_local_vector(), result[rank], 0);
+}
+
 #endif
```
yhmtsai, line 389:
```suggestion
}


#endif
```

yhmtsai (changes requested):
does the #1589 require this?
I think some of matrix function are deleted in another pr.

MarcelKoch:
@yhmtsai Yes #1589 needs this, because it removes some data members from the dist matrix class which the PGM accessed. Getting this data is not possible anymore when the dist matrix uses the row gatherer, since the data is hidden by the collective communicator from #1780. The issue is solved in the dist row gatherer pr, by letting the collective communicator handle the communication in the PGM.
I guess it should be possible in principle to also use the collective communicator in the PGM, without these changes, but I don't understand the current implementation, so I'm not able to work on that.

core/distributed/matrix.cpp
```diff
@@ -31,6 +42,62 @@ GKO_REGISTER_OPERATION(separate_local_nonlocal,
 }  // namespace matrix
 
 
+template <typename LocalIndexType, typename GlobalIndexType>
+void initialize_communication_pattern(
```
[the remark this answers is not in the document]
MarcelKoch: Not directly repeated, but reduced in complexity. But I will revisit the row gatherer PR to check if it can be removed completely.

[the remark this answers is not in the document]
MarcelKoch: I now removed it completely in #1589

core/multigrid/pgm.cpp
```diff
@@ -441,80 +379,79 @@ void Pgm<ValueType, IndexType>::generate()
...
+            // coarse global indices, so mapping it with the coarse index
+            // map results in the coarse non-local indices.
+            auto non_local_map = coarse_imap.map_to_local(
+                non_local_agg,
+                experimental::distributed::index_space::non_local);
+
             // build csr from row and col map
             // unlike non-distributed version, generate_coarse uses
-            // different row and col maps.
+            // differentrow and col maps.
```
pratikvn, line 427:
```suggestion
            // different row and col maps.
```

pratikvn (approved):
LGTM

core/distributed/matrix.cpp
```diff
@@ -7,16 +7,27 @@
...
 namespace gko {
+namespace conversion {
+namespace {
+
+
+GKO_REGISTER_OPERATION(convert, components::convert_precision);
+
+
+}  // anonymous namespace
+}  // namespace conversion
```
yhmtsai, line 30: unused

benchmark/test/reference/spmv_distributed.profile.stderr
```diff
@@ -52,6 +52,10 @@ DEBUG: begin components::fill_array
 DEBUG: end   components::fill_array
 DEBUG: begin components::fill_array
 DEBUG: end   components::fill_array
+DEBUG: begin components::fill_array
+DEBUG: end   components::fill_array
+DEBUG: begin components::fill_array
+DEBUG: end   components::fill_array
```
yhmtsai, line 58: where is the fill_array from?

core/multigrid/pgm.cpp
```diff
@@ -270,20 +279,30 @@ void Pgm<ValueType, IndexType>::communicate(
         send_agg.get_size(), local_agg.get_const_data(),
         gather_idxs.get_const_data(), send_agg.get_data()));
 
+    // There is no index map on the coarse level yet, so map the local indices
+    // to global manually
```
yhmtsai, line 283:
```suggestion
    // to global manually on the coarse partition
```

yhmtsai: Just to be sure. For example, 8 x 8 matrix with two partition and each contains 4. After aggregation, each aggregate to two nodes (i.e. 4 x 4 matrix as global coarse matrix)
first aggregation {0, 1} -> {0, 1}
second aggregation {0, 1} -> {2, 3}
i-th rank maps the local agg to global by local_idx + the coarse matrix offsets by rank (0 for first, 2 for the second in the example). Do I get the function meaning here correctly?

yhmtsai (approved):
LGTM. do you have an idea why the additional fill_array appear?

benchmark/test/reference/spmv_distributed.profile.stderr
MarcelKoch: Since this happens after the partition was created and before the device matrix data is created, I can only think of it being related to creating the index map in the matrix constructor. 
The index map contains two segmented arrays, each of which has an offsets array. This array always contains at least one element, so this is most likely from that.

core/multigrid/pgm.cpp
MarcelKoch: I don't really understand your question, so I will try to explain this more generally.
This operation is totally separate from the notion of aggregation. The only relevant part is that there are some local indices (which correspond to the local aggregation indices) and a partition of the global index space these local indices live in.
The local indices are mapped to the corresponding global indices. That's already all. The meaning of the indices, i.e. that they define the aggregation, doesn't really matter here.

PR #1641, opened 2024-07-04
Fix: transposed triangular solver uses the same algorithm as the original one

yhmtsai:
This PR changes the algorithm from transposed triangular solver from the default option to the source.
first commit adds the test which shows failure and then second commit changes the algorithm.


upsj (approved):
LGTM!

MarcelKoch (approved):

PR #1642, opened 2024-07-04
Fix inconsistent behavior of scalar Jacobi between reference and the others

yhmtsai:
We have the workaround for zero diagonal value in scalar Jacobi, but reference uses 1 but the others uses zero.
I currently make it be 1.
Note. it is not the proper inverse Jacobi matrix.

first commit adds the test and the second one fixes the issue.

### DIFF
--- a/reference/test/preconditioner/jacobi.cpp
+++ b/reference/test/preconditioner/jacobi.cpp
@@ -494,4 +494,24 @@ TYPED_TEST(Jacobi, ScalarJacobiGeneratesOnDifferentPrecision)
 }
 
 
+TYPED_TEST(Jacobi, ScalarJacobiHandleZero)
+{
+    using value_type = typename TestFixture::value_type;
+    using Vec = typename TestFixture::Vec;
+    using Bj = typename TestFixture::Bj;
+    auto mtx = gko::share(
+        gko::initialize<Vec>({{0, 0, 0}, {0, 2, 0}, {0, 0, 0}}, this->exec));
+    auto b = gko::initialize<Vec>({1, 2, 3}, this->exec);
+    auto x = Vec::create(this->exec, gko::dim<2>(3, 1));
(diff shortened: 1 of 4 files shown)


upsj:
@yhmtsai the second commit seems to be missing

yhmtsai:
@upsj Yes and sorry. I would like to wait for that some jobs of the first commit fail first

reference/test/preconditioner/jacobi.cpp
```diff
@@ -494,4 +494,24 @@ TYPED_TEST(Jacobi, ScalarJacobiGeneratesOnDifferentPrecision)
...
+TYPED_TEST(Jacobi, ScalarJacobiHandleZero)
+{
+    using value_type = typename TestFixture::value_type;
+    using Vec = typename TestFixture::Vec;
+    using Bj = typename TestFixture::Bj;
+    auto mtx = gko::share(
+        gko::initialize<Vec>({{0, 0, 0}, {0, 2, 0}, {0, 0, 0}}, this->exec));
+    auto b = gko::initialize<Vec>({1, 2, 3}, this->exec);
+    auto x = Vec::create(this->exec, gko::dim<2>(3, 1));
+    auto jacobi = this->scalar_j_factory->generate(mtx);
```
MarcelKoch, line 506: nit: I would say that this belongs to the ACT part, so it should be in the second block, and the apply would be in the third block.

MarcelKoch (approved):

reference/test/preconditioner/jacobi.cpp
yhmtsai: It's mixed actually because it was also used to check the apply operation result.
I will move the generation to the ACT part but not move the apply

yhmtsai: I am wrong. I mix it with another pr. The apply is already tested in other test. here only checks the generation

upsj (approved):

PR #1645, opened 2024-07-10
Fix communicator categorization

thoasm:
The default for `is_communicator_group` was `std::true_type` for some reason, which means that every type you pass in was considered a communicator.
This PR fixes it (so it defaults to `std::false_type`) and adds tests to prevent this from happening again.


MarcelKoch (approved):

upsj (approved):

yhmtsai (approved):
LGTM, but need to reformat the files

PR #1646, opened 2024-07-11
Add orthogonalization options to GMRES

nbeams:
Add orthogonalization choices for GMRES (CGS and CGS2 in addition to the MGS available before).

This adds a new unified kernel for performing the multidot operation for CGS/CGS2. It also changes the data layout of the Hessenberg matrix so that all the data that needs to be communicated after the multidot is stored contiguously, and can be handled with one communication.

TODO:
- [x] Benchmark check to make sure MGS performance is not hurt by the changes to the Hessenberg data layout.


nbeams:
Some initial benchmarks comparing old/new Hessenberg use in MGS for the affected kernels (`hessenberg_qr` and `solve_krylov`). These were run on Guyot, so A100 GPUs and 2 AMD EPYC 7742 CPUs. The `size` parameter is the size given to generate the stencil matrix, using the 7 pt stencil, with weak scaling for increased MPI ranks. The points with the same size have various MPI ranks (1, 2, 4, or 8) and `nrhs` (1, 10, or 100).

[CUDA - Hessenberg QR](https://github.com/user-attachments/files/16239822/cuda_hessenberg_qr_guyot.pdf)
[CUDA - Solve Krylov](https://github.com/user-attachments/files/16239823/cuda_sovle_krylov_guyot.pdf)
[OpenMP - Hessenberg QR](https://github.com/user-attachments/files/16239824/omp_hessenberg_qr_guyot.pdf)
[OpenMP - Solve Krylov](https://github.com/user-attachments/files/16239825/omp_solve_krylov_guyot.pdf)

To me, this seems to indicate that any difference between the old and new versions is basically just noise (especially as the absolute times for `solve_krylov` are so tiny). But I can run more tests if this isn't considered conclusive enough or looks odd to anybody.

core/solver/gmres.cpp
```diff
@@ -112,6 +142,155 @@ struct help_compute_norm {
...
+                       size_type num_rows, size_type num_rhs,
+                       size_type local_num_rows)
+{
+    for (size_type i = 0; i <= restart_iter; i++) {
+        // orthogonalize against krylov_bases(:, i):
+        // hessenberg(i, restart_iter) = next_krylov' * krylov_bases(:,
+        // i) next_krylov -= hessenberg(i, restart_iter) *
+        // krylov_bases(:, i)
+        auto hessenberg_entry = hessenberg_iter->create_submatrix(
+            span{0, 1}, span{i * num_rhs, (i + 1) * num_rhs});
```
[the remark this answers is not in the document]
MarcelKoch: To answer my own question, the `hessenber_iter` is a view on the `restart_iter` row of the full hessenberg matrix, so yes, using row `0` is correct.

core/solver/gmres_kernels.hpp
```diff
@@ -38,11 +38,19 @@ namespace gmres {
...
-#define GKO_DECLARE_ALL_AS_TEMPLATES             \
-    template <typename ValueType>                \
-    GKO_DECLARE_GMRES_RESTART_KERNEL(ValueType); \
-    template <typename ValueType>                \
-    GKO_DECLARE_GMRES_MULTI_AXPY_KERNEL(ValueType)
+#define GKO_DECLARE_GMRES_MULTI_DOT_KERNEL(_type)               \
+    void multi_dot(std::shared_ptr<const DefaultExecutor> exec, \
+                   const matrix::Dense<_type>* krylov_bases,    \
+                   const matrix::Dense<_type>* next_krylov,     \
+                   matrix::Dense<_type>* hessenberg_col)
```
MarcelKoch, line 45: nit: maybe `hessenberg_transposed_col` would be a better name, but I get that it might be too long

core/solver/gmres.cpp
```diff
@@ -312,36 +504,39 @@ void Gmres<ValueType>::apply_dense_impl(const VectorType* dense_b,
...
-        // hessenberg_iter = hessenberg(:, restart_iter);
-        auto hessenberg_iter = hessenberg->create_submatrix(
-            span{0, restart_iter + 2},
-            span{num_rhs * restart_iter, num_rhs * (restart_iter + 1)});
+        // Create view of current "column" in the hessenberg matrix:
+        // hessenberg_iter = hessenberg(:, restart_iter), which
+        // is actually stored as a row, hessenberg(restart_iter, :)
+        auto hessenberg_iter =
+            hessenberg->create_submatrix(span{restart_iter, restart_iter + 1},
+                                         span{0, num_rhs * (restart_iter + 2)});
```
MarcelKoch, line 512: If I understand the storage layout correctly, you could 'reshape' this row (aka column) into a `N_basis x N_rhs` matrix, right? Then some of the changes related to the new indexing might not be necessary. But I've not checked it fully to see how much it would actually help-

common/unified/solver/common_gmres_kernels.cpp
```diff
@@ -69,28 +69,30 @@ void hessenberg_qr(std::shared_ptr<const DefaultExecutor> exec,
...
                 return;
             }
             // increment iteration count
             final_iter_nums[rhs]++;
-            auto hess_this = hessenberg_iter(0, rhs);
-            auto hess_next = hessenberg_iter(1, rhs);
+            auto hess_this =
+                hessenberg_iter(0, rhs);  // hessenberg_iter(0, rhs);
+            auto hess_next =
+                hessenberg_iter(0, num_rhs + rhs);  // hessenberg_iter(1, rhs);
```
MarcelKoch, line 83: these are the kinds of index changes I was referring to earlier.

MarcelKoch (changes requested):
LGTM. I've left mostly minor comments, except for the shape of `hessenberg_iter`.

core/solver/gmres.cpp
nbeams: Do you mean by directly creating a `Dense` of shape `[restart_iter + 2, num_rhs]` with a view to the correctly offset starting point in `hessenberg`'s values for the data? That would mean the `hessenberg_qr` kernel could revert back to its original version, but `solve_krylov` would still need to be altered since it considers the whole Hessenberg matrix. 

I think it depends on whether we think it's more or less clear to "mix" the two in this way -- meaning, if the full Hessenberg matrix is viewed in its new layout, while the "for-this-iteration" slice of it is viewed in the "logical" way, with rows corresponding to iterations and columns to the rhs vector. Viewing `hessenberg_iter` in the "logical" way would probably make routines only involving `hessenberg_iter` more clear, but then is it more confusing overall? I could see the argument either way.

core/solver/gmres_kernels.hpp
nbeams: What about `hess_trans_col` or something like that? Unless we change the layout of `hessenberg_iter`, in which case `hessenberg_col` would be correct after all...

core/solver/gmres.cpp
MarcelKoch: I think it would be best to have both views available. They are free anyway, and I think if we name them properly they should be a net benefit.
The most natural way would probably be to just create the view on-the-fly, but this doesn't work with our separation of our core and device libraries. So we need to create the views beforehand.

core/solver/gmres_kernels.hpp
MarcelKoch: I guess this just up to personal preference. I would be ok with that.

core/solver/gmres.cpp
nbeams: Why would we need both views? What would we do with each?

MarcelKoch: Oh, I guess I misunderstood your comment. So again, the question is, which view is more useful for kernels, but I think you can judge that better. IMO, using indexing like `hess(i, j)` seems easier to me than `hess(0, i * N + j)` (or similar).

core/solver/gmres_kernels.hpp
```diff
@@ -38,11 +38,19 @@ namespace gmres {
...
+
```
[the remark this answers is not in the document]
nbeams: This was fixed in 1fb37be5d99ef0db3d861d3c100246eda8829a0b

core/solver/gmres.cpp
```diff
@@ -112,6 +142,155 @@ struct help_compute_norm {
...
+                       VectorType* krylov_bases, VectorType* next_krylov,
+                       array<char>& reduction_tmp, size_type restart_iter,
+                       size_type num_rows, size_type num_rhs,
+                       size_type local_num_rows)
+{
+    for (size_type i = 0; i <= restart_iter; i++) {
+        // orthogonalize against krylov_bases(:, i):
+        // hessenberg(i, restart_iter) = next_krylov' * krylov_bases(:,
+        // i) next_krylov -= hessenberg(i, restart_iter) *
+        // krylov_bases(:, i)
```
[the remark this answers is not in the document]
nbeams: Also fixed in 1fb37be5d99ef0db3d861d3c100246eda8829a0b

core/solver/gmres.cpp
nbeams: In 35e965c065fc190dcef09278a62741aca8afad4e I changed the `hessenberg_iter` layout.

core/solver/gmres.cpp
```diff
@@ -112,6 +142,155 @@ struct help_compute_norm {
...
+void finish_reduce(matrix::Dense<ValueType>* hessenberg_iter,
+                   experimental::distributed::Vector<ValueType>* next_krylov,
+                   const size_type num_rhs, const size_type restart_iter)
+{
+    auto exec = hessenberg_iter->get_executor();
+    const auto comm = next_krylov->get_communicator();
+    exec->synchronize();
+    // hessenberg_iter is the size of all non-zeros for this iteration -- but we
+    // are not setting the last values for each rhs (values that would be below
+    // the diagonal in the "full" matrix.
```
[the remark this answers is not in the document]
nbeams: I expanded the comment to try to make it more clear, let me know what you think.

core/solver/gmres_kernels.hpp
nbeams: Since `multi_dot` works with `hessenberg_iter` which is now viewed as columns again, I left the name for now.

MarcelKoch (approved):

test/solver/gmres_kernels.cpp
```diff
@@ -294,18 +339,27 @@ TEST_F(Gmres, GmresApplyOneRHSIsEquivalentToRef)
...
+            base_params.with_orthog_method(orthog).on(exec)->generate(d_mtx);
+        auto b = gen_mtx(m, n);
+        auto x = gen_mtx(m, n);
+        auto d_b = gko::clone(exec, b);
+        auto d_x = gko::clone(exec, x);
+
+        ref_solver->apply(b, x);
+        exec_solver->apply(d_b, d_x);
+
+        GKO_ASSERT_MTX_NEAR(d_b, b, 0);
```
pratikvn, line 362: not needed
```suggestion
```

test/solver/gmres_kernels.cpp
```diff
@@ -294,18 +339,27 @@ TEST_F(Gmres, GmresApplyOneRHSIsEquivalentToRef)
...
-    GKO_ASSERT_MTX_NEAR(d_x, x, r<value_type>::value * 1e3);
+    for (auto orthog :
+         {orthog_method::mgs, orthog_method::cgs, orthog_method::cgs2}) {
+        SCOPED_TRACE(orthog);
+        int m = 123;
+        int n = 5;
+        auto ref_solver =
+            base_params.with_orthog_method(orthog).on(ref)->generate(mtx);
+        auto exec_solver =
+            base_params.with_orthog_method(orthog).on(exec)->generate(d_mtx);
```
pratikvn, line 353: Do you know if the default restart parameters are the ones being used ? I think it would be nice to test for problems that need higher restart parameters as well.

pratikvn (approved):
LGTM! Some minor nits. Thanks for adding the performance comparisons.

core/solver/gmres.cpp
```diff
@@ -112,6 +142,156 @@ struct help_compute_norm {
     }
 };
 
+namespace {
```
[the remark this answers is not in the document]
nbeams: Not sure if it's technically necessary. I thought perhaps it would be preferred to keep the helper functions contained to this file; I noticed it was used for some helper functions in multigrid. But, now, as I look at more files, it seems multigrid is an outlier? I knew usually an anonymous namespace is used for registering the kernels, but in multigrid it's not, and then it's used for some other functions...are there some other rules for use of anonymous namespace in Ginkgo?

Anyway, I will remove it from the gmres helper functions.

include/ginkgo/core/solver/gmres.hpp
```diff
@@ -31,6 +31,29 @@ namespace solver {
 
 constexpr size_type gmres_default_krylov_dim = 100u;
 
+namespace gmres {
+/**
+ * Set the orthogonalization method for the Krylov subspace.
+ */
+enum class orthog_method {
```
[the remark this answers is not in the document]
nbeams: Ha, I went with `orthog` because `ortho` just makes me think of teeth, rather than orthogonalization :) 
Should we have a poll? Maybe someone has another better option than either of these?

test/solver/gmres_kernels.cpp
nbeams: I'm not sure I understand. For this test, I put the existing test "guts" inside a loop over each of the different orthogonalization methods -- if it was included before, why not now?

test/solver/gmres_kernels.cpp
nbeams: Yeah, the default krylov_dim should be used from what I understand. The default is pretty high IMO (100) compared to the sizes that are used in the tests.

test/solver/gmres_kernels.cpp
```diff
@@ -272,6 +273,50 @@ TEST_F(Gmres, GmresKernelMultiAxpyIsEquivalentToRef)
...
+    gko::kernels::GKO_DEVICE_NAMESPACE::gmres::multi_dot(
+        exec, d_krylov_basis.get(), d_next_krylov.get(),
+        d_hessenberg_iter.get());
+
+    // The multidot computation does not set the value below the diagonal
+    // in the Hessenberg matrix column(s), as that is done after the
+    // orthogonalization of the next basis vector. In this test, we
+    // are checking the column(s) created on the last iteration before the
+    // solver's restart would be triggered, so it is only the final row of
+    // the Hessenberg column(s) that we ignore.
```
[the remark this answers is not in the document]
nbeams: Yeah, good point. Originally, I thought it would be better to have this specifically marked/explained for future reference, but you're right, it's just unnecessarily complicated. I'll take it out.

include/ginkgo/core/solver/gmres.hpp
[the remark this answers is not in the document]
pratikvn: :) I dont have a strong preference. Both are okay with me.

test/solver/gmres_kernels.cpp
pratikvn: IMO the test should also test what might be changed by the function. `apply` is guaranteed to not modify `b/d_b`, as they are maked `const`. So there is no need to test that they are same.

test/solver/gmres_kernels.cpp
pratikvn: The matrix used here is a dense random matrix of size (123x123) with values distributed from a normal distribution with mean -1.0 and stddev 1. I am not sure the edge cases of the different orthogonalization methods are being properly tested ?

test/solver/gmres_kernels.cpp
nbeams: I agree it seems odd, but it's also done in the `cgs`, `gcr`, and `idr` tests (in test/solver), so I'm hesitant to change in this PR.

test/solver/gmres_kernels.cpp
nbeams: By "higher restart parameters" do you mean "needs to go all the way up to 100 (or higher if we set the dim higher)" or "needs to restart at least once" (which could be achieved with a small problem by dropping `krylov_dim`)?

pratikvn: I think it needs to restart atleast once.

nbeams: I put a temporary convergence logger in the test and saw that it was actually not converging -- maxes out at 246 iterations (set as `max_iters`). I don't know if that's the intended behavior for this test, but the end result is that it should be restarting twice as-is. I output the matrix and rhs vectors that are generated and did some playing around in scipy. The matrix is fairly non-normal (Frobenius norm of A^TA - AA^T ~3452) and the GMRES convergence stagnates pretty severely after restart. Even with Krylov dim = 122 it stagnates around 10e-2 (in the scipy gmres); only going up to full size for the Krylov dim (123) gets that last "drop" (and it does, as it should).

pratikvn: I see. Thanks for checking that. From the numbers you state, that should be sufficient then.

test/solver/gmres_kernels.cpp
pratikvn: Yeah, I would say it is not really necessary in those either. So, I would say you can remove them in all those tests as well.

nbeams: Should that be in this PR, though? For all the tests that have it, it seems to go all the way back to the original addition of the tests -- except GMRES, where it was added in 2022? (https://github.com/ginkgo-project/ginkgo/commit/963b1a3b152c797da61b76cbbc16e89d8f3ad3c9#diff-79b1284e17363ce1e01a13254f43119d1a20c7973d1477efe43c79c3938711fc). I think we should keep this PR limited to GMRES as far as code that it touches.

pratikvn: Then, I guess we can tackle it in a future PR, if necessary.

PR #1647, opened 2024-07-12
Unify factorization kernels

upsj:
This unifies the ParILUT/ICT and ILU(0)/IC(0) kernels and the unimplemented index_set kernels.


pratikvn (approved):
LGTM!

PR #1649, opened 2024-07-14
Unify Csr

upsj:
This unifies the `csr_kernels` files and removes all switches we had for older CUDA 10.x versions.


cuda/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.18 FATAL_ERROR)
 add_library(ginkgo_cuda $<TARGET_OBJECTS:ginkgo_cuda_device> "")
 include(${PROJECT_SOURCE_DIR}/cmake/template_instantiation.cmake)
-add_instantiation_files(. matrix/csr_kernels.instantiate.cu CSR_INSTANTIATE)
+add_instantiation_files(${PROJECT_SOURCE_DIR}/common/cuda_hip matrix/csr_kernels.instantiate.cpp CSR_INSTANTIATE)
```
MarcelKoch, line 4: wouldn't this lead to conflicts if both cuda and hip are enabled?

MarcelKoch (approved):

cuda/CMakeLists.txt
MarcelKoch: I think the issue (if it exists) could be prevented by instantiating the files in the `common/cuda_hip/CMakeLists.txt`

upsj: No, those are being instantiated in different folders. The reason we need to do it here is related to file properties being scoped to directories, so if we do it in cuda_hip, the dependency between the auto-generation of the file and the library being compiled from it is broken.

MarcelKoch: You're right, they are instantiated in the current binary directory, I was just looking at the wrong place.

common/cuda_hip/matrix/csr_kernels.template.cpp
```diff
@@ -2,6 +2,88 @@
...
+using compiled_kernels = syn::value_list<int, 3, 4, 6, 7, 8, 12, 14>;
+
+using classical_kernels =
+    syn::value_list<int, config::warp_size, 32, 16, 8, 4, 2, 1>;
+
+using spgeam_kernels =
+    syn::value_list<int, 1, 2, 4, 8, 16, 32, config::warp_size>;
+
+
+#include "common/cuda_hip/matrix/csr_common.hpp.inc"
```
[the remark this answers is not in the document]
upsj: no, `.hpp` should only be used for self-sufficient files (with rare exceptions), while `.hpp.inc` should be used for code that is meant to go inside a namespace

PR #1650, opened 2024-07-16
Additive read distributed

fritzgoebel:
This PR adds an option to communicate overlap in the distributed matrix' `read_distributed`. If the option is used, nonzero entries present in multiple ranks are added up on the owning rank rather than thrown away in `read_distributed`.

This can be useful e.g. if in a domain decomposed finite element setting, each rank assembles their local contribution to a global system matrix and when assembling the global system matrix information on the subdomain boundaries has to be exchanged. 

TODO:
- [x] Implement reference kernel tests
- [x] Implement device kernels
- [x] Implement device kernel tests


core/distributed/matrix.cpp
```diff
@@ -252,14 +258,105 @@ void Matrix<ValueType, LocalIndexType, GlobalIndexType>::read_distributed(
...
+        array<global_index_type> original_positions{exec, num_entries};
+        overlap_count.fill(0);
+        auto tmp_part = make_temporary_clone(exec, row_partition);
+        exec->run(matrix::make_count_overlap_entries(
+            data, tmp_part.get(), local_part, overlap_count, overlap_positions,
+            original_positions));
+
+        overlap_count.set_executor(exec->get_master());
+        std::vector<comm_index_type> overlap_send_sizes(
+            overlap_count.get_data(), overlap_count.get_data() + num_parts);
```
MarcelKoch, line 284: you could just use `overlap_count`, right?

core/distributed/matrix.cpp
```diff
@@ -252,14 +258,105 @@ void Matrix<ValueType, LocalIndexType, GlobalIndexType>::read_distributed(
...
+        if (use_host_buffer) {
+            overlap_send_row_idxs.set_executor(exec->get_master());
+            overlap_send_col_idxs.set_executor(exec->get_master());
+            overlap_send_values.set_executor(exec->get_master());
+            overlap_recv_row_idxs.set_executor(exec->get_master());
+            overlap_recv_col_idxs.set_executor(exec->get_master());
+            overlap_recv_values.set_executor(exec->get_master());
+        }
+        comm.all_to_all_v(
+            use_host_buffer ? exec : exec->get_master(),
```
MarcelKoch, line 320: we might want to think about merging these three calls. Basically that would mean transform the device_matrix_data back to an Array-of-Structs and then define a custom MPI datatype. After the communication, the received data needs to be unpacked again into a SoA.

core/distributed/matrix.cpp
```diff
@@ -252,14 +258,105 @@ void Matrix<ValueType, LocalIndexType, GlobalIndexType>::read_distributed(
...
+            overlap_recv_row_idxs.set_executor(exec);
+            overlap_recv_col_idxs.set_executor(exec);
+            overlap_recv_values.set_executor(exec);
+        }
+
+        array<global_index_type> all_row_idxs{exec, num_entries + n_recv};
+        array<global_index_type> all_col_idxs{exec, num_entries + n_recv};
+        array<value_type> all_values{exec, num_entries + n_recv};
+        exec->copy_from(exec, num_entries, data.get_const_row_idxs(),
+                        all_row_idxs.get_data());
```
MarcelKoch, line 344: For another time, but I think this would be clearer if we could do:
```
all_row_idxs.subview(0, num_entries) = data.row_idxs;
all_row_idxs.subview(num_entries, all_size) = overlap_recv_row_idxs;
```
But this requires to expose array acces in `device_matrix_data`.

reference/test/distributed/matrix_kernels.cpp
```diff
@@ -186,6 +186,101 @@ TYPED_TEST_SUITE(Matrix, gko::test::ValueLocalGlobalIndexTypes,
...
+
+    ca overlap_count{this->ref, static_cast<gko::size_type>(num_parts)};
+    ga overlap_positions{this->ref, input.get_num_stored_elements()};
+    ga original_positions{this->ref, input.get_num_stored_elements()};
+    for (gko::size_type i = 0; i < num_parts; i++) {
+        overlap_count.fill(0);
+        gko::kernels::reference::distributed_matrix::count_overlap_entries(
+            this->ref, input, partition.get(), i, overlap_count,
+            overlap_positions, original_positions);
+        GKO_ASSERT_ARRAY_EQ(overlap_count, overlap_count_ref[i]);
```
MarcelKoch, line 223: nit
```suggestion

        GKO_ASSERT_ARRAY_EQ(overlap_count, overlap_count_ref[i]);
```

reference/test/distributed/matrix_kernels.cpp
```diff
@@ -186,6 +186,101 @@ TYPED_TEST_SUITE(Matrix, gko::test::ValueLocalGlobalIndexTypes,
                  TupleTypenameNameGenerator);
 
 
+TYPED_TEST(Matrix, CountOverlapEntries)
+{
+    using lit = typename TestFixture::local_index_type;
+    using git = typename TestFixture::global_index_type;
+    using vt = typename TestFixture::value_type;
```
MarcelKoch, line 193:
```suggestion
```

reference/test/distributed/matrix_kernels.cpp
```diff
@@ -186,6 +186,101 @@ TYPED_TEST_SUITE(Matrix, gko::test::ValueLocalGlobalIndexTypes,
...
+    }
+}
+
+
+TYPED_TEST(Matrix, FillOverlapSendBuffers)
+{
+    using lit = typename TestFixture::local_index_type;
+    using git = typename TestFixture::global_index_type;
+    using vt = typename TestFixture::value_type;
+    using ca = gko::array<comm_index_type>;
```
MarcelKoch, line 235:
```suggestion
```

reference/test/distributed/matrix_kernels.cpp
```diff
@@ -186,6 +186,101 @@ TYPED_TEST_SUITE(Matrix, gko::test::ValueLocalGlobalIndexTypes,
...
+    for (gko::size_type i = 0; i < num_parts; i++) {
+        auto num_entries = overlap_row_idxs_ref[i].get_size();
+        overlap_row_idxs.resize_and_reset(num_entries);
+        overlap_col_idxs.resize_and_reset(num_entries);
+        overlap_values.resize_and_reset(num_entries);
+        gko::kernels::reference::distributed_matrix::fill_overlap_send_buffers(
+            this->ref, input, partition.get(), i, overlap_positions[i],
+            original_positions[i], overlap_row_idxs, overlap_col_idxs,
+            overlap_values);
+        GKO_ASSERT_ARRAY_EQ(overlap_row_idxs, overlap_row_idxs_ref[i]);
```
MarcelKoch, line 277:
```suggestion

        GKO_ASSERT_ARRAY_EQ(overlap_row_idxs, overlap_row_idxs_ref[i]);
```

reference/test/distributed/matrix_kernels.cpp
```diff
@@ -186,6 +186,101 @@ TYPED_TEST_SUITE(Matrix, gko::test::ValueLocalGlobalIndexTypes,
...
+
+    gko::array<git> overlap_row_idxs{this->ref};
+    gko::array<git> overlap_col_idxs{this->ref};
+    gko::array<vt> overlap_values{this->ref};
+    for (gko::size_type i = 0; i < num_parts; i++) {
+        auto num_entries = overlap_row_idxs_ref[i].get_size();
+        overlap_row_idxs.resize_and_reset(num_entries);
+        overlap_col_idxs.resize_and_reset(num_entries);
+        overlap_values.resize_and_reset(num_entries);
+        gko::kernels::reference::distributed_matrix::fill_overlap_send_buffers(
```
MarcelKoch, line 273:
```suggestion

        gko::kernels::reference::distributed_matrix::fill_overlap_send_buffers(
```

MarcelKoch (changes requested):
Mostly smaller stuff. Only the communication change would be significant, if that would be done.

core/distributed/matrix.cpp
```diff
@@ -252,14 +258,105 @@ void Matrix<ValueType, LocalIndexType, GlobalIndexType>::read_distributed(
...
+            overlap_send_values.get_const_data(), overlap_send_sizes.data(),
+            overlap_send_offsets.data(), overlap_recv_values.get_data(),
+            overlap_recv_sizes.data(), overlap_recv_offsets.data());
+        if (use_host_buffer) {
+            overlap_recv_row_idxs.set_executor(exec);
+            overlap_recv_col_idxs.set_executor(exec);
+            overlap_recv_values.set_executor(exec);
+        }
+
+        array<global_index_type> all_row_idxs{exec, num_entries + n_recv};
```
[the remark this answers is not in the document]
fritzgoebel: Yes, but they get ignored when the `local` and `non_local` matrices are created, just as before.

core/distributed/matrix.cpp
```diff
@@ -252,14 +258,105 @@ void Matrix<ValueType, LocalIndexType, GlobalIndexType>::read_distributed(
...
+        overlap_recv_offsets[0] = 0;
+
+        size_type n_send = overlap_send_offsets.back();
+        size_type n_recv = overlap_recv_offsets.back();
+        array<global_index_type> overlap_send_row_idxs{exec, n_send};
+        array<global_index_type> overlap_send_col_idxs{exec, n_send};
+        array<value_type> overlap_send_values{exec, n_send};
+        array<global_index_type> overlap_recv_row_idxs{exec, n_recv};
+        array<global_index_type> overlap_recv_col_idxs{exec, n_recv};
+        array<value_type> overlap_recv_values{exec, n_recv};
```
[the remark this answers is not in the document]
fritzgoebel: I think this would mean that we would have to add to add a `set_executor` to `device_matrix_data` for the `use_host_buffer` check, and having the separate arrays is consistent with the arrays for `local_row_idxs`, `local)col_idxs` and so on just below. I would prefer leaving these arrays.

MarcelKoch (approved):

yhmtsai (approved):
LGTM.
Because the kernel supports out-of-order, that will be great if you make the testing case out-of-order in test/mpi/matrix.cpp

core/distributed/matrix.cpp
```diff
@@ -252,14 +259,103 @@ void Matrix<ValueType, LocalIndexType, GlobalIndexType>::read_distributed(
...
+        }
+        comm.all_to_all_v(use_host_buffer ? exec : exec->get_master(),
+                          send_row_idxs.get_const_data(), send_sizes.get_data(),
+                          send_offsets.get_data(), recv_row_idxs.get_data(),
+                          recv_sizes.get_data(), recv_offsets.get_data());
+        comm.all_to_all_v(use_host_buffer ? exec : exec->get_master(),
+                          send_col_idxs.get_const_data(), send_sizes.get_data(),
+                          send_offsets.get_data(), recv_col_idxs.get_data(),
+                          recv_sizes.get_data(), recv_offsets.get_data());
+        comm.all_to_all_v(use_host_buffer ? exec : exec->get_master(),
```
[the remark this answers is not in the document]
MarcelKoch: I think the non-blocking versions are a good choice here. The latency cost could be hidden a bit by that, so I would prefer it.

test/mpi/matrix.cpp
```diff
@@ -60,9 +60,16 @@ class MatrixCreation : public CommonMpiTestFixture {
                      {3, 4, 7},
                      {4, 0, 9},
                      {4, 4, 10}}},
-          dist_input{{{size, {{0, 1, 1}, {0, 3, 2}, {1, 1, 3}, {1, 2, 4}}},
-                      {size, {{2, 1, 5}, {2, 2, 6}, {3, 3, 8}, {3, 4, 7}}},
-                      {size, {{4, 0, 9}, {4, 4, 10}}}}},
+          dist_input{
+              {{size,
+                {{0, 1, 1},
```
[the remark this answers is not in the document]
fritzgoebel: The `local_only` (original) `read_distributed` does not work for unordered data, so I don't think I would change this here.

reference/distributed/matrix_kernels.cpp
```diff
@@ -16,6 +19,88 @@ namespace reference {
...
+        row_part_ids_per_entry.get_data()[i] = row_part_id;
+        if (row_part_id != local_part) {
+            send_count.get_data()[row_part_id]++;
+            original_positions.get_data()[i] = i;
+        } else {
+            original_positions.get_data()[i] = -1;
+        }
+    }
+
+    auto comp = [row_part_ids_per_entry, local_part](auto i, auto j) {
```
MarcelKoch, line 51: To prevent a copy:
```suggestion
    auto comp = [&row_part_ids_per_entry, local_part](auto i, auto j) {
```

omp/distributed/matrix_kernels.cpp
```diff
@@ -20,6 +22,92 @@ namespace omp {
...
+        if (row_part_id != local_part) {
+#pragma omp atomic
+            send_count.get_data()[row_part_id]++;
+            original_positions.get_data()[i] = i;
+        } else {
+            original_positions.get_data()[i] = -1;
+        }
+    }
+
+    auto comp = [row_part_ids_per_entry, local_part](auto i, auto j) {
```
MarcelKoch, line 56: same here.

MarcelKoch (changes requested):

MarcelKoch (approved):

upsj (commented):
Not trying to hold this up, only a question because I made the incorrect assumption that this would be heavily ingrained with the rest of the assembly: Does it need to be an extension of the `read_distributed` function? It looks like it could also be a separate preprocessing function.

yhmtsai:
I had the similar thought with @upsj about putting these preprocessing function to `device_matrix_data`.
However, I am conviced by myself because it requires the communication pattern, which is only tied to matrix now.

upsj:
@yhmtsai At the time read_distributed is being called, the matrix doesn't have any distribution information yet, so I was thinking more of a free function like `device_matrix_data distributed::assemble(device_matrix_data, communicator, row_partition, col_partition)` (good name pending 😅)

fritzgoebel:
I'm sure it could also be a separate preprocessing function. What would the advantage of that be?

include/ginkgo/core/distributed/assembly_helpers.hpp
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * @param comm the communicator used to assemble the global matrix.
+ * @param input the device_matrix_data structure.
+ * @param partition the partition used to determine owndership.
+ *
+ * @return the globally assembled device_matrix_data structure for this MPI
+ * rank.
+ */
+template <typename ValueType, typename LocalIndexType, typename GlobalIndexType>
+device_matrix_data<ValueType, GlobalIndexType> assemble(
```
MarcelKoch, line 44: The name is not ideal. Maybe something like `add_non_local_entries`, `add_non_local_contributions`?

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -313,6 +348,21 @@ class Matrix
...
+    /**
+     * Reads a square matrix from the matrix_data structure and a global
+     * partition.
+     *
+     * @see read_distributed
+     *
+     * @note For efficiency it is advised to use the device_matrix_data
+     * overload.
+     */
+    void read_distributed(
```
MarcelKoch, line 360: For consistency you should also add the overloads with two partitions.

common/cuda_hip/distributed/assembly_helpers_kernels.cpp
```diff
@@ -0,0 +1,137 @@
...
+        },
+        num_parts, local_part, row_part_ptrs.get_data(), send_count.get_data());
+}
+
+GKO_INSTANTIATE_FOR_EACH_VALUE_AND_LOCAL_GLOBAL_INDEX_TYPE(
+    GKO_DECLARE_COUNT_NON_OWNING_ENTRIES);
+
+
+template <typename ValueType, typename LocalIndexType, typename GlobalIndexType>
+void fill_send_buffers(
```
MarcelKoch, line 98: I think you could also use the unified kernels for this one.

include/ginkgo/core/distributed/assembly_helpers.hpp
MarcelKoch, line 44: I would also prefer a test for this function. Right now, only the individual kernels, and the indirect call for the matrix read distributed.

MarcelKoch (changes requested):

include/ginkgo/core/distributed/assembly_helpers.hpp
```diff
@@ -0,0 +1,56 @@
...
+    mpi::communicator comm,
+    const device_matrix_data<ValueType, GlobalIndexType>& input,
+    ptr_param<const Partition<LocalIndexType, GlobalIndexType>> partition);
```
MarcelKoch, line 47: Maybe rename it to 
```suggestion
    ptr_param<const Partition<LocalIndexType, GlobalIndexType>> row_partition);
```
since only the ownership of the rows is relevant here. Alternatively, mention it in the doc.

upsj: If this is meant to be used in conjunction with a row-partitioned matrix, we should probably include that in the name. There are three types of distributions we can have (symmetric partitioning assumed): row-partitioned, column-partitioned and overlapping. This is probably for row-partitioning, we would need a slightly different routine for column-partitioned operations, and overlapping works without explicit assembly.

MarcelKoch: Just for my understanding, a row- and column partitioned matrix would mean that the matrix is decomposed into blocks, and each block is owned by a different process, right? In that case, overlapping might not be the right word to describe it.

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -299,6 +311,29 @@ class Matrix
...
+     * @param partition  The global row and column partition.
+     * @param assembly_mode  The mode of assembly.
+     *
+     * @return the index_map induced by the partitions and the matrix structure
+     */
+    void read_distributed(
+        const device_matrix_data<value_type, global_index_type>& data,
+        std::shared_ptr<const Partition<local_index_type, global_index_type>>
+            partition,
+        assembly_mode assembly_type);
```
yhmtsai, line 335:
```suggestion
    void read_distributed(
        const device_matrix_data<value_type, global_index_type>& data,
        std::shared_ptr<const Partition<local_index_type, global_index_type>>
            partition,
        assembly_mode assembly_type = assmbly_mode::communicate);
```
you do not need to create double interface, but you will need to move `read_distributed(device_matrix_data, row_partition, col_partition)` out of public interface with new name or add explicit interface like you did but only for that. It is to avoid the ambiguous function call. all the other function will end up with this call

include/ginkgo/core/distributed/assembly_helpers.hpp
```diff
@@ -0,0 +1,57 @@
...
+ * Assembles device_matrix_data entries owned by this MPI rank from other ranks
+ * and communicates entries located on this MPI rank owned by other ranks to
+ * their respective owners. This can be useful e.g. in a finite element code
+ * where each rank assembles a local contribution to a global system matrix and
+ * the global matrix has to be assembled by summing up the local contributions
+ * on rank boundaries. The partition used is only relevant for row ownership.
+ *
+ * @param comm the communicator used to assemble the global matrix.
+ * @param input the device_matrix_data structure.
+ * @param partition the partition used to determine row owndership.
```
yhmtsai, line 39:
```suggestion
 * @param comm  the communicator used to assemble the global matrix.
 * @param input  the device_matrix_data structure.
 * @param partition  the partition used to determine row owndership.
```
I am not sure whether it is doxygen format, but we use  two space between key and description `@param <key>  <description>`

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -299,6 +311,29 @@ class Matrix
...
+    /**
+     * Reads a square matrix from the device_matrix_data structure and a global
+     * partition.
+     *
+     * The global size of the final matrix is inferred from the size of the
+     * partition. Both the number of rows and columns of the device_matrix_data
+     * are ignored.
+     *
+     * @note The matrix data can contain entries for rows other than those owned
+     *        by the process. Entries for those rows are discarded.
```
yhmtsai, line 323:
```suggestion
     * @note The matrix data can contain entries for rows other than those owned
     *       by the process. Entries for those rows are discarded.
```

include/ginkgo/core/distributed/matrix.hpp
```diff
@@ -337,6 +387,32 @@ class Matrix
...
+    /**
+     * Reads a matrix from the device_matrix_data structure, a global row
+     * partition, and a global column partition.
+     *
+     * The global size of the final matrix is inferred from the size of the row
+     * partition and the size of the column partition. Both the number of rows
+     * and columns of the device_matrix_data are ignored.
+     *
+     * @note The matrix data can contain entries for rows other than those owned
+     *        by the process. Entries for those rows are discarded.
```
yhmtsai, line 399:
```suggestion
     * @note The matrix data can contain entries for rows other than those owned
     *       by the process. Entries for those rows are discarded.
```

reference/test/distributed/assembly_helpers_kernels.cpp
```diff
@@ -0,0 +1,153 @@
...
+#include <gtest/gtest.h>
+
+#include <ginkgo/core/base/device_matrix_data.hpp>
+#include <ginkgo/core/base/executor.hpp>
+#include <ginkgo/core/matrix/csr.hpp>
+
+#include "core/test/utils.hpp"
+
+
+namespace {
```
yhmtsai, line 19: nit: we do not use anonymuos namespace for test for new file now

test/distributed/assembly_helpers_kernels.cpp
```diff
@@ -0,0 +1,197 @@
...
+        gko::experimental::distributed::comm_index_type>(
+        num_rows,
+        std::uniform_int_distribution<
+            gko::experimental::distributed::comm_index_type>(0, num_parts - 1),
+        this->engine, this->ref);
+    auto input = gko::test::generate_random_device_matrix_data<
+        value_type, global_index_type>(
+        num_rows, num_cols,
+        std::uniform_int_distribution<int>(static_cast<int>(num_cols - 1),
+                                           static_cast<int>(num_cols - 1)),
```
yhmtsai, line 183: do you want to generate the full dense matrix?

test/distributed/assembly_helpers_kernels.cpp
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    std::default_random_engine engine;
+};
+
+TYPED_TEST_SUITE(AssemblyHelpers, gko::test::ValueLocalGlobalIndexTypes,
+                 TupleTypenameNameGenerator);
+
+
+TYPED_TEST(AssemblyHelpers, AddNonLocalEntriesDiagOffdiagEmptyIsSameAsRef)
```
yhmtsai, line 108:
```suggestion
TYPED_TEST(AssemblyHelpers, AddNonLocalEntriesEmptyIsSameAsRef)
```
it tests an empty matrix, right? or is it other meaning I miss?

test/mpi/assembly_helpers.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+
+TYPED_TEST(AssemblyHelpers, AddsNonLocalEntries)
+{
+    using value_type = typename TestFixture::value_type;
+    using local_index_type = typename TestFixture::local_index_type;
+    using global_index_type = typename TestFixture::global_index_type;
+    I<I<value_type>> res_local[] = {{{1, 1}, {0, 3}}, {{7, 1}, {0, 7}}, {{10}}};
+    I<I<value_type>> res_non_local[] = {
+        {{0, 2}, {4, 0}}, {{1, 5, 0}, {0, 0, 7}}, {{9}}};
```
yhmtsai, line 97: It is unused

yhmtsai (approved):
LGTM. leaving the default argument into the discussion decision

MarcelKoch (approved):

dpcpp/distributed/assembly_helpers_kernels.dp.cpp
```diff
@@ -0,0 +1,49 @@
...
+    comm_index_type local_part, array<comm_index_type>& send_count,
+    array<GlobalIndexType>& send_positions,
+    array<GlobalIndexType>& original_positions) GKO_NOT_IMPLEMENTED;
+
+GKO_INSTANTIATE_FOR_EACH_VALUE_AND_LOCAL_GLOBAL_INDEX_TYPE(
+    GKO_DECLARE_COUNT_NON_OWNING_ENTRIES);
+
+
+template <typename ValueType, typename LocalIndexType, typename GlobalIndexType>
+void fill_send_buffers(
```
yhmtsai, line 31: It is implemented by unified now. It leads the multiple definition in the pipeline

fritzgoebel: thanks!

include/ginkgo/core/distributed/assembly_helpers.hpp
```diff
@@ -0,0 +1,57 @@
+// SPDX-FileCopyrightText: 2017 - 2024 The Ginkgo authors
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+#ifndef GKO_PUBLIC_CORE_DISTRIBUTED_ASSEMBLY_HELPERS_HPP_
```
upsj, line 5: I would consider renaming assembly_helpers to just assembly, since the functions essentially do the full distributed part of the assembly, not just help in it.

upsj (approved):
LGTM!

PR #1651, opened 2024-07-23
Unify and simplify batch functionality: Multivector

pratikvn:
This PR is the first part of the batched kernel refactoring. It removes the .inc style files, moves to .hpp, making the functions available in a common namespace. It also uses a single source for both HIP and CUDA.

## TODO

- [x] Check if there are any issues with solver includes.
- [x] Remove existing redundant .inc files
- [ ] Differences in `__launch_bounds__` between CUDA and HIP needs to be resolved


MarcelKoch:
If you haven't noted that already, there is a naming inconsistency of the single-batch matrix-apply kernels for the reference and cuda/hip backend.
In the reference backend those are called `simple|advanced_apply_kernel`, while for cuda/hip they are called `simple|advanced_apply`.
I think that should also be unified.

upsj (approved):
LGTM, I'm wondering if we can get rid of some of the GKO_DEVICE_NAMESPACE:: annotations though - is there an ambiguity issue when removing them?

pratikvn:
@MarcelKoch , yes, I plan to unify those when refactoring the matrix format kernels.

common/cuda_hip/base/batch_multi_vector_kernels.hpp
```diff
@@ -20,8 +58,7 @@ __device__ __forceinline__ void scale(
 
 
 template <typename ValueType, typename Mapping>
-__global__
-__launch_bounds__(default_block_size, sm_oversubscription) void scale_kernel(
+__global__ __launch_bounds__(default_block_size) void scale_kernel(
```
yhmtsai, line 61: you have used 4 for sm_oversubscription on both cuda/hip.
I assume the cuda is the correct and hip just uses it.
if you want to compute in more accurate mapping, hip should use (min_blocks_multiprocessor (4) * max_threads_per_block (256) )/32 = 32 for hip.
you will need to distinguish it by macro

pratikvn: I havent fully benchmarked that yet. I agree that this will have to be a macro specialized for CUDA and HIP. But will done in a future PR. It has already been noted in https://github.com/ginkgo-project/ginkgo/issues/1376

yhmtsai (approved):

pratikvn:
format!

PR #1652, opened 2024-07-23
Temporarily disable optimized batched solver instantiations

pratikvn:
This PR temporarily disables the instantiations of the different batched solver kernels to reduce compilation time.

### DIFF
--- a/cuda/solver/batch_bicgstab_kernels.cu
+++ b/cuda/solver/batch_bicgstab_kernels.cu
@@ -167,6 +167,9 @@ public:
 
         value_type* const workspace_data = workspace.get_data();
 
+        // Only instantiate when full optimizations has been enabled. Otherwise,
+        // just use the default one with no shared memory.
+#ifdef GINKGO_BATCHED_FULL_OPTIMIZATIONS
         // Template parameters launch_apply_kernel<StopType, n_shared,
         // prec_shared>
         if (sconf.prec_shared) {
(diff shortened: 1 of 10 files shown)


cuda/solver/batch_bicgstab_kernels.cu
```diff
@@ -167,68 +167,69 @@ public:
...
-            case 9:
-                launch_apply_kernel<StopType, 9, false>(
-                    sconf, logger, prec, mat, b.values, x.values,
-                    workspace_data, block_size, shared_size);
-                break;
-            default:
-                GKO_NOT_IMPLEMENTED;
-            }
-        }
+        // if (sconf.prec_shared) {
```
MarcelKoch, line 173: nit: IMO it could just be deleted directly, we still have the code in the git log, so it won't be lost. But I'm also fine with leaving it as it is, which probably highlights that it's only temporary.

MarcelKoch (approved):

yhmtsai:
It currently forces the behavior.
Because you add the config.hpp, I thought you would like to use config.hpp (configured by CMake) to control the behavior.
Because it is used by some projects, do they use these optimized instantiation?

pratikvn:
Yes, currently we decided to temporarily remove the optimized instantiations. A more permanent solution would be something of the form in https://github.com/ginkgo-project/ginkgo/pull/1629

We decided not to go with the config options as that is maybe more annoying for users.

Currently a few projects use the optimized versions, but I think they rely on the previous release, so it should be okay. We will move to the split version soon anyway.

pratikvn:
Created an issue https://github.com/ginkgo-project/ginkgo/issues/1661, with the milestone 1.9.0, to track it.

upsj (approved):

cuda/solver/batch_bicgstab_kernels.cu
upsj: I feel the same way, but since this is temporary, it shouldn't matter much

PR #1653, opened 2024-07-24
Add early return to solvers that apply preconditioners before first residual check

fritzgoebel:
For the following solvers, in case of an alread good enough initial guess, we apply the preconditioner before checking the residual norm for the first time. That can be quite a bit of wasted effort, so I suggest adding a residual norm check to avoid it.

I added it for CG here, if we agree this is reasonable, it would also be applicable to

- BiCG
- FCG
- GCR


pratikvn (changes requested):
You need to update the benchmark and example outputs, but otherwise, I think this looks good.

PR #1656, opened 2024-07-30
Fix stride for GCR initialization

upsj:
Noticed by @nbeams, the GCR initialization kernel uses an incorrect stride. This fixes it and adds a test for the behavior.


MarcelKoch (approved):

pratikvn (approved):

yhmtsai (approved):
@nbeams good catch!
Just to be clear, assuming that b and residual have the same stride leads the issue, right?

upsj:
@yhmtsai exactly, the residual uses the default stride, while b can have an arbitrary stride

PR #1658, opened 2024-08-02
add Schwarz config

yhmtsai:
This PR adds the schwarz config.
Following the current decision, only allow the global index setting from the `type_descriptor`.
Because sizeof(global_index_type) >= sizeof(index_type), there is an invalid configuration (int64, int32) such that we can not use current dispatch. We only have this case, so I directly write it into the schwarz not create another generalized version for dispatch.

### DIFF
--- a/core/test/config/type_descriptor.cpp
+++ b/core/test/config/type_descriptor.cpp
@@ -21,6 +21,7 @@ TEST(TypeDescriptor, TemplateCreate)
 
         ASSERT_EQ(td.get_value_typestr(), "float64");
         ASSERT_EQ(td.get_index_typestr(), "int32");
+        ASSERT_EQ(td.get_local_index_typestr(), td.get_index_typestr());
         ASSERT_EQ(td.get_global_index_typestr(), "int64");
     }
     {
@@ -29,6 +30,7 @@ TEST(TypeDescriptor, TemplateCreate)
 
         ASSERT_EQ(td.get_value_typestr(), "float32");
         ASSERT_EQ(td.get_index_typestr(), "int32");
+        ASSERT_EQ(td.get_local_index_typestr(), td.get_index_typestr());
         ASSERT_EQ(td.get_global_index_typestr(), "int32");
     }
     {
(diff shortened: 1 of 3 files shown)


core/config/schwarz_config.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+
+template <>
+deferred_factory_parameter<gko::LinOpFactory> parse<LinOpFactoryType::Schwarz>(
+    const pnode& config, const registry& context, const type_descriptor& td)
+{
+    auto updated = update_type(config, td);
+    // We can not directly dispatch the global index type without consider local
+    // index type, which leads the invalid index type <int64, int32> in
+    // compile time.
```
pratikvn, line 26: This logic would be necessary for all distributed types with Global and Local Index types, right ?

core/test/config/type_descriptor.cpp
```diff
@@ -21,27 +21,40 @@ TEST(TypeDescriptor, TemplateCreate)
...
         ASSERT_EQ(td.get_index_typestr(), "int32");
+        ASSERT_EQ(td.get_global_index_typestr(), "int64");
+    }
+    {
+        SCOPED_TRACE("specify global indextype");
+        auto td = make_type_descriptor<float, int, int>();
+
+        ASSERT_EQ(td.get_value_typestr(), "float32");
+        ASSERT_EQ(td.get_index_typestr(), "int32");
+        ASSERT_EQ(td.get_global_index_typestr(), "int");
```
pratikvn, line 32: Maybe to be clear:
```suggestion
        ASSERT_EQ(td.get_global_index_typestr(), "int32");
```

include/ginkgo/core/config/type_descriptor.hpp
```diff
@@ -27,10 +29,9 @@ namespace config {
  * value `void` can be used to specify that no default type is provided. In this
  * case, the configuration has to provide the necessary template types.
  *
- * If the configuration specifies one of the fields (or both):
+ * If the configuration specifies one field (only allow value_type now):
  * ```
  * value_type: "some_value_type"
- * index_type: "some_index_type"
  * ```
  * these types will take precedence over the type_descriptor.
```
pratikvn, line 36:
```suggestion
 * this type will take precedence over the type_descriptor.
```

include/ginkgo/core/config/type_descriptor.hpp
```diff
@@ -59,9 +63,15 @@ class type_descriptor final {
      */
     const std::string& get_index_typestr() const;
 
+    /**
+     * Get the global index type string
+     */
+    const std::string& get_global_index_typestr() const;
```
pratikvn, line 69: Just `index` and `global_index` is a bit confusing IMO. It would be nice to change it to `local_index` and `global_index`, but I understand that that would be a breaking change.

pratikvn (approved):
LGTM!

core/config/schwarz_config.cpp
yhmtsai: yes, but no other cases require it currently.

core/test/config/type_descriptor.cpp
yhmtsai: interesting... "int" should be wrong but the test seems not to catch it

yhmtsai: I miss adding it in the CMakeLists..

core/config/schwarz_config.cpp
yhmtsai: we limit linopfactory only

MarcelKoch (approved):
LGTM. I would suggest to not instantiate `type_descriptor` versions, which are not allowed.

include/ginkgo/core/config/type_descriptor.hpp
```diff
@@ -42,12 +43,15 @@ class type_descriptor final {
      *
      * @param value_typestr  the value type string. "void" means no default.
      * @param index_typestr  the index type string. "void" means no default.
+     * @param global_index_typestr  the global index type string. "void" means
+     * no default.
```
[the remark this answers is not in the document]
yhmtsai: no, the value in `the value "void"` just mentions it is a possible value like float32, float64, int32 not it is for valuetype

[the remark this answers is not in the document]
yhmtsai: Yes, we can only set up value type in file now. with void in index or global index type, user will get error when they use some solver requiring the index type.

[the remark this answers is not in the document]
yhmtsai: It seems to be useful only when user wants to ensure solver only uses the type provided by user.
It will be quite rare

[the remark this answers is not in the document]
yhmtsai: I have removed them for now

PR #1662, opened 2024-08-06
add file config example with the configs

yhmtsai:
This PR adds the example to show the solver configuration based on the file.
This PR aims to set up config close to the existing example without any registry usage.
Note. the stopping criterion is different among these configs.
- cg.json: simple-solver
- blockjacobi-cg.json: preconditioned-solver
- ir.json: iterative-refinement
- parilu.json: ilu-preconditioned-solver (by using factoization parameter directly)
- pgm-multigrid-cg.json: multigrid-preconditioned-solver (set min_coarse_rows additionally due to this small example matrix)
- mixed-pgm-multigrid-cg.json: mixed-multigrid-preconditioned-solver (assuming there are always more than one level)

There will be another one example to show the registry usage.


examples/file-config-solver/config/blockjacobi-cg.json
MarcelKoch, line 1: IMO, all json files should be merged, and the filenames should be keywords instead. I think that is more realistic for acutal usecases.

examples/file-config-solver/build.sh
```diff
@@ -0,0 +1,16 @@
...
+    echo -e "Usage: $0 GINKGO_BUILD_DIRECTORY"
+    exit 1
+fi
+BUILD_DIR=$1
+THIS_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" &>/dev/null && pwd )
+
+source ${THIS_DIR}/../build-setup.sh
+
+# build
+${CXX} -std=c++14 -o ${THIS_DIR}/file-config-solver ${THIS_DIR}/file-config-solver.cpp \
```
MarcelKoch, line 14:
```suggestion
${CXX} -std=c++17 -o ${THIS_DIR}/file-config-solver ${THIS_DIR}/file-config-solver.cpp \
```
I guess this would need to be updated for all examples.

examples/file-config-solver/file-config-solver.cpp
```diff
@@ -0,0 +1,150 @@
...
+    //              directly)
+    // pgm-multigrid-cg.json: multigrid-preconditioned-solver (set
+    //                        min_coarse_rows additionally due to this small
+    //                        example matrix)
+    // mixed-pgm-multigrid-cg.json: mixed-multigrid-preconditioned-solver
+    //                              (assuming there are always more than one
+    //                              level)
+    auto config = gko::ext::config::parse_json_file(configfile);
+    // Create the registry, which allows passing the existing data into config
+    // This example does not show the usage
```
MarcelKoch, line 94: nit:
```suggestion
    // This example does use existing data.
```

examples/file-config-solver/file-config-solver.cpp
```diff
@@ -0,0 +1,150 @@
...
+    auto tic = std::chrono::steady_clock::now();
+    solver->apply(b, x);
+    exec->synchronize();
+    auto toc = std::chrono::steady_clock::now();
+    time += std::chrono::duration_cast<std::chrono::nanoseconds>(toc - tic);
+
+    // Print out the solver config
+    std::cout << "Config file: " << configfile << std::endl;
+    std::ifstream f(configfile);
+    std::cout << f.rdbuf() << std::endl;
```
MarcelKoch, line 129: You could output the json directly:
```
std::cout << std::setw(4) << nlohman::json::parse(configfile) << std::endl;
```
see: https://json.nlohmann.me/api/operator_ltlt/
Of course, it would make more sense if the json object was stored already before.

MarcelKoch (changes requested):
LGTM. I would prefer the json files merged.

examples/file-config-solver/config/blockjacobi-cg.json
yhmtsai: I thought user usually swap the file to choose the solver, which gives less confusing.

MarcelKoch: I would think a single file is less confusing, since it's only one file, where the keys are the previous filenames, and you can add comments for each configuration. (Ofc not in JSON, but that is arguably not a good format for this use case anyway.)

thoasm: After reading through this PR, I agree with @yhmtsai.
Having one file per configuration is much simpler because it makes it easier to swap them around.
Additionally, it is easier to understand when reading the code, which is important for this example.

examples/file-config-solver/file-config-solver.cpp
```diff
@@ -0,0 +1,150 @@
...
+    // ir.json: iterative-refinement
+    // parilu.json: ilu-preconditioned-solver (by using factoization parameter
+    //              directly)
+    // pgm-multigrid-cg.json: multigrid-preconditioned-solver (set
+    //                        min_coarse_rows additionally due to this small
+    //                        example matrix)
+    // mixed-pgm-multigrid-cg.json: mixed-multigrid-preconditioned-solver
+    //                              (assuming there are always more than one
+    //                              level)
+    auto config = gko::ext::config::parse_json_file(configfile);
```
thoasm, line 92: My main concern is that this only tests a single file as part of our automatic tests. It doesn't check the other files, so we couldn't detect if we broke the compatibility.

examples/CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ set(EXAMPLES_EXEC_LIST
 set(EXAMPLES_LIST
     ${EXAMPLES_EXEC_LIST}
     custom-stopping-criterion
+    file-config-solver
```
thoasm, line 23: Does this mean they are not run as part of our CI?
Is there a reason why we don't?

thoasm (commented):
Right now, I only comment until the discussion is resolved.
I would like all configuration files to be tested and part of the CI.
Other than that, I only have nits.

examples/file-config-solver/file-config-solver.cpp
yhmtsai: I assume you mean `This example does not use existing data.`

examples/file-config-solver/file-config-solver.cpp
yhmtsai: Thanks for this information.
I will put it when we have the example with loading json object before

examples/CMakeLists.txt
yhmtsai: I have added test which only checks it can run properly.

yhmtsai: I do not reuse the existing one because it needs to accept additional parameter

MarcelKoch (approved):
I still think that
1. JSON is not the right format
2. The files should be combined, and annotated with comments

But I won't hold up the PR.

MarcelKoch (changes requested):
you also need to add the file to `doc/examples/examples.hpp.in`

yhmtsai:
I think json does not support the comment in the file.

My thought on this example is to show user can provide any kind of new config and then they can get a new solver without recompilation.
Combining them and accessing by key is somehow in the half-way. When they add a new solver to try, they need to recompile the program for the new key which is conflicted with the file config benefit.

> Json is not suitable format 

It is another topic. this example does not tend to handle this discussion.

MarcelKoch (approved):

examples/file-config-solver/file-config-solver.cpp
```diff
@@ -0,0 +1,150 @@
...
+    solver->add_logger(logger);
+
+    // Solve system
+    exec->synchronize();
+    const auto tic = std::chrono::steady_clock::now();
+    solver->apply(b, x);
+    exec->synchronize();
+    const auto toc = std::chrono::steady_clock::now();
+    const auto time =
+        std::chrono::duration_cast<std::chrono::nanoseconds>(toc - tic);
```
greole, line 124: Maybe change to:
```suggestion
        std::chrono::duration_cast<std::chrono::milliseconds>(toc - tic);
```
you can remove the conversions later?

examples/file-config-solver/file-config-solver.cpp
```diff
@@ -0,0 +1,150 @@
...
+    // Print solver statistics
+    std::cout << "Solver iteration count:     " << logger->get_num_iterations()
+              << std::endl;
+    std::cout << "Solver generation time [ms]: "
+              << static_cast<double>(gen_time.count()) / 1000000.0 << std::endl;
+    std::cout << "Solver execution time [ms]: "
+              << static_cast<double>(time.count()) / 1000000.0 << std::endl;
+    std::cout << "Solver execution time per iteration[ms]: "
+              << static_cast<double>(time.count()) / 1000000.0 /
+                     logger->get_num_iterations()
```
greole, line 148: See comment above about milliseconds.

examples/file-config-solver/doc/intro.dox
```diff
@@ -0,0 +1,5 @@
+<a name="File Config Solver"></a>
+<h1>This example shows how to use file to configure solver.</h1>
+
+<h3> In this example, we first read in a matrix from a file. We read the file to configure the solver. The example features the generating time and runtime of the solver.</h3>
```
greole, line 4:
```suggestion
<h3> In this example, we first read in a matrix from a file. Then we read the config file to configure the solver. The example application reports the generation and run time of the solver.</h3>
```

examples/file-config-solver/file-config-solver.cpp
```diff
@@ -0,0 +1,150 @@
...
+    auto one = gko::initialize<vec>({1.0}, exec);
+    auto neg_one = gko::initialize<vec>({-1.0}, exec);
+    auto initres = gko::initialize<vec>({0.0}, exec);
+    A->apply(one, x, neg_one, b);
+    b->compute_norm2(initres);
+
+    // Copy b again
+    b->copy_from(host_b);
+
+    // Read the json file into ginkgo structure
```
greole, line 81:
```suggestion
    // Read the json config file to configure the ginkgo solver. The following files are available 
```

examples/file-config-solver/file-config-solver.cpp
```diff
@@ -0,0 +1,150 @@
...
+    // generate the linopfactory on the given executors
+    auto solver_gen = gko::config::parse(config, reg, td).on(exec);
+
+    // Create solver
+    const auto gen_tic = std::chrono::steady_clock::now();
+    auto solver = solver_gen->generate(A);
+    exec->synchronize();
+    const auto gen_toc = std::chrono::steady_clock::now();
+    const auto gen_time =
+        std::chrono::duration_cast<std::chrono::nanoseconds>(gen_toc - gen_tic);
```
greole, line 110: Maybe use ms to avoid conversion later.
```suggestion
        std::chrono::duration_cast<std::chrono::milliseconds>(gen_toc - gen_tic);
```

greole (approved):
LGTM.

examples/file-config-solver/file-config-solver.cpp
yhmtsai: good point!

PR #1665, opened 2024-08-11
Fix triangular solvers on Windows CUDA

upsj:
The alignment of uninitialized_array was currently only 1 based on `unsigned char`, and the whole setup screams _undefined behavior_ to me, so this makes it much safer for our current use cases (since casting between complex and real types is well-defined).

Additionally, MSVC doesn't properly find `isnan` as a device function, so we need to add our own implementation.


common/cuda_hip/matrix/csr_kernels.template.cpp
```diff
@@ -335,7 +335,7 @@ __device__ void merge_path_reduce(const IndexType nwarps,
             }
         }
     }
-    __shared__ uninitialized_array<IndexType, spmv_block_size> tmp_ind;
+    __shared__ IndexType tmp_ind[spmv_block_size];
```
upsj, line 338: we don't need uninitialized_array here

MarcelKoch (approved):

yhmtsai (approved):
I was wondering whether jacboi using any type whose size is not divisible by 4 for shared memory may lead an issue here. But Jacobi seems not to use it.

MarcelKoch (approved):

include/ginkgo/core/base/math.hpp
```diff
@@ -1220,10 +1236,12 @@ GKO_INLINE GKO_ATTRIBUTES T safe_divide(T a, T b)
  * @return `true` if the value is NaN.
  */
 template <typename T>
-GKO_INLINE GKO_ATTRIBUTES std::enable_if_t<!is_complex_s<T>::value, bool>
-is_nan(const T& value)
+GKO_DEPRECATED("is_nan can't be used safely on the device (MSVC+CUDA)")
```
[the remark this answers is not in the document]
upsj: I don't want to give that recommendation, since that misses most NaN values, this is only meant to recognize the value returned by `gko::nan`, while `is_nan` might also be useful in other cases. This is meant to discourage using the function at all.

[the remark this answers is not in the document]
MarcelKoch: Ok, then explicitly state that this function is deprecated without replacement. Otherwise, it could lead to confusion.

include/ginkgo/core/base/math.hpp
```diff
@@ -102,6 +103,21 @@ using std::sqrt;
...
+
+template <>
+struct float_to_bytes_impl<double> {
+    using type = uint64;
+};
+
+template <>
+struct float_to_bytes_impl<float> {
+    using type = uint32;
+};
```
thoasm, line 118: Please name it `float_to_uint_impl` because the type you are using is not `byte`.

include/ginkgo/core/base/math.hpp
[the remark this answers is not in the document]
thoasm: Do you really want to deprecate the whole function because one use case disappeared? To me, they fulfill two separate intentions because`is_nan_exact` can't detect all NaN values.
I would at least not mention the `is_nan_exact` alternative.

include/ginkgo/core/base/math.hpp
```diff
@@ -1274,6 +1297,48 @@ nan()
...
+ *
+ * @return `true` if any component of the complex number fulfills
+ * is_nan_exact(component).
+ */
+template <typename T>
+GKO_INLINE GKO_ATTRIBUTES std::enable_if_t<is_complex_s<T>::value, bool>
+is_nan_exact(const T& value)
+{
+    return is_nan_exact(value.real()) || is_nan_exact(value.imag());
+}
```
thoasm, line 1339: Can you move these function to `common_trs_kernels`?
To me, they are not general enough to justify putting them in the public header. The chances that they are misused are quite big (this function likely doesn't detect NaNs caused by actual computations).

thoasm (changes requested):
I would like to have the `is_nan_exact` function moved to the TRS kernel file to avoid future misuse.

include/ginkgo/core/base/math.hpp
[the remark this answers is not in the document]
upsj: It was the only use case, and the function itself cannot be used in CUDA device code, which makes it misplaced in this function

include/ginkgo/core/base/math.hpp
upsj: makes sense, I'll move it

include/ginkgo/core/base/math.hpp
```diff
@@ -1220,10 +1220,14 @@ GKO_INLINE GKO_ATTRIBUTES T safe_divide(T a, T b)
  * @return `true` if the value is NaN.
  */
 template <typename T>
-GKO_INLINE GKO_ATTRIBUTES std::enable_if_t<!is_complex_s<T>::value, bool>
-is_nan(const T& value)
+GKO_DEPRECATED(
+    "is_nan can't be used safely on the device (MSVC+CUDA), and will thus be "
+    "removed in a future release, without replacement")
```
yhmtsai, line 1225: We only use this function in trs apparently, so I am fine to deprecate this now.
If we have other usage, which does not rely on our initial value, I will prefer still providing a proper one (std::is_nan in general but a bit time/memory-consuming on msvc+cuda side)

include/ginkgo/core/base/math.hpp
```diff
@@ -1220,10 +1220,14 @@ GKO_INLINE GKO_ATTRIBUTES T safe_divide(T a, T b)
...
+GKO_INLINE GKO_ATTRIBUTES
+    std::enable_if_t<!is_complex_s<T>::value, bool> is_nan(const T& value)
 {
-    return std::isnan(value);
+    using std::isnan;
+    return isnan(value);
```
yhmtsai, line 1230: is `std::isnan(value)` different from the updated one?

yhmtsai (approved):

cuda/solver/common_trs_kernels.cuh
```diff
@@ -342,6 +343,46 @@ constexpr int default_block_size = 512;
...
+    using type = uint64;
+};
+
+template <>
+struct float_to_unsigned_impl<float> {
+    using type = uint32;
+};
+
+
+/** Checks if a floating point number is a quiet NaN (gko::nan()). */
```
thoasm, line 361:
```suggestion
/** Checks if a floating point number representation matches the representation of the quiet NaN of gko::nan() exactly. */
```

include/ginkgo/core/base/math.hpp
thoasm: I don't think so, but this has a better chance of using the CUDA `isnan` function.

cuda/solver/common_trs_kernels.cuh
```diff
@@ -342,6 +343,46 @@ constexpr int default_block_size = 512;
...
+    type nan_bytes{};
+    auto nan_value = nan<T>();
+    using std::memcpy;
+    memcpy(&value_bytes, &value, sizeof(value));
+    memcpy(&nan_bytes, &nan_value, sizeof(value));
+    return value_bytes == nan_bytes;
+}
+
+
+/** Checks if any component of a complex value is a quiet NaN (gko::nan). */
```
thoasm, line 377:
```suggestion
/** Checks if any component of the complex value matches the quiet NaN gko::nan() exactly. */
```

thoasm (approved):
Thank you for applying my suggestions.
I only have a minor nit regarding the documentation; the rest looks good!

include/ginkgo/core/base/math.hpp
upsj: unfortunately it still doesn't work with MSVC, which means I'll have to file a bug report some time ;)

PR #1666, opened 2024-08-12
Fix not-found tbb in the intel github action 

yhmtsai:
This pr fixes the not-found tbb module in the intel github action.
The library itself seems to be still available when loading the other intel module.

- [ ] wait for that the intel github action are passed


pratikvn:
So, it is automatically loaded now (within some other package), and still used ? Or not necessary at all ?

yhmtsai:
@pratikvn I think the dependence is from oneDPL. We do not use them directly, but we need to link it in our CMake.
I can check whether it is still needed now and probably report this.

yhmtsai:
I have checked the spack installation, the intel-oneapi-mkl package will also install intel-tbb package.
If it does not find the intel-oneapi-tbb, it will use tbb library from intel-tbb.

pratikvn (approved):

.github/workflows/intel.yml
```diff
@@ -33,20 +33,20 @@ jobs:
...
         source /etc/profile
-        module load intel-oneapi-compilers intel-oneapi-dpl intel-oneapi-mkl intel-oneapi-tbb cmake
+        module load intel-oneapi-compilers intel-oneapi-dpl intel-oneapi-mkl cmake
         cd build
         SYCL_DEVICE_FILTER=level_zero:gpu make install
         export GINKGO_PATH="$(pwd)/install_ginkgo/lib"
-        export LIBRARY_PATH=${GINKGO_PATH}:$LIBRARY_PATH
-        export LD_LIBRARY_PATH=${GINKGO_PATH}:$LD_LIBRARY_PATH
+        export LIBRARY_PATH=${ICL_INTEL_TBB_ROOT}/lib64:${GINKGO_PATH}:$LIBRARY_PATH
+        export LD_LIBRARY_PATH=${ICL_INTEL_TBB_ROOT}/lib64:${GINKGO_PATH}:$LD_LIBRARY_PATH
```
MarcelKoch, line 51: This seems a bit brittle, but as long as it works it's fine with me.
Have you tried using `find_package(...)` and `link_libraries(...)` to set the paths instead?

MarcelKoch (approved):

.github/workflows/intel.yml
yhmtsai: yes, I agree with you.
No, I haven't. I think it should be done from MKL/DPL if they put it in public linking interface.
There are two package name (for the same thing, AFAIK): TBB from oneMKL (current in machine) and oneTBB from usual setup.

PR #1667, opened 2024-08-15
Named lambda operation

MarcelKoch:
This PR enables the naming of lambda operations.

### DIFF
--- a/core/test/base/executor.cpp
+++ b/core/test/base/executor.cpp
@@ -521,4 +521,42 @@ TEST_F(ExecutorLogging, LogsOperation)
 }
 
 
+struct NameLogger : public gko::log::Logger {
+protected:
+    void on_operation_launched(const gko::Executor* exec,
+                               const gko::Operation* op) const override
+    {
+        op_name = op->get_name();
+    }
+
+public:
+    mutable std::string op_name;
+};
+
+
+TEST(LambdaOperation, CanSetName)
+{
+    auto name_logger = std::make_shared<NameLogger>();
+    auto exec = gko::ReferenceExecutor::create();
+    exec->add_logger(name_logger);
+
+    exec->run(
+        "name", [] {}, [] {}, [] {}, [] {});
+
+    ASSERT_EQ("name", name_logger->op_name);
+}
+
+
+TEST(LambdaOperation, HasDefaultName)
+{
+    auto name_logger = std::make_shared<NameLogger>();
+    auto exec = gko::ReferenceExecutor::create();
+    exec->add_logger(name_logger);
+
+    exec->run([] {}, [] {}, [] {}, [] {});
+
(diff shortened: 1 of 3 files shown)


core/test/base/executor.cpp
```diff
@@ -521,4 +521,42 @@ TEST_F(ExecutorLogging, LogsOperation)
...
+
+TEST(LambdaOperation, HasDefaultName)
+{
+    auto name_logger = std::make_shared<NameLogger>();
+    auto exec = gko::ReferenceExecutor::create();
+    exec->add_logger(name_logger);
+
+    exec->run([] {}, [] {}, [] {}, [] {});
+
+    ASSERT_EQ("unname", name_logger->op_name);
```
pratikvn, line 558:
```suggestion
    ASSERT_EQ("unnamed", name_logger->op_name);
```

pratikvn (changes requested):
Minor typo. Otherwise looks good!

core/test/base/executor.cpp
MarcelKoch: I think I'm going to delete this test. The 'default' name is not really important, since we don't specify it publicly. Testing that value seems rather pointless.

upsj: I think it makes sense to test that the value returned is a valid string literal and not an invalid pointer (which would otherwise be very unlikely to be found)

include/ginkgo/core/base/executor.hpp
```diff
@@ -659,6 +659,17 @@ class Executor : public log::EnableLogging<Executor> {
         this->run(op);
     }
 
+    template <typename ClosureOmp, typename ClosureCuda, typename ClosureHip,
+              typename ClosureDpcpp>
+    void run(std::string name, const ClosureOmp& op_omp,
```
upsj, line 664: Adding a new overload, would it make sense to extend this to include `ReferenceExecutor`? We can't change the other one, since it would break interface

upsj (approved):
LGTM!

include/ginkgo/core/base/executor.hpp
MarcelKoch: Any ordering preference? I would add it before `op_omp`, since I think we usually order them that way. But of course, this is then not compatible with the other overload.

upsj: Sounds sensible, since this is a new overload. We can't add `reference` to the other one though, since that would create ambiguities with the std::string parameter and the additional templated parameter.

MarcelKoch: I guess we could think about deprecating the other overload. Adding a name to an operation seems quite useful, especially since all ginkgo kernels are already named.

pratikvn (approved):

PR #1669, opened 2024-08-21
Unify and simplify batch functionality: Matrix formats (csr, dense, ell)

pratikvn:
Part 2 of the refactor: Unify and simplify batch functionality for matrix formats

## TODO

- [x] Reference and omp kernels
- [x] Dpcpp kernels


common/cuda_hip/base/batch_multi_vector_kernels.hpp
```diff
@@ -22,24 +23,20 @@
...
 namespace kernels {
 namespace GKO_DEVICE_NAMESPACE {
 namespace batch_single_kernels {
+namespace {
 
 
 constexpr auto default_block_size = 256;
 
 
+}
```
upsj, line 37: why this namespace?

upsj (approved):
LGTM! I don't see any functional changes at a cursory glance, are there any, or just renaming and code movement?

common/cuda_hip/base/types.hpp
```diff
@@ -8,8 +8,12 @@
 
 #if defined(GKO_COMPILING_CUDA)
 #include "cuda/base/types.hpp"
+#define device_type cuda_type
+#define as_device_type as_cuda_type
```
MarcelKoch, line 12: aren't those already defined in the `cuda/base/types.hpp` header? Same for hip.

common/cuda_hip/solver/batch_bicgstab_kernels.hpp.inc
```diff
@@ -27,8 +27,9 @@ __device__ __forceinline__ void initialize(
     __syncthreads();
 
     // r = b - A*x
-    advanced_apply(static_cast<ValueType>(-1.0), mat_entry, x_shared_entry,
-                   static_cast<ValueType>(1.0), r_shared_entry);
+    gko::kernels::GKO_DEVICE_NAMESPACE::batch_single_kernels::advanced_apply(
```
MarcelKoch, line 30: is the full qualified namespace necessary? I would have thought that only `batch_single_kernels` is necessary. Or maybe this can be simplified in the next PR.

reference/matrix/batch_csr_kernels.cpp
```diff
@@ -9,26 +9,23 @@
 #include <ginkgo/core/base/batch_multi_vector.hpp>
 #include <ginkgo/core/matrix/batch_csr.hpp>
 
+
+#define GKO_DEVICE_NAMESPACE reference
```
MarcelKoch, line 13: The OMP files don't have this. Is it set by CMake for OMP? If so, maybe also set it via CMake here.

MarcelKoch (changes requested):
I've got some minor question that I would like to get clarified.

common/cuda_hip/base/types.hpp
upsj: ah yeah, missed those. Those should not be macros, as that breaks other files

reference/matrix/batch_csr_kernels.cpp
pratikvn: These are defined in `common/unified/base/kernel_launch.hpp` for the rest of the executors, but not defined for reference.

common/cuda_hip/solver/batch_bicgstab_kernels.hpp.inc
pratikvn: Yes, it should be removed in the next PR for solvers.

reference/matrix/batch_csr_kernels.cpp
upsj: They should be removed there, since they (should) already get set via CMake for other files, e.g. cuda_hip unified files.

MarcelKoch (approved):

PR #1670, opened 2024-08-21
Fix ROCm 6.1 issues

upsj:
The CSR lookup tests and all usages of the lookup functionality fails with ROCm 6.1 for some reason. These workarounds were only necessary in older ROCm versions anyways.

TODO:
- [x] Fix sorting segfault
- [x] Fix ParILUT segfault (probably related to the sorting segfault)
- [x] Fix searching hang (not a hang, just a slow test)

### DIFF
--- a/common/cuda_hip/components/sorting.hpp
+++ b/common/cuda_hip/components/sorting.hpp
@@ -118,9 +118,9 @@ struct bitonic_warp {
             // workaround for ROCm 6.x segfaults on gfx906
 #ifdef GKO_COMPILING_CUDA
             auto other = __shfl_xor_sync(config::full_lane_mask, els[i],
-                                         num_threads / 2);
+                                         num_threads / 2, num_threads);
 #else
-            auto other = __shfl_xor(els[i], num_threads / 2);
+            auto other = __shfl_xor(els[i], num_threads / 2, num_threads);
 #endif
             bitonic_cas(els[i], other, new_reverse);
         }
--- a/common/cuda_hip/components/sorting.hpp
+++ b/common/cuda_hip/components/sorting.hpp
@@ -113,11 +113,15 @@ struct bitonic_warp {
 
     __forceinline__ __device__ static void merge(ValueType* els, bool reverse)
     {
-        auto tile =
-            group::tiled_partition<num_threads>(group::this_thread_block());
         auto new_reverse = reverse != upper_half();
         for (int i = 0; i < num_local; ++i) {
-            auto other = tile.shfl_xor(els[i], num_threads / 2);
+            // workaround for ROCm 6.x segfaults on gfx906
+#ifdef GKO_COMPILING_CUDA
+            auto other = __shfl_xor_sync(config::full_lane_mask, els[i],
+                                         num_threads / 2);
+#else
+            auto other = __shfl_xor(els[i], num_threads / 2);
+#endif
             bitonic_cas(els[i], other, new_reverse);
         }
         half::merge(els, reverse);
--- a/include/ginkgo/core/base/types.hpp
+++ b/include/ginkgo/core/base/types.hpp
@@ -51,29 +51,7 @@
 #endif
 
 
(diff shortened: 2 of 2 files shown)


include/ginkgo/core/base/types.hpp
```diff
@@ -51,30 +51,8 @@
...
-#ifdef NDEBUG
-#define GKO_ASSERT(condition) ((void)0)
-#else  // NDEBUG
-// Poor man's assertions on GPUs for MACs. They won't terminate the program
-// but will at least print something on the screen
-#define GKO_ASSERT(condition)                                               \
-    ((condition)                                                            \
-         ? ((void)0)                                                        \
-         : ((void)printf("%s: %d: %s: Assertion `" #condition "' failed\n", \
-                         __FILE__, __LINE__, __func__)))
```
MarcelKoch, line 66: Was this giving issues because of some extra `%` in the `condition`?

upsj: No, it used to be an issue in the batch kernels, but I refactored that to fix the warning. But the ROCm assertions were broken for a long time, now the workaround is causing issues

common/cuda_hip/components/sorting.hpp
```diff
@@ -113,11 +113,15 @@ struct bitonic_warp {
...
     {
-        auto tile =
-            group::tiled_partition<num_threads>(group::this_thread_block());
         auto new_reverse = reverse != upper_half();
         for (int i = 0; i < num_local; ++i) {
-            auto other = tile.shfl_xor(els[i], num_threads / 2);
+            // workaround for ROCm 6.x segfaults on gfx906
+#ifdef GKO_COMPILING_CUDA
+            auto other = __shfl_xor_sync(config::full_lane_mask, els[i],
+                                         num_threads / 2);
```
pratikvn, line 121: Can we not use `tile.shfl_xor` for CUDA ?

pratikvn (approved):

common/cuda_hip/components/sorting.hpp
upsj: we could, but that means we'd have to add `tile` again, which seems to be part of the reason why the compiler generates bad code in the first place

common/cuda_hip/components/sorting.hpp
```diff
@@ -113,11 +113,15 @@ struct bitonic_warp {
...
+#else
+            auto other = __shfl_xor(els[i], num_threads / 2);
```
yhmtsai, line 123: this changes should be the same as use `tiled_partition<config::warp_size>.shlf_xor` not `<num_thread>`
It will be similar to sycl now because sycl does not support subwarp operation and I need to perform it by assuming they are always perform full warp (which is true though)

upsj: yes, that is deliberate, since the full warp is always running this code, so there should be no effect of the smaller width. I'll check if there is any performance difference though

upsj: To make the diff smaller, I reverted it back to using `num_threads`

yhmtsai: sorry. my first comment might be confusing.
I thought you can change it to tile_partition<config::warp_size> for all.
With the current version, you can use the num_threads in width. so is the segfault from some cooperative group class? not from compiler compiles sub_warp wrongly?

yhmtsai (approved):

common/cuda_hip/components/sorting.hpp
upsj: You can change it to that, but I thought since we know this is a smaller shuffle, we can potentially give the hardware a hint for faster execution, just a guess though. The segfault comes from a weird combination of the member variables of thread_block_tile (which should be optimized away since they are unused) and the variable-size shuffle width. But compiler bugs like this are very fragile and can disappear randomly, so I wanted to remove as many things that I saw trigger the bug as possible

PR #1673, opened 2024-08-29
Remove hip path manipulation

MarcelKoch:
This PR removes the setting of the `HIP_PATH` environment variable. It also removes setting the paths for each rocm library individually. Instead, it is expected that user provide the correct `CMAKE_PREFIX_PATH` during configuration.
Additionally, a check is introduced to warn about rocm's faulty CMake setup. For rocm 5.0-5.6 the default installation (as provided in the rocm/dev images) has some bugs, which prevent even the simplest CMake setup to enable the HIP language.


cmake/hip_helpers.cmake
```diff
@@ -0,0 +1,43 @@
+function(ginkgo_find_hip_version)
```
yhmtsai, line 1: Could this be found in `CMAKE_HIP_COMPILER_VERSION`?

PR #1676, opened 2024-09-05
fix the race condition of bicgstab

yhmtsai:
This pr fixes the race condition of bicgstab.
Originally, we put the finalize and update together.
That means that some threads might get the finalized value from the other thread and the thread will not update the value.

We need to separate them to ensure update the stop_status after update values.

It should solve #1563

### DIFF
--- a/common/unified/solver/bicgstab_kernels.cpp
+++ b/common/unified/solver/bicgstab_kernels.cpp
@@ -174,11 +174,18 @@ void finalize(std::shared_ptr<const DefaultExecutor> exec,
                       auto stop) {
             if (stop[col].has_stopped() && !stop[col].is_finalized()) {
                 x(row, col) += alpha[col] * y(row, col);
-                stop[col].finalize();
             }
         },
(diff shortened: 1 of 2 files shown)


MarcelKoch (approved):

upsj (approved):

upsj:
Would it be possible to construct a test where this fails more reliably? E.g. many rhs, few rows

yhmtsai:
It needs more rows such that some have updated the stop_status, but the rest does not read the stop_status yet after the cache value is updated.
The failed test might be easier on the cpu side but not gpu side which relies on when threads can see the updated value from the cache/memory.
I have added a test with some description

yhmtsai:
@upsj @MarcelKoch please feel free to take a look at the test. I will wait for the test failed on our CI and then push the fix back again. If everything works correctly, I will merge this pr

yhmtsai:
The failed tests are in https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/pipelines/1446060008
except for the amd due to CI issue, others faced the issue

MarcelKoch (approved):

PR #1679, opened 2024-09-12
Restructure Documentation

MarcelKoch:
This PR splits the documentation into a part that is generated by sphinx and one that is generated by doxygen.
The doxygen part will cover only the reference API documentation, and anything else will be left to sphinx. The full separation is not done in this PR.

Changes:
- adds sphinx configuration
- make doxygen available for sphinx
- simplify doxygen setup
  - only the `usr` doc is available, the `dev` doc is removed. IMO there is no need for a `dev` doc, since we can expect developers to look into the Ginkgo source code and discover the documentation there.
- fix some existing documentation (grouping, markdown files)
- use read-the-docs instead of GH pages to deploy documentation

TODO:
- [ ] finalize the read-the-docs namespace


thoasm:
Sorry, clicked the close PR button by mistake.

thoasm:
> Link to the documentation at RTD: https://ginkgo-test.readthedocs.io/en/read-the-docs/

This link results in a 404 error. Could you regenerate it or update the link?

MarcelKoch:
This should work: https://ginkgo-test.readthedocs.io/doxysphinx/index.html
I think this contains some changes with are also in #1680

thoasm:
Thanks, that link works!

thoasm (changes requested):
I like the changes, and the documentation looks excellent!
However, when trying out the [new documentation](https://ginkgo-test.readthedocs.io/doxysphinx/), I encountered an issue: It seems to be possible to access the old doxygen documentation, which leads to all links on the page becoming broken.
The easiest way to go to the old page is by clicking one of the links on the `doxysphinx` tab (not sure if that is the right expression) on the bottom right:
Screenshot 2025-02-14 at 11-01-42 User Guide — Ginkgo v1 9 0
After clicking any of the links in there, I assume the old style is used, and all of the links on the Doxygen page don't work anymore.
Here is the Doxysphinx page (that you shared): https://ginkgo-test.readthedocs.io/doxysphinx/index.html
And here is the page with the broken links I got when clicking a link on the tab bottom right: https://ginkgo-test.readthedocs.io/user-guide/

Unfortunately, I don't know how to fix it (I don't know where this tab is coming from), but it must be addressed to improve the user experience (maybe by simply removing this tab altogether).

MarcelKoch:
closed in favor of #1813

PR #1684, opened 2024-09-26
ginkgo own ILU and IC

yhmtsai:
This pr enables the ginkgo's implementation ILU through the current LU implementation.
To enable that, we need to allow LU can ignore the missing fill-in entries. 
It is controlled by checked_lookup parameter, and user needs to pass the same sparsityCsr as the original one as the symbolic factor to get ILU.
Moreover, this pr also adds a parameter for algorithm selection. It will forward the information to LU and unpack them directly to return the same structure as the original sparselib.
Note. LU by default gives combined matrix but ILU gives the composition. 

The first commit shows that the hashmap lookup leads the hang issue and bitmap lookup leads illegal memory access (incorrect result with -1 index return).
The second commit shows the results are wrong if we do not ignore missing fill-in entries update after fixing infinite loop of hashmap lookup

After this pr, we provide another way to factorize ILU without relying on the sparse library.

TODO
- [x] cholesky (in this pr, too)


core/matrix/csr_lookup.hpp
```diff
@@ -262,15 +264,17 @@ struct device_sparsity_lookup {
...
             hash++;
             if (hash >= hashmap_size) {
                 hash = 0;
             }
-            GKO_ASSERT(hashmap[hash] >= 0);
+            out_idx = hashmap[hash];
             GKO_ASSERT(hashmap[hash] < row_nnz);
         }
-        const auto out_idx = hashmap[hash];
+        // out_idx is either correct or invalid_index, the hashmap sentinel
```
nbeams, line 277: I am confused here. With this change, `lookup_hash_unsafe` is the same as `lookup_hash` (except for one extra assertion for `unsafe` -- the `>=0` -- not sure why?). So...`lookup_hash_unsafe` isn't really unsafe anymore? Why do we need any changes to `unsafe` at all, when in the factorization, we switch to use the safe lookup if requested? (`lookup[col]` rather than `lookup.lookup_unsafe(col)`)

nbeams (commented):
I have some questions and comments, but overall I am excited about this! Thanks @yhmtsai

common/cuda_hip/factorization/lu_kernels.cpp
```diff
@@ -85,7 +85,7 @@ __global__ __launch_bounds__(default_block_size) void initialize(
 }
 
 
-template <typename ValueType, typename IndexType>
+template <bool checked_lookup, typename ValueType, typename IndexType>
 __global__ __launch_bounds__(default_block_size) void factorize(
```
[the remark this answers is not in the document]
upsj: I think I would prefer a name that states what it is meant to be used for, rather than a name describing an implementation detail. Something like `has_all_fillin`?

core/factorization/ilu.cpp
```diff
@@ -79,16 +93,36 @@ std::unique_ptr<Composition<ValueType>> Ilu<ValueType, IndexType>::generate_l_u(
...
+                exec, local_system_matrix->get_size(),
+                make_const_array_view(
+                    exec, local_system_matrix->get_num_stored_elements(),
+                    local_system_matrix->get_const_col_idxs()),
+                make_const_array_view(
+                    exec, local_system_matrix->get_size()[0] + 1,
+                    local_system_matrix->get_const_row_ptrs())));
+        ilu =
+            gko::experimental::factorization::Lu<ValueType, IndexType>::build()
+                .with_checked_lookup(true)
```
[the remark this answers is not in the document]
upsj: IMO this should rather be a new type, LU and ILU(...) have very different use cases. That would also allow us to build a new, improved ILU preconditioner interface based on the `Factorization` type instead of the slightly hacky `Composition`-like type ILU is currently.

core/factorization/ilu.cpp
```diff
@@ -79,16 +93,36 @@ std::unique_ptr<Composition<ValueType>> Ilu<ValueType, IndexType>::generate_l_u(
...
+                .with_symbolic_factorization(sparsity)
+                .on(exec)
+                ->generate(local_system_matrix)
+                ->get_combined();
+    } else {
+        exec->run(
+            ilu_factorization::make_compute_ilu(local_system_matrix.get()));
```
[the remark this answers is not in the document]
upsj: I guess based on our existing code, `sparselib_ilu` would be an appropriate name

upsj (dismissed):
Thanks Mike, I'm happy to see how simple this was to extend to ILU. Some design and naming suggestions I'd like to see discussed, and maybe we should build a roadmap for the future ILU preconditioner interface.

core/matrix/csr_lookup.hpp
yhmtsai: `lookup_hash_unsafe` is mostly equal to `lookup_hash` up to one assert as you mentioned.
in my mind. when you call `lookup_unsafe`, you will assume it always returns a valid value back without any checks.
when calling `lookup`, you will always checks whether it is valid or not.
I keep them because of the extra assert. unsafe should report the index is invalid in debug mode because it may not be checked in the code.
I want to change the unsafe lookup because the infinite loop leads debug difficult and this change does not affect the performance.
BTW, I do not think there's performance difference between safe and unsafe version, but I will keep this behavior first.
Then, we can see whether to use safe-only lookup.

core/factorization/ilu.cpp
```diff
@@ -79,16 +93,36 @@ std::unique_ptr<Composition<ValueType>> Ilu<ValueType, IndexType>::generate_l_u(
...
+                make_const_array_view(
+                    exec, local_system_matrix->get_size()[0] + 1,
+                    local_system_matrix->get_const_row_ptrs())));
+        ilu =
+            gko::experimental::factorization::Lu<ValueType, IndexType>::build()
+                .with_checked_lookup(true)
+                .with_symbolic_factorization(sparsity)
+                .on(exec)
+                ->generate(local_system_matrix)
+                ->get_combined();
```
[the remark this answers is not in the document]
yhmtsai: LU always generates the combined matrix first.
We can call `unpack` or rely on the original code here. I do not check there's any difference between the implementation.

common/cuda_hip/factorization/lu_kernels.cpp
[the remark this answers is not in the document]
yhmtsai: I also prefer not safe_lookup. I will think why I use unsafe version rather than safe version.

core/factorization/ilu.cpp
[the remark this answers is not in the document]
nbeams: Right, LU creates combined matrix, but my question is, why call `get_combined` directly (which returns a sparse matrix, right?) when we could store the `Factorization` and get the L and U factors from that? I guess there's no reason to expect the performance would be different between the `Factorization` method for splitting the factors from a combined matrix vs the kernels called for turning the cuSPARSE output into factors?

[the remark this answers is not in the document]
yhmtsai: That's right. I did it because it reuse the following part and make the condition block smaller.
I can also use early return to avoid a large condition block actually.
I am not big fan of large condition block

include/ginkgo/core/factorization/ilu.hpp
```diff
@@ -25,6 +25,14 @@ namespace gko {
 namespace factorization {
 
 
+/**
+ * A helper for algorithm selection in the incomplete factorization.
+ * sparselib is only available for cuda and hip.
+ * syncfree is Ginkgo's implementation through the Lu factorization with given
+ * sparsity.
+ */
+enum class factorize_algorithm { sparselib, syncfree };
```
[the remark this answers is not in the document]
yhmtsai: I will also use it to IC, so it will be for incomplete factorize algorithm at least.
maybe `incomplete_factorize_algorithm` if the original one is too general?

include/ginkgo/core/factorization/lu.hpp
```diff
@@ -97,6 +97,15 @@ class Lu
          * incorrect results or crash.
          */
         bool GKO_FACTORY_PARAMETER_SCALAR(skip_sorting, false);
+
+        /**
+         * The symbolic factoization should contains the fill-in information. If
```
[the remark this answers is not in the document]
yhmtsai: for the related note, it should work if the algorithm is okay with ignoring the value from A but missing in LU

core/factorization/ilu.cpp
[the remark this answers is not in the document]
yhmtsai: To enable this code path, I also add strategy argument to unpack function.

include/ginkgo/core/factorization/ilu.hpp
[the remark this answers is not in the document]
nbeams: That makes sense. I'm mostly thinking about code readability/logicality on the user side. I do think it would be good to specify that this is only for incomplete somehow, though `incomplete_factorize_algorithm` is pretty long. What if we had the same enum for both ilu and ic but with different names (`ilu_algorithm` and `ic_algorithm`)? But we can also go with the longer name. Maybe @upsj has thoughts on this?

common/cuda_hip/factorization/lu_kernels.cpp
```diff
@@ -252,19 +260,29 @@ template <typename ValueType, typename IndexType>
...
+                    lookup_storage, lookup_descs, diag_idxs,
+                    as_device_type(factors->get_values()), storage, num_rows);
+        } else {
+            kernel::factorize<false>
+                <<<num_blocks, default_block_size, 0, exec->get_stream()>>>(
+                    factors->get_const_row_ptrs(),
+                    factors->get_const_col_idxs(), lookup_offsets,
+                    lookup_storage, lookup_descs, diag_idxs,
+                    as_device_type(factors->get_values()), storage, num_rows);
+        }
```
MarcelKoch, line 285: These branches are identical.
I think it should be
```suggestion
        if (!has_all_fillin) {
            kernel::factorize<false>
                <<<num_blocks, default_block_size, 0, exec->get_stream()>>>(
                    factors->get_const_row_ptrs(),
                    factors->get_const_col_idxs(), lookup_offsets,
                    lookup_storage, lookup_descs, diag_idxs,
                    as_device_type(factors->get_values()), storage, num_rows);
        } else {
            kernel::factorize<true>
                <<<num_blocks, default_block_size, 0, exec->get_stream()>>>(
                    factors->get_const_row_ptrs(),
                    factors->get_const_col_idxs(), lookup_offsets,
                    lookup_storage, lookup_descs, diag_idxs,
                    as_device_type(factors->get_values()), storage, num_rows);
        }
```

yhmtsai: good catch!

core/factorization/factorization.cpp
```diff
@@ -31,10 +31,25 @@ GKO_REGISTER_OPERATION(initialize_l, factorization::initialize_l);
 
 template <typename ValueType, typename IndexType>
 std::unique_ptr<Factorization<ValueType, IndexType>>
-Factorization<ValueType, IndexType>::unpack() const
+Factorization<ValueType, IndexType>::unpack(
```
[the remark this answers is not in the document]
yhmtsai: I do not think 2.0 happens soon
if we do not want to introduce here, I can only unpack them by original code path not the unpack them.
@nbeams it should be fine, right? the unpacking codes are the same.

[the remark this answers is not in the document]
nbeams: You mean revert back to before 57c0051e76645ae002360774c0a216d9f574213f? It should be fine, as far as I know. I tested it before this commit, anyway.

nbeams (commented):
A few more minor comments, but most of my questions/issues have been addressed.

core/factorization/lu.cpp
```diff
@@ -69,6 +69,9 @@ Lu<ValueType, IndexType>::parse(const config::pnode& config,
     if (auto& obj = config.get("skip_sorting")) {
         params.with_skip_sorting(config::get_value<bool>(obj));
     }
+    if (auto& obj = config.get("has_all_fillin")) {
+        params.with_has_all_fillin(config::get_value<bool>(obj));
```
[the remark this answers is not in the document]
yhmtsai: sounds good to me.
Do you think it will be a bit misleading for the case we were discussing, the factorization does not contains all values from the matrices?

include/ginkgo/core/factorization/lu.hpp
```diff
@@ -97,6 +97,16 @@ class Lu
...
         bool GKO_FACTORY_PARAMETER_SCALAR(skip_sorting, false);
+
+        /**
+         * The symbolic factorization contains the fill-in for the matrix. If it
+         * does not have full fill-in, as in Ilu, this parameter must be set to
+         * false in order to avoid the possibility of hanging or illegal memory
+         * accesses during the factorization process. When this is true, the
+         * symbolic factorization must still contain the non-zero locations in
+         * the original matrix, at minimum.
+         */
```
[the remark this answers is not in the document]
yhmtsai: the later part is wrong. it tries to describe the information when the setting is true.
I have updated and reorganize it.

core/factorization/lu.cpp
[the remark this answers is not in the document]
nbeams: Also, based on the update comment in `lu.hpp` (the above coversation), a factorization that doesn't contain all the locations from the original matrix currently isn't allowed, anyway, right?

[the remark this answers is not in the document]
yhmtsai: right. The current version can not work with the factorization misses some entries from the original matrix. I just wonder whether should consider the situation when the factorization does not contains the nonzeros from the original matrix. To do so, we also need to adapt it in the `initialize` function. When we allow thissituation, we definitely do not want to introduce another option to enable it.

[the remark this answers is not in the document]
nbeams: Oh, I see -- just looking to the future. I don't think it's a problem, personally -- I think "fill-in" is pretty well-understood to mean "all the additional locations of the factorization vs original matrix" so I don't think there'd be any confusion that it's meant to differentiate between incomplete factorization with or without all the original matrix locations.

include/ginkgo/core/factorization/cholesky.hpp
```diff
@@ -67,6 +67,17 @@ class Cholesky
...
+        /**
+         * If the user provides the symbolic factorization, it should contain
+         * the fill-in for the matrix. i.e., When this is true, the symbolic
+         * factorization must contain the non-zero locations in the original
+         * matrix and the corresponding fill-in locations during factorization.
+         * If it does not have full fill-in, as in Ilu, this parameter must be
+         * set to false in order to avoid the possibility of hanging or illegal
+         * memory accesses during the factorization process.
+         */
+        bool GKO_FACTORY_PARAMETER_SCALAR(full_fillin, true);
```
upsj, line 80: Why this parameter? This goes against the name LU, and could as well be handled by ILU, if we hand in the sparsity pattern there as a new parameter.

include/ginkgo/core/factorization/lu.hpp
```diff
@@ -97,6 +97,19 @@ class Lu
...
+         * the fill-in for the matrix. i.e., When this is true, the symbolic
+         * factorization must contain the non-zero locations in the original
+         * matrix and the corresponding fill-in locations during factorization.
+         * If it does not have full fill-in, as in Ilu, this parameter must be
+         * set to false in order to avoid the possibility of hanging or illegal
+         * memory accesses during the factorization process. Also, the symbolic
+         * factorization still needs to contain all entries from the original
+         * matrix.
+         */
+        bool GKO_FACTORY_PARAMETER_SCALAR(full_fillin, true);
```
upsj, line 112: Same here

include/ginkgo/core/factorization/cholesky.hpp
yhmtsai: does it go against LU? I try to avoid copying half of functions from lu to ilu.
We do not need the sparsity pattern in ilu currently because we use the same sparsity of the matrix

upsj: In its current form they use the same algorithm, but adding this parameter prevents future evolution taking advantage of structural properties of an LU that an ILU doesn't have.

yhmtsai: could you elaborate more?
@nbeams I assume you use the ILU on your side now not the LU interface, right?
I will copy some function into ILU.
It should still work with ILU-only interface for the case we discussed above in the future

yhmtsai: have copied

include/ginkgo/core/factorization/incompleted_factorization.hpp
nbeams, line 1: Is there a reason why this is named `incompleted_factorization.hpp` and not `incomplete_factorization.hpp`?

reference/test/factorization/lu_kernels.cpp
```diff
@@ -216,7 +218,7 @@ TYPED_TEST(Lu, KernelFactorizeWorks)
         gko::kernels::reference::lu_factorization::factorize(
             this->ref, this->storage_offsets.get_const_data(),
             this->row_descs.get_const_data(), this->storage.get_const_data(),
-            diag_idxs.get_const_data(), this->mtx_lu.get(), tmp);
+            diag_idxs.get_const_data(), this->mtx_lu.get(), false, tmp);
```
nbeams, line 221: Since this is an `Lu` test, shouldn't the next-to-last parameter be `true`? (`full_fillin` = `true`)?

reference/test/factorization/lu_kernels.cpp
```diff
@@ -36,6 +36,8 @@ class Lu : public ::testing::Test {
     using index_type =
         typename std::tuple_element<1, decltype(ValueIndexType())>::type;
     using matrix_type = gko::matrix::Csr<value_type, index_type>;
+    using sparsity_pattern_type =
```
nbeams, line 39: Is this used anywhere anymore? I don't see it anywhere else in the file.

test/factorization/lu_kernels.cpp
```diff
@@ -194,11 +196,11 @@ TYPED_TEST(Lu, KernelFactorizeIsEquivalentToRef)
         gko::kernels::reference::lu_factorization::factorize(
             this->ref, this->storage_offsets.get_const_data(),
             this->row_descs.get_const_data(), this->storage.get_const_data(),
-            diag_idxs.get_const_data(), this->mtx_lu.get(), tmp);
+            diag_idxs.get_const_data(), this->mtx_lu.get(), false, tmp);
         gko::kernels::GKO_DEVICE_NAMESPACE::lu_factorization::factorize(
             this->exec, this->dstorage_offsets.get_const_data(),
             this->drow_descs.get_const_data(), this->dstorage.get_const_data(),
-            ddiag_idxs.get_const_data(), this->dmtx_lu.get(), dtmp);
+            ddiag_idxs.get_const_data(), this->dmtx_lu.get(), false, dtmp);
```
nbeams, line 203: Similar to the other LU test comment -- shouldn't these `false`s be `true`s?

include/ginkgo/core/factorization/incompleted_factorization.hpp
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * An enum class for algorithm selection in the incomplete factorization.
+ * `sparselib` is only available for CUDA and HIP.
+ * `syncfree` is Ginkgo's implementation through the Lu/Cholesky factorization
+ * with given sparsity.
+ */
+enum class factorize_algorithm { sparselib, syncfree };
+
+
```
nbeams, line 21: Isn't it confusing for this to be in its own incomplete factorization file but still just be in the `factorization` namespace and be named `factorize_algorithm`? Nothing in the name `gko::factorization::factorize_algorithm` indicates that it should be in its own file dedicated to incomplete factorizations.

nbeams (commented):
Looking good, but I have some more minor comments.

core/factorization/ic.cpp
```diff
@@ -79,15 +103,75 @@ std::unique_ptr<Composition<ValueType>> Ic<ValueType, IndexType>::generate(
     exec->run(ic_factorization::make_add_diagonal_elements(
         local_system_matrix.get(), false));
 
+    std::shared_ptr<const matrix_type> ic;
     // Compute LC factorization
-    exec->run(ic_factorization::make_compute(local_system_matrix.get()));
+    if (std::dynamic_pointer_cast<const OmpExecutor>(exec) ||
+        parameters_.algorithm == factorize_algorithm::syncfree) {
+        std::unique_ptr<gko::factorization::elimination_forest<IndexType>>
```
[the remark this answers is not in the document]
yhmtsai: I think it would make sense we use the syncfree as default not sparselib.
some default value not working sounds a bit weird to me.

core/factorization/ic.cpp
```diff
@@ -30,6 +36,13 @@ GKO_REGISTER_OPERATION(add_diagonal_elements,
 GKO_REGISTER_OPERATION(initialize_row_ptrs_l,
                        factorization::initialize_row_ptrs_l);
 GKO_REGISTER_OPERATION(initialize_l, factorization::initialize_l);
+// for gko syncfree implementation
```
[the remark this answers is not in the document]
yhmtsai: I think `sparselib_ic` is good because sparselib indeed completes everything for that

include/ginkgo/core/factorization/incompleted_factorization.hpp
yhmtsai: no, I forgot incomplete is already adj

include/ginkgo/core/factorization/incompleted_factorization.hpp
yhmtsai: I do not have a good idea other than `incomplete_factorize_algorithm`.
It is a bit long but should be descriptive.

upsj: How about `factorization::incomplete_algorithm`? Otherweise factorize would be redundant

core/factorization/ic.cpp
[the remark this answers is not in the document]
yhmtsai: I think ginkgo can be kind of sparselib for itself? We have that for reference check.

test/factorization/lu_kernels.cpp
yhmtsai: for some reason, I keep using it in reversed way

core/factorization/ic.cpp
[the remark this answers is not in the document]
nbeams: Personally, I'm not too keen on having "sparselib" mean two different things: an external library for CUDA/HIP, and Ginkgo itself for OpenMP(/Reference). Also I realize I got my comment wrong above: I meant "if you tried to NOT use `syncfree`". 
I guess my main concern is the fact that the parameter is basically meaningless for OpenMP -- but if you use it, you don't get a warning that you selected a meaningless option. Right now, if you try to create an Ilu/Ic preconditioner with OpenMP, it will tell you it's not implemented, right? Why not let it continue to do that in the default case (`sparselib`)?

I agree it would make more sense to have Ginkgo's version as the default once it's added, though I guess it's a bit tricky to change default behavior on users...

include/ginkgo/core/factorization/incompleted_factorization.hpp
nbeams: I like that. Or what about another namespace: `factorization::incomplete::algorithm`?

core/factorization/ic.cpp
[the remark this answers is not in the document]
yhmtsai: Yes, I have raised it in our meeting. We keep the original behavior: default is stillsparselib, but omp will failed.
We throw a descriptive error message to indicate user should use syncfree for omp executor.
I have updated it

include/ginkgo/core/factorization/incompleted_factorization.hpp
yhmtsai: I currently went for `incomplete_algorithm`.
I do not have strong preference on both.

nbeams (approved):

common/cuda_hip/factorization/cholesky_kernels.cpp
```diff
@@ -200,9 +202,16 @@ __global__ __launch_bounds__(default_block_size) void factorize(
             const auto upper_col = cols[upper_nz];
             if (upper_col >= row) {
                 const auto upper_val = vals[upper_nz];
-                const auto output_pos =
-                    lookup.lookup_unsafe(upper_col) + row_begin;
-                vals[output_pos] -= scale * upper_val;
+                if (!full_fillin) {
```
upsj, line 205: if constexpr?

common/cuda_hip/factorization/lu_kernels.cpp
```diff
@@ -130,8 +130,16 @@ __global__ __launch_bounds__(default_block_size) void factorize(
              upper_nz += config::warp_size) {
             const auto upper_col = cols[upper_nz];
             const auto upper_val = vals[upper_nz];
-            const auto output_pos = lookup.lookup_unsafe(upper_col) + row_begin;
-            vals[output_pos] -= scale * upper_val;
+            if (!full_fillin) {
```
upsj, line 133: same here

omp/factorization/cholesky_kernels.cpp
```diff
@@ -233,8 +233,15 @@ void factorize(std::shared_ptr<const DefaultExecutor> exec,
                 const auto col = cols[dep_nz];
                 if (col < row) {
                     const auto val = vals[dep_nz];
-                    const auto nz = row_begin + lookup.lookup_unsafe(col);
-                    vals[nz] -= scale * val;
+                    if (!full_fillin) {
```
upsj, line 236: if constexpr

reference/factorization/cholesky_kernels.cpp
```diff
@@ -204,8 +204,15 @@ void factorize(std::shared_ptr<const DefaultExecutor> exec,
                 const auto col = cols[dep_nz];
                 if (col < row) {
                     const auto val = vals[dep_nz];
-                    const auto nz = row_begin + lookup.lookup_unsafe(col);
-                    vals[nz] -= scale * val;
+                    if (!full_fillin) {
```
upsj, line 207: if constexpr

reference/factorization/lu_kernels.cpp
```diff
@@ -87,8 +87,15 @@ void factorize(std::shared_ptr<const DefaultExecutor> exec,
             for (auto dep_nz = dep_diag_idx + 1; dep_nz < dep_end; dep_nz++) {
                 const auto col = cols[dep_nz];
                 const auto val = vals[dep_nz];
-                const auto nz = row_begin + lookup.lookup_unsafe(col);
-                vals[nz] -= scale * val;
+                if (!full_fillin) {
```
upsj, line 90: if constexpr

upsj (commented):
I'm not 100% convinced it's a good choice to put both ILU and LU into the same gko::kernels::... function (It's fine if they use the same kernel, just the function itself), since the interface to LU will probably change significantly in the future. Otherwise only the suggestion of using `if constexpr` where it makes sense

yhmtsai:
I don't get the meaning about putting into `gko::kernels`, I think all kernel functions are in the namespace.
Could you elaborate it more?
I do not think this change give any changes to LU. It only touches ILU interface now.
Also, LU in experimental, although it is not good, we can still change the LU interface.
I would also like to avoid unclear future to stop this pr. There are a few pr hang by that comment, but they are resolved in long time more than one year.

upsj:
instead of adding a new parameter to factorize, forcing me to split that function later on when I implement more LU-specific preprocessing and optimizations, split it into `factorize` and `incomplete_factorize`. For now they use the same kernel, in the future they will likely not.

omp/factorization/cholesky_kernels.cpp
yhmtsai: this is not template, so I can not use constexpr

reference/factorization/cholesky_kernels.cpp
yhmtsai: this is not template, so I can not use constexpr

reference/factorization/lu_kernels.cpp
yhmtsai: this is not template, so I can not use constexpr

upsj (approved):
I'll likely have to split up the `factorize` function into two different kernels without `full_fillin` very soon, but I won't hold up the PR any further. Thanks for the updates

yhmtsai:
@upsj Yes, that makes sense if these kernels can not share most of part

PR #1685, opened 2024-10-02
Disable horeka CI jobs

MarcelKoch:
This PR disables CI jobs on horeka. In time for the release I would like to have our CI pass all stages.

Edit:
This also deletes the windows github jobs, although it doesn't fit the original PR topic ideally.


upsj:
This should probably be rebased. Otherwise LGTM

yhmtsai (approved):
If we do not have project quota issue in the future, then simple is always better.
I think you can change the title, if you feel the changes does not fit the title

PR #1688, opened 2024-10-09
Add Distributed VectorCache and use it for intermediate vector in and adapt `apply_uses_initial_guess` in Schwarz 

yhmtsai:
This adds a cache/workspace for the intermediate vector in advanced_apply of Schwarz.
Also, it makes apply_uses_initial_guess rely on the `local_solver->apply_uses_initial_guess()`.
We might need to revisit this function after adding refining step in Schwarz.

### DIFF
--- a/core/base/vector_cache.cpp
+++ b/core/base/vector_cache.cpp
@@ -1,0 +1,55 @@
+// SPDX-FileCopyrightText: 2017 - 2024 The Ginkgo authors
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include "ginkgo/core/base/vector_cache.hpp"
+
+#include <ginkgo/core/base/mpi.hpp>
+#include <ginkgo/core/distributed/vector.hpp>
+
+namespace gko {
+namespace detail {
+
+
+template <typename ValueType>
+void VectorCache<ValueType>::init(std::shared_ptr<const Executor> exec,
+                                  gko::experimental::mpi::communicator comm,
+                                  dim<2> global_size, dim<2> local_size) const
+{
+    if (!vec || vec->get_size() != global_size || vec->get_executor() != exec) {
+        vec = experimental::distributed::Vector<ValueType>::create(
+            exec, comm, global_size, local_size);
+    } else if (vec->get_local_vector()->get_size() != local_size) {
+        // handle locally to eliminate the mpi call
+        vec->local_ =
+            std::move(gko::matrix::Dense<ValueType>(exec, local_size));
+    }
+}
+
+
+template <typename ValueType>
+void VectorCache<ValueType>::init_from(
+    const experimental::distributed::Vector<ValueType>* template_vec) const
+{
+    if (!vec || vec->get_size() != template_vec->get_size() ||
+        vec->get_executor() != template_vec->get_executor()) {
+        vec =
+            experimental::distributed::Vector<ValueType>::create_with_config_of(
+                template_vec);
+    } else if (vec->get_local_vector()->get_size() !=
+               template_vec->get_local_vector()->get_size()) {
+        // handle locally to eliminate the mpi call
+        vec->local_ = std::move(gko::matrix::Dense<ValueType>(
+            template_vec->get_executor(),
+            template_vec->get_local_vector()->get_size(),
+            template_vec->get_local_vector()->get_stride()));
+    }
+}
+
+
+#define GKO_DECLARE_VECTOR_CACHE(_type) class VectorCache<_type>
+GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE(GKO_DECLARE_VECTOR_CACHE);
+
+
+}  // namespace detail
+}  // namespace gko
--- a/core/distributed/preconditioner/schwarz.cpp
+++ b/core/distributed/preconditioner/schwarz.cpp
@@ -51,6 +51,13 @@ Schwarz<ValueType, LocalIndexType, GlobalIndexType>::parse(
     return params;
 }
 
+template <typename ValueType, typename LocalIndexType, typename GlobalIndexType>
+bool Schwarz<ValueType, LocalIndexType,
+             GlobalIndexType>::apply_uses_initial_guess() const
+{
+    return this->local_solver_->apply_uses_initial_guess();
+}
+
 
 template <typename ValueType, typename LocalIndexType, typename GlobalIndexType>
 void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::apply_impl(
--- a/core/distributed/preconditioner/schwarz.cpp
+++ b/core/distributed/preconditioner/schwarz.cpp
@@ -84,12 +84,24 @@ void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::apply_impl(
 {
     precision_dispatch_real_complex_distributed<ValueType>(
         [this](auto dense_alpha, auto dense_b, auto dense_beta, auto dense_x) {
-            auto x_clone = dense_x->clone();
-            this->apply_dense_impl(dense_b, x_clone.get());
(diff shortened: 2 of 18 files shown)


core/distributed/preconditioner/schwarz.cpp
```diff
@@ -84,15 +91,27 @@ void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::apply_impl(
...
-            dense_x->add_scaled(dense_alpha, x_clone.get());
+            dense_x->add_scaled(dense_alpha, cache_.intermediate.get());
         },
         alpha, b, beta, x);
 }
 
 
+template <typename ValueType, typename LocalIndexType, typename GlobalIndexType>
+template <typename VectorType>
+void Schwarz<ValueType, LocalIndexType, GlobalIndexType>::set_cache_to(
```
MarcelKoch, line 105: You only need this function once, right? Then I would prefer to just add this directly, it's not that much IMO.

include/ginkgo/core/distributed/preconditioner/schwarz.hpp
```diff
@@ -150,6 +169,26 @@ class Schwarz
...
+     * Manages a vector as a cache, so there is no need to allocate one every
+     * time an intermediate vector is required.
+     * Copying an instance will only yield an empty object since copying the
+     * cached vector would not make sense.
+     *
+     * @internal  The struct is present so the whole class can be copyable
+     *            (could also be done with writing `operator=` and copy
+     *            constructor of the enclosing class by hand)
+     */
+    mutable struct cache_struct {
```
MarcelKoch, line 183: I think this should be a `DenseCache` instead.

MarcelKoch (changes requested):

pratikvn (approved):
I agree with the Marcel that it can be a `DenseCache` instead of `cache_struct`. But otherwise looks good to me. I approve it as Marcel has already requested that change.

include/ginkgo/core/distributed/preconditioner/schwarz.hpp
yhmtsai: I think it does not support distributed::Vector.

MarcelKoch: you're right. Perhaps it would make sense to add `VectorCache`, instead of using the cache construct.

core/base/vector_cache.cpp
```diff
@@ -0,0 +1,56 @@
...
+void VectorCache<ValueType>::init(std::shared_ptr<const Executor> exec,
+                                  gko::experimental::mpi::communicator comm,
+                                  dim<2> global_size, dim<2> local_size) const
+{
+    if (!vec || vec->get_size() != global_size || vec->get_executor() != exec) {
+        vec = experimental::distributed::Vector<ValueType>::create(
+            exec, comm, global_size, local_size);
+    } else if (vec->get_local_vector()->get_size() != local_size) {
+        // handle locally to eliminate the mpi call
+        std::cout << "local diff" << comm.rank() << std::endl;
```
pratikvn, line 24:
```suggestion
```

core/base/vector_cache.cpp
```diff
@@ -0,0 +1,56 @@
...
+        vec->local_ =
+            std::move(gko::matrix::Dense<ValueType>(exec, local_size));
```
pratikvn, line 26: nit: use the create function instead ?

core/base/vector_cache.cpp
```diff
@@ -0,0 +1,56 @@
...
+{
+    if (!vec || vec->get_size() != template_vec->get_size() ||
+        vec->get_executor() != template_vec->get_executor()) {
+        vec =
+            experimental::distributed::Vector<ValueType>::create_with_config_of(
+                template_vec);
+    } else if (vec->get_local_vector()->get_size() !=
+               template_vec->get_local_vector()->get_size()) {
+        // handle locally to eliminate the mpi call
+        vec->local_ = std::move(gko::matrix::Dense<ValueType>(
```
pratikvn, line 43: nit: use the create function instead ?

include/ginkgo/core/matrix/dense.hpp
```diff
@@ -122,6 +132,7 @@ class Dense
     friend class SparsityCsr<ValueType, int64>;
     friend class Dense<to_complex<ValueType>>;
     friend class experimental::distributed::Vector<ValueType>;
+    friend class gko::detail::VectorCache<ValueType>;
```
MarcelKoch, line 135: why is this necessary?

MarcelKoch (approved):

pratikvn (approved):
Nice work! LGTM!

include/ginkgo/core/base/vector_cache.hpp
```diff
@@ -0,0 +1,114 @@
...
+#include <ginkgo/config.hpp>
+#include <ginkgo/core/base/executor.hpp>
+#include <ginkgo/core/base/mpi.hpp>
+#include <ginkgo/core/distributed/vector.hpp>
+
+
+#if GINKGO_BUILD_MPI
+
+
+namespace gko {
```
pratikvn, line 20: I think it might make sense to put this into the `experimental` and `distributed` namespaces.

include/ginkgo/core/base/vector_cache.hpp
```diff
@@ -0,0 +1,114 @@
...
+        return vec.get();
+    }
+};
+
+
+#endif
+
+
+}  // namespace detail
+}  // namespace gko
```
pratikvn, line 111:
```suggestion
}  // namespace detail
}  // namespace gko


#endif
```

pratikvn (changes requested):
Some fixes needed

include/ginkgo/core/matrix/dense.hpp
yhmtsai: Because `distributed::Vector` holds the `Dense` object not `Dense` pointer, I need the constructor to construct the object directly and then use the move_to

core/base/vector_cache.cpp
yhmtsai: We hold the object not the pointer.
However, I can create the unique pointer and release the object out actually, right?

include/ginkgo/core/matrix/dense.hpp
yhmtsai: ah, I might use the `unique_ptr.release` from create function

MarcelKoch: No it's fine, I just got confused. I thought this was in the `distributed::Matrix`.

include/ginkgo/core/base/vector_cache.hpp
yhmtsai: I move it into `gko::experimental::distributed::detail`

core/base/vector_cache.cpp
pratikvn: No, It is a only a nit from my side. Using the constructor should also be okay.

pratikvn (approved):

include/ginkgo/core/distributed/preconditioner/schwarz.hpp
```diff
@@ -150,6 +161,8 @@ class Schwarz
     void set_solver(std::shared_ptr<const LinOp> new_solver);
 
     std::shared_ptr<const LinOp> local_solver_;
+
+    gko::experimental::distributed::detail::VectorCache<ValueType> cache_;
```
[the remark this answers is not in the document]
yhmtsai: it works. I did not test it before uploading, so I pick the one definitely works.
However, we need to use `gko::detail::...` not `detail::...` to call the function when in distributed namespace.
compiler only searches `distributed::detail` not `gko::detail`

PR #1694, opened 2024-10-14
Move the majority of jobs to the full pipeline

MarcelKoch:
This PR moves the majority of the gitlab jobs to the full pipeline.

IMO our CI could use a more thourough restructuring, since the job properties are very random, but that should be done later.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -92,7 +92,7 @@ trigger_pipeline:
 
 
 # cuda 11.4 and friends
-build/cuda114/nompi/gcc/cuda/debug/shared:
+build/cuda114/nompi/gcc/cuda/release/shared:
   extends:
     - .build_and_test_template
     - .default_variables
@@ -101,7 +101,7 @@ build/cuda114/nompi/gcc/cuda/debug/shared:
   variables:
     BUILD_OMP: "ON"
     BUILD_CUDA: "ON"
-    BUILD_TYPE: "Debug"
+    BUILD_TYPE: "Release"
     FAST_TESTS: "ON"
     # fix gtest issue https://github.com/google/googletest/issues/3514
     CXX_FLAGS: "-Wno-error=maybe-uninitialized"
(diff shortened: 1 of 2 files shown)


yhmtsai (approved):
I would still prefer the original grouping strategy
The configuration under same compiler version are slightly different, so grouping them together might be easier to configure or distinguish them.

MarcelKoch:
@yhmtsai can you explain the original grouping strategy? For me all options, except for device+compiler, seem to be choosen very randomly. I.e. wether it's debug or release, shared or static, fast test enabled or not, .... This became especially clear for the #1536 PR.

yhmtsai:
I mean grouping by device+compiler.
for example, we might have several configuration under cuda 12.2.
we might only adjust a bit for debug/release or some minor configuration settings.
putting them together makes it easier to see minor changes than separating them partially in quick test section and partially in full test.

pratikvn:
I think it would make sense to group jobs properly to be able to navigate and access them easily. The main groups could be `cuda`, `hip`, `sycl`, `mpi`. In each of these groups, the number of jobs should be fewer. I dont know if it is possible to be subgroups within groups, but gitlab does allow for creation of groups atleast: https://docs.gitlab.com/ee/ci/jobs/#group-jobs-in-a-pipeline

pratikvn (approved):

MarcelKoch:
@yhmtsai that grouping is still there, just now it's under the full pipeline. I don't think it's reasonable to have our current 'quick' pipeline run with 15+ jobs, while our 'full' pipeline has ~10 jobs. The quick pipeline is always run, even if the PR is still a bit under development. The 15+ jobs are not needed during that phase, so the CI is just clogged with meaningless tests.

MarcelKoch:
@pratikvn That looks like a nice feature.

Sidenote: I would not count `mpi` as separated from the other groups, instead `mpi` should be enabled in almost all of our jobs.

yhmtsai (approved):

PR #1701, opened 2024-10-17
Fix parameter ordering/naming for reference (omp) batch CG

MarcelKoch:
This PR fixes the parameter ordering of the `initialize` function for the reference+omp implementation of the batch CG. Additionally, a parameter name in `update_x_and_r` is changed to better reflect the actually used variable.

### DIFF
--- a/reference/solver/batch_cg_kernels.hpp
+++ b/reference/solver/batch_cg_kernels.hpp
@@ -32,8 +32,8 @@ inline void initialize(
     const BatchMatrixType_entry& A_entry,
     const gko::batch::multi_vector::batch_item<const ValueType>& b_entry,
     const gko::batch::multi_vector::batch_item<const ValueType>& x_entry,
-    const gko::batch::multi_vector::batch_item<ValueType>& rho_new_entry,
     const gko::batch::multi_vector::batch_item<ValueType>& rho_old_entry,
+    const gko::batch::multi_vector::batch_item<ValueType>& rho_new_entry,
     const gko::batch::multi_vector::batch_item<ValueType>& r_entry,
     const gko::batch::multi_vector::batch_item<ValueType>& p_entry,
     const gko::batch::multi_vector::batch_item<ValueType>& z_entry,
@@ -86,7 +86,7 @@ inline void update_p(
 
 template <typename ValueType>
 inline void update_x_and_r(
-    const gko::batch::multi_vector::batch_item<const ValueType>& rho_old_entry,
+    const gko::batch::multi_vector::batch_item<const ValueType>& rho_new_entry,
     const gko::batch::multi_vector::batch_item<const ValueType>& p_entry,
     const gko::batch::multi_vector::batch_item<const ValueType>& Ap_entry,
     const gko::batch::multi_vector::batch_item<ValueType>& alpha_entry,
@@ -96,7 +96,7 @@ inline void update_x_and_r(
     batch_single_kernels::compute_conj_dot_product_kernel<ValueType>(
         p_entry, Ap_entry, alpha_entry);
 
-    const ValueType temp = rho_old_entry.values[0] / alpha_entry.values[0];
+    const ValueType temp = rho_new_entry.values[0] / alpha_entry.values[0];
     for (int row = 0; row < r_entry.num_rows; row++) {
         x_entry.values[row * x_entry.stride] +=
             temp * p_entry.values[row * p_entry.stride];


reference/solver/batch_cg_kernels.hpp
```diff
@@ -96,7 +96,7 @@ inline void update_x_and_r(
     batch_single_kernels::compute_conj_dot_product_kernel<ValueType>(
         p_entry, Ap_entry, alpha_entry);
 
-    const ValueType temp = rho_old_entry.values[0] / alpha_entry.values[0];
+    const ValueType temp = rho_new_entry.values[0] / alpha_entry.values[0];
```
pratikvn, line 99: I dont think this is accurate. In this case here, `rho_old` is $<r_{j}, z_{j}>$, while `rho_new` is $<r_{j+1}, z_{j+1}>$. And `temp` here is equivalent to $\alpha_j$ in Algorithm 9.1 in Saad's book for example. So I think the current state in `develop` should be correct.

pratikvn (changes requested):

reference/solver/batch_cg_kernels.hpp
MarcelKoch: But we are passing `rho_new_entry` to this function. Maybe `rho_new` means $<r_j, z_j>$ and `rho_old` means $<r_{j-1}, z_j>$?

pratikvn: No, I think the order of the function calls needs to be re-arranged. From looking at Saad's book, I think here is the diff that should be correct. The CUDA/HIP/DPCPP kernels should be correct. 

```diff
diff --git a/reference/solver/batch_cg_kernels.hpp b/reference/solver/batch_cg_kernels.hpp
index d4a35e3d0..2c0309478 100644
--- a/reference/solver/batch_cg_kernels.hpp
+++ b/reference/solver/batch_cg_kernels.hpp
@@ -199,6 +199,10 @@ inline void batch_entry_cg_impl(
                  gko::batch::to_const(rho_old_entry),
                  gko::batch::to_const(z_entry), p_entry);
 
+        // rho_old = rho_new
+        batch_single_kernels::copy_kernel(gko::batch::to_const(rho_new_entry),
+                                          rho_old_entry);
+
         // Ap = A * p
         batch_single_kernels::simple_apply(
             A_entry, gko::batch::to_const(p_entry), Ap_entry);
@@ -207,12 +211,8 @@ inline void batch_entry_cg_impl(
         // x = x + temp * p
         // r = r - temp * Ap
         update_x_and_r(
-            gko::batch::to_const(rho_new_entry), gko::batch::to_const(p_entry),
+            gko::batch::to_const(rho_old_entry), gko::batch::to_const(p_entry),
             gko::batch::to_const(Ap_entry), alpha_entry, x_entry, r_entry);
-
-        // rho_old = rho_new
-        batch_single_kernels::copy_kernel(gko::batch::to_const(rho_new_entry),
-                                          rho_old_entry);
     }
 
     logger.log_iteration(batch_item_id, iter, res_norms_entry.values[0]);
```

yhmtsai: No, I think the original is correct.
the implementation is rotated a bit because we put the criterion right after updating r.
we move the rest after updating r in the begining of the loop by different initial value of p.

Use preconditioned algorithm in https://en.wikipedia.org/wiki/Conjugate_gradient_method 
image
we move the code after `if` to the beginning, so we do not need preconditioner and p = z out of loop in the picture by setting the p = 0 -> we get p = z in the first iteration.

MarcelKoch: This function is called with `rho_new_entry`. This is the same as for the non-batched CG, and it matches @yhmtsai explanation. The cuda/hip version is differently ordered, since the initialization is different (`rho_old` is set to $<r_0, z_0>$ instead of `1`).
So I think calling the function with `rho_new_entry` is correct, and the parameter name should be changed here.

pratikvn: Yes, you are right. Disregard my comments.

pratikvn (approved):

yhmtsai (approved):
LGTM.
one question might to @pratikvn : is there any reason that the device side follows the another algorithm more?

pratikvn:
I think we re-arranged it so that the convergence check is first. But I dont see a particular reason for that. I will maybe compare the performance for both and maybe unify the device side to the reference one as well.

PR #1706, opened 2024-10-22
Half base type

yhmtsai:
This pr moves the half out of extended_float.hpp with some modification and half is available in the public interface.

Currently, we can still mark some conversion/operation with GKO_ATTRIBUTE and GKO_INLINE.
However, I think it is good to ensure we only use the vendor's impl on device side.
To achieve it, Jacobi needs to use __half not half now.
This also removes the undefined behavior from `reinterpret_cast` to `std::memcpy`

~~one question: should we mark the half precision in experimental namespace?~~ in usual interface from discussion


include/ginkgo/core/base/half.hpp
```diff
@@ -0,0 +1,671 @@
...
+    // caused by something else in jacobi or isai.
+    constexpr half() noexcept : data_(0){};
+
+    template <typename T, typename = std::enable_if_t<std::is_scalar<T>::value>>
+    half(const T val) : data_(0)
+    {
+        this->float2half(static_cast<float>(val));
+    }
+
+    constexpr half(const half& val) : data_(0) { data_ = val.data_; };
```
MarcelKoch, line 302: does this have any advantage over
```suggestion
    constexpr half(const half& val) : data_(val.data_) { }
```

include/ginkgo/core/base/half.hpp
```diff
@@ -0,0 +1,671 @@
...
+        const auto bits = half2float(data_);
+        float ans(0);
+        std::memcpy(&ans, &bits, sizeof(float));
+        return ans;
+    }
+
+    // can not use half operator _op(const half) for half + half
+    // operation will cast it to float and then do float operation such that it
+    // becomes float in the end.
+#define HALF_OPERATOR(_op, _opeq)                                  \
```
MarcelKoch, line 322: best to undefine this and the macro below.

core/test/base/half.cpp
```diff
@@ -0,0 +1,285 @@
...
+    using type = float;
+};
+
+template <>
+struct floating_impl<64> {
+    using type = double;
+};
+
+template <std::size_t N>
+using floating = typename floating_impl<N>::type;
```
MarcelKoch, line 33: This is duplicated from `extended_float.cpp`. Maybe move this into a `floating.hpp` or `bit_helpers.hpp`, together with the `create_from_bits` and `get_bits` below.

MarcelKoch (approved):
Just to reiterate, I'm in favor of putting this directly into the `gko` namespace.

include/ginkgo/core/base/half.hpp
yhmtsai: no, it was changed step by step.
fixed it

yhmtsai:
@MarcelKoch I have made the half without types.hpp dependence such that we can include it directly without circular dependence and the instantiate function have the definition of half now

MarcelKoch (approved):

core/test/base/floating_bit_helper.hpp
```diff
@@ -0,0 +1,82 @@
...
+    using type = double;
+};
+
+
+template <std::size_t N>
+using floating = typename floating_impl<N>::type;
+
+
+template <std::size_t N>
+static floating<N - 1> create_from_bits(const char (&s)[N])
```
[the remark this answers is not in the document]
yhmtsai: good catch!

core/test/base/floating_bit_helper.hpp
```diff
@@ -0,0 +1,82 @@
...
+template <std::size_t N>
+static std::bitset<N - 1> get_bits(const char (&s)[N])
+{
+    return std::bitset<N - 1>(s);
+}
+
+
+}  // namespace floating_bit_helper
+
+#endif  // GKO_CORE_TEST_BASE_FLOATING_BIT_HELPER_HPP_
```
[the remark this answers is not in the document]
yhmtsai: I adds the end-of-file-fixer from pre-commit-hook to fix it now, too 😉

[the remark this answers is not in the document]
greole: just out of curiosity, why don't we use `#pragma once` as header guard?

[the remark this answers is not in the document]
yhmtsai: I guess the reason behind that is `#pragma once` is not the standard and need to rely on the compiler implementation

include/ginkgo/core/base/half.hpp
```diff
@@ -0,0 +1,683 @@
...
+    // caused by something else in jacobi or isai.
+    constexpr half() noexcept : data_(0){};
+
+    template <typename T, typename = std::enable_if_t<std::is_scalar<T>::value>>
+    half(const T val) : data_(0)
+    {
+        this->float2half(static_cast<float>(val));
+    }
+
+    constexpr half(const half& val) : data_(val.data_){};
```
MarcelKoch, line 307:
```suggestion
```
This has to be removed (or at least defaulted), because otherwise this class would not be trivially copyable, see https://en.cppreference.com/w/cpp/language/classes#Trivially_copyable_class and https://en.cppreference.com/w/cpp/language/copy_constructor#Trivial_copy_constructor. 
If the class is not trivially copyable, then `std::memcpy` may be undefined behavior.

MarcelKoch (changes requested):

MarcelKoch (approved):

thoasm (approved):
LGTM, only some nits

accessor/cuda_helper.hpp
```diff
@@ -17,7 +17,15 @@
 #include "utils.hpp"
 
 
+struct __half;
```
upsj, line 20: I don't think there's a way around the issue, but we should be aware that `__half` is a reserved identifier according to the C++ standard, so we should technically not be defining anything with that name. Also there might be a tiny potential of name mangling issues if this is ever handled as a `class` instead of a `struct` with MSVC.

core/test/base/half.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+// clang-format does terrible formatting of string literal concatenation
+// clang-format off
+
+
+TEST(FloatToHalf, ConvertsOne)
+{
+    half x = create_from_bits("0" "01111111" "00000000000000000000000");
+
+    ASSERT_EQ(get_bits(x), get_bits("0" "01111" "0000000000"));
```
upsj, line 32: FYI, since C++14 we can use binary literals `0b01111111`

cuda/base/types.hpp
```diff
@@ -124,6 +121,17 @@ struct culibs_type_impl<std::complex<double>> {
...
 
+
+template <>
+struct culibs_type_impl<half> {
+    using type = __half;
+};
+
+template <>
+struct culibs_type_impl<std::complex<half>> {
+    using type = __half2;
```
upsj, line 132: Just for my understanding, `__half2` is a vector type derived from `__half`?

dev_tools/scripts/gdb-ginkgo.py
```diff
@@ -220,12 +225,28 @@ def display_hint(self):
...
 
 
+class GkoHalfPrinter:
+    "Print a gko::half"
+
+    def __init__(self, val):
+        # GDB doesn't seem to consider the user-defined conversion in its Value.cast,
+        # so we need to call the conversion operator explicitly
+        address = hex(val.address)
+        self.float_val = gdb.parse_and_eval(f"reinterpret_cast<gko::half*>({address})->operator float()")
```
upsj, line 235: If that function is ever not exported, we may need to implement an xfunction for it

.pre-commit-config.yaml
```diff
@@ -1,4 +1,8 @@
 repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v5.0.0
+    hooks:
+    -   id: end-of-file-fixer
```
upsj, line 5: great idea! nit:
```suggestion
    - id: end-of-file-fixer
```

cuda/base/types.hpp
yhmtsai: you are right. only one function `__hcmadd` does complex operation, but the other are all vector type operation.

dev_tools/scripts/gdb-ginkgo.py
yhmtsai: I am not familar with that. This is done by @MarcelKoch  
could you implement that?

upsj: We likely don't need it now, just for future reference

cuda/base/types.hpp
yhmtsai: I have check cusparse they use __half2 to hold complex<__half>, too. https://docs.nvidia.com/cuda/cusparse/#cudadatatype-t

PR #1707, opened 2024-10-23
Add local to global index mapping

MarcelKoch:
This PR adds a local-to-global mapping to the `index_map` class. It is necessary for the follow-up PR #1639. This also introduces a device-side view of a partition, and adds another member to the partition which stores the range-ids segmented by their part-id.

(I know that the branch name is wrong...)

### DIFF
--- a/common/cuda_hip/distributed/index_map_kernels.cpp
+++ b/common/cuda_hip/distributed/index_map_kernels.cpp
@@ -302,20 +302,20 @@ void map_to_global(
     device_partition<const LocalIndexType, const GlobalIndexType> partition,
     device_segmented_array<const GlobalIndexType> remote_global_idxs,
     experimental::distributed::comm_index_type rank,
-    const array<LocalIndexType>& local_ids,
+    const array<LocalIndexType>& local_idxs,
     experimental::distributed::index_space is,
-    array<GlobalIndexType>& global_ids)
+    array<GlobalIndexType>& global_idxs)
 {
     auto range_bounds = partition.offsets_begin;
     auto starting_indices = partition.starting_indices_begin;
     const auto& ranges_by_part = partition.ranges_by_part;
-    auto local_ids_it = local_ids.get_const_data();
-    auto input_size = local_ids.get_size();
+    auto local_idxs_it = local_idxs.get_const_data();
+    auto input_size = local_idxs.get_size();
 
     auto policy = thrust_policy(exec);
 
-    global_ids.resize_and_reset(local_ids.get_size());
-    auto global_ids_it = global_ids.get_data();
+    global_idxs.resize_and_reset(local_idxs.get_size());
+    auto global_idxs_it = global_idxs.get_data();
 
     auto map_local = [rank, ranges_by_part, range_bounds, starting_indices,
                       partition] __device__(auto lid) {
@@ -330,11 +330,16 @@ void map_to_global(
         auto local_ranges_size =
             static_cast<int64>(local_ranges.end - local_ranges.begin);
 
-        auto it = binary_search(int64(0), local_ranges_size, [=](const auto i) {
-            return starting_indices[local_ranges.begin[i]] >= lid;
-        });
+        // the binary search finds the first local range, such that the starting
+        // index is larger than lid, thus lid is contained in the local range
+        // before that one
         auto local_range_id =
-            it != local_ranges_size ? it : max(int64(0), it - 1);
+            binary_search(int64(0), local_ranges_size,
+                          [=](const auto i) {
+                              return starting_indices[local_ranges.begin[i]] >
+                                     lid;
+                          }) -
+            1;
         auto range_id = local_ranges.begin[local_range_id];
 
         return static_cast<GlobalIndexType>(lid - starting_indices[range_id]) +
@@ -363,16 +368,16 @@ void map_to_global(
     };
 
     if (is == experimental::distributed::index_space::local) {
-        thrust::transform(policy, local_ids_it, local_ids_it + input_size,
-                          global_ids_it, map_local);
+        thrust::transform(policy, local_idxs_it, local_idxs_it + input_size,
+                          global_idxs_it, map_local);
     }
     if (is == experimental::distributed::index_space::non_local) {
-        thrust::transform(policy, local_ids_it, local_ids_it + input_size,
-                          global_ids_it, map_non_local);
+        thrust::transform(policy, local_idxs_it, local_idxs_it + input_size,
+                          global_idxs_it, map_non_local);
     }
     if (is == experimental::distributed::index_space::combined) {
-        thrust::transform(policy, local_ids_it, local_ids_it + input_size,
-                          global_ids_it, map_combined);
+        thrust::transform(policy, local_idxs_it, local_idxs_it + input_size,
+                          global_idxs_it, map_combined);
     }
 }
 
--- a/core/distributed/device_partition.hpp
+++ b/core/distributed/device_partition.hpp
@@ -2,8 +2,7 @@
 //
 // SPDX-License-Identifier: BSD-3-Clause
 
-#ifndef GINKGO_PARTITION_HPP
-#define GINKGO_PARTITION_HPP
+#pragma once
 
 #include <ginkgo/core/distributed/partition.hpp>
 
@@ -34,34 +33,7 @@ struct device_partition {
 
 
 /**
- * Create device_segmented_array from a segmented_array.
- */
-template <typename LocalIndexType, typename GlobalIndexType>
-constexpr device_partition<const LocalIndexType, const GlobalIndexType>
-to_device(
-    const experimental::distributed::Partition<LocalIndexType, GlobalIndexType>*
-        partition)
-{
-    auto num_ranges = partition->get_num_ranges();
-    auto num_parts = partition->get_num_parts();
-    return {num_parts,
-            partition->get_num_empty_parts(),
-            partition->get_size(),
-            partition->get_range_bounds(),
-            partition->get_range_bounds() + num_ranges + 1,
-            partition->get_range_starting_indices(),
-            partition->get_range_starting_indices() + num_ranges,
-            partition->get_part_sizes(),
-            partition->get_part_sizes() + num_parts,
-            partition->get_part_ids(),
-            partition->get_part_ids() + num_parts,
-            to_device(partition->get_ranges_by_part())};
-}
-
-/**
- * Explicitly create a const version of device_segmented_array.
- *
- * This is mostly relevant for tests.
+ * Explicitly create a const version of device_partition.
  */
 template <typename LocalIndexType, typename GlobalIndexType>
 constexpr device_partition<const LocalIndexType, const GlobalIndexType>
@@ -87,6 +59,3 @@ to_device_const(
 
 
 }  // namespace gko
-
-
-#endif  // GINKGO_PARTITION_HPP
--- a/core/distributed/index_map_kernels.hpp
+++ b/core/distributed/index_map_kernels.hpp
@@ -54,11 +54,11 @@ namespace kernels {
  * space defined by is. The resulting indices are stored in local_ids.
  * The index map is defined by the input parameters:
  *
- * - partition:  the global partition
+ * - partition: the global partition
  * - remote_target_ids: the owning part ids of each segment of
  *                      remote_global_idxs
  * - remote_global_idxs: the remote global indices, segmented by the owning part
- *   ids
+ *                       ids, and each segment sorted
  * - rank: the part id of this process
  *
  * Any global index that is not in the specified local index space is mapped
@@ -81,7 +81,7 @@ namespace kernels {
  *
  * The relevant input parameter from the index map are:
  *
- * - partition:  the global partition
+ * - partition: the global partition
  * - remote_global_idxs: the remote global indices, segmented by the owning part
  *                       ids
  * - rank: the part id of this process
@@ -95,8 +95,8 @@ namespace kernels {
         device_partition<const _ltype, const _gtype> partition,  \
         device_segmented_array<const _gtype> remote_global_idxs, \
         experimental::distributed::comm_index_type rank,         \
-        const array<_ltype>& local_ids,                          \
-        experimental::distributed::index_space is, array<_gtype>& global_ids)
+        const array<_ltype>& local_idxs,                         \
+        experimental::distributed::index_space is, array<_gtype>& global_idxs)
 
 
 #define GKO_DECLARE_ALL_AS_TEMPLATES                                      \
--- a/core/test/utils/assertions.hpp
+++ b/core/test/utils/assertions.hpp
@@ -24,6 +24,7 @@
 #include <ginkgo/core/base/math.hpp>
 #include <ginkgo/core/base/mtx_io.hpp>
 #include <ginkgo/core/base/name_demangling.hpp>
+#include <ginkgo/core/base/segmented_array.hpp>
 #include <ginkgo/core/matrix/dense.hpp>
 
 #include "core/base/batch_utilities.hpp"
@@ -1014,17 +1015,17 @@ template <typename ValueType>
                                    second.get_const_flat_data())
             .copy_to_array();
 
-    auto buffer_result = array_equal(first_expression, second_expression,
-                                     view_first, view_second);
-    if (buffer_result == ::testing::AssertionFailure()) {
-        return buffer_result << "Buffers of the segmented arrays mismatch";
-    }
-
     auto offsets_result =
         array_equal(first_expression, second_expression, first.get_offsets(),
                     second.get_offsets());
     if (offsets_result == ::testing::AssertionFailure()) {
         return offsets_result << "Offsets of the segmented arrays mismatch";
+    }
+
+    auto buffer_result = array_equal(first_expression, second_expression,
+                                     view_first, view_second);
+    if (buffer_result == ::testing::AssertionFailure()) {
+        return buffer_result << "Buffers of the segmented arrays mismatch";
     }
 
     return ::testing::AssertionSuccess();
@@ -1414,6 +1415,18 @@ T* plain_ptr(T* ptr)
     }
 
 
+/**
+ * Checks if two `gko::segmented_array`s are equal.
+ *
+ * Both the flat array buffer and the offsets of both arrays are tested
+ * for equality.
+ *
+ * Has to be called from within a google test unit test.
+ * Internally calls gko::test::assertions::segmented_array_equal().
+ *
+ * @param _array1 first segmented array
+ * @param _array2 second segmented array
+ */
 #define GKO_ASSERT_SEGMENTED_ARRAY_EQ(_array1, _array2)                      \
     {                                                                        \
         ASSERT_PRED_FORMAT2(::gko::test::assertions::segmented_array_equal,  \
--- a/dpcpp/distributed/partition_kernels.dp.cpp
+++ b/dpcpp/distributed/partition_kernels.dp.cpp
@@ -140,10 +140,7 @@ void build_ranges_by_part(std::shared_ptr<const DefaultExecutor> exec,
 
     range_ids.resize_and_reset(num_ranges);
     auto range_ids_ptr = range_ids.get_data();
-    // fill range_ids with 0,...,num_ranges - 1
-    run_kernel(
-        exec, [] GKO_KERNEL(auto i, auto rid) { rid[i] = i; }, num_ranges,
-        range_ids_ptr);
+    components::fill_seq_array(exec, range_ids_ptr, num_ranges);
 
     oneapi::dpl::stable_sort(policy, range_ids_ptr, range_ids_ptr + num_ranges,
                              [range_parts](const auto rid_a, const auto rid_b) {
--- a/omp/distributed/index_map_kernels.cpp
+++ b/omp/distributed/index_map_kernels.cpp
@@ -245,62 +245,50 @@ void map_to_global(
     device_partition<const LocalIndexType, const GlobalIndexType> partition,
     device_segmented_array<const GlobalIndexType> remote_global_idxs,
     experimental::distributed::comm_index_type rank,
-    const array<LocalIndexType>& local_ids,
+    const array<LocalIndexType>& local_idxs,
     experimental::distributed::index_space is,
-    array<GlobalIndexType>& global_ids)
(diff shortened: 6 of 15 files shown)


core/distributed/partition.cpp
```diff
@@ -126,6 +128,13 @@ void Partition<LocalIndexType, GlobalIndexType>::finalize_construction()
         get_num_parts(), num_empty_parts_, starting_indices_.get_data(),
         part_sizes_.get_data()));
     size_ = get_element(offsets_, get_num_ranges());
+    array<size_type> range_ids(exec);
+    array<int64> num_ranges_per_part(exec);
+    exec->run(partition::make_build_ranges_by_part(
+        part_ids_.get_const_data(), get_num_ranges(), get_num_parts(),
+        range_ids, num_ranges_per_part));
+    ranges_by_part_ = segmented_array<size_type>::create_from_sizes(
+        std::move(range_ids), num_ranges_per_part);
```
yhmtsai, line 137: currently, this rely on the atomic add to count the size.
then segmented_array will convert size to offset.
Maybe you can implement a similar kernel with convert_idxs_to_ptrs + additional index map from i to range_id[i] because `range_part[range_id[i]]` is sorted. then segmented_array can `create_from_offsets` directly

reference/distributed/partition_helpers.hpp
```diff
@@ -48,6 +51,49 @@ LocalIndexType map_to_local(
...
+        });
+    auto local_range_id = std::distance(local_ranges.begin, it) - 1;
+    return local_range_id;
+}
+
+
+template <typename LocalIndexType, typename GlobalIndexType>
+GlobalIndexType map_to_global(
+    LocalIndexType idx,
+    device_partition<const LocalIndexType, const GlobalIndexType> partition,
```
yhmtsai, line 87:
```suggestion
    const device_partition<const LocalIndexType, const GlobalIndexType>& partition,
```

reference/distributed/partition_helpers.hpp
```diff
@@ -48,6 +51,49 @@ LocalIndexType map_to_local(
...
+    size_type range_id)
+{
+    auto range_bounds = partition.offsets_begin;
```
yhmtsai, line 90: assert range_id is lower than num_range?

yhmtsai (commented):
not finish review yet.
why is it required for PGM?
after merging row and column, do they still need to map back to the global index?
you need to reindex all of them because the size will be the smaller than the original one after coarsening.

MarcelKoch:
> after merging row and column, do they still need to map back to the global index? 
> you need to reindex all of them because the size will be the smaller than the original one after coarsening.

@yhmtsai I don't understand your point, could you rephrase this?

MarcelKoch:
@yhmtsai I think this discussion makes more sense in #1639. I don't see much relevance to this PR.

yhmtsai:
yes, but if PGM does not need it, do you still need this function for something else?

MarcelKoch:
I think I need at least parts of it to keep the old distributed matrix constructor around. @fritzgoebel also seems to need it for his domain decomposition matrix.

Anyway, this still sounds like a discussion for #1639.

test/distributed/index_map_kernels.cpp
```diff
@@ -388,3 +406,256 @@ TEST_F(IndexMap, GetLocalWithCombinedIndexSpaceWithInvalidIndexSameAsRef)
 
     GKO_ASSERT_ARRAY_EQ(result, dresult);
 }
+
+
+TEST_F(IndexMap, GetGlobalWithLocalIndexSpaceSameAsRef)
```
pratikvn, line 411: Same here. I think just the test with both invalid and valid indices should be enough because the kernel you are calling is the same, and the code path in both cases is the same except for the bounds check.

test/distributed/partition_kernels.cpp
```diff
@@ -69,6 +69,16 @@ class Partition : public CommonTestFixture {
...
+
+        GKO_ASSERT_ARRAY_EQ(
+            gko::make_const_array_view(
+                this->ref, part->get_num_ranges(),
+                part->get_ranges_by_part().get_const_flat_data()),
+            gko::make_const_array_view(
+                this->exec, dpart->get_num_ranges(),
+                dpart->get_ranges_by_part().get_const_flat_data()));
+        GKO_ASSERT_ARRAY_EQ(part->get_ranges_by_part().get_offsets(),
+                            dpart->get_ranges_by_part().get_offsets())
```
pratikvn, line 81: If you add a `GKO_ASSERT_SEGEMENT_ARRAY_EQ` macro from the assert function in the other file, then you can use that here as well.

pratikvn (commented):
Some comments, but mostly looks good!

core/distributed/device_partition.hpp
```diff
@@ -0,0 +1,92 @@
...
+}
+
+/**
+ * Explicitly create a const version of device_segmented_array.
+ *
+ * This is mostly relevant for tests.
+ */
+template <typename LocalIndexType, typename GlobalIndexType>
+constexpr device_partition<const LocalIndexType, const GlobalIndexType>
+to_device_const(
```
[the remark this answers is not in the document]
MarcelKoch: Yes, right now I only need the `to_device_const` version, as I'm not modifying the partition. So I will keep only that version.

reference/distributed/partition_helpers.hpp
MarcelKoch: the `device_partition` type is meant to be handled as a value type. It only contains trivially copyable members.

core/distributed/device_partition.hpp
```diff
@@ -0,0 +1,92 @@
...
+to_device(
+    const experimental::distributed::Partition<LocalIndexType, GlobalIndexType>*
+        partition)
+{
+    auto num_ranges = partition->get_num_ranges();
+    auto num_parts = partition->get_num_parts();
+    return {num_parts,
+            partition->get_num_empty_parts(),
+            partition->get_size(),
+            partition->get_range_bounds(),
```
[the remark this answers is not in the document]
MarcelKoch: I think this is a bit of a temporary solution until we fully introduce device-side views for all of our (relevant) types. So I would leave handling that issue until then.

reference/test/distributed/index_map_kernels.cpp
```diff
@@ -195,3 +196,113 @@ TEST_F(IndexMap, CanGetLocalWithCombinedISWithInvalid)
...
+
+    gko::array<global_index_type> expected(ref, {14, 13, 12, 5, 4, 3, 13});
+    GKO_ASSERT_ARRAY_EQ(global_ids, expected);
+}
+
+
+TEST_F(IndexMap, CanGetGlobalWithLocalISWithInvalid)
+{
+    gko::array<global_index_type> global_ids(ref);
+    gko::array<local_index_type> local_ids(ref, {5, 4, 10, 3, 2, 1, 0, 100, 4});
```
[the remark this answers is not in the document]
MarcelKoch: The global_id=17 is not part of the local index set of rank 1, which is used here. I changed it to rank 0 for this test, and now it also outputs gid=17.

core/distributed/index_map_kernels.hpp
```diff
@@ -72,11 +76,36 @@ namespace kernels {
...
+ * The relevant input parameter from the index map are:
+ *
+ * - partition: the global partition
+ * - remote_global_idxs: the remote global indices, segmented by the owning part
+ *                       ids
+ * - rank: the part id of this process
+ *
+ * Any local index that is not part of the specified index space is mapped to
+ * invalid_index.
+ */
```
pratikvn, line 91: I think this is a nice way to capture the documentation of our kernels (And we have a lot of them). I wonder if it is possible to list all the kernels and capture their documentation in the developer documentation.

core/test/utils/assertions.hpp
```diff
@@ -999,6 +999,38 @@ ::testing::AssertionResult array_equal(const std::string& first_expression,
 }
 
 
+template <typename ValueType>
+::testing::AssertionResult segmented_array_equal(
```
pratikvn, line 1003: Can you also add a small test for this in `core/test/utils/assertions_test.cpp`  ?

test/distributed/index_map_kernels.cpp
pratikvn: I think this is not resolved yet ?

pratikvn (approved):
A couple more comments.

test/distributed/index_map_kernels.cpp
```diff
@@ -117,6 +117,11 @@ class IndexMap : public CommonTestFixture {
 
     IndexMap()
     {
+        std::random_device rd;
+        std::uniform_int_distribution<> engine_dist(
+            0, std::numeric_limits<int>::max());
```
yhmtsai, line 122: `rd` and `engine_dist` seems to be unused

reference/distributed/partition_helpers.hpp
```diff
@@ -48,6 +51,51 @@ LocalIndexType map_to_local(
...
+
+
+template <typename LocalIndexType, typename GlobalIndexType>
+GlobalIndexType map_to_global(
+    LocalIndexType idx,
+    device_partition<const LocalIndexType, const GlobalIndexType> partition,
+    size_type range_id)
+{
+    assert(range_id <
+           std::distance(partition.offsets_begin, partition.offsets_end) - 1);
```
yhmtsai, line 91:
```suggestion
    assert(range_id <
           std::distance(partition.offsets_begin, partition.offsets_end));
```
do you need -1?

reference/test/distributed/index_map_kernels.cpp
```diff
@@ -195,3 +196,108 @@ TEST_F(IndexMap, CanGetLocalWithCombinedISWithInvalid)
     gko::array<local_index_type> expected(ref, {2, 3, 0, 1, 2, 4, -1, 1});
     GKO_ASSERT_ARRAY_EQ(local_ids, expected);
 }
+
+
+TEST_F(IndexMap, CanGetGlobalWithLocalISWithInvalid)
```
yhmtsai, line 201:
```suggestion
TEST_F(IndexMap, CanGetGlobalWithLocalIsWithInvalid)
```
Is it supposted to be `IS`?

common/cuda_hip/distributed/index_map_kernels.cpp
```diff
@@ -296,6 +296,90 @@ GKO_INSTANTIATE_FOR_EACH_LOCAL_GLOBAL_INDEX_TYPE(
...
+
+        auto local_ranges = ranges_by_part.get_segment(rank);
+        auto local_ranges_size =
+            static_cast<int64>(local_ranges.end - local_ranges.begin);
+
+        auto it = binary_search(int64(0), local_ranges_size, [=](const auto i) {
+            return starting_indices[local_ranges.begin[i]] >= lid;
+        });
+        auto local_range_id =
+            it != local_ranges_size ? it : max(int64(0), it - 1);
```
yhmtsai, line 337: will this happen? it seems to be broken if we reach the else branch

common/cuda_hip/distributed/partition_kernels.cpp
```diff
@@ -132,6 +135,36 @@ GKO_INSTANTIATE_FOR_EACH_LOCAL_GLOBAL_INDEX_TYPE(
...
+    thrust::sequence(policy, range_ids_ptr, range_ids_ptr + num_ranges);
+
+    // mutable copy of range_parts such that it can be used as keys for sorting
+    array<int> range_parts_copy{exec, num_ranges};
+    thrust::copy_n(policy, range_parts, num_ranges,
+                   range_parts_copy.get_data());
+    auto range_parts_ptr = range_parts_copy.get_data();
+
+    thrust::stable_sort_by_key(policy, range_parts_ptr,
+                               range_parts_ptr + num_ranges, range_ids_ptr);
```
yhmtsai, line 156: it uses the truth that range ids is also sorted, right?
such that it is sorted by (part_id, range_id)

yhmtsai (changes requested):
the new testing method `GKO_ASSERT_SEGMENTED_ARRAY_EQ` needs a test in core/test/utils/assertions_test.cpp as @pratikvn mentioned
some comments from @pratikvn seems not to be checked yet

test/distributed/partition_kernels.cpp
```diff
@@ -69,6 +69,9 @@ class Partition : public CommonTestFixture {
             gko::make_array_view(
                 this->exec, dpart->get_num_parts(),
                 const_cast<local_index_type*>(dpart->get_part_sizes())));
+
+        GKO_ASSERT_SEGMENTED_ARRAY_EQ(part->get_ranges_by_part(),
```
upsj, line 73: Side-node: Since this is a function returning `void`, this will not prevent subsequent operations from executing.

reference/distributed/index_map_kernels.cpp
```diff
@@ -199,6 +199,77 @@ GKO_INSTANTIATE_FOR_EACH_LOCAL_GLOBAL_INDEX_TYPE(
...
 
+template <typename LocalIndexType, typename GlobalIndexType>
+void map_to_global(
+    std::shared_ptr<const DefaultExecutor> exec,
+    device_partition<const LocalIndexType, const GlobalIndexType> partition,
+    device_segmented_array<const GlobalIndexType> remote_global_idxs,
+    experimental::distributed::comm_index_type rank,
+    const array<LocalIndexType>& local_ids,
+    experimental::distributed::index_space is,
+    array<GlobalIndexType>& global_ids)
```
upsj, line 210: idxs and ids seems to be used interchangably here?

reference/distributed/index_map_kernels.cpp
```diff
@@ -199,6 +199,77 @@ GKO_INSTANTIATE_FOR_EACH_LOCAL_GLOBAL_INDEX_TYPE(
...
+            auto lid = local_ids.get_const_data()[i];
+            global_ids.get_data()[i] = map_non_local(lid);
+        }
+    }
+    if (is == experimental::distributed::index_space::combined) {
+        for (size_type i = 0; i < local_ids.get_size(); ++i) {
+            auto lid = local_ids.get_const_data()[i];
+            global_ids.get_data()[i] = map_combined(lid);
+        }
+    }
```
upsj, line 266: I think this might be more readable when pulling in the condition `is == ...`, the compiler can hoist this condition out when optimizing.

reference/distributed/index_map_kernels.cpp
```diff
@@ -199,6 +199,77 @@ GKO_INSTANTIATE_FOR_EACH_LOCAL_GLOBAL_INDEX_TYPE(
...
+    device_segmented_array<const GlobalIndexType> remote_global_idxs,
+    experimental::distributed::comm_index_type rank,
+    const array<LocalIndexType>& local_ids,
+    experimental::distributed::index_space is,
+    array<GlobalIndexType>& global_ids)
+{
+    const auto& ranges_by_part = partition.ranges_by_part;
+    auto local_ranges = ranges_by_part.get_segment(rank);
+
+    global_ids.resize_and_reset(local_ids.get_size());
```
upsj, line 215: local_ids.get_size() gets used multiple times, might be worth a variable?

omp/distributed/index_map_kernels.cpp
```diff
@@ -239,6 +239,77 @@ GKO_INSTANTIATE_FOR_EACH_LOCAL_GLOBAL_INDEX_TYPE(
...
+    auto local_ranges = ranges_by_part.get_segment(rank);
+
+    global_ids.resize_and_reset(local_ids.get_size());
+
+    auto local_size =
+        static_cast<LocalIndexType>(partition.part_sizes_begin[rank]);
+    auto remote_size = static_cast<LocalIndexType>(
+        remote_global_idxs.flat_end - remote_global_idxs.flat_begin);
+    size_type local_range_id = 0;
+    if (is == experimental::distributed::index_space::local) {
```
upsj, line 262: same here, I think this might be more readable pulling the condition in? Also can we copy the approach from `reference`, since it uses named lambdas instead of unnamed pieces of code

dpcpp/distributed/partition_kernels.dp.cpp
```diff
@@ -130,6 +131,35 @@ GKO_INSTANTIATE_FOR_EACH_LOCAL_GLOBAL_INDEX_TYPE(
...
+                          array<int64>& sizes)
+{
+    auto policy = onedpl_policy(exec);
+
+    range_ids.resize_and_reset(num_ranges);
+    auto range_ids_ptr = range_ids.get_data();
+    // fill range_ids with 0,...,num_ranges - 1
+    run_kernel(
+        exec, [] GKO_KERNEL(auto i, auto rid) { rid[i] = i; }, num_ranges,
+        range_ids_ptr);
```
upsj, line 146: `components::fill_seq_array`?

core/test/utils/assertions.hpp
```diff
@@ -999,6 +999,38 @@ ::testing::AssertionResult array_equal(const std::string& first_expression,
...
+    if (buffer_result == ::testing::AssertionFailure()) {
+        return buffer_result << "Buffers of the segmented arrays mismatch";
+    }
+
+    auto offsets_result =
+        array_equal(first_expression, second_expression, first.get_offsets(),
+                    second.get_offsets());
+    if (offsets_result == ::testing::AssertionFailure()) {
+        return offsets_result << "Offsets of the segmented arrays mismatch";
+    }
```
upsj, line 1028: I would probably check offsets first, since if they mismatch, the rest of the contents make even less sense.

core/distributed/index_map_kernels.hpp
upsj: I agree with Marcel's comment yesterday that Ginkgo developers would rather use code as documentation than the Doxygen, I think we should focus on writing good comments, regardless of whether they work precisely as Doxygen comments.

core/distributed/device_partition.hpp
```diff
@@ -0,0 +1,61 @@
...
+    device_segmented_array<const size_type> ranges_by_part;
+};
+
+
+/**
+ * Explicitly create a const version of device_partition.
+ */
+template <typename LocalIndexType, typename GlobalIndexType>
+constexpr device_partition<const LocalIndexType, const GlobalIndexType>
+to_device_const(
```
upsj, line 40: I like this approach!

core/distributed/device_partition.hpp
```diff
@@ -0,0 +1,61 @@
+// SPDX-FileCopyrightText: 2024 - 2025 The Ginkgo authors
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+#pragma once
```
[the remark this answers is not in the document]
upsj: There are only very obscure edge cases where pragma once doesn't work properly (because some filesystems don't make it easy to see if two files have the same identity), I think we could safely move from header guards to `#pragma once` if we wanted to. Every relevant compiler supports it.

upsj (changes requested):
Some minor change requests and comments

reference/distributed/partition_helpers.hpp
MarcelKoch: Yes, the offsets array has `#range-ids + 1` entries, since it's an offset array. For an empty partition, `offsets = {0}`, so the begin and end pointers are different.

reference/test/distributed/index_map_kernels.cpp
MarcelKoch: I think I'm usually using `IS` in these tests to denote the `index_space` enum. I would keep it like that.

common/cuda_hip/distributed/partition_kernels.cpp
MarcelKoch: Yes, although the `part_id` is not used after this anymore. It's only relevant that the `range_id` is first sorted by `part_id`, and then by `range_id`.

test/distributed/partition_kernels.cpp
MarcelKoch: This should be implemented the same way as the `GKO_ASSERT_ARRAY_EQ` macro. The `segmented_array_equal` function will return something, so I don't know if there is anything more to do here.

upsj: I mean the overall approach of implementing a custom `assert_something` function may cause tests to continue executing after the assertion fails, which may or may not be desirable. It's fine here, but something to keep in mind. To fix this, we would need to return `testing::AssertionResult` and wrap the call in a macro. But nothing to hold back the code here, just a comment that people should be aware of.

reference/distributed/index_map_kernels.cpp
MarcelKoch: I will now use `idxs` here.

reference/distributed/index_map_kernels.cpp
MarcelKoch: by hoisting the ifs into the loop this reduced to two times, so I will just leave it.

omp/distributed/index_map_kernels.cpp
MarcelKoch: I think one compiler didn't work with lambdas within the parallel for loop, maybe nvhpc. I will still hoist the ifs into the loop.

core/distributed/partition.cpp
MarcelKoch: Yes, for now I would just use atomics. I don't think this is a performance critical part, so it should be fine for a while.

omp/distributed/index_map_kernels.cpp
upsj: Intel's classic compiler used to have the opposite issue (you needed to put the loop body into an IIFE to work around some compiler crashes). Then maybe the code could use some comments describing the different cases briefly

omp/distributed/index_map_kernels.cpp
```diff
@@ -19,8 +19,6 @@ namespace gko {
 namespace kernels {
 namespace omp {
 namespace index_map {
-
-
```
yhmtsai, line 23: should be kept

reference/distributed/index_map_kernels.cpp
```diff
@@ -15,8 +15,6 @@ namespace gko {
 namespace kernels {
 namespace reference {
 namespace index_map {
-
-
 template <typename LocalIndexType, typename GlobalIndexType>
```
yhmtsai, line 18: should kept

yhmtsai (approved):

pratikvn (approved):
LGTM!

upsj (approved):
LGTM!

PR #1710, opened 2024-10-25
Sycl Half

yhmtsai:
This PR will be necessary to support half in sycl when we still use gko::half in the host.
It creates the mapping like cuda_type or hip_type in sycl part, and apply it to those variables with value type

It will be merged/checked when we have half and intel ci


MarcelKoch (approved):

CMakeLists.txt
```diff
@@ -34,7 +34,7 @@ option(GINKGO_TEST_NONDEFAULT_STREAM "Uses non-default streams in CUDA and HIP t
 option(GINKGO_MIXED_PRECISION "Instantiate true mixed-precision kernels (otherwise they will be conversion-based using implicit temporary storage)" OFF)
 option(GINKGO_ENABLE_HALF "Enable the use of half precision" ON)
 # We do not support MSVC. SYCL will come later
-if(MSVC OR GINKGO_BUILD_SYCL)
+if(MSVC)
```
upsj, line 37: The error message also needs to be fixed

dpcpp/base/math.hpp
```diff
@@ -0,0 +1,236 @@
...
+        result_f /= val_f;
+        real_ = result_f.real();
+        imag_ = result_f.imag();
+        return *this;
+    }
+
+// It's for MacOS.
+// TODO: check whether mac compiler always use complex version even when real
+// half
+#define COMPLEX_HALF_OPERATOR(_op, _opeq)                                  \
```
upsj, line 144: Has this TODO been addressed?

dpcpp/components/atomic.dp.hpp
```diff
@@ -266,15 +270,34 @@ template <sycl::access::address_space addressSpace = atomic::global_space,
           typename T>
 __dpct_inline__ T atomic_add(T* __restrict__ addr, T val)
 {
-    return detail::atomic_helper<addressSpace, T>::atomic_add(addr, val);
+    if constexpr (std::is_same_v<T, sycl::half> ||
+                  std::is_same_v<T, std::complex<sycl::half>>) {
+        // unsupported
+        auto old = *addr;
+        *addr += val;
+        return old;
```
upsj, line 278: If it's unsupported, we should not compile it

upsj (dismissed):

dpcpp/components/atomic.dp.hpp
upsj: For a workaround, we could use an atomic_cas loop

dpcpp/base/complex.hpp
MarcelKoch, line 1: Instead of this workaround, would creating a non-std complex implementation with `sycl::half` work, e.g. `gko::complex<sycl::half>`? Then you could do the usual device type mapping stuff to actually use this.

upsj: +1 to this idea, though we should keep the implementation internal to Ginkgo to allow future changes

MarcelKoch: Yes, of course, it should be a private type.

yhmtsai: the complex is internal type. They will not expose it to public.
We only add the dependence with `-fsycl`, which is only in the dpcpp backend.
I wonder it affect the performance especially the intel complex mapped std::complex<sycl::half> to __spv::complex_half, which might be their internal type?
In gko::half and sycl::half, I have seen 1.5 or 2x slowdown by using gko::half.

MarcelKoch: So can you also use `_spv::complex_half` here? It sounds a bit like you could just map `std::complex<gko::half>` to that type and don't need to introduce a new type at all.

dpcpp/base/complex.hpp
```diff
@@ -0,0 +1,275 @@
...
+#endif
+
+#else
+
+
+#include <complex>
+
+
+#endif
+
```
MarcelKoch, line 208: You should be able to just remove this and always include `<complex>`.
Some of the comments can also be adjusted.

yhmtsai: Yes, it will be a normal header again after everything works well

PR #1711, opened 2024-10-25
solver (mostly krylov solver) and the residual norm with half

yhmtsai:
This PR enable the solvers and the residual norm with half. It also includes the corresponding config


common/cuda_hip/solver/idr_kernels.cpp
```diff
@@ -555,7 +555,8 @@ void initialize(std::shared_ptr<const DefaultExecutor> exec,
     orthonormalize_subspace_vectors(exec, subspace_vectors);
 }
 
-GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE(GKO_DECLARE_IDR_INITIALIZE_KERNEL);
+GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE_WITH_HALF(
```
pratikvn, line 558: A general question: Why do we need another macro `_WITH_HALF` ? Is it not possible to just have `half` within the first macro ?

yhmtsai: It is possible but it will need to result it in #1257  because we need to finish everything all at once.

MarcelKoch (changes requested):
LGTM. Can you clarify the testing issues with IDR?

core/config/config_helper.hpp
```diff
@@ -200,7 +200,9 @@ get_value(const pnode& config)
  * This is specialization for floating point type
  */
 template <typename ValueType>
-inline std::enable_if_t<std::is_floating_point<ValueType>::value, ValueType>
+inline std::enable_if_t<std::is_floating_point<ValueType>::value ||
```
pratikvn, line 203: I think this structure will be useful everywhere. Maybe it would be nice to have our own `is_floating_point` which includes `half` ? So all you will need to do is `gko::is_floating_point<ValueType>::value` ?

pratikvn (approved):
I think it makes sense to remove `half` support for solvers which do not converge with half ? (IDR, GCR ?) Otherwise, looks good to me.

reference/test/solver/idr_kernels.cpp
```diff
@@ -57,6 +57,9 @@ class Idr : public ::testing::Test {
     std::unique_ptr<typename Solver::Factory> idr_factory_precision;
 };
 
+// Solves((Conj)Trans)DenseSystem((Mixed)Complex) does not work in some default
```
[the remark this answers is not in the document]
yhmtsai: Fixes it.
It is solved in the subspace solve. 
The second updated residual vector (t) is zero such that the omega is nan.
residual += omega * t
dense_x += omega * helper
I set the omega = 0 when norm(t) is zero.
(no update in residual, so no update in dense_x)
It sounds a reasonable workaround for me.

core/config/config_helper.hpp
yhmtsai: I only use it here because others usually need to be writen separately

reference/test/solver/idr_kernels.cpp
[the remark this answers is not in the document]
MarcelKoch: seems reasonable to me.

reference/test/solver/gcr_kernels.cpp
```diff
@@ -330,7 +332,7 @@ TYPED_TEST(Gcr, SolvesStencilSystemUsingAdvancedApplyMixed)
     solver->apply(alpha.get(), b.get(), beta.get(), x.get());
 
     GKO_ASSERT_MTX_NEAR(x, l({1.5, 5.0, 2.0}),
-                        (r_mixed<value_type, TypeParam>()) * 1e1);
+                        (r_mixed<value_type, TypeParam>()) * 1e2);
```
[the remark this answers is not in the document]
yhmtsai: I also check the other tol changes in this file. they are not needed at least in cuda10.2 image. need to check the other ci job

reference/test/solver/idr_kernels.cpp
```diff
@@ -57,7 +60,10 @@ class Idr : public ::testing::Test {
     std::unique_ptr<typename Solver::Factory> idr_factory_precision;
 };
 
-TYPED_TEST_SUITE(Idr, gko::test::ValueTypes, TypenameNameGenerator);
+// Solves((Conj)Trans)DenseSystem((Mixed)Complex) does not work in some default
+// random generator from different environments. All tests will SKIP half, so we
+// do not test half here.
```
MarcelKoch, line 65: The comment seems to be outdated.

MarcelKoch (approved):

PR #1712, opened 2024-10-25
Half factorization

yhmtsai:
this pr adds the factorization with half support.

Hip does not support atomic on the 16bits type currently

NVHPC 23.3 seems to handle assignment index with optimization wrongly on a custom class when IndexType is long. We set the index explicitly with volatile to solve it. NVHPC24.1 seem to fixed this issue.
https://godbolt.org/z/srYhGndKn

TODO:
- [x] add the fix of tri solve with half


hip/components/memory.hip.hpp
```diff
@@ -60,6 +60,12 @@ struct gcc_atomic_intrinsic_type_map<double> {
 };
 
 
+template <>
+struct gcc_atomic_intrinsic_type_map<__half> {
```
MarcelKoch, line 64: maybe leave this undefined, so we will get compiler errors, when we try to use hip fp16 atomics.

reference/factorization/par_ilut_kernels.cpp
```diff
@@ -191,7 +191,12 @@ void threshold_filter_approx(std::shared_ptr<const DefaultExecutor> exec,
     // pick splitters
     for (IndexType i = 0; i < bucket_count - 1; ++i) {
         // shift by one so we get upper bounds for the buckets
-        sample[i] = sample[(i + 1) * sampleselect_oversampling];
+        // TODO FIXME: NVHPC 23.3 seems to handle assignment index with
```
MarcelKoch, line 194: wow, this is pretty bad... BTW it is fixed in NVHPC 24.1, maybe also mention that.

test/factorization/lu_kernels.cpp
```diff
@@ -127,7 +127,7 @@ class Lu : public CommonTestFixture {
 };
 
 #ifdef GKO_COMPILING_OMP
-using Types = gko::test::ValueIndexTypes;
+using Types = gko::test::ValueIndexTypesWithHalf;
```
MarcelKoch, line 130: Shouldn't the types for the cuda tests also be adjusted?

MarcelKoch (changes requested):
Generally LGTM. I have a question regarding atomics and hip. The latest ROCm shows support for fp16 atomic operations: https://rocm.docs.amd.com/en/latest/reference/precision-support.html#atomic-operations-support, but TBH I can't figure out what operations exactly they mean with that. Did you try anything in that regard?

cuda/solver/common_trs_kernels.cuh
```diff
@@ -557,7 +567,7 @@ void sptrsv_naive_caching(std::shared_ptr<const CudaExecutor> exec,
     const auto nrhs = b->get_size()[1];
 
     // Initialize x to all NaNs.
-    dense::fill(exec, x, nan<ValueType>());
+    dense::fill(exec, x, ValueType(nan<ValueType>()));
```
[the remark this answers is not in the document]
yhmtsai: good catch!
It is the leftover from `numeric_limits` can only return float before `create_from_bit`

cuda/solver/common_trs_kernels.cuh
```diff
@@ -577,7 +587,13 @@ void sptrsv_naive_caching(std::shared_ptr<const CudaExecutor> exec,
                 as_device_type(x->get_values()), x->get_stride(), n, nrhs,
                 unit_diag, nan_produced.get_data(), atomic_counter.get_data());
     } else {
-        sptrsv_naive_caching_kernel<is_upper>
+        // TODO: need to investigate
+        // memory operation on the half-precision shared_memory seem to give
+        // wrong result. we use float in shared_memory.
+        using shared_value_type =
+            std::conditional_t<std::is_same<ValueType, gko::half>::value, float,
```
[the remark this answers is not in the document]
yhmtsai: Yes, it can be actually.
I thought we did it in pack-fasion, so it use 32-bit not 16-bit one. However, it is not true.

test/factorization/par_ilut_kernels.cpp
```diff
@@ -245,25 +233,31 @@ class ParIlut : public CommonTestFixture {
...
 };
 
-TYPED_TEST_SUITE(ParIlut, gko::test::ValueIndexTypes,
+TYPED_TEST_SUITE(ParIlut, gko::test::ValueIndexTypesWithHalf,
                  PairTypenameNameGenerator);
 
 
 TYPED_TEST(ParIlut, KernelThresholdSelectIsEquivalentToRef)
 {
+    using value_type = typename TestFixture::value_type;
```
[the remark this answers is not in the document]
yhmtsai: we do not support compute_l_u_factors in hip, but the others still works with half precision in HIP

cuda/solver/common_trs_kernels.cuh
```diff
@@ -212,13 +212,15 @@ struct CudaSolveStruct : gko::solver::SolveStruct {
 
         size_type work_size{};
 
+        // TODO: In nullptr is considered nullptr_t not casted to const
+        // it does not work in cuda110/100 images
```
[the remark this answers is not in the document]
yhmtsai: I think it is more on the host compiler side because it goes through our binding first with specfic type

cuda/solver/common_trs_kernels.cuh
```diff
@@ -577,7 +589,13 @@ void sptrsv_naive_caching(std::shared_ptr<const CudaExecutor> exec,
...
                 unit_diag, nan_produced.get_data(), atomic_counter.get_data());
     } else {
-        sptrsv_naive_caching_kernel<is_upper>
+        // TODO: need to investigate
+        // memory operation on the half-precision shared_memory seem to give
+        // wrong result. we use float in shared_memory.
+        using shared_value_type = std::conditional_t<
+            std::is_same<remove_complex<ValueType>, gko::half>::value, float,
+            ValueType>;
+        sptrsv_naive_caching_kernel<is_upper, device_type<shared_value_type>>
```
MarcelKoch, line 598: now this will also be `float` when using `std::complex<gko::half>`. That doesn't seem correct. You will loose any imaginary part that might be in the matrix or vectors.

MarcelKoch (changes requested):

cuda/solver/common_trs_kernels.cuh
yhmtsai: you are right.
Sorry, I do not pay enough attention when changing it.

MarcelKoch (approved):

test/factorization/par_ilut_kernels.cpp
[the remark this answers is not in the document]
yhmtsai: I got your meaning now

cuda/solver/common_trs_kernels.cuh
```diff
@@ -212,12 +212,16 @@ struct CudaSolveStruct : gko::solver::SolveStruct {
 
         size_type work_size{};
 
+        // nullptr is considered nullptr_t not casted to the function signature
+        // automatically Explicitly cast `nullptr` to `const ValueType*` to
```
upsj, line 216: nit:
```suggestion
        // automatically explicitly cast `nullptr` to `const ValueType*` to
```

cuda/solver/common_trs_kernels.cuh
```diff
@@ -389,15 +398,17 @@ is_nan_exact(const T& value)
 }
 
 
-template <bool is_upper, typename ValueType, typename IndexType>
+template <bool is_upper, typename SharedValueType, typename ValueType,
+          typename IndexType>
```
upsj, line 402: Could SharedValueType be deduced inside, instead of making it an additional template parameter? You should be able to pull the code from the kernel launch into here and add a type alias. Otherwise it is easier to accidentally call the kernel with inconsistent types.

reference/factorization/par_ilut_kernels.cpp
```diff
@@ -191,7 +191,12 @@ void threshold_filter_approx(std::shared_ptr<const DefaultExecutor> exec,
     // pick splitters
     for (IndexType i = 0; i < bucket_count - 1; ++i) {
         // shift by one so we get upper bounds for the buckets
-        sample[i] = sample[(i + 1) * sampleselect_oversampling];
+        // NVHPC 23.3 seems to handle assignment index with
+        // optimization wrongly on a custom class when IndexType is long. We set
+        // the index explicitly with volatile to solve it. NVHPC24.1 fixed this
+        // issue. https://godbolt.org/z/srYhGndKn
+        volatile auto index = (i + 1) * sampleselect_oversampling;
```
upsj, line 198: I'm not sure we should go this far to accommodate broken compilers. We have workarounds for compilation issues, but not really for this degree of broken-ness.

upsj:
For HIP 16 bit atomics, as long as you only use load and store, you could implement them as
- a 32 bit load_* plus a memcpy and,
- a 32 bit load_* plus an atomic CAS, similar to how we did atomicAdd in the past. For safety, we need to execute this twice, assuming that every memory location only ever gets written once (which is true for all algorithms that use atomics), because your write can either fail because the upper half changed, or the lower half changed, one of which belongs to you and can't change without your knowledge.

cuda/solver/common_trs_kernels.cuh
yhmtsai: good idea

yhmtsai:
using 32 bit memory operation for 16 bit, it will have illegal memory access in the tail or head if we do not handle it in a upper level.

upsj:
Theoretically that would be an easy fix: Make sure all allocations are at least 32 bits and round up the size to multiples of 4. But I believe most allocators already silently fulfill that assumption, and GPUs are unlikely to have 16 bit allocation boundaries for alignment purposes.

yhmtsai:
I do not like slight guarantee unless we have a way to ensure or at least check.
However, I would suggest we do not consider it for this pr and release such that we have enough time ensure that it works correctly on hip.

upsj:
I can give you a somewhat technical justification for this: `cudaMalloc` returns correctly aligned memory for `thrust::complex<double>`, despite not knowing anything about the type. So that means that the allocator is not using any space between those 16 byte-aligned allocations. Whether this is special-cased for allocations divisible by 16 or not I'm not sure (I would assume not, since people also allocate memory pools themselves), but again, we have an easy fix, which I would honestly consider useful in any case: round up the sizes raw_allow uses to at least be divisible by 4.

yhmtsai:
I know the idea, sometimes it is necessary for optimized half precision by packing them (so, we will have kind of natively 32 bit by enforcing packing structure requirement)
I will still say it is not easy and confident to say it will be correct in this short period.
For example, user allocates some memory with 16 bits type but only pass the odd number to array_view.
should we accept or throw the error? Of course, these memory operation will not change the value out of the actual array, but it is still illegal memory operation.

PR #1713, opened 2024-10-25
Half preconditioner, multigrid, log, and reorder

yhmtsai:
This PR adds the preconditioner, multigrid, and reorder with half precision support.
move multigrid here because multigrid require direct from #1712  and jacobi here
the reorder and log are not much, so I include them as well.


reference/test/reorder/mc64_kernels.cpp
```diff
@@ -118,7 +119,7 @@ class Mc64 : public ::testing::Test {
                                           0., 2., 3., 0.}},
           final_dual_u{ref, I<real_type>{0., 1., -1., -2., 0., 0.}},
           final_distance{ref, I<real_type>{inf(), inf(), 1., 0., inf(), 1.}},
-          zero_tol{1e-14}
+          zero_tol{1e-4}
```
MarcelKoch, line 122: perhaps `c * r<value_type>`?

reference/test/reorder/mc64_kernels.cpp
```diff
@@ -284,6 +286,7 @@ TYPED_TEST(Mc64, CreatesCorrectPermutationAndScalingExampleSum)
         gko::experimental::reorder::Mc64<value_type, index_type>::build()
             .with_strategy(
                 gko::experimental::reorder::mc64_strategy::max_diagonal_sum)
+            .with_tolerance(real_type{1e-4})
```
MarcelKoch, line 289: IMO, the default parameter of MC64 should be changed to depend on the value type instead.

reference/test/reorder/mc64_kernels.cpp
```diff
@@ -373,6 +379,8 @@ TYPED_TEST(Mc64, CreatesCorrectPermutationAndScalingLargeExampleProduct)
     using value_type = typename TestFixture::value_type;
     using matrix_type = typename TestFixture::matrix_type;
     using perm_type = typename TestFixture::perm_type;
+    // this example can not be finished in half precision
```
MarcelKoch, line 382: I guess that's because the matrix contains values outside the value range of half? Maybe add that to the comment.

reference/test/reorder/scaled_reordered.cpp
```diff
@@ -364,6 +364,8 @@ TYPED_TEST(ScaledReordered, AppliesWithRcmReordering)
 TYPED_TEST(ScaledReordered, SolvesSingleRhsWithOnlyInnerOperator)
 {
     using SR = typename TestFixture::SR;
+    using value_type = typename TestFixture::value_type;
+    SKIP_IF_HALF(value_type);
```
MarcelKoch, line 368: The system is fairly small, so I wouid expect that the system can be solved with half precision. Is that not the case?

MarcelKoch (changes requested):
Mostly LGTM. I would suggest to change the default threshold parameter for MC64. 
Additionally, I would prefer to add some explanaitions as to why tests are skipped.

core/distributed/helpers.hpp
```diff
@@ -122,11 +122,15 @@ void vector_dispatch(T* linop, F&& f, Args&&... args)
 {
 #if GINKGO_BUILD_MPI
     if (is_distributed(linop)) {
-        using type = std::conditional_t<
-            std::is_const<T>::value,
-            const experimental::distributed::Vector<ValueType>,
-            experimental::distributed::Vector<ValueType>>;
-        f(dynamic_cast<type*>(linop), std::forward<Args>(args)...);
+        if constexpr (std::is_same_v<remove_complex<ValueType>, half>) {
```
pratikvn, line 125: Is there a reason we dont support half with distributed ? Or will it come in a future PR ?

pratikvn (changes requested):
Same note here. If we are skipping tests or unable to solve/compute with half with the correct result, then I think we should not support that algorithm with half.

reference/test/reorder/mc64_kernels.cpp
yhmtsai: because it still passes the original tests with double or float, I keep 1e-4 for now

core/distributed/helpers.hpp
yhmtsai: half precision transfer is trivial but the reduce operation and one side accumulation is not trivial.
I have tried adding the custom `MPI_Op` to support reduce on custom type, which works.
However, one side accumulation does not allow any non-predefined operation.
We will only support half in this case.
Or, use relatively new compiler and MPI standard with its extension, so they can support half natively.
It will be future PR anyways. It might not be in this release, either.

reference/test/reorder/scaled_reordered.cpp
yhmtsai: We need the scaling to make it work. I have deleted skip_if_half in the following tests

reference/test/preconditioner/isai_kernels.cpp
```diff
@@ -1021,6 +1034,9 @@ TYPED_TEST(Isai, ReturnsCorrectInverseALongrowWithExcessSolver)
 {
     using value_type = typename TestFixture::value_type;
     using GeneralIsai = typename TestFixture::GeneralIsai;
+    // When using the other precision, we already eed to drastically reduce the
```
MarcelKoch, line 1037: nit:
```suggestion
    // When using the other precision, we already need to drastically reduce the
```

reference/test/reorder/mc64_kernels.cpp
MarcelKoch: This still needs to be addressed.

MarcelKoch (changes requested):

reference/test/reorder/mc64_kernels.cpp
yhmtsai: Ah, I tend to keep this in the test because it does not affect the test result with single/double precision.
Or, do you mean the default parameter of MC64?

MarcelKoch: Yes, the default parameter in `include/ginkgo/core/reorder/mc64.hpp`. The current default will round to zero for half, so we should replace it.

yhmtsai: I pick 50*epsilon which is close to 1e-14 for double precision

MarcelKoch: sounds reasonable.

pratikvn (approved):
Some things on remaining tolerance levels, but otherwise LGTM

reference/test/reorder/mc64.cpp
```diff
@@ -79,18 +81,21 @@ TYPED_TEST(Mc64, HasSensibleDefaults)
...
 TYPED_TEST(Mc64, CanBeCreatedWithReorderingStrategy)
 {
     using reorder_type = typename TestFixture::reorder_type;
+    using real_type = typename TestFixture::real_type;
 
     auto mc64 =
         reorder_type::build()
             .with_strategy(
                 gko::experimental::reorder::mc64_strategy::max_diagonal_sum)
+            .with_tolerance(real_type{1e-4})
```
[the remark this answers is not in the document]
yhmtsai: no, it can be passed with the default tolerance. It works in one of case at least. let's see whether it can also pass the other ci job

reference/test/reorder/mc64_kernels.cpp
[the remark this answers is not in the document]
yhmtsai: I also updated it to the same setup like tolerance now

dpcpp/test/preconditioner/jacobi_kernels.dp.cpp
```diff
@@ -401,7 +401,7 @@ TEST_F(Jacobi, DpcppScalarApplyEquivalentToRef)
     auto dense_data =
         gko::test::generate_random_matrix_data<value_type, index_type>(
             dim, dim, std::uniform_int_distribution<>(1, dim),
-            std::normal_distribution<>(1.0, 2.0), engine);
+            std::normal_distribution<value_type>(1.0, 2.0), engine);
```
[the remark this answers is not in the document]
yhmtsai: When the sycl test is back, I guess I used it to ensure the single test working. 
When the single_mode is on, the generation on double gives different for generation on single such that some tests are failed with single

PR #1714, opened 2024-10-28
Electrostatic amir

amirbous:
#example

Further work on code provided by @pratikvn going along with progress for using ginkgo in electrostatic

Structure and changes - all changes are done under example/electrostatic-abb:
Added new utils.hpp file to include user defined methods: read input from matrix file, read config file .... .
And moved some of the already existing functions to it
added readme.md for electrstatic specific usage

Purpose
Added Support for binary matrix files: added reader
Added config file (electrto.config) parser , to read parameters for Ginkgo configuration
Ginkgo configuration parameters are set in electro.config. More in the readme page for the electrostatic example

!! Warning !!:
Still uncertain about double and single precision as output files seem to be double precision in ascii format and single precision in binary format. For now when reading binary, reading as float and casting to double.


pratikvn:
Thanks @amirbous! I think the explicit cast to `double` should not be necessary, but shouldn't hurt. The `matrix_data` object is templated and should implicitly cast to its value_type. 

So, you should be able to automatically run different solvers for both `float` and `double` by changing the `using ValueType = ...` in the `electrostatic_abb.cpp` file and compiling it again.

pratikvn:
@upsj , I think he is not using `gko::write`, but using the write functionality in the casopt library.

amirbous:
@upsj, @pratikvn  Thank you for the check.
The input files I am using are generated externally from another fortran script, seperate from ginkgo, and are using different precision based on the format. I made sure to note it down, so whenever other files are being used.
I will proceed as described by @pratikvn where the user is aware of the datatype being used, and can speicify it in the ValueType.

amirbous:
. To reissue

examples/electrostatic-abb/electrostatic-abb.cpp
```diff
@@ -137,8 +152,6 @@ int main(int argc, char* argv[])
         solver->apply(b, x_clone);
```
pratikvn, line 152: You can reduce the warmup here in line 150 to 1 instead of 3 to speed things up.

pratikvn (approved):
This looks okay to me. If I can merge it, let me know.

PR #1716, opened 2024-10-29
batch with half

yhmtsai:
This PR enable batch funtionality with half precision


reference/test/solver/batch_bicgstab_kernels.cpp
```diff
@@ -136,7 +141,7 @@ TYPED_TEST(BatchBicgstab, CanSolveDenseSystem)
     using real_type = gko::remove_complex<value_type>;
     using Solver = typename TestFixture::solver_type;
     using Mtx = typename TestFixture::Mtx;
-    const real_type tol = 1e-5;
+    const real_type tol = 1e-4;
```
pratikvn, line 144: Maybe loosen the tolerance only for half ? I think it would be better to leave the tighter tolerance for the other types.

reference/test/solver/batch_bicgstab_kernels.cpp
```diff
@@ -105,8 +105,13 @@ TYPED_TEST(BatchBicgstab, StencilSystemLoggerLogsResidual)
         ASSERT_LE(
             res_log_array[i] / this->linear_system.host_rhs_norm->at(i, 0, 0),
             this->solver_settings.residual_tol);
-        ASSERT_NEAR(res_log_array[i], res.host_res_norm->get_const_values()[i],
-                    10 * this->eps);
+        if (!std::is_same<real_type, gko::half>::value) {
+            // There is no guarantee of this condition. We disable this check in
+            // half.
```
pratikvn, line 110: What do you mean there is no guarantee for this condition ? I would assume that they are atleast close.

pratikvn (commented):
Mostly looks good. Some questions on tolerance, but in general we need to really rethink how we set tolerance everywhere in Ginkgo.

reference/test/solver/batch_bicgstab_kernels.cpp
yhmtsai: the implicit residual norm can be quite different from the actual residual norm after many iterations.

reference/test/solver/batch_bicgstab_kernels.cpp
yhmtsai: I think 1e-5 is quite loose for double, so 1e-5 and 1e-4 does not make big difference there.
I can change them only in half precision.
Side note. this condition also means that it is more challenging to half than double/single precision on the machine eps perspective.

reference/test/solver/batch_bicgstab_kernels.cpp
MarcelKoch: IMO the check can be removed completely. We can only guarantee that the (implicit) residual will be less than the given tolerance, which is checked above. So this is a quite unrelated check.

MarcelKoch (changes requested):

reference/test/solver/batch_bicgstab_kernels.cpp
```diff
@@ -288,6 +293,10 @@ TYPED_TEST(BatchBicgstab, CanSolveDenseHpdSystem)
     using real_type = gko::remove_complex<value_type>;
     using Solver = typename TestFixture::solver_type;
     using Mtx = typename TestFixture::Mtx;
+    // Need to design a better random system. With different random value
+    // distribution, the solver can not solve the hpd matrix even with single
+    // precision
```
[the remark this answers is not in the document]
yhmtsai: Although it is not tridiag one, we had test with 3pt stencil problem in the other test in this file.

reference/test/solver/batch_bicgstab_kernels.cpp
MarcelKoch: I would still suggest to remove this check for all value types.

MarcelKoch (approved):

pratikvn (approved):

PR #1720, opened 2024-10-31
Fixes formulas for Sor

MarcelKoch:
This PR fixes the SOR (and Gauss-Seidel) documentation.

### DIFF
--- a/include/ginkgo/core/preconditioner/sor.hpp
+++ b/include/ginkgo/core/preconditioner/sor.hpp
@@ -26,16 +26,16 @@ namespace preconditioner {
  * $A = D + L + U$, where $L$ contains all entries below the diagonal, and $U$
  * contains all entries above the diagonal. The application of the
  * preconditioner is then defined as solving $M x = y$ with
- * $$
+ * \f[
  * M = \frac{1}{\omega} (D + \omega L), \quad 0 < \omega < 2.
- * $$
+ * \f]
  * $\omega$ is known as the relaxation factor.
  * The preconditioner can be made symmetric, leading to the SSOR preconitioner.
  * Here, $M$ is defined as
- * $$
+ * \f[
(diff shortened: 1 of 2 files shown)


include/ginkgo/core/preconditioner/sor.hpp
```diff
@@ -26,16 +26,16 @@ namespace preconditioner {
...
- * $$
+ * \f[
  * M = \frac{1}{\omega} (D + \omega L), \quad 0 < \omega < 2.
- * $$
+ * \f]
  * $\omega$ is known as the relaxation factor.
  * The preconditioner can be made symmetric, leading to the SSOR preconitioner.
  * Here, $M$ is defined as
- * $$
+ * \f[
```
yhmtsai, line 35: I thought `\[` is enough, does it need to `\f[`?

MarcelKoch: No `\[` doesn't work, it has to be `\f[`.

yhmtsai: I assume you also look this page https://www.doxygen.nl/manual/formulas.html
maybe we or they have the tranformation from `$` to `\f$`? I wonder the transformation makes `\f[` to `\ff[`

MarcelKoch: I was a bit confused by that, somehow `\f$` didn't work, but `$`works as expected.

yhmtsai: I found we have used `\f[` and the formula is shown in https://ginkgo-project.github.io/ginkgo-generated-documentation/doc/develop/classgko_1_1matrix_1_1Fft.html.
I think we can rely on `\f[`

yhmtsai (approved):

PR #1721, opened 2024-11-05
Add function to create gko::config::pnode from std::string

greole:
This PR adds a function to create `gko::config::pnode` directly from a `std::string`.

Motivation: We would like to use Ginkgo's file config solver within the pyGinkgo bindings. The current implementation of the config solver seems to rely on json files. However, creating temporary files is a nuisance which can cause all kinds issues and is not really necessary in our case.  Thus, just passing a `std::string`  to generate a property tree would simplify things a lot.

### DIFF
--- a/include/ginkgo/extensions/config/json_config.hpp
+++ b/include/ginkgo/extensions/config/json_config.hpp
@@ -80,6 +80,13 @@ inline gko::config::pnode parse_json_file(std::string filename)
     return parse_json(nlohmann::json::parse(fstream));
 }
 
+/**
+ * parse_json_string takes a json string to generate the property tree object
+ */
+inline gko::config::pnode parse_json_string(std::string json)
+{
+    return parse_json(nlohmann::json::parse(json));
+}
 
 }  // namespace config
 }  // namespace ext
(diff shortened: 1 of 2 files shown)


include/ginkgo/extensions/config/json_config.hpp
```diff
@@ -80,6 +80,13 @@ inline gko::config::pnode parse_json_file(std::string filename)
     return parse_json(nlohmann::json::parse(fstream));
 }
 
+/**
+ * parse_json_string takes a json string to generate the property tree object
+ */
+inline gko::config::pnode parse_json_string(std::string json)
```
upsj, line 86: I wonder if it's a good time to start introducing `std::string_view` to the code base, now that we support C++17

upsj (approved):
LGTM, I thought we had this already

yhmtsai (changes requested):
need to add a test for it in extensions/test/json_config
question for pyGinkgo:
do you have any intermediate layer in C or C++ between ginkgo and pyGinkgo?

greole:
> need to add a test for it in extensions/test/json_config

sure i'll add one

> question for pyGinkgo:
> do you have any intermediate layer in C or C++ between ginkgo and pyGinkgo?

we use pybind11 , which is c++, to generate any intermediate code to register to python.

yhmtsai (approved):

Issue #1654, opened 2024-07-25
mingw and clang error


3rav:
I try to build msys2/mingw packages I have this error only for clang with dynamic library (static are ok):

```
lib/libginkgo_device.dll.a  -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32 && cd ."
  clang++: warning: argument unused during compilation: '-Wa,-mbig-obj' [-Wunused-command-line-argument]
  ld.lld: error: undefined symbol: gko::batch::matrix::Ell<float, int>::Ell(std::__1::shared_ptr<gko::Executor const>, gko::batch_dim<2ull, unsigned long long> const&, int)
```

yhmtsai:
Thanks for reporting this.
Unfortunately, we did not try clang on windows before.
Could you share your configure setup with us?

3rav:
The error still occurs in clang64: https://github.com/3rav/MINGW-packages/actions/runs/11712157728/job/32622223182

```
  clang++: warning: argument unused during compilation: '-Wa,-mbig-obj' [-Wunused-command-line-argument]
  ld.lld: error: undefined symbol: gko::batch::matrix::Ell<float, int>::Ell(std::__1::shared_ptr<gko::Executor const>, gko::batch_dim<2ull, unsigned long long> const&, int)
  >>> referenced by reference/CMakeFiles/ginkgo_reference.dir/solver/batch_bicgstab_kernels.cpp.obj:(gko::EnablePolymorphicObject<gko::batch::matrix::Ell<float, int>, gko::batch::BatchLinOp>::create_default_impl(std::__1::shared_ptr<gko::Executor const>) const)
  >>> referenced by reference/CMakeFiles/ginkgo_reference.dir/solver/batch_bicgstab_kernels.cpp.obj:(gko::EnablePolymorphicObject<gko::batch::matrix::Ell<float, int>, gko::batch::BatchLinOp>::clear_impl())
  >>> referenced by reference/CMakeFiles/ginkgo_reference.dir/solver/batch_cg_kernels.cpp.obj
  
  ld.lld: error: undefined symbol: gko::batch::matrix::Dense<float>::Dense(std::__1::shared_ptr<gko::Executor const>, gko::batch_dim<2ull, unsigned long long> const&)
[...]

PR #1724, opened 2024-11-12
Fix dynamic cast leads the undefined symbol in clang/msys2/windows 

yhmtsai:
It is reported in #1654 . We have some undefined symbol when building the shared library (libginkgo_reference) by the clang-18 from msys2 in windows.

In clang 18 (at least), the compiler include `convert_to`/`move_to`/... virtual function as undefined symbols when we use `dynamic_cast`.  These symbols are from `ConvertibleTo<..>`, which contain general function like `convert_to` calling virtual functions, such that these functions are considered in the library.
It can be observed with clang-18 on linux, too. However, there are different behaviors between systems.
- Linux - allow any undefined symbol when building shared library. 
- MacOS - allow undefined symbol with `-undefined dynamic_lookup`
- Windows (specific to clang in msys2) - clang ld.lld has `--allow-shlib-undefined`, which is enabled by default, but it seems to be available in Linux https://man.archlinux.org/man/extra/lld/ld.lld.1.en In the environment, I check the help documentation and the linker also reports unknown argument when passing the argument.

The older compiler or gcc compiler put these function as weak symbol, so it does not complain undefined symbols.

This PR moves the matrix/preconditioner dispatch from the kernels to core such that there are no dynamic_cast in the codes.

It also needs to add `-fno-assume-unique-vtables` to clang/msys2/windowns to make the dynamic_cast to the class marked by final work.
https://reviews.llvm.org/D154658 should be relevant

TODO:
- [x] make the macros work in MSVC


core/base/batch_instantiation.hpp
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @param _macro  A macro which expands the template instantiation
+ *                (not including the leading `template` specifier).
+ *                Should take three arguments, where the first is replaced by
+ *                the value type, the second by the matrix, and the third by the
+ *                preconditioner.
+ *
+ * @note the second and third arguments only accept the base type.s
+ */
+#define GKO_INSTANTIATE_FOR_BATCH_VALUE_MATRIX_PRECONDITIONER(_macro)          \
```
MarcelKoch, line 31: I think this needs to be unified with the changes from #1629, but I don't immediately see how to do that.

MarcelKoch (changes requested):

core/base/batch_instantiation.hpp
yhmtsai: I reuse the part in types.hpp.

yhmtsai: I update the macro usage a bit such that make the for-stack easy and clear.
I think it can not reuse the main stack of dispatch type.
one dispatches the device data structure which are only available in `namespace cuda/hip/...`, but the other uses the main class.

core/solver/batch_dispatch.hpp
```diff
@@ -201,22 +195,12 @@ enum class log_type { simple_convergence_completion };
  * GKO_BATCH_INSTANTIATE will be prepended to the batch solver template
  * parameters.
  */
-#define GKO_BATCH_INSTANTIATE_VARGS(macro, ...) \
```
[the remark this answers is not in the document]
yhmtsai: I can remove it for these, because I do not need the macro explicitly in the first place of these macro.
In value_type variants, we need macro as an argument explicitly to call `macro(...)`.
Thus, we need one version for `macro(...)` and another VARGS version for `macro(..., __VA_ARGS__)`

core/base/batch_instantiation.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+namespace gko {
+namespace batch {
+
+
+// just make the call list more consistent
+#define GKO_CALL(_macro, ...) _macro(__VA_ARGS__)
+
+#define GKO_BATCH_INSTANTIATE_PRECONDITIONER(_next, ...)            \
+    _next(GKO_INDIRECT(__VA_ARGS__), gko::batch::matrix::Identity); \
```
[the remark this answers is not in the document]
yhmtsai: I was using it wrongly.
no, I think it is required here.
in MSVC, they consider the __VA_ARGS__ as one arguments by default.
`GKO_CALL(A_loop, B_loop, ...)` -> A_loop("B_loop, ...").
will just be `"B_loop, ..."(A_1)` `"B_loop, ..."(A_2)` -> only the first one get expanded properly, the B_loop, ... is considered as one arguments.  no `__VA_ARGS__` in A_loop.

I need to use `GKO_INDIRECT` to take the first arguments out of `__VA_ARGS__` when it may have the nested call.

MarcelKoch (approved):

dpcpp/solver/batch_bicgstab_kernels.dp.cpp
```diff
@@ -168,21 +169,21 @@ class kernel_caller {
 };
 
 
-template <typename ValueType>
+template <typename ValueType, typename BatchMatrixType, typename PrecType>
 void apply(std::shared_ptr<const DefaultExecutor> exec,
            const settings<remove_complex<ValueType>>& settings,
-           const batch::BatchLinOp* mat, const batch::BatchLinOp* precond,
+           const BatchMatrixType* mat, const PrecType* precon,
```
upsj, line 175: nit: I would leave it as `precond`

upsj (approved):
LGTM! Just a naming nit

Issue #1675, opened 2024-09-03
Unable to build on Windows


jrobcary:
I am trying to build windows static.  I configure with

'/winsame/cary/dev3/contrib-llvm17_vs2022/cmake-3.26.4-ser/bin/cmake.exe' \
  -DCMAKE_INSTALL_PREFIX:PATH=D:/winsame/cary/dev3/contrib-llvm17_vs2022/ginkgo-1.8.0-ser \
  -DCMAKE_BUILD_TYPE:STRING=Release \
  -DCMAKE_COLOR_MAKEFILE:BOOL=FALSE \
  -DCMAKE_VERBOSE_MAKEFILE:BOOL=TRUE \
  -DCMAKE_C_COMPILER:FILEPATH='C:/Program Files/LLVM17/bin/clang-cl.exe' \
  -DCMAKE_CXX_COMPILER:FILEPATH='C:/Program Files/LLVM17/bin/clang-cl.exe' \
  -DCMAKE_C_FLAGS:STRING='/W3 -Wno-implicit-function-declaration' \
  -DCMAKE_CXX_FLAGS:STRING='/W3 /GR /EHsc /Zc:__cplusplus /D_HAS_AUTO_PTR_ETC /bigobj -fms-compatibility-version=19 -Wno-register -Wno-implicit-function-declaration -Wno-microsoft-include' \
  -DBUILD_SHARED_LIBS=OFF \
  -DCMAKE_CUDA_COMPILER:FILEPATH='C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v11.8/bin/nvcc.exe' \
  -DCMAKE_CUDA_ARCHITECTURES:STRING='61;70;80' \
  -DGINKGO_BUILD_HIP:BOOL=OFF \
  -DGINKGO_BUILD_OMP:BOOL=OFF \
  -DGINKGO_BUILD_MPI:BOOL=OFF \
  -G 'NMake Makefiles JOM' D:/winsame/cary/xsd3/builds/ginkgo-1.8.0

But when I build (with jom), it dies at
```

[...]

PR #1725, opened 2024-11-14
fix constructor order, friend class, inconsistent function, inconsistent class/struct

yhmtsai:
From #1675 , we can not use clang-cl on windows to build ginkgo (maybe only static build, not sure the shared build)
The reason is the inconsistence between the declaration and declaration of definition.
For example, 
We declare `void func(int *a);`, which is seen by ginkgo_core. We also use it to instantiate the function.
However, we implement it like `void func(int *const a) {...}`.
The symbol in the library from clang-cl will use `void func(int *const a)` not `void func(int *a)`.
core only know `void func(int *a)` such that it can not find the symbol properly although the `int* a` should be fit into `int* const a` 

More than that, I also fix the constructor order is different from the data member order, microsoft does not like the friend class without namespace not in the nearest namespace, and inconsistent class/struct.

also, add some tricks to avoid alais undeclared in the a structure which is in another tempalted structure.
It is mentioned in https://github.com/llvm/llvm-project/issues/64996
we use `using .. = ..` in the nested struct as workaround here


upsj (approved):
LGTM! Great catch, I would have thought that `-Wreorder` can catch some of these issues, and I must have missed the name mangling capturing parameter const-ness.

yhmtsai:
I found the const-ness issue by lucky. I do not have an idea that any tool can help detecting this situation.

PR #1726, opened 2024-11-15
workaround for hanging issue on A770 action

yhmtsai:
The default sub-group size 8 for most kernels (I assume) is available on A770, but it can not run or terminiate successfully with this kind of kernels. It might be from some mismatch between driver/kernel/compiler.
We pass the `-fsycl-default-sub-group-size=16` such that it does not hang for now.
Interestingly, the cooperative group with sub-group size 8 works for now.
I was inspired by that because job with 8 hangs but job with 16 works.

### DIFF
--- a/.github/workflows/intel.yml
+++ b/.github/workflows/intel.yml
@@ -29,16 +29,17 @@ jobs:
     steps:
     - name: Checkout the latest code (shallow clone)
       uses: actions/checkout@v4
-
+    # We set -fsycl-default-sub-group-size=16 to avoid the A770 hanging issue.
+    # The default configuration makes that A770 hangs during running the kernel without specifying sub-group size
     - name: configure
       run: |
         source /etc/profile
-        module load intel-oneapi-compilers intel-oneapi-dpl intel-oneapi-mkl cmake
+        module load intel-oneapi-compilers intel-oneapi-dpl intel-oneapi-mkl cmake ninja
         mkdir build
         cd build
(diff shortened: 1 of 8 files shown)


.github/workflows/intel.yml
```diff
@@ -29,16 +29,17 @@ jobs:
...
       run: |
         source /etc/profile
-        module load intel-oneapi-compilers intel-oneapi-dpl intel-oneapi-mkl cmake
+        module load intel-oneapi-compilers intel-oneapi-dpl intel-oneapi-mkl cmake ninja
         mkdir build
         cd build
-        cmake .. -DCMAKE_INSTALL_PREFIX=install_ginkgo -DCMAKE_CXX_FLAGS="-Wpedantic -ffp-model=precise" -DCMAKE_CXX_COMPILER=${{ matrix.config.compiler }} -DCMAKE_BUILD_TYPE=${{ matrix.config.build_type }} -DGINKGO_MIXED_PRECISION=${{ matrix.config.mixed }} -DGINKGO_BUILD_CUDA=OFF -DGINKGO_BUILD_HIP=OFF -DGINKGO_BUILD_MPI=OFF -DGINKGO_DPCPP_SINGLE_MODE=ON
-        make -j8
-        ONEAPI_DEVICE_SELECTOR=level_zero:gpu ctest -j10 --output-on-failure
+        cmake .. -GNinja -DCMAKE_INSTALL_PREFIX=install_ginkgo -DCMAKE_CXX_FLAGS="-Wpedantic -ffp-model=precise -fsycl-default-sub-group-size=16 -Wno-unused-command-line-argument -Wno-deprecated" -DCMAKE_CXX_COMPILER=${{ matrix.config.compiler }} -DCMAKE_BUILD_TYPE=${{ matrix.config.build_type }} -DGINKGO_MIXED_PRECISION=${{ matrix.config.mixed }} -DGINKGO_BUILD_CUDA=OFF -DGINKGO_BUILD_HIP=OFF -DGINKGO_BUILD_MPI=OFF -DGINKGO_DPCPP_SINGLE_MODE=ON
```
upsj, line 40: I would keep `-Wdeprecated` around, especially now that we are starting to fix those deprecated usages.

upsj (approved):
LGTM!

yhmtsai:
the workaround is not needed anymore as the machine runs normally now

PR #1727, opened 2024-11-15
Always enable nlohmann_json with examples

pratikvn:
Enable the file-config example conditionally, as it depends on nlohmann-json.


upsj (changes requested):
find_package should ideally always happen on the topmost level, so that we don't look for different configurations accidentally. I would suggest moving the call to the root CMakeLists.txt. Is there a downside to always pulling in the dependency? I would like to avoid too many fragmented possible configurations depending on small environment differences.

examples/CMakeLists.txt
```diff
@@ -64,6 +63,14 @@ else()
     message(STATUS "No Kokkos found, disabling examples with Kokkos assembly.")
 endif()
 
+set(GKO_FILE_CONFIG_ENABLED 0)
```
yhmtsai, line 66: I think if you also add GINKGO_BUILD_EXAMPLES in the third_party CMakeLists, you do not need the change here

pratikvn: Then we always pull in nlohmann_json when examples is built. But I guess that is okay.

yhmtsai (approved):

examples/CMakeLists.txt
```diff
@@ -110,16 +110,15 @@ if(GINKGO_BUILD_TESTS)
...
-                     WORKING_DIRECTORY
-                     "$<TARGET_FILE_DIR:ginkgo>")
+            add_test(NAME example_file-config-solver_${config_name}_${executor}
+                COMMAND
+                "$<TARGET_FILE:file-config-solver>"
+                "${executor}" "${CMAKE_CURRENT_SOURCE_DIR}/file-config-solver/config/${config_name}.json"
+                "${CMAKE_CURRENT_SOURCE_DIR}/file-config-solver/data/A.mtx"
+                WORKING_DIRECTORY
+                "$<TARGET_FILE_DIR:ginkgo>")
             # Prevent performance issues with high core counts
```
upsj, line 120: This doesn't actually change anything, right? I would just revert it then for a cleaner diff

upsj (approved):
LGTM!

PR #1733, opened 2024-11-25
Update cuda arch 8.9

yhmtsai:
This PR adds Ada to sSMtoCores, adds ada and hopper in selector

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,7 +72,7 @@ gko_rename_cache(GINKGO_CUDA_COMPILER_FLAGS CMAKE_CUDA_FLAGS BOOL "Flags used by
 # load executor-specific configuration
 if(GINKGO_BUILD_CUDA)
     include(cmake/cuda.cmake)
-    if(GINKGO_CUDA_CUSTOM_THRUST_NAMESPACE AND CUDAToolkit_VERSION VERSION_LESS 11.6)
+    if(CUDAToolkit_VERSION VERSION_LESS 11.6)
         message(STATUS "Disable custom thrust namespace for cuda before 11.6 because it has no effect in the thrust shipped by cuda before 11.6")
         set(GINKGO_CUDA_CUSTOM_THRUST_NAMESPACE OFF)
     else()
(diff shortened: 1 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -108,6 +108,64 @@ build/cuda110/nompi/gcc/cuda/release/shared:
     # disable spurious unused argument warning
     EXTRA_CMAKE_FLAGS: "-DCMAKE_CUDA_FLAGS=-diag-suppress=177"
 
+# cuda 11.4
```
MarcelKoch, line 111: Why are these versions necessary? We already test 11.0, 11.7, 12.0. I could understand if we also add the latest cuda version (12.6), but these seem quite random and unnecessary to me.

yhmtsai: Not necessary but good to have.
I consider all versions are different, so it will be great if we test all.
Also, something in nvhpc is different from the normal setup in my opinion.
We do not do that actually just limited by the resource.
Also, I only pick up those in the docker hub now because we do not process the our container currently.

MarcelKoch: We can't test all versions, so my question still remains, are these versions so important that we need to add extra jobs, each might take ~40--60min, to our CI?  Until I see a compelling reason, I will block this particular change. 

Also, I wouldn't want to change our CI this close to the release. After the release, we should restructure our CI anyway to be a bit less random in which versions we are using.

MarcelKoch (changes requested):
The CI changes should be removed, see the comment thread.
Otherwise LGTM.

.gitlab-ci.yml
yhmtsai: I am seeing it in the opposite side. I do not have idea why we drop these version accendidently. The version jump is roughly 3 minor version in nvcc side. They change some components version also with updating the version. As #1730 , we do not test it a lot because only two of them only in nvhpc (I still think nvhpc is different from nvcc with the same cuda version). To get more confident with the thrust without giving the user option to turn it off by comments, which actually made something broken if we do not provide any fix, I manually run additionaly 4 version to check.

MarcelKoch (approved):

upsj (approved):

PR #1736, opened 2024-12-02
Rename half helper

yhmtsai:
This PR renames the temporary naming scheme for half.

- precision chain
`next_precision` -> `next_precision_base` (only float and double)
`next_precision_with_half` -> `next_precision` (It performs the same as the previous release if `GINKGO_ENABLE_HALF` is `OFF`)
- instantiation
`GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE` -> `GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE_BASE`
`GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE_WITH_HALF` -> `GKO_INSTANTIATE_FOR_EACH_VALUE_TYPE`
- testing type
`ValueTypes` -> `ValueTypesBase`
`ValueTypesWithHalf` -> `ValueTypes` 
- value type list (config)
`value_type_list()` -> `value_type_list_base()`
`value_type_list_with_half()` -> `value_type_list()`

We might go for the same way as what we did in the dpcpp_single_mode, which keeps the interface but throw the error for double precision implementation in the future. From the public interface perspective, we can go from the current status to there without hurting, but not good via the other direction. Thus, we keep the original plan for this release now.

The following script (not simple rule though)
```
# precision chain
regex="s/(next|previous)_precision(_impl)?</\1_precision_base\2</g;
s/(struct|using) (next|previous)_precision(_impl)? /\1 \2_precision_base\3 /g;
s/(next|previous)_precision_with_half(_impl)?</\1_precision\2</g;
s/(struct|using) (next|previous)_precision_with_half(_impl)? /\1 \2_precision\3 /g";
# instantiation
regex="s/(GKO_[a-zA-Z_0-9]*(VALUE|TEMPLATE|POD)[a-zA-Z_0-9]*[^F])([(,])/\1_BASE\3/g;s/__BASE/_BASE_/g;s/(GKO_[a-zA-Z_0-9]*(VALUE|TEMPLATE|POD)[a-zA-Z_0-9]*)_WITH_HALF([(,])/\1\3/g"
# testing type
regex="s/([a-zA-Z0-9]*(Value|POD)[a-zA-Z0-9]*Types)([ ,>;])/\1Base\3/g;
s/([a-zA-Z0-9]*(Value|POD)[a-zA-Z0-9]*Types)WithHalf([ ,>;])/\1\3/g; s/TwoValueIndexTypeWithHalf/TWithHalf/g;s/TwoValueIndexType/TwoValueIndexTypesBase/g;s/TWithHalf/TwoValueIndexTypes/g"
# config type
regex="s/value_type_list_with_half/T_with_half/g;s/value_type_list/value_type_list_base/g;s/T_with_half/value_type_list/g"

# clean newline
regex="$(echo $regex | tr -d '\n')"
```
using it with `sed -E`

```
find . -type f -not -path ./build* -not -path ./.* -not -name '*.sh' -not -name '*.png' -not -name '*.pdf' -not -name '*.ico' -exec ./rename.sh {} \;
```

the last commit contains tiny manual change to turn some unnecessary change as they are helper without type information actually


test/matrix/matrix.cpp
```diff
@@ -557,7 +557,7 @@ class Matrix : public CommonTestFixture {
     using Mtx = typename T::matrix_type;
     using index_type = typename Mtx::index_type;
     using value_type = typename Mtx::value_type;
-    using mixed_value_type = gko::next_precision<value_type>;
+    using mixed_value_type = gko::next_precision_base<value_type>;
```
MarcelKoch, line 560: What about `gko::next_precision` here?

MarcelKoch: I guess it doesn't matter, since `value_type` is always double. So I would prefer the non-base version.

test/solver/solver.cpp
```diff
@@ -520,7 +520,7 @@ class Solver : public CommonTestFixture {
     using Precond = typename T::precond_type;
     using Mtx = typename T::matrix_type;
     using value_type = typename Mtx::value_type;
-    using mixed_value_type = gko::next_precision<value_type>;
+    using mixed_value_type = gko::next_precision_base<value_type>;
```
MarcelKoch, line 523: Same here

test/matrix/csr_kernels2.cpp
```diff
@@ -34,7 +34,7 @@ class Csr : public CommonTestFixture {
 protected:
     using Arr = gko::array<int>;
     using Vec = gko::matrix::Dense<value_type>;
-    using Vec2 = gko::matrix::Dense<gko::next_precision<value_type>>;
+    using Vec2 = gko::matrix::Dense<gko::next_precision_base<value_type>>;
```
MarcelKoch, line 37:
```suggestion
    using Vec2 = gko::matrix::Dense<gko::next_precision<value_type>>;
```
Although I think this is not used at all. (But that is nothing to change in this PR.)

test/reorder/amd.cpp
```diff
@@ -40,7 +40,8 @@ class Amd : public CommonTestFixture {
     std::shared_ptr<matrix_type> dmtx;
 };
 
-TYPED_TEST_SUITE(Amd, gko::test::ValueIndexTypes, PairTypenameNameGenerator);
+TYPED_TEST_SUITE(Amd, gko::test::ValueIndexTypesBase,
+                 PairTypenameNameGenerator);
```
MarcelKoch, line 44: Necessary?

core/config/parse_macro.hpp
```diff
@@ -16,7 +16,8 @@
 
 
 // for value_type only
-#define GKO_PARSE_VALUE_TYPE_(_type, _configurator, _value_type_list)        \
+#define GKO_PARSE_VALUE_TYPE_BASE_(_type, _configurator,                     \
+                                   _value_type_list_base)                    \
```
MarcelKoch, line 20: Necessary?

core/config/parse_macro.hpp
```diff
@@ -28,21 +29,22 @@
         return gko::config::dispatch<gko::LinOpFactory, _configurator>(      \
             config, context, updated,                                        \
             gko::config::make_type_selector(updated.get_value_typestr(),     \
-                                            _value_type_list));              \
+                                            _value_type_list_base));         \
```
MarcelKoch, line 32: Same

core/config/parse_macro.hpp
```diff
@@ -28,21 +29,22 @@
...
-                          gko::config::value_type_list_with_half())
+#define GKO_PARSE_VALUE_TYPE(_type, _configurator)   \
+    GKO_PARSE_VALUE_TYPE_BASE_(_type, _configurator, \
+                               gko::config::value_type_list())
 
 // for value_type and index_type
-#define GKO_PARSE_VALUE_AND_INDEX_TYPE_(_type, _configurator,                 \
-                                        _value_type_list)                     \
+#define GKO_PARSE_VALUE_AND_INDEX_TYPE_BASE_(_type, _configurator,            \
+                                             _value_type_list_base)           \
```
MarcelKoch, line 47: Same

core/config/parse_macro.hpp
```diff
@@ -54,21 +56,21 @@
         return gko::config::dispatch<gko::LinOpFactory, _configurator>(       \
             config, context, updated,                                         \
             gko::config::make_type_selector(updated.get_value_typestr(),      \
-                                            _value_type_list),                \
+                                            _value_type_list_base),           \
```
MarcelKoch, line 59: same

MarcelKoch (changes requested):
I think these are the only instances I could find where the 'normal' version can be used instead if the '*Base' version.

I can't go through every file by itself, so I just searched where the '*Base' version still appear.

test/matrix/matrix.cpp
yhmtsai: It matters when enabling dpcpp_single_mode, `next_precision_base<float>` will be double, which is what we test now, but `next_precision<float>` will be half

test/reorder/amd.cpp
yhmtsai: good catch! It can pass with half. It was missed in the reorder commit

core/config/parse_macro.hpp
yhmtsai: No, I manually turn them back

test/reorder/amd.cpp
yhmtsai: I also change it for core/test/reorder/amd

test/matrix/csr_kernels2.cpp
yhmtsai: It is the same reason for keeping that. the value_type will be float with enabling dpcpp_single_mode.
I would rather than deleting them because they are not used.

PR #1742, opened 2024-12-05
support properly when disable half precision

yhmtsai:
Nvidia GPU before compute capability 5.3 does not support IEEE FP16 (half) precision operation.
However, cuda 12.2+ relax these some limitation, so we still support these GPUs with `GINKGO_ENABLE_HALF=ON` after cuda 12.2.
We only support these GPUs with `GINKGO_ENABLE_HALF=OFF` before cuda 12.2

- [x] wait for the check on the real GPU below the requirment

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -26,7 +26,8 @@ Ginkgo adds the following additional switches to control what is being built:
     Enabling this flag increases the library size, but improves performance of
     mixed-precision kernels.
 *   `-DGINKGO_ENABLE_HALF={ON, OFF}` enable half precision support in Ginkgo, default is `ON`.
-    It is `OFF` when the compiler is MSVC.
+    It is `OFF` when the compiler is MSVC. If compiling is done with the CUDA backend before CUDA 12.2, 
+    we only support half precision after compute capability 5.3. CUDA 12.2+ compilers waive the compute capbility limitation.
 *   `-DGINKGO_BUILD_TESTS={ON, OFF}` builds Ginkgo's tests
     (will download googletest), default is `ON`.
 *   `-DGINKGO_FAST_TESTS={ON, OFF}` reduces the input sizes for a few slow tests
--- a/common/cuda_hip/base/math.hpp
+++ b/common/cuda_hip/base/math.hpp
@@ -104,6 +104,9 @@ struct is_complex_or_scalar_impl<thrust::complex<T>>
 }  // namespace gko
 
 
+#if GINKGO_ENABLE_HALF
+
+
 GKO_THRUST_NAEMSPACE_PREFIX
 namespace thrust {
 
@@ -184,6 +187,9 @@ __device__ __forceinline__ bool is_finite(const thrust::complex<__half>& value)
 
 
 }  // namespace gko
+
+
+#endif  // GINKGO_ENABLE_HALF
 
 
 #endif  // GKO_COMMON_CUDA_HIP_BASE_MATH_HPP_


upsj:
Just a nit with the description: There is no 5.3, the next compute capability is 6.0.

upsj (approved):
LGTM after fixing 5.3

upsj:
I removed the last 5.2 device I had, because the system kept running into kernel panics. I would be fine if we didn't test this, or pushed the device requirements to 6.0, because that's the only architecture we actually have GPUs available for.

yhmtsai:
5.3 is listed in cuda 11.0

upsj:
That's the Jetson Nano, I don't think we need to consider them.

pratikvn:
The Tegra X1 and Jetson TX1 GPUs are 5.3. They officially do have half precision FP support, so technically with this PR we should be able to run on them. But I am not sure it is easy to confirm that, as we dont have access to those.

upsj:
I was thinking earlier about the term "supported" - what does it mean when we state that we support certain versions? We still list classical Intel compiler support, but we no longer have pipelines for it. Does it mean it _should_ work, so does it mean we will fix issues if it doesn't work, or does it mean we are actively monitoring it via CI?

MarcelKoch:
> I was thinking earlier about the term "supported" - what does it mean when we state that we support certain versions? We still list classical Intel compiler support, but we no longer have pipelines for it. Does it mean it _should_ work, so does it mean we will fix issues if it doesn't work, or does it mean we are actively monitoring it via CI?

I think it means the first on. We also list GCC 7+, so we expect that all versions 7.0-14.? work, but of course we are not (and will not) testing every version in our CI.

upsj:
Fixing issues means we should also be able to confirm when they are fixed. If we don't have the necessary hardware available, I would try to avoid claiming we can support them.

pratikvn:
Some libraries split them into two: `Minimum supported version (confirmed/tested with CI): x.x.x` and `Might work (untested with CI, but might work)`, but I guess we cant really call the latter "Supported", and can say that we support only versions/hardware we test on.

upsj:
Yes, I was thinking along the same lines. If we split up the list of supported versions, we can be more explicit.

upsj:
We have P100, Titan X, V100, RTX 2060, A2 (potentially A100 if we need it), L40S and H100 available, so we could cover the full range if we wanted to.

yhmtsai:
Also, I do not think additional claim on the testing hardware help anything.
Users run it with V100 or A100 for example and face issues. We will still need to fix it, right?

upsj:
For CPUs: Contrary to SASS, x86_64 is stable between hardware architectures (we should likely add a variant of aarch64), and we don't compile with `-march=native`, so beyond CPU bugs, it is unlikely to lead to different behaviors between microarchitectures.

INSTALL.md
```diff
@@ -26,7 +26,7 @@ Ginkgo adds the following additional switches to control what is being built:
     Enabling this flag increases the library size, but improves performance of
     mixed-precision kernels.
 *   `-DGINKGO_ENABLE_HALF={ON, OFF}` enable half precision support in Ginkgo, default is `ON`.
-    It is `OFF` when the compiler is MSVC.
+    It is `OFF` when the compiler is MSVC. If compiling it with CUDA backend, we only support half precision after compute capability 5.3.
```
thoasm, line 29: Nit (improving the text and limit width to 80 characters):
```suggestion
    It is `OFF` when the compiler is MSVC. If compiling is done with the CUDA
    backend, we only support half-precision after compute capability 5.3.
```

thoasm (approved):
LGTM

PR #1743, opened 2024-12-06
TSAN fixes

MarcelKoch:
This PR includes fixes for errors reported by the TSAN. The changes include:

- RCM: locking the write mutex when checking the queue size 
- SparsityCsr: using atomic load/store when checking of other threads report that the matrix is unsorted
- Par IC/ILU (t): use atomic load/stores in the same way as the CUDA/HIP implementations
- MG: 'fix' a write-after-write race in the kcycle stopping criteria. Honestly, I don't think this is necessary, but it stops TSAN from complaining.
- Batch Solvers: fixed the generation of the 3pt stencil matrix for tests, which caused data races.
- NOT FIXED: the PGM contains data races in the `match_edge` function. I didn't fix this, since that would require to extract the common kernels implementations into each backend, which I found a bit overkill. There is also a data race in the non-deterministic part of `assign_to_exist_agg`, but I ignored that, since it is marked as non-deterministic anyway.

### DIFF
--- a/omp/components/atomic.hpp
+++ b/omp/components/atomic.hpp
@@ -100,6 +100,18 @@ inline void store(float* addr, float val)
     *addr = val;
 }
 
+inline void store(int32* addr, int32 val)
+{
+#pragma omp atomic write
+    *addr = val;
+}
+
+inline void store(int64* addr, int64 val)
+{
+#pragma omp atomic write
+    *addr = val;
+}
+
 inline void store(half* addr, half val)
 {
     auto uint_addr = copy_cast<uint16_t*>(addr);
@@ -129,6 +141,22 @@ inline float load(float* addr)
 inline double load(double* addr)
 {
     double val;
+#pragma omp atomic read
+    val = *addr;
+    return val;
+}
+
+inline int32 load(int32* addr)
+{
+    float val;
+#pragma omp atomic read
+    val = *addr;
+    return val;
+}
+
+inline int64 load(int64* addr)
+{
+    float val;
 #pragma omp atomic read
     val = *addr;
     return val;
--- a/omp/components/atomic.hpp
+++ b/omp/components/atomic.hpp
@@ -52,7 +52,7 @@ inline ResultType copy_cast(const ValueType& val)
 
 
 template <>
-void atomic_add(half& out, half val)
+inline void atomic_add(half& out, half val)
 {
 #ifdef __NVCOMPILER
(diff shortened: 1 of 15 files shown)


omp/components/atomic.hpp
```diff
@@ -85,6 +85,115 @@ void atomic_add(half& out, half val)
 }
 
 
+// There is an error in Clang 17 which prevents us from merging the
+// implementation of double and float. The compiler will throw an error if the
+// templated version is implemented. GCC doesn't throw an error.
+template <typename T>
+struct atomic_store_helper;
```
upsj, line 92: Do we actually need this? Or can we get the same result with just overloaded functions? We only use atomics for a fixed set of types (value types and index types)

upsj (approved):
LGTM! Two points I would like to discuss (shouldn't impact the PR though):
1. naming - the functions in CUDA and HIP are named `load_...`, where the fact that a memory ordering is specified makes it clear that we are dealing with atomics. For consistency it would be best if OpenMP and CUDA/HIP use the same names. They can become pretty long though (`load_relaxed_local`), so I would like to avoid adding another `atomic_` prefix to them. Do we need to use it?
2. With PGM being the only kernel that has this issue, I would probably suggest splitting it into OpenMP, SYCL and CUDA/HIP again, also since there is some optimization potential in terms of load balancing and memory access patterns. The alternative (putting atomic operations into the unified headers) seems to invite using the unified kernels for too much.

omp/components/atomic.hpp
MarcelKoch: You're right. I first wanted to use the templated implementation, but as that doesn't work with clang, we can just use simple overloads.

MarcelKoch:
@upsj regarding 1.: I will go with just `load|store`, since OpenMP 4.5 doesn't support specifying the memory ordering.

Regarding 2.: I would also not support adding atomics for the unified kernels. I've added now the separate kernels, it now a bit more LOC, but the kernels are simple enough that it should be fine.

upsj:
I would consider still adding the `_relaxed` suffix, because this is the weakest possible atomic memory ordering, and likely close to what the default would be (just to avoid accidentally relying on too much ordering between operations)

MarcelKoch:
I would rather leave it out to indicate that we can't make assumptions on the memory ordering at all.

upsj:
relaxed is as close to "no assumptions" as possible in terms of memory ordering. But not pushing any harder than that 😄

PR #1746, opened 2024-12-09
Release 1.9.0

MarcelKoch:
The Ginkgo team is proud to announce the new Ginkgo minor release 1.9.0.

This release brings new features such as:
- Support for half precision (IEEE FP16). The type `gko::half` can now be selected in most instances as the value type
  of a matrix, solver, preconditioner, etc. If the selected backend supports FP16 as a native type, the native type is
  used within the kernels, otherwise an overhead might occur. The new behavior is enabled by default, but it can be
  turned off during configuration.
- New implementations of the ILU and IC factorization for CUDA, HIP, OpenMP, and Reference backends. These are
  available in addition to the existing implementations based on the vendor libraries cuSPARSE and hipSPARSE.
- New (S)SOR and Gauss-Seidel preconditioners.
- Simplifyied distributed matrix assembly by exchanging local rows between neighboring processes.

And more!

See [the Changelog](./CHANGELOG.md) for more information.


pratikvn (dismissed):

yhmtsai (approved):

MarcelKoch:
The check-format PR seems to fail because it uses some old configuration. I locally ran out pre-commit checks after locally merging develop into main, and no issues were reported. So, I would just continue.

pratikvn (approved):

PR #1748, opened 2024-12-09
Master release/1.9.0

MarcelKoch:
This PR updates `master` to the latest state of develop. This is done by cherry-picking all commit between the v1.8.0 and v1.9.0 release, using:
```bash
git log --format=%H --no-merges --skip=1 --reverse <prev-release>..v1.9.0 | git cherry-pick --stdin
```
The commit for `<prev-release>` was found using:
```bash
git log --oneline --merges --first-parent --skip=1 -n 1 v1.9.0
```
In future releases, this should not be necessary, as the release tags can just be used directly.

Some conflicts needed to be resolved manually, but those were very few and only related to the tag name.

Edit: It was necessary to cherry-pick also the first commit after v1.8.0, since this contained the version change in the CMake files. To apply it, some conflicts had to be resolved manually. Additionally one commit had to be cherry-picked again, because the conflict was resolved incorrectly. @upsj provided this command to easily check the difference between main and the new master release:
```
git diff origin/master-release/1.9.0 origin/main --word-diff --word-diff-regex=.
```


upsj:
Would using `git merge` be an option? The worktree should look exactly the same as `main`, so merging should be possible.

MarcelKoch:
No, `master` and `main` don't have a common history since the 1.0 release. So merging `main` into master would introduce ~6000 commits.

yhmtsai:
JOSS is failed due to old upload artifact.

pratikvn:
I think we can ignore the joss pandoc job for this and main branch. I think we can remove that job entirely in develop. I guess currently it has been setup to run on master and also on main somehow in the settings, which can probably be updated.

pratikvn (approved):

upsj (changes requested):
There seem to be some minor changes missing, e.g. version numbers in `examples`, see 
```bash
git diff origin/master-release/1.9.0 origin/main --word-diff --word-diff-regex=.
```

MarcelKoch:
@upsj should be fixed now.

upsj (approved):

PR #1747, opened 2024-12-09
Prepare Develop for 1.10 Release

MarcelKoch:
This PR prepares the develop branch for the next release.

PR #1750, opened 2024-12-10
Remove JOSS job

pratikvn:
Remove the job that compiled the paper for JOSS. The files used (`doc/joss/*`) have still been left as is for reference, but I think the job is not needed anymore.


upsj (approved):
LGTM! There are some people approaching software publications as "living papers" that evolve alongside the code and documentation, but I don't think we need to do that with the JOSS paper.

yhmtsai:
I think it is the JOSS idea, right?
How does it work with changed content? Originally, I thought we only need to update it by ourselve. Do we need to request a review from JOSS site?
We should update some outdated information like we use another json package not rapidjson anymore.

pratikvn:
I dont think they have support for version updates like Zenodo. It would have to be a new paper (with new reviews and so on), which I think is possible, but I dont think we need to do that, as I dont see any real benefits from that.

yhmtsai:
okay, it need to wait until #1747 is merged

PR #1751, opened 2024-12-11
[dist] remove distributed version of EnableLinOp

MarcelKoch:
This PR removes the `EnableDistributedLinOp` mixin class. Instead, the distributed linops can now directly derive from `EnableLinOp`. To make this work, the `EnablePolymorphicObject` now checks if the `ConcreteType` is derived from `DistributedBase`, and uses the `(exec, comm)` constructor in that case. (This is now possible due to the `if constexpr` in c++17.)


include/ginkgo/core/base/polymorphic_object.hpp
```diff
@@ -15,6 +15,15 @@
...
 namespace gko {
+namespace experimental {
+namespace distributed {
+
+
+class DistributedBase;
+
+
+}
+}  // namespace experimental
```
yhmtsai, line 26:
```suggestion
}  // namespace distributed
}  // namespace experimental
```

yhmtsai (approved):

pratikvn (approved):
Nice! LGTM!

PR #1755, opened 2024-12-15
Add cmake-format, and a pre-commit hook for it

pratikvn:
This PR adds a pre-commit hook to format CMakeLists,txt files. The actual format can still be discussed. But currently, I just formatted all CMakeLists.txt files with the default style. If necessary, we can adjust the style in the `.cmake-format.json` file.

Unfortunately, as develop does not have cmake-format hook, the current format job will not run the cmake-format pre-commit hook yet.


.cmake-format.json
```diff
@@ -0,0 +1,311 @@
...
+    "_help_separate_fn_name_with_space": [
+      "If true, separate function names from parentheses with a",
+      "space"
+    ],
+    "separate_fn_name_with_space": false,
+    "_help_dangle_parens": [
+      "If a statement is wrapped to more than one line, than dangle",
+      "the closing parenthesis on its own line."
+    ],
+    "dangle_parens": false,
```
upsj, line 94: This option would make for nicer diffs e.g. when adding files to the end of a list

.cmake-format.json
```diff
@@ -0,0 +1,311 @@
...
+    ],
+    "disable": false,
+    "_help_line_width": [
+      "How wide to allow formatted cmake files"
+    ],
+    "line_width": 80,
+    "_help_tab_size": [
+      "How many spaces to tab for indent"
+    ],
+    "tab_size": 2,
```
upsj, line 46: Maybe keep this in sync with our C++ code?
```suggestion
    "tab_size": 4,
```

.cmake-format.json
```diff
@@ -0,0 +1,311 @@
...
+          "BAR",
+          "BAZ"
+        ],
+        "kwargs": {
+          "HEADERS": "*",
+          "SOURCES": "*",
+          "DEPENDS": "*"
+        }
+      }
+    },
```
upsj, line 19: Here we should list the functions that use CMake argument parsing

upsj (approved):
LGTM! Format suggestions:

cmake/install_helpers.cmake
```diff
@@ -17,102 +16,116 @@ set(GINKGO_INSTALL_MODULE_DIR "${CMAKE_INSTALL_LIBDIR}/cmake/Ginkgo/Modules")
 # 3. GINKGO_INSTALL_RPATH_DEPENDENCIES : Allows adding any extra paths to the
 #    RPATH.
 #
-# @param name  the name of the target
-# @param ARGN  any external dependencies path to be added
+# @param name  the name of the target @param ARGN  any external dependencies
```
yhmtsai, line 19: wrong format

core/CMakeLists.txt
```diff
@@ -13,120 +13,119 @@ set(config_source
 if(GINKGO_BUILD_MPI)
     list(APPEND config_source config/schwarz_config.cpp)
 endif()
-# MSVC: LNK1189 issue
-# CLANG in MSYS2 (MINGW): too many exported symbols
-# We separate the library as a workaround to solve this issue
-# To make ginkgo still be the major library, we make the original to ginkgo_core in MSVC/shared
-# TODO: should think another way to solve it like dllexport or def file
+# MSVC: LNK1189 issue CLANG in MSYS2 (MINGW): too many exported symbols We
```
yhmtsai, line 16: wrong format

yhmtsai (changes requested):

yhmtsai:
alternative one is https://github.com/BlankSpruce/gersemi

pratikvn:
Gersemi is used now, as it also seems to provide a decent setup and has pre-commit support, in addition to being active. The comments needed to be guarded with `#gersemi: on/off`, but otherwise, the default formatting seems to work well enough.

upsj:
Is there a way to prevent that from happening? Comments are a pretty important and frequent part of the code, and other formatters manage to not mess them up that badly 😁

pratikvn:
Not all comments, but only the ones where we try to use multi-line doxygen style documentation comments.

pratikvn:
Turns out it is doing the right thing anyway, so disabling it with `off/on` is not necessary.

third_party/identify_stream_usage/CMakeLists.txt
```diff
@@ -6,42 +6,73 @@ if(GINKGO_BUILD_CUDA)
...
-    add_test(NAME default_stream_identification_cuda COMMAND test_stream_identification_cuda)
-    
-    set_tests_properties(default_stream_identification_cuda PROPERTIES
-                         ENVIRONMENT LD_PRELOAD=$<TARGET_FILE:identify_stream_usage_cuda>)
+            CXX_STANDARD
+            17
+            CXX_STANDARD_REQUIRED
+            ON
+            POSITION_INDEPENDENT_CODE
+            ON
```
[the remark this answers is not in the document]
pratikvn: Yes, I moved it above now.

cmake/CTestScript.cmake
```diff
@@ -147,49 +156,71 @@ if(NOT PROC_COUNT EQUAL 0)
...
+elseif(
+    (
+        NOT CTEST_MEMORYCHECK_TYPE STREQUAL "NONE"
+        AND NOT CTEST_MEMORYCHECK_TYPE STREQUAL "Valgrind"
+    )
+    OR CTEST_BUILD_CONFIGURATION STREQUAL "COVERAGE"
+)
+    set(GINKGO_CONFIGURE_OPTIONS
+        "-DGINKGO_DEVEL_TOOLS=OFF;-DGINKGO_BUILD_REFERENCE=ON;-DGINKGO_BUILD_OMP=ON;-DGINKGO_BUILD_CUDA=OFF;-DGINKGO_BUILD_HIP=OFF;-DGINKGO_BUILD_SYCL=OFF;-DCMAKE_BUILD_TYPE=${CTEST_BUILD_CONFIGURATION}"
+    )
```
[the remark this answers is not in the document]
pratikvn: I think it does not break single argument into multiple lines. It was the same before as well, so I think it should be okay.

yhmtsai (approved):

PR #1756, opened 2024-12-18
only use mkl dense gemm when all value pointers are valid

yhmtsai:
We notice an page fault error from accessing nullptr in mkl dense gemm.
The issue happens when we pass empty matrix (nullptr) and only with single right hand side, but the routine seems to still access the pointer.
It only happens with rolling version of intel runtime library, which starts throwing error on page fault I guess, not long-term-support channel with icpx/oneMKL 2023.1.0 and 2025.0.4.
We simply avoid this issue by only calling mkl gemm when all data pointer are vaild.


MarcelKoch (approved):
I'm fine with it as it is, but maybe we could use `a->get_size() && b->get_size() && c->get_size()` instead of the pointers?

upsj:
That may not be sufficient, as 0xn operators also have a `nullptr`. Does the issue also happen with rectangular empty matrices?

PR #1771, opened 2025-01-22
Add sized ctor and public add_operator to Combination

greole:
This PR makes the `Combination::add_operators` member public so that operators can be added after constructing a combination matrix. Additionally, a new constructor is added so that an empty combination can be created first and additional operators are added  later.


upsj:
Can you describe a use case for this? Combination creation is pretty lightweight, so I am hesitant to add more complexity there.

greole:
Sure and maybe there is a better way to do it. On the OGL side I have `std::vector<gko::LinOp>` of local and non_local linops representing local and non_local interfaces. Probably I could also use this ctr 
```
    explicit Combination(CoefficientIterator coefficient_begin,
                         CoefficientIterator coefficient_end,
                         OperatorIterator operator_begin,
                         OperatorIterator operator_end)
```
but there are cases (after repartitioning) where `operator_begin==operator_end` thus it would throw.

upsj:
Making the empty case work seems sensible, except for the fact that we don't have anything to base the size on then. BTW ctr is kind of a misleading abbreviation, it could be counter or constructor.

include/ginkgo/core/base/combination.hpp
```diff
@@ -112,6 +112,7 @@ class Combination : public EnableLinOp<Combination<ValueType>>,
         add_operators(std::forward<Rest>(rest)...);
     }
 
+protected:
```
MarcelKoch, line 115: I agree with @upsj that we should not expose this member. I think it would be enough to add a constructor, which takes the iterators, and just an dimension as an extra parameter. Then you could build the linop from that parameter, and just check that the operator dimensions match up.

MarcelKoch (changes requested):

PR #1773, opened 2025-01-29
Add primitive for warp load balancing

upsj:
This adds a primitive that allows the distribution of variable-sized chunks of work across a warp for better memory coalescing and warp utilization. This can be used as a component in operations like SpGEMM, symbolic Cholesky or, combined with a segmented reduction, could also be used to implement the inner loop in an SpMV similar to MergePath. Looking for feedback and performance benchmark suggestions :)


upsj:
If the performance is good, this could also be used as an internal component for merge path SpMV

PR #1780, opened 2025-02-14
Add specific communicator for neighborhood communication (Again)

MarcelKoch:
This is just #1588, which was marked as merged, because I reordered some commits, and GH interpreted this as the `neighborhood-communicator` branch being merged.
For comments and details, see the previous PR.


core/distributed/collective_communicator.cpp
```diff
@@ -0,0 +1,37 @@
+// SPDX-FileCopyrightText: 2024 - 2025 The Ginkgo authors
```
pratikvn, line 1: I think the range of years is somehow incorrect ? I think it should 2017 - 2025 ? Or is it different for each file ?

core/distributed/neighborhood_communicator.cpp
```diff
@@ -0,0 +1,273 @@
+// SPDX-FileCopyrightText: 2017 - 2025 The Ginkgo authors
```
pratikvn, line 1: Here it is 2017 - 2025.

pratikvn (changes requested):

yhmtsai (approved):
need to apply formatting on ginkgo.hpp again due to the order of pre-commit order.

core/distributed/collective_communicator.cpp
MarcelKoch: IMO it's not important that every file has the copyright 2017-2025. This file didn't exist 2017, so putting a copyright for that year into it seems a bit pointless. Also this is just what the reuse tool does, so we should stick to it.

pratikvn: Yes, I think that is fine, but `neighborhood_communicator.cpp` also didnt exist in 2017. So, I would prefer that we be uniform.

pratikvn (approved):

core/test/mpi/distributed/collective_communicator.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+
+TYPED_TEST(CollectiveCommunicator, CanCopyConstruct)
+{
+    auto spcomm = this->create_default_comm();
+
+    auto copy(spcomm);
+
+    ASSERT_EQ(copy, spcomm);
+    ;
```
yhmtsai, line 148:
```suggestion
```

core/test/mpi/distributed/collective_communicator.cpp
```diff
@@ -0,0 +1,264 @@
...
+TYPED_TEST(CollectiveCommunicator, CanCopyAssign)
+{
+    using communicator_type = typename TestFixture::communicator_type;
+    auto spcomm = this->create_default_comm();
+    communicator_type copy{this->comm};
+
+    copy = spcomm;
+
+    ASSERT_EQ(copy, spcomm);
+    ;
```
yhmtsai, line 161:
```suggestion
```

PR #1782, opened 2025-02-16
Generate `ginkgo.hpp` via CMake

### LINKED ISSUE #1781: Create `ginkgo.hpp` dynamically
We have a script that generates the `ginkgo.hpp` header, and commit it any time we add a new public header file. Instead we could use the script (or a similar tool) to generate the file at compile time, eliminating the need to update the file in git.

upsj:
Otherwise we always need to commit the header when adding new files. Additionally, it was being picked up by pre-commit, and the output of the script isn't properly formatted, so everything broke when adding a new header in #802 

Closes #1781


pratikvn (approved):

.pre-commit-config.yaml
```diff
@@ -30,10 +30,3 @@ repos:
         examples/external-lib-interfacing/external-lib-interfacing.cpp|
         core/base/workspace_aliases.hpp
       )$
-  - id: update-ginkgo-header
```
yhmtsai, line 33: if we only aim for avoid pre-commit issue, we can move this to early stage such that they are checked after generation

yhmtsai (changes requested):
I like you did it in cmake!
I think we should add ginkgo.hpp into .gitignore in our repo.
However, it raises two concerns which I need to confirm
1. when user pulls the change, will their old ginkgo.hpp they be stayed?
2. If the file stays, will our include_header logics use source one or build one first?

Also, add a fatal error to detect whether user has the file in source

upsj:
If users pull this change, their ginkgo.hpp will be deleted, so there is no danger of conflicting files (unless they changed ginkgo.hpp, then it's a merge conflict at rebase).

.pre-commit-config.yaml
upsj: To extend: I think as a general rule, different pre-commit steps shouldn't undo each other's changes.

yhmtsai: no, it will not. unless pre-commit -> build -> pre-commit -> build.
It will not undo anything if ginkgo.hpp.in is updated there too.

yhmtsai (approved):

.pre-commit-config.yaml
upsj: I just checked, it's even worse than that: Regardless of order, if one hook undoes the change by another hook, both of them are marked as failed and prevent the changes from being committed
```
update ginkgo header.....................................................Failed
- hook id: update-ginkgo-header
- files were modified by this hook
reuse-annotate...........................................................Failed
- hook id: reuse-annotate
- files were modified by this hook

Successfully changed header of include/ginkgo/ginkgo.hpp
Successfully changed header of include/ginkgo/core/test.hpp
```

PR #1783, opened 2025-02-16
Fall back to ILU/IC sparselib algorithm with CPU executors

upsj:
There is no reason to make the distinction between algorithms when we only have a single one available. This issue was found when running the custom multigrid example with OpenMP

### DIFF
--- a/core/factorization/ic.cpp
+++ b/core/factorization/ic.cpp
@@ -103,7 +103,8 @@ std::unique_ptr<Composition<ValueType>> Ic<ValueType, IndexType>::generate(
 
     std::shared_ptr<const matrix_type> ic;
     // Compute LC factorization
-    if (parameters_.algorithm == incomplete_algorithm::syncfree) {
+    if (parameters_.algorithm == incomplete_algorithm::syncfree ||
+        exec == exec->get_master()) {
         std::unique_ptr<gko::factorization::elimination_forest<IndexType>>
             forest;
         const auto nnz = local_system_matrix->get_num_stored_elements();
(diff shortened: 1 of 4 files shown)


pratikvn (changes requested):
The OMP tests need to be updated. Otherwise, LGTM

yhmtsai:
I think we have decided to stay the behavior which throw the error when OMP does not have syncfree algorithm. It is also based on we keep sparselib algorithm although it is not from sparse library in fact.
Refer to the discussion https://github.com/ginkgo-project/ginkgo/pull/1684#discussion_r1849229785

upsj:
The current default does not support OpenMP, so either we change the default, or we fix the breakage. With the CSR strategies, we fall back to a different algorithm if it's unsupported - why the difference here? Both compute the same thing, and have only runtime performance implications.

yhmtsai:
I had the same comments on that but it was still the decision we made.

nbeams:
I actually think it would be better if `syncfree` were the default, because it supports OpenMP + CUDA/HIP, vs. `sparselib` which only supports CUDA/HIP. But that changes the default behavior for anyone who was already using CUDA/HIP, perhaps in a "silent" way that could be easy to miss, which I think was the main reason not to do it. The current setup has the same default behavior as pre-#1684 for all three affected backends (use cuSPARSE/hipSPARSE or throw a not implemented error for OpenMP), except with the new error message for OpenMP, letting the user know that another option is available now.

MarcelKoch (approved):
IMO the approach is reasonable. Another alternative which doesn't change the current default, would be to introduce a new enum value, probably `default`, which defers to `sparselib` on gpu and `syncfree` on CPU. But maybe that is a bit overkill.

PR #1792, opened 2025-02-18
Check config key

yhmtsai:
This PR adds the check to verify the keys in config are allowed by predefined keys set.
It also avoids typo from user such that the configuring solver is not what users expect without notice.

It is WIP. before finalizing it, some thoughts will be nice.
`allowed_keys` preparation: predefine it as early as possible or insert the key into the set close to the condition
I am predefining it as early as possible. but I think the second way might makes sense when the class has long list of keys like multigrid. It makes the manual type together and reduce the chance that missing deleting one of condition and allowed key. The first one is somehow easy to set with constructor not many lines of `set.insert("key")`


MarcelKoch:
Suggestion: use the pnode to keep track of which keys of a map were accessed. If not all keys were accessed, then there are some invalid keys. 
The pnode will also keep track of what is the set of allowed keys, since the parser function will test for all valid keys.

upsj:
Implementation example:
Replace
```cpp
if (auto& obj = config.get("max_iters"))
```
by
```cpp
std::set<std::string> allowed_keys;
auto get_config_value = [](const char* key) {
    return config.get(key);
    allowed_keys.insert(key);
};
// ...
if (auto& obj = get_config_value("max_iters"))
```
The main goal is to specify every key name string only once, so we avoid accidental typos.

yhmtsai:
It merge the allowed keys preparation with trying getting nodes now

core/config/config_helper.cpp
```diff
@@ -113,6 +113,18 @@ parse_minimal_criteria(const pnode& config, const registry& context,
 
     type_descriptor updated_td = update_type(config, td);
 
+    // Although it can still be caught in the following map, it gives consistent
+    // error exception when some keys are not allowed.
+    std::set<std::string> allowed_keys{"time",
```
upsj: There is potential for typos here - can this be initialized from the map, or collect the string literals in a single place, either a constexpr variable or maybe a macro?

core/config/solver_config.hpp
```diff
@@ -17,19 +19,20 @@ namespace config {
...
 template <typename SolverParam>
-inline void common_solver_parse(SolverParam& params, const pnode& config,
-                                const registry& context,
-                                type_descriptor td_for_child)
+void common_solver_parse(SolverParam& params, const pnode& config,
+                         const registry& context, type_descriptor td_for_child,
+                         std::set<std::string>& allowed_keys)
 {
-    if (auto& obj = config.get("generated_preconditioner")) {
+    if (auto& obj =
```
upsj: We could simplify the diff and improve readability by wrapping the get_config_node call in a lambda that inserts the config and allowed_keys parameters. This goes for almost every change in this PR.

upsj (changes requested):
Looks mostly good, but I would like to eliminate some potential for bugs, and simplify the diffs

core/config/solver_config.hpp
yhmtsai: I think it will not reduce the diff.
It will introduce the same lambda to all parse function by copy-paste, which sounds easy to be inconsistent in the future.

core/config/config_helper.cpp
yhmtsai: maybe not the macro. I did it before, but it was not easy to read from macro.
I can remove it because it can be caught by normal map exception. we can think how to deal with it later or just catch the exception and throw more easier understanding error.

core/config/solver_config.hpp
upsj: It will simplify the individual config checks, since they don't need additional parameters to the get function call, so the changes are smaller in each line, only one bigger addition at the beginning of each function. We could potentially also put it into a macro, if you are worried about inconsistencies.

yhmtsai: To me, they sounds similar betweem changing to the lambda function name or function with allowed_keys
I do not think adding allowed_keys additionally will make the check complicated.
I think another possible approach just coming to my mind is to create a config_structure binding from config.
It provides `get` again with adding string to set internally and provide the check function.
(although check destructor is more automatic to me, usually destructor should avoid throwing the exception if I remember correctly.)

upsj: The difference is most visible for long names like `symbolic_factorization`, where the change forces the if condition into two lines, while previously we only had a single line.

I do like the config_structure idea, I would probably call it a view or a decorator from the class design perspective, but essentially they do the same thing.

Destructors avoiding throwing exceptions only matters if the exception is thrown when unrolling another exception, and that is not a place to be calling Gingko code anyways.

yhmtsai: two lines to me is not a big issue to prevent them.
Yes, I mean three of them are the same thing, which just give different way to pass allowed_keys.
Although I use macro a lot, I would also like to avoid the macro usage if there is an alternative one without introducing too many stacks.

core/config/config_helper.hpp
```diff
@@ -315,6 +316,32 @@ inline std::shared_ptr<typename Csr::strategy_type> get_strategy(
 }
 
 
+/**
+ * This is a decroator on top of config node. When we try to get a node by a
+ * key, it registers the keys as the allowed_keys. When this object is
+ * destroyed, it will check whether the config node only contains the key in
+ * allowed_keys. If some key is not in allowed_keys, it will throw an exception.
+ */
+class config_decorator {
```
upsj: I'm trying to look for a better name for this class - while it is a decorator, it should probably have a more specific name, maybe `config_check_decorator`?

core/config/solver_config.hpp
```diff
@@ -17,19 +19,18 @@ namespace config {
 
 
 template <typename SolverParam>
-inline void common_solver_parse(SolverParam& params, const pnode& config,
-                                const registry& context,
-                                type_descriptor td_for_child)
+void common_solver_parse(SolverParam& params, config_decorator& decorator,
```
upsj: I'm wondering if we should just call this parameter `config` itself, since it acts like the original config parameter, plus a little bit extra

core/distributed/preconditioner/schwarz.cpp
```diff
@@ -47,30 +48,30 @@ Schwarz<ValueType, LocalIndexType, GlobalIndexType>::parse(
     const config::type_descriptor& td_for_child)
 {
     auto params = Schwarz::build();
-
-    if (auto& obj = config.get("generated_local_solver")) {
+    config::config_decorator decorator(config);
```
upsj: decorator is not a particularly descriptive name, what if we named it `config` and renamed the parameter to something like `config_node`?

upsj (approved):
LGTM! I only would like to find some better names.

core/distributed/preconditioner/schwarz.cpp
yhmtsai: we already used config for the object to hold property tree, so I would like to have the name easy to distinguish the behavior. 
`config_checker checker` is okay to me or `config_tracker tracker`?

upsj: typename `config_check_decorator` and variable name `config_check`?

yhmtsai: looks good to me

core/config/trisolver_config.hpp
```diff
@@ -1,10 +1,12 @@
-// SPDX-FileCopyrightText: 2017 - 2024 The Ginkgo authors
+// SPDX-FileCopyrightText: 2017 - 2025 The Ginkgo authors
 //
 // SPDX-License-Identifier: BSD-3-Clause
 
 #ifndef GKO_CORE_CONFIG_TRISOLVER_CONFIG_HPP_
 #define GKO_CORE_CONFIG_TRISOLVER_CONFIG_HPP_
 
+#include <set>
```
upsj: nit: I think the set include is no longer used?

yhmtsai: good catch!

PR #1794, opened 2025-02-19
Add fallback to lower-memory cuSparse SpGEMM algorithm

izilberter:
Address an issue where the default CuSPARSE SpGEMM algorithm estimates an overly large memory buffer for matrices greater than ~4 million rows, causing a memory allocation exception regardless of the actual GPU memory capacity. Since CUDA 12.0, alternate, less memory-intensive algorithms for SpGEMM have been introduced to fix the issue.

The spgemm and advanced_spgemm cuda routines now attempt to compute the matrix product using the default CUSPARSE_SPGEMM_ALG1 algorithm, and if it fails, fall back to CUSPARSE_SPGEMM_ALG2. The CuSparse bindings for spgemm-related functions are updated to take the algorithm as an argument.

A simple test .cpp file
[spgemm.txt](https://github.com/user-attachments/files/18858877/spgemm.txt)
 is attached to show the creation of a Poisson operator by multiplying two matrices. The sample output pre-patch:

> ./spgemm cuda 4194305
> terminate called after throwing an instance of 'gko::CusparseError'
>  what():  ginkgo/cuda/base/cusparse_bindings.hpp:228: spgemm_compute: Unknown error

Post patch:
> ./spgemm cuda 4194305
> Matrix multiply time(s): 0.0102431

### DIFF
--- a/common/cuda_hip/matrix/csr_kernels.template.cpp
+++ b/common/cuda_hip/matrix/csr_kernels.template.cpp
@@ -2453,7 +2453,7 @@ void spgemm(std::shared_ptr<const DefaultExecutor> exec,
     } else {
         GKO_NOT_IMPLEMENTED;
     }
-#else   // GKO_COMPILING_CUDA
+#else  // GKO_COMPILING_CUDA
     auto a_vals = a->get_const_values();
     auto a_row_ptrs = a->get_const_row_ptrs();
     auto a_col_idxs = a->get_const_col_idxs();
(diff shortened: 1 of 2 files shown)


common/cuda_hip/matrix/csr_kernels.template.cpp
```diff
@@ -2488,26 +2488,63 @@ void spgemm(std::shared_ptr<const DefaultExecutor> exec,
...
+                                  c_descr, spgemm_descr, spgemm_alg,
+                                  buffer1.get_data(), buffer2_size, nullptr);
+        // compute spgemm
+        buffer2.resize_and_reset(buffer2_size);
+        sparselib::spgemm_compute(
+            handle, &alpha, a_descr, b_descr, &beta, c_descr, spgemm_descr,
+            spgemm_alg, buffer1.get_data(), buffer2_size, buffer2.get_data());
+    }
+
+    catch (const CusparseError& cse) {
```
upsj, line 2518: IMO we should be more specific here based on the error code: Only for `CUSPARSE_STATUS_INSUFFICIENT_RESOURCES` should we attempt to use a different algorithm

upsj (changes requested):
Thanks for your contribution! Can you also add yourself to the CONTRIBUTORS list (sorted alphabetically IIRC)? One change request from my side

common/cuda_hip/matrix/csr_kernels.template.cpp
izilberter: Good idea, I've updated accordingly. There was a minor bug where the error was being registered as "Unknown Error", which was fixed by including cuda.h in cuda/base/exception.cpp to set the CUDA_VERSION macro.

upsj: Good catch! Instead of doing a string comparison, IMO we should rather expose the actual error code integer in CusparseError (also `strstr` is a C-style function that should be avoided in C++ code)

upsj:
We are using a semi-linear merge workflow, so instead of merging the changes from `develop` into this branch, please rebase onto `develop` so your branch is just a linear sequence of commits on top of `develop`.
Did you encounter an easy way to trigger this error, so we can test the fall-back execution path?

izilberter:
> We are using a semi-linear merge workflow, so instead of merging the changes from `develop` into this branch, please rebase onto `develop` so your branch is just a linear sequence of commits on top of `develop`. Did you encounter an easy way to trigger this error, so we can test the fall-back execution path?

Got it, thanks.

The test in the file attached with the pull request (spgemm.txt) is the most reliable way I've found, and basically where the problem came up in the wild, so to speak: Multiply 2csr matrices where the result would contain >~ 13M nonzeros. In the attached example, this is triggered with ./spgemm cuda 4200000. There are further examples of failure cases in this thread:
https://github.com/NVIDIA/CUDALibrarySamples/issues/38

common/cuda_hip/matrix/csr_kernels.template.cpp
izilberter: Should be cleaned up now.

upsj:
Your example has a slight bug BTW, which causes the cuSPARSE transpose routine to do something slightly weird - you claim to have `2 * discretization_points` non-zeros, but only fill `2 * discretization_points - 1` of them, the resulting transposed matrix has an additional entry (1, 4200000) that doesn't belong.

izilberter:
> Your example has a slight bug BTW, which causes the cuSPARSE transpose routine to do something slightly weird - you claim to have `2 * discretization_points` non-zeros, but only fill `2 * discretization_points - 1` of them, the resulting transposed matrix has an additional entry (1, 4200000) that doesn't belong.

Wow, that is a good catch, and doesn't pop up in the reference executor. With the correct nnz specified, the fallback is never triggered and the multiply works until my card runs out of memory (at 30MM rows or so).

I had to rig up something closer to my actual application to reproduce the error; here I fill a discrete gradient matrix on an `NX*NX` grid, then compute `grad * grad^T` (NB: Not `grad^T * grad`, which would produce a standard Laplacian):

I get the following output:

> ./spgemm cuda 2000
> Matrix multiply time(s): 0.054936

> ./spgemm cuda 2500
> ginkgo/cuda/base/cusparse_bindings.hpp:247: spgemm_compute: CUSPARSE_STATUS_INSUFFICIENT_RESOURCES, 
> Falling back to Alg2
> Matrix multiply time(s): 0.129386

[spgemm.txt](https://github.com/user-attachments/files/18916158/spgemm.txt)

MicroZHY:
Hello, your contribution is excellent. I used the patch you provided. When using the ISAI preconditioner, it involves estimating the sparse pattern, which calls spgemm. For my matrix with nrows = 125025, ncols = 125025, and nnz = 13324468, the original algorithm throws an error. After applying your patch and using CUSPARSE_SPGEMM_ALG2, it still throws an error. Is it possible that the intermediate storage is still too high?
Error output:  nrows = 125025, ncols = 125025, nnz = 13324468
I am in CUSPARSE_SPGEMM_ALG2
terminate called after throwing an instance of 'gko::CusparseError'

izilberter:
> Hello, your contribution is excellent.
> 
> I used the patch you provided. When using the ISAI preconditioner, it involves estimating the sparse pattern, which calls spgemm.
> 
> For my matrix with nrows = 125025, ncols = 125025, and nnz = 13324468, the original algorithm throws an error. After applying your patch and using CUSPARSE_SPGEMM_ALG2, it still throws an error.
> 
> Is it possible that the intermediate storage is still too high?
> 
> Error output:
> 
> ```
> I am in CUSPARSE_SPGEMM_ALG2
> terminate called after throwing an instance of 'gko::CusparseError'
>   what():  /home/zhanghy/software/ginkgo-develop/cuda/base/cusparse_bindings.hpp:231: spgemm_estimate_memory: Unknown error
> ```
> 
> When using alg3, it also throws an error. The output is:
> 
> ```
> I am in CUSPARSE_SPGEMM_ALG3
> terminate called after throwing an instance of 'gko::AllocationError'
>   what():  /home/zhanghy/software/ginkgo-develop/cuda/base/memory.cpp:58: cuda: failed to allocate memory block of 91587203607B
> ```

Hello,

At that matrix density and nnz you are almost certainly running out of GPU memory. I would try ALG3 again, but change the chunk fraction in the spgemm_estimate_memory call in csr_kernels.template.cpp:

```
            sparselib::spgemm_estimate_memory(
                handle, &alpha, a_descr, b_descr, &beta, c_descr, spgemm_descr,
                spgemm_alg, 1.0f, buffer3_size, nullptr, nullptr);
            array<char> buffer3{exec, buffer3_size};
            sparselib::spgemm_estimate_memory(
                handle, &alpha, a_descr, b_descr, &beta, c_descr, spgemm_descr,
                spgemm_alg, 1.0f, buffer3_size, buffer3.get_data(),
                &buffer2_size);
```

The 1.0f can be replaced with a different number from (0,1] representing the fraction of work handled in each buffer. It looks like the total buffer estimate is 91GB, so depending on your card this should probably be 0.05-0.1.
Let me know how this works out. I am also starting to run into similar issues for very large problems and was going to revisit this patch.

PR #1799, opened 2025-02-28
`MPI_Comm` cannot expected to be a pointer

pratikvn:
Fixes an issue where it was expected that MPI_Comm is comparable to a pointer. But some implementations (like MPICH) just use a reference to an `int` and hence are not comparable to `std::nullptr_t`

### DIFF
--- a/include/ginkgo/core/base/mpi.hpp
+++ b/include/ginkgo/core/base/mpi.hpp
@@ -567,9 +567,6 @@ public:
      */
     bool is_identical(const communicator& rhs) const
     {
-        if (!get() || !rhs.get()) {
-            return get() == rhs.get();
-        }
         if (get() == MPI_COMM_NULL || rhs.get() == MPI_COMM_NULL) {
             return get() == rhs.get();
         }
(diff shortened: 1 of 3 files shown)


pratikvn:
I am actually not sure why the first `if` check is needed (in `is_identical` and in `is_congruent`.  Maybe @MarcelKoch can elaborate on this ?

include/ginkgo/core/base/mpi.hpp
```diff
@@ -567,7 +567,7 @@ class communicator {
      */
     bool is_identical(const communicator& rhs) const
     {
-        if (get() == nullptr || rhs.get() == nullptr) {
+        if (!get() || !rhs.get()) {
```
MarcelKoch, line 570: TBH, maybe the first check is not necessary at all? I think it was in there to check if the `comm_` is a nullptr, but I think this can't be the case anymore.

yhmtsai:
what is the empty state of MPICH? 0 here?

pratikvn:
@yhmtsai , yes, empty state is NULL, or 0.

yhmtsai (approved):
some typos are caught by spell-check. could you also change those?

MarcelKoch (approved):

PR #1801, opened 2025-03-05
Add pnode equality check (`operator ==` and `operator !=`)

yhmtsai:
This PR adds the pnode equality check. The equality of pnode only check the equality of current representing data.

### DIFF
--- a/core/test/config/property_tree.cpp
+++ b/core/test/config/property_tree.cpp
@@ -207,6 +207,12 @@ protected:
             // array3
             vector.emplace_back(
                 pnode(pnode::array_type{pnode{"3"}, pnode{4}, pnode{true}}));
+            // array4 with map
+            vector.emplace_back(pnode(pnode::array_type{
+                pnode(pnode::map_type{{"first", pnode{"1"}}})}));
+            // array5 with array
+            vector.emplace_back(
+                pnode(pnode::array_type{pnode(pnode::array_type{pnode{"1"}})}));
             // map
             vector.emplace_back(pnode(pnode::map_type{{"first", pnode{"1"}},
                                                       {"second", pnode{1.2}}}));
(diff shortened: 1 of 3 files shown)


core/test/config/property_tree.cpp
```diff
@@ -181,3 +181,134 @@ TEST(PropertyTree, UseInCondition)
...
+            // array
+            vector.emplace_back(
+                pnode(pnode::array_type{pnode{"1"}, pnode{4}, pnode{true}}));
+            // array2
+            vector.emplace_back(
+                pnode(pnode::array_type{pnode{"2"}, pnode{4}, pnode{true}}));
+            // array3
+            vector.emplace_back(
+                pnode(pnode::array_type{pnode{"3"}, pnode{4}, pnode{true}}));
+            // map
```
MarcelKoch, line 210: maybe have one of these maps also store a map (and maybe an array?). I think this should make the test `CheckEqualityUnderlyingNodeDiff` unnecessary.

core/test/config/property_tree.cpp
```diff
@@ -181,3 +181,134 @@ TEST(PropertyTree, UseInCondition)
...
+    pnode map_2{pnode::map_type{{"first", first.at(5)},
+                                {"second", first.at(7)},
+                                {"third", diff_content.at(8)},
+                                {"forth", first.at(9)}}};
+
+    ASSERT_NE(map_1, map_2);
+}
+
+
+TEST_F(PropertyTreeEquality, CheckEqualityOfDiffOrderMap)
```
MarcelKoch, line 300: I think checking that differently ordered arrays are not the same would also make sense.

core/test/config/property_tree.cpp
MarcelKoch, line 1: I think this is missing a simple tests that nodes with different tags are different.

MarcelKoch (changes requested):
Implementation looks good, just some comments on the tests.

core/test/config/property_tree.cpp
MarcelKoch: I see, that actually makes sense.

upsj (approved):
LGTM! I agree with Marcel's suggestion for a map-inside-map test.

core/test/config/property_tree.cpp
yhmtsai: Yes, will it be a bit confusing or hidden?
When I add these different map and different array, I feel CheckEquality else-branch for different tag and different content

MarcelKoch (approved):

PR #1809, opened 2025-03-17
Support for Kokkos Threads

greole:
Currently, only `Kokkos::omp` is supported. This PR enables `Kokkos::Threads` as possible memory space and maps to Ginkgos openmp executor. 

I added separate `#ifdef `checks even though using `#if defined(KOKKOS_ENABLE_OMP) || defined(KOKKOS_ENABLE_THREADS) ` would be more concise, but it allows to treat the Kokkos::Threads case differently if required.


greole:
format!

MarcelKoch (changes requested):
I'm a bit hesitant, because we can't assume that the `OmpExecutor` can be used. If it is really necessary that the mapping `create_executor(Kokkos::Threads)` exists, then it should return a `ReferenceExecutor`. This won't blow up at least.

PR #1811, opened 2025-03-19
Allow to use ValueType in template of preconditioner IC

yhmtsai:
This PR gives different way than adding a new type (https://github.com/ginkgo-project/ginkgo/pull/1029), which allows ValueType can be used in template LSolverType of preconditioner::Ic (similar to Ilu later).
Both PRs aims to reduce the challenges of instantiating many types for preconditioner::Ic and preconditioner::Ilu because we need to instantiate them first such that users can select one of them. 

when the SolverType is LinOp,
- no default solver and factorization factory in general

TODO:
- [x] reduce the instantiation type for config
- [ ] Do we still accept the same configuration previously? via the LinOp interface but create the default solver when they use l_solver_type as certain type?
- [x] ~~apply to ILU (will be next pr)~~ #1828


include/ginkgo/core/base/composition.hpp
```diff
@@ -16,6 +16,21 @@
 namespace gko {
 
 
+/**
+ * CompositeionBase provides the interface
```
upsj, line 20:
```suggestion
 * CompositionBase provides the interface
```

include/ginkgo/core/base/composition.hpp
```diff
@@ -16,6 +16,21 @@
...
+ */
+class CompositionBase {
```
upsj, line 22: This is only necessary because we don't have a value type for Ic and Ilu, right? What if we allowed the first template parameter to be _either_ ValueType _or_ SolverType? Then the value type would always be available. Slightly more complexity in extracting the actual types to use, but no need to add a new class whose only reason for existence is only necessary temporarily.

upsj (commented):
Suggestion for how we could make the changes less intrusive, and more consistent with other LinOpFactory types

include/ginkgo/core/base/composition.hpp
yhmtsai: It is a good idea! I might need to adapt CompositionBase or different way for the mixed-IC but it should doable.
the value_type seems to be more clear way.

include/ginkgo/core/preconditioner/ic.hpp
```diff
@@ -394,19 +481,28 @@ class Ic : public EnableLinOp<Ic<LSolverType, IndexType>>, public Transposable {
...
-            l_solver_ = generate_default_solver<l_solver_type>(exec, l_factor);
-            // If comp contains both factors: use the transposed factor to avoid
-            // transposing twice
-            if (comp->get_operators().size() == 2) {
-                auto lh_factor = comp->get_operators()[1];
-                GKO_ASSERT_EQUAL_DIMENSIONS(l_factor, lh_factor);
-                lh_solver_ = as<lh_solver_type>(l_solver_->conj_transpose());
+            if constexpr (std::is_same_v<l_solver_type, LinOp>) {
+                // linop can not have a default factory generation
+                GKO_NOT_IMPLEMENTED;
```
upsj, line 486: GKO_NOT_SUPPORTED

include/ginkgo/core/preconditioner/ic.hpp
```diff
@@ -368,19 +451,23 @@ class Ic : public EnableLinOp<Ic<LSolverType, IndexType>>, public Transposable {
...
+                            .on(exec);
+                }
+                auto fact = std::shared_ptr<const LinOp>(
+                    parameters_.factorization_factory->generate(lin_op));
+                // ensure that the result is a composition
+                comp = std::dynamic_pointer_cast<const Composition<value_type>>(
+                    fact);
+                if (!comp) {
+                    GKO_NOT_SUPPORTED(comp);
+                }
```
upsj, line 470: could this be `gko::as`?

include/ginkgo/core/preconditioner/ic.hpp
```diff
@@ -327,30 +414,26 @@ class Ic : public EnableLinOp<Ic<LSolverType, IndexType>>, public Transposable {
...
-            b, x);
+        this->set_cache_to(b);
+        if (l_solver_->apply_uses_initial_guess()) {
+            cache_.intermediate->copy_from(b);
+        }
+        l_solver_->apply(b, cache_.intermediate);
+        if (lh_solver_->apply_uses_initial_guess()) {
+            x->copy_from(cache_.intermediate);
+        }
+        lh_solver_->apply(cache_.intermediate, x);
```
upsj, line 425: This changes the mixed-precision behavior slightly, meaning we require more allocations, and may lose some precision. Is that intended?

upsj (dismissed):
Looks good for the most part, a handful of questions and suggestions, and one concern about mixed precision. Maybe change the title to remove ILU?

MarcelKoch (changes requested):
The change to accepting `ValueType` as first template parameter seems not to be done fully.

include/ginkgo/core/preconditioner/ic.hpp
```diff
@@ -368,19 +451,23 @@ class Ic : public EnableLinOp<Ic<LSolverType, IndexType>>, public Transposable {
...
-            }
-            auto fact = std::shared_ptr<const LinOp>(
-                parameters_.factorization_factory->generate(lin_op));
-            // ensure that the result is a composition
-            comp =
-                std::dynamic_pointer_cast<const Composition<value_type>>(fact);
-            if (!comp) {
-                GKO_NOT_SUPPORTED(comp);
+            if constexpr (std::is_same_v<value_type, void>) {
+                GKO_NOT_IMPLEMENTED;
```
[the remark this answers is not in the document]
yhmtsai: GKO_NOT_SUPPORTED require a object, so I currently put the GKO_NOT_IMPLEMENTED

include/ginkgo/core/preconditioner/ic.hpp
yhmtsai: I miss to delete the `copy_from` in the `set_cache_to`.
I think the current one will be less than the original one.
assume the LSolver have the same precision as IC, so cache vector is not needed to copy for precision

original one does not support mixed precision.
original: (no matter the underlying support mixed or not)
input (copy in), output (copy in)
Lsolver (always copy in set_cache_to without considering initial_guess), LHsolver no precision change
output (copy out)

now:
LSolver (copy input (in))
LHSolver (copy output (inout))
if the underlying solver support mixed precision, it will reduce the copy.

When we consider more different settings: the LSolver and LHSolver do not use the same precision 
original will need to pay Lsolver(input, cache) LHSolver(cache, output) additionally because it dispatches the precision first.
but the current one: only need to pay cache conversion in LSolver and LHsolver additionally, which is less than the original one.

This is one of reason for additional stuff may need for mixed-precision (internally not just outside)

include/ginkgo/core/preconditioner/ic.hpp
```diff
@@ -35,17 +35,93 @@ namespace preconditioner {
 namespace detail {
 
 
+// helper for handle the transposed type of concrete type and LinOp
 template <typename Type>
+struct transposed_type_impl {
```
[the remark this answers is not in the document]
yhmtsai: move to utils_helper.hpp

include/ginkgo/core/preconditioner/ic.hpp
yhmtsai: ah, do you only mean the `x->copy_from(cache)`?
If so, you are right, it should use the user input directly.

include/ginkgo/core/base/utils_helper.hpp
```diff
@@ -152,6 +152,79 @@ template <typename Pointer>
...
+struct factory_type_impl<LinOp> {
+    using type = LinOpFactory;
+};
+
+
+template <typename Type>
+using factory_type = typename factory_type_impl<Type>::type;
+
+template <typename Type>
+constexpr bool is_ginkgo_linop = std::is_convertible_v<Type*, LinOp*>;
```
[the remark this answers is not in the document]
yhmtsai: I thought `is_convertible_v` more appropriate check?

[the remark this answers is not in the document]
MarcelKoch: At least not to me. Here we want to know if `LinOp` is a base class of `Type`, which the first suggestion does very explicitly. For the other one you have to remember that this is equivalent to LinOp being a base class.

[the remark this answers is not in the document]
yhmtsai: I see your point from readability. 
If we do not face the some situation (like private or ambiguous base) mentioned by std, we can go for that

include/ginkgo/core/preconditioner/ic.hpp
yhmtsai: It is more than what I thought because of real apply on the complex vector. I will move this out of the current pr. the apply and set_cache_to should be the same function as the original one

include/ginkgo/core/preconditioner/ic.hpp
```diff
@@ -35,17 +32,14 @@ namespace preconditioner {
...
 
 
-template <typename Type>
+template <typename SolverTypeOrValueType>
 constexpr bool support_ic_parse =
-    is_instantiation_of<Type, solver::LowerTrs>::value ||
-    is_instantiation_of<Type, solver::Ir>::value ||
-    is_instantiation_of<Type, solver::Gmres>::value ||
-    is_instantiation_of<Type, preconditioner::LowerIsai>::value;
+    std::is_same_v<gko::detail::get_solver_type<SolverTypeOrValueType>, LinOp>;
```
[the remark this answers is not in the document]
yhmtsai: I change it a bit to only require the class not the first template, which will simplify the process for Ilu

include/ginkgo/core/preconditioner/ic.hpp
```diff
@@ -142,14 +141,16 @@ class Ic : public EnableLinOp<Ic<LSolverType, IndexType>>, public Transposable {
...
+                const gko::detail::factory_type<l_solver_type>>
                 solver)
         {
             return with_l_solver(std::move(solver));
         }
 
         parameters_type& with_l_solver(
-            deferred_factory_parameter<const typename l_solver_type::Factory>
+            deferred_factory_parameter<
+                const gko::detail::factory_type<l_solver_type>>
```
MarcelKoch, line 153: I think this function needs some documentation to clarify what the allowed inputs are.

MarcelKoch (changes requested):
Only two minor comments left.

MarcelKoch (approved):

PR #1816, opened 2025-03-26
Add Coo Transpose

yhmtsai:
This PR reuses the `device_matrix_data` `sort_row_major` by turning it to accept the pointer argument and add Transpose function to `Coo` under `Transposable` class.
I would also like to use this sort_row_major in pgm if the changes are acceptable. Otherwise, I will reuse the pgm::device_matrix_data in coo.

TODO:
- [x] reuse the sort_row_major in PGM or reuse it from PGM


core/matrix/coo.cpp
```diff
@@ -378,6 +380,27 @@ void Coo<ValueType, IndexType>::write(mat_data& data) const
...
+std::unique_ptr<LinOp> Coo<ValueType, IndexType>::transpose() const
+{
+    auto coo = this->clone();
+    std::swap(coo->row_idxs_, coo->col_idxs_);
+    auto size = this->get_size();
+    coo->set_size(dim<2>{size[1], size[0]});
+    coo->get_executor()->run(coo::make_sort_row_major(
+        coo->get_num_stored_elements(), coo->get_row_idxs(),
+        coo->get_col_idxs(), coo->get_values()));
+    return coo;
```
upsj, line 393: from the parallel algorithm perspective, an in-place sort is much harder than an out-of-place sort. I would suggest changing the interface to have separate inputs and outputs (which is also what the GPU version needs to do internally), then this wouldn't need any additional allocations, see also potential improvements stemming from #1821

upsj (approved):
LGTM!

core/matrix/coo.cpp
upsj: Thrust uses cub internally, and their RadixSort implementation allows fine control over the allocations if we wanted to optimize it. But for now, not a blocking suggestion from me.

PR #1817, opened 2025-03-28
Avoid CI run when only changes files in doc/

yhmtsai:
After #1813 , I believe we will have more PR just improving documentation.
All documentation build are handled in readthedoc, so we do not need to occupy our CI for that.
Some are still working like pre-commit check, spell check, pr label, etc without the compilation need.
Others like gitlab, osx, intel, abidiff which require compile ginkgo will not be executed when only changes the file under doc/
If some changes in other place or it is our main/develop branch, the CI still performs as usual.
gitlab ci currently does not support the ignore pattern unfortunately (related issue: https://gitlab.com/gitlab-org/gitlab/-/issues/198688)
We need to use the opposite way to check whether others file changed or not. It definitely goes crazy if we have more conditions. (see the gitlab changes)
Github action has path_ignore which makes thing clear.

**I also want to avoid mirror when it is not main/develop and no changes out of doc in github action rather than doing that in gitlab side. any comments on it?** 
It is done.
When use branch and path condition, both are required to be satisfied for running a job.
Also Github action does not allow two push event in one file, so I split it into two: one for the major branches, which always mirror, and the other for the rest, which only mirror when there are more than documentation changes


.github/workflows/bot-pr-updated.yml
```diff
@@ -63,3 +28,13 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ secrets.BOT_TOKEN }}
       run: .github/check-wiki-changelog.sh
+  
+  spell-check:
+    name: Spell Check with Typos
```
pratikvn, line 33: I think this would be needed when something in `doc/` changes ?

pratikvn (approved):
LGTM! I am fine with mirroring only when it is main/develop. I am also fine with doing that from the github side.

.github/workflows/bot-pr-updated.yml
yhmtsai: Yes, I forgot to move the ignore out of this file

pratikvn (approved):
LGTM!

PR #1818, opened 2025-03-31
Add a job to sync github repo with Codehub 

pratikvn:
This PR adds a github sync workflow to sync the github repo to codehub when the main/develop/release branches have been updated for easy CI/CD within the [CASTIEL-2 project](https://eurohpc-ju.europa.eu/research-innovation/our-projects/castiel-2_en). 

The objective is that eventually Ginkgo will be automatically deployed on the EU supercomputers and available to users through Easybuild and [EESSI](https://www.eessi.io/)

### DIFF
--- a/.github/workflows/github-to-codehub-push-sync.yml
+++ b/.github/workflows/github-to-codehub-push-sync.yml
@@ -1,0 +1,24 @@
+name: Codehub-sync
+
+on:
+  push:
+    branches:
+      - 'main'
+      - 'develop'
+      - 'releases/**'
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+    name: CASTIEL2 Codehub Sync
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+    - uses: cniethammer/git-repo-sync@cb1067d118c0c209bcded585b5c78e581bb0dfec
+      with:
+        target-url: 'https://codehub.hlrs.de/coes/microcard2/ginkgo.git'
+        target-username: ${{ secrets.CODEHUB_TOKEN_NAME }}
+        target-token: ${{ secrets.CODEHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


yhmtsai:
we only sync our code to their side. Do we need to provide any compilation step here?

.github/workflows/github-to-codehub-push-sync.yml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+    name: CASTIEL2 Codehub Sync
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+    - uses: cniethammer/git-repo-sync@cb1067d118c0c209bcded585b5c78e581bb0dfec
```
yhmtsai, line 20: does it require specific hash rather than a version number?

pratikvn:
@yhmtsai , currently, I only sync. The compilation is done with easybuild, so that will be handled externally. In a fashion similar to spack.

.github/workflows/github-to-codehub-push-sync.yml
pratikvn: I think the repo does not really have a version number. This is what they have in their docs as well, so I would keep that for now. We can update it later to a version number if necessary.

yhmtsai:
Does anyone else have access to the documentation? maybe @fritzgoebel ?
does the sync only syncs the branch which triggers the event or the entire repo?

pratikvn:
I think syncing the whole repo should not be an issue. We might want to trigger specific branches and/tags, and I believe there should be no specific downside to syncing all the branches and tags.

yhmtsai:
No, I do not mean avoiding sync the entire repo. 
more on checking which one is what you would like to do because the pr description sounds more on only push certain branch not entire one.

yhmtsai (approved):
I do not know the usage of arguments from git-repo-sync from CASTIEL2
Others look good to me

pratikvn:
I have updated the description.

PR #1819, opened 2025-04-02
Enable mixed precision dispatch in distributed matrix with ScalarCache, GenericDenseCache, and GenericVectorCache

yhmtsai:
This PR enables the mixed precision dispatch in distributed matrix.
Moreover, it adds the `ScalarCache` to handle the scalar with user-specified value but with different type (mainly for one scalar) and the `GenericDenseCache` and `GenericVectorCache` to reuse the workspace for Dense view with different value type for the communication buffer. 
Both them will prepare the data during the get not initialized.

Although it is based on the distributed rowGatherer now, it is not required.

### DIFF
--- a/benchmark/test/reference/spmv_distributed.profile.stdout
+++ b/benchmark/test/reference/spmv_distributed.profile.stdout
@@ -5,7 +5,7 @@
         "comm_pattern": "stencil",
         "spmv": {
             "csr-csr": {
-                "storage": 11476,
+                "storage": 11452,
                 "time": 1.0,
                 "repetitions": 1,
                 "completed": true
--- a/core/distributed/matrix.cpp
+++ b/core/distributed/matrix.cpp
@@ -4,6 +4,8 @@
 
 #include "ginkgo/core/distributed/matrix.hpp"
 
+#include <utility>
+
 #include <ginkgo/core/base/array.hpp>
 #include <ginkgo/core/base/precision_dispatch.hpp>
 #include <ginkgo/core/distributed/assembly.hpp>
@@ -416,21 +418,28 @@ void Matrix<ValueType, LocalIndexType, GlobalIndexType>::read_distributed(
 }
 
 
-template <typename LocalIndexType>
-void init_recv_buffers(std::shared_ptr<const Executor> exec,
-                       const RowGatherer<LocalIndexType>* row_gatherer,
-                       size_type num_cols,
-                       const detail::GenericVectorCache& buffer,
-                       const detail::GenericVectorCache& host_buffer)
+template <typename ValueType, typename LocalIndexType>
+std::pair<std::shared_ptr<Vector<ValueType>>,
+          std::shared_ptr<Vector<ValueType>>>
+init_recv_buffers(std::shared_ptr<const Executor> exec,
+                  const RowGatherer<LocalIndexType>* row_gatherer,
+                  size_type num_cols, const detail::GenericVectorCache& buffer,
+                  const detail::GenericVectorCache& host_buffer)
 {
+    auto comm =
+        row_gatherer->get_collective_communicator()->get_base_communicator();
     auto global_recv_dim =
         dim<2>{static_cast<size_type>(row_gatherer->get_size()[0]), num_cols};
     auto local_recv_dim = dim<2>{
         static_cast<size_type>(
             row_gatherer->get_collective_communicator()->get_recv_size()),
         num_cols};
-    buffer.init(exec, global_recv_dim, local_recv_dim);
-    host_buffer.init(exec->get_master(), global_recv_dim, local_recv_dim);
+
+    auto vector = buffer.template get<ValueType>(exec, comm, global_recv_dim,
+                                                 local_recv_dim);
+    auto host_vector = host_buffer.template get<ValueType>(
+        exec->get_master(), comm, global_recv_dim, local_recv_dim);
+    return std::make_pair(vector, host_vector);
 }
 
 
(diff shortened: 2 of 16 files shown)


include/ginkgo/core/base/dense_cache.hpp
```diff
@@ -90,6 +100,68 @@ struct DenseCache {
...
+ */
+struct ScalarCache {
+    ScalarCache(std::shared_ptr<const Executor> executor, double scalar_value);
+    ~ScalarCache() = default;
+    ScalarCache(const ScalarCache& other);
+    ScalarCache(ScalarCache&& other) noexcept;
+    ScalarCache& operator=(const ScalarCache& other);
+    ScalarCache& operator=(ScalarCache&& other) noexcept;
+    std::shared_ptr<const Executor> exec;
+    double value;
```
MarcelKoch, line 151: can you make this private? Changing it will not invalidate the stored dense vectors, so perhaps some safety would be nice.

include/ginkgo/core/base/dense_cache.hpp
```diff
@@ -90,6 +100,68 @@ struct DenseCache {
...
+ *            not be copied when the outer object gets copied.
+ */
+struct GenericDenseCache {
+    GenericDenseCache() = default;
+    ~GenericDenseCache() = default;
+    GenericDenseCache(const GenericDenseCache&);
+    GenericDenseCache(GenericDenseCache&&) noexcept;
+    GenericDenseCache& operator=(const GenericDenseCache&);
+    GenericDenseCache& operator=(GenericDenseCache&&) noexcept;
+    mutable array<char> workspace;
```
MarcelKoch, line 120: same here.

core/distributed/matrix.cpp
```diff
@@ -408,20 +406,24 @@ void Matrix<ValueType, LocalIndexType, GlobalIndexType>::apply_impl(
                            row_gatherer_->get_collective_communicator()
                                ->get_recv_size()),
                        dense_b->get_size()[1]};
-            recv_buffer_.init(exec, recv_dim);
-            host_recv_buffer_.init(exec->get_master(), recv_dim);
+            auto recv_buffer =
+                recv_buffer_.template get<b_value_type>(exec, recv_dim);
```
MarcelKoch, line 410: shouldn't this be
```suggestion
                recv_buffer_.template get<x_value_type>(exec, recv_dim);
```
I think I would rather have the communication use different value types than the second advanced apply.

core/test/base/dense_cache.cpp
```diff
@@ -194,4 +191,300 @@ TYPED_TEST(DenseCache, VectorIsNotMoveAssigned)
...
+    // Different executor always regenerate different workspace
+    ASSERT_NE(array_ptr, this->cache.workspace.get_const_data());
+}
+
+
+TYPED_TEST(GenericDenseCache, WorkspaceIsNotCopied)
+{
+    using value_type = typename TestFixture::value_type;
+    auto buffer = this->cache.template get<value_type>(this->ref, this->size);
+    gko::detail::GenericDenseCache cache(this->cache);
```
MarcelKoch, line 325: nit: (also for the other tests)
```suggestion

    gko::detail::GenericDenseCache cache(this->cache);
```

MarcelKoch (changes requested):
LGMT, just some minor remarks.

include/ginkgo/core/base/dense_cache.hpp
yhmtsai: I make them private now. I think there is no easy way to change the value now. Do you have something in mind for the safety check?

core/distributed/matrix.cpp
yhmtsai: I think the distributed RowGatherer does not support mixed precision.
It takes the value type from b and create a temporary clone on x if the type is not matched.
Also, the matrix can have the mixed precision.

MarcelKoch: Yes, I looked a bit into it, and this sort of mixed precision is not allowed by the MPI standard. So my original comment was pointless.

MarcelKoch:
@yhmtsai can you rebase again?

include/ginkgo/core/base/dense_cache.hpp
MarcelKoch: Thanks, that is already enough.

yhmtsai:
@MarcelKoch sure, it is done

MarcelKoch (approved):

yhmtsai:
@MarcelKoch Because buffer is handled by distributed::Vector now, I add GenericVectorCache for that.
The last commit is to combine the init and get<type> and I return the vector directly from `init_recv_buffers` now.
Before last commit, it follows the same pattern as the current develop, initializing the configuration in `init_recv_buffers` and get the vector later. I think it is a bit weird because initializing in GenericVectorCache does nothing and it splits the configuration to two places.

core/test/mpi/distributed/vector_cache.cpp
```diff
@@ -321,3 +322,262 @@ TYPED_TEST(VectorCache, VectorIsNotMoveAssigned)
...
+          rank(comm.rank()),
+          num(comm.size()),
+          default_local_size(rank + 1, 3),
+          default_global_size((num + 1) * num / 2, 3)
+    {}
+
+    gko::experimental::mpi::communicator comm;
+    std::shared_ptr<gko::ReferenceExecutor> ref;
+    int rank;
+    int num;
```
MarcelKoch, line 348: nit:
```suggestion
    int num_ranks;
```

include/ginkgo/core/base/dense_cache.hpp
```diff
@@ -1,17 +1,27 @@
...
 
-#include <ginkgo/core/base/executor.hpp>
+#include <ginkgo/core/base/array.hpp>
+#include <ginkgo/core/base/dim.hpp>
 
 
 namespace gko {
+
+
+class Executor;
```
MarcelKoch, line 20: Is this necessary? `array.hpp` already includes `executor.hpp`.

include/ginkgo/core/base/precision_dispatch.hpp
MarcelKoch, line 1: These changes look like they are reverting some previous commits. Is that actually wanted?

MarcelKoch (changes requested):
Are the changes to `precision_dispatch.hpp` intentional? 
Otherwise, LGTM.

benchmark/test/reference/spmv_distributed.profile.stdout
```diff
@@ -5,7 +5,7 @@
         "comm_pattern": "stencil",
         "spmv": {
             "csr-csr": {
-                "storage": 11476,
+                "storage": 11452,
```
yhmtsai, line 8: It is from DenseCache -> ScalarCache, we allocate the memory when using it.
It is 3 * sizeof(ValueType) because we use 3 processes for mpi

PR #1820, opened 2025-04-02
Add bitvector data structure

upsj:
Pulled out of the symbolic Cholesky work:
This adds a bitvector data structure that can be used to map from a full index range to a subset of indices in constant time $\mathcal O(1)$ instead of $\mathcal O(\log n)$. It can be used to either represent an increasing sequence of integers, or a boolean predicate on an index range. It can be used to get rid of a lot of binary searches.


core/components/bitvector.hpp
```diff
@@ -0,0 +1,103 @@
...
+    constexpr bool get(index_type i) const
+    {
+        assert(i >= 0);
+        assert(i < size());
+        const auto block = i / block_size;
+        const auto local = i % block_size;
+        return bool((bits_[block] >> local) & 1);
+    }
+
+    constexpr index_type rank(index_type i) const
```
upsj, line 44: we could potentially also add `rank_inclusive` that adds 1 if the bit is set, then we can use that to represent row_ptrs

common/cuda_hip/components/bitvector_kernels.cpp
```diff
@@ -0,0 +1,55 @@
...
+    using bv = device_bitvector<IndexType>;
+    using storage_type = typename bv::storage_type;
+    const auto num_blocks = ceildiv(size, bv::block_size);
+    thrust::fill_n(policy, bits, num_blocks, 0u);
+    thrust::for_each_n(
+        policy, indices, num_indices, [bits] __device__(IndexType idx) {
+            constexpr auto block_size = device_bitvector<IndexType>::block_size;
+            const auto block = idx / block_size;
+            const auto local = idx % block_size;
+            atomicOr(bits + block, storage_type{1} << local);
```
yhmtsai, line 39: We can atomicOr in the shared memory rather than the global memory.
The global write will not need the initialization and the atomic on the global memory

common/cuda_hip/components/bitvector_kernels.hpp
```diff
@@ -0,0 +1,84 @@
...
+    IndexType* __restrict__ popcounts, DevicePredicate predicate)
+{
+    constexpr auto block_size = device_bitvector<IndexType>::block_size;
+    const auto subwarp_id = thread::get_subwarp_id_flat<block_size>();
+    const auto subwarp_base = subwarp_id * block_size;
+    if (subwarp_base >= size) {
+        return;
+    }
+    const auto subwarp =
+        group::tiled_partition<block_size>(group::this_thread_block());
```
yhmtsai, line 42: It will give an issue when the stoarge type is uint64 not 32 in cuda.

yhmtsai (changes requested):

common/cuda_hip/components/bitvector_kernels.hpp
upsj: good point, we should use a static_assert there, but for now it's always uint32

common/unified/components/bitvector_kernels.generic.hpp
```diff
@@ -0,0 +1,65 @@
...
+    const auto num_blocks = static_cast<size_type>(ceildiv(size, block_size));
+    array<uint32> bits{exec, num_blocks};
+    array<IndexType> ranks{exec, num_blocks};
+    run_kernel(
+        exec,
+        [] GKO_KERNEL(auto block_i, auto size, auto device_predicate, auto bits,
+                      auto ranks) {
+            const auto base_i = block_i * block_size;
+            storage_type mask{};
+            if (base_i + block_size <= size) {
```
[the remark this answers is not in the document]
upsj: the last index handled by the loop is `(base_i + block_size - 1)`, so this needs to be `<=`

common/unified/components/bitvector_kernels.generic.hpp
```diff
@@ -0,0 +1,65 @@
...
+                    mask |= bit << local_i;
+                }
+            } else {
+                int local_i = 0;
+                for (int local_i = 0; base_i + local_i < size; local_i++) {
+                    const storage_type bit =
+                        device_predicate(base_i + local_i) ? 1 : 0;
+                    mask |= bit << local_i;
+                }
+            }
```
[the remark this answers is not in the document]
upsj: I wanted the top one to potentially allow for full unrolling, since the loop trip count is known. But I'm not actually sure if the compiler can manage that always.

MarcelKoch (changes requested):
I have a suggestion for the device type: make the data members public. This could allow to implement the current member functions as free-standing functions, which would allow for backend specific implementations. It would also allow to add a `set` function, which will require backend specific atomics.
(This suggestion is not blocking however.)

upsj:
@MarcelKoch the device view is meant as a read-only representation for the bitvector, I don't really want to encourage people building them by hand without the helper algorithms.

dpcpp/components/bitvector.dp.hpp
```diff
@@ -0,0 +1,113 @@
...
+    const auto queue = exec->get_queue();
+    queue->submit([&](sycl::handler& cgh) {
+        cgh.parallel_for(num_blocks, [=](sycl::id<1> block_i) {
+            const auto base_i = static_cast<IndexType>(block_i) * block_size;
+            storage_type mask{};
+            if (base_i + block_size <= size) {
+                for (int local_i = 0; local_i < block_size; local_i++) {
+                    const storage_type bit =
+                        device_predicate(base_i + local_i) ? 1 : 0;
+                    mask |= bit << local_i;
```
MarcelKoch, line 45: maybe extract this as a lambda, since it gets repeated?

omp/components/bitvector.hpp
```diff
@@ -0,0 +1,124 @@
...
+    array<IndexType> rank_array{exec, num_blocks};
+    const auto bits = bit_array.get_data();
+    const auto ranks = rank_array.get_data();
+#pragma omp parallel for
+    for (IndexType block_i = 0; block_i < num_blocks; block_i++) {
+        const auto base_i = block_i * block_size;
+        storage_type mask{};
+        if (base_i + block_size <= size) {
+            for (int local_i = 0; local_i < block_size; local_i++) {
+                const storage_type bit =
```
MarcelKoch, line 39: same, but I guess some compiler will not allow it (maybe nvhpc?)

reference/components/bitvector.hpp
```diff
@@ -0,0 +1,83 @@
...
+    array<index_type> ranks{exec, static_cast<size_type>(num_blocks)};
+    std::fill_n(bits.get_data(), num_blocks, 0);
+    assert(std::is_sorted(it, it + count));
+    for (auto i : irange{count}) {
+        const auto value = it[i];
+        const auto block = value / block_size;
+        const auto local = value % block_size;
+        const auto mask = storage_type{1} << local;
+        assert((bits.get_data()[block] & mask) == 0);
+        bits.get_data()[block] |= mask;
```
MarcelKoch, line 65: I still think there should be a convenience function for all of this.

reference/test/components/bitvector.cpp
```diff
@@ -0,0 +1,103 @@
...
+                                                                   33,   40,
+                                                                   63,   64,
+                                                                   65,   127,
+                                                                   128,  129,
+                                                                   1000, 1024,
+                                                                   2000}
+    {}
+
+    std::vector<index_type> create_random_values(index_type num_values,
+                                                 index_type size)
```
MarcelKoch, line 39: this won't (always) return a vector with `num_values` entries. Maybe call it `max_num_values`.
(If you want to always return `num_values`, then you need to use iota+shuffle+take.)

test/components/bitvector.cpp
```diff
@@ -0,0 +1,208 @@
...
+    using storage_type = typename bitvector::storage_type;
+    constexpr static auto block_size = bitvector::block_size;
+
+    Bitvector()
+        : rng{67193}, sizes{0,    1,    2,    16,    31,    32,  33,
+                            40,   63,   64,   65,    127,   128, 129,
+                            1000, 1024, 2000, 10000, 100000}
+    {}
+
+    gko::array<index_type> create_random_values(index_type num_values,
```
MarcelKoch, line 45: same here.

test/components/bitvector.cpp
```diff
@@ -0,0 +1,208 @@
...
+                                    dvalues.get_size(), size);
+
+            this->assert_bitvector_equal(bv, dbv);
+        }
+    }
+}
+
+
+template <typename IndexType>
+std::pair<gko::bitvector<IndexType>, gko::bitvector<IndexType>> run_predicate(
```
MarcelKoch, line 112: why is this its own function? It's only used once, so I don't see any benefit.

test/components/bitvector.cpp
```diff
@@ -0,0 +1,208 @@
...
+        },
+        dbv.size(), dbv, doutput_bools, doutput_ranks);
+    for (auto i : gko::irange{bv.size()}) {
+        output_bools.get_data()[i] = bv.get(i);
+        output_ranks.get_data()[i] = bv.rank(i);
+    }
+}
+
+
+TYPED_TEST(Bitvector, AccessIsEquivalentToRef)
```
MarcelKoch, line 176: This test has no assertions.

MarcelKoch (changes requested):
For me the biggest issue is the header in `common/unified`. I don't think we should start putting headers in there, especially for kernels that are available for all backends.

common/unified/components/bitvector.hpp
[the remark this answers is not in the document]
upsj: We have the same kind of dispatch for `kernel_launch`, I feel like this would start pulling kernel code into `core`, which we may want to avoid?

test/components/bitvector.cpp
upsj: Same as in the other test, nvcc has issues with device lambdas defined inside template member functions

core/components/bitvector.hpp
```diff
@@ -0,0 +1,160 @@
...
+     *              `popcount(bits[i])`.
+     * @param size  the number of bits we consider part of this bitvector.
+     */
+    constexpr device_bitvector(const storage_type* bits,
+                               const index_type* ranks, index_type size)
+        : size_{size}, bits_{bits}, ranks_{ranks}
+    {}
+
+    /** Returns the number of bits stored in this bitvector. */
+    constexpr index_type size() const { return size_; }
```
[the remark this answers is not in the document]
upsj: This is modeled more after a C++ container than other Ginkgo types. IMO the `get_` here would introduce unnecessary noise and inconsistency, especially since the other functions don't use it.

test/components/bitvector.cpp
MarcelKoch: :'-(

dpcpp/components/bitvector.dp.hpp
upsj: done

reference/test/components/bitvector.cpp
upsj: good suggestion, thanks!

test/components/bitvector.cpp
upsj: done

test/components/bitvector.cpp
upsj: fixed

reference/components/bitvector.hpp
upsj: I added `get_block_and_mask` to `device_bitvector` and used it throughout the code

common/unified/components/bitvector.hpp
MarcelKoch, line 1: This is still a blocker for me. I just don't see that we should put kernel headers (that are not used exclusively in `common/unified`) into `common/unified`.

MarcelKoch (changes requested):

reference/components/bitvector_kernels.cpp
```diff
@@ -0,0 +1,49 @@
...
+{
+    using bv = device_bitvector<IndexType>;
+    using storage_type = typename bv::storage_type;
+    const auto num_blocks = ceildiv(size, bv::block_size);
+    std::fill_n(bits, num_blocks, 0u);
+    for (auto i : irange{num_indices}) {
+        const auto index = indices[i];
+        assert(index >= 0);
+        assert(index < size);
+        bits[index / bv::block_size] |= storage_type{1}
```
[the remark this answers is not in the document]
upsj: added `device_bitvector::get_block_and_mask`

common/unified/components/bitvector.hpp
upsj: With a uniform `transform_iterator` implementation, I can also implement #1832 in `common/unified`. Would that be sufficient justification for putting it there?

common/unified/components/bitvector_kernels.generic.hpp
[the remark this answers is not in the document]
yhmtsai: I do not get it. Isn't the last index of `base_i + block_size <= size` is `base_i+block_size` not `base_i + block_size - 1`?

reference/test/components/bitvector.cpp
```diff
@@ -0,0 +1,101 @@
...
+            ASSERT_EQ(bv.get_num_blocks(), num_blocks);
+            ASSERT_EQ(dbv.num_blocks(), num_blocks);
+            auto it = values.begin();
+            index_type rank{};
+            for (auto i : gko::irange{size}) {
+                const auto block = i / block_size;
+                const auto local = i % block_size;
+                ASSERT_EQ(dbv.rank(i), rank);
+                if (it != values.end() && *it == i) {
+                    ASSERT_TRUE(bool(bv.get_bits()[block] &
```
yhmtsai, line 88:
```suggestion
                    ASSERT_TRUE(static_cast<bool>(bv.get_bits()[block] &
```

test/components/bitvector.cpp
```diff
@@ -0,0 +1,211 @@
...
+        std::sort(values.begin(), values.end());
+        return gko::array<index_type>{this->ref, values.begin(), values.end()};
+    }
+
+    void assert_bitvector_equal(const bitvector& bv, const bitvector& dbv)
+    {
+        ASSERT_EQ(bv.get_size(), dbv.get_size());
+        const auto num_blocks =
+            static_cast<gko::size_type>(bv.get_num_blocks());
+        const auto bits =
```
yhmtsai, line 63: we have also dealed the const_array_view in `GKO_ASSERT_ARRAY_EQ`, so you do not need to cast them back to non-const

omp/components/bitvector.hpp
```diff
@@ -0,0 +1,124 @@
...
+    assert(std::is_sorted(it, it + count));
+#pragma omp parallel num_threads(num_threads)
+    {
+        const auto tid = omp_get_thread_num();
+        const auto begin = std::min<index_type>(tid * work_per_thread, count);
+        const auto end = std::min<index_type>(begin + work_per_thread, count);
+        if (begin < end) {
+            const auto first_block = it[begin] / block_size;
+            const auto last_block = it[end - 1] / block_size;
+            storage_type word{};
```
yhmtsai, line 88:
```suggestion
            storage_type word{0};
```

yhmtsai: the other seem to use mask?

omp/components/bitvector.hpp
```diff
@@ -0,0 +1,124 @@
...
+            auto block = first_block;
+            for (auto i : irange{begin, end}) {
+                const auto value = it[i];
+                const auto new_block = value / block_size;
+                const auto local = value % block_size;
+                if (new_block != block) {
+                    assert(new_block > block);
```
yhmtsai, line 95: I think it assumes no duplicated index to make this assert.

common/cuda_hip/components/bitvector.hpp
```diff
@@ -0,0 +1,181 @@
...
+{
+    using index_type = typename std::iterator_traits<IndexIterator>::value_type;
+    using storage_type = typename device_bitvector<index_type>::storage_type;
+    constexpr auto block_size = device_bitvector<index_type>::block_size;
+    const auto num_blocks = static_cast<size_type>(ceildiv(size, block_size));
+    const auto policy = thrust_policy(exec);
+    array<storage_type> bits_compact{exec, num_blocks};
+    array<index_type> bits_position{exec, num_blocks};
+    array<storage_type> bits{exec, num_blocks};
+    array<index_type> ranks{exec, num_blocks};
```
yhmtsai, line 151: compared to the others, it uses double memory allocation.
does it give better performance?

dpcpp/components/bitvector.dp.hpp
```diff
@@ -0,0 +1,114 @@
...
+    queue->submit([&](sycl::handler& cgh) {
+        cgh.parallel_for(num_blocks, [=](sycl::id<1> block_i) {
+            const auto base_i = static_cast<IndexType>(block_i) * block_size;
+            storage_type mask{};
+            const auto local_op = [&](int local_i) {
+                const storage_type bit =
+                    device_predicate(base_i + local_i) ? 1 : 0;
+                mask |= bit << local_i;
+            };
+            if (base_i + block_size <= size) {
```
yhmtsai, line 46: same comment for the index range

dpcpp/components/bitvector.dp.hpp
```diff
@@ -0,0 +1,114 @@
...
+                for (int local_i = 0; local_i < block_size; local_i++) {
+                    local_op(local_i);
+                }
```
yhmtsai, line 49:
```suggestion
#pragma unroll
                for (int local_i = 0; local_i < block_size; local_i++) {
                    local_op(local_i);
                }
```
to make them indeed different from known size.
the later one can be still merge into this unroll loop by wrapping local_op into condition.
the #check should be the same as the current one.
```
for (...) {
  if (base_i + local_i < size) {
    local_op(local_i);
  }
}
```

core/components/bitvector.hpp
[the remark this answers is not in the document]
yhmtsai: but bitvector uses `get_*` because it is the thing should be used in ginkgo core?

yhmtsai (commented):
some questions
- Do we need to use iterator for `from_sorted_indices`?
- What's the purpose of transformer iterator?
- the index check of (base_i + block_size <= size())

yhmtsai:
By the way, I am not sure how do you plan to use it in symbolic cholesky.
I guess you will have some base as the start point to use smaller size than the #columns of matrix.
Otherwise, the memory will request #columns bits * # rows

upsj:
The most common use case in symbolic elimination tree computation is mapping between a set [0, n) of vertices and a subset (e.g. representatives of connected components, endpoints of tree edges, ...), and speeding up the CSR-to-COO conversion, which ran into some edge cases in the algorithm that caused a 100x-1000x slowdown compared to this improved algorithm.

common/cuda_hip/components/bitvector.hpp
```diff
@@ -0,0 +1,181 @@
...
+struct bitvector_popcnt_functor {
+    using storage_type = typename device_bitvector<IndexType>::storage_type;
+    __device__ IndexType operator()(storage_type mask)
+    {
+        return gko::detail::popcount(mask);
+    }
+};
+
+
+template <typename IndexIterator>
```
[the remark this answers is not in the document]
upsj: yes, for example in #1832

dpcpp/components/bitvector.dp.hpp
```diff
@@ -0,0 +1,114 @@
...
+                           storage_type{});
+    const auto bits = bit_array.get_data();
+    const auto ranks = rank_array.get_data();
+    const auto queue = exec->get_queue();
+    queue->submit([&](sycl::handler& cgh) {
+        cgh.parallel_for(count, [=](sycl::id<1> i) {
+            auto value = it[i];
+            const auto [block, mask] =
+                device_bitvector<index_type>::get_block_and_mask(value);
+            sycl::atomic_ref<storage_type, sycl::memory_order::relaxed,
```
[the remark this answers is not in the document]
upsj: not sure I follow. This would require a segmented scan to get rid of some of the atomic operations

common/unified/components/bitvector_kernels.generic.hpp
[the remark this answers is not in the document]
upsj: When base_i + block_size == size, then the loop below iterates through [0, block_size), just like the other loop would, only this one can be unrolled

common/unified/components/bitvector.hpp
MarcelKoch: My issue is not that it includes non-uniform implementations from different backends, but that I don't really understand how this header is going to be used. If the kernels are only called from kernels implemented in `common/unified`, then it's fine with me. If this file is going to be included anywhere else, then I think it should be put under `core`.
In any case, I hope I explained my reservation enough, and I will not hold up the PR because of it anymore. However, if the header gets included outside `common/unified`, I will bring it up again.

MarcelKoch (approved):

common/cuda_hip/components/bitvector.hpp
upsj: It has a better worst-case behavior, since it doesn't suffer from atomic collisions. If we implemented our own kernel for it, we could also get rid of the second array, but `reduce_by_key` has an interface that prevents us from scattering the results to the output already.

common/unified/components/bitvector.hpp
upsj: For tests and kernels implemented in `unified`, I would use this header, for other tests and kernels I would use the backend-specific one.

omp/components/bitvector.hpp
upsj: No, this assumes sortedness. I assert that there are no duplicates in other places though

dpcpp/components/bitvector.dp.hpp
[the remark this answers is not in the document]
yhmtsai: I think it should almost be like the kernel in cuda_hip.
```
mask = ballot(device_predicate(base_i + local_i)). // subgroup needs to be 32
masks[base_i/bv::blocksize] = mask
```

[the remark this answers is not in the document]
upsj: This is `from_sorted_indices`, not `from_predicate`

common/unified/components/bitvector_kernels.generic.hpp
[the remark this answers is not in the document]
yhmtsai: I see. thanks for the explanation

yhmtsai:
(It is not related to this PR itself but the usage)
If it always takes care of [0, n), we will need $\frac{2^{20} * 2^{20}}{2^3}$ byte -> $2^{37}$ bytes ~ 128 GB for $2^{20} \times 2^{20}$ matrix (~1e6 size).
It takes quite a lot for a small matrix. Do I calculate something wrong?

upsj:
I think you fundamentally misunderstand how this is used - it is an algorithmic component that is usually of size `n` or `nnz`, but would never be used to represent a dense matrix.

yhmtsai:
That's why I ask how do you plan to use it.
Originally, I thought it is to represent the index in a row. no matter how sparse of this row, it always take at least n bits, right?
And then you need to have a bitvector for each row, so you need to n * n bits.

upsj:
Please take a look at https://github.com/ginkgo-project/ginkgo/pull/1832 for an example how the data structure is used. I am not using it in symbolic Cholesky, only for the elimination forest computation and the aforementioned conversion so far. Other applications are operations similar to `copy_if` that need to map between a full index range and a subset repeatedly.
You will just have to take my word on it that this is a useful data structure 🙃

dpcpp/components/bitvector.dp.hpp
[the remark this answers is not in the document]
yhmtsai: I put it in wrong place. I mean for the predicate case.

yhmtsai:
I see. It is more clear now. the index range on the nnz not the columns and symbolic Cholesky.

dpcpp/components/bitvector.dp.hpp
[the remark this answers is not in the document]
upsj: Yes, I realized that after writing my response 😄 We could do that, but I wanted do avoid the complexity of subgroup size dispatch, and since subgroups are sometimes smaller than 32, we would need to assemble the mask out of multiple elements. So that definitely makes sense as a future optimization, but I wanted to keep the complexity in this PR smaller.

upsj:
If I'm filtering nodes or edges, the index space is of size n or nnz, if I'm representing row_ptrs, it is of size nnz + n.

common/cuda_hip/components/bitvector.hpp
```diff
@@ -0,0 +1,181 @@
...
+
+
+template <typename IndexType, typename DevicePredicate>
+__global__ __launch_bounds__(default_block_size) void from_predicate(
+    IndexType size,
+    typename device_bitvector<IndexType>::storage_type* __restrict__ bits,
+    IndexType* __restrict__ popcounts, DevicePredicate predicate)
+{
+    constexpr auto block_size = device_bitvector<IndexType>::block_size;
+    static_assert(block_size <= config::warp_size);
```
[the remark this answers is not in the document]
yhmtsai: actually no need. we have already check it in the get_subwarp_id_flat.

yhmtsai (approved):
three nits left. others looks good to me or not necessary to block this pr.
1. some uint32 should use storage_type from bitvector.
2. add the unroll to the for loop with known fixed size
3. using `static_cast<bool>` rather than `bool()`

core/components/bitvector.hpp
```diff
@@ -0,0 +1,103 @@
...
+        return (size() + block_size - 1) / block_size;
+    }
+
+    constexpr bool get(index_type i) const
+    {
+        assert(i >= 0);
+        assert(i < size());
+        const auto block = i / block_size;
+        const auto local = i % block_size;
+        return bool((bits_[block] >> local) & 1);
```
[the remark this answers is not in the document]
upsj: I think static_cast makes sense to highlight places where casts play an important role, e.g. casting between different integer types. But here to me this is unnecessarily verbose.

core/components/bitvector.hpp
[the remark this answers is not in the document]
MarcelKoch: +1 from me.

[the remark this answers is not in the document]
upsj: done

PR #1821, opened 2025-04-02
Add bucketsort kernels

upsj:
Extracted from the symbolic Cholesky (but might also be useful for other things, e.g. parallel OpenMP COO sorting)

### DIFF
--- a/omp/components/sorting.hpp
+++ b/omp/components/sorting.hpp
@@ -1,0 +1,79 @@
+// SPDX-FileCopyrightText: 2025 The Ginkgo authors
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <numeric>
+
+#include <omp.h>
+
+#include <ginkgo/core/base/array.hpp>
+#include <ginkgo/core/base/types.hpp>
+
+namespace gko {
+namespace kernels {
+namespace omp {
+
+
+template <int num_buckets, typename InputIterator, typename OutputIterator,
+          typename BucketIndexOp>
+std::array<int64, num_buckets + 1> bucket_sort(InputIterator begin,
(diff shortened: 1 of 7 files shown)


omp/components/sorting.hpp
```diff
@@ -0,0 +1,79 @@
...
+    const auto size = end - begin;
+    const auto num_threads = omp_get_max_threads();
+    const auto tmp_size = num_threads * num_buckets;
+    if (tmp.get_size() < tmp_size) {
+        tmp.resize_and_reset(tmp_size);
+    }
+    const auto sums = tmp.get_data();
+    std::fill_n(sums, tmp_size, 0);
+    std::array<int64, num_buckets + 1> global_offsets{};
+#pragma omp parallel
```
upsj, line 36:
```suggestion
#pragma omp parallel num_threads(num_threads)
```

omp/components/sorting.hpp
```diff
@@ -0,0 +1,79 @@
...
+            std::array<int64, num_buckets> offsets{};
+            for (int tid = 0; tid < num_threads; tid++) {
+                for (int i = 0; i < num_buckets; i++) {
+                    const auto value = sums[tid * num_buckets + i];
+                    sums[tid * num_buckets + i] = offsets[i];
+                    offsets[i] += value;
+                }
+            }
+            std::copy_n(offsets.begin(), num_buckets, global_offsets.begin());
+            std::exclusive_scan(global_offsets.begin(), global_offsets.end(),
```
upsj, line 60: This seems to be unsupported in one of the libstdc++ versions we are using (ROCm 4.5 job), but I think I would rather require a fully C++17 compliant compiler than work around this.

PR #1824, opened 2025-04-09
Add PIPECG reference implementation

gojakuch:
This PR adds the reference implementation for the pipelined CG method.


gojakuch:
I've put the "Ready for review" label, but I'm not sure if I should've put the "Requires feedback" one instead, since this is my first PR here and it's hard for me to assess the state of it. any feedback would be considered of course :)

core/solver/pipe_cg.cpp
```diff
@@ -0,0 +1,301 @@
...
+    this->get_system_matrix()->apply(z, w);
+    // m = preconditioner * w
+    this->get_preconditioner()->apply(w, m);
+    // n = A * m
+    this->get_system_matrix()->apply(m, n);
+    // rho = dot(r, z)
+    r->compute_conj_dot(z, rho, reduction_tmp);
+    // delta = dot(w, z)
+    w->compute_conj_dot(z, delta, reduction_tmp);
+    // can these two dot products above be merged in some way?
```
yhmtsai, line 141:
```suggestion
    // TODO: can these two dot products above be merged in some way?
```
the current support should allow [r w] and [z z], so you need to duplicate the z unfortunately.
It might be possible if we introduce conjugate transposed apply on dense it will be like [r w]' * z

pratikvn: Yes, that would be the plan, but we will do that in a later PR.

pratikvn (changes requested):
Nice work! Mostly looks good to me. Only the documentation needs to be updated.

core/solver/pipe_cg.cpp
gojakuch: @pratikvn so should I remove this comment for now? I've put it there so that we don't forget about it

yhmtsai: I would say we can keep that for reminder

reference/solver/pipe_cg_kernels.cpp
```diff
@@ -0,0 +1,164 @@
...
+    // x = x + tmp * p
+    // r = r - tmp * q
+    // z = z - tmp * f
+    // w = w - tmp * g
+    for (size_type i = 0; i < p->get_size()[0]; ++i) {
+        for (size_type j = 0; j < p->get_size()[1]; ++j) {
+            if (stop_status->get_const_data()[j].has_stopped()) {
+                continue;
+            }
+            if (is_nonzero(beta->at(j))) {
```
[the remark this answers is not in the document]
gojakuch: yes, good point, that's a problem I haven't noticed. I just did what's done in the CG in this case, which wouldn't work out for the PIPECG logic. If beta becomes zero somewhere throughout the iteration, for PIPECG, it is gonna stay 0 forever and no other values are going to be changed. I think one way to address this would be to check if beta is zeroed out by step_2 and prevent this from happening there, possibly by just letting `beta = delta`, which is a valid value, as it's what we initialise beta with and what we set beta to in step_1 in its corner case.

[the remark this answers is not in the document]
pratikvn: Yes, I think that is a reasonable choice.

include/ginkgo/core/solver/pipe_cg.hpp
```diff
@@ -0,0 +1,184 @@
...
+
+
+/**
+ * PIPE_CG or the pipelined conjugate gradient method is an iterative type
+ * Krylov subspace method which is suitable for symmetric positive definite
+ * methods. It improves upon the CG method by allowing computation of inner
+ * products and norms to be overlapped with operator and preconditioner
+ * application. The pipelined method scales up to the 10^6 nodes of the assumed
+ * exascale machine, while its standard counterpart level off about one order of
+ * magnitude earlier, as suggested in the referenced paper (see below).
```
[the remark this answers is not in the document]
gojakuch: added

pratikvn (changes requested):
Great! Thanks for making the changes. One last thing before we merge it is that you add yourself to the contributors list here after Marcel Koch: https://github.com/ginkgo-project/ginkgo/blob/develop/contributors.txt#L20 with the following info. `LastName FirstName <email> Affiliation`

pratikvn (approved):
LGTM! Thank you for your contribution!

core/test/solver/pipe_cg.cpp
```diff
@@ -0,0 +1,272 @@
...
+#include <ginkgo/core/matrix/dense.hpp>
+#include <ginkgo/core/solver/pipe_cg.hpp>
+#include <ginkgo/core/stop/combined.hpp>
+#include <ginkgo/core/stop/iteration.hpp>
+#include <ginkgo/core/stop/residual_norm.hpp>
+
+#include "core/test/utils.hpp"
+
+
+namespace {
```
yhmtsai, line 19:
```suggestion
```
we do not need the anonymous namespace for test now

core/test/solver/pipe_cg.cpp
```diff
@@ -0,0 +1,272 @@
...
+                       .with_criteria(stop_factory)
+                       .with_preconditioner(precond_factory)
+                       .on(this->exec);
+
+    ASSERT_EQ(factory->get_parameters().criteria.front(), stop_factory);
+    ASSERT_EQ(factory->get_parameters().preconditioner, precond_factory);
+}
+
+
+}  // namespace
```
yhmtsai, line 272: same here

reference/test/solver/pipe_cg_kernels.cpp
```diff
@@ -0,0 +1,694 @@
...
+    this->small_stop.get_data()[1].reset();
+
+    gko::kernels::reference::pipe_cg::step_2(
+        this->exec, this->small_beta.get(), this->small_p.get(),
+        this->small_q.get(), this->small_f.get(), this->small_g.get(),
+        this->small_z.get(), this->small_w.get(), this->small_m.get(),
+        this->small_n.get(), this->small_prev_rho.get(), this->small_rho.get(),
+        this->small_delta.get(), &this->small_stop);
+
+    GKO_ASSERT_MTX_NEAR(this->small_beta, this->small_delta, 0);
```
yhmtsai, line 333: also need to test the other vector

reference/test/solver/pipe_cg_kernels.cpp
```diff
@@ -0,0 +1,694 @@
...
+    this->small_p->fill(4);
+    this->small_q->fill(3);
+    this->small_f->fill(2);
+    this->small_g->fill(1);
+
+    this->small_rho->at(0) = 2;
+    this->small_rho->at(1) = 3;
+    this->small_beta->at(0) = 8;
+    this->small_beta->at(1) = 3;
+    this->small_stop.get_data()[1] = this->stopped;
```
yhmtsai, line 186: follow AAA rules, but you can use comment to distinguish the group

reference/test/solver/pipe_cg_kernels.cpp
```diff
@@ -0,0 +1,694 @@
...
+          pipe_cg_factory(
+              Solver::build()
+                  .with_criteria(
+                      gko::stop::Iteration::build().with_max_iters(400u),
+                      gko::stop::Time::build().with_time_limit(
+                          std::chrono::seconds(6)),
+                      gko::stop::ResidualNorm<value_type>::build()
+                          .with_reduction_factor(r<value_type>::value))
+                  .on(exec)),
+          mtx_big(gko::initialize<Mtx>(
```
yhmtsai, line 45: the actual initiation follows the member declaration.
here you need to make it to follow the same order of member.

reference/test/solver/pipe_cg_kernels.cpp
yhmtsai: same for the rest of test in this file

reference/test/solver/pipe_cg_kernels.cpp
```diff
@@ -0,0 +1,694 @@
...
+    SKIP_IF_HALF(value_type);
+    auto solver = this->pipe_cg_factory_big->generate(this->mtx_big);
+    auto b = gko::initialize<Mtx>(
+        {1300083.0, 1018120.5, 906410.0, -42679.5, 846779.5, 1176858.5},
+        this->exec);
+    auto x = gko::initialize<Mtx>({0.0, 0.0, 0.0, 0.0, 0.0, 0.0}, this->exec);
+
+    solver->apply(b, x);
+
+    GKO_ASSERT_MTX_NEAR(x, l({81.0, 55.0, 45.0, 5.0, 85.0, -10.0}),
```
yhmtsai, line 543: it is 50 times the CG test

reference/test/solver/pipe_cg_kernels.cpp
```diff
@@ -0,0 +1,694 @@
...
+    auto solver = this->pipe_cg_factory_big->generate(this->mtx_big);
+    auto b = gko::initialize<Mtx>(
+        {886630.5, -172578.0, 684522.0, -65310.5, 455487.5, 607436.0},
+        this->exec);
+    auto x = gko::initialize<Mtx>({0.0, 0.0, 0.0, 0.0, 0.0, 0.0}, this->exec);
+
+    solver->apply(b, x);
+
+    GKO_ASSERT_MTX_NEAR(x, l({33.0, -56.0, 81.0, -30.0, 21.0, 40.0}),
+                        r<value_type>::value * 2 * 1e5);
```
yhmtsai, line 563: it is 2000 times of CG test. it sounds like we need to check something again.

reference/test/solver/pipe_cg_kernels.cpp
```diff
@@ -0,0 +1,694 @@
...
+    auto x = gko::initialize<Mtx>({0.0, 0.0, 0.0, 0.0, 0.0, 0.0}, this->exec);
+
+    solver->conj_transpose()->apply(b, x);
+
+    GKO_ASSERT_MTX_NEAR(x, l({81.0, 55.0, 45.0, 5.0, 85.0, -10.0}),
+                        r<value_type>::value * 5 * 1e4);
+}
+
+
+}  // namespace
```
yhmtsai, line 694: same here

core/solver/pipe_cg.cpp
```diff
@@ -0,0 +1,301 @@
...
+}  // namespace pipe_cg
+
+
+template <typename ValueType>
+typename PipeCg<ValueType>::parameters_type PipeCg<ValueType>::parse(
+    const config::pnode& config, const config::registry& context,
+    const config::type_descriptor& td_for_child)
+{
+    auto params = solver::PipeCg<ValueType>::build();
+    common_solver_parse(params, config, context, td_for_child);
```
yhmtsai, line 43: need to add corresponding test in config core/test/config/solver.cpp and add it into the main parse
core/config/solver_config.cpp and core/config/registry.cpp

core/solver/pipe_cg.cpp
```diff
@@ -0,0 +1,301 @@
...
+    GKO_SOLVER_STOP_REDUCTION_ARRAYS();
+
+    // r = b
+    // prev_rho = 1.0
+    exec->run(pipe_cg::make_initialize_1(gko::detail::get_local(dense_b),
+                                         gko::detail::get_local(r), prev_rho,
+                                         &stop_status));
+    // r = r - Ax
+    this->get_system_matrix()->apply(neg_one_op, dense_x, one_op, r);
+    // z = preconditioner * r
```
yhmtsai, line 129: if there is no specific reason, I will suggest the variable follows the paper notation.

core/solver/pipe_cg.cpp
```diff
@@ -0,0 +1,301 @@
...
+        this->get_system_matrix()->apply(m, n);
+        // prev_rho = rho
+        swap(prev_rho, rho);
+        // TODO: merge these two dot products:
+        // rho = dot(r, z)
+        r->compute_conj_dot(z, rho, reduction_tmp);
+        // delta = dot(w, z)
+        w->compute_conj_dot(z, delta, reduction_tmp);
+        // check
+        ++iter;
```
yhmtsai, line 186: I think the criterion check should be earlier?

yhmtsai (changes requested):
need to add small section into test/test_install/test_install.cpp

reference/test/solver/pipe_cg_kernels.cpp
gojakuch: this is a weird thing that I don't get. the numbers that my PIPECG implementation outputs here are the same as the tested values, and yet it fails this check. I explained this to Pratik and he fixed it by adjusting these tolerances. I've got no other idea what that could be. the output numbers are literally the same here. any suggestions?

core/solver/pipe_cg.cpp
gojakuch: if you're talking about variable naming, Ginkgo I chose names compatible with the CG implementation in Ginkgo. the naming in the paper contradicts the naming in CG in Ginkgo

core/solver/pipe_cg.cpp
gojakuch: alright, I'll add these tests

reference/test/solver/pipe_cg_kernels.cpp
yhmtsai: do you mean x is [33.0, -56.0, 81.0, -30.0, 21.0, 40.0]' on output?

pratikvn:
It seems nvhpc compilers still have some issues with the pipecg tolerances, so it might need to be increased even more: https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/9693465904

reference/test/solver/pipe_cg_kernels.cpp
gojakuch: yeah. or do I not get what these tolerances are? :sweat_smile:

core/solver/pipe_cg.cpp
gojakuch: so are you suggesting that I put the check before computing rho in the loop?

gojakuch:
> It seems nvhpc compilers still have some issues with the pipecg tolerances, so it might need to be increased even more: https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/9693465904

I think @yhmtsai suggests that we find another solution

reference/test/solver/pipe_cg_kernels.cpp
gojakuch: yeah nevermind, I misinterpreted some things in my test example, it's out of the tolerance bounds

pratikvn:
The tolerance problem is not limited to PipeCG, but is prevalent across **all** solvers in Ginkgo. We have arbitrarily increased the tolerances for different solvers to ensure that they pass the tests. The matrices we use to test the solvers are also for the most part arbitrary (we have some control: hpd-ness, sparsity, but not over condition number for example.) 

A proper solution IMO would be to have full control over the matrix and vector inputs: 
1. Generate matrices with specific condition numbers and eigenvalue distributions.
2. Define vectors with controlled vector norms.

Then we can correlate solvers with inputs and tailor the inputs rather than tailor the output tolerances. This approach requires a more sophisticated sparse matrix generation, which we currently do not have.

Therefore, IMO, the current approach should be okay, and we should revisit the controlled inputs approach at a later point (with a new PR) and update all the solvers and their tests.

gojakuch:
@yhmtsai please consider @pratikvn's comment when reviewing. I rebased this and tried to address all the other things you've mentioned except for the tolerances

reference/test/solver/pipe_cg_kernels.cpp
```diff
@@ -0,0 +1,678 @@
...
+    SKIP_IF_HALF(value_type);
+    auto solver = this->pipe_cg_factory_big->generate(this->mtx_big);
+    auto b = gko::initialize<Mtx>(
+        {886630.5, -172578.0, 684522.0, -65310.5, 455487.5, 607436.0},
+        this->exec);
+    auto x = gko::initialize<Mtx>({0.0, 0.0, 0.0, 0.0, 0.0, 0.0}, this->exec);
+
+    solver->apply(b, x);
+
+    GKO_ASSERT_MTX_NEAR(x, l({33.0, -56.0, 81.0, -30.0, 21.0, 40.0}),
```
yhmtsai, line 549: I suggest to add a TODO here.

yhmtsai (changes requested):
I do not fully agree Pratik's comments.
Pratik's comments should only for the general testing case for different executor.
In the reference test, everything is initialized manually, so everything is under control.
I have checked the implementation with the paper and I do not find any difference between them. Because the paper also showed less convergence than CG, I do not hold this PR due to the relaxed error check on big system.

The change request is for the rest of my comments.

yhmtsai:
still miss the first check when having the first rho

gojakuch:
> still miss the first check when having the first rho and the test_install

sorry, I don't think I understood what you mean here

reference/test/solver/pipe_cg_kernels.cpp
```diff
@@ -0,0 +1,678 @@
...
+    auto solver = this->pipe_cg_factory_big->generate(this->mtx_big);
+    auto b = gko::initialize<Mtx>(
+        {1300083.0, 1018120.5, 906410.0, -42679.5, 846779.5, 1176858.5},
+        this->exec);
+    auto x = gko::initialize<Mtx>({0.0, 0.0, 0.0, 0.0, 0.0, 0.0}, this->exec);
+
+    solver->apply(b, x);
+
+    GKO_ASSERT_MTX_NEAR(x, l({81.0, 55.0, 45.0, 5.0, 85.0, -10.0}),
+                        r<value_type>::value * 5 * 1e3);
```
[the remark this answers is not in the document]
gojakuch: ok, so what should I put as the tolerance then?

reference/test/solver/pipe_cg_kernels.cpp
yhmtsai: I think the sentence until ...design better tests should be good enough

core/solver/pipe_cg.cpp
yhmtsai: one additional check after the first rho computation

gojakuch:
rebased the PR because the formatting bot complained about something unrelated. seems to have fixed that

yhmtsai (changes requested):
I think it still misses the part in ` test/test_install/test_install.cpp`.
It needs a section like
```
// core/solver/pipe_cg.hpp
{
    using Solver = gko::solver::PipeCg<>;
    check_solver<Solver>(exec, A_raw, b, x);
}
```
I also add the place which should contains the check.
you can use only one check but you need to rearrange the algorithm like CG
Otherwise, use one out of the loop and another in the loop after you get the updated residual norm.

gojakuch:
> I think it still misses the part in ` test/test_install/test_install.cpp`. It needs a section like
> 
> ```
> // core/solver/pipe_cg.hpp
> {
>     using Solver = gko::solver::PipeCg<>;
>     check_solver<Solver>(exec, A_raw, b, x);
> }
> ```
As I've mentioned in my comment before: I think another check is failing now, I suspect that it's because of me adding PipeCG to test_install like Mike suggested, while the unified kernels are not implemented in this PR. I think I'll just add it to the test_install in my next PR with kernel implementation

meaning that I've tried adding exactly this block of code but a new check started failing after this. I'll try this again now, but I can just do this in my next PR where I implement the kernels.

as for the check, I don't see how adding two checks is better, but I'll add this.

yhmtsai (approved):
LGTM. sorry for missing your comment and confusing suggestion.
Yes, test_install indeed also tests for different backend, so it will be failed when the other backend are enabled.
Only one thing left is the iteration count needs to be moved a little.
Thanks for implementing it!

gojakuch:
> LGTM. sorry for missing your comment and confusing suggestion. Yes, test_install indeed also tests for different backend, so it will be failed when the other backend are enabled. Only one thing left is the iteration count needs to be moved a little. Thanks for implementing it!

@yhmtsai thanks! I've rebased the PR and moved the iter counter before the second check, so that we don't check with iter==0 twice

pratikvn:
Thank you @gojakuch for your contribution!

PR #1826, opened 2025-04-15
Test labels and reduced fast test sizes

pratikvn:
This PR adds the ability to pass in labels to tests which can then be used to run tests matching specific labels `ctest -L <>`. It also reduces the sizes of some tests to speedup when `GINKGO_FAST_TESTS` is enabled.

I found the labels feature, and thought it would be interesting to have, but I think it is not necessary.


reference/test/matrix/fft_kernels.cpp
```diff
@@ -39,9 +39,15 @@ class Fft : public ::testing::Test {
     Fft()
         : exec(gko::ReferenceExecutor::create()),
           rng{7381},
+#ifdef GINKGO_FAST_TESTS
+          n1{2},
```
upsj, line 43: I think to test the complex arithmetic, it would be best if each dimension is at least 4

MarcelKoch:
What benefit does using the label option of ctest have compared to the regex option? It should already be possible to, for example, run all reference tests just by using `ctest -R reference`.

cmake/create_test.cmake
```diff
@@ -219,6 +221,12 @@ function(ginkgo_add_test test_name test_target_name)
             PROPERTIES ENVIRONMENT LD_PRELOAD=${test_preload}
         )
     endif()
+    if(add_test_LABEL)
+        set_tests_properties(
+            ${REL_BINARY_DIR}/${test_binary_name}
+            PROPERTIES LABELS "${add_test_LABEL}"
+        )
+    endif()
```
upsj, line 229: nice find - seems supported since CMake 3.16, so I think this could be useful. We could also use it to add executor information

upsj (approved):
LGTM!

upsj:
@MarcelKoch Labels can be a list of keywords, thus carry more information than just the test name.

MarcelKoch:
@upsj sure, but the question is rather, what labels could we have that are not already found in the test name?

upsj:
I think it would be cleaner to use labels for things like MPI usage, executor type, differentiation between core/device, since things like `mpi` or `omp` are also part of words that might pop up inside Ginkgo, like `compression` (or less likely: `compiler`)

yhmtsai:
you can also use `/omp` or `_omp` for that

upsj:
Those are two patterns that match different subsets of our tests. What I'm saying is that we are using the names for something they are not meant to do, labels would be a cleaner solution

PR #1830, opened 2025-04-17
Data validation for simple format

JEFFjeffJJ:


core/base/validation.hpp
```diff
@@ -0,0 +1,183 @@
...
+{
+    return is_finite_scalar(value.real()) && is_finite_scalar(value.imag());
+}
+
+
+template <typename IndexType>
+bool is_sorted(const gko::array<IndexType>& row_ptrs)
+{
+    const auto host_row_ptrs =
+        make_temporary_clone(row_ptrs.get_executor()->get_master(), &row_ptrs);
```
upsj, line 44: Since #1835, you can turn the data into a host `std::vector` with that function directly

core/base/validation.hpp
```diff
@@ -0,0 +1,183 @@
...
+}
+
+
+template <typename IndexType>
+bool is_within_bounds(const gko::array<IndexType>& col_idxs,
+                      const std::int64_t& lower_bound,
+                      const size_t& upper_bound)
+{
+    const auto host_col_idxs =
+        make_temporary_clone(col_idxs.get_executor()->get_master(), &col_idxs);
```
upsj, line 57: `col_idxs.copy_to_host()`

core/base/validation.hpp
```diff
@@ -0,0 +1,183 @@
...
+        host_col_idxs->get_const_data() + host_col_idxs->get_size());
+    return (*min >= lower_bound || *min == -1) &&
+           *max < upper_bound;  //-1 for padding? Probably.
+}
+
+
+template <typename ValueType>
+bool is_finite(const gko::array<ValueType>& values)
+{
+    const auto host_values =
```
upsj, line 83: `copy_to_host()`

core/base/validation.hpp
```diff
@@ -0,0 +1,183 @@
...
+
+
+template <typename IndexType>
+bool has_unique_ptrs(const gko::array<IndexType>& row_ptrs,
+                     const gko::array<IndexType>& col_idxs)
+{
+    const auto host_row_ptrs =
+        make_temporary_clone(row_ptrs.get_executor()->get_master(), &row_ptrs);
+    const auto host_col_idxs =
+        make_temporary_clone(col_idxs.get_executor()->get_master(), &col_idxs);
```
upsj, line 103: `copy_to_host()`

core/base/validation.hpp
```diff
@@ -0,0 +1,183 @@
...
+        result = result && unique_ptrs.size() == size;
+    }
+    return result;
+}
+
+template <typename IndexType>
+bool has_unique_perm_idxs(const gko::array<IndexType>& permutation_)
+{
+    const auto host_perm_idxs = make_temporary_clone(
+        permutation_.get_executor()->get_master(), &permutation_);
```
upsj, line 127: `copy_to_host()`

core/base/validation.hpp
```diff
@@ -0,0 +1,183 @@
...
+
+    const auto size = host_perm_idxs->get_size();
+    std::unordered_set<IndexType> unique_ptrs(
+        host_perm_idxs->get_const_data(),
+        host_perm_idxs->get_const_data() + size);
```
upsj, line 132: Here we additionally also want to make sure that the indices are actually the values 0, ..., size - 1, not just any unique values.

core/base/validation.hpp
```diff
@@ -0,0 +1,183 @@
...
+//     const auto row_ptrs = host_row_ptrs->get_const_data();
+//     const auto col_idxs = host_col_idxs->get_const_data();
+//     const auto values = host_values->get_const_data();
+
+//     if(row_ptrs->get_size() != col_idxs->get_size())
+//     {
+//         return false;
+//     }
+
+//     if (values->get_size() > num_rows * (num_rows -1))  //pigeonhole
```
upsj, line 161: I think until we need them, we can remove these functions instead of leaving them commented out

core/base/validation.hpp
```diff
@@ -0,0 +1,183 @@
...
+    return std::is_sorted(
+        host_row_ptrs->get_const_data(),
+        host_row_ptrs->get_const_data() + host_row_ptrs->get_size());
+}
+
+
+template <typename IndexType>
+bool is_within_bounds(const gko::array<IndexType>& col_idxs,
+                      const std::int64_t& lower_bound,
+                      const size_t& upper_bound)
```
upsj, line 54: I think those two types should be consistent with the value type of the array, at worst both should be int64, maybe even
```suggestion
                      IndexType lower_bound,
                      IndexType upper_bound)
```
Additionally, I'm not sure if we even need the lower_bound parameter, since it's always set to 0

core/base/validation.hpp
```diff
@@ -0,0 +1,183 @@
...
+        host_col_idxs->get_const_data(),
+        host_col_idxs->get_const_data() + host_col_idxs->get_size());
+    return *min >= lower_bound && *max < upper_bound;
+}
+
+
+template <typename IndexType>
+bool ell_is_within_bounds(const gko::array<IndexType>& col_idxs,
+                          const std::int64_t& lower_bound,
+                          const size_t& upper_bound)
```
upsj, line 68: Same here: IndexType or int64

core/base/validation.hpp
```diff
@@ -0,0 +1,183 @@
...
+{
+    const auto host_col_idxs =
+        make_temporary_clone(col_idxs.get_executor()->get_master(), &col_idxs);
```
upsj, line 71: `copy_to_host()`

core/base/validation.hpp
```diff
@@ -0,0 +1,183 @@
...
+            return false;
+        }
+    }
+
+    return true;
+}
+
+
+template <typename IndexType>
+bool has_unique_ptrs(const gko::array<IndexType>& row_ptrs,
```
upsj, line 97: This checks whether the column indices are unique, so maybe we should rather name it 
```suggestion
bool has_unique_columns(const gko::array<IndexType>& row_ptrs,
```

core/matrix/coo.cpp
```diff
@@ -47,6 +48,23 @@ GKO_REGISTER_OPERATION(aos_to_soa, components::aos_to_soa);
...
+    GKO_VALIDATE(
+        validation::is_within_bounds(row_idxs_, 0, this->get_size()[0]),
+        "row_idxs must be within bounds");
+
+    GKO_VALIDATE(
+        validation::is_within_bounds(col_idxs_, 0, this->get_size()[1]),
+        "col_idxs must be within bounds");
+
+    GKO_VALIDATE(validation::is_finite(values_),
+                 "matrix must contain only finite values");
```
upsj, line 66: I don't think we need to separate them by empty lines, also we can simplify the code a bit
```suggestion
    using namespace gko::validation;
    GKO_VALIDATE(is_sorted(row_idxs_), "row_idxs must be non-decending");
    GKO_VALIDATE(is_within_bounds(row_idxs_, 0, this->get_size()[0]), "row_idxs must be within bounds");
    GKO_VALIDATE(is_within_bounds(col_idxs_, 0, this->get_size()[1]), "col_idxs must be within bounds");
    GKO_VALIDATE(is_finite(values_), "matrix must contain only finite values");
```

core/matrix/csr.cpp
```diff
@@ -104,6 +105,28 @@ GKO_REGISTER_OPERATION(aos_to_soa, components::aos_to_soa);
...
+
+    GKO_VALIDATE(
+        validation::is_within_bounds(col_idxs_, 0, this->get_size()[1]),
+        "col_idxs must be within bounds");
+
+    GKO_VALIDATE(validation::is_finite(values_),
+                 "matrix must contain only finite values");
+
+    GKO_VALIDATE(validation::has_unique_ptrs(row_ptrs_, col_idxs_),
+                 "each row of col_idxs must contain unique values");
```
upsj, line 126: same as above: using namespace and no empty lines between

upsj (commented):
Hi Jeff, here my first pass over the code. What I looked at looks really good, I only have a few small suggestion (usually called nits) that could improve the quality of the code, e.g. simplify it or make it easier or clearer to read.

I would suggest that you walk through the individual comments, look at whether they make sense to you, and then after making each individual change, mark the corresponding thread as "Resolved". That allows us to see which changes you already did. Finally, you can create a new commit that includes all of these changes. Generally, we try to annotate the commits with `Co-authored-by: Reviewer Name <reviewer.mail@address.com>` in the commit message so we see who made the corresponding suggestions in the commit history

core/matrix/ell.cpp
```diff
@@ -54,6 +55,18 @@ GKO_REGISTER_OPERATION(outplace_absolute_array,
...
+template <typename ValueType, typename IndexType>
+void Ell<ValueType, IndexType>::validate_data() const
+{
+    GKO_VALIDATE(validation::is_finite(values_),
+                 "matrix must contain only finite values");
+
+    GKO_VALIDATE(
+        validation::ell_is_within_bounds(col_idxs_, 0, this->get_size()[1]),
+        "col_idxs must be within bounds");
+}
```
upsj, line 67: I just noticed: This function doesn't yet check that col_idxs_ doesn't contain duplicate indices, or has indices that appear after padding. We can take a closer look at what this means together, but in general, what you need to do is for each row look at the values `col_idxs_[row + i * this->get_stride()]` for `i = 0, ..., this->get_num_stored_elements_per_row() - 1`. Find the first `invalid_index<IndexType>()` value (if it exists), and all values before that should be unique in that row, all values afterwards should also be `invalid_index<IndexType>()`.

core/base/validation.hpp
```diff
@@ -0,0 +1,183 @@
...
+    bool result = true;
+
+    for (IndexType row = 0; row < num_rows_; row++) {
+        const auto begin = row_ptrs_[row];
+        const auto end = row_ptrs_[row + 1];
+        const auto size = end - begin;
+        std::unordered_set<IndexType> unique_ptrs(col_idxs_ + begin,
+                                                  col_idxs_ + end);
+
+        result = result && unique_ptrs.size() == size;
```
upsj, line 118: here you could to a short circuit
```suggestion
        if (unique_ptrs.size() < size) {
            return false;
        }
```

upsj (commented):
Two more comments on the algorithm implementation

core/base/validation.hpp
```diff
@@ -0,0 +1,167 @@
...
+    const auto host_col_idxs = col_idxs.copy_to_host();
+    const auto [min, max] =
+        std::minmax_element(host_col_idxs.begin(), host_col_idxs.end());
+
+    return *min >= 0 && *max < upper_bound;
+}
+
+
+template <typename IndexType>
+bool ell_has_unique_idxs(const gko::array<IndexType>& col_idxs,
```
upsj: These functions probably fit better into ell.cpp

core/base/validation.hpp
```diff
@@ -0,0 +1,167 @@
...
+                         const IndexType num_rows, const size_type nnz,
```
upsj: I would suggest a clearer name for the second parameter
```suggestion
                         const IndexType num_rows, const size_type nnz_per_row,
```

core/base/validation.hpp
```diff
@@ -0,0 +1,167 @@
...
+        for (size_type j = 0; j < nnz; ++j) {
+            const auto idx = host_col_idxs[i + stride * j];
+            if (idx == -1) {
+                continue;
+            }
+            if (idx != -1 && idx < num_rows) {
+                if (!unique_idxs.insert(idx).second) {
+                    return false;
+                }
+            }
```
upsj: This is slightly more lenient than what we are using in practice: A row with `1, -1, 1` would be invalid to us, since all values after a `-1` should also be `-1`

core/base/validation.hpp
```diff
@@ -0,0 +1,167 @@
...
+bool ell_has_unique_idxs(const gko::array<IndexType>& col_idxs,
+                         const IndexType num_rows, const size_type nnz,
+                         const size_type stride)
+{
+    const auto host_col_idxs = col_idxs.copy_to_host();
+    for (size_type i = 0; i < num_rows; ++i) {
+        std::unordered_set<IndexType> unique_idxs;
+        for (size_type j = 0; j < nnz; ++j) {
+            const auto idx = host_col_idxs[i + stride * j];
+            if (idx == -1) {
```
upsj: Instead of using a magic number -1, I would write
```suggestion
            if (idx == invalid_index<IndexType>()) {
```

core/base/validation.hpp
```diff
@@ -0,0 +1,167 @@
...
+            }
+        }
+    }
+    return true;
+}
+
+
+template <typename IndexType>
+bool ell_is_within_bounds(const gko::array<IndexType>& col_idxs,
+                          const IndexType num_rows, const size_type nnz,
```
upsj:
```suggestion
                          const IndexType num_rows, const size_type nnz_per_row,
```

core/base/validation.hpp
```diff
@@ -0,0 +1,167 @@
...
+                          const size_type stride)
+{
+    const auto host_col_idxs = col_idxs.copy_to_host();
+    for (size_type i = 0; i < num_rows; ++i) {
+        bool padding = false;
+        for (size_type j = 0; j < nnz; ++j) {
+            const auto idx = host_col_idxs[i + stride * j];
+            if (idx == -1) {
```
upsj:
```suggestion
            if (idx == invalid_index<IndexType>()) {
```

upsj (commented):
I just have some final comments on the ELL checks, otherwise it looks good to me :) I think you can also turn the PR from a draft into a ready PR by clicking on "Ready for review" at the bottom

core/base/validation.hpp
```diff
@@ -0,0 +1,167 @@
+// SPDX-FileCopyrightText: 2017 - 2025 The Ginkgo authors
```
yhmtsai: for the new one, it should be 2025 not 2017 - 2025

core/matrix/dense.cpp
```diff
@@ -106,6 +107,13 @@ GKO_REGISTER_OPERATION(add_scaled_identity, dense::add_scaled_identity);
 }  // anonymous namespace
 }  // namespace dense
 
+template <typename ValueType>
```
yhmtsai, line 116:
```suggestion


template <typename ValueType>
```
nit

core/matrix/csr.cpp
```diff
@@ -104,6 +105,24 @@ GKO_REGISTER_OPERATION(aos_to_soa, components::aos_to_soa);
...
+    GKO_VALIDATE(validation::is_sorted(row_ptrs_),
+                 "row_ptrs must be non-descending");
+    GKO_VALIDATE(validation::is_within_bounds(
+                     row_ptrs_, static_cast<IndexType>(this->get_size()[0])),
+                 "row_ptrs must be within bounds");
+    GKO_VALIDATE(validation::is_within_bounds(
+                     col_idxs_, static_cast<IndexType>(this->get_size()[1])),
+                 "col_idxs must be within bounds");
+    GKO_VALIDATE(validation::is_finite(values_),
+                 "matrix must contain only finite values");
```
yhmtsai, line 120: the error message seem to be strongly related to the function. do you need another `GKO_VALIDATE`?
I will probably throw it in the function with error message directly.
the calling stack can be done with gdb.

core/base/validation.hpp
```diff
@@ -0,0 +1,167 @@
...
+
+        if (unique_ptrs.size() < size) {
+            return false;
+        }
+    }
+    return result;
+}
+
+template <typename IndexType>
+bool has_unique_perm_idxs(const gko::array<IndexType>& permutation_)
```
yhmtsai: should be in cpp file

core/test/matrix/dense.cpp
```diff
@@ -441,4 +441,21 @@ TEST(DenseView, CustomViewKeepsRuntimeType)
...
+        5.0, 6.0, -1.0};
+    // clang-format on
+
+    auto m = gko::matrix::Dense<TypeParam>::create(
+        this->exec, gko::dim<2>{3, 2},
+        gko::make_array_view(this->exec, 9, data), 3);
+
+    ASSERT_THROW(m->validate_data(), gko::InvalidData);
+}
+
```
yhmtsai, line 460: you can delete the anonymous namespace

core/test/matrix/diagonal.cpp
```diff
@@ -193,4 +193,14 @@ TYPED_TEST(Diagonal, GeneratesCorrectMatrixData)
...
+TYPED_TEST(Diagonal, RecognizesInfiniteValue)
+{
+    using value_type = typename TestFixture::value_type;
+    auto m = gko::matrix::Diagonal<TypeParam>::create(this->exec);
+    m->read(
+        gko::matrix_data<TypeParam>{{3, 3}, {{2, 2, INFINITY}, {0, 0, 1.0}}});
+
+    ASSERT_THROW(m->validate_data(), gko::InvalidData);
+}
+
```
yhmtsai, line 205: same here

core/test/matrix/ell.cpp
```diff
@@ -231,3 +231,49 @@ TYPED_TEST(Ell, CanBeReadFromMatrixAssemblyData)
...
+
+    auto mtx = gko::matrix::Ell<value_type, index_type>::create(
+        this->exec, gko::dim<2>{3, 2},
+        gko::make_array_view(this->exec, 8, values),
+        gko::make_array_view(this->exec, 8, col_idxs), 2, 4);
+
+    ASSERT_THROW(mtx->validate_data(), gko::InvalidData);
+}
+
+TYPED_TEST(Ell, RecognizesDupplicateIndex)
```
yhmtsai, line 249:
```suggestion


TYPED_TEST(Ell, RecognizesDupplicateIndex)
```
nit

core/test/matrix/ell.cpp
```diff
@@ -231,3 +231,49 @@ TYPED_TEST(Ell, CanBeReadFromMatrixAssemblyData)
...
+
+
+TYPED_TEST(Ell, RecognizesInfiniteValue)
+{
+    using value_type = typename TestFixture::value_type;
+    using index_type = typename TestFixture::index_type;
+    value_type values[] = {INFINITY, 3.0, 4.0, -1.0, 2.0, 0.0, 0.0, -1.0};
+    index_type col_idxs[] = {0, 1, 0, -1, 1, 0, 0, -1};
+
+    auto mtx = gko::matrix::Ell<value_type, index_type>::create(
```
yhmtsai:
```suggestion
    index_type col_idxs[] = {0, 1, 0, -1, 1, 0, 0, -1};
    auto mtx = gko::matrix::Ell<value_type, index_type>::create(
```

Because your action and assert are in the same line, the assign part should not have a space.
you can use comment to distinguish if you want.

core/test/matrix/ell.cpp
```diff
@@ -231,3 +231,49 @@ TYPED_TEST(Ell, CanBeReadFromMatrixAssemblyData)
...
+}
+
+TYPED_TEST(Ell, RecognizesDupplicateIndex)
+{
+    using value_type = typename TestFixture::value_type;
+    using index_type = typename TestFixture::index_type;
+    value_type values[] = {1.0, 3.0, 4.0, -1.0, 2.0, 0.0, 0.0, -1.0};
+    index_type col_idxs[] = {0, 0, 0, -1, 1, 0, 0, -1};
+
+    auto mtx = gko::matrix::Ell<value_type, index_type>::create(
```
yhmtsai:
```suggestion
    index_type col_idxs[] = {0, 0, 0, -1, 1, 0, 0, -1};
    auto mtx = gko::matrix::Ell<value_type, index_type>::create(
```

core/test/matrix/ell.cpp
```diff
@@ -231,3 +231,49 @@ TYPED_TEST(Ell, CanBeReadFromMatrixAssemblyData)
...
+}
+
+TYPED_TEST(Ell, RecognizesUnboundedColumnIndex)
+{
+    using value_type = typename TestFixture::value_type;
+    using index_type = typename TestFixture::index_type;
+    value_type values[] = {1.0, 3.0, 4.0, -1.0, 2.0, 0.0, 0.0, -1.0};
+    index_type col_idxs[] = {4, 1, 0, -1, 1, 0, 0, -1};
+
+    auto mtx = gko::matrix::Ell<value_type, index_type>::create(
```
yhmtsai, line 267:
```suggestion
    index_type col_idxs[] = {4, 1, 0, -1, 1, 0, 0, -1};
    auto mtx = gko::matrix::Ell<value_type, index_type>::create(
```

include/ginkgo/core/base/exception.hpp
```diff
@@ -54,7 +56,7 @@ namespace gko {
  *
  * @ingroup error
  */
-class Error : public std::exception {
+class Error : public std::runtime_error {
```
yhmtsai, line 59: any reason for that?

include/ginkgo/core/matrix/coo.hpp
```diff
@@ -247,6 +247,11 @@ class Coo : public EnableLinOp<Coo<ValueType, IndexType>>,
     const LinOp* apply2(ptr_param<const LinOp> alpha, ptr_param<const LinOp> b,
                         ptr_param<LinOp> x) const;
 
+    /**
+     * Throws gko::Invalid exception if the data inside the arrays is invalid.
+     */
+    void validate_data() const;
```
yhmtsai: I think you can create a feature class contain validate_data virtual function and then override it here.

include/ginkgo/core/matrix/coo.hpp
```diff
@@ -247,6 +247,11 @@ class Coo : public EnableLinOp<Coo<ValueType, IndexType>>,
     const LinOp* apply2(ptr_param<const LinOp> alpha, ptr_param<const LinOp> b,
                         ptr_param<LinOp> x) const;
 
+    /**
+     * Throws gko::Invalid exception if the data inside the arrays is invalid.
```
yhmtsai:
```suggestion
     * Throws gko::InvalidData exception if the data inside the arrays is invalid.
```

include/ginkgo/core/base/exception.hpp
upsj: That was a change I added: std::runtime_error has its own resource management for `what()` that works correctly with C++ exception semantics, std::string doesn't handle all edge cases properly IIRC.

> Because copying std::runtime_error is not permitted to throw exceptions, this message is typically stored internally as a separately-allocated reference-counted string. This is also why there is no constructor taking std::string&&: it would have to copy the content anyway.

core/test/matrix/dense.cpp
JEFFjeffJJ: I am not sure if I sure leave it there because most of other .cpp here, such as csr, or fbcsr, still contains the anonymous namespace and I didn't add that namespace. Thus, I don't know is it useful for others.

upsj: I assume the anonymous namespaces are there to avoid creating excess symbols, but since we don't link together multiple test files into a single binary, the impact on build performance should be small. This probably comes from Google's development best practices, where build performance matters on the large scale. As soon as we handle symbol visibility ourselves (like #715), we will also be able to achieve the same effect without the anonymous namespace.

core/base/validation.hpp
```diff
@@ -0,0 +1,142 @@
...
+}
+
+
+template <typename IndexType>
+bool is_sorted(const gko::array<IndexType>& row_ptrs)
+{
+    const auto host_row_ptrs = row_ptrs.copy_to_host();
+    for (size_t i = 0; i + 1 < host_row_ptrs.size(); ++i) {
+        if (host_row_ptrs[i] > host_row_ptrs[i + 1]) {
+            std::printf(
```
yhmtsai: `printf` is C function to print the message. I will suggest to use cout and you do not need to consider the variable type if you still need to print.
Here, I do not think it should print something in the check and sorting property may be required from the other function.
- if it is only for row_ptr, change the function name to `is_sorted_row_ptrs` to match the error message. Otherwise, change the error message
- I do not think it should print the error. I think it return the index of error should be enough
`std::optional` may help when designed it reversely, like `is_not_sorted`.
always returning a pair works but it will not be available in if condition directly.
Create custom type works perfectly but needs more work

core/base/validation.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+
+template <typename ValueType>
+bool is_finite(const gko::array<ValueType>& values)
+{
+    const auto host_values = values.copy_to_host();
+    for (size_t i = 0; i < host_values.size(); ++i) {
+        if (!is_finite_scalar(host_values[i])) {
+            // throw std::invalid_argument("matrix must contain only finite
+            // values");
```
yhmtsai:
```suggestion
```
I think you can output the location out of bound

core/base/validation.hpp
```diff
@@ -0,0 +1,142 @@
...
+            return false;
+        }
+    }
+    return true;
+}
+
+
+template <typename IndexType>
+bool has_unique_columns(const gko::array<IndexType>& row_ptrs,
```
yhmtsai:
```suggestion
bool has_unique_index(const gko::array<IndexType>& row_ptrs,
```
I think unique_index or unique_location is more suitable for this

core/base/validation.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+    const auto num_rows_ = host_row_ptrs.size() - 1;
+    bool result = true;
+
+    for (IndexType row = 0; row < num_rows_; row++) {
+        const auto begin = host_row_ptrs[row];
+        const auto end = host_row_ptrs[row + 1];
+        const auto size = end - begin;
+        std::unordered_set<IndexType> unique_ptrs(host_col_idxs.begin() + begin,
+                                                  host_col_idxs.begin() + end);
```
yhmtsai, line 110:
```suggestion
        std::unordered_set<IndexType> unique_cols(host_col_idxs.begin() + begin,
                                                  host_col_idxs.begin() + end);
```
nit: `unique_ptrs` is too close to `unique_ptr`

core/matrix/dense.cpp
```diff
@@ -107,6 +108,14 @@ GKO_REGISTER_OPERATION(add_scaled_identity, dense::add_scaled_identity);
 }  // namespace dense
 
 
+template <typename ValueType>
+void Dense<ValueType>::validate_data() const
+{
+    GKO_VALIDATE(validation::is_finite(values_),
```
yhmtsai: It should not check the values in the padding area (col >= #cols but < stride)

core/base/validation.hpp
```diff
@@ -0,0 +1,142 @@
...
+bool is_finite_scalar(const ValueType& value)
+{
+    return std::isfinite(static_cast<double>(value));
+}
+
+template <typename ValueType>
+bool is_finite_scalar(const std::complex<ValueType>& value)
+{
+    return is_finite_scalar(value.real()) && is_finite_scalar(value.imag());
+}
```
yhmtsai: we already have `is_finite` in math.hpp

core/matrix/ell.cpp
```diff
@@ -413,6 +430,51 @@ Ell<ValueType, IndexType>::create_const(
...
+{
+    const auto host_col_idxs = col_idxs.copy_to_host();
+    for (size_type i = 0; i < num_rows; ++i) {
+        std::unordered_set<IndexType> unique_idxs;
+        for (size_type j = 0; j < num_non_zero_per_row; ++j) {
+            const auto idx = host_col_idxs[i + stride * j];
+            if (idx == -1) {
+                continue;
+            }
+            if (idx != -1 && idx < num_rows) {
```
yhmtsai, line 485:
```suggestion
            if (idx < num_rows) {
```
it is checked before

yhmtsai: if you want to put them together, remove the previous check

core/matrix/ell.cpp
```diff
@@ -413,6 +430,51 @@ Ell<ValueType, IndexType>::create_const(
...
+                }
+            }
+        }
+    }
+    return true;
+}
+
+
+template <typename ValueType, typename IndexType>
+bool Ell<ValueType, IndexType>::is_within_bounds(
```
yhmtsai: the function did something different then the name.
The implementation is check whether there is a value not `invalid_index` after the first `invalid_index`.

core/matrix/permutation.cpp
```diff
@@ -180,6 +188,28 @@ Permutation<IndexType>::create_const(
         exec, gko::detail::array_const_cast(std::move(perm_idxs))}};
 }
 
+template <typename IndexType>
+bool Permutation<IndexType>::has_unique_idxs(
+    const gko::array<IndexType>& permutation_)
+{
+    const auto host_perm_idxs = permutation_.copy_to_host();
+    const auto size = host_perm_idxs.size();
+    std::unordered_set<IndexType> unique_ptrs(host_perm_idxs.begin(),
```
yhmtsai:
```suggestion
    std::unordered_set<IndexType> unique_idxs(host_perm_idxs.begin(),
```
nit

core/matrix/permutation.cpp
```diff
@@ -180,6 +188,28 @@ Permutation<IndexType>::create_const(
...
+
+    if (unique_ptrs.size() < size) {
+        return false;
+    }
+
+    for (IndexType i = 0; i < static_cast<IndexType>(size); ++i) {
+        if (unique_ptrs.find(i) == unique_ptrs.end()) {
+            return false;
+        }
+    }
```
yhmtsai: It is out of scope from the function name.
you can check this implicitly by bound check before the unique_idxs check.
unique_idxs and all in bound will be one to one mapping

core/matrix/sellp.cpp
```diff
@@ -49,7 +49,6 @@ GKO_REGISTER_OPERATION(outplace_absolute_array,
 }  // anonymous namespace
 }  // namespace sellp
 
-
```
yhmtsai, line 52: you do not need remove this line.
no validation?

core/test/matrix/coo.cpp
```diff
@@ -398,4 +398,68 @@ TYPED_TEST(Coo, GeneratesCorrectMatrixData)
...
+    auto values = gko::array<value_type>(this->exec, 4);
+    auto m = Mtx::create(this->exec, gko::dim<2>{2, 3}, values.as_view(),
+                         col_idxs.as_view(), row_idxs.as_view());
+
+    m->read(
+        {{2, 3}, {{0, 0, INFINITY}, {0, 1, 3.0}, {0, 2, 2.0}, {1, 1, 5.0}}});
+    ASSERT_THROW(m->validate_data(), gko::InvalidData);
+}
+
+TYPED_TEST(Coo, RecognizesUnboundedRowIndex)
```
yhmtsai, line 418:
```suggestion
}


TYPED_TEST(Coo, RecognizesUnboundedRowIndex)
```
same for the others

include/ginkgo/core/matrix/ell.hpp
```diff
@@ -140,6 +140,10 @@ class Ell : public EnableLinOp<Ell<ValueType, IndexType>>,
 
     void compute_absolute_inplace() override;
 
+    // bool has_unique_idxs(const array<index_type>& col_idxs, index_type
+    // num_rows,
+    //     size_type num_non_zero_per_row, size_type stride) const;
```
yhmtsai:
```suggestion
```

include/ginkgo/core/matrix/ell.hpp
```diff
@@ -331,6 +341,30 @@ class Ell : public EnableLinOp<Ell<ValueType, IndexType>>,
         gko::detail::const_array_view<IndexType>&& col_idxs,
         size_type num_stored_elements_per_row, size_type stride);
 
+    /**
+     * Checks if all column indices in the given array are unique.
+     * @param col_idxs  the column index array of the matrix
```
yhmtsai:
```suggestion
     * Checks if all column indices in the given array are unique.
     * 
     * @param col_idxs  the column index array of the matrix
```

include/ginkgo/core/matrix/ell.hpp
```diff
@@ -331,6 +341,30 @@ class Ell : public EnableLinOp<Ell<ValueType, IndexType>>,
...
+     * @param num_rows  the number of rows in the matrix
+     * @param num_non_zero_per_row  the number of nonzeros per row
+     * @param stride  the column-stride of the value and column index array
+     */
+    static bool has_unique_idxs(const array<IndexType>& col_idxs,
```
yhmtsai: I think these function can be internal function in ell.cpp not in the class.
At least, it should not be in the public interface

include/ginkgo/core/matrix/ell.hpp
```diff
@@ -331,6 +341,30 @@ class Ell : public EnableLinOp<Ell<ValueType, IndexType>>,
...
+                                const IndexType num_rows,
+                                const size_type num_non_zero_per_row,
+                                const size_type stride);
+
+    /**
+     * Checks if all column indices in the given array are unique.
+     * @param col_idxs  the column index array of the matrix
```
yhmtsai:
```suggestion
     * Checks if all column indices in the given array are unique.
     * 
     * @param col_idxs  the column index array of the matrix
```

include/ginkgo/core/matrix/permutation.hpp
```diff
@@ -257,6 +263,11 @@ class Permutation : public EnableLinOp<Permutation<IndexType>>,
     static std::unique_ptr<const Permutation> create_const(
         std::shared_ptr<const Executor> exec,
         gko::detail::const_array_view<IndexType>&& perm_idxs);
+    /**
+     * Check if the permutation indices are unique.
+     * @param permutation_  the permutation indices
```
yhmtsai:
```suggestion
     * Check if the permutation indices are unique.
     * 
     * @param permutation_  the permutation indices
```

include/ginkgo/core/matrix/permutation.hpp
```diff
@@ -257,6 +263,11 @@ class Permutation : public EnableLinOp<Permutation<IndexType>>,
...
+     */
+    static bool has_unique_idxs(const gko::array<IndexType>& permutation_);
```
yhmtsai: same comments

include/ginkgo/core/matrix/coo.hpp
yhmtsai: I think it should be in feature class, not in any of base class like LinOp.

include/ginkgo/core/base/polymorphic_object.hpp
```diff
@@ -57,6 +57,8 @@ class PolymorphicObject : public log::EnableLogging<PolymorphicObject> {
                                                                     this);
     }
 
+    virtual void validate_data() const {}
+
```
yhmtsai, line 65: It should be in new feature class and the class inherit the feature class when they implement it.
It is good if we indeed provide `validate_data` to all ginkgo class.
However, it is not case, user can call it on any ginkgo object even if the object does not validate itself.
It will mislead users about the valid state of the object.

core/base/validation.hpp
JEFFjeffJJ: Should the value also be outputted? Then handling complex type is necessary, since the matrix contains different types and I haven't found a convention to output them. Need to do it manually.

core/matrix/dense.cpp
JEFFjeffJJ: Isn't most of paddings -1? Then they will probably be validated, right? So is it still necessary to check the paddings? Also, since "is_finite" function is used by other matrices, some doesn't have paddings, e.g. csr, we can only check it by implementing variances.

include/ginkgo/core/base/polymorphic_object.hpp
upsj: The goal should be to support this with every class if possible. If we don't have it available in all classes, we need special-case handling for every child inside a more complex class, e.g. Composition, solvers, multigrid etc, which I think would overcomplicate things.

core/base/validation.hpp
upsj: I think a custom type that is convertible to `bool` should do the trick, this would allow us to just return success/failure in simple cases, and actual location information for more complex cases.

Also the error should ideally happen at the location of the `validate_data` function, since we have `__LINE__` and type information for `this` available at that place.

core/matrix/permutation.cpp
JEFFjeffJJ: The reason that I implement a if and a for loop is because if the Indices are duplicated, we can terminate the function sooner. Then the for loop checks if the permutation is valid, since even if the array has the same size the specified, an out of bound index might still appear. I think a better way to deal is re-naming it.

core/matrix/sellp.cpp
yhmtsai: This unnecessary change is not reverted yet

core/test/matrix/coo.cpp
```diff
@@ -398,4 +398,71 @@ TYPED_TEST(Coo, GeneratesCorrectMatrixData)
...
+    auto col_idxs = gko::array<index_type>(this->exec, 4);
+    auto values = gko::array<value_type>(this->exec, 4);
+    auto m = Mtx::create(this->exec, gko::dim<2>{2, 3}, values.as_view(),
+                         col_idxs.as_view(), row_idxs.as_view());
+
+    m->read({{2, 3}, {{1, 1, 5.0}, {0, 0, 1.0}, {0, 1, 3.0}, {0, 2, 2.0}}});
+
+    ASSERT_THROW(m->validate_data(), gko::InvalidData);
+}
+
```
yhmtsai, line 467:
```suggestion
}


```
If you keep the anonymous namespace, our format guideline requires the code needs two empty line before namespace scope

core/test/matrix/ell.cpp
yhmtsai: It is not resolved yet

core/test/matrix/permutation.cpp
```diff
@@ -155,4 +155,15 @@ TYPED_TEST(Permutation, CanBeCleared)
...
+    using index_type = typename TestFixture::index_type;
+    index_type data[] = {1, 0, 2, 2};
+
+    auto m = gko::matrix::Permutation<index_type>::create_const(
+        this->exec, gko::array<index_type>::const_view(this->exec, 4, data));
+
+    ASSERT_THROW(m->validate_data(), gko::InvalidData);
+}
+
 }  // namespace
```
yhmtsai, line 170:
```suggestion
}


}  // namespace
```
same here. ref: https://github.com/ginkgo-project/ginkgo/wiki/Contributing-guidelines#whitespace

include/ginkgo/core/base/polymorphic_object.hpp
yhmtsai: We should not make it available to all class when we do not implement that.
The `validate_data` is available in `PolymorphicObject` with default empty implementation. User can call this function on Composition, solvers, multigrid and do not get any exception.
The pass misleads user that the object setup is correct although we do not validate that actually

include/ginkgo/core/base/polymorphic_object.hpp
```diff
@@ -57,6 +57,8 @@ class PolymorphicObject : public log::EnableLogging<PolymorphicObject> {
                                                                     this);
     }
 
+    virtual void validate_data() const {}
```
yhmtsai, line 64:
```suggestion
    /**
     * Throws gko::InvalidData exception if the data inside the arrays is
     * invalid.
     */
    virtual void validate_data() const {}
```
you can move the description here.
Then you do not need the description when overriding it.

include/ginkgo/core/matrix/csr.hpp
```diff
@@ -1113,6 +1110,12 @@ class Csr : public EnableLinOp<Csr<ValueType, IndexType>>,
         this->inv_scale_impl(make_temporary_clone(exec, alpha).get());
     }
 
+    /**
+     * Throws gko::InvalidData exception if the data inside the arrays is
+     * invalid.
+     */
+    void validate_data() const override;
```
yhmtsai:
```suggestion
    void validate_data() const override;
```
no need description when move the description to the virtual function

core/base/validation.hpp
```diff
@@ -0,0 +1,127 @@
...
+                                       std::to_string(result.exceptionIndex) + \
+                                       ". " _message " (" #_expression ")");   \
+        }                                                                      \
+    }
+
+// Now I have these questions, first i am a little bit stuck at how to print the
+// exception location through the new custom type, and do i print it or throw it
+// with exception message? Also, in the comment, why shoudln't i check the
+// padding area for dense?
+
```
yhmtsai: I need to check the current implementation for the first question.
I can reply the second question first.
Padding is not in the actual matrix (or physically), so we do not touch and validate it.
Padding can be uninitialized value or some specific value for later use.

yhmtsai (changes requested):

core/base/validation.hpp
yhmtsai: In general, we need to wrap information into exception rather than printing something else.
User can deal with exception by themselves, so some output on the screen will be weird  if user decide ignore or do additional stuff

core/base/validation.hpp
```diff
@@ -0,0 +1,127 @@
...
+
+struct ValidationResult {
+    bool isValid;
+    size_t exceptionIndex;
+
+    operator bool() const noexcept { return isValid; }
```
yhmtsai:
```suggestion
    explicit operator bool() const noexcept { return isValid; }
```
It will still work with if condition but it avoids implicit casting.

core/matrix/ell.cpp
```diff
@@ -443,6 +472,48 @@ Ell<ValueType, IndexType>::Ell(std::shared_ptr<const Executor> exec,
...
+    const size_type num_non_zero_per_row, const size_type stride)
+{
+    const auto host_col_idxs = col_idxs.copy_to_host();
+    for (size_type i = 0; i < num_rows; ++i) {
+        std::unordered_set<IndexType> unique_idxs;
+        for (size_type j = 0; j < num_non_zero_per_row; ++j) {
+            const auto idx = host_col_idxs[i + stride * j];
+            if (idx != -1 && idx < num_rows) {
+                if (!unique_idxs.insert(idx).second) {
+                    return {false, static_cast<size_t>(idx)};
```
yhmtsai:
```suggestion
                    return {false, static_cast<size_t>(i + stride * j)};
```
Because the error message is "at index ...", so it should be the location

core/base/validation.hpp
```diff
@@ -0,0 +1,127 @@
...
+
+// Now I have these questions, first i am a little bit stuck at how to print the
+// exception location through the new custom type, and do i print it or throw it
+// with exception message? Also, in the comment, why shoudln't i check the
+// padding area for dense?
+
+
+struct ValidationResult {
+    bool isValid;
+    size_t exceptionIndex;
```
yhmtsai:
```suggestion
    bool isValid;
    std::string exception_message;
```
I think Index is okay-ish now, but exception_message might be easy for customized message for different kind of validation

core/base/validation.hpp
```diff
@@ -0,0 +1,122 @@
...
+        if (host_idxs_array[i] > host_idxs_array[i + 1]) {
+            return {false, static_cast<size_t>(i)};
+        }
+    }
+    return {true, 0};
+}
+
+
+template <typename IndexType>
+ValidationResult is_within_bounds(const gko::array<IndexType>& idxs_array,
```
yhmtsai:
```suggestion
ValidationResult is_within_nonnegative_bounds(const gko::array<IndexType>& idxs_array,
```
Because it only works with nonnegative number

core/base/validation.hpp
```diff
@@ -0,0 +1,122 @@
...
+                                                  host_col_idxs.begin() + end);
+
+        if (unique_ptrs.size() < size) {
+            return {false, static_cast<size_t>(row)};
+        }
+    }
+    return {true, 0};
+}
+
+}  // namespace validation
```
yhmtsai, line 120:
```suggestion
}


}  // namespace validation
```

core/test/matrix/dense.cpp
```diff
@@ -441,4 +441,21 @@ TEST(DenseView, CustomViewKeepsRuntimeType)
 }
 
 
+TYPED_TEST(Dense, RecognizesInfiniteValue)
```
yhmtsai, line 444: need to add another test on the matrix with larger stride, which contains not finite value in the padding part. (no throw in this case)

core/matrix/dense.cpp
```diff
@@ -107,6 +108,14 @@ GKO_REGISTER_OPERATION(add_scaled_identity, dense::add_scaled_identity);
 }  // namespace dense
 
 
+template <typename ValueType>
+void Dense<ValueType>::validate_data() const
+{
+    GKO_VALIDATE(validation::assert_array_is_finite(values_),
+                 "matrix must contain only finite values");
```
yhmtsai: do not need to check the value in padding but not in the matrix

core/base/validation.hpp
JEFFjeffJJ: For this I am thinking if message tells something, but this parameter now only provide some information. Maybe we can name it otherwise?

core/matrix/dense.cpp
```diff
@@ -2079,6 +2093,31 @@ Dense<ValueType>::Dense(std::shared_ptr<const Executor> exec,
 }
 
 
+template <typename ValueType>
+validation::ValidationResult dense_matrix_values_are_finite(
+    const Dense<ValueType>& mtx)
```
yhmtsai:
```suggestion
    const Dense<ValueType>* mtx)
```
the signature is different

core/matrix/dense.cpp
```diff
@@ -107,6 +108,19 @@ GKO_REGISTER_OPERATION(add_scaled_identity, dense::add_scaled_identity);
...
 
+template <typename ValueType>
+validation::ValidationResult dense_matrix_values_are_finite(
+    const Dense<ValueType>* mtx);
+
+
+template <typename ValueType>
+void Dense<ValueType>::validate_data() const
+{
+    GKO_VALIDATE(dense_matrix_values_are_finite(*this),
```
yhmtsai:
```suggestion
    GKO_VALIDATE(dense_matrix_values_are_finite(this),
```

core/matrix/dense.cpp
```diff
@@ -2079,6 +2093,31 @@ Dense<ValueType>::Dense(std::shared_ptr<const Executor> exec,
...
 
+template <typename ValueType>
+validation::ValidationResult dense_matrix_values_are_finite(
+    const Dense<ValueType>& mtx)
+{
+    if constexpr (std::is_integral<ValueType>::value) {
+        return {true, 0};
+    } else {
+        const auto host_mtx =
+            gko::clone(mtx.get_executor()->get_master(), &mtx);
```
yhmtsai:
```suggestion
        const auto host_mtx = mtx->clone(mtx->get_executor()->get_master());
```

core/test/matrix/dense.cpp
yhmtsai: still miss data
consider the data
[2.0 -1.0 inf 2.0 -1.0 1.0] 2 x 2 matrix with stride 3.
inf is in the padding, but the matrix is valid [2.0 -1.0; 2.0 -1.0]

PR #1832, opened 2025-04-20
Improved CSR-to-COO conversion

upsj:
This conversion can suffer from load imbalance in imbalanced matrices, this new algorithm is always perfectly load-balanced.

### DIFF
--- a/common/unified/components/format_conversion_kernels.cpp
+++ b/common/unified/components/format_conversion_kernels.cpp
@@ -8,6 +8,7 @@
 
 #include "common/unified/base/kernel_launch.hpp"
 #include "common/unified/components/bitvector.hpp"
+#include "core/base/index_range.hpp"
 #include "core/base/iterator_factory.hpp"
 #include "core/components/fill_array_kernels.hpp"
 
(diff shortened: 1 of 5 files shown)


common/unified/components/format_conversion_kernels.cpp
```diff
@@ -21,16 +24,23 @@ void convert_ptrs_to_idxs(std::shared_ptr<const DefaultExecutor> exec,
                           const RowPtrType* ptrs, size_type num_blocks,
                           IndexType* idxs)
 {
+    const auto num_elements = exec->copy_val_to_host(ptrs + num_blocks);
```
yhmtsai, line 27: maybe we just paste the idxs size from outside. We should have it already when allocating the idxs array

common/unified/components/format_conversion_kernels.cpp
```diff
@@ -21,16 +24,23 @@ void convert_ptrs_to_idxs(std::shared_ptr<const DefaultExecutor> exec,
...
+    // transform the ptrs to a bitvector in unary delta encoding, i.e.
+    // every row with n elements is encoded as 1 0 ... n times ... 0
```
yhmtsai, line 29:
```suggestion
    // every row with n elements is encoded as 1 0 ... n times ... 0
    // we only process the value when bv is zero, the prefix-sum of bv - 1, which is get_rank() - 1, is the row index.
```

yhmtsai (approved):
LGTM!

PR #1834, opened 2025-04-20
Show correct line numbers in generated source lines for common tests

upsj:
For CUDA/HIP and DPCPP, we create separate copies of the source files using configure_file(...) to apply different CMake file properties to them. For the compiler to report error messages with the correct source file name, this adds a #line preprocessor statement to the top of these generated files.

### DIFF
--- a/test/base/index_range.cpp
+++ b/test/base/index_range.cpp
@@ -1,6 +1,8 @@
-// SPDX-FileCopyrightText: 2017 - 2024 The Ginkgo authors
+// SPDX-FileCopyrightText: 2017 - 2025 The Ginkgo authors
 //
 // SPDX-License-Identifier: BSD-3-Clause
+
+/*@GKO_PREPROCESSOR_FILENAME_HELPER@*/
 
 #include "core/base/index_range.hpp"
 
--- a/test/base/intrinsics.cpp
+++ b/test/base/intrinsics.cpp
@@ -2,6 +2,8 @@
 //
 // SPDX-License-Identifier: BSD-3-Clause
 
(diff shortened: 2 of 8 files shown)


test/base/index_range.cpp
```diff
@@ -1,7 +1,9 @@
-// SPDX-FileCopyrightText: 2017 - 2024 The Ginkgo authors
+// SPDX-FileCopyrightText: 2017 - 2025 The Ginkgo authors
 //
 // SPDX-License-Identifier: BSD-3-Clause
 
+/*@GKO_PREPROCESSOR_FILENAME_HELPER@*/
```
upsj, line 5: I am using this weird comment structure to make sure both the original file and the generated file are valid C++ code. The ´#line` directive can start after a comment, but afaik it can't end with a comment, so that needs to be in the next line.

test/base/intrinsics.cpp
```diff
@@ -2,6 +2,8 @@
 //
 // SPDX-License-Identifier: BSD-3-Clause
 
+/*@GKO_PREPROCESSOR_FILENAME_HELPER@*/
```
pratikvn, line 5: Maybe I am missing something, wouldn't this be now after preprocessing be `/*/*#line 5 .... */*/` ?

pratikvn: Okay, the above comment clarifies it. :)

pratikvn (approved):

test/base/intrinsics.cpp
upsj: The actual string will be
```
/**/
#line 5 "file.cpp"
/**/

PR #1835, opened 2025-04-22
Add `array::copy_to_host` function

upsj:
For debugging and development purposes, this makes it easier to get a host view of device data.


MarcelKoch (approved):
LGTM. A bit confused that we don't have a unified array test, but that is not really relevant to the PR.

PR #1836, opened 2025-04-27
Add transform_iterator

upsj:
This adds a transform_iterator that can be used with standard library algorithms and other functions like the ones from #1820

### DIFF
--- a/core/test/base/iterator_factory.cpp
+++ b/core/test/base/iterator_factory.cpp
@@ -510,10 +510,36 @@ TYPED_TEST_SUITE(TransformIterator, gko::test::ComplexAndPODTypes,
                  TypenameNameGenerator);
 
 
+struct identity_transform {
+    template <typename T>
+    T operator()(T v) const
+    {
+        return v;
+    }
+};
+
+
+template <typename T>
+struct scale_transform {
+    T operator()(T v) const { return scale * v; }
+
+    T scale;
+};
+
+
+struct double_transform {
+    template <typename T>
+    T operator()(T v) const
+    {
+        return T{2} * v;
+    }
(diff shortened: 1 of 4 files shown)


core/test/base/iterator_factory.cpp
```diff
@@ -511,4 +512,141 @@ TYPED_TEST(PermuteIterator, DecreasingIterator)
...
+TYPED_TEST_SUITE(TransformIterator, gko::test::ComplexAndPODTypes,
+                 TypenameNameGenerator);
+
+
+TYPED_TEST(TransformIterator, EmptyIterator)
+{
+    auto test_iter = gko::detail::make_transform_iterator<TypeParam*>(
+        nullptr, [](TypeParam v) { return v; });
+
+    ASSERT_NO_THROW((void)std::find(test_iter, test_iter, TypeParam{}));
```
yhmtsai, line 530: does it check anything? `std::find` does not do anything because first and end are the same.
also what is `(void)` for?

upsj: No, this is just a copy of the same test for permute_iterator. I needed to use a read-only operation, so find was the first thing that came to mind. The (void) is there to silence the `[[nodiscard]]` warning.

yhmtsai (approved):
LGTM from the implementation, but I am not familiar with the iterator implementation.
If someone can review it to check whether it match the std iterator requirement or you can give the reference. I only check the iterator_traits and the random access iterator requirement.
I assume it will just give a compilation error when the Iterator is not random accessable because the following member functions uses the iterator as random access iterator.

ftschirpke:
format!

PR #1837, opened 2025-04-30
Add OpenMP segmented prefix sum

upsj:
Again part of #1758, this provides the equivalent of `thrust::exclusive_scan_by_key`

### DIFF
--- a/omp/components/prefix_sum.hpp
+++ b/omp/components/prefix_sum.hpp
@@ -1,0 +1,113 @@
+// SPDX-FileCopyrightText: 2017 - 2025 The Ginkgo authors
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+#ifndef GKO_OMP_COMPONENTS_PREFIX_SUM_HPP_
+#define GKO_OMP_COMPONENTS_PREFIX_SUM_HPP_
+
+#include <algorithm>
+#include <iterator>
+#include <limits>
+#include <string>
+
+#include <omp.h>
+
+#include "core/base/allocator.hpp"
+#include "core/base/iterator_factory.hpp"
+
+
+namespace gko {
+namespace kernels {
+namespace omp {
+namespace components {
+
+
+/*
+ * Similar to prefix_sum, only reduces within runs of the same key value (each
+ * key run must only occur once, otherwise the scan operation is not necessarily
+ * associaive). It also doesn't ignore the last value!
+ * Similar to thrust::exclusive_scan_by_key
+ */
+template <typename KeyIterator, typename Iterator,
+          typename ScanOp =
+              std::plus<typename std::iterator_traits<Iterator>::value_type>>
+void segmented_prefix_sum(
+    std::shared_ptr<const OmpExecutor> exec, KeyIterator key, Iterator it,
+    const size_type num_entries,
+    typename std::iterator_traits<KeyIterator>::value_type key_init = {},
+    typename std::iterator_traits<Iterator>::value_type init = {},
+    ScanOp op = {})
+{
+    using key_type = typename std::iterator_traits<KeyIterator>::value_type;
+    using value_type = typename std::iterator_traits<Iterator>::value_type;
+    // the operation only makes sense for arrays of size at least 2
+    if (num_entries < 2) {
+        if (num_entries == 0) {
+            return;
+        } else {
+            *it = init;
+            return;
+        }
+    }
+
+    const int nthreads = omp_get_max_threads();
+    vector<value_type> proc_sums(nthreads, init, {exec});
+    vector<key_type> proc_first_key(nthreads, key_init, {exec});
+    vector<key_type> proc_last_key(nthreads, key_init, {exec});
+    const size_type def_num_witems = (num_entries - 1) / nthreads + 1;
+
+#pragma omp parallel
+    {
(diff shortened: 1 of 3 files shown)


omp/components/prefix_sum.hpp
```diff
@@ -0,0 +1,113 @@
...
+    using value_type = typename std::iterator_traits<Iterator>::value_type;
+    // the operation only makes sense for arrays of size at least 2
+    if (num_entries < 2) {
+        if (num_entries == 0) {
+            return;
+        } else {
+            *it = init;
+            return;
+        }
+    }
```
yhmtsai, line 51:
```suggestion
    if (num_entries == 0) {
        return;
    } else if (num_entries == 1) {
        *it = init;
        return;
    }
    
```

omp/components/prefix_sum.hpp
```diff
@@ -0,0 +1,113 @@
...
+ * associaive). It also doesn't ignore the last value!
+ * Similar to thrust::exclusive_scan_by_key
+ */
+template <typename KeyIterator, typename Iterator,
+          typename ScanOp =
+              std::plus<typename std::iterator_traits<Iterator>::value_type>>
+void segmented_prefix_sum(
+    std::shared_ptr<const OmpExecutor> exec, KeyIterator key, Iterator it,
+    const size_type num_entries,
+    typename std::iterator_traits<KeyIterator>::value_type key_init = {},
```
yhmtsai, line 37: does `exclusive_scan_by_key` have key_init?

omp/components/prefix_sum.hpp
```diff
@@ -0,0 +1,113 @@
...
+    typename std::iterator_traits<Iterator>::value_type init = {},
+    ScanOp op = {})
```
yhmtsai, line 39: As it call prefix_sum, we do not need the `ScanOp`.

omp/components/prefix_sum.hpp
yhmtsai: Is it for the situation when KeyIterator is float or double?
I think we only need to consider it is integer (or bool/char) type because float and double as the key having the same value sounds unusual to me

omp/components/prefix_sum.hpp
```diff
@@ -0,0 +1,113 @@
...
+
+#pragma omp parallel
+    {
+        const int thread_id = omp_get_thread_num();
+        const size_type startidx = thread_id * def_num_witems;
+        const size_type endidx =
+            std::min(num_entries, (thread_id + 1) * def_num_witems);
+
+        auto partial_sum = init;
+        auto cur_key = startidx < num_entries ? key[startidx] : key_init;
```
yhmtsai, line 67: I think key_init can be the last key.
It may introduce one or two operation in `proc_sums[i + 1] = op(proc_sums[i], proc_sums[i + 1]);`, but I think it should be okay.

omp/components/prefix_sum.hpp
```diff
@@ -0,0 +1,113 @@
...
+namespace kernels {
+namespace omp {
+namespace components {
+
+
+/*
+ * Similar to prefix_sum, only reduces within runs of the same key value (each
+ * key run must only occur once, otherwise the scan operation is not necessarily
+ * associaive). It also doesn't ignore the last value!
+ * Similar to thrust::exclusive_scan_by_key
```
yhmtsai, line 29: I do not get it. if it does not ignore the last value, should it be the inclusive one?

omp/test/components/prefix_sum.cpp
```diff
@@ -0,0 +1,74 @@
...
+                                    [&] { return value_dist(this->rand); });
+                    std::copy(input.begin() + start, input.end(),
+                              std::back_inserter(ref_result));
+                    std::exclusive_scan(
+                        ref_result.begin() + start, ref_result.end(),
+                        ref_result.begin() + start, index_type{});
+                }
+
+                gko::kernels::omp::components::segmented_prefix_sum(
+                    this->exec, keys.cbegin(), input.begin(), keys.size());
```
yhmtsai, line 68:
```suggestion
                gko::kernels::omp::components::segmented_prefix_sum(
                    this->exec, keys.begin(), input.begin(), keys.size());
```

omp/test/components/prefix_sum.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+    std::shared_ptr<const gko::OmpExecutor> exec;
+    std::default_random_engine rand;
+    gko::size_type total_size;
+};
+
+TYPED_TEST_SUITE(PrefixSum, gko::test::IndexTypes, TypenameNameGenerator);
+
+
+TYPED_TEST(PrefixSum, SegmentedPrefixSumWorks)
```
yhmtsai, line 38: another test for checking the init is not 0

omp/components/prefix_sum.hpp
```diff
@@ -0,0 +1,113 @@
...
+ */
+template <typename KeyIterator, typename Iterator,
+          typename ScanOp =
+              std::plus<typename std::iterator_traits<Iterator>::value_type>>
+void segmented_prefix_sum(
```
yhmtsai, line 34: segmented_scan_by_key?

yhmtsai (changes requested):

PR #1839, opened 2025-05-07
Add complex type support for hipBLAS GEAM

nbeams:
`cgeam` and `zgeam` were added to hipBLAS in [ROCm 3.6.0](https://github.com/ROCm/hipBLAS/blob/develop/CHANGELOG.md#hipblas-0300-for-rocm-360), so we can expand the hipBLAS bindings.

(Needed over in #1814)

### DIFF
--- a/hip/base/hipblas_bindings.hip.hpp
+++ b/hip/base/hipblas_bindings.hip.hpp
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: 2017 - 2024 The Ginkgo authors
+// SPDX-FileCopyrightText: 2017 - 2025 The Ginkgo authors
 //
 // SPDX-License-Identifier: BSD-3-Clause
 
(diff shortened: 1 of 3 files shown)


yhmtsai (approved):

upsj:
Do we have any public-facing function that uses it? Maybe dense complex transpose? In that case I would say yes, otherwise no.

nbeams:
Hmm, yeah, it does look like it's used in dense transpose. I'll add it to the changelog, then -- once that's done, are there any complaints if I go ahead and merge this?

nbeams:
For `Fbcsr`, I enabled all the tests for complex types, which passed when I tested locally (with the half precision tests skipped). For `Dense`, it was just the conjugate transpose tests that were guarded for HIP. 

In the header file/doc strings for `Fbcsr`, I didn't see any specific note about it not being available for complex types with HIP, but since it wasn't tested, is that true? (I am wondering how much needs to go in the changelog.)

yhmtsai:
I think something like `enable complex support of Dense transpose and FbCsr on AMD GPU` in the changelog should be enough.

upsj (approved):

PR #1841, opened 2025-05-12
Enable more jobs relying on the server's spack and module

yhmtsai:
This PR enables more jobs running on our server and they are reusing the packages and spack from the system rather than building everything through spack.
- before_script: setup the environment to use the system packages.
- script: additionally allows environment variables `MODULE_LOAD` and `SPACK_LOAD` to load the package through the variables. They are simply expanded to `module load ${MODULE_LOAD}` and `spack load ${SPACK_LOAD}` if they contains context. cuda module from spack does not include `LD_LIBRARY_PATH`, so the script contains extending `LD_LIBRARY_PATH` before test_install
- image-tags: tags contains "s" like nvidia-gpus is the new settings. Do we have any specific tags additionally for the server or do we need to have?

TODO:
- [ ] move rocky_tum to some registry? It is a local image for apptainer now. It uses the same linux version as the server's system with the necessary components
- [ ] move the jobs to appropriate place (just show them together in quick condition for easy check)
- [ ] add more version from the current package sets


.gitlab-ci.yml
```diff
@@ -149,6 +148,76 @@ build/nvhpc227/cuda117/nompi/nvcpp/debug/shared:
...
+    BUILD_MPI: "ON"
+    BUILD_SHARED_LIBS: "OFF"
+    BUILD_TYPE: "Release"
+    MODULE_LOAD: "cmake/3.18.6 cuda/12.0.1 gcc/12.4.0 openmpi/4.1.8"
+
+build/cuda122/openmpi/gcc/cuda/release/shared:
+  extends:
+    - .build_and_test_tum_template
+    - .default_variables
+    - .quick_test_condition
```
yhmtsai: have a mpi test in quick pipeline

pratikvn: Maybe as this is in the quick test one, maybe having cuda is not necessary ? Maybe just MPI+ half is enough ?

pratikvn (commented):
Its a bit weird that the OSX jobs are failing.

.gitlab-ci.yml
yhmtsai: make sense

.gitlab-ci.yml
```diff
@@ -149,6 +148,76 @@ build/nvhpc227/cuda117/nompi/nvcpp/debug/shared:
...
+    - .full_test_condition
+    - .use_tum-nvidia
+  variables:
+    BUILD_CUDA: "ON"
+    BUILD_HWLOC: "OFF"
+    ENABLE_HALF: "ON"
+    BUILD_TYPE: "Release"
+    MODULE_LOAD: "cmake/3.20.6 cuda/12.2.2 gcc/12.4.0 openmpi/5.0.7"
+
+build/cuda124/mpich/gcc/cuda/release/shared:
```
pratikvn, line 192: I feel so many jobs might overload the TUM system ? Maybe we can:
1. Switch off MPI or CUDA wherever possible ? 
2. Not test multiple versions of the same CUDA major version ?

yhmtsai: It is not the scope of this pr.
We will allow a GPU can serve multiple instance. 
also, we are working on another project to make the setup more widely cover with limited resource

PR #1842, opened 2025-05-15
Fix inverse neighborhood comm

MarcelKoch:
This PR fixes the generation of the inverse of a neighborhood communicator. Instead of the one-sided communication, it uses an all-to-all communication.

Some MPI implementations seem to take issues with the one-sided implementation.

### DIFF
--- a/core/distributed/neighborhood_communicator.cpp
+++ b/core/distributed/neighborhood_communicator.cpp
@@ -7,6 +7,8 @@
 #include <ginkgo/core/base/precision_dispatch.hpp>
 #include <ginkgo/core/matrix/dense.hpp>
 
+#include "core/base/allocator.hpp"
+
 
 namespace gko {
 namespace experimental {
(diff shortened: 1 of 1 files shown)


pratikvn:
Which MPI implementations have issues?

MarcelKoch:
@yhmtsai noticed it. Perhaps it's also based on UCX.

yhmtsai:
It is OpenMPI 5.0.3 with GCC 13.3.0.
but it only introduces issue with certain #nodes and size.

core/distributed/neighborhood_communicator.cpp
```diff
@@ -31,15 +33,13 @@ communicate_inverse_envelope(std::shared_ptr<const Executor> exec,
                              const std::vector<comm_index_type>& sizes)
 {
     auto host_exec = exec->get_master();
```
pratikvn, line 35: I think the doc of the function needs to be updated ?

pratikvn (changes requested):
As this is used only when creating the neighborhood communicator, this is probably not a bottleneck and should be fine, but I think the doc needs to be updated. Otherwise, looks good to me.

pratikvn (approved):

PR #1843, opened 2025-05-16
Move CI job and fix cuda12.2 cusparse matrix, coo exception, workspace reallcation

yhmtsai:
This PR extracts some part from #1841 . This PR does not add the spack availability or more jobs.

- move the job from amdci to tum
  - ~~cuda below 11.8 only check build not test now because current setup can not always get the gpu supported by these compiler~~
  - ~~cudamemcheck task uses cuda12 docker image and compute-sanitizer not cuda11 with the same reason. TODO: check whether it work or not~~
  - some old cuda jobs can be moved to P100 now, so we do not have above changes
  - wait for cuda memcheck from https://gitlab.com/ginkgo-project/ginkgo-public-ci/-/jobs/10059330364
- fix coo exception on empty matrix
- fix row gatherer workspace reallocation
- fix cuda12.2 csr does not accept nullptr as row ptr in the beginning.
- use csr for non-local matrix in the test because some of backend does not support Coo spmv on 16bit operation

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -662,9 +662,6 @@ cudamemcheck:
     - .default_variables
     - .deploy_condition
     - .use_gko_cuda110-mvapich-gnu9-llvm9
-  tags:
-    - private_ci
-    - nvidia-gpu
   script:
     - ctest -V -S cmake/CTestScript.cmake -DCTEST_BUILD_CONFIGURATION=RelWithDebInfo
       -DCTEST_MEMORYCHECK_TYPE=CudaMemcheck
--- a/.gitlab/image.yml
+++ b/.gitlab/image.yml
@@ -1,7 +1,6 @@
 .use_status-job-settings:
   image: ginkgohub/cpu:openmpi-gnu9-llvm8
   tags:
-    - private_ci
     - status-jobs
 
 .use_gko-nocuda-openmpi-gnu9-llvm8:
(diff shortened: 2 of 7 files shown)


.gitlab/image.yml
```diff
@@ -1,8 +1,8 @@
 .use_status-job-settings:
   image: ginkgohub/cpu:openmpi-gnu9-llvm8
   tags:
-    - private_ci
-    - status-jobs
+    - cpu
+    - tum
```
upsj, line 5: I set up a separate VM for status jobs, you can use the tag `status-jobs` for this, it shouldn't run on slurm.

yhmtsai: moved

.gitlab-ci.yml
```diff
@@ -656,15 +656,12 @@ undefinedsanitizer:
       -DCTEST_MEMORYCHECK_TYPE=UndefinedBehaviorSanitizer
 
 cudamemcheck:
-  stage: QoS_tools
+  stage: build
```
MarcelKoch, line 659: should this stay?

.gitlab-ci.yml
```diff
@@ -656,15 +656,12 @@ undefinedsanitizer:
...
   extends:
     - .before_script_template
     - .default_variables
-    - .deploy_condition
+    - .quick_test_condition
```
MarcelKoch, line 663: same here

MarcelKoch (changes requested):

.gitlab-ci.yml
yhmtsai: I would like to ensure that it is working before I merge this.
But I wonder whether it exceed the time limit of slurm.

yhmtsai: I will reverse it before merging when the job is finished once.

yhmtsai: It is passed. I have also restructured the commit history with reversing the temporary check.

MarcelKoch (approved):

PR #1660, opened 2024-08-05
Fix issues with distributed preconditioners, add to benchmarks 

pratikvn:
This PR adds benchmarking support for distributed preconditioners. It also fixes a few things that were causing increased allocations for distributed preconditioners.

PR #1845, opened 2025-05-16
Add distributed preconditioners to benchmarks

pratikvn:
This PR adds distributed preconditioners to benchmarks.

Extracted from: https://github.com/ginkgo-project/ginkgo/pull/1660

### DIFF
--- a/benchmark/utils/preconditioners.hpp
+++ b/benchmark/utils/preconditioners.hpp
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: 2017 - 2024 The Ginkgo authors
+// SPDX-FileCopyrightText: 2017 - 2025 The Ginkgo authors
 //
 // SPDX-License-Identifier: BSD-3-Clause
 
@@ -22,7 +22,11 @@ DEFINE_string(preconditioners, "none",
               "A comma-separated list of preconditioners to use. "
               "Supported values are: none, jacobi, paric, parict, parilu, "
               "parilut, ic, ilu, paric-isai, parict-isai, parilu-isai, "
-              "parilut-isai, ic-isai, ilu-isai, sor, overhead");
+              "parilut-isai, ic-isai, ilu-isai, sor, overhead"
+#ifdef GINKGO_BUILD_MPI
+              ", schwarz-jacobi, schwarz-ilu, schwarz-ic, schwarz-lu, dist-mg"
+#endif
+);
 
 DEFINE_uint32(parilu_iterations, 5,
               "The number of iterations for ParIC(T)/ParILU(T)");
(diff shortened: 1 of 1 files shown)


benchmark/utils/preconditioners.hpp
```diff
@@ -22,7 +22,11 @@ DEFINE_string(preconditioners, "none",
               "A comma-separated list of preconditioners to use. "
               "Supported values are: none, jacobi, paric, parict, parilu, "
               "parilut, ic, ilu, paric-isai, parict-isai, parilu-isai, "
-              "parilut-isai, ic-isai, ilu-isai, sor, overhead");
+              "parilut-isai, ic-isai, ilu-isai, sor, overhead"
+#ifdef GINKGO_BUILD_MPI
+              ", schwarz-jacobi, schwarz-ilu, schwarz-ic, schwarz-lu, dist-mg"
```
MarcelKoch, line 27: Shouldn't `mg` just be its own thing? You can use it also without mpi.

MarcelKoch, line 27: IMO it should just be `schwarz` and then another option to set the local preconditioner for schwarz.

benchmark/utils/preconditioners.hpp
```diff
@@ -55,6 +59,18 @@ DEFINE_double(sor_relaxation_factor, 1.0,
...
+            "Use deterministic computation of the aggregated group within PGM");
+
+DEFINE_uint32(
+    mg_max_num_levels, false,
+    "The maximum number of levels to use for the Multigrid preconditioner");
+
+DEFINE_double(mg_tolerance, false, "The tolerance for the coarse solver");
+
+DEFINE_uint32(mg_max_iters, false,
+              "The max number of iterations for the coarse solver");
```
MarcelKoch, line 72: To me, this looks like we reached the limit of what we can do through the CLI. I think for now this is fine, but we should consider in the long term to also allow our json config to define these parameters.

MarcelKoch (changes requested):
The changes are mostly fine (I would want to see the `mg` also for non-distributed), but there are two more levels of generality we could achieve:
1. define the inner schwarz preconditioner through a separate option;
2. define everything directly through our json format.

I think 2. is definitely not in the scope of this PR. For 1., however, I could see this getting added.

benchmark/utils/preconditioners.hpp
pratikvn: After looking into it a bit, I think this makes it more annoying to manage, because the local_solver also needs to be generated at the same time due to the way we set this up. 

For now, I would prefer to keep it the way it is and overhaul this with the JSON setup.

pratikvn: I updated it to use the default options. The default is Gmres+Schwarz-Jacobi as coarse solver for distributed, which may be overkill.

MarcelKoch: This still has the old `dist-mg`.

PR #1847, opened 2025-05-19
Support METIS with built-in GKLib

upsj:
Previously, we only supported METIS 5.x when it was exposed as two separate libraries libmetis.so and libGKLib.so
This is how METIS is available in Arch and when built manually, but Spack, Ubuntu, Rocky and others combine the two together into a single library.

cc @fritzgoebel


yhmtsai (approved):
LGTM. that will be great if @fritzgoebel can also confirm it works for his environment before merging this.

fritzgoebel (approved):
LGTM, thanks for fixing this!

PR #1850, opened 2025-05-23
fix the race condition in lu factorization

yhmtsai:
This PR fixes the race condition in lu factorization.

we need a sync before grabbing `val[lower_nz]` because the warp can modify the entry in the previous iteration.
We only need the warp sync. The wait(dep) implicitly has that if I understand it correctly, so I just move it after the wait(dep).
After move, we need another sync before assigning scale to ensure every thread gets the data before modification.

### DIFF
--- a/common/cuda_hip/factorization/lu_kernels.cpp
+++ b/common/cuda_hip/factorization/lu_kernels.cpp
@@ -112,19 +112,21 @@ __global__ __launch_bounds__(default_block_size) void factorize(
         storage,  row_descs, static_cast<size_type>(row)};
     // for each lower triangular entry: eliminate with corresponding row
     for (auto lower_nz = row_begin; lower_nz < row_diag; lower_nz++) {
+        // prevent data races between updates of vals for this row in the
+        // previous loop iteration and following reads of vals we can load the
+        // value without waiting for any dependencies because this warp is the
+        // only one writing to this row.
+        warp.sync();
+        const auto val = vals[lower_nz];
         const auto dep = cols[lower_nz];
         const auto diag_idx = diag_idxs[dep];
         const auto dep_end = row_ptrs[dep + 1];
         scheduler.wait(dep);
-        // We need to load vals after synchronize.
-        // the next lower_nz might be modified if the dep row has the same col
-        // as next lower_nz's col.
-        const auto val = vals[lower_nz];
         const auto diag = vals[diag_idx];
-        // we need sync to ensure all threads get the data before assigning to
-        // scale.
-        warp.sync();
         const auto scale = val / diag;
+        // prevent data races between preceding read and following write of
+        // vals[lower_nz]
+        warp.sync();
         if (lane == 0) {
             vals[lower_nz] = scale;
         }
--- a/common/cuda_hip/factorization/lu_kernels.cpp
+++ b/common/cuda_hip/factorization/lu_kernels.cpp
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: 2017 - 2024 The Ginkgo authors
+// SPDX-FileCopyrightText: 2017 - 2025 The Ginkgo authors
 //
 // SPDX-License-Identifier: BSD-3-Clause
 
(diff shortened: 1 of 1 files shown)


common/cuda_hip/factorization/lu_kernels.cpp
```diff
@@ -113,14 +113,17 @@ __global__ __launch_bounds__(default_block_size) void factorize(
...
-        // updates only go past the diagonal of the dependency row, i.e. at
-        // least column dep + 1
-        const auto val = vals[lower_nz];
         const auto diag_idx = diag_idxs[dep];
         const auto dep_end = row_ptrs[dep + 1];
         scheduler.wait(dep);
+        // We need to load vals after synchronize.
+        // the next lower_nz might be modified if the dep row has the same col
+        // as next lower_nz's col.
+        const auto val = vals[lower_nz];
```
upsj, line 122: Not sure I follow - each warp only modifies memory locations belonging to its row, so there are only data races between threads of the same warp. So is this reordering actually necessary?

common/cuda_hip/factorization/lu_kernels.cpp
```diff
@@ -113,14 +113,17 @@ __global__ __launch_bounds__(default_block_size) void factorize(
...
         const auto diag = vals[diag_idx];
+        // we need sync to ensure all threads get the data before assigning to
+        // scale.
+        warp.sync();
```
upsj, line 126: makes sense, good catch!

common/cuda_hip/factorization/lu_kernels.cpp
upsj: What you essentially want to do is add another warp sync at the end of each iteration of the outer loop, to prevent the modification of previous loops racing with ready from the following loop iterations? I would prefer having that sync happen explicitly at the end of the loop than hidden inside the scheduler wait function.

yhmtsai: Exactly. It is to ensure that the modification of the same warp can be seen from the others when getting it.
Yes, it is what I mean in the PR description. We only need warp sync.
I thought the wait(dep) should also imply that because it make the memory visible in block at least.
I just move that rather than introducing the warp sync

common/cuda_hip/factorization/lu_kernels.cpp
upsj: I think this still does some superfluous changes that hide the actual fix: The move around `scheduler.wait(dep)` is unnecessary, the only thing that needs to be added is a `warp.sync()` at the top or bottom of the loop (so one iteration doesn't interleave with another iteration)

yhmtsai: yes, I have mentioned that, but is it not done implicitly by sceduler.wait(dep) together? 
If we would like to make them more clear. them we need two warp.sync() one before and after get val

upsj: Most of the time is spent in the scheduler.wait function, the additional warp sync doesn't really matter in terms of performance. I think we should also be more precise in our descriptions here, this is mixing up two different levels of dependencies.
- The warp syncs are necessary because of independent thread scheduling and the fact that wait(dep) uses only a single thread to wait
- The values for the current row will only be modified by threads from the same warp, so there is no potential for data races once we add a sync
- The wait function is necessary to synchronize with other warps and their row's values. The fact that it calls syncwarp is an implementation detail we should not rely on, as it might change in the future.

common/cuda_hip/factorization/lu_kernels.cpp
```diff
@@ -112,16 +112,19 @@ __global__ __launch_bounds__(default_block_size) void factorize(
...
+        // diagonal of the dependency row, i.e. at least column dep + 1
         const auto val = vals[lower_nz];
+        const auto dep = cols[lower_nz];
         const auto diag_idx = diag_idxs[dep];
         const auto dep_end = row_ptrs[dep + 1];
         scheduler.wait(dep);
         const auto diag = vals[diag_idx];
         const auto scale = val / diag;
+        // we need sync to ensure all threads get the data before assigning to
+        // scale.
```
upsj:
```suggestion
        // prevent data races between preceding read and following write of vals[lower_nz]
```

common/cuda_hip/factorization/lu_kernels.cpp
```diff
@@ -112,16 +112,19 @@ __global__ __launch_bounds__(default_block_size) void factorize(
         storage,  row_descs, static_cast<size_type>(row)};
     // for each lower triangular entry: eliminate with corresponding row
     for (auto lower_nz = row_begin; lower_nz < row_diag; lower_nz++) {
-        const auto dep = cols[lower_nz];
-        // we can load the value before synchronizing because the following
-        // updates only go past the diagonal of the dependency row, i.e. at
-        // least column dep + 1
+        // we can load the value before waiting the dependence with ensuring the
+        // previous run is done because the following updates only go past the
+        // diagonal of the dependency row, i.e. at least column dep + 1
```
upsj: I need to think about this again, I believe this comment (and my original comment) may be slightly misleading

common/cuda_hip/factorization/lu_kernels.cpp
```diff
@@ -141,6 +144,9 @@ __global__ __launch_bounds__(default_block_size) void factorize(
                 }
             }
         }
+        // we need sync to avoid the read in next round not during the current
+        // update
+        warp.sync();
```
upsj:
```suggestion
```

common/cuda_hip/factorization/lu_kernels.cpp
upsj: I think I would simplify the PR and documentation a bit by putting all references to synchronization in here
```suggestion
        // prevent data races between updates of vals for this row in the previous loop iteration
        // and following reads of vals
        // we can load the value without waiting for any dependencies because this warp is the only one writing to this row.
        warp.sync();
```
Otherwise we have two places that explain the same issue, one at the top of the loop and one at the bottom.

PR #1853, opened 2025-05-28
Properly handle cuda arch for unsupported function

yhmtsai:
For example, bfloat16 are natively supported after CC 80, we need to throw an exception or avoid failed comfiguration.
It is mostly handled by cmake option when user only compiles for one cuda arch.
Cuda allows to compile library for different arch.
However, `CUDA_ARCH` is only available in device code not host code, so using macro on host side does not have effect actually.
There is one host macro but will give the entire list.
We can only rely on the runtime dispatch on CC in this case and throw an exception when they are not available.
To achieve that, we need to provide a working version of atomic add just for compilation.

Side note: there are still some issue that compiling bfloat16 kernel in templated lambda after 12.2 on the architecture not natively supporting bfloat16 leads unknown device kernel in runtime. but if duplicate the kernel with full specialization will work. This requires further investigation.

### DIFF
--- a/common/cuda_hip/components/atomic.hpp
+++ b/common/cuda_hip/components/atomic.hpp
@@ -165,7 +165,7 @@ GKO_BIND_ATOMIC_ADD(__nv_bfloat16);
 
 
 #if defined(__CUDA_ARCH__) && __CUDA_ARCH__ < 700
-#define GKO_BIND_UNSUPPORT_ATOMIC_ADD(ValueType)     \
+#define GKO_BIND_UNSUPPORTED_ATOMIC_ADD(ValueType)   \
     __forceinline__ __device__ ValueType atomic_add( \
         ValueType* __restrict__ addr, ValueType val) \
     {                                                \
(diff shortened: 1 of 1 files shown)


common/cuda_hip/matrix/coo_kernels.cpp
```diff
@@ -323,10 +333,20 @@ void advanced_spmv2(std::shared_ptr<const DefaultExecutor> exec,
...
     }
     // not support 16 bit atomic
-#if !(defined(CUDA_VERSION) && (__CUDA_ARCH__ >= 700))
+#if !defined(CUDA_VERSION)
     if constexpr (sizeof(remove_complex<ValueType>) == sizeof(int16)) {
         GKO_NOT_SUPPORTED(c);
     } else
+#else
+    auto compute_capability = as<CudaExecutor>(exec)->get_major_version() * 10 +
+                              as<CudaExecutor>(exec)->get_minor_version();
```
pratikvn: Would be cleaner to make this a member of `CudaExecutor`, something like `get_compute_capability()`

yhmtsai: sounds good. done

PR #1855, opened 2025-05-30
Fix the reference behavior on factorization algorithm

yhmtsai:
In https://github.com/ginkgo-project/ginkgo/pull/1783, we fallback the algorithm to syncfree when sparselib is unsupported.
We also have sparselib implementation in reference so we should use that when algorithm is sparselib.

### DIFF
--- a/core/factorization/ic.cpp
+++ b/core/factorization/ic.cpp
@@ -102,9 +102,10 @@ std::unique_ptr<Composition<ValueType>> Ic<ValueType, IndexType>::generate(
         local_system_matrix.get(), false));
 
     std::shared_ptr<const matrix_type> ic;
-    // Compute LC factorization
+    // Compute IC factorization
     if (parameters_.algorithm == incomplete_algorithm::syncfree ||
-        exec == exec->get_master()) {
+        (!std::dynamic_pointer_cast<const ReferenceExecutor>(exec) &&
+         exec == exec->get_master())) {
         std::unique_ptr<gko::factorization::elimination_forest<IndexType>>
             forest;
         const auto nnz = local_system_matrix->get_num_stored_elements();
--- a/reference/test/factorization/ic_kernels.cpp
+++ b/reference/test/factorization/ic_kernels.cpp
@@ -1,20 +1,36 @@
-// SPDX-FileCopyrightText: 2017 - 2024 The Ginkgo authors
+// SPDX-FileCopyrightText: 2017 - 2025 The Ginkgo authors
 //
 // SPDX-License-Identifier: BSD-3-Clause
 
 #include <algorithm>
 #include <memory>
+#include <string>
 #include <vector>
 
 #include <gtest/gtest.h>
 
 #include <ginkgo/core/base/executor.hpp>
 #include <ginkgo/core/factorization/ic.hpp>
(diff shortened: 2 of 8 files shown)


upsj (changes requested):
Does this actually make a difference? The conditions should be equivalent, since ReferenceExecutor has OmpExecutor as a base class. Do we have a test for this behavior?

core/factorization/ic.cpp
```diff
@@ -104,7 +104,7 @@ std::unique_ptr<Composition<ValueType>> Ic<ValueType, IndexType>::generate(
     std::shared_ptr<const matrix_type> ic;
     // Compute LC factorization
     if (parameters_.algorithm == incomplete_algorithm::syncfree ||
-        exec == exec->get_master()) {
+        std::dynamic_pointer_cast<const OmpExecutor>(exec)) {
```
MarcelKoch: @upsj is right. It needs to be
```suggestion
        std::dynamic_pointer_cast<const OmpExecutor>(exec) && !std::dynamic_pointer_cast<const ReferenceExecutor>(exec)) {
```
¯\\_(ツ)_/¯

MarcelKoch (changes requested):

core/factorization/ic.cpp
yhmtsai: forgot reference is also OmpExecutor

core/factorization/ic.cpp
```diff
@@ -104,7 +104,7 @@ std::unique_ptr<Composition<ValueType>> Ic<ValueType, IndexType>::generate(
     std::shared_ptr<const matrix_type> ic;
     // Compute LC factorization
     if (parameters_.algorithm == incomplete_algorithm::syncfree ||
-        exec == exec->get_master()) {
+        !std::dynamic_pointer_cast<const ReferenceExecutor>(exec)) {
```
MarcelKoch: I think you still need the check for `OmpExecutor`, because otherwise this will also trigger for cuda/hip/dpcpp executor with sparselib selected.

MarcelKoch (changes requested):

reference/test/factorization/ic_kernels.cpp
```diff
@@ -118,6 +135,44 @@ TYPED_TEST(Ic, SetStrategy)
 }
 
 
+TYPED_TEST(Ic, UsesSyncFreeAlgorithm)
+{
+    using value_type = typename TestFixture::value_type;
+    using index_type = typename TestFixture::index_type;
+    bool contains_sparselib = false;
+    auto hooks = make_hooks(contains_sparselib);
+    this->ref->add_logger(gko::log::ProfilerHook::create_custom(
```
MarcelKoch: nit: I think a custom logger type that only logs operations would be easier to understand.

MarcelKoch (approved):

reference/test/factorization/ic_kernels.cpp
yhmtsai: I change it. does that match?

MarcelKoch: yes, thanks for that.

MarcelKoch (approved):

PR #1857, opened 2025-06-07
Update tpls for v1.10

MarcelKoch:
These updates are due to deprecated CMake required versions in the TPLs.

### DIFF
--- a/extensions/test/config/CMakeLists.txt
+++ b/extensions/test/config/CMakeLists.txt
@@ -5,7 +5,7 @@ if(NOT yaml-cpp_FOUND)
     FetchContent_Declare(
         yaml-cpp
         GIT_REPOSITORY https://github.com/jbeder/yaml-cpp.git
-        GIT_TAG 0.8.0
+        GIT_TAG 2f86d13775d119edbb69af52e5f566fd65c6953b
     )
 
     # Turn off additional tool in yaml-cpp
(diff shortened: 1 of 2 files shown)


extensions/test/config/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ if(NOT yaml-cpp_FOUND)
     FetchContent_Declare(
         yaml-cpp
         GIT_REPOSITORY https://github.com/jbeder/yaml-cpp.git
-        GIT_TAG 0.8.0
+        GIT_TAG 2f86d13775d119edbb69af52e5f566fd65c6953b
```
upsj, line 8: Maybe add a comment describing the git tag this commit hash is based on?

upsj (approved):

extensions/test/config/CMakeLists.txt
MarcelKoch: There is no corresponding tag. It seems like yaml-cpp doesn't create releases (or tags) anymore.

yhmtsai (approved):
I guess it is from cmake 4+.
ref: https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html
> Changed in version 4.0: Compatibility with versions of CMake older than 3.5 is removed. Calls to [cmake_minimum_required(VERSION)](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#command:cmake_minimum_required) or [cmake_policy(VERSION)](https://cmake.org/cmake/help/latest/command/cmake_policy.html#version) that do not specify at least 3.5 as their policy version (optionally via ...<max>) will produce an error in CMake 4.0 and above.

